PR #426, opened 2017-03-02
Simplified Federation Control Plane Deployment.

madhusudancs:
This was copied from https://github.com/kubernetes/kubernetes/pull/34484 which is already LGTM'ed.


madhusudancs:
@matchstick this was already LGTM'ed in  kubernetes/kubernetes#34484. Could you please merge this?

PR #428, opened 2017-03-03
Fix example text for cgroup name

derekwaynecarr:
fix from a prior nit


derekwaynecarr:
per request from @vishh

vishh:
LGTM

PR #430, opened 2017-03-06
fix wrong cmd

tanshanshan:
when we run integration test,  the script will find_integration_test_dirs first. 
ref  [test-integration.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/make-rules/test-integration.sh#L41-L48)


contributors/devel/testing.md
```diff
@@ -197,8 +197,7 @@ for those internal etcd instances with the `TEST_ETCD_DIR` environment variable.
 
 The integration tests are run using `make test-integration`.
 The Kubernetes integration tests are writting using the normal golang testing
```
grodrigues3: can you fix the "writting" typo as well in this PR.

PR #431, opened 2017-03-06
List of the SIG-OpenStack members deleted

idvoretskyi:
It doesn't reflect the actual state of the membership and it's difficult to continuously maintain it to make in in a sync state.

The actual state can be found at https://github.com/orgs/kubernetes/teams/sig-openstack-misc.


xsgordon:
LGTM

PR #436, opened 2017-03-07
Propose allowing explicit elb subnet selection

calebamiles:
### User Story

Customer has many subnets spanning multiple AZs in which they'd like to explicitly define which two subnets a given Load balancer Service's ELB should be created. For example, one load balancer service can create an ELB attached to subnets A and B while another load balancer service can create an ELB attached to subnets C and D. Currently, Kubernetes supports subnet tagging allowing users to ensure ELBs are attached to specific subnets but this need comes out of a desire for greater subnet granularity in subnet selection on a per-service basis.

### DIFF
--- a/contributors/design-proposals/aws_under_the_hood.md
+++ b/contributors/design-proposals/aws_under_the_hood.md
@@ -158,6 +158,18 @@ ELB at the other end of its connection) when forwarding requests.
 TCP and SSL will select layer 4 proxying: the ELB will forward traffic without
 modifying the headers.
 
+#### Explicit Subnet Selection
+
+To explicitly select which subnets should be associated with an ELB an AWS 
+specific annotation can be added to a service
+
+```
+aws.kubernetes.io/elb-subnets="subnet-012345a6,subnet-912345b6"
+```
+
+which will cause the created ELB to have _only_ the specified subnets
+attached.
+
 ### Identity and Access Management (IAM)
 
 kube-proxy sets up two IAM roles, one for the master called


calebamiles:
cc: @philips, @joshrosso

philips:
Yes, this is a customer need we have encountered twice now. Would be great to get this proposal merged in time for the work to be done in 1.7.

justinsb:
Can you explain why the user wants different subnets for the two ELBs?  I don't particularly mind having an annotation for this, but it seems odd :-)

philips:
@justinsb they have kube workers in different VPCs based on their "security level". And some of those VPCs span different subnets because of direct connect. Thus, the need.

justinsb:
It sounds like you are saying they are running a single cluster with nodes in multiple VPCs ?  I don't think we support that (e.g. https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L2186 )

I'm not sure I understand the use-case sufficiently here to comment on whether we should support this.  But have you tested that this actually works if you do guide the subnet selection?

philips:
@justinsb we may need to fix that too.

joshrosso:
@justinsb, The current use case that derives this need is a single VPC deployment. This cluster has many subnets, used by multiple ASGs. Nodes in each ASG are dedicated, using taints, to certain uses with specific security requirements.

Currently in the AWS integration, to the best of my knowledge, we support tagging subnets to ensure Loadbalancers land in what we've tagged. Due to security requirements, we need more granularity to ensure a type LoadBalancer for a specific service lands in precisely the subnets that we deem the Public-In (Inbound) tier.

philips:
@joshrosso could you walk through a concrete example and then we can copy/paste that into the proposal.

calebamiles:
Thanks for the elaboration @joshrosso and please let me know if I mangle the use case. A user may have a cluster with nodes spread out over multiple subnets that are considered various "security domains" (e.g. some may have an internet gateway associated with them while others may not) and it is desired to be able to explicitly select which subnets an ELB will be attached to rather than _all_ subnets where nodes are running. One could imagine auditing all ELBs to ensure an ELB is never attached to a "protected" subnet that would cause problems with the way the AWS CloudProvider works today.

I can think of another relatively common scenario where the operations team responsible for creating AWS infrastructure (e.g. VPCs and the subnets within) is _not_ the team responsible for deploying into the provisioned AWS resources and it would simply be easier for the operations team deploying Kubernetes to specify ARNs rather than filing a ticket to have the operations team tag the AWS resources directly. I've worked with an unfortunate number of organizations with this structure (often using a single VPC for their entire organization). Are there other issues you'd like addressed, @justinsb?

**cc: @philips**

PR #437, opened 2017-03-07
[Federation] Federated statefulsets design proposal

irfanurrehman:
Design proposal for federated statefulsets.

@kubernetes/sig-federation-misc @kubernetes/sig-federation-proposals @kubernetes/sig-federation-pr-reviews

cc @quinton-hoole @deepak-vij @shashidharatd @dhilipkumars

Please feel free to cc any probable reviewers, not part of the targeted groups.

### DIFF
--- a/contributors/design-proposals/federated-statefulsets.md
+++ b/contributors/design-proposals/federated-statefulsets.md
@@ -3,53 +3,70 @@ irfan.rehman@huawei.com, dhilip.kumar.s@huawei.com
 
 # Goal
 
-To outline the use cases for the need of federated Statefulset and determine suitable design/implementation approach for the same.
+To outline the use cases for the need of federated Statefulset and determine 
+suitable design/implementation approach for the same.
 
 # Background
 
-The original design proposal of the statefulset in kubernetes, including the use cases and example applications, can be found [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/stateful-apps.md).
+The original design proposal of the statefulset in kubernetes, including the 
+use cases and example applications, can be found [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/stateful-apps.md).
 
 
 # Use Cases
 
-1 – A stateful app, for the reasons of high availability, wants the stateful pods distributed in different clusters, such that the set can withstand cluster failures. This represents an app with one single global quorum.
+1 – A stateful app, for the reasons of high availability, wants the stateful 
+pods distributed in different clusters, such that the set can withstand cluster failures. 
+This represents an app with one single global quorum.
 
-2 – A stateful app, wants replicas distributed in multiple clusters, such that it can form multiple smaller sized local clusters using only local replicas.
+2 – A stateful app, wants replicas distributed in multiple clusters, such that it can 
+form multiple smaller sized local clusters using only local replicas.
 
 # Design Requirements
 
 If we consider the use cases listed above, the main design requirements can roughly be listed as:
 
-1 – An unique, consistent and discoverable identity for each replica/instance across the federated clusters.
+1 – An unique, consistent and discoverable identity for each replica/instance across the 
+federated clusters.
 
-2 – It must be possible for the federated statefulset to first safely form an initial quorum (probably using cluster local replicas and identities).
-It should further also be possible to add members (probably from other clusters and global identities) to this initial quorum.
+2 – It must be possible for the federated statefulset to first safely form an initial quorum 
+(probably using cluster local replicas and identities).
+It should further also be possible to add members (probably from other clusters and global 
+identities) to this initial quorum.
 
 3 – The ability to scale, across clusters in some deterministic fashion.
 
-4 – An extended design requirement (which might not be a necessary requirement for this phase for work) can be an ability to migrate a pod from one node (in one cluster) to another node (possibly in another cluster), with stable network identity along with its attached volumes.
+4 – An extended design requirement (which might not be a necessary requirement for this 
+phase for work) can be an ability to migrate a pod from one node (in one cluster) to 
+another node (possibly in another cluster), with stable network identity along with its attached volumes.
 
 # Design Proposals
 
 The Statefulset object and APIs are already defined in k8s.
-The same needs to be exposed from federation to ensure that the interfaces remain consistent with a stand-alone k8s cluster.
+The same needs to be exposed from federation to ensure that the interfaces remain consistent 
+with a stand-alone k8s cluster.
 A separate controller at federation would however be useful.
 
-We considered two alternative designs for this proposal, both of which are further described in this document.
-We find the Design alternative 2, better in many aspects, and find it more suitable to be pursued for implementation further.
+We considered two alternative designs for this proposal, both of which are further described 
+in this document. We find the Design alternative 2, better in many aspects, and find it more 
+suitable to be pursued for implementation further.
 
 ## Design Alternative 1
 
-The federated replicset  controller should be able to spin off and monitor individual pods into federated clusters directly in a similar way to how an in-cluster controller would spin off and monitor the pods in a stand-alone cluster.
-The federation controller should be able to calculate and assign ordinal identities to these pods.
-The federation controller should also enforce the creation sequence, the same way an in-cluster controller would.
+The federated replicset  controller should be able to spin off and monitor individual pods 
+into federated clusters directly in a similar way to how an in-cluster controller would spin
+off and monitor the pods in a stand-alone cluster.
+The federation controller should be able to calculate and assign ordinal identities to these 
+pods. The federation controller should also enforce the creation sequence, the same way 
+an in-cluster controller would.
 
 ### Replica distribution (across federated clusters)
 
-The federated controller will interact with each cluster and create and monitor individual pod objects in the clusters.
-The pod&#39;s spec and state will be stored in individual cluster&#39;s etcd, as it would happen on creating a pod in the given cluster.
+The federated controller will interact with each cluster and create and monitor individual 
+pod objects in the clusters. The pod&#39;s spec and state will be stored in individual cluster&#39;s 
+etcd, as it would happen on creating a pod in the given cluster.
 The controller also assigns each ordinal replica to a specific cluster.
-As the primary intention of this feature is to provide a solution for high availability, it makes great sense to ensure that the instances are distributed into maximum clusters possible.
+As the primary intention of this feature is to provide a solution for high availability, 
+it makes great sense to ensure that the instances are distributed into maximum clusters possible.
 The initial proposal is to simply hash the replica number into the number of healthy clusters.
 A simple modulo calculation can be used to do the distribution.
 
@@ -59,65 +76,108 @@ In the case of in-cluster statefulset, pods discover each other using the in-clu
 A headless service with selectors enables creation of dns records against the pod names.
 This cannot work across clusters, as local pod IPs cannot be reached across the clusters.
 
-The proposal to make it work across the clusters is to assign a service type &#39;LoadBalancer&#39; for each pod instance that is created in the clusters.
+The proposal to make it work across the clusters is to assign a service type &#39;LoadBalancer&#39; 
+for each pod instance that is created in the clusters.
 The service name will be based on the ordinal pod name assigned to each pod.
-Additionally a set of label selectors will be added to each individual pod to ensure that the cluster routes the specific service traffic to the specific pod.
+Additionally a set of label selectors will be added to each individual pod to ensure that the
+cluster routes the specific service traffic to the specific pod.
 The service type &#39;LoadBalancer&#39; will ensure that the data traffic can be routed across clusters.
 The network dns names for each pod will remain very similar to the network names used by the in-cluster statefulsets.
 
 As an example:
 
-For a federation with 3 clusters all in the same DNS zone _federation.example_ and the statefulset name _web_ the dns names for instances might look like:
+For a federation with 3 clusters all in the same DNS zone _federation.example_ and the statefulset 
+name _web_ the dns names for instances might look like:
 ```
 web-0.mynamespace.myfederation.svc.federation.example
 web-1.mynamespace.myfederation.svc.federation.example
 Web-2.mynamespace.myfederation.svc.federation.example
 ```
 
-The difference compared to an in-cluster dns name for an instance of a statefulset can be noted where the dns id starts with **$(podname).$(governing service domain)**, whereas when federated, in this approach, it will start only with **$(service name).**
-This will be achieved by federated controller creating and monitoring individual services in specific clusters (not a federated service) against each pod, with the same name as that pod&#39;s ordinal identity.
-It will further need to be ensured that the pods be assigned the same hostname as the pod name (ordinal identity) and service name to keep it consistent with the in-cluster statefulsets topology.
+The difference compared to an in-cluster dns name for an instance of a statefulset can
+be noted where the dns id starts with **$(podname).$(governing service domain)**, whereas when 
+federated, in this approach, it will start only with **$(service name).**
+This will be achieved by federated controller creating and monitoring individual services in 
+specific clusters (not a federated service) against each pod, with the same name as that pod&#39;s 
+ordinal identity. It will further need to be ensured that the pods be assigned the same hostname 
+as the pod name (ordinal identity) and service name to keep it consistent with the in-cluster 
+statefulsets topology.
 
 There is one interesting catch about the service creation, here.
-The service needs to exist locally in the cluster (1 service per replica pod across clusters), which is unlike a federated service.
-If we use the existing federated service mechanism, to create this service, and the service with a particular name (for example web-0 in example above) is created in federation, it will propagate the same to all federated clusters.
-This is unnecessary for the solution mentioned in this section (or for further reading in the following section), and would be extreme waste of resources.
-Some simple proposals to tackle this problem are mentioned at the end of this document under section &quot; **Handling additional LB Services**&quot;.
+The service needs to exist locally in the cluster (1 service per replica pod across clusters), 
+which is unlike a federated service. 
+If we use the existing federated service mechanism, to create this service, and the service with 
+a particular name (for example web-0 in example above) is created in federation, it will 
+propagate the same to all federated clusters. 
+This is unnecessary for the solution mentioned in this section (or for further reading in 
+the following section), and would be extreme waste of resources. 
+Some simple proposals to tackle this problem are mentioned at the end of this document under 
+section &quot; **Handling additional LB Services**&quot;.
 
 ## Design Alternative 2
 
-In this approach, the federated statefulset controller will behave quite similar to federated replicaset or the federated deployment controller.
-The federated controller would create and monitor individual statefulsets (rather then pods directly) partitioning and distributing the total stateful replicas across the federated clusters.
+In this approach, the federated statefulset controller will behave quite similar to federated 
+replicaset or the federated deployment controller.
+The federated controller would create and monitor individual statefulsets (rather then pods 
+directly) partitioning and distributing the total stateful replicas across the federated clusters.
 
 As a proposal in this design we suggest the possibility of the pods having multiple identities.
-This facilitates application pods being able to participate in different quora, and leaves a choice with application replicas to join either the local or the global quora, as need be.
-As of now we can envision the need of two possible identities for each pod (relevant to use case 2), enabled by this approach.
-The ordinal number assigned to the pod will be calculated and assigned in exactly the same way, it is assigned in the k8s cluster as described [here](https://kubernetes.io/docs/concepts/abstractions/controllers/statefulsets/#pod-identity) and will be unique.
+This facilitates application pods being able to participate in different quora, and leaves a 
+choice with application replicas to join either the local or the global quora, as need be.
+As of now we can envision the need of two possible identities for each pod (relevant to use case 2), 
+enabled by this approach.
+The ordinal number assigned to the pod will be calculated and assigned in exactly the same way, 
+it is assigned in the k8s cluster as described 
+[here](https://kubernetes.io/docs/concepts/abstractions/controllers/statefulsets/#pod-identity) and will be unique.
 It however will be possible for each pod to take multiply dns identities.
 This is detailed in the &#39;instance identity and discovery&#39; section below.
 
-For this phase of implementation, it is proposed that the statefulset need not guarantee the order of creation of the pod instances across federation (_we believe more discussion will follow this statement_).
+For this phase of implementation, it is proposed that the statefulset need not guarantee the 
+order of creation of the pod instances across federation (_we believe more discussion will follow this statement_).
 
 ### Replica distribution (across federated clusters)
 
-The proposed default behaviour of the federation controller, when a statefulset creation request is sent to federation API server, would be to partition the statefulset replicas and create a statefulset into each of the clusters with the reduced replica number (after partitioning), quite similar to the behaviour of the replicaset or daemonset controllers of the k8s federation.
-If the replica count assigned to the federated statefulset is equal to or smaller than the total healthy clusters in the federation, then there will be at most one replica assigned to the statefulset created in each underlying cluster (and some clusters might not get a statefulset object at all in case of replica number is smaller than total healthy clusters).
-For the case of replica numbers being larger than the total number of clusters, the proposed default behaviour will be, to partition the replica count equally into healthy clusters.
-Annotations could be provided to override this behaviour in exactly the same fashion annotations are provided in the spec of federated replicasets ([_frs preferences_](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/federated-replicasets.md#federatereplicaset-preferences)).
-The applications which are being deployed as the statefulset can choose to connect to the other replica instances locally within the cluster or globally across the clusters to form the quorum using the respective dns names.
-Applications (in case they support such a behaviour) can also communicate across quora (for example a local db cluster connects to another cluster/instance outside the cluster for periodic backups) using the global dns identities.
+The proposed default behaviour of the federation controller, when a statefulset creation request is 
+sent to federation API server, would be to partition the statefulset replicas and create a statefulset 
+into each of the clusters with the reduced replica number (after partitioning), quite similar to the 
+behaviour of the replicaset or daemonset controllers of the k8s federation.
+If the replica count assigned to the federated statefulset is equal to or smaller than the total 
+healthy clusters in the federation, then there will be at most one replica assigned to the statefulset 
+created in each underlying cluster (and some clusters might not get a statefulset object at all 
+in case of replica number is smaller than total healthy clusters).
+For the case of replica numbers being larger than the total number of clusters, the proposed default 
+behaviour will be, to partition the replica count equally into healthy clusters.
+Annotations could be provided to override this behaviour in exactly the same fashion annotations 
+are provided in the spec of federated replicasets ([_frs preferences_](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/federated-replicasets.md#federatereplicaset-preferences)).
+The applications which are being deployed as the statefulset can choose to connect to the other 
+replica instances locally within the cluster or globally across the clusters to form the quorum 
+using the respective dns names.
+Applications (in case they support such a behaviour) can also communicate across quora (for example 
+a local db cluster connects to another cluster/instance outside the cluster for periodic backups) 
+using the global dns identities.
 
 ### Instance identity and discovery
 
-Together with the statefulset object, the user will also need to submit the request for a headless service corresponding to the statefulset.
-When such requests (one for statefulset and another for the headless service) are sent to the federation control plane, the federation service controller will create a matching headless service in each of the clusters.
-It will further partition the total number of replicas and create statefulsets with partitioned replica numbers into at least 1 or more clusters.
-The noteworthy point is the proposal that federated stateful controller would additionally modify the statefulset name by appending the cluster name to the statefulset name into whichever cluster the partitioned statefulset is getting created.
-This will ensure that each pod in the federated statefulset maintains an unique identity across all clusters, including a stable non-changing hostname for each pod even with the ordinal numbers being generated local to the clusters.
+Together with the statefulset object, the user will also need to submit the request for a headless 
+service corresponding to the statefulset.
+When such requests (one for statefulset and another for the headless service) are sent to the 
+federation control plane, the federation service controller will create a matching headless service 
+in each of the clusters.
+It will further partition the total number of replicas and create statefulsets with partitioned 
+replica numbers into at least 1 or more clusters.
+The noteworthy point is the proposal that federated stateful controller would additionally modify 
+the statefulset name by appending the cluster name to the statefulset name into whichever cluster 
+the partitioned statefulset is getting created.
+This will ensure that each pod in the federated statefulset maintains an unique identity across 
+all clusters, including a stable non-changing hostname for each pod even with the ordinal numbers 
+being generated local to the clusters.
 
-In case of clusters getting statefulsets with more than 1 replica, pods will be able to discover each other using the in-cluster dns identity provided by the headless in-cluster service (as described [here](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#using-stable-network-identities)).
+In case of clusters getting statefulsets with more than 1 replica, pods will be able to discover 
+each other using the in-cluster dns identity provided by the headless in-cluster service 
+(as described [here](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#using-stable-network-identities)).
 
-For an example, let&#39;s say we have 3 federated clusters and the following yamls are submitted to create a federated statefulset:
+For an example, let&#39;s say we have 3 federated clusters and the following yamls are submitted 
+to create a federated statefulset:
 ```
 ---
 apiVersion: v1
@@ -169,9 +229,12 @@ The service will be created in each of the 3 clusters unmodified.
 
 The default replica partitioning will mean a distribution of **2, 2 and 1** replicas to the 3 clusters.
 
-The statefulset spec submitted to the clusters will have the statefulset metadata.name modified to _**web+&lt;cluster-name&gt;**_.
+The statefulset spec submitted to the clusters will have the statefulset metadata.name modified 
+to _**web+&lt;cluster-name&gt;**_.
 
-Using the example above, with statefulset name _web_, service name _nginx_ and domain name as _federation.example_ the dns names created against each instance would be as below, discoverable both locally and globally (assuming the registered cluster names being c1, c2 and c3):
+Using the example above, with statefulset name _web_, service name _nginx_ and domain name as
+_federation.example_ the dns names created against each instance would be as below, discoverable 
+both locally and globally (assuming the registered cluster names being c1, c2 and c3):
 ```
 web-c1-0.nginx.mynamespace.svc.federation.example
 web-c1-1.nginx.mynamespace.svc.federation.example
@@ -179,10 +242,15 @@ web-c2-0.nginx.mynamespace.svc.federation.example
 web-c2-1.nginx.mynamespace.svc.federation.example
 web-c3-0.nginx.mynamespace.svc.federation.example
 ```
-This above is fine for the case of multiple local quora (one in each cluster), but these identities cannot work across the clusters, as the local pod IPs cannot be reached across the clusters.
+This above is fine for the case of multiple local quora (one in each cluster), but these identities 
+cannot work across the clusters, as the local pod IPs cannot be reached across the clusters.
 
-Similar to _design alternative 1_, the proposal to make it work across the clusters is to create an additional service of type &#39;LoadBalancer&#39; for each pod instance that is created in the clusters. This will be automatically created by the federated statefulset controller, on demand, proposed to be controlled by an annotation on the statefulset.
-If the annotation specifies so (users choice), the federation controller will create the additional service of type &#39;LoadBalancer&#39; against each pod replica.
+Similar to _design alternative 1_, the proposal to make it work across the clusters is to create an 
+additional service of type &#39;LoadBalancer&#39; for each pod instance that is created in the clusters. 
+This will be automatically created by the federated statefulset controller, on demand, proposed to be 
+controlled by an annotation on the statefulset.
+If the annotation specifies so (users choice), the federation controller will create the additional 
+service of type &#39;LoadBalancer&#39; against each pod replica.
 The service name for service, one each against the pod will be the same name as the pod identity.
 The LB service creation at federation has the same drawback as discussed in _design alternative 1_.
 The proposed solution is elaborated in section **Handling additional LB Services**.
@@ -197,32 +265,46 @@ There are two ways a federated statefulset can be assigned a persistent storage.
 
 Either of the 2 options above can be used/specified by the user to attach a persistent store to the pods.
 
-It is known, that as of now, the same persistent volume, even if can be used in a different cluster, k8s directly does not provide an API, yet, which can aid the same. In the absence of no direct way of quick migration of the storage data from one zone to another, or from one cloud provider to another if the case be, the proposal is to disallow migration of pods across clusters.
+It is known, that as of now, the same persistent volume, even if can be used in a different cluster, 
+k8s directly does not provide an API, yet, which can aid the same. In the absence of no direct way of 
+quick migration of the storage data from one zone to another, or from one cloud provider to another if the 
+case be, the proposal is to disallow migration of pods across clusters.
 
 ## Scale up/Scale down
 
-When a federated statefulset is scaled up or down, the distribution of replicas will be recalculated and the replica numbers in each cluster will be updated independently.
+When a federated statefulset is scaled up or down, the distribution of replicas will be recalculated and 
+the replica numbers in each cluster will be updated independently.
 
 The behaviour of the scaling on being targeted using a **federated hpa** will be discussed in a separate design.
 
 ## Handling additional LB services
 
-As listed in the approaches, the current option of exposing the statefulset pods across clusters is to assign a LB service to each replica pod.
-If the federated service controller is used (by creating the service in federation) to create these services, then as per the current behaviour the service will be created in each cluster. 
+As listed in the approaches, the current option of exposing the statefulset pods across clusters is to 
+assign a LB service to each replica pod.
+If the federated service controller is used (by creating the service in federation) to create these services, 
+then as per the current behaviour the service will be created in each cluster. 
 This leads to exactly one unused service per statefulset per cluster.
 There are two alternatives to circumvent this problem:
 
-1 - Cluster affinity is introduced at least for services such that the controller creates the service only in needed cluster. (The need of cluster affinity and anti-affinity in general is discussed [elsewhere](https://github.com/kubernetes/kubernetes/issues/41442) also)
+1 - Cluster affinity is introduced at least for services such that the controller creates the service only 
+in needed cluster. (The need of cluster affinity and anti-affinity in general is discussed 
+[elsewhere](https://github.com/kubernetes/kubernetes/issues/41442) also)
 
-2 - Federated statefulset controller handles the creation of cluster local services, rather than the federated service controller. It also will need to handle the dns records for each LB service into the federation (or cloud provider specific) public dns server.
+2 - Federated statefulset controller handles the creation of cluster local services, rather than the 
+federated service controller. It also will need to handle the dns records for each LB service into the 
+federation (or cloud provider specific) public dns server.
 
-We propose using alternative 1 listed here, as it fits broader scheme of things and is more consistent with the user expectation of being able to query all needed resources from the federation control plane and is less confusing to use at the same time.
+We propose using alternative 1 listed here, as it fits broader scheme of things and is more consistent with 
+the user expectation of being able to query all needed resources from the federation control plane and is 
+less confusing to use at the same time.
 
 # Conclusion
 
-Alternative 1 can solve only 1 of the listed use cases (scenario 1), and does not remain consistent with the existing federated controllers.
+Option 1 above can solve only 1 of the listed use cases (scenario 1), and does not remain consistent with 
+the existing federated controllers.
 
-Alternative 2 can solve both the use cases, is consistent with the existing federated controllers and makes more sense in the scheme of things.
+Option 2 above can solve both the use cases, is consistent with the existing federated controllers 
+and makes more sense in the scheme of things.
 
 # Limitations (to be discussed)
 
@@ -242,7 +324,9 @@ I think we&#39;re left with a few options:
 
 (2) disallow migration of replicas between clusters if they use non-node-local volumes (simple, but perhaps overly restrictive)
 
-(3) allow migration of replicas between clusters provided that either the replica uses node-local-volumes, or the clusters are in the same zone (less restrictive, but more complex to implement, and for the user to understand). I suspect that the right answer here might be to start with (2) and add 3 in a later phase.
+(3) allow migration of replicas between clusters provided that either the replica uses node-local-volumes, or the 
+clusters are in the same zone (less restrictive, but more complex to implement, and for the user to understand). 
+I suspect that the right answer here might be to start with (2) and add 3 in a later phase.
 
 _3 – What happens if a cluster dies_
 
--- a/contributors/design-proposals/federated-statefulsets.md
+++ b/contributors/design-proposals/federated-statefulsets.md
@@ -1,249 +1,200 @@
 # Federated StatefulSet design
-irfan.rehman@huawei.com, dhilip.kumar.s@huawei.com
+irfan.rehman@huawei.com
 
 # Goal
 
-To outline the use cases for the need of federated Statefulset and determine 
+To outline the use cases for the need of federated Statefulset and determine
 suitable design/implementation approach for the same.
 
 # Background
 
-The original design proposal of the statefulset in kubernetes, including the 
+The original design proposal of the statefulset in kubernetes, including the
 use cases and example applications, can be found [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/stateful-apps.md).
 
 # Use Cases
 
-1 – A stateful app, wants replicas distributed in multiple clusters, such that it can 
-form multiple smaller sized local app clusters using only local replicas, each serving 
+1 – A stateful app, wants replicas distributed in multiple clusters, such that it can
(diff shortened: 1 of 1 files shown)


liggitt:
cc @kubernetes/sig-federation-misc @smarterclayton

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,213 @@
...
+2 – A stateful app, wants replicas distributed in multiple clusters, such that it can form multiple smaller sized local clusters using only local replicas.
+
+
+# Design Requirements
+
+If we consider the use cases listed above, the main design requirements can roughly be listed as:
+
+1 – An unique, consistent and discoverable identity for each replica/instance across the federated clusters.
+
+2 – Predictability on the number of instances and sequentialization of pod creation (for example instance 1 creation starts only when instance 0 is up and running).
```
smarterclayton: We've discussed relaxing this.

smarterclayton:
Please break your lines at some fixed interval, the proposal is unreviewable asnis

irfanurrehman:
> Please break your lines at some fixed interval, the proposal is unreviewable asnis

Thanks @smarterclayton for having a look. I have added line break on all the lines. PTAL.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: Yes, I have specified (later in the doc) that we might not be able to guarantee this requirement for the federated statefulsets.

dhilipkumars:
CC : @erictune @kargakis @kow3ns

madhusudancs:
Some high level questions:

1. What happens when a new cluster joins or a cluster leaves federation? Do you rebalance?
2. Say I have 2 clusters A and B, and `replicas: 5`. Cluster A gets 3 and B gets 2 replicas. Cluster A dies. According to your limitation _#3_, there will be only 2 replicas and there is no quorum now. This means the stateful set cannot function even though there is enough capacity in the other cluster. Is that correct?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,248 @@
...
+
+1 – A stateful app, for the reasons of high availability, wants the stateful pods distributed in different clusters, such that the set can withstand cluster failures. This represents an app with one single global quorum.
+
+2 – A stateful app, wants replicas distributed in multiple clusters, such that it can form multiple smaller sized local clusters using only local replicas.
+
+# Design Requirements
+
+If we consider the use cases listed above, the main design requirements can roughly be listed as:
+
+1 – An unique, consistent and discoverable identity for each replica/instance across the federated clusters.
```
smarterclayton: I'd like a new design requirement.

2. It must be possible for the federated sets to SAFELY form an initial quorum that adds the rest of the set.  

Any design which doesn't allow (for instance) cluster 1 to form an initial quorum and then safely add members in cluster 2 (or any other order, really), is a non-starter because then stateful set at federation has a different set of guarantees.

I think that's implicit in some of your comments, but it needs to be explicit, obvious, and impossible to break.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,248 @@
...
+
+The original design proposal of the statefulset in kubernetes, including the use cases and example applications, can be found [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/stateful-apps.md).
+
+
+# Use Cases
+
+1 – A stateful app, for the reasons of high availability, wants the stateful pods distributed in different clusters, such that the set can withstand cluster failures. This represents an app with one single global quorum.
+
+2 – A stateful app, wants replicas distributed in multiple clusters, such that it can form multiple smaller sized local clusters using only local replicas.
+
```
chrislovecnm: What about two stateful sets that need to comminicate with each other. Many applications such as Cassandra, Kafka and elasticsearch include the capability to span physical data centers?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,248 @@
...
+Similar to _design alternative 1_, the proposal to make it work across the clusters is to create an additional service of type &#39;LoadBalancer&#39; for each pod instance that is created in the clusters. This will be automatically created by the federated statefulset controller, on demand, proposed to be controlled by an annotation on the statefulset.
+If the annotation specifies so (users choice), the federation controller will create the additional service of type &#39;LoadBalancer&#39; against each pod replica.
+The service name for service, one each against the pod will be the same name as the pod identity.
+The LB service creation at federation has the same drawback as discussed in _design alternative 1_.
+The proposed solution is elaborated in section **Handling additional LB Services**.
+
+## Storage volumes
+
+There are two ways a federated statefulset can be assigned a persistent storage.
```
chrislovecnm: How are we handling storage classes?  Am I my missing that section?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,248 @@
...
+
+_1 – In in case of a particular replica or the cluster containing particular replicas die, what should be the behavior._
+
+It can be resurrected with the same identity in the original cluster only
+
+_2 – Is there a migration scenario? If there is, in case a replica needs migration, how do we handle this?_
+
+elaborated by @quinton-hoole
+
+Strictly speaking, migration between clusters in the same zone is quite feasible.
```
chrislovecnm: Please add more clarity to this statement.

chrislovecnm: More content as well.

contributors/design-proposals/federated-statefulsets.md
chrislovecnm: When you mention quorum what quorum are you referring to?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,248 @@
...
+
+# Design Requirements
+
+If we consider the use cases listed above, the main design requirements can roughly be listed as:
+
+1 – An unique, consistent and discoverable identity for each replica/instance across the federated clusters.
+
+2 – Predictability on the number of instances and sequentialization of pod creation (for example instance 1 creation starts only when instance 0 is up and running).
+
+3 – The ability to scale, across clusters in some deterministic fashion.
```
chrislovecnm: Can this be optional? Some technologies do not allow more than one pod at a time.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,248 @@
...
+
+(1) only allow the use of node-local persistent (bad),
+
+(2) disallow migration of replicas between clusters if they use non-node-local volumes (simple, but perhaps overly restrictive)
+
+(3) allow migration of replicas between clusters provided that either the replica uses node-local-volumes, or the clusters are in the same zone (less restrictive, but more complex to implement, and for the user to understand). I suspect that the right answer here might be to start with (2) and add 3 in a later phase.
+
+_3 – What happens if a cluster dies_
+
+Nothing, statefulset would need to run with lesser replicas
```
chrislovecnm: One section that I am not seeing is networking. Often ha distributed applications use patterns such as gossip. That can require every pod to talk to every other pod. Thoughts around how we are going to handle those patterns?

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: The scenario you mention is infact solved by design alternative 2 specified below, where a particular statefulset instance gets a local and global identity both. Applications can choose to use either identities or both at the same time. The use cases I mentioned are for reference alone and not exhaustive. Do you think its important to mention another use case specifying the scenario u mention?

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: @smartclayton I have mentioned the requirement explicitly as you suggested. I have also dropped the sequentialization requirement, because we are not honoring that as of now.
@chrislovecnm I think @smarterclayton is referring to the quorum application pod instances form after discovering each other.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: I have now dropped this requirement altogether from this section.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: The scaling feature will be  on demand, but I think we need to have the ability in the federated statefulsets.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: I did we need not mention anything related to the storage classes. The chosen design proposal intends to reuse the cluster local statefulsets implementation, when the statefulsets are deployed in individual clusters.
One catch might be that the different clusters might not have the same storage classes available. I think for now, we leave it to the user to ensure that the specified storage class is available across all federated clusters that he/she mentions in the federated statefulset spec. The behavior if this does not happen is also deterministic for now. The volume provisioning for the stateful pods in a particular cluster which does not have that class will fail.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: We handle this using the dns names similar to current k8s statefulset implementation. The "Instance identity and discovery" sections specify the details.
Apps need to discover and use the dns names to communicate across instances and not the IPs.

irfanurrehman:
> What happens when a new cluster joins or a cluster leaves federation? Do you rebalance?

In the current proposal, when a new cluster joins and the federation already has a running statefulset, nothing will happen (meaning no rebalance). However if the statefulset is scaled after the cluster joins it might get replica(s).

irfanurrehman:
> Some high level questions:
> 
> What happens when a new cluster joins or a cluster leaves federation? Do you rebalance?

In the current proposal, when a new cluster joins and the federation already has a running statefulset, nothing will happen (meaning no rebalance). However if the statefulset is scaled after the cluster joins it might get replica(s).

> Say I have 2 clusters A and B, and replicas: 5. Cluster A gets 3 and B gets 2 replicas. Cluster A dies. According to your limitation #3, there will be only 2 replicas and no quorum now. This means the stateful set cannot function even though there is enough capacity in the other cluster. Is that correct?

What you mention is a valid scenario, but this is probably something we need to live with as of now. I believe the federated statefulsets will still be pretty useful even with this constraint.
Also this constraint (no migration) is mainly because there is no direct or automated way of migrating the persistent storage as is across k8s clusters.  Any suggestions relating this are certainly welcome.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: Sure, Will update with something more explanatory.

irfanurrehman:
Thanks @smarterclayton @chrislovecnm @madhusudancs for the comments!
Some updates PTAL!

contributors/design-proposals/federated-statefulsets.md
chrislovecnm: Should we say that we are going to test against them?

contributors/design-proposals/federated-statefulsets.md
chrislovecnm: But how will the routing work?  Does CNI provide the capability to route between a headless service in the uk and a headless service in the US?  Kinda confused a bit.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,249 @@
...
+The controller also assigns each ordinal replica to a specific cluster.
+As the primary intention of this feature is to provide a solution for high availability, it makes great sense to ensure that the instances are distributed into maximum clusters possible.
+The initial proposal is to simply hash the replica number into the number of healthy clusters.
+A simple modulo calculation can be used to do the distribution.
+
+### Instance identity and discovery
+
+In the case of in-cluster statefulset, pods discover each other using the in-cluster dns names.
+A headless service with selectors enables creation of dns records against the pod names.
+This cannot work across clusters, as local pod IPs cannot be reached across the clusters.
```
chrislovecnm: Is this true with all CNI providers or are we just saying that this is the expectation?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,249 @@
...
+
+The proposal to make it work across the clusters is to assign a service type &#39;LoadBalancer&#39; for each pod instance that is created in the clusters.
```
chrislovecnm: Is there any way to do this differently?  Think about this at scale.  Cost, quotas and provisioning would get seriously fun :)  How are we doing guaranteed network identity for each ELB?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,249 @@
...
+If we use the existing federated service mechanism, to create this service, and the service with a particular name (for example web-0 in example above) is created in federation, it will propagate the same to all federated clusters.
+This is unnecessary for the solution mentioned in this section (or for further reading in the following section), and would be extreme waste of resources.
+Some simple proposals to tackle this problem are mentioned at the end of this document under section &quot; **Handling additional LB Services**&quot;.
+
+## Design Alternative 2
+
+In this approach, the federated statefulset controller will behave quite similar to federated replicaset or the federated deployment controller.
+The federated controller would create and monitor individual statefulsets (rather then pods directly) partitioning and distributing the total stateful replicas across the federated clusters.
+
+As a proposal in this design we suggest the possibility of the pods having multiple identities.
```
chrislovecnm: Doesn't this break the contract that statefuls sets already have?  I probably need more details to fully understand this design case.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,249 @@
...
+This leads to exactly one unused service per statefulset per cluster.
+There are two alternatives to circumvent this problem:
+
+1 - Cluster affinity is introduced at least for services such that the controller creates the service only in needed cluster. (The need of cluster affinity and anti-affinity in general is discussed [elsewhere](https://github.com/kubernetes/kubernetes/issues/41442) also)
+
+2 - Federated statefulset controller handles the creation of cluster local services, rather than the federated service controller. It also will need to handle the dns records for each LB service into the federation (or cloud provider specific) public dns server.
+
+We propose using alternative 1 listed here, as it fits broader scheme of things and is more consistent with the user expectation of being able to query all needed resources from the federation control plane and is less confusing to use at the same time.
+
+# Conclusion
```
chrislovecnm: With some applications, you can only create one pod at a time.  How does this design proposal maintain the ordinal order as defined by the stateful set contract?

Does this proposal define the capability of migrating from a regular stateful set to a federated stateful set?

contributors/design-proposals/federated-statefulsets.md
chrislovecnm: Also one of the amazing advantages is to be able to have a baremetal stateful set talking to a cloud stateful set.  How are we going to allow for that?  Think about expanding your footprint during a busy shopping season.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,249 @@
...
+
+1 - A node local volume attached to the container
+
+2 - A persistent volume claim from the cloud provider.
+
+Either of the 2 options above can be used/specified by the user to attach a persistent store to the pods.
+
+It is known, that as of now, the same persistent volume, even if can be used in a different cluster, k8s directly does not provide an API, yet, which can aid the same. In the absence of no direct way of quick migration of the storage data from one zone to another, or from one cloud provider to another if the case be, the proposal is to disallow migration of pods across clusters.
+
+## Scale up/Scale down
```
mstump: For the database and search systems that I can think of this would be undesirable. k8s isn't workload aware, what it means to scale up, re-shard data and all the implications of those actions. Additionally, workloads across DCs are not always homogenous or even distributed, you want to be able to scale them independently.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,249 @@
...
+
+elaborated by @quinton-hoole
+
+Strictly speaking, migration between clusters in the same zone is quite feasible.
+
+I think we&#39;re left with a few options:
+
+(1) only allow the use of node-local persistent (bad),
+
+(2) disallow migration of replicas between clusters if they use non-node-local volumes (simple, but perhaps overly restrictive)
```
mstump: This is the preferred method. Don't attempt to move storage across clusters. Most distributed databases have their own notion of identity, consistency and replication. This is an app specific concern, don't overcomplicate k8s by trying to manage replication for them.

contributors/design-proposals/federated-statefulsets.md
mstump: I concur, this is a horrible idea. Imagine trying to manage a Cassandra cluster of hundreds, of nodes per DC. This isn't feasible. Do one headless service per k8s cluster, and then create a VPN link between virtual networks. Setup DNS resolution across clusters.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,249 @@
...
+As of now we can envision the need of two possible identities for each pod (relevant to use case 2), enabled by this approach.
+The ordinal number assigned to the pod will be calculated and assigned in exactly the same way, it is assigned in the k8s cluster as described [here](https://kubernetes.io/docs/concepts/abstractions/controllers/statefulsets/#pod-identity) and will be unique.
+It however will be possible for each pod to take multiply dns identities.
+This is detailed in the &#39;instance identity and discovery&#39; section below.
+
+For this phase of implementation, it is proposed that the statefulset need not guarantee the order of creation of the pod instances across federation (_we believe more discussion will follow this statement_).
+
+### Replica distribution (across federated clusters)
+
+The proposed default behaviour of the federation controller, when a statefulset creation request is sent to federation API server, would be to partition the statefulset replicas and create a statefulset into each of the clusters with the reduced replica number (after partitioning), quite similar to the behaviour of the replicaset or daemonset controllers of the k8s federation.
```
mstump: I don't understand why this is desirable or needed. Let each statefulset be managed independently, but create network links between the federated sets.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: As of now (and in my limited knowledge) the CNI overlays are limited within the cluster. However there probably isn't much of a technical bottleneck which stops the overlay to work across clusters. Perhaps that might be another proposal for cluster federation in the pipeline if the need arises. Right now this is true with all CNI providers (and again in my limited knowledge :) ).

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: @chrislovecnm you are completely right about the cost involved in provisioning ELB's. Right now the options (using k8s available features) to communicate across clusters are: 
- ELB's provisioned from the cloud provider
- Ingress (autoprovisioning is tricky, and is quite cloud specific)
- Nodeport

As of today, the federated services have out of the box support for ELB only and a proper working for ingress and nodeport for federated services is still evolving (somebody please correct me if I am wrong). This proposal in this design in fact is to adhere to the same facility/evolution of federated services for communication across clusters.
The point I am trying to make is that, there already are use cases which demand cross cluster/federated  statefulsets and would benefit from, even with the current proposal in place.
Better solutions can evolve over time.
@mstump thanks for your suggestion. Can u please elaborate a little bit more with the point of this functionality getting into some existing k8s/federation feature, or it being a new feature in itself. The point is the overlay u mention ideally need to be possible as an auto provisioned method of communication across clusters. If it makes sense in near term and is doable, I would not mind pursuing the same.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: No it doe not. In a normal in-cluster statefulset a stateful pod gets 2 identities:
- the dns name accessible within the cluster
- the hostname visible to the stateful app

The suggestion in this design retains both and then adds another dns name accessible across the cluster. I specify the additional dns name as another identity, thus multiple identities. I dont know of any such use case as of now, but if needed the same design can be extended to have even more dns names visible locally or globally reflecting more identities.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: If u see line 99 (and as suggested by @smarterclayton ) the proposal is to not have a hard requirement to preserve the order of stateful pod creation.
This document has not handled the migration of a regular in-cluster statefulset to a federated one. Do u think its needed to address it in this design.. ?

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: I did not get the question completely.
Are you saying that scale up/down of a statefulset is undesirable, or an overall scale up/down of  a federated statefulset whose pods are distributed across clusters undesirable?
In either case I think its wise to have a scale function atleast available to the users rather then not having this function at all.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: agree.. ! That is what this doc also proposes.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: The point is that the federation user needs to see a consolidated view of the statefulset. Also, the current k8s federation design philosophy is that the federation ideally appears as another k8s cluster, where some users might not even know that they are talking to federation and not a normal k8s cluster.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: This routing is not handled by CNI.
This routing will be handled very similar to the way it is handled for federated services.
(check one of my other reply to your comments about need of ELB).

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: I am sorry; I did not get you here.. ? You mean test against storage classes.. ?

irfanurrehman:
cc @kubernetes/sig-apps-misc @kubernetes/sig-apps-feature-requests

quinton-hoole:
@irfanurrehman I think that the design looks good overall.  It would be useful to add a few non-trivial, concrete examples to illustrate both useful applications, and some of the explicit limitations.  Off the top of my head, these examples might include:

1. A federated etcd statefulset of 3 replicas across >2 clusters (using maxReplicas=1, PVC's) - this creates a global quorum able to withstand the temporary failure of any one cluster, as well as the permanent failure of any node in any cluster. Replicas would move around in each cluster, but not be moved between clusters.  You already allude to this, but you could explain in some detail how the replicas would discover each other, and recover from failure.
2. A federated cassandra statefulset of 100 replicas across, say 5 clusters.  If they use local disks, replicas could move between clusters.
3. A federated etcd statefulset of 30 replicas across 10 clusters (using minreplicas=3, maxreplicas=3) - this creates a local quorum in each of the 10 clusters.  The potential global quorum is ignored in this example.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,333 @@
...
+### Instance identity and discovery
+
+Together with the statefulset object, the user will also need to submit the request for a headless 
+service corresponding to the statefulset.
+When such requests (one for statefulset and another for the headless service) are sent to the 
+federation control plane, the federation service controller will create a matching headless service 
+in each of the clusters.
+It will further partition the total number of replicas and create statefulsets with partitioned 
+replica numbers into at least 1 or more clusters.
+The noteworthy point is the proposal that federated stateful controller would additionally modify
```
soltysh: With this you might hit name limit problem. When any of the names is close to max and you concatenate them you might exceed allowed limit. It is an important problem that should be described in this proposal, imho.

soltysh:
https://github.com/kubernetes/community/pull/503 seems also very relevant for this proposal.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: Thanks for the specific suggestion, and apologies for the delayed response.
I have added as solution for this implementable as an admission controller, but given that its a difficult to hit problem, it might not be a necessary implementation in the first phase solution. Is that ok?

irfanurrehman:
> #503 seems also very relevant for this proposal.

yes, It is indeed relevant. My suggestion however for now is to treat the federated statefulset update design separate (or a later extension) from the federated statefulset feature. The same way its happening for this feature in local k8s.

irfanurrehman:
@marun, @madhusudancs, I have updated this on the lines of your comments, and tried to flush out as much detail as I could. PTAL!!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+specific clusters against each pod, with the same name as that pod&#39;s
+ordinal identity. It will further need to be ensured that the pods be assigned the same hostname
+as the pod name (ordinal identity) and service name to keep it consistent with the in-cluster
+statefulsets topology.
+
+The details about creation and handling of the LB services and updating the dns records will
+be similar to that elaborated in the previous design.
+
+This alternative has following drawbacks:
+ - As the control over the stateful pods is with the federation statefulset controller, the
```
marun: This limitation could be overcome if the federated statefulsets controller delegated responsibility for monitoring stateful pods to a federation-specific controller in each targeted member cluster.  I think this proposal needs to be fleshed out more before it is disqualified as not being viable.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+ monitoring of the stateful pods will not happen, if for some reason a cluster (or all clusters)
+ are disconnected from the federation control plane.
+ - Does not solve both the use cases listed at the beginning of this doc.
```
marun: Please provide more detail as to why this is the case.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+Its thus proposed to use the primary design alternative listed earlier in this doc.
```
marun: Given the amount of detail provided for the alternatives, I think a separate section is suggested that discusses in more detail the rational for making a decision.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+The in cluster statefulsets update proposal is under implementation as per [this proposal](https://github.com/kubernetes/community/pull/503)
+The suggestion is to handle it as a separate design proposal, once the same is implemented and stabilised.
+
+
+# Limitations (to be discussed)
+
+_1 – In case of a particular replica or the cluster containing particular replicas die, what should be the behavior._
+
+It can be resurrected with the same identity in the original cluster only
```
marun: What is the implication of not being able to resurrect an identity?  i.e. if a cluster goes away and doesn't come back?

contributors/design-proposals/federated-statefulsets.md
marun: More detail is still required here.  If storage can be replicated between clusters (at least those hosted by the same infrastructure e.g. gce), then including the cluster name in the identity doesn't seem like such a good idea.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+(2) disallow migration of replicas between clusters (simple, but perhaps overly restrictive)
+
+(3) allow migration of replicas between clusters provided that either the replica uses node-local-volumes, or the
+clusters are in the same zone (less restrictive, but more complex to implement, and for the user to understand).
+I suspect that the right answer here might be to start with (2) and add 3 in a later phase.
+
+As a proposal in this design, and this phase of implementation replicas once created in a cluster cannot be moved
+to another cluster. If a particular statefulset replica dies and the cluster and other stateful replicas in that particular cluster
+are still alive, the behavior will be within the purview of local statefulset controller to handle.
+If a cluster dies, which had stateful replicas, then the statefulset will need to run with remaining replicas.
```
marun: What if the cluster comes back?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+# Background
+
+The original design proposal of the statefulset in kubernetes, including the
+use cases and example applications, can be found [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/stateful-apps.md).
+
+# Use Cases
+
+1 – A stateful app, wants replicas distributed in multiple clusters, such that it can
+form multiple smaller sized local app clusters using only local replicas, each serving
+users local to the cluster (using in cluster stateful app identities).
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+2 – A stateful app, for the reasons of high availability, wants the stateful pods
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+distributed in different clusters, such that the set can withstand cluster failures.
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+This represents an app with one single global quorum. This use case can be an extension
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+to the previous, where the same app instances get multiple identities (local and global).
```
irfanurrehman: Its not really an implementation detail alone; I kept it as a use case to be able to support applications which support this kind of communication.
- A striking example is the use case of multi data centre consul. Details [here](https://www.consul.io/docs/internals/architecture.html#10-000-foot-view) and [here](https://www.consul.io/docs/guides/datacenters.html)
- Etcd backup and restore; Although coreos has implemented etcd operator to ease etcd use case, but etcd has a feature where, data could be synced using backup from one remote cluster (in-cluster statufulset, which also has dns identity accessible from another cluster) and restored into another.

This is not the exhaustive list; I had put in the concrete use case of galera db for multi datacentre app needing multiple identities, updated with the use case of consul now, because of the clarity in their documentation links.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+# Concrete use case examples
```
irfanurrehman: Done! Partitioned into:
- examples for local identities only
- examples for global identities only
- examples needing multiple identities

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+1 - A federated etcd statefulset of 30 replicas across 10 clusters which creates a local quorum
+in each of the 10 clusters. The potential global quorum is ignored in this example. If a cluster
+ goes down, the quorum in that cluster is dead, but the quora in all the other clusters remain
+ up accessible locally to the in-cluster apps. (Similar to use case 1 above).
+
+2 - A federated etcd statefulset of 3 or more replicas across 2 or more then 2 clusters
```
irfanurrehman: Updated!

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: If that is true and if there is a solution, which can easily enable federated objects to use or access networks across k8s clusters; I would want to pursue it.
In the absence of such available or usable solution, we go ahead with the current proposal, and improve the same when a cross cluster overlay is easy to achieve or inherently built in as cross cluster communication (probably another proposal within the scope of federation)

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+## Detailed Design (Primary Alternative)
+
+In this approach, the federated statefulset controller will partly behave quite similar to federated
+replicaset or the federated deployment controller.
+The federated controller would create and monitor individual statefulsets partitioning and distributing
+the total stateful replicas across the federated clusters.
+
+As an additional proposal in this design we suggest the possibility of the pods having multiple identities.
+This facilitates application pods being able to participate in different quora, and leaves a
```
irfanurrehman: I have given some details In reply to one your comments (https://github.com/kubernetes/community/pull/437#discussion_r124242716)
Additionally, this certainly is not something which will force a pod to have multiple identities. As mentioned the app pod "can" if apps need be. 
The two identities however are not exclusive, but orthogonal (as in our earlier discussions and per this proposal). The app can have local identity alone or local + global if it needs so. The app which needs to use "only" global identity needs to ignore the local one.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+The ordinal number assigned to the pod will be calculated and assigned in exactly the same way,
+it is assigned in the k8s cluster as described
+[here](https://kubernetes.io/docs/concepts/abstractions/controllers/statefulsets/#pod-identity) and will
+be unique.
+The details of how the replicas will be distributed across federated clusters is elaborated in
+&#39;Replica distribution&#39; section below.
+The details of the multiple dns identities for each pod is separately elaborated in
+&#39;instance identity and discovery&#39; sections below.
+
+For this phase of implementation, it is proposed that the statefulset need not guarantee the
```
irfanurrehman: I think there has been some discussion is app sig, and given the complexity of the same, I strongly think that this can be relaxed (atleast for the current phase of impl). https://github.com/kubernetes/community/pull/437#pullrequestreview-25676538

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+order of creation of the pod instances across federation.
+
+### Replica distribution (across federated clusters)
+
+The proposed default behaviour of the federation controller, when a statefulset creation request is
```
irfanurrehman: Acknowledge. I think this can be a common feature or impl across the federated objects/controllers which map to pods (probably a separate proposal).

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+behaviour will be, to partition the replica count equally into healthy clusters. This default behaviour
+in future can be overridden by applying user preferences very similar to replicaset preferences used for
+federated replicasets today. It is proposed that these preferences are kept outside the scope of the
+first phase (alpha) of the federated statefulsets implementation.
+
+### Instance identity and discovery (local identities)
+
+In the case of in-cluster statefulset, pods discover each other using the in-cluster dns names.
+In k8s, a headless service with selectors enables creation of dns records against the pod names
+for an in cluster statefulset.
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+As an inherent design requirement in this proposal, the applications which are being deployed as
```
irfanurrehman: https://github.com/kubernetes/community/pull/437#discussion_r124248774

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+the statefulset can choose to connect to the other replica instances locally (just like the local
+statefulset) within the cluster or globally across the clusters (additional federation identity) to
+form the quorum using the respective dns names.
+Applications (in case they support such a behaviour) can also communicate across quora (for example
```
irfanurrehman: https://github.com/kubernetes/community/pull/437#discussion_r124248774

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+a local db cluster connects to local replicas forming a quorum and then subsequently connects to other
+cluster/instance outside the cluster for periodic backups) using the global dns identities.
+This section deals with the cluster local stateful identities.
+
+Together with the statefulset object, the user will also need to submit the request for a headless
```
irfanurrehman: Updated!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+global (federated) dns server. The results will however be different, if the global identities
+are also requested by the user as elaborated in the next section.
+
+### Instance identity and discovery (global identities)
+
+The above is fine for the case of multiple local quora (one in each cluster), but these
+identities cannot work across the clusters, as the local pod IPs cannot be reached across
+the clusters.
+
+The proposal to make it work across the clusters is to create an additional service of type
```
irfanurrehman: I have put in the details of these additional LB service handling in the section **Handling additional LB services**.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+&#39;LoadBalancer&#39; for each stateful pod instance that is created in the clusters. The
+name of each service instance will be same as the stateful instance name assigned to the
+stateful replicas. This will be automatically created by the federated statefulset controller,
+on demand, proposed to be controlled by an annotation on the statefulset. If the annotation
+specifies so (users choice), the federation controller will create the additional service of
+type &#39;LoadBalancer&#39; against each pod replica. The key name for the annotation will be:
+```federation.kubernetes.io/statefulset-lb-needed```
+The values could be ```true``` or ```false```. The default value will be ```false```, considering
```
irfanurrehman: I think a similar solution can be achieved using ingress or nodeports. A stable federated ingress solution out of the box seems not available across cloud providers. Nodeport based solution might not be portable across cloud provider specific dns servers. I might be missing somethings. Suggestions are welcome.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+that the external load balancer is an expensive resource.
+
+The service name for the LB service, one each against the stateful pod will be the same name as
+the stateful pod unique name. Also note that these particular LB services, although federated,
+will ideally resolve to 1 IP per service. The LB services will be created with proper label
+selectors to ensure the network traffic can reach the designated pods. This would also mean
+that each stateful pod will need to be updated with an additional unique label selector which
+can be matched with by the individual LB services. The suggestion is to use
+```federated-statefulset: <stateful replica unique name>``` as the label selector.
```
irfanurrehman: Yes can do; but "name=" seems like a label in common usage for app names? I wanted to keep some label avoiding interference from common stuff. If you think "name=" is ok to use (or some other better suggestion), I can update accordingly.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+The details of creation and handling the LB services are elaborated in section **Handling
+additional LB Services** later in this doc.
+
+The trick for individual pods being able to discover this additional dns identity through the LB
+service is the federated controller first creating SRV records for each stateful pod in the
+federation dns server against the governing service domain
+```store.mynamespace.myfederation.svc.example.com``` and then later updating an A record of the IP
```
irfanurrehman: Thanks for the info.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+resolved for the specific LB service (for example ```store-c1-0.mynamespace.myfederation.svc.example.com```)
+against each of these SRV records.
+A peer pod in a stateful set then can discover its federated peers by doing a dns query against
```
irfanurrehman: I think even in the case of in-cluster statefulsets the sequence of creation does not guarantee that the dns name for the previous stateful replica will necessarily resolve to its ip when the next in sequence replica is spun off.
I think its eventual name resolution, and most of the apps ideally are resilient if the dns name does not resolve to its actual ip in the first hit. In case of federated stateful app, I am not sure if this can really be a problem. As stated in the doc earlier, its proposed already to relax the requirement of the guaranteeing the creation sequence in case of federated statefulsets.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+the governing service domain ```store.mynamespace.myfederation.svc.example.com```.
+It should be noted that as of now the federation service controller ignores service records which
+are not of type Loadbalancer, and does not update them in the federation dns server. This behaviour
+will be aptly modified to ensure that the dns records against the headless services are also updated
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+in the federated dns server if the need be. The federation service controller will determine this based
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+on the user annotation (```federation.kubernetes.io/statefulset-lb-needed```) with value ```true``` present
+on the federated headless service. Please not that this is the same annotation which is used on the
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+federated statefulset to determine, if this statefulset needs LB services or not. The federation
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+statefulset controller will update this on the federated headless service while creating it, or if
```
irfanurrehman: Updated!

irfanurrehman: Updated to -> "the stateful set controller will synchronize the headless service with the annotation value"

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+its applied/updated later, will update it accordingly.
+Additionally the information about the stateful replicas (and the same name LB services) for which the
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+service controller should update the dns records against the headless service will also need to be passed
+on to the headless service. The proposal is to pass this information through annotations on the headless
+service as below:
+```
+federation.kubernetes.io/stateful-replica-list: <replica-name-1>, <replica-name-2>, ..., <replica-name-n>
+```
+The federation service controller, if finds the names in this list against a federated headless service,
+will look for matching federated LB services and update the CNAME record for the headless service with the
+relevant SRV records for all the LB services in the federation dns server.
+This list can certainly be updated to the first class API data, if the federation API objects and clients
```
irfanurrehman: I was trying to state that if in future we get federation specific data as an extention in the api object definitions, this can be added there, until then it remains as the annotation. Considering this is true with all annotations, removed this sentence as a specific here.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+are updated and maintained separately to have additional fields/data then that present in k8s API.
+
+Once this happens, the stateful pods can discover peers by querying the governing service domain (federated
+headless service). Taking the same example listed earlier if the peer (or a user) queries the service
+```store.mynamespace.myfederation.svc.example.com```, the below SRV records will be returned.
+```
+store-c1-0.etcd.mynamespace.myfederation.svc.example.com
```
irfanurrehman: Have added some text relating the storage at the location describing the naming in "Instance identity and discovery (local identities)".

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+store-c2-2.etcd.mynamespace.myfederation.svc.example.com
+store-c2-3.etcd.mynamespace.myfederation.svc.example.com
+store-c3-0.etcd.mynamespace.myfederation.svc.example.com
+store-c3-1.etcd.mynamespace.myfederation.svc.example.com
+store-c3-2.etcd.mynamespace.myfederation.svc.example.com
+```
+The global federation dns service will have SRV records and relevant A records listing IPs of the LBs.
+The local in-cluster kube-dns will have the SRV records and the relevant A records listing the local
+stateful set pod IPs. The local kube-dns will be able to respond with both the local and global records,
+when a query hits the local kube-dns. The peers can choose to connect to the local identities or the
```
irfanurrehman: I think its app specific and app decides, which identity to use. We just provide a feature where a federated app can have that facility.

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+global ones as the need be.
+
+#### Handling additional LB services
+
+As listed in the above approach, the current option of exposing the statefulset pods across clusters is to
+assign a LB service to each replica pod.
+If the federated service controller is used (by creating the service in federation) to create these services,
+then as per the current behaviour the service will be created in each cluster.
+This leads to exactly one unused service per statefulset per cluster.
```
irfanurrehman: Actually that is not what I did intend to state. The default behaviour of federated service creation is create the same object in all federated clusters. 
So for example if the service named etcd-cluster1-0 is created in federation, all clusters will have one service each with this name.
The cluster selector now facilitates the behaviour needed for this proposal. 
Did you mean to say, that the service will be of use only in the cluster hosting the pod ?

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+1 - Cluster affinity is introduced for services such that the controller creates the service only
+in needed cluster. The need of cluster affinity and anti-affinity in general is discussed
+[elsewhere](https://github.com/kubernetes/kubernetes/issues/41442) also, and as of today is already
+available in federation.
+
+2 - Federated statefulset controller handles the creation of cluster local services, rather than the
+federated service controller. It also will need to handle the dns records for each LB service into the
+federation (or cloud provider specific) public dns server.
+
+We propose using alternative 1 listed here, as it fits broader scheme of things and is more consistent with
```
irfanurrehman: I don't think these objects need to be treated any way different.
Say, a pod created by a deployment can be queried and listed like any other pod.
Same with the lb type automatically created by federated statefulset. 
Also, if a service with the identical name (but with other properties exists already), it will be overwritten, similar to the cascaded objects behaviour (cascading from federation into clusters or cascading from an object to another object)

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+
+There are two ways a federated statefulset can be assigned a persistent storage.
+
+1 - A node local volume attached to the container
+
+2 - A persistent volume claim from the cloud provider.
+
+Either of the 2 options above can be used/specified by the user to attach a persistent store to the pods.
+
+It is known, that as of now, the same persistent volume, even if can be used in a different cluster,
```
irfanurrehman: Updated!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+k8s directly does not provide an API, yet, which can aid the same. In the absence of no direct way of
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+quick migration of the storage data from one zone to another, or from one cloud provider to another if the
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
```diff
@@ -0,0 +1,491 @@
...
+case be, the proposal is to disallow migration/movement of pods across clusters.
```
irfanurrehman: Done!

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: We can probably discuss this once, if you think this approach might make more sense.
To me it has few inherent drawbacks which hold me back from elaborating it as a third (or extension of the second alternative):
- One more controller over the same objects (k8s statefulset objects which can be queried from individual cluster), seems two point of control on the same object.
- If the new local controller monitors only stateful pods assigned to its cluster, then the statefulset object is not locally query-able.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: Probably, this statement was not entirely correct, have removed it.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: Have updated an additional section as "Conclusion" and tried listing pros cons of both alternatives.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: As listed further, down the statefulset will need to run without that particular replica ( or replicas if there were more in that cluster). If and when the cluster joins back, the stateful replicas in that cluster will join back to the app quorum.

contributors/design-proposals/federated-statefulsets.md
irfanurrehman: As above!

contributors/design-proposals/federated-statefulsets.md
chrislovecnm: @irfanurrehman because this is internal to the cluster using a networking solution that can mesh or expose pod ips, is a possible solution.  The networking is complex, so for smaller deployments loadbalancers may be another solution as well.

This is a challenging problem to say the least ;) But I have done a POC with weave and have gotten the network setup that I would need.

jwaldrip:
Why was this closed?

kow3ns:
It auto closed because it was not merged and has not received attention. I don't think this proposal is complete enough to implement.

irfanurrehman:
@jwaldrip, as @kow3ns pointed out correctly, it was autoclosed. Meanwhile sig multicluster relooked at its priority list a little while back, and we refocused our efforts to moving the federation code out of core first and then moving the existing features to GA as the top priorities, rather then implementing advanced features like this now. This does not mean, we are not going to implement this; however it will take some time until we get back to this (probably a quarter). Reopening this, as this will be implemented anyhow.

irfanurrehman:
@kubernetes/sig-multicluster-feature-requests

dionysius:
Don't know if I'm allowed to respond here. I am waiting for this feature, so I don't want this issue to be closed, otherwise it gets completely forgotten...

quinton-hoole:
@dionysius This is not dead.  It's being pursued in https:://github.com/kubernetes/federation . Don't worry :-)

PR #88, opened 2016-09-27
move kube-mesos-framework to incubator project

deads2k:
The code inside of `kubernetes/contrib/mesos` should get promoted to its own repository.  It is built on top of Kubernetes, but it adapts those concepts for use as a mesos framework.  It is distinct from the core Kubernetes code base.

We've discussed the idea in https://github.com/kubernetes/kubernetes/issues/33283 and since it's already part of the Kubernetes source tree, we have an existing OWNERS file to use to prime the new repo.

@k82cn @smarterclayton @thockin

PR #446, opened 2017-03-13
Proposal: Public IPs per Pod

slaskawi:
Hey!

Please have a look at a design proposal for solving kubernetes/kubernetes#28660

It is very likely that the newly created interface for External IPs would also be affected by kubernetes/features#88. However, because of an early stage of #88, it has not been included in this proposal.

Thanks,
Sebastian Łaskawiec

### DIFF
--- a/contributors/design-proposals/public-identities-for-pods.md
+++ b/contributors/design-proposals/public-identities-for-pods.md
@@ -1,0 +1,141 @@
+# Public identities for Pods
+
+* Authors: Sebastian Łaskawiec (@slaskawi)
+* Last edit: [2017-03-13](#history)
+* Status: design
+
+Approvers:
+* [ ] smarterclayton
+* [ ] bprashanth
+* [ ] enisoc
+* [ ] foxish
+* [ ] janetkuo
+* [ ] kargakis
+* [ ] kow3ns
+* [ ] thockin
+
+**Table of Contents**
+
+* [Goals](#goals)
+* [Non-goals](#non-goals)
+* [API](#api)
+* [Behavior](#behavior)
+* [Upgrading](#upgrading)
+* [Implementation](#implementation)
+* [Alternatives](#alternatives)
+* [History](#history)
+
+# Goals
+
+* The main goal of Public Identities for Pods is to expose each Pod as an external IP address.
+  This approach allows sophisticated applications to communicate with the outside world using custom protocols.
+  A examples which require this approach are:
+  * Game Servers (which often use RTP, UDP)
+  * Data Grids (with L7 routing, such as [Infinispan](http://infinispan.org/docs/dev/user_guide/user_guide.html#using_hot_rod_server))
+  * Media Servers
+
+* The secondary goal is to create an API similar to `LoadBalancer` to manage external IP addresses.
+
+* The implementation is targeted (but not limited to) StatefulSets. Stable hostnames allow to identify the same
+  Stateful Pod after crash or restart (it should have a stable hostname). However it also makes sense to expose
+  ordinary deployments this way.
+
+# Existing solution
+
+* The above goal is partially possible to achieve with Loadbalanced Services, NodePort Services or
+  Ingress with session stickiness (if HTTP is the main protocol used for communication) but very often
+  this is not good enough and application clients assume there is only one application instance
+  behind an IP address (in other words there 1 to 1 mapping between IP addresses and Application
+  instances). Therefore a specialized implementation exposing an IP address per Pod is required.
+
+# API
+
+A new type of Service needs to be created in order to indicate that an external IP address will be exposed per Pod:
+
+```go
+const (
+    //Existing Service types
+	ServiceTypeClusterIP ServiceType = "ClusterIP"
+	ServiceTypeNodePort ServiceType = "NodePort"
+	ServiceTypeLoadBalancer ServiceType = "LoadBalancer"
+	ServiceTypeExternalName ServiceType = "ExternalName"
+
+	//new Service type - expose IP per Pod
+	ServiceTypeIPPerPod ServiceType = "IPPerPod"
+)
+```
+
+A new interface in `cloud.go` will be created to allow creating an external IP addresses.
+
+```go
+// ExternalIP is an abstract, pluggable interface for external ips.
+type ExternalIP interface {
+	// GetExternalIP returns whether the specified external IP exists, and
+	// if so, what its status is.
+	// Implementations must treat the *v1.Service parameter as read-only and not modify it.
+	// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+	GetExternalIP(clusterName string, service *v1.Service) (status *v1.ExternalIPStatus, exists bool, err error)
+	// EnsureExternalIP creates a new external IP 'name', or updates the existing one.
+	// Returns the status of the external IP.
+	// Implementations must treat the *v1.Service
+	// parameters as read-only and not modify them.
+	// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+	EnsureExternalIP(clusterName string, service *v1.Service) (*v1.ExternalIPStatus, error)
+	// EnsureExternalIPDeleted deletes the specified external IP if it
+	// exists, returning nil if the external IP specified either didn't exist or
+	// was successfully deleted.
+	// Implementations must treat the *v1.Service parameter as read-only and not modify it.
+	// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+	EnsureExternalIPDeleted(clusterName string, service *v1.Service) error
+}
+```
+
+Finally a new Annotation will be created to allow falling from "IP per Pod" back into "LoadBalancer per Pod" solution
+(more abot this in the following paragraphs).
+
+```go
+const FallbackToLoadBalancerPerPod = "service.beta.kubernetes.io/fallback-loadbalancer-per-pod"
+```
+
+# Behavior
+
+The logic implemented behind assigning and updating external IPs will be similar to Load Balanced Services.
+
+A ServiceController will periodically iterate though Pods governed by a Service and assign an external IP to each of them.
+Note that there is no `UpdateExternalIP` function in `ExternalIP` interface. This means that a single IP address will not
+be reassigned into another Pod. It will be explicitly deleted (by `EnsureExternalIPDeleted`).
+
+Since `ExternalIP` interface might not be implemented and supported by all Cloud Vendors, we allow creating a Load Balancer
+per Pod. This should however be considered as a last-resort approach. In that case, a user will need to add
+`FallbackToLoadBalancerPerPod` annotation on a Service.
+
+# Implementation
+
+Checked items had been completed at the time of the [last edit](#history) of
+this proposal.
+
+* [ ] Add `ServiceTypeIPPerPod`, `ExternalIP` and `FallbackToLoadBalancerPerPod`
+* [ ] Add proper validation (e.g. a nil `ClusterIP` for the Service etc)
+* [ ] Implement logic inside `ServiceController`
+* [ ] Implement e2e tests `ServiceController`
+* [ ] Add `FallbackToLoadBalancerPerPod` fallback mechanism
+* [ ] Implement e2e tests for `FallbackToLoadBalancerPerPod`
+
+# Alternatives
+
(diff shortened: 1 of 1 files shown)


slaskawi:
cc @bprashanth @enisoc @foxish @janetkuo @kargakis @kow3ns @smarterclayton

justinsb:
cc @kubernetes/sig-network-misc

contributors/design-proposals/public-identities-for-pods.md
```diff
@@ -0,0 +1,132 @@
+# Public identities for Pods
+
+* Authors: Sebastian Łaskawiec (@slaskawi)
+* Last edit: [2017-03-13](#history)
+* Status: design
+
+Approvers:
```
thockin: please put me in this list

contributors/design-proposals/public-identities-for-pods.md
```diff
@@ -0,0 +1,132 @@
...
+	ServiceTypeNodePort ServiceType = "NodePort"
+	ServiceTypeLoadBalancer ServiceType = "LoadBalancer"
+	ServiceTypeExternalName ServiceType = "ExternalName"
+
+	//new Service type - expose IP per Pod
+	ServiceTypeIPPerPod ServiceType = "IPPerPod"
+)
+```
+
+A new interface in `cloud.go` will be created to allow creating an external IP addresses.
```
thockin: CloudProvider is a mostly frozen API and is being disassembled. We should not rely on it being ubiquitous.  It's fine as a model, though.

contributors/design-proposals/public-identities-for-pods.md
```diff
@@ -0,0 +1,132 @@
...
+	// if so, what its status is.
+	// Implementations must treat the *v1.Service parameter as read-only and not modify it.
+	// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+	GetExternalIP(clusterName string, service *v1.Service) (status *v1.ExternalIPStatus, exists bool, err error)
+	// EnsureExternalIP creates a new external IP 'name', or updates the existing one.
+	// Returns the status of the external IP.
+	// Implementations must treat the *v1.Service and *v1.Node
+	// parameters as read-only and not modify them.
+	// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+	EnsureExternalIP(clusterName string, service *v1.Service, nodes []*v1.Node) (*v1.ExternalIPStatus, error)
```
thockin: I don't understand this API.  What is the service argument for?  Who writes a status back to Service?  What is the nodes argument for?  How do I know which backend pod this IP is supposed to be for?

I am having a hard time imagining the implementation of this..

contributors/design-proposals/public-identities-for-pods.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Behavior
+
+The logic implemented behind assigning and updating external IPs will be similar to Load Balanced Services.
+
+A ServiceController will periodically iterate though Pods governed by a Service and assign an external IP to each of them.
+Note that there is no `UpdateExternalIP` function in `ExternalIP` interface. This means that a single IP address will not
+be reassigned into another Pod. It will be explicitly deleted (by `EnsureExternalIPDeleted`).
+
+Since `ExternalIP` interface might not be implemented and supported by all Cloud Vendors, we allow creating a Load Balancer
```
thockin: Is it implementable in ANY provider?  GCP's ForwardingRule is close, but it is a load-balancer underneath.  ELB is a non-transparent LB, which depends on nodePorts (which are not addressed in this doc at all).

contributors/design-proposals/public-identities-for-pods.md
```diff
@@ -0,0 +1,132 @@
...
+per Pod. This should however be considered as a last-resort approach. In that case, a user will need to add
```
thockin: Why is it a last resort?

philips:
I want to echo @thockin's sentiment: I don't get this.

It would be great if a section could be written about the alternatives that could be considered for this proposal and why they are untenable for your use case.

contributors/design-proposals/public-identities-for-pods.md
kow3ns: My guess would be $$$. LBs are not cheap.

contributors/design-proposals/public-identities-for-pods.md
slaskawi: Sure, no problem.

contributors/design-proposals/public-identities-for-pods.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Table of Contents**
+
+* [Goals](#goals)
+* [Non-goals](#non-goals)
+* [API](#api)
+* [Behavior](#behavior)
+* [Upgrading](#upgrading)
+* [Implementation](#implementation)
+* [Alternatives](#alternatives)
```
slaskawi: Sorry, I think I messed something up when editing the file. Let me add it.

contributors/design-proposals/public-identities-for-pods.md
```diff
@@ -0,0 +1,132 @@
...
+# Goals
+
+* The main goal of Public Identities for Pods is to expose each Pod as an external IP address.
+  This approach allows sophisticated applications to communicate with the outside world using custom protocols.
+  A examples which require this approach are:
+  * Game Servers (which often use RTP, UDP)
+  * Data Grids (with L7 routing, such as [Infinispan](http://infinispan.org/docs/dev/user_guide/user_guide.html#using_hot_rod_server))
+  * Media Servers
+
+* The secondary goal is to create an API similar to `LoadBalancer` to manage external IP addresses.
```
slaskawi: Yes and no. I wanted to propose a similar API to Load Balancers and give cloud vendors a chance to allocate an IP address available externally. I imagine some of them (cloud vendors) might not want to abuse Load Balancers (they might be expensive and @smarterclayton mentioned that checking their health might be tricky).

contributors/design-proposals/public-identities-for-pods.md
slaskawi: Thanks for letting me know. 

Let me also explain myself a little bit. This is my first proposal to Kubernetes and don't have much hands-on experience with its code base (I work on Data Grid solutions on daily basis). Please forgive me any obvious mistakes that you might see. I'm also open for any suggestions to improve this proposal.

So as I mentioned in a comments a little bit higher, I wanted to give a chance to cloud vendors to implement a kind of External IP Allocator. This might be a much cheaper solution opposed to abusing Load Balancers. Unfortunately that would require adding some bits into `cloud.go`. Of course I might be totally wrong here and there might be a different mechanism for adding such capabilities. In that case, could you please advice me how can I achieve this?

contributors/design-proposals/public-identities-for-pods.md
slaskawi: I think I made a mistake here. I wanted to pass a clusterName, a Service (which governs a clustered app deployed in Kubernetes, such as Data Grid) and a list of Pod (not Nodes, I'm sorry). I'm pretty sure the cloud vendor could easily extract a list of Pods from governing Service, so let me remove the last argument.

The reason for returning a status here is that some cloud vendors might not implement this api and we will return an error here. Then we could fall back to abusing Load Balancers for the instance.

contributors/design-proposals/public-identities-for-pods.md
slaskawi: I would kindly ask for your advice here. I imagine this could be implemented very easily by a pool of externally available IP addresses and a simple allocator. But my (very naive) vision might not be implementable. What do you think?

contributors/design-proposals/public-identities-for-pods.md
slaskawi: Exactly!

slaskawi:
@philips Thanks for the feedback. The alternatives section has been added.

slaskawi:
I hope I address most of the concerns in my comments. I also added `alternatives` section which was missing.

Just to let you know, I also started to work on a proof-of-concept which allocates a Load Balancer per Pod without touching Kubernetes codebase. A very dirty code might be found [here](https://github.com/slaskawi/external-ip-proxy). The demo creates an additional binary proxy between Load Balancer and target Pod (I required it for some testing and will be removing it soon).

psycotica0-shopify:
I have also written a controller that creates a LoadBalancer service for each pod, scoped by adding an ordinal tag to each pod.
It works, but it'd be cool if the system had something built in for it.

My use case is Kafka, which is really great with statefulsets, but users of the system must be able to individually address brokers. Can't just throw them all behind a simple LB.

contributors/design-proposals/public-identities-for-pods.md
thockin: The problem is that, at least for some clouds, an external IP == a load-balancer.  We already have the primitives you need to get an IP assigned, without making a new API.

contributors/design-proposals/public-identities-for-pods.md
thockin: It doesn't map to any cloud I know.  I get what you're trying to express but it is more complicated than you imagine.  It will required either iptables rules, routing rules, or node ports (or combinations thereof) - which is exactly what Service provides.

contributors/design-proposals/public-identities-for-pods.md
thockin: I don't know what the first-resort is - I'm looking for even one case where this is implemntable in a way that doesn't exactly equal Services.

kow3ns:
@slaskawi are you proposing to provision elastic IPs (AWS) or ephemeral/static external IPs (GCP) as opposed to LBs on a per Pod basis and to forward traffic from the  external address to the Pod?

psycotica0-shopify:
Not to hijack the proposal, but if there are concerns that this might be similar to Services, what if there was a field on `StatefulSet` called `serviceTemplate`, similar to `volumeClaimTemplates` where you can put a service that will be created and scoped to each pod individually.

Then, if you want to use an LB, you can put that as the type in the spec.
But, if some cloud provides a service type that represents a route from ExternalIP directly to a pod somehow, then you can put that there instead?

Service spec already allows specifying a static IP from a pool you've already got allocated. (Though indexing from the generated service into that might be a little tricky compared to how things already work... I don't know if there are patterns for this)

slaskawi:
@kow3ns Yes, my initial though was to create a new interface for cloud provider which would allow them to provision an Elastic IPs or Static External IPs. But it seems there might be problems to implement it on cloud provider side. See https://github.com/kubernetes/community/pull/446#discussion_r110556059 and https://github.com/kubernetes/community/pull/446#discussion_r110556262

@psycotica0-shopify I actually think those `serviceTemplate`s could really do the job here. Plus, as you mentioned, they would look consistent to `volumeClaimTemplates`s. @thockin convinced me that we should be really looking more into load balancer-based solution rather than trying to expose an External IP (please correct me if I got it wrong). So adding those two together seems an idea to move this design forward.

contributors/design-proposals/public-identities-for-pods.md
slaskawi: @kow3ns mentioned in the comment below about Elastic IP (AWS) and Static External IPs (GCP). Is it possible to use those things somehow?

thockin:
A GCP External IP *is* a load-balancer.  There's really no difference.  I
am not deeply familiar with how AWS elastic IPs work, but to get a
VM-oriented IP into a container requires some work in the node beyond just
assigning an IP.  That's part of why I don't like this - we already do that
work, it's called Service.

If you want to explore it, I can't stop you.  I can say that I don't think
we want a generic construct of external IP for pods.  This is why
Kubernetes has (and is adding more) extension points.  You should be able
to write a controller and per-node agent to manage this all.

slaskawi:
I think I collected enough feedback to summarize this discussion and design. Thank you all for lots and lots of interesting comments.

So the bottom line is that an External IP is a Load Balancer for most of the cloud vendors. So there is no need to duplicate this sort of functionality.

Just before I close this Pull Request, I would like to ask one last question. @thockin Could you please tell me if there are any plans for implementing something similar to `serviceTemplate`s that @psycotica0-shopify  proposed? If not I would be happy to investigate this approach.

thockin:
I don't think there are PLANS per se, but it's not the first time it has come up, so maybe worth exploring with sig-apps (StatefulSet)

slaskawi:
@thockin Ok, great! Thanks for all the guidance!

PR #447, opened 2017-03-13
Update identifiers.md

zhangxiaoyu-zidif:
change "Since" to "If"


philips:
The old language is correct as it is talking about empty fields that are provided, not if provided.

PR #450, opened 2017-03-14
Expand details on security fix development and release process

ixdy:
(adding all folks in the PST list as reviewers)

I've added details on the development and release process for security fixes, which should guide the changes necessary to our test and release infrastructure.

As a starting point, I've made some opinionated choices, partially based on how regular public releases have worked in the past. 

There are a number of TODOs (especially around disclosure timelines) that I don't yet have answers for and would appreciate advice.

### DIFF
--- a/contributors/devel/security-release-process.md
+++ b/contributors/devel/security-release-process.md
@@ -115,29 +115,33 @@ The communication to users should be actionable. They should know when to block 
 
 **Fix Release Day** (Completed within 1-21 days of Disclosure)
 
-- The Release Managers will ensure all the binaries are built,
-  publicly available, and functional before the Release Date.
-    - Note: we ship source with our binary releases, so as soon as binary
-      artifacts are publicly available, the fixes (and vulnerabilities) are
-      effectively public.
+- Before the Release Date, the Release Manager(s) will ensure all the binaries
+  are built, publicly available, and functional.
+  - Note: since we ship source with our binary releases, as soon as binary
+    artifacts are publicly available, the fixes (and vulnerabilities) are
+    effectively public.
   - CI testing of the private patch release branch should provide confidence in
     the release.
-- The Release Manager will use release tooling (i.e. `anago`) to
-  - Compute the version for the new security patch release; in the previous
-    example, the `release-2.4.7` branch would produce a `v2.4.8` release.
-  - Verify that no public patch release has occurred since the security branch
-    was cut (in this example, make sure that `v2.4.8` has not been tagged
-    publicly).
-  - Perform a normal release build from the private release branch, staging
-    binary artifacts locally (or possibly in a private bucket).
-  - Push the new patch release tag to the private repo.
-  - The tooling should **not** upload any artifacts to public storage buckets or
-    create a draft release on the public GitHub repo.
-- As soon as we are ready to make the fix known publicly (TODO: when?),
-  the Release Manager will use tooling to
-  - Publish results.
-    - Upload binary artifacts to GCS, gcr.io, and GitHub. (TODO: are these
-      prebuilt from the previous step?)
+  - The Release Manager will use release tooling (i.e. `anago`) to
+    - Compute the version for the new security patch release; in the previous
+      example, the `release-2.4.7` branch would produce a `v2.4.8` release.
+    - Verify that no public patch release has occurred since the security branch
+      was cut (in this example, make sure that `v2.4.8` has not been tagged
+      publicly).
+    - Perform a normal release build from the private release branch, staging
+      binary artifacts locally (or possibly in a private bucket).
+    - Push the new patch release tag to the private repo.
+    - The tooling should **not** upload any artifacts to public storage buckets or
+      create a draft release on the public GitHub repo.
+- On the Release Date, the Release Manager will use tooling to
+  - Verify that nothing new has been pushed to the private release repository,
+    and that no new public patch release has occurred.
+      - If there are new commits on the private release branch, or if a new
+        public release has occurred, the previously build artifacts will be
+        discarded.
+  - Publish the release.
+    - Upload prebuilt binary artifacts (from previous step) to GCS, gcr.io, and
+      GitHub.
     - Update the `CHANGELOG.md` in the **public** `master` branch
     - Create a new release on the **public** kubernetes GitHub
     - Send email announcing the release
@@ -146,7 +150,7 @@ The communication to users should be actionable. They should know when to block 
     - In our continuing example, this would merge the private `release-2.4.7`
       branch into the public `release-2.4` branch, effectively advancing the
       latter to `2.4.9-beta.0+`. The `git describe --tags` will update to
-      '2.4.9-beta.0+` and any changes in the `release-2.4` branch will
+      `2.4.9-beta.0+` and any changes in the `release-2.4` branch will
       now ship with `v2.4.9`.
 - The Fix Lead will request a CVE from [DWF](https://github.com/distributedweaknessfiling/DWF-Documentation) and include the CVSS and release details.
 - The Fix Lead will email kubernetes-{dev,users,announce,security-announce}@googlegroups.com now that everything is public announcing the new releases, the CVE number, the location of the binaries, and the relevant merged PRs to get wide distribution and user action. As much as possible this email should be actionable and include links how to apply the fix to users environments; this can include links to external distributor documentation.
--- a/contributors/devel/security-release-process.md
+++ b/contributors/devel/security-release-process.md
@@ -67,16 +67,20 @@ These steps should be completed within the 1-7 days of Disclosure.
    `v2.4`, a new branch will be created in the private fork called
    `release-2.4.7`, branching at the `v2.4.7` tag. If `HEAD` were to be used
    instead, the security release might include non-security fixes.
-  - TODO: continuous integration testing should be set up for the branch
+   - Alternatively, this branch could be named something like `issue-N`,
+     referencing the GitHub issue tracking the vulnerability and fix.
+  - TODO: continuous integration testing should be set up for the branch,
+    ideally with only a very small configuration change.
 - If a security incident affects multiple supported releases, then patch
-  release branches will be created for all affected releases.
-- Similarly, if a security issue affects `master`, then a private `master`
-  branch will be created, though this will be synced to public `HEAD`, rather
-  than from a particular tag.
-- The Fix Team will submit fixes to the private branch(es). If a fix is only
-  needed for a published release (and not `master`), then it is fine to merge
-  directly into the private release branch (rather than following the
-  cherry-pick process).
+  release branches will be created for all affected releases. Vulnerabilities
+  affecting `master` will be fixed via cherry-picks after fixes to the release
+  branches have been publicized.
+- If multiple security issues are reported simultaneously, they will be combined
+  into a single security release target, unless it is deemed infeasible to fix
+  all vulnerabilities at once (e.g. if one vulnerability will take significantly
+  longer to fix).
+- The Fix Team will merge fixes directly to the private branch(es), rather than
+  following the normal cherry-pick process
   - As is feasible, the normal PR review process (including LGTM and OWNERS
     approval) should be followed for all fixes, though the submit queue
     infrastructure may not be operational on the private branches.
@@ -87,8 +91,7 @@ These steps should be completed within the 1-7 days of Disclosure.
   releases while security fixes are ongoing. Should it be necessary to create a
   new public patch release before the security fixes are complete, the private
   security branch will need to be rebased (and renamed?) on the new latest
-  patch release.
-- TODO: how frequently should the private `master` branch be updated?
+  patch release. Only one security release will be built.
 
 If the CVSS score is under 4.0 ([a low severity score](https://www.first.org/cvss/specification-document#i5)) the Fix Team can decide to slow the release process down in the face of holidays, developer bandwidth, etc. These decisions must be discussed on the kubernetes-security mailing list.
 
@@ -139,20 +142,21 @@ The communication to users should be actionable. They should know when to block 
     - Create a new release on the **public** kubernetes GitHub
     - Send email announcing the release
   - Merge the private patch release branch into the **public** release branch
-    including the fixes and release tag.
+    including both the commits and release tag.
     - In our continuing example, this would merge the private `release-2.4.7`
       branch into the public `release-2.4` branch, effectively advancing the
-      latter to `2.4.9-beta.0+`. Any changes in the `release-2.4` branch will
+      latter to `2.4.9-beta.0+`. The `git describe --tags` will update to
+      '2.4.9-beta.0+` and any changes in the `release-2.4` branch will
       now ship with `v2.4.9`.
-  - If fixes were necessary to `master`, merge the private `master` branch into
-    the public `master` branch. No release will be built or published for
-    `master`.
-    - TODO: should this be a PR or a push?
 - The Fix Lead will request a CVE from [DWF](https://github.com/distributedweaknessfiling/DWF-Documentation) and include the CVSS and release details.
 - The Fix Lead will email kubernetes-{dev,users,announce,security-announce}@googlegroups.com now that everything is public announcing the new releases, the CVE number, the location of the binaries, and the relevant merged PRs to get wide distribution and user action. As much as possible this email should be actionable and include links how to apply the fix to users environments; this can include links to external distributor documentation.
 - The Fix Lead will remove the Fix Team from the private security repo.
-- The Fix Lead will (possibly?) delete the patch release branch from the private
-  security repo.
+- The Fix Lead will delete the patch release branch from the private security
+  repo.
+- If fixes are needed for `master`, the Fix Lead (or chosen delegate) will open
+  PRs against the public `master` branch containing cherry-picks of the commits
+  now merged into the release branch(es). No release will be built or published
+  for `master`.
 
 ### Retrospective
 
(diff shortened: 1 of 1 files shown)


contributors/devel/security-release-process.md
```diff
@@ -58,7 +58,37 @@ These steps should be completed within the first 24 hours of Disclosure.
...
+- To facilitate this goal, the Fix Lead will [create a patch release branch](https://github.com/kubernetes/release/blob/master/docs/branching.md#branching-from-a-tag)
+  in a private fork of the public kubernetes.git repository, branched from the
+  most recent released tag for the release being fixed.
+  - As a hypothetical example, if `v2.4.7` is the latest public patch release for
+   `v2.4`, a new branch will be created in the private fork called
+   `release-2.4.7`, branching at the `v2.4.7` tag. If `HEAD` were to be used
+   instead, the security release might include non-security fixes.
+  - TODO: continuous integration testing should be set up for the branch
+- If a security incident affects multiple supported releases, then patch
+  release branches will be created for all affected releases.
```
timstclair: What if there are multiple simultaneous security issues? I imagine we would develop them in the same branches, and release them together?

contributors/devel/security-release-process.md
```diff
@@ -58,7 +58,37 @@ These steps should be completed within the first 24 hours of Disclosure.
...
+- Similarly, if a security issue affects `master`, then a private `master`
+  branch will be created, though this will be synced to public `HEAD`, rather
+  than from a particular tag.
```
timstclair: I think this is too much of a moving target. Since master is not intended for production use, I recommend releasing the patches for the patch releases, and then cherrypicking the patch to master *after* the public disclosure (in public, through the regular patch process). WDYT?

contributors/devel/security-release-process.md
```diff
@@ -58,7 +58,37 @@ These steps should be completed within the first 24 hours of Disclosure.
...
+    approval) should be followed for all fixes, though the submit queue
+    infrastructure may not be operational on the private branches.
+- The Fix Team will notify the Fix Lead that work on the fix branch(es) is
+  complete once all relevant changes have been merged into the private
+  branch(es).
+- Release Managers are encouraged to refrain from creating new public patch
+  releases while security fixes are ongoing. Should it be necessary to create a
+  new public patch release before the security fixes are complete, the private
+  security branch will need to be rebased (and renamed?) on the new latest
+  patch release.
```
timstclair: :+1: Since the patch would only be for the latest release, admins will be required to pick up the most recent patch release first. Actually, should we consider building a patch for both versions in this case? (e.g. in your above example, `v2.4.8` is released without the security fix, so we create `v2.4.9` as a security releases, and also `v2.4.7-r2`)

contributors/devel/security-release-process.md
ixdy: That would probably be ideal, though it might be reasonable in certain cases to have separate releases, e.g. if it's expected that one issue may take significantly longer to fix.

contributors/devel/security-release-process.md
ixdy: That's a good point. The reason I suggested this here was to ensure that the fixes landed in `master` promptly and aren't forgotten. I guess we can add a step to this process somewhere; what is a reasonable deadline for merging fixes in to `master` after the vulnerability is publicized?

(Note: our cherrypick script probably doesn't support this type of cherry-picking yet and will need to be updated too.)

contributors/devel/security-release-process.md
ixdy: You raise a good point.

I think we should only build a single security patch release for each issue - so if we branch from `v2.4.7`, and `v2.4.8` is released without a fix, we ship only `v2.4.9` with the fix (after rebasing on `v2.4.8`). I would guess that the vulnerability would've existed in all `v2.4` patch releases, and we certainly don't want to build and release a `v2.4.0-r2`, `v2.4.1-r2`, etc. (I guess this raises questions about support commitments - do we support users at every patch release for a given X.Y release? Or only the latest patch release? We certainly don't regularly test anything besides latest patch release.)

The challenge here is that delaying the non-security patch release may reveal the existence of a vulnerability, though that's probably OK, especially if the disclosure of fix (detailed in the next section) has already occurred.

If there is a serious bug affecting production clusters which needs the non-security patch immediately, then this is a reasonable case to break the "no non-security fixes" rule.
In such cases, we'd probably need to send a follow-up to the original disclosure email, noting that the public patch release doesn't include the fix, but recommending an upgrade to make the security upgrade go as smoothly as possible.

contributors/devel/security-release-process.md
```diff
@@ -58,7 +58,37 @@ These steps should be completed within the first 24 hours of Disclosure.
...
+- To encourage fast adoption of the security fix, the new patched release should
+  contain only security fixes.
+- To facilitate this goal, the Fix Lead will [create a patch release branch](https://github.com/kubernetes/release/blob/master/docs/branching.md#branching-from-a-tag)
+  in a private fork of the public kubernetes.git repository, branched from the
+  most recent released tag for the release being fixed.
+  - As a hypothetical example, if `v2.4.7` is the latest public patch release for
+   `v2.4`, a new branch will be created in the private fork called
+   `release-2.4.7`, branching at the `v2.4.7` tag. If `HEAD` were to be used
+   instead, the security release might include non-security fixes.
+  - TODO: continuous integration testing should be set up for the branch
```
liggitt: agree on scoping the fixes, but I'd like to avoid having reconfigure N CI jobs every time a fix is needed.

contributors/devel/security-release-process.md
liggitt: I would expect so

contributors/devel/security-release-process.md
liggitt: agree.

contributors/devel/security-release-process.md
liggitt: I'd expect just v2.4.9

contributors/devel/security-release-process.md
ixdy: My hope is that at some point the test infrastructure is sophisticated enough that we just tell it which branches we want to test, and it sets up a predetermined set of jobs against that branch. 

(This would be mainly used for the public release branches, but we should also be able to set it up for private branches.)

Setting up this test-infra would likely again reveal the presence of a security vulnerability (since the test configs are all public), but hopefully we would have disclosed its existence already (as previously noted).

contributors/devel/security-release-process.md
```diff
@@ -58,7 +58,37 @@ These steps should be completed within the first 24 hours of Disclosure.
...
+    infrastructure may not be operational on the private branches.
+- The Fix Team will notify the Fix Lead that work on the fix branch(es) is
+  complete once all relevant changes have been merged into the private
+  branch(es).
+- Release Managers are encouraged to refrain from creating new public patch
+  releases while security fixes are ongoing. Should it be necessary to create a
+  new public patch release before the security fixes are complete, the private
+  security branch will need to be rebased (and renamed?) on the new latest
+  patch release.
+- TODO: how frequently should the private `master` branch be updated?
```
liggitt: I'd expect all the branches in the private repo to be fast-forwarded on an as-needed basis

contributors/devel/security-release-process.md
```diff
@@ -58,7 +58,37 @@ These steps should be completed within the first 24 hours of Disclosure.
...
 - The Fix Lead and the Fix Team will create a [CVSS](https://www.first.org/cvss/specification-document) using the [CVSS Calculator](https://www.first.org/cvss/calculator/3.0). The Fix Lead makes the final call on the calculated CVSS; it is better to move quickly than make the CVSS prefect.
-- The Fix Team will notify the Fix Lead that work on the fix branch is complete once there are LGTMs on all commits in the private repo from one or more relevant assignees in the relevant OWNERS file.
+- To encourage fast adoption of the security fix, the new patched release should
+  contain only security fixes.
+- To facilitate this goal, the Fix Lead will [create a patch release branch](https://github.com/kubernetes/release/blob/master/docs/branching.md#branching-from-a-tag)
+  in a private fork of the public kubernetes.git repository, branched from the
+  most recent released tag for the release being fixed.
+  - As a hypothetical example, if `v2.4.7` is the latest public patch release for
+   `v2.4`, a new branch will be created in the private fork called
+   `release-2.4.7`, branching at the `v2.4.7` tag. If `HEAD` were to be used
```
liggitt: I know there's always a bump commit that updates release version files (like base.go) to a x.y.beta.0... should we branch from that or the last actual tag?

contributors/devel/security-release-process.md
```diff
@@ -82,14 +112,47 @@ The communication to users should be actionable. They should know when to block
...
+    create a draft release on the public GitHub repo.
+- As soon as we are ready to make the fix known publicly (TODO: when?),
+  the Release Manager will use tooling to
+  - Publish results.
+    - Upload binary artifacts to GCS, gcr.io, and GitHub. (TODO: are these
+      prebuilt from the previous step?)
+    - Update the `CHANGELOG.md` in the **public** `master` branch
+    - Create a new release on the **public** kubernetes GitHub
+    - Send email announcing the release
+  - Merge the private patch release branch into the **public** release branch
```
liggitt: We need to do both of these things:
* push the released tag to the public repo (so the released sha is available via the given tag)
* merge the private patch release branch into the public release branch as described

contributors/devel/security-release-process.md
timstclair: I think it should be pretty fast. Ideally within a day. I think the fix lead (or someone they delegate to) should be responsible for opening a PR against master. In most cases it should be pretty straightforward, and the cherry-pick can be done manually if necessary.

contributors/devel/security-release-process.md
```diff
@@ -82,14 +112,47 @@ The communication to users should be actionable. They should know when to block
...
+    including the fixes and release tag.
+    - In our continuing example, this would merge the private `release-2.4.7`
+      branch into the public `release-2.4` branch, effectively advancing the
+      latter to `2.4.9-beta.0+`. Any changes in the `release-2.4` branch will
+      now ship with `v2.4.9`.
+  - If fixes were necessary to `master`, merge the private `master` branch into
```
liggitt: I'd open these as a normal PR with cherry-picks and tag as lgtm/approved (since it was already reviewed)

contributors/devel/security-release-process.md
```diff
@@ -82,14 +112,47 @@ The communication to users should be actionable. They should know when to block
...
+    the public `master` branch. No release will be built or published for
+    `master`.
+    - TODO: should this be a PR or a push?
 - The Fix Lead will request a CVE from [DWF](https://github.com/distributedweaknessfiling/DWF-Documentation) and include the CVSS and release details.
 - The Fix Lead will email kubernetes-{dev,users,announce,security-announce}@googlegroups.com now that everything is public announcing the new releases, the CVE number, the location of the binaries, and the relevant merged PRs to get wide distribution and user action. As much as possible this email should be actionable and include links how to apply the fix to users environments; this can include links to external distributor documentation.
 - The Fix Lead will remove the Fix Team from the private security repo.
+- The Fix Lead will (possibly?) delete the patch release branch from the private
```
liggitt: the tag holds the pointer to the released sha, I don't see a need to pollute with point release branches

contributors/devel/security-release-process.md
timstclair: Agreed. Patch releases are supposed to be pretty stable, which is why we only support the latest one and expect users to upgrade pretty quickly. I think this is reasonable.

contributors/devel/security-release-process.md
liggitt: could name the branch `release-2.4.x`, it's slightly confusing to work on what will become 2.4.8 in a release-2.4.7 branch

ixdy: The name of the branch is just matching what is described in the [release doc](https://github.com/kubernetes/release/blob/master/docs/branching.md#branching-from-a-tag), which is at least partially implemented in `anago` right now. (It doesn't support cross-repo yet.)

It's not great, since it's a little confusing, as you note. Using the target or source tag as the branch name also doesn't work well if the branch has to be rebased. We could instead name these something like issue-N, referring to the GitHub issue tracking the vulnerability?

contributors/devel/security-release-process.md
```diff
@@ -82,14 +115,48 @@ The communication to users should be actionable. They should know when to block
...
+    example, the `release-2.4.7` branch would produce a `v2.4.8` release.
+  - Verify that no public patch release has occurred since the security branch
+    was cut (in this example, make sure that `v2.4.8` has not been tagged
+    publicly).
+  - Perform a normal release build from the private release branch, staging
+    binary artifacts locally (or possibly in a private bucket).
+  - Push the new patch release tag to the private repo.
+  - The tooling should **not** upload any artifacts to public storage buckets or
+    create a draft release on the public GitHub repo.
+- As soon as we are ready to make the fix known publicly (TODO: when?),
```
philips: The when should be scheduled by the Fix Lead

contributors/devel/security-release-process.md
```diff
@@ -82,14 +115,48 @@ The communication to users should be actionable. They should know when to block
...
+  the Release Manager will use tooling to
+  - Publish results.
+    - Upload binary artifacts to GCS, gcr.io, and GitHub. (TODO: are these
+      prebuilt from the previous step?)
```
philips: they should be, yes.

contributors/devel/security-release-process.md
ixdy: updated, thanks

contributors/devel/security-release-process.md
ixdy: updated

ixdy:
updated the remaining two TODOs in https://github.com/kubernetes/community/pull/450/commits/086065014f73faacd63a6fa79d1d86f1324ff371.

ixdy:
I will update anago documentation once this process is actually implemented in code.

ixdy:
Any remaining feedback? I updated some of the details on merging security fixes back into the public release branch based on the recent v1.5.5 security release (which didn't follow this process).

contributors/devel/security-release-process.md
liggitt: After tagging an official release (like `v2.4.7`), anago bumps the version to a beta of the next point release and creates a tag (like `v2.4.8-beta.0`). I think we should branch from that point, and name the branch for the upcoming release (e.g. `release-2.4.8`).

I *think* if you run anago on that branch with `--official`, it would then compute `v2.4.8` as the official release.

liggitt: Can we make the branch by hand? We don't really want the announcements and changelog that accompany creating a branch, do we? Is there other magic anago is doing in the branch creation step? Do we need that magic?

contributors/devel/security-release-process.md
```diff
@@ -82,14 +115,54 @@ The communication to users should be actionable. They should know when to block
...
+    the release.
+  - The Release Manager will use release tooling (i.e. `anago`) to
+    - Compute the version for the new security patch release; in the previous
+      example, the `release-2.4.7` branch would produce a `v2.4.8` release.
+    - Verify that no public patch release has occurred since the security branch
+      was cut (in this example, make sure that `v2.4.8` has not been tagged
+      publicly).
+    - Perform a normal release build from the private release branch, staging
+      binary artifacts locally (or possibly in a private bucket).
+    - Push the new patch release tag to the private repo.
```
liggitt: anago currently also adds a version bump commit and tags the beta of the next version. I don't mind the version bump commit, but we'd want to tag the result of merging that into the public release branch (e.g. `v2.4.9-beta.0`)

contributors/devel/security-release-process.md
```diff
@@ -82,14 +115,54 @@ The communication to users should be actionable. They should know when to block
...
+  - Verify that nothing new has been pushed to the private release repository,
+    and that no new public patch release has occurred.
+      - If there are new commits on the private release branch, or if a new
+        public release has occurred, the previously build artifacts will be
+        discarded.
+  - Publish the release.
+    - Upload prebuilt binary artifacts (from previous step) to GCS, gcr.io, and
+      GitHub.
+    - Update the `CHANGELOG.md` in the **public** `master` branch
+    - Create a new release on the **public** kubernetes GitHub
```
liggitt: this means pushing the `v2.4.8` tag from the private repo to the public repo?

contributors/devel/security-release-process.md
```diff
@@ -82,14 +115,54 @@ The communication to users should be actionable. They should know when to block
...
+    - Send email announcing the release
+  - Merge the private patch release branch into the **public** release branch
```
liggitt: > including both the commits and release tag

the release tag would already have been pushed exactly, not merged, as part of creating the v2.4.8 release

I suppose the commit that bumps the version to v2.4.8 could be merged, though it's misleading if someone checks out that commit and builds it

> Advance the public release branch to the next beta.0 release, by updating tags and metadata as necessary.

If anago creates the bump commit to `v2.4.9-beta.0` as part of the `v2.4.8` release, do we get the commit for free by merging in the hotfix branch? All we need to do then is tag the result of that merge `v2.4.9-beta.0`, right?

cblecker:
/assign @philips @liggitt @jessfraz @tallclair @cjcullen
/unassign @idvoretskyi @calebamiles 
Unsure if this is still relevant. Assigning to PST members

jessfraz:
@ixdy if you want I can apply the comments from here into a new proposal if you don't have time just so we don't lose this

jessfraz:
oh shoot thanks will try and find time

PR #451, opened 2017-03-14
update nodeaffinity.md

zhangxiaoyu-zidif:
add a "," to make the sentence clear.
The sentence is really confused at first sight. Adding a "," may be helpful.


davidopp:
Thanks for your contribution, but this comma makes the sentence more confusing. I think it's just a weird subtlety of English. Basically, "Unlike A, when B C" means "B always implies C, and this is unlike the situation with A" whereas "Unlike A, when B, C" means "C is unlike the situation when A and B are both true" (it makes the "when B" modify A).

Sorry but I can't give a better explanation than that. :) It's just weird.

PR #454, opened 2017-03-14
[Deprecated] Proposal: Encrypt secrets (and potentially others) at rest

andrewsykim:
Deprecated in favour of https://github.com/kubernetes/community/pull/607

### DIFF
--- a/contributors/design-proposals/encryption.md
+++ b/contributors/design-proposals/encryption.md
@@ -1,0 +1,186 @@
+# Encryption
+
+## Abstract
+
+The scope of this proposal is to ensure that resources can be encrypted at the datastore layer. Encrypting data at REST and via third party vendors is a desired feature but outside the scope of this proposal. Encryption will be optional for any resource but we suspect it will be used for the Secret resource in most cases.
+
+Allowing sensitive data to be encrypted adheres to best practices as well as other requirements such as HIPAA.
+
+How encryption keys are delivered to the machine running the Kubernetes apiserver is of relevance - we assume that the encryption at rest pattern is loosely coupled to how those keys are delivered and secured on disk.
+
+In general, full disk encryption of the volumes storing etcd data is generally preferred - this proposal focuses on scenarios where additional protection is desired against malicious parties gaining read access to the etcd API or a running etcd instance without access to memory of the etcd process.
+
+## High level design
+Before a resource is written to etcd and before it is read, an encryption provider will take the plaintext data and encrypt it.  These providers will be able to be created and turned on depending on the users needs or requirements and will adhere to an encryption interface. This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.  It should be possible for the distribution of keys to the apiserver to be separated out (injected in by a higher level security process) or to be directly requested by the provider implementation. For the first iteration, a default provider will be developed and will run as part of the kube-apiserver.
+
+## Kubernetes Storage Changes
+Kubernetes requires that an update that does not change the serialized form of object not be persisted to etcd to prevent other components from seeing no-op updates.  This must be done within the Kubernetes storage interfaces - we will introduce a new API to the Kube storage layer that transforms the serialized object into the desired at-rest form and provides hints as to whether no-op updates should still persist (when key rotation is in effect).
+```go
+// ValueTransformer allows a string value to be transformed before being read from or written to the underlying store. The methods
+// must be able to undo the transformation caused by the other.
+type ValueTransformer interface {
+	// TransformFromStorage may transform the provided data from its underlying storage representation or return an error.
+	// Stale is true if the object on disk is stale and a write to etcd should be issued, even if the contents of the object
+	// have not changed.
+	TransformFromStorage([]byte) (data []byte, stale bool, err error)
+	// TransformToStorage may transform the provided data into the appropriate form in storage or return an error.
+	TransformToStorage([]byte) (data []byte, err error)
+}
+``` 
+
+When the storage layer of Kubernetes is initialized for some resource`, an implementation of this interface that manages encryption will be passed down.  Other resources can use a no-op provider by default.
+
+## Encryption Provider
+An encryption provider implements the ValueTransformer interface. Out of the box this proposal will implement encryption using a standard AES-GCM performing AEAD, using the standard Go library for AES-GCM, with the key configuration provided at process startup.
+
+Other encryption or key distribution implementations are possible such as AWS KMS, Google Cloud KMS and Hashicorp Vault just for example depending on the environment the cluster is deployed into as well as its capabilities. 
+
+The encryption provider implementation should work the same where a set of flags will be added to the API server specifying which (if any) provider is requested, which keys should be used for decryption, and which key is the primary encryption key. Each encryption provider will have a unique string identifier to ensure versioning of contents on disk and to allow future schemes to be replaced.
+
+It should be possible to separate the mechanism by which we encrypt the data in etcd from the mechanism whereby the keys to perform that mechanism are loaded onto the appropriate master.  The provider MAY perform dynamic retrieval of keys from a hardware security module, retrieve a set of keys from a cloud-native security provider, or prompt the user for a password to decrypt the encryption keys from rest on disk, or some combination of the above.
+
+During encryption, only a single provider is required.  During decryption, multiple providers or keys may be in use (when migrating from an older version of a provider, or when rotating keys), and thus the ValueTransformer implementation must be able to delegate to the appropriate provider.
+
+For the near term, we identify a single encryption provider out of the box:
+
+### AES-GCM provider
+The simplest possible provider is an AES-GCM encrypter/decrypter using AEAD, where we create a unique nonce on each new write to etcd, use that as the IV for AES-GCM of the value (the JSON or protobuf data) along with a set of authenticated data to create the ciphertext, and then on decryption use the nonce and the authenticated data to decode.  The keys come from configuration on the local disk (potentially decrypted at startup time using a stronger password protected key, or in the future from an alternative source).
+
+The provider will be assigned a versioned identifier to uniquely pair the implementation with the data at rest, such as “k8s-aes-gcm-v1”.  Any implementation that attempts to decode data associated with this provider id must follow a known structure and apply a specific algorithm.
+
+The provider would take a set of keys and unique key identifiers from the command line, with the key values stored on disk.  One key is identified as the write key, all others are used to decrypt data from previous keys.  Keys must be rotated more often than every 2^32 writes
+
+The provider should use the recommended Go defaults for all crypto settings unless otherwise noted.  We should use AES-256 keys (32 bytes).
+
+Process for transforming a value (object encoded as JSON or protobuf) to and from stable storage will look like the following:
+
+Layout as written to etcd2 (json safe string only):
+``` 
+NONCE := read(/dev/urandom)
+PLAIN_TEXT := <VALUE>
+AUTHENTICATED_DATA := ETCD_KEY
+CIPHER_TEXT := aes_gcm_encrypt(KEY, IV:NONCE, PLAIN_TEXT, A:AUTHENTICATED_DATA)
+BASE64_DATA := base64(<NONCE><CIPHER_TEXT>)
+STORED_DATA := <PROVIDER>:<KEY_ID>:<BASE64_DATA>
+```
+
+Layout as written to etcd3 (bytes):
+```
+NONCE := read(/dev/urandom)
+PLAIN_TEXT := <VALUE>
+AUTHENTICATED_DATA := ETCD_KEY
+CIPHER_TEXT := aes_gcm_encrypt(KEY, IV:NONCE, PLAIN_TEXT, A:AUTHENTICATED_DATA)
+STORED_DATA := <PROVIDER_ID>:<KEY_ID>:<NONCE><AUTHENTICATED_DATA><CIPHER_TEXT>
+```
+
+Pseudo-code for encrypt (golang):
+```go
+block := aes.NewCipher(primaryKeyString)
+aead := cipher.NewGCM(c.block)
+keyId := primaryKeyId
+
+// for each write
+nonce := make([]byte, block.BlockSize())
+io.ReadFull(crypto_rand.Reader, nonce)
+authenticatedData := ETCD_KEY
+cipherText := aead.Seal(nil, nonce, value, authenticatedData)
+storedData := providerId + keyId + base64.Encode(nonce + authenticatedData + cipherText)
+```
+
+Pseudo-code for decrypt (golang):
+```go
+// for each read
+providerId, keyId, base64Encoded := // slice provider and key from value
+// ensure this provider is the one handling providerId
+
+aead := // lookup an aead instance for keyId or error
+bytes := base64.Decode(base64Encoded)
+nonce, authenticatedData, cipherText := // slice from bytes
+out, err := aead.Open(nil, nonce, cipherText, authenticatedData)
+```
+
+## Key Rotation 
+
+In order to take an API driven approach for key rotation, new API objects will be defined:
+* Key Encryption Key (KEK) - key used to unlock the Data Encryption Key. Stored on API server nodes.
+* Data Encryption Key (DEK) - long-lived secret encrypted with a KEK. Stored in etcd encrypted. Unencrypted in-memory in API servers.
+* KEK Slot - to support KEK rotation there will be an ordered list of KEKs stored in the KEK DB. The current active KEK slot number, is stored in etcd for consistency.
+* KEK DB - a file with N KEKs in a JSON list. KEK[0], by definition, is null.
+
+```go
+type DataEncryptionKey struct {
+    ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Value string    // Encrypted
+}
+```
+
+```go
+type KeyEncryptionKeySlot struct {
+    ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Slot int
+}
+```
+
+```go
+type KeyEncryptionKeyDatabase struct {
+    metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Keys []string 
+}
+```
+
+To enable encryption a user will issue a PUT to an endpoint such as `/rotate`. If this is the first time this API has ever been called the API server will generate a key (unecrypted DEK), encrypt it with the KEK in slot 1, and encrypt all secrets with the DEK. If this is the second + N time the API has been called the API server will encrypt the DEK with the KEK in slot N+1 and do a compare-and-swap on the DEK stored in etcd. If this API is called and a KEK doesn't exist in the KEK DB at slot N+1 a 400 is returned. `kubectl` will have support for calling the rotate API endpoint eventually. To account for failure scenarios during key rotation, the old and new DEK will be stored in etcd during the rotation.  
+
+## Master Configuration
+In order to enable encryption, a user must first create a KEK DB file and tell the API server to use it with --key-encryption-key-db-path=/path/to/kekdb/file. The file will be a simple YAML file that lists all of the keys:
+
+```yaml
+kind: KeyEncryptionKeyDatabase
+version: v1
+keys:
+ - foo
+ - bar
+ - baz
+```
+
+The user will also need to specify the encryption provider and the resources to encrypt as follows:
+```yaml
+--key-encryption-key-db-path=/path/to/key-encryption-key/db
+--encryption-provider=default
+--encrypt-resource=v1/Secrets
+--encrypt-resource=v1/ConfigMap
+```
+
+As mentioned previously, the first pass at encryption will only support the default encryption provider.
+
+## Additional Considerations
+
+### Other encryption providers
+
+It should be easy to introduce new variations on the basic provider - such as those that require a password via STDIN on startup to decrypt the keys on disk, or wish to exchange a temporary token on disk for the actual keys from a third party security server.
+
+We expect that the provider implementation may therefore be layered or composed and that as much as possible providers should be separated behind interfaces.
+
+### Potential providers
+
+* A local HSM implementation that retrieves the keys from the secure enclave prior to reusing the AES-GCM implementation (initialization of keys only)
+* Exchanging a local temporary token for the actual decryption tokens from a networked secret vault
+* Decrypting the AES-256 keys from disk using asymmetric encryption combined with a user input password
+
+### Backwards Compatibility 
+Once a user encrypts any resource in etcd, they are locked to that Kubernetes version and higher unless they choose to manually decrypt that resource in etcd. This will be discouraged. It will be highly recommended that users discern if their Kubernetes cluster is on a stable version before enabling encryption.  
+
+### Performance
+
+Introducing even a relatively well tuned AES-GCM implementation is likely to have performance implications for Kubernetes.  Fortunately, existing optimizations occur above the storage layer and so the highest penalty will be incurred on writes when secrets are created or updated.  In multi-tenant Kube clusters secrets tend to have the highest load factor (there are 20-40 resources types per namespace, but most resources only have 1 instance where secrets might have 3-9 instances across 10k namespaces).  Writes are uncommon, creates usually happen only when a namespace is created, and reads are somewhat common. 
+
+### Actionable Items / Milestones
+
+* [p0] Add ValueTransformer to storage 
+* [p0] Create a default implementation of AES-GCM interface 
+* [p0] Add encryption flags on kube-apiserver and key rotation API
+* [p1] Add kubectl command to call /rotate endpoint
+* [p1] Audit of default implementation for safety and security 
+* [p2] E2E and performance testing 
+* [p2] Documentation and users guide
+* [p2] Read cache layer if encrypting/decrypting Secrets adds too much load on kube-apiserver
+
--- a/contributors/design-proposals/encryption.md
+++ b/contributors/design-proposals/encryption.md
@@ -2,20 +2,20 @@
 
 ## Abstract
 
-The scope of this proposal is to ensure that resources can be encrypted at the datastore layer. Encrypting data at REST and via third party vendors is a desired feature but outside the scope of this proposal. Encryption will be optional for any resource but we suspect it will be used for the Secret resource in most cases.
+The scope of this proposal is to ensure that resources can be encrypted at the datastore layer. Encrypting data over the network and via third party vendors is a desired feature but outside the scope of this proposal. There is future work to be done to enable end to end encryption, until then clients of the kubernetes API should be using TLS connections. Encryption will be optional for any resource but we suspect it will be used for the Secret resource in most cases.
 
 Allowing sensitive data to be encrypted adheres to best practices as well as other requirements such as HIPAA.
 
 How encryption keys are delivered to the machine running the Kubernetes apiserver is of relevance - we assume that the encryption at rest pattern is loosely coupled to how those keys are delivered and secured on disk.
 
-In general, full disk encryption of the volumes storing etcd data is generally preferred - this proposal focuses on scenarios where additional protection is desired against malicious parties gaining read access to the etcd API or a running etcd instance without access to memory of the etcd process.
+In general, full disk encryption of the volumes storing etcd data is preferred - this proposal focuses on scenarios where additional protection is desired against malicious parties gaining read access to the etcd API or its backups or a running etcd instance without access to memory of the etcd process.
 
 ## High level design
 Before a resource is written to etcd and before it is read, an encryption provider will take the plaintext data and encrypt it.
 These providers will be able to be created and turned on depending on the users needs or requirements and will adhere to an encryption interface.
 This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.
 It should be possible for the distribution of keys to the apiserver to be separated out (injected in by a higher level security process) or to be directly
-requested by the provider implementation. For the first iteration, a default provider will be developed and will run as part of the kube-apiserver.
+requested by the provider implementation. For the first iteration, a default provider that handles encryption in-process using a locally stored key will be developed.
 
 ## Kubernetes Storage Changes
 Kubernetes requires that an update that does not change the serialized form of object not be persisted to etcd to prevent other components from seeing no-op updates.
@@ -140,7 +140,7 @@ type KeyEncryptionKeyDatabase struct {
 
 To enable encryption a user will issue a PUT to an endpoint such as `/rotate`. If this is the first time this API has ever been called the API server
 will generate a key (unecrypted DEK), encrypt it with the KEK in slot 1, and encrypt all secrets with the DEK. If this is the second + N time the API
-has been called the API server will encrypt the DEK with the KEK in slot N+1 and do a compare-and-swap on the DEK stored in etcd.
+has been called the API server will encrypt the DEK with the KEK in slot N+1 and do a compare-and-swap on the DEK stored in etcd. There will be one DEK per database.
 If this API is called and a KEK doesn't exist in the KEK DB at slot N+1 a 400 is returned. `kubectl` will have support for calling the
 rotate API endpoint eventually. To account for failure scenarios during key rotation, the old and new DEK will be stored in etcd during the rotation.
 
--- a/contributors/design-proposals/encryption.md
+++ b/contributors/design-proposals/encryption.md
@@ -11,10 +11,16 @@ How encryption keys are delivered to the machine running the Kubernetes apiserve
 In general, full disk encryption of the volumes storing etcd data is generally preferred - this proposal focuses on scenarios where additional protection is desired against malicious parties gaining read access to the etcd API or a running etcd instance without access to memory of the etcd process.
 
 ## High level design
-Before a resource is written to etcd and before it is read, an encryption provider will take the plaintext data and encrypt it.  These providers will be able to be created and turned on depending on the users needs or requirements and will adhere to an encryption interface. This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.  It should be possible for the distribution of keys to the apiserver to be separated out (injected in by a higher level security process) or to be directly requested by the provider implementation. For the first iteration, a default provider will be developed and will run as part of the kube-apiserver.
+Before a resource is written to etcd and before it is read, an encryption provider will take the plaintext data and encrypt it.
+These providers will be able to be created and turned on depending on the users needs or requirements and will adhere to an encryption interface.
+This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.
+It should be possible for the distribution of keys to the apiserver to be separated out (injected in by a higher level security process) or to be directly
+requested by the provider implementation. For the first iteration, a default provider will be developed and will run as part of the kube-apiserver.
 
 ## Kubernetes Storage Changes
-Kubernetes requires that an update that does not change the serialized form of object not be persisted to etcd to prevent other components from seeing no-op updates.  This must be done within the Kubernetes storage interfaces - we will introduce a new API to the Kube storage layer that transforms the serialized object into the desired at-rest form and provides hints as to whether no-op updates should still persist (when key rotation is in effect).
+Kubernetes requires that an update that does not change the serialized form of object not be persisted to etcd to prevent other components from seeing no-op updates.
+This must be done within the Kubernetes storage interfaces - we will introduce a new API to the Kube storage layer that transforms the serialized object
+into the desired at-rest form and provides hints as to whether no-op updates should still persist (when key rotation is in effect).
 ```go
 // ValueTransformer allows a string value to be transformed before being read from or written to the underlying store. The methods
 // must be able to undo the transformation caused by the other.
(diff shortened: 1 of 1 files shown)


smarterclayton:
Nit: break lines at 100 or 120 char to make line reviewing easier.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+
+As mentioned previously, the first pass at encryption will only support the default encryption provider.
+
+## Additional Considerations
+
+### Other encryption providers
+
+It should be easy to introduce new variations on the basic provider - such as those that require a password via STDIN on startup to decrypt the keys on disk, or wish to exchange a temporary token on disk for the actual keys from a third party security server.
+
+We expect that the provider implementation may therefore be layered or composed and that as much as possible providers should be separated behind interfaces.
```
jcbsmpsn: What do you mean layered or composed? Like, multiple encryption providers at the same time?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
+# Encryption
```
destijl: Hi @andrewsykim, this is an interesting proposal. I'm doing some work to figure out the direction for secrets on the google side this quarter, it would be great to discuss this over VC with @smarterclayton and @jcbsmpsn.

I'm particularly interested in the use cases driving building this into kubernetes vs. taking advantage of the various external stores. Especially given there's a key management system for every cloud, and ones like vault that are already brokers for many different authn/z types and storage backends. If we could easily authenticate to a store that already handles key storage, encryption, rotation, and logging, would we still build it into kubernetes?

andrewsykim: You raise some good questions, I do believe there's been some talk around this already in google docs. 

It's true that most clouds do support key management systems out of the box, hence why we want the encryption interface to be pluggable for those use cases. @smarterclayton may have stronger/better reasonings than I do on this but I had thought we want to support a builtin solution for this because:

1) We need to support this for bare metal setups
2) Using third party vendors like Vault would require support for the storage backends it depends on. 
3) Using third party tools like Vault means any pods that require Secrets heavily depend on the availability of those vendors. This adds operational overhead for cluster admins.

For 2) and 3) I can see Kubernetes supporting a highly available cluster add-on for Vault that would work in most setups (similar to kube-dns, not required but highly recommended and a highly available setup is provided by the kubernetes community). 

For 2) etcd is an optional storage backend for Vault so we may not have to worry about adding an additional storage backend, but there are some issues we should address if we go down that path (e.g. Vault putting too much load on etcd servers will make master server unavailable). 

Would love to discuss this with you in more details, feel free to ping me on kubernetes slack to setup a time :)

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Abstract
+
+The scope of this proposal is to ensure that resources can be encrypted at the datastore layer. Encrypting data at REST and via third party vendors is a desired feature but outside the scope of this proposal. Encryption will be optional for any resource but we suspect it will be used for the Secret resource in most cases.
```
andrewsykim: Yes, just the implementation is outside the scope. When we develop the built in provider, it should be behind a pluggable interface. 

I think my understanding of "encryption at rest" may be wrong here. My understanding of "encryption at rest" is that HTTP calls between a client and the server will have secrets encrypted. This requires a lot more work since the `kubelet` now needs to have the encryption keys as well. For this proposal we meant that the secret stored in etcd will be encrypted. I'll make sure to clarify that.

jcbsmpsn: I think you can say encryption over the network is out of scope. API users should be using TLS connections. The kubelet can do that and there is work in flight to improve it.

contributors/design-proposals/encryption.md
destijl: Generally I think the direction should be to make it as easy as possible to use a pluggable external secret store, and kubernetes should avoid re-implementing features, and accruing maintenance cost, of those stores. 

Adding encryption and associated key rotation as described here is a step down this re-implementation path. Probably followed by very detailed audit logging, followed by fancier authn/authz features. We need to draw a line under how far we want to travel down this path. If we want to include this, we should be explicit that it's as far as we're going re-inventing the wheel.

There's some danger that adding encryption here will tip some people into using k8s secrets when really the risk profile of what they are storing means they should be moving to an external store that has stronger audit trail guarantees and various other security features. Larger organizations will probably be motivated to use external stores regardless, because they are already storing other secrets there and just want to centralize on one system.

Re plugging in other stores, I think the right way to do this is to give pods (and perhaps nodes) identities they can use to talk to the stores directly, but this is out of scope for this proposal.

Re HA: Even if you are air-gapped and running your own vault installation you can make it HA with some work. The kubernetes piece of availability might be providing some flexibility in allowing nodes to access and cache HA secrets vs. the normal least-privilege path where only the pod can retrieve the secret. But this is also out of scope here.

I'm definitely still interested in meeting to talk about this.

smarterclayton: Secrets are so fundamental to applications, and the diversity of key stores so broad, that we need to be able to do two things:

1. Store secrets and help distribute them to pods, in a basic and secure fashion
2. Enable deep integration with sophisticated key store, management, and audit solutions

So those are the 80% and the 20%.  The 80% has to work - the 20% has to be correct.  It has to be possible to transition from 1 to 2.  The items on the secrets roadmap roughly talk about enabling the remainder of the 80% and the transition to the 20%.

contributors/design-proposals/encryption.md
smarterclayton: Yes, we have an end to end encrypted communication story today.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+Allowing sensitive data to be encrypted adheres to best practices as well as other requirements such as HIPAA.
+
+How encryption keys are delivered to the machine running the Kubernetes apiserver is of relevance - we assume that the encryption at rest pattern is loosely coupled to how those keys are delivered and secured on disk.
+
+In general, full disk encryption of the volumes storing etcd data is generally preferred - this proposal focuses on scenarios where additional protection is desired against malicious parties gaining read access to the etcd API or a running etcd instance without access to memory of the etcd process.
+
+## High level design
+Before a resource is written to etcd and before it is read, an encryption provider will take the plaintext data and encrypt it.
+These providers will be able to be created and turned on depending on the users needs or requirements and will adhere to an encryption interface.
+This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.
```
smarterclayton: Possibly not API mechanism.  At a minimum all at rest data stored with some encryption patterns MUST support key rotation.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+
+During encryption, only a single provider is required.  During decryption, multiple providers or keys may be in use (when migrating from an older version of a provider, or when rotating keys), and thus the ValueTransformer implementation must be able to delegate to the appropriate provider.
+
+For the near term, we identify a single encryption provider out of the box:
+
+### AES-GCM provider
+The simplest possible provider is an AES-GCM encrypter/decrypter using AEAD, where we create a unique nonce on each new write to etcd, use that as the IV for AES-GCM of the value (the JSON or protobuf data) along with a set of authenticated data to create the ciphertext, and then on decryption use the nonce and the authenticated data to decode.  The keys come from configuration on the local disk (potentially decrypted at startup time using a stronger password protected key, or in the future from an alternative source).
```
smarterclayton: Authenticated data is part of the AEAD definition and ensures that someone with access to ciphertext can't trick the server into using the value in a different context.  https://en.wikipedia.org/wiki/Authenticated_encryption

The AD we are proposing is the etcd key - that means an attacker with access to etcd can't take a cipher text from one key, put it under another key (under the attacker's control) and then see the contents in the API (where the apiserver decodes it for that user).

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+providerId, keyId, base64Encoded := // slice provider and key from value
+
+// ensure this provider is the one handling providerId
+aead := // lookup an aead instance for keyId or error
+bytes := base64.Decode(base64Encoded)
+nonce, authenticatedData, cipherText := // slice from bytes
+out, err := aead.Open(nil, nonce, cipherText, authenticatedData)
+```
+
+## Key Generation & Rotation
```
smarterclayton: I think this section is somewhat optional to accomplish the primary goal.

1. store details encrypted
2. enable rotation of keys
3. make keys easy to manage

I don't think 3 is required to accomplish 1-2 - I'd prefer to see the proposal reflect that (focus on the requirements for rotation, then talk about future work to reduce the complexity).  Rotation requires a set of core patterns (behavior in the storage) and then should have clear layers for key management.  Key management is not core.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+We expect that the provider implementation may therefore be layered or composed and that as much as possible providers should be separated behind interfaces.
+
+### Potential providers
+
+* A local HSM implementation that retrieves the keys from the secure enclave prior to reusing the AES-GCM implementation (initialization of keys only)
+* Exchanging a local temporary token for the actual decryption tokens from a networked secret vault
+* Decrypting the AES-256 keys from disk using asymmetric encryption combined with a user input password
+
+### Backwards Compatibility
+Once a user encrypts any resource in etcd, they are locked to that Kubernetes version and higher unless they choose to manually decrypt that resource in etcd. This will be discouraged. It will be highly recommended that users discern if their Kubernetes cluster is on a stable version before enabling encryption.
```
smarterclayton: I think this requires section on these two topics:

1. How rotation in storage happens
2. How encryption is enabled

Ideally 1 and 2 should follow the same pattern (adding a new key is a lot like turning on encryption at first).

contributors/design-proposals/encryption.md
smarterclayton: So:

1. How to do rotation
2. How to do so with keys passed in externally
3. How to do so with a managed set of internal keys
4. How to do so with an external KSM correctly and safely.

I'm most concerned with 1-2 and the implications of 4.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+```go
+type KeyEncryptionKeyDatabase struct {
+    metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Keys []string
+}
+```
+
+To enable encryption a user will issue a PUT to an endpoint such as `/rotate`. If this is the first time this API has ever been called the API server
+will generate a key (unecrypted DEK), encrypt it with the KEK in slot 1, and encrypt all secrets with the DEK. If this is the second + N time the API
+has been called the API server will encrypt the DEK with the KEK in slot N+1 and do a compare-and-swap on the DEK stored in etcd.
```
philips: One per database was my initial thinking.

philips:
Overall I think this proposal looks good. Although, I did review an early draft and help with the rotation/rekey stuff :)

Would love another set of eyes from @ericchiang or @lpabon

contributors/design-proposals/encryption.md
andrewsykim: Multiple encryption providers at the same time only for migration purposes.

contributors/design-proposals/encryption.md
andrewsykim: Would taking an API driven approach for key rotation be part of "3. make keys easy to manage" or "2. enable rotation of keys"?

andrewsykim: i.e. are we okay with users having to manually restart api servers after adding a new KEK?

andrewsykim: My impression from the last meeting was that we're okay with users having to restart servers themselves to reduce the complexity of encryption. Just making sure I understood correctly.

smarterclayton: I think in stage 1, restarting services themselves is the best approach, followed by layering more powerful options down the road.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+If this API is called and a KEK doesn't exist in the KEK DB at slot N+1 a 400 is returned. `kubectl` will have support for calling the
+rotate API endpoint eventually. To account for failure scenarios during key rotation, the old and new DEK will be stored in etcd during the rotation.
+
+## Master Configuration
+In order to enable encryption, a user must first create a KEK DB file and tell the API server to use it with --key-encryption-key-db-path=/path/to/kekdb/file. The file will be a simple YAML file that lists all of the keys:
```
andrewsykim: Any objections to having the encryption provider config file also contain the resources to encrypt?

andrewsykim: How's something like this?
```yaml
kind: EncryptionProvider
version: crypto/v1alpha
metadata:
  name: kube-secrets-encryption
spec:
  provider: default
  keys:
   - foo
   - bar
   - baz
  resources:
   - v1/Secrets
```

jcbsmpsn: I'm not sure on this one. It might make sense for the resources to encrypt to be outside the encryption provider. Would putting the resources inside the encryption provider mean the encryption provider would have to be configured for all resources so that it could decide if it should or shouldn't encrypt?

contributors/design-proposals/encryption.md
jcbsmpsn: So are you envisioning a chain, something like:

    REST Request -> ... -> Encryption Provider 1 -> Encryption Provider 2 -> etcd

Or more like something (where only one of the encryption providers is invoked):

    REST Request -> ... -> (Encryption Provider 1 | Encryption Provider 2) -> etcd

contributors/design-proposals/encryption.md
jcbsmpsn: Restating so I understand, the thinking is, key rotation will be implemented, probably be accomplished by modifying config files and restarting the API server to pick up changes. Making it easy/pleasant is future work.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+## Key Generation & Rotation
+
+To account for easy key rotation, an additional layer of abstraction is introduced where keys used for resource encryption is encrypted by another set of keys (Key Encryption Keys aka KEK).
+The key used for encrypting kubernetes resources (Data Encryption Keys, aka DEK) are generated by the apiserver and stored in etcd once encrypted with one of the KEKs.
+KEKs should never be stored in etcd and in most cases be on the same volume as the apiserver. This provides consistency for the apiserver but still protects against compromises to the etcd volume.
+
+In order to take an API driven approach for key rotation, new API objects will be defined:
+* Key Encryption Key (KEK) - key used to unlock the Data Encryption Key. Stored on API server nodes.
```
jcbsmpsn: Instead of 'Stored on API server nodes' how about 'Stored on API server nodes or remotely depending on the encryption provider implementation.'

I think for HSM and Google/Amazon KSM encryption providers, the HSM/KSM will hold the KEK.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+```go
+type KeyEncryptionKeyDatabase struct {
+    metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Keys []string
+}
+```
+
+To enable encryption a user will issue a PUT to an endpoint such as `/rotate`. If this is the first time this API has ever been called the API server
+will generate a key (unecrypted DEK), encrypt it with the KEK in slot 1, and encrypt all secrets with the DEK. If this is the second + N time the API
+has been called the API server will encrypt the DEK with the KEK in slot N+1 and do a compare-and-swap on the DEK stored in etcd. There will be one DEK per database.
```
jcbsmpsn: Perhaps this is already what you are thinking and I've just misunderstood, but can you outline the procedure for rotating the DEK as well, not just the procedure for re-encrypting the same DEK with the new KEK?

Or is that what you mean in the next paragraph where it talks about the new and the old DEK. That there will be multiple DEKs stored in etcd (and I don't mean the same DEK encrypted with different KEKs).

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+Once a user encrypts any resource in etcd, they are locked to that Kubernetes version and higher unless they choose to manually decrypt that resource in etcd. This will be discouraged. It will be highly recommended that users discern if their Kubernetes cluster is on a stable version before enabling encryption.
+
+### Performance
+
+Introducing even a relatively well tuned AES-GCM implementation is likely to have performance implications for Kubernetes.
+Fortunately, existing optimizations occur above the storage layer and so the highest penalty will be incurred on writes when secrets are created or updated.
+In multi-tenant Kube clusters secrets tend to have the highest load factor (there are 20-40 resources types per namespace, but most resources
+only have 1 instance where secrets might have 3-9 instances across 10k namespaces).  Writes are uncommon, creates usually happen
+only when a namespace is created, and reads are somewhat common.
+
```
jcbsmpsn: In the meeting @smarterclayton indicated that there are two secrets access patterns that are most common. I think it was getting and listing. Can we include that so that alternate encryption provider implementers can see why keeping most encryption/decryption operations in process in the API server is preferable?

contributors/design-proposals/encryption.md
mheese: 1. does this mean that every secret could potentially be encrypted with a different provider?
2. with regards to rotation and encryption mechanism ... if this is not triggered from the API, is it at least exposed on the object? (like e.g. this key has been in use for 35d, it is a 128bit AES key, and is encrypted using AES-GCM)
3. how do you ensure that key rotation is indeed implemented / supported?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.
+It should be possible for the distribution of keys to the apiserver to be separated out (injected in by a higher level security process) or to be directly
+requested by the provider implementation. For the first iteration, a default provider that handles encryption in-process using a locally stored key will be developed.
+
+## Kubernetes Storage Changes
+Kubernetes requires that an update that does not change the serialized form of object not be persisted to etcd to prevent other components from seeing no-op updates.
+This must be done within the Kubernetes storage interfaces - we will introduce a new API to the Kube storage layer that transforms the serialized object
+into the desired at-rest form and provides hints as to whether no-op updates should still persist (when key rotation is in effect).
+```go
+// ValueTransformer allows a string value to be transformed before being read from or written to the underlying store. The methods
```
mheese: I guess you mean `byte slice value` instead of `string value`?

contributors/design-proposals/encryption.md
mheese: that sounds like sensible and secure defaults for a "default" provider 👍

contributors/design-proposals/encryption.md
mheese: Is this process meant for rotating the KEK only? Or also for DEKs?
I agree that for rotating KEKs, a restart first is fine.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+bytes := base64.Decode(base64Encoded)
+nonce, authenticatedData, cipherText := // slice from bytes
+out, err := aead.Open(nil, nonce, cipherText, authenticatedData)
+```
+
+## Key Generation & Rotation
+
+To account for easy key rotation, an additional layer of abstraction is introduced where keys used for resource encryption is encrypted by another set of keys (Key Encryption Keys aka KEK).
+The key used for encrypting kubernetes resources (Data Encryption Keys, aka DEK) are generated by the apiserver and stored in etcd once encrypted with one of the KEKs.
+KEKs should never be stored in etcd and in most cases be on the same volume as the apiserver. This provides consistency for the apiserver but still protects against compromises to the etcd volume.
```
mheese: I hope this section would apply for the "default" provider only because ...

Try to imagine the implementation of an HSM provider:
1. When you try to integrate with an HSM (e.g. with PKCS#11), the preferred way of encryption/decryption would be that you generate the DEK on the HSM, and also never export the DEK from the HSM.
2. Key rotation would mean, you generate a new key on the HSM and tell your API that you now have this new key
3. You would *not want* to let the DEKs being generated by the apiserver
4. There is only a need for a KEK if the DEKs are indeed generated by the apiserver and you basically only store the KEKs in your  HSM

Alternatively to support this approach an HSM provider implementation could look like:
- KEKs are generated and stored on the HSM
- kubelet startup loads them (extracts them - which means that it would need to get "wrapped" with the HSM (another KEK layer) and "unwrapped" in software again
- now this could decrypt the DEKs and basically use the default provider approach (golang AES-GCM) for doing the actual encryption/decryption work

contributors/design-proposals/encryption.md
smarterclayton: Yeah, originally was strings in etcd2.  Now byte slice.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+
+Other encryption or key distribution implementations are possible such as AWS KMS, Google Cloud KMS and Hashicorp Vault just for example depending on the environment the cluster is deployed into as well as its capabilities.
+
+Each encryption provider will have a unique string identifier to ensure versioning of contents on disk and to allow future schemes to be replaced.
+
+It should be possible to separate the mechanism by which we encrypt the data in etcd from the mechanism whereby the keys to perform that mechanism are loaded onto the appropriate master.  The provider MAY perform dynamic retrieval of keys from a hardware security module, retrieve a set of keys from a cloud-native security provider, or prompt the user for a password to decrypt the encryption keys from rest on disk, or some combination of the above.
+
+During encryption, only a single provider is required.  During decryption, multiple providers or keys may be in use (when migrating from an older version of a provider, or when rotating keys), and thus the ValueTransformer implementation must be able to delegate to the appropriate provider.
```
smarterclayton: Only when writing a new version of the resource.  The design of the API today is that we migrate at rest content by trigging a no-op PUT - if the encoded data is != the underlying data in etcd we would reencrypt.  In the write path, both the API layer (the plaintext) and the cipher process can signal that a write is required, so either content changes or a new configured encryption key can be used to ensure that a write is performed.

contributors/design-proposals/encryption.md
destijl: Implementation started here:
https://github.com/kubernetes/kubernetes/pull/41939

destijl: I know this question is coming late to the game, but was encrypting the whole etcd database ever considered? I think it would make attacks like this harder and delivers similar security benefits. Docker seems to do it at this level:
https://docs.docker.com/engine/swarm/swarm_manager_locking/

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+
+In general, full disk encryption of the volumes storing etcd data is generally preferred - this proposal focuses on scenarios where additional protection is desired against malicious parties gaining read access to the etcd API or a running etcd instance without access to memory of the etcd process.
+
+## High level design
+Before a resource is written to etcd and before it is read, an encryption provider will take the plaintext data and encrypt it.
+These providers will be able to be created and turned on depending on the users needs or requirements and will adhere to an encryption interface.
+This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.
+It should be possible for the distribution of keys to the apiserver to be separated out (injected in by a higher level security process) or to be directly
+requested by the provider implementation. For the first iteration, a default provider will be developed and will run as part of the kube-apiserver.
```
destijl: +1, lets resolve how the key is going to get delivered for the default implementation. Are we only supporting a file on disk? What do we do if it is missing or the wrong key? Should we support passing it in "manually" somehow? If so, will the cluster still work while we're waiting on the key?

contributors/design-proposals/encryption.md
stevesloka: I believe this came up before in discussion, but the overhead was thought to be too much to encrypt everything. 

We also discussed at one point storing the database on an encrypted volume, but that doesn't fit all deployment scenarios and breaks many setups.

contributors/design-proposals/encryption.md
stevesloka: I have a big interest in this as well dealing with secrets. Happy to join in on your discussions @andrewsykim @destijl @jcbsmpsn

destijl: Thanks, I think we're all in agreement to get this done and we only have a few weeks left before 1.7 code freeze.

contributors/design-proposals/encryption.md
lpabon: Quick question, if etcd had encryption-at-rest support, would this be necessary? Also, if this is a normal question asked by new people like me :-) , could we add a section to the document explaining why that direction was not taken?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+
+For the near term, we identify a single encryption provider out of the box:
+
+### AES-GCM provider
+The simplest possible provider is an AES-GCM encrypter/decrypter using AEAD, where we create a unique nonce on each new write to etcd, use that as the IV for AES-GCM of the value (the JSON or protobuf data) along with a set of authenticated data to create the ciphertext, and then on decryption use the nonce and the authenticated data to decode.  The keys come from configuration on the local disk (potentially decrypted at startup time using a stronger password protected key, or in the future from an alternative source).
+
+The provider will be assigned a versioned identifier to uniquely pair the implementation with the data at rest, such as “k8s-aes-gcm-v1”.  Any implementation that attempts to decode data associated with this provider id must follow a known structure and apply a specific algorithm.
+
+The provider would take a set of keys and unique key identifiers from the command line, with the key values stored on disk.  One key is identified as the write key, all others are used to decrypt data from previous keys.  Keys must be rotated more often than every 2^32 writes
```
lpabon: Could `from the command line` be expanded? Seems vague.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,204 @@
...
+// ensure this provider is the one handling providerId
+aead := // lookup an aead instance for keyId or error
+bytes := base64.Decode(base64Encoded)
+nonce, authenticatedData, cipherText := // slice from bytes
+out, err := aead.Open(nil, nonce, cipherText, authenticatedData)
+```
+
+## Key Generation & Rotation
+
+To account for easy key rotation, an additional layer of abstraction is introduced where keys used for resource encryption is encrypted by another set of keys (Key Encryption Keys aka KEK).
```
lpabon: This seems like a great idea for any application. Does it need to be part of Kubernetes? Can it be its own pkg and be included by Kubernetes?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+```go
+type KeyEncryptionKeyDatabase struct {
+    metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Keys []string
+}
+```
+
+To enable encryption a user will issue a PUT to an endpoint such as `/rotate`. If this is the first time this API has ever been called the API server
```
lpabon: This is probably one of the key usability definitions in this paper. Could this be expanded in its own section?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+will generate a key (unecrypted DEK), encrypt it with the KEK in slot 1, and encrypt all secrets with the DEK. If this is the second + N time the API
+has been called the API server will encrypt the DEK with the KEK in slot N+1 and do a compare-and-swap on the DEK stored in etcd. There will be one DEK per database.
+If this API is called and a KEK doesn't exist in the KEK DB at slot N+1 a 400 is returned. `kubectl` will have support for calling the
+rotate API endpoint eventually. To account for failure scenarios during key rotation, the old and new DEK will be stored in etcd during the rotation.
+
+## Master Configuration
+In order to enable encryption, a user must first create a KEK DB file and tell the API server to use it with --key-encryption-key-db-path=/path/to/kekdb/file. The file will be a simple YAML file that lists all of the keys:
```
lpabon: When I think `user` I think the user of Kubernetes. I think `user` here means `the administrator of Kubernetes`, right?

contributors/design-proposals/encryption.md
lpabon: I'm guessing the second, but this needs to be highlighted in the document.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+## Kubernetes Storage Changes
+Kubernetes requires that an update that does not change the serialized form of object not be persisted to etcd to prevent other components from seeing no-op updates.
+This must be done within the Kubernetes storage interfaces - we will introduce a new API to the Kube storage layer that transforms the serialized object
+into the desired at-rest form and provides hints as to whether no-op updates should still persist (when key rotation is in effect).
+```go
+// ValueTransformer allows a string value to be transformed before being read from or written to the underlying store. The methods
+// must be able to undo the transformation caused by the other.
+type ValueTransformer interface {
+	// TransformFromStorage may transform the provided data from its underlying storage representation or return an error.
+	// Stale is true if the object on disk is stale and a write to etcd should be issued, even if the contents of the object
```
smarterclayton: stale usually means "the data is encrypted with an older primary key"

contributors/design-proposals/encryption.md
NathanMcCauley: We haven't found it to be too much overhead in Swarm. Also greatly simplifies the model to just encrypt everything, as there are definitely other important items in the store that can get the benefit of the authenticity as well as encryption.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,186 @@
...
+
+The provider will be assigned a versioned identifier to uniquely pair the implementation with the data at rest, such as “k8s-aes-gcm-v1”.  Any implementation that attempts to decode data associated with this provider id must follow a known structure and apply a specific algorithm.
+
+The provider would take a set of keys and unique key identifiers from the command line, with the key values stored on disk.  One key is identified as the write key, all others are used to decrypt data from previous keys.  Keys must be rotated more often than every 2^32 writes
+
+The provider should use the recommended Go defaults for all crypto settings unless otherwise noted.  We should use AES-256 keys (32 bytes).
```
NathanMcCauley: Recommend using https://godoc.org/golang.org/x/crypto/nacl/secretbox

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,186 @@
...
+```yaml
+kind: KeyEncryptionKeyDatabase
+version: v1
+keys:
+ - foo
+ - bar
+ - baz
+```
+
+The user will also need to specify the encryption provider and the resources to encrypt as follows:
```
NathanMcCauley: This is overly complicated. k8s admins shouldn't have to choose which resources to encrypt. Either we should pick which resources should be encrypted or they should all be encrypted.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,186 @@
+# Encryption
+
+## Abstract
+
+The scope of this proposal is to ensure that resources can be encrypted at the datastore layer. Encrypting data at REST and via third party vendors is a desired feature but outside the scope of this proposal. Encryption will be optional for any resource but we suspect it will be used for the Secret resource in most cases.
+
+Allowing sensitive data to be encrypted adheres to best practices as well as other requirements such as HIPAA.
+
+How encryption keys are delivered to the machine running the Kubernetes apiserver is of relevance - we assume that the encryption at rest pattern is loosely coupled to how those keys are delivered and secured on disk.
```
NathanMcCauley: Could use a similar pattern to Swarm secrets for lock/unlock of the cluster: https://docs.docker.com/engine/swarm/swarm_manager_locking/ -- this pattern really helps for having an option to keep the key-encrypting keys protected.

NathanMcCauley: Building at least the UX for this in the first version allows for expansion to something like secret sharing / key splitting a la https://github.com/hashicorp/vault and https://github.com/cloudflare/redoctober

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
+# Encryption
+
+## Abstract
+
+The scope of this proposal is to ensure that resources can be encrypted at the datastore layer. Encrypting data over the network and via third party vendors is a desired feature but outside the scope of this proposal. There is future work to be done to enable end to end encryption, until then clients of the kubernetes API should be using TLS connections. Encryption will be optional for any resource but we suspect it will be used for the Secret resource in most cases.
```
diogomonica: I think we should really strive to have the secret resource come encrypted by default instead of being optional. IMHO "we suspect it will be used for the Secret resource in most cases" isn't good enough.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+
+Allowing sensitive data to be encrypted adheres to best practices as well as other requirements such as HIPAA.
+
+How encryption keys are delivered to the machine running the Kubernetes apiserver is of relevance - we assume that the encryption at rest pattern is loosely coupled to how those keys are delivered and secured on disk.
+
+In general, full disk encryption of the volumes storing etcd data is preferred - this proposal focuses on scenarios where additional protection is desired against malicious parties gaining read access to the etcd API or its backups or a running etcd instance without access to memory of the etcd process.
```
diogomonica: A quick comment here that the recommendation should also include encrypted swap (no swap is also common in production deployments).

contributors/design-proposals/encryption.md
diogomonica: Automated key rotation might be desirable but IMO not necessary. What we did in swarm was piggyback on transitions from an "unlocked" cluster to a "locked" cluster, rotating both the KEK and the DEK (and doing a snapshot w/ the new DEK). This allows users to force a full rotation of keys/re-encryption of data, as a byproduct of the operation that will need it the most (transitioning from a state where a KEK is on disk to a state where the KEK only lives in memory)

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+
+## High level design
+Before a resource is written to etcd and before it is read, an encryption provider will take the plaintext data and encrypt it.
+These providers will be able to be created and turned on depending on the users needs or requirements and will adhere to an encryption interface.
+This interface will provide the abstraction to allow various encryption mechanisms to be implemented, as well as for the method of encryption to be rotated over time.
+It should be possible for the distribution of keys to the apiserver to be separated out (injected in by a higher level security process) or to be directly
+requested by the provider implementation. For the first iteration, a default provider that handles encryption in-process using a locally stored key will be developed.
```
diogomonica: What are the plans on the default mode? locally stored key by default?

contributors/design-proposals/encryption.md
diogomonica: `Primary key` seems like an awkward name for this. Why primary?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+	TransformFromStorage([]byte) (data []byte, stale bool, err error)
+	// TransformToStorage may transform the provided data into the appropriate form in storage or return an error.
+	TransformToStorage([]byte) (data []byte, err error)
+}
+```
+
+When the storage layer of Kubernetes is initialized for some resource, an implementation of this interface that manages encryption will be passed down. Other resources can use a no-op provider by default.
+
+## Encryption Provider
+An encryption provider implements the ValueTransformer interface. Out of the box this proposal will implement encryption using a standard AES-GCM performing AEAD, using the standard Go library for AES-GCM, with the key configuration provided at process startup.
```
diogomonica: I recommend not using `AES-GCM`, and using `XSalsa20` and `Poly1305` instead. The software implementation of the latter in general CPUs is faster and general implementations of GCM are vulnerable to cache-timing attacks. Also the higher level interface of https://godoc.org/golang.org/x/crypto/nacl/secretbox is probably what we want to use. A lot less room for error with just `Open()` and `Seal()`

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+
+Other encryption or key distribution implementations are possible such as AWS KMS, Google Cloud KMS and Hashicorp Vault just for example depending on the environment the cluster is deployed into as well as its capabilities.
```
diogomonica: Not sure if we ever want to have a remote encryption API for every write due to performance reasons. The right thing to do here is probably to have the Data Encryption Key encrypted at rest by a Key Encrypting Key, and ask for the decryption at boot from the remote system (KMS, etc).

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,205 @@
...
+
+Each encryption provider will have a unique string identifier to ensure versioning of contents on disk and to allow future schemes to be replaced.
+
+It should be possible to separate the mechanism by which we encrypt the data in etcd from the mechanism whereby the keys to perform that mechanism are loaded onto the appropriate master.  The provider MAY perform dynamic retrieval of keys from a hardware security module, retrieve a set of keys from a cloud-native security provider, or prompt the user for a password to decrypt the encryption keys from rest on disk, or some combination of the above.
```
diogomonica: This is what swarm does. Either you call an API to unlock the swarm w/ the corresponding key, or the user provides it:

```
➜  ~ docker swarm update --autolock=true
Swarm updated.
To unlock a swarm manager after it restarts, run the `docker swarm unlock`
command and provide the following key:

    SWMKEY-1-G5+GnzpF6hlgsakXUjPV0Pf4XsXgBoa9CLEQmn2mpOE

Please remember to store this key in a password manager, since without it you
will not be able to restart the manager.
[ RESTART Docker ]
➜  ~ docker swarm unlock
Please enter unlock key:
```

contributors/design-proposals/encryption.md
diogomonica: Why `2^32`? And are you referring to the soft block limit or to the nonce limit? IIRC, nonce limit is dependent on the size of the payload being encrypted. 

If we're using 96bit IVs we should have guarantees with a higher nonce limit than `2^32`, but we should probably ping lvh to see what he things.

contributors/design-proposals/encryption.md
lpabon: ah, thanks

destijl:
Hi all, Andrew mentioned he didn't have time to devote to this PR and he was happy for us to take it over to move it forward. So I've split out this content into a new PR:
https://github.com/kubernetes/community/pull/607 

so we can get it to resolution. Andrew thanks for the help and your understanding here. If you could modify the description to mark this as deprecated and point to the new one that would be appreciated.

Thanks,
Greg

php-coder:
Let's close it.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/454)

PR #467, opened 2017-03-16
preserve order of list in strategicMergePatch

mengqiy:
This is kind of a small proposal of how to preserve the order of list in strategicMergePatch.

Old behavior: we never preserve the order for list with merge `patchStrategy`. Instead we sort them by the `mergeKey`.

Related issue: #40373

cc: @pwittrock @lavalamp

### DIFF
--- a/contributors/devel/api-conventions.md
+++ b/contributors/devel/api-conventions.md
@@ -595,6 +595,16 @@ Strategic Merge Patch also supports special operations as listed below.
 
 ### List Operations
 
+#### Specifying element ordering for a strategic merge patch on a list [Alpha]
+
+By default, patching a list using the Strategic Merge strategy will provide no guarantees to the final ordering of elements in the patched list.
+
+As of Kubernetes 1.7, Alpha support exists for explicitly specifying the ordering of elements in a during a Strategic Merge Patch.
+This can be done by providing a parallel list containing all of the patchMergeKeys of elements that need to be ordered.
+Any elements missing from the parallel list are guaranteed to appear after elements appearing in the parallel list.
+
+When patching Strategic Merge lists, kubectl apply sends a parallel list containing keys for each of the elements in the configuration file.
+
 #### List of Maps
 
 To override the container list to be strictly replaced, regardless of the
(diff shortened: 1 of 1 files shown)


ymqytw:
@lavalamp Ping

ymqytw:
Updated.

pwittrock:
@apelisse Could you give this a review?

contributors/devel/api-conventions.md
```diff
@@ -595,6 +595,16 @@ Strategic Merge Patch also supports special operations as listed below.
 
 ### List Operations
 
+#### Specifying element ordering for a strategic merge patch on a list [Alpha]
+
+By default, patching a list using the Strategic Merge strategy will provide no guarantees to the final ordering of elements in the patched list.
+
+As of Kubernetes 1.7, Alpha support exists for explicitly specifying the ordering of elements in a during a Strategic Merge Patch.
+This can be done by providing a parallel list containing all of the patchMergeKeys of elements that need to be ordered.
+Any elements missing from the parallel list are guaranteed to appear after elements appearing in the parallel list.
```
apelisse: Not sure if this is a relevant use-case, but what happens if the server adds many?

Original:
```yaml
- A
```

Server adds B to F:
```yaml
- A
- B
- C
- D
- E
- F
```

User applies:
```yaml
- A
- B
```

 We only guarantee that A and B will keep order?

liggitt: Or if the server reorders things in the meantime. Intent gets really muddled here. 

Original:
`
A
B
C
D
E
`

how does a client indicate it want to move A to the end, tolerating any server-side reordering of the other elements?

contributors/devel/api-conventions.md
```diff
@@ -595,6 +595,16 @@ Strategic Merge Patch also supports special operations as listed below.
 
 ### List Operations
 
+#### Specifying element ordering for a strategic merge patch on a list [Alpha]
```
lavalamp: It is super confusing to put this here. This section of the doc is about when the verb is "list" (i.e., GET on a collection). Please make this part of the strategic merge patch section. Also should we really be editing the doc before it's implemented? I guess I would have expected a proposal.

contributors/devel/api-conventions.md
```diff
@@ -595,6 +595,16 @@ Strategic Merge Patch also supports special operations as listed below.
...
+
+By default, patching a list using the Strategic Merge strategy will provide no guarantees to the final ordering of elements in the patched list.
+
+As of Kubernetes 1.7, Alpha support exists for explicitly specifying the ordering of elements in a during a Strategic Merge Patch.
+This can be done by providing a parallel list containing all of the patchMergeKeys of elements that need to be ordered.
```
lavalamp: I don't understand what this means. An example would be helpful. Something that explicitly states the syntax of the patch would be best.

contributors/devel/api-conventions.md
lavalamp: Wait, sorry, I am wrong.

lavalamp: I think we should specify strategic merge patch elsewhere. This is not the right doc to define it.

contributors/devel/api-conventions.md
```diff
@@ -620,18 +630,82 @@ containers:
...
+```
+
+After server applying the patch:
+```yaml
+env:
+  - name: ENV1
+    value: foo
+  - name: ENV2
+    value: bar
+  - name: ENV5
```
lavalamp: I am not sure that this is a useful feature. It is very complex and I'm not sure why it is necessary, or even why it is an improvement? I can see adding an "append" command.

contributors/devel/api-conventions.md
```diff
@@ -620,18 +630,82 @@ containers:
...
+  - name: ENV4
+    value: server-added
+```
+
 #### List of Primitives
 
 We have two patch strategies for lists of primitives: replace and merge.
 Replace is the default patch strategy, which will replace the whole list on update and it will preserve the order;
-while merge strategy works as an unordered set. In this section, we call a primitive list with merge strategy an unordered set.
+while merge strategy works as an ordered set. In this section, we call a primitive list with merge strategy an ordered set.
```
lavalamp: Will it still be possible to treat these items as an unordered set?

lavalamp:
Apologies for delay in looking at this. I am lost in a sea of github notifications, next time chat me up on slack (or internally).

Is there a reason why https://github.com/kubernetes/kubernetes/issues/40373 can't just use the replace strategy? We don't have random things mutating the environment variables, I don't think.

contributors/devel/api-conventions.md
```diff
@@ -620,18 +630,82 @@ containers:
...
 The patch strategy is defined in the go struct tag of the API objects,
 e.g. `finalizers` uses `merge` as patch strategy.
 ```go
 Finalizers []string `json:"finalizers,omitempty" patchStrategy:"merge" protobuf:"bytes,14,rep,name=finalizers"`
 ```
 
-##### Unordered Set
+##### Ordered Set
 
 There are 3 operations: add, delete, replace.
```
liggitt: If it's truly an ordered set, then there should be more than three operations (e.g. insert)

liggitt:
> Is there a reason why kubernetes/kubernetes#40373 can't just use the replace strategy? We don't have random things mutating the environment variables, I don't think.

`kubectl edit` submits a patch computed by `CreateTwoWayMergePatch`, which means clients have baked-in structs with merge strategies, so they will only send the added/deleted items, which means we can't change the strategy to replace server-side or we'll drop all the envvars they didn't change.

lavalamp:
Yeah that's a fair point, we should stop computing these from compiled in
structs. I talked with Phil & Mengqi some yesterday and suggested explicit
inserts, deletes, and moves instead of this ordering suggestion.

contributors/devel/api-conventions.md
lavalamp: Yeah, we shouldn't get rid of the current unordered set, we should add a way to deal with lists instead.

contributors/devel/api-conventions.md
mengqiy: We guaranteed the order from the user's config will always be preserved in the server.

> We only guarantee that A and B will keep order?

Yes

> how does a client indicate it want to move A to the end, tolerating any server-side reordering of the other elements?

If the server add some items in the list, a client cannot move something to the end without specify the whole list using this approach.

contributors/devel/api-conventions.md
mengqiy: Yes. Pulled strategic merge patch doc out.
https://github.com/kubernetes/community/pull/535

mengqiy:
Move to a proposal PR in https://github.com/kubernetes/community/pull/537. PTAL

PR #473, opened 2017-03-22
Updated the interface for admission controller

ravisantoshgudimetla:
Updated the documentation to have latest admission controller interface.

### DIFF
--- a/contributors/design-proposals/admission_control.md
+++ b/contributors/design-proposals/admission_control.md
@@ -44,20 +44,41 @@ An **AdmissionControl** plug-in is an implementation of the following interface:
 ```go
 package admission
 
-// Attributes is an interface used by a plug-in to make an admission decision
-// on a individual request.
+// Attributes is an interface used by AdmissionController to get information about a request
+// that is used to make an admission decision.
 type Attributes interface {
-  GetNamespace() string
-  GetKind() string
-  GetOperation() string
-  GetObject() runtime.Object
+	// GetName returns the name of the object as presented in the request.  On a CREATE operation, the client
+	// may omit name and rely on the server to generate the name.  If that is the case, this method will return
+	// the empty string
+	GetName() string
+	// GetNamespace is the namespace associated with the request (if any)
+	GetNamespace() string
+	// GetResource is the name of the resource being requested.  This is not the kind.  For example: pods
+	GetResource() schema.GroupVersionResource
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/admission_control.md
```diff
@@ -44,20 +44,41 @@ An **AdmissionControl** plug-in is an implementation of the following interface:
...
 type Interface interface {
-  // Admit makes an admission decision based on the request attributes
-  // An error is returned if it denies the request.
-  Admit(a Attributes) (err error)
+	// Admit makes an admission decision based on the request attributes
+	Admit(a Attributes) (err error)
+
+	// Handles returns true if this admission controller can handle the given operation
+	// where operation can be one of CREATE, UPDATE, DELETE, or CONNECT
+	Handles(operation Operation) bool
```
bsalamat: I think "Admissible" would be a better name for the function.

deads2k: > I think "Admissible" would be a better name for the function.

This looks to be sync-ing the actual impl, not suggesting any changes.

ravisantoshgudimetla: Thanks @bsalamat, @deads2k for reviewing- The only question that I have is should we include how to enable it in local-cluster script to test locally

derekwaynecarr:
Hmm, rather than let this get out of sync, I think the interface is stable enough that we could just point to the file at some point...

derekwaynecarr:
I guess the main point is this is a design-proposal document (from the earliest days of kubernetes) and not a user guide.  So explaining things like local clusters, etc. feels more tilted to user guide stuff.

ravisantoshgudimetla:
>Hmm, rather than let this get out of sync, I think the interface is stable enough that we could just point to the file at some point...

Makes sense. I can make that change.

>I  guess the main point is this is a design-proposal document (from the earliest days of kubernetes) and not a user guide. So explaining things like local clusters, etc. feels more tilted to user guide stuff.

The incoming links to page on github are from developer docs, if this is a design doc, we should remove incoming links from developer docs or have a new doc which talks about how to add adm controller. The other day when I was writing custom admission controller, I had to look at older PR's to know which files are to be touched for testing.

derekwaynecarr:
@ravisantoshgudimetla -- good point, there is a doc that states what admission controllers to enable/disable and in what order.  let me see if i can locate that.

PR #475, opened 2017-03-22
embed aggregator inside of the kube-apiserver

deads2k:
The aggregator will be embedded into the `kube-apiserver` during 1.7.  This updates our aggregated-apiserver doc to match our intent from https://docs.google.com/document/d/1lU1SnVtEec2iIfYx5U3L5N0za2YhfEOik0uPen276Ks/edit#

@kubernetes/sig-api-machinery-misc

### DIFF
--- a/contributors/design-proposals/aggregated-api-servers.md
+++ b/contributors/design-proposals/aggregated-api-servers.md
@@ -41,8 +41,7 @@ The following are related but are not the goals of this specific proposal:
 
 ## High Level Architecture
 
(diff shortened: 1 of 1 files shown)


deads2k:
@luxas you were asking about the simplification.  I think the rest of the doc stands and this is just a change to how the gateway mode will work.

lavalamp:
This LGTM, maybe I should make a few additional edits (want to rename single-user/test to 'standalone' mode.)

PR #476, opened 2017-03-23
Proposal: support multiple keys as merge key

ymqytw:
cc: @pwittrock


contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,336 @@
...
+
+Support a key set as the merge key in Strategic Merge Patch.
+
+## Motivation
+
+*Background*: In Strategic Merge Patch, Merge Key is the key to distinguish the entries in the list of non-primitive types.
+It must always be present and unique to perform the merge on the list of non-primitive types,
+and will be preserved.
+
+The issue in practice is that sometimes we cannot find a key that always be present and unique.
```
pwittrock: I'd replace this with:

>
The current implementation requires a single field that uniquely identifies each element in a list.  For some element Kinds, the identity is defined using multiple fields.  An example of this is the `x.x.xPort`, which is identified by `protocol` and `port`

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,336 @@
...
+[#39188](https://github.com/kubernetes/kubernetes/issues/39188) is a typical example. We need to
+support a key set as Merge Key. A key set must contain at least 1 key.
+
+## Proposed Change
+
+### API Change
+
+We will update the struct tags for the API resources whose Merge Key is not guaranteed to be always
+present and unique. The keys will be seperated by "|", i.e. `patchMergeKey:"<key1>|<key2>|<key3>"`.
```
pwittrock: Why '|' instead of ','?  What is more consistent with other tags and comments?

mengqiy: After checking types.go, I agree we should use ','.

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Open API
+
+Update Open API schema to have a string that contains multiple keys for `patchMergeKey`.
+The change should be trivial.
+
+### Strategic Merge Patch pkg
+
+Entries are considered as the same if and only if all the keys in the key set are identical.
+We allow keys to be missing in the key set as long as it is not empty.
```
pwittrock: What does this mean that it is ok for the field to be missing by not empty?  Can you give an example?

pwittrock: I see this is answered following :)

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,336 @@
...
+Take `Ports` as an example:
+
+Suppose we are using `name` and `port` as merge key as mentioned in section [API Change](#api-change)
+
+We have live config:
+```yaml
+spec:
+  type: NodePort
+  ports:
+    - protocol: UDP
```
pwittrock: comment that the name is missing here

contributors/design-proposals/key-set-as-merge-key.md
pwittrock: How about

> We allow keys to be unspecified, in which case they default to an empty value.

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,336 @@
...
+
+Document what the developer should consider when adding an API with `mergeKey`.
+
+## Version Skew
+
+Some changes are not backward compatible, e.g. moving the `mergeKey` and `patchStrategy` from field level to the list level.
+
+The plan to not break backward compatibility:
+
+In 1.7, add the logic support for key set, but keep the APIs' tags unchanged.
```
pwittrock: Yay

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,336 @@
...
+In 1.8, we introduce the changes to APIs' tags. The changes will be backward compatible.
+
+## Impacted APIs
+
+Impacted APIs are grouped into two groups:
+- Broken APIs
+- Need auditing APIs
+
+### Broken APIs
+(1a) `ContainerPort`: Change merge key from `containerPort` to `name|containerPort`.
```
pwittrock: update to ','

pwittrock:
I think this is ready for a broader review.  Would you present at the next sig-cli?

pwittrock:
@liggitt Would you be willing to act as a reviewer for this proposal for supporting multiple-fields as a mergeKey?

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: >If one or more of the mergeKeys is unspecified, it is interpreted as an empty value for purposes of uniquely identifying the object.

IMO we still want to distinguish unspecified and empty
```
- port: 12345
```
and
```
- name: ""
  port: 12345
```
The wording you give seems not support this?

Can you give me another wording?

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: Done.

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: Done.

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: Done.

mengqiy:
> I think this is ready for a broader review. Would you present at the next sig-cli?

Sure.

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+      port: 30420
+      name: tcpport
+      nodePort: 30420
+```
+
+### Docs
+
+Document what the developer should consider when adding an API with `mergeKey`.
+
+## Version Skew
```
liggitt: again, the 1 minor version skew policy is kubectl-specific, not for the API in general. I'm not sure how to evolve mergeKey on existing fields compatibly without making all clients do field-level API discovery which sounds pretty painful.

mengqiy: I understood.
The usage of OpenAPI is not just for kubectl. It is actually for all clients. We want all clients to use OpenAPI instead of baked-in types in the future. It is also useful and important for TPR and service catalog.

Ref: https://github.com/kubernetes/kubernetes/pull/44531

liggitt: > It is also useful and important for TPR and service catalog.

For discovering arbitrary resources, sure, but for a client that wants to patch exactly one thing on a known version of a known resource, forcing them to do openapi discovery to see if the meaning of that field in that version has changed is a pretty high bar.

smarterclayton: Yeah, in general.  MergeKeys are forever I think.

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Motivation
+
+*Background*: In Strategic Merge Patch, Merge Key is the key to distinguish the entries in the list of non-primitive types.
+It must always be present and unique to perform the merge on the list of non-primitive types,
+and will be preserved.
+
+The current implementation requires a single field that uniquely identifies each element in a list.
+For some element Kinds, the identity is defined using multiple fields.
+An [example](https://github.com/kubernetes/kubernetes/issues/39188 of this is the service.spec.ports, which is identified by `protocol` and `port`.
```
jamiehannaford: Link MD is not terminated

jamiehannaford: I'd emphasise both here: 

> identified by both `protocol` and `port`

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+We need to support a key set as Merge Key. A key set must contain at least 1 key.
```
jamiehannaford: How about:

> As a result we need to also support a set of keys as a Merge Key. A key set must be a list of strings at least 1 element long

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Proposed Change
+
+### API Change
+
+We will update the struct tags for the API resources whose Merge Key is not guaranteed to be always
```
jamiehannaford: I'm wondering whether this sentence makes sense. How about:

> for API resources that cannot be effectively merged with a single merge key

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+present and unique. The keys will be seperated by ",", i.e. `patchMergeKey:"<key1>,<key2>,<key3>"`.
+
+E.g. `Ports` in `ServiceSpec`.
```
jamiehannaford: Can we link to the source code struct?

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+  Ports []ServicePort `json:"ports,omitempty" patchStrategy:"merge" patchMergeKey:"name,port" protobuf:"bytes,1,rep,name=ports"`
+  ...
+}
+```
+
+All the impacted APIs are listed in section [Impacted APIs](#impacted-apis)
+
+### Open API
+
+Update Open API schema to have a string that contains multiple keys for `patchMergeKey`.
```
jamiehannaford: What does this mean, that the type of `patchMergeKey` needs to change? Can you give an example of the OpenAPI change?

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+### Strategic Merge Patch pkg
+
+Entries are considered as the same if and only if all the keys in the key set are identical.
+We allow keys to be missing in the key set as long as it is not empty.
+
+Take `Ports` as an example:
+
+Suppose we are using `name` and `port` as merge key as mentioned in section [API Change](#api-change)
+
+We have live config:
```
jamiehannaford: > This config currently exists:

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+spec:
+  type: NodePort
+  ports:
+    - protocol: UDP
+      # name is missing here
+      port: 30420
+      nodePort: 30420
+```
+
+We want to add another port, and the user's local config file becomes:
```
jamiehannaford: > The user then wants to add a new port. They will send this manifest:

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+      port: 30420
+      name: udpport
+      nodePort: 30420
+    - protocol: TCP
+      port: 30420
+      name: tcpport
+      nodePort: 30420
+```
+
+The patch we send should be:
```
jamiehannaford: > The patch manifest that will be sent is:

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+We want to add another port, and the user's local config file becomes:
+```yaml
+spec:
+  type: NodePort
+  ports:
+    - protocol: UDP
+      port: 30420
+      name: udpport
```
jamiehannaford: What happens if the user omits the name in their manifest? Will the existing `UDP` port stay with an empty name?

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: > but for a client that wants to patch exactly one thing on a known version of a known resource, forcing them to do openapi discovery to see if the way they are supposed to patch that field in that version has changed is a pretty high bar.

@liggitt Do you have specific client(s) in your mind?

pwittrock: @thockin FYI

To clarify, this is an issue generally with changing merge keys, not specifically with supporting tuples as merge keys.

A couple thoughts:
- The ship may have sailed on this already, but I would think that the client should sent the merge keys and merge strategies as part of the patch request so it could tell the server how the patch request is intended to be merged.  
- It maybe possible to introduce additional keys for fields while keeping backwards compatibility.  Would need to think through the combinatorics, but I am wondering if some of the merge key tuple is missing from the patch request (e.g. the new part), if the server could fallback on the remaining pieces as the merge key.
- If merge keys really are "forever", and we needed to change the one referenced in this case, would this then require bumping the version for everything that has a PodTemplate in it - e.g. every controller type?  The cure may be worse than the disease in this case.
- Given that we don't publish what the merge keys actually are for resource kinds in any user facing documentation besides the `types.go` files (correct me if I am wrong here), are they considered first class pieces of the API, or are they an implementation detail?  Does our API definition include metadata about fields that is only defined in code?  Do we have any guess as to which clients besides kubectl calculate patch requests?

contributors/design-proposals/key-set-as-merge-key.md
pwittrock: What about ints?

contributors/design-proposals/key-set-as-merge-key.md
smarterclayton: Let me qualify - mergeKeys are forever if a patch which previously "worked" (in the sense that a client uses it to accomplish a concrete mutation successfully) is broken when we change the mergeKey.  If the patch was impossible or broken (envVar is a clear example of "we were totally broken because order is meaningful in env vars by our spec"), then I don't think a mergeKey change counts as a break.

We could consider versioning the patch strategy per type.  Or the micro version approach Dan was championing for fields, but with restricted scope - v1.Pod has mergeKey "foo" for field "bar" at version 1, but "baz" for field "bar" at version 2.  And we simply default the merge version.

pwittrock: Thanks for the clarification.  IMHO the `ServicePort` is broken in a similar fashion as envVar - that is - the patch is impossible / broken for things we support in our spec - e.g. having a list containing 2 different `ServicePort`s with the same `port` value and different `protocol` values.  However some clients may not be encountering these issues because they don't rely on those capabilities.

pwittrock: Thinking on this more, I am pretty sure we can introduce additional fields for merge keys while retaining backwards compatibility by having the server fallback to use the present merge keys in the event the additional merge key fields are missing from the patch request.

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: The key is always a string, because we require all maps to be `map[string]interface{}`.

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: The operation will fail because of the validation error. Name is required if there are more than one ports.

mengqiy:
Added edge cases.
@pwittrock There is case when deleting maybe you will want to think about it. https://github.com/kubernetes/community/pull/476/commits/5e9717fe4f1e590615db0ad09b5c6ce3d1925525#diff-7bf73a741414e64cb422a883e2de7f0eR182

mengqiy:
PTAL

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,417 @@
...
+}
+```
+
+We also need to introduce a new optional directive `$patchMergeKey` in the patch.
+This directive contains a list of strings which are the merge keys used in this patch.
+
+An example patch will look like:
+```yaml
+list:
+- $patchMergeKey:
```
liggitt: in this example, `$patchMergeKey` is included in the first item's definition... is that intended?

contributors/design-proposals/key-set-as-merge-key.md
```diff
@@ -0,0 +1,417 @@
...
+- foo: a
+  bar: x
+  other: val
+```
+
+#### When `$patchMergeKey` only has part of all merge keys
+
+When `$patchMergeKey` only has part of all merge keys,
+the server will apply the patch to all the matching items.
+So an update or a deletion may apply to multiple items in the list.
```
liggitt: Accepting a `$patchMergeKey` that does not uniquely identify elements is what this proposal is trying to fix... I don't think a `$patchMergeKey` key that does not uniquely identify items should be accepted.

If it was, how would this interact with the proposed `$setElementOrder` directive, which is supposed to contain a list of keys? Keys that identify more than one item in the list mean that directive becomes ambiguous:

server items:
```yaml
ports:
- port: 8080
  protocol: TCP
  name: http-tcp
- port: 9090
  protocol: TCP
- port: 8080
  protocol: UDP
  name: http-udp
```

patch:
```yaml
$setElementOrder/ports:
- port: 8080
- port: 8080
- port: 9090
ports:
- port: 8080
  protocol: UDP
  name: http-udp-renamed
```

It's not clear which 8080 port was ordered where.

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: I think `$patchMergeKey` should be in each entry in the list. This simple example only has one entry.

contributors/design-proposals/key-set-as-merge-key.md
mengqiy: > I don't think a $patchMergeKey key that does not uniquely identify items should be accepted.

You are right.

`$setElementOrder` should always use all of the recommended merge keys that are present, so each entry in `$setElementOrder` list can uniquely identify an entry in the live list.
```yaml
$setElementOrder/ports:
- port: 8080
  protocol: TCP
- port: 8080
  protocol: UDP
- port: 9090
  protocol: TCP
ports:
- port: 8080
  protocol: UDP
  name: http-udp-renamed
```

mengqiy: I will add a section for `$setElementOrder`.

pwittrock:
The majority of the content in this doc is now about how to change a merge key value.  Lets pull the multi-field merge key into a separate design doc and review that separately from the design around to how to change merge keys for fields and how to control the merge key from the client side

mengqiy:
Updating:
- Addressed comments 
- Pull supporting multi-fields merge key logic to https://github.com/kubernetes/community/pull/610 per https://github.com/kubernetes/community/pull/476#issuecomment-300311478
- Add content of how to move merge keys that are at wrong level (child level)

liggitt:
cc @kubernetes/sig-cli-proposals @kubernetes/sig-api-machinery-proposals

contributors/design-proposals/changing-merge-key.md
```diff
@@ -0,0 +1,670 @@
...
+An [example](https://github.com/kubernetes/kubernetes/issues/39188) is the service.spec.ports,
+which is identified by both `protocol` and `port`.
+Therefore, we should change the merge key to multi-fields.
+
+There are some existing APIs that put the merge key tag at the wrong place.
+We should also correct them.
+
+## Goal
+
+Propose a solution of changing the merge keys for existing APIs without breaking backward compatibility.
```
bgrant0607: We need to take a step back and think about this systematically.

lavalamp: I think we should consider building a more principled form of the patch syntax. We keep making ad-hoc fixes.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/476)

contributors/design-proposals/changing-merge-key.md
smarterclayton: Yeah, the existing RFCs have their limitations but evolving one of those with the improvements from SMR seems like a stronger path than retrofitting into existing SMR.

bgrant0607:
@mengqiy @pwittrock @lavalamp I think we should close this.

lavalamp:
Agree, we're fixing this class of things with a different mechanism entirely.

PR #384, opened 2017-02-19
Refine the Deployment proposal and move away from hashing

kargakis:
Alternative proposal to https://github.com/kubernetes/community/pull/477

@kubernetes/sig-apps-api-reviews @kubernetes/api-reviewers

PR #477, opened 2017-03-23
Refine the Deployment proposal and switch hashing algorithm

kargakis:
Alternative proposal to https://github.com/kubernetes/community/pull/384

@kubernetes/sig-apps-api-reviews @kubernetes/api-reviewers

### DIFF
--- a/contributors/design-proposals/deployment.md
+++ b/contributors/design-proposals/deployment.md
@@ -3,14 +3,13 @@
 ## Abstract
 
 A proposal for implementing a new resource - Deployment - which will enable
-declarative config updates for Pods and ReplicationControllers.
-
-Users will be able to create a Deployment, which will spin up
-a ReplicationController to bring up the desired pods.
-Users can also target the Deployment at existing ReplicationControllers, in
-which case the new RC will replace the existing ones. The exact mechanics of
-replacement depends on the DeploymentStrategy chosen by the user.
-DeploymentStrategies are explained in detail in a later section.
+declarative config updates for ReplicaSets. Users will be able to create a
+Deployment, which will spin up a ReplicaSet to bring up the desired Pods.
+Users can also target the Deployment to an existing ReplicaSet either by
+rolling back an existing Deployment or creating a new Deployment that can
+adopt an existing ReplicaSet. The exact mechanics of replacement depends on
+the DeploymentStrategy chosen by the user. DeploymentStrategies are explained
+in detail in a later section.
 
 ## Implementation
 
@@ -33,10 +32,10 @@ type Deployment struct {
 type DeploymentSpec struct {
   // Number of desired pods. This is a pointer to distinguish between explicit
   // zero and not specified. Defaults to 1.
-  Replicas *int
+  Replicas *int32
 
-  // Label selector for pods. Existing ReplicationControllers whose pods are
-  // selected by this will be scaled down. New ReplicationControllers will be
+  // Label selector for pods. Existing ReplicaSets whose pods are
+  // selected by this will be scaled down. New ReplicaSets will be
   // created with this selector, with a unique label `pod-template-hash`.
   // If Selector is empty, it is defaulted to the labels present on the Pod template.
   Selector map[string]string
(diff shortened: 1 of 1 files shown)


kargakis:
WIP: https://github.com/kubernetes/kubernetes/pull/38714

janetkuo:
@kargakis we'd need to support migration if switching hashing algorithm, correct?

kargakis:
@janetkuo correct, I am linking the implementation in the comment above yours.

kargakis:
@janetkuo @kow3ns @bgrant0607 @smarterclayton @erictune updated as per our discussion yesterday

cc: @kubernetes/sig-apps-api-reviews

contributors/design-proposals/deployment.md
```diff
@@ -65,7 +67,7 @@ const (
   // Kill all existing pods before creating new ones.
   RecreateDeploymentStrategyType DeploymentStrategyType = "Recreate"
 
-  // Replace the old RCs by new one using rolling update i.e gradually scale down the old RCs and scale up the new one.
+  // Replace the old RSs by new one using rolling update i.e gradually scale down the old RSs and scale up the new one.
```
janetkuo: RSs --> ReplicaSets

contributors/design-proposals/deployment.md
```diff
@@ -163,52 +169,71 @@ LastTransitionTime to PodCondition.
...
-For ex: consider the following case:
-- User creates a deployment to rolling-update 10 pods with image:v1 to
-  pods with image:v2.
-- User then deletes this deployment while the old and new RCs are at 5 replicas each.
-  User will end up with 2 RCs with 5 replicas each.
-User can then create the same deployment again in which case, DeploymentController will
-notice that the second RC exists already which it can ramp up while ramping down
+Users can cancel a rollout by doing a non-cascading deletion of the Deployment
+before it is complete. Recreating the same Deployment will resume it.
+For example, consider the following case:
```
janetkuo: This example shows how to pause and resume a rollout using non-cascading deletion, not cancel a rollout. However the first sentence goes "users can cancel a rollout..."

contributors/design-proposals/deployment.md
kargakis: fixed

contributors/design-proposals/deployment.md
kargakis: fixed

janetkuo:
As a side note: this proposal = `fnv` + https://github.com/kubernetes/community/pull/384#issuecomment-292659819

erictune:
I know that more than one person contributed significant time and ideas to this proposal.  I think it would be a good idea to list each of them as authors at the top of the proposal.

kargakis:
> I know that more than one person contributed significant time and ideas to this proposal. I think it would be a good idea to list each of them as authors at the top of the proposal.

Agreed. Added commenters from the initial proposal plus anybody else that participated in the recent discussions around hashing vs templateGeneration. Let me know if I am missing somebody.

erictune:
Thanks for updating this.  Please don't include me, though, as I didn't
spend much time on this.

kargakis:
All comments addressed, I am going to merge this one and follow-up with the implementation which is in-flight in kubernetes/kubernetes#44774

smarterclayton:
So I don't see this in the doc, but I think a rather significant use case for collisions needs to be clarified: initializers on replica sets.  Effectively anything that mutates or defaults the replicaset (so that those defaults are applied exactly once, and each version is consistent).  We hit this on a prototype initializer that resolves image tags to fully qualified hashes, which causes the deployment controller to not recognize the template as it's own.  I would not want collision avoidance to apply in this scenario...

bgrant0607:
@smarterclayton I commented on this scenario here: https://github.com/kubernetes/kubernetes/issues/1697#issuecomment-320347887

And in the PR that (erroneously) changed image from optional to required.

I'd expect the Deployment's pod template to be updated, in which case there would be no collision.

The user needs to specify the unresolved image tag in a different field.

For instance, we could add an imageStream field (or struct) specifies the repo+image+tag. That would be automatically translated to the digest form into the existing image field in the DeploymentSpec, which would trigger the creation of a new ReplicaSet whose hash wouldn't collide with others'.

smarterclayton:
I was also thinking of other examples as well - autolabellers or auto annotators, snapshotting config maps (at deployment time automatically copy config maps and set owner refs on them to a particular replica set), possibly sidecar injectors (I think injecting some sidecars at pod scope is wrong).  Also node selectors and tolerations, conceivably. 

Are we ready to eliminate the possibility that replica set templates can be filled in (and thus preserve an atomic snapshot of rollout that can be reverted to under pressure)?  I can think of lots of "on demand" policies that we can benefit from doing at the replica set level, where the natural action of the deployment (stamp a new desired RS out) can capture that point in time.

bgrant0607:
The hash shouldn't be used for comparison. It should just be a uniquifier. Strategic merge diff should be used to determine whether any of the ReplicaSets matches the current Deployment template.

The model I had in mind was that any automatic injection that we wanted to trigger a rollout would also modify DeploymentSpec fields that were not specified by the user.

PR #478, opened 2017-03-23
Introduce stable pods

kargakis:
@kubernetes/sig-apps-api-reviews @kubernetes/api-reviewers

### DIFF
--- a/contributors/design-proposals/stable-pods.md
+++ b/contributors/design-proposals/stable-pods.md
@@ -1,0 +1,100 @@
+# Introduce Stable Pods
+
+**Author**: @kargakis
+
+**Status**: Proposal
+
+## Abstract
+Today, Deployments, ReplicaSets, and DaemonSets use MinReadySeconds in order to include
+an additional delay on top of readiness checks and facilitate more robust rollouts. The
+ReplicaSet controller decides how many Ready Pods a ReplicaSet runs for at least
+MinReadySeconds and that number is reflected in the ReplicaSetStatus as AvailableReplicas.
+The Deployment controller when scaling down old ReplicaSets, will not proceed if the
+minimum number of Pods that are required to run are not Ready for at least MinReadySeconds.
+If MinReadySeconds is not specified then the Pods are included in AvailableReplicas as
+soon as they are Ready. 
+
+Problems that have been identified so far with the current state of things:
+1. A Pod is marked Ready by the kubelet as soon as it passes its readiness check. The
+ReplicaSet controller runs as part of master and estimates when a Pod is stable by
+comparing the time the Pod became Ready (as seen by the kubelet) with MinReadySeconds.
+[Clock-skew between master and nodes will affect these availability checks](https://github.com/kubernetes/kubernetes/issues/29229). 
+2. PodDisruptionBudget is working with Ready Pods and has no notion of MinReadySeconds
+when used by workload controllers.
+3. The workload controllers cannot take advantage of readiness checks when checking the
+state of their underlying Pods.
+4. It's very hard to reason about which pods are not stable just by looking in the status
+of a controller, because it only reports aggregated data. Imagine a ReplicaSet with 200
+replicas, and only 3 of them are unstable.
+
+## Design
+All problems above can be solved by moving MinReadySeconds in the PodSpec. Once a kubelet
+observes that a Pod has been Ready for at least MinReadySeconds without any of its
+containers crashing and passes its readiness check successfully for another time, it will
+update the PodStatus with an Stable Condition set to Status=True. Higher-level
+orchestrators running on different machines such as the ReplicaSet or the
+PodDisruptionBudget controller will merely need to look at the Stable condition that is set
+in the status of a Pod.
+
+Services should not be affected by the implementation of Stable Pods. Endpoints should
+continue to be added or removed when a Pod transitions to or from Ready.
+
+### API changes
+
+A new field is proposed in the PodSpec:
+```go
+	// Minimum number of seconds for which a newly created Pod should be Ready
+	// without any of its container crashing, for it to be considered Stable.
+	// Defaults to 0 (pod will be considered Stable as soon as it is Ready)
+	// +optional
+	MinReadySeconds *int32 `json:"minReadySeconds,omitempty"`
+```
+and a new PodConditionType:
+```go
+	// PodStable is added in a Pod after it has been Ready for at least MinReadySeconds.
+	// The Pod should already have an Endpoint and serve requests, this condition lets
+	// higher-level orchestrators know that the Pod is stable after some amount of time
+	// without having any of its containers crashed.
+	PodStable PodConditionType = "Stable"
+```
+
+The reason for adding a separate Condition instead of extending the existing Ready
+condition is mainly for allowing traffic in Pods. Users can have robust rollouts if their
+applications expose metrics which are then fed into readiness checks.
+
+Additionally:
+* Deployments/ReplicaSets/DaemonSets already use MinReadySeconds in their Spec so we
+should probably deprecate those fields in favor of the field in the PodTemplateSpec and
+remove them in a future version. 
+* All the aforementioned controllers will not propagate MinReadySeconds from their Spec
+down to their PodTemplateSpec because that may lead in differences in the PodTemplateSpec
+between a Deployment and a ReplicaSet resulting in new rollouts. If MinReadySeconds is
+specified both in the Spec and PodTemplateSpec for a controller, the field in the
+PodTemplateSpec will stomp the field in the Spec. If it is specified only in the Spec,
+Pods can be created using it without updating the controller's PodTemplateSpec.
+
+### kubelet changes
+For a Pod that specifies MinReadySeconds, kubelet will need to rerun the readiness check
+after MinReadySeconds and also check if any of its containers has crashed in the meantime.
+If everything is ok, the kubelet will switch the Pod's Stable condition to Status=True.
+Pods that don't specify MinReadySeconds will be considered Stable as soon as they are
+Ready.
+
+### Controller manager changes
+The ReplicaSet and DaemonSet controllers will create new Pods by setting MinReadySeconds in
+the PodSpec if it is specified either in their Spec or PodTemplateSpec. The value in the
+PodTemplateSpec should stomp the value in the Spec. For Pods that do not specify
+MinReadySeconds (those running on old kubelets), the controllers will continue to use the
+current approach for estimating stability. Eventually, we should switch all controllers that
+work with Pods to use virtual clocks to avoid clock-skew with older kubelets.
+
+The PDB controller will need to be extended to recognize the Stable Condition in Pods.
+It may also need to get into the business of estimating stability for ReplicaSets that
+already use MinReadySeconds because some of their underlying Pods may run on old kubelets.
+More discussion about the PDB controller can be found [here](https://github.com/kubernetes/kubernetes/issues/34776).
+
+## Future work
+Since we will move forward with Stable as the name of the Condition for a Pod that is Ready
+for at least MinReadySeconds, we should also rename the status fields in ReplicaSet,
+Deployment, and DaemonSet from AvailableReplicas to StableReplicas. Unfortunately this
+cannot be done for v1 due to backwards-compatibility.


contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+pod template because that will lead in differences in the pod templates between a
+Deployment and a ReplicaSet resulting in new rollouts. If MinReadySeconds is specified
+both in the spec and pod template for a Deployment/ReplicaSet and it's not the same value,
+a validation error will be returned (tentative). API defaulting can set MinReadySeconds
+in the spec, if it's specified only in the PodTemplate (tentative).
+* ReplicaSets that specify MinReadySeconds only in the ReplicaSetSpec, can create new Pods
+by specifying MinReadySeconds in their PodSpec (w/o updating the ReplicaSet pod template).
+
+## kubelet changes
+For a Pod tha specifies MinReadySeconds, kubelet will need to check (after MinReadySeconds)
```
derekwaynecarr: i think in the absence of min ready seconds, available condition would just map with ready, no?   i would rather all things report a condition instead of just a few...

contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+1. A Pod is marked ready by the kubelet as soon as it passes its readiness check. The
+ReplicaSet controller runs as part of master and estimates when a Pod is available by
+comparing the time the Pod became ready (as seen by the kubelet) with MinReadySeconds.
+[Clock-skew between master and nodes will affect the availability checks](https://github.com/kubernetes/kubernetes/issues/29229). 
+2. PodDisruptionBudget is working with ready Pods and has no notion of MinReadySeconds
+when used by a Deployment/ReplicaSet. 
+
+Both problems above can be solved by moving MinReadySeconds in the PodSpec. Once kubelet
+observes that a Pod has been ready for at least MinReadySeconds without any of its containers
+crashing, it will update the PodStatus with an Available condition set to Status=True.
```
thockin: Can I still ask for a better name?  Is "Stable" or "StableReady" or "ReadyStable" problematic?

contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+
+## API changes
+
+A new field is proposed in the PodSpec:
+```go
+	// Minimum number of seconds for which a newly created pod should be ready
+	// without any of its container crashing, for it to be considered available.
+	// Defaults to 0 (pod will be considered available as soon as it is ready)
+	// +optional
+	MinReadySeconds *int32 `json:"minReadySeconds,omitempty"`
```
thockin: Correspondingly "StableReadySeconds", or Min... ?

contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+Deployment and a ReplicaSet resulting in new rollouts. If MinReadySeconds is specified
+both in the spec and pod template for a Deployment/ReplicaSet and it's not the same value,
+a validation error will be returned (tentative). API defaulting can set MinReadySeconds
+in the spec, if it's specified only in the PodTemplate (tentative).
+* ReplicaSets that specify MinReadySeconds only in the ReplicaSetSpec, can create new Pods
+by specifying MinReadySeconds in their PodSpec (w/o updating the ReplicaSet pod template).
+
+## kubelet changes
+For a Pod tha specifies MinReadySeconds, kubelet will need to check (after MinReadySeconds)
+if any of the Pod containers has crashed. If not, it will switch the Available condition to
```
thockin: Does this flip to false every time a container crashes, or is it latched to true once it is achieved?

thockin: after the initial period, if a container has crashed, how long does it wait?  We don't have enough info (I think) to know WHEN it crashed.  Maybe restate this as kubelet will reset the timer every time a container crashes until this condition is achieved?

thockin: Is this really EVERY container?  What if it is OK for my helper container to crash occasionally?  Reality is that it happens...

contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+Status=True in the status of the Pod. Pods that don't specify MinReadySeconds, won't have
+the Available condition set in their status.
```
thockin: Most conditions are assumed to have their default value == the Go zero value, and are named as such, I think.  Brian has better context on this, but it seems weird to me that nonexistent implies true

contributors/design-proposals/minreadyseconds.md
kargakis: Stable sgtm

contributors/design-proposals/minreadyseconds.md
kargakis: I would expect this to be similar to the Ready condition. If a Pod is Ready and then one of its container crashes, the Pod stops being Ready.

contributors/design-proposals/minreadyseconds.md
kargakis: Yes, similar to Derek's comment, we probably want to add the condition to all Pods and default to false. I was thinking though that such an action would be disruptive but then again in order to upgrade a Node you need to drain it first so I am probably wrong.

contributors/design-proposals/minreadyseconds.md
kargakis: Agreed, will update.

contributors/design-proposals/minreadyseconds.md
kargakis: @smarterclayton @bgrant0607 thoughts? This makes sense if we roll with "Stable" but this entails that we then need to update the status fields in our controllers from AvailableReplicas to StableReplicas (something to keep in mind for v2 I guess at this point)

contributors/design-proposals/minreadyseconds.md
thockin: I don't think that's true, is it?

contributors/design-proposals/minreadyseconds.md
thockin: Why?  You don't need to use the same name everywhere - names are contextual.

contributors/design-proposals/minreadyseconds.md
kargakis: I just double-checked because it felt for a moment like I am crazy but yes, this is true. Once a container crashes, the Ready condition switches back to False. Of course, if the container is restarted and comes back to a healthy state the condition will transition back to True and then you end up crashlooping.

contributors/design-proposals/minreadyseconds.md
kargakis: Trying to convey the same thing via different names is at least confusing to me as a user.

smarterclayton: You must be "Ready" for "X Seconds" in order to be "Stable".  "StableAfterReadySeconds" is awkward but is somewhat closer.

contributors/design-proposals/minreadyseconds.md
smarterclayton: If Stable is just a longer Ready.... why not increase the amount of time before you are Ready?

kargakis: That's a good question, today MinReadySeconds is just that, an additional delay on top of Ready.

kargakis: And the answer is that longer wait for Ready means you are not receiving traffic.

contributors/design-proposals/minreadyseconds.md
kargakis: I think I still prefer MinReadySeconds ("Minimum Seconds the pod needs to be Ready") but I won't hang on it if you guys feel strong about changing it.

thockin:
I just think it doesn't convey to a naive user what it does.

thockin:
really....I do not think that is what SHOULD happen...  BUt I accept that
it might be what DOES happen.

kargakis:
> really....I do not think that is what SHOULD happen... BUt I accept that it might be what DOES happen.

I remember that this didn't use to be the case but you don't want to continue sending traffic in broken Pods is my understanding of it. Furthermore, it may be just your sidecar that is failing, but there is no way today to separate between the main and the sidecard via the API... Or at least not a way I am aware of.

thockin:
Well, we control the API, so we could have added a flag indicating whether
a particular container should flip "ready" or not.

contributors/design-proposals/minreadyseconds.md
smarterclayton: I don't know that I love any of the new names.

kargakis:
Yes, I believe it's reasonable. Today effectively all containers are
affecting Ready by default. By adding that flag you can block switching
Ready (and Stable) back n forth :)

contributors/design-proposals/minreadyseconds.md
smarterclayton: I guess you need to take traffic before you know whether you're ok.  If you have a liveness or readiness check that *also* checks your business metrics, you can use that to signal a problem.

1. created
2. check readiness (no traffic, so no errors in your metrics)
3. take traffic, users getting 500's
4. check readiness (got errors, fail)
5. go out of rotation
6. ???

What happens at step six?  You just oscillate in and out, causing some percentage of users to get failures?  In openshift we can fail the deployment config, and I guess Deployments can hold the deployment and backoff creation.  Stability would be false, since we're oscillating.

kargakis: So there seem to be two types of errors an application can have thus a 500 that a user will see: 1) those that will make the application exit. In this case, the pod goes out of rotation anyway, it never transitions to Stable, the Deployment blocks. We can simply verify these by looking at the container statuses. 2) Bugs that may not cause the application to exit. Pod is not going out of rotation. Deployment still blocks. Can't look at container statuses, readiness checks may not be enough unless business metrics are involved.

Both cases are covered by rerunning the readiness check I guess. I need to call this out. Does it also mean that the endpoints controller will need to remove endpoints for Pods that never reach Stable?

6 does not seem to be something that needs to be handled in this proposal. We have ProgressDeadlineSeconds and there is AutoRollback that needs to be implemented.

contributors/design-proposals/minreadyseconds.md
davidopp: SecondsRequiredStableBeforeReady?

contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+```go
+	// Minimum number of seconds for which a newly created pod should be ready
+	// without any of its container crashing, for it to be considered available.
+	// Defaults to 0 (pod will be considered available as soon as it is ready)
+	// +optional
+	MinReadySeconds *int32 `json:"minReadySeconds,omitempty"`
+```
+and a new PodConditionType:
+```go
+	// PodAvailable is added in a ready pod that has MinReadySeconds specified. The pod
```
davidopp: The second sentence of this comment is confusing. BTW is it worth stating somewhere in this doc how Ready vs Available are treated by services?

contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+## Controller manager changes
+The ReplicaSet controller will create new Pods by setting MinReadySeconds in their PodSpec
+if it's specified in the ReplicaSetSpec (and not in the ReplicaSet pod template). For Pods
+that don't specify MinReadySeconds, it can switch to use a virtual clock and continue using
+the current approach of estimating availability. This will also help in keeping new servers
+backwards-compatible with old kubelets.
+
+## Future work
+The PDB controller will need to be extended to recognize the Available condition in Pods.
+It may also need to get into the bussiness of estimating availability for Deployments/ReplicaSets
```
davidopp: business

contributors/design-proposals/minreadyseconds.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Controller manager changes
+The ReplicaSet controller will create new Pods by setting MinReadySeconds in their PodSpec
+if it's specified in the ReplicaSetSpec (and not in the ReplicaSet pod template). For Pods
+that don't specify MinReadySeconds, it can switch to use a virtual clock and continue using
+the current approach of estimating availability. This will also help in keeping new servers
+backwards-compatible with old kubelets.
+
+## Future work
+The PDB controller will need to be extended to recognize the Available condition in Pods.
```
davidopp: Might want to cite #34776 where we had some discussion of this topic.

contributors/design-proposals/minreadyseconds.md
kargakis: @davidopp isn't an api convention to follow the *Seconds pattern? I think it is but I may be wrong.

contributors/design-proposals/minreadyseconds.md
kargakis: Tried to rephrase this one. Also added a sentence where I explicitly mention that Services are not going to be affected by this change.

contributors/design-proposals/minreadyseconds.md
kargakis: done

contributors/design-proposals/minreadyseconds.md
kargakis: done

kargakis:
Addressed most the comments, ptal

Still tentative is the name in the PodSpec. I still prefer MinReadySeconds over all current suggestions because 1) it's less ugly and 2) it's already used by three different controllers (so people should already be familiar with it).

kargakis:
cc: @mfojtik since somebody from PM may need to drive the implementation :)

kargakis:
@kubernetes/sig-node-api-reviews

contributors/design-proposals/stable-pods.md
```diff
@@ -0,0 +1,93 @@
...
+If everything is ok, the kubelet will switch the Pod's Stable condition to Status=True.
+Pods that don't specify MinReadySeconds will be considered Stable as soon as they are
+Ready.
+
+### Controller manager changes
+The ReplicaSet and DaemonSet controllers will create new Pods by setting MinReadySeconds in
+the PodSpec if it is specified either in their Spec or PodTemplateSpec. The value in the
+PodTemplateSpec should stomp the value in the Spec. For Pods that do not specify
+MinReadySeconds (those running on old kubelets), the controllers will continue to use the
+current approach for estimating stability. Eventually, we should switch all controllers that
```
janetkuo: Can we make PodTemplate.Spec.MinReadySeconds default to whatever is set in the controller spec when creating pods (we don't need to touch PodTemplate so won't cause additional rollouts)? It overwrites controller Spec.MinReadySeconds only when it's set (not `nil`).

contributors/design-proposals/stable-pods.md
```diff
@@ -0,0 +1,93 @@
...
+
+Problems that have been identified so far with the current state of things:
+1. A Pod is marked Ready by the kubelet as soon as it passes its readiness check. The
+ReplicaSet controller runs as part of master and estimates when a Pod is stable by
+comparing the time the Pod became Ready (as seen by the kubelet) with MinReadySeconds.
+[Clock-skew between master and nodes will affect these availability checks](https://github.com/kubernetes/kubernetes/issues/29229). 
+2. PodDisruptionBudget is working with Ready Pods and has no notion of MinReadySeconds
+when used by workload controllers.
+3. The workload controllers cannot take advantage of readiness checks when checking the
+state of their underlying Pods.
```
kargakis: Agreed. Updated the proposal.

contributors/design-proposals/stable-pods.md
kargakis: This is what I mean by "The ReplicaSet and DaemonSet controllers will create new Pods by setting MinReadySeconds in the PodSpec if it is specified either in their Spec or PodTemplateSpec. The value in the PodTemplateSpec should stomp the value in the Spec."

kargakis:
@tnozicka

contributors/design-proposals/stable-pods.md
janetkuo: gotya

kargakis: I also mention it above in the API changes section.

contributors/design-proposals/stable-pods.md
```diff
@@ -0,0 +1,100 @@
...
+	PodStable PodConditionType = "Stable"
+```
+
+The reason for adding a separate Condition instead of extending the existing Ready
+condition is mainly for allowing traffic in Pods. Users can have robust rollouts if their
+applications expose metrics which are then fed into readiness checks.
+
+Question: Do we need a separate kind of probe (stabilityProbe?) or should we re-use
+the readinessProbe? What if a user needs to take into account different kinds of metrics
+when doing the stability check than those used in her readiness probe?
```
janetkuo: Reusing readinessProbe makes sense to me.

kargakis: sgtm

kargakis:
@wanghaoran1988 is going to start working on the kubelet side of things

contributors/design-proposals/stable-pods.md
```diff
@@ -0,0 +1,100 @@
...
+remove them in a future version. 
+* All the aforementioned controllers will not propagate MinReadySeconds from their Spec
+down to their PodTemplateSpec because that may lead in differences in the PodTemplateSpec
+between a Deployment and a ReplicaSet resulting in new rollouts. If MinReadySeconds is
+specified both in the Spec and PodTemplateSpec for a controller, the field in the
+PodTemplateSpec will stomp the field in the Spec. If it is specified only in the Spec,
+Pods can be created using it without updating the controller's PodTemplateSpec.
+
+### kubelet changes
+For a Pod that specifies MinReadySeconds, kubelet will need to rerun the readiness check
```
wanghaoran1988: @kargakis Why we need rerun the readiness check? Can we just wait  "minReadySeconds" seconds after it's ready and check if the pod is still ready?

kargakis: My understanding is that ideally that check would be able to identify failures that are not caught by readiness (by using metrics exposed by the app), so we may want to think about introducing a net-new probe. @kubernetes/sig-node-proposals @janetkuo @smarterclayton Or do you guys think re-running the readiness check is enough?

kargakis:
I'm not able to move this forward anymore so I am going to close.

kow3ns:
@kargakis why are you closing this?

kargakis:
> @kargakis why are you closing this?

> I'm not able to move this forward

kargakis:
https://trello.com/c/g3Dly9Xi/803-5-upstream-introduce-stable-pods

davidopp:
FWIW I think this is a good idea in general and we should try to implement it, or something like it, eventually.

mattmb:
I think this is still a good idea and want to share some context. I'm also interested if there are any related developments elsewhere that anyone can link me to or if we think the right SIGs/contributors know about this suggestion?

At Yelp we've had one form or another of this problem since before we even used kubernetes! I would describe it like so: "When we launch a new container it takes some time to be discoverable in whatever service mesh technology you choose. In large clusters, proxies or iptables rules or whatever might have to be updated on many nodes or inside many pods. If you remove old versions of the container before this happens your service will be under capacity and likely to send errors".

Before kubernetes we had our own deployment code that checked a sample of our hosts to check that the new containers had converged in our mesh before the deployment proceeded to kill the old containers. In kubernetes we "abuse" the readiness check by doing a similar check of our service mesh before saying a Pod is ready. We could alternatively use min ready seconds but in cases where our mesh infrastructure goes wrong or is really slow for some reason this timer would expire and rather than a stalled deployment we have a service outage. I think in general people would prefer a stalled deployment. So I do think a second condition is needed for the deployment controller (or replica set controller?) to know when it's safe to proceed. I think it should also be possible to define a separate probe because I think it's dependent on your choice of service mesh as to what you should check before proceeding. I don't know enough about kubernetes internals to comment with authority but it feels like it could be possible in a backwards compatible way where the deployment controller falls back to depending on readiness condition when this new condition isn't defined? And then users have a choice between min ready seconds which defines this new condition or they can define a custom check that overrides that?

thockin:
We have this notion of pod readiness gates - which allows a 3rd party to
offer an opinion about whether a pod is ready, so the higher-order
controllers can proceed.  Does that get to what you are describing?

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate

mattmb:
I think probably it doesn't (but certainly possible I misunderstand something). If both the deployment controller and the service/endpoint controller both wait for all the conditions to be ready then we still have the same problem. Hypothetically if we could tell a deployment to only proceed with one readiness gate and a service to only mark an endpoint ready on a different readiness gate then I think that would probably work. But I don't think that's currently possible? I actually can't find the documentation that says when an endpoint is added/removed from a service...

thockin:
On Wed, Sep 21, 2022 at 9:34 AM Matthew Mead-Briggs <
***@***.***> wrote:

> I think probably it doesn't (but certainly possible I misunderstand
> something). If both the deployment controller and the service/endpoint
> controller both wait for all the conditions to be ready then we still have
> the same problem. Hypothetically if we could tell a deployment to only
> proceed with one readiness gate and a service to only mark an endpoint
> ready on a different readiness gate then I think that would probably work.
> But I don't think that's currently possible? I actually can't find the
> documentation that says when an endpoint is added/removed from a service...
>

Endpoints are added when the corresponding pod is ready, as indicated by
the Ready condition.  I think I see what you are asking for - you want pods
to be included in the service when they are passing readiness probes, but
you don't want rollouts to proceed until that service is confirmed
"healthy" in some statistical manner.  Is that right?

One way to do that would be to set `Service.spec.publishNotReadyAddresses`
but that is not REALLY what you want.

The problem is that readiness gates are material to the pod Ready
condition.  We can't really add a pod to a Service before Ready (e.g. as
soon as readiness probe passes), because we don't know what readinessGates
might actually mean.

So what you'd really need is something in ReplicaSet (and Deployment, which
configures RS) which says "which pod condition should I look at to
determine pod-ready" (the proverbial "another level of indirection").  Then
you could set that to a "service is happy" indicator, which you can set by
your own heuristics.  Looking at the code for RS controller, that actually
seems quite simple to implement, but I am not sure what other
considerations are required - that subsystem is not my expertise.  We would
need a sig-apps KEP, for sure, though it should be pretty short
(relatively).

Tim

mattmb:
> Endpoints are added when the corresponding pod is ready, as indicated by
the Ready condition.  I think I see what you are asking for - you want pods
to be included in the service when they are passing readiness probes, but
you don't want rollouts to proceed until that service is confirmed
"healthy" in some statistical manner.  Is that right?

Yep exactly, and indeed I think it depends on what service mesh or routing people use as to what that check would be. But the overall problem in my mind is generic to any mesh technology

> So what you'd really need is something in ReplicaSet (and Deployment, which
configures RS) which says "which pod condition should I look at to
determine pod-ready" (the proverbial "another level of indirection").  Then
you could set that to a "service is happy" indicator, which you can set by
your own heuristics.  Looking at the code for RS controller, that actually
seems quite simple to implement, but I am not sure what other
considerations are required - that subsystem is not my expertise.  We would
need a sig-apps KEP, for sure, though it should be pretty short
(relatively).

I think that sounds pretty sensible and would work well for our case. I'll have some discussions about this, I'm not sure how urgently we need to fix our current approach but I suspect we'll have to tackle this eventually so would be nice to get it described in a KEP and contribute back if we can.

thockin:
On Wed, Sep 21, 2022 at 10:48 AM Matthew Mead-Briggs <
***@***.***> wrote:

> Endpoints are added when the corresponding pod is ready, as indicated by
> the Ready condition. I think I see what you are asking for - you want pods
> to be included in the service when they are passing readiness probes, but
> you don't want rollouts to proceed until that service is confirmed
> "healthy" in some statistical manner. Is that right?
>
> Yep exactly, and indeed I think it depends on what service mesh or routing
> people use as to what that check would be. But the overall problem in my
> mind is generic to any mesh technology
>
> So what you'd really need is something in ReplicaSet (and Deployment, which
> configures RS) which says "which pod condition should I look at to
> determine pod-ready" (the proverbial "another level of indirection"). Then
> you could set that to a "service is happy" indicator, which you can set by
> your own heuristics. Looking at the code for RS controller, that actually
> seems quite simple to implement, but I am not sure what other
> considerations are required - that subsystem is not my expertise. We would
> need a sig-apps KEP, for sure, though it should be pretty short
> (relatively).
>
> I think that sounds pretty sensible and would work well for our case. I'll
> have some discussions about this, I'm not sure how urgently we need to fix
> our current approach but I suspect we'll have to tackle this eventually so
> would be nice to get it described in a KEP and contribute back if we can.
>

Awesome.  I'm happy to look at a KEP, though I should not be approver for
it, probably.  Honestly, it feels like one field on RS, one on Deployment,
and maybe one on StatefulSet.  Then some plumbing work to pass the
indirection around and some good devex (events, logs, etc) and tests.

PR #363, opened 2017-02-13
Proposal: Server-side `kubectl get`

smarterclayton:
This proposal discusses how the `kubectl get` operation can be moved to the server to allow multiple clients to more easily show simple lists of objects.

Part of https://github.com/kubernetes/kubernetes/issues/12143 and general work under api-machinery.

Depends on:
* https://github.com/kubernetes/community/pull/123

Related to:
* https://github.com/kubernetes/kubernetes/issues/24855

Prototype:
* https://github.com/kubernetes/kubernetes/pull/40848

Some WIP items left, including a sketch of describe to contrast the two approaches.

PR #481, opened 2017-03-24
Proposal - kubectl commands for the Service Catalog

fabianofranz:
Proposes how `kubectl` will interact with the API for the [Service Catalog](https://github.com/kubernetes-incubator/service-catalog/). 

Trying to cover all described [use-cases](https://github.com/kubernetes-incubator/service-catalog/blob/master/docs/v1/use-cases.md), some by the use of existing commands and some by the introduction of new commands. TBD if new ones are going to be part of the main binary, or [plugins](https://github.com/kubernetes/kubernetes/pull/37499).

@kubernetes/sig-cli-proposals @smarterclayton @pmorie @jwforres looking for feedback.

### DIFF
--- a/contributors/design-proposals/sig-cli/kubectl-service-catalog.md
+++ b/contributors/design-proposals/sig-cli/kubectl-service-catalog.md
@@ -1,0 +1,227 @@
+# MVP - kubectl commands for the Service Catalog
+
+Status: Pending
+
+## Motivation
+
+The client binary for the service catalog must cover every [use case](use-cases.md) and provide 
+tools for every [persona](use-cases.md#personas) described. 
+
+While in some use cases we will most likely rely on the base commands already present on `kubectl` - like 
+`create`, `get`, `describe`, and `delete` - in others we want to have subcommands dedicated 
+to interacting with the service catalog. It is expected that the "end-users" (the *Service Consumer* 
+persona, for example), have a set of subcommands that are simple to use and allow them to easily check 
+information about the services available on a catalog, request instances, bind them to apps, and 
+manage their installed instances in a simple and user-friendly way.
+
+This document proposes the way every use case is fulfilled by `kubectl` subcommands and/or extensions.
+
+## Proposal
+
+Below the [identified personas](use-cases.md#personas) and the commands that are expected to fulfill the use
+cases for each one of them.
+
+Note that when relying on existing commands, some improvements to printers and describers of the involved 
+resources might be needed. For example, in the describer of a service instance we might want to display the 
+number of bindings that currently exists and other information useful to catalog operators.
+
+### Service consumer
+
+For this persona we want simple and user-friendly commands. Service consumers must be able to easily search the entire
+catalog for services available, check the metadata for each service, request an instance of a given service and
+bind it to apps, only through the command line without the need of json or yaml files. 
+
+So the proposal is to have a new set of subcommands called
+
+```
+kubectl catalog
+```
+
+which could be aliased as `kubectl (catalog | servicecatalog | service-catalog)` and distributed as part of the main 
+`kubectl` binary or, preferably, as `kubectl` [binary plugins](#distribution) 
+(TBD, this feature is under discussion and not yet available).
+
+Note that `instances` and `bindings` are user-scoped, so they are applied in the context of current namespace (or what's
+provided by the `--namespace` flag, etc).
+
+The list of proposed subcommands under `kubectl catalog` is:
+
+#### List and search the service catalog and existing service instances
+
+```
+Usage:
+
+kubectl catalog list (available | instances) [--search=TERM]
+
+Examples:
+
+# List all services in the catalog available to be instantiated
+$ kubectl catalog list available
+
+## List all service instances in use by current namespace
+$ kubectl catalog list instances
+
+# List all service instances in use by the "mine" namespace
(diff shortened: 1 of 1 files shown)


davidopp:
Sorry for the drive-by question, but I was wondering: do service instances run in the cluster belonging to the application that is consuming the service, or in a cluster belonging to the service provider (broker)?

jwforres:
@davidopp short answer: yes to both, it depends on the broker.

The ServiceInstance object is a namespaced resource that tells the service catalog to talk to a particular broker to provision a service.  That broker may then choose to create things in that same namespace, in a different namespace, in a different cluster, or may not actually be creating any k8s resources at all (e.g. provision just creates an account/space for you with some third party SaaS provider)

The Binding object is what then makes the ServiceInstance consumable by your application by providing the connection info / credentials / etc.

More info here https://github.com/kubernetes-incubator/service-catalog#introduction

contributors/design-proposals/sig-cli/kubectl-service-catalog.md
```diff
@@ -0,0 +1,227 @@
...
+catalog for services available, check the metadata for each service, request an instance of a given service and
+bind it to apps, only through the command line without the need of json or yaml files. 
+
+So the proposal is to have a new set of subcommands called
+
+```
+kubectl catalog
+```
+
+which could be aliased as `kubectl (catalog | servicecatalog | service-catalog)` and distributed as part of the main
```
deads2k: I don't think we start allowing our "favorite" non-core projects into `kubectl`.  A binary-plugin seems more appropriate.

philips: I just want to echo the sentiment that this proposal should only consider a binary plugin not making it built-in.

fabianofranz: Makes sense to me, we'll have V0 for plugins [pretty soon](https://github.com/kubernetes/kubernetes/pull/37499) so this looks like a great first use of it.

jbeda:
One comment -- it might be useful for everyone to build a CLI for the service broker outside of kubectl as a way to iterate on the UX.  This can then be converted to a plug-in as that becomes available.  That way if the plug-in mechanism doesn't work there is *something* that works here.  While not 100% the same, both `kubeadm` and `kubefed` are separate CLIs for this reason.

contributors/design-proposals/sig-cli/kubectl-service-catalog.md
pmorie: @deads2k since binary plugins shouldn't be able to change the meaning of primitives like 'get' and 'describe', do you have in mind that we will have separate subcommands for these, like:

```
kubectl service-catalog get
kubectl service-catalog describe
```

I would strongly prefer to avoid doing this, and just make `get` and `describe` driven off of openapi/swagger so that we would get basic functionality for free in any API server that serves swagger/openapi.

cc @fabianofranz

pmorie:
@jbeda 

>  CLI for the service broker 

A service broker is something that offers a set of services that the catalog consumes.  We want to build a CLI for the catalog, not a broker.

contributors/design-proposals/sig-cli/kubectl-service-catalog.md
liggitt: three initial categories of additions make sense to me:
* get/describe for custom objects accomplished server-side - https://github.com/kubernetes/community/pull/363
* generators for object-specific creation nested under `kubectl create <mykind>` that support all the conventions of the other `kubectl create` generators (--dry-run, different --output formats, etc). This seems like it should a really structured plugin (plugin can be interrogated for args/flags/usage/help info, user invokes kubectl, kubectl invokes plugin with args/flags, plugin outputs yaml or json that kubectl can transform/create/output in all the standard ways)
* completely independent commands that nest within kubectl

pmorie:
I think this should be moved to the service-catalog incubator repo, and discussed there.  Since we have kubectl extensions now, we can package it as a product of that repository.  There's a lot of interest in this in service catalog SIG.

pwittrock:
Why do we need `servicecatalog get`?  Can't we use extension mechanisms to get what we need?  Same question for describe.

pmorie:
I have a feeling that this proposal is somewhat outdated, as the generic resource support has improved a lot in kubectl.  Perhaps I'm wrong.

fabianofranz:
> I have a feeling that this proposal is somewhat outdated

Definitely, I'll work on a review and moving this to the appropriate repo and having in mind the [WIP implementation](https://github.com/kubernetes-incubator/service-catalog/pull/840).

pmorie:
@fabianofranz awesome, thanks a million

pmorie:
should we close this for now?

fabianofranz:
BTW, I'm having some trouble finding a way to expose the API to plugins that finds wide acceptance by other groups. The proposal originally ([PR here](https://github.com/kubernetes/kubernetes/pull/47151)) wanted to expose it through a proxy (by using the underlying `kubectl proxy` mechanism) but it's not being well accepted due to security concerns. 

So most likely, at least for now plugins will be responsible to access the API by themselves (using one of the Kube clients), or by invoking the `kubectl` binary.

I'll open a PR to the [WIP implementation](https://github.com/kubernetes-incubator/service-catalog/pull/840) reflecting that.

PR #483, opened 2017-03-25
[Proposal] deferContainers - Pod Termination Semantics

dhilipkumars:
Proposal for deferContainers a complimenting feature to init-containers that enable graceful pod termination. 

@quinton-hoole @deepak-vij @shashidharatd @irfanurrehman @kevin-wangzefeng

@smarterclayton @kargakis @erictune @kow3ns @hectorj2f @davidopp 

It wold be great to get your feedback on this, also could someone tag sig-apps please.

### DIFF
--- a/contributors/design-proposals/defer-containers.md
+++ b/contributors/design-proposals/defer-containers.md
@@ -1,0 +1,248 @@
+# Defer Containers
+
+dhilip.kumar.s@huawei.com
+
+March 2017
+
+## Prerequisite
+This is a continuation of [initContainers](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/container-init.md)
+
+## Motivation
+Since the introduction to Statefulset and Daemonset PODs that consist of stateful containers are now fully supported by
+Kubernetes.  Such stateful pods benefit from constructor and destructor semantics. InitContainers provide constructor
+semantics. The proposed DeferContainers will provide destructor semantics for such stateful PODs
+
+## Abstract
+This introduces the concept of ‘deferContainers’, inspired by golang’s keyword ‘defer’. This will allow one to define a
+set of Containers to be executed at the time of POD termination. The defined containers will be executed sequentially and
+in the specified order. It will bring destructor() capability to Pods.
+
+## Use Cases
+
+### Cleanup:
+Most stateful workloads require certain Cleanup activity before or after the appContainers are terminated such as
+* Sync/Flush the Disk or logs before the POD is evicted off a node.
+* Delete or Update Global Configuration / Release Application level Lock.
+* Some Legacy applications may not respond to SIGTERM signals sent by docker daemon, they might require special command/procedure
+  to initiate graceful shutdown.
+
+### Shard rebalancing.
+In a sharded workload configured as a statefulset, if the application is scaled down, then the exiting Pod should trigger
+a re-balance of keys from the current shard to the rest of them. This needs to be done befoe the shard goes down.  
+‘deferContainers’ could handle such scenarios with ease.
+
+### Statefulset Upgrading
+If in future, statefulsets support rolling updates, an update request might attempt to replace Pods one after the other. 
+With deferContainers each Pod could be removed gracefully without having to implement complicated sidecar logic to prevent data loss.
+
+### Master-Slave or Leader-Follower Statefulset down-size
+Consider a Stateful set consisting a Master-Slave or Leader-follower type application running N replicas. Ordinal index 
+ranging from 0 to N-1.  If you scale it down (reduce the number of replicas), the statefulset controller would attempt 
+to bring down the last replica (N-1th pod in this case).  If that pod is the elected Leader or Master, then this would 
+disrupt the service intermittently. With deferContainers, this handover or re-election could be programmed gracefully/elegantly.
+
+### SHUTDOWN Sequence
+* Traditional relational  databases typically support a reasonable shutdown sequence, for instance, Oracle has 4 types of 
+shutdown such as NORMAL, IMMEDIATE, TRANSACTIONAL and ABORT. ‘deferContainers’ will allow us to program and wait for such
+complex shutdown scenarios. 
+* In future when kuberenetes supports Virtual Machine runtime (eg: hyperContainer) for better isolation, we should shutdown
+  the VMs instead of killing them abruptly.  ‘deferContainers’ could help us run such shutdwon commands.
+
+## Limitation with the current system
+Container pre-stop hooks are not sufficient for all termination cases:
+* They cannot easily coordinate complex conditions across containers in multi-container pods
+* They can only function with code in the image or code in a shared volume, which would have to be statically linked 
+  (not a common pattern in wide use)
+* They cannot be implemented with the current Docker implementation
+
+## Design Requirements
+Most of the requirements are very similar to initContainers.  They are replicated and modified as necessary.
+* deferContainers should be able to:
+  * Use the same volume (PV) as appContainers such that it can
+    * Perform cleanup of shared volume, such as delete several temp directories.
+    * Delete unwanted files before a final sync is initiated.
+    * Update Configuration files about the changes in the distributed system so that the next pod getting attached to this PV 
+	  will benefit from it. (like a new leader/master etc).
+    * Deleted secrets or security related files before the pod de-couples from the PV.
+  * Delay the termination of application containers until operations are complete
+  * De-Register the pod with other components of the system
+
+* Reduce coupling:
+  * Between application images, eliminating the need to customize those images for Kubernetes generally or specific roles
+  * Inside of images, by specializing which containers perform which tasks (install git into init container, use filesystem 
+    contents in web container)
+  * Between termination steps, by supporting multiple sequential cleanup containers
+* Pre-Exit and Post-Exit workflow
+  * Post - specify that the controller can continue to delete the appContainers but wait for deferContainer’s to complete 
+    its execution before marking as the Pod is deleted.
+  * Pre - specify that the controller should wait until deferContainers completes its execution first and then proceed to 
+    delete the appContainers if needed
+* deferContainers should allow us to specify
+  * if certain containers in deferContainer list need to be re-started on failures.
+* GracePeriod
+  * It should be possible to mention overall terminationGracePeriod for defercontainers.
+* Run-once and run-forever pods should be able to use deferContainers
+* Reduce Complexity, 
+  * It should be possible to use a generic container as a deferContainer,
+  * A deferContainer should be independently invokable, ie:- should not require code in the same image as the appContainers.
+
+## Design
+This proposed pod spec would look like below.
+```yaml
+pod:
+  spec:
+    initContainers: ...
+    containers: ...
+    deferContainers:
+    - name: defer-container1
+      image: ...
+      ...
+    - name: defer-container2
+    ...
+  status:
+    initContainersStatuses: ...
+    containerStatuses: ...
+    deferContainerStatuses:
+    - name: defer-container1
+      ...
+    - name: defer-container2
+      ...
+```
+Most of the design elements are similar to initContainers such as below
+* Will have 0...N containers and will be executed in sequence (specified order)
+* Restart policy could be either ‘restartAlways’ or ‘restartNever’
+* Adding a new phase ‘Terminating’ in the Pod lifecycle
+
+### Terminating Phase / Defer Phase
+* A POD reaches terminating phase when it is about to be removed from Kubernetes Cluster
+* During this phase, the appContainers are not restarted if they get terminated/killed
+* deferContainers will be executed one-after-the-other in the same sequence as they were specified in the POD Spec. From the above
+  pod spec example execution sequence will be defer-Container1, then defer-Container2, …, etc.,
+* we will either move-on from a failed deferContainer or restart it depending upon its restart policy, the default behavior will be 
+  restartNever
+* if the user specifies `kubectl delete pod foo --grace-period=0 --force` to delete a pod deferContainers will not be executed.
+```
+Example status output when a pod is being terminated.
+   NAME      READY     STATUS     RESTARTS   AGE
+    foo-0     2/2       Defer:0/4   0          7m
+```
+* Failure of one or all deferContainers will not trigger a POD restart.
+
+### TerminationGracePeriod
+If this flag is Specified along with deferContainers the grace period is applied to deferContainers execution time rather than appContainers.
+If terminationGracePeriod is reached currently executing deferConatiner will be terminated, all the appContainers will be terminated (if not killed already).
+
+### Pre/Post Termination triggers
+All the deferContainers will behave like a preExit trigger, it should be easier program deferContainers in such a way that it does both 
+pre and post Termination tasks.
+```yaml
+pod:
+  spec:
+    terminationGracePeriod: 60
+    initContainers: ...
+    containers: ...
+    deferContainers:
+#pre exit operations
+    - name: remove-shard
+      image: dbUtils
+      command: ["/bin/sh", "-C", "remove-shard", "--name=${POD_NAME}"]
+    - name: wait-for-rebalance
+      image: dbUtils
+      command: ["/bin/sh", "-C", "while[[ 1 ]];do sleep 1; if key_rebalance_complete.sh; then exit 0; fi; done"]
+    - name: kill-db
+      image: dbUtils
+      command: ["/bin/sh", "-C", "shutdown-db", "--name=${POD_NAME}"]
+#post exit operation
+    - name: clean-update
+      image: dbUtils
+      command: ["/bin/sh", "-C", "./disk_cleanup.sh"]
+      ...
+```
+### Short running Job / Pod
+For PODs which run and exit gracefully themselves if deferContainers are configured they will act as PostExit triggers, 
+an internal flag should indicate if  deferContainers have already been called or not.
+
+## Implementation Plan
+Development and release lifecycle of this feature will follow other kubernetes experimental feature. This will be originally
+rolled out as alpha as annotations based on its usefulness and community feedback it will graduate to PodSpec.
+
+## Examples 
+### Cleanup
+```yaml
+pod:
+  spec:
+    terminationGracePeriod: 60
+    initContainers: ...
+    containers: ...
+    deferContainers:
+#pre exit operations
+    - name: fsync
+      image: my-utils
+#Something like this https://docs.mongodb.com/manual/reference/command/fsync/
+      comamnd: ["/bin/sh", "-C", "./syncDisk.sh"] 
+      name: killApp
+      image: my-utils
+      command: ["/bin/sh", "-C", "./killAndWait.sh", "--name=${POD_NAME}"]  
+#post exit operation
+    - name: rm-tmpdir
+      image: my-utils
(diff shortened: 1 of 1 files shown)


kargakis:
@kubernetes/sig-apps-feature-requests  @kubernetes/sig-node-feature-requests

erictune:
@kubernetes/sig-node-api-reviews would need to comment on this too since it would be implemented in the node.

contributors/design-proposals/defer-containers.md
```diff
@@ -0,0 +1,248 @@
...
+      image: db-utils
+      #run a script that will wait until the DB is down.
+      command: ["/bin/sh", "-C", "./waitForDB.sh"]  
+      ...
+```
+
+## Caviate
+This Design primarily focuses on handling graceful termination cases.  If the Node running a deferContainer configured pod
+crashes abruptly, then this design does not guarantee that cleanup was performed gracefully.  This still requires community
+feedback on how such scenarios are handled and how important it is for deferContainers to handle that situation.
```
pigmej: I think this is quite crucial. In case of initContainers you have a guarantee if container is running => Init container was executed and it executed successfully. In case of deferContainers we have *no* guarantees about that. Not to mention fetching logs from deleted pods (no pod => no pod name in case of controllers => no way to get deferContainer logs).

Probably all of that has been already mentioned somewhere else, but I wanted to raise it also here.

contributors/design-proposals/defer-containers.md
```diff
@@ -0,0 +1,248 @@
...
+semantics. The proposed DeferContainers will provide destructor semantics for such stateful PODs
+
+## Abstract
+This introduces the concept of ‘deferContainers’, inspired by golang’s keyword ‘defer’. This will allow one to define a
+set of Containers to be executed at the time of POD termination. The defined containers will be executed sequentially and
+in the specified order. It will bring destructor() capability to Pods.
+
+## Use Cases
+
+### Cleanup:
```
pigmej: How are we going to handle situation:
- you want to delete pod because it's broken or whatever
- it has deferContainer that will hung because of "some reasons"

How am I supposed to *not* wait for deferContainer? Are we going to add `kubectl delete .... --ignore-defer-containers`?

dhilipkumars: was thinking we should probably force the delete in that case we wont need to run deferContainers
```
`kubectl delete pod foo --grace-period=0 --force
```
But this new option is cool too, it seperates out the funtionality.

contributors/design-proposals/defer-containers.md
dhilipkumars: Honestly, we could not think of an elegant way to solve this problem, we really need to get communities feedback on how this is done today but we believe even without this deferContainers would be useful in many cases.

contributors/design-proposals/defer-containers.md
liggitt: doesn't backwards compatibility mean that if someone previously ran `delete pod foo --grace-period=0` to immediately terminate without waiting, that should skip deferContainers as well?

dhilipkumars: then it might force `--grace-period` to all the other deferContainers, predicting the running-time of a certain cleaup/termination activity might be difficult.  Or may be we could have `--grace-period=-1` for unlimited termination time for a POD?

contributors/design-proposals/defer-containers.md
```diff
@@ -0,0 +1,248 @@
...
+### SHUTDOWN Sequence
+* Traditional relational  databases typically support a reasonable shutdown sequence, for instance, Oracle has 4 types of 
+shutdown such as NORMAL, IMMEDIATE, TRANSACTIONAL and ABORT. ‘deferContainers’ will allow us to program and wait for such
+complex shutdown scenarios. 
+* In future when kuberenetes supports Virtual Machine runtime (eg: hyperContainer) for better isolation, we should shutdown
+  the VMs instead of killing them abruptly.  ‘deferContainers’ could help us run such shutdwon commands.
+
+## Limitation with the current system
+Container pre-stop hooks are not sufficient for all termination cases:
+* They cannot easily coordinate complex conditions across containers in multi-container pods
```
kow3ns: Can you provide a specific use case where the current Pod hooks are not sufficient?

contributors/design-proposals/defer-containers.md
```diff
@@ -0,0 +1,248 @@
...
+* They can only function with code in the image or code in a shared volume, which would have to be statically linked
```
kow3ns: When would I want to pull down new code to turn down a storage system. If the primary use cases are those expressed [above](https://github.com/kubernetes/community/pull/483/files#diff-a658ee76ff92bc1cfac0f3963b7e06a2R24), isn't all the code in my container already?

contributors/design-proposals/defer-containers.md
```diff
@@ -0,0 +1,248 @@
...
+complex shutdown scenarios. 
+* In future when kuberenetes supports Virtual Machine runtime (eg: hyperContainer) for better isolation, we should shutdown
+  the VMs instead of killing them abruptly.  ‘deferContainers’ could help us run such shutdwon commands.
+
+## Limitation with the current system
+Container pre-stop hooks are not sufficient for all termination cases:
+* They cannot easily coordinate complex conditions across containers in multi-container pods
+* They can only function with code in the image or code in a shared volume, which would have to be statically linked 
+  (not a common pattern in wide use)
+* They cannot be implemented with the current Docker implementation
```
kow3ns: Can you please elaborate on what is meant by this?

kow3ns:
If I understand your design, it seems you want to move to a model where we run deferContainers during terminationGracePeriod seconds to provide a more eloquent abstraction than handling cleanup from the containers entry point.
1. Other than being "more eloquent," isn't this functionally equivalent with what we can do now with the exception that you could potentially pull down new code in your deferContainer?
1. When would you need to pull down a new image to cleanup an existing image? I think its a little dangerous to not package your cleanup code in the same container as the application. If you have an image pull failure for your defer container, you will not be able to cleanup successfully.
1. I would think running commands in the same container to clean up a stateful application would be a more beneficial abstraction than pulling down a new image.
1. deferContainers will make it more difficult to reason about terminationGracePeriod's. You will now have to budget image pull time into grace period. This can be highly variable based on the environment.

contributors/design-proposals/defer-containers.md
dhilipkumars: The objective is not always to pull a new image, the objective is to seperate out/de-couple Termination sequence. Most often we wont have to pull a new image all togather.  I can also imagine simple cleanups like 'rm -rf /tmp/dir1' wrapped around bash or busyboxy images which are lighter.

If your pod is part of a CI engine and running 1000s of testcases, it should be easy to configure a deferContainer (from a different image/source) simply upload (curl POST) currently processed test results to a different server when the POD is getting evicted out of that node.  
If curl is not available in your primary container you dont have to re-package and create a new image just to run curl at the termination - which will be the case for preStop hooks.

dhilipkumars:
**_1. Other than being "more eloquent," isn't this functionally equivalent with what we can do now with the exception that you could potentially pull down new code in your deferContainer?_**

Yes de-couple the termination logic, handle complex termination sequence, upgrade/change the primary container images without affecting initialization or termination scripts.  More predictable termination behaviour

**_2. When would you need to pull down a new image to cleanup an existing image? I think its a little dangerous to not package your cleanup code in the same container as the application. If you have an image pull failure for your defer container, you will not be able to cleanup successfully._**

It is not always necessary to pull a new image for cleanup,  Most cleanups might use the same image but just that now it is more predictable.  Modularize the termination sequence and ability to restart only failed steps.   Currently, if preStop hook failed the (only way to trigger cleanup) nothing much can be done.  Imagine in a master-slave type workload a terminating master is promoting a slave and if that command failed, it is important to retry this command.  The user has to implement logic to handle this functionality.  Below code doesnt attempt to retry preStop hooks if failed. 
```go
// executePreStopHook runs the pre-stop lifecycle hooks if applicable and returns the duration it takes.
func (m *kubeGenericRuntimeManager) executePreStopHook(pod *v1.Pod, containerID kubecontainer.ContainerID, containerSpec *v1.Container, gracePeriod int64) int64 {
	glog.V(3).Infof("Running preStop hook for container %q", containerID.String())

	start := metav1.Now()
	done := make(chan struct{})
	go func() {
		defer close(done)
		defer utilruntime.HandleCrash()
		if msg, err := m.runner.Run(containerID, pod, containerSpec, containerSpec.Lifecycle.PreStop); err != nil {
			glog.Errorf("preStop hook for container %q failed: %v", containerSpec.Name, err)
			m.generateContainerEvent(containerID, v1.EventTypeWarning, events.FailedPreStopHook, msg)
		}
	}()

```
**_3. I would think running commands in the same container to clean up a stateful application would be a more beneficial abstraction than pulling down a new image._**

deferContainers would allow us to run an arbitrary set of containers during termination (preStop) for the entire pod.  If one develops an operator for a particular TRP workload which has reasonable termination sequence then one can use the primary image like redis or [oracle](https://hub.docker.com/_/oraclelinux/) directly from the source instead of repackaging that image with the termination scripts,  without affecting each other the terimnation scripts/sequences can be improved and the primary image can be upgraded. 

**_4. deferContainers will make it more difficult to reason about terminationGracePeriod's. You will now have to budget image pull time into grace period. This can be highly variable based on the environment._**

I was thinking we should add a seperateflag to handle such rare cases where one would want to use a different image which is very heavy. 
```yaml
prePullDeferImages: true
```
such that as soon as the POD is started kublet can download and pre-populate these heavy images in the node?

contributors/design-proposals/defer-containers.md
dhilipkumars: The container hooks are not generic to POD itself they are specific to containers, imagine a multi-container pod with 1 App Container and 1 Side Car 
SideCar : Membership/Gossip Container
AppContainer: A Stateful Sharded workload
Termination Sequence: First one needs to re-shard/rebalance and then come out of membership
Below code suggests that both the preStop hooks (if configured for both the containers) will be started in parllel
```go
for _, container := range runningPod.Containers {
		go func(container *kubecontainer.Container) {
			defer utilruntime.HandleCrash()
			defer wg.Done()

			killContainerResult := kubecontainer.NewSyncResult(kubecontainer.KillContainer, container.Name)
			if err := m.killContainer(pod, container.ID, container.Name, "Need to kill Pod", gracePeriodOverride); err != nil {
				killContainerResult.Fail(kubecontainer.ErrKillContainer, err.Error())
			}
			containerResults <- killContainerResult
		}(container)
	}
```
1) hard to co-ordinate if both have different preStop hooks, 
2) hard consolidate this logic to one preStop script if both are from different images. Individually upgrading the images becomes a problem
3) If only one preStop command failed only that should be retried and not the whole script(currently preStop hooks restarts are not available user shoudl include them in the script)

kargakis:
> if preStop hook failed the (only way to trigger cleanup) nothing much can be done. Imagine in a master-slave type workload a terminating master is promoting a slave and if that command failed, it is important to retry this command. The user has to implement logic to handle this functionality. Below code doesnt attempt to retry preStop hooks if failed.

Can't we fix pre-stop hooks to be retriable?

dhilipkumars:
```
Can't we fix pre-stop hooks to be retriable?
```
yes in my opinion that would make preStop hook more functional but still the termination script is tighgtly coupled with the AppImage and are container specific hooks.

stp-ip:
With this proposal we definitely make termination logic more descriptive and extensive. I do like the syntactic ideas as they are similar to init containers for startup logic.

The main issue I have with this proposal is yet another way to handle something. If deferContainers are accepted, we should think about "deprecating" or at least advise against preStop hooks.

smarterclayton:
deferContainers wouldn't obviate the need for preStop hooks - for instance,
anything that has to manage per container shutdown state can't be done from
outside the container.

I'll have more time to review this later this week, but in general
termination guarantees are a lot harder than initialization guarantees (and
it's taken a while to ensure that initialization happens correctly).  It's
best to focus on use cases that can't be accomplished any other way.

contributors/design-proposals/defer-containers.md
pigmej: I think adding this without any idea how to solve that problem is bad idea, it may create some illusion that some functionality is provided, while on production it may work not as "expected". ( compared with initContainers).

contributors/design-proposals/defer-containers.md
pigmej: How are we going to solve "force shutdown" "force shutdown + ignore defer"? Pls, don't get me wrong but it seems to be a bit inconsistent.

contributors/design-proposals/defer-containers.md
dhilipkumars: Termination constructs are important but providing absolute guarantee like their initialization counterparts are difficult to achieve, especially when there are ungraceful terminations like node crash.

In programming languages C++'s destructor, golang's defer(), [atexit](http://man7.org/linux/man-pages/man3/atexit.3.html) in C.  None of this is guaranteed to work if the process crashes (ungraceful termination).  This does not mean they are not useful.  Their initialization counterpart offers higher level of guarantee.  

We think the current design is a good starting point for providing better support for graceful POD termination. We might be able to improve it along the way as the users try it and provide more feedback.  Initial implementation will be alpha status so should caution community from trying this in Production without enough tests.

What do you think?

contributors/design-proposals/defer-containers.md
dhilipkumars: Agreed.  I now see the inconsistency,  to keep it more predictable we should stick with, 
```
kubectl delete pod foo --grace-period=0 --force
```
as [originally](https://github.com/dhilipkumars/community/blob/7532950e04d8963e74f6b555877a686ce1aafdac/contributors/design-proposals/defer-containers.md#terminating-phase--defer-phase) proposed.  Even now this is the way to avoid running PreStop hooks.

contributors/design-proposals/defer-containers.md
dhilipkumars: Currently, Docker does not allow us to configure any termination hooks.  I think its better to re-phrase or remove it to avoid confusion.

dhilipkumars:
Ref https://github.com/kubernetes/kubernetes/issues/3585

contributors/design-proposals/defer-containers.md
smarterclayton: We'd have to come up with some other signal.  --grace-period=1 means "delete fast" and --grace-period=0 means "give up strong consistency and risk split brain".  You don't want to do the latter.

I'm not sure there is a backward compatible way to allow defer containers to run without being bounded in time by the grace-period.  I'm also not sure we want defer containers to run unbounded, since that was a key part of the graceful deletion design.

dhilipkumars: @smarterclayton I have added some design thoughts on how we handle time boundness for deferContainers,  this was something i wanted to bring up in the SIG-APPs meeting yesterday.   PTAL. 

Also this particular discussion is about deleting the POD without running deferContainers or preStopHooks,  an equivalent of `kill -9 PID`.  Could you please elaborate why should we come up with a different signal for this?

dhilipkumars:
/assign @erictune 
Hi Eric and Clayton, 
   It would be great to get your attention on this proposal this week.
Thanks in advance

kargakis:
Dropping in a couple of links because this seems to be related to how people today manage to run zero-downtime deployments:
kubernetes/kubernetes#43576
kubernetes/ingress#322
kubernetes/kubernetes#20473

timoreimann:
What's the status of this proposal? From what I understand, it has the potential to simplify the termination process significantly as it can be outsourced to a dedicated pod. I'd love to have that.

dhilipkumars:
@timoreimann Not outsourced to a separate POD but a container in the same POD.  Very similar to [initContainers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/).

timoreimann:
@dhilipkumars yes, of course, typo. New Year's Eve is taking its toll on me. :-) Thanks for the correction.

contributors/design-proposals/defer-containers.md
kfox1111: What about renaming it an eventContainer, and then passing in event kind as an env var?
something like: POD_EVENT=stop, POD_EVENT=recover, etc. We only need to implement the first, stop, for now, but can reuse the machinery later for things like, maybe informing a statefulset pod it will be shutting down but will be replaced. so like, in cephs case, the statefulset might want to set no-out during the upgrade as it will come right back. but not set no-out during a delete as it might not come back and the cluster should recover.

lukasheinrich:
Just joining the discussio to add a use case: https://github.com/kubernetes/kubernetes/issues/62117#issuecomment-378649712

For big-data analysis scenarios (such as higher-energy physics) the data management concerns and the actual data-processing concerns are tended to by different teams and separating them is very useful. The sequence we would like to implement as part of a k8s `Job` object, is

```
initContainer: copy data from object storage into a local, fast I/O disk attached as a a dynamically provisioned volume
main Container: a FROM scratch container consisting only of a binary that an operate on the local data via POSIX, without having to have support / or knowledge / credentials to the object store from which the data originated
deferContainer: a container that takes output files written to the local disk by the main container and uploads it back to object storage
```

crucially, this would allow the data management teams to independently develop their stage-in/stage-out containers including authn/authz while keeping the ability to keep the main workload container as lean as possible (in an extreme case a statically compiled binary packaged as `FROM scratch`) and oblivious to authentication / credentials etc.

lukasheinrich:
Hi, I discussed this isse with a couple of people at KubeCon, how can we move this forward?

idealhack:
/reopen

I'm not sure if we need to translate this proposal into a KEP to move it forward.

smarterclayton:
I’m not convinced we have enough reason to implement this at this time.  It
increases pod complexity in a direction we have historically resisted
moving pods.  I would prefer to focus on the other smaller proposals around
sidecars and pod restart logic in the near term, then come back to this.

On Nov 21, 2018, at 3:13 AM, k8s-ci-robot <notifications@github.com> wrote:

@dhilipkumars <https://github.com/dhilipkumars>: The following test *failed*,
say /retest to rerun them all:
Test name Commit Details Rerun command
pull-community-verify 9e81de3
<https://github.com/kubernetes/community/commit/9e81de385df8355af78b32ec0dd596cecda86a8a>
link
<https://gubernator.k8s.io/build/kubernetes-jenkins/pr-logs/pull/community/483/pull-community-verify/4051/>
/test
pull-community-verify

Full PR test history <https://gubernator.k8s.io/pr/community/483>. Your PR
dashboard <https://gubernator.k8s.io/pr/dhilipkumars>. Please help us cut
down on flakes by linking to
<https://git.k8s.io/community/contributors/devel/flaky-tests.md#filing-issues-for-flaky-tests>
an open issue
<https://github.com/kubernetes/community/issues?q=is:issue+is:open> when
you hit one in your PR.

Instructions for interacting with me using PR comments are available here
<https://git.k8s.io/community/contributors/guide/pull-requests.md>. If you
have questions or suggestions related to my behavior, please file an issue
against the kubernetes/test-infra
<https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:>
repository. I understand the commands that are listed here
<https://go.k8s.io/bot-commands>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/kubernetes/community/pull/483#issuecomment-440574201>,
or mute the thread

furkanmustafa:
Dear @fejta-bot can you please reopen?
/reopen
/remove-lifecycle rotten

PR #485, opened 2017-03-28
Fix the link of /pkg/kubectl/cmd/logs.go

yu-song:


yu-song:
cc @gnufied @childsb

grodrigues3:
Thanks for fixing!

/lgtm

PR #486, opened 2017-03-28
Fix the error url show in  sig-storage/contributing.md

yu-song:
Fix broken links.


yu-song:
cc @gnufied @childsb @saad-ali

PR #487, opened 2017-03-28
Fix mount options proposal to be accurate

gnufied:
cc @childsb @saad-ali


gnufied:
@childsb can we merge this? This PR simply brings the design document inline with code that already is written and merged.

PR #491, opened 2017-03-29
Update buildcop instructions with how to force a build.

mml:
Update buildcop instructions with how to force a build.

### DIFF
--- a/contributors/devel/on-call-build-cop.md
+++ b/contributors/devel/on-call-build-cop.md
@@ -113,6 +113,8 @@ pre-requisites are being merged at a reasonable rate:
 
 * When the build is broken, roll back the PRs responsible ASAP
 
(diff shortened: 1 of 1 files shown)


krousey:
/lgtm

@spxtr Why isn't this job on a timer as well?

spxtr:
We don't actually want to give most build cops access to Jenkins. Really, if the build breaks then ping the test-infra oncall to kick it.
cc @fejta

> @spxtr Why isn't this job on a timer as well?

It might be because it will reupload to the same place which might race with e2e jobs that are pulling the latest build.

krousey:
@spxtr Is there a way to trigger a build if the previous one failed? I don't like that this job won't auto-recover from a failure like this. It's especially problematic on the weekend when build cop won't be watching this.

spxtr:
> Is there a way to trigger a build if the previous one failed?

Only Googlers can access Jenkins and manually trigger a build. This is bad, and we're moving towards a better world. Until we're there, just ask the test-infra oncall to kick it.

PR #495, opened 2017-03-30
Add new client library procedure proposal

mbohlool:
Adding approved proposal for new client libraries

@kubernetes/sig-api-machinery-misc


lavalamp:
LGTM

PR #498, opened 2017-04-01
devel: add recommended time to release process

philips:
I am simply copying Golang which has selected a reasonable time: https://groups.google.com/forum/#!topic/golang-announce/YOqTqcJtiJI


grodrigues3:
@kubernetes/sig-release-pr-reviews

tallclair:
OK, but depending on the urgency the announcement shouldn't be delayed a day to hit that time. I guess that's why it's recommended.
/lgtm

PR #499, opened 2017-04-02
Update aws_under_the_hood.md to clarify pod vs node instance

diwu1989:
I don't think it's the POD that's getting a /24 CIDR but rather the virtual machine itself, a POD has a single address, not a range.

Please close PR if my understanding is incorrect, thanks!


diwu1989:
see related PR https://github.com/kubernetes/kops/pull/852/files on this exact misunderstanding

justinsb:
Absolutely correct - thanks!

/lgtm

PR #501, opened 2017-04-03
Proposal: IAM networking proposal

justinsb:
Proposal to add to the networking rules introduced in https://github.com/kubernetes/kubernetes/pull/42192, to support IAM proxies like kube2iam.

A little hacky / convention-based right now, but I think that's the right first step.

cc @thockin @Q-Lee @mikedanese @cjcullen

### DIFF
--- a/contributors/design-proposals/networking-iam.md
+++ b/contributors/design-proposals/networking-iam.md
@@ -1,0 +1,125 @@
+# Network strategy for IAM well-known IPs
+
+## Introduction
+
+AWS and GCE both define a well-known IP, 169.254.169.254, on which a HTTP
+server that provides system information can be obtained.  Typically this
+includes a script for booting a machine, data about the instance type and
+configuration, and also often authorization information for access to the
+full cloud API.
+
+The metadata service information is therefore either sensitive or inaccurate
+for many pods.  Sensitive because credentials are intended for the node, not
+the pod; and inaccurate because the metadata might report more memory than the
+pod has available.  As such it is often desirable to block access to the
+metadata server.  In addition it can be helpful to serve alternative
+information, such as alternative cloud access credentials, so that a pod can
+have a defined role for accessing cloud APIs.  By using the metadata service,
+existing code will work unaltered and transparently.
+
+This proposal documents a simple "agreed configuration" which will allow IAM
+proxies to operate consistently, and describes a few future enhancements for
+consideration once the value has been demonstrated.
+
+## Existing work
+
+[#42192](https://github.com/kubernetes/kubernetes/pull/42192) in Kubernetes 1.6
+on GCE created some firewall rules that by default prevent pods from reaching
+169.254.169.254.  It does not affect processes not running in a container, nor
+pods with `hostNetwork=true`.  It does not support proxying to a service,
+though it may support proxying to a local daemonset.
+
+```
+iptables -N KUBE-METADATA-SERVER
+iptables -A FORWARD -p tcp -d 169.254.169.254 --dport 80 -j KUBE-METADATA-SERVER
+iptables -A KUBE-METADATA-SERVER -j DROP
+```
+
+[kube2iam](https://github.com/jtblin/kube2iam) is widely used on AWS to provide
+pods with a particular IAM role.  It relies on a daemonset running on every
+node, and an iptables rule such as
+
+```
+iptables \
+  --append PREROUTING \
+  --protocol tcp \
+  --destination 169.254.169.254 \
+  --dport 80 \
+  --in-interface docker0 \
+  --jump DNAT \
+  --table nat \
+  --to-destination `curl 169.254.169.254/latest/meta-data/local-ipv4`:8181
+```
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/networking-iam.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Proposal
+
+This proposal is intended to allow for integration of a solution such as
+kube2iam into the kubernetes networking infrastructure.
+
+Specifically:
+
+* The rules in [#42192](https://github.com/kubernetes/kubernetes/pull/42192)
+  should be applied at the node level by the installation system, before
```
philips: what is an installation system?

contributors/design-proposals/networking-iam.md
```diff
@@ -0,0 +1,125 @@
...
+  kube-proxy is running.  This acts as a "last line of defense" in case pods are
+  started before kube-proxy rules are fully in place.
+
+* We propose adding a "well known service" for IAM, as we have for DNS.  The
+  DNS service currently runs on `.10`, so we propose reserving `.9` for IAM.
```
philips: do we have a registry for these things? We are also going to need to reserve an IP for etcd service soon too.

philips:
Why is this called IAM networking instead of "instance metadata firewall"? Maybe I am missing something.

contributors/design-proposals/networking-iam.md
justinsb: That which gets you from zero to hero

contributors/design-proposals/networking-iam.md
justinsb: Ah, apparently it is merely a gentlemen's agreement.  The only thing stopping you grabbing .10 is that the installation tool likely got their first.

justinsb:
Thanks @philips .  I don't particularly care about the name, so if you think "instance metadata firewall" is clearer, that's great.

We already have the instance metadata firewall merged (#42192), now we need the additional rules to _allow_ access to IAM.   But I'm certainly mixing implementation strategy into the title there.  Really want to try to get agreement on an implementation approach though.  I think the general approach of sending 169.254.169.254 to k8s service is the correct one, and while I don't love the short-term implementation strategy I think it is good enough to get mileage on the approach while we decide if we want to allow out-of-range ClusterIPs or specially named iptables rule etc.

colemickens:
We'll need this for Azure as well (maybe now, eventually for sure) as our IMDS expands. (cc: @brendandburns)

Two small questions:

1. Is this only about accessTokens, or also about "real" instance metadata as well. If it's the latter, how is that going to work if you proxy all requests through a deployment on some subset of nodes (master or otherwise), instead of through a local per-node proxy as `kube2iam` does? Is the thought that the cloudproviders would fully emulate the metadata API by calling the top level control plane to get the same data and then munge it to look like the metadata response?

2. To check my understanding, the service iptable suffix is a hash of _only_ (name of the service + name of namespace + name of ports)? And that's how you can make the future additions out-of-scope?

and a more philosophical one: Why is this needed as an upstream proposal at all, rather than just letting cluster installers choose to or not to use things like kube2iam? I don't see any extra functionality that falls out of this proposal that isn't possible already with `kube2iam`. Or is it just an effort to standardize the service hash suffix for consistency's sake?

justinsb:
@colemickens 
> Is this only about accessTokens, or also about "real" instance metadata as well. If it's the latter, how is that going to work if you proxy all requests through a deployment on some subset of nodes (master or otherwise), instead of through a local per-node proxy as kube2iam does? Is the thought that the cloudproviders would fully emulate the metadata API by calling the top level control plane to get the same data and then munge it to look like the metadata response?

Good point - yes my current belief is that the IAM service would in fact emulate everything.  You could do a connection back I guess, but it isn't clear that that is the right thing to do anyway.  For example, if we report the instance type as being an `c4.large` on AWS, that implies memory and CPU which the pod might not have.  And the pod isn't reachable on the node's external IP (normally).

An alternative would be a daemonset with selective proxying of just the IAM requests to a privileged central node.  The downside there is that we need an agent running on every node, and if we don't want to give the worker nodes full permissions we still need the central service.  But maybe the firewall rules would be easier.

To check my understanding, the service iptable suffix is a hash of only (name of the service + name of namespace + name of ports)?
To check my understanding, the service iptable suffix is a hash of only (name of the service + name of namespace + name of ports)?

justinsb:
Actually, just going to do this privately.

colemickens:
@justinsb Not sure what "privately" means, but I am in total agreement about having a forum to discuss these things among cloud/k8s providers. There was mention of a `sig-cloud` (maybe with child `wg-{azure,aws,gce}` or something) from @philips... it might be relevant in this context.

thockin:
I agree we can/should do something here.  @Q-Lee is looking at it too.

philips:
Ping @justinsb on what privately means. We would like to see this happen in 1.7. cc @lpabon and @ericchiang

justinsb:
I would also like to see this happen in 1.7.  I'm just going to send a PR.

philips:
Great! Happy to help review.

PR #504, opened 2017-04-03
WIP - Add a threat model to the proposals list

smarterclayton:
Cover our primary threats, our boundaries, and our philosophy for
defense.

### DIFF
--- a/contributors/design-proposals/security-threat-model.md
+++ b/contributors/design-proposals/security-threat-model.md
@@ -1,0 +1,89 @@
+# Security Threat Model
+
+As a cluster manager Kubernetes is expected to defend against a wide variety of threats. This document
+attempts to define the high level actors, potential threats, and our primary layers of defense.
+
+
+## Brief architectural description.
+
+Kubernetes is a cluster manager that runs application workloads inside of containers.  Workloads run as
+**pods** on **nodes** which communicate with the **api-server** to retrieve the description and list of
+pods. Pods may communicate with each other over the network, or contact virtual proxies known as
+**services** from each node at a unique IP or DNS address. Traffic to pods from outside the cluster
+may flow via an **ingress** router/proxy. Pods may have additional resources attached at runtime,
+including persistent volumes, credentials that let them act against the API, secrets used for contacting
+other systems, or configuration files for their processes.
+
+Users and administrators access the apiserver to deploy applications, control access, or make
+changes to existing applications. **Controllers** are automated agents of the system that use the
+definitions in the api-server to ensure the desired state matches reality. Controllers act on-behalf-of
+users to make these changes and have access privileges regular users will not.
+
+
+## Actors
+
+The primary actors and their roles on the system are:
+
+1. Users deploy applications onto the platform via the APIs related to application definition.
+   1. Some users are highly limited in the actions they may take, such as only creating a subset of available resources and being unable to edit.
+2. Administrators manage security, allocate resources, and conduct maintenance on cluster components.
+   1. Organization admins control a subset of resources on the cluster, but are not privileged to act on the whole cluster.
+   2. Infrastucture admins manage both the cluster and the underlying systems that run the cluster.
+3. Each node offers resources to the cluster and accepts pods that have been scheduled on it.
+4. Scheduler controllers assign pods to nodes based on workload definitions and must know details of both to effectively schedule.
+5. Workload controllers create pods that have certain lifecycle rules.
+6. Network providers use information about the topology of the cluster (physical and logical) to give pods IP address and control their access.
+7. Secret controllers create or update secrets for use by users and applications.
+8. Federations need to run workloads on behalf of users on individual clusters.
+9. Other integrations may need access to read or write definitions on some or all of the namespaces in the cluster.
+
+
+## Security boundaries
+
+Several important security boundaries exist in Kubernetes that reduce the possibility of interference or exploit by consumers.
+
+1. Between the container processes within a pod and the node, and between different pods on a node
+2. At the network interfaces between different pods, from the outside of the cluster to inside (ingress), and from the inside of the cluster to outside (egress)
+3. Between infrastructure components and the API
+4. Between end users and the API
+5. Between the API servers and etcd
+6. Between infrastructure monitoring infrastructure and the infrastructure components
+7. Between different tenants, primarily modelled as different namespaces or nodes
+8. Between the federation control plane and the cluster control plane
+
+
+## Multi-tenancy
+
+Kubernetes should be capable of subdividing the cluster for use by distinct participants with their own security boundaries.
+
+1. Single-tenant: A small number of administrators who are also users deploying applications, and are often infrastructure admins
+2. Collaborative Multi-tenant: Two or more teams acting together on the same cluster without strong boundaries between their roles
+3. Multi-tenant: Many users with limited permissions whose applications may need to interact or be isolated, and administrators to configure and enforce that tenancy
+4. Organizational multi-tenant: Several large groups of users and applications (organizations) where the organization admin is granted a subset of cluster administrative permissions in order to subdivide responsibility.
+
+TODO: link to multi-tenancy doc
+
+
+## Threats
+
+The primary threats to Kubernetes are described here:
+
+1. Escaping container and pod isolation to the node or to other containers on the node
+2. Accessing APIs as a user without proper authorization or authentication
+3. Consuming disproportionate resources on the system to deny other users access
+4. Crashing or wedging the system so that no workloads can be processed
+5. Encouraging controllers to act as confused deputies and perform actions at a higher level of trust
+6. Accessing secret data entrusted to the system without appropriate permission for escalating access to the cluster or other systems
+7. Using access to the cluster API to gain elevated permission on the nodes
+8. Using access to the federation API to gain elevated permission on the nodes
+9. Disguising or concealing malicious actions to prevent forensic assessment after an incident
+10. Reusing cluster resources (like services, namespaces, or secret names) after they have been deleted to masquerade as the previous user
+
+TODO: list threats that are not considered
+
+
+### Defending against threats
+
+TODO: categorize how we approach this
+
+TODO: list important upcoming work


smarterclayton:
@kubernetes/sig-auth-misc fairly early, but would appreciate comments on structure, things you'd expect to see vs not, etc.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+
+The primary threats to Kubernetes are described here:
+
+1. Escaping container and pod isolation to the node or to other containers on the node
+2. Accessing APIs as a user without proper authorization or authentication
+3. Consuming disproportionate resources on the system to deny other users access
+4. Crashing or wedging the system so that no workloads can be processed
+5. Encouraging controllers to act as confused deputies and perform actions at a higher level of trust
+6. Accessing secret data entrusted to the system without appropriate permission for escalating access to the cluster or other systems
+7. Using access to the cluster API to gain elevated permission on the nodes
```
timstclair: goes both ways: also using elevated node permissions to gain elevated API permissions.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+8. Using access to the federation API to gain elevated permission on the nodes
+9. Disguising or concealing malicious actions to prevent forensic assessment after an incident
+10. Reusing cluster resources (like services, namespaces, or secret names) after they have been deleted to masquerade as the previous user
+
```
timstclair: - MitM, both for cluster internal traffic, and external traffic
- Threats against storage? Reusing volumes....

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+## Threats
+
+The primary threats to Kubernetes are described here:
+
+1. Escaping container and pod isolation to the node or to other containers on the node
+2. Accessing APIs as a user without proper authorization or authentication
+3. Consuming disproportionate resources on the system to deny other users access
+4. Crashing or wedging the system so that no workloads can be processed
+5. Encouraging controllers to act as confused deputies and perform actions at a higher level of trust
```
timstclair: Also addons, nodes, etc.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+6. Accessing secret data entrusted to the system without appropriate permission for escalating access to the cluster or other systems
+7. Using access to the cluster API to gain elevated permission on the nodes
+8. Using access to the federation API to gain elevated permission on the nodes
+9. Disguising or concealing malicious actions to prevent forensic assessment after an incident
+10. Reusing cluster resources (like services, namespaces, or secret names) after they have been deleted to masquerade as the previous user
+
+TODO: list threats that are not considered
```
timstclair: - Information disclosure through side-channel attacks
- Threats to the underlying infrastructure (or should this be included?), e.g. hardware access
    - Do we want to include threats to the node from outside pods? e.g. through ssh
- Threats to user credentials, e.g. stealing the admin cert
- Application security (to some extent), e.g. XSS in your web app hosted on k8s
- Threats to other systems on the same network (or should this be included?), e.g. cloud services, non-k8s VMs, etc.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+6. Between infrastructure monitoring infrastructure and the infrastructure components
+7. Between different tenants, primarily modelled as different namespaces or nodes
+8. Between the federation control plane and the cluster control plane
+
+
+## Multi-tenancy
+
+Kubernetes should be capable of subdividing the cluster for use by distinct participants with their own security boundaries.
+
+1. Single-tenant: A small number of administrators who are also users deploying applications, and are often infrastructure admins
```
timstclair: Might be worth adding: single tenant, but with workloads that have different trust levels (e.g. DB handling PII and 3p software with a history of bad vulnerabilities)

smarterclayton:
> How do we prevent the list of threats from being a laundry list of every attack we can brainstorm against the system? Or is that what it should be? Ditto for defense.

I'd say if we expect a contributor in the Kube community to think about a particular attack angle during normal development, or to ask about it in a review, it's reasonable to put in this doc.  We definitely can't list everything, but I think a goal I might say is reasonable as well is "do I need to understand this risk to be an effective Kubernetes operator?".  Some threats (side-channel attacks against crypto?) are generally not within that angle, and could be covered by us linking to reference material

> List mitigations per threat, and/or defenses per boundary?

I think if it doesn't get exhaustive.

> What level of detail do we want this to go into? On my todo list is doing a much more in-depth threat model of the node, and similarly I imagine you could go into much more depth for every component of the system.

I think that could be a separate document, and probably is worthwhile to do regardless.  I would expect node change reviewers to apply extra node considerations as well (but maybe not others).

Another possible rule of thumb - if we think we'll need to make a system design change to defend against it, we should have it in the threat model as a high level topic at minimum.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+2. Collaborative Multi-tenant: Two or more teams acting together on the same cluster without strong boundaries between their roles
+3. Multi-tenant: Many users with limited permissions whose applications may need to interact or be isolated, and administrators to configure and enforce that tenancy
+4. Organizational multi-tenant: Several large groups of users and applications (organizations) where the organization admin is granted a subset of cluster administrative permissions in order to subdivide responsibility.
```
smarterclayton: Mostly because you want to delegate more infrastructure/cluster admin tasks to a set of organization admins without necessarily making those admins cluster admins.  Examples would be - an organization admin could define RBAC policy per namespace, but only on a set of namespaces.  Or an organization admin could edit namespaces, but not edit the label on the namespace `organization=org1`, or edit nodes that have label `organization=org1` but not remove them from that organization.  Today, we could go quite far without introducing another scoping concept just with clever application of labels and policy.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+2. At the network interfaces between different pods, from the outside of the cluster to inside (ingress), and from the inside of the cluster to outside (egress)
+3. Between infrastructure components and the API
+4. Between end users and the API
+5. Between the API servers and etcd
+6. Between infrastructure monitoring infrastructure and the infrastructure components
+7. Between different tenants, primarily modelled as different namespaces or nodes
+8. Between the federation control plane and the cluster control plane
+
+
+## Multi-tenancy
```
smarterclayton: The link to multi-tenant needs to exist and soon - David you or I need to turn the various half docs into real docs.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+attempts to define the high level actors, potential threats, and our primary layers of defense.
+
+
+## Brief architectural description.
+
+Kubernetes is a cluster manager that runs application workloads inside of containers.  Workloads run as
+**pods** on **nodes** which communicate with the **api-server** to retrieve the description and list of
+pods. Pods may communicate with each other over the network, or contact virtual proxies known as
+**services** from each node at a unique IP or DNS address. Traffic to pods from outside the cluster
+may flow via an **ingress** router/proxy. Pods may have additional resources attached at runtime,
```
smarterclayton: Probably not, definitely worth calling out here.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+5. Workload controllers create pods that have certain lifecycle rules.
+6. Network providers use information about the topology of the cluster (physical and logical) to give pods IP address and control their access.
+7. Secret controllers create or update secrets for use by users and applications.
+8. Federations need to run workloads on behalf of users on individual clusters.
+9. Other integrations may need access to read or write definitions on some or all of the namespaces in the cluster.
+
+
+## Security boundaries
+
+Several important security boundaries exist in Kubernetes that reduce the possibility of interference or exploit by consumers.
```
destijl: I assume consumers is the superset of all the actors above. To reduce confusion I'd drop it and just use:

```
Several important security boundaries exist in Kubernetes.
```

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+1. Escaping container and pod isolation to the node or to other containers on the node
+2. Accessing APIs as a user without proper authorization or authentication
+3. Consuming disproportionate resources on the system to deny other users access
+4. Crashing or wedging the system so that no workloads can be processed
+5. Encouraging controllers to act as confused deputies and perform actions at a higher level of trust
+6. Accessing secret data entrusted to the system without appropriate permission for escalating access to the cluster or other systems
+7. Using access to the cluster API to gain elevated permission on the nodes
+8. Using access to the federation API to gain elevated permission on the nodes
+9. Disguising or concealing malicious actions to prevent forensic assessment after an incident
+10. Reusing cluster resources (like services, namespaces, or secret names) after they have been deleted to masquerade as the previous user
```
destijl: Agree with this, I proposed some categories above. Happy to contribute more threat brainstorming when we have some categories to structure around. Couple I thought of on first read: misconfigurations leading to unauthorized access, vulnerabilities in system containers.

contributors/design-proposals/security-threat-model.md
destijl: User credential theft should be in. It's one of the most likely methods of compromise. Defences are detailed logging, alerting on unusual access patterns, fast revocation.

philips:
Thanks for doing this Clayton. I won't pile on more feedback as there is plenty to address. But can you ping the comments once you push a new update?

contributors/design-proposals/security-threat-model.md
davidopp: @smarterclayton I finally have time to get back to this now. Let's talk about what scope we want for an initial doc.

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+4. Between end users and the API
+5. Between the API servers and etcd
+6. Between infrastructure monitoring infrastructure and the infrastructure components
+7. Between different tenants, primarily modelled as different namespaces or nodes
+8. Between the federation control plane and the cluster control plane
+
+
+## Multi-tenancy
+
+Kubernetes should be capable of subdividing the cluster for use by distinct participants with their own security boundaries.
```
davidopp: I'm not sure I completely agree with this taxonomy. I certainly don't object to putting it in this doc but I think we can probably do something more a bit more complete later. For example the distinction between users and applications is a bit blurred, and also the "hardness" (you could have teams that are not "acting together" but still don't require hard multi-tenancy... actually you're multi-tenant doesn't really  speak to the "hardness")  ... I think maybe there are multiple axes.

contributors/design-proposals/security-threat-model.md
davidopp: @smarterclayton Do you have anything written up on the latest thoughts on using labels in access control?

contributors/design-proposals/security-threat-model.md
davidopp: "Encouraging" is a bit anthropomorphic :)

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. Single-tenant: A small number of administrators who are also users deploying applications, and are often infrastructure admins
+2. Collaborative Multi-tenant: Two or more teams acting together on the same cluster without strong boundaries between their roles
+3. Multi-tenant: Many users with limited permissions whose applications may need to interact or be isolated, and administrators to configure and enforce that tenancy
```
jessfraz: What level of isolation is this guaranteeing?

contributors/design-proposals/security-threat-model.md
```diff
@@ -0,0 +1,89 @@
...
+7. Between different tenants, primarily modelled as different namespaces or nodes
+8. Between the federation control plane and the cluster control plane
+
+
+## Multi-tenancy
+
+Kubernetes should be capable of subdividing the cluster for use by distinct participants with their own security boundaries.
+
+1. Single-tenant: A small number of administrators who are also users deploying applications, and are often infrastructure admins
+2. Collaborative Multi-tenant: Two or more teams acting together on the same cluster without strong boundaries between their roles
```
jessfraz: could also be called "soft" idk on terminology... if it matters

jessfraz:
is there a plan of action to combine #532 and #551 into one doc? its a little confusing having the discussion sprawled in a few places :)

smarterclayton:
Ugh, I wish there were more hours in the day.  I'll queue this up for next sig-auth meeting.

bgrant0607:
Nit: This will need to be moved into a sig-specific subdirectory, such as sig-auth.

gkarthiks:
Hi @smarterclayton Since we have a KEP process, can we please merge this soon or close this and create a KEP?

mrbobbytables:
As the design proposal process has been superseded by the [KEP process](https://github.com/kubernetes/enhancements) -- I'm going to go ahead and close this out. If this is still relevant, please consider adapting it to a KEP.

/close

PR #506, opened 2017-04-04
Create resource-dependent-pod-preset.md

therc:
A policy to automatically inject changes into a pod when specific resources are requested.

cc @jessfraz @derekwaynecarr

### DIFF
--- a/contributors/design-proposals/resource-dependent-pod-preset.md
+++ b/contributors/design-proposals/resource-dependent-pod-preset.md
@@ -1,0 +1,182 @@
+# Resource-dependent Pod Preset
+
+<!-- vscode-markdown-toc -->
+* 1. [Abstract](#Abstract)
+* 2. [Motivation](#Motivation)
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
cc @pmorie

bgrant0607:
cc @pwittrock

bgrant0607:
cc @vishh @davidopp

bgrant0607:
ref https://github.com/kubernetes/kubernetes/issues/17097

and

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/metadata-policy.md

derekwaynecarr:
is this something we can experiment doing with custom resource types and initializers?

the service catalog community that initiated development of pod preset has been challenged to think along similar lines rather than do this in the core.

bgrant0607:
I agree with @derekwaynecarr. I cannot foresee adding this in the next few releases.

jessfraz:
I agree with initializers :)

PR #507, opened 2017-04-04
Amend cloudprovider metric proposal for new metric names

gnufied:
We are going to use same name for all metrics. But each api request
being made will be added as a dimension.

This has the benefit that - aggregating metrics is way easier without
without resorting to wildcards and stuff.

xref - https://github.com/kubernetes/features/issues/182

cc @kubernetes/sig-storage-pr-reviews @saad-ali @piosz @brancz @fabxc

### DIFF
--- a/contributors/design-proposals/cloudprovider-storage-metrics.md
+++ b/contributors/design-proposals/cloudprovider-storage-metrics.md
@@ -49,37 +49,88 @@ Since we are interested in count(or rate) and latency percentile metrics of API 
 the external Cloud Provider - we will use [Histogram](https://prometheus.io/docs/practices/histograms/) type for
 emitting these metrics.
 
-We will be using `HistogramVec` type so as we can attach dimensions at runtime. Whenever available
-`namespace` will reported as a dimension with the metric.
+We will be using `HistogramVec` type so as we can attach dimensions at runtime. All metrics will contain API action
+being taken as a dimension. The cloudprovider maintainer may choose to add additonal dimensions as needed. If a
+dimension is not available at point of emission sentinel value `<n/a>` should be emitted as a placeholder.
+
+We are also interested in counter of cloudprovider API errors. `NewCounterVec` type will be used for keeping
+track of API errors.
 
 ### GCE Implementation
 
-For GCE we simply use `gensupport.RegisterHook()` to register a function which will be called
-when request is made and response returns.
-
 To begin with we will start emitting following metrics for GCE. Because these metrics are of type
-`Summary` - both count and latency will be automatically calculated.
+`Histogram` - both count and latency will be automatically calculated.
 
-1. gce_instance_list
-2. gce_disk_insert
-3. gce_disk_delete
-4. gce_attach_disk
-5. gce_detach_disk
-6. gce_list_disk
+#### GCE Latency metrics
 
-A POC implementation can be found here - https://github.com/kubernetes/kubernetes/pull/40338/files
+All gce latency metrics will be named - `cloudprovider_gce_api_request_duration_seconds`. api request
+being made will be reported as dimensions.
+
+
+To begin we will start emitting following metrics:
+
+```
+cloudprovider_gce_api_request_duration_seconds { request = "instance_list"}
+cloudprovider_gce_api_request_duration_seconds { request = "disk_insert"}
+cloudprovider_gce_api_request_duration_seconds { request = "disk_delete"}
+cloudprovider_gce_api_request_duration_seconds { request = "attach_disk"}
+cloudprovider_gce_api_request_duration_seconds { request = "detach_disk"}
+cloudprovider_gce_api_request_duration_seconds { request = "list_disk"}
+```
+
+#### GCE API error metrics.
+
+All gce error metrics will be named `cloudprovider_gce_api_request_errors`. api request being made will be
+reported as a dimension.
+
+To begin with we expect to report following error metrics:
+
+```
+cloudprovider_gce_api_request_errors { request = "instance_list"}
+cloudprovider_gce_api_request_errors { request = "disk_insert"}
+cloudprovider_gce_api_request_errors { request = "disk_delete"}
+cloudprovider_gce_api_request_errors { request = "attach_disk"}
+cloudprovider_gce_api_request_errors { request = "detach_disk"}
+cloudprovider_gce_api_request_errors { request = "list_disk"}
+```
+
 
 ### AWS Implementation
 
 For AWS currently we will use wrapper type `awsSdkEC2` to intercept all storage API calls and
 emit metric datapoints.  The reason we are not using approach used for `aws/log_handler` is - because AWS SDK doesn't uses Contexts and hence we can't pass custom information such as API call name or namespace to record with metrics.
 
+
+#### AWS Latency metrics
+
+All aws API metrics will be named - `cloudprovider_aws_api_request_duration_seconds`. `request` will be reported as dimensions.
+AWS maintainer may choose to add additional dimensions as needed.
+
 To begin with we will start emitting following metrics for AWS:
 
-1. aws_attach_volume
-2. aws_create_tags
-3. aws_create_volume
-4. aws_delete_volume
-5. aws_describe_instance
-6. aws_describe_volume
-7. aws_detach_volume
+```
+cloudprovider_aws_api_request_duration_seconds { request = "attach_volume"}
+cloudprovider_aws_api_request_duration_seconds { request = "detach_volume"}
+cloudprovider_aws_api_request_duration_seconds { request = "create_tags"}
+cloudprovider_aws_api_request_duration_seconds { request = "create_volume"}
+cloudprovider_aws_api_request_duration_seconds { request = "delete_volume"}
+cloudprovider_aws_api_request_duration_seconds { request = "describe_instance"}
+cloudprovider_aws_api_request_duration_seconds { request = "describe_volume"}
+```
+
+#### AWS Error metrics
+
+All aws error metrics will be named `cloudprovider_aws_api_request_errors`. api request being made will be
+reported as a dimension.
+
+To begin with we expect to report following error metrics:
+
+```
+cloudprovider_aws_api_request_errors { request = "attach_volume"}
+cloudprovider_aws_api_request_errors { request = "detach_volume"}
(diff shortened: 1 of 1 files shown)


gnufied:
cc @justinsb

brancz:
I think it is now less clear what the namespace stands for, can you elaborate in the doc?

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -58,28 +58,39 @@ For GCE we simply use `gensupport.RegisterHook()` to register a function which w
...
-`Summary` - both count and latency will be automatically calculated.
+`Histogram` - both count and latency will be automatically calculated.
 
-1. gce_instance_list
-2. gce_disk_insert
-3. gce_disk_delete
-4. gce_attach_disk
-5. gce_detach_disk
-6. gce_list_disk
+All gce metrics will be named - `gce_api_latency_duration_seconds`. Namespace and api request
```
fabxc: GCE and API for readability. Why the "–"?

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -58,28 +58,39 @@ For GCE we simply use `gensupport.RegisterHook()` to register a function which w
...
+being made will be reported as dimensions.
 
+The complete list of time series will look like:
+
+```
+gce_api_latency_duration_seconds { namespace: "", api_request: "instance_list"}
```
fabxc: The `:` should be `=` if we follow the format of the /metrics endpoints.

contributors/design-proposals/cloudprovider-storage-metrics.md
fabxc: Isn't `latency_duration` kind of doubling down on the meaning? Seems either of them would be enough but `request` should be in there somehow. 

How about `gce_api_request_duration_seconds`?

gnufied:
@brancz @fabxc namespace is kubernetes namespace. Now that I think of, we can explicitly name dimension to be `kube_namespace` so as to avoid any confusion

brancz:
I understand it's the kubernetes namespace, but I'm wondering why this information is necessary at all. If we provide a namespace, when we should also provide the object this request was made for, no?

gnufied:
In a multi tenant environment there is a desire to calculate which namespace is using most api requests.  This way we can identify namespaces which are using the cloudprovider api most. 

To be fair namespace information is not always available but my idea is to emit that information whenever it is available.

I can amend the metrics to emit object for which api request is being made for. But similar to namespaces - we do not always have that information available at call site where metrics are created.  Moreover - sometimes a api call can be made against 2 different types of objects and we can't uniquely identify against which object the api request was made (`describeVolume` for example). To capture this information - we have to make larger refactor of cloudprovider code which I am hesitant to do at this point.

fabxc:
That's fine. But if the info is not available it should be filled with a sentinel value rather than the empty string, e.g. `namespace="<unknown>"`.

brancz:
Makes sense. To stay conform to the instrumentation guide for kubernetes components we should definitely stick to `namespace`, and @fabxc's comment also makes sense, we've done something similar in other components, such as the rescheduler.

gnufied:
I have updated the design proposal as per latest discussion me, @justinsb and @bowei had. I have dropped namespace as a dimension. I have also added bit about cloudprovider maintainer adding dimension as needed.

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -49,37 +49,46 @@ Since we are interested in count(or rate) and latency percentile metrics of API
...
-2. gce_disk_insert
-3. gce_disk_delete
-4. gce_attach_disk
-5. gce_detach_disk
-6. gce_list_disk
-
+`Histogram` - both count and latency will be automatically calculated.
+
+All gce metrics will be named - `cloudprovider_gce_api_duration_seconds`. api request
+being made will be reported as dimensions. The cloudprovider maintainer may choose to add additional dimensions
```
bowei: This statement seems like it should be not specific to GCE?

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -49,37 +49,46 @@ Since we are interested in count(or rate) and latency percentile metrics of API
...
+as needed.
+
+The complete list of time series will look like:
```
bowei: Remove "complete"? This is not a complete list, it's a list of examples? It'll make the document hard to maintain if we have to keep it in sync with the code.

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -49,37 +49,46 @@ Since we are interested in count(or rate) and latency percentile metrics of API
 the external Cloud Provider - we will use [Histogram](https://prometheus.io/docs/practices/histograms/) type for
 emitting these metrics.
 
-We will be using `HistogramVec` type so as we can attach dimensions at runtime. Whenever available
-`namespace` will reported as a dimension with the metric.
+We will be using `HistogramVec` type so as we can attach dimensions at runtime.
```
bowei: Should we almost keep track of errors? E.g. the error count in the GCE proposed implementation...

gnufied: updated.

gnufied:
@bowei I have updated proposal to include error counter.

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -49,37 +49,87 @@ Since we are interested in count(or rate) and latency percentile metrics of API
...
-when request is made and response returns.
+### GCE Implementation
 
 To begin with we will start emitting following metrics for GCE. Because these metrics are of type
-`Summary` - both count and latency will be automatically calculated.
+`Histogram` - both count and latency will be automatically calculated.
+
+#### GCE Latency metrics
+
+All gce latency metrics will be named - `cloudprovider_gce_api_duration_seconds`. api request
```
bowei: You will need to add the naming for the error counter as well

cloudprovider_gce_api_errors

bowei: whoops ignore this comment

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -49,37 +49,87 @@ Since we are interested in count(or rate) and latency percentile metrics of API
...
+being made will be reported as dimensions.
+
+
+To begin we will start emitting following metrics:
+
+```
+cloudprovider_gce_api_duration_seconds { api_request = "instance_list"}
```
bowei: api_ in `api_request` is redundant (the metric name already has `api` in it)

gnufied: lets just call it just `request` ? I don't have strong opinions on this one.

bowei: request works for me

gnufied: renamed

contributors/design-proposals/cloudprovider-storage-metrics.md
bowei: Now thinking about it, we may not want all of the metrics to be put into the same name, as this makes all of the metrics need to share the same dimensions (e.g. zone, region, etc). The dimensions may not make sense for most of the metrics.

For example, if you look at my PR, disk has (zone, node), but other metrics only have (zone) or ().

It probably would better to have:
cloudprovider_{gce,aws, ...}_{disk,loadbalancer, ...}

to avoid having too many unused properties.

What do you think?

gnufied: What other properties we are considering btw? I was thinking these metrics are somewhat lower level and meant to only track API latency and rate/quota usage.

If we are going to need to classify the metrics between `disk` or `loadbalancer`  or something else, it might be better to use API request name with metrics itself rather than moving it to dimension. It is needless mental overload having to classify the metrics.

gnufied: We spoke about this offline and we agreed to use same metric name for all metrics.  A cloudprovider may choose to emit a dimension's value as `n/a` if it is not available at metric emission point.

gnufied:
@piosz @fabxc  PTAL. Since last update - have made following changes to the proposal:

1. Use same metric name for all cloudprovider metrics from same cloudprovider. Use `request` as a dimension to determine API action being taken.
2. Drop `namespace` as a requirement from cloudprovider metrics. The only dimension which will be always present is `request` dimension. Beyond that, a cloudprovider may choose to emit any dimensions they need.
3. If a dimension's value is not available (and declared metric defines the dimension). Its value can be set to `n/a` during the emission.

contributors/design-proposals/cloudprovider-storage-metrics.md
```diff
@@ -49,37 +49,88 @@ Since we are interested in count(or rate) and latency percentile metrics of API
...
+#### GCE Latency metrics
+
+All gce latency metrics will be named - `cloudprovider_gce_api_duration_seconds`. api request
+being made will be reported as dimensions.
+
+
+To begin we will start emitting following metrics:
+
+```
+cloudprovider_gce_api_duration_seconds { request = "instance_list"}
```
fabxc: I'd tend to include make the suffix `request_duration_seconds` regardless. It's not a super hard requirement but certainly the common approach. Plus, "API duration" reads kind of awkward – when just looking at the metric name, it's unclear that it's about a request.

gnufied: Do you mean the name ought to be `cloudprovider_gce_request_duration_seconds`  or `cloudprovider_gce_api_request_duration_seconds` ? I was mostly concerned with making the name too big - but I do not have any objections to changing it.

brancz: I agree that it makes the name pretty long, but it is much more understandable what the metric represents so I'd vote for the convention/best practice of `request_duration_seconds`.

gnufied: Not sure I am completely agree with keeping such long names. But in the interest of getting this shipped - I will lay down my guns.

gnufied:
@saad-ali can you approve and merge?

PR #508, opened 2017-04-04
sig-list: update sig lists

euank:
I stepped down from sig-node a while ago and missed removing my name
from here, sorry!

I've also removed sig-rktnetes from the list since none of the listed
leads are active on that sig.

cc @philips


derekwaynecarr:
LGTM

PR #510, opened 2017-04-05
Note to skip Disruptive when testing [Volume]

jeffvance:
Small addition to e2e-tests doc based on experience.


jeffvance:
Since every test should now be owned by a SIG and have a corresponding [sig-<name>] label, the use of  the `[Volumes]` tag should become deprecated, and therefore the note added by this pr is unnecessary.

PR #516, opened 2017-04-06
Specification for multi-label subdomain support in CNAME records

CsatariGergely:
We would like to propose a feature to support multi-label subdomain support in CNAME records.

### DIFF
--- a/contributors/design-proposals/multi-label-CNAME.md
+++ b/contributors/design-proposals/multi-label-CNAME.md
@@ -2,7 +2,7 @@
 
 Author: Laszlo Janosi (@janosi), Gergely Csatari (@CsatariGergely)
 
-Date: August 2017
+Date: April 2017
 
 Status: New
 
@@ -103,8 +103,8 @@ type ServiceSpec struct {
 	LoadBalancerSourceRanges []string
 
 +	// The name of the service which is resolved by kubedns or equivalent to a canonical name.
-+ // Enable the definition of multi-label CNAME records. But also can be used to
-+ // define a single-label service name, in which case it takes the role of the "name" label in metadata part.
++	// Enable the definition of multi-label CNAME records. But also can be used to
++	// define a single-label service name, in which case it takes the role of the "name" label in metadata part.
 +	// When Name is resolved, kubedns or equivalent returns with the canonical name defined in ExternalName field above.
 +	// Must be a valid DNS name and requires Type to be ExternalName.
 + Name string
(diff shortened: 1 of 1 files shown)


kargakis:
@kubernetes/sig-network-api-reviews @smarterclayton

contributors/design-proposals/multi-label-CNAME.md
```diff
@@ -0,0 +1,127 @@
...
+# Goal
+
+Enable the usage of multi-label subdomains in CNAME records via ExternalName Service.
+Example: `my-service.my-cluster` could be mapped to `your-service.your-cluster`,
+while with the current implementation only `my-service` can be mapped.
+
+# Motivation
+DNS CNAME records are specified in [RFC1034](https://www.ietf.org/rfc/rfc1034.txt)
+and clarified in Section 10 of [RFC2181](https://www.ietf.org/rfc/rfc2181.txt).
+These allow the usage of multi-label subdomains in CNAME records, and not only
```
smarterclayton: Can you provide more use cases to support this?  The concept of a single service label is very deeply embedded in Kubernetes architecture, and it's going to be very expensive to change direction.  I'd like to understand what scenarios you can't achieve today and how you see this functional change enabling new use cases that improve Kubernetes.  Otherwise, it's going to be very difficult.

CsatariGergely: @smarterclayton thank for the comments. 
Our belief was, that this change will affect only ExternalName, so the change would not be that expensive. 
Are the two usecases listed in the Examples section not enough?

thockin:
This allows all sorts of collisions and ambiguities to arise with no real way to detect or prevent them.  I don't think we should do this.

@bowei (has done most of the work in upstream DNS lately)

janosi:
@thockin could you please provide some examples for those collisions and ambiguities? Maybe there is a misunderstanding about our goals here.

thockin:
if I have a service whose Name field (as specified here) is `kubernetes.default` I have just broken my cluster.  I can hijack any name in any other namespace.

janosi:
@thockin even if this Name is put only into kubeDNS as CNAME? So, if I go and enter this CNAME into kubeDNS directly now, then my cluster gets down? Do we talk about generic CNAME related security issues then, not specific to k8s?

janosi:
@thockin OK, forget my previous comment.  I was wrong. I understand now that k8s actually supports this use case I mentioned. I.e. mapping a service_name.current_namespace.svc.cluster.local to another hostname.
And that is the way it provides some level of security against hijacking anything from anywhere.

janosi:
@thockin I understand your concerns. Please close this PR.

PR #520, opened 2017-04-07
Update godep.md

cblecker:
Update with proper hack commands and procedures to update godep.md

### DIFF
--- a/contributors/devel/godep.md
+++ b/contributors/devel/godep.md
@@ -54,7 +54,7 @@ export GOPATH=$KPATH
 
 ```sh
 cd $KPATH/src/k8s.io/kubernetes
-godep restore
+./hack/godep-restore.sh
 ```
 
 4) Next, you can either add a new dependency or update an existing one.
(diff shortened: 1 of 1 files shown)


cblecker:
I've tested this personally, as well as had another user test them [here](https://github.com/kubernetes/kubernetes/pull/44082#issuecomment-292698640).

cblecker:
cc @sttts who has helped previously with the godep scripts

contributors/devel/godep.md
```diff
@@ -69,24 +69,19 @@ godep get $DEP/...
...
 # state.
-rm -rf $KPATH/src/$DEP # repo root
-godep get $DEP/...
-# Change code in Kubernetes, if necessary.
-rm -rf Godeps
-rm -rf vendor
+cd $KPATH/src/$DEP # repo root
+git checkout <$REVISION/$TAG/$BRANCH>
+cd $KPATH/src/k8s.io/kubernetes
+godep update $DEP/...
```
sttts: Will this work if `$DEP` depends on one of our staging repos? Like k8s.io/metrics.

sttts: Why do we need this update call after all? Does it speed up the following godep-save? Can we leave vendor/ and Godeps/Godeps.json when running godep-save.sh?

cblecker: Godeps/Godeps.json doesn't pick up the intention to upgrade without it. And yes, with this procedure, I'm not blowing away vendor/ or Godeps/

sttts: I am still not convinced that we have to bring godep-update into play here if we call `./hack/godep-save.sh` afterwards. Therefore my question again: does it speed up `./hack/godep-save.sh` considerably?

Also note that we pin the godep version in godep-save/restore.sh. If we call godep-update directly, it's not necessarily the pinned version.

cblecker: The second point you make is excellent -- you're absolutely right that this doesn't work if we're pinning the version.

Let me test this out some more, and I'll report back with findings. Appreciate the review, @sttts!

cblecker:
I'm going to close these as these scripts really just need to be improved rather than worked around.

PR #523, opened 2017-04-10
Add Extending SLO docs to sig-scalability

gmarek:
Google-doc-made-md - a copy of the doc shared with the community a month ago, thus ready to be made "official". This is the best place we were able to find to put this, but if it there is a better place please let me know. cc @thockin for better idea on where to put this.

@kubernetes/sig-scalability-pr-reviews

### DIFF
--- a/sig-scalability/README.md
+++ b/sig-scalability/README.md
@@ -26,11 +26,12 @@ For more details about our objectives please review [Scaling And Performance Goa
 - [kubernetes-sig-scale](https://groups.google.com/forum/#!forum/kubernetes-sig-scale)
 
 ## Docs
+- [Extending SLO definitions](extending_slo.md)
 - [Scaling And Performance Goals](goals.md)
 
-## Scalability SLAs
+## Scalability SLOs
 
-We officially support two different SLAs:
+We officially support two different SLOs:
 
 1. "API-responsiveness":
    99% of all API calls return in less than 1s
@@ -38,8 +39,8 @@ We officially support two different SLAs:
 1. "Pod startup time:
    99% of pods (with pre-pulled images) start within 5s
 
-This should be valid on appropriate hardware up to a 1000 node cluster with 30 pods/node.  We eventually want to expand that to 100 pods/node.
+This should be valid on appropriate hardware up to a 5000 node cluster with 30 pods/node.  We eventually want to expand that to 100 pods/node.
 
 For more details how do we measure those, you can look at: http://blog.kubernetes.io/2015_09_01_archive.html
 
-In the future we may want to add more SLAs (e.g. scheduler throughput), but we are not there yet.
+We are working on refining existing SLOs and defining more for other areas of the system.
--- a/sig-scalability/extending_slo.md
+++ b/sig-scalability/extending_slo.md
@@ -1,0 +1,69 @@
+# Extended Kubernetes scalability SLOs
+
+## Goal
+The goal of this effort is to extend SLOs which Kubernetes cluster has to meet to support given number of Nodes. As of April 2017 we have only two SLOs:
+- API-responsiveness: 99% of all API calls return in less than 1s
+- Pod startup time: 99% of Pods (with pre-pulled images) start within 5s
+which are enough to guarantee that cluster doesn't feel completely dead, but not enough to guarantee that it satisfies user's needs.
+
+We're going to define more SLOs based on most important indicators, and standardize the format in which we speak about our objectives. Our SLOs need to have two properties:
+- They need to be testable, i.e. we need to have a benchmark to measure if it's met,
+- They need to be expressed in a way that's possible to understand by a user not intimately familiar with the system internals, i.e. formulation can't depend on some arcane knowledge.
+On the other hand we do not require that:
+- SLOs are possible to monitor in a running cluster, i.e. not all SLOs need to be easily translatable to SLAs. Being able to benchmark is enough for us.
+
+## Split metrics from environment
+Currently what me measure and how we measure it is tightly coupled. This means that we don't have good environmental constraint suggestions for users (e.g. how many Pods per Namespace we support, how many Endpoints per Service, how to setup the cluster etc.). We need to decide on what's reasonable and make the environment explicit.
+
+## Split SLOs by kind
+Current SLOs implicitly assume that the cluster is in a "steady state". By this we mean that we assume that there's only some, limited, number of things going during benchmarking. We need to make this assumption explicit and split SLOs into two categories: steady-state SLOs and burst SLOs.
+
+## Steady state SLOs
+With steady state SLO we want to give users the data about system's behavior during normal operation. We define steady state by limiting the churn on the cluster. 
+
+This includes current SLOs:
+- API call latency
+- E2e Pod startup latency
+
+By churn we understand a measure of amount changes happening in the cluster. It's formal(-ish) definition will follow, but informally it can be thought about as number of user-issued requests per second plus number of pods affected by those requests. 
+
+More formally churn per second is defined as:
+#Pod creations + #PodSpec updates + #user originated requests in a given second
+The last part is necessary only to get rid of situations when user is spamming API server with various requests. In ordinary circumstances we expect it to be in the order of 1-2. 
+
+## Burst SLOs
+With burst SLOs we want to give user idea on how system behaves under the heavy load, i.e. when one want the system to do something as quickly as possible, not caring too much about response time for a single request. Note that this voids all steady-state SLOs.
+
+This includes the new SLO:
+- Pod startup throughput
+
+## Environment
+Kubernetes cluster in which we benchmark SLOs need to meet following criteria:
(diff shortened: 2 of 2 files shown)


jeremyeder:
Good first step.

I think the most important gap remaining (from the control-plane/scheduler-centric POV) is that we deploy more complex applications that cause different controllers and features to show up.

We cannot ignore the fact that there are significantly lower scalability limits for different objects within the system (such as services or PVs) than there are for pause pods.

wojtek-t:
> We cannot ignore the fact that there are significantly lower scalability limits for different objects within the system (such as services or PVs) than there are for pause pods.

Yes of course. I'm working on the proposal for a new set of SLOs that I think we should introduce (obviously a bunch of them we won't be meeting when introducing them) and I hope will be able to share it next week or so. But I obviously agree with that.

gmarek:
@jeremyeder - goal is to create a framework, so that we'll know what we talk about, and easily exchange some parts of the definition.

jeremyeder:
Thank you for the follow-ups, we're on the same page :-)

sig-scalability/README.md
```diff
@@ -26,20 +26,21 @@ For more details about our objectives please review [Scaling And Performance Goa
...
 1. "Pod startup time:
    99% of pods (with pre-pulled images) start within 5s
 
-This should be valid on appropriate hardware up to a 1000 node cluster with 30 pods/node.  We eventually want to expand that to 100 pods/node.
+This should be valid on appropriate hardware up to a 5000 node cluster with 30 pods/node.  We eventually want to expand that to 100 pods/node.
 
 For more details how do we measure those, you can look at: http://blog.kubernetes.io/2015_09_01_archive.html
 
-In the future we may want to add more SLAs (e.g. scheduler throughput), but we are not there yet.
+In the future we may want to add more SLOs (e.g. scheduler throughput).
```
wojtek-t: We are working on refining existing SLOs and defining more for other areas of the system.

sig-scalability/extending_slo.md
```diff
@@ -0,0 +1,71 @@
+# Extended Kubernetes scalability SLOs
+###### Authors: gmarek@google.com jeder@redhat.com
```
wojtek-t: I think that we generally not putting authors on any docs and code.

sig-scalability/extending_slo.md
```diff
@@ -0,0 +1,71 @@
...
+###### Last edit: 2017-04-10
+
+## Goal
+The goal of this effort is to extend SLOs which Kubernetes cluster has to meet to support given number of Nodes. Currently we have only two SLOs:
```
wojtek-t: s/Currently/As of 2017-04-10/

sig-scalability/extending_slo.md
```diff
@@ -0,0 +1,71 @@
...
+
+This includes the new SLO:
+- Pod startup throughput
+
+## Environment
+Kubernetes cluster in which we benchmark SLOs need to meet following criteria:
+- Run a single master machine appropriately sized
+- Main etcd runs as a single instance on the master machine
+- Events are stored in a separate etcd instance running on the master machine
+- Kubernetes version is at least 1.7
```
wojtek-t: s/1.7/1.X.Y/

sig-scalability/extending_slo.md
```diff
@@ -0,0 +1,71 @@
+# Extended Kubernetes scalability SLOs
```
wojtek-t: Since we will be adding more SLOs, I suggest creating a directory for SLO-related documents.

sig-scalability/README.md
gmarek: Done.

sig-scalability/extending_slo.md
gmarek: I'm going to do it in the second PR (with the throughput SLO doc)

sig-scalability/extending_slo.md
gmarek: Done.

sig-scalability/extending_slo.md
gmarek: Done.

sig-scalability/extending_slo.md
gmarek: Done.

sig-scalability/extending_slo.md
```diff
@@ -0,0 +1,69 @@
...
+The last part is necessary only to get rid of situations when user is spamming API server with various requests. In ordinary circumstances we expect it to be in the order of 1-2. 
+
+## Burst SLOs
+With burst SLOs we want to give user idea on how system behaves under the heavy load, i.e. when one want the system to do something as quickly as possible, not caring too much about response time for a single request. Note that this voids all steady-state SLOs.
+
+This includes the new SLO:
+- Pod startup throughput
+
+## Environment
+Kubernetes cluster in which we benchmark SLOs need to meet following criteria:
```
timothysc: I do think we should rally on a common deployment such as kubeadm, b/c the number produced would only reflect GKE, b/c no one else deploys this way.

sig-scalability/extending_slo.md
```diff
@@ -0,0 +1,69 @@
...
+- Main etcd runs as a single instance on the master machine
+- Events are stored in a separate etcd instance running on the master machine
+- Kubernetes version is at least 1.X.Y
+- Components configuration = _?_
+
+_TODO: NEED AN HA CONFIGURATION AS WELL_
+
+## SLO template
+All our performance SLOs should be defined using the following template:
+
```
timothysc: ?

sig-scalability/extending_slo.md
countspongebob: We'll need to figure out how to stage these out release by release. Thinking we'll cover that in the feature repo, though. This is looking good.

gmarek: We certainly should think where to put it and how to advertise it. We also need to work on SIG meeting on filling up cluster configuration for tests. I'm thinking about putting there our sizing guides, to have anything to begin discussion.

sig-scalability/extending_slo.md
```diff
@@ -0,0 +1,71 @@
...
+###### Authors: gmarek@google.com jeder@redhat.com
+###### Last edit: 2017-04-10
+
+## Goal
+The goal of this effort is to extend SLOs which Kubernetes cluster has to meet to support given number of Nodes. Currently we have only two SLOs:
+- API-responsiveness: 99% of all API calls return in less than 1s
+- Pod startup time: 99% of Pods (with pre-pulled images) start within 5s
+which are enough to guarantee that cluster doesn't feel completely dead, but not enough to guarantee that it satisfies user's needs.
+
+We want to add few most important SLOs to our list and standardize the format in which we speak about them. Our SLOs need to have two properties:
```
gmarek: PTAL.

sig-scalability/extending_slo.md
gmarek: ? ?

sig-scalability/extending_slo.md
gmarek: Are you relating to single-master setup, or defining it as a set of requirements?

I know virtually nothing about HA setups (nor kops or kubeadm), so I need someone else to write the recommendations for that. Is someone interested in helping with that? @timothysc @justinsb @countspongebob @mikedanese.

sig-scalability/extending_slo.md
gmarek: This is what I see:

slo-template

sig-scalability/extending_slo.md
timothysc: > Are you relating to single-master setup. 

I'm referring to your setup defined which is pretty much different then everyone else now-a-days.  e.g. events etcd + single master ...

gmarek: Agree, we just need some stable definition to act as benchmark, as we can't test every possible configuration. I don't want to depend on anything external (e.g. kubeadm), as this would mean, that we'd need to rerun all tests for all new kubeadm versions. I'm OKish with freezing kubeadm version (as long as it'll support creating clusters from head), but this is as arbitrary configuration as ours. The main goal here is to give some intuition about resource requirements, and give easy-enough way to rerun our experiments.

wojtek-t: @timothysc - I definitely agree that "single master" shouldn't be the only setup we are testing (so clustered etcd and more then 1 apiserver is definitely something we also want to test). That said, I don't agree that HA configuration should be the only one we test.

It's maybe not clear from this, but the plan is to have defined two configuration:
- 1 for single-master setup
- 1 for HA setups (which obviously has more master machines)
@gmarek - maybe this should be more clear.

Etcd for events is a separate discussion which I would like to avoid now :)

gmarek: I have a `_TODO: NEED AN HA CONFIGURATION AS WELL_` in the doc - I'm not sure how much clearer I can get:)

gmarek:
OK, as no one has any better ideas, I'm merging it here.

PR #524, opened 2017-04-10
Proposal: TPR to beta proposal

deads2k:
There are a number of important issues with the alpha version of ThirdPartyResources that we wish to address to move TPR to beta. The list is tracked [here](https://github.com/kubernetes/features/issues/95), and also includes feedback from existing Kubernetes ThirdPartyResource users. This proposal covers the steps that are necessary to move TPR to beta and to prevent future challenges in upgrading.

@kubernetes/sig-api-machinery-misc @erictune @lavalamp @brendandburns @philips

### DIFF
--- a/contributors/design-proposals/thirdpartyresources.md
+++ b/contributors/design-proposals/thirdpartyresources.md
@@ -1,0 +1,253 @@
+# Moving ThirdPartyResources to beta
+
+## Background
+There are a number of important issues with the alpha version of
+ThirdPartyResources that we wish to address to move TPR to beta. The list is
+tracked [here](https://github.com/kubernetes/features/issues/95), and also
+includes feedback from existing Kubernetes ThirdPartyResource users. This
+proposal covers the steps we believe are necessary to move TPR to beta and to
+prevent future challenges in upgrading.
+
+
+## Goals
+1. Ensure ThirdPartyResource APIs operate consistently with first party
+Kubernetes APIs.
+2. Enable ThirdPartyResources to specify how they will appear in API
+discovery to be consistent with other resources and avoid naming confilcts
+3. Move TPR into their own API group to allow the extensions group to be
+[removed](https://github.com/kubernetes/kubernetes/issues/43214)
+4. Support cluster scoped TPR resources
+5. Identify other features required for TPR to become beta
+6. Minimize the impact to alpha ThirdPartyResources consumers and define a
+process for how TPR migrations / breaking changes can be accomplished (for
+both the cluster and for end users)
+
+Non-goals
+1. Solve automatic conversion of TPR between versions or automatic migration of
+existing TPR
+
+### Desired API Semantics
+TPRs are intended to look like normal kube-like resources to external clients.
+In order to do that effectively, they should respect the normal get, list,
+watch, create, patch, update, and delete semantics.
+
+In "normal" Kubernetes APIs, if I have a persisted resource in the same group
+with the same name in v1 and v2, they are backed by the same underlying object.
+A change made to one is reflected in the other. API clients, garbage collection,
+namespace cleanup, version negotiation, and controllers all build on this.
+
+The convertibility of Kubernetes APIs provides a seamless interaction between
+versions.  A TPR does not have the ability to convert between versions, which
+focuses on the primary role of TPR as an easily extensible and simple mechanism
+for adding new APIs. Conversion primarily allows structural, but not backwards
+incompatible, changes. By not supporting conversion, all TPR use cases are
+preserved, but a large amount of complexity is avoided for consumers of TPR.
+
+Allowing a single, user specified version for a given TPR will provide this
+semantic by preventing server-side versioning altogether.  All instances of a
+single TPR must have the same version or the Kubernetes API semantic of always
+returning a resource encoded to the matching version will not be maintained.
+Since conversions (even native Kubernetes conversions) cannot be used to handle
+behavioral changes, the same effect can be achieved for TPRs client-side with
+overlapping serialization changes.
+
+
+### Avoiding Naming Problems
+There are several identifiers that a Kubernetes API resource has which share
+value-spaces within an API group and must not conflict.  They are:
+1. Resource-type value space
+  1. plural resource-type name - like "configmaps"
+  2. singular resource-type name - like "configmap"
+  3. short names - like "cm"
+2. Kind-type value space - for group "example.com"
+  1. Kind name - like "ConfigMap"
+  2. ListKind name - like "ConfigMapList"
+If these values conflict within their value-spaces then no client will be able
+to properly distinguish intent.
+
+The actual name of the TPR-registration (resource that describes the TPR to
+create) resource can only protect one of these values from conflict.  Since
+Kubernetes API types are accessed via a URL that looks like `/apis/<group>/<version>/namespaces/<namespace-name>/<plural-resource-type>`,
+the name of the TPR-registration object will be `<plural-resource-type>.<group>`.
+
+Conflicts with other parts of the value-space can not be detected with static
+validation, so there will be a spec/status split with `status.conditions` that
+reflect the acceptance status of a TPR-registration.  For instance, you cannot
+determine whether two TPRs in the same group have the same short name without
+inspecting the current state of existing TPRs.
+
+Parts of the value-space will be "claimed" by making an entry in TPR.status to
+include the accepted names which will be served.  This prevents a new TPR from
+disabling an existing TPR's name.
+
+
+## New API
+In order to: 
+1. eliminate opaquely derived information - deriving camel-cased kind names
+from lower-case dash-delimited values as for instance.
+1. allow the expression of complex transformations - not all plurals are easily
+determined (ox and oxen) and not all are English.  Fields for complete
+specification eliminates ambiguity.
+1. handle TPR-registration value-space conflicts
+1. [stop using the extensions API group](https://github.com/kubernetes/kubernetes/issues/43214)
+
+We can create a type `ThirdPartyResource.apiextension.k8s.io`.
+```go
+// ThirdPartyResourceSpec describe how a user wants their resource to appear
+type ThirdPartyResourceSpec struct {
+	// Group is the group this resource belongs in
+	Group string `json:"group" protobuf:"bytes,1,opt,name=group"`
+	// Version is the version this resource belongs in
+	Version string `json:"version" protobuf:"bytes,2,opt,name=version"`
+	// Names holds the information about the resource and kind you have chosen which is
+	// surfaced through discovery.
+	Names ThirdPartyResourceNames
+
+	// Scope indicates whether this resource is cluster or namespace scoped.  Default is namespaced
+	Scope ResourceScope `json:"scope" protobuf:"bytes,8,opt,name=scope,casttype=ResourceScope"`
+}
+
+type ThirdPartyResourceNames struct {
+	// Plural is the plural name of the resource to serve.  It must match the name of the TPR-registration
+	// too: plural.group
+	Plural string `json:"plural" protobuf:"bytes,3,opt,name=plural"`
+	// Singular is the singular name of the resource.  Defaults to lowercased <kind>
+	Singular string `json:"singular,omitempty" protobuf:"bytes,4,opt,name=singular"`
+	// ShortNames are short names for the resource.
+	ShortNames []string `json:"shortNames,omitempty" protobuf:"bytes,5,opt,name=shortNames"`
+	// Kind is the serialized kind of the resource
+	Kind string `json:"kind" protobuf:"bytes,6,opt,name=kind"`
+	// ListKind is the serialized kind of the list for this resource.  Defaults to <kind>List
+	ListKind string `json:"listKind,omitempty" protobuf:"bytes,7,opt,name=listKind"`
+}
+
+type ResourceScope string
+
+const (
+	ClusterScoped    ResourceScope = "Cluster"
+	NamespaceScoped  ResourceScope = "Namespaced"
+)
+
+type ConditionStatus string
+
+// These are valid condition statuses. "ConditionTrue" means a resource is in the condition.
+// "ConditionFalse" means a resource is not in the condition. "ConditionUnknown" means kubernetes
+// can't decide if a resource is in the condition or not. In the future, we could add other
+// intermediate conditions, e.g. ConditionDegraded.
+const (
+	ConditionTrue    ConditionStatus = "True"
+	ConditionFalse   ConditionStatus = "False"
+	ConditionUnknown ConditionStatus = "Unknown"
+)
+
+// ThirdPartyResourceConditionType is a valid value for ThirdPartyResourceCondition.Type
+type ThirdPartyResourceConditionType string
+
+const (
+	// NameConflict means the resource or kind names chosen for this ThirdPartyResource conflict with others in the group.
+	// The first TPR in the group to have the name reflected in status "wins" the name.
+	NameConflict ThirdPartyResourceConditionType = "NameConflict"
+	// Terminating means that the ThirdPartyResource has been deleted and is cleaning up.
+	Terminating ThirdPartyResourceConditionType = "Terminating"
+)
+
+// ThirdPartyResourceCondition contains details for the current condition of this ThirdPartyResource.
+type ThirdPartyResourceCondition struct {
+	// Type is the type of the condition.
+	Type ThirdPartyResourceConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=ThirdPartyResourceConditionType"`
+	// Status is the status of the condition.
+	// Can be True, False, Unknown.
+	Status ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status,casttype=ConditionStatus"`
+	// Last time the condition transitioned from one status to another.
+	// +optional
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,4,opt,name=lastTransitionTime"`
+	// Unique, one-word, CamelCase reason for the condition's last transition.
+	// +optional
+	Reason string `json:"reason,omitempty" protobuf:"bytes,5,opt,name=reason"`
+	// Human-readable message indicating details about last transition.
+	// +optional
+	Message string `json:"message,omitempty" protobuf:"bytes,6,opt,name=message"`
+}
+
+// ThirdPartyResourceStatus indicates the state of the ThirdPartyResource
+type ThirdPartyResourceStatus struct {
+	// Conditions indicate state for particular aspects of a ThirdPartyResource
+	Conditions []ThirdPartyResourceCondition `json:"conditions" protobuf:"bytes,1,opt,name=conditions"`
+
+	// AcceptedNames are the names that are actually being used to serve discovery
+	// They may not be the same as names in spec.
+	AcceptedNames ThirdPartyResourceNames
+}
+
+// +genclient=true
+
+// ThirdPartyResource represents a resource that should be exposed on the API server.  Its name MUST be in the format
+// <.spec.plural>.<.spec.group>.
+type ThirdPartyResource struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Spec describes how the user wants the resources to appear
+	Spec ThirdPartyResourceSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+	// Status indicates the actual state of the ThirdPartyResource
+	Status ThirdPartyResourceStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// ThirdPartyResourceList is a list of ThirdPartyResource objects.
+type ThirdPartyResourceList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Items individual ThirdParties
+	Items []ThirdPartyResource `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+```
+
+
+## Behavior
+### Create
+When a new TPR is created, no synchronous action is taken.
+A controller will run to confirm that value-space of the reserved names doesn't
+collide and sets the "KindNameConflict" condition to `false`.
+
+A custom `http.Handler` will look at request and use the parsed out
+GroupVersionResource information to match it to a ThirdPartyResource.  The ThirdPartyResource
+will be checked to make sure its valid enough in .Status to serve and will
+response appropriated.  If there is no ThirdPartyResource defined, it will delegate
+to the next handler in the chain.
+
+### Delete
+When a TPR-registration is deleted, it will be handled as a finalizer like a
+namespace is done today.  The `Terminating` condition will be updated (like
+namespaces) and that will cause mutating requests to be rejected by the REST
+handler (see above).  The finalizer will remove all the associated storage.
+Once the finalizer is done, it will delete the TPR-registration itself.
+
+
+## Migration from existing TPR
+Because of the changes required to meet the goals, there is not a silent
+auto-migration from the existing TPR to the new TPR.  It will be possible, but
+it will be manual.  At a high level, you simply:
+ 1. Stop all clients from writing to TPR (revoke edit rights for all users) and
+ stop controllers.
+ 2. Get all your TPR-data.  
+ `$ kubectl get TPR --all-namespaces -o yaml > data.yaml`
+ 3. Delete the old TPR-data.  Be sure you orphan!  
+ `$ kubectl delete TPR --all --all-namespaces --cascade=false`
+ 4. Delete the old TPR-registration.  
+ `$ kubectl delete TPR/name`
+ 5. Create a new TPR-registration with the same GroupVersionKind as before.  
+ `$ kubectl create -f new_tpr.name`
+ 6. Recreate your new TPR-data.  
+ `$ kubectl create -f data.yaml`
+ 7. Restart controllers.
+
+There are a couple things that you'll need to consider:
+ 1. Garbage collection.  You may have created links that weren't respected by
+ the GC collector in 1.6.  Since you orphaned your dependents, you'll probably
+ want to re-adopt them like the Kubernetes controllers do with their resources.
+ 2. Controllers will observe deletes.  Part of this migration actually deletes
+ the resource.  Your controller will see the delete.  You ought to shut down
+ your TPR controller while you migrate your data.  If you do this, your
+ controller will never see a delete.
+


contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+In order to: 
+1. eliminate opaquely derived information - deriving camel-cased kind names
+from lower-case dash-delimited values as for instance.
+1. allow the expression of complex transformations - not all plurals are easily
+determined (ox and oxen) and not all are English.  Fields for complete
+specification eliminates ambiguity.
+1. handle TPR-registration value-space conflicts
+1. [stop using the extensions API group](https://github.com/kubernetes/kubernetes/issues/43214)
+
+We can create a type `ThirdPartyResource.apiextension.k8s.io`.
```
philips: nit: kubernetes.io?

philips:
cc @aaronlevy @derekparker @lpabon @ericchiang @yifan-gu

philips:
cc @xiang90 @hongchaodeng

contributors/design-proposals/thirdpartyresources.md
liggitt: API groups are k8s.io

philips: gah, right. consistency FTW.

deads2k:
> @deads2k can you outline how the on-the-wire API changes in order to accommodate these suggested changes? e.g. http://127.0.0.1:8080/apis/etcd.coreos.com/v1beta1/namespaces/default/clusters/example-etcd-cluster becomes ...?

That format will not change.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+In "normal" Kubernetes APIs, if I have a persisted resource in the same group
+with the same name in v1 and v2, they are backed by the same underlying object.
+A change made to one is reflected in the other. API clients, garbage collection,
+namespace cleanup, version negotiation, and controllers all build on this.
+
+The convertibility of Kubernetes APIs provides a seamless interaction between
+versions.  A TPR does not have the ability to convert between versions, which
+focuses on the primary role of TPR as an easily extensible and simple mechanism
+for adding new APIs. Conversion primarily allows structural, but not backwards
+incompatible, changes. By not supporting conversion, all TPR use cases are
```
lpabon: Q: I thought the version conversions was a good thing, but you mention this is a bad thing and should be avoided because it does not preserve behavior, right?

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// ThirdPartyResourceList is a list of ThirdPartyResource objects.
+type ThirdPartyResourceList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Items individual ThirdParties
+	Items []ThirdPartyResource `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
```
lpabon: specs LGTM

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+### Create
+When a new TPR is created, no synchronous action is taken.
+A controller will run to confirm that value-space of the reserved names doesn't
+collide and sets the "KindNameConflict" condition to `false`.
+
+A custom `http.Handler` will look at request and use the parsed out
+GroupVersionResource information to match it to a ThirdPartyResource.  The ThirdPartyResource
+will be checked to make sure its valid enough in .Status to serve and will
+response appropriated.  If there is no ThirdPartyResource defined, it will delegate
+to the next handler in the chain.
```
lpabon: In issue https://github.com/kubernetes/features/issues/95 it mentions that it could take up to 10 seconds to see a new TPR. How will _Create_ help reduce that? Maybe not one of the goals for this change?

deads2k: > In issue kubernetes/features#95 it mentions that it could take up to 10 seconds to see a new TPR. How will Create help reduce that? Maybe not one of the goals for this change?

That's more of an implementation detail, but given a spec/status split, it's possible write a series of small, "normal", controllers that are easy to reason about and write a different kind of RESTHandler (re-using our impl code) to drive it all from a shared cache that would end up more consistent.

contributors/design-proposals/thirdpartyresources.md
deads2k: > Q: I thought the version conversions was a good thing, but you mention this is a bad thing and should be avoided because it does not preserve behavior, right?

Conversions can allow a change of expression over time (rename fieldA to fieldB for example), but they can't change behavior (do something different with fieldA).  Someone who owns a TPR could change expression over time client-side by being tolerant for a couple releases and rewriting the objects periodically over that time.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+ 5. Create a new TPR-registration with the same GroupVersionKind as before.  
+ `$ kubectl create -f new_tpr.name`
+ 6. Recreate your new TPR-data.  
+ `$ kubectl create -f data.yaml`
+ 7. Restart controllers.
+
+There are a couple things that you'll need to consider:
+ 1. Garbage collection.  You may have created links that weren't respected by
+ the GC collector in 1.6.  Since you orphaned your dependents, you'll probably
+ want to re-adopt them like the Kubernetes controllers do with their resources.
```
enisoc: It's unlikely but possible that someone else may have adopted them in the meantime. For example, suppose my TPR-based controller created ReplicaSets that happen to match the selector of some other Deployment.

There may not be an easy way to systematically prevent this (hard ways include reparenting to a dummy object), but it should probably be mentioned as a caveat for anyone following these migration instructions.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+for adding new APIs. Conversion primarily allows structural, but not backwards
+incompatible, changes. By not supporting conversion, all TPR use cases are
+preserved, but a large amount of complexity is avoided for consumers of TPR.
+
+Allowing a single, user specified version for a given TPR will provide this
+semantic by preventing server-side versioning altogether.  All instances of a
+single TPR must have the same version or the Kubernetes API semantic of always
+returning a resource encoded to the matching version will not be maintained.
+Since conversions (even native Kubernetes conversions) cannot be used to handle
+behavioral changes, the same effect can be achieved for TPRs client-side with
```
enisoc: I think it would be helpful to write out an example flow for handling version upgrades of a TPR client-side, similar to the tpr.extensions->tpr.apiextension migration at the bottom.

One thing that may be worth having a better plan for is how TPR clients (e.g. Operators) can upgrade TPR data safely. For the initial migration to tpr.apiextension, it's perhaps reasonable to require offline migration, but version bumps within a given TPR (e.g. EtcdCluster v1beta1 -> v1beta2) will be more common.

For example, I'd want to make sure the framework exists to let an Operator do its own auto-upgrade of the stored TPR data. If that must be done by downloading all TPR data before deregistering the old TPR version, where is it going to store that data to survive crashes?

deads2k: > version bumps within a given TPR (e.g. EtcdCluster v1beta1 -> v1beta2) will be more common.

The point of my description is that its actually possible to mutate within a rolling version of your choosing which preserves backwards compatibility.  That is essentially what we do with server-side conversions, but with just slightly more sugar.

If you need to change behavior or wish to break backwards compatibility, TPRs end up in the same boat as normal resources have to deal with a logical move to a different resource or some kind of off-line change.  It's the same regardless of TPR-ness or not.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+	// Group is the group this resource belongs in
+	Group string `json:"group" protobuf:"bytes,1,opt,name=group"`
+	// Version is the version this resource belongs in
+	Version string `json:"version" protobuf:"bytes,2,opt,name=version"`
+	// Plural is the plural name of the resource
+	Plural string `json:"plural" protobuf:"bytes,3,opt,name=plural"`
+	// Singular is the singular name of the resource.  Defaults to lowercased <kind>
+	Singular string `json:"singular,omitempty" protobuf:"bytes,4,opt,name=singular"`
+	// ShortNames are short names for the resource.
+	ShortNames []string `json:"shortNames,omitempty" protobuf:"bytes,5,opt,name=shortNames"`
```
deads2k: > Does the plumbing exist yet to actually make this work?

Yes, `kubectl` respects it.  Since shortNames are group scoped, there's no need for a central registry (each group owns their own) and given `status` TPRs do the same.

contributors/design-proposals/thirdpartyresources.md
enisoc: To make sure I understand, is the following what you're proposing?

1. Create TPR with some arbitrary k8s-level API version string, but don't change it for anything that's backwards compatible, if you want to be able to upgrade seamlessly (automated by the Operator).
2. If you need to track data version to help you automatically migrate between backwards-compatible representations, you should record it in your own field within the TPR data.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+proposal covers the steps we believe are necessary to move TPR to beta and to
+prevent future challenges in upgrading.
+
+
+## Goals
+1. Ensure ThirdPartyResource APIs operate consistently with first party
+Kubernetes APIs.
+2. Enable ThirdPartyResources to specify how they will appear in API
+discovery to be consistent with other resources and avoid naming confilcts
+3. Move TPR into their own API group to allow the extensions group to be
```
lavalamp: arguably "extensions" is a good place for TPRs? Move everything else out instead? :)

contributors/design-proposals/thirdpartyresources.md
enisoc: Ah right I forgot that long names are allowed to overlap too, and kubectl guesses which one you mean if you don't specify a group (based on discovery order?).

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+[removed](https://github.com/kubernetes/kubernetes/issues/43214)
+4. Support cluster scoped TPR resources
+5. Identify other features required for TPR to become beta
+6. Minimize the impact to alpha ThirdPartyResources consumers and define a
+process for how TPR migrations / breaking changes can be accomplished (for
+both the cluster and for end users)
+
+Non-goals
+1. Solve automatic conversion of TPR between versions or automatic migration of
+existing TPR
```
lavalamp: It's unclear to me if this and previous references to TPR migration are talking about TPR data objects or TPR spec objects.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+namespace cleanup, version negotiation, and controllers all build on this.
+
+The convertibility of Kubernetes APIs provides a seamless interaction between
+versions.  A TPR does not have the ability to convert between versions, which
+focuses on the primary role of TPR as an easily extensible and simple mechanism
+for adding new APIs. Conversion primarily allows structural, but not backwards
+incompatible, changes. By not supporting conversion, all TPR use cases are
+preserved, but a large amount of complexity is avoided for consumers of TPR.
+
+Allowing a single, user specified version for a given TPR will provide this
```
lavalamp: *shrug* I really don't have a problem with allowing the user to register v1 and v2 and manually convert. I think *we* will need to do that for some resources eventually.

contributors/design-proposals/thirdpartyresources.md
ncdc: Ha, I've had this exact same thought

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+to properly distinguish intent.
+
+The actual name of the TPR-registration (resource that describes the TPR to
+create) resource can only protect one of these values from conflict.  Since
+Kubernetes API types are accessed via a URL that looks like `/apis/<group>/<version>/namespaces/<namespace-name>/<plural-resource-type>`,
+the name of the TPR-registration object will be `<plural-resource-type>.<group>`.
+
+Conflicts with other parts of the value-space can not be detected with static
+validation, so there will be a spec/status split with `status.conditions` that
+reflect the acceptance status of a TPR-registration.  For instance, you cannot
```
lavalamp: I guess I would expect to do this with an admission controller / initializer.

contributors/design-proposals/thirdpartyresources.md
liggitt: that makes TPR behave differently than every other kube resource, and breaks things like garbage collection and namespace cleanup, right?

contributors/design-proposals/thirdpartyresources.md
enisoc: The comment on https://github.com/kubernetes/kubernetes/issues/43214#issuecomment-287143011 seems to imply that there's something special about `extensions` such that "it doesn't support multiple API versions." Is that true?

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Avoiding Naming Problems
+There are several identifiers that a Kubernetes API resource has which share
+value-spaces within an API group and must not conflict.  They are:
+1. Resource-type value space
+  1. plural resource-type name - like "configmaps"
+  2. singular resource-type name - like "configmap"
+  3. short names - like "cm"
+2. Kind-type value space - for group "example.com"
+  1. Kind name - like "ConfigMap"
```
lavalamp: It is not important for this to be unique, since multiple TPRs could use the same kind.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+
+Conflicts with other parts of the value-space can not be detected with static
+validation, so there will be a spec/status split with `status.conditions` that
+reflect the acceptance status of a TPR-registration.  For instance, you cannot
+determine whether two TPRs in the same group have the same short name without
+inspecting the current state of existing TPRs.
+
+Parts of the value-space will be "claimed" by making an entry in TPR.status to
+include the accepted names which will be served.  This prevents a new TPR from
+disabling an existing TPR's name.
```
lavalamp: short name is not important because no one should write scripts using the short name, if two objects collide then the spec/discovery assembly code can just serve neither short name (or both and let clients complain).

Where is the singular resource name used? I am not convinced we need to write a big system to keep it unique.

Honestly this is a lot of work/complexity all to handle a case where two different plural resource types happen to try and use the same singular name.

liggitt: plural resource and kind must be unique.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+	Version string `json:"version" protobuf:"bytes,2,opt,name=version"`
+	// Plural is the plural name of the resource
+	Plural string `json:"plural" protobuf:"bytes,3,opt,name=plural"`
+	// Singular is the singular name of the resource.  Defaults to lowercased <kind>
+	Singular string `json:"singular,omitempty" protobuf:"bytes,4,opt,name=singular"`
+	// ShortNames are short names for the resource.
+	ShortNames []string `json:"shortNames,omitempty" protobuf:"bytes,5,opt,name=shortNames"`
+	// Kind is the serialized kind of the resource
+	Kind string `json:"kind" protobuf:"bytes,6,opt,name=kind"`
+	// ListKind is the serialized kind of the list for this resource.  Defaults to <kind>List
```
lavalamp: Do we have any exceptions to this derivation? I'm not sure we want to allow customizing this?

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+	ListKind string `json:"listKind,omitempty" protobuf:"bytes,7,opt,name=listKind"`
+
+	// ClusterScoped indicates that this resource is cluster scoped as opposed to namespace scoped
+	ClusterScoped bool `json:"clusterScoped" protobuf:"bytes,8,opt,name=clusterScoped"`
```
lavalamp: This should be Scope = {Cluster, Namespaced} enum

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+We can create a type `ThirdPartyResource.apiextension.k8s.io`.
+```go
+// ThirdPartyResourceSpec describe how a user wants their resource to appear
+type ThirdPartyResourceSpec struct {
+	// Group is the group this resource belongs in
+	Group string `json:"group" protobuf:"bytes,1,opt,name=group"`
+	// Version is the version this resource belongs in
+	Version string `json:"version" protobuf:"bytes,2,opt,name=version"`
+	// Plural is the plural name of the resource
+	Plural string `json:"plural" protobuf:"bytes,3,opt,name=plural"`
```
lavalamp: must match name of resource

contributors/design-proposals/thirdpartyresources.md
deads2k: > I guess I would expect to do this with an admission controller / initializer.

Admission is opt-in by the cluster-admin. This behavior shouldn't be enable/disable-able (admission is).  Also, I prefer the locality.

contributors/design-proposals/thirdpartyresources.md
deads2k: > It is not important for this to be unique, since multiple TPRs could use the same kind.

Given how a restmapper works with serialized objects, the kind must be unique per group or weird/bad things will happen.

contributors/design-proposals/thirdpartyresources.md
deads2k: > Do we have any exceptions to this derivation? I'm not sure we want to allow customizing this?

Non-english comes to mind.

contributors/design-proposals/thirdpartyresources.md
deads2k: > Where is the singular resource name used? I am not convinced we need to write a big system to keep it unique.

Singular is used on the CLI.  The size is subjective, this ends up being pretty small and keeps us consistent with the first party resources.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+// ThirdPartyResourceConditionType is a valid value for ThirdPartyResourceCondition.Type
+type ThirdPartyResourceConditionType string
+
+const (
+	// ResourceNameConflict means the resource names chosen for this ThirdPartyResource conflict with others in the group.
+	// The first TPR in the group to have the name reflected in status "wins" the name.
+	ResourceNameConflict ThirdPartyResourceConditionType = "ResourceNameConflict"
+	// KindNameConflict means the kind names chosen for this ThirdPartyResource conflict with others in the group.
+	// The first TPR in the group to have the name reflected in status "wins" the name.
+	KindNameConflict ThirdPartyResourceConditionType = "KindNameConflict"
```
lavalamp: maybe just one name conflict condition, and the reason/message can say which specifically is wrong.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+type ConditionStatus string
+
+// These are valid condition statuses. "ConditionTrue" means a resource is in the condition.
+// "ConditionFalse" means a resource is not in the condition. "ConditionUnknown" means kubernetes
+// can't decide if a resource is in the condition or not. In the future, we could add other
+// intermediate conditions, e.g. ConditionDegraded.
+const (
+	ConditionTrue    ConditionStatus = "True"
+	ConditionFalse   ConditionStatus = "False"
+	ConditionUnknown ConditionStatus = "Unknown"
```
lavalamp: Is this stuff not defined by the v1 api?

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+	Message string `json:"message,omitempty" protobuf:"bytes,6,opt,name=message"`
+}
+
+// ThirdPartyResourceStatus indicates the state of the ThirdPartyResource
+type ThirdPartyResourceStatus struct {
+	// Conditions indicate state for particular aspects of a ThirdPartyResource
+	Conditions []ThirdPartyResourceCondition `json:"conditions" protobuf:"bytes,1,opt,name=conditions"`
+
+	// Singular is the singular name of the resource.  Defaults to lowercased <kind>
+	Singular string `json:"singular,omitempty" protobuf:"bytes,2,opt,name=singular"`
```
lavalamp: Consider grouping all of these into a "naming" struct, which can be present in both status & spec instead of copying everything individually.

contributors/design-proposals/thirdpartyresources.md
adohe: all third party resources will live in the same group?

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Goals
+1. Ensure ThirdPartyResource APIs operate consistently with first party
+Kubernetes APIs.
+2. Enable ThirdPartyResources to specify how they will appear in API
+discovery to be consistent with other resources and avoid naming confilcts
+3. Move TPR into their own API group to allow the extensions group to be
+[removed](https://github.com/kubernetes/kubernetes/issues/43214)
+4. Support cluster scoped TPR resources
+5. Identify other features required for TPR to become beta
```
adohe: like multi-version support, more custom specific validation.

contributors/design-proposals/thirdpartyresources.md
adohe: Ha, I have the same question.

contributors/design-proposals/thirdpartyresources.md
adohe: seems we don't plan to support multi version :(

deads2k: > seems we don't plan to support multi version

I'm not fundamentally opposed, but I did lay out the REST API semantics that prevent it from working like the first-party kube APIs.  @lavalamp indicated that he wanted to change that restriction in first-party kube APIs, but if that's the case I think the first-party APIs should move first and the third-party APIs can change afterwards.

contributors/design-proposals/thirdpartyresources.md
deads2k: > Is this stuff not defined by the v1 api?

I think I would encourage API groups to define their own types unless they actually need to interoperate.

contributors/design-proposals/thirdpartyresources.md
deads2k: > all third party resources will live in the same group?

No, this is just the description of the TPR you want to store.  TPR-registration.

deads2k: > The comment on kubernetes/kubernetes#43214 (comment) seems to imply that there's something special about extensions such that "it doesn't support multiple API versions." Is that true?

It's a really confusing group.  Not everything it's version is actually beta.  I agree with that issue where Brian says we should just retire it.

contributors/design-proposals/thirdpartyresources.md
deads2k: > Consider grouping all of these into a "naming" struct, which can be present in both status & spec instead of copying everything individually.

done

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,251 @@
...
+
+
+## Goals
+1. Ensure ThirdPartyResource APIs operate consistently with first party
+Kubernetes APIs.
+2. Enable ThirdPartyResources to specify how they will appear in API
+discovery to be consistent with other resources and avoid naming confilcts
+3. Move TPR into their own API group to allow the extensions group to be
+[removed](https://github.com/kubernetes/kubernetes/issues/43214)
+4. Support cluster scoped TPR resources
```
deads2k: > we did want this, not every resource is namespace scoped, does this mean we may want a generic mechanism to define cluster scoped resources?

Yes, its described in this doc.

contributors/design-proposals/thirdpartyresources.md
deads2k: > maybe just one name conflict condition, and the reason/message can say which specifically is wrong.

done

deads2k:
struct updated for comments.

contributors/design-proposals/thirdpartyresources.md
lavalamp: If you implement these in their own apiserver, that doesn't need to be true.

contributors/design-proposals/thirdpartyresources.md
lavalamp: Personally, I don't think it makes a lot of sense to internationalize the type names.

contributors/design-proposals/thirdpartyresources.md
lavalamp: I disagree, I think we want conditions to be as close to being generically programmatically consumable as possible.

contributors/design-proposals/thirdpartyresources.md
deads2k: > If you implement these in their own apiserver, that doesn't need to be true.

Technically true, but very confusing.  Especially as we look to compose API servers.

In addition, most resources (all except RBAC where its a security problem?), handle "any chance of working" validation synchronously and "is it working now" asynchronously.  I think the model works well and applies cleanly here.

contributors/design-proposals/thirdpartyresources.md
deads2k: > I disagree, I think we want conditions to be as close to being generically programmatically consumable as possible.

That sounds more like a polymorphic `/conditions` subresource that could be programatically discovered and used like HPA than attempting to force a particular serialization format on kinds.

contributors/design-proposals/thirdpartyresources.md
deads2k: > Personally, I don't think it makes a lot of sense to internationalize the type names.

Derived, uninspectable, unchangeable, unfixable field tend to bite people.  This is information we need to properly serve the endpoint. I don't mind defaulting it (noted in the doc), but I really don't want to say that there's no way for a user to specify it if there is some need.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,253 @@
...
+includes feedback from existing Kubernetes ThirdPartyResource users. This
+proposal covers the steps we believe are necessary to move TPR to beta and to
+prevent future challenges in upgrading.
+
+
+## Goals
+1. Ensure ThirdPartyResource APIs operate consistently with first party
+Kubernetes APIs.
+2. Enable ThirdPartyResources to specify how they will appear in API
+discovery to be consistent with other resources and avoid naming confilcts
```
derekwaynecarr: nit: s/confilcts/conflicts

contributors/design-proposals/thirdpartyresources.md
derekwaynecarr: manual conversion means we stored two different distinct values?  i hope we don't do that... i could see offering some transformation hook, but hope we do not start persisting distinct entities per version.  if we did that how would resource version work?  possible i am misunderstanding what @lavalamp was implying.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,253 @@
...
+	ConditionUnknown ConditionStatus = "Unknown"
+)
+
+// ThirdPartyResourceConditionType is a valid value for ThirdPartyResourceCondition.Type
+type ThirdPartyResourceConditionType string
+
+const (
+	// NameConflict means the resource or kind names chosen for this ThirdPartyResource conflict with others in the group.
+	// The first TPR in the group to have the name reflected in status "wins" the name.
+	NameConflict ThirdPartyResourceConditionType = "NameConflict"
```
enisoc: I would expect in most cases, the real Condition clients care about is, "Is this TPR actually installed in the API server?". A name conflict seems more like a particular Reason for the `Installed` condition to have a status of `ConditionFalse`. In the future, there may be other Reasons besides name conflict that a TPR would have condition `Installed:False`, and clients shouldn't need to know to check for those specifically.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,253 @@
...
+1. [stop using the extensions API group](https://github.com/kubernetes/kubernetes/issues/43214)
+
+We can create a type `ThirdPartyResource.apiextension.k8s.io`.
+```go
+// ThirdPartyResourceSpec describe how a user wants their resource to appear
+type ThirdPartyResourceSpec struct {
+	// Group is the group this resource belongs in
+	Group string `json:"group" protobuf:"bytes,1,opt,name=group"`
+	// Version is the version this resource belongs in
+	Version string `json:"version" protobuf:"bytes,2,opt,name=version"`
```
enisoc: If we later support multiple versions (e.g. because the first-party APIs evolve to allow the content of different versions to differ), can this Spec be changed to allow multiple versions in a backwards-compatible way?

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,253 @@
...
+reflect the acceptance status of a TPR-registration.  For instance, you cannot
+determine whether two TPRs in the same group have the same short name without
+inspecting the current state of existing TPRs.
+
+Parts of the value-space will be "claimed" by making an entry in TPR.status to
+include the accepted names which will be served.  This prevents a new TPR from
+disabling an existing TPR's name.
+
+
+## New API
```
enisoc: Since we are moving to a new API group, this seems like the time to ask whether everyone is happy with the name ThirdPartyResource. Is it going to make sense given all the ways we now expect this to be used in the future?

For example, I've seen TPR proposed as a way to implement new first-party (included in Kubernetes) objects like [BatchJob](https://docs.google.com/document/d/1-H2hnZap7gQivcSU-9j4ZrJ8wE_WwcfOkTeAGjzUyLA/edit) in order to avoid bloating the core API. TPR has also been proposed as a way to store data for objects implemented by aggregated API servers, in cases when those aggregated servers don't want to run their own datastore.

Given that, the defining feature of TPR is not that the resources are necessarily third-party, but rather that they are not compiled in, and can be added and removed on a running cluster.

Just brainstorming:
* DynamicResource (cf. [dynamic client](https://godoc.org/k8s.io/client-go/dynamic))
* ExtensionResource
* PluginResource

ncdc: While a new name might make sense, I'd recommend we defer any discussion about it until all the other issues are resolved.

philips:
On Tue, Apr 11, 2017 at 4:46 AM David Eads <notifications@github.com> wrote:

> @deads2k <https://github.com/deads2k> can you outline how the on-the-wire
> API changes in order to accommodate these suggested changes? e.g.
> http://127.0.0.1:8080/apis/etcd.coreos.com/v1beta1/namespaces/default/clusters/example-etcd-cluster
> becomes ...?
>
> That format will not change.
>
What would prevent the API server from not doing the data dump and restore
internally to do a migration?

Brandon

deads2k:
> What would prevent the API server from not doing the data dump and restore
> internally to do a migration?

I'm not sure I understand the question, but I think you're asking why the migration steps are manual instead of automatic?

Because of the interaction with controllers, there needs to be a coordinated change.  Coordinating with external processes is beyond the scope of what our API server does.  I want to make sure the migration is crisply described, but I don't think it can be baked into the API server itself.

contributors/design-proposals/thirdpartyresources.md
deads2k: > I would expect in most cases, the real Condition clients care about is, "Is this TPR actually installed in the API server?". A name conflict seems more like a particular Reason for the Installed condition to have a status of ConditionFalse. In the future, there may be other Reasons besides name conflict that a TPR would have condition Installed:False, and clients shouldn't need to know to check for those specifically.

Not all name conflicts prevent the TPR from being served.  ShortNames are a good example.  If you added a conflicting ShortName, your resource is still servable, you just can't use the ShortName to refer to it.

I'm fine adding an `Available` condition, but this isn't it and available doesn't have sufficient power to describe the states.

contributors/design-proposals/thirdpartyresources.md
deads2k: > can this Spec be changed to allow multiple versions in a backwards-compatible way?

`AdditionalVersions []string` along with a new TPR version (v1alpha2?) which combines them into a single field.  It is definitely extensible.

contributors/design-proposals/thirdpartyresources.md
foxish: It looks like the short names for the exsiting resource types are in [kubectl](https://github.com/kubernetes/kubernetes/blob/2aa4abb73b305c7c868107831bc007aa1c952f86/pkg/kubectl/kubectl.go#L46) for the default resources. I remember having a conversation with @bgrant0607 about moving these out into the discovery API.

contributors/design-proposals/thirdpartyresources.md
foxish: Yeah, makes sense to me as well and we should discuss it after the other parts of this proposal are agreed on.
I'd like to add 'APIExtension' to the list of options. (I'm in favor of dropping the 'Resource' suffix.)

contributors/design-proposals/thirdpartyresources.md
deads2k: > It looks like the short names for the exsiting resource types are in kubectl for the default resources. I remember having a conversation with @bgrant0607 about moving these out into the discovery API.

A community member, @p0lyn0mial , moved the shortNames to the api discovery endpoint in 1.6.

contributors/design-proposals/thirdpartyresources.md
lpabon: Cool, thanks

contributors/design-proposals/thirdpartyresources.md
lpabon: Cool

deads2k:
I think we've made it through the structural concerns.  I think we can continue hashing out names when the types pull gets opened up.  We get all the way until code freeze for changes there.

enisoc:
Merging the proposal so we can propose working on it in the 1.7 cycle. We can continue to iterate on the details in later reviews.

contributors/design-proposals/thirdpartyresources.md
```diff
@@ -0,0 +1,253 @@
...
+namespaces) and that will cause mutating requests to be rejected by the REST
+handler (see above).  The finalizer will remove all the associated storage.
+Once the finalizer is done, it will delete the TPR-registration itself.
+
+
+## Migration from existing TPR
+Because of the changes required to meet the goals, there is not a silent
+auto-migration from the existing TPR to the new TPR.  It will be possible, but
+it will be manual.  At a high level, you simply:
+ 1. Stop all clients from writing to TPR (revoke edit rights for all users) and
```
ericchiang: @deads2k is it going to be a requirement to coordinate this cluster wide for all TPRs, or will individual controllers be able to migrate on a per-TPR basis?

enisoc: As I understand it, the plan is that existing TPR (`extensions/v1beta1`) will continue to work side-by-side with the new TPR (`apiextensions/v1beta1`) in v1.7. So, it will be possible to migrate on a per-TPR or per-controller basis, assuming you don't have interdependencies between multiple TPRs/controllers.

deads2k: > As I understand it, the plan is that existing TPR (extensions/v1beta1) will continue to work side-by-side with the new TPR (apiextensions/v1beta1) in v1.7. So, it will be possible to migrate on a per-TPR or per-controller basis, assuming you don't have interdependencies between multiple TPRs/controllers.

Exactly right.

adohe:
@deads2k @enisoc any specific work items list? I would like to help with TPR related things.

deads2k:
> @deads2k @enisoc any specific work items list? I would like to help with TPR related things.

I think the biggest pain point is likely to be the migration from original TPR to new TPR.  @enisoc is out this week, so we won't be able to coordinate fully, but I was hoping he could take the lead on making the transition easier.  You have significant user experience expertise too, so help with the design (I just sketched it in this doc) and implementation of actually transitioning a TPR from old to new is an open area.

I've got most of an implementation running and I've started making small-ish pulls to bring it in.  If you have review time, getting 90% types https://github.com/kubernetes/kubernetes/pull/45115 and iterating on them would be useful.

adohe:
Get it and thanks very much @deads2k  :)

PR #527, opened 2017-04-11
Proposal for DaemonSet history and rollback

janetkuo:
In summary:
1. Use PodTemplate to store DaemonSet history 
1. Use "pod-template-generation" to label both pods and PodTemplates, so that DaemonSet controller can map a pod to a PodTemplate (history), and than compare PodTemplate with DaemonSet template 
1. Each PodTemplate is a snapshot of DaemonSet template + annotation + templateGeneration at the moment. A DaemonSet can have multiple PodTemplates with the same template (each with a unique pod-template-generation label)
1. Default clean up policy for DaemonSet history is 2. Only PodTemplates with no existing pods map to it can be cleaned up (so that a pod can always be mapped to a PodTemplate). 
1. Rollback will be done server side using RollbackConfig, similar to Deployment 

An alternative is to create a DaemonSetHistory resource, and/or use template hash instead of template generation. 

@erictune @kow3ns @foxish @enisoc @kargakis @lukaszo @smarterclayton @bgrant0607 @kubernetes/sig-apps-api-reviews

### DIFF
--- a/contributors/design-proposals/daemonset-update.md
+++ b/contributors/design-proposals/daemonset-update.md
@@ -126,25 +126,8 @@ type DaemonSetSpec struct {
 
 	// The number of old history to retain to allow rollback.
 	// This is a pointer to distinguish between explicit zero and not specified.
-	// Defaults to 2.
+	// Defaults to 10.
 	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty"`
-
-	// The config this DaemonSet is rolling back to. Will be cleared after
-	// rollback is done.
-	// +optional
-	RollbackTo *RollbackConfig `json:"rollbackTo",omitempty`
-}
-
-// DaemonSetRollback stores the information required to rollback a DaemonSet.
-type DaemonSetRollback struct {
-	metav1.TypeMeta `json:",inline"`
-	// Required: This must match the Name of a DaemonSet.
-	Name string `json:"name"`
-	// The annotations to be updated to a DaemonSet
-	// +optional
-	UpdatedAnnotations map[string]string `json:"updatedAnnotations,omitempty"`
-	// The config of this DaemonSet rollback.
-	RollbackTo RollbackConfig `json:"rollbackTo"`
 }
 
 // DaemonSetStatus represents the current status of a daemon set.
@@ -170,11 +153,11 @@ type DaemonSetStatus struct {
 	// +optional
 	NumberUnavailable int32 `json:"numberUnavailable"`
 
-	// A monotonically increasing counter that tracks hash collisions for
-	// the DaemonSet. Used as a collision avoidance mechanism by the 
-	// DaemonSet controller.
+	// Count of hash collisions for the DaemonSet. The DaemonSet controller
+	// uses this field as a collision avoidance mechanism when it needs to
+	// create the name for the newest ControllerRevision.
 	// +optional
-	Uniquifier *int64 `json:"uniquifier,omitempty"`
+	CollisionCount *int64 `json:"collisionCount,omitempty"`
 }
 
 const (
@@ -187,7 +170,7 @@ const (
 	// DefaultDaemonSetUniqueLabelKey is the default label key that is added
 	// to existing DaemonSet pods to distinguish between old and new
 	// DaemonSet pods during DaemonSet template updates.
-	DefaultDaemonSetUniqueLabelKey string = "pod-template-hash"
+	DefaultDaemonSetUniqueLabelKey string = "daemonset-controller-hash"
 )
 ```
 
@@ -198,57 +181,61 @@ const (
 The DaemonSet Controller will make DaemonSet updates happen. It will watch
 DaemonSets on the apiserver. 
 
-DaemonSet controller manages [Controller History](controller_history.md) for
+DaemonSet controller manages [`ControllerRevisions`](controller_history.md) for
 DaemonSet revision introspection and rollback. It's referred to as "history"
 throughout the rest of this proposal.
 
 For each pending DaemonSet updates, it will:
 
-1. Find all pods controlled by DaemonSet
+1. Reconstruct DaemonSet history:
+   - List existing DaemonSet history controlled by this DaemonSet
+   - Find the history of DaemonSet's current target state, and create one if
+     not found:
+     - The `.name` of this history will be unique, generated from pod template
+       hash with hash collision resolution. If history creation failed:
+       - If it's because of name collision:
+         - Compare history with DaemonSet current target state:
+           - If they're the same, we've already created the history
+           - Otherwise, bump DaemonSet `.status.collisionCount` by 1, exit and
+             retry in the next sync loop
+       - Otherwise, exit and retry again in the next sync loop.
+     - The history will be labeled with `DefaultDaemonSetUniqueLabelKey`.
+     - DaemonSet controller will add a ControllerRef in the history
+       `.ownerReferences`.
+   - Current history should have the largest `.revision` number amonst all
+     existing history. Update `.revision` if it's not (e.g. after a rollback.)
+   - If more than one current history is found, remove duplicates and relabel
+     their pods' `DefaultDaemonSetUniqueLabelKey`.
+1. Sync nodes:
+   - Find all nodes that should run these pods created by this DaemonSet.
+   - Create daemon pods on nodes when they should have those pods running but not
+     yet. Otherwise, delete running daemon pods that shouldn't be running on nodes.
+   - Label new pods with current `.spec.templateGeneration` and
+     `DefaultDaemonSetUniqueLabelKey` value of current history when creating them.
 1. Check `DaemonSetUpdateStrategy`:
    - If `OnDelete`: do nothing
    - If `RollingUpdate`:
-     - Reconstruct DaemonSet history: filter and sort existing DaemonSet history
-       owned by this DaemonSet
-     - Find the history of DaemonSet's current target state, and create one if
-       not found:
-       - The `.name` of this history will be unique, generated from pod template
-         hash with hash collision resolution. If history creation failed:
-         - If it's because of name collision:
-           - Compare history with DaemonSet current target state:
-             - If they're the same, we've already created the history
-             - Otherwise, bump DaemonSet `.status.uniquifier` by 1, exit and
-               retry in the next sync loop
-         - Otherwise, exit and retry again in the next sync loop.
-       - The history will be labeled with `pod-template-hash`.
-       - DaemonSet controller will add a ControllerRef in the history
-         `.ownerReferences`.
      - For all pods owned by this DaemonSet:
        - If its `pod-template-generation` label value equals to DaemonSet's 
          `.spec.templateGeneration`, it's a new pod (don't compare
-         `pod-template-hash`, for backward compatibility).
-         - Add `pod-template-hash` label to the new pod based on current
+         `DefaultDaemonSetUniqueLabelKey`, for backward compatibility).
+         - Add `DefaultDaemonSetUniqueLabelKey` label to the new pod based on current
            history, if the pod doesn't have this label set yet. 
        - Otherwise, if the value doesn't match, or the pod doesn't have a
-         `pod-template-generation` label, check its `pod-template-hash` label:
-         - If the value matches any of the history's `pod-template-hash` label,
+         `pod-template-generation` label, check its `DefaultDaemonSetUniqueLabelKey` label:
+         - If the value matches any of the history's `DefaultDaemonSetUniqueLabelKey` label,
            it's a pod generated from that history.
            - If that history matches the current target state of the DaemonSet,
              it's a new pod.
            - Otherwise, it's an old pod. 
-         - Otherwise, if the pod doesn't have a `pod-template-hash` label, or no
+         - Otherwise, if the pod doesn't have a `DefaultDaemonSetUniqueLabelKey` label, or no
            matching history is found, it's an old pod. 
      - If there are old pods found, compare `MaxUnavailable` with DaemonSet
        `.status.numberUnavailable` to see how many old daemon pods can be
        killed. Then, kill those pods in the order that unhealthy pods (failed,
        pending, not ready) are killed first.
-     - Clean up history based on `.spec.revisionHistoryLimit`
-       - Always keep live history
-1. Find all nodes that should run these pods created by this DaemonSet.
-1. Create daemon pods on nodes when they should have those pods running but not
-   yet. Otherwise, delete running daemon pods that shouldn't be running on nodes.
-   - Label new pods with current `.spec.templateGeneration` and
-     `pod-template-hash` value of current history when creating them.
+1. Clean up old history based on `.spec.revisionHistoryLimit`
+   - Always keep live history and current history
 1. Cleanup, update DaemonSet status  
    - `.status.numberAvailable` = the total number of DaemonSet pods that have
      become `Ready` for `MinReadySeconds`
@@ -263,6 +250,8 @@ In DaemonSet strategy (pkg/registry/extensions/daemonset/strategy.go#PrepareForU
 increase DaemonSet's `.spec.templateGeneration` by 1 if any changes is made to
 DaemonSet's `.spec.template`.
 
+This was originally implmeneted in 1.6, and kept in 1.7 for backward compatibility.
+
 ### kubectl 
 
 #### kubectl rollout 
@@ -274,14 +263,6 @@ Users can use `kubectl rollout` to monitor DaemonSet updates:
 - `kubectl rollout history daemonset/<DaemonSet-Name>`: to view the history of
   DaemonSet updates.
 - `kubectl rollout undo daemonset/<DaemonSet-Name>`: to rollback a DaemonSet
-
-#### API
-
-Implement a subresource for DaemonSet history (`daemonsets/foo/history`) that
-summarizes the information in the history.
-
-Implement a subresource for DaemonSet rollback (`daemonsets/foo/rollback`) that
-triggers a DaemonSet rollback.
 
 ## Updating DaemonSets mid-way
 
@@ -356,6 +337,14 @@ In the future, we may:
 - Allow user-defined DaemonSet unique label key
 - Support pausing DaemonSet rolling update
 - Support auto-rollback DaemonSets
+
+### API
+
+Implement a subresource for DaemonSet history (`daemonsets/foo/history`) that
+summarizes the information in the history.
+
+Implement a subresource for DaemonSet rollback (`daemonsets/foo/rollback`) that
+triggers a DaemonSet rollback.
 
 ### Tests
 
--- a/contributors/design-proposals/daemonset-update.md
+++ b/contributors/design-proposals/daemonset-update.md
@@ -118,18 +118,19 @@ type DaemonSetSpec struct {
 	// +optional
 	MinReadySeconds int32 `json:"minReadySeconds,omitempty"`
 
+	// DEPRECATED.
 	// A sequence number representing a specific generation of the template.
 	// Populated by the system. Can be set at creation time. Read-only otherwise. 
 	// +optional
 	TemplateGeneration int64 `json:"templateGeneration,omitempty"`
 
-	// The number of old PodTemplates to retain to allow rollback.
+	// The number of old history to retain to allow rollback.
 	// This is a pointer to distinguish between explicit zero and not specified.
 	// Defaults to 2.
 	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty"`
 
 	// The config this DaemonSet is rolling back to. Will be cleared after
-        // rollback is done.
+	// rollback is done.
 	// +optional
 	RollbackTo *RollbackConfig `json:"rollbackTo",omitempty`
 }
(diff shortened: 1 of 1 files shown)


smarterclayton:
@erictune @bgrant0607 if we go down this path we'll need to figure out how we could do "safe" pod templates (the original use case).  I proposed in the other issue that the isolation could be preserved if a user would have to activate a pod template before it could mint pod templates (thus the controllers could create them, but only a user could authorize them to instantiate).  How concerned are we that using pod templates for this limits the options for future changes along these lines?

contributors/design-proposals/daemonset-update.md
```diff
@@ -206,6 +244,18 @@ Users can use `kubectl rollout` to monitor DaemonSet updates:
...
 - `kubectl rollout status daemonset/<DaemonSet-Name>`: to see the DaemonSet
   upgrade status 
+- `kubectl rollout history daemonset/<DaemonSet-Name>`: to view the history of
+  DaemonSet updates. We use `PodTemplate` created by DaemonSets to store update
+  history.
+- `kubectl rollout undo daemonset/<DaemonSet-Name>`: to rollback a DaemonSet
+
+#### API
+
+Implement a subresource for DaemonSet history (`daemonsets/foo/history`) that
```
smarterclayton: Is this generic (pod template list, intended for generic clients) or type specific (array of daemonsets, with only some fields preserved)?

kow3ns:
@smarterclayton @erictune @bgrant0607 
I think we should consider a per Kind object that aggregates history as well as PodTemplate creation.

My primary concern about PodTemplates actually stems from the ability to mitigate inter controller (e.g. DaemonSet and StatefulSet with the same name) history overlap. A History object per controller Kind does provides adequate mitigation.

My primary concern about a per controller Kind History object is that third party operators or controllers have no way to integrate with it at the moment. However TPRs and sub resources are currently insufficient to communicate `.Status` and `.Spec`, and it seems we have already decided to fix this @foxish. 

Changing the method or persistence really doesn't impact the overall design much (provided we do so prior to implementation). If we use a per Kind History object we could make the following claim about controllers.

contributors/design-proposals/daemonset-update.md
```diff
@@ -122,6 +122,28 @@ type DaemonSetSpec struct {
...
+// DaemonSetRollback stores the information required to rollback a DaemonSet.
+type DaemonSetRollback struct {
+	metav1.TypeMeta `json:",inline"`
+	// Required: This must match the Name of a DaemonSet.
+	Name string `json:"name"`
+	// The annotations to be updated to a DaemonSet
+	// +optional
+	UpdatedAnnotations map[string]string `json:"updatedAnnotations,omitempty"`
+	// The config of this DaemonSet rollback.
+	RollbackTo RollbackConfig `json:"rollbackTo"`
```
kargakis: This is the same struct used by Deployments, right?

contributors/design-proposals/daemonset-update.md
kargakis: Can we move this into a separate proposal since a history subresource is not needed for upgrades and it's something we'll need to do for all the workload APIs?

contributors/design-proposals/daemonset-update.md
lukaszo: Right

contributors/design-proposals/daemonset-update.md
janetkuo: @smarterclayton something generic if we use PodTemplates to store history
@kargakis you meant moving the `daemonsets/foo/history` part to a separate proposal, right? Sure.

contributors/design-proposals/daemonset-update.md
janetkuo: Yes just reuse `RollbackConfig`

contributors/design-proposals/daemonset-update.md
```diff
@@ -165,22 +187,38 @@ DaemonSets on the apiserver.
...
    - If `RollingUpdate`:
-     - Find all daemon pods that belong to this DaemonSet using label selectors.
-       - Pods with "pod-template-generation" value equal to the DaemonSet's
-         `.spec.templateGeneration` are new pods, otherwise they're old pods.
-       - Note that pods without "pod-template-generation" labels (e.g. DaemonSet
-         pods created before RollingUpdate strategy is implemented) will be
-         seen as old pods.
+     - Find existing PodTemplates owned by DaemonSet, and sort them by the value
+       of `pod-template-generation` label
+     - Clean up PodTemplates based on `.spec.revisionHistoryLimit`
```
lukaszo: I think cleanup should be the last step

contributors/design-proposals/daemonset-update.md
```diff
@@ -165,22 +187,38 @@ DaemonSets on the apiserver.
...
+       - Always keep the PodTemplate if its `pod-template-generation` label has
+         the same value of any existing pods owned by the DaemonSet
+       - Remove PodTemplates from the ones with the smallest `pod-template-generation`
+         to the highest, until `.spec.revisionHistoryLimit` is hit or no
+         PodTemplates are allowed to be removed 
+     - Create a PodTemplate based on DaemonSet
```
lukaszo: Maybe we should first check if it's not a "rollback by update". User may update image x to y, then y to x. 
I think we should first check all PodTemplates and if we find one use it instead of creating new one.
label pod-template-generation should be updated then also for all pods which are still using this template we should also update their label.
Basically it's normal rollback.

janetkuo: This proposal doesn't differentiate rollback by update vs. by calling /rollback. 

Relabeling a pod is already included (see L221).

The reason for duplicating PodTemplates is to keep track of templateGeneration vs. PodSpec mapping. For example, let's say current templateGeneration is 5, v1 = v3 = v5 and there are v1 and v3 existing pods. If we reuse PodTemplates, we only know that v3 = v5 (or v1 = v5, depends on current label of the PodTemplate), and will have to kill v1 (or v3) pods. Alternatively, we can use revision annotation (instead of relabeling Pods and PodTemplates) to solve this, but templateGeneration + revision may be confusing since they have similar meaning (from the name).

contributors/design-proposals/daemonset-update.md
janetkuo: I don't think there's much difference, but sure we can move it to be the last step. 

One a side note, we don't need to keep history for `OnDelete` DaemonSets, right?

lukaszo: Right

contributors/design-proposals/daemonset-update.md
lukaszo: Ok, I'm convinced :) I'm not sure if this approach is optimal but for sure it's much safer.

contributors/design-proposals/daemonset-update.md
kargakis: > One a side note, we don't need to keep history for OnDelete DaemonSets, right?

Seems that people can use OnDelete in the same way Paused can be used for Deployments, what is our stance on that?

contributors/design-proposals/daemonset-update.md
kargakis: Why can't we deduplicate and keep a label in the latest PodTemplate with all the templateGenerations it has served? We already do that in Deployments for the revision annotation.

janetkuo: Yes we can deduplicate, and then we need a way to remember previous PodTemplate templateGenerations.

One way is to use revision annotations like I mentioned above. 

Another way is to use a label or annotation, say something like `previous-template-generation="1,3,5"`, and then we append more generations in the new PodTemplate after a rollout (`previous-template-generation="1,3,5,7"`). It may eventually become too long so we'd want to clean it up when no pods with that generation exist anymore. We'll also need to parse the string and search for the generation of all pods to see if they're from this PodTemplate. 

I thought about these alternatives, and prefer duplicating PodTemplate because it's simpler.

contributors/design-proposals/daemonset-update.md
janetkuo: `OnDelete` is mainly for backward-compatibility. Without history, it's still backward compatible. 

If we decide to maintain history for `OnDelete`, this history will be available when switching to `RollingUpdate`. The logic for `OnDelete` would be similar to `RollingUpdate`, except for the pod killing / creating part. 

I'm not opposed to keeping history for `OnDelete`, but just not sure who are really using `OnDelete` to update their DaemonSets and want to keep the history.

contributors/design-proposals/daemonset-update.md
kargakis: It's unlikely though that a PodTemplate will have a history greater than two.

contributors/design-proposals/daemonset-update.md
kargakis: Now if I want to apply multiple updates to a DaemonSet template w/o automatically rolling it out (use a flow like the following), I'll have to use OnDelete and then switch back to Rolling thus use the switch as a trigger to rollout all of my updates in a single unit.
```sh
# using OnDelete
# updated image
kubectl set image ds foo foo-sidecar=sidecar:0.2
# updated config
kubectl set volume --configmap-name=foo-cm-fa4dbh --overwrite
# maybe configure a missing probe
kubectl set probe --readiness ...
# switch back to Rolling
```
I guess what I am arguing for is keeping this consistent with Deployments and not have two different ways to do the same thing.

kargakis: The cat is out of the bag though, right? :)

kargakis: Note that the `set volume` and `set probe` subcommands already exist downstream in `oc` and we have been meaning to upstream them for some time now ^_^

contributors/design-proposals/daemonset-update.md
janetkuo: If several rollbacks are done consecutively (before the previous ones finish), it's possible that we end up having more than two generations of the same template.

contributors/design-proposals/daemonset-update.md
janetkuo: In your example, there will be 3 more PodTemplates created for `OnDelete`. Is that expected? In this case, it seems that a single history with all three changes make more sense.

contributors/design-proposals/daemonset-update.md
kargakis: > If several rollbacks are done consecutively (before the previous ones finish), it's possible that we end up having more than two generations of the same template.

Sounds like a user error or is it something on our part of how we are doing rollbacks currently? How can somebody end up doing that if the former?

contributors/design-proposals/daemonset-update.md
janetkuo: @kargakis another question, do people generally use `pause` before triggering a rollout (your example), or in the middle of it? Will this usage be different in Deployments vs. DaemonSet?

contributors/design-proposals/daemonset-update.md
janetkuo: Before a rollout completes (v1 & v2) (or maybe v2 is a bad rollout), rollback, update, and then rollback again. You'll get v1 = v3 = v5  in the system. I know this seems rare that someone would do that. Can we just ignore this case?

contributors/design-proposals/daemonset-update.md
kargakis: > In your example, there will be 3 more PodTemplates created for OnDelete. Is that expected? In this case, it seems that a single history with all three changes make more sense (since you want to trigger all 3 changes at once).

No new PodTemplate should be generated in the OnDelete case. At least, when a Deployment is paused, we won't create a new ReplicaSet for it.

> @kargakis another question, do people generally use pause before triggering a rollout (your example), or in the middle of it?

Based on feedback from others and on my own usage, most common is the flow I explained above: use `pause` before triggering. Ideally, for manual deployments (eg. prod rollouts), I think we want to run one-off rollouts and always run with `paused=true`. Pausing in the middle or at any point in the rollout is auto-pause (https://github.com/kubernetes/kubernetes/issues/11505) and that ability solves many other use cases (canaries, A/B, hooks).

> Will this usage be different in Deployments vs. DaemonSet?

IMO, the flow described above applies equally to DaemonSets and StatefulSets. Auto-pause will probably differ slightly.

jayunit100:
From afar I wonder if this belongs in the ecosystem, or at least, some portion of it does belong outside of kubernetes core ?

contributors/design-proposals/daemonset-update.md
janetkuo: >> In your example, there will be 3 more PodTemplates created for OnDelete. Is that expected? In this case, it seems that a single history with all three changes make more sense (since you want to trigger all 3 changes at once).
> 
> No new PodTemplate should be generated in the OnDelete case. At least, when a Deployment is paused, we won't create a new ReplicaSet for it.

If we don't create history for `OnDelete`, a history with all three changes will be created when switching to `RollingUpdate`. If we create history for `OnDelete`, there should be one history created each time templateGeneration is updated. So you agree on not creating history for `OnDelete`? Or you have other thoughts?

contributors/design-proposals/daemonset-update.md
kargakis: I think it's highly unlike that this update pattern will run more than once for the same template but we can solve that case by not relabeling in case a previous pod-template-generation label already exists in the PodTemplate. Assuming your example, all Pods would be labeled as pod-template-generation=1, even though ds.spec.generation=5. The DS template matches with the 1st PodTemplate so we simply continue to mint Pods from that template w/o any changes to it and its existing Pods. Reminder: the original reason for using templateGeneration was to avoid naming collisions introduced by hashing.

contributors/design-proposals/daemonset-update.md
kargakis: Yes, I think `OnDelete` should not create templates but we should probably allow cleanup of older templates, if we can delete any.

contributors/design-proposals/daemonset-update.md
janetkuo: Document something brought up in the sig-apps meeting today: 

If we reuse PodTemplates, we need to relabel them with a different templateGeneration when rolling back, and the name of the PodTemplate will be different with its label (regarding templateGeneration number) -- possible user confusion.

If we duplicate PodTemplates, it's hard for Deployment-like controllers to follow this pattern.

lukaszo: Will this approach work:
- create new PodTemplate
- start deleting old pods. If for old pod its PodTemplate == new PodTemplate. Don't delete it, just relabel
- If PodTemplate != new PodTemplate, delete the pod.

Ps. Is anything more which needs to be discussed before  implementation?

kargakis: After agreeing on continuing with hashing and the avoidance mechanism for Deployments, are we going to switch DaemonSets to use hashing too? If we want to switch then it'll be much easier to do it now rather than in the future. Hashing also would obsolete the need to relabel pods when we adopt them as part of a rollback.

kargakis:
> From afar I wonder if this belongs in the ecosystem, or at least, some portion of it does belong outside of kubernetes core ?

@jayunit100 based on experience we have so far with Deployments (and DeoploymentConfigs in OpenShift), providing sensible update strategies works for most use cases. Extensible strategies is https://github.com/kubernetes/kubernetes/issues/14510 + https://github.com/kubernetes/kubernetes/issues/39337

contributors/design-proposals/daemonset-update.md
lukaszo: Yeah. I agree. We should deprecate TemplateGeneration and start using hashing.

janetkuo: Yep switching to hash and deprecating templateGeneration is the plan. Will update this proposal when the other "history resource" proposal is out

lukaszo:
I have one more question. What should we do if user deletes PodTemplate, current or from history?
Or more general: should we watch for PodTemplate changes?

kargakis:
> Or more general: should we watch for PodTemplate changes?

Not sure but on a first thought I don't think it's necessary. The only thing that we could do is recreate the current PodTemplate if a delete event comes down the watch but apart from that, not much else. We should definitely list them to clean them up regurarly.

janetkuo:
@lukaszo thanks for bringing this up. We should always create a current/new history that matches current DS template if it doesn't exist yet (and if it's a rolling update DS). This will be done in the sync loop (list all existing history and compare) but not watch. I don't think we need to watch for history deletion/updates yet, and if the users mess up with history it's seen as user error.

kargakis:
> @lukaszo thanks for bringing this up. We should always create a current/new history that matches current DS template if it doesn't exist yet (and if it's a rolling update DS). This will be done in the sync loop (list all existing history and compare) but not watch. I don't think we need to watch for history deletion/updates yet, and if the users mess up with history it's seen as user error.

Why not make history read-only?

janetkuo:
Yes, the plan is to make the history object read-only. The previous comment is to answer the question what to do if users can change history. We can make the history read-only and disallow users from deleting it (they can only delete history implicitly via changing clean up policy). Even if we choose not to disallow them, we'll only recreate current history (not via watch, but from normal sync loop)

lukaszo:
> Yes, the plan is to make the history object read-only.

Where can I read more about this plan? And what is a history object? 
Is this proposal still valid?

krmayankk:
@kargakis @janetkuo @kow3ns Is the proposal for storing history using AnyState serving a different purpose than this one , or its an alternative to this one ?

kargakis:
> Where can I read more about this plan? And what is a history object?

The proposal that handles history is https://github.com/kubernetes/community/pull/594

janetkuo:
> Where can I read more about this plan? And what is a history object?
Is this proposal still valid?

Controller history (#594) will only replace PodTemplate in this proposal

janetkuo:
Updated the proposal to deprecate templateGeneration (use hash instead), replace PodTemplate with controller history, and add hash collision avoidance mechanism. 

Note that this proposal is not finalized until controller history is finalized.

kargakis:
> History resource struct

#594

> History name and unique labels

#594 + plus the labels/annotations we already use for Deployments but customized for DaemonSets?

> How we hash a history

fnv + uniquifier?

> Rollback: server side or client side?

+1 for server-side so other clients can take advantage of it.

contributors/design-proposals/daemonset-update.md
```diff
@@ -146,13 +169,25 @@ type DaemonSetStatus struct {
...
 	// DaemonSetTemplateGenerationKey is the key of the labels that is added
-	// to daemon set pods to distinguish between old and new pod templates
+	// to daemon set pods to distinguish between old and new pod
 	// during DaemonSet template update.
 	DaemonSetTemplateGenerationKey string = "pod-template-generation"
+
+	// DefaultDaemonSetUniqueLabelKey is the default label key that is added
+	// to existing DaemonSet pods to distinguish between old and new
+	// DaemonSet pods during DaemonSet template updates.
+	DefaultDaemonSetUniqueLabelKey string = "pod-template-hash"
```
kargakis: Now that's why I was asking for making this namespaced. This collides with the label added in Deployments pods with the potential for cross-controller overlaps.

contributors/design-proposals/daemonset-update.md
kargakis: This section is obsolete?

contributors/design-proposals/daemonset-update.md
janetkuo: This is not final. #594 proposed hashing `AnyState.Data`. If we decided to hash data instead of template, this won't be a proper name. If all controllers (except for Deployments) will use `AnyStates`, we can just use `AnyState.Name` which is guaranteed to be unique.

janetkuo:
>> History name and unique labels

> #594 + plus the labels/annotations we already use for Deployments but customized for DaemonSets?

Should all controllers follow the same "copying labels/annotations" pattern? If so, #594 needs to include this.

>> How we hash a history

> fnv + uniquifier?

#594 proposed fnv and SHA-2. #594 proposed hashing data, so we should include uniquifier in the data.

>> Rollback: server side or client side?

> +1 for server-side so other clients can take advantage of it.

I'm +1 for server side

contributors/design-proposals/daemonset-update.md
kargakis: > we can just use AnyState.Name as the label value which is guaranteed to be unique.

How is this guaranteed?

kargakis:
> Should all controllers follow the same "copying labels/annotations" pattern? If so, #594 needs to include this.

I don't feel strong about this but people seem to expect that metadata such as labels and annotations are inherited.

> #594 proposed fnv and SHA-2. #594 proposed hashing data, so we should include uniquifier in the data.

I don't think we need a cryptographic algo for hashing and fnv is faster while still being more stable than what we need.

> I'm +1 for server side too, but there are some debates about it. The main concern is that controller is modifying its own spec (spec.rollbackTo).

I can see the argument but I am on the fence for this one. I like the way rollbackTo works but I agree that mutating spec is not good. I think we can have a server-side implementation w/o the need to mutate spec.

contributors/design-proposals/daemonset-update.md
janetkuo: You can't have two AnyStates with the same name

kargakis: I mean how do we guarantee that a history object for a DS won't ask for the same name as a history object owned by a StatefulSet.

janetkuo: > I mean how do we guarantee that a history object for a DS won't ask for the same name as a history object owned by a StatefulSet.

Each history has OwnerRef and labels adopted from controller's selector. Discussed with @kow3ns yesterday, the label key will be consistent (format-wise) across controllers, something like "statefulset-history-revision", "daemonset-history-revision"

janetkuo:
DaemonSet controller history is ready for review: https://github.com/kubernetes/kubernetes/pull/45924

In this PR, I have the server side rollback ready regarding API changes and /rollback endpoint, but we may need to do the rollback client-side, considering the "controller is modifying its own spec" drawback.

kubectl rollback PR is here (but not ready for review yet): https://github.com/kubernetes/kubernetes/pull/46144

contributors/design-proposals/daemonset-update.md
janetkuo: On second thought we still need history for OnDelete, otherwise once switching a OnDelete DS to RollingUpdate, all its pods will be killed and recreated. We also need history to compare hash.

janetkuo:
Updated proposal. PTAL

janetkuo:
If no one objects by the end of this week, I'm going to merge this. This is already implemented.

PR #532, opened 2017-04-13
High-level discussion of k8s' security model for user-facing surfaces.

Q-Lee:
This is a little different from Clayton's threat model (#504), but has a non-zero intersection.

I would like to nail down the direction for Kubernetes' security model.

### DIFF
--- a/contributors/design-proposals/security-model.md
+++ b/contributors/design-proposals/security-model.md
@@ -1,0 +1,190 @@
+# Requirements
+
+This document considers "customer facing" entities. For our purposes these are
+the Kubernetes API, and running workloads.
+
+1. Kubernetes needs to be safe in a multi-tenant environment
+2. Securing Kubernetes should be easy
+
+This document uses the "[CIA triad](https://en.wikipedia.org/wiki/Information_security#Key_concepts)" for analyzing these goals.
+
+# Kubernetes API access controls
+
+Kubernetes needs access controls capable of protecting the CIA of cluster
+resources in a multi-tenant environment.
+
+## Concerns with method based access controls
+
+Many Kubernetes deployments provide method based access control (e.g.,
+list pods). This is akin to securing \*nix with permissions on the "open"
+syscall. To allow more than privileged user, ACLs should be placed on nouns, not
+verbs.
+
+## Concerns with implicit relationships
+
+Kubernetes' implicit object relationships create subtle dangers.
+
+1. The "create pod" privilege allows a user to mount a secret, even if the user
+   cannot "get secrets". This is a failure of confidentiality.
+2. The "create pod" privilege can be used to place a pod behind a service, even
+   if the user cannot "update service". This is a failure of integrity.
+3. The "create pod" privilege can be used to place a pod behind a deployment,
+   which can be used to starve legitimate replicas. This is a failure of
+   integrity, and availability.
+
+This isn't a complete list of such dangers, but it's sufficient to show that
+access controls on an object are insufficient to guarantee CIA. Access controls
+are also needed on the relationships between objects.
+
+## Label ACLs (implicit ownership)
+
+Label ACLs are sometimes suggested as a method to provide access control on
+object-object relationships. Special permissions would be required to create or
+query an ACL'ed label. This approach is both complex and insufficient.
+
+1. The accidental omission of a label means that the system fails open.
+2. It becomes increasingly difficult to reason about permissions, especially if
+   set-based label selectors are used.
+3. Can ACLs be created on existing labels?
+   1. Yes: clients can be locked out of their workloads.
+   2. No: workloads must be restarted to change permissions.
+
+This approach also perturbs the existing behavior of label selectors. Does the
+"list pod" permission allow me to see pods with contentious ACLs? What about
+"list secrets"?
+
+Finally, this approach is insufficent, because not all relationships are defined
+by label selectors. Notably, secrets cannot be protected by label ACLs.
+
+## Explicit "ownership" (recommended)
+
+Explicit owners can be used to limit access to both objects and relationships. A
+service would reject both unauthorized changes to its body, and unauthorized
+pods as backends.
+
+The simplicity and correctness of this method recommend it.
+
+### Namespace as "owner"
+
+By scoping method permissions to namespaces, a crude ownership model can be
+constructed (all objects are owned by a namespace, and object-object
+relationships must have the same owner). The advantage of this approach is that
+it's possible today.
+
+There are also disadvantages to this method.
+
+1. Not all resources are namespaced (e.g., nodes and volumes)
+2. The myriad escalations and CIA failures make it harder to [separate
+   privileges](https://en.wikipedia.org/wiki/Privilege_separation)
+   within a namespace.
+3. The mapping between namespace and true ownership must be enforced by the
+   cluster admin.
+
+### Intentional ownership
+
+Rather than overloading namespace, Kubernetes can introduce an explicit concept
+of ownership. For example, a service owned by group\* "foo" could only be updated by members
+of "foo". The service's backends would also be limited to pods owned by a group.
+
+1. All resources would be owned.
+2. Scoping down permissions is simple with control over both object and relationship access\*\*.
+3. These owners could derive from an identity management system, so the onus of
+enforcement isn't on the cluster admin.
+
+A true ownership model solves the disadvantages of namespace-based ownership.
+But it's not clear that the improvements are worth the effort of implementation.
+
+\* Group is only used as an example. The system could be built to support
+users, roles, service accounts, or any other identifiable entity.
+
+\*\* It's also reasonable to control relationships with a reflective or membership
+constraint. This simplifies permissions management at the expense of
+flexibility, and gives no real improvement over problem (2) with namespaces.
+
+# Workloads
+
+The other face of Kubernetes exposed to customers is the workload environment.
+This section focuses on protecting workloads from the environment, and the
+environment from workloads.
+
+## Scheduled isolation
+
+Many users have use cases for scheduled isolation. This is where Kubernetes
+separates sensitive workloads from each other.
+
+1. A sensitive application might not tolerate other workloads on the same node.
+2. A user may bring a dedicated node.
+3. An application may need to run privileged containers.
+
+Both affinity and taints can achieve this today, but neither is protected. Both
+affinities and tolerations are specific on the "PodSpec", so anyone with "create
+pod" permissions can circumvent the mechanism.
+
+### Dedicated nodes
+
+Dedicated nodes require scheduling protection, because A's safety shouldn't be
+dependent on B's goodwill. Taint ACLs would suffer from the same problem as
+label ACLs. However, the "ownership" model can be applied successfully.
+
+For instance, a special "allowed-namespaces" taint would forbid pods outside of
+authorized namespaces. The pods would still be responsible for selecting the
+nodes.
+
+Alternatively, membership in a group, or a role, could be used to permit or
+forbid execution on a node. This might make more sense than forcing namespaces
+onto nodes, but probably requires the intentional ownership model.
+
+### Privileged pods
+
+A PodSecurityPolicy can prevent users from creating privileged containers.
+However, some use cases may be that privileged pods can only be created on
+dedicated nodes.
+
+For instance, a game server may want to run with host networking. These will
+need to be created with a privileged PSP. But many workloads may not tolerate
+sharing a host with privileged user pods.
+
+1. Privileged containers could taint a node.
+2. A cluster-level policy could restrict privileged containers to dedicated
+   nodes.
+3. A "NodeSecurityPolicy" could be introduced, which could forbid privileged
+   user containers.
+
+## Identity and Secrets
+
+Kubernetes is not intended to be a key management system. Customers should run a
+key management system (KMS) that emphasizes security (e.g., pinned memory, encrypted
+at rest, etc.). The chosen KMS can deliver appropriate secrets to containers.
+
+Bootstrapping identity may still fall to Kubernetes. Either containers are
+created with identifying resources (e.g., a cert), or the bootstrap protocol can
+physically verify the origin of a request (e.g., by inspecting OS resources).
+
+### Separation of privileges
+
+Imagine a pod with the following members and permissions:
+
+Permissions: logging (W), metrics (W), database (R/W)
+Containers: logger, metric agent, web service
+
+The pod's privileges can be further separated, giving each container only the
+permissions needed to function. This suggests that a KMS needs to be capable of
+distinguishing individual containers.
+
+### Bootstrapping
+
+An IP-based metadata server (e.g., kube2iam) is insufficient to separate
+privileges, because all containers within a pod share a network namespace.
+
+## Host isolation
+
+### Security Modules
+
+Many users are interested in using LSM plugins to further increase security in a
+deployment, or even LSM replacements (e.g., RSBAC).
+
+## Isolated networks
+
+If namespace is repurposed as a fine-grained security boundary, then a
+namespaced NetworkPolicy seems woefully inadequate. A large organization would
+need at least one per project, and they would need to be kept in sync.


contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+Many Kubernetes deployments provide method based access control (e.g.,
+list pods). This is akin to securing \*nix with permissions on the "open"
+syscall. To allow more than privileged user, ACLs should be placed on nouns, not
+verbs.
+
+## Concerns with implicit relationships
+
+Kubernetes' implicit object relationships create subtle dangers.
+
+1. The "create pod" privilege allows a user to mount a secret, even if the user
```
smarterclayton: Service accounts can limit what secrets can be mounted.  However, since the point of most secrets is to be used in pods, I think the danger is probably more the other way "secrets not intended for use in pods can be accessed by pods".  Both directions are valid, of course.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+constructed (all objects are owned by a namespace, and object-object
+relationships must have the same owner). The advantage of this approach is that
+it's possible today.
+
+There are also disadvantages to this method.
+
+1. Not all resources are namespaced (e.g., nodes and volumes)
+2. The myriad escalations and CIA failures make it harder to [separate
+   privileges](https://en.wikipedia.org/wiki/Privilege_separation)
+   within a namespace.
```
smarterclayton: Note that separating privileges within a namespace has never been a core goal of Kubernetes - the original security model explicitly describes namespaces as a security boundary, rather than within a namespace.  Sub-namespace separation should be possible, although I don't know that I think it is required if namespace separation can better model actual user intent.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+3. The mapping between namespace and true ownership must be enforced by the
+   cluster admin.
+
+### Intentional ownership
+
+Rather than overloading namespace, Kubernetes can introduce an explicit concept
+of ownership. For example, a service owned by group\* "foo" could only be updated by members
+of "foo". The service's backends would also be limited to pods owned by a group.
+
+1. All resources would be owned.
```
smarterclayton: As an alternative - why not model namespaces as being "owned"?

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+2. Scoping down permissions is simple with control over both object and relationship access\*\*.
+3. These owners could derive from an identity management system, so the onus of
+enforcement isn't on the cluster admin.
+
+A true ownership model solves the disadvantages of namespace-based ownership.
```
smarterclayton: A true ownership model however requires that a user leaving the organization transfer ownership of all resources.  There are two predominant models of tenancy:

1. resources owned by the user
2. resources owned by the organization

The former is best described as "when you leave, these are deleted".  The latter is defined as "when you leave, everything keeps working but you no longer have access".

If namespaces are owned by individuals (1) or groups (2), is there a significant difference between the granular ownership model?

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+Many users have use cases for scheduled isolation. This is where Kubernetes
+separates sensitive workloads from each other.
+
+1. A sensitive application might not tolerate other workloads on the same node.
+2. A user may bring a dedicated node.
+3. An application may need to run privileged containers.
+
+Both affinity and taints can achieve this today, but neither is protected. Both
+affinities and tolerations are specific on the "PodSpec", so anyone with "create
+pod" permissions can circumvent the mechanism.
```
smarterclayton: Only because no permission model applies to either the actor or the namespace.  If namespaces or the owning service account required access to affinities and tolerations before pods could use them, this would not be an issue.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+access controls on an object are insufficient to guarantee CIA. Access controls
+are also needed on the relationships between objects.
+
+## Label ACLs (implicit ownership)
+
+Label ACLs are sometimes suggested as a method to provide access control on
+object-object relationships. Special permissions would be required to create or
+query an ACL'ed label. This approach is both complex and insufficient.
+
+1. The accidental omission of a label means that the system fails open.
```
smarterclayton: Only if users are granted permission on all resources.  Granting access to all labels is oversharing - if access is only granted by labels for resources across contexts, then the system fails closed.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+2. It becomes increasingly difficult to reason about permissions, especially if
+   set-based label selectors are used.
+3. Can ACLs be created on existing labels?
+   1. Yes: clients can be locked out of their workloads.
+   2. No: workloads must be restarted to change permissions.
+
+This approach also perturbs the existing behavior of label selectors. Does the
+"list pod" permission allow me to see pods with contentious ACLs? What about
+"list secrets"?
```
smarterclayton: List secrets is already a special case, in that secrets are escalating resources.  I shouldn't be able to see contents unless I have permission to do so.  The act of creating a secret without protection is the problem, not secrets implicitly.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+
+Finally, this approach is insufficent, because not all relationships are defined
+by label selectors. Notably, secrets cannot be protected by label ACLs.
```
smarterclayton: Secrets can be protected by label ACLs if a few other conditions are met:

1. creation of a secret requires user to specify intent, or intent is defaulted to a restrictive mode
2. access to secret contents is restricted by intent or context
3. secrets are write only

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+2. The myriad escalations and CIA failures make it harder to [separate
+   privileges](https://en.wikipedia.org/wiki/Privilege_separation)
+   within a namespace.
+3. The mapping between namespace and true ownership must be enforced by the
+   cluster admin.
+
+### Intentional ownership
+
+Rather than overloading namespace, Kubernetes can introduce an explicit concept
+of ownership. For example, a service owned by group\* "foo" could only be updated by members
```
smarterclayton: Ownership doesn't defend against all threats - confused deputy, for example, is just as difficult with ownership.  A controller that has access to update a set of services likely does so in response to other inputs, which then means that controller can propagate updates outside of group 'foo' as easily as in the current model.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+
+Kubernetes needs access controls capable of protecting the CIA of cluster
+resources in a multi-tenant environment.
+
+## Concerns with method based access controls
+
+Many Kubernetes deployments provide method based access control (e.g.,
+list pods). This is akin to securing \*nix with permissions on the "open"
+syscall. To allow more than privileged user, ACLs should be placed on nouns, not
+verbs.
```
liggitt: both noun and verb dimensions are supported by authorization rules

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+
+For instance, a game server may want to run with host networking. These will
+need to be created with a privileged PSP. But many workloads may not tolerate
+sharing a host with privileged user pods.
+
+1. Privileged containers could taint a node.
+2. A cluster-level policy could restrict privileged containers to dedicated
+   nodes.
+3. A "NodeSecurityPolicy" could be introduced, which could forbid privileged
+   user containers.
```
smarterclayton: This may not be sufficient for all scenarios, especially with self hosted components like controllers that run with rights across many resources.  Policies controlling privileged access to the node must at some level be imposed by the hosts, rather than the cluster.  It's easier to build a ring of moats with decreasing privilege than to attempt to cross lines.  Outside of the core rings, placement policy can be enforced safely by the apiserver.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Identity and Secrets
+
+Kubernetes is not intended to be a key management system. Customers should run a
+key management system (KMS) that emphasizes security (e.g., pinned memory, encrypted
+at rest, etc.). The chosen KMS can deliver appropriate secrets to containers.
```
smarterclayton: I would rephrase this as:

Kubernetes is not intended to be a key storage solution.  It is intended to assist in the transport, rotation, and access to keys.

I would agree that Kubernetes should limit the investment in secret management - but a KMS still must actually trust the Kubernetes system for two reasons: identity and placement (as you note below), and thus an external KMS can only partially limit the spread of an attack.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+### Separation of privileges
+
+Imagine a pod with the following members and permissions:
+
+Permissions: logging (W), metrics (W), database (R/W)
+Containers: logger, metric agent, web service
+
+The pod's privileges can be further separated, giving each container only the
+permissions needed to function. This suggests that a KMS needs to be capable of
+distinguishing individual containers.
```
smarterclayton: Can be - the question is whether that separation is valuable enough to justify the additional complexity.  The classic unix "drop privileges" approach is relevant and can be modeled solely with keys delivered to an init container.  A KMS can enforce policy that only delivers secrets to pods with a particular shape (an init container that has access to the pod identity volume and is running an image signed with a particular cert).

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+### Security Modules
+
+Many users are interested in using LSM plugins to further increase security in a
+deployment, or even LSM replacements (e.g., RSBAC).
+
+## Isolated networks
+
+If namespace is repurposed as a fine-grained security boundary, then a
+namespaced NetworkPolicy seems woefully inadequate. A large organization would
+need at least one per project, and they would need to be kept in sync.
```
smarterclayton: Why is this a problem?

contributors/design-proposals/security-model.md
Q-Lee: That's a good point. I wasn't trying to create a comprehensive list, but just show some examples. Do you any others?

contributors/design-proposals/security-model.md
smarterclayton: Hosts in existing clusters already require denormalized network policy - denormalizing network policy into namespaces is no worse.

contributors/design-proposals/security-model.md
davidopp: Dumb question - what does it mean to say a service account owns an object? Just that there is an RBAC rule that allows the service account to perform some verbs on the object?

contributors/design-proposals/security-model.md
deads2k: > Dumb question: How does that work?

ServiceAccounts have a list of `secrets` which are allowed to be mounted by a pod using that ServiceAccount.  There's an admission plugin which inspects the secrets being mounted by the pod and any secret not in the ServiceAccount.Secrets list will cause a rejection.

deads2k: > That's a good point. I wasn't trying to create a comprehensive list, but just show some examples. Do you any others?

Mutation of ReplicationControllers allows the ability to delete pods the user couldn't create.  Without a concept of ownership (difficult in a multi-mutator system), solving the "problem" is intractable.

deads2k: > Without a concept of ownership (difficult in a multi-mutator system),

Ah, I see you've tried to define it below. A namespace level owner doesn't seem to fix this concern and the "intentional" owner from below doesn't seem to provide value above RBAC since, "group of people who can edit" is the same as "people bound to role which can edit".

contributors/design-proposals/security-model.md
deads2k: How is this providing value above knowing the set of people bound to roles which allow them to update the resource.  It seems to be a different way of describing the same information.  Only doing it twice.

contributors/design-proposals/security-model.md
Q-Lee: You're right: protecting k8s' objects is insufficient to protect their integrity. The relationships between objects must also be protected (e.g., services <> pods, deployment <> pods, pods <> secrets).

Scoping permissions to a namespace protects most inter-object relationships, because most are bound to a namespace. There are some difficulties with this approach

1. Objects not bound to a namespace

1. The loss of versatility could make separating privileges hard

1. The cluster administrator is responsible for keeping namespaces in sync with ownership

The advantages with namespace are that it can be done today with RBAC.

contributors/design-proposals/security-model.md
Q-Lee: That's true, but I dread a k8s where I need permissions for _every_ label. I don't see how access management is simplified by doing this.

contributors/design-proposals/security-model.md
Q-Lee: Arguably, all objects need similar protections. It's so much more important for secrets that we had to do something to protect them earlier.

contributors/design-proposals/security-model.md
Q-Lee: ".... The service's backends would also be limited to pods owned by a group."

Explicit ownership makes access control _between_ objects possible. Namespace satisfies this requirement with the caveat that objects can only relate to others with the same "owner".

If we want to avoid special cases (e.g., secrets, nodes, and volumes), or allow separation of privileges within a namespace (a nice, but admittedly dubious, property), then all objects will need to know an owner.

contributors/design-proposals/security-model.md
Q-Lee: That's essentially what the namespace argument boils down to. Most objects belong to a namespace, and you can own a namespace.

contributors/design-proposals/security-model.md
Q-Lee: I think we've always intended to support container scope privileges. If we want to bootstrap a KMS with k8s secrets, then it's pretty straight forward.

@mikedanese 
@cjcullen 

https://kubernetes.io/docs/concepts/configuration/secret/#use-case-secret-visible-to-one-container-in-a-pod

contributors/design-proposals/security-model.md
Q-Lee: On reflection, I prefer the denormalized model. It beats cutting a ticket to networking, and the scope helps it retain some its purpose (so cruft is more manageable).

In fact... I like it so much that now I want access defined on Services and Deployments :(

contributors/design-proposals/security-model.md
Q-Lee: If we make namespace the unit of security, then we'll need to special case every separation within a namespace, and every non-namespaced object.

Maybe that's not so bad. Most objects are namespaced, and few objects will ever be as precious as secrets. But it will need to be explicit in our security model, so that we can check for special cases when we're reviewing new features.

contributors/design-proposals/security-model.md
Q-Lee: @smarterclayton - Exactly, we'll need to special access controls for non-namespaced objects.

@davidopp - I think "ownership" is being kept intentionally vague, since it's a concept under discussion. For instance, an admission controller could check that you have permissions to use affinities and tolerations.

contributors/design-proposals/security-model.md
Q-Lee: Let me see if I understand your points correctly.

1. Intent: some form of "SecretSecurityPolicy" (e.g., {podConsumable: true, initContainer: true, runas: foo-service}? This would default to a fairly inaccessible secret.
1. An admission controller prevents creation of pods that mount a secret without matching the policy.
1. I'm not sure why secrets need to be immutable. Is it because I can "upgrade" the classification (e.g., locking down the runas), but the secret will still be accessible by a grandfathered object?

Rather than special casing access controls for every object-uses-object relationship, I would prefer a generic means of control. It could then be re-used for pod-uses-secret, pod-uses-node, and even service-uses-pod.

contributors/design-proposals/security-model.md
Q-Lee: This is exactly why there needs to be control on object-uses-object relationships. Whether this is implicit with namespace (with special cases for non-namespaced objects), or done by run-time permission checking (e.g., when a service wants to add a pod as a backend) is less important than that it happens at all.

contributors/design-proposals/security-model.md
Q-Lee: The security model should list what practices are required.

contributors/design-proposals/security-model.md
Q-Lee: I meant to exclude controllers with "user containers". If that's unclear, can you suggest another phrasing?

Are we considering nodes that don't trust the apiserver? It sounds fine for nodes to be the final authority on what they execute, and therefore reject pods.

1. I don't see why the cluster's control plane shouldn't know these restrictions for placement. If 90% of my fleet wants to reject privileged containers, then I would have quite a few mishaps before I found a stable home.

1. Nodes could still be the authority. For instance, the kubelet could pull a "NodeSecurityPolicy" out of a local yaml file, and then write it back to the apiserver. The apiserver will respect the node's policy, but if they get out of sync, then the node will still reject forbidden workloads.

contributors/design-proposals/security-model.md
Q-Lee: Yes, but I've been hearing quite a lot of discussion on how identity should be established.

@mikedanese @destijl @cjcullen - any input?

contributors/design-proposals/security-model.md
smarterclayton: The model being proposed is that access to secrets is gated by label.  Integrations do not have access to read all secrets, they only have access to read secrets by a particular label.  For instance, label `kubernetes.io/for-use-by-nodes=1`.  This would then not be special case - it would be normal case.  All "for use by" relationships can be modeled via ACL on labels for secrets.  

I do agree that targeted "-uses-" relationships are valuable, but only a small subset of those relationships matter, so I'm not sure that generic "uses" has significant value to users (I think it has to be proved that uses is a strictly better model than buckets + roles for the majority case, unless you're proposing to add "uses" as an additional dimension).  I also don't see that "-uses-" addresses confused deputy on controllers (unless I missed that in this doc), which is the more pressing concern.  And with controllers, if you grant a controller access to create anything that can use anything else, you also have to enable it to either transitively refer to all uses (thus requiring a fairly large set of permissions and reducing to the current model) or explicitly specify all uses (which for many users will result in them bulk selecting "all related resources" to grant access to).

contributors/design-proposals/security-model.md
smarterclayton: :)

contributors/design-proposals/security-model.md
derekwaynecarr: Is it safe to assume that if in the service catalog, we want to be able to reference secrets for use in service `Instance` or `Binding` parameters, we would potentially need our users to add a label `kubernetes.io/for-use-by-service-catalog=1` label, and then any secrets created by the service catalog as a result of a `Binding` object being created would need to have a way of defining labels to apply on the secret the catalog provisions?

/cc @kubernetes/sig-service-catalog-misc @pmorie

Q-Lee: Let's set up a call to discuss [just] access controlled labels. I think this _calls_ for something with a faster back and forth than a PR.

@deads2k @smarterclayton @davidopp @cjcullen @destijl

davidopp: Please include me in the call. (But please notify me though some means other than Github, which I don't atch enough of.)

contributors/design-proposals/security-model.md
feiskyer: Does this include host path volumes? hostPath should also be controlled for arbitrary mounting.

contributors/design-proposals/security-model.md
```diff
@@ -0,0 +1,190 @@
...
+\*\* It's also reasonable to control relationships with a reflective or membership
+constraint. This simplifies permissions management at the expense of
+flexibility, and gives no real improvement over problem (2) with namespaces.
+
+# Workloads
+
+The other face of Kubernetes exposed to customers is the workload environment.
+This section focuses on protecting workloads from the environment, and the
+environment from workloads.
+
```
feiskyer: I think a section of container runtime isolation should also be added, e.g. via hypervisor-based runtimes (refer [frakti](https://github.com/kubernetes/frakti)). 

Hypervisors could provide a more consolidate isolation boundary than namespace-based runtimes. And it could ensure the minimal attack surface in multi-tenant environments.

contributors/design-proposals/security-model.md
destijl: I think we need to design for container identity. There are already use cases, and trying to bolt it in later will cause problems. We can encourage people to split into more pods wherever it's possible to reduce the need for container identity, but in some cases it won't be. A SPIFFE doc had a good use case: a migration of a large monolithic app into a set of microservices. First container deployment might have only broken up the app into a couple of services, so pods have containers with very different roles that need different identities but have to run together on the same pod as the devs work on splitting it out into separate services.

contributors/design-proposals/security-model.md
destijl: I've been thinking of k8s secrets as supporting two approaches:

1. a built-in, functional secrets store for basic use; and
2. a simple way to interface with enterprise-grade external stores such as Google/Amazon KMS and Hashicorp Vault etc.

davidopp:
This doc raises good issues but I'm worried it could be confusing to have too many docs about multi-tenancy checked into our repo. @Q-Lee maybe you and I could work together with @smarterclayton on something comprehensive rather than attacking it piecemeal (this PR, [my doc](https://docs.google.com/document/d/15w1_fesSUZHv-vwjiYa9vN_uyc--PySRoLKTuDhimjc/edit) on multi-tenancy models, and Clayton's [doc](https://github.com/kubernetes/community/pull/504) have some overlap, while none by itself is comprehensive, and I'm worried people will just get confused if we merge them as separate docs).

PR #535, opened 2017-04-13
Strategic Merge Patch doc

mengqiy:
Pull Strategic Merge Patch out of API convention doc.
And refactor it.

Use the format what I discussed with @pwittrock offline.
More content need to be fill by @pwittrock in a followup PR.

PR #537, opened 2017-04-14
Proposal: preserving order in strategic merge patch

mengqiy:
Move from https://github.com/kubernetes/community/pull/467, since we want to move strategic merge patch doc out of the api-convention doc. [WIP PR](https://github.com/kubernetes/community/pull/535)

Addressed comments in https://github.com/kubernetes/community/pull/467.

cc: @pwittrock @lavalamp @liggitt @apelisse

### DIFF
--- a/contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
+++ b/contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
@@ -1,0 +1,566 @@
+# Preserve Order in Strategic Merge Patch
+
+Author: @mengqiy
+
+## Motivation
+
+Background of the Strategic Merge Patch is convered [here](../devel/strategic-merge-patch.md).
+
+The Kubernetes API may apply semantic meaning to the ordering of items within a list,
+however the strategic merge patch does not keeping the ordering of elements.
+Ordering has semantic meaning for Environment variables,
+as later environment variables may reference earlier environment variables,
+but not the other way around.
+
+One use case is the environment variables. We don't preserve the order which causes
+issue [40373](https://github.com/kubernetes/kubernetes/issues/40373).
+
+## Proposed Change
+
+We will use the following notions through the doc.
+Notion:
+list to be merged: same as live list, which is the list current in the server.
+parallel list: the list with `$setElementOrder` directive in the patch.
+patch list: the list in the patch that contains the value changes.
+
+Changes are all in strategic merge patch package.
+The proposed solution is similar to the solution used for deleting elements from lists of primitives.
+
+Add to the current patch, a directive ($setElementOrder) containing a list of element keys -
+either the patch merge key, or for primitives the value. When applying the patch,
+the server ensures that the relative ordering of elements matches the directive.
+
+The server will reject the patch if it doesn't satisfy the following 2 requirement.
+- the relative order of any two items in the `$setElementOrder` list
+matches that in the patch list if they present.
+- the items in the patch list must be a subset or the same as the `$setElementOrder` list if the directive presents.
+
+The relative order of two items are determined by the following order:
+
+1. relative order in the $setElementOrder if both items are present
+2. else relative order in the patch if both items are present
+3. else relative order in the server-side list if both items are present
+4. else append to the end
+
+If the relative order of the live config in the server is different from the order of the parallel list,
+the user's patch will always override the order in the server.
+
+Here is an simple example of the patch format:
+
+Suppose we have a type called list. The patch will look like below.
+The order from the parallel list ($setElementOrder/list) will be respected.
+
+```yaml
+$setElementOrder/list:
+- A
+- B
+- C
+list:
+- A
+- C
+```
+
+All the items in the server's live list but not in the parallel list will be come before the parallel list.
+The relative order between these appended items are kept.
+
+The patched list will look like:
+
+```
+mergingList:
+- serverOnlyItem1 \
+  ...              |===> items in the server's list but not in the parallel list
+- serverOnlyItemM /
+- parallelListItem1 \
+  ...                |===> items from the parallel list
+- parallelListItemN /
+```
+
+### When $setElementOrder is not present and patching a list
+
+The new directive $setElementOrder is optional.
+When the $setElementOrder is missing,
+relative order in the patch list will be respected.
+
+Examples where A and C have been changed, B has been deleted and D has been added.
+
+Patch:
+
+```yaml
+list:
+- A'
+- B'
+- D
+```
+
+Live:
+
+```yaml
+list:
+- B
+- C
+- A
+```
+
+Result:
+
+```yaml
+list:
+- C # server-only item comes first
+- A'
+- B'
+- D
+```
+
+### `$setElementOrder` may contain elements not present in the patch list
+
+The $setElementOrder value may contain elements that are not present in the patch
+but present in the list to be merge to reorder the elements as part of the merge.
+
+Example where A & B have not changed:
+
+Patch:
+
+```yaml
+$setElementOrder/list:
+- A
+- B
+```
+
+Live:
+
+```yaml
+list:
+- B
+- A
+```
+
+Result:
+
+```yaml
+list:
+- A
+- B
+```
+
+### When the list to be merged contains elements not found in `$setElementOrder`
+
+If the list to be merged contains elements not found in $setElementOrder,
+they will come before all elements defined in $setElementOrder, but keep their relative ordering.
+
+Example where A & B have been changed:
+
+Patch:
+
+```yaml
+$setElementOrder/list:
+- A
+- B
+list:
+- A
+- B
+```
+
+Live:
+
+```yaml
+list:
+- C
+- B
+- D
+- A
+- E
+```
+
+Result:
+
+```yaml
+list:
+- C
+- D
+- E
+- A
+- B
+```
+
+### When `$setElementOrder` contains elements not found in the list to be merged
+
+If `$setElementOrder` contains elements not found in the list to be merged,
+the elements that are not found will be ignored instead of failing the request.
+
+Patch:
+```yaml
+$setElementOrder/list:
+- C
+- A
+- B
+list:
+- A
+- B
+```
+
+Live:
+```yaml
+list:
+- A
+- B
+```
+
+Result:
+
+```yaml
+list:
+- A
+- B
+```
+
+## Version Skew and Backwards Compatibility
+
+The new version patch is always a superset of the old version patch.
+The new patch has one additional parallel list which will be dropped by the old server.
+
+As mentioned [above](#when-setelementorder-is-not-present-and-patching-a-list),
+the new directive is optional.
+Patch requests without the directive will change a little,
+but still be fully backward compatible.
+
+### kubectl
+If an old kubectl sends a old patch to a new server,
+the server will honor the order in the list as mentioned above.
+The behavior is a little different from before but is not a breaking change.
+
+If a new kubectl sends a new patch to an old server, the server doesn't recognise the parallel list and will drop it.
+So it will behave the same as before.
+
+## Example
+
+### List of Maps
+
+We take environment variables as an example.
+Environment variables is a list of maps with merge patch strategy.
+
+Suppose we define a list of environment variables and we call them
+the original environment variables:
+
+```yaml
+env:
+- name: ENV1
+  value: foo
+- name: ENV2
+  value: bar
+- name: ENV3
+  value: baz
+```
+
+Then the server appends two environment variables and reorder the list:
+
+```yaml
+env:
+- name: ENV2
+  value: bar
+- name: ENV5
+  value: server-added-2
+- name: ENV1
+  value: foo
+- name: ENV3
+  value: baz
+- name: ENV4
+  value: server-added-1
+```
+
+Then the user wants to change it from the original to the following using `kubectl apply`:
+
+```yaml
+env:
+- name: ENV1
+  value: foo
+- name: ENV2
+  value: bar
+- name: ENV6
+  value: new-env
+```
+
+The old patch without parallel list will looks like:
+
+```yaml
+env:
+- name: ENV3
+  $patch: delete
+- name: ENV6
+  value: new-env
+```
+
+The new patch will looks like below. It is the
+
+```yaml
+$setElementOrder/env:
+- name: ENV1
+- name: ENV2
+- name: ENV6
+env:
+- name: ENV3
+  $patch: delete
+- name: ENV6
+  value: new-env
+```
+
+After server applying the new patch:
+
+```yaml
+env:
+- name: ENV5
+  value: server-added-2
+- name: ENV4
+  value: server-added-1
+- name: ENV1
+  value: foo
+- name: ENV2
+  value: bar
+- name: ENV6
+  value: new-env
+```
+
+### List of Primitives
+
+We take finalizers as an example.
+finalizers is a list of strings.
+
+Suppose we define a list of finalizers and we call them
+the original finalizers:
+
+```yaml
+finalizers:
+- a
+- b
+- c
+```
+
+Then the server appends two finalizers and reorder the list:
+
+```yaml
+finalizers:
+- b
+- e
+- a
+- c
+- d
+```
+
+Then the user wants to change it from the original to the following using `kubectl apply`:
+
+```yaml
+finalizers:
+- a
+- b
+- f
+```
+
+The old patch without parallel list will looks like:
+
+```yaml
+$deleteFromPrimitiveList/finalizers:
+- c
+finalizers:
+- f
+```
+
+The new patch will looks like below. It is the
+
+```yaml
+$setElementOrder/finalizers:
+- a
+- b
+- f
+$deleteFromPrimitiveList/finalizers:
+- c
+finalizers:
+- f
+```
+
+After server applying the patch:
+
+```yaml
+finalizers:
+- e
+- d
+- a
+- b
+- f
+```
+
+
+# Alternative Considered
+
+# 1. Use the patch list to set order
+
+## Proposed Change
+
+This approach can considered as merging the parallel list and patch list into one single list.
+
+For list of maps, the patch list will have all entries that are
+either a map that contains the mergeKey and other changes
+or a map that contains the mergeKey only.
+
+For list of primitives, the patch list will be the same as the list in users' local config.
+
+## Reason of Rejection
+
+It cannot work correctly in the following concurrent writers case,
+because PATCH in k8s doesn't use optimistic locking, so the following may happen.
+
+Live config is:
+
+```yaml
+list:
+- mergeKey: a
+  other: A
+- mergeKey: b
+  other: B
+- mergeKey: c
+  other: C
+```
+
+Writer foo first GET the object from the server.
+It wants to delete B, so it calculate the patch and is about to send it to the server:
+
+```yaml
+list:
+- mergeKey: a
+- mergeKey: b
+  $patch: delete
+- mergeKey: c
+```
+
+Before foo sending the patch to the server,
+writer bar GET the object and it want to update A.
+
+Patch from bar is:
+
+```yaml
+list:
+- mergeKey: a
+  other: A'
+- mergeKey: b
+- mergeKey: c
+```
+
+After the server first applying foo's patch and then bar's patch,
+the final result will be wrong.
+Because entry b has been recreated which is not desired.
+
+```yaml
+list:
+- mergeKey: a
+  other: A
+- mergeKey: b
+- mergeKey: c
+  other: C
+```
+
+# 2. Use $position Directive
+
+## Proposed Change
+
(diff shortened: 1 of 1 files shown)


mengqiy:
Anyone help to review this?

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
+# Preserve Order in Strategic Merge Patch
+
+author: @mengqiy
+
+## Motivation
+
+The order of lists with a merge strategy is important, because an item in the list may depends on an item before it.
```
pwittrock: >The Kubernetes API may apply semantic meaning the ordering of items within a list, however the strategic merge patch does not keeping the ordering of elements.  Ordering has semantic meaning for Environment variables, as later environment variables may reference earlier environment variables, but not the other way around.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+An use case is the environment variables. We don't preserve the order which causes
+issue [40373](https://github.com/kubernetes/kubernetes/issues/40373)
+
+## Proposed Change
+
+Changes are all in strategic merge patch package.
+It will be similar to how we solve the problem of deleting items from a list of primitives.
```
pwittrock: It will be similar to how we solve the problem of deleting items from a list of primitives.

> The proposed solution is similar to the solution used for deleting elements from lists of primitives.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+
+Always send a parallel list along with the patch for list:
```
pwittrock: > Add to the current patch, a directive containing a list of element keys - either the patch merge key, or for primitives the value.  When applying the patch, the server we ensure that the relative ordering of elements matches the directive.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+mergingList:
+- parallelListItem1 \
+  ...                |===> items from the parallel list
+- parallelListItemN /
+- otherItem1 \
+  ...         |===> items in the server's list but not in the parallel list
+- otherItemN /
+```
+
+## Version Skew
```
pwittrock: > Version Skew and Backwards Compatibility

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+Always send a parallel list along with the patch for list:
+
+- When patching a list of maps, the parallel list contains a list of items. Each item contains only the merge key.
+- When patching a list of primitives, the parallel list contains a full list from user's config file.
+
+All the items in the server's live list but not in the parallel list will be append to the end of the parallel list.
+The relative order between these appended items are kept.
+If the relative order of live config in the server is different from the order of the parallel list,
+user's patch will always override the order in the server.
+
```
pwittrock: Add an example of what the new patch would look like vs the old patch.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+So it will behave the same as before.
+
+## Example
+
+We take environment variables as an example.
+Environment variables is a list of maps with merge patch strategy.
+
+Suppose we define a list of environment variables and we call them
+the original environment variables:
+```yaml
```
pwittrock: add space before '`'

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+  - name: ENV2
+    value: bar
+  - name: ENV6
+    value: new-env
+  - name: ENV5
+    value: server-added-2
+  - name: ENV4
+    value: server-added-1
+```
+
```
pwittrock: Add example for primitives as well

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+Changes are all in strategic merge patch package.
+It will be similar to how we solve the problem of deleting items from a list of primitives.
+
+Always send a parallel list along with the patch for list:
+
+- When patching a list of maps, the parallel list contains a list of items. Each item contains only the merge key.
+- When patching a list of primitives, the parallel list contains a full list from user's config file.
+
+All the items in the server's live list but not in the parallel list will be append to the end of the parallel list.
```
pwittrock: Before this give a simple example

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,226 @@
...
+The relative order between these appended items are kept.
```
pwittrock: Add these as separate sections:

### Behavior when the ordering list contains only some elements

Explain why this would happen.  Explain what happens.

### Behavior when the ordering lists contains elements missing from live value

Explain why this would happen.  Explain what happens.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,231 @@
...
+    value: foo
+  - name: ENV2
+    value: bar
+  - name: ENV6
+    value: new-env
+```
+
+The patch will looks like:
+```yaml
+$preserveOrderList/env:
```
pwittrock: $setElementOrder

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,422 @@
...
+
+Background of the Strategic Merge Patch is convered [here](../devel/strategic-merge-patch.md).
+
+The Kubernetes API may apply semantic meaning to the ordering of items within a list,
+however the strategic merge patch does not keeping the ordering of elements.
+Ordering has semantic meaning for Environment variables,
+as later environment variables may reference earlier environment variables,
+but not the other way around.
+
+An use case is the environment variables. We don't preserve the order which causes
```
pwittrock: > An use case

One use case

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,422 @@
...
+issue [40373](https://github.com/kubernetes/kubernetes/issues/40373).
+
+## Proposed Change
+
+Changes are all in strategic merge patch package.
+The proposed solution is similar to the solution used for deleting elements from lists of primitives.
+
+Add to the current patch, a directive containing a list of element keys -
+either the patch merge key, or for primitives the value. When applying the patch,
+the server we ensure that the relative ordering of elements matches the directive.
```
pwittrock: > the server we ensure 

the server ensures

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,422 @@
...
+
+All the items in the server's live list but not in the parallel list will be append to the end of the parallel list.
+The relative order between these appended items are kept.
+If the relative order of live config in the server is different from the order of the parallel list,
```
pwittrock: > If the relative order of live config

If the relative order of the live config

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,422 @@
...
+user's patch will always override the order in the server.
```
pwittrock: > user's patch will al

the user's patch will al

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,422 @@
...
+mergingList:
+- parallelListItem1 \
+  ...                |===> items from the parallel list
+- parallelListItemN /
+- otherItem1 \
+  ...         |===> items in the server's list but not in the parallel list
+- otherItemN /
+```
+
+### `$setElementOrder` may contain elements not present in the patch list
```
pwittrock: First introduce this as the directive name.  I don't think we have done that yet.  Give a simple example first.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,422 @@
...
+- B
+```
+
+## Version Skew and Backwards Compatibility
+
+The new version patch is always a superset of the old version patch.
+The new patch has one additional parallel list which will be dropped by the old server.
+
+The logic for old version patch is still kept,
+so the patch from any arbitrary old clients used to work before will continue to function.
```
pwittrock: > so the patch from any arbitrary old clients used to work before will continue to function.

The new directive is optional.  Patch requests without the directive will retain the behavior from previous releases.  This will ensure backward compatibility with old clients.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,422 @@
...
+finalizers:
+- a
+- b
+- c
+```
+
+Then the server appends two finalizers and reorder the list:
+
+```yaml
+finalizers:
```
pwittrock: This is a really helpful example.  Good job.

pwittrock:
Minor nits on wording.  Otherwise LGTM.

mengqiy:
Updated.

pwittrock:
This looks good to me.

mengqiy:
Squashed commits.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,437 @@
...
+- B
+```
+
+## Version Skew and Backwards Compatibility
+
+The new version patch is always a superset of the old version patch.
+The new patch has one additional parallel list which will be dropped by the old server.
+
+The new directive is optional.
+Patch requests without the directive will retain the behavior from previous releases.
```
pwittrock: We should do something better

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,437 @@
...
+```yaml
+list:
+- A
+- B
+- C
+- D
+- E
+```
+
+### When `$setElementOrder` contains elements not found in the list to be merged
```
pwittrock: Add a section for 

#### When `$setElementOrder` is not present and patching a list.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,437 @@
...
+finalizers:
+- a
+- b
+- f
+- e
+- d
+```
+
+
+# Alternative Considered
```
pwittrock: Another alternative, is to send the mergeKeys as empty elements in the list, and have the server maintain the order of the provided items.  In this case, we don't need to introduce any new directives, and the patch should be totally compatible.  The patch list would always be reordered to come before all elements missing from the patch.

pwittrock: Thinking on this more, if we did that, we might end up creating elements that we just want to order - e.g. if we want to modify "A", and send "A, B, C" as the main list to keep ordering, "C" would get recreated if it was deleted or missing, which we would not want.

mengqiy:
Updated. PTAL.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
adohe:
```
For example ordering has semantic meaning for environment variables.
```
I would like to change to this, semantic meaning is not very common.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,547 @@
...
+
+One use case is the environment variables. We don't preserve the order which causes
+issue [40373](https://github.com/kubernetes/kubernetes/issues/40373).
+
+## Proposed Change
+
+Changes are all in strategic merge patch package.
+The proposed solution is similar to the solution used for deleting elements from lists of primitives.
+
+Add to the current patch, a directive ($setElementOrder) containing a list of element keys -
```
adohe: not every list wants to keep ordering, whether we will consider add a directive to type definition?

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,547 @@
...
+either the patch merge key, or for primitives the value. When applying the patch,
```
adohe: any consideration about multiple key as merge key?

mengqiy: The mergeKey is a key-value pair in the map. If we are not using any special format, I think new additional merge keys will simply be new  key-value pair in the map.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
mengqiy: > add a directive to type definition

I don't understand what do you mean. Can you give me an example?

pwittrock: @adohe I think you mean that the ordering is not semantically important for every list, and should we add a comment tag to the field definition indicating whether order has semantic meaning.  (e.g. The containers list in a PodTemplate does not have semantically meaningful order).

I think this would be be  interesting to do in a follow up.  Since this is totally optional and sent by the client, the client should decide whether it wants to control the order or not for a given patch (yes it should want to do so)

pwittrock:
@fabianofranz @liggitt I will be merging this today if there are no other comments.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,547 @@
...
+### When $setElementOrder is not present and patching a list
+
+The new directive $setElementOrder is optional.
+The new server applying a patch request without the directive will be
+a best-effort operation to keep the order of elements that are already in the list:
+
+For changes and deletions, the server will just merge the change without sorting them.
+So the items in the list will retain the order.
+
+For additions, the items will be appended to the end of the list.
```
liggitt: If the server has `[A,B,C]`
and the client sends `[X,A,B,C]`
always appending to the end has this counter-intuitive result: `[A,B,C,X]`

Are old clients known to scramble the order when computing the patch, or was the scrambling happening server-side?

I would have expected the order of two items to be determined by something like the following:
* relative order in the $setElementOrder if both items are present
* else relative order in the patch if both items are present
* else relative order in the server-side list if both items are present
* else append to the end

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,547 @@
...
+```yaml
+list:
+- A
+- B
+```
+
+### When the list to be merged contains elements not found in `$setElementOrder`
+
+If the list to be merged contains elements not found in $setElementOrder,
+they will come after all elements defined in $setElementOrder, but keep their relative ordering.
```
liggitt: it seems like an error for $setElementOrder to be missing items that are present in the list (or to have items in a different order than the list)

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
```diff
@@ -0,0 +1,547 @@
...
+The new version patch is always a superset of the old version patch.
+The new patch has one additional parallel list which will be dropped by the old server.
+
+As mentioned [above](#when-setelementorder-is-not-present-and-patching-a-list),
+the new directive is optional.
+Patch requests without the directive will change a little,
+but still be fully backward compatible.
+
+### kubectl
+If an old kubectl sends a old patch to a new server, the server will not preserve the order which behave as an old server.
```
liggitt: can we not do better here by honoring relative order of items in the patch?

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
mengqiy: > Are old clients known to scramble the order when computing the patch, or was the scrambling happening server-side?

Both. We sort the list by `mergeKey` when diffing and merging in current strategic merge patch implementation.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
mengqiy: As mentioned above https://github.com/kubernetes/community/pull/537#discussion_r114874780, a patch sent by old kubectl doesn't has any useful order in it. Because it has been sorted by mergeKey.
What should we do in this case?

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
mengqiy: > it seems like an error for $setElementOrder to be missing items that are present in the list

I'm OK to fail the operation. I will update it soon.

> or to have items in a different order than the list

Do you mean we should always verify the order in patch list and the order in $setElementOrder match before merging?

mengqiy:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/537)

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
pwittrock: @liggitt I was thinking about this as well.  

I think you are saying, can you just use the ordering in the patch list instead of sending another directive.

The issue here is: If you have a list of [a,b,c] and want to patch b you would either send a patch [b] or send a patch [a,b,c].  The former lacks the ordering, and the latter will recreate [a,c] if they were deleted from the server.

liggitt: I don't think it removes the need for the separate order-indicating directive, but it seems really counter-intuitive to ignore relative order in the list

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
mengqiy: @liggitt What should the result be when applying patch [c,b,a] to server's list [a,x,b,y,c]?

mengqiy: I cannot find a way to keep relative order of (a,x) and (x,b) at the same time.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
pwittrock: I think @liggitt was suggesting both are required.  I think that is probably fine.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
pwittrock: @liggitt Your later comment addresses this I think - all items in the list must appear in the orderList

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
pwittrock: @liggitt Now sure how much of an issue you think this is.  I think it is an edge case, but one that be annoying to track down if it is encountered.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
liggitt: I was expecting something like https://play.golang.org/p/tJLW1Zj0h7 where the directive is used if present, otherwise the relative order in the patch list is honored for item pairs which are both present, and finally the relative server-side order is honored.

that is less compelling if old clients sent randomly ordered lists

mengqiy: Thanks a lot. This is super helpful.

mengqiy:
PTAL

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
adohe: @pwittrock yep, that's exactly what I mean. We could do this in a follow up and I am just wondering how the client can decide whether or not to keep list order.

contributors/design-proposals/preserve-order-in-strategic-merge-patch.md
pwittrock: SGTM

mengqiy:
Squashed commits.

pwittrock:
@liggit merging this since it looks like you comments have been addressed

PR #247, opened 2017-01-02
Allow pvc.Selector and pvc.Class both to be set.

jsafrane:
To simplify initial implementation we made selector and class mutually
exclusive. Now it's time to finish the implementation and let users to
choose e.g. a zone where a volume should be provisioned in pvc.Selector.

@kubernetes/sig-storage-misc

PR #543, opened 2017-04-18
Proposal for user configuration of dynamic provisioning

jsafrane:
This is counter-proposal for https://github.com/kubernetes/community/pull/247. I've been talking to @msau42 and @saad-ali and with upcoming scheduler changes it would be much better not to allow generic selector as proposed in #247, which is hard to evaluate and allow only exact configuration of dynamic provisioning using ConfigMaps in user namespace(s).

I leave both PRs open and close one when we decide which way to go.

Pros of ConfigMaps:

  *  PVCs are platform independent, as long as user supplies updated `ConfigMap(s)` when deploying to another platform.

  * It is strict and thus easier to handle in the provisioner and in scheduler. No selectors like "zone NOT IN [east, west]"

Cons:

 * It works only with dynamic provisioning.
   * With Selectors, user might request a PV with certain labels and a PV would be provisioned *only* if no existing PV matched.
   * With `ConfigMap`, a PV must be always dynamically provisioned as we don't have any way how to check if an existing PV conforms to a `ConfigMap`.

 * One extra object to care about.

 * External dynamic provisioners must be able to read all ConfigMaps on the system. This should be OK as long as users don't use ConfigMaps for their mysql passwords.

Again, "zone" is used here just for an example, it may be phased out by upcoming scheduler changes, but we still need a way how to configure encryption or nr. of Gluster replicas.

cc: @thockin @vishh @jwforres @derekwaynecarr @eparis 
@kubernetes/sig-storage-api-reviews

### DIFF
--- a/contributors/design-proposals/volume-provisioning-configuration.md
+++ b/contributors/design-proposals/volume-provisioning-configuration.md
@@ -1,0 +1,207 @@
+## Abstract
+
+This proposal gives regular users which cannot edit StorageClasses possibility to configure some aspects of dynamic provisioning.
+
+## Overview
+
+In Kubernetes 1.4 we introduced StorageCasses as a way how admins can configure dynamic provisioning of PersistentVolumes. This requires admins to create a StorageClass instance for every possible configuration, which is cumbersome. In addition, regular users that can require a PV with specific properties typically cannot create StorageClasses.
+
+## Terminology
+
+* *admin* - user that can edit StorageClass objects.
+* *regular user* - user that can create PersistentVolumeClaims and cannot edit StorageClasses.
+
+## Use cases
+
+* Regular user wants get a PV of storage class "fast" in availability zone "east".
+* Admin wants to allow users to create PVs with storage class "fast" only in availability zones "east" and "north".
+* Regular user wants to discover what tunable properties for storage class "fast" are available and as result he can see that parameter "availability zone" with allowed values "east" and "north" is configurable.
+
+While availability zone is used as example in use cases, provisioners can choose any property such as encryption key, IOPS or replica count as tunable.
+
+## Design
+
+* Users can use `PersistentVolumeClaim.spec.provisioningOptions` with a pointer to `ConfigMap` in the same namespace. This `ConfigMap` contains options for the dynamic provisioning of a PV for this PVC.
+   ```go
+   type PersistentVolumeClaimSpec struct {
+      ...
+      // ProvisiongConfiguration contains a reference to ConfigMap with configuration of dynamic provisioning for this claim.
+      ProvisiongConfiguration *LocalObjectReference
+   }
+   ```
+   
+    * This way, PVCs are platform-independent and can be deployed elsewhere, as long as the user updates `ConfigMap(s)` with platform-dependent details of the provisioning.
+    * PV controller would ignore existing PVs and it would **always provision a new PV** for a PVC with non-empty `spec.provisioningOptions` - we do not have any way how to check if an existing PV conforms to `spec.provisioningOptions` so dynamic provisioning is the only way.
+    * All provisioners (both internal and external) must provision a PV that satisfies both `StorageClass.parameters` and `PVC.spec.provisioningOptions` or throw an error if such combination of options is not possible.
+    * When user does not specify any `provisioningOptions` or the referred `ConfigMap` does not specify all allowed parameters, the provisioner either uses some defaults or throws an error that some parameters in the ConfigMap are mandatory.
+
+* StorageClass will get a new field to tell regular users what are allowed options in `PVC.spec.provisioningOptions`:
+    ```go
+    type StorageClass struct {
+        // <snip>
+
+        // AllowedProvisioningOptions contains details about provisioning options that are allowed by this StorageClass.
+        AllowedProvisioningOption  []ProvisioningOptionSpec
+    }
+
+    // ProvisioningOption describes one option that can be used in PVC provisioningOptions
+    // when requesting a volume of a particular StorageClass.
+    type ProvisioningOptionSpec struct {
+        // Name of key in ConfigMap
+        Key string
+
+        // Optional default value if it is unspecified in the ConfigMap.
+        DefaultValue *string
+
+        // Optional flag whether this selector key is mandatory in the ConfigMap.
+        // Defaults to "false".
+        Required bool
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/volume-provisioning-configuration.md
```diff
@@ -0,0 +1,207 @@
...
+    type StorageClass struct {
+        // <snip>
+
+        // AllowedProvisioningOptions contains details about provisioning options that are allowed by this StorageClass.
+        AllowedProvisioningOption  []ProvisioningOptionSpec
+    }
+
+    // ProvisioningOption describes one option that can be used in PVC provisioningOptions
+    // when requesting a volume of a particular StorageClass.
+    type ProvisioningOptionSpec struct {
```
jwforres: this is a similar approach to how we handled Parameters in OpenShift templates, but we've found it generally ended up being insufficient / we kept needing to add fields to it to improve the experience.  In the open service broker API we decided instead to define parameters via JSON schema v4 since that will provide more flexibility and there are a plethora of JSON schema validators (and GUI form builders) out there.

Although from the perspective of the storage admin that is setting this up, it is more complex to define a JSON schema than a limited set of fields like is proposed here.

contributors/design-proposals/volume-provisioning-configuration.md
```diff
@@ -0,0 +1,207 @@
...
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1500Mi
+  provisioningOptions:
+    name: my-provisioning-config
+```
+
+* Note that this ConfigMap is used only during provisioning. Changes in the ConfigMap are not reflected in already provisioned PVs.
```
jwforres: This is actually why I don't like this approach.  From a GUI perspective a user is probably not going to be picking an existing ConfigMap, I see us instead providing a form to fill out the parameter values and then generating a new ConfigMap to go along with the PVC they are creating.  But this extra ConfigMap will sit around forever, and probably never be used again. This just adds clutter to the GUI for ConfigMaps the users actually DO care about, like their application configs.  I also don't like that this means two requests to set up the PVC correctly, 1 for the PVC and 1 to create a config map.  It's increasing the surface area for failure.

msau42: Some alternatives are:
1. (Current method) Create a different storageclass for each permutation of parameters.  This causes clutter to the StorageClasses, but it eliminates the two-step PVC process.
2. Have the parameters as a map in the PVC.  This breaks the PVC "portability".  
3. Have each storage feature exposed as a first class field in the PVC.  It could be difficult to represent all the features generically, and still has the question of portability.  What happens if you have zone in the PVC, but then you take the spec on a platform that doesn't support zones?

davidopp:
ref/ kubernetes/kubernetes#41442

jsafrane:
@thockin, can you please take a look and compare with https://github.com/kubernetes/community/pull/247

contributors/design-proposals/volume-provisioning-configuration.md
```diff
@@ -0,0 +1,207 @@
...
+
+```yaml
+apiVersion: storage.k8s.io/v1beta1
+kind: StorageClass
+metadata:
+  name: slow
+provisioner: kubernetes.io/aws-ebs
+parameters:
+  type: gp2
+  zones: us-east-1a, us-east-1b, us-east-1c, us-east-2a, us-east-2b, us-east-2c
```
krmayankk: why do you repeat all the allowed values for zones here ? If the PVC doesnt specify a zone , then what is the default value you we get  or the user would always have to specify a zone in the PVC using config map ?

krmayankk: i see below you mean the admin restricts to these zones. if this doesnt repeated everything in the allowed, it will look more realistic example. My question about is there a default still stands ?

jsafrane: if PVC does not specify a parameter value, e.g. a zone, then the provisioner is free to choose one. With zone, we do semi-random distribution of PVCs across all zones, for other parameter the default may be different.

contributors/design-proposals/volume-provisioning-configuration.md
jsafrane: The idea is that most people won't care about configuration of storage at all. Admin prepares a handful of storage classes for generic usage and user pick one of them and that's it. ConfigMap is for those who need specific options for their data. This proposal assumes there will be a limited amount of those, otherwise it gets indeed complicated.

contributors/design-proposals/volume-provisioning-configuration.md
```diff
@@ -0,0 +1,207 @@
...
+
+        // Optional default value if it is unspecified in the ConfigMap.
+        DefaultValue *string
+
+        // Optional flag whether this selector key is mandatory in the ConfigMap.
+        // Defaults to "false".
+        Required bool
+
+        // Optional list of values that are allowed for this key in the ConfigMap.
+        AllowedValues []string
```
lpabon: also, what if the necessary value is a number value? Like` [ Key:"CrazyNumberOfDisks", Value: "34" ]` ? Should one of the AllowedValues strings be something like "0-100"? Not necessary for the first version.

contributors/design-proposals/volume-provisioning-configuration.md
```diff
@@ -0,0 +1,207 @@
...
+  name: myclaim
+spec:
+  class: slow
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1500Mi
+  provisioningOptions:
+    name: my-provisioning-config
```
lpabon: :+1:

contributors/design-proposals/volume-provisioning-configuration.md
lpabon: I'm not really sure this would cause an issue that wouldn't be solved by other methods in a GUI, but I understand what you mean. It could get cluttered.

I really like this idea. There is a connection from ConfigMap--PVC--StorageClass--PV. I think the ConfigMap (which a UI could create for the user) would be an easy way for that user to adjust the PVC request.

jsafrane:
I am closing this proposal, we as sig-storage have other priorities and we did not find people to lead this effort. Volunteers are welcome at https://github.com/kubernetes/features/issues/244 !

PR #545, opened 2017-04-18
Container Device Interface (CDI) proposal

urzds:
We are interested in using various types of devices, e.g. GPUs, in our containers running on Kubernetes. Thus we propose a generic "Container Device Interface" (CDI), that shall provide an extensible way to easily add support for new device types to Kubernetes.

We would love to hear feedback from the community on this proposal.

### DIFF
--- a/contributors/design-proposals/container-device-interface.md
+++ b/contributors/design-proposals/container-device-interface.md
@@ -2,48 +2,231 @@
 
 **Author**: Dennis Schridde <dennis.schridde@uni-heidelberg.de> (@urzds)
 
-**Version**: 0.0.1
+**Version**: 0.0.2
 
-**Last edit**: [2017-04-18](#history)
+**Last edit**: [2017-05-05](#history)
 
 **Status**: Draft proposal
 
-## Motivation
+## Introduction
 
-We use Kubernetes to schedule scientific 3D visualisation applications onto a cluster. Hence our containers use GPUs / DRM and other device nodes (e.g. TTYs).  Scheduling of Pods to Nodes with a capacity for such devices can be dealt with using [Opaque Integer Resources (OIR)](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#opaque-integer-resources-alpha-feature), where scripts external to Kubernetes handle device discovery and capacity reporting (see e.g. [kube-gpu-resources](https://github.com/urzds/kube-gpu-resources)).  The [Container Runtime Interface (CRI)](https://github.com/kubernetes/community/blob/master/contributors/devel/container-runtime-interface.md) already supports [binding host device nodes](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/api/v1alpha1/runtime/api.proto#L586) to a container.  What is currently missing is a way for the Kubelet to generically allocate the available devices of a certain type to a container while it is being started.  The Container Device Interface (CDI) described here shall bridge this gap.
+### Motivation
 
-## Description
+We use Kubernetes to schedule scientific 3D visualisation applications onto a cluster. Hence our containers use GPUs with the Linux Direct Rendering Manager (DRM) and other device nodes (e.g. TTYs).  Scheduling of Pods to Nodes with a capacity for such devices can be dealt with using [Opaque Integer Resources (OIR)](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#opaque-integer-resources-alpha-feature), where scripts external to Kubernetes handle device discovery and capacity reporting (see e.g. [kube-gpu-resources](https://github.com/urzds/kube-gpu-resources)).  The [Container Runtime Interface (CRI)](https://github.com/kubernetes/community/blob/master/contributors/devel/container-runtime-interface.md) already supports [binding host device nodes](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/api/v1alpha1/runtime/api.proto#L586) to a container.  What is currently missing is a way for the Kubelet to generically allocate the available devices of a certain type to a container while it is being started.  The Container Device Interface (CDI) described here shall bridge this gap.
 
-CDI is modelled after CNI in that it consist of an API between the Kubelet and CDI binaries that transports instructions through environment variables and configuration and data through JSON chunks transmitted through stdin/stdout.  As in CNI, which binaries to call and their configuration is described through JSON configuration files located in `/etc/cdi/<device-type>.d/<binary-name>.conf`.  The principal commands are `ADD` and `DEL`, which allocate a device to a container or return it to the node's pool.  All commands expect a `CDI_VERSION` environment variable, to ensure the version used by the Kubelet (environment variable), the CDI config (JSON chunk on stdin) and the binary match.
+We propose out-sourcing the management of device node allocations to external binaries, to easily allow to support different device types and provide the possibility for future extensions e.g. using device partitioning.  (E.g. PCIe SR-IOV could be used to dynamically allocate a certain amount of timeslices or VRAM.)  We propose an API similar to CNI, since its design is already well tested within Kubernetes and capable of this task.  This specification also contains a command that can be used for device discovery, which can be used to implement or supplant OIR for device nodes.
+
+### Overview
+
+CDI is modelled after CNI in that it specifies an API between the Kubelet and CDI binaries, which transports requests through environment variables and configuration and responses through JSON chunks transmitted via stdin/stdout.  As with CNI, which binaries to call and their configuration is described via JSON configuration files located in `/etc/cdi/<resource-type>.d/<binary-name>.conf`.  The binaries themselves are to be located in `/opt/cdi/bin`, similar to CNI.
+
+### Terminology
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+The terms `resource-type`, `resource-subtype`, `resource-spec` and `request-id` shall be defined by the [ABNF rules below](#common-abnf-rules).
+
+We generally distinguish the terms "device" (meaning the physical hardware) and "device node" (meaning the file-like entity in the Linux file-system).
+
+## Configuration
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/container-device-interface.md
```diff
@@ -0,0 +1,138 @@
...
+## Usage examples
+
+### GPU / DRM
+
+```
+# cat /etc/cdi/gpu.d/drm.conf
+{
+  "cdiVersion": "0.0.1",
+  "name": "drm-gpus",
+  "type": "drm",
```
squeed: Hi there, CNI maintainer here.
One thing we learned from CNI was that calling this field `type` was super confusing to our users. I strongly suggest calling this `plugin-name`.

squeed:
It looks like this is built on CNI, which is great! As a CNI maintainer, there are a few things we've done recently that you might want to think about:

* Plugin chaining: we have a cool way to chain multiple plugins in a single execution - it's proving to be super useful for more specialized use cases that you might not want to put in every plugin
* If you choose to provide some reference plugin implementation, make sure they're in a separate repository from the spec

Feel free to reach out to the CNI team if you've got more questions.

vishh:
It would be awesome if the kubelet can discover additional devices through this proposed means and allocate it to pods and/or containers instead of going through the CRI.
OIR in its current state is not meant to tackle compute resources that require special operations like that of adding a device file.
CRI in it's current state is not expected to handle compute resource discovery, isolation and management.
What if kubelet can expose a bunch of generic devices as compute resources through this CDI like interface? It can then take into account PCIe locality when necessary and provide pod level shared access to devices too.

davidopp:
@ConnorDoyle

ConnorDoyle:
This way of exposing devices to Kubelet is pretty nice: easy to understand and explain.

To expand on the theme of what's been said so far, most of the questions that come to mind are around UX for pod spec authors. Could you elaborate on how you see this interacting with OIRs (if at all)? Maybe some sample container specs would help show how users might a) consume an exclusively allocated device b) consume a shareable device c) consume discrete fractions of a device.

contributors/design-proposals/container-device-interface.md
urzds: Thanks for the suggestion! The next version will have this changed. I chose `plugin` instead of `plugin-name` to prevent confusion with the `name` field.

urzds:
@squeed Could you explain why in CNI you chose to pass the config via stdin and the request via env-vars? I went with the same scheme, because that's what CNI did, but passing a link to the config via a `CDI_CONFIG=/path/to/file.json`, sending the request through stdin and receiving the plugin's response on stdout seems more natural to me. That would even allow to pass the request through to another plugin / instance, possibly via network to another machine, without having to worry about loosing some information that was unknown to the plugin, but important to the actual handler. This already works for plugins on the same machine, if one inherits its environment to the other, but passing the request via stdin would work in a more general sense. If you allow special cases like `CDI_CONFIG=fd:3`, you could still pass the configuration through a FIFO on a file-descriptor, in case direct access to a file is not possible or there is no actual file to begin with.

I am taking your comments regarding plugin chaining into consideration and will try to have a sample implementation (of the CDI, without the Kubelet changes) ready before the meeting on Tuesday.

@derekwaynecarr Adding NUMA locality seems to be a simple extension. I would add a `CDI_PREFER_NUMA_LOCALITY=0,1,6,7` environment variable to instruct the plugin of your request. The plugin could probably figure out which devices are local to which NUMA node by examining the PCI bus IDs of the devices. The response could then include a `"numa_locality":[0,1]` JSON field, to inform the Kubelet of what location was actually chosen, so the Kubelet could adjust the call to the CRI accordingly.

@vishh Sharing devices between multiple containers (and thus CDI ADD requests) should be possible by adding another environment variable `CDI_SHARE_ID=<pod-id>`. The CDI plugin could then open a new "share-group" and all future ADD requests with the same `CDI_SHARE_ID` would receive the same device, without actually allocating additional devices. The Kubelet / scheduler would then have to ensure that the different containers do not request varying numbers of devices, since then it would be difficult to decide which device shall actually be shared and which not.

@ConnorDoyle Right now my idea is that something else (could be the CDI plugin as response to an `INIT` command, but it could be something else entirely, like a [kube-gpu-resources](https://github.com/urzds/kube-gpu-resources) static pod) informs the api-server / scheduler via OIR of the number of available devices on this node. Any pod that arrives on the node through this means gets handed over to the CDI plugin responsible for the specific OIR (one could use the "name" part of the OIR to look up the corresponding config file - i.e. `pod.alpha.kubernetes.io/opaque-int-resource-gpu` would look at the config files `/etc/cdi/gpu.d/*.json`).

So far I did not think about sharing devices between multiple containers (that was not a use-case for us). My suggestion would be to request the OIRs on the pod-level (not sure whether that is possible already) and then allow the containers to piggy-back on that when the pod actually arrives at a node.

Something like this:
```yaml
apiVersion: fantasy/v1alpha
kind: Pod
metadata:
  name: gpu-sharing-pod
spec:
  resources:
    request:
      pod.alpha.kubernetes.io/opaque-int-resource-gpu: 1
  containers:
  - name: pod-1
    resources:
      shared:
      - pod.alpha.kubernetes.io/opaque-int-resource-gpu
  - name: pod-2
    resources:
      shared:
      - pod.alpha.kubernetes.io/opaque-int-resource-gpu
```

The scheduler would look at the pod-level `request` to find a suitable node. The Kubelet on the node would look at the `shared` resources to call the corresponding CDI with the values it derives from the pod-level `request`.

In this draft I chose to not allow `shared` resources to specify a quantity, to prevent the problem of "varying numbers of devices", mentioned above in this comment. The number of GPU resources both containers would share would be exclusively determined through the resource `request` of the pod.

Through this means one could even have some OIRs `shared` among multiple containers in the pod, while some containers request additional OIRs through the regular container-level `request`. Since some devices would belong to a "share-group" (s.a. in this comment), while others do not, two calls to the CDI plugin would have to be made - one with `CDI_SHARE_ID` set and another without. Since the mechanism is so terribly simple, with the CDI only informing the Kubelet of what device nodes / files to use, this would be possible. The CDI spec would have to mention that plugins may be called multiple times with the same `CDI_CONTAINER_ID`, with the meaning that the union of all requested devices shall be allocated.

vishh:
@urzds 
I still think the act of injecting devices cannot be `Opaque` to the system. The kubelet has to orchestrate that as a first class thing. So I strongly prefer not following the proposed approach of using OIRs as it is today. We are working on a better representation for compute resources and representing arbitrary devices is a priority there. I'd recommend waiting for a proposal on that before tackling extensible device management.

derekwaynecarr:
@vishh +1 on the `Opaque` comment.

pod spec authors could just say make a resource requirement for `device-{name}`.  any name that followed that convention would be considered exclusive to the pod and would be expressed in integral values with no overcommit (i.e. request==limit).

@ConnorDoyle - if we need to support consumption of fractional devices, maybe those devices should have a convention of `shared-device-{name}` and support fractional value 0..1?

we have previously discussed introducing resource requirements at the pod spec which has a potentially nice side-effect of allowing `G` quality of service spread across `Bu/Be` containers within the pod.  For `device-{name}` resources, I would want to prevent over-commit semantics, so if a container wanted a unique and shared device in the pod spec, the sum would need to equal the pod request.  Since we have pod level cgroups on-by default now, I am less apprehensive about adding pod level resource requirements, but I need to reason through the edge cases around `cpu` and `memory`.  we would need to define defaulting semantics around pod resource requirements, but I think they can follow the same general convention we have on the local qos enforcement on the node for how we manifest pod cgroup sandbox.

derekwaynecarr:
@urzds -- if we pursue the separate binary route, I would desire the binary that handles device scheduling to also support an endpoint for discovering numbers of allocatable devices.  since this dictionary is already present, why not extend it to take over discovery that the kubelet integrates against?

RenaudWasTaken:
Hi !

Reading this proposal I'm a bit confused as to when in the CRI process would the binary be called ?
There is a CDI_CONTAINERID variable implying that it would be called after the container creation right ?

You are also talking about "virtualisation techniques" and I'm wondering if there is a way to know if the runtime behind CRI is creating VMs or containers ? 

My last question is related to the GPU use case :) (and more specifically nvidia) how do you tackle the problem of loading nvidia_uvm and the correct libs ?
What if you need to do different things depending on the runtime ?

squeed:
Regarding configuration being file vs. stdin, ultimately the justification is that it's simple. Given that plugins will be written in other languages, anything that increases the scope that plugin & runtime authors have to implement is a bad thing.

Ultimately, there is a bit of awkwardness commingling "static" configuration with "dynamic" or "runtime" information, but it's not so bad and is easy to implement.

urzds:
@RenaudWasTaken 

> Reading this proposal I'm a bit confused as to when in the CRI process would the binary be called ?
> There is a CDI_CONTAINERID variable implying that it would be called after the container creation right ?

Naming it `CDI_CONTAINERID` was a mistake. I fixed that in the next version of the specification (0.0.2) and also added a paragraph explaining when the CDI plugin is supposed to be called.

> You are also talking about "virtualisation techniques" and I'm wondering if there is a way to know if the runtime behind CRI is creating VMs or containers ?

I have never thought of that. I assume such a CRI would not be able to deal with `kubecontainer.DeviceInfo`?

> My last question is related to the GPU use case :) (and more specifically nvidia) how do you tackle the problem of loading nvidia_uvm and the correct libs ?

I don't. As far as I know Nvidia has a Docker plugin, which would detect the requirement for mounting the appropriate libraries from the host. I assume they could detect that `/dev/nvidia0` is being passed in `kubecontainer.DeviceInfo` and then take appropriate measure?

> What if you need to do different things depending on the runtime ?

Hopefully I don't. My assumption was that all CRIs are required to handle `kubecontainer.DeviceInfo`, and hence I could just piggy-back on that.

> Is that something left to kubelet ?
> I can imagine this problem for other devices so you could probably generalize the question to other hardware :D

Could you please elaborate?

@squeed 

* Plugin chaining: I have to think more about that to make it fully useful. As a first example I would like to have an example work-flow for the PCIe SR-IOV case, with a CDI plugin that creates a VF and binds it in response a CDI request, and then hands it over to the more generic DRM CDI plugin.
* Config / stdin / envvars: I am currently, and especially after reading about CNI configuration lists, in favour of passing the config with the command through stdin. Those two are closely coupled anyway, especially if we start chaining plugins.

urzds:
In version 0.0.2 I tried to address most of your concerns. If you have more suggestions, please keep them coming.

What I left out for now:
* @RenaudWasTaken's concerns about interaction with CRIs that use VMs
  - I need more explanations about the actual problem and what CRI actually specifies and what it leaves open
* @squeed's suggestion for chain-loading
  - It's just a matter of time until I find a method that convinces me

fabiand:
How does this proposal relate to #695?

cmluciano:
@urzds Do you think we should close this in favor of #695 ?

PR #549, opened 2017-04-19
Be more precise about times of SIG-Node meetings

urzds:
* 10 PT probably means 10 AM PT
* Also add time-zone as UTC for our European friends, for whom PT is less common


xiangpengzhao:
@urzds FYI: the time has already been updated in the latest doc.

cblecker:
Thanks for this @urdz, but yes, it's already fixed on head. :)

PR #551, opened 2017-04-19
devel: add threat model to security release process

philips:
As discussed we likely won't consider certain issues as security issues
at this time.


timstclair:
I think we should just wait for https://github.com/kubernetes/community/pull/504 to merge, and reference it from here. Alternatively, reference the PR from this document and state that it's a work in progress.

philips:
Good idea, will wait for #504

jessfraz:
Awesome!

PR #552, opened 2017-04-19
Proposal for GCE L4 load-balancer health check

MrHohn:
From kubernetes/kubernetes#14661.

I guess it may be better to have this L4 healthcheck feature in-place before promoting ESIPP to GA, just to make sure we won't need to change semantics for OnlyLocal service in future.

@nicksardo @thockin @bowei

### DIFF
--- a/contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
+++ b/contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
@@ -1,0 +1,61 @@
+# GCE L4 load-balancers' health checks for nodes
+
+## Goal
+Set up health checks for GCE L4 load-balancer to ensure it is only
+targeting healthy nodes.
+
+## Motivation
+On cloud providers which support external load balancers, setting the
+type field to "LoadBalancer" will provision a L4 load-balancer for the
+service ([doc](https://kubernetes.io/docs/concepts/services-networking/service/#type-loadbalancer)),
+which load-balances traffic to k8s nodes. As of k8s 1.6, we don't
+create health check for L4 load-balancer by default, which means all
+traffic will be forwarded to any one of the nodes blindly.
+
+This is undesired in cases:
+- k8s components including kubelet dead on nodes. Nodes will be flipped
+to unhealthy after a long propagation (~40s), even if we remove nodes
+from target pool at that point it is too slow.
+- kube-proxy dead on nodes while kubelet is still alive. Requests will
+be continually forwarded to nodes that may not be able to properly route
+traffic.
+
+For now, the only case health check will be created is for
+[OnlyLocal Service](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-typeloadbalancer).
+We should have a node-level health check for load balancers that are used
+by non-OnlyLocal services.
+
+## Design
+Healthchecking the kube-proxys seems to be the best choice:
+- kube-proxy runs on every nodes and it is the pivot for service traffic
+routing.
+- Port 10249 on nodes is currently used for both kube-proxy's healthz and
+pprof.
+- We already have a similar mechanism for healthchecking OnlyLocal services
+in kube-proxy.
+
+The plan is to enable health check on all LoadBalancer services (if use GCP
+as cloud provider).
+
+## Implementation
+kube-proxy
+- Separate healthz from pprof (/metrics) to use a different port and bind it
+to 0.0.0.0. As we will only allow traffic from load-balancer source IPs, this
+wouldn't be a big security concern.
+- Make healthz check timestamp in iptables mode while always returns "ok" in
+other modes.
+
+GCE cloud provider (through kube-controller-manager)
+- Manage `k8s-l4-healthcheck` firewall and healthcheck resources.
+These two resources should be shared among all non-OnlyLocal LoadBalancer
+services.
+- Add a new flag to pipe in the healthz port num as it is configurable on
+kube-proxy.
+
+Version skew:
+- Running higher version master (with L4 health check feature enabled) with
+lower version nodes (without kube-proxy exposing healthz port) should fall
+back to the original behavior (no health check).
+- Rollback shouldn't be a big issue. Even if health check is left on Network
+load-balancer, it will fail on all nodes and fall back to blindly forwarding
+traffic.


contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
```diff
@@ -0,0 +1,57 @@
...
+Set up health checks for GCE L4 load-balancer to ensure it is only
+targeting healthy nodes.
+
+## Motivation
+On cloud providers which support external load balancers, setting the
+type field to "LoadBalancer" will provision a L4 load-balancer for the
+service ([doc](https://kubernetes.io/docs/concepts/services-networking/service/#type-loadbalancer)),
+which load-balances traffics to k8s nodes. As of k8s 1.6, we don't
+create health check for L4 load-balancer by default, which means all
+traffics will be forwarded to any one of the nodes blindly.
```
bowei: traffic (no s)

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
```diff
@@ -0,0 +1,57 @@
...
+Eventually this annotation will be deprecated and all load-balancers should
+come with health check.
+
+kube-proxy
+- Make healthz bind to 0.0.0.0 by default (for now we only bind localhost).
+As we only allow traffics from load-balancer source IPs to healthz port,
+this wouldn't be a big security concern.
+
+GCE cloud provider
+- Manage `k8s-l4-default-healthcheck` firewall and healthcheck resources.
```
nicksardo: I'm not a big fan of having `default` in the name as someone will assume it means namespace-specific. 

What's the name the firewall that cloud provider generates for each OnlyLocal service?

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Implementation
+
+API
+- Define an annotation on service for opting out load-balancer health check.
+Eventually this annotation will be deprecated and all load-balancers should
+come with health check.
+
+kube-proxy
+- Make healthz bind to 0.0.0.0 by default (for now we only bind localhost).
```
bowei: I think we should have a separate web server binding for the healthz port. I *think* /metrics and other metadata is currently available off of this port.

nicksardo: Didn't we once encounter a problem where the proxy server was responding but iptables failed to be updated? Should/could we do something low cost to verify iptables is being actively updated? For instance checking a time of last update was in the past five minutes?

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
MrHohn: Did a quick check and surprisingly found we don't create any firewall rules for OnlyLocal service health check. Still looking into how.

MrHohn: Nick told me offline, health check for L4 load-balancer currently comes from 169.254.169.254. In the future it will come from other external IPs so we will need additional firewall rule.

And the naming is a good point :)

nicksardo: https://github.com/kubernetes/kubernetes/issues/40778

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
MrHohn: > I think we should have a separate web server binding for the healthz port.

Make sense to separate /healthz and /metriz into different ports.

MrHohn: > Should/could we do something low cost to verify iptables is being actively updated?

Good point, sounds like this could be easily achieved via a timestamp.

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
MrHohn: Done.

MrHohn:
Thanks for the comments, updated the doc a bit.

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
```diff
@@ -0,0 +1,59 @@
...
+type field to "LoadBalancer" will provision a L4 load-balancer for the
+service ([doc](https://kubernetes.io/docs/concepts/services-networking/service/#type-loadbalancer)),
+which load-balances traffic to k8s nodes. As of k8s 1.6, we don't
+create health check for L4 load-balancer by default, which means all
+traffic will be forwarded to any one of the nodes blindly.
+
+This is undesired in cases:
+- k8s components including kubelet dead on nodes. Nodes will be flipped
+to unhealthy after a long propagation (~40s), even if we remove nodes
+from target instance group at that point it is too slow.
```
nicksardo: This is still "target pool" until we migrate to backend services.

MrHohn: Thank, done.

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
```diff
@@ -0,0 +1,59 @@
...
+API
+- Define an annotation on service for opting out load-balancer health check.
+Eventually this annotation will be deprecated and all load-balancers should
+come with health check.
+
+kube-proxy
+- Separate healthz from pprof (/metric) to use a different port and bind it
+to 0.0.0.0. As we will only allow traffic from load-balancer source IPs, this
+wouldn't be a big security concern.
+- Do something low cost to verify iptables is being actively updated (probably
```
MrHohn: @nicksardo I'm also thinking about maybe we should make this generic instead of iptable mode specific?

nicksardo: Sounds great, what did you have in mind? I'm only aware of the iptables mode.

MrHohn: In the short term I'd like to make healthz check timestamp in iptables mode while always returns "ok" in other modes. In this way all proxy modes would be compatible. We could re-visit the healthcheck logic later if needed :)

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
```diff
@@ -0,0 +1,59 @@
...
+- We already have a similar mechanism for healthchecking OnlyLocal services
+in kube-proxy.
+
+We should enable health check by default for non-OnlyLocal LoadBalancer
+service but provide a way for opting out initially.
+
+## Implementation
+
+API
+- Define an annotation on service for opting out load-balancer health check.
```
thockin: why?

MrHohn: @bowei Or maybe we won't need this? I couldn't think of a strong reason now...

MrHohn: Removed opt-out annotation.

contributors/design-proposals/gce-l4-loadbalancer-healthcheck.md
```diff
@@ -0,0 +1,58 @@
...
+
+GCE cloud provider (through kube-controller-manager)
+- Manage `k8s-l4-healthcheck` firewall and healthcheck resources.
+These two resources should be shared among all non-OnlyLocal LoadBalancer
+services.
+- Add a new flag to pipe in the healthz port num as it is configurable on
+kube-proxy.
+
+Version skew:
+- Running higher version master (with L4 health check feature enabled) with
```
MrHohn: I'm now leaning toward create no health check if there is any lower version nodes running (without kube-proxy exposing healthz) instead of depending on Network loadbalancer's behavior --- "If all health checks are failing, traffic goes to all VMs".

One scenario I thought about is a cluster running one newer version node with many older version nodes. If we do create the health check, all traffic will go to that new node.

thockin:
Status on this?  Are you done updating?  Idea LGTM

MrHohn:
> Status on this? Are you done updating? Idea LGTM

@thockin I don't have other things to add. Squashed commits.

PR #553, opened 2017-04-19
Fix yaml links in flexvolume document

ivan4th:
Updated the links to point to k8s master branch - is this ok or should I use e.g. release-1.6 branch instead?


gyliu513:
Thanks @ivan4th , I think it is good to use master branch here, we already have some reference as https://raw.githubusercontent.com/ivan4th/community/9acf7fc5b6a3f029c67dba7d36f984de2ae4733e/sig-storage/contributing.md

/assign

PR #559, opened 2017-04-21
Add Throughput SLO doc

gmarek:
cc @kubernetes/sig-scalability-pr-reviews


gmarek:
Ping @timothysc 
cc @jeremyeder

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
+# SLO: Kubernetes cluster of size at least X is able to start Y Pods in Z minutes
+
+## Burst Pod Startup Throughput SLO
+### Summary
+Kubernetes cluster of size at least X is able to start Y Pods in Z minutes.
+### User Stories
+- User is running a workload of X total pods and wants to ensure that it can be started in Y time. 
+- User wants to understand how quickly they can react to a dramatic change in workload profile - e.g. they need to begin scheduling new pods when QPS meets W% of max to accommodate a spike due to a Black Friday Sale.
```
jeremyeder: What is W%?  What will the tests choose for this value?  How is QPS metric being fed to scheduler?

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
...
+- User is running a huge serving app on a huge cluster. He wants to know how quickly he can recreate his whole setup in case of a serious disaster which will bring the whole cluster down.
+
+Current steady state SLOs are do not provide enough data to make these assessments about burst behavior. 
+## SLO definition (full)
+### Test setup
+Standard performance test kubernetes setup.
+### Test scenario is following:
+- Start with a healthy (all nodes ready, all cluster addons already running) cluster with N running pause Pods/Node
```
jeremyeder: What is the value of N?  What will be the cluster size used by this test?

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
...
+- Create smallest number of Deployments necessary to create X Pods, with one Deployment running at most 5000 Pods.
+- Each Deployment runs in it's own Namespace
+- Each Deployment is covered by a single Service
+- Each Pod in any Deployment contains two pause containers, one secret other than ServiceAccount and one ConfigMap
```
jeremyeder: Assuming that all necessary images will be pre-pulled to all nodes.

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
...
+- Measure the time between starting the test and moment when watch saw all Pods Ready.
```
jeremyeder: Ready for pause pods is different than ready for real applications.  Just a note/clarification, I guess.

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
...
+### Definition
+Kubernetes cluster of size at least X adhering to the environment definition, when running the specified test, 99th percentile of time necessary to start Y pods from the time when user created all controllers to the time when user-side watch observes all Pod as Ready is no greater than Z minutes, assuming that all images are already present on all Nodes.
```
jeremyeder: Does kube provide a method/best practice for pre-pulling images?

sig-scalability/throughput_burts_slo.md
jeremyeder: I'd like to know if the value of N is zero or not.  Also, more complicated pods would affect this test; those that have resource limits that require the scheduler to do a little more work in fitting, use different scheduler features, etcetera.

jeremyeder:
Hi @gmarek -- took a read through.  Overall it is concise and mostly clear.  Just a few points I'd like to understand more.

sig-scalability/throughput_burts_slo.md
gmarek: It probably should be Z - I was just running out of variable names. This is just a user story for why throughput is important, it assumes that user knows what the max throughput of his service is, and  wants to increase it quickly.

sig-scalability/throughput_burts_slo.md
gmarek: N is >0, but I don't want to use any advanced scheduling features (requests will be set of course, as I don't treat it as 'advanced feature'). All concrete numbers are to be filled as soon as we know what kind of values are feasible.

sig-scalability/throughput_burts_slo.md
gmarek: Pause containers have this property (kubernetes/pause is always present on the machine)

sig-scalability/throughput_burts_slo.md
gmarek: True. I added clarification.

sig-scalability/throughput_burts_slo.md
gmarek: Not that I know of, but I saw this question few times. I guess it's the question for the node sig.

sig-scalability/throughput_burts_slo.md
jeremyeder: What will the tests choose for this value? How is QPS metric being fed to scheduler?

gmarek: They won't - it's completely unrelated.

gmarek:
@wojtek-t @timothysc ping

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
...
+- User wants to understand how quickly they can react to a dramatic change in workload profile - e.g. they need to begin scheduling new pods when QPS meets Z% of maximum throughput of users system - to accommodate a spike due to a Black Friday Sale.
+- User is running a huge serving app on a huge cluster. He wants to know how quickly he can recreate his whole setup in case of a serious disaster which will bring the whole cluster down.
+
+Current steady state SLOs are do not provide enough data to make these assessments about burst behavior. 
+## SLO definition (full)
+### Test setup
+Standard performance test kubernetes setup.
+### Test scenario is following:
+- Start with a healthy (all nodes ready, all cluster addons already running) cluster with N (>0) running pause Pods/Node.
+- Create smallest number of Deployments necessary to create X Pods, with one Deployment running at most 5000 Pods.
```
wojtek-t: I'm not convinced about this. I would say that we want both large and small deployments, and most probably more than one per namespace.
Maybe something like:
- we would like to have 5000 pods / namespace split across multiple deployments (from large ones (5000-pod one) to small ones (1? 5?)).

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
+# SLO: Kubernetes cluster of size at least X is able to start Y Pods in Z minutes
+
+## Burst Pod Startup Throughput SLO
+### Summary
+Kubernetes cluster of size at least X is able to start Y Pods in Z minutes.
+### User Stories
+- User is running a workload of X total pods and wants to ensure that it can be started in Y time. 
+- User wants to understand how quickly they can react to a dramatic change in workload profile - e.g. they need to begin scheduling new pods when QPS meets Z% of maximum throughput of users system - to accommodate a spike due to a Black Friday Sale.
```
wojtek-t: I think i don't really understand this. Can you clarify?

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
+# SLO: Kubernetes cluster of size at least X is able to start Y Pods in Z minutes
+
+## Burst Pod Startup Throughput SLO
+### Summary
+Kubernetes cluster of size at least X is able to start Y Pods in Z minutes.
```
wojtek-t: I wondering about this - do we need it? It's the same as title of the doc (and I think it will be the case in most of situations?)

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
...
+### User Stories
+- User is running a workload of X total pods and wants to ensure that it can be started in Y time. 
+- User wants to understand how quickly they can react to a dramatic change in workload profile - e.g. they need to begin scheduling new pods when QPS meets Z% of maximum throughput of users system - to accommodate a spike due to a Black Friday Sale.
+- User is running a huge serving app on a huge cluster. He wants to know how quickly he can recreate his whole setup in case of a serious disaster which will bring the whole cluster down.
+
+Current steady state SLOs are do not provide enough data to make these assessments about burst behavior. 
+## SLO definition (full)
+### Test setup
+Standard performance test kubernetes setup.
```
wojtek-t: We should have a link to the doc containing this here (and we already have this doc - this is your previous one, even though there are multiple TODOs there).

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
+# SLO: Kubernetes cluster of size at least X is able to start Y Pods in Z minutes
```
wojtek-t: I vote for creating a dedicated directory for those documents, probably sig-scalability/slo/ (or sth like that).

sig-scalability/throughput_burts_slo.md
```diff
@@ -0,0 +1,23 @@
...
+
```
wojtek-t: Can we mention that this is not something we are meeting, just something we are planning to have.

sig-scalability/throughput_burts_slo.md
gmarek: +1. Done.

sig-scalability/throughput_burts_slo.md
gmarek: I think that having `Z` instead of real number means exactly that, but I added a disclaimer as well.

sig-scalability/throughput_burts_slo.md
gmarek: Maybe... I'll remove it from this one.

sig-scalability/throughput_burts_slo.md
gmarek: This was written by @timothysc:) I tried to rephrase it.

sig-scalability/throughput_burts_slo.md
gmarek: Done.

sig-scalability/throughput_burts_slo.md
gmarek: I want this to be as specific as possible. I'll make this change, but it'll make this point much longer:)

gmarek: If we want to go smaller (i.e. more Deployments) we need to think what do to with Services.

wojtek-t:
I'm fine with this - thanks!

If needed, we can iterate on it later.

/lgtm

PR #563, opened 2017-04-21
Update godep.md

mbohlool:
`godep restore` may not work because of staging folder. `hack/godep-restore.sh` is the way to go.


cblecker:
I had opened #520 to update the "updating an existing dependency" part as well as this, but I'm okay with getting this bit in there, and I can still fix up the section.
/lgtm
cc: @sttts

mbohlool:
That supersedes this, it is fine to just finish up that one.

PR #572, opened 2017-04-23
Fix broken link to strategic-merge in api-conventions.md

aslakknutsen:


aslakknutsen:
Signed. 

/rant-start
Mind you, the most ridiculous CLA process I've ever seen. Sign an online doc with a signature font of your liking? Really? How did you get any more valid info from me then what you already had 3 steps before, a name and an email.

And for what, a 3 char doc change? What possible legal ramifications could you get into based on that change...
/rant-end

PR #576, opened 2017-04-24
add the node client support for promises (async/await) and Low-level …

shenshouer:
…URL building

The nodejs client for kubernetes:

- Support for promises (async/await)
- Low-level URL building (makes it a little more future-proof)


cblecker:
This was moved to https://kubernetes.io/docs/reference/client-libraries/

Issue #544, opened 2017-04-18
SIG README formats are inconsistent and sometimes incomplete


jamiehannaford:
Would it be possible to standardise a template for all SIGs to use in their READMEs? Something like:

```
# Title 

Mission statement

## Leads

## Contact us

## Meetings 

## Other links 
```

I'd be willing to take this on but wanted some feedback first.

castrojo:
I'm wondering if a template would work? So instead of manually updating READMEs we could have one template with the appropriate metadata that could just autogenerate the READMEs for each SIG for consistency and drop them in the right directory. And then as a bonus also generate one-big-page to replace the list of SIGs (which are updated by hand) and also generate a one-line snippet for SIG leads to use as a topic in the slack channel. 

I've been in the situation before where a SIG meeting was listed differently in slack, on the SIG page, and the list-of-SIGs page, so perhaps thinking about a single source of truth for all the sig metadata and then autogenerating the rest would be worth investigating?

jamiehannaford:
@castrojo I like that idea. Where would be best to store the SIG-specific info? Perhaps a JSON, YAML or MD file, like so?

```yaml
sigs:
  sig-openstack: 
     mission-statement: >
       This is the community page of the Kubernetes OpenStack SIG, a special interest group
       coordinating the cross-community efforts of the OpenStack and Kubernetes communities. 
     leads:
     - name: Foo 
       github: @foo
     - name: Bar
       email: contact@bar.com
     meetings:
     - day: monday
       utc: 0900
     - day: wednesday
       utc: 2100
     contact:
       slack: #sig-openstack
       mailing-list: ...
     other-links:
     - google.com
     - openstack.org
```

castrojo:
I think a YAML file to match everything else makes sense, though it's probably prudent to see what other people think.  

Meetings would need `meeting_url` for the zoom hangout, and perhaps a `meeting_archive_url` for the groups that maintain youtube playlists of their meetings, looking good!

PR #577, opened 2017-04-24
Add automation tooling to generate SIG documentation

jamiehannaford:
As discussed in #544, it'd be nice to have consistently formatted and comprehensive documentation for our SIGs. Right now each group has to manually update or create their own content, which often leads to missing sections, duplication and an unnecessary maintenance burden. 

This PR introduces a single metadata YAML file which structures SIG data in a consistent way. The small Go app in the `build` folder generates READMEs based off this metadata, and also updates the `sig_list.md` in the project root. I tried to use existing software to do this (mustache CLI, jq and pandoc), but trying to integrate them soon became unworkable when compared to a small Go binary that could do the same thing. 

We could at a later date incorporate this into a CI hook which either tests the user's changeset against what's expected, or automatically builds the SIG docs per PR.


jamiehannaford:
cc/ @castrojo

castrojo:
Thanks for the PR Jamie, @jbeda can you take a look?

build/README.md
```diff
@@ -0,0 +1,19 @@
...
+```
+
+Based off the `sigs.yaml` metadata file.
+
+## How to use 
+
+To (re)build documentation for all the SIGs, run these commands:
+
+```bash
+docker build -t sigdocs -c build/Dockerfile build
```
kensimon: did you mean `-f build/Dockerfile`?  `-c` is cpu-shares.

build/README.md
```diff
@@ -0,0 +1,19 @@
...
+docker run -v $(pwd):/go/src/app sigdocs
```
kensimon: This is a no-op, did you want `docker run -v $(pwd):/go/src/app sigdocs go run build/app.go`?

build/README.md
```diff
@@ -0,0 +1,19 @@
+# SIG Doc builder
+
+This script will generate the following documentation files:
+
+```
+sig-*/README.md
+sig_list.md
```
kensimon: it's `sig-list.md` in the repo right now, not `sig_list.md`

build/app.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/cbroglie/mustache"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	sigsYamlFile   = "sigs.yaml"
+	templateDir    = "build"
+	indexTemplate  = fmt.Sprintf("%s/sig_index.mustache", templateDir)
+	listTemplate   = fmt.Sprintf("%s/sig_list.mustache", templateDir)
+	sigListOutput  = "sig_list.md"
```
kensimon: `sig-list.md`?

build/README.md
jamiehannaford: My understanding was that the `golang:1.6-onbuild` image already does this. From [the docs](https://hub.docker.com/_/golang/):

> This image includes multiple ONBUILD triggers which should cover most applications. The build will COPY . /go/src/app, RUN go get -d -v, and RUN go install -v. This image also includes the CMD ["app"] instruction which is the default command when running the image without arguments.

jamiehannaford:
@kensimon Thanks for the review. Per @ahmetb's feedback I added a few more features:

- a footer which indicates last generation date
- a comment discouraging users from editing the generated files
- a `Makefile` to unify the commands

build/README.md
kensimon: Looks like it works for me now. I thought I had tried it last time and saw that it didn't, but maybe that's because the build didn't work properly.

build/app.go
```diff
@@ -0,0 +1,182 @@
...
+	Day       string
+	UTC       string
+	PST       string
+	Frequency string
+}
+
+type Contact struct {
+	Slack       string
+	MailingList string `yaml:"mailing_list"`
+	GitHubTeam  string `yaml:"github_team"`
```
jbeda: Github team has gotten more complicated.  See https://github.com/kubernetes/community/blob/master/sig-creation-procedure.md.  Not every SIG has done this dance yet though.

Idea: have a bool ("FullGitHubTeams") and if that is try then we can expand out the team according to the link above.  If false, just don't list any github teams.

build/app.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+type Sig struct {
```
jbeda: Consider: Much of this should/could be consistently derived from the sig name.

build/app.go
```diff
@@ -0,0 +1,182 @@
...
+			return err
+		}
+		if !exists {
+			err = os.Mkdir(sig.Dir, 0755)
+			if err != nil {
+				return err
+			}
+		}
+
+		footer, err := constructFooter()
```
jbeda: Does it make sense to put the footer in the template vs. do the concat here?  Just make it a sub template?  Don't know how hard that is to do with mustache but it is easy with golang templates.

build/app.go
```diff
@@ -0,0 +1,182 @@
...
+		if err != nil {
+			return err
+		}
+
+		filePath := fmt.Sprintf("%s/%s", sig.Dir, sigIndexOutput)
+		err = ioutil.WriteFile(filePath, []byte(data+footer), 0644)
```
jbeda: feature request:  Only generate a section of the README.md.  Perhaps from the top of the file along with metadata.  Let users hand edit stuff below that.  That would require some comment guards so we can identify what parts are generated and which aren't.

build/app.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func main() {
+	yamlData, err := ioutil.ReadFile(sigsYamlFile)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var ctx SigEntries
+	err = yaml.Unmarshal(yamlData, &ctx)
```
jbeda: feature request/TODO: sort based on sig name here.

build/app.go
```diff
@@ -0,0 +1,182 @@
...
+	MailingList string `yaml:"mailing_list"`
+	GitHubTeam  string `yaml:"github_team"`
+}
+
+type Sig struct {
+	Name              string
+	Dir               string
+	MissionStatement  string `yaml:"mission_statement"`
+	Leads             []Lead
+	Meetings          []Meeting
```
jbeda: Not sure why this is an array.  Are we using it as such anywhere?

build/app.go
jamiehannaford: Sounds like the best way, but I can't find any SIG except `sig-openstack` which does this as expected. The only other one is `@sig-onprem-*` which doesn't match its `.Dir` value (sig-on-prem) or name (On Premise). 

How about we use a `GithubPrefix` or similar field to indicate what the team prefix will be? The default value would be the SIG name in kebab case. Logic could be:

```golang
if sig.Contact.FullGitHubTeams {
   prefix := sig.Contact.GithubPrefix
   if prefix == "" {
      // default is kebab cased name
      prefix = strings.ToLower(strings.Replace(sig.Name, " ", "-", -1))
   }
   teamNames := []string{"misc", "test-failures", "bugs", "feature-requests", "proposals", "pr-reviews", "api-reviews"}
   for _, teamName := range teamNames {
      sig.Contact.GithubTeams = append(sig.Contact.GithubTeams, fmt.Sprintf("%s-%s", prefix, teamName)
   }
}

File: build/app.go (hunk 6 above)
Comment (jamiehannaford) [inline build/app.go]: Agreed. What other fields except `.Dir` can be removed do you think?

File: build/app.go (hunk 10 above)
Comment (jamiehannaford) [inline build/app.go]: `sig-scheduling` AFAIK has alternating meetings: biweekly Mondays 20:00 UTC and Wednesdays 07:30 UTC. I'm not sure how else we can express that than as a list of discrete meetings.

**Comment by jamiehannaford** [comment]
@jbeda Thanks for the great feedback. I've incorporated most of your suggestions (text/template, ordering, Github Teams etc.) and left comments where necessary.

In terms of the feature request to only generate some of the README.md files, I think it's a great idea, but can I implement in a follow-up PR? It's quite tricky to get right and I don't want to distract from the main focus of this PR.

**Comment by bgrant0607** [comment]
cc @sarahnovotny @philips @smarterclayton @brendandburns @thockin @jbeda

File: generator/header.tmpl [hunk 11]
```diff
@@ -0,0 +1,10 @@
+{{- define "header" -}}
+<!---
+This is an autogenerated file!
+
+Please do not edit this file directly, but instead make changes to the
+sigs.yaml file in the project root.
+
+To understand how this file is generated, see build/README.md.
```
jbeda: Nit: update this to `generator/README.md`

generator/sig_index.tmpl
```diff
@@ -0,0 +1,27 @@
+{{- template "header" }}
+# {{.Name}} SIG
+
+{{ .MissionStatement }}
+## Meetings
+{{- range .Meetings }}
+* [{{.Day}}s at {{.UTC}} UTC]({{$.MeetingURL}}) ({{.Frequency}})
```
jbeda: Suggestion: time zones are hard.  Suggestion: perhaps link the time to some handy service like http://www.thetimezoneconverter.com/event_link.html.  I just pulled that one off of Google but it looks nice enough.

jamiehannaford: Great idea 👍

jamiehannaford:
@jbeda Thanks again for the great review, I've added those two things

jbeda:
LGTM!  Merging now.  I think the next step will be to send mail to kubernetes-dev and then regen the pages and check them in.

generator/README.md
```diff
@@ -0,0 +1,18 @@
...
+This script will generate the following documentation files:
+
+```
+sig-*/README.md
+sig-list.md
+```
+
+Based off the `sigs.yaml` metadata file.
+
+## How to use
```
idvoretskyi: I would request having more detailed documentation on how to use this tool (together with extended description).

jamiehannaford: @idvoretskyi Good idea. I'll submit a PR now.

PR #579, opened 2017-04-24
Fix broken Analytics in generating-clientset

spzala:
The current doc page for generating-clientset shows a broken Analytics link.
Remove duplicated and unformatted text and fix it.


sarahnovotny:
thanks!

PR #583, opened 2017-04-26
Backoff policy and failed pod limit

soltysh:
This update addresses problems raised in https://github.com/kubernetes/kubernetes/issues/30243 and https://github.com/kubernetes/kubernetes/issues/43964.

@erictune I've mentioned this to you during last sig-apps, ptal
@kubernetes/sig-apps-feature-requests ptal
@lukasheinrich @nickschuch @Yancey1989 @sstarcher @maximz fyi since you all were interested in this

### DIFF
--- a/contributors/design-proposals/job.md
+++ b/contributors/design-proposals/job.md
@@ -18,12 +18,38 @@ Several existing issues and PRs were already created regarding that particular s
 1. Be able to get the job status.
 1. Be able to specify the number of instances performing a job at any one time.
 1. Be able to specify the number of successfully finished instances required to finish a job.
+1. Be able to specify backoff policy, when job is continuously failing.
 
 
 ## Motivation
 
 Jobs are needed for executing multi-pod computation to completion; a good example
 here would be the ability to implement any type of batch oriented tasks.
+
+
+## Backoff policy and failed pod limit
+
+By design, Jobs do not have any notion of failure, other than Pod's `restartPolicy`
+which is mistakenly taken as Job's restart policy ([#30243](https://github.com/kubernetes/kubernetes/issues/30243),
+[#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
+situation where one wants to fail a Job after some amount of retries over certain
+period of time, due to a logical error in configuration etc.  To do so we are going
+following fields will be introduced, which will control the exponential backoff
+when retrying Job: number of retries and time to retry.  The two fields will allow
+creating a fine grain control over the backoff policy, limiting the number of retries
+over specified period of time.  In the case when only one of them is specified
+an exponential backoff with duration of 10 seconds and factor of 2 will be applied
+in such a way that either time or number is reached.  After reaching the limit
+a Job will be marked as failed.
+
+Additionally, to help debug the issue with a job, and limit the impact of having
+too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243))
+we are going to introduce a field which will allow specifying the maximum number
+of failed pods to keep around.  This number will also take effect if none of the
+limits, described above, are set.
+
+All of the above fields will be optional and will apply no matter which `restartPolicy`
+is set on a `PodTemplate`.
 
 
 ## Implementation
@@ -82,6 +108,15 @@ type JobSpec struct {
     // Optional duration in seconds relative to the startTime that the job may be active
     // before the system tries to terminate it; value must be positive integer
     ActiveDeadlineSeconds *int
+
+    // Optional number of retries, before marking this job failed.
+    BackoffLimit *int
+
+    // Optional time (in seconds), how log a job should be retried before marking it failed.
+    BackoffDeadlineSeconds *int
+
+    // Optional number of failed pods to retain.
+    FailedPodsLimit *int
 
     // Selector is a label query over pods running a job.
     Selector LabelSelector
--- a/contributors/design-proposals/job.md
+++ b/contributors/design-proposals/job.md
@@ -79,8 +79,12 @@ type JobSpec struct {
     // job should be run with. Defaults to 1.
     Completions *int
 
+    // Optional duration in seconds relative to the startTime that the job may be active
+    // before the system tries to terminate it; value must be positive integer
+    ActiveDeadlineSeconds *int
+
     // Selector is a label query over pods running a job.
-    Selector map[string]string
+    Selector LabelSelector
 
     // Template is the object that describes the pod that will be created when
     // executing a job.
@@ -109,12 +113,12 @@ type JobStatus struct {
     // Active is the number of actively running pods.
     Active int
 
-    // Successful is the number of pods successfully completed their job.
-    Successful int
+    // Succeeded is the number of pods successfully completed their job.
+    Succeeded int
 
-    // Unsuccessful is the number of pods failures, this applies only to jobs
+    // Failed is the number of pods failures, this applies only to jobs
     // created with RestartPolicyNever, otherwise this value will always be 0.
-    Unsuccessful int
+    Failed int
 }
 
 type JobConditionType string
--- a/contributors/design-proposals/job.md
+++ b/contributors/design-proposals/job.md
@@ -34,14 +34,12 @@ which is mistakenly taken as Job's restart policy ([#30243](https://github.com/k
 [#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
 situation where one wants to fail a Job after some amount of retries over a certain
 period of time, due to a logical error in configuration etc.  To do so we are going
-to introduce following fields, which will control the backoff policy: a number of
-retries and a time to retry (counted from the first failure).  The two fields will
-allow fine-grained control over the backoff policy, limiting the number of retries
-over a specified period of time.  If only one of the two fields is supplied,
-a backoff with an intervening duration of ten seconds and a factor of two will be
-applied, such that either:
-* the number of retries will not exceed a specified count, if present, or
-* the maximum time elapsed will not exceed the specified duration, if present.
+to introduce the following fields, which will control the backoff policy: a number of
+retries and an initial time of retry.  The two fields will allow fine-grained control
+over the backoff policy.  Each of the two fields will use a default value if none
+is provided,  `BackoffLimit` is set by default to 6 and `BackoffSeconds` to 10s.
+This will result in the following retry sequence: 10s, 20s, 40s, 1m20s, 2m40s,
+5m20s.  After which the job will be considered failed.
 
 Additionally, to help debug the issue with a Job, and limit the impact of having
 too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243)),
@@ -56,6 +54,9 @@ is set on a `PodTemplate`.  The only difference applies to how failures are coun
 For restart policy `Never` we count actual pod failures (reflected in `.status.failed`
 field). With restart policy `OnFailure` we take an approximate value of pod restarts
 (as reported in `.status.containerStatuses[*].restartCount`).
+When `.spec.parallelism` is set to a value higher than 1, the failures are an
+overall number (as coming from `.status.failed`) because the controller does not
+hold information about failures coming from separate pods.
 
 
 ## Implementation
@@ -118,10 +119,12 @@ type JobSpec struct {
     ActiveDeadlineSeconds *int64
 
     // Optional number of retries before marking this job failed.
+    // Defaults to 6.
     BackoffLimit *int32
 
-    // Optional time (in seconds) specifying how long a job should be retried before marking it failed.
-    BackoffDeadlineSeconds *int64
+    // Optional time (in seconds) specifying how long the initial backoff will last.
+    // Defaults to 10s.
+    BackoffSeconds *int64
 
     // Optional number of failed pods to retain.
     FailedPodsLimit *int32
--- a/contributors/design-proposals/job.md
+++ b/contributors/design-proposals/job.md
@@ -34,11 +34,11 @@ which is mistakenly taken as Job's restart policy ([#30243](https://github.com/k
 [#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
 situation where one wants to fail a Job after some amount of retries over a certain
 period of time, due to a logical error in configuration etc.  To do so we are going
-to introduce following fields, which will control the exponential backoff when
-retrying a Job: number of retries and time to retry.  The two fields will allow
-fine-grained control over the backoff policy, limiting the number of retries over
-a specified period of time.  If only one of the two fields is supplied, an exponential
-backoff with an intervening duration of ten seconds and a factor of two will be
+to introduce following fields, which will control the backoff policy: a number of
+retries and a time to retry (counted from the first failure).  The two fields will
+allow fine-grained control over the backoff policy, limiting the number of retries
+over a specified period of time.  If only one of the two fields is supplied,
+a backoff with an intervening duration of ten seconds and a factor of two will be
 applied, such that either:
 * the number of retries will not exceed a specified count, if present, or
 * the maximum time elapsed will not exceed the specified duration, if present.
@@ -47,13 +47,15 @@ Additionally, to help debug the issue with a Job, and limit the impact of having
 too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243)),
 we are going to introduce a field which will allow specifying the maximum number
 of failed pods to keep around.  This number will also take effect if none of the
-limits described above are set.
+limits described above are set. By default it will take value of 1, to allow debugging
+job issues, but not to flood the cluster with too many failed jobs and their
+accompanying pods.
 
 All of the above fields will be optional and will apply no matter which `restartPolicy`
 is set on a `PodTemplate`.  The only difference applies to how failures are counted.
 For restart policy `Never` we count actual pod failures (reflected in `.status.failed`
-field). With restart policy `OnFailure` we look at pod restarts (calculated from
-`.status.containerStatuses[*].restartCount`).
+field). With restart policy `OnFailure` we take an approximate value of pod restarts
+(as reported in `.status.containerStatuses[*].restartCount`).
 
 
 ## Implementation
@@ -103,24 +105,26 @@ type JobSpec struct {
     // run at any given time. The actual number of pods running in steady state will
     // be less than this number when ((.spec.completions - .status.successful) < .spec.parallelism),
     // i.e. when the work left to do is less than max parallelism.
-    Parallelism *int
+    Parallelism *int32
 
     // Completions specifies the desired number of successfully finished pods the
     // job should be run with. Defaults to 1.
-    Completions *int
+    Completions *int32
 
     // Optional duration in seconds relative to the startTime that the job may be active
     // before the system tries to terminate it; value must be a positive integer.
-    ActiveDeadlineSeconds *int
+    // It applies to overall job run time, no matter of the value of completions
+    // or parallelism parameters.
+    ActiveDeadlineSeconds *int64
 
     // Optional number of retries before marking this job failed.
-    BackoffLimit *int
+    BackoffLimit *int32
 
     // Optional time (in seconds) specifying how long a job should be retried before marking it failed.
-    BackoffDeadlineSeconds *int
+    BackoffDeadlineSeconds *int64
 
     // Optional number of failed pods to retain.
-    FailedPodsLimit *int
+    FailedPodsLimit *int32
 
     // Selector is a label query over pods running a job.
     Selector LabelSelector
@@ -150,14 +154,14 @@ type JobStatus struct {
     CompletionTime unversioned.Time
 
     // Active is the number of actively running pods.
-    Active int
+    Active int32
 
     // Succeeded is the number of pods successfully completed their job.
-    Succeeded int
+    Succeeded int32
 
     // Failed is the number of pods failures, this applies only to jobs
     // created with RestartPolicyNever, otherwise this value will always be 0.
-    Failed int
+    Failed int32
 }
 
 type JobConditionType string
--- a/contributors/design-proposals/job.md
+++ b/contributors/design-proposals/job.md
@@ -49,11 +49,14 @@ limits described above are set. By default it will take value of 1, to allow deb
 job issues, but not to flood the cluster with too many failed jobs and their
 accompanying pods.
 
-All of the above fields will be optional and will apply no matter which `restartPolicy`
-is set on a `PodTemplate`.  The only difference applies to how failures are counted.
-For restart policy `Never` we count actual pod failures (reflected in `.status.failed`
-field). With restart policy `OnFailure` we take an approximate value of pod restarts
-(as reported in `.status.containerStatuses[*].restartCount`).
+All of the above fields will be optional and will apply when `restartPolicy` is
+set to `Never` on a `PodTemplate`.  With restart policy `OnFailure` only `BackoffLimit`
+applies. The reason for that is that failed pods are already restarted by the
+kubelet with an [exponential backoff](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy).
+Additionally, failures are counted differently depending on `restartPolicy`
+setting.  For `Never` we count actual pod failures (reflected in `.status.failed`
+field). With `OnFailure`, we take an approximate value of pod restarts (as reported
+in `.status.containerStatuses[*].restartCount`).
 When `.spec.parallelism` is set to a value higher than 1, the failures are an
 overall number (as coming from `.status.failed`) because the controller does not
 hold information about failures coming from separate pods.
--- a/contributors/design-proposals/job.md
+++ b/contributors/design-proposals/job.md
@@ -34,12 +34,12 @@ which is mistakenly taken as Job's restart policy ([#30243](https://github.com/k
 [#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
 situation where one wants to fail a Job after some amount of retries over a certain
 period of time, due to a logical error in configuration etc.  To do so we are going
-to introduce following fields will be introduced, which will control the exponential
-backoff when retrying a Job: number of retries and time to retry.  The two fields
-will allow creating a fine-grained control over the backoff policy, limiting the
-number of retries over a specified period of time.  If only one of the two fields
-is supplied, an exponential backoff with an intervening duration of ten seconds
-and a factor of two will be applied, such that either:
+to introduce following fields, which will control the exponential backoff when
+retrying a Job: number of retries and time to retry.  The two fields will allow
+fine-grained control over the backoff policy, limiting the number of retries over
+a specified period of time.  If only one of the two fields is supplied, an exponential
+backoff with an intervening duration of ten seconds and a factor of two will be
+applied, such that either:
 * the number of retries will not exceed a specified count, if present, or
 * the maximum time elapsed will not exceed the specified duration, if present.
 
--- a/contributors/design-proposals/job.md
+++ b/contributors/design-proposals/job.md
@@ -18,7 +18,7 @@ Several existing issues and PRs were already created regarding that particular s
 1. Be able to get the job status.
 1. Be able to specify the number of instances performing a job at any one time.
 1. Be able to specify the number of successfully finished instances required to finish a job.
-1. Be able to specify backoff policy, when job is continuously failing.
+1. Be able to specify a backoff policy, when job is continuously failing.
 
 
 ## Motivation
@@ -29,27 +29,31 @@ here would be the ability to implement any type of batch oriented tasks.
 
 ## Backoff policy and failed pod limit
 
-By design, Jobs do not have any notion of failure, other than Pod's `restartPolicy`
+By design, Jobs do not have any notion of failure, other than a pod's `restartPolicy`
 which is mistakenly taken as Job's restart policy ([#30243](https://github.com/kubernetes/kubernetes/issues/30243),
 [#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
-situation where one wants to fail a Job after some amount of retries over certain
+situation where one wants to fail a Job after some amount of retries over a certain
 period of time, due to a logical error in configuration etc.  To do so we are going
-following fields will be introduced, which will control the exponential backoff
-when retrying Job: number of retries and time to retry.  The two fields will allow
-creating a fine grain control over the backoff policy, limiting the number of retries
-over specified period of time.  In the case when only one of them is specified
-an exponential backoff with duration of 10 seconds and factor of 2 will be applied
-in such a way that either time or number is reached.  After reaching the limit
-a Job will be marked as failed.
+to introduce following fields will be introduced, which will control the exponential
+backoff when retrying a Job: number of retries and time to retry.  The two fields
+will allow creating a fine-grained control over the backoff policy, limiting the
+number of retries over a specified period of time.  If only one of the two fields
+is supplied, an exponential backoff with an intervening duration of ten seconds
+and a factor of two will be applied, such that either:
+* the number of retries will not exceed a specified count, if present, or
+* the maximum time elapsed will not exceed the specified duration, if present.
 
-Additionally, to help debug the issue with a job, and limit the impact of having
-too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243))
+Additionally, to help debug the issue with a Job, and limit the impact of having
+too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243)),
 we are going to introduce a field which will allow specifying the maximum number
 of failed pods to keep around.  This number will also take effect if none of the
-limits, described above, are set.
+limits described above are set.
 
 All of the above fields will be optional and will apply no matter which `restartPolicy`
-is set on a `PodTemplate`.
+is set on a `PodTemplate`.  The only difference applies to how failures are counted.
+For restart policy `Never` we count actual pod failures (reflected in `.status.failed`
+field). With restart policy `OnFailure` we look at pod restarts (calculated from
+`.status.containerStatuses[*].restartCount`).
 
 
 ## Implementation
@@ -106,13 +110,13 @@ type JobSpec struct {
     Completions *int
 
     // Optional duration in seconds relative to the startTime that the job may be active
-    // before the system tries to terminate it; value must be positive integer
+    // before the system tries to terminate it; value must be a positive integer.
     ActiveDeadlineSeconds *int
 
-    // Optional number of retries, before marking this job failed.
+    // Optional number of retries before marking this job failed.
     BackoffLimit *int
 
-    // Optional time (in seconds), how log a job should be retried before marking it failed.
+    // Optional time (in seconds) specifying how long a job should be retried before marking it failed.
     BackoffDeadlineSeconds *int
 
     // Optional number of failed pods to retain.


contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+a Job will be marked as failed.
+
+Additionally, to help debug the issue with a job, and limit the impact of having
+too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243))
+we are going to introduce a field which will allow specifying the maximum number
+of failed pods to keep around.  This number will also take effect if none of the
+limits, described above, are set.
+
+All of the above fields will be optional and will apply no matter which `restartPolicy`
+is set on a `PodTemplate`.
```
soltysh: I was struggling with this idea, but I think we have necessary information when failure happens in both cases. When `Never` we count failures ourselves, with `OnFailure` we can rely on no of restarts. Alternatively, I was thinking to address that only for `Never`, but I want to hear from others.

contributors/design-proposals/job.md
```diff
@@ -79,8 +105,21 @@ type JobSpec struct {
...
     Completions *int
 
+    // Optional duration in seconds relative to the startTime that the job may be active
+    // before the system tries to terminate it; value must be positive integer
+    ActiveDeadlineSeconds *int
+
+    // Optional number of retries, before marking this job failed.
+    BackoffLimit *int
+
+    // Optional time (in seconds), how log a job should be retried before marking it failed.
```
nickschuch: long*

contributors/design-proposals/job.md
diegodelemos: For our use case it would be useful to rely on `OnFailure` policy and then counting number of restarts (this is in fact what we are doing from outside, watching status over stream connections). Since we aim to submit large amounts of `jobs`, it would be better to have only one container retrying rather that having a new `pod` per failure spawned by the `job controller`. Unless you think this is not going to end up creating performance issues.

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
 
+## Backoff policy and failed pod limit
+
+By design, Jobs do not have any notion of failure, other than Pod's `restartPolicy`
+which is mistakenly taken as Job's restart policy ([#30243](https://github.com/kubernetes/kubernetes/issues/30243),
+[#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
+situation where one wants to fail a Job after some amount of retries over certain
+period of time, due to a logical error in configuration etc.  To do so we are going
+following fields will be introduced, which will control the exponential backoff
+when retrying Job: number of retries and time to retry.  The two fields will allow
```
seh: - [ ] s/retrying Job/retrying a Job/  
  or maybe s/retying Job/retrying Jobs/

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+creating a fine grain control over the backoff policy, limiting the number of retries
```
seh: - [ ] s/a fine grain/fine-grained/

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+over specified period of time.  In the case when only one of them is specified
```
seh: - [ ] s/over specified/over a specified/

seh: - [ ] s/is specified/is specified,/

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+an exponential backoff with duration of 10 seconds and factor of 2 will be applied
+in such a way that either time or number is reached.  After reaching the limit
```
seh: - [ ] Can you clarify what you mean by, "either time or number is reached?"

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+a Job will be marked as failed.
+
+Additionally, to help debug the issue with a job, and limit the impact of having
```
seh: - [ ] s/job/Job/

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243))
```
seh: - [ ] s/))/)),/

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+we are going to introduce a field which will allow specifying the maximum number
+of failed pods to keep around.  This number will also take effect if none of the
+limits, described above, are set.
```
seh: - [ ] s/, described above, /—described above—/
  or s/, described above, / described above /

contributors/design-proposals/job.md
```diff
@@ -79,8 +105,21 @@ type JobSpec struct {
     // job should be run with. Defaults to 1.
     Completions *int
 
+    // Optional duration in seconds relative to the startTime that the job may be active
+    // before the system tries to terminate it; value must be positive integer
```
seh: - [ ] s/be positive integer/be a positive integer./  
- [ ] Did you consider using type `*uint` instead?  
  That still requires an admonishment about a nonzero value.

contributors/design-proposals/job.md
```diff
@@ -79,8 +105,21 @@ type JobSpec struct {
...
+    ActiveDeadlineSeconds *int
+
+    // Optional number of retries, before marking this job failed.
```
seh: - [ ] s /retries,/retries/

contributors/design-proposals/job.md
seh: - [ ] s/), how/) specifying how/
- [ ] Did you consider using type `*uint` instead?
  That could encourage unintentional overflow when used with [`time`](https://golang.org/pkg/time/).

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
 here would be the ability to implement any type of batch oriented tasks.
 
 
+## Backoff policy and failed pod limit
+
+By design, Jobs do not have any notion of failure, other than Pod's `restartPolicy`
```
seh: - [ ] s/than Pod's/than a pod's/

contributors/design-proposals/job.md
```diff
@@ -18,6 +18,7 @@ Several existing issues and PRs were already created regarding that particular s
 1. Be able to get the job status.
 1. Be able to specify the number of instances performing a job at any one time.
 1. Be able to specify the number of successfully finished instances required to finish a job.
+1. Be able to specify backoff policy, when job is continuously failing.
```
seh: - [ ] s/specify backoff/specify a backoff/

contributors/design-proposals/job.md
seh: I interpret the _Job_ restart limit to be predicated on the number of pod failures. If a pod has a restart policy of `OnFailure`, by my reading of [the documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase), a pod in which its sole container terminates in failure means the whole pod transitions to the "Failed" state, perhaps only momentarily, even if the container gets restarted and the pod transitions back to the "Running" state. However, [the "Example states" section](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#example-states) at the end of the same document never mentions such a transition.

That means that to count the number of attempts for a pod template with restart policy of `Never`, we'd have to count the number of pods created for the _Job_. For a policy of `OnFailure`, that count is the number of pods created (assuming it's possible for a _Job_ to lose a pod and have to schedule a replacement) plus the number of container restarts. (Each pod creation implies each container starting once.)

However, for a pod with multiple containers, we're then treating each failure of a container to imply one failure of the containing pod. That's confusing to reason about. The alternative is to say that a pod with a restart policy of `OnFailure` only participates in this accounting when the entire pod fails, such as due to the disk failing or the hosting node is removed from the cluster.

soltysh:
> All this makes perfect sense to me. Would you like me to start demonstrating this in my PR #41451?

@nickschuch let's wait for this to get to a stable shape (more or less) and then you can go ahead an open a fresh one for it. I'll close the old one.

contributors/design-proposals/job.md
soltysh: The reason I went with both actually is that I don't want to connect pod restart policy with job backoff policy, in the first place. Additionally, if pod's restart policy is never the kubelet will keep on creating new pods along the old ones (that's the job contract), which in turns the `FailedPodsLimit` can address. 
Currently `Never` is the only one that returns the pod failure back to job controller which creates a new pod to satisfy completions. With `OnFailure` we're passing that responsibility to kubelet. After talking to @kargakis I'm starting to think we should've restricted jobs to only use `Never`, but it's too late now because it would break to many folks. That's why I'd go with the backoff policy not being connected with pod's restart policy to limit the amount of confusion we already caused allowing both restart policies. Does that make sense?

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
 here would be the ability to implement any type of batch oriented tasks.
 
 
+## Backoff policy and failed pod limit
+
+By design, Jobs do not have any notion of failure, other than Pod's `restartPolicy`
+which is mistakenly taken as Job's restart policy ([#30243](https://github.com/kubernetes/kubernetes/issues/30243),
+[#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
+situation where one wants to fail a Job after some amount of retries over certain
```
soltysh: a certain period, will fix.

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,31 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+period of time, due to a logical error in configuration etc.  To do so we are going
+following fields will be introduced, which will control the exponential backoff
```
soltysh: Yup ;)

contributors/design-proposals/job.md
soltysh: When creating exponential backoff with only number, I'll use the number as how many retries should be performed (using 10s as the time between retries, with factor of 2) see [Backoff and ExponentialBackoff](https://github.com/kubernetes/kubernetes/blob/e917f4ed39967e02edcd64d0574f22edc4d7e76f/staging/src/k8s.io/apimachinery/pkg/util/wait/wait.go#L139-L172) for more info about these. In the case with time, I'll use the time as the max duration of the backoff. Does it make sense now?

contributors/design-proposals/job.md
soltysh: I went with `int`, similarly to other `*Seconds` fields, this will actually be `int64` same as `ActiveDeadlineSeconds`.

contributors/design-proposals/job.md
seh: Yes, I understand it after you explained it. Consider rewording that sentence to something like this:
> If only one of the two fields is supplied, an exponential backoff with an intervening duration of ten seconds and a factor of two will be applied, such that either
> - the number of retries will not exceed a specified count, if present, or
> - the maximum time elapsed will not exceed the specified duration, if present.

contributors/design-proposals/job.md
seh: It does make sense to me. It's important that users understand, though, that with `OnFailure` a container failure counts against the _Job_ retry accounting.

If a _Job_ uses pods with three containers, and two of the containers fail "at the same time" and get restarted by the _kubelet_, does that count as two retries of the _Job_?

soltysh: I'll add that explanation, in that case.

soltysh:
I've addressed first batch of comments.

contributors/design-proposals/job.md
```diff
@@ -29,27 +29,31 @@ here would be the ability to implement any type of batch oriented tasks.
...
 of failed pods to keep around.  This number will also take effect if none of the
-limits, described above, are set.
+limits described above are set.
 
 All of the above fields will be optional and will apply no matter which `restartPolicy`
-is set on a `PodTemplate`.
+is set on a `PodTemplate`.  The only difference applies to how failures are counted.
+For restart policy `Never` we count actual pod failures (reflected in `.status.failed`
+field). With restart policy `OnFailure` we look at pod restarts (calculated from
+`.status.containerStatuses[*].restartCount`).
```
seh: 👍

contributors/design-proposals/job.md
```diff
@@ -29,27 +29,31 @@ here would be the ability to implement any type of batch oriented tasks.
...
-over specified period of time.  In the case when only one of them is specified
-an exponential backoff with duration of 10 seconds and factor of 2 will be applied
-in such a way that either time or number is reached.  After reaching the limit
-a Job will be marked as failed.
-
-Additionally, to help debug the issue with a job, and limit the impact of having
-too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243))
+to introduce following fields will be introduced, which will control the exponential
+backoff when retrying a Job: number of retries and time to retry.  The two fields
+will allow creating a fine-grained control over the backoff policy, limiting the
```
seh: - [ ] Reading this again, I think this would read better as  
  > will allow fine-grained control over

contributors/design-proposals/job.md
```diff
@@ -106,13 +110,13 @@ type JobSpec struct {
...
-    // before the system tries to terminate it; value must be positive integer
+    // before the system tries to terminate it; value must be a positive integer.
     ActiveDeadlineSeconds *int
 
-    // Optional number of retries, before marking this job failed.
+    // Optional number of retries before marking this job failed.
     BackoffLimit *int
 
-    // Optional time (in seconds), how log a job should be retried before marking it failed.
+    // Optional time (in seconds) specifying how long a job should be retried before marking it failed.
```
seh: - [ ] Obviously this duration shouldn't be negative (in which case I assume we'd ignore it, or reject it), but what about zero?  
  Since it's a pointer, the pointer being nil means that it's not set, but what if it's present but zero? Is that the same as it not being set, or is that a valid duration meaning that we won't wait any amount of time for a job complete (that it must complete immediately)?

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,35 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
 here would be the ability to implement any type of batch oriented tasks.
 
 
+## Backoff policy and failed pod limit
+
+By design, Jobs do not have any notion of failure, other than a pod's `restartPolicy`
+which is mistakenly taken as Job's restart policy ([#30243](https://github.com/kubernetes/kubernetes/issues/30243),
+[#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
+situation where one wants to fail a Job after some amount of retries over a certain
+period of time, due to a logical error in configuration etc.  To do so we are going
```
agonzalezro: - [ ] I think you are repeating "introduce{,d}". It should be something like: "To do so we are going to introduce the following fields"

contributors/design-proposals/job.md
```diff
@@ -79,8 +109,21 @@ type JobSpec struct {
     // job should be run with. Defaults to 1.
     Completions *int
 
+    // Optional duration in seconds relative to the startTime that the job may be active
+    // before the system tries to terminate it; value must be a positive integer.
+    ActiveDeadlineSeconds *int
```
agonzalezro: - [ ] Can `uint` be used, also in the other fields?

contributors/design-proposals/job.md
```diff
@@ -79,8 +109,21 @@ type JobSpec struct {
     // job should be run with. Defaults to 1.
     Completions *int
 
+    // Optional duration in seconds relative to the startTime that the job may be active
```
agonzalezro: - [ ] I think the comments for exported fields should start with the name of the field: "ActiveDeadlineSeconds is an optional..."

contributors/design-proposals/job.md
soltysh: We usually require those being >= 0, it'll be checked in validation.

contributors/design-proposals/job.md
soltysh: We used ints, for consistency and then do appropriate validation so this is non-negative and bigger than 0.

contributors/design-proposals/job.md
soltysh: Not anymore, the def will have those in place.

soltysh:
@erictune / @janetkuo any objections for merging this in?

contributors/design-proposals/job.md
kargakis: *int32

kargakis: > *int32

Actually [*int64](https://github.com/kubernetes/kubernetes/blob/ee39d359dd0896c4c0eccf23f033f158ad3d3bd7/pkg/apis/batch/v1/types.go#L80)

kargakis: Can you also update the rest of the fields to their canonical types?

contributors/design-proposals/job.md
```diff
@@ -79,8 +109,21 @@ type JobSpec struct {
...
+    // before the system tries to terminate it; value must be a positive integer.
+    ActiveDeadlineSeconds *int
+
+    // Optional number of retries before marking this job failed.
+    BackoffLimit *int
+
+    // Optional time (in seconds) specifying how long a job should be retried before marking it failed.
+    BackoffDeadlineSeconds *int
```
kargakis: ProgressDeadlineSeconds to match the field in Deployments? Seems like it's the same concept although in Deployment we merely add a condition in the DeploymentStatus.

contributors/design-proposals/job.md
```diff
@@ -79,8 +109,21 @@ type JobSpec struct {
...
+
+    // Optional number of failed pods to retain.
+    FailedPodsLimit *int
```
kargakis: When are pods left around? I guess when pods have restartPolicy=Never? It feels weird to add a feature that is not related directly to the failurePolicy and also is useful for a specific case in Jobs. Can we leave it out of this proposal?

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,35 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+
+By design, Jobs do not have any notion of failure, other than a pod's `restartPolicy`
+which is mistakenly taken as Job's restart policy ([#30243](https://github.com/kubernetes/kubernetes/issues/30243),
+[#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
+situation where one wants to fail a Job after some amount of retries over a certain
+period of time, due to a logical error in configuration etc.  To do so we are going
+to introduce following fields, which will control the exponential backoff when
+retrying a Job: number of retries and time to retry.  The two fields will allow
+fine-grained control over the backoff policy, limiting the number of retries over
+a specified period of time.  If only one of the two fields is supplied, an exponential
```
kargakis: It's not clear what happens if both fields are specified. Is only the number of retries limited or both of the fields limit each other?

kargakis: Ok, it's clearer after rereading the bullet points below.

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,35 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+backoff with an intervening duration of ten seconds and a factor of two will be
```
erictune: I don't think we should promise our users exponential backoff -- we should give ourselves the freedom to do other things in the future. Just say that we will retry at a time that the system choses, and that in any case the number of retries won't exceed the number specified by the user.

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,35 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+applied, such that either:
+* the number of retries will not exceed a specified count, if present, or
+* the maximum time elapsed will not exceed the specified duration, if present.
+
+Additionally, to help debug the issue with a Job, and limit the impact of having
+too many failed pods left around (as mentioned in [#30243](https://github.com/kubernetes/kubernetes/issues/30243)),
+we are going to introduce a field which will allow specifying the maximum number
+of failed pods to keep around.  This number will also take effect if none of the
+limits described above are set.
```
erictune: Is the intent that users will typically set this to 0 unless they expect to need to debug?    Or do we typically want to keep around at least 1 failed pod so the user can get the logs?

 not save all pods without

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,35 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+
+All of the above fields will be optional and will apply no matter which `restartPolicy`
+is set on a `PodTemplate`.  The only difference applies to how failures are counted.
+For restart policy `Never` we count actual pod failures (reflected in `.status.failed`
+field). With restart policy `OnFailure` we look at pod restarts (calculated from
+`.status.containerStatuses[*].restartCount`).
```
erictune: This cannot be synchronous, though, so the limit has to be approximate.  Document that it is precise with Never, and approximate with OnFailure?

contributors/design-proposals/job.md
erictune: I think this is the deadline for the entire job to complete, regardless of whether there are restarts, right?  Document that.

contributors/design-proposals/job.md
```diff
@@ -79,8 +109,21 @@ type JobSpec struct {
     // job should be run with. Defaults to 1.
     Completions *int
 
+    // Optional duration in seconds relative to the startTime that the job may be active
+    // before the system tries to terminate it; value must be a positive integer.
+    ActiveDeadlineSeconds *int
+
+    // Optional number of retries before marking this job failed.
+    BackoffLimit *int
```
erictune: Why BackoffLimit instead of RetryLimit or TriesLimit?

contributors/design-proposals/job.md
erictune: How would a user typically use ActiveDeadlineSeconds with BackoffDeadlineSeconds?

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,35 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
 
 
+## Backoff policy and failed pod limit
+
+By design, Jobs do not have any notion of failure, other than a pod's `restartPolicy`
+which is mistakenly taken as Job's restart policy ([#30243](https://github.com/kubernetes/kubernetes/issues/30243),
+[#[43964](https://github.com/kubernetes/kubernetes/issues/43964)]).  There are
+situation where one wants to fail a Job after some amount of retries over a certain
+period of time, due to a logical error in configuration etc.  To do so we are going
+to introduce following fields will be introduced, which will control the exponential
```
sdminonne: ~~will be introduced~~

contributors/design-proposals/job.md
sdminonne: @erictune @soltysh if I understand correctly the only check system may run is: `ActiveDeadlineSeconds` must be strictly bigger than `BackoffDeadlineSeconds`, missing something?

contributors/design-proposals/job.md
sdminonne: I vote to have this defaulted to 1

contributors/design-proposals/job.md
soltysh: Correct.

soltysh: It clearly states it's relative to job's startTime.

contributors/design-proposals/job.md
soltysh: @kargakis this is different than `ProgressDeadlineSecods`, in deployments world PDS is how long you are waiting for the job to make progress and than fail. With Job `BackoffDeadlineSeconds` applies at any point in time. This means that if a Job expects to reach 5 completions, 4 of them can succeed and only the 5th Pod might be having troubles and BDS should still apply. That's why I'd rather not to use the same name, it's intentional.

soltysh: @erictune @sdminonne ADS is the overall time for a Job (max life-time). Whereas BDS is counted from the moment the first failure happens. But yes, usually BDS will be smaller than ADS, otherwise you'll never hit BDS in case of a failure.

contributors/design-proposals/job.md
soltysh: This is for debugging purposes. Leaving pods should allow you to figure out what went wrong. Without it and with restart policy Never you might end up with many failed pods, if you set the limits high. That was one of the original requests wrt to the backoff policy, so I think it's crucial to address it here.

contributors/design-proposals/job.md
soltysh: 1 seems reasonable default.

contributors/design-proposals/job.md
kargakis: Well, PDS is not that different. We start counting from the last time we saw progress and once the deadline is exceeded we merely switch the Progressing condition to False. "Failing the deployment" should be something that a higher-level orchestrator decides based on the condition.

contributors/design-proposals/job.md
soltysh: Yup, makes sense.

contributors/design-proposals/job.md
soltysh: I went with Backoff as a prefix for every field related to that, since we're talking about Backoff policy.

soltysh:
@erictune @kargakis I've addressed your comments, ptal once more. @sdminonne mind pointing your guy who will be implementing this to have a look at it, as well.

sdminonne:
@soltysh Cedric's GH account is @clamoriniere1A

sdminonne:
@clamoriniere1A mind have a look at this and eventually comment?

clamoriniere1A:
@sdminonne sure, I have started to look at it.

clamoriniere1A:
I started to work on the Job Backoff policy: https://github.com/kubernetes/kubernetes/pull/48075

contributors/design-proposals/job.md
```diff
@@ -73,14 +105,29 @@ type JobSpec struct {
...
+    Completions *int32
+
+    // Optional duration in seconds relative to the startTime that the job may be active
+    // before the system tries to terminate it; value must be a positive integer.
+    // It applies to overall job run time, no matter of the value of completions
+    // or parallelism parameters.
+    ActiveDeadlineSeconds *int64
+
+    // Optional number of retries before marking this job failed.
+    BackoffLimit *int32
```
erictune: Any reason to pick Backoff rather than Retry or Restart?

erictune:
Is it correct that: with Pod restart policy `OnFailure`the backoff amount is determined by the Kubelet, not Job controller and it [is documented](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy) as *exponential back-off delay (10s, 20s, 40s …) capped at five minutes, and is reset after ten minutes of successful execution*.

erictune:
2)
Is it correct that,  with Pod restart policy `Never`, the backoff amount is determined by the Job controller?  

3)
  If `BackoffDeadlineSeconds` is not set, then is the backoff sequence is (10s, 20s, 40s, ...).  

4)
If `BackoffDeadlineSeconds` is also set to N second, then is the backoff sequence is (N, 2N, 4N, ...).

5)

 In the case of a parallel job is there one backoff counter for the whole job or one per node?
(I'm fine with inconsistency for parallel jobs, as I think these are not being widely used.)

erictune:
BTW, I like fixing this, just trying to see it be well specified, or else document if it is intentionally not fully specified.

soltysh:
> Is it correct that: with Pod restart policy OnFailurethe backoff amount is determined by the Kubelet, not Job controller and it is documented as exponential back-off delay (10s, 20s, 40s …) capped at five minutes, and is reset after ten minutes of successful execution.

True, but can still be overwritten by the job controller values. Do you think we should only apply to restart policy `Never`? Which seems reasonable, but still would worth some explanation in the docs.

> Is it correct that, with Pod restart policy Never, the backoff amount is determined by the Job controller?

Yes.

> 
> If Pod restart policy is Never and BackoffDeadlineSeconds is not set, then is the backoff sequence is (10s, 20s, 40s, ...).

Yes.

> If Pod restart policy is Never and BackoffDeadlineSeconds is also set to N second, then is the backoff sequence is (N, 2N, 4N, ...).

Initially, the sequence was designed to hold overall value. Don't know why currently. But thinking about it right now it does make sense to be just that what you're saying. Lemme update the doc. 

> In the case of a parallel job, with Pod restart policy set to Never, is there one backoff counter for the whole job or one per node?
> (I'm fine with inconsistency for parallel jobs, as I think these are not being widely used.)

A whole job. We don't count failures per pod, so we're not able to distinguish between them here, either. Additionally, due to https://github.com/kubernetes/kubernetes/issues/28486 we're not able to implement this, yet. 

> BTW, I like fixing this, just trying to see it be well specified, or else document if it is intentionally not fully specified.

+1

soltysh:
> If Pod restart policy is Never and BackoffDeadlineSeconds is not set, then is the backoff sequence is (10s, 20s, 40s, ...).
> 
> Yes.

For this to happen, you need to set `BackoffLimit`. When neither `BackoffLimit` nor `BackoffDeadlineSeconds` is set, no backoff is applied to this job. That's very important to note.

> If Pod restart policy is Never and BackoffDeadlineSeconds is also set to N second, then is the backoff sequence is (N, 2N, 4N, ...).
> 
> Initially, the sequence was designed to hold overall value. Don't know why currently. But thinking about it right now it does make sense to be just that what you're saying. Lemme update the doc.

OK, now that I sat and almost started rewriting this I remember my intention. I (after talking to you) wanted to hide the details of the exponential backoff and give users two simple values driving it:
1. the number of retries (`BackoffLimit`),
2. maximum time of retries (`BackoffDeadlineSeconds`). 
What and how the actual implementation will handle this is out of scope of this proposal. Obviously setting low values of `BackoffDeadlineSeconds` doesn't make sense and should be verified during validation. 

I hope that this make sense?

erictune:
>  Do you think we should only apply to restart policy Never? Which seems reasonable, but still would worth some explanation in the docs.

Seems reasonable to me.  Yes, would want to explain in design and user docs.

erictune:
> We don't count failures per pod, so we're not able to distinguish between them here, either. 

Sounds fine.  Just need to document it.

erictune:
> For this to happen, you need to set BackoffLimit. When neither BackoffLimit nor BackoffDeadlineSeconds is set, no backoff is applied to this job. That's very important to note.

Why not?  Out of concern that it would be a change in default behavior?

erictune:
> OK, now that I sat and almost started rewriting this I remember my intention. I (after talking to you) wanted to hide the details of the exponential backoff and give users two simple values driving it...

I also like this approach of hiding the backoff from users.  But:
  - If you are hiding the exact backoff sequence from users, then the user should not have to think about the number of retries even.   So, `BackoffLimit` field is not needed.
 -  if a user wants to stop after some amount of time, then `ActiveDeadlineSeconds` can accomplish that too.  Not sure how different `BackoffDeadlineSeconds ` is.

WDYT?

Okay, this different approach also makes sense to me.

soltysh:
> Do you think we should only apply to restart policy Never? Which seems reasonable, but still would worth some explanation in the docs.
> 
> Seems reasonable to me. Yes, would want to explain in design and user docs.

After more digging, it doesn't make much sense. The backoff here will fail the job at some point, no matter which restart policy is set. If we apply this only for `Never` the jobs with `OnFailure` will never get the chance to be killed and will keep restarting forever. 
From what I've just checked the backoff for pods only goes up 10s, 20s... until it reaches 5m, but it never stops, whereas here we want the job to fail at some point. Eventually, we could apply only `BackoffLimit` for `OnFailure`, whereas both would be needed for `Never`, wdyt?

> Sounds fine. Just need to document it.

Added paragraph. See last commit. 


> For this to happen, you need to set BackoffLimit. When neither BackoffLimit nor BackoffDeadlineSeconds is set, no backoff is applied to this job. That's very important to note.
> 
> Why not? Out of concern that it would be a change in default behavior?

Mostly yes, but I guess most importantly we want to protect the cluster. So I'm fine with having some defaults. Let's then use the same as pods have, iow. 10s, 20s, 40s, 1m20s, 2m40s, and 5m which gives us `BackoffLimit` at 6, and `BackoffDuration` to 10, wdyt?

> I also like this approach of hiding the backoff from users. But:
> 
>     If you are hiding the exact backoff sequence from users, then the user should not have to think about the number of retries even. So, BackoffLimit field is not needed.
>     if a user wants to stop after some amount of time, then ActiveDeadlineSeconds can accomplish that too. Not sure how different BackoffDeadlineSeconds is.
> 
> WDYT?

OK, so let's go with `BackoffLimit` as number of retries and `BackoffDuration` being the initial retry value which is then multiplied by factor of 2, wdyt?

soltysh:
I've just pushed the last commit which:
1. expands the failure counting procedure
2. replaces `BackoffDeadlineSeconds` with `BackoffSeconds`

@erictune lemme know what you think about the current state

erictune:
Thanks for the update to the doc.  This is getting clearer.  

Just to make sure I understand, let me run through some examples.

Assume a pod always runs briefly then fails.

| `BackoffLimit` | `BackoffSeconds` | `ActiveDeadlineSeconds` | Result |
|----------------|--------------------|---------------------------|---------------------------|
| unspecified      |   unspecified           |   unspecified                       | 10s,  20s, 40s, 1m20s, 2m40s, 5m20s, Job Failed. |
| 0      |   unspecified           |   unspecified                       | Job terminated after (approximately) first Pod Failure  -- must use with `Never` to get close to Run-Once behavior |
| 1000000000      |   unspecified           |   unspecified                       | 10s, 20s, 40s, 1m20s, 2m40s, 5m20s,  5m20s,  5m20s,  5m20s, ...... Job Running basically forever |
| 0      |   unspecified           |   3600s                       | 10s, 20s, 40s, 1m20s, 2m40s, 5m20s,  5m20s,  5m20s,  5m20s, 5m20s, 5m20s, 5m20s, 5m20s, 5m20s, Job Fails after 1 hour |
| unspecified      |   unspecified           |   60s                       | 10s, 20s, 40s Job Failed |

soltysh:
Yup, that's pretty much what it is. 
I've also updated the doc with the information that `BackoffLimit` and `BackoffSeconds` applies to `Never`, whereas only `BackoffLimit` applies to `OnFailure`, as mentioned before. 
Lemme know if this is good for final sign off.

paulwalker:
This recent development looks good, but If I may ask, what is the current best config/pattern for a try-once job in 1.7.3? TIA

soltysh:
@paulwalker there's nothing unfortunately. You'd need to implement similar mechanism inside of your job manually, I'm afraid 😞

kargakis:
Or run a pod with RestartPolicy=Never.

chirag04:
@erictune can you elaborate on `Job terminated after (approximately) first Pod Failure -- must use with Never to get close to Run-Once behavior`?

why is that a `close` to Run-once behavior?

erictune:
With these changes, a job with "RestartPolicy=Never" and BackoffLimit=0
with run once under almost all circumstances.

However, the communication between the job controller, apiserver, kubelet
and docker is not atomic.  So, it is possible that under some very unlikely
combination of crashes/restarts
in the kubelet, docker and/or job controller, that the program in the Job's
pod template might be started twice.

Basically, I would use it for any task that has built-in idempotence, and
for most types of tasks that don't have built-in idempotence.  I would not
use it for say, transferring money, without some additional
application-level check to avoid duplicates.

chirag04:
@erictune thanks for the explanation and that makes perfect sense. 👍

contributors/design-proposals/job.md
```diff
@@ -26,6 +27,41 @@ Jobs are needed for executing multi-pod computation to completion; a good exampl
...
+situation where one wants to fail a Job after some amount of retries over a certain
+period of time, due to a logical error in configuration etc.  To do so we are going
+to introduce the following fields, which will control the backoff policy: a number of
+retries and an initial time of retry.  The two fields will allow fine-grained control
+over the backoff policy.  Each of the two fields will use a default value if none
+is provided,  `BackoffLimit` is set by default to 6 and `BackoffSeconds` to 10s.
+This will result in the following retry sequence: 10s, 20s, 40s, 1m20s, 2m40s,
+5m20s.  After which the job will be considered failed.
+
+Additionally, to help debug the issue with a Job, and limit the impact of having
```
chenchun: Why not also limit the number of successful pods which may takes a lot of resource either ?

PR #443, opened 2017-03-10
List/Watch/Get of objects associated with node

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/40476

@lavalamp @liggitt @kubernetes/sig-api-machinery-misc 
@dchen1107

PR #585, opened 2017-04-26
Scoped Kubelet API Access

liggitt:
Proposal to subdivide kubelet API access to read secrets/configmaps, and to mutate node and pod resources

### DIFF
--- a/contributors/design-proposals/kubelet-authorizer.md
+++ b/contributors/design-proposals/kubelet-authorizer.md
@@ -1,0 +1,179 @@
+# Scoped Kubelet API Access
+
+Author: Jordan Liggitt (jliggitt@redhat.com)
+
+## Overview
+
+Kubelets are primarily responsible for:
+* creating and updating status of their Node API object
+* running and updating status of Pod API objects bound to their node
+* creating/deleting "mirror pod" API objects for statically-defined pods running on their node
+
+To run a pod, a kubelet must have read access to the following objects referenced by the pod spec:
+* Secrets
+* ConfigMaps
+* PersistentVolumeClaims (and any bound PersistentVolume or referenced StorageClass object)
+
+As of 1.6, kubelets have read/write access to all Node and Pod objects, and 
+read access to all Secret, ConfigMap, PersistentVolumeClaim, and PersistentVolume objects.
+This means that compromising a node gives access to credentials that allow modifying other nodes,
+pods belonging to other nodes, and accessing confidential data unrelated to the node's pods.
+
+This document proposes limiting a kubelet's API access using a new node authorizer, admission plugin, and additional API validation:
+* Node authorizer
+  * Authorizes requests from nodes using a fixed policy identical to the default RBAC `system:node` cluster role
+  * Further restricts secret and configmap access to only allow reading objects referenced by pods bound to the node making the request
+* Node admission
+  * Limit nodes to only be able to mutate their own Node API object
+  * Limit nodes to only be able to create mirror pods bound to themselves
+  * Limit nodes to only be able to mutate mirror pods bound to themselves
+  * Limit nodes to not be able to create mirror pods that reference API objects (secrets, configmaps, service accounts, persistent volume claims)
+* Additional API validation
+  * Reject mirror pods that are not bound to a node
+  * Reject pod updates that remove mirror pod annotations
+
+## Alternatives considered
+
+**Can this just be enforced by authorization?**
+
+Authorization does not have access to request bodies (or the existing object, for update requests),
+so it could not restrict access based on fields in the incoming or existing object.
+
+**Can this just be enforced by admission?**
+
+Admission is only called for mutating requests, so it could not restrict read access.
+
+**Can an existing authorizer be used?**
+
+Only one authorizer (RBAC) has in-tree support for dynamically programmable policy.
+
+Manifesting RBAC policy rules to give each node access to individual objects within namespaces
+would require large numbers of frequently-modified roles and rolebindings, resulting in
+significant write-multiplication.
+
+Additionally, not all clusters will use RBAC, but all useful clusters will have nodes.
+A node-specific authorizer allows cluster admins to continue to use their authorization mode of choice.
+
+## Node identification
+
+The first step is to identify whether a particular API request is being made by 
+a node, and if so, from which node.
+
+The proposed node authorizer and admission plugin will take a `NodeIdentifier` interface:
+
+```go
+type NodeIdentifier interface {
+  // IdentifyNode determines node information from the given user.Info.
+  // nodeName is the name of the Node API object associated with the user.Info,
+  // and may be empty if a specific node cannot be determined.
+  // isNode is true if the user.Info represents an identity issued to a node.
+  IdentifyNode(user.Info) (nodeName string, isNode bool)
+}
+```
+
+The default `NodeIdentifier` implementation:
+* `isNode` - true if the user groups contain the `system:nodes` group
+* `nodeName` - populated if `isNode` is true, and the user name is in the format `system:node:<nodeName>`
+
+This group and user name format match the identity created for each kubelet as part of [kubelet TLS bootstrapping](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/).
+
+## Node authorizer
+
+A new node authorizer will be inserted into the authorization chain:
+* API server authorizer (existing, authorizes "loopback" API clients used by components within the API server)
+* Node authorizer (new)
+* User-configured authorizers... (e.g. ABAC, RBAC, Webhook)
+
+The node authorizer does the following:
+1. If a request is not from a node (`IdentifyNode()` returns isNode=false), reject
+2. If a request is not allowed by the rules in the default `system:node` cluster rule, reject
+3. If a specific node cannot be identified (`IdentifyNode()` returns nodeName=""):
+    * If in compatibility-mode (default), allow. This lets nodes that don't use node-specific identities continue to work with the broad authorization rules in step 2.
+    * If in strict-mode, reject. This lets deployments that provision all nodes with individual identities to indicate that only identifiable nodes should be allowed.
+4. If a request is for a secret, configmap, persistent volume or persistent volume claim, reject unless the verb is `get`, and the requested object is related to the requesting node:
+    
+    * node -> pod
+    * node -> pod -> secret
+    * node -> pod -> configmap
+    * node -> pod -> pvc
+    * node -> pod -> pvc -> pv
+    * node -> pod -> pvc -> pv -> secret
+5. For other resources, allow
+
+Subsequent authorizers in the chain can run and choose to allow requests rejected by the node authorizer.
+
+## Node admission
+
+A new node admission plugin is made available that does the following:
+
+1. If a request is not from a node (`IdentifyNode()` returns isNode=false), allow the request
+2. If a specific node cannot be identified (`IdentifyNode()` returns nodeName=""):
+  * If in compatibility-mode (default), allow. This lets nodes that don't use node-specific identities continue to work.
+  * If in strict-mode, reject. This lets deployments that provision all nodes with individual identities to indicate that only identifiable nodes should be allowed.
+3. For requests made by identifiable nodes:
+  * Limits `create` of node resources:
+    * only allow the node object corresponding to the node making the API request
+  * Limits `create` of pod resources:
+    * only allow pods with mirror pod annotations
+    * only allow pods with nodeName set to the node making the API request
+    * do not allow pods that reference any API objects (secrets, serviceaccounts, configmaps, or persistentvolumeclaims)
+  * Limits `update` of node and nodes/status resources:
+    * only allow updating the node object corresponding to the node making the API request
+  * Limits `update` of pods/status resources:
+    * only allow reporting status for pods with nodeName set to the node making the API request
+  * Limits `delete` of node resources:
+    * only allow deleting the node object corresponding to the node making the API request
+  * Limits `delete` of pod resources:
+    * only allow deleting pods with nodeName set to the node making the API request
+
+## API Changes
+
+Change Pod validation for mirror pods:
+  * Reject `create` of pod resources with mirror pod annotations that do not specify a nodeName
+  * Reject `update` of pod resources with mirror pod annotations that modify or remove the mirror pod annotation
+
+## RBAC Changes
+
+As of 1.6, the `system:node` cluster role is automatically bound to the `system:nodes` group when using RBAC.
+
+Because the node authorizer accomplishes the same purpose, with the benefit of additional restrictions
+on secret and configmap access, this binding is no longer needed, and will no longer be set up automatically.
+
+The `system:node` cluster role will continue to be created when using RBAC,
+for compatibility with deployment methods that bind other users or groups to that role.
+
+## Migration considerations
+
+### Kubelets outside the `system:nodes` group
+
+Kubelets outside the `system:nodes` group would not be authorized by the node authorizer,
+and would need to continue to be authorized via whatever mechanism currently authorizes them.
+The node admission plugin would not restrict requests from these kubelets.
+
+### Kubelets with undifferentiated usernames
+
+In some deployments, kubelets have credentials that place them in the `system:nodes` group,
+but do not identify the particular node they are associated with.
+Those kubelets would be broadly authorized by the node authorizer,
+but would not have secret and configmap requests restricted.
+The node admission plugin would not restrict requests from these kubelets.
+
+### Upgrades from previous versions
+
+Versions prior to 1.7 that have the `system:node` cluster role bound to the `system:nodes` group would need to
+remove that binding in order for the node authorizer restrictions on secret and configmap access to be effective.
+
+## Future work
+
+Node and pod mutation, and secret and configmap read access are the most critical permissions to restrict.
+Future work could further limit a kubelet's API access:
+* only get persistent volume claims and persistent volumes referenced by a bound pod
+* only write events with the kubelet set as the event source
+* only get/list/watch pods bound to the kubelet's node (requires additional list/watch authorization capabilities)
+* only get/list/watch it's own node object (requires additional list/watch authorization capabilities)
+
+Features that expand or modify the APIs or objects accessed by the kubelet will need to involve the node authorizer.
+Known features in the design or development stages that might modify kubelet API access are:
+* [Dynamic kubelet configuration](https://github.com/kubernetes/features/issues/281)
+* [Local storage management](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/local-storage-overview.md)
+* [Bulk watch of secrets/configmaps](https://github.com/kubernetes/community/pull/443)


liggitt:
https://github.com/kubernetes/features/issues/279
https://github.com/kubernetes/kubernetes/issues/40476

liggitt:
cc @kubernetes/sig-auth-proposals

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+## API Changes
+
+None
+
+## RBAC Changes
+
+Currently, the `system:node` cluster role is automatically bound to the `system:nodes` group.
+
+Because the node authorizer accomplishes the same purpose, with the benefit of additional restrictions
+on secret and configmap access, this binding is no longer needed, and will no longer be set up automatically.
```
ericchiang: This is going to impact anyone using `system:node` x509 certs that haven't been able to dynamically make the CN `system:node:<nodeName>`.

liggitt: It shouldn't. See the section below that addresses this.

ericchiang: I mean that those deployment tools, to switch to 1.7, will either have to start creating that ClusterRoleBinding or dynamically setting the CN.

For example look at kube-up's x509 generation https://github.com/kubernetes/kubernetes/blob/274df99e9bf63213325d9813eec84288d3c5ab73/cluster/common.sh#L996

I don't know how prevalent this is, and I don't think it's a big problem, but it's probably worth checking popular tools for this pattern over this release cycle.

liggitt: No, the node authorizer would authorize against the broad rules for `system:nodes` members (as the bound cluster role does today), then further subdivide if the identity allowed us to determine the specific node

ericchiang: Ah, sorry. Misread that last section. Thanks.

destijl:
Thanks for working on this, LGTM.

contributors/design-proposals/kubelet-authorizer.md
liggitt: No problem, I'll call it out to make it clearer

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+  // nodeName is the name of the Node API object associated with the user.Info,
+  // and may be empty if a specific node cannot be determined.
+  // isNode is true if the user.Info represents an identity issued to a node.
+  IdentifyNode(user.Info) (nodeName string, isNode bool)
+}
+```
+
+The default `NodeIdentifier` implementation:
+* `isNode` - true if the user groups contain the `system:nodes` group
+* `nodeName` - populated if `isNode` is true, and the user name is in the format `system:node:<nodeName>`
```
liggitt: Will clarify that this dovetails with the client certs requested by the kubelet TLS bootstrap process (and kubeadm join process)

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+Kubelets are responsible for:
+* creating and updating status of their Node API object
+* updating status of Pod API objects bound to their node
+* creating/deleting "mirror pod" API objects for statically-defined pods running on their node
+* reading Secret and ConfigMap objects referenced by pod specs bound to their node
+
+Currently, kubelets have read/write access to all Node and Pod objects, and read access to all Secret and ConfigMap objects.
+This means that compromising a node gives access to credentials with power to make escalating API calls that affect other nodes.
+
+This proposal limits kubelets' API access using a new node authorizer and admission plugin:
```
smarterclayton: Expected an alternatives section here or below describing why rbac isn't enough and other alternatives considered (I can't remember others).  Also a caution about future use of this pattern for another client

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+* Node authorizer
+  * Authorizes requests from nodes using the existing policy rules developed for the `system:node` cluster role
+  * Further restricts secret and configmap requests to only authorize reading objects referenced by pods bound to the node
+* Node admission
+  * Limits a node to mutating its own Node API object
+  * Limits a node to mutating pods bound to itself
```
smarterclayton: Would add words to the effect of "we represent a policy that is a peer to rbac with code" or something that gives this a conceptual hat hook

contributors/design-proposals/kubelet-authorizer.md
smarterclayton: Maybe a quick blurb about whether future extension is anticipated (we don't expect there to be things that are like nodes that don't fit this reasonable pattern for cert and naming)

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+* only allow pods with nodeName set to the node making the API request
+* only allow pods with no secret references
+
+Limits `update`,`patch`,`delete` of node and nodes/status resources by identifiable nodes:
+* only allow modifying the node object corresponding to the node making the API request
+
+Limits `update`,`patch`,`delete` of pod and pod/status resources by identifiable nodes:
+* only allow modifying pods with nodeName set to the node making the API request (requires fetching the pod on delete)
+* do not allow removing a mirror pod annotation
+
```
timstclair: Also create/update/patch of events, limited to `InvolvedObject`s bound to the node

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+The node authorizer is responsible for ensuring:
+* The request is made by a node (`IdentifyNode()` returns isNode=true)
+* The request is allowed by the rules contained in the default `system:node` cluster role
+
+Requests that do not meet those conditions are forbidden by this authorizer.
+Subsequent authorizers in the chain can run and choose to allow the request.
+
+Requests from identifiable nodes (`IdentifyNode()` returns nodeName != "") for secrets and configmaps are further restricted:
+* Requests for secrets are limited to `get`, and the requested secret must be related to the requesting node by one of the following relationships:
```
timstclair: Add a cross reference to https://github.com/kubernetes/community/pull/443. With that addition, this should include `list`

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+  * node -> pod -> secret
+  * node -> pod -> pvc -> pv -> secret
+  * node -> pod -> pvc -> storageclass -> secret
+* Requests for configmaps are limited to `get`, and the requested configmap must be related to the requesting node by one of the following relationships:
+  * node -> pod -> configmap
+
```
timstclair: Also various other read (get/list/watch) requests should be authorized for objects bound to the node, including:
- pods
- pvc & pv

contributors/design-proposals/kubelet-authorizer.md
timstclair: There are a few other things granted by the `system:node` clusterrole not listed here:
```go
// Needed to check API access.  These creates are non-mutating
rbac.NewRule("create").Groups(authenticationGroup).Resources("tokenreviews").RuleOrDie(),
rbac.NewRule("create").Groups(authorizationGroup).Resources("subjectaccessreviews", "localsubjectaccessreviews").RuleOrDie(),
// Needed to build serviceLister, to populate env vars for services
rbac.NewRule(Read...).Groups(legacyGroup).Resources("services").RuleOrDie(),

// TODO: restrict to pods scheduled on the bound node once supported
rbac.NewRule(Read...).Groups(legacyGroup).Resources("pods").RuleOrDie(),

// TODO: restrict to namespaces of pods scheduled on bound node once supported
// TODO: change glusterfs to use DNS lookup so this isn't needed?
// Needed for glusterfs volumes
rbac.NewRule("get").Groups(legacyGroup).Resources("endpoints").RuleOrDie(),
// Used to create a certificatesigningrequest for a node-specific client certificate, and watch
// for it to be signed. This allows the kubelet to rotate it's own certificate.
rbac.NewRule("create", "get", "list", "watch").Groups(certificatesGroup).Resources("certificatesigningrequests").RuleOrDie(),
```

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+  // isNode is true if the user.Info represents an identity issued to a node.
+  IdentifyNode(user.Info) (nodeName string, isNode bool)
+}
+```
+
+The default `NodeIdentifier` implementation:
+* `isNode` - true if the user groups contain the `system:nodes` group
+* `nodeName` - populated if `isNode` is true, and the user name is in the format `system:node:<nodeName>`
+
+## Node authorizer
```
timstclair: Can you clarify the logic for what goes into the authorizer vs the admission plugin?

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+* creating/deleting "mirror pod" API objects for statically-defined pods running on their node
+* reading Secret and ConfigMap objects referenced by pod specs bound to their node
+
+Currently, kubelets have read/write access to all Node and Pod objects, and read access to all Secret and ConfigMap objects.
+This means that compromising a node gives access to credentials with power to make escalating API calls that affect other nodes.
+
+This proposal limits kubelets' API access using a new node authorizer and admission plugin:
+* Node authorizer
+  * Authorizes requests from nodes using the existing policy rules developed for the `system:node` cluster role
+  * Further restricts secret and configmap requests to only authorize reading objects referenced by pods bound to the node
```
liggitt: I wouldn't expect to let a node establish a relationship from a mirror pod to any API object (secret, configmap, pvc, serviceaccount, etc)

contributors/design-proposals/kubelet-authorizer.md
liggitt: no, the nodeName is used to filter pods, and the kubelet already knows it (since it is responsible for updating the Node status and querying bound pods).

contributors/design-proposals/kubelet-authorizer.md
liggitt: we'll never let nodes list or watch secrets, since that exposes the entire list. the bulk watch of individual resources will likely perform `get` authorization checks on the individual resources.

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+
+Limits `create` of pod resources by identifiable nodes:
+* only allow pods with mirror pod annotations
+* only allow pods with nodeName set to the node making the API request
+* only allow pods with no secret references
+
+Limits `update`,`patch`,`delete` of node and nodes/status resources by identifiable nodes:
+* only allow modifying the node object corresponding to the node making the API request
+
+Limits `update`,`patch`,`delete` of pod and pod/status resources by identifiable nodes:
```
liggitt: I'm pretty sure kubelets do the final deletion of a gracefully-deleted pod, after they've run the shutdown hooks and sent the termination signal. @smarterclayton?

contributors/design-proposals/kubelet-authorizer.md
liggitt: we can extend this further in the future if we want to, but I consider subdividing those reads out of scope for this proposal. pods are list/watched across all namespaces via a field selector, which is currently not available to authorization to enforce. PVC and PV objects should not contain sensitive information (technically configmaps should not either, but experience suggests they will)

contributors/design-proposals/kubelet-authorizer.md
timstclair: I think that would be a breaking change, since static pods can currently use PVCs and configmaps. I suppose we could make it a hard requirement for adopting the node admission plugin, but that might make adoption more difficult.

liggitt: If a node can simply create a static pod referencing any configmap, we cannot restrict node access to any configmaps (same for pod>pvc>secret references). It was my understanding that static pods were not supposed to reference API resources

timstclair: I aggree, I'm just not sure we ever documented or enforced that anywhere. This is where we enforce it for secrets & service accounts: https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/serviceaccount/admission.go#L149

liggitt: The node restricting admission plugin will forbid those. You'll have to choose whether you want to be able to lock down nodes or use static pods that use those features (which I don't think are intended to be used today). For compatibility, if the enforcement is via admission, they can opt in or out

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,119 @@
...
+
+Author: Jordan Liggitt (jliggitt@redhat.com)
+
+## Overview
+
+Kubelets are responsible for:
+* creating and updating status of their Node API object
+* updating status of Pod API objects bound to their node
+* creating/deleting "mirror pod" API objects for statically-defined pods running on their node
+* reading Secret and ConfigMap objects referenced by pod specs bound to their node
```
alex-mohr: +1 to (attempting?) to make a semi-complete list, even if they're not in-scope for v1.7 or v1.8 -- be nice to see a what's *sufficient* to get where we want to be, not just *known-necessary*.

Also @jbeda mentioned that we should eventually make kubelet/nodes have whitelisted access to the specific set of things they need access to?  And if so, having that list of necessary permissions would be a pre-requisite?

alex-mohr: Also, I think kubelets rely on an (implicit?) list of all Pods in the system via a watch to pods?  If so, then Kubelets have an additional requirement to
* get a list of Pods that are scheduled to that kubelet.

That is, the problem of allowing *read* access to a specific pod is different from a list query to find all visible-to-me pods?

timstclair:
This has some overlap with https://github.com/kubernetes/kubernetes/pull/29459

We might need some additional mechanism for granting nodes permission to objects that are used directly by the Kubelet, and not referenced by any pod.

@mtaufen

contributors/design-proposals/kubelet-authorizer.md
liggitt: all of those rules would be used as the coarse-grained authorization check, then additional restrictions places on specific resources

contributors/design-proposals/kubelet-authorizer.md
liggitt: I'm not sure how valuable that is, relative to the cost of evaluation on a high-volume API

contributors/design-proposals/kubelet-authorizer.md
liggitt: done

contributors/design-proposals/kubelet-authorizer.md
liggitt: added section for future work

liggitt:
comments addressed. @alex-mohr @smarterclayton @timstclair PTAL

derekwaynecarr:
this generally lgtm, agree depending on how #29459 shakes out, it may need to be tweaked further.

saad-ali:
CC @kubernetes/sig-node-proposals

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,169 @@
...
+pods belonging to other nodes, and accessing confidential data unrelated to the node's pods.
+
+This document proposes limiting a kubelet's API access using a new node authorizer and admission plugin:
+* Node authorizer
+  * Authorizes requests from nodes using a fixed policy identical to the default RBAC `system:node` cluster role
+  * Further restricts secret and configmap access to only allow reading objects referenced by pods bound to the node making the request
+* Node admission
+  * Limits nodes to only be able to mutate their own Node API object
+  * Limits nodes to only be able to mutate pods bound to themselves
+  * Limits nodes to only be able to create mirror pods
```
smarterclayton: Create mirror pods that are bound to themselves

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,169 @@
...
+  * Prevents creating mirror pods that reference API objects (secrets, configmaps, persistent volume claims)
+  * Prevents creating mirror pods that are not bound to nodes
```
smarterclayton: Bound to themselves

contributors/design-proposals/kubelet-authorizer.md
liggitt: the previous point implies that, but will make that explicit

smarterclayton:
Other than a few comments LGTM

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,169 @@
...
+
+Kubelets are primarily responsible for:
+* creating and updating status of their Node API object
+* running and updating status of Pod API objects bound to their node
+* creating/deleting "mirror pod" API objects for statically-defined pods running on their node
+
+To run a pod, a kubelet must have read access to the following objects referenced by the pod spec:
+* Secrets
+* ConfigMaps
+* PersistentVolumeClaims (and any bound PersistentVolume or referenced StorageClass object)
```
vishh: Note: As part of local storage management, kubelet will be writing to PersistentVolume and PersistentVolumeClaim objects

liggitt: Good to know. Is that just proposing nodes perform status update operations, or also create/delete operations?

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,169 @@
...
+This means that compromising a node gives access to credentials that allow modifying other nodes,
+pods belonging to other nodes, and accessing confidential data unrelated to the node's pods.
+
+This document proposes limiting a kubelet's API access using a new node authorizer and admission plugin:
+* Node authorizer
+  * Authorizes requests from nodes using a fixed policy identical to the default RBAC `system:node` cluster role
+  * Further restricts secret and configmap access to only allow reading objects referenced by pods bound to the node making the request
+* Node admission
+  * Limits nodes to only be able to mutate their own Node API object
+  * Limits nodes to only be able to mutate pods bound to themselves
```
deads2k: Will it prevent `ObjectMeta` mutation?

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,169 @@
...
+
+For requests made by identifiable nodes:
+  * Limits `create` of node resources:
+    * only allow the node object corresponding to the node making the API request
+  * Limits `create` of pod resources:
+    * only allow pods with mirror pod annotations
+    * only allow pods with nodeName set to the node making the API request
+  * Limits `update`,`delete` of node and nodes/status resources:
+    * only allow modifying the node object corresponding to the node making the API request
+  * Limits `update`,`delete` of pod and pod/status resources:
```
deads2k: The kubelet changes the spec of a pod object?

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,169 @@
...
+    * only allow modifying pods with nodeName set to the node making the API request (requires fetching the pod on delete)
+
+For requests made by any user:
```
deads2k: This information looks statically determinant.  Are we doing this in admission instead of validation because we want it to be optional?  Why not some sort `disableSomeValidation` flag instead?

contributors/design-proposals/kubelet-authorizer.md
timstclair: Sorry, I'm still confused by this. Yes, those permissions can still be granted as course-grained without additional restrictions, but they need to be granted somewhere. They could be left as RBAC rules, or granted at a course level in the node authorizer.

liggitt: As proposed, the built-in node authorizer would broadly authorize users in the `system:nodes` group that do not have a username that permits determining the specific node (just like the current RBAC binding of the `system:nodes` group to the `system:node` does)

timstclair: It would also need to do this for _some_ APIs for all users in the `system:nodes` group, not just those without a node-specific identity, right? I'm talking about things like "list services" and "get endpoints".

liggitt: yes... clearly I'm not explaining this well :)

This is how the node authorizer would work, in order:
1. If a request is not from a node, reject
2. If a request is not allowed by the `system:node` rules, reject
3. If a specific node cannot be identified, allow (this is what lets nodes that don't use node-specific identities continue to work with the broad authorization rules in step 2)
4. If resources we care about subdividing access to (secrets, configmaps, etc), determine if the requested object is related to the requesting node, and reject the request if it isn't
5. For other resources, allow

FWIW, this is my current authorizer implementation:

```go
func (r *NodeAuthorizer) Authorize(attrs authorizer.Attributes) (bool, string, error) {
	nodeName, isNode := r.identifier.NodeIdentity(attrs.GetUser())
	if !isNode {
		// reject requests from non-nodes
		return false, "", nil
	}
	if len(nodeName) == 0 && r.requireNodeName {
		// reject requests from unidentifiable nodes if in strict mode
		return false, "unknown node", nil
	}

	if !rbac.RulesAllow(attrs, r.nodeRules...) {
		// reject requests that don't fit within the broad authorization rules
		return false, "", nil
	}

	if len(nodeName) == 0 {
		// allow requests from unidentifiable nodes if not in strict mode
		return true, "", nil
	}

	// subdivide access to specific resources

	if attrs.GetAPIGroup() != api.GroupName {
		// the resources we care about subdividing happen to be in the core API group.
		// access to resources in other API groups is not subdivided.
		return true, "", nil
	}

	switch attrs.GetResource() {
	case "configmaps":
		if attrs.GetVerb() != "get" || len(attrs.GetName()) == 0 || len(attrs.GetNamespace()) == 0 {
			glog.V(2).Infof("NODE DENY: %#v", attrs)
			return false, "can only get individual configmaps", nil
		}
		return r.hasPathFrom(nodeName, ConfigMapNode, attrs.GetNamespace(), attrs.GetName()), "", nil

	case "secrets":
		if attrs.GetVerb() != "get" || len(attrs.GetName()) == 0 || len(attrs.GetNamespace()) == 0 {
			glog.V(2).Infof("NODE DENY: %#v", attrs)
			return false, "can only get individual secrets", nil
		}
		return r.hasPathFrom(nodeName, SecretNode, attrs.GetNamespace(), attrs.GetName()), "", nil

	case "persistentvolumeclaims":
		if attrs.GetVerb() != "get" || len(attrs.GetName()) == 0 || len(attrs.GetNamespace()) == 0 {
			glog.V(2).Infof("NODE DENY: %#v", attrs)
			return false, "can only get individual persistentvolumeclaims", nil
		}
		return r.hasPathFrom(nodeName, PVCNode, attrs.GetNamespace(), attrs.GetName()), "", nil

	case "persistentvolumes":
		if attrs.GetVerb() != "get" || len(attrs.GetName()) == 0 {
			glog.V(2).Infof("NODE DENY: %#v", attrs)
			return false, "can only get individual persistentvolumes", nil
		}
		return r.hasPathFrom(nodeName, PVNode, "", attrs.GetName()), "", nil

	default:
		// Access to other resources is not subdivided
		return true, "", nil
	}
}
```

timstclair: Thank you, this clears it up.

contributors/design-proposals/kubelet-authorizer.md
liggitt: no, clarified this is [update, pod/status] and [delete, pod]

contributors/design-proposals/kubelet-authorizer.md
liggitt: only pod/status updates are currently allowed

contributors/design-proposals/kubelet-authorizer.md
liggitt: split this into a validation section and an admission section

restricting mirror pods from referencing API objects is more related to not letting nodes expand their related objects graph, so I'd probably keep it in the admission plugin

liggitt:
comments addressed

liggitt:
@timstclair @deads2k PTAL

deads2k:
lgtm

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,167 @@
...
+
+As of 1.6, kubelets have read/write access to all Node and Pod objects, and 
+read access to all Secret, ConfigMap, PersistentVolumeClaim, and PersistentVolume objects.
+This means that compromising a node gives access to credentials that allow modifying other nodes,
+pods belonging to other nodes, and accessing confidential data unrelated to the node's pods.
+
+This document proposes limiting a kubelet's API access using a new node authorizer and admission plugin:
+* Node authorizer
+  * Authorizes requests from nodes using a fixed policy identical to the default RBAC `system:node` cluster role
+  * Further restricts secret and configmap access to only allow reading objects referenced by pods bound to the node making the request
```
mtaufen: Maybe add the ability to read Secrets (well, maaaaaaaybe Secrets) and ConfigMaps from a specific set of namespaces, whether or not Pods bound to the Node reference them. This has to do with the Kubelet needing access to the namespace containing ConfigMaps for Kubelet/Node configuration, wrt https://github.com/kubernetes/kubernetes/pull/29459

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,167 @@
...
+* Node admission
+  * Limit nodes to only be able to mutate their own Node API object
```
mtaufen: +1

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,167 @@
...
+* Additional API validation
+  * Reject mirror pods that are not bound to a node
+  * Reject pod updates that remove mirror pod annotations
+
+## Alternatives considered
+
+**Can this just be enforced by authorization?**
+
+Authorization does not have access to request bodies (or the existing object, for update requests),
+so it could not restrict access based on fields in the incoming or existing object.
```
mtaufen: There seem to be more and more use-cases for per-field access control. It would sure make things like this easier.

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,167 @@
...
+
+Admission is only called for mutating requests, so it could not restrict read access.
+
+**Can an existing authorizer be used?**
+
+Only one authorizer (RBAC) has in-tree support for dynamically programmable policy.
+
+Manifesting RBAC policy rules to give each node access to individual objects within namespaces
+would require large numbers of frequently-modified roles and rolebindings, resulting in
+significant write-multiplication.
```
mtaufen: Labels and selectors for ConfigMaps and Secrets might make writing the roles easier. I swear I saw an issue or something about this somewhere...

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,167 @@
...
+A new node authorizer will be inserted into the authorization chain:
+* API server authorizer (existing, authorizes "loopback" API clients used by components within the API server)
+* Node authorizer (new)
+* User-configured authorizers... (e.g. ABAC, RBAC, Webhook)
+
+The node authorizer does the following:
+1. If a request is not from a node (`IdentifyNode()` returns isNode=false), reject
+2. If a request is not allowed by the rules in the default `system:node` cluster rule, reject
+3. If a specific node cannot be identified (`IdentifyNode()` returns nodeName=""), allow (this lets nodes that don't use node-specific identities continue to work with the broad authorization rules in step 2)
+4. If a request is for a secret, configmap, persistent volume or persistent volume claim, reject unless the verb is `get`, and the requested object is related to the requesting node:
```
mtaufen: +1 only allowing read requests for these types

I think it makes sense for "related to the requesting node" to also include a set of arbitrary namespaces that are considered to be system-level (or just node-level).

contributors/design-proposals/kubelet-authorizer.md
liggitt: I'd like to wait until configmap/secret access for kubelets to read their own config formalizes before including it in this authorizer, since it is not meant to be configurable. I'll add an item to the future work list to track it.

In the meantime, access can be granted using whatever authorization method is in place for the cluster.

contributors/design-proposals/kubelet-authorizer.md
liggitt: > I think it makes sense for "related to the requesting node" to also include a set of arbitrary namespaces that are considered to be system-level (or just node-level).

This also seems like something that would vary by deployment. I'd like to see more use cases around this before making it part of the fixed node authorizer.

contributors/design-proposals/kubelet-authorizer.md
liggitt: They wouldn't help with transitive relationships (you can access secrets referenced by pods which reference your node)

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -0,0 +1,167 @@
...
+### Kubelets outside the `system:nodes` group
+
+Kubelets outside the `system:nodes` group would not be authorized by the node authorizer,
+and would need to continue to be authorized via whatever mechanism currently authorizes them.
+The node admission plugin would not restrict requests from these kubelets.
+
+### Kubelets with undifferentiated usernames
+
+In some deployments, kubelets have credentials that place them in the `system:nodes` group,
+but do not identify the particular node they are associated with.
```
timstclair: I think we should add an override option to disable this exception. I can't think of any way it would work, but I'm worried this leaves an opening for a sort of down-grade attack. For a deployment with identifiable node identities, removing this exception would provide additional assurance (and is probably also what we want eventually).

contributors/design-proposals/kubelet-authorizer.md
liggitt: requiring decoding objects prior to authorization is unlikely, especially given API aggregation

contributors/design-proposals/kubelet-authorizer.md
liggitt: agree on a strict mode option, will add

PR #589, opened 2017-04-28
Proposal: containerized mount utilities in pods

jsafrane:
@kubernetes/sig-storage-proposals  @kubernetes/sig-node-proposals

### DIFF
--- a/contributors/design-proposals/storage/containerized-mounter-pod.md
+++ b/contributors/design-proposals/storage/containerized-mounter-pod.md
@@ -1,0 +1,149 @@
+# Containerized mounter using volume utilities in pods
+
+## Goal
+Kubernetes should be able to run all utilities that are needed to provision/attach/mount/unmount/detach/delete volumes in *pods* instead of running them on *the host*. The host can be a minimal Linux distribution without tools to create e.g. Ceph RBD or mount GlusterFS volumes.
+
+## Secondary objectives
+These are not requirements per se, just things to consider before drawing the final design.
+* CNCF designs Container Storage Interface (CSI). So far, this CSI expects that "volume plugins" on each host are long-running processes with a fixed gRPC API. We should aim the same direction, hoping to switch to CSI when it's ready. In other words, there should be one long-running container for a volume plugin that serves all volumes of given type on a host.
+* We should try to avoid complicated configuration. The system should work out of the box or with very limited configuration.
+
+## Terminology
+
+**Mount utilities** for a volume plugin are all tools that are necessary to use a volume plugin. This includes not only utilities needed to *mount* the filesystem (e.g. `mount.glusterfs` for Gluster), but also utilities needed to attach, detach, provision or delete the volume, such as `/usr/bin/rbd` for Ceph RBD.
+
+## User story
+Admin wants to run Kubernetes on a distro that does not ship `mount.glusterfs` that's needed for GlusterFS volumes.
+1. Admin installs and runs Kubernetes in any way.
+1. Admin deploys a DaemonSet that runs a pod with `mount.glusterfs` on each node. In future, this could be done by installer.
+1. User creates a pod that uses a GlusterFS volume. Kubelet finds a pod with mount utilities on the node and uses it to mount the volume instead of expecting that `mount.glusterfs` is available on the host.
+
+- User does not need to configure anything and sees the pod Running as usual.
+- Admin just needs to deploy the DaemonSet.
+- It's quite hard to update the DaemonSet, see below.
+
+## Alternatives
+### Sidecar containers
+We considered this user story:
+* Admin installs Kubernetes.
+* Admin configures Kubernetes to use sidecar container with template XXX for glusterfs mount/unmount operations and pod with template YYY for glusterfs provision/attach/detach/delete operations. These templates would be yaml files stored somewhere.
+* User creates a pod that uses a GlusterFS volume. Kubelet find a sidecar template for gluster, injects it into the pod and runs it before any mount operation. It then uses `docker exec mount <what> <where>` to mount Gluster volumes for the pod. After that, it starts init containers and the "real" pod containers.
+* User deletes the pod. Kubelet kills all "real" containers in the pod and uses the sidecar container to unmount gluster volumes. Finally, it kills the sidecar container.
+
+-> User does not need to configure anything and sees the pod Running as usual.
+-> Admin needs to set up the templates.
+
+Similarly, when attaching/detaching a volume, attach/detach controller would spawn a pod on a random node and the controller would then use `kubectl exec <the pod> <any attach/detach utility>` to attach/detach the volume. E.g. Ceph RBD volume plugin needs to execute things during attach/detach. After the volume is attached, the controller would kill the pod.
+
+Advantages:
+* It's probably easier to update the templates than update the DaemonSet.
+
+Drawbacks:
+* Admin needs to store the templates somewhere. Where?
+* Short-living processes instead of long-running ones that would mimic CSI (so we could catch bugs early or even redesign CSI).
+* Needs some refactoring in kubelet - now kubelet mounts everything and then starts containers. We would need kubelet to start some container(s) first, then mount, then run the rest. This is probably possible, but needs better analysis (and I got lost in kubelet...)
+
+### Infrastructure containers
+
+Mount utilities could be also part of infrastructure container that holds network namespace (when using Docker). Now it's typically simple `pause` container that does not do anything, it could hold mount utilities too.
+
+Advantages:
+* Easy to set up
+* No extra container running
+
+Disadvantages:
+* One container for all mount utilities. Admin needs to make a single container that holds utilities for e.g. both gluster and nfs and whatnot.
+* Needs some refactoring in kubelet - now kubelet mounts everything and then starts containers. We would need kubelet to start some container(s) first, then mount, then run the rest. This is probably possible, but needs better analysis (and I got lost in kubelet...)
+* Short-living processes instead of long-running ones that would mimic CSI (so we could catch bugs early or even redesign CSI).
+* Infrastructure container is implementation detail and CRI does not even allow executing binaries in it.
+
+**We've decided to go with long running DaemonSet pod as described below.**
+
+## Design
+
+* Pod with mount utilities puts a registration JSON file into `/var/lib/kubelet/plugin-containers/<plugin name>.json` on the host with name of the container where mount utilities should be executed:
+  ```json
+  {
+      "podNamespace": "kubernetes-storage",
+      "podName": "gluster-daemon-set-xtzwv",
+      "podUID": "5d1942bd-7358-40e8-9547-a04345c85be9",
+      "containerName": "gluster"
+  }
+  ```
+  * Pod UID is used to avoid situation when a pod with mount utilities is terminated and leaves its registration file on the host. Kubelet should not assume that newly started pod with the same namespace+name has the same mount utilities.
+  * All slashes in `<plugin name>` must be replaced with tilde, e.g. `kubernetes.io~glusterfs.json`.
+  * Creating the file must be atomic so kubelet cannot accidentally read partly written file.
+
+ * All volume plugins use `VolumeHost.GetExec` to get the right exec interface when running their utilities.
+
+ * Kubelet's implementation of `VolumeHost.GetExec` looks at `/var/lib/kubelet/plugin-containers/<plugin name>.json` if it has a container for given volume plugin.
+   * If the file exists and referred container is running, it returns `Exec` interface implementation that leads to CRI's `ExecSync` into the container (i.e. `docker exec <container> ...`)
+   * If the file does not exist or referred container is not running, it returns `Exec` interface implementation that leads to `os.Exec`. This way, pods do not need to remove the registration file when they're terminated.
+   * Kubelet does not cache content of `plugin-containers/`, one extra `open()`/`read()` with each exec won't harm and it makes Kubelet more robust to changes in the directory.
+
+* In future, this registration of volume plugin pods should be replaced by a gRPC interface based on Device Plugin registration.
+
+## Requirements on DaemonSets with mount utilities
+These are rules that need to be followed by DaemonSet authors:
+* One DaemonSet can serve mount utilities for one or more volume plugins. We expect that one volume plugin per DaemonSet will be the most popular choice.
+* One DaemonSet must provide *all* utilities that are needed to provision, attach, mount, unmount, detach and delete a volume for a volume plugin, including `mkfs` and `fsck` utilities if they're needed.
+    * E.g. `mkfs.ext4` is likely to be available on all hosts, but a pod with mount utilities should not depend on that nor use it.
+    * Kernel modules should be available in the pod with mount utilities too. "Available" does not imply that they need to be shipped in a container, we expect that binding `/lib/modules` from host to `/lib/modules` in the pod will be enough for all modules that are needed by Kubernetes internal volume plugins (all distros I checked incl. the "minimal" ones ship scsi.ko, rbd.ko, nfs.ko and fuse). This will allow future flex volumes ship vendor-specific kernel modules. It's up to the vendor to ensure that any kernel module matches the kernel on the host.
+    * The only exception is udev (or similar device manager). Only one udev can run on a system, therefore it should run on the host. If a volume plugin needs to talk to udev (e.g. by calling `udevadm trigger`), they must do it on the host and not in a container with mount utilities.
+* It is expected that these daemon sets will run privileged pods that will see host's `/proc`, `/dev`, `/sys`, `/var/lib/kubelet` and such. Especially `/var/lib/kubelet` must be mounted with shared mount propagation so kubelet can see mounts created by the pods.
+* The pods with mount utilities should run some simple init as PID 1 that reaps zombies of potential fuse daemons.
+* The pods with mount utilities must put a file into `/var/lib/kubelet/plugin-containers/<plugin name>.json` for each volume plugin it supports. It should overwrite any existing file - it's probably leftover from older pod.
+   * Admin is responsible to run only one pod with utilities for one volume plugin on a single host. When two pods for say GlusterFS are scheduled on the same node they will overwrite the registration file of each other.
+   * Downward API can be used to get pod's name and namespace.
+   * Root privileges (or CAP_DAC_OVERRIDE) are needed to write to `/var/lib/kubelet/plugin-containers/`.
+
+To sum it up, it's just a daemon set that spawns privileged pods, running a simple init and registering itself into Kubernetes by placing a file into well-known location.
+
+**Note**: It may be quite difficult to create a pod that see's host's `/dev` and `/sys`, contains necessary kernel modules, does the initialization right and reaps zombies. We're going to provide a template with all this.
+
+### Upgrade
+Upgrade of DaemonSets with pods with fuse-based mount utilities needs to be done node by node and with extra care. Killing a pod with fuse daemon(s) inside will un-mount all volumes that are used by other pods on the host and may result in data loss.
+
+In order to update the fuse-based DaemonSet (=GlusterFS or CephFS), admin must do for every node:
+* Mark the node as tainted. Only the pod with mount utilities can tolerate the taint, all other pods are evicted. As result, all volumes are unmounted and detached.
+* Update the pod.
+* Remove the taint.
+
+Is there a way how to make it with DaemonSet rolling update? Is there any other way how to do this upgrade better?
+
+### Containerized kubelet
+
+Kubelet should behave the same when it runs inside a container:
+* Use `os.Exec` to run mount utilities inside its own container when no pod with mount utilities is registered. This is current behavior, `mkfs.ext4`, `lsblk`, `rbd` and such are executed in context of the kubelet's container now.
+* Use `nsenter <host> mount` to mount things when no pod with mount utilities is registered. Again, this is current behavior.
+* Use CRI's `ExecSync` to execute both utilities and the final `mount` when a pod with mount utilities is registered so everything is executed in this pod.
+
+## Open items
+
+* How will controller-manager talk to pods with mount utilities?
+
+  1. Mount pods expose a gRPC service.
+      * controller-manager must be configured with the service namespace + name.
+      * Some authentication must be implemented (=additional configuration of certificates and whatnot).
+      * -> seems to be complicated.
+
+  2. Mount pods run in a dedicated namespace and have labels that tell which volume plugins they can handle.
+      * controller manager scans a namespace with a labelselector and does `kubectl exec <pod>` to execute anything in the pod.
+      * Needs configuration of the namespace.
+      * Admin must make sure that nothing else can run in the namespace (e.g. rogue pods that would steal volumes).
+      * Admin/installer must configure access to the namespace so only pv-controller and attach-detach-controller can do `exec` there.
+
+  3. We allow pods to run on hosts that run controller-manager.
+
+      * Usual socket in `/var/lib/kubelet/plugin-sockets` will work.
+      * Can it work on GKE?
+
+We do not implement any of these approaches, as we expect that most volume plugins are going to be moved to CSI soon-ish. The only affected volume plugins are:
+
+* Ceph dynamic provisioning - we can use external provisioner during tests.
+* Flex - it has its own dynamic registration of flex drivers.
+
+## Implementation notes
+As we expect that most volume plugins are going to be moved to CSI soon, all implementation of this proposal will be guarded by alpha feature gate "MountContainers" which is never going leave alpha. Whole implementation of this proposal is going to be removed when the plugins are fully moved to CSI.
+
+Corresponding e2e tests for internal volume plugins will initially run only when with the feature gate is enabled and they will continue running when we move the volume plugins to CSI to ensure we won't introduce regressions.


contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Controller
+* There will be new parameter to kube-controller-manager and kubelet:
+    * `--experimental-mount-namespace`, which specifies a dedicated namespace where all pods with mount utilities reside.
+    * `--experimental-mount-plugins`, which contains comma-separated list of all volume plugins that should run their utilities in pods instead on the host. The list must include also all flex volume drivers. Without this option, controllers and kubelet would not know if a plugin should use a pod with mount utilites or directly host, especially on startup when the daemon set may not yet be fully deployed on all nodes.  * If so, it finds a pod in the dedicated namespace with label `mount.kubernetes.io/foo=true` (or `mount.kubernetes.io/flexvolume/foo=true` for flex volumes) and calls the volume plugin with `VolumeExec` pointing to the pod. All utilities that are executed by the volume plugin for attach/detach/provision/delete are executed in the pod as `kubectl exec <pod> xxx` (of course, we'll use clientset interface instead of executing `kubectl`).
```
jsafrane: I admit I don't like this `experimental-mount-plugins` cmdline option. Can anyone find a bulletproof way, how kubelet / a controller can reliably find if a volume plugin should execute its utilities on the host or in a pod? Especially, when kubelet starts in a fresh cluster, the pod with mount utilities may not be running yet and kubelet *must* know if it should wait for it or not. Kubelet must not try to run the utilities on the host, because there may be wrong version or wrong configuration of the utilities.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Secondary objectives
+These are not requirements per se, just things to consider before drawing the final design.
+* CNCF designs Container Storage Interface (CSI). So far, this CSI expects that "volume plugins" on each host are long-running processes with a fixed (gRPC?) API. We should aim the same direction, using exec instead of gRPC, hoping to switch to CSI when it's ready. In other words, there should be one long-running container for a volume plugin that serves all volumes of given type on a host.
+* We should try to avoid complicated configuration. The system should work out of the box or with very limited configuration.
+
+## User story
+Admin wants to run Kubernetes on a distro that does not ship `mount.glusterfs` that's needed for GlusterFS volumes.
+1. Admin installs Kubernetes in any way.
+2. Admin modifies controller-manager and kubelet command line to include `--experimental-mount-namespace=foo` and `--experimental-mount-plugins=kubernetes.io/glusterfs` so Kubernetes knows what volume plugins should use utilities in pods and in which namespace to find these pods. This should be probably part of installation in the future.
```
vishh: Why not leave discovery of appropriate mount plugins be a vendor specific requirement. Kubelet execs a script or binary that knows which container or service to talk to for each type of storage.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Alternatives
+### Sidecar containers
+We considered this user story:
+* Admin installs Kubernetes.
+* Admin configures Kubernetes to use sidecar container with template XXX for glusterfs mount/unmount operations and pod with template YYY for glusterfs provision/attach/detach/delete operations. These templates would be yaml files stored somewhere.
+* User creates a pod that uses a GlusterFS volume. Kubelet find a sidecar template for gluster, injects it into the pod and runs it before any mount operation. It then uses `docker exec mount <what> <where>` to mount Gluster volumes for the pod. After that, it starts init containers and the "real" pod containers.
+* User deletes the pod. Kubelet kills all "real" containers in the pod and uses the sidecar container to unmount gluster volumes. Finally, it kills the sidecar container.
+
+-> User does not need to configure anything and sees the pod Running as usual.
```
vishh: Have you considered packaging mount scripts into the infra container? Kubelet will then have to exec into the infra container to mount a volume. This will alter the pod lifecycle in the kubelet where volumes are now setup prior to starting a pod.
The advantage is that all storage related processes belonging to a pod is contained with a pod's boundary and it's lifecycle is tied to that of the pod.

jsafrane: 1) rkt does not use infrastructure container, they "hold" network NS in other way. 
2) using long-running pods better reflects CSI, as it will run one long-running process on each node. @saad-ali, can you confirm?

I will add a note about it to the proposal

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,109 @@
...
+In order to update the DaemonSet, admin must do for every node:
+* Mark the node as tainted. Only the pod with mount utilities can tolerate the taint, all other pods are evicted. As result, all volumes are unmounted and detached.
+* Update the pod.
+* Remove the taint.
+
+Is there a way how to make it with DaemonSet rolling update? Is there any other way how to do this upgrade better?
+
+
+## Implementation notes
+Flex volumes won't be changed in alpha implementation of this PR. Flex volumes will still need their utilities (and binaries in /usr/libexec/kubernetes) on all hosts.
```
jingxu97: Is there some reason for this flex volume note?

gnufied: As mentioned above we are hoping that flex utils will eventually be moved to pods as well - with label `mount.kubernetes.io/flexvolume/foo=true`  but we are not considering that as part of alpha implementation.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: You then need to deploy the script to all nodes and masters and that's exactly I'd like to avoid. Otherwise I can deploy the mount utilities directly there, right? I see GCI and Atomic Host and CoreOS as mostly immutable images with some configuration in /etc that just starts Kubernetes with the right options (and even that is complicated enough!)

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: Note added. To be hones, infrastructure container looks compelling to me if we did not want to mimic long-running processes of CSI.

contributors/design-proposals/containerized-mounter-pod.md
vishh: All these Container optimizes distros do have some writable stateful partitions. That would be necessary for other parts of the system like CNI. How does this align with CSI?

jsafrane: CSI does not dictate any specific way how its drivers will run. @saad-ali expects they will run as a daemonset.

With CNI approach (a script in /opt/cni/bin), we would need a way how to deploy it on a master. This is OK for OpenShift, would it be fine for GKE, where user does not have access to masters so they can't install a attach/detach/provision/delete script for their favorite storage? And how would it talk to Kubernetes to find the right pod where to do the attaching/provisioning?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Secondary objectives
+These are not requirements per se, just things to consider before drawing the final design.
+* CNCF designs Container Storage Interface (CSI). So far, this CSI expects that "volume plugins" on each host are long-running processes with a fixed (gRPC?) API. We should aim the same direction, using exec instead of gRPC, hoping to switch to CSI when it's ready. In other words, there should be one long-running container for a volume plugin that serves all volumes of given type on a host.
+* We should try to avoid complicated configuration. The system should work out of the box or with very limited configuration.
+
+## User story
+Admin wants to run Kubernetes on a distro that does not ship `mount.glusterfs` that's needed for GlusterFS volumes.
+1. Admin installs Kubernetes in any way.
+2. Admin modifies controller-manager and kubelet command line to include `--experimental-mount-namespace=foo` and `--experimental-mount-plugins=kubernetes.io/glusterfs` so Kubernetes knows what volume plugins should use utilities in pods and in which namespace to find these pods. This should be probably part of installation in the future.
```
yujuhong: I think the discovery process is far from ideal. One would need to enumerate a list of plugins via a kubelet flag (which is static) before kubelet starts and before the (dynamic) DaemonSet pods are created. Any change to the plugin list will require restarting kubelet. Can we try finding other discovery methods?

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: PV controller on master needs a way how to execute Ceph utilities when creating a volume +  attach/detach controller uses the same utilities to attach/detach the volume. Now it uses plain exec. When we move Ceph utilities from master somewhere else we need to tell controller-manager *where* the utilities are.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: I agree that the discovery is not ideal. What are other options? AFAIK there is no magic way how to configure kubelet dynamically. Is it possible to have a config object somewhere where *all* kubelets and controller-manager would reliably get list of volume plugins that are supposed to be containerized?

The list is needed only at startup, where kubelet gets its first pods from scheduler - a pod that uses e.g. a gluster volume may be scheduled before daemon set for gluster is started or daemon set controller spawns a pod on the node. With the list kubelet knows that it should wait. Without the list, it blindly tries to mount the Gluster volume on the host, which is likely to fail with something as ugly as `wrong fs type, bad option, bad superblock on 192.168.0.1:/foo missing codepage or helper program, or other error`. mount stderr and exit codes are not helpful at all here.

When all daemon sets are up and running we don't need `--experimental-mount-plugins` at all and dynamic discovery works.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,122 @@
...
+Mount utilities could be also part of infrastructure container that holds network namespace (when using Docker). Now it's typically simple `pause` container that does not do anything, it could hold mount utilities too.
+
+Advantages:
+* Easy to set up
+* No extra container running
+
+Disadvantages:
+* One container for all mount utilities. Admin needs to make a single container that holds utilities for e.g. both gluster and nfs and whatnot.
+* Needs some refactoring in kubelet - now kubelet mounts everything and then starts containers. We would need kubelet to start some container(s) first, then mount, then run the rest. This is probably possible, but needs better analysis (and I got lost in kubelet...)
+* Short-living processes instead of long-running ones that would mimic CSI.
```
jsafrane: The advantage is that it mimics our current design of CSI and we can catch bugs or even discover that it's not ideal before CSI is standardized.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: Note added (and thanks for spotting this)

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,122 @@
...
+
+Mount utilities could be also part of infrastructure container that holds network namespace (when using Docker). Now it's typically simple `pause` container that does not do anything, it could hold mount utilities too.
+
+Advantages:
+* Easy to set up
+* No extra container running
+
+Disadvantages:
+* One container for all mount utilities. Admin needs to make a single container that holds utilities for e.g. both gluster and nfs and whatnot.
+* Needs some refactoring in kubelet - now kubelet mounts everything and then starts containers. We would need kubelet to start some container(s) first, then mount, then run the rest. This is probably possible, but needs better analysis (and I got lost in kubelet...)
```
jsafrane: Added to drawbacks.

jsafrane:
I updated the proposal with current development:

- Added Terminology section to clear some confusion
- Removed `--experimental-mount-plugins` option
- During alpha (hopefuly 1.7), this feature must be explicitly enabled using `kubelet --experimental-mount-namespace=foo` so we don't break working clusters accidentally. This may change during beta/GA!
- During alpha, no controller changes will be done, as it is only Ceph RBD provisioner who needs to execute stuff on master. I may implement it if time permits, I am just not sure...

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: I removed `--experimental-mount-plugins` for now, but it will behave exactly as I described in the previous comment - weird errors may appear in pod events during kubelet startup before a pod with mount utilities is scheduled and started.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Controller
+* There will be new parameter to kube-controller-manager and kubelet:
+    * `--experimental-mount-namespace`, which specifies a dedicated namespace where all pods with mount utilities reside. It would default to `kube-mount`.
+* Whenever PV or attach/detach controller needs to call a volume plugin, it looks for *any* running pod in the specified namespace with label `mount.kubernetes.io/foo=true` (or `mount.kubernetes.io/flexvolume/foo=true` for flex volumes) and calls the volume plugin so it all mount utilities are executed as `kubectl exec <pod> xxx` (of course, we'll use clientset interface instead of executing `kubectl`).
+* If such pod does not exist, it executes the mount utilities on the host as usual.
+* During alpha, no controller-manager changes will be done. That means that Ceph RBD provisioner will still require `/usr/bin/rbd` installed on the master. All other volume plugins will work without any problem, as they don't execute any utility when attaching/detaching/provisioning/deleting a volume.
```
kfox1111: the rbd provisioner has been pulled out to here:
https://github.com/kubernetes-incubator/external-storage/tree/master/ceph/rbd

so the container can be built with the right ceph version already.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,145 @@
...
+## Prerequisites
+[HostPath volume propagation](propagation.md) must be implemented first.
+
+## Requirements on DaemonSets with mount utilities
+These are rules that need to be followed by DaemonSet authors:
+* One DaemonSet can serve mount utilities for one or more volume plugins. We expect that one volume plugin per DaemonSet will be the most popular choice.
+* One DaemonSet must provide *all* utilities that are needed to provision, attach, mount, unmount, detach and delete a volume for one volume plugin, including `mkfs` and `fsck` utilities if they're needed.
+    * E.g. `mkfs.ext4` is likely to be available on all hosts, but a pod with mount utilities should not depend on that nor use it.
+    * The only exception are kernel modules. They are not portable across distros and they should be on the host.
+* It is expected that these daemon sets will run privileged pods that will see host's `/proc`, `/dev`, `/sys`, `/var/lib/kubelet` and such. Especially `/var/lib/kubelet` must be mounted with shared mount propagation so kubelet can see mounts created by the pods.
```
lucab: Are mounts constrained to be performed under `/var/lib/kubelet`? If so, this seems to be a contract detail between controller/kubelet/daemonset that should be mentioned.

jsafrane: No, any directory can be shared. It's up to the system admin + author of the privileged pods to make sure it can be shared (i.e. it's on a mount with shared mount propagation) and it's safe to share (e.g. systemd inside a container does not like /sys/fs/cgroup to be shared to the host, I don't remember exact error message but it simply won't start).

jsafrane:
I updated the proposal with latest discussion om sig-node and with @tallclair.

- Kubelet won't use `docker exec <pod> mount <what> <where>` to mount things in a pod, it will talk to a gRPC service running in the pod instead (following docker shim example).
  - Allows for much easier discovery of mount pods by kubelet, no magic namespaces or labels.
  - Opens a question how controller-manager will talk to these pods, see open items at the bottom.

This is basically a new proposal and needs a complete re-review. I left the original proposal as a separate commit so we can roll back easily.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+- User does not need to configure anything and sees the pod Running as usual.
+- Admin just needs to deploy the DaemonSet.
+- It's quite hard to update the DaemonSet, see below.
+
+## Alternatives
+### Sidecar containers
+We considered this user story:
+* Admin installs Kubernetes.
+* Admin configures Kubernetes to use sidecar container with template XXX for glusterfs mount/unmount operations and pod with template YYY for glusterfs provision/attach/detach/delete operations. These templates would be yaml files stored somewhere.
+* User creates a pod that uses a GlusterFS volume. Kubelet find a sidecar template for gluster, injects it into the pod and runs it before any mount operation. It then uses `docker exec mount <what> <where>` to mount Gluster volumes for the pod. After that, it starts init containers and the "real" pod containers.
```
humblec: Do we need to extend the pod spec for doing this( sidecar template injection) operation or can it be done by existing pod spec or  achievable by kube jobs or similar mechanism?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+* User deletes the pod. Kubelet kills all "real" containers in the pod and uses the sidecar container to unmount gluster volumes. Finally, it kills the sidecar container.
+
+-> User does not need to configure anything and sees the pod Running as usual.
+-> Admin needs to set up the templates.
+
+Similarly, when attaching/detaching a volume, attach/detach controller would spawn a pod on a random node and the controller would then use `kubectl exec <the pod> <any attach/detach utility>` to attach/detach the volume. E.g. Ceph RBD volume plugin needs to execute things during attach/detach. After the volume is attached, the controller would kill the pod.
```
humblec: Can it be random node ? or it should be the same node where the pod is getting scheduled ?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+
+Advantages:
+* It's probably easier to update the templates than update the DaemonSet.
```
humblec: I have one doubt here, how we are going to control the version of required mount utils ? for eg# if the mount utils need to be of a particular version can we specify that in the template ? Does that also mean there can be more than one sidecar container if user wish to ?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+
+Drawbacks:
+* Admin needs to store the templates somewhere. Where?
```
humblec: Cant we make use of configmap or similar mechanism for this templates ? Just a thought 👍

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+## Prerequisites
+[HostPath volume propagation](propagation.md) must be implemented first.
+
+## Requirements on DaemonSets with mount utilities
+These are rules that need to be followed by DaemonSet authors:
+* One DaemonSet can serve mount utilities for one or more volume plugins. We expect that one volume plugin per DaemonSet will be the most popular choice.
+* One DaemonSet must provide *all* utilities that are needed to provision, attach, mount, unmount, detach and delete a volume for a volume plugin, including `mkfs` and `fsck` utilities if they're needed.
+    * E.g. `mkfs.ext4` is likely to be available on all hosts, but a pod with mount utilities should not depend on that nor use it.
+    * The only exception are kernel modules. They are not portable across distros and they *should* be on the host.
+* It is expected that these daemon sets will run privileged pods that will see host's `/proc`, `/dev`, `/sys`, `/var/lib/kubelet` and such. Especially `/var/lib/kubelet` must be mounted with shared mount propagation so kubelet can see mounts created by the pods.
```
humblec: if we have one Daemonset per volume plugin and if we share `/dev` amoung these containers , there is a risk or security concern, Isnt it ?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+       If this startup adjustment fails, kubelet refuses to start with a clear error message.
+* Kubelet does not do any other special check. We assume that kubelet either runs on the host and in the same mount namespace as docker and the host, or it runs in a container that sees /var/lib/kubelet as it is on the host and with shared mount propagation.
+
+This ensures that kubelet runs out of the box on any distro without any configuration done by the cluster admin.
+
+### Volume plugins
+* All volume plugins need to be updated to use a new `mount.Exec` interface to call external utilities like `mount`, `mkfs`, `rbd lock` and such. Implementation of the interface will be provided by caller and will lead either to simple `os.exec` on the host or a gRPC call to a socket in `/var/lib/kubelet/plugin-sockets/` directory.
+
+### Controllers
+TODO: how will controller-manager talk to a remote pod? It's relatively easy to do something like `kubectl exec <mount pod>` from controller-manager, however it's harder to *discover* the right pod.
```
humblec: may be we could make use of labelling/selector mechanism based on the pod content.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Upgrade
+Upgrade of the DaemonSet with pods with mount utilities needs to be done node by node and with extra care. The pods may run fuse daemons and killing such pod with glusterfs fuse daemon would kill all pods that use glusterfs on the same node.
+
+In order to update the DaemonSet, admin must do for every node:
+* Mark the node as tainted. Only the pod with mount utilities can tolerate the taint, all other pods are evicted. As result, all volumes are unmounted and detached.
+* Update the pod.
+* Remove the taint.
+
+Is there a way how to make it with DaemonSet rolling update? Is there any other way how to do this upgrade better?
```
humblec: iirc, rolling update is yet to come in DS , need to check the current status though. However there is an option called `--cascade=false` and possible to do a rolling update by manually or in a scripted way, not sure is that you are looking for.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
+# Containerized mounter using volume utilities in pods
+
+## Goal
+Kubernetes should be able to run all utilities that are needed to provision/attach/mount/unmount/detach/delete volumes in *pods* instead of running them on *the host*. The host can be a minimal Linux distribution without tools to create e.g. Ceph RBD or mount GlusterFS volumes.
+
+## Secondary objectives
```
tallclair: What are the goals around adding (or removing) volume plugins dynamically? In other words, do you expect the pods serving the volume plugins to be deployed at cluster creation time, or at a later time? How about removing plugins?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+
+**We've decided to go with long running DaemonSet pod as described below.**
+
+## Prerequisites
+[HostPath volume propagation](propagation.md) must be implemented first.
+
+## Requirements on DaemonSets with mount utilities
+These are rules that need to be followed by DaemonSet authors:
+* One DaemonSet can serve mount utilities for one or more volume plugins. We expect that one volume plugin per DaemonSet will be the most popular choice.
+* One DaemonSet must provide *all* utilities that are needed to provision, attach, mount, unmount, detach and delete a volume for a volume plugin, including `mkfs` and `fsck` utilities if they're needed.
```
tallclair: > that are needed to provision

Provisioning is a "cluster level" operation, and is handled by the volume controller rather than the Kubelet, right? In that case, I don't think they need to be handled by the same pod. In practice its probably often the same utilities that handle both, but I don't think it should be a hard requirement.

contributors/design-proposals/containerized-mounter-pod.md
tallclair: > Especially `/var/lib/kubelet` must be mounted with shared mount propagation so kubelet can see mounts created by the pods.

This only applies if the Kubelet is running in a container, right? Also, it needs slave mount propogation, not shared, right? (Pardon my ignorance of this subject)

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+    * E.g. `mkfs.ext4` is likely to be available on all hosts, but a pod with mount utilities should not depend on that nor use it.
+    * The only exception are kernel modules. They are not portable across distros and they *should* be on the host.
+* It is expected that these daemon sets will run privileged pods that will see host's `/proc`, `/dev`, `/sys`, `/var/lib/kubelet` and such. Especially `/var/lib/kubelet` must be mounted with shared mount propagation so kubelet can see mounts created by the pods.
+* The pods with mount utilities should run some simple init as PID 1 that reaps zombies of potential fuse daemons.
```
tallclair: > that reaps zombies of potential fuse daemons.

What does this mean? I believe the zombie process issue was fixed in 1.6 (https://github.com/kubernetes/kubernetes/pull/36853)

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+* The pods with mount utilities run a daemon with gRPC server that implements `VolumExecService` defined below.
```
tallclair: > `VolumExecService`

nit: I'd prefer `VolumePluginService`, or some other variation. I think `Exec` in this case is a bit unclear.

contributors/design-proposals/containerized-mounter-pod.md
tallclair: What are the tradeoffs of using exec vs. http to serve this? My hunch is that this should just be a service model, with a Kubernetes service that provides the volume plugin (how the controller manager identifies the service could be up for debate - predefined name? labels? namespace?). The auth{n/z} is a bit more complicated with that model though.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Kubelet
+* When kubelet talks to a volume plugin, it looks for a socket named `/var/lib/kubelet/plugin-sockets/<plugin-name>`. This allows for easier discovery of flex volume drivers - probe in https://github.com/kubernetes/community/pull/833 needs to scan `/var/lib/kubelet/plugin-sockets/` too and find sockets in any new subdirectories.
+  * If the socket does not exist, kubelet gives the volume plugin plain `os.Exec` as implementation of `mount.Exec` interface and all mount utilities are executed on the host.
+  * If the socket exists, kubelet gives the volume plugin `GRPCExec` as implementation of `mount.Exec` and all mount utilities are executed via gRPC on the socket which presumably leads to a pod with mount utilities running a gRPC server.
+
+As consequence, kubelet will try to run mount utilities on the host when it starts and has not received pods with mount utilities yet (and thus `/var/lib/kubelet/plugin-sockets/` is empty). This is likely to fails with a cryptic error:
```
tallclair: I don't understand this paragraph. Is this the design, or the current state? Why does the Kubelet exercise every volume plugin before receiving pods (and what arguments will it pass to the utilities)?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+```
+mount: wrong fs type, bad option, bad superblock on 192.168.0.1:/test_vol,
+       missing codepage or helper program, or other error
+```
+
+Kubelet will periodically retry mounting the volume and it will eventually succeed when pod with mount utilities is scheduled and running on the node.
+
+### gRPC API
+
+`VolumeExecService` is a simple gRPC service that allows to execute anything via gRPC:
```
tallclair: Is there a CSI API proposal out? Does this align with that? It might be worth using the CSI API in it's current state, if it's sufficient.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+
+```protobuf
+service VolumeExecService {
+    // Exec executes a command and returns its output.
+    rpc Exec(ExecRequest) returns (ExecResponse) {}
+}
+
+message ExecRequest {
+    // Command to execute
+    string cmd = 1;
```
tallclair: This should be abstracted so that the Kubelet doesn't need to understand the specifics of the volume type. I believe this is what the volume interfaces defined in https://github.com/kubernetes/kubernetes/blob/4a73f19aed1f95b3fde1177074aee2a8bec1196e/pkg/volume/volume.go do? In that case, this API should probably mirror those interfaces.

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+  * `hack/update-generated-runtime.sh` will be updated to generate go files for this API.
+
+    * Should it be renamed to `update-generated-grpc-apis.sh`?
+
+* Kubernetes will ship a daemon with server implementation of this API in `cmd/volume-exec`. This implementation simply calls `os.Exec` for each `ExecRequest` it gets and returns the right response.
+
+  * Authors of container images with mount utilities can then add this `volume-exec` daemon to their image, they don't need to care about anything else.
+
+### Upgrade
+Upgrade of the DaemonSet with pods with mount utilities needs to be done node by node and with extra care. The pods may run fuse daemons and killing such pod with glusterfs fuse daemon would kill all pods that use glusterfs on the same node.
```
tallclair: Would it kill the pods, or just cause IO errors?

contributors/design-proposals/containerized-mounter-pod.md
tallclair: DaemonSets support rolling update as of 1.6 (https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/)

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,200 @@
...
+
+  * `hack/update-generated-runtime.sh` will be updated to generate go files for this API.
+
+    * Should it be renamed to `update-generated-grpc-apis.sh`?
+
+* Kubernetes will ship a daemon with server implementation of this API in `cmd/volume-exec`. This implementation simply calls `os.Exec` for each `ExecRequest` it gets and returns the right response.
+
+  * Authors of container images with mount utilities can then add this `volume-exec` daemon to their image, they don't need to care about anything else.
+
+### Upgrade
```
tallclair: What happens if the kubelet can't reach the pod serving a volume plugin (either do to update, or some other error) when a pod with a volume is deleted? Will the Kubelet keep retrying until it is able to unmount the volume? What are the implications of being unable to unmount the volume?

contributors/design-proposals/containerized-mounter-pod.md
kfox1111: I think it may be driver specific. For some drivers, its probably best to exec into the container on the host going to have the volume?

For my k8s systems, I tend to have user provided code running in containers. I usually segregate these into differently labeled nodes then the control plane. In this configuration, the container doing the reach back to, say, openstack to move around volumes from vm to vm should never run on the user reachable nodes, as access to the secret for volume manipulation would be really bad. with k8s 1.7+, the secret is inaccessable to nodes that don't have a pod referencing the secret. So targeted exec would be much much better for that use case.

contributors/design-proposals/containerized-mounter-pod.md
kfox1111: not sure its any worse then what is there today.

kfox1111: no, shared is needed.

slave:
(u)mount events on the host show up in containers. events on the containers dont affect the host.

shared:
(u)mount events that are initiated from either the host or the container show up on the other side.

If you want an (u)mount event in the mount utility container to show up to kubelet, it needs shared.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: Volume plugins are not a real plugins, they're hardcoded in Kubernetes.

It does not really matter when the pods with mount utilities are deployed - I would expect that they should be deployed during Kubernetes installation because cluster admin plans storage ahead (e.g. has existing NFS server) , however I can imagine that cluster admin could deploy pods for Gluster volumes later as the NFS server becomes full or so. 

The only exception are flex plugin drivers. In 1.7, they needed to be installed before kubelet and controller-manager started. In #833 we're trying to change it to a more dynamic model, where flex drivers can be added/removed dynamically and this proposal could be easily extended with flex drivers running in pods. So admins could dynamically install/remove flex drivers running in pods. Again. I would expect that this would be mostly done during installation of a cluster. And #833 is better place to discuss it.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: Yes, technically it does not need to be the same pod.

On the other hands, the only internal volume plugin that needs to execute something during provisioning or attach/detach (i.e. initiated by controller-manager) is Ceph RBD that needs /usr/bin/rbd. The same utility is then needed by kubelet to finish attachment of the device.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: > This only applies if the Kubelet is running in a container, right

No. Kubelet running on the host must see mounts mounted by a pod. Therefore we need shared mount propagation from the pod  to the host. With slave propagation in the pod the mount would be visible only in the pod and not on the host.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: @yujuhong says in https://github.com/kubernetes/community/pull/589#discussion_r116596309 that infrastructure pod ("pause") is implementation detail of docker integration and other container engines may not use it.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: I am trying to say that when kubelet starts and it does not run any pods yet (it's a new node), it may receive pods in any order. It can happen that an application pod that requires say GlusterFS volume is placed there *before* our DaemonSet with mount utilities for Gluster is scheduled. 

Our fresh kubelet then looks at /var/lib/kubelet/plugin-sockets, does not see GlusterFS socket there and tries to run mount.glusterfs on the host. This will fail with confusing error message.

Only after a while DaemonSet sees a new node, puts a pod with GlusterFS utilities there and this pod creates a socket in /var/lib/kubelet/plugin-sockets.

Kubelet tries to mount volumes pod periodically, finds the socket there and runs mount.glusterfs in the pod  and everything self-heals. Still, the pod startup is delayed (that's IMO OK) and user can see some errors (which may cause some support calls).

In some old version of this proposal I suggested new kubelet and controller-manager parameter with list of plugins for which kubelet or controller-manager **must** wait for a pod with mount utilities and never try to run them on the host. This was turned down as too complicated.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: CSI is too complicated. Also, this would require a completely new implementation of at least gluster, nfs, CephFS, Ceph RBD, git volume, iSCSI, FC and ScaleIO volume plugins which is IMO too much. Keeping the plugins as they are, just using an interface that would defer `os.Exec` to a pod where appropriate is IMO much simpler and without risk of breaking existing (and tested!) volume plugins.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: Again, that would require me to rewrite the volume plugins. Volume plugins need e.g. access to CloudProvider or SecretManager, I can't put them into pods easily. And this pod would have access to all Kubernetes secrets...

Whole idea of ExecRequest/Response is to take existing and tested volume plugins and replace all `os.Exec` calls with `<abstract exec interface>.Exec`. Kubelet would provide the right interface implementation, leading to `os.Exec` or gRPC. No big changes in the volume plugins*, simple changes in Kubelet, one common VolumeExec server daemon for all pods with mount utilities.

It does not leak any specific volume knowledge to kubelet / controller-manager. It's dumb exec interface, common to all volume plugins.

*) one or two plugins would still need nontrivial refactoring to pass the interface from place where it's available to place where it's needed, but that's another story.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: Yes, kubelet tries indefinitely.

And if the pod with mount utilities is not available for a longer time... I checked volume plugins, most (if not all) run `umount` on the host. So the volume gets unmounted cleanly and data won't be corrupted. Detaching an iSCSI/FC/Ceph RBD disk may be a different story. The disk may be attached forever and then it depends on the backend if it supports attaching the volume to a different node.

As I wrote, update of the daemon set is a very tricky operation and the node should be drained first.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: IO errors. I guess health probe should fail and the pod should be rescheduled (or deployment / replication set will create a new one).

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: I am asking if there are some tricks how to do DaemonSet rolling update that would drain a node first before updating the pod. Otherwise I need to fall back to `--cascade=false` and do the update manually as @humblec suggests.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: `kubectl exec` is easy to implement, does not need a new protocol and can be restrained by RBAC. With HTTP, we need to define and maintain the protocol, its implementation, have a db for auth{n,z}, generate certificates, ...

> how the controller manager identifies the service could be up for debate - predefined name? labels? namespace

getting rid of namespaces / labels was the reason why we have gRPC over UNIX sockets. If we have half of the system using gPRC, second half with `kubectl exec`, why don't we use `kubectl exec` (or gRPC) for everything?

contributors/design-proposals/containerized-mounter-pod.md
liggitt: would the pod need to be a highly privileged pod, likely with hostpath volume mount privileges?

jsafrane: yes

liggitt: this is the first instance I'm aware of where a controller would be required to have the ability to create privileged pods. not necessarily a blocker, but that is a significant change.

contributors/design-proposals/containerized-mounter-pod.md
tallclair: Yes, but if zombie processes is an issue for other runtimes, they should have a built-in way of dealing with them. It shouldn't be necessary to implement reaping in the pod, unless it's expected to generate a lot of zombie processes, I believe. ( @yujuhong does this sound right? )

contributors/design-proposals/containerized-mounter-pod.md
tallclair: Got it. In that case, we should at least detect this condition and give a better error message. That was the part that was most confusing to me.

contributors/design-proposals/containerized-mounter-pod.md
tallclair: Does unmounting block pod deletion? I.e. will the pod be stuck in a terminated state until the volume utility pod is able to be reached?

tallclair:
I misunderstood the original intent of this proposal. I thought the goal was to get much closer to the desired end-state of true CSI plugins. However, I now see that this is just providing the binary utilities for the existing (hard-coded) plugins.

Given that, I'm afraid I want to go back on my original suggestions. Since this really is an exec interface, I think the original proposal of using the native CRI exec (specifically, [`ExecSync`](https://github.com/kubernetes/kubernetes/blob/a1c0510d006ccff9be8478f86635c86658c9bf73/pkg/kubelet/apis/cri/v1alpha1/runtime/api.proto#L66)) makes sense.

contributors/design-proposals/containerized-mounter-pod.md
humblec: iic, for mounts we only need `CAP_SYS_ADMIN`, however if we export `/dev/ we need privilege pods.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: No, unmounting happens after a pod is deleted.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: I agree that the error message looks ugly, however it's quite hard to give a better one. Kubelet does not have any configuration regarding mount pods, so it does not know in advance which volume plugins should execute their stuff on the host and which should wait for pods with mount utilities (and kubelet should therefore report a nice error that it's waiting for a pod).

Taking Gluster as an example, its volume plugin just calls `/bin/mount -t glusterfs <what> <where>`. `/bin/mount` returns universal exit code 32, which can mean many things and missing `/sbin/mount.glusterfs` is just one of them. Another one can be missing fuse kernel module. IMO, it's impossible to distinguish if the error is caused by a tool that could be provided by a pod (and thus we could report "please install a daemonset with GlusterFS utilities") or something else is broken.

What I could do is to 1) check for presence of `/sbin/mount.glusterfs` in the volume plugin and 2) scan stderr for some known strings like "unknown filesystem type 'glusterfs'". If one of these happens, mount plugin returns an error "Missing mount utilities for GlusterFS" and kubelet sends it as event on the pod.  Of course, all plugins need to be updated like this, Gluster is just an example. This is not 100% heuristics, but better than nothing.


What's worst, there *can* be mount.glusterfs on the host, but system admin wants to use the one from a pod, because the pod can have a newer version or some important bug fixed or the right file in pod's /etc. And here the mount can just fail with any error message.

jsafrane:
@tallclair `ExecSync` looks usable.

I'd like to revisit the UNIX sockets. We still need a way how to run stuff in pods with mount utilities from controller-manager, which cannot use UNIX sockets. So there must be a way (namespaces, labels) to find these pods. Why can't kubelet use the same mechanism instead of UNIX sockets? It's easy to do `kubectl exec <pod> mount -t glusterfs ...` from kubelet (using `ExecSync` in the background) and it's currently the easiest way how to reach pods from controller-manager.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: For Attach/Detach operation a random node is IMO OK. All state is kept in attach/detach controller and volume plugins, not in the utilities that are executed by a volume plugin. Note that there is only Ceph RBD that executes something during attach.

For chasing secrets, that's actually benefit of a pod with mount utilities - any secrets that are needed to talk to backend storage can be easily available to the pod via env. variables or Secret volume. And since only `os.exec` will be delegated to a pod, whole command line will be provided to the pod incl. all necessary credentials.

jsafrane:
@tallclair I just had meeting with @saad-ali and @thockin and we agreed that UNIX sockets are better for now, we care about `mount` in 1.8 and we'll see if we ever need to implement attach/detach and how.

So, `ExecSync` indeed looks usable. I am not sure about whole `service RuntimeService` - can the gRPC endpoint that runs in a pod with mount utilities implement just `ExecSync` part of it? Would it be better to create a new `service RemoteExecService` just with `ExecSync` in it? Is it possible to have two services  having the same `ExecSync` rpc function?

jsafrane:
* Updated to correspond with the latest mount propagation proposal (#659), i.e. removed duplicate items
* Using `ExecSync` call from CRI. Created a new `ExecService` for it.

adelton:
Is there any expectation that with this change, some of the native volume plugins will become flexVolume plugins / drivers?

jsafrane:
I have implementation of this proposal in in fork: https://github.com/jsafrane/kubernetes/commits/mount-container3-08-16. I cherry picked kubernetes/kubernetes#46458 and kubernetes/kubernetes#46458 before they're merged.

Interesting parts (links are probably unstable):

* [CRI change](https://github.com/kubernetes/kubernetes/compare/master...jsafrane:mount-container3-08-16?expand=1#diff-6e8c1262dc9341dd851e5f7b636be64fR89) - to be honest, I don't like re-using CRI for communication between kubelet and a daemon running in a container. It's no longer Container Runtime Interface.

* [New command in cmd/volumeexec](https://github.com/kubernetes/kubernetes/compare/master...jsafrane:mount-container3-08-16?expand=1#diff-4b341f4bd7d8e0b41b44fd3ebfdc4f5b)

* The rest is just code that puts it together.

* Testing is pending until the proposal is merged.

contributors/design-proposals/containerized-mounter-pod.md
fabiand: There are some things to take care of when mounting `/dev` into a container, i.e. oyu need to take care of the `pts` device to not break the console. And there are other things to take care of as well.

Because of this I wonder if it makes sense to add an API flag to signal that a container should get the hosts proc, sys, and dev paths. If we had such a flag it would be much more well defined what a container will get, if he is told to get the _host view_ of these three directories.

Also a side note, we can not prevent it, but mounting the host's `dev` directory into a privileged container can cause quite a lot confusion (actually any setup where more than one udev is run can cause problems).

contributors/design-proposals/containerized-mounter-pod.md
fabiand: Could it be an option to provide a base container for these mount util containers, which has a sane pid 1?

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,158 @@
...
+    * E.g. `mkfs.ext4` is likely to be available on all hosts, but a pod with mount utilities should not depend on that nor use it.
+    * The only exception are kernel modules. They are not portable across distros and they *should* be on the host.
+* It is expected that these daemon sets will run privileged pods that will see host's `/proc`, `/dev`, `/sys`, `/var/lib/kubelet` and such. Especially `/var/lib/kubelet` must be mounted with shared mount propagation so kubelet can see mounts created by the pods.
+* The pods with mount utilities should run some simple init as PID 1 that reaps zombies of potential fuse daemons.
+* The pods with mount utilities run a daemon with gRPC server that implements `ExecService` defined below.
+  * Upon starting, this daemon puts a UNIX domain socket into `/var/lib/kubelet/plugin-sockets/` directory on the host. This way, kubelet is able to discover all pods with mount utilities on a node.
+  * Kubernetes will ship implementation of this daemon that creates the socket on the right place and simply executes anything what kubelet asks for.
+
```
fabiand: Does it also need ot be defined as which user the daemonset (rather container) is run?

I can imagine that problems can arise if the uid's between containers and host mismatch.

jsafrane: It's up to authors of these DaemonSets to define the UID in `SecurityContext` in the pod definition.

contributors/design-proposals/containerized-mounter-pod.md
jsafrane: I'd like to stay distro agnostic here and let the DaemonSet authors use anything they want. For NFS, simple Alpine Linux + buysbox init could be enough, for Gluster and Ceph a more powerful distro is needed.

contributors/design-proposals/containerized-mounter-pod.md
fabiand: To be more precise on the "things" - Please take a look here to see the workarounds *cough*hacks*cough* we need to do to have libvirt running "properly" (we don't use all features, just a subset, and they work well so far) in a container.

jsafrane: We can't influence how Docker (or other container runtime) creates/binds /dev and /sys. Once this flag is available in Docker/Moby and CRI we could expose it via Kubernetes API, but it's a long process. Until then we're stick to workarounds done *inside* the container.

I'll make sure we ship a well documented sample of such mount container. That's why it's alpha feature and we know all these workarounds before going to beta/stable.

jsafrane:
@adelton, no, current volume plugins won't be affected by this proposal. On the other hand, there are people working on CSI that may eventually replace internal volume plugins in future, see https://github.com/container-storage-interface/spec/blob/master/spec.md

jsafrane:
* Added a note that udev should run on the host (because it should run only once per machine). This iplies that if a volume plugin needs to talk to udev, it must do it on the host. E.g. `udevadm trigger` should be executed on the host and not in a container with mount utilities.

* Added a possibility that `volume-exec` binary (gRPC endpoint that runs inside a container and executes mount utilities there as requested by kubelet / volume plugin) could run as pid 1 and consume / ignore SIGCHLD (and therefore reap zombies of fuse daemons). Implementation would be IMO trivial.

jbw976:
Nice write up @jsafrane.  With our efforts to create a volume plugin in the [Rook](https://github.com/rook/rook) project, we have landed on a design with similarities to the approach outlined here, but has no requirement for new features in the Kubernetes control plane (besides FlexVolume plugin discovery, which isn't strictly required, but makes the experience much more seamless).

The full design write up can be found at https://github.com/rook/rook/pull/924, with easy to read rendered markup at https://github.com/jbw976/rook/blob/agent-design/design/local-node-agent.md.

A PR for the working implementation can be found at https://github.com/rook/rook/pull/882.

The tl;dr summary is that Rook is implementing a Flexvolume driver and an "agent" pod that runs on all nodes in the cluster.  The driver calls into the agent pod over a Unix domain socket to perform the various volume plugin functions (attach/detach/mount/unmount).  To communicate with a remote node, e.g. to tell a remote node to perform an attach, CustomResourceDefinitions will be used that the agents in the cluster will be watching.

As an alternative for out of tree storage providers in the Kubernetes ecosystem, I'm wondering if this Rook design/pattern may work well.  Some things I like about the Rook Flexvolume and Agent design:
* only uses existing Kubernetes functionality and management (Flexvolume, CRDs).  No new functionality needs to go into the Kubernetes tree.
* containerized deployment of all the tools and utilities needed for the volume plugin.  Nothing needs to be installed on the host machine.
* Kubelet doesn't need any modification to the way it is started (e.g. no need to mount /lib/modules or /usr/sbin/modprobe)
* fully out of tree implementation so it can be versioned and released independently from Kubernetes.
* 100% automated UX for storage admins. There's no manual fields like secrets, endpoints, etc. that need to be copy pasted into yaml files.
* long running daemon that can perform other tasks, such as proxying I/O traffic from kernel to user space.

I wanted to provide some visibility on that design in case it could be helpful to other storage providers in the community.  We welcome any feedback on the design that anyone has.

jsafrane:
@jbw976, I know about existing flex updates and how it can be used to inject flex drivers on the fly, but it can't work on deployments that can't run pods on master nodes such as GKE. In addition, it requires using flex for everything, even for simple NFS. With my proposal, anyone could use NFS, Gluster and Ceph as well as flex everywhere with full attach/detach functionality (in a future releases, the controler part needs some work)

bassam:
@jsafrane it seems to me that if we are willing to run containers on master that we should be willing to schedule pods. The bigger point here is wether we can use Kubernetes itself to run storage plugins.

I see your point about having NFS and others work without having to install/run anything. I think that could be achieved by introducing a "storage-addon" like we do with for DNS for example. Kubeadm and installers could run the relevant pods on installation.

jsafrane:
Trying to resurrect the discussion, I am still interested in this proposal.

@tallclair, looking at [device plugin gRPC API](https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/apis/deviceplugin/v1alpha1), it looks better to me to follow this approach and introduce "container exec API" with `ExecSync` call instead of re-using CRI. What do you think about it?

tallclair:
The device plugin api is a higher level abstraction than just arbitrary exec. I wasn't a part of the meeting where it was decided to stick with a socket interface, but I don't see the value in implementing an alternative arbitrary command exec interface rather than relying on `ExecSync`. I'm happy to join in another discussion.

jsafrane:
Reworked according to result of the latest discussion:

- Use CRI ExecSync to execute stuff in pods with mount utilities (= no new gRPC interface)
- Use files in `/var/lib/kubelet/plugin-containers/<plugin name>` for discovery of pods with mount utilities.

- Added a note about containerized kubelet for completeness, no extra changes necessary.

@tallclair @dchen1107 @thockin @vishh PTAL

jsafrane:
Implementation of this proposal is at https://github.com/kubernetes/kubernetes/pull/53440 - it's quite small and well contained.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/589)

contributors/design-proposals/containerized-mounter-pod.md
```diff
@@ -0,0 +1,149 @@
...
+* The pods with mount utilities must put a file into `/var/lib/kubelet/plugin-containers/<plugin name>.json` for each volume plugin it supports. It should overwrite any existing file - it's probably leftover from older pod.
+   * Admin is responsible to run only one pod with utilities for one volume plugin on a single host. When two pods for say GlusterFS are scheduled on the same node they will overwrite the registration file of each other.
+   * Downward API can be used to get pod's name and namespace.
+   * Root privileges (or CAP_DAC_OVERRIDE) are needed to write to `/var/lib/kubelet/plugin-containers/`.
+
+To sum it up, it's just a daemon set that spawns privileged pods, running a simple init and registering itself into Kubernetes by placing a file into well-known location.
+
+**Note**: It may be quite difficult to create a pod that see's host's `/dev` and `/sys`, contains necessary kernel modules, does the initialization right and reaps zombies. We're going to provide a template with all this. During alpha, it is expected that this template will be polished as we encounter new bugs, corner cases, systemd / udev / docker weirdness.
```
tallclair: > During alpha,

Is this expected to ever leave alpha? I thought this was a temporary hack while we wait for CSI?

jsafrane: I removed all notes about alpha in the text and added a note about feature gate and that it's going to be alpha forever.

jsafrane:
I squashed all the commits, the PR is ready to be merged.

On personal meeting with @tallclair and @saad-ali we agreed that all volume plugins are going to be moved to CSI eventually, so this proposal has limited lifetime. CSI drivers will have different discovery mechanism and all kubelet changes proposed here won't be needed.

I still think this PR is useful, as it allows us to create tests for internal volume plugins so we can check their CSI counterparts for regressions in e2e tests. Wherever the CSI drivers will live, Kubernetes still needs to keep its backward compatibility and make sure that old PVs keep working.

jsafrane:
Why is this not merged? "pull-community-verify — Waiting for status to be reported"
/test all

PR #601, opened 2017-05-05
Move youtube instructions to comm/comm

grodrigues3:
Youtube collab instructions were in wrong folder.  Moved.


czahedi:
LGTM thank you!

PR #602, opened 2017-05-05
Kubelet server certificate bootstrap and rotation

jcbsmpsn:


ericchiang:
@kubernetes/sig-auth-proposals @kubernetes/sig-node-proposals

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+    - generate a key
+    - create a certificate signing request
+    - request a signed certificate from the API server
+    - wait for a response
+    - Store the new cert/key pairs in the directory specified by `--cert-dir`,
+      or the default value if it is unspecified.
+
+### As Expiration Approaches
+
+Rotating a Kubelet server certificate will work by generating a new private key,
```
ericchiang: Any reason it's regenerating a new private key rather then reusing the old one?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+   default value if it is unspecified. This will allow the kubelet to have a
+   place for storing the multiple cert/key pairs that it might have available at
+   any given moment (because of a rotation in progress).
+    - When cert/key data is directly embedded in the kubelet config file
+      (client-certificate-data, client-key-data)
+        - The values be used only if there are no updated cert/key pairs due to
+          rotation
+    - When cert/key files are specified in kubeconfig, or with the
+      `--tls-cert-file` and `--tls-private-key-file`:
+        - These values be used only if there are no updated cert/key pairs due
```
ericchiang: Having trouble parsing this section. Does this imply that the kubelet will try to request new serving certs even if you've configured them through flags?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+          [`GetClientCertificate`](https://golang.org/pkg/crypto/tls/#Config)
+          callbacks so that the connection dynamically requests the certificate
+          as needed.
+    - Recovering from kubelet crashes or restarts that occur while certificate
+      transitions are in flight (request issued, but not yet signed, etc)
+1. The RBAC role for nodes will be updated to allow nodes to request new
+   certificates.
+1. Have the CertificateManager repeat the CSR process as certificate expiration
+   approaches.
+    - New certificates will be requested when the configured duration threshold
```
ericchiang: Is this actually configurable through a command line flag or something?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+        - Private key for requests in flight are held in memory. In case of
+          interruption, they will be abandoned.
+        - When the corresponding signed certificate is received, the cert/key
+          pair will be written to a single file, eg.
+          kubelet-server-<timestamp>.pem.
+        - Write kubelet-server-updated.pem symlink to point to the new cert/key
+          pair.
+        - Delete kubelet-server-current.pem
+        - Move kubelet-server-updated.pem to kubelet-server-current.pem
+1. Currently the Certificate Request Signing API is hard coded to issue
```
ericchiang: s/Certificate Request Signing API/Certificate Request Signing controller/

ericchiang: Also this doesn't seem like a step in the renewal process. Maybe a "future work" section that could include testing and approval?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+      has been exceeded.
+    - Crash safe file structure
+        - Private key for requests in flight are held in memory. In case of
+          interruption, they will be abandoned.
+        - When the corresponding signed certificate is received, the cert/key
+          pair will be written to a single file, eg.
+          kubelet-server-<timestamp>.pem.
+        - Write kubelet-server-updated.pem symlink to point to the new cert/key
+          pair.
+        - Delete kubelet-server-current.pem
```
ericchiang: nit: Delete is redundant? e.g. the move already does this?

luxas:
@kubernetes/sig-cluster-lifecycle-feature-requests @timothysc @jbeda

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
coresolve: how do you go about loading the new key? Is the application watching the file? Or is there some other part of this where we HUP the process when after the move to kubelet-server-current.pem?

ericchiang: @coresolve it's discussed in the section about the "CertificateManager".

Go can dynamically present a certificate on a per connection basis. The certificate manager code reads the cert/key from disk and kubelet will start using it.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+### During Kubelet Boot Sequence
+
+1. Look on disk for an existing cert/key pair.
+1. If there is an existing cert/key pair, load them.
+1. If there is no existing cert/key pair, at the point where the kubelet
+   currently generates a key and certificate, the kubelet will:
+    - generate a key
+    - create a certificate signing request
+    - request a signed certificate from the API server
+    - wait for a response
```
deads2k: If it doesn't get approved, what will it do?  What if it doesn't get an answer?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
deads2k: Optional or required?  I suspect optional, but I'd like it called out as an optional feature at the top level.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+issuing a new Certificate Signing Request to the API Server, safely updating the
+cert/key pair on disk, begin using the new cert/key pair.
+ 
+1. Store the new cert/key pairs in the directory specified by `--cert-dir`, or the
+   default value if it is unspecified. This will allow the kubelet to have a
+   place for storing the multiple cert/key pairs that it might have available at
+   any given moment (because of a rotation in progress).
+    - When cert/key data is directly embedded in the kubelet config file
+      (client-certificate-data, client-key-data)
+        - The values be used only if there are no updated cert/key pairs due to
```
deads2k: This override behavior seems backwards.  I'd expect the explicit values in a kubeconfig to win and specifying both to be an error.

deads2k: > This override behavior seems backwards. I'd expect the explicit values in a kubeconfig to win and specifying both to be an error.

I guess explicit is relative.  Do our CLI flags work this way too?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+          rotation
+    - When cert/key files are specified in kubeconfig, or with the
+      `--tls-cert-file` and `--tls-private-key-file`:
+        - These values be used only if there are no updated cert/key pairs due
+          to rotation
+1. There will be a feature gate to enable certificate bootstrapping and
+   rotation, `RotateKubeletServerCertificate`
+1. Centralize certificate access within the kubelet code to the
+   CertificateManager. The CertificateManager will be responsible for:
+    - Providing the correct certificate to use for establishing TLS connections.
```
deads2k: This is the CA bundle used to verify a connection to the `kube-apiserver`?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+    - Generating new private keys and requesting new certificates when the
+      current certificate approaches expiry.
+    - Since certificates can rotate at any time, all other parts of the kubelet
+      should ask the CertificateManager for the correct certificate each time a
+      certificate is used. No certificate caching except by the
```
deads2k: How will you prevent the use of the "normal" connection caches we have in the client?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
mikedanese: Is there a reason not to?

liggitt: generally, you want young keys... the older a key, the longer the exposure to factoring. if you have an auto-renewal system, using a new key each time is good.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
liggitt: this is all under a heading of serving-certificate management... serving certs are not embedded in kubeconfig files anywhere

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
```diff
@@ -0,0 +1,97 @@
...
+ 
+1. Store the new cert/key pairs in the directory specified by `--cert-dir`, or the
+   default value if it is unspecified. This will allow the kubelet to have a
+   place for storing the multiple cert/key pairs that it might have available at
+   any given moment (because of a rotation in progress).
+    - When cert/key data is directly embedded in the kubelet config file
+      (client-certificate-data, client-key-data)
+        - The values be used only if there are no updated cert/key pairs due to
+          rotation
+    - When cert/key files are specified in kubeconfig, or with the
```
liggitt: serving certs are only specified via --tls-cert-file/--tls-private-key-file

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
liggitt: I'd expect either an explicitly specified cert or an auto-rotated cert, not both

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
ericchiang: On the client side the certificate manager overrides the restclient.Config GetClientCertificate method[0].

[0] https://github.com/kubernetes/kubernetes/pull/41912/files#diff-e0160f895346669c2e4495acfd858ea7R466

liggitt: making the client certificate dynamic is problematic for the transport cache at https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/transport/cache.go#L41

ericchiang: Ah I see that #41912 actually added the GetClientCertificate field to the TLS config.

Does it work for the cache to ignore configs that set this new field? I'm under the impression that the kubelet is a lot better about keeping a centralized client instead of constantly creating new ones like kubectl.

liggitt: yeah, those have to be uncacheable. I'm not sure how many transports get created with a call to something like `clientset.NewForConfig`.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: These were pre-existing flags and I'm trying to find a way to nicely incorporate their behavior.

The way I see this, section to to use this cert/key pair if it is there, and be able to move on to rotation when the time is right.

I could do as Jordan suggests, but it makes it harder to transition an existing cluster. If someone has a cluster, with this cert/key specified, then these are the valid authentication mechanisms for connecting to the API server in order to do a rotation. However, once there is a rotation, that cert/key pair will be newer, and these parameters will be ignored. Analogous to how bootstrapping works, a cert/key pair specified in the kube config file is valid, as long as there is no more up to date cert/key pair.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: Duration is chosen by the Certificate Request Signing API. There is currently no flag for changing the duration that will be assigned, but I've been thinking of adding one to facilitate testing.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: It doesn't proceed.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: Done.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: Done.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: Do you mean certificate rotation should be optional? Currently it is gated by a feature flag due to it's alpha status. Would you like it gated by a command line parameter and rotation of expiring certificates would be an optional feature of a kubelet?

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: This is the certificate that is used for any incoming connections to port 10250. CAs are not affected by this proposal.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
liggitt: even once the rotation feature is no longer alpha, not all deployments will use cert rotation. I would expect to be able to continue providing a cert/key explicitly to the kubelet and have it simply use it.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: Updated to remove the reference to keys in the kubeconfig files.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: Removed this.

contributors/design-proposals/kubelet-server-certificate-bootstrap-rotation.md
jcbsmpsn: There is now a flag (`--experimental-cluster-signing-duration`) on the API server to set the certificate duration for signed certificates.

https://github.com/kubernetes/kubernetes/pull/46058

luxas:
Can this be merged now when the code has been merged to master?

gkarthiks:
hi @jcbsmpsn Since we have a KEP process, can we please merge this soon or close this and create a KEP?

PR #606, opened 2017-05-08
Clarify when to use conditions versus fields in status

pmorie:
This is an attempt to clarify the answer to the question "when should I add a field to a resource's status, andd when should I add a condition".  In SIG service catalog, there were a number of times when this was not clear to us while building the service catalog API resources.

See https://github.com/kubernetes-incubator/service-catalog/pull/717 for some background.

### DIFF
--- a/contributors/devel/api-conventions.md
+++ b/contributors/devel/api-conventions.md
@@ -232,7 +232,7 @@ the resource may be deleted prior to this time.
 categorize objects (see [the labels docs](https://kubernetes.io/docs/user-guide/labels/))
 * annotations: a map of string keys and values that can be used by external
 tooling to store and retrieve arbitrary metadata about this object (see
-[the annotations docs](https://kubernetes.io/docs/user-guide/annotations/)
+[the annotations docs](https://kubernetes.io/docs/user-guide/annotations/)).
 
 Labels are intended for organizational purposes by end users (select the pods
 that match this label query). Annotations enable third-party automation and
@@ -300,8 +300,10 @@ information) and the status is the response. For these RPC like objects the only
 operation may be POST, but having a consistent schema between submission and
 response reduces the complexity of these clients.
 
-
 ##### Typical status properties
+
+There are two fundamental types of status information: conditions and normal
+API fields.
 
 **Conditions** represent the latest available observations of an object's
 current state. Objects may report multiple conditions, and new types of
@@ -317,14 +319,14 @@ following fields, but must contain at least `type` and `status` fields:
   // +optional
   LastHeartbeatTime  unversioned.Time  `json:"lastHeartbeatTime,omitempty" description:"last time we got an update on a given condition"`
   // +optional
-  LastTransitionTime unversioned.Time  `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
+  LastTransitionTime unversioned.Time  `json:"lastTransitionTime,omitempty" description:"last time the condition transitioned from one status to another"`
   // +optional
   Reason             string            `json:"reason,omitempty" description:"one-word CamelCase reason for the condition's last transition"`
   // +optional
   Message            string            `json:"message,omitempty" description:"human-readable message indicating details about last transition"`
 ```
 
-Additional fields may be added in the future.
+Additional fields may be added to the prototypical condition in the future.
 
 Conditions should be added to explicitly convey properties that users and
 components care about rather than requiring those properties to be inferred from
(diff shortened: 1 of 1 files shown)


pmorie:
Note to self: update the updated time in this doc if this PR is accepted

pmorie:
I think this doc needs to have a specific decision plane in it about this question like:

IF X -> add a condition
IF Y -> add a field to status

vaikas-google:
In case it helps, the one specific state in the SIG Service Catalog was that a Service Instance can be in a state, where it is making an asynchronous request to a Service Broker. That operation can be very long lasting (hours ++). While an asynchronous request is ongoing, no additional requests can be made against that Service Instance. We decided to put this into a Status field as a boolean field, but the discussion that lead up to that design decision left us wondering if the decision on whether to use a condition vs. Status field could be made easier for any further developers tackling a similar design decision by clarifying the documentation.

contributors/devel/api-conventions.md
```diff
@@ -317,14 +319,14 @@ following fields, but must contain at least `type` and `status` fields:
   // +optional
   LastHeartbeatTime  unversioned.Time  `json:"lastHeartbeatTime,omitempty" description:"last time we got an update on a given condition"`
```
deads2k: This field causes a lot of update traffic.

contributors/devel/api-conventions.md
```diff
@@ -379,10 +381,15 @@ objects, with possibly a reference from the original object. This helps to
 ensure that GETs and watch remain reasonably efficient for the majority of
 clients, which may not need that data.
 
-Some resources report the `observedGeneration`, which is the `generation` most
-recently observed by the component responsible for acting upon changes to the
-desired state of the resource. This can be used, for instance, to ensure that
-the reported status reflects the most recent desired status.
+Sometimes not every aspect of an object's status can be modeled as a
```
deads2k: Generally speaking, I expect to see status API fields used when the information is structured and will be used for taking other programmatic action.  For example, the number of replicas in various states on the replicaset status or the acceptednames on the customresource status.

contributors/devel/api-conventions.md
pmorie: Agree -- I can't think of an API that we actually have that field in, and was surprised to see it.

contributors/devel/api-conventions.md
pmorie: That is a good way to articulate the decision plane

pmorie:
@deads2k updated

contributors/devel/api-conventions.md
davidopp: But we use NodeCondition to decide to evict pods...

contributors/devel/api-conventions.md
davidopp: IIRC the whole concept of conditions was invented for NodeCondition and so some of the fields only make sense there (like this one).

bgrant0607:
@pmorie what distinction are you trying to make? I looked at the change, but it was not obvious.

Aside: having resources go out to lunch is unprecedented in the API

contributors/devel/api-conventions.md
pmorie: I will take this particular condition out.

vaikas-google:
@bgrant0607 We're trying to determine what guideline a developer should utilize when adding a new field. As an example, we had a hard time determining whether a condition should represent the state or if it should be stashed into the Status as a field. Fact that it took a few rounds to decide seems like there could be some clarifications made so that to a developer it would be more obvious on whether to utilize Conditions or whether to keep state in the Status field. I tried to give the concise version of what we were trying to do (more details in the link @pmorie gave) in case it helps to illustrate a problem we were trying to solve, and how we ended up solving it. It's still not obvious to me that putting this state into a Status field is any more correct than adding a Condition based on the current wording of the document.
@pmorie is trying to clarify the documentation so that a developer can really just have an easier way to decide where to keep some state.

@bgrant0607 when you say, go out to lunch, do you mean the blocking aspect of asynchronous requests?

pmorie:
I am also having trouble interpreting the out to lunch comment, or determining which part of the changeset it pertains to :)

pmorie:
> I am also having trouble interpreting the out to lunch comment, or determining which part of the changeset it pertains to :)

Oh my god what a pun.

pmorie:
@vaikas-google https://github.com/kubernetes/community/pull/606#discussion_r115835829

pmorie:
@bgrant0607 

> @pmorie what distinction are you trying to make? I looked at the change, but it was not obvious.

The more I think about this, the more I think the question should be asked: if we were starting over, would conditions typically have optional fields, possible with union-type behavior, for extra values beyond 'Reason' and 'Message' that might be relevant attributes of the object's status?  For example, would replica sets have a 'ReadyReplicas' condition reason that was associated with an optional integer field of the condition type for replica set?

I think what is not obvious now is given some aspect of an object's status, when to add fields to the `ThingStatus` and when to add new condition reasons.

bgrant0607:
@pmorie @vaikas-google 

There is more explanation of the motivation/origin for conditions here:
https://github.com/kubernetes/kubernetes/issues/7856

Conditions were created as a way to wean contributors off of the idea of state machines with mutually exclusive states (called "phase").

Each condition represents a non-orthogonal "state" of a resource -- is the resource in that state or not (or unknown). They should be complementary to events pertaining to the resource.

No, I wouldn't pack additional condition-specific data into the conditions, nor add additional value types. 

Conditions are intended to be handled uniformly, across multiple conditions within a resource and across resources, and extensibly -- it should be possible for kubectl to extract status.conditions from any resource and display them.

Arbitrarily typed/structured status properties specific to a particular resource should just be fields.

Note that we're converting most Node conditions to taints, which can express more clear and more sophisticated properties, which are specific to Nodes only.

Between the issue I referenced and the above text, there should be quite a bit of material to further clarify the intent.

bgrant0607:
Following up with some more thoughts that came out of a discussion with @erictune ...

* The question of what should be an ordinary status field vs. a condition has come up a number of times.
* Conditions are less discoverable and more cumbersome to use than fields.
* The tri-state nature of conditions (i.e., they may not be present) has been found to be confusing, though this can occur for status fields, as well, across releases, and due to asynchrony, when populated only by a controller.
* Conditions have proven to be expensive to update. In particular, the LastHeartbeatTime / LastProbeTime / LastUpdateTime fields are usually not used as a result. 
* Time skew across components prevents the time fields from being used for decision-making.
* LastTransitionTime and Message are primarily for debugging, and overlap enough with Events to reconsider them.
* As discussed in the ["Make Kubernetes Events Great Again" doc](https://docs.google.com/document/d/13BeJlrEcJhSKgsHOHWmHdJGqXrjSm_o9XxOtwcN6yNg/edit) comments, object status fields do need to represent information that controllers would want to watch in order to take some kind of action.
* As discussed in the [API orchestration issue](https://github.com/kubernetes/kubernetes/issues/34363), we will want to be able to interpret common statuses polymorphically across resource types, but conditions are neither necessary nor sufficient for that, and it's likely that subresources aren't necessary, either. I'm currently inclined towards more of a duck-typing approach (which would allow observedGeneration to remain part of status while being handled generically).
* As mentioned above, we're converting most/all Node conditions to taints, for more robust and generic handling in schedulers.
* In `kubectl get` output, we found that just a boolean (true/false) wasn't sufficiently informative. Users found it useful to pack more information into a single field, such as CrashLoopBackoff. I do think that retaining subcategorization of exceptional conditions is valuable to users.

Will think about a specific example.

bgrant0607:
Moving this to https://github.com/kubernetes/kubernetes/issues/7856 until we have a proposal.

PR #607, opened 2017-05-08
Encryption of secrets at the database layer

destijl:
This replaces https://github.com/kubernetes/community/pull/454 and is an updated version so we could continue to move the design forward as andrew didn't have bandwidth to continue.

I've updated the doc with most of the feedback and split out a few design options of increasing complexity so we can decide where to go next.

Based on the work that's been done in:
https://docs.google.com/document/d/1lFhPLlvkCo3XFC2xFDPSn0jAGpqKcCCZaNsBAv8zFdE/.

Related:
kubernetes/features#92
kubernetes/kubernetes#32579
kubernetes/kubernetes#12742

@kubernetes/sig-auth-misc

### DIFF
--- a/contributors/design-proposals/auth/encryption.md
+++ b/contributors/design-proposals/auth/encryption.md
@@ -1,0 +1,443 @@
+# Encryption
+
+## Abstract
+
+The scope of this proposal is to ensure that resources can be encrypted at the
+datastore layer with sufficient metadata support to enable integration with
+multiple encryption providers and key rotation. Encryption will be optional for
+any resource, but will be used by default for the Secret resource. Secrets are
+already protected in transit via TLS.
+
+Full disk encryption of the volumes storing etcd data is already expected as
+standard security hygiene.  Adding the proposed encryption at the datastore
+layer defends against malicious parties gaining access to:
+
+- etcd backups; or
+- A running etcd instance without access to memory of the etcd process.
+
+Allowing sensitive data to be encrypted adheres to best practices as well as
+other requirements such as HIPAA.
+
+## High level design
+
+Before a resource is written to etcd and after it is read, an encryption
+provider will take the plaintext data and encrypt/decrypt it.  These providers
+will be able to be created and turned on depending on the users needs or
+requirements and will adhere to an encryption interface.  This interface will
+provide the abstraction to allow various encryption mechanisms to be
+implemented, as well as for the method of encryption to be rotated over time.
+
+For the first iteration, a default provider that handles encryption in-process
+using a locally stored key on disk will be developed.
+
+## Kubernetes Storage Changes
+
+Kubernetes requires that an update that does not change the serialized form of
+object not be persisted to etcd to prevent other components from seeing no-op
+updates.
+
+This must be done within the Kubernetes storage interfaces - we will introduce a
+new API to the Kube storage layer that transforms the serialized object into the
+desired at-rest form and provides hints as to whether no-op updates should still
+persist (when key rotation is in effect).
+
+```go
+// ValueTransformer allows a string value to be transformed before being read from or written to the underlying store. The methods
+// must be able to undo the transformation caused by the other.
+type ValueTransformer interface {
+        // TransformFromStorage may transform the provided data from its underlying
+        // storage representation or return an error.  Stale is true if the object
+        // on disk is stale (encrypted with an older key) and a write to etcd
+        // should be issued, even if the contents of the object have not changed.
+        TransformFromStorage([]byte) (data []byte, stale bool, err error)
+        // TransformToStorage may transform the provided data into the appropriate form in storage or return an error.
+        TransformToStorage([]byte) (data []byte, err error)
+}
+```
+
+When the storage layer of Kubernetes is initialized for some resource, an
+implementation of this interface that manages encryption will be passed down.
+Other resources can use a no-op provider by default.
+
+## Encryption Provider
+
+An encryption provider implements the ValueTransformer interface. Out of the box
+this proposal will implement encryption using a standard AES-GCM performing
+AEAD, using the standard Go library for AES-GCM.
+
+Each encryption provider will have a unique string identifier to ensure
+versioning of the ciphertext in etcd, and to allow future schemes to be added.
+
+During encryption, only a single provider is required.  During decryption,
+multiple providers or keys may be in use (when migrating from an older version
+of a provider, or when rotating keys), and thus the ValueTransformer
+implementation must be able to delegate to the appropriate provider.
+
+Note that the ValueTransformer is a general storage interface and not related to
+encryption directly. The AES implementation linked below combines
+ValueTransformer and encryption provider.
+
+### AES-GCM Encryption provider
+
+Implemented in [#41939](https://github.com/kubernetes/kubernetes/pull/41939).
+
+The simplest possible provider is an AES-GCM encrypter/decrypter using AEAD,
+where we create a unique nonce on each new write to etcd, use that as the IV for
+AES-GCM of the value (the JSON or protobuf data) along with a set of
+authenticated data to create the ciphertext, and then on decryption use the
+nonce and the authenticated data to decode.
+
+The provider will be assigned a versioned identifier to uniquely pair the
+implementation with the data at rest, such as “k8s-aes-gcm-v1”.  Any
+implementation that attempts to decode data associated with this provider id
+must follow a known structure and apply a specific algorithm.
+
+Various options for key generation and management are covered in the following
+sections. The provider implements one of those schemes to retrieve a set of
+keys. One is identified as the write key, all others are used to decrypt data
+from previous keys.  Keys must be rotated more often than every 2^32 writes.
+
+The provider will use the recommended Go defaults for all crypto settings
+unless otherwise noted.  We should use AES-256 keys (32 bytes).
+
+Process for transforming a value (object encoded as JSON or protobuf) to and
+from stable storage will look like the following:
+
+Layout as written to etcd2 (json safe string only):
+```
+NONCE := read(/dev/urandom)
+PLAIN_TEXT := <VALUE>
+AUTHENTICATED_DATA := ETCD_KEY
+CIPHER_TEXT := aes_gcm_encrypt(KEY, IV:NONCE, PLAIN_TEXT, A:AUTHENTICATED_DATA)
+BASE64_DATA := base64(<NONCE><CIPHER_TEXT>)
+STORED_DATA := <PROVIDER>:<KEY_ID>:<BASE64_DATA>
+```
+
+Layout as written to etcd3 (bytes):
+```
+NONCE := read(/dev/urandom)
+PLAIN_TEXT := <VALUE>
+AUTHENTICATED_DATA := ETCD_KEY
+CIPHER_TEXT := aes_gcm_encrypt(KEY, IV:NONCE, PLAIN_TEXT, A:AUTHENTICATED_DATA)
+STORED_DATA := <PROVIDER_ID>:<KEY_ID>:<NONCE><AUTHENTICATED_DATA><CIPHER_TEXT>
+```
+
+Pseudo-code for encrypt (golang):
+```go
+block := aes.NewCipher(primaryKeyString)
+aead := cipher.NewGCM(c.block)
+keyId := primaryKeyId
+
+// string prefix chosen over a struct to minimize complexity and for write
+// serialization performance.
+// for each write
+nonce := make([]byte, block.BlockSize())
+io.ReadFull(crypto_rand.Reader, nonce)
+authenticatedData := ETCD_KEY
+cipherText := aead.Seal(nil, nonce, value, authenticatedData)
+storedData := providerId + keyId + base64.Encode(nonce + authenticatedData + cipherText)
+```
+
+Pseudo-code for decrypt (golang):
+```go
+// for each read
+providerId, keyId, base64Encoded := // slice provider and key from value
+
+// ensure this provider is the one handling providerId
+aead := // lookup an aead instance for keyId or error
+bytes := base64.Decode(base64Encoded)
+nonce, authenticatedData, cipherText := // slice from bytes
+out, err := aead.Open(nil, nonce, cipherText, authenticatedData)
+```
+
+### Alternative Considered: SecretBox
+
+Using [secretbox](https://godoc.org/golang.org/x/crypto/nacl/secretbox) would
+also be a good choice for crypto. We decided to go with AES-GCM for the first
+implmentation since:
+
+- No new library required.
+- We'd need to manage AEAD ourselves.
+- The cache attack is not much of a concern on x86 with AES-NI, but is more so
+  on ARM
+
+There's no problem with adding this as an alternative later.
+
+## Configuration
+
+We will add the following options to the API server. At API server startup the
+user will specify:
+
+```yaml
+--encryption-provider-config=/path/to/config
+--encryption-provider=default
+--encrypt-resource=v1/Secrets
+```
+
+The encryption provider will check it has the keys it needs and if not, generate
+them as described in the following section.
+
+## Key Generation, Distribution and Rotation
+
+To start with we want to support a simple user-driven key generation,
+distribution and rotation scheme. Automatic rotation may be achievable in the
+future.
+
+To enable key rotation a common pattern is to have keys used for resource
+encryption encrypted by another set of keys (Key Encryption Keys aka KEK).  The
+keys used for encrypting kubernetes resources (Data Encryption Keys, aka DEK)
+are generated by the apiserver and stored encrypted with one of the KEKs.
+
+In future versions, storing a KEK off-host and off-loading encryption/decryption
+of the DEK to AWS KMS, Google Cloud KMS, Hashicorp Vault etc. should be
+possible. The decrypted DEK would be cached locally after boot.
+
+Using a remote encrypt/decrypt API offered by an external store will be limited
+to encrypt/decrypt of keys, not the actual resources for performance reasons.
+
+Incremental deliverable options are presented below.
+
+### Option 1: Simple list of keys on disk
+
+In this solution there is no KEK/DEK scheme, just single keys in a list on disk.
+They will live in a file specified by the --encryption-provider-config, which
+can be an empty file when encryption is turned on.
+
+If the key file is empty or the user calls PUT on a /rotate API endpoint keys
+are generated as follows:
+
+1. A new encryption key is created.
+1. The key is added to a file on the API master with metadata including an ID
+   and an expiry time. Subsequent calls to rotate will prepend new keys to the
+   file such that the first key is always the key to use for encryption.
+1. The list of keys being used by the master is updated in memory so that the
+   new key is in the list of read keys.
+1. The list of keys being used by the master is updated in memory so that the
+   new key is the current write key.
+1. All secrets are re-encrypted with the new key.
+
+Pros:
+
+ - Simplicity.
+ - The generate/write/read interfaces can be pluggable for later replacement
+   with external secret management systems.
+ - A single master shouldn't require API Server downtime for rotation.
+ - No unseal step on startup since the file is already present.
+ - Attacker with access to /rotate is a DoS at worst, it doesn't return any
+   keys.
+
+Cons:
+
+ - Coordination of keys between a deployment with multiple masters will require
+   updating the KeyEncryptionKeyDatabase file on disk and forcing a re-read.
+ - Users will be responsible for backing up the keyfile from the API server
+   disk.
+
+### Option 2: User supplied encryption key
+
+In this solution there is no KEK/DEK scheme, just single keys managed by the
+user. To enable encryption a user specifies the "user-supplied-key" encryption
+provider at api startup. Nothing is actually encrypted until the user calls PUT
+on a /rotate API endpoint:
+
+1. A new encryption key is created.
+1. The key is provided back to the caller for persistent storage. Within the
+   cluster, it only lives in memory on the master.
+1. The list of keys being used by the master is updated in memory so that the
+   new key is in the list of read keys.
+1. The list of keys being used by the master is updated in memory so that the
+   new key is is the current write key.
+1. All secrets are re-encrypted with the new key.
+
+On master restart the api server will wait until the user supplies the list of
+keys needed to decrypt all secrets in the database. In most cases this will be a
+single key unless the re-encryption step was incomplete.
+
+Pros:
+
+ - Simplicity.
+ - A single master shouldn't require API Server downtime.
+ - User is explicitly in control of managing and backing up the encryption keys.
+
+Cons:
+
+ - Coordination of keys between a deployment with multiple masters is not
+   possible. This would have to be added as a subsequent feature using a
+   consensus protocol.
+ - API master needs to refuse to start and wait on a decrypt key from the user.
+ - /rotate API needs to be strongly protected: if an attacker can cause a
+   rotation and get the new key, it might as well not be encrypted at all.
+
+### Option 3: Encrypted DEKs in etcd, KEKs on disk
+
+In order to take an API driven approach for key rotation, new API objects (not
+exposed over REST) will be defined:
+
+* Key Encryption Key (KEK) - key used to unlock the Data Encryption Key. Stored
+  on API server nodes.
+* Data Encryption Key (DEK) - long-lived secret encrypted with a KEK. Stored in
+  etcd encrypted. Unencrypted in-memory in API servers.
+* KEK Slot - to support KEK rotation there will be an ordered list of KEKs
+  stored in the KEK DB. The current active KEK slot number, is stored in etcd
+  for consistency.
+* KEK DB - a file with N KEKs in a JSON list. KEK[0], by definition, is null.
+
+```go
+type DataEncryptionKey struct {
+    ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Value string    // Encrypted
+}
+```
+
+```go
+type KeyEncryptionKeySlot struct {
+    ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Slot int
+}
+```
+
+```go
+type KeyEncryptionKeyDatabase struct {
+    metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+    Keys []string
+}
+```
+
+To enable encryption a user must first create a KEK DB file and tell the API
+server to use it with `--encryption-provider-config=/path/to/config`.  The
+file will be a simple YAML file that lists all of the keys:
+
+```yaml
+kind: KeyEncryptionKeyDatabase
+version: v1
+keys:
+ - foo
+ - bar
+ - baz
+```
+
+The user will also need to specify the encryption provider and the resources to
+encrypt as follows:
+```yaml
+--encryption-provider-config=/path/to/key-encryption-key/db
+--encryption-provider=default
+--encrypt-resource=v1/Secrets
+--encrypt-resource=v1/ConfigMap
+```
+
+Then a user calls PUT on a /rotate API endpoint the first time:
+
+1. A new encryption key (unencrypted DEK) is created.
+1. Encrypt DEK with KEK[1]
+1. The list of DEKs being used by the master is updated in memory so that the
+   new key is in the list of read keys.
+1. The list of DEKs being used by the master is updated in etcd so that the
+   new key is in the list of read keys available to all masters.
+1. Confirm that all masters have the new DEK for reading. Key point here is that
+   all readers have the new key before anyone writes with it.
+1. The list of DEKs being used by the master is updated in memory so that the
+   new key is is the current write key.
+1. The list of DEKs being used by the master is updated in etcd so that the new
+   key is the current write key and is available to all masters. It doesn't
+   matter if there's some masters using the new key and some using the old key,
+   since we know all masters can read with the new key. Eventually all masters
+   will be writing with the new key.
+1. All secrets are re-encrypted with the new key.
+
+After N rotation calls:
+
+1. A new encryption key (unencrypted DEK) is created.
+1. Encrypt DEK with KEK[N+1]
+
+Each rotation generates a new KEK and DEK. Two DEKs will be in-use temporarily
+during rotation, but only one at steady-state.
+
+Pros:
+
+ - Most closely matches the pattern that will be used for integrating with
+   external encryption systems. Hashicorp Vault, Amazon KMS, Google KMS and HSM
+   would eventually serve the purpose of KEK storage rather than local disk.
+
+Cons:
+
+ - End state is still KEKs on disk on the master. This is equivalent to the much
+   simpler list of keys on disk in terms of key management and security.
+   Complexity is much higher.
+ - Coordination of keys between a deployment with multiple masters will require
+   manually generating and providing a key in the key file then calling rotate
+   to have the config re-read. Same as keys on disk.
+
+### Option 4: Protocol for KEK agreement between masters
+
+TODO: write a proposal for coordinating KEK agreement among multiple masters and
+having the KEK be either user supplied or backed by external store.
+
+
+## External providers
+
+It should be easy for the user to substitute a default encryption provider for
+one of the following:
+
+* A local HSM implementation that retrieves the keys from the secure enclave
+  prior to reusing the AES-GCM implementation (initialization of keys only)
+* Exchanging a local temporary token for the actual decryption tokens from a
+  networked secret vault
+* Decrypting the AES-256 keys from disk using asymmetric encryption combined
+  with a user input password
+* Sending the data over the network to a key management system for encryption
+  and decryption (Google KMS, Amazon KMS, Hashicorp Vault w/ Transit backend)
+
+### Backwards Compatibility
+
+Once a user encrypts any resource in etcd, they are locked to that Kubernetes
+version and higher unless they choose to manually decrypt that resource in etcd.
+This will be discouraged. It will be highly recommended that users discern if
+their Kubernetes cluster is on a stable version before enabling encryption.
+
+### Performance
+
+Introducing even a relatively well tuned AES-GCM implementation is likely to
+have performance implications for Kubernetes.  Fortunately, existing
+optimizations occur above the storage layer and so the highest penalty will be
+incurred on writes when secrets are created or updated.  In multi-tenant Kube
+clusters secrets tend to have the highest load factor (there are 20-40 resources
+types per namespace, but most resources only have 1 instance where secrets might
+have 3-9 instances across 10k namespaces).  Writes are uncommon, creates usually
+happen only when a namespace is created, and reads are somewhat common.
+
+### Actionable Items / Milestones
+
+* [p0] Add ValueTransformer to storage (Done in [#41939](https://github.com/kubernetes/kubernetes/pull/41939))
+* [p0] Create a default implementation of AES-GCM interface (Done in [#41939](https://github.com/kubernetes/kubernetes/pull/41939))
+* [p0] Add encryption flags on kube-apiserver and key rotation API
+* [p1] Add kubectl command to call /rotate endpoint
+* [p1] Audit of default implementation for safety and security
+* [p2] E2E and performance testing
+* [p2] Documentation and users guide
+* [p2] Read cache layer if encrypting/decrypting Secrets adds too much load on kube-apiserver
+
+
+## Alternative Considered: Encrypting the entire etcd database
+
+It should be easy for the user to substitute a default encryption provider for
+one of the following:
+
+Rather than encrypting individual resources inside the etcd database, another
+approach is to encrypt the entire database.
+
+Pros:
+
+ - Removes the complexity of deciding which types of things should be encrypted
+   in the database.
+ - Protects any other sensitive information that might be exposed if etcd
+   backups are made public accidentally or one of the other desribed attacks
+   occurs.
+
+Cons:
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+Other resources can use a no-op provider by default.
+
+## Encryption Provider
+
+An encryption provider implements the ValueTransformer interface. Out of the box
+this proposal will implement encryption using a standard AES-GCM performing
+AEAD, using the standard Go library for AES-GCM.
+
+Each encryption provider will have a unique string identifier to ensure
+versioning of contents on disk and to allow future schemes to be replaced.
```
jcbsmpsn: versioning of the [cipher text] on disk and to allow future schemes to be [added].

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+
+During encryption, only a single provider is required.  During decryption,
+multiple providers or keys may be in use (when migrating from an older version
+of a provider, or when rotating keys), and thus the ValueTransformer
+implementation must be able to delegate to the appropriate provider.
```
jcbsmpsn: Isn't the ValueTransformer implementation the same as an encryption provider?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+block := aes.NewCipher(primaryKeyString)
+aead := cipher.NewGCM(c.block)
+keyId := primaryKeyId
+
+// for each write
+nonce := make([]byte, block.BlockSize())
+io.ReadFull(crypto_rand.Reader, nonce)
+authenticatedData := ETCD_KEY
+cipherText := aead.Seal(nil, nonce, value, authenticatedData)
+storedData := providerId + keyId + base64.Encode(nonce + authenticatedData + cipherText)
```
jcbsmpsn: Maybe this is just the pseudo code, but couldn't this use a struct to store these separate pieces instead of string munging/byte manipulation?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+
+### Option 1: Simple list of keys on disk
+
+In this solution there is no KEK/DEK scheme, just single keys in a list on disk.
+To enable encryption a user calls PUT on a /rotate API endpoint:
+
+1. A new encryption key is created.
+1. The key is added to a file on the API master with metadata including an ID
+   and an expiry time. Subsequent calls to rotate will prepend new keys to the
+   file such that the first key is always current.
```
jcbsmpsn: such that the first key is always the [key to use for encryption].

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+ - Users will be responsible for backing up the keyfile from the API server
+   disk.
+
+### Option 2: User supplied encryption key
+
+In this solution there is no KEK/DEK scheme, just single keys managed by the
+user. To enable encryption a user calls PUT on a /rotate API endpoint:
+
+1. A new encryption key is created.
+1. The key is provided back to the caller for persistent storage. It only lives
```
jcbsmpsn: for persistent storage. [Within the cluster, it] only lives

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+keys:
+ - foo
+ - bar
+ - baz
+```
+
+The user will also need to specify the encryption provider and the resources to
+encrypt as follows:
+```yaml
+--key-encryption-key-db-path=/path/to/key-encryption-key/db
```
jcbsmpsn: s/--key-encryption-key-db-path/--encryption-provider-config/

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+   new key is in the list of read keys.
+1. The list of DEKs being used by the master is updated in etcd so that the
+   new key is in the list of read keys available to all masters. TODO: seems
+   racey?
+1. The list of DEKs being used by the master is updated in memory so that the
+   new key is is the current write key.
+1. The list of DEKs being used by the master is updated in etcd so that the new
+   key is the current write key and is available to all masters. TODO: Is there
+   a race here where the etcd write is finished but not all masters have the
+   data?
```
jcbsmpsn: Yes, there is a lag where some master may not be using the new write key, but as long as they all have it as a read key, it isn't important. Some will write with the new key, some will write with the old key, but all will be able to read with either key.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+
+Then a user calls PUT on a /rotate API endpoint the first time:
+
+1. A new encryption key (unencrypted DEK) is created.
+1. Encrypt DEK with KEK[1]
+1. The list of DEKs being used by the master is updated in memory so that the
+   new key is in the list of read keys.
+1. The list of DEKs being used by the master is updated in etcd so that the
+   new key is in the list of read keys available to all masters. TODO: seems
+   racey?
```
jcbsmpsn: Is the race condition because not all masters will have the read key at the same time? If so, as long as no one is writing with that key, it doesn't matter. This step prepares all the masters for the safe transition to writing with the new key.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+
+1. A new encryption key (unencrypted DEK) is created.
+1. Encrypt DEK with KEK[N+1]
+
+Each rotation generates a new KEK and DEK. Two DEKs will be in-use temporarily
+during rotation, but only one at steady-state.
+
+Pros:
+
+ - ?
```
jcbsmpsn: Most closely matches the pattern that will be used for integrating with external encryption systems. Amazon KMS, Google KMS and HSM will serve the purpose of KEK storage rather than local disk.

contributors/design-proposals/encryption.md
smarterclayton: In this case yes, although future work might compress at rest.  We should clarify value transformer is a storage detail, which is not strictly limited to encryption although intended for

contributors/design-proposals/encryption.md
smarterclayton: The stored data being a struct has been discussed - it comes with additional complexity in cost and perf.  Prefixing is the cheapest option that requires no new serialization struct.

Our proto at rest format follows this pattern (magic prefix to make the bytes recognizable).

contributors/design-proposals/encryption.md
smarterclayton: Yes.  Make new key available everywhere before anyone writes.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+any resource, but will be used by default for the Secret resource. Secrets are
+already protected in transit via TLS.
+
+Full disk encryption of the volumes storing etcd data is already expected as
+standard security hygiene.  Adding the proposed encryption inside the database
+defends against malicious parties gaining access to:
+
+- The etcd API
+- etcd backups; or
+- A running etcd instance without access to memory of the etcd process.
```
philips: the keys are encrypted before hitting etcd, how does it change with or without memory access?

contributors/design-proposals/encryption.md
destijl: Done.

contributors/design-proposals/encryption.md
destijl: Added.

destijl:
@andrewsykim @stevesloka I think you may have had use cases for this. Can you comment on the three options above? In particular if option 1 satisfies any hard requirements you may have?

destijl:
I just added a configuration section to call out the api server options we're proposing adding. @jcbsmpsn @smarterclayton are we ready to make the call on going with option 1? FWIW it gets my vote.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,403 @@
...
+of a provider, or when rotating keys), and thus the ValueTransformer
+implementation must be able to delegate to the appropriate provider.
+
+### AES-GCM Encryption provider
+
+Implemented in [#41939](https://github.com/kubernetes/kubernetes/pull/41939).
+
+TODO: Decide if
+[secretbox](https://godoc.org/golang.org/x/crypto/nacl/secretbox) should be used
+instead.
```
smarterclayton: Other than 

1. pulling in a newer library (which has some small risk) which is not part of go standard crypto
2. requiring us to manage AEAD ourselves (bigger concern)
3. the cache attack is not much of a concern on x86 with AES-NI, but is more so on ARM

I'm not terribly concerned.  I'm actually less concerned with having two impls than just having secretbox in the short term.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+STORED_DATA := <PROVIDER_ID>:<KEY_ID>:<NONCE><AUTHENTICATED_DATA><CIPHER_TEXT>
+```
+
+Pseudo-code for encrypt (golang):
+```go
+block := aes.NewCipher(primaryKeyString)
+aead := cipher.NewGCM(c.block)
+keyId := primaryKeyId
+
+// string prefix chosen over a struct to mimimize complexity and for write
```
kargakis: typo

destijl: thanks, fixed.

stevesloka:
@destijl did my comments show up? I don't see now, but did from phone, wanted to make sure I replied to your comment.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+out, err := aead.Open(nil, nonce, cipherText, authenticatedData)
+```
+
+## Configuration
+
+We will add the following options to the API server. At API server startup the
+user will specify:
+
+```yaml
+--encryption-provider-config=/path/to/config
```
kargakis: Provide an example configuration?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+ - API master needs to refuse to start and wait on a decrypt key from the user.
+ - /rotate API needs to be strongly protected: if an attacker can cause a
+   rotation and get the new key, it might as well not be encrypted at all.
+
+### Option 3: Encrypted DEKs in etcd, KEKs on disk
+
+In order to take an API driven approach for key rotation, new API objects will
+be defined:
+
+* Key Encryption Key (KEK) - key used to unlock the Data Encryption Key. Stored
```
kargakis: Is this an API object?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+1. A new encryption key (unencrypted DEK) is created.
+1. Encrypt DEK with KEK[1]
+1. The list of DEKs being used by the master is updated in memory so that the
+   new key is in the list of read keys.
+1. The list of DEKs being used by the master is updated in etcd so that the
+   new key is in the list of read keys available to all masters.
+1. Confirm that all masters have the new DEK for reading. Key point here is that
+   all readers have the new key before anyone writes with it.
+1. The list of DEKs being used by the master is updated in memory so that the
+   new key is is the current write key.
```
kargakis: typo

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+
+1. A new encryption key is created.
+1. The key is added to the config file on the API master with metadata including
+   an ID and an expiry time. Subsequent calls to rotate will prepend new keys to
+   the file such that the first key is always the key to use for encryption.
+1. The list of keys being used by the master is updated in memory so that the
+   new key is in the list of read keys.
+1. The list of keys being used by the master is updated in memory so that the
+   new key is is the current write key.
+1. All secrets are re-encrypted with the new key.
```
mayakacz: What's the expected volume of secrets here? Is there a performance impact of choosing only a single layer of keys vs. envelope encryption (DEK/KEK)?
A lot of customers I talk to are in the 100-1000 secrets range, so this should be fine, but worth considering.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+   will be writing with the new key.
+1. All secrets are re-encrypted with the new key.
+
+After N rotation calls:
+
+1. A new encryption key (unencrypted DEK) is created.
+1. Encrypt DEK with KEK[N+1]
+
+Each rotation generates a new KEK and DEK. Two DEKs will be in-use temporarily
+during rotation, but only one at steady-state.
```
mayakacz: Why does the rotation need to generate a new DEK? Or rather, in a non-emergency rotation (say, with automated rotation), this adds a lot of complexity.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+Pros:
+
+ - Most closely matches the pattern that will be used for integrating with
+   external encryption systems. Amazon KMS, Google KMS and HSM would eventually
+   serve the purpose of KEK storage rather than local disk.
+
+Cons:
+
+ - End state is still KEKs on disk on the master. This is equivalent to the much
+   simpler list of keys on disk in terms of key management and security.
```
mayakacz: My preferred end state is a combination of 2 & 3 - local DEK are generated and sent elsewhere to be encrypted with KEK. In this implementation, it may be a local KEK on disk, but longer term, there could be other backends and others. This also circumvents the KEK agreement bit because in 4 because they just need to agree on a single source which maintains the KEKs, rather than agree on the KEKs themselves.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+having the KEK be either user supplied or backed by external store.
+
+
+## External providers
+
+It should be easy for the user to substitute a default encryption provider for
+one of the following:
+
+* A local HSM implementation that retrieves the keys from the secure enclave
+  prior to reusing the AES-GCM implementation (initialization of keys only)
```
mayakacz: Is this 1 with some sort of root key that unlocks all the disk keys?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+future.
+
+To enable key rotation a common pattern is to have keys used for resource
+encryption encrypted by another set of keys (Key Encryption Keys aka KEK).  The
+keys used for encrypting kubernetes resources (Data Encryption Keys, aka DEK)
+are generated by the apiserver and stored encrypted with one of the KEKs.
+
+In future versions, storing a KEK off-host and off-loading encryption/decryption
+of the DEK to AWS KMS, Google Cloud KMS, Hashicorp Vault etc. should be
+possible. The decrypted DEK would be cached locally after boot.
```
mayakacz: Is the DEK forcibly cached in memory? Or do we want to put an external store in the serving path of a secret? (this likely also depends on the sensitivity of the workload, where it's hosted/ latency to external store, etc.). I would propose that if one of the DEK/KEK solutions is pursued, this should be a design option.

contributors/design-proposals/encryption.md
philips: The introductory sentence doesn't make sense to me. The proposal is adding encryption _before_ the database, yet this paragraph says: "Adding the proposed encryption inside the database".

Line 24 below says the exact opposite.

contributors/design-proposals/encryption.md
philips: It is an API object but won't be exposed over REST. It would be like a component config or kubeconfig.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+1. Encrypt DEK with KEK[N+1]
+
+Each rotation generates a new KEK and DEK. Two DEKs will be in-use temporarily
+during rotation, but only one at steady-state.
+
+Pros:
+
+ - Most closely matches the pattern that will be used for integrating with
+   external encryption systems. Amazon KMS, Google KMS and HSM would eventually
+   serve the purpose of KEK storage rather than local disk.
```
philips: I took this design almost exactly from hashicorp vault, so it should map there as well.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+
+Cons:
+
+ - End state is still KEKs on disk on the master. This is equivalent to the much
+   simpler list of keys on disk in terms of key management and security.
+   Complexity is much higher.
+ - Coordination of keys between a deployment with multiple masters will require
+   shutting down all masters, and manually generating and providing a key in the
+   key file. Same as keys on disk.
```
philips: Why do the masters need to shut down? It should just be replacing the KeyEncryptionKeyDatabase file on disk and then when rotate is called the database is re-read. It should make coordination easier without downtime.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+
+Before a resource is written to etcd and after it is read, an encryption
+provider will take the plaintext data and encrypt/decrypt it.  These providers
+will be able to be created and turned on depending on the users needs or
+requirements and will adhere to an encryption interface.  This interface will
+provide the abstraction to allow various encryption mechanisms to be
+implemented, as well as for the method of encryption to be rotated over time.
+
+For the first iteration, a default provider that handles encryption in-process
+using a locally stored key on disk will be developed.
```
lpabon: Should this be an interface for Key retrieval?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+TODO: Decide if
+[secretbox](https://godoc.org/golang.org/x/crypto/nacl/secretbox) should be used
+instead.
+
+The simplest possible provider is an AES-GCM encrypter/decrypter using AEAD,
+where we create a unique nonce on each new write to etcd, use that as the IV for
+AES-GCM of the value (the JSON or protobuf data) along with a set of
+authenticated data to create the ciphertext, and then on decryption use the
+nonce and the authenticated data to decode.  The keys come from configuration on
+the local disk (potentially decrypted at startup time using a stronger password
```
lpabon: Keys coming from local disk seems like a limitation to the design. Wouldn't this be difficult to manage by administrators of the cluster?

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,429 @@
...
+protected key, or in the future from an alternative source).
+
+The provider will be assigned a versioned identifier to uniquely pair the
+implementation with the data at rest, such as “k8s-aes-gcm-v1”.  Any
+implementation that attempts to decode data associated with this provider id
+must follow a known structure and apply a specific algorithm.
+
+The provider would take a set of keys and unique key identifiers from the
+command line, with the key values stored on disk.  One key is identified as the
```
lpabon: I had this question on 454 also. Do you mind describing a little more by what it means "from the command line"?

contributors/design-proposals/encryption.md
destijl: Trying to reason through this (feel free to correct me):

1. Rotation of just the KEK time-limits exposure of the previous KEK where the attacker didn't also retrieve the DEKs.

2. DEKs need to be rotated every 2^32 writes somehow.

3. We need to rotate everything if DEKs and KEKs got exposed.

Given we need to write code for 2 and 3, is there much advantage shooting just for 1?

contributors/design-proposals/encryption.md
destijl: yes, I think doing it like this only really makes sense if you use the external store. I'll reword this one when I get a chance.

contributors/design-proposals/encryption.md
destijl: yep, that works.

contributors/design-proposals/encryption.md
destijl: I think? this is saying there's only one layer of keys, and they are retrieved from the HSM. So it's like option 1 with the keys in the HSM instead of on disk. @philips wrote this originally I believe.

contributors/design-proposals/encryption.md
destijl: needs more thinking, but yes.

contributors/design-proposals/encryption.md
destijl: added.

destijl:
@stevesloka I don't see any. I saw your comment on the issue about HIPAA...?

destijl:
@philips let me coordinate with @jcbsmpsn who is working on code and choose the option.

contributors/design-proposals/encryption.md
smarterclayton: We routinely deal with clusters that have 100k and up secrets.  In practice, it doesn't matter how many they are, they still have to be rotated.  However, they don't have to be rotated synchronously - the whole point of having multiple keys is to allow that process to be done in the background.  You have to execute a no-op PUT to every secret and verify you get a resource change (no resource change means the secret is already stored with the new key).  Once you've completed that, you can remove the old key (it's zero cost to leave it, other than search order though).

smarterclayton: (this is the same feature we use to move stored API resources to a new storage version)

mayakacz: ack. This seems like me to a push to envelope encryption model

contributors/design-proposals/encryption.md
mayakacz: re: 2, ahh I see my confusion. Recommendation would be to generate a new DEK every write (not change on reads)
re: 3, yes, but this is (hopefully) a rare case. We still need to write code but it shouldn't be built into the 'default' rotation model.

contributors/design-proposals/encryption.md
mayakacz: I think that we have caching be the default, but might want to make this an option (not high priority).

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,436 @@
...
+  networked secret vault
+* Decrypting the AES-256 keys from disk using asymmetric encryption combined
+  with a user input password
+* Sending the data over the network to a key management system for encryption
+  and decryption (Google KMS, Amazon KMS, Hashicorp Vault w/ Transit backend)
+
+### Backwards Compatibility
+
+Once a user encrypts any resource in etcd, they are locked to that Kubernetes
+version and higher unless they choose to manually decrypt that resource in etcd.
```
smarterclayton: I just realized - we probably need to at least think about how to disable encryption.  That means we need a config way to have decrypter keys in rotation, but ensure the write key is "none"

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,436 @@
...
+ - Attacker with access to /rotate is a DoS at worst, it doesn't return any
+   keys.
+
+Cons:
+
+ - Coordination of keys between a deployment with multiple masters will require
+   shutting down all masters, and manually generating and providing a key in the
+   key file.
+ - Users will be responsible for backing up the keyfile from the API server
+   disk.
```
simo5: You can avoid this con with a kvno (key version number) and allowing a master that does not have a new version key to call out to get it dynamically.
The only problem left is that until all masters get to the new key you may get new secrets stored with an old key.
Other software deal with this by preloading a new key, but not activating it for encryption until there is confirmation all masters have the new key, at that point all switch to use the new key, and after a short delay a batch job to rotate all keys is started.

contributors/design-proposals/encryption.md
simo5: you should rather have a special none provider that has to be very explicitly configured, allowing "none" keys is generally dangerous.

jessfraz:
So one problem we might want to think about is this is solving the problem of secrets being encrypted at rest. But if a user was to set a secret as en environment variable and then say use docker or runc to ultimately run their container, the secret will be decrypted in the runc config used by containerd, in the containerd implementation in docker this file is stored on disk so essentially the secret is decrypted on disk

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,436 @@
...
+
+The provider will use the recommended Go defaults for all crypto settings
+unless otherwise noted.  We should use AES-256 keys (32 bytes).
+
+Process for transforming a value (object encoded as JSON or protobuf) to and
+from stable storage will look like the following:
+
+Layout as written to etcd2 (json safe string only):
+```
+NONCE := read(/dev/urandom)
```
tiran: A CSPRNG is **not** a safe nonce generator for AES-GCM. For GCM you **must** ensure that two messages are never encrypted with the same key and nonce. Nonce misuse breaks at least semantic security, probably much more. Adam Langley's blog post https://www.imperialviolet.org/2017/05/14/aesgcmsiv.html explains some of the issues with GCM and a new GCM variant that does not break catastrophically.

I assume mandatory key rotation before 2^32 writes is to avoid nonce misuse. The issue should be stated in your threat analysis of our design.

smarterclayton: That was called out in the original proposal, but was omitted.  Yes, we are relying on birthday collision defense for a random IV.  

Given that we have no AES-GCM-SIV implementation at this point, and the expected write volumes, AND that to make this attack feasible you have to observe a lot of writes in order to find reuse (and that reuse wouldn't be on the same cipher text), AND that the attacker would have to have persistent read access to a running system for that period of time (to observe reuse, and the threat model is still targeted at someone observing the store at rest), it was judged sufficient.  That should be in the doc.

Agree that SIV would be superior.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,436 @@
...
+datastore layer with sufficient metadata support to enable integration with
+multiple encryption providers and key rotation. Encryption will be optional for
+any resource, but will be used by default for the Secret resource. Secrets are
+already protected in transit via TLS.
+
+Full disk encryption of the volumes storing etcd data is already expected as
+standard security hygiene.  Adding the proposed encryption at the datastore
+layer defends against malicious parties gaining access to:
+
+- The etcd API
```
smarterclayton: It does not.  Someone gaining write access to etcd is effectively root on the entire cluster, the same as any other core database.  Secrets are the least of the problems at that point.  Agree should be separated.

contributors/design-proposals/encryption.md
smarterclayton: That's in the implementation now.

contributors/design-proposals/encryption.md
```diff
@@ -0,0 +1,436 @@
...
+// ensure this provider is the one handling providerId
+aead := // lookup an aead instance for keyId or error
+bytes := base64.Decode(base64Encoded)
+nonce, authenticatedData, cipherText := // slice from bytes
+out, err := aead.Open(nil, nonce, cipherText, authenticatedData)
+```
+
+### Alternative Considered: SecretBox
+
+Using [secretbox](https://godoc.org/golang.org/x/crypto/nacl/secretbox) would
```
tiran: SecretBox is a viable alternative to AES-CBC. Although SecretBox is based on a stream cipher and stream ciphers like AES-GCM are in general not NMR (nonce-misuse resistant), SecretBox is safe. It uses XSalsa20, which is a variant of Salsa20 like ChaCha20. XSalsa20 is especially crafted to be resilient against nonce reuse. It takes a 256bit key and a 196bit nonce. The 128bit of the nonce are used to derive a new subkey from the input key. The remaining 64bit of the nonce and the subkey are then used for encryption with Salsa20 and Poly1305 for authentication. The 196 bit nonce space makes it safe to generate nonces from a CSPRNG. (For comparison, that's more nonces than atoms in the solar system).

There is one catch:
Contrary to AES-GCM, SecretBox is not an AEAD. It only guarantees integrity. A malicious admin can copy an existing secret to another etcd name without violating integrity check of the encryption layer. You'd have to come up with a custom scheme to include integrity (if desired).

tiran: PS: AFAIK Salsa20 is slower than AES-CBC on modern machines with AES-NI.

smarterclayton: I'll do a quick prototype of secretbox and benchmark against the GCM with NI.  I was seeing ~2-3us to encode which is well within our budget.  If secretbox is within 5-6us i think we can afford it, if it's substantially simpler to leverage than CBC.

Our current threat model does not assume an admin able to copy arbitrary cipher text into a key space under the attackers control (the secrets in a namespace that belongs to them) because write access to etcd trivially allows the user to escalate to viewing the secrets.  So for an attacker with read access to etcd we can live without AEAD.

contributors/design-proposals/encryption.md
smarterclayton: Given the concerns raised with GCM, I'll do a quick prototype of SecretBox and (maybe) AES-CBC and verify we're within the bounds.

contributors/design-proposals/encryption.md
tiran: SecretBox has the advantage that you don't have to take care of padding and integrity checks yourself. Downside besides lack of hardware acceleration is missing FIPS / CommonCriteria conformity, since neither Salsa nor Poly1305 are approved algorithms. CC is relevant for three-letter-acronym users with a ``.gov`` or ``.mil`` TLD. :)

Since your threat model does not include arbitrary modifications of encrypted secrets and all communication is already secured and integrity validated by TLS, you may use plain AES-CBC with PKCS7v1.5 padding and no MAC. If you prefer MAC, then I recommend to use the Fernet standard algorithm, https://github.com/fernet/spec/blob/master/Spec.md. It's AES-CBC with HMAC-SHA256 encrypt-than-mac.

smarterclayton: Pushed an alternative to https://github.com/kubernetes/kubernetes/pull/46916.  2x slower which is acceptable.

@jcbsmpsn @sakshamsharma if we decide to go with secretbox we'll need to update the config change to make this the preferred outcome.

mayakacz: I would push us towards something like AES-GCM or -CBC due to wider acceptance in regulated industries (even if whatever particularly implementation is here is not FIPS or CC compliant), and hardware acceleration. There will be noticeable performance improvement at scale.
If the issue is just getting support from crypto folks for proper implementation, I think we can solve that.

simo5: AES-GCM has nonce reuse issues, but AES CBC does not.
AEs-CBS is not aead, so, on its own it does not give you integrity nor authentication protection, but afaik, these are not really needed.
First of all, no other data in ETCD is integrity protected, so it doesn't seem like a big priority to get integrity protection just for one piece of data in the database. And athentication is really generally only needed when an attackare can manipulate data either by injecting random data in a stream, or by swapping packets or data in a stream or on a disk. Neither of these cases seem to be a concern for this use case, as all communications are already protected by TLS, and we assume that if an attacker can get write access to the ETCD API, we have much bigger issues to care about than integrity or authentication for the secrets.
Even if an attacker can cause the API server to write encrypted packets (by writing random secrets) and has somehow read access to the ETCD API so it can observe all secrets, the API server is still using a random nonce to encrypt each secret.
Therefore even just using AES-CBC without integrity or authentication, the attacker would have to try to write more than 2^N nonce to have a chance to see a collision and generate the same block, and this is for each possible secret combination. In other words, this is the same as doing a full brute force on a secret, which is unfeasible in the first place and can be done completely offline anyway if someone has read access.

smarterclayton:
I tried CBC - it was significantly slower.  I think based on the feedback
above we should consider having two OOTB modes for 1.7 - AES-CBC (2-5x
slower than accelerated GCM), and Secretbox (2x slower than accelerated
GCM).  We can then either remove the GCM mode or recommend against it.

Given that we are using the Golang CBC and it's unlikely in the near term
we'll have accelerated CBC without reverting to gcc-go (which is a giant
can of worms that is impractical for lots of other reasons) I think we'll
simply say you can do "slow and CC compliant" or "faster and not CC
compliant".

simo5:
I think it is ok to offer multiple options, I would lean on making AES-CBC the default as it is "safer" than AES-GCM (does not require key rotation), and leave to the user to decide if they want "faster but riskier" option.

Other than that the code looks ok as well, so up to you what you want to set as default.

tiran:
@smarterclayton Does Golang use native instructions only for GCM? AES-GCM and AES-CBC without integrity checks should have the performance characteristics on AES-NI capable machines. Both modes use the same AES instructions like ``AESENC``.

contributors/design-proposals/encryption.md
jcbsmpsn: I believe the point here is encrypting the data at rest within etcd defends against someone who has direct etcd access from listing the secrets. It is not actually about protecting the etcd API.

smarterclayton:
> Does Golang use native instructions only for GCM?

No, it uses it for CBC as well.  Which is bizarre because the numbers are so divergent.

smarterclayton:
To clarify here - the linked PR has AES-CBC with PKCS#7 padding.

simo5:
I only had really a nitpick about naming, but otherwise PKCS#7 padding looks correct.

enj:
@smarterclayton @tiran CBC is probably slower because it has this extra step (which is only optimized with native instructions on IBM s390x):

```go
func (x *cbcEncrypter) CryptBlocks(dst, src []byte) {
	if len(src)%x.blockSize != 0 {
		panic("crypto/cipher: input not full blocks")
	}
	if len(dst) < len(src) {
		panic("crypto/cipher: output smaller than input")
	}

	iv := x.iv

	for len(src) > 0 {
		// Write the xor to dst, then encrypt in place.
		xorBytes(dst[:x.blockSize], src[:x.blockSize], iv)
		x.b.Encrypt(dst[:x.blockSize], dst[:x.blockSize])

		// Move to the next block with this block as the next iv.
		iv = dst[:x.blockSize]
		src = src[x.blockSize:]
		dst = dst[x.blockSize:]
	}

	// Save the iv for the next CryptBlocks call.
	copy(x.iv, iv)
}
```

edknapp:
Hey folks, could someone explain (formally) what security properties AES-CBC has that are lacked by AES-GCM? 

While it's true that AES-GCM lacks CCA2 security in presence of nonce collisions, unauthenticated AES-CBC is *not* CCA2 secure in the first place. *authenticated* AES-CBC is CCA2 security, but not with nonce collisions. 

My suggestion would be to drop AES-CBC entirely in favor of AES-GCM and secretbox.

simo5:
In GCM nonce collisions are catastrophic, plaintext is revealed, in AES CBC they are not.
Given the distributed nature of ETCD, it is impossible to guarantee nonce non-reuse.

tiran:
@edknapp AES-GCM turns AES into a stream cipher. When two cipher texts are encrypted with the same key and IV, then the XOR of both encrypted cipher texts is equal to the XOR of both plain texts: ``ct1 ^ ct2 == pt1 ^ pt2``. The encrypted is cancelled out. An attacker is able to learn a lot about the structure of your data. If an attacker also happens to know one of the plain texts, then he gets the other plain text for free: ``ct1 ^ ct2 ^ pt1 = pt2``.

That's why the IV for any counter mode must be a nonce (number only used once) and not be controllable by anybody but the encrypter. A couple of weeks ago I visualized the issue with three pictures, https://github.com/tiran/militarygrade/

tiran:
What is your attack vector against AES-CBC that requires an AEAD or message integrity? In our threat analysis we rely on the fact that all communication is protected against modification and eves-dropping by TLS. We assume that TLS provides strong encryption and authentication. AES-CBC is used to implement encryption at rest to protect against offline attacks (e.g. stolen backups of etcd database). If an attacker is able to break TLS, break authentication or write to etcd database directly, then the cluster is compromised no matter what. An AE or MAC won't help either.

edknapp:
Thank you for the clarification. I'm informed that write access to the database effectively gives code execution, so the authentication isn't important. 

I'm curious about favoring a AES-CBC over secretbox, can someone comment on that? 

Regardless, I don't see any particular advantage in dropping AES-CBC, other than secretbox having a larger nonce size (thus fewer collisions), and perhaps a philosophical preference to minimize supported algorithms and prefer modern crypto (all else equal).

simo5:
AES can be accelerated via AES-NI extensions, and I would not say AES is "not modern", I do not know if there is HW acceleration available for XSalsa20 and Poly1305 on common HW yet. AES is also NIST/FIPS approves, secretbox isn't.

edknapp:
Your performance argument is convincing, continuing to support AES-CBC makes sense. Thank you for time. 

(I was referring to unauthenticated encryption as "not modern". I agree that AES is modern.)

tiran:
@edknapp Secretbox is XSalsa20 + Poly1305. XSalsa20 is not free from the nonce misuse problems. It merely uses a very large nonce. In combination with a strong, high quality random source (CSPRNG like getrandom() syscall), a nonce collision is unlikely to happen. [1]. In theory a 192bit nonce is safe for the remaining lifetime of our universe. In practice you have to be careful that your virtualization software injects good randomness or HW random generator into your VMs.

[1] https://download.libsodium.org/doc/advanced/xsalsa20.html

smarterclayton:
I forgot this hasn't merged.  We need to freshen it and get it in.

philips:
Bump, we do need to get this merged!

philips:
Also, I still don't see the final recommendation of what is implemented in Kubernetes 1.8. Did that get added? Link to the line number? https://github.com/kubernetes/community/pull/607#pullrequestreview-38984817

sakshamsharma:
[Issue about addition of Google Cloud KMS integration](
https://github.com/kubernetes/kubernetes/issues/48522)

[Encryption at rest user guide from v1.7](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/)

liggitt:
@destijl can we get this merged. would like to fold https://docs.google.com/document/d/1S_Wgn-psI0Z7SYGvp-83ePte5oUNMr4244uanGLYUmw/edit# into this doc

contributors/design-proposals/encryption.md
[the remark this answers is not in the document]
destijl: What jacob said. Intent was if you have access to etcd API but not the master pieces doing encrypt/decrypt then you only see encrypted secrets. But it's probably just confusing. Removed.

liggitt:
agree on merging to capture current state, and to provide a base for future KEPs to reference/extract from

/lgtm
/approve

PR #609, opened 2017-05-09
Add information on running local cluster

crimsonfaith91:
- also include tip to check logs

### DIFF
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -117,6 +117,18 @@ Kubernetes maintains state in [`etcd`][etcd-latest], a distributed key store.
 
 Please [install it locally][etcd-install] to run local integration tests.
 
+### Development using local cluster
+After installing etcd, we recommend running a local cluster for development purpose.
+```
+cd go/src/k8s.io/kubernetes
+hack/local-up-cluster.sh
+```
+
+Tip: After adding new glog messages, use the command below to access the logs:
+```
+tail -f /tmp/<log-file-name>
+```
+
 ### Go
 
 Kubernetes is written in [Go](http://golang.org).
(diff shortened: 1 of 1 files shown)


contributors/devel/development.md
```diff
@@ -117,6 +117,9 @@ Kubernetes maintains state in [`etcd`][etcd-latest], a distributed key store.
 
 Please [install it locally][etcd-install] to run local integration tests.
 
+### Development using local cluster
+After installing etcd, we recommend running a local cluster for development purpose. Refer this [tutorial](https://kubernetes-v1-4.github.io/docs/getting-started-guides/locally/) for more information.
```
MHBauer: Format in shorter lines, please.

That doesn't seem like a very helpful link to me.

It immediately says, stop, go use minikube, and minikube cannot be used for development so, it is deceptive, if anything.

Possibly link to 
https://kubernetes-v1-4.github.io/docs/getting-started-guides/locally/#starting-the-cluster
or 
https://kubernetes-v1-4.github.io/docs/getting-started-guides/locally/#i-changed-kubernetes-code-how-do-i-run-it

That whole document should probably go away and and any mention of `hack/local-up-cluster.sh` should be mentioned inline.

crimsonfaith91:
@MHBauer Both local cluster and minikube bring up one-node cluster. The local cluster is the suggested way to start a development cluster, while the minikube is more for "trying out" a stable release rather than for development.

I am following up with cluster lifecycle team about the best way to set up a cluster with multiple nodes. You can check the update here: https://groups.google.com/forum/#!topic/kubernetes-sig-cluster-lifecycle/Zp3Y4uNHnPQ

crimsonfaith91:
@foxish are we going to merge this?

crimsonfaith91:
@MHBauer Will this be fine for merging?

crimsonfaith91:
i will close this due to no reply. feel free to open it again with feedback, and i will continue working on it. :)
thanks!

PR #617, opened 2017-05-12
Proposal: Watchable Discovery API

sttts:
This is about an improvement of the discovery story in a dynamic apiserver worlds where group come and go much more often.

A first alpha version is planned for 1.7. We can use a feature toggle to switch the controllers. Goal is to have this as the default for the controllers in 1.8.

Prototyping is happening in https://github.com/kubernetes/kubernetes/pull/45659.

@kubernetes/sig-api-machinery-api-reviews @kubernetes/sig-api-machinery-misc @deads2k @caesarxuchao @smarterclayton @lavalamp @liggitt

### DIFF
--- a/contributors/design-proposals/watchable-discovery.md
+++ b/contributors/design-proposals/watchable-discovery.md
@@ -1,0 +1,231 @@
+# Watchable API Group Discovery
+
+The Kubernetes API server provides a number of HTTP endpoinds to
+discover the available API groups, versions and resources:
+
+- `/apis`: returns a list of groups with available versions (compare [metav1.APIGroupList](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L639))
+- `/apis/<group>`: returns a list of versions, **without** resources (compare [metav1.APIVersions](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L623))
+- `/apis/<group>/<version>`: returns a list of resources (compare [metav1.APIResourceList](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L718))
+
+This discovery information is used by many API clients like kubectl
+and controllers like the namespace controller. The garbage
+collection controller is supposed to use discovery as well, [but
+doesn't yet](https://github.com/kubernetes/kubernetes/blob/6f5337cff7baffc2b27d85036cb1ecb6a4c1f73b/cmd/kube-controller-manager/app/core.go#L174).
+
+Kubectl will cache the discovery information in
+`~/.kube/cache` for ten minutes and queries the API server again after
+that time or on a cache miss on RESTMapper lookups. In the namespace controller
+a [polling logic is implemented](https://github.com/kubernetes/kubernetes/blob/6f5337cff7baffc2b27d85036cb1ecb6a4c1f73b/cmd/kube-controller-manager/app/core.go#L115).
+
+With ThirdPartyResources and API aggregation the discovery information
+is not static anymore within a deployed Kubernetes version. It can
+change at any time, either because additional API servers are added to
+or removed from the cluster, or because the user creates a
+ThirdPartyResource. For consistent handling of the new resources in
+kubectl and the controllers, they have to adapt to the changed API
+groups, prefereably in real-time or at least seconds.
+
+The discovery API is not tailored for many API groups and versions. To
+get a list of all GroupVersionResources in the cluster, a client has
+to query the available groups (with their versions through `/apis`)
+and then the resources for each GroupVersion. In addition, retry logic
+is involved to make this reliable. This can easily lead to
+20 and more discovery requests, slowing down especially an interactive
+tool like kubectl. This was the reason to add the cache as a
+workaround, with the danger that discovery infos is outdates for up to
+10 minutes.
+
+The discovery API is not versioned. Incompatible changes will break
+old client therefore.
+
+The discovery API does not use the API semantics of API group resource
+endpoints. I.e. it is not possible to query with selectors, to watch
+for changes or to use protobuf for more efficient transport. Moreover,
+discovery info does not carry a `ResourceVersion` such that it's
+impossible to compare cache information with the last version in the
+API server.
+
+## Goals
+
+- near real-time updates without polling
+- better scalability to more groups and versions with less HTTP requests
+  - one request which returns all GroupVersionResources
+- versioning
+- default Kubernetes API semantics including:
+  - watchability
+  - resource versioning
+  - protobuf support
+  - compatibility with all existing machinery like clients,
+    ListWatches, informers, caching, etc.
+- compatibility with kube-aggregator, third-party apiservers and
+  ThirdPartyResources
+
+### Non-goals
+
+- different view for different clients: e.g. kubectl 1.5 won't see
+  `batch/v1`
+- ETag like response without content for even better scalability and
+  responsiveness
+- readability of certain `Group` objects for unauthenticated users
+  (e.g. to show those APIs used by kubelet and kubeadm bootstrapping
+  before both-way trust is established).
+
+### Desired API Semantics
+
+The discovery information should be served like any other resource in
+the Kubernetes API, i.e. as its own resource within a versioned API
+group.
+
+The API should be read-only, i.e. only GET, LIST and WATCH will be
+allowed verbs.
+
+The default RBAC rules will make the resources readable for every
+authenticated user.
+
+It would be possible to add a flag for unauthenticated readability,
+i.e. as a counterpart to the `kube-public` API group that is used
+during kubelet bootstrapping. But this is a non-goal for now.
+
+The discovery information includes its own API group.
+
+## New API Types
+
+The proposed API group name is `discovery.k8s.io`, starting with a
+`v1alpha1` version.
+
+The only new Kind is `Group`, containing all version information and
+all resources. The name of group objects is the API group name. The
+Kind is cluster-scoped.
+
+```golang
+// +genclient=true
+// +nonNamespaced=true
+
+// Group describes an API group with all its versions and resources.
+type Group struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+	// Status of the group.
+	Status GroupStatus `json:"spec,omitempty" protobuf:"bytes,2,opt,name=status"`
+}
+
+// GroupStatus describes the API group with all its versions and resources.
+type GroupStatus struct {
+	// versions are the versions supported in this group, order by preference. The first one is the preferred
+	// version for the group.
+	Versions []GroupVersion `json:"versions" protobuf:"bytes,1,rep,name=versions"`
+	// a map of client CIDR to server address that is serving this group.
+	// This is to help clients reach servers in the most network-efficient way possible.
+	// Clients can use the appropriate server address as per the CIDR that they match.
+	// In case of multiple matches, clients should use the longest matching CIDR.
+	// The server returns only those CIDRs that it thinks that the client can match.
+	// For example: the master will return an internal IP CIDR only, if the client reaches the server using an internal IP.
+	// Server looks at X-Forwarded-For header or X-Real-Ip header or request.RemoteAddr (in that order) to get the client IP.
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+  field. In a distributed environment, this requires some kind of
+  state. This state would naturally be provided by etcd. At this
+  points, it's not that far from storing the `Group` objects
+  persistently in etcd and inherit all the machinery. E.g. we can use
+  normal informers to watch existing API groups, versions and
+  resources.
+- We could extend the current endpoints to be watchable (the main goal
+  of this proposal). But this would mean to reinvent a lot of
+  infrastructure we already have.
+- We could re-use the `meta.k8s.io` group name instead of
```
smarterclayton: I don't want to add things to meta.k8s.io that aren't about raw API server mechanics.  Discovery is a layer on top of that.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+e.g. an API server update shouldn't take longer than a minute.
+
+We could introduce a server version for each API group. Then an older
+server could be less aggressive about reconciliation of new
+groups. But, this is not part of the API types yet.
+
+### Alternative Architectures
+
+- We could implement the discovery resource in the aggregator only and
+  use polling to the sub-API-servers. This would look mostly the same
```
smarterclayton: Given how complex this proposal is, I don't think you've adequately justified why simple polling is worse than big-complex-new-thing.

contributors/design-proposals/watchable-discovery.md
sttts: totally agree, see my comment at the end.

lavalamp:
I haven't read this but people can just watch the API registration api and the TPR api right now if they really need up-to-the-second data on this. Ideally it'd just be one thing to watch but we can improve this in the future.

sttts:
@lavalamp We will soon have two TPR kinds (for some time) and the APIService. In the controller manager, we already have an ugly polling logic, this will grow now with aggregation even more if we want to watch everything that can contribute to new API group.

And we know that kubectl's caching is often broken. Without caching it's slow as hell (OpenShift has yet another set of groups and versions that must be queried). So we have an immediate need for something better when TPR and aggregation becomes more popular.

This proposal is about "improving this in the future" and to have a clean versioned API and to use shared informers for the discovery info like for everything else. Why do you see this for the future and not now? We have to start at some point in time: our polling is broken for multiple major versions and watchable discovery is a topic at least since the idea of aggregation was born and the pain with TPRs with GC and namespace deletion is well known.

The implementation of this is straight forward and fits perfectly well with the aggregation design. I expect a first working implementation next week. So this is not rocket science.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+## Behavior
+
+Each API server in the cluster serves its own `discovery.k8s.io` API
+group. By default, this is backed by normal etcd storage, i.e. the
+semantics of the `ResourceVersion` is the one of the underlaying etcd
+storage backend.
+
+Each API server has its own `DiscoveryUpdateController` that's
```
lavalamp: controllers like this should not run out of apiserver.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// List of group.
+	Items []Group `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+```
+
+## Behavior
+
+Each API server in the cluster serves its own `discovery.k8s.io` API
```
lavalamp: I don't agree with each apiserver having a private copy of this, nothing else in the system is done this way?

I don't think apiserver should be doing anything here, actually. I tentatively agree with the discovery types and group, but there should just be a few controllers syncing them up with what is currently stored in TPR & and apiregistration groups.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+
+Each API server has its own `DiscoveryUpdateController` that's
+regularly reconciles the storage state of the discovery API group with
+its internal registered/announce API group registry. Moreover, when
+a new API groups is installed or removed in its HTTP handler, the
+corresponding API group object is updated synchronously.
+
+### Aggegration
+
+The aggregator watches the discovery API group of each of its proxied
```
lavalamp: the controllers populating the discovery information need to be pointing at the aggregator and not the main apiserver.

I'd also be OK with adding these discovery types/group to the aggregator type tree.

Other than that, I don't think the aggregator code should change.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+- only showing those GroupVersions that are accepted by the aggregator
+  (i.e. which have no naming conflict).
+
+The aggregator will update its own discovery API group resources
+triggered by changes in the sub-API-servers.
+
+### Embedded Aggregator in kube-apiserver
+
+In this mode, both the aggregator and the kube-apiserver share the
+storage. To avoid conflicts about updating the API group, the `Group`
```
lavalamp: Literally nothing should change in this scenario.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+objects of kube-apiserver are marked as "local" with an
+annotation. The aggregator will not touch them and the other way
+around for the kube-apiserver for the not-marked `Group` objects.
+
+### HA
+
+API servers in HA mode will share a storage. Hence, they will share
+the state of the discovery API group. Their controllers will fight for
```
lavalamp: Yeah this demonstrates why the plan of each of them doing its own thing is not good. (honestly when reading the previous section I thought you were going to have each apiserver store the data in a separate path to avoid this collision)

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+the correct state of the discovery information.
+
+During normal operation the desired discovery state is equal in an HA
+setup. So no real race will take place.
+
+During rolling updates of the control plane, an old and a new API
+server version might be active. Their reconciliation loops will race
```
lavalamp: Yeah, this would be very chaotic.

contributors/design-proposals/watchable-discovery.md
deads2k: > Given how complex this proposal is, I don't think you've adequately justified why simple polling is worse than big-complex-new-thing.

We'd be talking about polling from inside of GC and the namespace cleanup controller.  We tried doing just the namespace cleanup controller and it resulted in burning through the QPS and burst that I "fixed" by bumping them significantly.

If we don't get notified, you can't simply wait since that would result in significant window of dangling resources.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,231 @@
...
+flipping. But, this is not a regression of the cluster behaviour as it
+already happens now if a load-balancer distributes requests among the
+cluster API servers, each returning different discovery information.
+
+A long reconciliation period will reduce the effect of the flipping,
+e.g. an API server update shouldn't take longer than a minute.
+
+We could introduce a server version for each API group. Then an older
+server could be less aggressive about reconciliation of new
+groups. But, this is not part of the API types yet.
```
lavalamp: This entire section isn't needed if we go the ordinary controller syncing route like I suggest.

contributors/design-proposals/watchable-discovery.md
sttts: The already run now: APIServiceAutoRegistration and of course the core logic of the kube-apiextension-apiserver which is a controller as well.

contributors/design-proposals/watchable-discovery.md
sttts: The aggregator is the only one who decides which GVR is visible. It's the place to do the aggregation. It also does that now with the polling btw: it selectively forwards discovery calls to the right API server, hiding those which are not to be exposed. The only difference is that it also takes care of the resources, not only GV.

contributors/design-proposals/watchable-discovery.md
lavalamp: I don't know what either of those things are. Controllers should (must?) either be 100% safe to run in parallel (e.g. apiserver endpoint publisher) or use a hot/warm standby model like controller manager does.

contributors/design-proposals/watchable-discovery.md
sttts: We have the same with APIService. I am actually just following that example. In a HA setup they race for the right APIService object.

contributors/design-proposals/watchable-discovery.md
sttts: Only the aggregator could do that. Compare my discussion under "Alternative Architectures", first paragraph. That's basically your "one controller" design.

lavalamp:
can we put a hold on this proposal? Clearly I need to take my time and explain myself better, but I have other things that are more immediately pressing. Is this targeted at 1.7?

contributors/design-proposals/watchable-discovery.md
sttts: Compare first paragraph under "Alternative Architectures" at the bottom (https://github.com/kubernetes/community/pull/617/files#diff-64d72530983c20925139eba9b44a495dR208). The variant with watches there matches what you describe here:
- one controller
- one copy of discovery.k8s.io (under the aggregator etcd key prefix)
- and watches for TPRs.

contributors/design-proposals/watchable-discovery.md
sttts: For reference: https://github.com/kubernetes/kubernetes/blob/55042b0ba9d9452366c2091a73d0bc4d7c50bf09/staging/src/k8s.io/kube-aggregator/pkg/controllers/autoregister/autoregister_controller.go

@deads2k can probably comment on the HA behavior of this.

sttts:
@lavalamp for thirdparty resources to be promoted to beta we need a better discovery story, on the controller side and in kubectl. @caesarxuchao is waiting for that for the GC controller to support non-kube resources.

So I guess those efforts might be blocked by this. If they don't make it for 1.7, there is a version after that.

Of course, there are always ways to do a quick'n'dirty solutions like the one of the namespace controller. It was quick, but was biting us eventually, e.g. compare https://github.com/kubernetes/community/pull/617#discussion_r116271478 and the broken use-cases with ThirdPartyResources.

k8s-mirror-api-machinery-api-reviews:
I need the GC to work on these asap, it blocks api aggregation which needs
to go out in 1.7. I will probably just manually fix it though, I don't want
to rush a big system in to fix this.

contributors/design-proposals/watchable-discovery.md
deads2k: > I don't agree with each apiserver having a private copy of this, nothing else in the system is done this way?
> 
> I don't think apiserver should be doing anything here, actually. I tentatively agree with the discovery types and group, but there should just be a few controllers syncing them up with what is currently stored in TPR & and apiregistration groups.

Each API server needs to service discovery information and ought to be able to have a watchable API that describes it.  In addition, to avoid some of the coupling we currently have in our API installer path, driving the actual `/apis` endpoints from an API like this will reduce internal coupling.

In addition, by having the constituent API servers provide `discovery.k8s.io`, the aggregated API server would be able to watch them to drive a unified view easily an cleanly via composition.

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,251 @@
+# Watchable API Group Discovery
+
+The Kubernetes API server provides a number of HTTP endpoinds to
```
kargakis: typo

contributors/design-proposals/watchable-discovery.md
```diff
@@ -0,0 +1,251 @@
...
+- `/apis/<group>/<version>`: returns a list of resources (compare [metav1.APIResourceList](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L718))
+
+This discovery information is used by many API clients like kubectl
+and controllers like the namespace controller. The garbage
+collection controller is supposed to use discovery as well, [but
+doesn't yet](https://github.com/kubernetes/kubernetes/blob/6f5337cff7baffc2b27d85036cb1ecb6a4c1f73b/cmd/kube-controller-manager/app/core.go#L174).
+
+Kubectl will cache the discovery information in
+`~/.kube/cache` for ten minutes and queries the API server again after
+that time or on a cache miss on RESTMapper lookups. In the namespace controller
```
kargakis: May this not be a cache miss but a misspelled resource name by a user?

sttts: Can be a misspelled resource name, can be a new TPR name. Otherwise, the 10 min apply.

PR #627, opened 2017-05-17
API: avoid bool fields

thockin:
As requested.  My thoughts.

### DIFF
--- a/contributors/devel/api-conventions.md
+++ b/contributors/devel/api-conventions.md
@@ -433,21 +433,25 @@ selectors, annotations, data), as opposed to sets of subobjects.
 #### Primitive types
 
 * Avoid floating-point values as much as possible, and never use them in spec.
-Floating-point values cannot be reliably round-tripped (encoded and re-decoded)
-without changing, and have varying precision and representations across
-languages and architectures.
+  Floating-point values cannot be reliably round-tripped (encoded and
+  re-decoded) without changing, and have varying precision and representations
+  across languages and architectures.
 * All numbers (e.g., uint32, int64) are converted to float64 by Javascript and
-some other languages, so any field which is expected to exceed that either in
-magnitude or in precision (specifically integer values > 53 bits) should be
-serialized and accepted as strings.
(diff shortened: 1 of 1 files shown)


contributors/devel/api-conventions.md
```diff
@@ -433,21 +433,24 @@ selectors, annotations, data), as opposed to sets of subobjects.
...
 * Do not use enums. Use aliases for string instead (e.g., `NodeConditionType`).
 * Look at similar fields in the API (e.g., ports, durations) and follow the
-conventions of existing fields.
+  conventions of existing fields.
 * All public integer fields MUST use the Go `(u)int32` or Go `(u)int64` types,
-not `(u)int` (which is ambiguous depending on target platform). Internal types
-may use `(u)int`.
+  not `(u)int` (which is ambiguous depending on target platform). Internal
+  types may use `(u)int`.
+* Think twice about `bool` fields. Many ideas start as boolean but eventually
```
davidopp: Recommended alternative? string field with defined set of constants (initially "True" and "False")

smarterclayton: I actually think True and False are worse, because then you get `True|False|FileNotFound` and that way lies madness.

`*Policy` is usually our way out - come up with words that describe the policy, even if there are only two:

```
// TerminationMessagePolicy describes how termination messages are retrieved from a container.
type TerminationMessagePolicy string

const (
	// TerminationMessageReadFile is the default behavior and will set the container status message to
	// the contents of the container's terminationMessagePath when the container exits.
	TerminationMessageReadFile TerminationMessagePolicy = "File"
	// TerminationMessageFallbackToLogsOnError will read the most recent contents of the container logs
	// for the container status message when the container exits with an error and the
	// terminationMessagePath has no contents.
	TerminationMessageFallbackToLogsOnError TerminationMessagePolicy = "FallbackToLogsOnError"
)
```

thockin:
re-pushed with that example.

smarterclayton:
LGTM, will wait for others to look too.

davidopp:
LGTM

PR #628, opened 2017-05-17
Update the token discovery proposal

luxas:
@jbeda @mikedanese @justinsb @timothysc @roberthbailey @lukemarsden @liggitt 

A first draft of the v1.7 updates for hopefully bringing Bootstrap Tokens to beta.
Minor changes but important

### DIFF
--- a/contributors/design-proposals/bootstrap-discovery.md
+++ b/contributors/design-proposals/bootstrap-discovery.md
@@ -63,7 +63,7 @@ For the `kubeadm` flow, the command line might look like:
 kubeadm join --discovery-file=my-cluster.yaml
 ```
 
-After loading the ClusterInfo from a file, the client MAY look for updated information from the server by reading the `kube-public` `cluster-info` ConfigMap defined below.  However, when retrieving this ConfigMap the client MUST validate the certificate chain when talking to the API server.
+After loading the ClusterInfo from a file, the client MAY look for updated information from the server by reading the `kube-public/cluster-info` ConfigMap defined below.  However, when retrieving this ConfigMap the client MUST validate the certificate chain when talking to the API server.
 
 **Note:** TLS bootstrap (which establishes a way for a client to authenticate itself to the server) is a separate issue and has its own set of methods.  This command line may have a TLS bootstrap token (or config file) on the command line also.  For this reason, even thought the `--discovery-file` argument is in the form of a `kubeconfig`, it MUST NOT contain client credentials as defined above.
 
@@ -72,16 +72,16 @@ After loading the ClusterInfo from a file, the client MAY look for updated infor
 If the ClusterInfo information is hosted in a trusted place via HTTPS you can just request it that way.  This will use the root certificates that are installed on the system.  It may or may not be appropriate based on the user's constraints.  This method MUST use HTTPS.  Also, even though the payload for this URL is the `kubeconfig` format, it MUST NOT contain client credentials.
 
 ```
-kubeadm join --discovery-url="https://example/mycluster.yaml"
+kubeadm join --discovery-file="https://example/mycluster.yaml"
 ```
 
 This is really a shorthand for someone doing something like (assuming we support stdin with `-`):
 
 ```
-curl https://example.com/mycluster.json | kubeadm join --discovery-file=-
+curl https://example.com/mycluster.yaml | kubeadm join --discovery-file=-
 ```
 
-After loading the ClusterInfo from a URL, the client MAY look for updated information from the server by reading the `kube-public` `cluster-info` ConfigMap defined below.  However, when retrieving this ConfigMap the client MUST validate the certificate chain when talking to the API server.
+After loading the ClusterInfo from a URL, the client MAY look for updated information from the server by reading the `kube-public/cluster-info` ConfigMap defined below.  However, when retrieving this ConfigMap the client MUST validate the certificate chain when talking to the API server.
 
 **Note:** support for loading from stdin for `--discovery-file` may not be implemented immediately.
 
@@ -101,7 +101,7 @@ An interesting aspect here is that this information is often easily obtained bef
 The user experience for joining a cluster would be something like:
 
 ```
-kubeadm join --token=ae23dc.faddc87f5a5ab458 <address>
+kubeadm join --token=ae23dc.faddc87f5a5ab458 <address>:<port>
 ```
 
 **Note:** This is logically a different use of the token used for authentication for TLS bootstrap.  We harmonize these usages and allow the same token to play double duty.
@@ -112,14 +112,13 @@ kubeadm join --token=ae23dc.faddc87f5a5ab458 <address>
 
 * `kubeadm` connects to the API server address specified over TLS.  As we don't yet have a root certificate to trust, this is an insecure connection and the server certificate is not validated.  `kubeadm` provides no authentication credentials at all.
   * Implementation note: the API server doesn't have to expose a new and special insecure HTTP endpoint.
+  * (D)DoS concern: Before this flow is secure to use/enable publicly (when not bootstrapping), the API Server must support rate-limiting. There are a couple of ways rate-limiting can be implemented to work for this use-case, but defining the rate-limiting flow in detail here is out of scope. One simple idea is limiting unauthenticated requests to come from clients in RFC1918 ranges.
 * `kubeadm` requests a ConfigMap containing the kubeconfig file defined above.
   * This ConfigMap exists at a well known URL: `https://<server>/api/v1/namespaces/kube-public/configmaps/cluster-info`
   * This ConfigMap is really public.  Users don't need to authenticate to read this ConfigMap.  In fact, the client MUST NOT use a bearer token here as we don't trust this endpoint yet.
 * The API server returns the ConfigMap with the kubeconfig contents as normal
   * Extra data items on that ConfigMap contains JWS signatures.  `kubeadm` finds the correct signature based on the `token-id` part of the token.  (Described below).
-* `kubeadm` verifies the JWS and can now trust the server.  Further
-  communication is simpler as the CA certificate in the kubeconfig file can be
-  trusted.
+* `kubeadm` verifies the JWS and can now trust the server.  Further communication is simpler as the CA certificate in the kubeconfig file can be trusted.
 
 
 #### NEW: Bootstrap Token Structure
@@ -138,17 +137,17 @@ The `token-id` must be 6 characters and the `token-secret` must be 16 characters
 
 #### NEW: Bootstrap Token Secrets
 
-Bootstrap tokens are stored and managed via Kubernetes secrets in the `kube-system` namespace.  They have type `bootstrap.kubernetes.io/token`.
+Bootstrap tokens are stored and managed via Kubernetes secrets in the `kube-system` namespace.  They have the type `bootstrap.kubernetes.io/token`.
 
 The following keys are on the secret data:
 * **token-id**. As defined above.
 * **token-secret**. As defined above.
 * **expiration**. After this time the token should be automatically deleted.  This is encoded as an absolute UTC time using RFC3339.
 * **usage-bootstrap-signing**. Set to `true` to indicate this token should be used for signing bootstrap configs.  If this is missing from the token secret or set to any other value, the usage is not allowed.
-* **usage-bootstrap-authentication**. Set to true to indicate that this token should be used for authenticating to the API server.  If this is missing from the token secret or set to any other value, the usage is not allowed.  The bootstrap token authenticagtor will use this token to auth as a user that is `system:bootstrap:<token-id>` in the group `system:bootstrappers`.
+* **usage-bootstrap-authentication**. Set to true to indicate that this token should be used for authenticating to the API server.  If this is missing from the token secret or set to any other value, the usage is not allowed.  The bootstrap token authenticator will use this token to auth as a user that is `system:bootstrap:<token-id>` in the group `system:bootstrappers`.
 * **description**. An optional free form description field for denoting the purpose of the token.  If users have especially complex token management neads, they are encouraged to use labels and annotations instead of packing machined readable data in to this field.
+* **auth-groups**. A comma-separated list of which groups the token should be authenticated as. All groups must have the `system:bootstrappers:` prefix.
 
-**Future**: At some point in the future we may add the ability to specify a set of groups that this token part of during authentication.  This will allow users to segment off which tokens are allowed to bootstrap which nodes.  However, we will restrict these groups under `system:bootstrappers:*` to discourage usage outside of bootstrapping.
 
 These secrets MUST be named `bootstrap-token-<token-id>`.  If a token doesn't adhere to this naming scheme it MUST be ignored.  The secret MUST also be ignored if the `token-id` key in the secret doesn't match the name of the secret.
 
@@ -165,7 +164,8 @@ The JWS specification [describes how to encode](https://tools.ietf.org/html/rfc7
 
 #### NEW: `kube-public` namespace
 
-Kubernetes ConfigMaps are per-namespace and are generally only visible to principals that have read access on that namespace.  To create a config map that *everyone* can see, we introduce a new `kube-public` namespace.  This namespace, by convention, is readable by all users (including those not authenticated).
+Kubernetes ConfigMaps are per-namespace and are generally only visible to principals that have read access on that namespace.  To create a config map that *everyone* can see, we introduce a new `kube-public` namespace.  This namespace, by convention, is readable by all users (including those not authenticated). Note that is a convention
+(to expose everything in `kube-public`), not something that's done by default in Kubernetes. `kubeadm` does _solely_ expose the `cluster-info` ConfigMap, not anything else.
 
 In the initial implementation the `kube-public` namespace (and the `cluster-info` config map) will be created by `kubeadm`. That means that these won't exist for clusters that aren't bootstrapped with `kubeadm`.  As we have need for this configmap in other contexts (self describing HA clusters?) we'll make this be more generally available.
 
@@ -179,7 +179,7 @@ A new controller (`bootstrapsigner`) is introduced that will watch for both new/
 
 Another controller (`tokencleaner`) is introduced that deletes tokens that are past their expiration time.
 
-Logically these controllers could run as a component in the control plane.  But, for the sake of efficiency, they are bundeled as part of the Kubernetes controller-manager.
+Logically these controllers could run as a separate component in the control plane.  But, for the sake of efficiency, they are bundeled as part of the Kubernetes controller-manager.
 
 ## `kubeadm` UX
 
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -220,16 +219,27 @@ Only one of `--discovery-url`, `--discovery-file` or `--discovery-token` can be
 
 ## Implementation Details
 
-Our documentations (and output from `kubeadm`) should stress to users that when the token is configured for authenitication and used for TLS bootstrap (using `--insecure-experimental-approve-all-kubelet-csrs-for-group`) it is essentially a root password on the cluster and should be protected as such.  Users should set a TTL to limit this risk.  Or, after the cluster is up and running, users should delete the token using `kubeadm token delete`.
+Our documentations (and output from `kubeadm`) should stress to users that when the token is configured for authentication and used for TLS bootstrap is a pretty powerful credential due to that any person with access to it can claim to be a node.
+The highest risk regarding being able to claim a credential in the `system:nodes` group is that it can read all Secrets in the cluster, which basically compromises the cluster, until the [Node Authorizer](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/kubelet-authorizer.md) is fully implemented.
```
liggitt: even once the node authorizer is in place, any credential that can be used to obtain credentials for any node (which the bootstrapper can) can still access secrets used by any pod

luxas:
> even once the node authorizer is in place, any credential that can be used to obtain credentials for any node (which the bootstrapper can) can still access secrets used by any pod

@mikedanese @liggitt @ericchiang @jbeda @jcbsmpsn What do you think of the proposal to make the group approver not auto-approve CSRs with a node name (CN=`system:node:<name>`) that has already have been approved under the condition the user is a bootstrap token?

Note that certificate rotation wouldn't be touched by this change -- there the kubelet already uses its own credential(s).

That way we could remove the possibility for a not trusted person being able to get client credentials for a node in the cluster that already exists and has been bootstrapped once before. By doing that, an untrusted person with a hijacked token wouldn't be able to get client creds for the master and that way get to read the super-valuable Secrets attached to it.

What do you think?

Came to think about this now... a validation of the idea would be cool ;)

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -138,17 +137,17 @@ The `token-id` must be 6 characters and the `token-secret` must be 16 characters
...
-* **usage-bootstrap-authentication**. Set to true to indicate that this token should be used for authenticating to the API server.  If this is missing from the token secret or set to any other value, the usage is not allowed.  The bootstrap token authenticagtor will use this token to auth as a user that is `system:bootstrap:<token-id>` in the group `system:bootstrappers`.
+* **usage-bootstrap-authentication**. Set to true to indicate that this token should be used for authenticating to the API server.  If this is missing from the token secret or set to any other value, the usage is not allowed.  The bootstrap token authenticator will use this token to auth as a user that is `system:bootstrap:<token-id>` in the group `system:bootstrappers`.
 * **description**. An optional free form description field for denoting the purpose of the token.  If users have especially complex token management neads, they are encouraged to use labels and annotations instead of packing machined readable data in to this field.
+* **auth-groups**. A comma-separated list of which groups the token should be authenticated as. All groups must have the `system:bootstrappers:` prefix.
```
luxas: I'm still considering whether this is worth it or not. The rationale would be to to not give any Bootstrap Token CSR Post access (the `system:bootstrappers` group) by default, but instead have a specific group (`system:bootstrappers:<foo>`) that would be authorized to make CSR requests.

But I still haven't decided whether that makes sense

ericchiang: It seems like it makes sense if the usage of these token's is going to expand past simple CSR use cases. I'd be in favor of putting on this for now. It doesn't seem like it'd be necessary for beta.

liggitt: > I'd be in favor of putting on this for now. It doesn't seem like it'd be necessary for beta.

putting *off*?

ericchiang: blah, "punting on this for now"

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -193,12 +193,11 @@ We extend kubeadm with a set of flags and helper commands for managing and using
 ### `kubeadm join` flags
 
 * `--token` This sets the token for both discovery and bootstrap auth.
-* `--discovery-url` If set this will grab the cluster-info data (a kubeconfig) from a URL. Due to the sensitive nature of this data, we will only support https URLs.  This also supports `username:password@host` syntax for doing HTTP auth.
-* `--discovery-file` If set, this will load the cluster-info from a file.
+* `--discovery-file` If set, this will load the cluster-info from a file on disk or from a HTTPS URL (the HTTPS requirement due to the sensitive nature of the data)
 * `--discovery-token` If set, (or set via `--token`) then we will be using the token scheme described above.
 * `--tls-bootstrap-token` (not officially part of this spec) This sets the token used to temporarily authenticate to the API server in order to submit a CSR for signing.  If `--insecure-experimental-approve-all-kubelet-csrs-for-group` is set to `system:bootstrappers` then these CSRs will be approved automatically for a hands off joining flow.
```
ericchiang: This flag is going away. Worth updating this proposal too?

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -72,16 +72,16 @@ After loading the ClusterInfo from a file, the client MAY look for updated infor
 If the ClusterInfo information is hosted in a trusted place via HTTPS you can just request it that way.  This will use the root certificates that are installed on the system.  It may or may not be appropriate based on the user's constraints.  This method MUST use HTTPS.  Also, even though the payload for this URL is the `kubeconfig` format, it MUST NOT contain client credentials.
 
 ```
-kubeadm join --discovery-url="https://example/mycluster.yaml"
+kubeadm join --discovery-file="https://example/mycluster.yaml"
```
roberthbailey: why is this flag called -file when it's clearly able to load URLs?

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -101,7 +101,7 @@ An interesting aspect here is that this information is often easily obtained bef
 The user experience for joining a cluster would be something like:
 
 ```
-kubeadm join --token=ae23dc.faddc87f5a5ab458 <address>
+kubeadm join --token=ae23dc.faddc87f5a5ab458 <address>:<port>
```
roberthbailey: is port optional? is address a fully formed URL (which will imply a port) or a raw IP address / DNS name?

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -112,14 +112,13 @@ kubeadm join --token=ae23dc.faddc87f5a5ab458 <address>
 
 * `kubeadm` connects to the API server address specified over TLS.  As we don't yet have a root certificate to trust, this is an insecure connection and the server certificate is not validated.  `kubeadm` provides no authentication credentials at all.
   * Implementation note: the API server doesn't have to expose a new and special insecure HTTP endpoint.
+  * (D)DoS concern: Before this flow is secure to use/enable publicly (when not bootstrapping), the API Server must support rate-limiting. There are a couple of ways rate-limiting can be implemented to work for this use-case, but defining the rate-limiting flow in detail here is out of scope.
```
roberthbailey: @jbeda also mentioned during the sig meeting today that an intermediate case, which should largely protect against DDoS is to only accept connections from RFC 1918 IP ranges.

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -165,7 +164,8 @@ The JWS specification [describes how to encode](https://tools.ietf.org/html/rfc7
 
 #### NEW: `kube-public` namespace
 
-Kubernetes ConfigMaps are per-namespace and are generally only visible to principals that have read access on that namespace.  To create a config map that *everyone* can see, we introduce a new `kube-public` namespace.  This namespace, by convention, is readable by all users (including those not authenticated).
+Kubernetes ConfigMaps are per-namespace and are generally only visible to principals that have read access on that namespace.  To create a config map that *everyone* can see, we introduce a new `kube-public` namespace.  This namespace, by convention, is readable by all users (including those not authenticated). Note that is a convention
```
roberthbailey: Do we expect the entire namespace to be readable w/o authentication as the convention, or only this single config map (which is what kubeadm does)?

contributors/design-proposals/bootstrap-discovery.md
luxas: No actually it isn't optional, since it's hard for the community to agree on a default port.
kubeadm uses 6443, GKE 443 IIRC, minikube maybe 8443, etc.

contributors/design-proposals/bootstrap-discovery.md
luxas: Since it's utilizing the option to bring a file. Whether you read it from the internet using HTTPS or from disk doesn't _really_ matter. There were quite a long debate on this on the last edit to the proposal and on Slack I think, and we ended up with this option but the proposal wasn't reflected it seems.

contributors/design-proposals/bootstrap-discovery.md
luxas: cc @kubernetes/sig-auth-feature-requests

contributors/design-proposals/bootstrap-discovery.md
luxas: @ericchiang With some proposals in flight etc., this might actually be a requirement to graduate to beta.
This would make it possible to use Bootstrap Tokens for more than the kubelet TLS Bootstrap, now the kubeadm implementation ties them unconditionally together with RBAC rules (binding the `system:bootstrappers` Group to CSR create access (ok) and auto-approving all kubelet CSRs (not great default))

Specifically thinking about enabling API Server TLS Bootstrapping, there we want to be able to scope down tokens to have different privs.

contributors/design-proposals/bootstrap-discovery.md
luxas: I think it's preferable to keep everything put there really public as the convention, but I don't think we should default to having a "read-all" Role bound to anonymous users. Instead, I expect implementors to create fine-grained rules for their use-case to open things up.

Should I clarify that in the text here?

contributors/design-proposals/bootstrap-discovery.md
luxas: I will, thanks

contributors/design-proposals/bootstrap-discovery.md
luxas: Yeah, you need to know the Node name, but sure. Can you think of an easy path to lock that down?

luxas:
> @mikedanese @liggitt @ericchiang @jbeda @jcbsmpsn What do you think of the proposal to make the group approver not auto-approve CSRs with a node name (CN=system:node:<name>) that has already have been approved under the condition the user is a bootstrap token?

> Note that certificate rotation wouldn't be touched by this change -- there the kubelet already uses its own credential(s).

> That way we could remove the possibility for a not trusted person being able to get client credentials for a node in the cluster that already exists and has been bootstrapped once before. By doing that, an untrusted person with a hijacked token wouldn't be able to get client creds for the master and that way get to read the super-valuable Secrets attached to it.

ping @liggitt @ericchiang @jbeda @roberthbailey on that idea.

I'm not sure how good it would be to make the certificate controller able of listing nodes, but are we willing to trade that so that only CSRs for new nodes are auto-approved?

I'll update this soon...

liggitt:
That seems like it is trying to allow you to survive compromise of a powerful shared secret bootstrap credential... if that happened, you wouldn't easily be able to trust any node identities obtained with that credential. I'd rather see the shared secret expired/removed/revoked after bootstrapping is done than try to add on a "first-request-per-node-wins" strategy.

luxas:
> That seems like it is trying to allow you to survive compromise of a powerful shared secret bootstrap credential... if that happened, you wouldn't easily be able to trust any node identities obtained with that credential.

So I was thinking about the worst-case scenario where you don't even know it has leaked...
But I now realized that it wouldn't actually help much as the unauthorized person may be able to directly create a "master node" by using the right labels, taints, etc., and get Pod bindings to the relevant Secrets via the control plane DaemonSets.

> I'd rather see the shared secret expired/removed/revoked after bootstrapping is done than try to add on a "first-request-per-node-wins" strategy.

There are auto-expiration mechanisms already in place, and you can easily revoke the token at any time using `kubeadm token delete <token-id>` for example.

contributors/design-proposals/bootstrap-discovery.md
ericchiang: Kubectl does the same thing too.

```
kubectl create --filename=https://foo.com/manifest.yaml
```

contributors/design-proposals/bootstrap-discovery.md
ericchiang: Would limiting unauthenticated users to only use the API server's cache be a more reasonable step forward? Or letting users register non-resource URLs with static information? A full blown rate limiting solution built into the API server seems like a can of worms.

contributors/design-proposals/bootstrap-discovery.md
ericchiang: Yeah, adding this feature works for me.

contributors/design-proposals/bootstrap-discovery.md
ericchiang: As a note, I think other component already use this namespace (simple grep shows it refer to by kube-aggregator, for example).

ericchiang:
> I'm not sure how good it would be to make the certificate controller able of listing nodes, but are we willing to trade that so that only CSRs for new nodes are auto-approved?

If I remember @gtank's original motivation for the certificates API, the bootstrapping token was only supposed to be use as DoS prevention, not actual credentials that identified the node. The primary issue we were trying to solve was the auto-scaling group problem, where workers come up with identical user-data. At the point where you can pass each node a unique secret, why not just give them credentials?

Isn't auto-approval is kind of avoiding the issue of node attestation, for example @mikedanese's proposal for ACME style challenges[1]? I always saw that as the end goal instead of creating richer management features for bootstrapping tokens.

[1] https://docs.google.com/document/d/1l85qgbH-CoyCst7Ksuc12VyOr7rvDmhPJ5kmB1fDnS4/edit?ts=590a3fa5#heading=h.lhk2tnjcns8x

contributors/design-proposals/bootstrap-discovery.md
jbeda: What is this token being replaced with?

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -220,16 +219,27 @@ Only one of `--discovery-url`, `--discovery-file` or `--discovery-token` can be
...
+Our documentations (and output from `kubeadm`) should stress to users that when the token is configured for authentication and used for TLS bootstrap is a pretty powerful credential due to that any person with access to it can claim to be a node.
+The highest risk regarding being able to claim a credential in the `system:nodes` group is that it can read all Secrets in the cluster, which basically compromises the cluster, until the [Node Authorizer](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/kubelet-authorizer.md) is fully implemented.
+
+Users should set a TTL on the token to limit the above mentioned risk. Or, after the cluster is up and running, users should delete the token using `kubeadm token delete`.
```
jbeda: Do we want to consider setting a default TTL for kubeadm tokens?  It would be a breaking change but would be much more secure.  We could default it to something like 6 hours?

contributors/design-proposals/bootstrap-discovery.md
```diff
@@ -220,16 +219,27 @@ Only one of `--discovery-url`, `--discovery-file` or `--discovery-token` can be
...
+
+This proposal assumes RBAC to lock things down in a couple of ways.  First, it will open up `cluster-info` ConfigMap in `kube-public` so that it is readable by unauthenticated users (user `system:anonymous`).  Next, it will make it so that the identities in the `system:bootstrappers` group can only be used with the certs API to submit CSRs.  After a TLS certificate is created, the groupapprover controller will approve the CSR and the CSR identity can be used instead of the bootstrap token.
 
-This proposal assumes RBAC to lock things down in a couple of ways.  First, it will open up `cluster-info` ConfigMap in `kube-public` so that it is readable by unauthenticated users.  Next, it will make it so that the identities in the `system:bootstrappers` group can only be used with the certs API to submit CSRs.  After a TLS certificate is created, that identity should be used instead of the bootstrap token.
+The end goal is to make it possible to delegate the client TLS Bootstrapping part to the kubelet, so `kubeadm join`'s function will solely be to verify the validity of the token and fetch the CA bundle.
```
jbeda: But are we passing the bootstrap token to the kubelet so the kubelet can then do the cert dance?  Or are we getting the first cert and then let the kubelet drive renews?

We should think through the case where people aren't using the built in CA and make sure we can do that with configs/phases.

jbeda:
As for "why not give nodes credentials?" -- there is a usability issue around that.  If you can copy arbitrary data around then you are good to go.  If you are copy/pasting command lines around, we need it to be short and somewhat secure.  Ideally we could use the token along with other mechanisms.

(Like, we could make the token good for `n` number of CSR signatures.  Or we could include the cloud identity document, etc.)

contributors/design-proposals/bootstrap-discovery.md
luxas: @jbeda Can you open a kubernetes/kubernetes issue with more details about that and assign someone?

contributors/design-proposals/bootstrap-discovery.md
luxas: @ericchiang Yes, it's used by aggregated API servers as well

liggitt: > As a note, I think other component already use this namespace (simple grep shows it refer to by kube-aggregator, for example).
> 
> Yes, it's used by aggregated API servers as well

I'm not aware of any expected uses of this namespace by the aggregator

luxas: Gah, was indeed thinking about the apiserver-extension configmap in the kube-system namespace. Recalled incorrectly

contributors/design-proposals/bootstrap-discovery.md
luxas: We've now done this :+1:

contributors/design-proposals/bootstrap-discovery.md
luxas: In what case wouldn't this work with the own-CA thing? The kubelet will request its identity from the CSR API in exactly the same manner as `kubeadm join` would.

> But are we passing the bootstrap token to the kubelet so the kubelet can then do the cert dance? Or are we getting the first cert and then let the kubelet drive renews?

I don't have a strong opinion here. WDYT?

contributors/design-proposals/bootstrap-discovery.md
luxas: ping @jbeda

contributors/design-proposals/bootstrap-discovery.md
luxas: A ClusterRoleBinding between a ClusterRole (like these: https://github.com/kubernetes/kubernetes/pull/49284) and a group that the token got authenticated into.

luxas:
@jbeda @ericchiang @liggitt @roberthbailey Updated.
More details on how to get this to beta can also be found here: https://github.com/kubernetes/features/issues/130#issuecomment-316739665

cc @mattmoyer

contributors/design-proposals/bootstrap-discovery.md
roberthbailey: it's weird but i guess it's consistent.

luxas:
Thanks Joe! I'm just gonna merge this now so @mattmoyer can update it for the new cert pinning thing he has been working on.

PR #629, opened 2017-05-17
Add proposal for kubeadm-local-cluster tool

ivan4th:
Moving the proposal from kubernetes/kubeadm#236 per sig-cluster-lifecycle suggestion.

A version of the local cluster tool described here exists in the form of [kubeadm-dind-cluster](https://github.com/Mirantis/kubeadm-dind-cluster) project (Go version is not here yet). The differences are described closer to the end of the document.

It's yet to be decided where the tool should live (keep the old location in Mirantis github repo, move to kubernetes-incubator or some other option).

I'll be showing a demo of existing DIND solution (`kubeadm-dind-cluster`) today on ContribX meeting (17 May 2017).

### DIFF
--- a/contributors/design-proposals/kubeadm-local-cluster.md
+++ b/contributors/design-proposals/kubeadm-local-cluster.md
@@ -1,0 +1,424 @@
+# Proposal: local cluster tool based on kubeadm and Docker-in-Docker
+
+## Abstract
+
+The `kubeadm-local-cluster` tool creates local multinode Kubernetes
+clusters using `kubeadm` tool and the Docker-in-Docker technique. Most
+of its functionality is already implemented in Mirantis'
+[kubeadm-dind-cluster](https://github.com/Mirantis/kubeadm-dind-cluster)
+project. The plan is to rewrite the shell scripts that comprise
+`kubeadm-dind-cluster` in Go language.
+
+The `kubeadm-local-cluster` tool supports building Kubernetes from
+source, but also can utilize prebuilt binaries. The tool can work
+against both remote and local docker engines. It's well suited for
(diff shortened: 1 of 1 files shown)


kargakis:
@kubernetes/sig-testing-feature-requests assuming you guys are responsible for hack/local-up-cluster - this should be of interest to you

timothysc:
I'm all for replacing local-up-cluster . 
@kubernetes/sig-cluster-lifecycle-feature-requests @ncdc

feiskyer:
local-up-cluster.sh supports all runtimes (including CRI runtimes) today, how kubeadm-local-cluster could support other runtimes besides docker, since dind is only for docker?

kargakis:
@feiskyer I don't think local-up-cluster is going away, this should just replace the docker parts of it.

ivan4th:
@feiskyer re CRI implementations, actually we use DIND just fine for [Virtlet](https://github.com/Mirantis/virtlet) work, also Virtlet's demo script is based on DIND, too. We can do this thanks to the CRI proxy mechanism which makes it possible to run CRI implementations as DaemonSet and deploy them via single `kubectl create -f ...` command. Although depending on the purpose of particular CRI implementation this may not be always the best approach, we can extract this component from Virtlet easily if other CRI implementation writers are interested in it.

@kargakis yes, I think the DIND mechanism cannot completely replace local-up-cluster. Yet I think that after Go rewrite it'll be possible to make multiple backends for this local cluster tool, including one that works in a way similar to local-up-cluster. This is currently out of scope of this proposal though.

danehans:
https://github.com/kubernetes/kubernetes/issues/47666

PR #637, opened 2017-05-18
Correct information about sig-storage leads in sigs.yaml

saad-ali:
Fixing what looks like a copy/paste error (`sig-service-catalog` to `sig-storage`).

Did not run `make all` since the generated files are not yet live.

CC @jamiehannaford


jamiehannaford:
👍 LGTM

PR #644, opened 2017-05-19
Correct wrong path for generic_scheduler.go file

LinEricYang:
Added missing "core" directory name into the path of
generic_scheduler.go file.


LinEricYang:
Already signed CLA.

PR #649, opened 2017-05-22
Propose a feature to troubleshoot running pods

verb:
This feature allows troubleshooting of running pods by running a new "Debug Container" in the pod namespaces.

This proposal was originally opened and reviewed in kubernetes/kubernetes#35584.

This proposal needs LGTM by the following SIGs:
- [ ] SIG Node
- [ ] SIG CLI
- [ ] SIG Auth
- [x] API Reviewer

Work in Progress:
- [x] Prototype `kubectl attach` for debug containers
- [x] Talk to sig-api-machinery about `/debug` subresource semantics


derekwaynecarr:
I think this proposal needs wider review since it comes with API changes.

derekwaynecarr:
Related to feature https://github.com/kubernetes/features/issues/277

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+allocation may cause it to be evicted.
+
+### Debug Container Status
+
+The status of a *debug container* is reported in a new field in `v1.PodStatus`:
+
+```
+type PodStatus struct {
+        ...
+        DebugContainerStatuses []ContainerStatus
```
derekwaynecarr: since this feature is alpha, doesn't the field name need to reflect that?

derekwaynecarr:
I think a section that covers the security concerns associated with this feature would be appreciated for future readers.

contributors/design-proposals/troubleshoot-running-pods.md
verb: That would make sense, I just didn't know that was how to do it. I know there's been a lot of discussion in #30819, but I haven't followed it. It doesn't look like there's consensus, but I can try to cherry-pick something that might work.

gyliu513: I think that we do not need to update the filed name but how about using `annotations` like `init container`? Such as adding an annotation as `pod.alpha.kubernetes.io/debug-containers`? You can get more detail from https://github.com/kubernetes/community/blob/master/contributors/design-proposals/container-init.md

liggitt: We should definitely not store something as important as debug container info in an annotation. Those are also compatibility nightmares when they transition to supported fields. 

@kubernetes/sig-api-machinery-misc for guidance on alpha field names and gating

gyliu513: @liggitt can you please show more detail for this? I found that the `init container` was using this logic to be from alpha->beta->graduate, why cannot the `debug container`?

liggitt: It did, and it has been a very painful transition, with very poor user experience (c.f. https://github.com/kubernetes/kubernetes/issues/45627#issuecomment-301863792)

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+
+The runtime manager stores metadata about containers in the runtime via labels
+(e.g. docker labels). These labels are used to populate the fields of
+`kubecontainer.ContainerStatus`. Since the runtime manager needs to handle Debug
+Containers differently in a few situations, we must add a new piece of metadata
+to distinguish Debug Containers from regular containers.
+
+`startContainer()` will be updated to write a new label
+`io.kubernetes.container.type` to the runtime. Existing containers will be
+started with a type of `REGULAR` or `INIT`. When added in a subsequent step,
```
liggitt: does this mean a 1.7 kubelet managing containers without this label (that were started by a 1.6 kubelet) will be confused about whether they are debug containers?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+
+1.  `KillPod()` already operates on all running containers returned by the
+    runtime.
+1.  Containers created prior to this feature being enabled will have a
+    `containerType` of `""`. Since this does not match `"DEBUG"` the special
+    handling of Debug Containers is backwards compatible.
+
+### Additional Constraints
+
+1.  Non-interactive workloads are explicitly supported. There are no plans to
```
liggitt: If the container run and attach steps are distinct, how is stdout/stderr coordinated so that the attach request obtains the first byte written to each? Is the first attach special? How do subsequent or additional attaches behave w.r.t. previous output from the debug container's entrypoint?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+    supported detached workloads, but doing so would be trivial with an
+    `attach=false` flag.
+1.  There are no guaranteed resources for ad-hoc troubleshooting. If
```
liggitt: this seems like it would make debugging a pod that was memory constrained pretty difficult.

contributors/design-proposals/troubleshoot-running-pods.md
gyliu513: Got it, thanks @liggitt

liggitt:
how do debug containers interact with graceful termination of pods?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+
+*   `kubectl describe pod` will not report the running command, only container
+    status
+*   Exited Debug Containers will be garbage collected as regular containers and
+    may disappear from the list of Debug Container Statuses.
+*   There's no specific integration with admission controller and pod security
+    policy.
+*   Explicit reattaching isn't implemented. Instead a `kubectl debug` invocation
+    will implicitly reattach if there is an existing, running container with the
+    same name. In this case container configuration will be ignored.
```
verb: I agree 100%

contributors/design-proposals/troubleshoot-running-pods.md
verb: No, it's backwards compatible (and there's a test). The `REGULAR` and `INIT` labels aren't used by anything initially and the kubelet behavior only differs when `Type == "DEBUG"`. (Type will be an empty string for a container that existed prior to the feature being enabled.)

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+Debug Containers will start with with the type `DEBUG`.
+
+##### Step 2: Creation and Handling of Debug Containers
+
+This step adds methods for creating debug containers, but doesn't yet modify the
+kubelet API. Since the runtime manager discards runtime (e.g. docker) labels
+after populating `kubecontainer.ContainerStatus`, the label value will be stored
+in a the new field `ContainerStatus.Type` so it can be used by `SyncPod()`.
+
+The kubelet gains a `RunDebugContainer()` method which accepts a `v1.Container`
```
verb: Yes, the pod spec is needed to generate the podsandbox configuration. It wasn't my intention to imply that a v1.Container was all that is required. Here's a staged copy of the change:

https://github.com/verb/kubernetes/pull/3/files#diff-44f60dd6d99cb695e5f333647ebd0703R753

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+    associated `PodDebugOptions`
+1.  `v1.PodStatus` gains a `DebugContainerStatuses` field
+
+Validation should disallow the following fields which are incompatible with
+Debug Containers:
+
+*   Resources
+*   LivenessProbe
+*   ReadinessProbe
+*   Lifecycle
```
verb: There's nothing inherently incompatible about volume mounts in a Debug Container, though there might be edge cases.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+Reattach Debug Container | Does Not Exist          | *ERROR*
+"                        | Running                 | Attach
+"                        | Exited                  | *ERROR*
+
+When supported by a runtime, multiple clients can attach to a single debug
+container and share the terminal. This is supported by Docker.
+
+### Killing Debug Containers
+
+Debug containers will not be killed automatically until the pod (specifically,
```
verb: That didn't occur to me as something we would want, though it does seem like a reasonable compromise. I'll look into it.

verb:
@liggitt Debug containers receive the same signals as other containers in the pod for lifecycle events. They differ only in that they aren't deleted when syncing pod spec while the pod is alive.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+to its console. Users give Debug Containers a name (e.g. "debug" or "shell")
+which is used in subsequent interactions and is reported in `kubectl describe`.
+
+### Creating Debug Containers
+
+The process for creating a Debug Container is:
+
+1.  `kubectl` constructs a `v1.Container` based on command line flags and
+    `POST`s it to `/api/v1/namespaces/$NS/pods/$POD/debug`.
+1.  The API server performs admission control and proxies the connection to the
```
liggitt: > I would hope the plugins will deny resources they don't recognize.

they do not, they select the resources/subresources they guard.

contributors/design-proposals/troubleshoot-running-pods.md
liggitt: if @ncdc has learned anything, it is that if an output buffer race can happen, it will, and that if it might be a problem, it will.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+
+Functionality in the kubelet will be hidden behind an alpha feature flag and
+disabled by default. The following are explicitly out of scope for the 1.7 alpha
+release:
+
+*   `kubectl describe pod` will not report the running command, only container
+    status
+*   Exited Debug Containers will be garbage collected as regular containers and
+    may disappear from the list of Debug Container Statuses.
+*   There's no specific integration with admission controller and pod security
```
liggitt: API resources are registered in https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/rest/storage_core.go#L187. Omitting an item from that map means it is not surfaced via the API.

contributors/design-proposals/troubleshoot-running-pods.md
liggitt: so if this moves from alpha to beta to stable, how do you maintain skew compatibility with older kubectl clients' debug implementations without propagating the implicit throw-away-info-and-reattach behavior into the stable version of the feature?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+
+1.  `SyncPod()` must not kill any Debug Container even though it is not part of
+    the pod spec.
+1.  As an exception to the above, `SyncPod()` will kill Debug Containers when
+    the pod sandbox changes since a Debug Container in a no-longer-used sandbox
+    is not useful. Debug Containers are not automatically restarted in the new
+    sandbox.
+1.  `convertStatusToAPIStatus()` must sort Debug Containers status into
+    `DebugContainerStatuses` similar to as it does for `InitContainerStatuses`
+1.  The kubelet must preserve information on debug containers for reporting.
```
liggitt: the entire v1.Container spec is placed into a label?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+
+For example, the following command would attach to a newly created container in
+a pod:
+
+```
+kubectl debug -it -c debug-shell --image=debian target-pod -- bash
+```
+
+It would be reasonable for Kubernetes to provide a default container name and
+image1, making the minimal possible debug command:
```
ncdc: image

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+
+```
+kubectl debug -it target-pod
```
ncdc: Do you think you'd ever want to run `kubectl debug` and _not_ attach stdin and use a tty?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+```
+
+This creates an interactive shell in a pod which can examine and signal all
+processes in the pod. It has access to the same network and IPC as processes in
+the pod. It can access the filesystem of other processes by `/proc/$PID/root`,
+and enter aribitrary namespaces of another container via `nsenter` when
```
ncdc: Kubernetes master with a new enough docker version uses a shared pid namespace by default now.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+resource pressure. This is acceptable for the alpha but must be resolved to
+graduate from alpha status.
+
+### Restarting and Reattaching Debug Containers
+
+Debug Containers will never be restarted automatically. It is possible to
+replace a Debug Container that has exited by re-using a Debug Container name. It
+is an error to attempt to replace a Debug Container that is still running.
+
+One way in which `kubectl debug` differs from `kubectl exec` is the ability to
```
ncdc: `kubectl attach` supports attaching to both init and normal containers. Would you want to expand it to support debug containers too? That would require the least amount of coding.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+"                        | Running                 | Attach
+"                        | Exited                  | *ERROR*
+
+When supported by a runtime, multiple clients can attach to a single debug
+container and share the terminal. This is supported by Docker.
+
+### Killing Debug Containers
+
+Debug containers will not be killed automatically until the pod (specifically,
+the pod sandbox) is destroyed. Unlike `kubectl exec`, Debug Containers will not
```
ncdc: There is roundabout support in newer docker versions for killing exec sessions. It now records the pid of the process that was exec'd, and we could use that information to do a kill.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+receive an EOF if their connection is interrupted. Instead, Debug Containers
+must be reattached to exit a running process. This could be tricky if the
+process does not allocate a TTY, in this case a second Debug Container could be
```
ncdc: Same comment as above re why would you not allocate a tty?

contributors/design-proposals/troubleshoot-running-pods.md
ncdc: @liggitt the typical process for starting a docker container is (outside of kubernetes):

1. create container
2. attach to container
3. start container

If you must start the container prior to attaching (which tends to be the case for things likes `kubectl run`), then your only option to make sure you see all prior output is to specify `logs=true` when attaching. This has downsides: last time I checked, you can't limit the output to e.g. the last 100 lines, and if you have a TTY, I'm not sure what happens in that case. Also note, this isn't currently available in the version of the docker api vendored in to kubernetes.

contributors/design-proposals/troubleshoot-running-pods.md
ncdc: I think if we just expand `kubectl attach` to support debug containers, that would solve the problem.

contributors/design-proposals/troubleshoot-running-pods.md
derekwaynecarr: @ncdc it requires docker 1.13+ i think, correct?

ncdc: @derekwaynecarr yes, plus we'd need to bump the docker APIs in kube (no small feat, since kube currently uses the now-deprecated engine-api repo).

contributors/design-proposals/troubleshoot-running-pods.md
verb: either way this has to be resolved prior to moving out of alpha

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+```
+
+It would be reasonable for Kubernetes to provide a default container name and
+image1, making the minimal possible debug command:
+
+```
+kubectl debug -it target-pod
+```
+
+This creates an interactive shell in a pod which can examine and signal all
```
ericchiang: What does "in a pod" mean? Each container has a different rootfs and aren't individual containers in a pods in unique PID namespaces?

contributors/design-proposals/troubleshoot-running-pods.md
ncdc: The only true way to ensure you don't miss any output is create-container, attach-container, start-container, in that order. That's how `docker run` works. For something like `kubectl run` and probably `kubectl debug` too, we can't really do that, because of the way the kubelet sync loop works (kubectl waits until it sees the pod is Running before attaching). Well, I guess we _could_ potentially do that, but it would require pausing a sync loop iteration until the remote client (kubectl) attaches, which isn't ideal.

contributors/design-proposals/troubleshoot-running-pods.md
ncdc: I would view `kubectl debug` as an interactive utility for debugging a running pod, whereas I see `kubectl exec` as a tool in my toolbox that might or might not require user interaction. Although the more I think about it, typically when I do use `kubectl exec` it's to get a shell, in which case I almost always do `-it`. There's a recent issue proposing that exec defaults these to true: #46300.

contributors/design-proposals/troubleshoot-running-pods.md
ncdc: I just wrote in another comment about this, but I think given the way the kubelet sync loop works, it will be very difficult to achieve create/attach/start without blocking one of the kubelet workers. We will need to discuss with sig-node if we want to pursue this.

contributors/design-proposals/troubleshoot-running-pods.md
verb: @ncdc @liggitt since `kubectl debug` is mainly a tool for getting a shell with a TTY, how much do we care about output stream coordination when `kubectl` will already say "press enter to get a prompt"? I think it's reasonable to defer to a future general solution rather than trying to solve this for debug.

ncdc: No objections to deferring

contributors/design-proposals/troubleshoot-running-pods.md
verb: Updated the document to specify that -i & -t will be enabled by default.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+From the perspective of the client, there's a new subresource of pod, /`debug`,
+that creates a Debug Container based on a provided `v1.Container` and attaches
+to its console. Users give Debug Containers a name (e.g. "debug" or "shell")
+which is used in subsequent interactions and is reported in `kubectl describe`.
+
+### Creating Debug Containers
+
+The process for creating a Debug Container is:
+
+1.  `kubectl` constructs a `v1.Container` based on command line flags and
```
verb: Added description of a new object `PodDebugContainer`

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+    is used because `/debug` is already used by the kubelet. `/podDebug` was
+    chosen to parallel existing endpoints like `/containerLogs`.
+1.  The kubelet instructs the Generic Runtime Manager (this feature is only
+    implemented for the CRI) to create a Debug Container.
+1.  The runtime manager uses the existing `startContainer()` method to create a
+    container in an existing pod. `startContainer()` has one modification for
+    Debug Containers: it creates a new runtime label (e.g. a docker label) that
+    identifies this container as a Debug Container.
+1.  The kubelet performs an attach operation and upgrades the client connection
+    to streaming.
```
verb: Correct, this shows up in pod status shortly after the container is created (assuming api server is reachable, etc). I've updated the doc to specify this.

Re: attach, I've updated the doc to clarify that the existing attach handler and protocol semantics are used.

contributors/design-proposals/troubleshoot-running-pods.md
verb: Added this info to the doc, thanks!

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,570 @@
...
+`DebugContainerRunner` interface.
+
+The Generic Runtime Manager's `RunDebugContainer()` calls `startContainer()` to
+create the Debug Container. Additionally, `SyncPod()` is modified to skip Debug
+Containers unless the sandbox is restarted.
+
+##### Step 3: kubelet API changes
+
+The kubelet gains a new streaming endpoint `/debugPod` and `server.getDebug()`
+which accepts a `v1.Container` via `POST` and calls `RunDebugContainer()` if
```
verb: Done.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,649 @@
...
+    Container named `$NAME` running in pod `$POD_NAME`.
+1.  A `DELETE` of `/api/v1/namespaces/$NS/pods/$POD_NAME/debug/$NAME` will stop
+    the Debug Container `$NAME` in pod `$POD_NAME`.
+
+Once created, a client can attach to the console of a debug container using the
+existing attach endpoint, `/api/v1/namespaces/$NS/pods/$POD_NAME/attach`. Status
+of Debug Containers is reported via a new field in `v1.PodStatus`, described in
+a subsequent section.
+
+#### Alternative: Extending `/exec`
```
verb: @dchen1107 @lavalamp @smarterclayton @pwittrock 

For pod troubleshooting we need to choose between the object-based approach described above (suggested by @smarterclayton ) and the exec-based approach described below (suggested by @lavalamp and resembling the "image exec" approach of the original proposal, interestingly).

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,666 @@
...
+
+```
+% kubectl debug -it -m debian neato-5thn0 -- bash
+root@debug-image:~# ps x
+  PID TTY      STAT   TIME COMMAND
+    1 ?        Ss     0:00 /pause
+   13 ?        Ss     0:00 bash
+   26 ?        Ss+    0:00 /neato
+  107 ?        R+     0:00 ps x
+root@debug-image:~# cat /proc/26/root/etc/resolv.conf
```
mikedanese: nit: I don't think this works with docker.

verb:
@mikedanese it has worked on the docker installs I've tried. Could you
provide details of the install where it doesn't work? I'd like to
investigate.

verb:
Feedback from API review with @lavalamp:

1. Don't use `debug` anywhere in the API. API names should reflect behavior rather than intended use.
2. This is not a good use case for a new imperative call that's passed-through to the kubelet. From an API perspective it would be better to have these debug containers reflected in the spec, which could be updated with a PATCH.

The consensus of sig-node is that the general container spec should be immutable, so we're at an impasse.

We could introduce a new field in the spec, something like `RunOnceContainers []Container`, and make that the only field allowed to be patched, but anything declared in the spec is easily mistaken for a regular container. From the perspective of the kubelet, this is not a general purpose container. It is an ephemeral container with no lifecycle or resource guarantees.

If we were to add to the additional complexity to the kubelet to make these general purpose, then we just have in the spec a container list that you're allowed to update and the one you're not allowed to update.

I'm happy to not use the word "debug" in the API, but I'm trying to build a better `exec` and I think the imperative call is justified. We need a compromise @derekwaynecarr @dchen1107 @vishh @smarterclayton

lavalamp:
Allowing mutations to containers marked `ephemeral: true` sounds pretty
good to me, actually. Documentation can clarify that ephemeral containers
have very few guarantees, etc...

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,666 @@
...
+good" or automated debugging environment to a pod.
+
+## Requirements
+
+A solution to troubleshoot arbitrary container images MUST:
+
+*   troubleshoot arbitrary running containers with minimal prior configuration
+*   allow access to shared pod namespaces and the file systems of individual
+    containers
+*   fetch troubleshooting utilities at debug time rather than at the time of pod
```
derekwaynecarr: fetch troubleshooting utilities at debug time rather than at the time of pod creation because...
- want to avoid image pulling?
- concerns about pod startup latency?

more detail here would be helpful.

sjenning: Yes, if this requirement is dropped, this becomes a lot simpler.  I know that istio is looking to enable sidecarring through an admission controller server-side https://github.com/istio/kube-inject (no code yet).  Currently they have a tool, `istio kube-inject`, that injects the istio sidecar into the pod spec client-side.

verb: @sjenning that's a clever solution that hadn't occurred to me, but I think it's solving a different problem. That sounds like a solution to a development or platform integration problem. Besides, let's not drop requirements to just make implementation easier. Let's solve the harder problem and delight the user.

Debug containers is an attempt to solve an operations and security problem. The operations problem is that I want to troubleshoot a running minimal container image (built from SCRATCH) without restarting it or having had to predict in advance which tools I would need.

Adding the container at pod creation has the following downsides:

- One of the motivations for minimal container images is to avoid deploying a full OS stack with every application instance to reduce attack surface.
- Specifying the debugging image at debug time allows freedom of choice between arbitrary debugging solutions. For example, I could choice between a "take snapshot and restart" container image or one that gives me an interactive shell.
- Choice of debugging image enables more useful cluster policy. For example, the "take snapshot and restart" image doesn't generate an audit alert while an interactive shell requires a human to review.
- Due to constraints of the runtime, a debugging container is only useful when it's running. Running an idle shell in hundreds of thousands of pods is inefficient and, frankly, feels like a hack.
- I want the latest debug tools rather than the old ones from when this pod was created.

These things are already mentioned in the proposal, but I understand that the proposal is too large for people to read because I've integrated hundreds (160 at last count, but that was before this PR was opened) of comments and suggestion.

Clearly I could be better at organizing information for this audience, and I would certainly appreciate any suggestions to that effect.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,666 @@
...
+
+We'll create a new API Object:
+
+```
+// PodDebugContainer describes a container to attach to a running pod for troubleshooting.
+type PodDebugContainer struct {
+        metav1.TypeMeta
+
+        // Template defines the pods that will be created from this pod template
+        Container v1.Container
```
derekwaynecarr: per our discussion, I do not support this being a v1.Container as that brings in resource requirements concerns which are not required for this use case (unless I am mistaken).  For example, I do not imagine a debug container would need access to its own GPU device.

derekwaynecarr:
this feature also has potential challenges with autoscaling, cc @DirectXMan12

contributors/design-proposals/troubleshoot-running-pods.md
verb: @derekwaynecarr I had planned on handling this with validation that the resources section is empty so that people can reuse the familiar `v1.Container`, but it's certainly not necessary and I'm happy to use a simpler object.

derekwaynecarr: @verb - it may be worth to just have a small section explaining it.  for example, i would not imagine that a debug container would change the QoS class of a pod.

@sjenning @vishh -- if this feature was enabled, and i had static cpu pinning enabled on the kubelet, what cpu should run this pod if it was a G workload?

ConnorDoyle: **Re: cpu manager**

If:
* Static cpu manager policy is enabled
* Pod QoS classification is Guaranteed
* Container requests an integer number _N_ of CPUs

Then the container will run in its own CPU set spanning _N_ logical CPUs.

Otherwise, it will run in the shared pool together with other Burstable, Best-Effort, and non-integral-cpu Guaranteed containers.

sjenning: @derekwaynecarr yes, to add to @ConnorDoyle comment, the decision to pin is a container-level decision, not pod-level.  So an added container will just be subjected to the same rules as the containers in the pod spec.

verb:
Updated proposal with results of API review

derekwaynecarr:
This still requires shared pid, should we rally on this proposal to identify a plan to address concerns that prevented that being the default in 1.8?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+
+## Implementation Details
+
+The implementation of `kubectl debug` closely mirrors the implementation of
+`kubectl exec`, with most of the complexity implemented in the `kubelet`. How
+functionality like this best fits into Kubernetes API has been contentious. In
+order to make progress, we will start with the smallest possible API change,
+extending `/exec` to support Debug Containers, and iterate.
+
+From the perspective of the user, there's a new command, `kubectl debug`, that
```
smarterclayton: I'd like to ensure any work on a debug command is compatible with future copy pod behavior (which is useful in many cases where this particular command is not).

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+}
+```
+
+After creating the Debug Container, the kubelet will upgrade the connection to
+streaming and perform an attach to the container's console. If disconnected, the
+Debug Container can be reattached using the pod's `/attach` endpoint with
+`DebugName`.
+
+Debug Containers cannot be removed via the API and instead the process must
+terminate. While not ideal, this parallels existing behavior of `kubectl exec`.
```
smarterclayton: kubectl exec in practice ends up closing commands early because those commands fail to write to stdout / stderr and then exit.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+To kill a Debug Container one would `attach` and exit the process interactively
+or create a new Debug Container to send a signal with `kill(1)` to the original
```
smarterclayton: That's pretty annoying / ugly.  If a user needs to guarantee the process exits, then debug containers become a significant liability.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+### Restarting and Reattaching Debug Containers
+
+Debug Containers will never be restarted automatically. It is possible to
+replace a Debug Container that has exited by re-using a Debug Container name. It
+is an error to attempt to replace a Debug Container that is still running, which
+is detected by both the API server and the kubelet.
+
+One can reattach to a Debug Container using `kubectl attach`. When supported by
+a runtime, multiple clients can attach to a single debug container and share the
+terminal. This is supported by Docker.
```
smarterclayton: What about other container runtimes?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+
+### Killing Debug Containers
+
+Debug containers will not be killed automatically until the pod (specifically,
+the pod sandbox) is destroyed. Debug Containers will stop when their command
+exits, such as exiting a shell. Unlike `kubectl exec`, processes in Debug
+Containers will not receive an EOF if their connection is interrupted.
```
smarterclayton: Why not?

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+Many people have proposed alternate solutions to this problem. This section
+discusses how the proposed solution meets all of the stated requirements and is
+intended to contrast the alternatives listed below.
+
+**Troubleshoot arbitrary running containers with minimal prior configuration.**
+This solution requires no prior configuration.
+
+**Access to all pod namespaces and the file systems of individual containers.**
+This solution runs a container in the shared pod namespaces. It relies on the
+behavior of `/proc/<pid>/root` (and therefore shared PID namespace) to provide
```
smarterclayton: I'm not convinced that most people are going to move to a shared pid namespace rapidly (because it's a significant change to how a multi container pod works).  Can you articulate why the solution requires a shared pid namespace for the pod and not just having the CRI share the pid namespace of the target container?

verb:
@derekwaynecarr It requires PID namespace sharing, but not necessarily with the entire pod. When the pod isn't sharing a single PID namespace the Debug Container will instead join the PID namespace of the target container and it will still be useful. I think it's safe for these two to proceed in parallel.

When namespace sharing is completely unavailable (e.g. docker < 1.13.1) then I think we should fail to create the Debug Container with an obvious error message (not supported by this runtime, etc) rather than create a crippled Debug Container in order to avoid confusing the user and/or making a bad impression.

contributors/design-proposals/troubleshoot-running-pods.md
verb: Agreed, this is important to me, too. Last I heard the desire for copy pod was to implement it entirely client side in `kubectl run`. It was important to SIG CLI to keep the concepts separate. As of right now I know of no conflict.

contributors/design-proposals/troubleshoot-running-pods.md
verb: Sorry, I'm not following. You're saying a common occurrence is that commands are getting errors when writing to stdout and so they immediately exit? Is it because users are exec'ing without Stdout=False?

We were hoping with Debug Containers to remove some of the confusion around stdin/stdout/stderr/TTY by specifying them as interactive containers that always have input/output streams and TTYs. Is this in line with your concerns?

I'll have to give some thought to the interaction with the other params in PodExecOptions, though.

contributors/design-proposals/troubleshoot-running-pods.md
verb: Agreed, but I've been stuck for almost a year. Lots of people have insightful, valid and conflicting opinions about what the API should be. I know it's not ideal, but can we start with this and iterate?

contributors/design-proposals/troubleshoot-running-pods.md
verb: I haven't tested them. I think this feature is useful enough to leave it open (at least for the alpha) and address problems in the runtimes. The runtimes don't have to support this, but I think it's reasonable to expect them to handle it gracefully since the situation is not unique to debug containers.

contributors/design-proposals/troubleshoot-running-pods.md
verb: By design. We want to support reattaching after an interrupted connection.

contributors/design-proposals/troubleshoot-running-pods.md
verb: We'll no longer require the entire pod to share the PID namespace. It will work as you describe.

When this part was written pod PID namespace sharing had already been implemented for all pods and we weren't expecting objections. I'll update it.

contributors/design-proposals/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+
+Debug Containers have no additional privileges above what is available to any
+`v1.Container`. It's the equivalent of configuring an shell container in a pod
+spec but created on demand.
+
+Admission plugins that guard `/exec` must be updated for the new parameters. In
+particular, they should enforce the same container image policy on the `Image`
+parameter as is enforced for regular containers. During the alpha phase we will
+additionally support a container image whitelist as a kubelet flag to allow
+cluster administrators to easily constraint debug container images.
```
verb: @derekwaynecarr I added a image whitelist as a flag to address your concerns about container image provenance, but I'm not sure I completely understand them. Could you tell me if this addresses your concerns sufficiently for the alpha? One will be able to, for example, allow only debian and alpine debug container images with something like `--debug-image-whitelist=debian,alpine`

verb:
Agreement from sig node yesterday is this can be submitted after @lavalamp signs off on API. I believe I've addressed all feedback. If I've lost track of anything please let me know.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,739 @@
...
+an HTTP parameter, so we must extend `v1.PodExecOptions` with the specific
+fields required for creating a Debug Container:
+
+```
+// PodExecOptions is the query options to a Pod's remote exec call
+type PodExecOptions struct {
+        ...
+        // DebugName is the name of the Debug Container. Its presence will cause
+        // exec to create a Debug Container rather than performing a runtime exec.
+        DebugName string `json:"debugName,omitempty" ...`
```
lavalamp: I vote for making a sub section:

```go 
type PodExecOptions struct {
  ...
  EphemeralContainer *PodExecEphemeralContainerSpec
}

type PodExecEphemeralContainerSpec struct {
  Name string
  Image string
}
```

verb: SGTM, though in [my naive prototype](https://github.com/verb/kubernetes/commit/c6a72efad4c72b2f7815c07d295b7ac9141199f5#diff-c73f80ad83608f18657d22a06950d929R475) the sub section wasn't populated from the HTTP params. 

I'm not very familiar with the api machinery so I probably just missed something. I see where [queryparams flattens the struct based on JSON names](https://github.com/kubernetes/apimachinery/blob/master/pkg/conversion/queryparams/convert.go#L129). Do I need to write a custom converter somewhere to get it back from params to an object? Could you point me in the right direction?

liggitt: yeah, nested structs parsed from query params won't work cleanly (https://github.com/kubernetes/kubernetes/issues/21476)

lavalamp: Earlier in the text it said this was not using query params--I thought it was using a message sent after the SPDY channel was opened.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,749 @@
...
+subsequently be used to reattach and is reported by `kubectl describe`.
+
+### Kubernetes API Changes
+
+#### Chosen Solution: "exec++"
+
+We will extend `v1.Pod`'s `/exec` subresource to support "executing" container
+images. The current `/exec` endpoint must implement `GET` to support streaming
+for all clients. We don't want to encode a (potentially large) `v1.Container` as
+an HTTP parameter, so we must extend `v1.PodExecOptions` with the specific
```
lavalamp: Oh, I see. I misunderstood this line.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,749 @@
...
+fields required for creating a Debug Container:
+
+```
+// PodExecOptions is the query options to a Pod's remote exec call
+type PodExecOptions struct {
+        ...
+        // Run Command in an ephemeral container which shares some namespaces with Container.
+        EphemeralContainer PodExecEphemeralContainerSpec
```
lavalamp: This needs to be a pointer. Uh, but I will suggest something else since sub structs are indeed currently broken.

lavalamp:
```go
// PodExecOptions is the query options to a Pod's remote exec call
type PodExecOptions struct {
  ...
  // EphemeralContainerName is the name of an ephemeral container in which the
  // command ought to be run. Either both EphemeralContainerName and
  // EphemeralContainerImage fields must be set, or neither.
  EphemeralContainerName *string `json:"ephemeralContainerName,omitempty" ...`

  // EphemeralContainerImage is the image of an ephemeral container in which the command
  // ought to be run. Either both EphemeralContainerName and EphemeralContainerImage
  // fields must be set, or neither.
  EphemeralContainerImage *string `json:"ephemeralContainerImage,omitempty" ...`
}
```

verb: Renamed as suggested

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+justified.
+
+### Debug Container Status
+
+The status of a Debug Container is reported in a new field in `v1.PodStatus`:
+
+```
+type PodStatus struct {
+        ...
+        DebugStatuses []DebugStatus
```
liggitt: should this mirror the exec option names? ephemeral statuses, etc?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+*   A `kubectl debug` command to initiate a debug container
+*   `kubectl describe pod` will list status of debug containers running in a pod
+
+Functionality will be hidden behind an alpha feature flag and disabled by
+default. The following are explicitly out of scope for the 1.9 alpha release:
+
+*   Exited Debug Containers will be garbage collected as regular containers and
+    may disappear from the list of Debug Container Statuses.
+*   Security Context for the Debug Container is not configurable. It will always
+    be run with `CAP_SYS_PTRACE` and `CAP_SYS_ADMIN`.
```
smarterclayton: So only a cluster admin should ever use debug containers?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+
+`DebugStatuses` is populated by the kubelet in the same way as regular and init
+container statuses. This is sent to the API server and displayed by `kubectl
+describe pod`.
+
+### Creating Debug Containers
+
+1.  `kubectl` invokes the debug API as described in the preceding section.
+1.  The API server checks for name collisions with existing containers, performs
+    admission control and proxies the connection to the kubelet's
```
liggitt: note that all admission plugins that do anything related to checking containers, images, etc, would need to be updated to check ephemeral images specified in exec options now

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+implicit attach maintains consistent semantics across `/exec` rather than
+varying behavior based on parameters.
+
+The apiserver detects container name collisions with both containers in the pod
+spec and other running Debug Containers by checking `DebugStatuses`. In a race
+to create two Debug Containers with the same name, the API server will pass both
+requests and the kubelet must return an error to all but one request.
+
+There are no limits on the number of Debug Containers that can be created in a
+pod, but exceeding a pod's resource allocation may cause the pod to be evicted.
```
liggitt: this also would bypass admission plugins that set resource limits/range on containers. please describe the container spec that would result from an ephemeral exec request

contributors/design-proposals/node/troubleshoot-running-pods.md
smarterclayton: We should use consistent naming

contributors/design-proposals/node/troubleshoot-running-pods.md
smarterclayton: Also, pods don't have security context set in many cases.  Exec implicitly escalating to root is bad.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+    reporting.
+1.  Debug Containers must be excluded from calculation of pod phase and
+    condition
+
+It's worth noting some things that do not change:
+
+1.  `KillPod()` already operates on all running containers returned by the
+    runtime.
+1.  Containers created prior to this feature being enabled will have a
+    `containerType` of `""`. Since this does not match `"DEBUG"` the special
```
liggitt: DEBUG or EPHEMERAL?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+
+Debug Containers have no additional privileges above what is available to any
+`v1.Container`. It's the equivalent of configuring an shell container in a pod
+spec but created on demand.
+
+Admission plugins that guard `/exec` must be updated for the new parameters. In
+particular, they should enforce the same container image policy on the `Image`
+parameter as is enforced for regular containers. During the alpha phase we will
+additionally support a container image whitelist as a kubelet flag to allow
+cluster administrators to easily constraint debug container images.
```
liggitt: what security context settings (uid/gid, selinux, apparmor) will the debug container have? how will admission plugins that constraint/force those (like PodSecurityPolicy) govern an ephemeral container

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+
+Functionality will be hidden behind an alpha feature flag and disabled by
+default. The following are explicitly out of scope for the 1.9 alpha release:
+
+*   Exited Debug Containers will be garbage collected as regular containers and
+    may disappear from the list of Debug Container Statuses.
+*   Security Context for the Debug Container is not configurable. It will always
+    be run with `CAP_SYS_PTRACE` and `CAP_SYS_ADMIN`.
+*   Image pull policy for the Debug Container is not configurable. It will
+    always be run with `PullAlways`.
```
liggitt: this prevents offline installations with pre-pulled images

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: Argh, yes, I was in a rush and didn't see this section. @verb can you modify to

```go
type PodStatus struct {
  
  EphemeralContainers []EphemeralContainerStatus
}
```

lavalamp: Hm. Actually were you trying to represent all exec actions with this?

lavalamp: (note I edited my comment above. I can't see anything you need that isn't already in v1.ContainerStatus.)

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -0,0 +1,743 @@
...
+
+The status of a Debug Container is reported in a new field in `v1.PodStatus`:
+
+```
+type PodStatus struct {
+        ...
+        DebugStatuses []DebugStatus
+}
+
+type DebugStatus struct {
```
lavalamp: I don't see the necessity of this.

If you wanted to represent the way(s) in which the container is "dirty" after exec/attach/portrforward etc, I don't think this is the way to go.

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: Yes please change references in the API to "ephemeral" everywhere.

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: I thought @verb had worked this out with @derekwaynecarr already?

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: @lavalamp I'd like to at least have command and args, which aren't part of ContainerStatus.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Yes, that's something we want to represent along with details about what command was run by exec (whether traditional exec or in an ephemeral container). This isn't a blocker for the alpha implementation, though, so if you think this is the wrong approach then I'll remove this bit from the proposal and we can figure out the correct way later.

/cc @dchen1107 @thockin

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: This is noted in the "Security Considerations" section.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: To be clear, this is a private label in the kubelet's runtime manager and not part of the API. I've updated it to EPHEMERAL for consistency, though.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: This should be configurable at debug time, but that depends on the API. Some of my proposed API changes addressed this, but SIG Node and the API reviewers deadlocked on which was best. Our compromise was to proceed to alpha with the minimum possible API change, which doesn't include a configurable security context.

Created https://github.com/kubernetes/kubernetes/issues/53188 to track this.

/cc @thockin

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Good point. Created https://github.com/kubernetes/kubernetes/issues/53189 to track this.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: @liggitt Compatibility with admission plugins is a top priority and a strict requirement. The implementation will depend a little bit on how the Kubernetes API settles, but it will be one of:

- The client may end up providing tje v1.Container that creates the Ephemeral Container
- If we stick with the imperative, exec-style API we can do exactly as you suggest and provide the v1.Container based on the PodExecOptions.

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: > Our compromise was to proceed to alpha with the minimum possible API change, which doesn't include a configurable security context.

you have a pointer to where that discussion happened? no one from the auth/psp side was involved afaik

verb: I don't understand, you've been involved since the first draft of the proposal 10 months ago, and your input has always been most welcome. It's not too late, what would you like to see changed?

liggitt: > It's not too late, what would you like to see changed?

there has to be control over the security aspects (uid/gid, capabilities, pid/network/ipc namespace, etc) of containers that are launched on the nodes. Omitting that control ("Security Context for the Debug Container is not configurable"), and defaulting to privileged levels ("It will always be run with `CAP_SYS_PTRACE` and `CAP_SYS_ADMIN`.") is a particular problem.

verb: Agreed, and I created an issue to track adding a configurable security context. Until then this feature is an alpha feature requiring cluster admins to enable it.

I don't think it's a good idea to encode an entire securityContext as an HTTP query parameter. Are there any fields you can't live without during the alpha? What about a toggle for SYS_ADMIN?

liggitt: > I created an issue to track adding a configurable security context. Until then this feature is an alpha feature requiring cluster admins to enable it.

When there are known gaps required for exiting alpha status that have yet-undetermined implications on the API for this feature, it doesn't seem wise to start making changes to stable APIs in master.

> I don't think it's a good idea to encode an entire securityContext as an HTTP query parameter. Are there any fields you can't live without during the alpha? What about a toggle for SYS_ADMIN?

I agree that encoding is ugly, and I don't a great idea there, but I think it's a mistake to not deal with this in terms of a standard `Container` definition:
* many parts of the security context has utility in some debug scenario:
  * running many debug tools require elevated capabilities
  * other scenarios would want to run as a specific uid/gid (plenty of overlap with https://github.com/kubernetes/kubernetes/issues/30656)
* admission plugins that make decisions on pods do so in terms of PodSpec and `Container` structs. What is the benefit of coming up with a different structure to describe the same thing?

contributors/design-proposals/node/troubleshoot-running-pods.md
thockin: I think we do want some way to represent the taintedness (also should do for exec )

contributors/design-proposals/node/troubleshoot-running-pods.md
thockin: Jordan,

Is your concern mainly that the structure is not a v1.Container ?  Or is the concern around the (ab)use of /exec?  I am not super keen on the /exec bits, but I want this to unstick and forward progress feels like a better tradeoff than another LITERAL year of wheel-spinning.

If we work through the changes required to plumb security down, even though not fully implementing for alpha, does that satisfy?  If not, can you explain exactly what you'd like to see for forward motion?

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: @r2d4 Can you work on the best way to represent taintedness in v1.PodStatus?

r2d4: @verb ack looking into it

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: > Is your concern mainly that the structure is not a v1.Container ? Or is the concern around the (ab)use of /exec? 

Mostly the former (the latter is ugly, but not fatal). I have three main concerns:

1. This proposes to build the container spec implicitly in the kubelet, given an image. That means admission doesn't have the information needed to allow/deny/modify that container spec, and if the kubelet changes how it defines the container spec release-to-release, admission can't even know what the kubelet will do.
2. There's not a clear path forward to support more aspects of the container spec (control over pull policy, uid/gid, capabilities, env, etc, all seem likely and reasonable)
3. The proposed defaults don't match the defaults for Container specs (specifically around security settings / capabilities), which would make transitioning to dealing with this as a `v1.Container` difficult

If we don't deal with this as a `v1.Container` consistently through the stack (api/admission/kubelet), we're making every layer have to understand two ways containers can be expressed, which seems like a mistake.

thockin: @verb how bad would it be to pass a full v1.Container?

verb: @thockin Not bad, It's what the kubelet does internally and I've had it working in a prototype.

We would use the API described in [Alternative 1](https://github.com/kubernetes/community/pull/649/files#diff-5cfb31b40ca47511743d0545d5697aa0R144) to POST a v1.Container (wrapped in a new top level object) to a new /debug subresource. Then the client would perform a separate /attach.

The API reviewers had concerns about this being a novel use of the API. Nothing else POSTs to a subresource.

Both the API reviewers and SIG Node reviewers had concerns about using v1.Container being confusing or communicating the wrong intent to the user. Debug Containers are not general purpose containers and should not be used to build services or for routine operations. Some prefer to consider "extended exec" rather than "configurable container".

Most of the fields of v1.Container do not apply to Debug Containers and should be rejected if configured (lifecycle, livenessProbe, ports, readinessProbe, resources, stdin, stdinOnce, terminationMessagePath, terminationMessagePolicy, tty, volumeMounts). We can do this with a validation whitelist, though it would be simpler to pass a securityContext rather than a full v1.Container.

davidopp: > The API reviewers had concerns about this being a novel use of the API. Nothing else POSTs to a subresource.

I thought scheduler POSTs to pod/binding subresource, clients POST to pod/eviction subresource, etc. Or is there a distinction between POST and PUT here? (Which I can never keep straight.)

verb: @davidopp Oh, that's good news then. When I prototyped this ~6 months ago I recall needing a couple of changes in the apiserver in order to make it work, but maybe those were specific to upgrading a connection to streaming after a POST. It's been a long ride.

Let's figure out a way to move this forward. Since we already had agreement among the reviewers at the time, and now we want to renegotiate that agreement based on new reviewers, I suggest that take the form of a new PR to amend the proposal where the new and old reviewers can work out their conflicting requirements. I'll prepare a diff.

liggitt: >> The API reviewers had concerns about this being a novel use of the API. Nothing else POSTs to a subresource.
>
> I thought scheduler POSTs to pod/binding subresource, clients POST to pod/eviction subresource, etc. Or is there a distinction between POST and PUT here? (Which I can never keep straight.)

Not sure about subresources, but there are issues with both PUT and POST as part of *websocket* requests (not all clients support them)

ndeloof:
Hi @verb , I'm glad to see your repeated effort to get this feature added to k8s.

I understand the initial use-case for this effort is related to diagnostic on a running Pod, but would like to share with you another use-case which would benefit this exact same improvement:

I'm working on Jenkins integration in k8s, we run containerized builds as pods. During the build execution, a new container might be required. In many case this can be identified _before_ the build is scheduled and as such set as part of the Pod's spec, but in some cases the required image is dynamically selected as part of the build. Also, being able to run containers as part of the build just like developer do on their workstation helps to make the build script reproducible and portable.

with a new API to add (transient) containers to a Pod I could provide the glue code for build script to control such additional containers. The current usage for most users is to run a privileged (!) DinD container to host the build, or to bind mount docker.sock from host (!). As you can guess I'd prefer we don't rely on this :-\

Hope this will help understand potential use-cases this feature could support.

smarterclayton:
Note that I'm specifically saying:

Once ephemeral containers are available, we have the rough shape of an api
to further subdivide a pod (subcontainer division).  We could potentially
put a whole class of problems ("how do i orchestrate a set of containers
within the pod") into the same pot by having the user run a container that
spawns ephemeral containers (in whatever fashion the user wants) and using
a combination of process and container controls to manage subdivision.
I.e. instead of kube implementing a task like mechanism within a pod, make
it possible for the user to safely implement that themselves.  The
ephemeral pod API then becomes an "in pod orchestrator" which exposes the
container runtime in a kube-like fashion to the user, rather than forcing
the user to implement nested containers (which are invisible to kube) or
having kube continue to add feature after feature to approximate a directed
task graph.

avanier:
I have a keen interest for this feature. Does anyone have movement to report on this outside of this ticket?

Also, to echo @ndeloof mentions for debuging Jenkins, I have to say this is indeed quite useful. I've been running [Concourse](https://concourse-ci.org/) at a few shops over recent years, and using Pivotal's Garden OCI runtime they achieve exactly that.

verb:
@avanier kubernetes/features#277 might be better for tracking progress of this feature. The API change is under review in kubernetes/kubernetes#59416, once the API changes there should be quick progress.

PR #650, opened 2017-05-22
Add proposal for volume spec reconstruction

chakri-nelluri:
Add proposal for volume spec reconstruction.
Addresses: https://github.com/kubernetes/kubernetes/issues/44737

### DIFF
--- a/contributors/design-proposals/volume-spec-reconstruction.md
+++ b/contributors/design-proposals/volume-spec-reconstruction.md
@@ -1,0 +1,34 @@
+# Abstract
+Today, Kubelet dynamically re-constructs volume spec during runtime. It reconstructs using volume mount path. But for many plugins volume mount path is not enough to reconstruct all the fields/options. As a result, some of the plugin API which depend on these fields/options are broken, especially in the cleanup code. The goal of this exercise is to store volume specific meta-data and use that information to reconstruct the spec.
+
+# High Level Design
+Store volume specific meta-data during volume mountdevice/mount/update logic and use it to reconstruct the volume spec. If it is an attachable plugin, store the meta-data during mountdevice and remove after a successful unmountdevice. For plugins, which use only mount & unmount API, store the meta-data during the mount API and remove during the umount API. Meta-data should also be updated, once we add support for volume spec update.
+
+## Meta-data format
+
+### Option 1:
+ Store the volume object source in JSON format. Example: for GCE plugin store "GCEPersistentDiskVolumeSource".
+ #### Pros:
+ * Simpler to implement.
+ #### Cons:
+ * Volume object source is not versioned and upgrades will be a problem going forward.
+### Option 2:
+ Store Persistent volume spec/Volume Source in JSON format. Persistent volume spec is stored if the volume is backed by a Persistent Volume Object. Volume source is stored if the volume source is inlined in Pod spec. We can use the different filenames \<volume name\>~pv.json and \<volume name\>~vs.json to distinguish these objects.
+ #### Pros:
+ * Persistent volume spec is versioned.
+ #### Cons:
+ * Complicated naming.
+ * VolumeSource is still not versioned.
+ * Persistent volume does not have the namespace information for the secretref.
+### Option 3:
+ Implement a per plugin specific API to store the meta-data relevant to the plugin. We can provide a sensible default using Option 2/Option 1 if plugin does not want to implement these API.
+ #### Pros:
+ * Plugins are free to implement/experiment with the information they want to store.
+ #### Cons:
+ * Versioning the meta-data is offloaded to the plugin.
+  
+## Meta-data location:
+Store the meta-data file \<volume name\>.json in the plugin path.
+Ex:
+For plugin: ```kubernetes.io/lvm``` store in the following directory
+```/var/lib/kubelet/plugins/kubernetes.io/flexvolume/kubernetes.io/lvm```


chakri-nelluri:
@saad-ali @jsafrane @rootfs PTAL when you get a chance.

contributors/design-proposals/volume-spec-reconstruction.md
```diff
@@ -0,0 +1,34 @@
...
+ * Persistent volume does not have the namespace information for the secretref.
+### Option 3:
+ Implement a per plugin specific API to store the meta-data relevant to the plugin. We can provide a sensible default using Option 2/Option 1 if plugin does not want to implement these API.
+ #### Pros:
+ * Plugins are free to implement/experiment with the information they want to store.
+ #### Cons:
+ * Versioning the meta-data is offloaded to the plugin.
+  
+## Meta-data location:
+Store the meta-data file \<volume name\>.json in the plugin path.
```
jsafrane: controller-manager can run on several machines in hot-standby HA, where one controller-manager is master (and runs controllers) and the other controller-managers just wait for the master to die so they can quickly resume. The metadata won't be available to a new master on another machine if you store the metadata on the filesystem. I am afraid that these metadata must be stored in API server.

chakri-nelluri: The information will be stored as part of mountdevice call, which is only executed on worker nodes. Do we need this information and spec reconstruction in controller-manager code path too?

jsafrane: oh, ok, I think we don't need it in the controller-manager, @tsmetana or @jingxu97 can confirm

jsafrane: and would you please add a note that this will be stored only on nodes and not on controller-manager?

chakri-nelluri: Sure will do.

saad-ali:
CC @kubernetes/sig-storage-feature-requests @kubernetes/sig-storage-misc 

> Option 1:
> Store the volume object source in JSON format. Example: for GCE plugin store "GCEPersistentDiskVolumeSource".
>
> Option 2:
> Store Persistent volume spec/Volume Source in JSON format. Persistent volume spec is stored if the volume is backed by a Persistent Volume Object. Volume source is stored if the volume source is inlined in Pod spec. We can use the different filenames <volume name>~pv.json and <volume name>~vs.json to distinguish these objects.
> 
> Option 3:
> Implement a per plugin specific API to store the meta-data relevant to the plugin. We can provide a sensible default using Option 2/Option 1 if plugin does not want to implement these API.

I think option 3 is a non-starter. That will lead to fragmentation and counter to our end goal of providing k8s kubelet a reliable way to reconstruct lost internal state.

Option 1 is insufficient for all the information we may need, and even option 2 does not go far enough.

The information we need for kubelet volume reconstruction is spread across 3 API objects:
1) PV Object
  * This includes the `PersistentVolumeSource`. Which we must have for PV defined objects to be able to run `GetVolumeName(...)` and the PV name required for generating the `InnerVolumeSpecName`.
  * It also includes things like `AccessMode` and `Capacity` which are likely not useful for kubelet volume reconstruction.
2) PVC Object
  * Most likely not useful for kubelet volume reconstruction
3) Pod API Object
  * Contains the `Pod.Spec.Volumes[x].VolumeSource` along with the user defined `Pod.Spec.Volumes[x].Name` -- both are required to run `GetVolumeName(...)` and generating the `InnerVolumeSpecName` and   `OuterVolumeSpecName`.
  * Also contains `Pod.Spec.Containers[x].VolumeMounts` which is likely not useful for kubelet volume reconstruction.

A 4th option I did not see proposed, is to figure out what fields Kubernetes needs and to store only those. I admit that would be more difficult to implement and harder to maintain (you need to worry about backwards compatibility, you need to find a way to encode volume plugin specific information, etc.).

So, it sounds like the easiest options would be to just save the entity of the involved objects (pod and PV) so that we are sure to have everything we need at reconstruction time. At a minimum we'd need to store some part of the Pod API object and the PV object (if not direct reference volume). But, like you pointed out, we need to worry about versioning and the easiest way around that would be to store the entire object (including version). Not sure if this is kosher, however, esp for pod objects.

chakri-nelluri:
@shilpamayanna PTAL

chakri-nelluri:
Thanks @saad-ali. I prefer storing the PV Object too, but if the spec is inlined in pod spec, we have to store the entire pod spec. I don't like that part. Lets pick a choice, before it's too late.
@rootfs @jsafrane any preferences?

Between, @ShilpaMayanna is working on implementing this support.

chakri-nelluri:
@saad-ali PTAL. Waiting on you for LGTM.
I am on PTO this week and @shilpamayanna is working on this.

saad-ali:
Freeze date for 1.7 was June 1, so we missed the boat. Let's prioritize it high for 1.8.

In the meantime please make sure that any bug fixes that need to go in for 1.7 to mitigate https://github.com/kubernetes/kubernetes/issues/44737 are in the master/1.7 branches.

chakri-nelluri:
@saad-ali PTAL. Waiting on you for LGTM.
@shilpamayanna is back from PTO. She is planning to work on it this week.

saad-ali:
Spoke with @thockin for some ideas. His suggestion was to use Kubernetes Component Config which gives us versioned APIs, backward compatibility, deprecation policies, documentation, etc for arbitrary configuration files. He also was leaning towards a per volume hook.

My concern with a per volume hook is whether it would be sufficient to capture all the information that the volume code would need (PV name, pod volume name, etc.). I feel like we may need to store the entire PV, PVC, and lots (if not all) the pod (volumes, mounts, etc.). Tim's suggestion was to store only what we need, and create our own versioned API to do that. Problem with building our own API is that it is one more thing to maintain, and as the external k8s volume API (or internal volume state management) changes, we'd need to make changes (and maintain backwards compatibility) in this layer.

@shilpamayanna what are your thoughts?

+@msau42

contributors/design-proposals/volume-spec-reconstruction.md
j-griffith: Can we just introduce a standard 'provider_info' key in json format and the plugin can determine what it needs to store here on it's own?  In addition, it would be up to the driver to reconnect if needed and it would utilize that provider_info data.

childsb:
Kubelet checkpointing is similar concept for all of kubelet: https://github.com/kubernetes/kubernetes/issues/489

jingxu97:
I think disable some volume plugins reconstructVolume is still needed. For example, in glusterfs
1. ConstructVolumeSpec function in glusterfs.go uses volumeName as endpoint and path which is wrong.
https://github.com/kubernetes/kubernetes/blob/release-1.6.3/pkg/volume/glusterfs/glusterfs.go#L197
2. reconstructVolume function in reconciler will try to call  plugin.NewMounter() to get the mounter. This NewMounter function will in turn call API server to get the endpoint object which will always fail. https://github.com/kubernetes/kubernetes/blob/release-1.6.3/pkg/volume/glusterfs/glusterfs.go#L143
In this you can see many error message in the log saying "glusterfs: failed to get endpoints". Also since this reconstructVolume is called periodically, doing this is wasting the API calls because endpoint is the fake name.

shilpamayanna:
For now, we will not implement GetVolumeName in flex volume to avoid volumes from getting unmounted and update flex volume documentation with limitations for inline volumes in pod spec.

PR #654, opened 2017-05-23
Add CPU manager proposal.

ConnorDoyle:
**Notes for reviewers**

First proposal submitted to the community repo, please advise if something's not right with the format or procedure, etc.

cc @flyingcougar @sjenning @derekwaynecarr

### DIFF
--- a/contributors/design-proposals/cpu-manager.md
+++ b/contributors/design-proposals/cpu-manager.md
@@ -1,0 +1,343 @@
+# CPU Manager
+
+_Authors:_
+
+* @ConnorDoyle - Connor Doyle &lt;connor.p.doyle@intel.com&gt;
+* @flyingcougar - Szymon Scharmach &lt;szymon.scharmach@intel.com&gt;
+* @sjenning - Seth Jennings &lt;sjenning@redhat.com&gt;
+
+**Contents:**
+
+* [Overview](#overview)
+* [Proposed changes](#proposed-changes)
+* [Operations and observability](#operations-and-observability)
+* [Practical challenges](#practical-challenges)
+* [Implementation roadmap](#implementation-roadmap)
+* [Appendix A: cpuset pitfalls](#appendix-a-cpuset-pitfalls)
+
+## Overview
+
+_Problems to solve:_
+
+1. Poor or unpredictable performance observed compared to virtual machine
+   based orchestration systems. Application latency and lower CPU
+   throughput compared to VMs due to cpu quota being fulfilled across all
+   cores, rather than exclusive cores, which results in fewer context
+   switches and higher cache affinity.
+1. Unacceptable latency attributed to the OS process scheduler, especially
+   for “fast” virtual network functions (want to approach line rate on
+   modern server NICs.)
+
+_Solution requirements:_
+
+1. Provide an API-driven contract from the system to a user: "if you are a
+   Guaranteed pod with 1 or more cores of cpu, the system will try to make
+   sure that the pod gets its cpu quota primarily from reserved core(s),
+   resulting in fewer context switches and higher cache affinity".
+1. Support the case where in a given pod, one container is latency-critical
+   and another is not (e.g. auxillary side-car containers responsible for
+   log forwarding, metrics collection and the like.)
+1. Do not cap CPU quota for guaranteed containers that are granted
+   exclusive cores, since that would be antithetical to (1) above.
+1. Take physical processor topology into account in the CPU affinity policy.
+
+### Related issues
+
+* Feature: [Further differentiate performance characteristics associated
+  with pod level QoS](https://github.com/kubernetes/features/issues/276)
+
+## Proposed changes
+
+### CPU Manager component
+
+The *CPU Manager* is a new software component in Kubelet responsible for
+assigning pod containers to sets of CPUs on the local node. In the
+future, it may be expanded to control shared processor resources like
+caches.
+
+The CPU manager interacts directly with the kuberuntime. The CPU Manager
+is notified when containers come and go, before delegating container
+creation via the container runtime interface and after the container's
+destruction respectively. The CPU Manager emits CPU settings for
+containers in response.
+
+#### Discovering CPU topology
+
+The CPU Manager must understand basic topology. First of all, it must
+determine the number of logical CPUs (hardware threads) available for
+allocation. On architectures that support [hyper-threading][ht], sibling
+threads share a number of hardware resources including the cache
+hierarchy. On multi-socket systems, logical CPUs co-resident on a socket
+share L3 cache. Although there may be some programs that benefit from
+disjoint caches, the policies described in this proposal assume cache
+affinity will yield better application and overall system performance for
+most cases. In all scenarios described below, we prefer to acquire logical
+CPUs topologically. For example, allocating two CPUs on a system that has
+hyper-threading turned on yields both sibling threads on the same
+physical core. Likewise, allocating two CPUs on a non-hyper-threaded
+system yields two cores on the same socket.
+
+##### Options for discovering topology
+
+1. Read and parse the virtual file [`/proc/cpuinfo`][procfs] and construct a
+   convenient data structure.
+1. Execute a simple program like `lscpu -p` in a subprocess and construct a
+   convenient data structure based on the output. Here is an example of
+   [data structure to represent CPU topology][topo] in go. The linked package
+   contains code to build a ThreadSet from the output of `lscpu -p`.
+1. Execute a mature external topology program like [`mpi-hwloc`][hwloc] --
+   potentially adding support for the hwloc file format to the Kubelet.
+
+#### CPU Manager interfaces (sketch)
+
+```go
+type CPUManagerPolicy interface {
+  Init(driver CPUDriver, topo CPUTopo)
+  Add(c v1.Container, qos QoS) error
+  Remove(c v1.Container, qos QoS) error
+}
+
+type CPUDriver {
+  GetPods() []v1.Pod
+  GetCPUs(containerID string) CPUList
+  SetCPUs(containerID string, clist CPUList) error
+  // Future: RDT L3 and L2 cache masks, etc.
+}
+
+type CPUTopo TBD
+
+type CPUList string
+
+func (c CPUList) Size() int {}
+
+// Returns a CPU list with size n and the remainder or
+// an error if the request cannot be satisfied, taking
+// into account the supplied topology.
+//
+// @post: c = set_union(taken, remaining),
+//        empty_set = set_intersection(taken, remainder)
+func (c CPUList) Take(n int, topo CPUTopo) (taken CPUList,
+                                            remainder CPUList,
+                                            err error) {}
+
+// Returns a CPU list that includes all CPUs in c and d and no others.
+//
+// @post: result = set_union(c, d)
+func (c CPUList) Add(d CPUList) (result CPUList) {}
+```
+
+Kubernetes will ship with three CPU manager policies. Only one policy is
+active at a time on a given node, chosen by the operator via Kubelet
+configuration. The three policies are **no-op**, **static** and **dynamic**.
+Each policy is described below.
+
+#### Policy 1: "no-op" cpuset control [default]
+
+This policy preserves the existing Kubelet behavior of doing nothing
+with the cgroup `cpuset.cpus` and `cpuset.mems` controls. This “no-op”
+policy would become the default CPU Manager policy until the effects of
+the other policies are better understood.
+
+#### Policy 2: "static" cpuset control
+
+The "static" policy allocates exclusive CPUs for containers if they are
+included in a pod of "Guaranteed" [QoS class][qos] and the container's
+resource limit for the CPU resource is an integer greater than or
+equal to one.
+
+When exclusive CPUs are allocated for a container, those CPUs are
+removed from the allowed CPUs of every other container running on the
+node. Once allocated at pod admission time, an exclusive CPU remains
+assigned to a single container for the lifetime of the pod (until it
+becomes terminal.)
+
+##### Implementation sketch
+
+```go
+// Implements CPUManagerPolicy
+type staticManager struct {
+  driver CPUDriver
+  topo   CPUTopo
+  // CPU list assigned to non-exclusive containers.
+  shared CPUList
+}
+
+func (m *staticManager) Init(driver CPUDriver, topo CPUTopo) {
+  m.driver = driver
+  m.topo = topo
+}
+
+func (m *staticManager) Add(c v1.Container, qos QoS) error {
+  if p.QoS == GUARANTEED && numExclusive(c) > 0 {
+    excl, err := allocate(numExclusive(c))
+    if err != nil {
+      return err
+    }
+    m.driver.SetCPUs(c.ID, excl)
+    return nil
+  }
+
+  // Default case: assign the shared set.
+  m.driver.SetCPUs(c.ID, m.shared)
+  return nil
+}
+
+func (m *staticManager) Remove(c v1.Container, qos QoS) error {
+  m.free(m.driver.GetCPUs(c.ID))
+}
+
+func (m *staticManager) allocate(n int) (CPUList, err) {
+  excl, remaining, err := m.shared.Take(n, m.topo)
+  if err != nil {
+    return "", err
+  }
+  m.setShared(remaining)
+  return excl, nil
+}
+
+func (m *staticManager) free(c CPUList) {
+  m.setShared(m.shared.add(c))
+}
+
+func (m *staticManager) setShared(c CPUList) {
+  prev := m.shared
+  m.shared = c
+  for _, pod := range m.driver.GetPods() {
+    for _, container := range p.Containers {
+      if driver.GetCPUs(container.ID) == prev {
+        driver.SetCPUs(m.shared)
+      }
+    }
+  }
+}
+
+// @pre: container_qos = guaranteed
+func numExclusive(c v1.Container) int {
+  if c.resources.requests["cpu"] % 1000 == 0 {
+    return c.resources.requests["cpu"] / 1000
+  }
+  return 0
+}
+```
+
+##### Example pod specs and interpretation
+
+| Pod                                        | Interpretation                 |
+| ------------------------------------------ | ------------------------------ |
+| Pod [Guaranteed]:<br />&emsp;A:<br />&emsp;&emsp;cpu: 0.5 | Container **A** is assigned to the shared cpuset. |
+| Pod [Guaranteed]:<br />&emsp;A:<br />&emsp;&emsp;cpu: 2.0 | Container **A** is assigned two sibling threads on the same physical core (HT) or two physical cores on the same socket (no HT.)<br /><br /> The shared cpuset is shrunk to  make room for the exclusively allocated CPUs. |
+| Pod [Guaranteed]:<br />&emsp;A:<br />&emsp;&emsp;cpu: 1.0<br />&emsp;A:<br />&emsp;&emsp;cpu: 0.5 | Container **A** is assigned one exclusive CPU and container **B** is assigned to the shared cpuset. |
+| Pod [Guaranteed]:<br />&emsp;A:<br />&emsp;&emsp;cpu: 1.5<br />&emsp;A:<br />&emsp;&emsp;cpu: 0.5 | Both containers **A** and **B** are assigned to the shared cpuset. |
+| Pod [Burstable] | All containers are assigned to the shared cpuset. |
+| Pod [BestEffort] | All containers are assigned to the shared cpuset. |
+
+#### Policy 3: "dynamic" cpuset control
+
+_TODO: Describe the policy._
+
+##### Implementation sketch
+
+```go
+// Implements CPUManagerPolicy.
+type dynamicManager struct {}
+
+func (m *dynamicManager) Init(driver CPUDriver, topo CPUTopo) {
+  // TODO
+}
+
+func (m *dynamicManager) Add(c v1.Container, qos QoS) error {
+  // TODO
+}
+
+func (m *dynamicManager) Remove(c v1.Container, qos QoS) error {
+  // TODO
+}
+```
+
+##### Example pod specs and interpretation
+
+| Pod                                        | Interpretation                 |
+| ------------------------------------------ | ------------------------------ |
+|                                            |                                |
+|                                            |                                |
+
+## Operations and observability
+
+* Checkpointing assignments
+  * The CPU Manager must be able to pick up where it left off in case the
+    Kubelet restarts for any reason.
+* Read effective CPU assinments at runtime for alerting. This could be
+  satisfied by the checkpointing requirement.
+* Configuration
+  * How does the CPU Manager coexist with existing kube-reserved
+    settings?
+  * How does the CPU Manager coexist with related Linux kernel
+    configuration (e.g. `isolcpus`.) The operator may want to specify a
+    low-water-mark for the size of the shared cpuset. The operator may
+    want to correlate exclusive cores with the isolated CPUs, in which
+    case the strategy outlined above where allocations are taken
+    directly from the shared pool is too simplistic. We could allow an
+    explicit pool of cores that may be exclusively allocated and default
+    this to the shared pool (leaving at least one core fro the shared
+    cpuset to be used for OS, infra and non-exclusive containers.
+
+## Practical challenges
+
+1. Synchronizing CPU Manager state with the container runtime via the
+   CRI. Runc/libcontainer allows container cgroup settings to be updtaed
+   after creation, but neither the Kubelet docker shim nor the CRI
+   implement a similar interface.
+    1. Mitigation: [PR 46105](https://github.com/kubernetes/kubernetes/pull/46105)
+
+## Implementation roadmap
+
+### Phase 1
+
+* Internal API exists to allocate CPUs to containers
+  ([PR 46105](https://github.com/kubernetes/kubernetes/pull/46105))
+* Kubelet configuration includes a CPU manager policy (initially only no-op)
+* No-op policy is implemented.
+* All existing unit and e2e tests pass.
+* Initial unit tests pass.
+
+### Phase 2
+
+* Kubelet can discover "basic" CPU topology (HT-to-physical-core map)
+* Static policy is implemented.
+* Unit tests for static policy pass.
+* e2e tests for static policy pass.
+* Performance metrics for one or more plausible synthetic workloads show
+  benefit over no-op policy.
+
+### Phase 3
+
+* Dynamic policy is implemented.
+* Unit tests for dynamic policy pass.
+* e2e tests for dynamic policy pass.
+* Performance metrics for one or more plausible synthetic workloads show
+  benefit over no-op policy.
+
+### Phase 4
+
+* Kubelet can discover "advanced" CPU topology (NUMA).
+
+## Appendix A: cpuset pitfalls
+
+1. `cpuset.sched_relax_domain_level`
+1. Child cpusets must be subsets of their parents. If B is a child of A,
+   then B must be a subset of A. Attempting to shrink A such that B
+   would contain allowed CPUs not in A is not allowed (the write will
+   fail.) Nested cpusets must be shrunk bottom-up. By the same rationale,
+   nested cpusets must be expanded top-down.
+1. Dynamically changing cpusets by directly writing to the sysfs would
+   create inconsistencies with container runtimes.
+1. The `exclusive` flag. This will not be used. We will achieve
+   exclusivity for a CPU by removing it from all other assigned cpusets.
+1. Tricky semantics when cpusets are combined with CFS shares and quota.
+
+[ht]: http://www.intel.com/content/www/us/en/architecture-and-technology/hyper-threading/hyper-threading-technology.html
+[hwloc]: https://www.open-mpi.org/projects/hwloc
+[procfs]: http://man7.org/linux/man-pages/man5/proc.5.html
+[qos]: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-qos.md
+[topo]:
+http://github.com/intelsdi-x/swan/tree/master/pkg/isolation/topo
--- a/contributors/design-proposals/cpu-manager.md
+++ b/contributors/design-proposals/cpu-manager.md
@@ -149,7 +149,7 @@ document][node-allocatable] for details. The CPU manager will claim
 `ceiling(node.status.allocatable.cpu)` as the number of CPUs available to
 assign to pods, starting from the highest-numbered physical core and
 descending topologically. It is recommended to configure an integer value for
-`node.status.allocatable.cpus` when the CPU manager is enabled.
+`node.status.allocatable.cpu` when the CPU manager is enabled.
 
 Operator documentation will be updated to explain how to configure the
 system to use the low-numbered physical cores for kube-reserved and
@@ -181,14 +181,13 @@ becomes terminal.)
 
 ```go
 func (p *staticPolicy) Start(s State) {
-  // Iteration starts at index `1` here because CPU `0` is reserved
-  // for infrastructure processes.
-  // TODO(CD): Improve this to align with kube/system reserved resources.
-  shared := NewCPUSet()
-  for cpuid := 1; cpuid < p.topology.NumCPUs; cpuid++ {
-    shared.Add(cpuid)
-  }
-  s.SetDefaultCPUSet(shared)
+	fullCpuset := cpuset.NewCPUSet()
+	for cpuid := 0; cpuid < p.topology.NumCPUs; cpuid++ {
+		fullCpuset.Add(cpuid)
+	}
+	// Figure out which cores shall not be used in shared pool
+	reserved, _ := takeByTopology(p.topology, fullCpuset, p.topology.NumReservedCores)
+	s.SetDefaultCPUSet(fullCpuset.Difference(reserved))
 }
 
 func (p *staticPolicy) RegisterContainer(s State, pod *Pod, container *Container, containerID string) error {
--- a/contributors/design-proposals/cpu-manager.md
+++ b/contributors/design-proposals/cpu-manager.md
@@ -138,29 +138,31 @@ type CPUTopology TBD
 
 Kubernetes will ship with three CPU manager policies. Only one policy is
 active at a time on a given node, chosen by the operator via Kubelet
-configuration. The three policies are **noop**, **static** and **dynamic**.
+configuration. The three policies are **none**, **static** and **dynamic**.
 
 The active CPU manager policy is set through a new Kubelet
-configuration value `--cpu-manager-policy`. The default value is `noop`.
+configuration value `--cpu-manager-policy`. The default value is `none`.
 
-The number of CPUs that pods may run on is set using the existing
-node-allocatable configuration settings. See the [node allocatable proposal
-document][node-allocatable] for details. The CPU manager will claim
+The number of CPUs that pods may run on can be implicitly controlled using the
+existing node-allocatable configuration settings. See the [node allocatable
+proposal document][node-allocatable] for details. The CPU manager will claim
 `ceiling(node.status.allocatable.cpu)` as the number of CPUs available to
 assign to pods, starting from the highest-numbered physical core and
-descending topologically. It is recommended to configure an integer value for
-`node.status.allocatable.cpu` when the CPU manager is enabled.
+descending topologically. It is recommended to configure `kube-reserved`
+and `system-reserved` such that their sum is an integer when the CPU manager
+is enabled. This ensures that `node.status.allocatable.cpu` is also an
+integer.
 
 Operator documentation will be updated to explain how to configure the
 system to use the low-numbered physical cores for kube-reserved and
-system-reserved slices.
+system-reserved cgroups.
 
 Each policy is described below.
 
-#### Policy 1: "no-op" cpuset control [default]
+#### Policy 1: "none" cpuset control [default]
 
 This policy preserves the existing Kubelet behavior of doing nothing
-with the cgroup `cpuset.cpus` and `cpuset.mems` controls. This “no-op”
+with the cgroup `cpuset.cpus` and `cpuset.mems` controls. This "none"
 policy would become the default CPU Manager policy until the effects of
 the other policies are better understood.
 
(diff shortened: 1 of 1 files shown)


ConnorDoyle:
Attn @kubernetes/sig-node-proposals

derekwaynecarr:
fyi @jeremyeder

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,343 @@
...
+   for “fast” virtual network functions (want to approach line rate on
+   modern server NICs.)
+
+_Solution requirements:_
+
+1. Provide an API-driven contract from the system to a user: "if you are a
+   Guaranteed pod with 1 or more cores of cpu, the system will try to make
+   sure that the pod gets its cpu quota primarily from reserved core(s),
+   resulting in fewer context switches and higher cache affinity".
+1. Support the case where in a given pod, one container is latency-critical
```
derekwaynecarr: i assume this is the case where a pod has a container that uses integral cores and another container that uses fractional?

ConnorDoyle: Yes, that's how the static policy as written satisfies this requirement. Does that need clarification here?

flyingcougar:
CRI update needed by CPU Manager - https://github.com/kubernetes/kubernetes/pull/46105

chino:
Will it be possible to reserve a core for a pod but not actually assign a "container" to it?  We actually dedicate cores to very specific application "threads".  We currently do this by setting aside cores at boot time using `isolcpus`, disabling interrupts on them, and then using `taskset` from within the container at runtime.  This way a single thread is never context switched.

Not disturbing the reactor thread is critical for our low latency / low jitter workloads.

Instead of using taskset it might make more sense to perhaps communicate with the manager at runtime from the container?

flyingcougar:
@chino Making CPU Manager aware of `isolcpus` and using only those as exclusive makes sense as a part of static policy or new policy. 
In any case IMHO removing the need for user to manually manage cpus and prefix container entrypoint (or pod manifest) with `taskset` is an advantage.
Which cores on which nodes are in `isolcpus`, which are not used at particular point in time  - can be troublesome if you have a cluster of size i.e. more than 10 nodes.

ConnorDoyle:
@chino there has some debate about similar use cases so far. This proposal represents a simplified compromise between the most sophisticated scenarios and some level of support that will benefit many users. We at Intel have worked with users with exactly this requirement, and the strategy you described is pretty common among the low latency crowd.

Process scheduler load balancing is off for those CPUs, so applications need awareness to call sched_setaffinity() (or wrap with taskset/numactl) in order to use more than one. You could argue that the additional complexity breaks fungibility, so one way to look at isolcpus could be as a separate resource. In that case, one of the policies could be expanded to discover, advertise and make node-level allocation decisions for an isolated-cpu resource. If the use case isn't a natural fit with the upstream CPU Manager policies, then it may motivate a plugin interface so more specialized policies can be added by users. Before we retreat to that alternative, we should try to see how it could be satisfied in this proposal.

chino:
Yes, I would like to replace the isolcpus method by working with the manager to achieve the same result.  I was only describing what we currently do as an example.  

The main point is that we want to lock a specific app "thread" to a core and not have anything else but that thread on the core.  

We don't know what the thread-id is until the process starts so we have a wrapper script that handles that in the container.

ConnorDoyle:
@chino by coincidence, a similar point just came up in the resource management workgroup weekly while discussing constraining cache partitions for garbage collector threads in a JVM.

How about exploring the other direction of communication first? That is, Kubelet/CPU Manager could communicate down to the container the intersection of its CPU mask and isolcpus? That way the subprocess or thread could continue to set its own affinity without needing to call back into the CRI or Kubelet. The container would gain the advantage of allowing the CPU Manager to track the assigned cpusets, therefore guaranteeing that it won't accidentally collide with another user of an isolated CPU.

My gut reaction is that implementing a bidirectional API for Kubelet/CRI to collaborate with the user application for these settings would get pretty complicated, but maybe we just haven't seen the right abstractions for it yet.

derekwaynecarr:
@ConnorDoyle @chino -- the cpu assigned by the kubelet to a container needs to be discoverable via downward API (should be in this spec, if not there yet).

chino:
@ConnorDoyle 

Yes, exactly.  I figured the simplest solution which wouldn't introduce too much complexity would be to `reserve a core BUT not assign it`.  Meaning the Manager will pull a core out from scheduling (hopefully disabling interrupts can become an option too) and then our container will use taskset it self to pin the specific thread-id which is only known at runtime. 

We generally have 1 core on each socket set to handle interrupts and os/background threads.  Sometimes we need to leave a few extra free to handle the interrupts on them fast enough for disk and what not.  Obviously in a true multi-tenant cluster it would be nice to leave cores for os/k8 and then also group up cores into "shared pools" between certain pods and as noted "exclusive cores" for specific containers/threads.

It could be interesting if there was someway to specify how to find the right thread to pin but it seems complicated and in our case that could be java (hence need to use jstack) or could be a native application where you need to use another method to identify the right thread.

Another issue we currently have is we'd like to share ipc namespace and mmap files across sets of pods in the same tenant.  Right now the only way to do that is to share it to the entire host which isn't very multi-tenant friendly/secure.

contributors/design-proposals/cpu-manager.md
derekwaynecarr: its probably ok, i have too much history in the discussion that as phrased here, it signals something deeper.

chino:
@ConnorDoyle I like that idea, as long as we can specify easily shared vs dedicated.  We previously used a very simple set of cgroups pre-docker but container runtimes create a whole set of different cgroup hierarchies and sometimes the container might not have write access.  If something like cgroup namespaces or something formal can make this process easy to manage inside the private region than that's great.  Again to be clear, I think multiple pods will want to share the same cores at times for background threads because there isn't enough cores to go around and controlling interrupt settings will be important.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,343 @@
...
+1. Poor or unpredictable performance observed compared to virtual machine
+   based orchestration systems. Application latency and lower CPU
+   throughput compared to VMs due to cpu quota being fulfilled across all
+   cores, rather than exclusive cores, which results in fewer context
+   switches and higher cache affinity.
+1. Unacceptable latency attributed to the OS process scheduler, especially
+   for “fast” virtual network functions (want to approach line rate on
+   modern server NICs.)
+
+_Solution requirements:_
```
derekwaynecarr: add a requirement that a container must be able to discover the set of cpus it has been assigned.

chino:
@derekparker yes, that would be great

The main thing that worries me is we literally turn off interrupts and even high performance tick timers.  We cannot have the top level container pid touching the core at all.  Hence why I asked if we could dedicate but not assign anything to the core so that we can taskset our selves and know that it's completely void of anything running on it.

When you work with low latency apps in the nano/micros time scale then any interrupt will cause measurable blips and jitter in the otherwise steady message rates.  Especially in distributed applications this can cause race conditions and outliers.

contributors/design-proposals/cpu-manager.md
sjenning: I'm not sure I agree that this should be a requirement.  I see the value, where people are wanting to do thread affinity inside the pod, but it comes with baggage.

It pins us in when we go to add the dynamic cpuset manager later.

It also creates problem in the kubelet restart case.  If our contract is "I'll put you on an exclusive core", we have to freedom to "reschedule" on kubelet restart rather than writing a nasty cgroup parser to reconstitute the cpuset manager state.

chino: Reschedule as in possibly shift a container to a new core after it's already been running? That would be rather scary for many workloads\users if that happened.

The cpuset is just a bit flag which is easy to parse?

Could the assigned cores be saved to etcd or a local cache so it's recoverable perhaps?

ConnorDoyle: @chino the CPU mask would only be changed after process start if the dynamic CPU Manager policy is enabled. The static policy guarantees never to do that.

@derekwaynecarr and @sjenning: Inside the container, workloads that care about their CPU mask can easily get it by reading `/proc/self/status` (see below). Is that sufficient?

```
connor:~/ $ grep -i cpus /proc/self/status
Cpus_allowed:   77
Cpus_allowed_list:      0-2,4-6
```

vishh:
APologies for this delay. I'll review this proposal next week.

contributors/design-proposals/cpu-manager.md
derekwaynecarr: For a dynamic policy, we should send a send a signal.  For static, I agree it's not needed.

ConnorDoyle: Backfilling from off-list discussion: we talked today about providing a signal in the following way. We could project (a subset of) the CPU manager state into a volume visible to selected containers. User workloads could subscribe to update events in a normal Linux manner (e.g. inotify.) We also decided, more or less, that this won't be necessary until the dynamic policy is implemented. We can document that user containers in the shared pool must avoid setting their own CPU affinity, since it will be overwritten without notice any time the membership of the shared pool changes.

fabiand:
I do not think that it's required in a first step, but: Some applications need a way to specify how threads are assigned to cores, to optimize performance. qemu/libvirt is one such example. Thus it might be valuable if at some point an workload can express how it expects the threads to be pinned to cores.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,348 @@
...
+}
+```
+
+##### Example pod specs and interpretation
+
+| Pod                                        | Interpretation                 |
+| ------------------------------------------ | ------------------------------ |
+| Pod [Guaranteed]:<br />&emsp;A:<br />&emsp;&emsp;cpu: 0.5 | Container **A** is assigned to the shared cpuset. |
+| Pod [Guaranteed]:<br />&emsp;A:<br />&emsp;&emsp;cpu: 2.0 | Container **A** is assigned two sibling threads on the same physical core (HT) or two physical cores on the same socket (no HT.)<br /><br /> The shared cpuset is shrunk to  make room for the exclusively allocated CPUs. |
+| Pod [Guaranteed]:<br />&emsp;A:<br />&emsp;&emsp;cpu: 1.0<br />&emsp;A:<br />&emsp;&emsp;cpu: 0.5 | Container **A** is assigned one exclusive CPU and container **B** is assigned to the shared cpuset. |
```
teferi: nit: here and next line Pod column mentions 2 *A* containers. easy to spot on rendered page, but not here

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,348 @@
...
+threads share a number of hardware resources including the cache
+hierarchy. On multi-socket systems, logical CPUs co-resident on a socket
+share L3 cache. Although there may be some programs that benefit from
+disjoint caches, the policies described in this proposal assume cache
+affinity will yield better application and overall system performance for
+most cases. In all scenarios described below, we prefer to acquire logical
+CPUs topologically. For example, allocating two CPUs on a system that has
+hyper-threading turned on yields both sibling threads on the same
+physical core. Likewise, allocating two CPUs on a non-hyper-threaded
+system yields two cores on the same socket.
```
teferi: What should happen if there are enough CPUs on a node, but they're far away from each other (I'm requesting 2 CPUs, but there ones left are on different cores). For some workloads admitting such Pod may be acceptable, for some not. Would it be a good idea to make this behaviour configurable?

contributors/design-proposals/cpu-manager.md
ConnorDoyle: Thank you, will fix.

contributors/design-proposals/cpu-manager.md
ConnorDoyle: For now, our plan is to have the first policy (static) provide the "best" allocation when assigning dedicated cores. So, at higher G pod density it would be possible to observe the topology fragmentation you described.

If it were configurable in the policy, do you have ideas about how you would like that to look (name for the knob, values, defaults etc.)?

teferi: The more I think about it — the more I realize how complicated things actually are =/
Generally my perception is that **by default** CPUs of a single container inside a Pod should be as close to each other as possible, while the CPUs of a different containers inside a Pod should be as far as possible from each other. The only case I can imagine when I would want to pack as much dedicated containers on a single socket as possible is when I would want to launch multiple small (1-CPU) containers and then launch a single giant one, that asks for half the CPUs of a core.

For a container if we want to allow to control how close the CPUs are, I imagine it might be a double parameter in the pod spec. Smth like:
```
corePolicy:
  placement: dense|separate
  strict: true|false   
```
with dense and false as defaults. `separate` would mean to attempt to spread CPUs as much as possible and `dense` would mean to pack as close as possible. With `strict: false` meaning that workload can tolerate if the requirements are not met. And `strict: true` on the other side should mean that the container has to be rescheduled. 

And then there is a whole new level with HT enabled. I can imagine smth like:
```
coreThreadPolicy:
  placement: dense|separate|separate_restrict
  strict: true|false
```
with `dense` and `false` as defaults. `dense` here would mean allow giving a container multiple sibling CPUs. `separate` would mean attempt to spread CPUs across non-siblings. `spearate_restict` would mean spread CPUs across non-siblings and reserve the sibling, so that it is not assigned to anyone.
I'm afraid that I might be over-thinking the whole problem =)

As a small disclaimer I'd like to mention that I come from OpenStack, so my perception migh be tainted with how OS does it =) [Here](https://specs.openstack.org/openstack/nova-specs/specs/juno/approved/virt-driver-cpu-pinning.html#proposed-change) is a link to a doc that describes how OS approaches similar problems.
(cc @ashish-billore)

flyingcougar: @teferi I like the idea of `corePolicy`. 
As for `coreThreadPolicy` - isn't it a little bit too much ? i.e. not all environments are homogeneous, some nodes may not have HT enabled (probably not much). Besides `separate_restrict` can be accomplished by requesting i.e. 2 cores. If node has HT and 
```
corePolicy:
  placement: dense
  strict: true
```
then it would result in 2 CPUs on 1 physical core.

fabiand:
Let me just raise our use-case: When spawning VMs it is important to map the virtual topology to the real one, to achieve optimal performance.
I wonder, if this proposal should cover a mechanism which allows a container to tell the manager how to place the processes/threads on the real hardware.

/cc @berrange

contributors/design-proposals/cpu-manager.md
teferi: @flyingcougar separate_restrict is more about not allowing others to use core's sibling. It probably would be very difficult to implement, since k8s would still think that we have those cores available for scheduling. So anyway it was probably a weird idea.

`corePolicy` is about packing the CPUs as tightly as possible on cores in the same socket, while `coreThreadPolicy` is about packing CPUs as tightly as possible within one core. So imo both of those have their uses. I think I need to draw a couple of diagrams to illustrate these ideas...

teferi: cc @fabiand I guess you should participate in ^^ part of the discussion

fabiand: > The more I think about it — the more I realize how complicated things actually are =/

Yes :)


I'm not an expert on this domain, but …
It's pretty clear that we - kubevirt - need the 'strict dense placement' option as a first step.
I actually wonder if additional differentiations are needed: Either you care, and you require an optimal placement, or you don't and get best effort.

Some workloads, like kubevirt (again), might want to manage parts of numa nodes themselfes, in a more fine granular fashion. In those cases it might make sense, to be able to tell the kubelet, that some nodes, sockets, core, should be reserved/exclusive to a specific container. A process can then take those reserved parts and manage them as needed (i.e. in our case, libvirt would take care of managing those nodes).
Long story short, we might want a mechanism to lend nodes to a process.

Another thing is that in future we need - and thus should already think of - to combine the placement with ResourceClasses #782, because in reality you want a process close to the device it uses, to avoid numa-node remote memory access.
Closely related to this is actually also the relationship to huge-pages.

I do not think that we need to solve all of this now, but we should understand the picture, to not create something which will already contradict with what we see coming up.

@berrange @mpolednik thoughts?

sjenning:
@derekwaynecarr @vishh @ConnorDoyle is there any objection with merging this proposal?  It seems feedback has tailed off and I don't see comments that express show-stopping concerns.  Mostly just indication that it doesn't go far enough.  But we have to start somewhere!

contributors/design-proposals/cpu-manager.md
ConnorDoyle: @fabiand there's exactly that topic on the resource management workgroup agenda for this morning. Besides the device binding logic and the hugetlb controller settings, CNI plugins could also take topology into account on a multi-socket system with multiple network interfaces. Hopefully we can get to an agreement that a solution is needed to unify affinity somehow in the v1.9 or v1.10 releases.

On the topic of the CPU manager itself, policies that operators want can get really complicated. Hopefully we can get a good default scoped and implemented first and then extend with more advanced policies. In the resource management face-to-face where this component was conceived, we had a pretty long discussion about implementing a more flexible policy driven by explicit pool configuration. Maybe we could think of ways to represent the behaviors needed for the advanced cases in the config domain. In my mind, a rich configurable policy may buy enough flexibility so that we don't need another plugin system for external policies. At the same time, for v1.8 at least, changes to the pod spec API are probably out of scope.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,348 @@
...
+  * The CPU Manager must be able to pick up where it left off in case the
+    Kubelet restarts for any reason.
+* Read effective CPU assinments at runtime for alerting. This could be
+  satisfied by the checkpointing requirement.
+* Configuration
+  * How does the CPU Manager coexist with existing kube-reserved
+    settings?
+  * How does the CPU Manager coexist with related Linux kernel
+    configuration (e.g. `isolcpus`.) The operator may want to specify a
+    low-water-mark for the size of the shared cpuset. The operator may
```
ConnorDoyle: In the PoC branch we have the static policy unconditionally reserving a core for infra processes, with banning BE if the shared pool becomes empty as a TODO. However, containers in the Burstable class also might lack cpu requests (a single-container pod that requires only memory is in the Burstable class.)

@sjenning what do you think about replacing the notion of infra-reserved cores with a min size for the shared pool in the static policy? One thing I wouldn't want is to effectively disable a significant fraction of available threads on small systems (e.g. 4 or 8 cores.)

sjenning: @ConnorDoyle I was thinking we would remove `ceiling(kube-reserved cpu + system-reserved cpu)` cores from the initial cpu pool with core 0 never being in the pool.  We do allow the shared pool to go to zero, at which point we evict BE pod and Burstable pods with no CPU requests, and set a CPUPressure condition true on the node that prevents the scheduler from assigning pods that don't have a cpu request.

But you are talking about the case where the user wants to ensure that the shared pool does _not_ go to zero.  I guess that value can't be implied by any tunable the kubelet exposes today.  That might have to be a new one.

Does that make sense? (not rhetorical)  Did I address your question?  My brain has been everywhere lately.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,307 @@
...
+    want to correlate exclusive cores with the isolated CPUs, in which
+    case the strategy outlined above where allocations are taken
+    directly from the shared pool is too simplistic. We could allow an
+    explicit pool of cores that may be exclusively allocated and default
+    this to the shared pool (leaving at least one core fro the shared
```
balajismaniam: s/fro/for

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,307 @@
...
+    cpuset to be used for OS, infra and non-exclusive containers.
+
+## Practical challenges
+
+1. Synchronizing CPU Manager state with the container runtime via the
+   CRI. Runc/libcontainer allows container cgroup settings to be updtaed
```
balajismaniam: s/updtaed/updated

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,307 @@
...
+* Performance metrics for one or more plausible synthetic workloads show
+  benefit over no-op policy.
+
+### Phase 5: NUMA
+
+* Kubelet can discover "advanced" CPU topology (NUMA).
+
+## Appendix A: cpuset pitfalls
+
+1. `cpuset.sched_relax_domain_level`
```
balajismaniam: Might be useful to add a link and a gist of the effect.

ConnorDoyle:
xref https://github.com/kubernetes/community/pull/654

contributors/design-proposals/cpu-manager.md
```diff
@@ -158,6 +163,26 @@ node. Once allocated at pod admission time, an exclusive CPU remains
 assigned to a single container for the lifetime of the pod (until it
 becomes terminal.)
 
+##### Configuration
+
+Operators can set the number of CPUs that pods may run on through a new
+Kubelet configuration setting `--cpu-manager-static-num-cpus`, which
```
flyingcougar: `--cpu-manager-static-num-cpus` suggests that it's valid only for **static** policy - it that the case ?

ConnorDoyle: Yes

sjenning: This seem in conflict with CPU Allocatable.  If I set `--cpu-manager-static-num-cpus=2` on a node with CPU Allocatable of 4, the scheduler thinks there are 4 allocatable cores.  However if it tries to schedule more that 2 cores worth of pinned pods, those pods... bounce? It isn't clear what happens.

What is the motivation for exposing this to users?  I thought we talked about implicitly setting this to `Capacity - Kube Resv - System Resv`.

ConnorDoyle: Linking parallel discussion in @flyingcougar 's PR (https://github.com/ConnorDoyle/kubernetes/pull/5).

Will update this to go back to using Allocatable for this value, agree it makes more sense now. Sorry for the confusion.

ConnorDoyle: Rewrite available for review: https://github.com/ConnorDoyle/community/pull/1/files

contributors/design-proposals/cpu-manager.md
derekwaynecarr: catching up on doc, so this may be covered elsewhere, but we should say something about signalling in a future iteration of this proposal.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+#### Configuring the CPU Manager
+
+Kubernetes will ship with three CPU manager policies. Only one policy is
+active at a time on a given node, chosen by the operator via Kubelet
+configuration. The three policies are **noop**, **static** and **dynamic**.
+
+The active CPU manager policy is set through a new Kubelet
+configuration value `--cpu-manager-policy`. The default value is `noop`.
+
+The number of CPUs that pods may run on is set using the existing
```
sjenning: nit: s/is set/can be implicitly controlled/

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+type CPUSet map[int]struct{} // set operations and parsing/formatting helpers
+
+type CPUTopology TBD
+```
+
+#### Configuring the CPU Manager
+
+Kubernetes will ship with three CPU manager policies. Only one policy is
+active at a time on a given node, chosen by the operator via Kubelet
+configuration. The three policies are **noop**, **static** and **dynamic**.
```
derekwaynecarr: prefer we s/noop/none

derekwaynecarr: mostly because i am not sure how well noop is understand to non-native english speakers.  it reads as "nupe"

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+The active CPU manager policy is set through a new Kubelet
+configuration value `--cpu-manager-policy`. The default value is `noop`.
+
+The number of CPUs that pods may run on is set using the existing
+node-allocatable configuration settings. See the [node allocatable proposal
+document][node-allocatable] for details. The CPU manager will claim
+`ceiling(node.status.allocatable.cpu)` as the number of CPUs available to
+assign to pods, starting from the highest-numbered physical core and
+descending topologically. It is recommended to configure an integer value for
+`node.status.allocatable.cpu` when the CPU manager is enabled.
```
sjenning: Not sure how to say this more precisely, but I think this might be better as the node allocatable CPU is not explicitly set by the user:

It is recommended that the user configure integer CPU values for `kube-reserved` and/or `system-reserved` as this will result in an integer value for `node.status.allocatable.cpu`.  This avoids a fractional core existing in both the pod shared pool and kube/system reserved scheduling domains.

contributors/design-proposals/cpu-manager.md
ConnorDoyle: OK to change it.

contributors/design-proposals/cpu-manager.md
ConnorDoyle: Thanks, let's add exactly that text. It's a big improvement.

contributors/design-proposals/cpu-manager.md
ConnorDoyle: Will add it to the dynamic policy section.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+```
+
+#### Configuring the CPU Manager
+
+Kubernetes will ship with three CPU manager policies. Only one policy is
+active at a time on a given node, chosen by the operator via Kubelet
+configuration. The three policies are **noop**, **static** and **dynamic**.
+
+The active CPU manager policy is set through a new Kubelet
+configuration value `--cpu-manager-policy`. The default value is `noop`.
```
derekwaynecarr: Just recording here for future awareness, but I prefer we not prefix the name of this flag with experimental and instead state similar to node-labels that specific options are Alpha feature.

contributors/design-proposals/cpu-manager.md
derekwaynecarr: its really the sum of kube-reserved and system-reserved should be integral.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+node-allocatable configuration settings. See the [node allocatable proposal
+document][node-allocatable] for details. The CPU manager will claim
+`ceiling(node.status.allocatable.cpu)` as the number of CPUs available to
+assign to pods, starting from the highest-numbered physical core and
+descending topologically. It is recommended to configure an integer value for
+`node.status.allocatable.cpu` when the CPU manager is enabled.
+
+Operator documentation will be updated to explain how to configure the
+system to use the low-numbered physical cores for kube-reserved and
+system-reserved slices.
```
derekwaynecarr: nit: s/slices/cgroup in case users are not on systemd.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+       topological-best-fit.
+    1. Static policy updates the state, adding an assignment for the new
+       container and removing those CPUs from the default pool.
+    1. CPU manager reads container assignment from the state.
+    1. CPU manager updates the container resources via the CRI.
+    1. Kuberuntime calls the CRI delegate to start the container.
+
+1. _A container that was assigned exclusive cores terminates._
+    1. Kuberuntime unregisters the container with the CPU manager.
+    1. CPU manager unregisters the contaner with the static policy.
```
derekwaynecarr: nit: container

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+    1. Static policy adds the container's assigned CPUs back to the default
+       pool.
+    1. Kuberuntime calls the CRI delegate to remove the container.
+    1. Asynchronously, the CPU manager's reconcile loop updates the
+       cpuset for all containers running in the shared pool.
+
+1. _The shared pool becomes empty._
```
derekwaynecarr: we previously discussed a node condition for cpu pressure.  is the taint the replacement?

derekwaynecarr: for pre-emption, will the taint be understood?

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+| ------------------------------------------ | ------------------------------ |
+|                                            |                                |
+|                                            |                                |
+
+## Operations and observability
+
+* Checkpointing assignments
+  * The CPU Manager must be able to pick up where it left off in case the
+    Kubelet restarts for any reason.
+* Read effective CPU assinments at runtime for alerting. This could be
```
derekwaynecarr: assignments

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+  satisfied by the checkpointing requirement.
+
+## Practical challenges
+
+1. Synchronizing CPU Manager state with the container runtime via the
+   CRI. Runc/libcontainer allows container cgroup settings to be updated
+   after creation, but neither the Kubelet docker shim nor the CRI
+   implement a similar interface.
+    1. Mitigation: [PR 46105](https://github.com/kubernetes/kubernetes/pull/46105)
```
derekwaynecarr: this is needed independent of cpu policy.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+1. Compatibility with the `isolcpus` Linux kernel boot parameter. The operator
+   may want to correlate exclusive cores with the isolated CPUs, in which
+   case the static policy outlined above, where allocations are taken
+   directly from the shared pool, is too simplistic.
+    1. Mitigation: defer supporting this until a new policy tailored for
+       use with `isolcpus` can be added.
+
+## Implementation roadmap
+
+### Phase 1: No-op policy
```
derekwaynecarr: make clear kube 1.8 targets phase 1 and 2.

derekwaynecarr:
@ConnorDoyle @sjenning -- we need to capture what happens when all cpus are used, and make clear that we need to evict a BE pod if already resident on the node.

contributors/design-proposals/cpu-manager.md
ConnorDoyle: Apologies, lost track of the progress to convert node conditions to taints. Will update.

contributors/design-proposals/cpu-manager.md
ConnorDoyle: Any update required in this text? (what are the other things that need it, VPA?)

ConnorDoyle:
@derekwaynecarr updated

contributors/design-proposals/cpu-manager.md
ConnorDoyle: updated

contributors/design-proposals/cpu-manager.md
ConnorDoyle: added

ConnorDoyle:
@derekwaynecarr and @vishh (approvers per the rm working group planning spreadsheet),

Aside from the node condition details that @sjenning is currently working to add, are there any other blockers to accept this proposal?

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+
+The active CPU manager policy is set through a new Kubelet
+configuration value `--cpu-manager-policy`. The default value is `noop`.
+
+The number of CPUs that pods may run on is set using the existing
+node-allocatable configuration settings. See the [node allocatable proposal
+document][node-allocatable] for details. The CPU manager will claim
+`ceiling(node.status.allocatable.cpu)` as the number of CPUs available to
+assign to pods, starting from the highest-numbered physical core and
+descending topologically. It is recommended to configure an integer value for
```
vikaschoudhary16: @ConnorDoyle Can you please give an example here illustrating with sample topology,  `kube-reserved`, `system-reserved` and `node.status.allocatable.cpu`?

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+`node.status.allocatable.cpu` when the CPU manager is enabled.
+
+Operator documentation will be updated to explain how to configure the
+system to use the low-numbered physical cores for kube-reserved and
+system-reserved slices.
+
```
vikaschoudhary16: @ConnorDoyle Will it make sense to add minimal details in this document as well for example "user-story"?

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+    1. Kuberuntime calls the CRI delegate to remove the container.
+    1. Asynchronously, the CPU manager's reconcile loop updates the
+       cpuset for all containers running in the shared pool.
+
+1. _The shared pool becomes empty._
+    1. The CPU manager adds a taint with effect NoSchedule, NoExecute
+       that prevents BestEffort and Burstable QoS class pods from
+       running on the node.
+
+1. _The shared pool becomes nonempty._
```
vikaschoudhary16: nit: numbering.

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+    1. Static policy adds the container's assigned CPUs back to the default
+       pool.
+    1. Kuberuntime calls the CRI delegate to remove the container.
+    1. Asynchronously, the CPU manager's reconcile loop updates the
+       cpuset for all containers running in the shared pool.
+
+1. _The shared pool becomes empty._
+    1. The CPU manager adds a taint with effect NoSchedule, NoExecute
+       that prevents BestEffort and Burstable QoS class pods from
+       running on the node.
```
vikaschoudhary16: is reading the state and updating the container resources via CRI synchronous with "start the container"? If yes, how kuberuntime gets to know when to invoke "start"?

contributors/design-proposals/cpu-manager.md
```diff
@@ -0,0 +1,366 @@
...
+1. Provide an API-driven contract from the system to a user: "if you are a
+   Guaranteed pod with 1 or more cores of cpu, the system will try to make
+   sure that the pod gets its cpu quota primarily from reserved core(s),
+   resulting in fewer context switches and higher cache affinity".
+1. Support the case where in a given pod, one container is latency-critical
+   and another is not (e.g. auxillary side-car containers responsible for
+   log forwarding, metrics collection and the like.)
+1. Do not cap CPU quota for guaranteed containers that are granted
+   exclusive cores, since that would be antithetical to (1) above.
+1. Take physical processor topology into account in the CPU affinity policy.
```
sjenning: This is github markdown.  When viewed as markdown, not source, the numbering is correct.

contributors/design-proposals/cpu-manager.md
sjenning: same as before re: github markdown

derekwaynecarr:
@ConnorDoyle -- i will take one final pass by eod tomorrow.  right now, this generally lgtm.

derekwaynecarr:
@ConnorDoyle @sjenning -- please make sure to keep updated with implementation progress, but I think this proposal is fine to proceed implementing.  Thanks!

/lgtm
/approve

sjenning:
@derekwaynecarr looks like merging is manual in this repo.  Should I just git the Big Green Button?

resouer:
@sjenning Yes, feel free to manually merge this :)

cliffburdick:
@sjenning @derekwaynecarr @vishh I believe your PR covers this, but the EPYC processor from AMD has some new (somewhat odd) differences between the Xeons when it comes to core affinity and NUMA. You can see in the diagram here that there are multiple latency paths that are possible when you cross NUMA boundaries, depending on which core originated the request:

http://www.anandtech.com/show/11544/intel-skylake-ep-vs-amd-epyc-7000-cpu-battle-of-the-decade/2

sjenning:
@cliffburdick it is true that LLC latency is not uniform among cores on the same socket for the new AMD CCX based processors.  If the a CCX must access a remote CCXs LLC, there is additional latency.  It is basically a topology level between cores and sockets.

Note this is only an issue if you have two threads sharing data.  If the working sets are independent, this is not an issue.

Right now, we can't make the distinction as cadvisor does not make the distinction.  But this is a last 5% optimization in my mind.  The existing cpuset allocator should work well, even without explicit knowledge of the CCXs.

jeremyeder:
@cliffburdick there's some historical context here, too.  We agreed to defer a bunch of what this feature was originally spec'd out to be, out to a later release.  That includes some of the more advanced topology detection and affinity (incl. PCI locality).  It is a roadmap item for the working group.  From RH perspective and I would say others too, it is a very important item to tackle to make sure this cpumanager delivers on the intended use-cases.

The compromise we made in order to even get this far was that the complexity of topology discovery and pinning would be hidden behind QoS tiers.  So users request a QoS tier and a core count in whole integers (rather than millicores) to get the performance they want.

I would love it if you have more concrete requirements that you write them up, send them to our mailing list and hopefully participate directly in the next resource management working group meeting.

You can find all of the mailing list, meeting and agenda details here:
https://github.com/kubernetes/community/tree/master/wg-resource-management#contact-us

cliffburdick:
Hi @jeremyeder, I'd be glad to list a couple here, and some of it is based on your work. We're running DPDK in Docker and have IO cards that should be affined to the same NUMA zone/CPU. For DPDK, isolcpus is the first obvious thing to turn on for all the packet processing and worker threads; we do not want any Linux housekeeping or other applications floating there. I've seen some recent threads on other message boards related to using isolcpus with containers, and it appears that it either works, or there are workaround to make it work. Taking it a step further, Intel recommends disabling the RCU threads on any isolcpus cores. I don't believe this changes anything, but it's another thing enabled on the kernel boot line. Lastly I wanted to experiment with the tickless kernel since many applications would benefit greatly by a single thread per core. I don't know how well this will work in a container since I haven't tried it, but I believe if two threads try to run on the same core, the kernel will revert to the normal scheduler for that core. I think it should work based on looking at the processes started by Docker, though.

In my case, I almost never will be crossing the NUMA boundary. All of the work will come in a NIC attached to a single processor, and all of the work will be done on that processor or accelerator cards attached to that processor before going back out the same NIC.

Of course, DPDK has other implications as well, which I'm currently working through.

ConnorDoyle:
@cliffburdick this is a fantastic discussion, but the comments in a closed PR is the wrong place for it. I'm afraid all this good info will be lost. Could we please move to the ML?

cliffburdick:
@ConnorDoyle sure. Thanks.

Issue #634, opened 2017-05-18
Allow doc generator to work with custom content


jamiehannaford:
As suggested in https://github.com/kubernetes/community/pull/577#discussion_r114610769.

We will need some kind of expected delimiter, for example `<!-- END OF GENERATED CONTENT -->`. Any content after this point is always appended to generated content.

PR #655, opened 2017-05-23
Add custom content blocks to generator

jamiehannaford:
Fixes https://github.com/kubernetes/community/issues/634.

cc/ @saad-ali @jbeda

### DIFF
--- a/generator/app.go
+++ b/generator/app.go
@@ -31,15 +31,17 @@ import (
 )
 
 var (
-	sigsYamlFile    = "sigs.yaml"
-	templateDir     = "generator"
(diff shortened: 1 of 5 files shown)


generator/app.go
```diff
@@ -105,33 +101,81 @@ func createDirIfNotExists(path string) error {
...
+	content, err := getExistingContent(outputPath)
+	if err != nil {
+		return err
+	}
+
+	// ensure file is empty
 	f.Truncate(0)
 
-	// write template output to file
+	// generated content
```
jbeda: Nit: I would have passed existing content into the template vs. having three functions append to the same file.  But this works well too.

PR #306, opened 2017-01-30
[Proposal] Improve Local Storage Management

vishh:
A note to reviewers: A detailed design proposal will be posted once the overall design in this proposal has been accepted by the community. So kindly hold on to specific design questions or suggestions that are not relevant to the overall high level design for local storage. 

cc @kubernetes/sig-storage-proposals  @kubernetes/sig-node-proposals @kubernetes/sig-apps-proposals @kubernetes/sig-scheduling-proposals 

cc @msau42  

For https://github.com/kubernetes/features/issues/121

TODO:
- [x] Handle scheduling of `runtime` primary partition capacity
- [x] Propose design choices for mitigating or avoiding Disk IO issues for system daemons, container logs and writable layer. EmptyDir IO isolation is out of scope because local PV can be used instead.

PR #657, opened 2017-05-23
Proposal for growing persistent volumes

gnufied:
cc @kubernetes/sig-storage-pr-reviews @liggitt @eparis

### DIFF
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -66,11 +66,11 @@ For volume types that only require volume plugin based api call, this will be on
 A new controller called `volume_expand_controller` will listen for pvc size expansion requests and take action as needed. The steps performed in this
 new controller will be:
 
-* Watch for pvc update requests and add pvc to controller's desired state of world if a increase in volume size was requested. Once PVC is added to
-  controller's desired state of world - `pvc.Status.Conditions` will be updated with `ResizeStarted: True`.
+* Watch for pvc update requests and add pvc to controller's work queue if a increase in volume size was requested. Once PVC is added to
+  controller's work queue - `pvc.Status.Conditions` will be updated with `ResizeStarted: True`.
 * For unbound or pending PVCs - resize will trigger no action in `volume_expand_controller`.
 * If `pv.Spec.Capacity` already is of size greater or equal than requested size, similarly no action will be perfomed by the controller.
-* A reconciler will read desired state of world and perform corresponding volume resize operation. If there is a resize operation in progress
+* A separate goroutine will read work queue and perform corresponding volume resize operation. If there is a resize operation in progress
   for same volume then resize request will be pending and retried once previous resize request has completed.
 * Controller resize in effect will be level based rather than edge based. If there are more than one pending resize request for same PVC then
   new resize requests for same PVC will replace older pending request.
@@ -80,30 +80,38 @@ new controller will be:
 ```go
 func (og *operationGenerator) GenerateExpandVolumeFunc(
     pvcWithResizeRequest *expandcache.PvcWithResizeRequest,
-    dsow expandcache.DesiredStateOfWorld) (func() error, error) {
+    resizeMap expandcache.VolumeResizeMap) (func() error, error) {
 
     volumePlugin, err := og.volumePluginMgr.FindExpandablePluginBySpec(pvcWithResizeRequest.VolumeSpec)
+    expanderPlugin, err := volumePlugin.NewExpander(pvcWithResizeRequest.VolumeSpec)
 
-    if err != nil {
-        return nil, fmt.Errorf("Error finding plugin for expanding volume: %q with error %v", pvcWithResizeRequest.UniquePvcKey(), err)
-    }
-
-    expanderPlugin, err := volumePlugin.NewExpander()
-
-    if err != nil {
-        return nil, fmt.Errorf("Error creating expander plugin for volume %q with error %v", pvcWithResizeRequest.UniquePvcKey(), err)
-    }
 
     expandFunc := func() error {
-        expandErr := expanderPlugin.ExpandVolumeDevice(pvcWithResizeRequest.VolumeSpec, pvcWithResizeRequest.ExpectedSize, pvcWithResizeRequest.CurrentSize)
+        expandErr := expanderPlugin.ExpandVolumeDevice(pvcWithResizeRequest.ExpectedSize, pvcWithResizeRequest.CurrentSize)
 
         if expandErr != nil {
-            glog.Errorf("Error expanding volume through cloudprovider : %v", expandErr)
+            og.recorder.Eventf(pvcWithResizeRequest.PVC, v1.EventTypeWarning, kevents.VolumeResizeFailed, expandErr.Error())
+            resizeMap.MarkResizeFailed(pvcWithResizeRequest, expandErr.Error())
             return expandErr
         }
-        dsow.MarkAsResized(pvcWithResizeRequest)
 
+        // CloudProvider resize succeded - lets mark api objects as resized
+        if expanderPlugin.RequiresFSResize() {
+            err := resizeMap.MarkForFileSystemResize(pvcWithResizeRequest)
+            if err != nil {
+                og.recorder.Eventf(pvcWithResizeRequest.PVC, v1.EventTypeWarning, kevents.VolumeResizeFailed, err.Error())
+                return err
+            }
+        } else {
+            err := resizeMap.MarkAsResized(pvcWithResizeRequest)
+
+            if err != nil {
+                og.recorder.Eventf(pvcWithResizeRequest.PVC, v1.EventTypeWarning, kevents.VolumeResizeFailed, err.Error())
+                return err
+            }
+        }
         return nil
+
     }
     return expandFunc, nil
 }
@@ -121,7 +129,7 @@ func (og *operationGenerator) GenerateExpandVolumeFunc(
   successful.
 
 * To consider cases of missed PVC update events, an additional loop will reconcile bound PVCs with PVs. This additional loop will loop through all PVCs
-  and match `pvc.spec.resources.requests` with `pv.spec.capacity` and add PVC in `volume_expand_controller`'s desired state of world if `pv.spec.capacity` is less
+  and match `pvc.spec.resources.requests` with `pv.spec.capacity` and add PVC in `volume_expand_controller`'s work queue if `pv.spec.capacity` is less
   than `pvc.spec.resources.requests`.
 
 * There will be additional checks in controller that grows PV size - to ensure that we do not make volume plugin API calls that can reduce size of PV.
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -25,12 +25,12 @@ Enable users to increase size of PVs that their pods are using. The user will up
 | ----------------| :---------------: | :--------------------------:| :----------------------: |
 | EBS             | Yes               | Yes                         | Yes                      |
 | GCE PD          | Yes               | Yes                         | Yes                      |
-| Azure Disk      | Yes               | Yes                         | No                       |
+| GlusterFS       | Yes               | No                          | Yes                      |
 | Cinder          | Yes               | Yes                         | Yes                      |
 | Vsphere         | Yes               | Yes                         | No                       |
 | Ceph RBD        | Yes               | Yes                         | No                       |
 | Host Path       | No                | No                          | No                       |
-| GlusterFS       | Yes               | No                          | Yes                      |
+| Azure Disk      | Yes               | Yes                         | No                       |
 | Azure File      | No                | No                          | No                       |
 | Cephfs          | No                | No                          | No                       |
 | NFS             | No                | No                          | No                       |
@@ -63,13 +63,15 @@ For volume types that only require volume plugin based api call, this will be on
 A new controller called `volume_expand_controller` will listen for pvc size expansion requests and take action as needed. The steps performed in this
 new controller will be:
 
-* Watch for pvc update requests and add pvc to controller's desired state of world if a increase in volume size was requested.
+* Watch for pvc update requests and add pvc to controller's desired state of world if a increase in volume size was requested. Once PVC is added to
+  controller's desired state of world - `pvc.Status.Conditions` will be updated with `ResizeStarted: True`.
+* For unbound or pending PVCs - resize will trigger no action in `volume_expand_controller`.
 * A reconciler will read desired state of world and perform corresponding volume resize operation. If there is a resize operation in progress
   for same volume then resize request will be pending and retried once previous resize request has completed.
 * Controller resize in effect will be level based rather than edge based. If there are more than one pending resize request for same PVC then
   new resize requests for same PVC will replace older pending request.
 * Resize will be performed via volume plugin interface, executed inside a goroutine spawned by `operation_exectutor`.
-* A new plugin interface called `volume.Exander` will be added to volume plugin interface. The controller call to expand the PVC will look like:
+* A new plugin interface called `volume.Expander` will be added to volume plugin interface. The controller call to expand the PVC will look like:
 
 ```go
 func (og *operationGenerator) GenerateExpandVolumeFunc(
@@ -103,11 +105,11 @@ func (og *operationGenerator) GenerateExpandVolumeFunc(
 }
 ```
 
-* Once volume expand is successful, the volume will be marked as expanded and new size will be updated in `pv.spec.capacity`. Any errors will be
-reported as *events* on PVC object.
+* Once volume expand is successful, the volume will be marked as expanded and new size will be updated in `pv.spec.capacity`. Any errors will be reported as *events* on PVC object.
+* If resize failed in above step, in addition to events - `pvc.Status.Conditions` will be updated with `ResizeFailed: True`. Corresponding error will be added to condition field as well.
 * Depending on volume type next steps would be:
 
-    * If volume is of type that does not require file system resize, then `pvc.status.capacity` will be immediately updated to reflect new size. This would conclude the volume expand operation.
+    * If volume is of type that does not require file system resize, then `pvc.status.capacity` will be immediately updated to reflect new size. This would conclude the volume expand operation. Also `pvc.Status.Conditions` will be updated with `Ready: True`.
     * If volume if of type that requires file system resize then a file system resize will be performed on kubelet. Read below for steps that will be performed for file system resize.
 
 * If volume plugin is of type that can not do resizing of attached volumes (such as `Cinder`) then `ExpandVolumeDevice` can return error by checking for
@@ -115,20 +117,52 @@ reported as *events* on PVC object.
   successful.
 
 * To consider cases of missed PVC update events, an additional loop will reconcile bound PVCs with PVs. This additional loop will loop through all PVCs
-  and match `pvc.spec.capactiy` with `pv.spec.capacity` and add PVC in `volume_expand_controller`'s desired state of world if `pv.spec.capacity` is less
-  than `pvc.spec.capacity`.
+  and match `pvc.spec.resources.requests` with `pv.spec.capacity` and add PVC in `volume_expand_controller`'s desired state of world if `pv.spec.capacity` is less
+  than `pvc.spec.resources.requests`.
 
 * There will be additional checks in controller that grows PV size - to ensure that we do not make volume plugin API calls that can reduce size of PV.
 
 ### File system resize on kublet
 
+A File system resize will be pending on PVC until a new pod that uses this volume is scheduled somewhere. While theoretically we *can* perform
+online file system resize if volume type and file system supports it - we are leaving it for next iteration of this feature.
+
 * When calling `MountDevice` or `Setup` call of volume plugin, volume manager will in addition compare `pv.spec.capacity` and `pvc.status.capacity` and if `pv.spec.capacity` is greater
   than `pvc.status.spec.capacity` then volume manager will additionally resize the file system of volume.
 * The call to resize file system will be performed inside `operation_generator.GenerateMountVolumeFunc`.  `VolumeToMount` struct will be enhanced to store PVC as well.
 * Any errors during file system resize will be added as *events* to Pod object and mount operation will be failed.
+* If there are any errors during file system resize `pvc.Status.Conditions` will be updated with `ResizeFailed: True`. Any errors will be added to
+  `Conditions` field.
 * File System resize will not be performed on kubelet where volume being attached is ReadOnly. This is similar to pattern being used for performing formatting.
-* After file system resize is successful, `pvc.status.capacity` will be updated to match `pv.spec.capacity` and volume expand operation will be considered complete.
+* After file system resize is successful, `pvc.status.capacity` will be updated to match `pv.spec.capacity` and volume expand operation will be considered complete. Also `pvc.Status.Conditions` will be updated with `Ready: True`.
 
+#### Reduce coupling between resize operation and file system type
+
+A file system resize in general requires presence of tools such as `resize2fs` or `xfs_growfs` on the host where kubelet is running. There is a concern
+that open coding call to different resize tools direclty in Kubernetes will result in coupling between file system and resize operation. To solve this problem
+we have considered following options:
+
+1. Write a library that abstracts away various file system operations, such as - resizing, formatting etc.
+
+   Pros:
+   * Relatively well known pattern
+
+   Cons:
+   * Depending on version with which Kubernetes is compiled with, we are still tied to which file systems are supported in which version
+     of kubernetes.
+2. Ship a wrapper shell script that encapsulates various file system operations and as long as the shell script supports particular file system
+   the resize operation is supported.
+   Pros:
+   * Kubernetes Admin can easily replace default shell script with her own version and thereby adding support for more file system types.
+
+   Cons:
+   * I don't know if there is a pattern that exists in kube today for shipping shell scripts that are called out from code in Kubernetes. Flex is
+     different because, none of the flex scripts are shipped with Kuberntes.
+3. Ship resizing tools in a container.
+
+
+Of all options - #3 is our best bet but we are not quite there yet. Hence, I would like to propose that we ship with support for
+most common file systems in curent release and we revisit this coupling and solve it in next release.
 
 ## API and UI Design
 
@@ -172,6 +206,43 @@ spec:
 ### PVC API Change
 
 `pvc.spec.resources.requests.storage` field of pvc object will become mutable after this change.
+
+In addition to that PVC's status will have a `Conditions []PvcCondition` - which will be used
+to communicate the status of PVC to the user.
+
+So the `PersistentVolumeClaimStatus` will become:
+
+```go
+type PersistentVolumeClaimStatus struct {
+    Phase PersistentVolumeClaimPhase
+    AccessModes []PersistentVolumeAccessMode
+    Capacity ResourceList
+    // New Field added as part of this Change
+    Conditions []PVCCondition
+}
+
+// new API type added
+type PVCCondition struct {
+   Type PVCConditionType
+   Status ConditionStatus
+   LastProbeTime metav1.Time
+   LastTransitionTime metav1.Time
+   Reason string
+   Message string
+}
+
+// new API type
+type PVCConditionType string
+
+// new Constants
+const (
+   PVCReady PVCConditionType = "Ready"
+   PVCResizeStarted PVCConditionType = "ResizeStarted"
+   PVCResizeFailed  PVCResizeFailed = "ResizeFailed"
+)
+```
+
+
 
 ### Other API changes
 
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -36,7 +36,6 @@ Enable users to increase size of PVs that their pods are using. The user will up
 | NFS             | No                | No                          | No                       |
 | Flex            | Yes               | Maybe                       | No                       |
 | LocalStorage    | Yes               | Yes                         | No                       |
-| Block device    | Yes               | No                          | No                       |
 
 
 ## Implementation Design
@@ -54,6 +53,7 @@ For volume types that only require volume plugin based api call, this will be on
   can be validated by node authorizer.
 * This feature will be protected by an alpha feature gate, so as API changes needed for it.
 
+
 ### Admission Control and Validations
 
 * Resource quota code has to be updated to take into account PVC expand feature.
@@ -62,6 +62,7 @@ For volume types that only require volume plugin based api call, this will be on
 * Not all PVCs will be resizable even if underlying volume plugin allows that. Only dynamically provisioned volumes
 which are explicitly enabled by an admin will be allowed to be resized. A plugin in admission controller will forbid
 size update for PVCs for which resizing is not enabled by the admin.
+* The design proposal for raw block devices should make sure that, users aren't able to resize raw block devices.
 
 
 ### Controller Manager resize
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -19,19 +19,19 @@ Enable users to increase size of PVs that their pods are using. The user will up
 ## Volume Plugin Matrix
 
 
-| Volume Plugin   | Supports Resize   | Requires File system Resize |
-| ----------------| :---------------: | :--------------------------:|
-| EBS             | Yes               | Yes                         |
-| GCE PD          | Yes               | Yes                         |
-| Azure Disk      | Yes               | Yes                         |
-| Cinder          | Yes               | Yes                         |
-| Vsphere         | Yes               | Yes                         |
-| Ceph RBD        | Yes               | Yes                         |
-| Host Path       | No                | No                          |
-| GlusterFS       | Yes               | No                          |
-| Azure File      | No                | No                          |
-| Cephfs          | No                | No                          |
-| NFS             | No                | No                          |
+| Volume Plugin   | Supports Resize   | Requires File system Resize | Supported in 1.7 Release |
+| ----------------| :---------------: | :--------------------------:| :----------------------: |
+| EBS             | Yes               | Yes                         | Yes                      |
+| GCE PD          | Yes               | Yes                         | Yes                      |
+| Azure Disk      | Yes               | Yes                         | No                       |
+| Cinder          | Yes               | Yes                         | Yes                      |
+| Vsphere         | Yes               | Yes                         | No                       |
+| Ceph RBD        | Yes               | Yes                         | No                       |
+| Host Path       | No                | No                          | No                       |
+| GlusterFS       | Yes               | No                          | Yes                      |
+| Azure File      | No                | No                          | No                       |
+| Cephfs          | No                | No                          | No                       |
+| NFS             | No                | No                          | No                       |
 
 
 ## Implementation Design
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -34,6 +34,9 @@ Enable users to increase size of PVs that their pods are using. The user will up
 | Azure File      | No                | No                          | No                       |
 | Cephfs          | No                | No                          | No                       |
 | NFS             | No                | No                          | No                       |
+| Flex            | Yes               | Maybe                       | No                       |
+| LocalStorage    | Yes               | Yes                         | No                       |
+| Block device    | Yes               | No                          | No                       |
 
 
 ## Implementation Design
@@ -49,7 +52,7 @@ For volume types that only require volume plugin based api call, this will be on
 * `pvc.spec.resources.requests.storage` field of pvc object will become mutable after this change.
 * #sig-api-machinery has agreed to allow pvc's status update from kubelet as long as pvc and node relationship
   can be validated by node authorizer.
-* This feature will be protected by an alpha feature gate.
+* This feature will be protected by an alpha feature gate, so as API changes needed for it.
 
 ### Admission Control and Validations
 
@@ -66,6 +69,7 @@ new controller will be:
 * Watch for pvc update requests and add pvc to controller's desired state of world if a increase in volume size was requested. Once PVC is added to
   controller's desired state of world - `pvc.Status.Conditions` will be updated with `ResizeStarted: True`.
 * For unbound or pending PVCs - resize will trigger no action in `volume_expand_controller`.
+* If `pv.Spec.Capacity` already is of size greater or equal than requested size, similarly no action will be perfomed by the controller.
 * A reconciler will read desired state of world and perform corresponding volume resize operation. If there is a resize operation in progress
   for same volume then resize request will be pending and retried once previous resize request has completed.
 * Controller resize in effect will be level based rather than edge based. If there are more than one pending resize request for same PVC then
@@ -210,6 +214,10 @@ spec:
 In addition to that PVC's status will have a `Conditions []PvcCondition` - which will be used
 to communicate the status of PVC to the user.
 
+The API change will be protected by Alpha feature gate and api-server will not allow PVCs with
+`Status.Conditions` field if feature is not enabled. `omitempty` in serialization format will
+prevent presence of field if not set.
+
 So the `PersistentVolumeClaimStatus` will become:
 
 ```go
@@ -248,6 +256,3 @@ const (
 
 This proposal relies on ability to update PVC status from kubelet. While updating PVC's status
 a PATCH request must be made from kubelet to update the status.
-
-Also - an Admin can directly edit the PV and specify new size but controller will not perform
-any automatic resize of underlying volume or file system in such cases.
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -134,6 +134,11 @@ online file system resize if volume type and file system supports it - we are le
 * When calling `MountDevice` or `Setup` call of volume plugin, volume manager will in addition compare `pv.spec.capacity` and `pvc.status.capacity` and if `pv.spec.capacity` is greater
   than `pvc.status.spec.capacity` then volume manager will additionally resize the file system of volume.
 * The call to resize file system will be performed inside `operation_generator.GenerateMountVolumeFunc`.  `VolumeToMount` struct will be enhanced to store PVC as well.
+* The flow of file system resize will be as follow:
+    * Perform a resize based on file system used inside block device.
+    * If resize succeeds, proceed with mounting the device as usual.
+    * If resize failed with an error that shows no file system exists on the device, then log a warning and proceed with format and mount.
+    * If resize failed with any other error then fail the mount operation.
 * Any errors during file system resize will be added as *events* to Pod object and mount operation will be failed.
 * If there are any errors during file system resize `pvc.Status.Conditions` will be updated with `ResizeFailed: True`. Any errors will be added to
   `Conditions` field.
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -59,6 +59,9 @@ For volume types that only require volume plugin based api call, this will be on
 * Resource quota code has to be updated to take into account PVC expand feature.
 * In case volume plugin doesn’t support resize feature. The resize API request will be rejected and PVC object will not be saved. This check will be performed via an admission controller plugin.
 * In case requested size is smaller than current size of PVC. A validation will be used to reject the API request. (This could be moved to admission controller plugin too.)
+* Not all PVCs will be resizable even if underlying volume plugin allows that. Only dynamically provisioned volumes
+which are explicitly enabled by an admin will be allowed to be resized. A plugin in admission controller will forbid
+size update for PVCs for which resizing is not enabled by the admin.
 
 
 ### Controller Manager resize
@@ -75,7 +78,18 @@ new controller will be:
 * Controller resize in effect will be level based rather than edge based. If there are more than one pending resize request for same PVC then
   new resize requests for same PVC will replace older pending request.
 * Resize will be performed via volume plugin interface, executed inside a goroutine spawned by `operation_exectutor`.
-* A new plugin interface called `volume.Expander` will be added to volume plugin interface. The controller call to expand the PVC will look like:
+* A new plugin interface called `volume.Expander` will be added to volume plugin interface. The `Expander` interface
+  will also define if volume requires a file system resize:
+
+  ```go
+    type Expander interface {
+      // ExpandVolume expands the volume
+      ExpandVolumeDevice(spec *Spec, newSize resource.Quantity, oldSize resource.Quantity) error
+      RequiresFSResize() bool
+    }
+  ```
+
+* The controller call to expand the PVC will look like:
 
 ```go
 func (og *operationGenerator) GenerateExpandVolumeFunc(
@@ -138,6 +152,15 @@ func (og *operationGenerator) GenerateExpandVolumeFunc(
 
 A File system resize will be pending on PVC until a new pod that uses this volume is scheduled somewhere. While theoretically we *can* perform
 online file system resize if volume type and file system supports it - we are leaving it for next iteration of this feature.
+
+#### Prerequisite of File system resize
+
+* `pv.spec.capacity` must be greater than `pvc.status.spec.capacity`.
+* A fix in pv_controller has to made to fix `claim.Status.Capacity` only during binding. See comment by jan here - https://github.com/kubernetes/community/pull/657#discussion_r128008128
+* A fix in attach_detach controller has to be made to prevent fore detaching of volumes that are undergoing resize.
+This can be done by checking `pvc.Status.Conditions` during force detach. `AttachedVolume` struct doesn't hold a reference to PVC - so PVC info can either be directly cached in `AttachedVolume` along with PV spec or it can be fetched from PersistentVolume's ClaimRef binding info.
+
+#### Steps for resizing file system available on Volume
 
 * When calling `MountDevice` or `Setup` call of volume plugin, volume manager will in addition compare `pv.spec.capacity` and `pvc.status.capacity` and if `pv.spec.capacity` is greater
   than `pvc.status.spec.capacity` then volume manager will additionally resize the file system of volume.
@@ -263,7 +286,22 @@ const (
 )
 ```
 
+### StorageClass API change
 
+A new field called `AllowVolumeExpand` will be added to StorageClass. The default of this value
+will be `false` and only if it is true - PVC expansion will be allowed.
+
+```go
+type StorageClass struct {
+    metav1.TypeMeta
+    metav1.ObjectMeta
+    Provisioner string
+    Parameters map[string]string
+    // New Field added
+    // +optional
+    AllowVolumeExpand bool
+}
+```
 
 ### Other API changes
 
--- a/contributors/design-proposals/grow-volume-size.md
+++ b/contributors/design-proposals/grow-volume-size.md
@@ -15,11 +15,13 @@ Enable users to increase size of PVs that their pods are using. The user will up
 * As a user I am running Mysql on a 100GB volume - but I am running out of space, I should be able to increase size of volume mysql is using without losing all my data. (*online and with data*)
 * As a user I created a PVC requesting 2GB space. I am yet to start a pod with this PVC but I realize that I probably need more space. Without having to create a new PVC, I should be able to request more size with same PVC. (*offline and no data on disk*)
 * As a user I was running a rails application with 5GB of assets PVC. I have taken my application offline for maintenance but I would like to grow asset PVC to 10GB in size. (*offline but with data*)
+* As a user I am running an application on glusterfs. I should be able to resize the gluster volume without losing data or mount point. (*online and with data and without taking pod offline*)
+* In the logging project we run on dedicated clusters, we start out with 187Gi PVs for each of the elastic search pods. However, the amount of logs being produced can vary greatly from one cluster to another and its not uncommon that these volumes fill and we need to grow them.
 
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,159 @@
...
+
+* In case requested size is smaller than current size of PVC. A validation will be used to reject the API request. (This could be moved to admission controller plugin too.)
+
+* There will be additional checks in controller that grows PV size - to ensure that we do not make cloudprovider API calls that can reduce size of PV.
+
+* To consider cases of missed PVC update events, an additional loop will reconcile bound PVCs with PVs.
+
+* Resource Quota code in admission controller has to be updated to consider PVC updates.
+
+* The resize of file system will be performed on kubelet. If there is a running pod - no operation will be performed. Only when a new pod is started using same PVC - then kubelet will match device size and size of pv and attempt a resize of file system.  resizing filesystem will be a volume plugin function. It is upto volume plugin maintainer to correctly implement this. In following cases no resize will be necessary and hence volume plugin can return success without actually doing anything.
```
thockin: Isn't a resize specific to a filesystem?  Do we expect all volume drivers to bundle all tools for all filesystems?  That seems untenable.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,159 @@
...
+
+  * If disk being attached to the pod is unformatted. In which case since kubelet formats the disk, no resize is necessary.
+  * If PVC being attached to pod is of volume type that requires no file system level resize. Such as glusterfs.
+
+  Once file system resize is successful - kubelet will update `pv.spec.status.capacity` and `pvc.spec.status.capacity`field to reflect updated size. Kubelet will also
```
thockin: I dislike that the status is set in 2 different places - can we refine it to centralize that?

saad-ali:
* How do we keep track of the status of a resize operation?
* What happens if the resize operation fails? How is that communicated to end user?
* What happens if the user resizes multiple times in rapid succession? If you prevent a resize operation from starting while an existing resize operation is pending how will you keep track of that? What if the user removes and readds the annotation?
* A subresource on the PVC might be a good API?

contributors/design-proposals/grow-volume-size.md
childsb: This is similar to specifying the FS type of the PV and having kubelet format it for you and expect that the fs tools are on the node.  Are we expecting anything besides XFS and EXT4? 

The weird case may be azure or other Samba volumes.  But either way I think the fs tools required is minimal.  I would love it if the tools were containerized, but we dont require that now to format PVs....

gnufied: I went ahead and had a look at various distributions and how they ship the file system utils, and one thing which I could confirm is if formatting utils for `xfs` or `ext` are available on the host - then resize utils are available too. I looked at - centos7, ubuntu 16.04 (server and minimal), Opensuse-Leap 42.2 .

On all these distros if you install formatting package (such `xfsprogs`) then it comes with resizing utility too.

I haven't confirmed this for all file system types and neither on all distros, so I wouldn't call my information conclusive.  But - what I propose is, we start with supporting ext4 and xfs for resizing. That would cover majority of use cases.  For more exhaustive solution - we may want to wait for containarization proposal on which @jsafrane is working on. 

@thockin @childsb what you think? Is there any thing else we are missing?

gnufied: Also - we document the fact that, growing file system is only supported for `ext4` and `xfs` in case of block device storage.

gnufied:
@saad-ali my original plan was to add a field called `pv.status.conditions []PvCondition` which is similar to `PodCondition`. Something like:

```go
type PvCondition {
  Type PvConditionType ("ResizeStarted|ResizeInProgress|ResizeComplete|ResizeFailed) 
  Status pvConditionStatus (True|False|Unknwon)
  Reason string
  Message string
}

type PersistentVolumeStatus {
  Condition []PvCondition
}
```

which would store different "status" of resize operation. The problem is - this feature being Alpha, we will  have to either serialize this whole struct into a json string and then put it in annotation, or we will have to create new version of PersistentVolume apis. 

For Alpha release:

> * How do we keep track of the status of a resize operation?

We have alpha annotation `volume.alpha.kuberenetes.io/fs-resize-pending` and `pv.status.capacity`. Do we need additional information to track status of resize operation?

> * What happens if the resize operation fails? How is that communicated to end user?

If cloudprovider resize failed, we are going to attach an event to the pvc. If file system resize failed - we are going  to attach an event to the pod as well as pvc.

> *What happens if the user resizes multiple times in rapid succession? If you prevent a resize operation from starting while an existing resize operation is pending how will you keep track of that? What if the user removes and readds the annotation?

This shouldn't be a problem - because file system resize is only performed when a new pod is started. So all rapid resizes are limited to cloudprovider api calls and they update `pv.spec.capacity` as expected.  We don't stop users from doing resize when an existing resize is pending. 

However if user manually edits the PV and removes the annotation - we do have a problem. Isn't this expected that users aren't supposed to touch kubernetes.io annotations?


> A subresource on the PVC might be a good API?

I don't know what you mean. A subresource to represent what?

contributors/design-proposals/grow-volume-size.md
thockin: do those tools all have the same flags?  Or do we need to open-code a list of supported filesystems and their tooling?

gnufied: That is somewhat problematic thing about resize(when compared to mkfs). Resize command for each file system is going to be different and their flags too. So, we will have to pretty much compose appropriate command via switch statement or something like that (or best case - define some interface functions)

thockin:
blech!!  can we find a way orthogonalize filesystems and storage subsystems?

gnufied:
hmm - orthogonalize at what level? There are two distinct problems:

1. Having to open-code support for each file system (or at least ext4 and xfs to begin with). We can mitigate some of this by - abstracting away the whole file system tooling in a library. So as a call from k8s boils down to:

```go
fs := NewFileSystem("xfs")
fs.resize(device_id, "100GB")
fs.format(device_id)
```

But ultimately this library has to hide that detail. We can implement this library as part of this proposal, so as support for different file systems can be expanded in k8s without possibly changing k8s core. 

2. Unavailability of tools on host.  That is something we can't easily solve without containarizing all the tools.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,159 @@
...
+| GCE PD          | Yes               | Yes                         | Yes                      |
+| Azure Disk      | Yes               | Yes                         | No                       |
+| Cinder          | Yes               | Yes                         | Yes                      |
+| Vsphere         | Yes               | Yes                         | No                       |
+| Ceph RBD        | Yes               | Yes                         | No                       |
+| Host Path       | No                | No                          | No                       |
+| GlusterFS       | Yes               | No                          | Yes                      |
+| Azure File      | No                | No                          | No                       |
+| Cephfs          | No                | No                          | No                       |
+| NFS             | No                | No                          | No                       |
```
taherv: flexvolume seems to be missing from this table .... is that intentional ? What about flexvolume with respect to this proposal ?

gnufied: If flex api adds support for a resize call - it can be added to the list. But for now - since it doesn't, I haven't added flex to the list.

jsafrane:
There is `/usr/sbin/fsadm` which can resize ext2/3/4, xfs and RaiserFS. It comes as part of LVM utilities and I don't know if it's available in "minimal" distros.

contributors/design-proposals/grow-volume-size.md
xingzhou: does it mean a common utility tool for all the volume plugin to use?

aglitke: Please also consider how resize would work for PVCs exposed as raw block devices.  While we don't have this feature today it is also being designed and discussed as part of the [local storage proposal](https://github.com/kubernetes/community/pull/306).  In the case of raw devices we can also allow online resize since the pod itself would be responsible for rescanning the device.

gnufied: @xingzhou I don't think a common utility for all volume plugins, but rather a common utility which can abstract away resize function for all/most file system types.

contributors/design-proposals/grow-volume-size.md
npramod: How about storage provisioned by out-of-tree provisioners that support Dynamic Provisioning(DP) today? Any thoughts on how resize would work there. For legacy/non-cloud native storage systems, Storage subsystem side operations might be required before a host side filesystem resize.

gnufied: Any external controller can implement `resize` mechanism of cloudprovider and leave host side file system resize to kubelet. Correct me if I am wrong but I don't think - this proposal stops external provisioners from doing that.

contributors/design-proposals/grow-volume-size.md
gnufied: `pvc.status.capacity` already exists, this proposal adds an additional field in PVs. but I will think if we can centralize this.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,159 @@
...
+
+
+## Implementation Design
+
+For volume type that supports growing the PV size, this will be a two step operation:
+
+* A controller in master-controller will listen for PVC events and perform corresponding cloudprovider operation. If successful - controller will store new device size in PV. Some cloudproviders (such as cinder) - do not allow resizing of attached volumes. In such cases - it is upto volume plugin maintainer to decide appropriate behaviour. Volume Plugin maintainer can choose to ignore resize request if disk is attached to a pod (and add appropriate error events to PVC object).  Resize request will keep failing until user corrects the error. User can take necessary action in such cases (such as scale down the pod) which will allow resize to proceed normally.
+
+  In case where volume type requires no file system resize, both PV & PVC objects will be updated accordingly and `status.capacity` of both objects will reflect new size.
```
dyaniv: Would the size update even if the change was done outside Kube?
If the admin did the change for the user via the storage management API.

contributors/design-proposals/grow-volume-size.md
fabiand: @aglitke Even if a pod is consuming the raw block device, I'd still assume that the host/kubelet is owning it - i.e. attaching/detaching it, and also rescanning it if necessary.
But if that's the case, the this could lead to surprises within the pod - if the device just becomes larger from one second to the other.

gnufied: @fabiand what kind of surprises you are talking about - if local device pod was using became larger? As proposed in this proposal - only device size will become larger. The size available to the pod will still be the same until file system has been resized too. 

Also, I am not sure - if there is a way to centrally resize a device for local storage. It has to be done from daemonset (as per current local storage proposal) which actually does provisioning of local storage.  All in all - I don't expect PV expand to be supported in local disks for while.  What I would be worried about is - if current design somehow inhibits that, which I don't think it does. cc @msau42 for local storage bits.

contributors/design-proposals/grow-volume-size.md
gnufied: If admin modifies volume outside kube, as per this design size update will not be observed by Kubernetes. I would argue that - it is conscious decision that we have taken, if admin modifies volume objects outside, he is entirely responsible for resizing file systems or making corresponding changes in volume objects.

contributors/design-proposals/grow-volume-size.md
msau42: For local storage, we're trying to put most of the logic in an external daemonset, to minimize kubelet dependencies, and to be able to support all the myriad ways to configure local storage, so each admin can customize it to their environment.  So, if say someone wants to use lvm as their local storage backend, then they can have all the lvm tools in the daemonset and do the resizing from there.  The daemonset needs some sort of signal from the system when to resize.

contributors/design-proposals/grow-volume-size.md
xingzhou: so if user uses LVM as storage pool, the resize operation will be rejected in this case? do we need another feature request for FlexVolume to add resize call support?

taherv: +1 for including flexvol support (either in this proposal, or as a sister-proposal).

xingzhou: there are some requirements in supporting LVM from our side, I can help create a proposal on FlexVolume for this and take care of the development, what's your idea @gnufied?

taherv: i can be an extra hand for flexvol support, too, @xingzhou

gnufied: Yeah, I think we have to have another proposal for resize of LVM volumes via Flex.  It wouldn't hurt to start a parallel proposal for flex that implements resizing based on this proposal. cc @chakri-nelluri

xingzhou: Sure, let me prepare and submit a new proposal on this, thx
/cc @taherv

contributors/design-proposals/grow-volume-size.md
fabiand: @gnufied Right, that's true for volumes thus file-systems. But @aglitke was speaking of raw block volumes, and there the size change is - IIRC - known once we rescan the bus, at least in the case of (i)SCSI devices.

contributors/design-proposals/grow-volume-size.md
ddysher: My understanding is that external controller can work like in-tree controller by just watching update event and resize volume accordingly, but we do need to clearly outline the 'protocol' we used (e.g. `volume.alpha.kubernetes.io/fs-resize-pending` annotation).

ddysher:
Why is this a problem, I think there is quite a few features using json string?
> The problem is - this feature being Alpha, we will have to either serialize this whole struct into a json string and then put it in annotation, or we will have to create new version of PersistentVolume apis.

@gnufied 
I'm interested in extending pv/pvc to have some sort of health status report by adding taints or conditions to PVs, so there might be some overlap related to PV conditions described here. For example, if a PV becomes unhealthy (e.g. lost connection, local volume partition error), then we'll want to taint the PV and only PVCs tolerate the taint can use it; for already bound PV/PVC, we'll do an unbind and reschedule PVC.

For node case, kubelet reports condition and node controller taints the node.  For volume, maybe we can have volume plugin reports the condition and pv controller to taint the node, is this align with ur plan?

contributors/design-proposals/grow-volume-size.md
dyaniv: If the service admin makes a change by request for a volume I argue this should be detected.
Maybe the FS resize should be manually triggered, but the sizing should reflect the storage service state, not make it inconsistent with Kube.
The storage service admin might not care for specifically Kube as a application and would expect it to be updated automatically like in dealing with OSes using volumes.

contributors/design-proposals/grow-volume-size.md
gnufied: Yep that is correct. I think as long as protocol being specified here is followed, any external controller can perform resize. This shouldn't be a problem.

dyaniv: This sounds like what I asked for below.
Don't we want to have the ability to resize the FS on external resize as well? Reactive fs resize.

gnufied: As long as - necessary fields/annotation that exist on PV that tells kubelet that a fs resize is needed - Kubelet will perform fs resize.  Ideally - we don't want admins to modify those annotations/fields manually - but we are not going to put additional restrictions if they want to!

So in a sense, if admin has resized a volume manually (via ebs resize command) and did not also update the PV object with new size and `volume.alpha.kubernetes.io/fs-resize-pending` annotation then no FS resize will be performed on kubelet. But if admin manually edits the PV and makes same changes that resize controller would do, then we can't distinguish manual changes from resize controller change and hence a FS resize will be performed.  But just because you can, doesn't mean you should.

gnufied:
> then we'll want to taint the PV and only PVCs tolerate the taint can use it; for already bound PV/PVC, we'll do an unbind and reschedule PVC.

What you mean by reschedule the PVC? You mean unbind from current PV and bind it to new one? We have to be careful because - PVs usually contain data and say if a MySQL PVC is bound to a new PV with no data in it, we have problem.

contributors/design-proposals/grow-volume-size.md
dyaniv: I feel this is no standard. Similar to ISCSI or NFS resizing. 
It is expected that volume size metadata is updated (that is why there are events), even if the FS is not resized on top.

gnufied: > It is expected that volume size metadata is updated (that is why there are events), even if the FS is not resized on top.

I am sorry, I don't understand what that means. Can you elaborate? May be perhaps with some examples.

ddysher:
Yes, unbind PVC from current PVC.  Starting from v1.8, scheduling will take PV/PVC into consideration (now they are two separate process).

Agree we should be careful about this.  So like node NotReady taint, we might add tolerations to pods in admission controller by default, unless user explicitly says otherwise, in which case, we can freely do an unbind.

contributors/design-proposals/grow-volume-size.md
gnufied: I have removed this duplication now.

gnufied:
@saad-ali @thockin PTAL. I have flexed out the details of the proposal quite a bit. Improved error handling.

gnufied:
also @liggitt let me know if everything looks good from #api-machinery and authorization perspective.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+
+
+
+### Other API changes
+
+This proposal relies on ability to update PVC status from kubelet. While updating PVC's status
+a PATCH request must be made from kubelet to update the status.
+
+Also - an Admin can directly edit the PV and specify new size but controller will not perform
+any automatic resize of underlying volume or file system in such cases.
```
ddysher: Is this true for file system resize as well?  My understanding is that kubelet will observe `pv.spec.capacity` greater than `pvc.status.spec.capacity` and perform a file system resize.

gnufied: Yeah that is a good point, I think we can't avoid doing file system resize. I do not see any particular harm in doing so too.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+* #sig-api-machinery has agreed to allow pvc's status update from kubelet as long as pvc and node relationship
+  can be validated by node authorizer.
+* This feature will be protected by an alpha feature gate.
+
+### Admission Control and Validations
+
+* Resource quota code has to be updated to take into account PVC expand feature.
+* In case volume plugin doesn’t support resize feature. The resize API request will be rejected and PVC object will not be saved. This check will be performed via an admission controller plugin.
+* In case requested size is smaller than current size of PVC. A validation will be used to reject the API request. (This could be moved to admission controller plugin too.)
+
```
jsafrane: Mising a note about external provisioners - perhaps the (internal) admission controller should ignore those and let the external provisioner implement its own (external) admission service.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+| GlusterFS       | Yes               | No                          | Yes                      |
+| Cinder          | Yes               | Yes                         | Yes                      |
+| Vsphere         | Yes               | Yes                         | No                       |
+| Ceph RBD        | Yes               | Yes                         | No                       |
+| Host Path       | No                | No                          | No                       |
+| Azure Disk      | Yes               | Yes                         | No                       |
+| Azure File      | No                | No                          | No                       |
+| Cephfs          | No                | No                          | No                       |
+| NFS             | No                | No                          | No                       |
+
```
jsafrane: Missing Flex (with no support for resize?)

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+    }
+    return expandFunc, nil
+}
+```
+
+* Once volume expand is successful, the volume will be marked as expanded and new size will be updated in `pv.spec.capacity`. Any errors will be reported as *events* on PVC object.
+* If resize failed in above step, in addition to events - `pvc.Status.Conditions` will be updated with `ResizeFailed: True`. Corresponding error will be added to condition field as well.
+* Depending on volume type next steps would be:
+
+    * If volume is of type that does not require file system resize, then `pvc.status.capacity` will be immediately updated to reflect new size. This would conclude the volume expand operation. Also `pvc.Status.Conditions` will be updated with `Ready: True`.
```
jsafrane: `If volume is of type that does not require file system resize` - how do you know that?

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+* To consider cases of missed PVC update events, an additional loop will reconcile bound PVCs with PVs. This additional loop will loop through all PVCs
+  and match `pvc.spec.resources.requests` with `pv.spec.capacity` and add PVC in `volume_expand_controller`'s desired state of world if `pv.spec.capacity` is less
+  than `pvc.spec.resources.requests`.
+
+* There will be additional checks in controller that grows PV size - to ensure that we do not make volume plugin API calls that can reduce size of PV.
+
+### File system resize on kublet
+
+A File system resize will be pending on PVC until a new pod that uses this volume is scheduled somewhere. While theoretically we *can* perform
+online file system resize if volume type and file system supports it - we are leaving it for next iteration of this feature.
```
jsafrane: So the user can wait for a quite long time until the resize is complete and the application can use more disk space. Ok, that sounds reasonable, however we should somehow signal the user that PV resize is done and all that we're waiting for is restart of a pod so we can do fsresize.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+
+* When calling `MountDevice` or `Setup` call of volume plugin, volume manager will in addition compare `pv.spec.capacity` and `pvc.status.capacity` and if `pv.spec.capacity` is greater
```
jsafrane: This rings a bell... right now, PV controller blindly puts PV.Spec.Capacity to claim.Status.Capacity every periodic resync - it tries to fix any damage done by user or other controller. We need to fix it and fill claim.Status.Capacity only during binding.

https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/pv_controller.go#L622

contributors/design-proposals/grow-volume-size.md
jsafrane: Also missing LocalStorage, assuming "no resize support" - these PVs are created by external daemonset and cannot be dynamically provisioned now.

liggitt:
this seems to be putting the control over the PV in the hands of the PVC spec... in general, I wouldn't have expected that. Dynamically provisioned PVs are "made-to-order" for a particular PVC, but existing PVs don't change themselves to suit a PVC. Is this proposal limited to dynamically provisioned PVs? It doesn't seem like PVs an admin provisioned independently from PVCs should be able to be resized by end-users, right?

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+   - ReadWriteOnce
+ resources:
+   requests:
+     storage: 1Gi
+```
+
+Users can request new size of underlying PV by simply editing the PVC and requesting new size:
+
+```
+~> kubectl edit pvc volume-claim
```
jsafrane: Maybe kubectl patch would be shorter?

```
kubectl patch pvc volume-claim -p '{"spec":{"resources":{"requests":{"storage":"10Gi"}}}}'
```

BTW, should we add `kubectl pvc resize volume-claim 10Gi` eventually?

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+* In case volume plugin doesn’t support resize feature. The resize API request will be rejected and PVC object will not be saved. This check will be performed via an admission controller plugin.
+* In case requested size is smaller than current size of PVC. A validation will be used to reject the API request. (This could be moved to admission controller plugin too.)
+
+
+### Controller Manager resize
+
+A new controller called `volume_expand_controller` will listen for pvc size expansion requests and take action as needed. The steps performed in this
+new controller will be:
+
+* Watch for pvc update requests and add pvc to controller's desired state of world if a increase in volume size was requested. Once PVC is added to
```
jsafrane: I am not sure if it's real "desired state of world", to me it sounds like a work queue.

Desired state of the world would contain pvc.status.capacity, current state of world would contain pv.status.capacity and reconciler would periodically compare these two...

contributors/design-proposals/grow-volume-size.md
gnufied: Isn't that handled by event and `Conditions` field on pvc?

contributors/design-proposals/grow-volume-size.md
gnufied: We would implement plugin level functions to check if volume type requires file system resize.

contributors/design-proposals/grow-volume-size.md
gnufied: okay that seems fair, we can name the internal data structure here since we don't really have an actual state of world.

gnufied:
@liggitt we only take action on bound pvcs (obviously), so I guess it seems reasonable that user should be able to change size of PV they already have. Otherwise, what is stopping them from just requesting another PV with larger size? The only problem of course with that is - they have to find a way to migrate the data.

contributors/design-proposals/grow-volume-size.md
gnufied: ack, I wil file a task for this to happen. thanks for catching it.

contributors/design-proposals/grow-volume-size.md
gnufied: may be yeah - but I think it can be considered later? I don't have very strong opinions on this one either way, as long as we support resizing via API.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+  controller's desired state of world - `pvc.Status.Conditions` will be updated with `ResizeStarted: True`.
+* For unbound or pending PVCs - resize will trigger no action in `volume_expand_controller`.
```
gnufied: I am thinking Conditions for pending or unbound pvcs would be simply empty. But pending in this context generally means pending phase.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+* In case requested size is smaller than current size of PVC. A validation will be used to reject the API request. (This could be moved to admission controller plugin too.)
+
+
+### Controller Manager resize
+
+A new controller called `volume_expand_controller` will listen for pvc size expansion requests and take action as needed. The steps performed in this
+new controller will be:
+
+* Watch for pvc update requests and add pvc to controller's desired state of world if a increase in volume size was requested. Once PVC is added to
+  controller's desired state of world - `pvc.Status.Conditions` will be updated with `ResizeStarted: True`.
```
msau42: Is the purpose of this field so that multiple resizes don't happen at the same time on the same volume?  What happens if the resize process crashes in the middle?

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+### File system resize on kublet
+
+A File system resize will be pending on PVC until a new pod that uses this volume is scheduled somewhere. While theoretically we *can* perform
+online file system resize if volume type and file system supports it - we are leaving it for next iteration of this feature.
+
+* When calling `MountDevice` or `Setup` call of volume plugin, volume manager will in addition compare `pv.spec.capacity` and `pvc.status.capacity` and if `pv.spec.capacity` is greater
+  than `pvc.status.spec.capacity` then volume manager will additionally resize the file system of volume.
+* The call to resize file system will be performed inside `operation_generator.GenerateMountVolumeFunc`.  `VolumeToMount` struct will be enhanced to store PVC as well.
+* Any errors during file system resize will be added as *events* to Pod object and mount operation will be failed.
+* If there are any errors during file system resize `pvc.Status.Conditions` will be updated with `ResizeFailed: True`. Any errors will be added to
```
msau42: Do we want to distinguish between volume resize failures and fs resize failures?

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,253 @@
...
+
+* There will be additional checks in controller that grows PV size - to ensure that we do not make volume plugin API calls that can reduce size of PV.
+
+### File system resize on kublet
+
+A File system resize will be pending on PVC until a new pod that uses this volume is scheduled somewhere. While theoretically we *can* perform
+online file system resize if volume type and file system supports it - we are leaving it for next iteration of this feature.
+
+* When calling `MountDevice` or `Setup` call of volume plugin, volume manager will in addition compare `pv.spec.capacity` and `pvc.status.capacity` and if `pv.spec.capacity` is greater
+  than `pvc.status.spec.capacity` then volume manager will additionally resize the file system of volume.
```
msau42: It is kubelet volume manager that is doing fs resize, instead of volume plugin?  How does it know what type of fs the volume has, and which tools to use?

contributors/design-proposals/grow-volume-size.md
msau42: Is there a race condition, where you update `pv.spec.capacity` first and before you update the pvc capacity, kubelet tries to do a fs resize?

contributors/design-proposals/grow-volume-size.md
gnufied: Purpose of `Conditions` field is entirely for informing the user about progress of resize. This field is not used in any state machine transition.

If resize crashes in the middle, I would expect that it would be caught by populator loop and resize would be attempted again.

ddysher:
pvc can bind to a pv with larger size, if user requests a larger size but still falls under the size of bound pv, are we stilling going to update `ResizeStarted: True` and go through the whole process?

We can check the size and if new requested size still falls under PV size, we do nothing. However, I'm not sure if fs resize is still needed here?  If it's required, then we have to go through the whole process to update PV size; however, this is effectively a `decrease` request from volume plugin's point of view. e.g. original pvc.spec.capacity=1Gi, and is bound to a pv.spec.capacity=5Gi (thus pvc.status.capacity=5Gi), and new request is pvc.spec.capacity=3Gi.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,159 @@
...
+## Non Goals
+
+* Reducing size of Persistent Volumes: We realize that, reducing size of PV is way riskier than increasing it. Reducing size of a PV could be a destructive operation and it requires support from underlying file system and volume type. In most cases it also requires that file system being resized is unmounted.
+
+* Rebinding PV and PVC: Kubernetes will only attempt to resize the currently bound PV and PVC and will not attempt to relocate data  from a PV to a new PV and rebind the PVC to newly created PV.
+
+## Use Cases
+
+* As a user I am running Mysql on a 100GB volume - but I am running out of space, I should be able to increase size of volume mysql is using without losing all my data. (*online and with data*)
+* As a user I created a PVC requesting 2GB space. I am yet to start a pod with this PVC but I realize that I probably need more space. Without having to create a new PVC, I should be able to request more size with same PVC. (*offline and no data on disk*)
```
j-griffith: I would omit this as a valid use case, if you created it and have yet to start using it just create a new one and try and get it right the second time?

gnufied:
@ddysher yeah in that case, no resize should be necessary. We can compare `pv.spec.capacity` against `pvc.spec.requests.storage` and if PV already has necessary size - we can simply skip the resize operation.  I will add a note for this in the proposal.

contributors/design-proposals/grow-volume-size.md
gnufied: No the kubelet's volume_manager will call into certain volume plugin's function that will do actual file system resize. I haven't added the name of the function or interface detail for doing that here. I can expand on that.

contributors/design-proposals/grow-volume-size.md
gnufied: ack

contributors/design-proposals/grow-volume-size.md
gnufied: I will update that in the chart. but as discussed elsewhere as long as necessary API/protocol exists on PVC, it wouldn't be hard to implement in future.

contributors/design-proposals/grow-volume-size.md
gnufied: ack

contributors/design-proposals/grow-volume-size.md
gnufied: If admin manually modifies `pv.spec.capacity` - I am thinking, we won't do any cloudprovider call to resize if the new size meets the critirea of PVCs. i.e - before considering a volume for resize in controller-manager, we will also compare pv.spec.capacity with requested size. It may be that, pv already satisfies the requested size and no cloudprovider based call to resize is necessary. 

This isn't captured in proposal, but I will add it. hopefully, that covers potential race you are talking abou.t

contributors/design-proposals/grow-volume-size.md
gnufied: I dunno - we are never going to do file system resize if volume resize failed right? so, we can never have both field present, I think.

gnufied:
Just to summarize today's discussion and items that still need resolution are:

1. For volume types where a disk could be detached while mount/resize is in progress - resize shouldn't corrupt the file system. This is especially applicable to something like GCE-PD which has 6 minute timeout for force detaching a volume, if user has deleted the pod and unmount is not successful.

Resolution : I am going to look into solving this gracefully and without introducing too much complexity.

2. Conditions field introduced in design https://github.com/kubernetes/community/pull/657/files#diff-d9927f780b642a63774acfe069abb30bR221 is not protected by Alpha feature gate.

Resolution : Talk to #sig-api-machinery and see if there is a precedence for this or if there is an alternative. We considered using annotations for Conditions in alpha release, but in my previous conversation with #sig-api-machinery, they are planning to tighten those permissions and make sure only status is updatable from kubelet. It is just a bug/mis-feature that annotation and status permissions are taken together right now but #sig-api-machinery aims to fix that.

saad-ali:
We had a good meeting about this design today. The outstanding issues (in order of most important to least important):

1. Cluster admin control
    * An admin should be able to prevent users from resizing.
    * Proposal (2 parts):
      1. Only allow resizing on dynamically provisioned volumes
      2. Add a `AllowResize` field to StorageClass (default `false`), so that the cluster admin consciously enables the behavior for their users.
2. New API fields for alpha feature doesn't seem right
    * This feature is alpha, but it is proposing adding new fields to stable API objects. If annotations are no longer the recommended way to add alpha functionality, we should get guidance on how to add alpha level fields. We can't have an alpha feature go out with new first class fields (i.e. changes to the v1 API).
3. Resize for raw block PVCs
    * Resizing a block volume that is in use via a file system is apparently ok, but the same may not be true for direct block volume access.
4. Recovery behavior for FSResize failure
    * If storage backend resize succeeds, but filesystem resize continuously fails, user will not be able to mount the volume and their pod will be stuck in pending state. How do they recover from this state? Can they "cancel" the resize, and say just mount it as is?
5. FSResize with force detach could result in data corruption.
    * If a filesystem resize is pending, and the volume is forcibly detached, this could result in data corruption.

gnufied:
@saad-ali about API field for Alpha feature the guidance we got is:

We feature gate PVC creation with `Status.Conditions`  and do not allow PVC with this field to be created or updated if this feature is not enabled. Similarly - we use `omitempty` in json/protbuf serialization so that, the field is not present in GET responses if it is empty. cc @liggitt

gnufied:
I think we have solution to all issues discussed in the call. 

1. **Cluster Admin Control:**
    Add a top level field to `StorageClass` called `AllowVolumeExpand: True`  which will be false by default. Only for dynamically provisioned PVs with matching storageclass this feature would be enaled. The field will be protected by same feature gate that governs all of this feature and it will not be enabled (i.e users can't see or set the field) if the feature gate is disabled. 

2. **New API fields for alpha feature doesn't seem right**
   We feature gate PVC creation with Status.Conditions and do not allow PVC with this field to be created or updated if this feature is not enabled. Similarly - we use omitempty in json/protbuf serialization so that, the field is not present in GET responses if it is empty. 

3. **Resize for raw block PVCs**
    I think for now, we can document the fact that - we don't support resizing of raw block PVCs.  The existing block devices can't be used directly by any application running on top of kubernetes (it would require a file system on top of it). 

4. **Recovery behavior for FSResize failure**
    Currently we don't have any way of stopping or cancelling resize.  This feature being alpha, if it becomes a problem we can introduce an annotation that can be added to PVC to request cancellation of file system resize operation. But I think, we would be fine with properly reporting the error to end user and letting admin intervene with operation.

5. **FSResize with force detach could result in data corruption.**
    Since we know if a PVC is being resized via `pvc.Status.Conditions` field, I am thinking we can use that field to prevent force detaching volumes that are undergoing resize. This would be a simple 10 or 15 line change in attach/detach controller to prevent force detach of volumes undergoing a resize. 

@saad-ali @kubernetes/sig-storage-api-reviews let me know if this addresses most of outstanding issues.

liggitt:
> Only for dynamically provisioned PVs with matching storageclass this feature would be enabled. 

What is the effect of the feature not being enabled on a particular PV? It seems like the user would still be able to modify the PVC spec. Would the PV (and therefore PVC status) just not react to that change? What API field indicates whether a resize is pending or not in that case?

msau42:
> Resize for raw block PVCs

More than document, should we block it too?

gnufied:
@liggitt No the plan is to block size update of PVC which weren't dynamically provisioned. And even if they were dynamically provisioned  and StorageClass field disallows resizing then again PVC size update will not be permitted.  This is being implemented via an admission control plugin. @NickrenREN

NickrenREN:
Yeah, i am working on pvc resize admission control plugin

gnufied:
@msau42 by block it, you mean don't support it? That is the plan.

gnufied:
I have added design proposal with resolution to oustanding issues raised in above comment. PTAL.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -59,6 +59,9 @@ For volume types that only require volume plugin based api call, this will be on
 * Resource quota code has to be updated to take into account PVC expand feature.
 * In case volume plugin doesn’t support resize feature. The resize API request will be rejected and PVC object will not be saved. This check will be performed via an admission controller plugin.
 * In case requested size is smaller than current size of PVC. A validation will be used to reject the API request. (This could be moved to admission controller plugin too.)
```
NickrenREN: Now this check will be performed by validation, right ?

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,301 @@
...
+| GlusterFS       | Yes               | No                          | Yes                      |
+| Cinder          | Yes               | Yes                         | Yes                      |
+| Vsphere         | Yes               | Yes                         | No                       |
+| Ceph RBD        | Yes               | Yes                         | No                       |
+| Host Path       | No                | No                          | No                       |
+| Azure Disk      | Yes               | Yes                         | No                       |
+| Azure File      | No                | No                          | No                       |
+| Cephfs          | No                | No                          | No                       |
+| NFS             | No                | No                          | No                       |
+| Flex            | Yes               | Maybe                       | No                       |
```
msau42: Wouldn't flex resize depend on the underlying flex driver?

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,301 @@
...
+| LocalStorage    | Yes               | Yes                         | No                       |
+| Block device    | Yes               | No                          | No                       |
```
msau42: Block device isn't really a volume plugin.  It's a new way of being able to consume volumes.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,301 @@
...
+| Cinder          | Yes               | Yes                         | Yes                      |
+| Vsphere         | Yes               | Yes                         | No                       |
+| Ceph RBD        | Yes               | Yes                         | No                       |
+| Host Path       | No                | No                          | No                       |
+| Azure Disk      | Yes               | Yes                         | No                       |
+| Azure File      | No                | No                          | No                       |
+| Cephfs          | No                | No                          | No                       |
+| NFS             | No                | No                          | No                       |
+| Flex            | Yes               | Maybe                       | No                       |
+| LocalStorage    | Yes               | Yes                         | No                       |
```
msau42: This is only true for an lvm-based approach to local storage, which is not implemented at the moment.

contributors/design-proposals/grow-volume-size.md
gnufied: Yes, it would be peformed via API validation for now.

contributors/design-proposals/grow-volume-size.md
gnufied: Yeah - I think I should put both fields to have `Maybe`. But @xingzhou has opened a feature to track adding resize support for flex volumes - https://github.com/kubernetes/features/issues/304

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Admission Control and Validations
+
+* Resource quota code has to be updated to take into account PVC expand feature.
+* In case volume plugin doesn’t support resize feature. The resize API request will be rejected and PVC object will not be saved. This check will be performed via an admission controller plugin.
+* In case requested size is smaller than current size of PVC. A validation will be used to reject the API request. (This could be moved to admission controller plugin too.)
+* Not all PVCs will be resizable even if underlying volume plugin allows that. Only dynamically provisioned volumes
+which are explicitly enabled by an admin will be allowed to be resized. A plugin in admission controller will forbid
+size update for PVCs for which resizing is not enabled by the admin.
+
```
humblec: Are we blocking pvc deletion when PV resize is in progress ?

gnufied: @humblec no we don't.  Can you explain what scenario you had in mind that would cause problem if user deletes PVC while resize is in progress?

saad-ali:
> I think we have solution to all issues discussed in the call. 
> 
> 1. **Cluster Admin Control:**
>    Add a top level field to `StorageClass` called `AllowVolumeExpand: True`  which will be false by default. Only for dynamically provisioned PVs with matching storageclass this feature would be enaled. The field will be protected by same feature gate that governs all of this feature and it will not be enabled (i.e users can't see or set the field) if the feature gate is disabled. 

SGTM 

> 2. **New API fields for alpha feature doesn't seem right**
>   We feature gate PVC creation with Status.Conditions and do not allow PVC with this field to be created or updated if this feature is not enabled. Similarly - we use omitempty in json/protbuf serialization so that, the field is not present in GET responses if it is empty.  That is the guidance we got from #sig-api-machinery

This is still concerning to me because when you introduce a new first class field in the v1 API, it can not be removed. The idea with alpha features is that we should be able to completely remove them without breaking our backwards compatibility promises. @kubernetes/sig-api-machinery-misc any suggestions: in the past we would use annotations to test new fields before adding them; apparently that is not recommended anymore? If so, what is the recommended way to add alpha fields?

> 3. **Resize for raw block PVCs**
>    I think for now, we can document the fact that - we don't support resizing of raw block PVCs.  The existing block devices can't be used directly by any application running on top of kubernetes (it would require a file system on top of it). 

> More than document, should we block it too?

I forgot what we agreed on here. That block volumes must never be able to be resized through this API (enforced at validation) or that we allow it but don't do the resize until the volume is not attached? +@msau42 Should run this by Erin Boyd as well.

> 4. **Recovery behavior for FSResize failure**
>    Currently we don't have any way of stopping or cancelling resize.  This feature being alpha, if it becomes a problem we can introduce an annotation that can be added to PVC to request cancellation of file system resize operation. But I think, we would be fine with properly reporting the error to end user and letting admin intervene with operation.

Ack.

> 5. **FSResize with force detach could result in data corruption.**
>    Since we know if a PVC is being resized via `pvc.Status.Conditions` field, I am thinking we can use that field to prevent force detaching volumes that are undergoing resize. This would be a simple 10 or 15 line change in attach/detach controller to prevent force detach of volumes undergoing a resize. 

Not sure how I feel about this, because it could cause major regressions, but I don't think this should be a blocker for alpha.

msau42:
> > 3. **Resize for raw block PVCs**
> >    I think for now, we can document the fact that - we don't support resizing of raw block PVCs.  The existing block devices can't be used directly by any application running on top of kubernetes (it would require a file system on top of it). 
> 
> > More than document, should we block it too?
> 
> I forgot what we agreed on here. That block volumes must never be able to be resized through this API (enforced at validation) or that we allow it but don't do the resize until the volume is not attached? +@msau42 Should run this by Erin Boyd as well.

Yes, with this initial phase, we should enforce that you cannot do raw block resize until we fully understand the ramifications on applications accessing raw block.  cc @erinboyd

ianchakeres:
@gnufied @saad-ali @msau42 @erinboyd - I'm a little confused by the comments about block PVCs being prohibited from expansion. 

Many of the volume types for different volume plugins are backed by block volumes, that are being formatted with a file system. For example, GCE PD and Ceph RBD. 

Why would we prevent resize of block volumes when the filesystem isn't being managed by the plugin? 

I'd expect a volume to be resizable, if the backing storage system and related plugin supports it.

liggitt:
> This is still concerning to me because when you introduce a new first class field in the v1 API, it can not be removed. The idea with alpha features is that we should be able to completely remove them without breaking our backwards compatibility promises. @kubernetes/sig-api-machinery-misc any suggestions: in the past we would use annotations to test new fields before adding them; apparently that is not recommended anymore? If so, what is the recommended way to add alpha fields?

discussion in https://github.com/kubernetes/kubernetes/issues/34508 and https://docs.google.com/document/d/1wuoSqHkeT51mQQ7dIFhUKrdi3-1wbKrNWeIL4cKb9zU/edit#

short version:

1. Alpha fields should be first-class fields, not annotations (annotations allow "time-bomb" data persisted against a previous version of the apiserver, and are impossible to migrate to first-class fields in a coherent way)
2. Alpha fields should be optional, omitempty, and should not be sent by the client and should be rejected by the server unless the alpha feature is enabled. That ensures that by default, we do not enable persisting the alpha data without cluster admin knowledge. The linked proposal discusses "field gates" as a mechanism for doing that, but those have not been implemented yet, so for now, we're just rejecting in validation based on a feature gate.
3. If the schema of an alpha field changes, it must do so in a way that is compatible with previously released versions of that field, or a new field name must be chosen.

msau42:
@ianchakeres just like this proposal is not handling online resizing of filesystems while a pod is using it, there is also similar concern about doing an online resize of a block volume when an pod is directly accessing the raw block device (when we start supporting raw block access).  Of course, if this is needed, we could also allow it and just say that they better make sure their application will be ok with an online resize.

contributors/design-proposals/grow-volume-size.md
gnufied: Yes, I looked at block storage proposal in some detail now and I think since it adds an addtional `VolumeType` field to PVC spec - I would assume at plugin level we would have capability to decide if we are mounting a raw block device or a formatted block device. That flag could be passed as an parameter or something else I assume.

For now, we would just not support raw block devices.  I will keep an eye on raw block storage proposal to make sure this is handled correctly.

gnufied:
@msau42 @ianchakeres As I mentioned before, when Pods would be able to consume raw block devices - the information that volume is of type unformatted block device will be available somewhere in the plugin code. We would use this information to differentiate between formatted block device and unformatted raw block device. 

Volume types like EBS, GCE-PD would be resizable  as long as they have file system on it. We just need to make sure that raw block devices consumable within pods as defined in spec - https://github.com/kubernetes/community/pull/805/files remain non-resizable.

saad-ali:
> > This is still concerning to me because when you introduce a new first class field in the v1 API, it can not be removed. The idea with alpha features is that we should be able to completely remove them without breaking our backwards compatibility promises. @kubernetes/sig-api-machinery-misc any suggestions: in the past we would use annotations to test new fields before adding them; apparently that is not recommended anymore? If so, what is the recommended way to add alpha fields?
> 
> discussion in https://github.com/kubernetes/kubernetes/issues/34508 and https://docs.google.com/document/d/1wuoSqHkeT51mQQ7dIFhUKrdi3-1wbKrNWeIL4cKb9zU/edit#
> 
> short version:
> 
> 1. Alpha fields should be first-class fields, not annotations (annotations allow "time-bomb" data persisted against a previous version of the apiserver, and are impossible to migrate to first-class fields in a coherent way)
> 2. Alpha fields should be optional, omitempty, and should not be sent by the client and should be rejected by the server unless the alpha feature is enabled. That ensures that by default, we do not enable persisting the alpha data without cluster admin knowledge. The linked proposal discusses "field gates" as a mechanism for doing that, but those have not been implemented yet, so for now, we're just rejecting in validation based on a feature gate.
> 3. If the schema of an alpha field changes, it must do so in a way that is compatible with previously released versions of that field, or a new field name must be chosen.

@liggitt Requiring backwards compatibility with alpha seems counter to the purpose of alpha. I should be able to throw away alpha functionality or completely redesign it in subsequent releases without penalty.

liggitt:
> @liggitt Requiring backwards compatibility with alpha seems counter to the purpose of alpha. I should be able to throw away alpha functionality or completely redesign it in subsequent releases without penalty.

You can, you just can't reuse the same field name.

saad-ali:
> You can, you just can't reuse the same field name.

Ack. So depreciation of alpha fields is allowed. That still requires more certainty for an alpha feature right (i.e. be sure that this field's schema is defined the way you intend to use it, otherwise you could "waste" a good field name)? I guess we have to draw the line somewhere, to prevent breaking existing clients thought. So, ok. I'm fine with this approach.

jolestar:
Can volume support auto resize ?  it also a kind of auto scale.

jsafrane:
@jolestar Probably not in alpha / beta release, personally I'd like us to take small steps.

Once we have API for resize, then anyone can use it and I can imagine some (external?) controller that does some autoresize magic.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,301 @@
...
+For volume type that requires both file system expansion and a volume plugin based modification, growing persistent volumes will be two
+step process.
+
+
+For volume types that only require volume plugin based api call, this will be one step process.
+
+### Prerequisite
+
+* `pvc.spec.resources.requests.storage` field of pvc object will become mutable after this change.
+* #sig-api-machinery has agreed to allow pvc's status update from kubelet as long as pvc and node relationship
```
wongma7: @liggitt did we ever resolve this, how to validate a PVC-node relationship? If not, then FSResize can't update the PVC.status.capacity. For alpha release, I think we have to settle for an uglier solution. For example, what if the kubelet sent a special "FSResizeDone" Event that the Expand Controller watched for, and the Expand Controller updated the PVC on kubelet's behalf. The worst that could happen is PVC is updated before it's actually had its FS resized if a bad event were sent.

liggitt: Event delivery is not reliable and should not be used as a message bus. The node-pod-pvc connection seemed reasonable, especially if the node's update was limited to the pvc status stanza

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,301 @@
...
+    pvcWithResizeRequest *expandcache.PvcWithResizeRequest,
+    dsow expandcache.DesiredStateOfWorld) (func() error, error) {
+
+    volumePlugin, err := og.volumePluginMgr.FindExpandablePluginBySpec(pvcWithResizeRequest.VolumeSpec)
+
+    if err != nil {
+        return nil, fmt.Errorf("Error finding plugin for expanding volume: %q with error %v", pvcWithResizeRequest.UniquePvcKey(), err)
+    }
+
+    expanderPlugin, err := volumePlugin.NewExpander()
```
wongma7: imo we should pass in spec to be consistent with NewDeleter & Delete, so that the expander can store relevant spec fields in itself

wongma7: this means not passing it in ExpandVolumeDevice, to be clear

contributors/design-proposals/grow-volume-size.md
liggitt: https://github.com/kubernetes/kubernetes/pull/51202 is an example of using the node->pod->pvc relationship to gate status update access

contributors/design-proposals/grow-volume-size.md
gnufied: yeah that is fine as well. We can take some of the implementation details with pinch of salt.

gnufied:
@saad-ali @msau42 do we have any other pending object for this design? can we merge it?

cc @childsb

msau42:
I would add a line in the non-goals that we won't be supporting resize of raw block devices for now.  Otherwise, lgtm

gnufied:
@msau42 done.

contributors/design-proposals/grow-volume-size.md
```diff
@@ -0,0 +1,310 @@
...
+   LastTransitionTime metav1.Time
+   Reason string
+   Message string
+}
+
+// new API type
+type PVCConditionType string
+
+// new Constants
+const (
```
gnufied: @jsafrane provided a feedback based on actual implementation that, we need only one Condition not three. We just need a single `ResizeInProgress` (or called `ResizeStarted` ) condition and once resize is finished we can delete the Condition from pvc. 

All messaging to user can be provided via events rather Conditions. I don't massively disagree and we can drop other Conditions since errors are also available via events too.

saad-ali: I'm ok with this

childsb:
this has gone a few iterations of review and is in the final stages of impl.

/lgtm

PR #658, opened 2017-05-24
support coredump in kubernetes cluster

CaoShuFeng:
This proposal aims at managing coredump files generated in cluster containers.


mikedanese:
@kubernetes/sig-node-feature-requests

CaoShuFeng:
@yujuhong 
@thockin 
Can you take a look at this pull request?
Thanks!

smarterclayton:
In general these are very specific integrations (other debugging tools might have similar requirements, like snapshotting a particular file inside the container, or grabbing a TAR of the contents of the top layer).  It seems like we won't be able to effectively add support for each of these to the API without making Kubernetes end up with a lot of complexity.  

Is there a way to make this proposal not require changes to the Kube API, but instead to allow a privileged pod running on the node to capture this and report it via a different channel?  Making coredumps available to apps seems like a cluster wide capability, similar to logging.  Historical logging today is handled outside of the Kubernetes API, rather than in the Kube API.

fabiand:
It looks like this could be close to the problems addressed with https://github.com/kubernetes/node-problem-detector

CaoShuFeng:
@fabiand I opened a issue in node-problem-detector to ask whether it's possible to add this coredump detector to that add-on.

CaoShuFeng:
@fabiand @smarterclayton 
Thanks for your review.

dchen1107:
Node-problem-detector (NPD) is designed to detect hardware, kernel, filesystem, etc. issues, which requires the remedy / repair actions from the cluster admin or auto management system. It is not designed for debugging application issues. Also NPD reports those problems to kube-apiserver, which defeats the purpose to report the applications' core dumps via a different channel commented by @smarterclayton at https://github.com/kubernetes/community/pull/658#issuecomment-312353397

dchen1107:
@CaoShuFeng Instead, you can build your own daemonset / privileged pod to achieve this.

fabiand:
@CaoShuFeng My bad - @dchen1107 reasoning is good, my mistake was that I did not see the purpose of the node problem detector it as clear as I see it now.

PR #659, opened 2017-05-24
Redesign mount propagation

### LINKED ISSUE #648: Proposed mount propagation won't work with Docker
[propagation.md](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/propagation.md) says that we should "Make HostPath shared for privileged containers, slave for non-privileged."

I hacked this into kubelet (inspired by https://github.com/kubernetes/kubernetes/pull/41683) and tried to build a container that would mount iSCSI, CephRBD and Gluster volumes without having `iscsiadm`, `rbd` and `mount.glusterfs` installed on the node.

The resulting pod looks like:

```
      spec:
        hostNetwork: true
        containers:
         - name: mounter
           image: jsafrane/mounter-daemonset:latest
           securityContext:
             privileged: true
           volumeMounts:
             - name: kubelet
               mountPath: /var/lib/kubelet
             - name: sys
               mountPath: /sys
             - name: dev
               mountPath: /dev
             - name: iscsi
               mountPath: /etc/iscsi
             - name: iscsilock
               mountPath: /run/lock/iscsi
             - name: modules
               mountPath: /lib/modules
        volumes:
          - name: kubelet
            hostPath:
              path: /var/lib/kubelet
          - name: sys
            hostPath:
              path: /sys
          - name: dev
            hostPath:
              path: /dev
          - name: iscsi
            hostPath:
              path: /etc/iscsi
          - name: iscsilock
            hostPath:
              path: /run/lock/iscsi
          - name: modules
            hostPath:
              path: /lib/modules
```

What the pod really needs with shared mount propagation is `/var/lib/kubelet`. All the rest **should** be either rslave or private and it breaks things when it's rshared:

* Docker creates `/dev/shm` when running the container. `/dev/` is shared between the host and the container -> anything that the host had in `/dev/shm` is lost. In addition, docker does not unmount it from some reason when the container dies. Private or rshared mount propagation would be enough for `/dev`.
* I need `/sys` in the container to be able to talk to Ceph kernel module, however systemd in the container can't work with `/sys/fs/cgroup` as shared. It simply refuses to start. I need systemd to kill reap zombies of fuse daemons and to start NFS client daemons and `iscsid` during container startup. Now I am stuck with non-systemd init. Private or rshared mount propagation would be enough for `/sys`.
* I have PR to make `/var/lib/kubelet/ as shared during kubelet startup (https://github.com/kubernetes/kubernetes/pull/45724). This won't be enough, because now I need `/etc`, `/dev`, `/sys` and `/run` as shared too.

IMO, exporting a HostPath as shared should be opt-in per VolumeHost (or VolumeMount), it should not be enabled by default for all HostPath volumes as agreed in the proposal.

Adding random people who were active in mount propagation PRs:
@euank @lpabon @lucab @thockin @ivan4th @vishh @lvlv 
@kubernetes/sig-node-proposals @kubernetes/sig-storage-proposals

jsafrane:
The proposal won't work as it was merged, it makes too many directories as shared (see #648).

A different approach is needed, I've chosen 'Add an option in VolumeMount API', but I would be fine also with 'Add an option in HostPathVolumeSource', there is only very small difference to me.

The proposal also describes how it will be implemented, especially during alpha phase.

Fixes #648

@kubernetes/sig-node-proposals @kubernetes/sig-storage-proposals

### DIFF
--- a/contributors/design-proposals/propagation.md
+++ b/contributors/design-proposals/propagation.md
@@ -65,10 +65,25 @@ and references to network namespaces persist.
 The new `VolumeMount` will look like:
 
 ```go
+type MountPropagationMode string
+
 const (
-	MountPropagationRShared  MountPropagationMode = "RShared"
-	MountPropagationRSlave   MountPropagationMode = "RSlave"
-	MountPropagationPrivate  MountPropagationMode = "Private"
+	// MountPropagationPrivate means that the volume in a container won't receive
+	// new mounts from host and also no mounts in this volume will be propagated
+	// to the host.
+	MountPropagationPrivate MountPropagationMode = "private"
+	// MountPropagationSlave means that the volume in a container will receive
+	// new mounts from the host or other containers, but its own mounts won't
+	// be propagated to the host or other containers.
+	// Note that this mode is recursively applied to all mounts in the volume
+	// ("rslave" in Linux terminology)
+	MountPropagationSlave  MountPropagationMode = "slave"
+	// MountPropagationShared means that the volume in a container will receive
+	// new mounts from the host or other containers, and its own mounts will
+	// be propagated from the container to the host or other containers.
+	// Note that this mode is recursively applied to all mounts in the volume
+	// ("rshared" in Linux terminology)
+	MountPropagationShared MountPropagationMode = "shared"
 )
 
 type VolumeMount struct {
@@ -78,13 +93,18 @@ type VolumeMount struct {
 	ReadOnly bool `json:"readOnly,omitempty"`
 	// Required.
 	MountPath string `json:"mountPath"`
+	// mountPropagation is the propagation mode how is the directory
+	// made available to a container in a pod. See description of individual
+	// modes for details.
+	// This field is alpha in 1.8 and can be reworked or removed in a future
+	// release.
 	// Optional.
-	Propagation PropagationMode `json:"propagation,omitempty"`
+	MountPropagation MountPropagationMode `json:"mountPropagation,omitempty"`
 }
 ```
 
-Default would be `RSlave`, which should not break backward compatibility,
-`RShared` must be explicitly requested.
+Default would be `Slave`, which should not break backward compatibility,
+`Shared` must be explicitly requested.
 
 Opinion against this:
 
@@ -99,22 +119,42 @@ propagation could be rejected.
 The new `HostPathVolumeSource` will look like:
 
 ```go
+type MountPropagationMode string
+
 const (
-	MountPropagationRShared  MountPropagationMode = "RShared"
-	MountPropagationRSlave   MountPropagationMode = "RSlave"
-	MountPropagationPrivate  MountPropagationMode = "Private"
+	// MountPropagationPrivate means that the volume in a container won't receive
+	// new mounts from host and also no mounts in this volume will be propagated
+	// to the host.
+	MountPropagationPrivate MountPropagationMode = "private"
+	// MountPropagationSlave means that the volume in a container will receive
+	// new mounts from the host or other containers, but its own mounts won't
+	// be propagated to the host or other containers.
+	// Note that this mode is recursively applied to all mounts in the volume
+	// ("rslave" in Linux terminology)
+	MountPropagationSlave  MountPropagationMode = "slave"
+	// MountPropagationShared means that the volume in a container will receive
+	// new mounts from the host or other containers, and its own mounts will
+	// be propagated from the container to the host or other containers.
+	// Note that this mode is recursively applied to all mounts in the volume
+	// ("rshared" in Linux terminology)
+	MountPropagationShared MountPropagationMode = "shared"
 )
 
 type HostPathVolumeSource struct {
 	Path string `json:"path"`
-	// Mount the host path with propagation mode specified. Docker only.
-	MountPropagation MountPropagationMode `json:"propagation,omitempty"`
+	// mountPropagation is the propagation mode how is the directory
+	// made available to a container in a pod. See description of individual
+	// modes for details.
+	// This field is alpha in 1.8 and can be reworked or removed in a future
+	// release.
+	// Optional.
+	MountPropagation MountPropagationMode `json:"mountPropagation,omitempty"`
 }
 ```
 
-The default mount propagation is `rslave`. Any HostPath can ask for `private`.
-Only privileged containers can use HostPath with `rshared` mount propagation -
-kubelet silently downgrades the propagation to `rslave` when running `rshared`
+The default mount propagation is `slave`. Any HostPath can ask for `private`.
+Only privileged containers can use HostPath with `shared` mount propagation -
+kubelet silently downgrades the propagation to `slave` when running `shared`
 HostPath in a non-privileged container.
 
 Opinion against this:
@@ -140,14 +180,14 @@ privileged, or we can introduce a new option in SecurityContext to control this.
 The propagation mode could be determined by the following logic:
 
 ```go
-// Environment check to ensure "rshared" is supported.
+// Environment check to ensure "shared" is supported.
 if !dockerNewerThanV110 || !mountPathIsShared {
 	return ""
 }
 if container.SecurityContext.Privileged {
-	return "rshared"
+	return "shared"
 } else {
-	return "rslave"
+	return "slave"
 }
 ```
 
@@ -175,9 +215,9 @@ and something prevents it from starting if `/sys` is shared.
 
 * We will take 'Add an option in HostPathVolumeSource API'
   * With an alpha feature gate in 1.8.
-  * Only privileged containers can use `rshared` mount propagation.
-    * When non-privileged container uses `rshared` HostPath, it silently
-	  downgrades it to `rslave`.
+  * Only privileged containers can use `shared` mount propagation.
+    * When non-privileged container uses `shared` HostPath, it silently
+	  downgrades it to `slave`.
 * Kubelet will make sure that at least `/var/lib/kubelet` can be share-able into
   containers and it will refuse to start if it's unsuccessful
   * kubernetes/kubernetes#45724
@@ -194,7 +234,7 @@ and something prevents it from starting if `/sys` is shared.
   `kubelet --feature-gates=MountPropagation=true`
   It will be used only for testing of volume plugins in e2e tests and
   Mount propagation may be redesigned or even removed in any future release.
-  * When the feature is enabled the default mount propagation will be `rslave`,
+  * When the feature is enabled the default mount propagation will be `slave`,
     which is different to current `private`. Extensive testing is needed!
 
 ## Extra Concerns
--- a/contributors/design-proposals/propagation.md
+++ b/contributors/design-proposals/propagation.md
@@ -57,6 +57,16 @@ survive we need to bind mount a dir from host mount namespace to container one
 with shared flag, so that all bind mounts are propagated across mount namespaces
 and references to network namespaces persist.
 
+1. (From https://github.com/kubernetes/kubernetes/issues/46643) I expect the
+   container to start and any fuse mounts it creates in a volume that exists on
+   other containers in the pod (that are using :slave) are available to those
+   other containers.
+
+   In other words, two containers in the same pod share an EmptyDir. One
+   container mounts something in it and the other one can see it. The first
+   container must have (r)shared mount propagation to the EmptyDir, the second
+   one can have (r)slave.
+
 
 ## Implementation Alternatives
 
@@ -68,22 +78,19 @@ The new `VolumeMount` will look like:
 type MountPropagationMode string
 
 const (
-	// MountPropagationPrivate means that the volume in a container won't receive
-	// new mounts from host and also no mounts in this volume will be propagated
-	// to the host.
-	MountPropagationPrivate MountPropagationMode = "private"
-	// MountPropagationSlave means that the volume in a container will receive
-	// new mounts from the host or other containers, but its own mounts won't
-	// be propagated to the host or other containers.
+	// MountPropagationHostToContainer means that the volume in a container will
+	// receive new mounts from the host or other containers, but filesystems
+	// mounted inside the container won't be propagated to the host or other
+	// containers.
 	// Note that this mode is recursively applied to all mounts in the volume
-	// ("rslave" in Linux terminology)
-	MountPropagationSlave  MountPropagationMode = "slave"
-	// MountPropagationShared means that the volume in a container will receive
-	// new mounts from the host or other containers, and its own mounts will
-	// be propagated from the container to the host or other containers.
+	// ("rslave" in Linux terminology).
+	MountPropagationHostToContainer  MountPropagationMode = "HostToContainer"
+	// MountPropagationBidirectional means that the volume in a container will
+	// receive new mounts from the host or other containers, and its own mounts
+	// will be propagated from the container to the host or other containers.
 	// Note that this mode is recursively applied to all mounts in the volume
-	// ("rshared" in Linux terminology)
-	MountPropagationShared MountPropagationMode = "shared"
+	// ("rshared" in Linux terminology).
+	MountPropagationBidirectional MountPropagationMode = "Bidirectional"
 )
 
 type VolumeMount struct {
@@ -93,18 +100,24 @@ type VolumeMount struct {
 	ReadOnly bool `json:"readOnly,omitempty"`
 	// Required.
 	MountPath string `json:"mountPath"`
-	// mountPropagation is the propagation mode how is the directory
-	// made available to a container in a pod. See description of individual
-	// modes for details.
+	// mountPropagation is the mode how are mounts in the volume propagated from
+	// the host to the container and from the container to the host.
+	// When not set, MountPropagationHostToContainer is used.
 	// This field is alpha in 1.8 and can be reworked or removed in a future
 	// release.
 	// Optional.
-	MountPropagation MountPropagationMode `json:"mountPropagation,omitempty"`
+	MountPropagation *MountPropagationMode `json:"mountPropagation,omitempty"`
 }
 ```
 
-Default would be `Slave`, which should not break backward compatibility,
-`Shared` must be explicitly requested.
+Default would be `HostToContainer`, i.e. `rslave`, which should not break
+backward compatibility, `Bidirectional` must be explicitly requested.
+Using enum instead of simple `PropagateMounts bool` allows us to extend the
+modes to `private` or non-recursive `shared` and `slave` if we need so in
+future.
+
+Only privileged containers are allowed to use `Bidirectional` for their volumes.
+This will be enforced during validation.
 
 Opinion against this:
 
@@ -122,46 +135,54 @@ The new `HostPathVolumeSource` will look like:
 type MountPropagationMode string
 
 const (
-	// MountPropagationPrivate means that the volume in a container won't receive
-	// new mounts from host and also no mounts in this volume will be propagated
-	// to the host.
-	MountPropagationPrivate MountPropagationMode = "private"
-	// MountPropagationSlave means that the volume in a container will receive
-	// new mounts from the host or other containers, but its own mounts won't
-	// be propagated to the host or other containers.
+	// MountPropagationHostToContainer means that the volume in a container will
+	// receive new mounts from the host or other containers, but filesystems
+	// mounted inside the container won't be propagated to the host or other
+	// containers.
 	// Note that this mode is recursively applied to all mounts in the volume
-	// ("rslave" in Linux terminology)
-	MountPropagationSlave  MountPropagationMode = "slave"
-	// MountPropagationShared means that the volume in a container will receive
-	// new mounts from the host or other containers, and its own mounts will
-	// be propagated from the container to the host or other containers.
+	// ("rslave" in Linux terminology).
+	MountPropagationHostToContainer  MountPropagationMode = "HostToContainer"
+	// MountPropagationBidirectional means that the volume in a container will
+	// receive new mounts from the host or other containers, and its own mounts
+	// will be propagated from the container to the host or other containers.
 	// Note that this mode is recursively applied to all mounts in the volume
-	// ("rshared" in Linux terminology)
-	MountPropagationShared MountPropagationMode = "shared"
+	// ("rshared" in Linux terminology).
+	MountPropagationBidirectional MountPropagationMode = "Bidirectional"
 )
 
 type HostPathVolumeSource struct {
 	Path string `json:"path"`
-	// mountPropagation is the propagation mode how is the directory
-	// made available to a container in a pod. See description of individual
-	// modes for details.
+	// mountPropagation is the mode how are mounts in the volume propagated from
+	// the host to the container and from the container to the host.
+	// When not set, MountPropagationHostToContainer is used.
 	// This field is alpha in 1.8 and can be reworked or removed in a future
 	// release.
 	// Optional.
-	MountPropagation MountPropagationMode `json:"mountPropagation,omitempty"`
+	MountPropagation *MountPropagationMode `json:"mountPropagation,omitempty"`
 }
 ```
 
-The default mount propagation is `slave`. Any HostPath can ask for `private`.
-Only privileged containers can use HostPath with `shared` mount propagation -
-kubelet silently downgrades the propagation to `slave` when running `shared`
-HostPath in a non-privileged container.
+Default would be `HostToContainer`, i.e. `rslave`, which should not break
+backward compatibility, `Bidirectional` must be explicitly requested.
+Using enum instead of simple `PropagateMounts bool` allows us to extend the
+modes to `private` or non-recursive `shared` and `slave` if we need so in
+future.
+
+Only privileged containers can use HostPath with `Bidirectional` mount
+propagation - kubelet silently downgrades the propagation to `HostToContainer`
+when running `Bidirectional` HostPath in a non-privileged container. This allows
+us to use the same `HostPathVolumeSource` in a pod with two containers, one
+non-privileged with `HostToContainer` propagation and second privileged with
+`Bidirectional` that mounts stuff for the first one.
 
 Opinion against this:
 
 1. This need API change, which is discouraged.
 
 1. All containers use this volume will share the same propagation mode.
+
+1. Silent downgrade from `Bidirectional` to `HostToContainer` for non-privileged
+   containers.
 
 1. (From @jonboulle) May cause cross-runtime compatibility issue.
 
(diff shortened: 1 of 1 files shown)


jsafrane:
For alpha annotation, I know there is attempt to have real fields in alpha in https://docs.google.com/document/d/1wuoSqHkeT51mQQ7dIFhUKrdi3-1wbKrNWeIL4cKb9zU/edit#, but it seems it has not been implemented yet. Alpha annotation is still described as preferred way to introduce new fields in https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md

yujuhong:
Adding a couple of folks who were on the original proposal.

/cc @vishh @euank

dchen1107:
We discussed at sig-node meeting, and agreed with the redesign the proposal scope. But this feature is not targeted for 1.7 release.

inc0:
This is required change for openstack-on-kubernetes too. Could we please add sig-openstack to this review?

kfox1111:
+1 for sig-openstack.

I would also greatly benefit from it for some high energy physics work I do.

If the patch set is ready for 1.7 will it go in?

Is there anything I can do to help get it ready faster?

ddysher:
This is also required for local volume provisioner.  The provisioner is running as a daemonset to monitor local volume and post PV to kubernetes.  To correctly detect volume capacity, the daemonset pod must mount hostpath (path where volume is detected) as rshared.

Are we targeting alpha in 1.8 now?

jsafrane:
yes, we missed 1.7

ddysher:
Thanks for confirm.  I'd be happy to test it out when implementation is out.

jsafrane:
I updated the PR with the latest development:

- added links to Kubernetes PRs (waiting for review)
- added note that the default propagation should be `rslave`. This is what was agreed in the previous proposal, now it's more explicit.
- I tested my PRs with various distros and I stumbled upon Google COS (former GCI). It runs Docker in its own mount namespace with slave mount propagation. I.e. it can't ever run a pod with shared mount propagation. There should be a check in kubelet that tests if it can run a pod with shared mount propagation, at least on `/var/lib/kubelet`, and refuse to start if it can't.
  - This test is quite tricky, the best would be to run some dummy container (`gcr.io/*/pause`?) with `docker run -v /var/lib/kubelet:/tmp:rshared` and check if it failed.
  - Note that COS-60 fixes this issue, however the test might be useful if someone runs another old or ~~weird~~ unsupported distro.

contributors/design-proposals/propagation.md
```diff
@@ -142,13 +153,41 @@ distros.
...
+  anything. Other container engines should follow the suit.
+* Node conformance suite will check that mount propagation in /var/lib/kubelet
+  works.
+* During alpha, all the behavior above must be explicitly enabled by
+  `kubelet --test-enable-mount-propagation`
+  It will be used only for testing of volume plugins in e2e tests and
+  it will be marked as deprecated from the beginning.
+  Developers / testers can enable it in their clusters manually.
+  Mount propagation may be redesigned or even removed in any future release.
+* The default mount propagation will be `rslave`, which is different to current
```
msau42: Earlier, it says the default is `private`, to not break backwards compatiblity. Why is it `rslave` here?

jsafrane: thanks for spotting this, the default should be rslave, which *should* not break backward compatibility (testing needed!) and enables this use case https://github.com/kubernetes/kubernetes/pull/46444#issuecomment-304439683

kfox1111:
Ping. What is the status of getting this into 1.8? I don't see it marked for 1.8.

I'm stuck on k8s 1.4 currently, as I need this kind of functionality and docker 1.9 is the last version to default mounts to shared and k8s 1.4 is the last version to support docker 1.9.

I would be happy if this was merged but hidden behind an experimental flag or something, just so those of us that need to make forward progress can do so.

kfox1111:
ping

luxas:
@jsafrane It would indeed be nice to have this as alpha in v1.8

jsafrane:
@luxas, all that's missing is review of #46444

saad-ali:
/lgtm
/approve 

We discussed this design in a meeting on Monday July 24. I am fine with this.

One thing I just remembered is that there is a pending design to add "HostPathType" to hostpath volumesource: https://github.com/kubernetes/kubernetes/pull/46597
Let's make sure that these two options (propagation and hostPathType play well together)

CC @timstclair @dixudx

dixudx:
@saad-ali `propagation` will not conflict with `HostPathType`.

jsafrane:
@timstclair was there any discussion about feature gate vs --exeprimental cmdline argument? I am happy with both, I just need a decision.

contributors/design-proposals/propagation.md
```diff
@@ -142,13 +153,41 @@ distros.
...
+Similarly, systemd running in a container is very picky about `/sys/fs/cgroup`
+and something prevents it from starting if `/sys` is shared.
 
 ## Decision
 
-We will take 'Make HostPath shared for privileged containers, slave for
-non-privileged', an environment check and an WARNING log will be emitted about
-whether propagation mode is supported.
-
+* We will take 'Add an option in VolumeMount API' (with an annotation during
```
tallclair: I disagree with this decision. I think this should be added to the HostPathVolumeSource, since it only applies to HostPath volumes. I think it's a relatively small edge case that there are multiple containers in the same pod mounting the same HostPath volume, where one wants it to be shared and the other wants slave. In that case, if one of the containers was not privileged, it would get the non-shared mount anyway.

contributors/design-proposals/propagation.md
```diff
@@ -142,13 +153,41 @@ distros.
...
+* Kubelet will make sure that at least `/var/lib/kubelet` can be share-able into
+  containers and it will refuse to start if it's unsuccessful
+  * kubernetes/kubernetes#45724
+* Kubelet's Docker shim layer will check that it is able to run a container with
+  shared mount propagation on `/var/lib/kubelet` during startup and refuse to
+  start otherwise. This ensures that both Docker and kubelet see the same
+  `/var/lib/kubelet` and it can be shared into containers.
+  E.g. Google COS-58 runs Docker in a separate mount namespace with slave
+  propagation and thus can't run a container with shared propagation on
+  anything. Other container engines should follow the suit.
```
tallclair: I'm confused by this. Isn't this saying that containers with shared mounts cannot be run on COS, which would prevent the Kubelet from starting under this new model?

contributors/design-proposals/propagation.md
```diff
@@ -142,13 +153,41 @@ distros.
...
+* Node conformance suite will check that mount propagation in /var/lib/kubelet
+  works.
+* During alpha, all the behavior above must be explicitly enabled by
+  `kubelet --experimental-enable-mount-propagation`
```
tallclair: Use a feature gate instead, with a new "experimental" release track (https://github.com/kubernetes/kubernetes/blob/9af1ff3dce1f71f31c051b5b84b6c187ebdc6476/staging/src/k8s.io/apiserver/pkg/util/feature/feature_gate.go#L64)

contributors/design-proposals/propagation.md
```diff
@@ -142,13 +153,41 @@ distros.
...
+and something prevents it from starting if `/sys` is shared.
 
 ## Decision
 
-We will take 'Make HostPath shared for privileged containers, slave for
-non-privileged', an environment check and an WARNING log will be emitted about
-whether propagation mode is supported.
-
+* We will take 'Add an option in VolumeMount API' (with an annotation during
+  alpha instead of real VolumeMount field):
```
liggitt: -1 on alpha annotation, feature-gated API field is preferred

jsafrane: I do know about https://docs.google.com/document/d/1wuoSqHkeT51mQQ7dIFhUKrdi3-1wbKrNWeIL4cKb9zU/, but it seems to me as it's work in progress.

* I haven't noticed any announce of new API style on kubernetes-devel

* https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#adding-unstable-features-to-stable-versions still recommends alpha fields for unstable features.

* Nobody uses `// +gate` comment in any `types.go` in Kubernetes

contributors/design-proposals/propagation.md
jsafrane: COS 58 runs docker 1.12 and its systemd service runs docker daemon it its own mount namespace with slave mount propagation from the host. So a container can never export anything to the host, only to the docker mount namespace. This is fixed in COS 60 with docker 1.13.

contributors/design-proposals/propagation.md
jsafrane: There were people who wound like to run EmptyDir with shared mount propagation so a pod could have two containers, one privileged one that mounts something into the EmptyDir so the second container can consume it. https://github.com/kubernetes/kubernetes/pull/46444#issuecomment-310857111

IMO it's an useful use case and we could implement it when going out of alpha.

contributors/design-proposals/propagation.md
jsafrane: ok, switched to feature gate

contributors/design-proposals/propagation.md
jsafrane: Discussed with @liggitt elsewhere, I'm dropping the annotation and real field with a feature gate will be used.

contributors/design-proposals/propagation.md
tallclair: I'm just confused by the `Other container engines should follow the suit.` line... I think what you're trying to say is:
```
Google COS-60 and above support shared propagation, and other Kubernetes node OSs should follow suit.
```

contributors/design-proposals/propagation.md
tallclair: There are several fields [1] that apply to multiple VolumeSources, and the fields are replicated amongst the applicable sources. On the other hand, all the fields in VolumeMount are universally applicable to all volumes. I think it would be more consistent in this case to just include the propagation mode in both the HostPath and EmptyDir VolumeSources.

[1] FSType, ReadOnly, DefaultMode

contributors/design-proposals/propagation.md
jsafrane: I have very limited knowledge about rkt and its support for mount propagation. It should follow Docker's model.

contributors/design-proposals/propagation.md
jsafrane: Ok, that makes sense. On the other hand, with `MountPropagation` as `HostPathVolumeSource` field we can't validate if container is privileged when running with a HostPath with shared mount propagation - propagation is in a PersistentVolume object and privileged mode is set in Pod and validator sees only one of these objects. Last time I checked we did not do cross-object validation.

jsafrane:
Just to make it clear, if we go with `HostPathVolumeSource`:

```go
type HostPathVolumeSource struct {
	Path string `json:"path"`
	// Mount the host path with propagation mode specified.
        Propagation PropagationMode `json:"propagation,omitempty"`
}
```

`HostPathVolumeSource.Propagation` is **string** and not **pointer to a string**, right? It's empty when omitted. Should it have a default value if it's alpha?

contributors/design-proposals/propagation.md
tallclair: So it's a bit weird, but my proposal is that it should be OK to have a shared hostpath volume mounted into an unprivileged container, but that container doesn't get shared (it gets slave). This let's you have the state where there are 2 containers in a pod, and only one gets the shared propagation. I suppose you could validate that if it's set to shared, there must be at least 1 privileged container, but I think it's ok without that. We should make sure this behavior is documented on the field though.

jsafrane:
Pushed new version

- `struct Propagation` renamed to `MountPropagation`, who knows what other kinds of propagation will be there
- moved to `HostPathVolumeSource`

I hope it's the final version.

contributors/design-proposals/propagation.md
```diff
@@ -89,18 +100,23 @@ The new `HostPathVolumeSource` will look like:
...
 
 type HostPathVolumeSource struct {
 	Path string `json:"path"`
 	// Mount the host path with propagation mode specified. Docker only.
-	Propagation PropagationMode `json:"propagation,omitempty"`
+	MountPropagation MountPropagationMode `json:"propagation,omitempty"`
 }
 ```
 
+The default mount propagation is `rslave`. Any HostPath can ask for `private`.
```
thockin: Why would anyone ask for private?  What value does the extra surface area provide?

contributors/design-proposals/propagation.md
```diff
@@ -73,13 +79,18 @@ type VolumeMount struct {
...
 }
 ```
 
+Default would be `RSlave`, which should not break backward compatibility,
+`RShared` must be explicitly requested.
+
 Opinion against this:
 
-1. This will affect all volumes, while only HostPath need this.
+1. This will affect all volumes, while only HostPath need this. It could be
```
thockin: This is a really nasty implication.

contributors/design-proposals/propagation.md
```diff
@@ -109,6 +125,12 @@ Opinion against this:
 
 1. (From @jonboulle) May cause cross-runtime compatibility issue.
 
+1. It's not possible to validate a pod + mount propagation. Mount propagation
```
thockin: The validators are just code - we can change this, or we can add another validation at the end of ValidatePodSpec() that cross-checks privileged concerns, right?

contributors/design-proposals/propagation.md
```diff
@@ -142,13 +164,38 @@ distros.
...
 ## Decision
 
-We will take 'Make HostPath shared for privileged containers, slave for
-non-privileged', an environment check and an WARNING log will be emitted about
-whether propagation mode is supported.
-
+* We will take 'Add an option in HostPathVolumeSource API'
+  * With an alpha feature gate in 1.8.
+  * Only privileged containers can use `rshared` mount propagation.
+    * When non-privileged container uses `rshared` HostPath, it silently
```
thockin: silent failure is bad - we should just fail

contributors/design-proposals/propagation.md
thockin: I am confused by this too.  Does this mean that this model can share mounts between docker containers (shared would propagate up to docker's mountns) but not up to the host OS?

contributors/design-proposals/propagation.md
jsafrane: At the time of validation of a pod the PVCs it uses may not be bound to a PVs yet, so there is nothing to validate. Binding to PVs can happen later.

contributors/design-proposals/propagation.md
jsafrane: Yes, containers can shared mounts among each other (and to docker mount NS), but kubelet won't see these mounts. Fixed in Docker 1.13 and I think 1.12 is going to be deprecated when Kubernetes 1.8 is released.

contributors/design-proposals/propagation.md
jsafrane: Do you mean setting `PodStatus.Phase: Failed`? Well... I can do it (if someone shows me the right place in kubelet), however it looks too offensive to me.

contributors/design-proposals/propagation.md
jsafrane: Dunno, it's just safety hatch when something breaks. Maybe a pod that uses whole "/" as HostPath may not want to get further mounts from the host.

jsafrane:
> What if we defined a whole new volume type that was abstractly "a place to mount volumes" - and we handle it how we need in kubelet

I thought that's what `struct VolumeSource` is for.

On sig-storage, we have embargo on new volume plugins and creating a new one just to expose extra Linux flags does not sound good to me. I would not expect that PVs are portable at all, we put too much Linux stuff to them already - e.g. mount options.

kfox1111:
Uh, last I looked the distro's were only shipping 1.12 still?

jsafrane:
@kfox1111 what distros? I know RHEL/CentOS 7 and COS (or GCI) < 60 ships 1.12. COS 59 is going to die soon-ish and I am working on RHEL 7.

kfox1111:
el7 in particular I guess. I didn't think redhat was going to support 1.13? You probably know more about that then I do though.

contributors/design-proposals/propagation.md
thockin: I don't think we should design to hypotheticals, especially when it involves complex API surface.

contributors/design-proposals/propagation.md
thockin: Ahh, sorry, I missed that this was about PV.  We can reject at runtime, if we pursue this.  That said, HostPath as PV is still a terrible idea, and nobody should use it.

contributors/design-proposals/propagation.md
thockin: We need to define some sane behavior.  Kubelet could reject it, or just leave it pending with an event

contributors/design-proposals/propagation.md
kfox1111: isn't local volumes using it? pv generated with hostpath and injected attribute to pin it to a particular host?

thockin:
re: embargo on new volumes.  It has to be more nuanced than a blanket embargo.  Some volumes don't make sense being built-in, and some just do.  In this case, we're exposing a pretty complex API surface, with weird corner cases and failure modes, that barely we can really understand just to avoid a new volume?  No.

I would like to understand what the pros/cons of a new volume type would be.  Would it prevent anything that we know or strongly suspect we will need?  Would it need to be exposed as PV at all (I think not)?  Does it simplify the impl or make it more complex or neutral?  Is it easier or harder to document?

It is not my intent to drag my feet, but this is sort of important.

@saad-ali @jingxu97

contributors/design-proposals/propagation.md
kfox1111: I disagree only hostpath needs propagation. I totally could see it with at least emtpyDir too. Say I want to have a cvmfs sidecar to my main pod? I could use cvmfs with its automounter to mount cvmfs sumbmounts in the emptydir and expose the whole /cvmfs automounter to the other containers in the pod.

I still would really like to see a namedEmptyDir implementation that could use it too.

kfox1111:
@thockin Like I said above, I can see other volumes that may want to share the same functionality. I think adding another volume type would be a step in the wrong direction as it couldn't be shared between the other volume types.

kfox1111:
Also, several different groups are really hamstrung by the lack of this functionality. Can we please get it into 1.8 as alpha while we explore the way to do it 100% proper in beta in a following release? I'm currently stuck on k8s 1.4 because of this.

contributors/design-proposals/propagation.md
jsafrane: > HostPath as PV is still a terrible idea, and nobody should use it.

+1000

> isn't local volumes using it?

Local has its own volume type named `Local` (`LocalVolumeSource`)
￼

contributors/design-proposals/propagation.md
jsafrane: `Pending` + event is easy, I am not sure about rejection. Does it mean pod admission? Are all PVCs bound at this time? @tallclair @derekwaynecarr

contributors/design-proposals/propagation.md
kfox1111: oh. I didn't realize that. Does that mean the cool annotation thing only works with local volumes?

That would be a bummer. as the annotation could be useful in other ways, such as tagging a cinder based pv so that the pod it gets attached to only lands on vm's when you have a hybrid openstack/baremetal cluster (I do)

msau42: The PV node affinity is designed so that it could be used on other volume types, but right now it is behind the local volume feature gate.

msau42: But even then, it will not work out of the box for hostpath PVs.  There are limitations in how the hostpath volume plugin is implemented that prevent it from working in a multi-node cluster.

contributors/design-proposals/propagation.md
tallclair: The modivation for the silent downgrade is to cover the case where you have a Pod with a shared hostpath volume shared by 2 containers, 1 privileged that does the mounting, and 1 non-privileged that simply reads from the mounted volume. Since we decided to move the mount propagation setting into the volume definition (instead of the volume mount). I don't like making the behavior conditional on privileged either, but I think I prefer it to having the setting in the VolumeMount (since it only applies to a small subset of volumes).

thockin: I don't understand the use case.  Does it not work?

contributors/design-proposals/propagation.md
thockin: This is an interesting use case in its own, and if we want to consider emptyDir with propagation, we need to pay extra attention to cleanup.  I have observed that this can lead to "leaked" mounts.  It definitely changes the calculus on this approach.

What is namedEmptyDir ?

thockin:
I am sorry that efforts are hamstrung, but this is real API surface.

contributors/design-proposals/propagation.md
jsafrane: Consider a pod that has two containers, one privileged "sidecar" and one unprivileged mysql. The sidecar one just mounts some volume for the mysql and unmounts it when the pod dies. If I add admission handler to kubelet, it will see a pod with an unprivileged container using a HostPath ( EmptyDir would be better in this case) with rshared propagation and it will reject it.

contributors/design-proposals/propagation.md
```diff
@@ -65,6 +65,12 @@ and references to network namespaces persist.
 The new `VolumeMount` will look like:
 
 ```go
+const (
+	MountPropagationRShared  MountPropagationMode = "RShared"
+	MountPropagationRSlave   MountPropagationMode = "RSlave"
```
bgrant0607: Abbreviation is discouraged.

jsafrane:
> I would like to understand what the pros/cons of a new volume type would be.

So, let there be `LinuxVolumeSource`:

```go
type LinuxVolume struct {
    MountPropagation string
    WrappedVolume *VolumeSource
}

type VolumeSource struct {
   ... aws, gce, hostPath, emptyDir, git, ...
    LinuxVolume *LinuxVolumeSource
}
```

1. It's recursive. Recursive structures in API are IMO ugly.
2. Does not help with application portability at all. If an application needed mount propagation it won't work on macOS / Windows at all, regardless if `LinuxVolume` or `HostPath` was used.
3. I do not see any added value other than increased complexity and confusion.

In my world, `VolumeSource` describes *where* to get the data from. I still personally think that it should not include things like `readOnly` or `mountPropagation`. It should be in `VolumeMount struct` which says *how* the data should be made available. It's much easier to validate it, it goes well along with `VolumeMount.readOnly`. I can accept that `HostPathVolumeSource` and `EmptyDirVolumeSource` is the right place, it will work well for all use cases that I know about. I don't see any reason why to invent a new volume type.

bgrant0607:
For reference:
https://lwn.net/Articles/689856/
https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt

kfox1111:
I agree. Properties on how something gets mounted is a totally different thing than what the thing storing the data is. hostPath, emptyDir, etc will need the flag. not, recreate volume types for rsharedHostPath and rsharedEmptyDir. Once namedEmptyDir becomes a thing, it will want it too.

kfox1111:
@thockin yeah, I get the desire for a correct api. Totally a good thing to get right. Just saying, for an alpha, getting it perfect is not so critical. I'm totally fine accepting the risk of writing software against an api that is expected to drastically change in future releases. That part is easy. Living without this feature is really really hard. I'm facing a 1.4 -> 1.8 migration path already, with broken steps in between. :/ Several of these releases are already unsupported.

As for openstack on k8s, it can't really be a complete thing until this feature merges in some form or another. Neutron really needs it.

contributors/design-proposals/propagation.md
bgrant0607: It doesn't have to be the case that adding the option to VolumeMount would imply that it would have to be supported for all volumes. The supported volumes could be documented on the field and whitelisted in validation.

contributors/design-proposals/propagation.md
bgrant0607: Is there a compelling reason why this couldn't be a simple boolean?

contributors/design-proposals/propagation.md
```diff
@@ -142,13 +164,38 @@ distros.
...
+  shared mount propagation on `/var/lib/kubelet` during startup and refuse to
+  start otherwise. This ensures that both Docker and kubelet see the same
+  `/var/lib/kubelet` and it can be shared into containers.
+  E.g. Google COS-58 runs Docker in a separate mount namespace with slave
+  propagation and thus can't run a container with shared propagation on
+  anything. Other container engines should follow the suit.
+* Node conformance suite will check that mount propagation in /var/lib/kubelet
+  works.
+* During alpha, all the behavior above must be explicitly enabled by
+  `kubelet --feature-gates=MountPropagation=true`
```
bgrant0607: ICYMI

Field gate proposal: https://github.com/kubernetes/community/pull/869

contributors/design-proposals/propagation.md
```diff
@@ -73,13 +79,18 @@ type VolumeMount struct {
 	// Required.
 	MountPath string `json:"mountPath"`
 	// Optional.
-	Propagation string `json:"propagation"`
+	Propagation PropagationMode `json:"propagation,omitempty"`
```
bgrant0607: This field name is very vague. So is the file name, by the way. Propagation of what? In what direction?

contributors/design-proposals/propagation.md
kfox1111: because there are three options at present. rslave, rshared, and private.

contributors/design-proposals/propagation.md
kfox1111: @thockin I've been toying around with the idea of contributing a 'namedEmptyDir' volume type. what it does, is similar to an emptyDir, but it is given a name unique to a namespace, and its lifecycle doesn't have it deleted until the last pod on the same host that references it goes away.
So, it would be stored somwhere like /var/lib/kubelet/volumes/namedEmptydir/<namespace name>/<name from user>/ (or whatever)

This would allow multiple k8s objects such as deamonsets or daemonsets/deployments/etc to coexist on the host and share some transient storage amongst them but without having to book keep where to store them using a hostpath and ensuring different name spaces don't stop on each others data.

Its use works in concert with this proposal in that say you have a fuse based system such as cvmfs that has a backing cache. You could have several different cvmfs daemonsets or pods consume the cache dir from a namedEmptyDir on the host via "cvmfs-cache" and have another one called "cvmfs" with mount propagation set to shared where cvmfs automounter is running.  Any pod in the namespace can then mount the namedEmptyDir 'cvmfs' and the the automount cvmfs volume would work optimally.

Ideally for security, you would have the volume mounted as shared on the autofs fuse filesystem container and have it slave on the things consuming it. Even more reason for having it as a flag and not a volumetype.

contributors/design-proposals/propagation.md
jsafrane: `MountPropagationRecursiveShared` sounds really ugly and it's not used nowhere in kernel / docker  documentation. Renamed to `Shared` with a note that it's applied recursively (imo it's still quite confusing...)

contributors/design-proposals/propagation.md
jsafrane: Thanks!

contributors/design-proposals/propagation.md
jsafrane: > This field name is very vague... Propagation of what? In what direction?

Renamed to `MountPropagation` with more documentation around. Still, I don't want to put whole kernel documentation there.

> So is the file name

What file name?

thockin:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/659)

jsafrane:
We had a productive meeting yesterday with @thockin and @saad-ali and we're back at "Add an option in VolumeMount API".

* While we do not see any use case for any other volume than HostPath or EmptyDir to be `rshared`, we will not block it, at least not in alpha. It must be explicitly enabled and having rshared GCE volume is not more harmful than HostPath.

* Renamed `(r)shared` and `(r)slave` in the API to `Bidirectional` and `HostToContainer`.

* Removed `private` as we do not see any use case for it.

* We use enum instead of simple bool to be able to add `private` and non-recursive `shared` and `slave` if we need so. We can't extend `bool` so easily.

I did more edits describing what exactly will be in alpha, see the last commit.

@bgrant0607, can you please look at the API again? PR with Kubernetes changes will follow very shortly to catch 1.8

saad-ali:
Revised proposal:

/lgtm
/approve

Thanks @jsafrane

contributors/design-proposals/propagation.md
bgrant0607: The proposal filename.

bgrant0607:
Github notifications are useless :-(

PR #663, opened 2017-05-25
Propose feature for kubectl unset subcommand

xilabao:
move from https://github.com/kubernetes/kubernetes/pull/46278
@kubernetes/sig-cli-proposals

### DIFF
--- a/contributors/design-proposals/kubectl-unset.md
+++ b/contributors/design-proposals/kubectl-unset.md
@@ -1,0 +1,81 @@
+# Kubectl Unset Subcommand
+
+## Goals
+
+`kubectl unset` is used to remove specific features from objects. To be contrasted with `kubectl set` subcommand,
+`unset` commands should be parallel with `set` commands. They aren't quite the same.
+
+## Design
+
+The `unset` subcommand helps users modify the existing application resources. Using `kubectl unset Object Resource`
+to remove features from `Object` of `Resource`(e.g. `kubectl unset subject rolebinding`).
+
+However, there is a big difference between `set` and `unset`. `set` commands will have more required fields than
+`unset` commands, since `unset` only needs to identify the "thing" to be removed, as opposed to defining what the
+"thing" is. In other ways, This has an impact on auto-complete, help, and validation.
+
+In addition, `kubectl unset` is not mutually exclusive to a `kubectl set --remove` or some such since the `--remove`
+could actually build the `kubectl unset` options.
+
+### Generic Code Template
+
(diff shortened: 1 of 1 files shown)


xilabao:
@smarterclayton @liggitt @deads2k @fabianofranz @adohe @pwittrock please take a look.

contributors/design-proposals/kubectl-unset.md
```diff
@@ -0,0 +1,70 @@
+# Kubectl Unset Subcommand
+
+## Goals
+
+`kubectl unset` is used to remove specific features from objects. To be contrasted with `kubectl set` subcommand, `unset` commands should has the same ability as `set` commands.
```
deads2k: unset commands should parallel set commands.  They aren't quite the same.

contributors/design-proposals/kubectl-unset.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `unset` subcommand helps users modify the existing application resources. Using `kubectl unset Object Resource` to remove features from `Object` of `Resource`(e.g. `kubectl unset subject rolebinding`).
+
+### Generic Code Template
+
+Now that `unset` subcmmand is highly similar to `set` subcommand, we could design a generic code template to reuse code.
+
+In set_object.go, the part to change object should be separated from options.Run(), and we should make it as a callback funtion for options.Run().
+
+For instance, we could pass setObjects() into options.Run(setObjects) in set_object.go. In the other side, we could create unsetObjects() as a callback funtion in unset_object.go.
```
deads2k: good thought.

contributors/design-proposals/kubectl-unset.md
```diff
@@ -0,0 +1,70 @@
+# Kubectl Unset Subcommand
+
+## Goals
+
+`kubectl unset` is used to remove specific features from objects. To be contrasted with `kubectl set` subcommand, `unset` commands should has the same ability as `set` commands.
+
+## Design
```
deads2k: spend some time in this section explaining why `unset` is different using some concrete examples.  Generally speaking, `set` commands will have more required fields than `unset` commands, since `unset` only needs to identify a thing to be removed, as opposed to defining what the the "thing" is.  This has an impact on auto-complete, help, and validation.

In addition, `kubectl unset` is not mutually exclusive to a `kubectl set --remove` or some such since the `--remove` could actually build the `kubectl unset` options.

deads2k:
I think the proposal needs some more detail around the alternatives we have and the advantages of this approach, but I'm generally in favor of the idea.  I certainly want the power, however we end up expressing it.

xilabao:
@fabianofranz ptal

PR #672, opened 2017-05-30
Update node authorization proposal

liggitt:
Updated the node authorization proposal based on implementation feedback:
* Recognition of nodes from userinfo is always strict... unless an incoming request has the `system:nodes` group AND a username in the format `system:node:<nodeName>`, the node authorizer and admission will ignore it
* Authorizer is opt-in as an authorization mode rather than a feature gate
* Admission is opt-in as an admission plugin
* Included PV and PVC objects as subdivided objects
* Added plan for deprecating the `system:nodes` group -> `system:node` role RBAC auto-binding

### DIFF
--- a/contributors/design-proposals/kubelet-authorizer.md
+++ b/contributors/design-proposals/kubelet-authorizer.md
@@ -21,13 +21,13 @@ pods belonging to other nodes, and accessing confidential data unrelated to the 
 
 This document proposes limiting a kubelet's API access using a new node authorizer, admission plugin, and additional API validation:
 * Node authorizer
-  * Authorizes requests from nodes using a fixed policy identical to the default RBAC `system:node` cluster role
-  * Further restricts secret and configmap access to only allow reading objects referenced by pods bound to the node making the request
+  * Authorizes requests from identifiable nodes using a fixed policy identical to the default RBAC `system:node` cluster role
+  * Further restricts secret, configmap, persistentvolumeclaim and persistentvolume access to only allow reading objects referenced by pods bound to the node making the request
 * Node admission
-  * Limit nodes to only be able to mutate their own Node API object
-  * Limit nodes to only be able to create mirror pods bound to themselves
-  * Limit nodes to only be able to mutate mirror pods bound to themselves
-  * Limit nodes to not be able to create mirror pods that reference API objects (secrets, configmaps, service accounts, persistent volume claims)
+  * Limit identifiable nodes to only be able to mutate their own Node API object
+  * Limit identifiable nodes to only be able to create mirror pods bound to themselves
+  * Limit identifiable nodes to only be able to mutate mirror pods bound to themselves
+  * Limit identifiable nodes to not be able to create mirror pods that reference API objects (secrets, configmaps, service accounts, persistent volume claims)
 * Additional API validation
   * Reject mirror pods that are not bound to a node
   * Reject pod updates that remove mirror pod annotations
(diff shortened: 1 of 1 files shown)


liggitt:
@kubernetes/sig-auth-proposals @timstclair @deads2k updated based on discussion in the authorizer PR (https://github.com/kubernetes/kubernetes/pull/46076)

ericchiang:
/lgtm

thanks for the update

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -146,29 +142,39 @@ for compatibility with deployment methods that bind other users or groups to tha
...
 but do not identify the particular node they are associated with.
-Those kubelets would be broadly authorized by the node authorizer,
-but would not have secret and configmap requests restricted.
-The node admission plugin would not restrict requests from these kubelets.
+
+These kubelets would not be authorized by the `Node` authorization mode,
+and would need to continue to be authorized via whatever mechanism currently authorizes them.
+
+The node admission plugin *would* restrict requests from these kubelets.
+That plugin should not be used in a deployment with kubelets that do not identify themselves.
```
timstclair: What are the implications of running the node authorizer without the admission piece? I think the lack of the admission plugin enables a simple way to bypass the authorizer (e.g. create a fake pod bound to itself that references the secret you want to read), right?

Given that there isn't a good fallback option for admission, would it make sense to keep the non-strict mode option there?

contributors/design-proposals/kubelet-authorizer.md
```diff
@@ -79,37 +79,30 @@ This group and user name format match the identity created for each kubelet as p
...
-    * node -> pod -> pvc -> pv -> secret
-5. For other resources, allow
+2. If a specific node cannot be identified (`IdentifyNode()` returns nodeName=""), reject
+3. If a request is for a secret, configmap, persistent volume or persistent volume claim, reject unless the verb is `get`, and the requested object is related to the requesting node:
+
+    * node <-pod
+    * node <-pod-> secret
+    * node <-pod-> configmap
+    * node <-pod-> pvc
+    * node <-pod-> pvc <-pv
```
timstclair: nit: I'm confused by the arrow directions here. Isn't it the `pvc` that references the `pv`?

liggitt: PVC and PV both reference each other, but our graph is only watching PV objects and inferring the PVC

contributors/design-proposals/kubelet-authorizer.md
liggitt: After talking with @deads2k, I think it's worse to have an admission plugin that claims to restrict nodes, but actually is completely permissive with some nodes. A deployment that is ok with a mix of identifiable nodes and indistinguishable nodes, but also cares about segmenting what those nodes can do seems odd to me. I'd rather start by saying this is a tool for clusters that have identifiable nodes. We can always add an "enforce, but not really sometimes" mode if there is demand for it.

timstclair: OK. Maybe add a warning (here, or in some future user-facing documentation) that running only one of the plugin does not grant any protection.

liggitt: after discussing the roll-out plan with @mikedanese, @cjcullen, and @deads2k, I'm ok with changing the node identifier to require both a group and username format match in order for the authorizer and admission plugin to pay attention to the request. See https://github.com/kubernetes/kubernetes/pull/46796/

Updated the proposal to match.

liggitt:
behavior as described in this update merged in https://github.com/kubernetes/kubernetes/pull/47218, merging

PR #675, opened 2017-06-01
Updates to development.md wrt Bazel

### LINKED ISSUE #597: development.md#workflow should emphasize that local repo location is not optional
I used to be able to work on Kubernetes when its location on my laptop was `$GOPATH/src/github.com/${my github profile name}/kubernetes` (an obvious choice, IMO).  Now that _does not_ work.  The location stated in https://github.com/kubernetes/community/blob/master/contributors/devel/development.md#workflow --- namely `$GOPATH/src/k8s.io/kubernetes` --- is _required_ .  If that location is not used then the `hack/update-bazel.sh` script fails (e.g., see https://github.com/kubernetes/kubernetes/pull/45263#issuecomment-299088432).  It would be good to update development.md#workflow to make it clear that the location it states *MUST* be used.

MikeSpreitzer:
Emphasized that the local repo location is not optional.

Added a pointer to the page about building with Bazel.

Resolves #597

### DIFF
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -172,6 +172,13 @@ dependencies](godep.md).
 
 
 
+## Build with Bazel/Gazel
+
+Building with Bazel is currently experimental.  For more information,
+see [Build with Bazel].
+
+
+
 ## Workflow
 
 ![Git workflow](git_workflow.png)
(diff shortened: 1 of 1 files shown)


contributors/devel/development.md
```diff
@@ -172,6 +172,13 @@ dependencies](godep.md).
 
 
 
+## Build with Bazel/Gazel
+
+Building with Bazel is currently experimental.  For more information,
+see [Build with Bazel].
```
mikedanese: Is this meant to have a link?

MikeSpreitzer: Yes, the link is defined at the end of the file.
Previewing on my fork looks right
screen shot 2017-06-02 at 4 38 43 pm
and functions correctly.

mikedanese: Interesting. I did not know this was a part of markdown.

PR #676, opened 2017-06-01
Update `contributors/devel/README.md` to reference `testing.md`

MikeSpreitzer:
Added a missing reference.

Resolves https://github.com/kubernetes/kubernetes/issues/35253

### DIFF
--- a/contributors/devel/README.md
+++ b/contributors/devel/README.md
@@ -27,6 +27,8 @@ Guide](http://kubernetes.io/docs/admin/).
 
 * **Development Guide** ([development.md](development.md)): Setting up your development environment.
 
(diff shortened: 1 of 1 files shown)


contributors/devel/README.md
```diff
@@ -27,6 +27,8 @@ Guide](http://kubernetes.io/docs/admin/).
 
 * **Development Guide** ([development.md](development.md)): Setting up your development environment.
 
+* **Testing** ([testing.md](testing.md)): Testing.
```
gyliu513: How about s/Testing/How to run unit, integration and end to end tests.

MikeSpreitzer:
@gyliu513 thanks for the suggestion.  I have added that, and one more detail.

PR #683, opened 2017-06-03
Add "make help"

xiangpengzhao:
Since we have `make help`, it would be good to add it in development.md

Ref: https://github.com/kubernetes/kubernetes/issues/46911

/cc @cblecker 
/assign @krzyzacy

### DIFF
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -298,6 +298,8 @@ make test-e2e
 See the [testing guide](testing.md) and [end-to-end tests](e2e-tests.md)
 for additional information and scenarios.
 
+Run `make help` for additional information on these make targets.
+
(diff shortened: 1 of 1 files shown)


contributors/devel/development.md
```diff
@@ -298,6 +298,8 @@ make test-e2e
 See the [testing guide](testing.md) and [end-to-end tests](e2e-tests.md)
 for additional information and scenarios.
 
+Run `make help` for help.
```
cblecker: "Run `make help` for additional information on these make targets."

xiangpengzhao: Thanks @cblecker！ This is exactly what I want!

xiangpengzhao:
@cblecker addressed as you suggested. PTAL.

PR #687, opened 2017-06-05
Add CI scripts

jamiehannaford:
This PR adds the following features:

- Users can now generate docs for 1 sig. They do this by running `make SIG=sig-testing gen-doc`
- CI scripts have been added (see below)
- All SIG docs have been generated for the first time

### CI script 

The script will perform the following:

- Generate docs into a temporary dir (based off the project YAML file)
- Go through each SIG doc and verify that the one which exists in the PR changeset matches the expected version
- If a mismatch is detected (i.e. the contributor tried to edit the generated docs instead of the YAML file) an error is generated and the PR fails


jamiehannaford:
@cmluciano I can try experimenting on a fork tomorrow

jamiehannaford:
@cmluciano I've created some PR scenarios:

1. [User updates generated doc file by hand](https://github.com/jamiehannaford/test/pull/4) (failure)
2. [User updates YAML but doesn't generate docs](https://github.com/jamiehannaford/test/pull/5) (failure)
3. [User updates YAML and generates docs](https://github.com/jamiehannaford/test/pull/3) (success)

jamiehannaford:
Note: This PR should probably also contain all of the generated docs, because if it doesn't, any PR after it will fail because the docs are inconsistent. I'll add a new commit for these.

Makefile
```diff
@@ -1,12 +1,15 @@
...
 reset-docs:
 	git checkout HEAD -- sig-list.md sig-*
 
-build-sigdocs:
+build-image:
 	docker build -t sigdocs -f generator/Dockerfile generator
 
-run-sigdocs:
+gen-doc:
+	docker run -e SIG=${SIG} -v $(shell pwd):/go/src/app sigdocs
```
jbeda: Protip: You can pass through env variables with just `-e SIG` and it'll grab the same named thing and pass it through.

generator/app.go
```diff
@@ -179,9 +179,15 @@ func writeLastGenerated(f *os.File) {
 }
 
 func createReadmeFiles(ctx SigEntries) error {
+	selectedSig := os.Getenv("SIG")
```
jbeda: nit: It is a little wacky to pass this through as an env variable vs. a command line flag. But it probably isn't worth changing at this point.

jamiehannaford: Yeah, I'll send a follow-up PR to refactor how this is done

Makefile
jamiehannaford: Good to know 😄

PR #690, opened 2017-06-06
Remove old references to SIG service catalog leads

pmorie:
Looks like I missed some previously.


jamiehannaford:
@pmorie Can you please also update [sigs.yaml](https://github.com/kubernetes/community/blob/master/sigs.yaml)? That's now the definitive source for sig information. Future docs will be generated from there :)

jamiehannaford:
@pmorie Sorry, didn't see that you did! Please ignore

PR #692, opened 2017-06-07
support ipvs mode for kube-proxy

m1093782566:
Implement IPVS-based in-cluster service load balancing. It can provide some performance enhancement and some other benefits to kube proxy while comparing iptables and userspace mode. Besides, it also support more sophisticated load balancing algorithms than iptables (least conns, weighted, hash and so on).

related issue: kubernetes/kubernetes#17470 kubernetes/kubernetes#44063

related PR: kubernetes/kubernetes#46580 kubernetes/kubernetes#48994

@thockin @quinton-hoole @wojtek-t

### DIFF
--- a/contributors/design-proposals/ipvs-proxy.md
+++ b/contributors/design-proposals/ipvs-proxy.md
@@ -1,0 +1,605 @@
+# Support IPVS Load Balancing Mode in Kubernetes
+
+**Authors**
+
+* dujun (@m1093782566)
+* haibin xie (@haibinxie)
+
+## Abstract
+
+This proposal summarizes what’s expected in alpha version IPVS load balancing support, includes Kubernetes user behavior changes, build and deployment changes, and the test validation planned. 
+
+## Motivation
+
+There’s been a lot discussion and voice of enabling IPVS as in-cluster service load balancing mode.  IPVS performs better than iptables, meanwhile supports more sophisticated load balancing algorithms than iptables (least load, least connections, locality, weighted) as well as other useful features (e.g. health checking, retries etc).
+
+IPVS is an alternative of iptables as load balancer, it’s assumed reader of this proposal is familiar with IPTables load balancer mode. 
+
+##  Kubernetes behavior change
+
+### Changes to kube-proxy startup parameter
+
+**1. Proxy mode**
+
+In addition to existing userspace and iptables mode, ipvs mode is configured via `--proxy-mode=ipvs`.
+
+**2. Expose different  load balancing  algorithms**
+
+A new kube-proxy parameter `--ipvs-scheduler` will be added to specify IPVS load balancing algorithm. Below is list of supported values, if it’s not configured `rr` is default value, if it’s incorrectly configured kube-proxy will exit with error message.
+
+- rr: round-robin
+
+- lc: least connection
+
+- dh: destination hashing
+
+- sh: source hashing
+
+- sed: shortest expected delay
+
+- nq: never queue
+
+For more details about it, refer to [http://kb.linuxvirtualserver.org/wiki/Ipvsadm](http://kb.linuxvirtualserver.org/wiki/Ipvsadm)
+
+**3. clean up inactive rules**
+
+In order to cleaning up inactive rules(includes iptables rules and ipvs rules), we will introduce a new  kube-proxy parameter `--cleanup-proxyrules ` and mark the older `--cleanup-iptables` deprecated. 
+
+**4. ipvs maximum sync period**
+
+Introduce a new parameter ` --ipvs-sync-period` to specify the maximum interval of how often ipvs rules are refreshed (e.g. '5s', '1m', '2h22m').  Must be greater than 0.
+
+**5. ipvs minimum sync period**
+
+Introduce a new parameter ` --ipvs-min-sync-period`  to the minimum interval of how often the ipvs rules can be refreshed as endpoints and services change (e.g. '5s', '1m', '2h22m').
+
+### Changes to build
+
+No changes at all. The IPVS implementation is built on [docker/libnetwork](https://godoc.org/github.com/docker/libnetwork/ipvs) ipvs library, which is a pure-golang implementation and talks to kernel via socket communication.
+
+### Changes to deployment
+
+IPVS kernel module installation is beyond kubernetes, it’s assumed IPVS kernel modules are installed on the node before running kube-proxy. When kube-proxy starts, if proxy mode is IPVS kube-proxy would validate if IPVS modules are installed on the node, if it’s not installed kube-proxy will fall back to userspace proxy mode.
+
+## Other design considerations
+
+### Clean up inactive rules
+
+It seems difficult to distinguish if an ipvs service is created by ipvs proxier or other processes. Currently we assume ipvs rules will be created only by ipvs proxier in a kubernetes node, so we can clear all ipvs rules in a kubernetes node. Probably we should add warnings in document or flag comment. 
+
+### When fall back on iptables
+
+IPVS proxier will employ iptables in doing packet filtering, SNAT and supporting NodePort type service. Specifically, ipvs proxier will fall back on iptables in the following 4 scenarios.
+
+* kube-proxy start with --masquerade-all=true
+* Specify cluster CIDR in kube-proxy startup
+* Load Balancer Source Ranges is specified for LB type service
+* Support NodePort type service
+
+And, IPVS proxier will maintain 5 kubernetes-specific chains in nat table
+
+- KUBE-POSTROUTING 
+- KUBE-MARK-MASQ
+- KUBE-MARK-DROP
+- KUBE-MARK-SERVICES
+- KUBE-NODEPORTS
+
+`KUBE-POSTROUTING`, `KUBE-MARK-MASQ`, ` KUBE-MARK-DROP` are maintained by kubelet and ipvs proxier won't create them. IPVS proxier will make sure chains `KUBE-MARK-SERVICES` and `KUBE-NODEPORTS` exist in its sync loop.
+
+**1. kube-proxy start with --masquerade-all=true**
+
+If kube-proxy start with `--masquerade-all=true`, ipvs proxier will masquerade all traffic accessing service Cluster IP, which behaves same as what iptables proxier does. Suppose there is a serivice with Cluster IP `10.244.5.1` and port `8080`, then the iptables installed by ipvs proxier should be like what is shown below.
+
+```shell
+# iptables -t nat -nL
+
+Chain PREROUTING (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
+Chain OUTPUT (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
+Chain POSTROUTING (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-POSTROUTING  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes postrouting rules */
+
+Chain KUBE-POSTROUTING (1 references)
+target     prot opt source               destination         
+MASQUERADE  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service traffic requiring SNAT */ mark match 0x4000/0x4000
+
+Chain KUBE-MARK-DROP (0 references)
+target     prot opt source               destination         
+MARK       all  --  0.0.0.0/0            0.0.0.0/0            MARK or 0x8000
+
+Chain KUBE-MARK-MASQ (6 references)
+target     prot opt source               destination         
+MARK       all  --  0.0.0.0/0            0.0.0.0/0            MARK or 0x4000
+
+Chain KUBE-SERVICES (2 references)
+target     prot opt source               destination         
+KUBE-MARK-MASQ  tcp  -- 0.0.0.0/0        10.244.5.1            /* default/foo:http cluster IP */ tcp dpt:8080
+```
+
+**2. Specify cluster CIDR in kube-proxy startup**
+
+If kube-proxy start with `--cluster-cidr=<cidr>`, ipvs proxier will masquerade off-cluster traffic accessing service Cluster IP, which behaves same as what iptables proxier does. Suppose kube-proxy is provided with the cluster cidr `10.244.16.0/24`, and service Cluster IP is `10.244.5.1` and port is `8080`, then the iptables installed by ipvs proxier should be like what is shown below.
+
+```shell
+# iptables -t nat -nL
+
+Chain PREROUTING (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
+Chain OUTPUT (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
+Chain POSTROUTING (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-POSTROUTING  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes postrouting rules */
+
+Chain KUBE-POSTROUTING (1 references)
+target     prot opt source               destination         
+MASQUERADE  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service traffic requiring SNAT */ mark match 0x4000/0x4000
+
+Chain KUBE-MARK-DROP (0 references)
+target     prot opt source               destination         
+MARK       all  --  0.0.0.0/0            0.0.0.0/0            MARK or 0x8000
+
+Chain KUBE-MARK-MASQ (6 references)
+target     prot opt source               destination         
+MARK       all  --  0.0.0.0/0            0.0.0.0/0            MARK or 0x4000
+
+Chain KUBE-SERVICES (2 references)
+target     prot opt source               destination         
+KUBE-MARK-MASQ  tcp  -- !10.244.16.0/24        10.244.5.1            /* default/foo:http cluster IP */ tcp dpt:8080
+```
+
+**3. Load Balancer Source Ranges is specified for LB type service**
+
+When service's `LoadBalancerStatus.ingress.IP` is not empty and service's `LoadBalancerSourceRanges` is specified, ipvs proxier will install iptables which looks like what is shown below. 
+
+Suppose service's `LoadBalancerStatus.ingress.IP` is `10.96.1.2` and service's `LoadBalancerSourceRanges` is `10.120.2.0/24`.
+
+```shell
+# iptables -t nat -nL
+
+Chain PREROUTING (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
+Chain OUTPUT (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
+Chain POSTROUTING (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-POSTROUTING  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes postrouting rules */
+
+Chain KUBE-POSTROUTING (1 references)
+target     prot opt source               destination         
+MASQUERADE  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service traffic requiring SNAT */ mark match 0x4000/0x4000
+
+Chain KUBE-MARK-DROP (0 references)
+target     prot opt source               destination         
+MARK       all  --  0.0.0.0/0            0.0.0.0/0            MARK or 0x8000
+
+Chain KUBE-MARK-MASQ (6 references)
+target     prot opt source               destination         
+MARK       all  --  0.0.0.0/0            0.0.0.0/0            MARK or 0x4000
+
+Chain KUBE-SERVICES (2 references)
+target     prot opt source       destination         
+ACCEPT  tcp  -- 10.120.2.0/24    10.96.1.2       /* default/foo:http loadbalancer IP */ tcp dpt:8080
+DROP    tcp  -- 0.0.0.0/0        10.96.1.2       /* default/foo:http loadbalancer IP */ tcp dpt:8080
+```
+
+**4. Support NodePort type service**
+
+Please check the section below.
+
+### Support NodePort type service
+
+For supporting NodePort type service, iptables will recruit the exsiting implementation in iptables proxier. For example, 
+
+```shell
+# kubectl describe svc nginx-service
+Name:			nginx-service
+...
+Type:			NodePort
+IP:			    10.101.28.148
+Port:			http	3080/TCP
+NodePort:		http	31604/TCP
+Endpoints:		172.17.0.2:80
+Session Affinity:	None
+
+# iptables -t nat -nL
+
+[root@100-106-179-225 ~]# iptables -t nat -nL
+Chain PREROUTING (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
+Chain OUTPUT (policy ACCEPT)
+target     prot opt source               destination         
+KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
+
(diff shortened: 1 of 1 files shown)


dhilipkumars:
@m1093782566 Thank you for reacting quickly,  Could we add some statistics we collected during our experiment?  Like a table that compares IpTables vs IPVS with few thousand services created.

m1093782566:
@dhilipkumars I think @haibinxie has the original statistics.

m1093782566:
**IPVS vs. IPTables Latency to Add Rules**

Measured by iptables and ipvsadm, observation:

* In iptables mode, latency to add rules increase significantly when number of service increases

* In IPVS mode, latency to add VIP and backend IPs does not increase when number of service increases

| number of services        | 1           | 5000  |  20000  |
| ------------- |:-------------:| -----:|-----:|
| number of rules      | 8 | 40000 |160000|
| iptables      | 2ms      |   11min |5hours|
| ipvs | are neat      |   2ms |2ms| 2ms|


@dhilipkumars I am not sure if the statistics is sufficient.

spiffxp:
related to the google doc here? https://docs.google.com/document/d/1YEBWR4EWeCEWwxufXzRM0e82l_lYYzIXQiSayGaVQ8M

spiffxp:

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+Similar to iptables proxy, IPVS proxy will do full sync loop every 10 seconds by default. Besides, every update on kubernetes service and endpoint will trigger an ipvs service and destination update.
+
+### Support Only NodeLocal Endpoints
+
+Similar to iptables proxy, IPVS proxy support Only NodeLocal Endpoints feature. When a service has the annotation,  ipvs proxy will do traffic proxy only for local endpoints.
+
+### Network policy
+
+For IPVS NAT mode to work, **all packets from the realservers to the client must go through the director**. It means ipvs proxy should do SNAT in a L3-overlay network(such as flannel) for cross-host network communication. When a container request a cluster IP to visit an endpoint in another host, we should enable `--masquer-all` for ipvs proxy, **which will break network policy**.
```
cmluciano: Breaking a soon-to-be GA feature seems like an edge case that we probably cannot overlook. Even with something in alpha form, I think we should be sure that all existing functionality is supported.

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+  -  host -> container (same host)
+
+  -  host -> container (cross host)
+
+  -  container -> host (same host)
+
+  -  container -> host (cross host)
+
+
+## TODO
```
cmluciano: I'm not sure if the TODO is necessary since it should be part of the PR steps.

m1093782566:
@spiffxp Yes. @haibinxie wrote the doc and I translated it to markdown file and added some details

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+- sed: shortest expected delay
+
+
+- nq: never queue
+
+For more details about it, refer to [http://kb.linuxvirtualserver.org/wiki/Ipvsadm](http://kb.linuxvirtualserver.org/wiki/Ipvsadm)
+
+In order to clean up inactive rules(includes iptables rules and ipvs rules), we will introduce a new  kube-proxy parameter `--cleanup-proxyrules ` and mark the older `--cleanup-iptables` deprecated. Unfortunately, there is no way to distinguish whether an ipvs service is created by ipvs proxy or other process, `--cleanup-proxyrules` will clear all ipvs service in a host.
+
+### Change to build
```
m1093782566: Yes. We changed to use libnetwork to avoid cgo and libnl dependency.

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Session affinity
+
+ipvs support client IP session affinity, it's called persistent connection. It's the `-p` flag in ipvsadm, like
+
+```shell
+ipvsadm -A -t 10.244.1.100:8080 -s rr -p [timeout]
+```
+
+When a service specify session affinity, ipvs proxy will assign a timeout value(180min by default) to ipvs service.
```
m1093782566: The current default value for iptables proxy is 180min, see https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/iptables/proxier.go#L205

m1093782566:
@ThomasZhou @haibinxie @fisherxu

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+
+- sed: shortest expected delay
+
+
+- nq: never queue
+
+For more details about it, refer to [http://kb.linuxvirtualserver.org/wiki/Ipvsadm](http://kb.linuxvirtualserver.org/wiki/Ipvsadm)
+
+In order to clean up inactive rules(includes iptables rules and ipvs rules), we will introduce a new  kube-proxy parameter `--cleanup-proxyrules ` and mark the older `--cleanup-iptables` deprecated. Unfortunately, there is no way to distinguish whether an ipvs service is created by ipvs proxy or other process, `--cleanup-proxyrules` will clear all ipvs service in a host.
```
m1093782566: @ddysher Do you have any good idea?

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+We will create some ipvs services for each kubernetes service. The VIP of ipvs service corresponding to the accessable IP(such as cluster IP, external IP, nodeIP, ingress IP, etc.) of kubernetes service. Each destination of an ipvs service corresponding to an kubernetes service endpoint.
+
+### Port remapping 
+
+There are 3 proxy mode in ipvs, NAT, IPIP and DR. Only NAT mode support port remapping.
+
+### NodePort type service support
+
+For NodePort type service, IPVS proxy will take all accessable IPs in a host as the virtual IP of ipvs service. Specifically, accessable IP excludes `lo`, `docker0`, `vethxxx`, `cni0`, `flannel0`, etc. Currently, we assume they are IPs bound to `eth{i}`.
```
ddysher: Sorry this comment somehow lost.

Why do we enforce interface name?  e.g. assuming ethx won't work for predictable network interface name for newer systemd.

contributors/design-proposals/ipvs-proxy.md
haibinxie: SNAT is required for cross host communication, there should be compromise in between, probably an awareness of the situation in release note.  I am also open to any solution/fix that we can work towards.

contributors/design-proposals/ipvs-proxy.md
ddysher: Not that I know of.  It's probably ok to just call out this in document or flag comment.  User with existing ipvs services should use with caution.  If one just wants to clean up kubernetes related ipvs services, just start the proxy and clear any unwanted k8s services.

fisherxu: Yes. We can add this to document or flag comment.

contributors/design-proposals/ipvs-proxy.md
ddysher: apart from soon to be a GA feature, network policy is really important that people probably would trade performance for it.  A draft idea is to not do snat, rather, on each host, add a new routing table and fwmark all service traffic to this table. The routing table will take care of routing back the packet.  There's a lot of caveats, i don't know if it will ever work at all.

haibinxie:
@thockin 
Let me know if this helps.

Is this built around exec ipvsadm?
[Haibin Michael Xie] this is built on top of libnetwork, which talks to kernel via socket communication. Not on top of ipvsadm.

I'd like to see pseudo-code explaining how the resync loop works, and what the intermediate state looks like.

How do you prevent dropped packets during updates?
[Haibin Michael Xie] do you mean OS updates? I don't know the full picture of how iptables handles it, IMO this is with no difference from iptables.

How do you do cleanups across proxy restarts, where you might have lost information? (e.g. create service A and B, you crash, service A gets deleted, you restart, you get a sync for B - what happens to A?)
[Haibin Michael Xie] There is periodic full resync and in memory cache based diff. This should be handled already in iptables, and there is no difference in this regard.

How does this scale (if I have 10,000 services and 5 backends each, is this 50,000 exec calls?)
[Haibin Michael Xie] Same to above libnetwork use socket to talk to kernel which is very efficient.

How will you use/expose the different algorithms?
[Haibin Michael Xie] If it's LB algorithm, it is already mentioned in the proposal. kube-proxy has a new parameter --ipvs-scheduler.

contributors/design-proposals/ipvs-proxy.md
m1093782566: I am not sure if L3 overlay(flannel with vxlan backend is the matter). Maybe I need a reverse traffic goes through IPVS proxy so that traffic reaches the source pod (after performing DNAT). I will re-test it in my environment.

m1093782566: @murali-reddy do you have any idea?

murali-reddy: Agree with @thockin, there is no need for SNAT, atleast not for all traffics. So when a pod access cluster IP or node port, IPVS does DNAT. On reverse path pretty much (can not think of any pod networking solution where its not true) route to source pod is back to source pod node.

Where its gets tricky, is when a node port is accessed from outside the cluster. Node on which destination pod is running may route traffic directly to client through default gateway. To prevent that we need to SNAT traffic, so the return traffic goes through the same node through which client accessed node port.  We do loose the source IP in this case. But AFAIK this is not unique to use of IPVS but even iptable kube-proxy will have to do the SNAT. 

FWIW, i have implemented [logic](https://github.com/cloudnativelabs/kube-router/blob/master/app/controllers/network_services_controller.go#L358) in [kube-router](https://github.com/cloudnativelabs/kube-router) to deal with external client accessing node port.

I just tested Flannel VXLAN + IPVS service proxy + network policy, i dont see any issue. Dont see reason to do SNAT. Please test it with your POC, and see you can remove this restriction.

m1093782566: Thanks @murali-reddy.

m1093782566: I re-tested in my environment(Flannel VXLAN + IPVS service proxy) and found something different.

There is an ipvs service with 2 destinations which are in the different host.
```
[root@SHA1000130405 home]# ipvsadm -ln
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn     
TCP  10.102.128.4:3080 rr
  -> 10.244.0.235:8080            Masq    1      0          0         
  -> 10.244.2.123:8080            Masq    1      0          0 
```

There was no SNAT rules applied and I `curl VIP` in each container.

```
# In container 10.244.0.235

$curl 10.102.128.4:3080
# get response

tcpdumped in another host to see what is source IP is.

$ tcpdump -i flannel.1

20:44:48.021765 IP 10.244.0.235.36844 > 10.244.2.123.webcache: Flags [S], seq 519767460, win 28200, options [mss 1410,sackOK,TS val 416
20:44:48.021998 IP 10.244.2.123.webcache > 10.244.0.235.36844: Flags [S.], seq 1131844123, ack 519767461, win 27960, options [mss 1410,76,nop,wscale 7], length 0
```

So, it seems that **there is no need to do SNAT for cross-host communication**.

However, when ipvs scheduled the self container, there is no response returned. It seems that the packet is dropped when the traffic reached the self container. 

I can confirm that it has nothing to do with `same host vs. cross host`, the issue is about **whether the request hit the same container.**

When I applied SNAT rules as iptables proxy does, it can reach self container via VIP.

```
iptables -t nat -A PREROUTING -s 10.244.0.235 -j KUBE-MARK-MASQ
```

Unfortunately, the source container lost the source container IP and the source IP became flannel.1 IP.


Anyway, it's a good news to me that SNAT is unnecessary for cross-host communication though I need more knowledge to fix the issue I mentioned above.

murali-reddy: Again this is nothing specic to IPVS, please search for hairpin related issues with kube-proxy/kubelet https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/#a-pod-cannot-reach-itself-via-service-ip

m1093782566:
@thockin 

As I re-tested and found something different. SNAT is not required for cross-host communication. So it won't break network policy. It's really a big finding to me :)

I find the packet will be dropped when container vist VIP and the real backend is self. I have no idea now but will try to find out why.

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
+# Alpha Version IPVS Load Balancing Mode in Kubernetes
```
m1093782566: Okay, will fix.

contributors/design-proposals/ipvs-proxy.md
m1093782566: Can you only clean rules with RC1918 addresses?
[m1093782566] It still has a possibility of clearing ipvs rules created by other process although the the possibility is low.
[m1093782566] Kube-apiserver knows the service cluster IP range through its--service-cluster-ip-range parameter. However, kube-proxy knows nothing about that. I don't suggest to add a new --service-cluster-ip-range flag in kube-proxy since it will easily conflicts with kube-apiserver's flag. Even though kube-proxy clearing all ipvs rules in the service cluster IP range, it may leave some ipvs rules in external IP, exteral LB ingress IP and Node IP.

Clearing all ipvs rules with RFC1918 addresses probably is easier to implement and has lower possibility of clearing user's existing ipvs rules. 

@fisherxu What do you think about it?

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+```shell
+ipvsadm -A -t 10.244.1.100:8080 -s rr -p [timeout]
+```
+
+When a service specify session affinity, ipvs proxy will assign a timeout value(180min by default) to ipvs service.
+
+### Sync period
+
+Similar to iptables proxy, IPVS proxy will do full sync loop every 10 seconds by default. Besides, every update on kubernetes service and endpoint will trigger an ipvs service and destination update.
```
m1093782566: Changing seesion affinity will call [UpdateService API](https://github.com/docker/libnetwork/blob/master/ipvs/ipvs.go#L109) and won't take a service down.

m1093782566:
I will update the proposal and try to fix the review comments in newer proposal. 

@thockin I will add pseudo-code explaining how the resync loop works,

m1093782566:
@thockin @ddysher @cmluciano I update the proposals according review comments and add more details. PTAL.

Any comments are welcomed. Thanks.

/cc @haibinxie

contributors/design-proposals/ipvs-proxy.md
m1093782566: @dhilipkumars did a test and found ipvs update did not disrupt the service or even the existing connection.

m1093782566:
@thockin 

> How do you prevent dropped packets during updates?

According to @dhilipkumars 's  test result. It shows ipvs update did not disrupt the service or even the existing connection

```
sudo ipvsadm -L -n
[sudo] password for d:
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  10.11.12.13:6379 wlc persistent 10
  -> 10.192.0.1:32768             Masq    1      1          0
```
the real backedn service is redis-alpine conecting to the service
```
docker run --net=host -it redis:3.0-alpine redis-cli -h 10.11.12.13 -p 6379
10.11.12.13:6379> info Clients
# Clients
connected_clients:1
client_longest_output_list:0
client_biggest_input_buf:0
blocked_clients:0
```
In a parllel session if i update the scheduler algorithm or persistence timeout the service is not disrupted.    Ipvsadm and libnetwor's ipvs pkg works in the same principle firing netlink message to the kernal so the behaviour should be the same.

Udpate persistance timout still no
```
$sudo ipvsadm -E -t 10.11.12.13:6379 -p 60
$sudo ipvsadm -L -n --persistent-conn
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port            Weight    PersistConn ActiveConn InActConn
  -> RemoteAddress:Port
TCP  10.11.12.13:6379 wlc persistent 60
  -> 10.192.0.1:32768             1         1           1          0
```

what other parameters should we test?

haibinxie:
@thockin Please take another look when you get a chance, this should be in good shape now.

m1093782566:

contributors/design-proposals/ipvs-proxy.md
m1093782566: What about take all UP state network interface(except the "vethxxx"s) address as the Node IP?

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+Some of the existing deployment script might need update to install ipvsadm, which is beyond the scope of alpha version.
+
+```shell
+apt-get install ipvsadm
+```
+
+## Other design considerations
+
+### IPVS setup and network topology
```
m1093782566: Yes, I give a section "when fall back on iptables" to explain the interaction between ipvs and iptables. Thanks!

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+IPVS is a replacement or IPTables as load balancer, it’s assumed reader of this proposal is familiar with IPTables load balancer mode. We will create a dummy interface and assign all service Cluster IPs to the dummy interface(maybe called `kube0`). In alpha version, we will implicitly use NAT mode.
```
m1093782566: Fixed. Thanks for reminding.

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+apt-get install ipvsadm
+```
+
+## Other design considerations
+
+### IPVS setup and network topology
+
+IPVS is a replacement or IPTables as load balancer, it’s assumed reader of this proposal is familiar with IPTables load balancer mode. We will create a dummy interface and assign all service Cluster IPs to the dummy interface(maybe called `kube0`). In alpha version, we will implicitly use NAT mode. 
+
+We will create some ipvs services for each kubernetes service. The VIP of ipvs service corresponding to the accessable IP(such as cluster IP, external IP, nodeIP, ingress IP, etc.) of kubernetes service. Each destination of an ipvs service corresponding to an kubernetes service endpoint.
```
m1093782566: Suppose kubernetes service is NodePort type. Then ipvs Proxier will create two ipvs serivice, one is NodeIP:NodePort and the other one is ClusterIP:Port.

Of course, I will explain it in the doc. Thanks.

contributors/design-proposals/ipvs-proxy.md
m1093782566: Fixed. Thanks.

contributors/design-proposals/ipvs-proxy.md
m1093782566: Fixed. Thanks!

contributors/design-proposals/ipvs-proxy.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Support Only NodeLocal Endpoints
+
+Similar to iptables proxy, IPVS proxy support Only NodeLocal Endpoints feature. When a service has the annotation,  ipvs proxy will do traffic proxy only for local endpoints.
+
+### Network policy
+
+For IPVS NAT mode to work, **all packets from the realservers to the client must go through the director**. It means ipvs proxy should do SNAT in a L3-overlay network(such as flannel) for cross-host network communication. When a container request a cluster IP to visit an endpoint in another host, we should enable `--masquer-all` for ipvs proxy, **which will break network policy**.
+
+## Test validation
```
m1093782566: super!

spiffxp:
pinging @kubernetes/sig-network-proposals (k8s-bot should re-mention this in the future thanks to https://github.com/kubernetes/test-infra/issues/2956)

m1093782566:
ping @thockin

m1093782566:
@thockin 

Thanks your awesome comments. Almost all of comments are addressed(at least updated, I think) except 2 issues as I am not sure which solution is better.

* Node Port type service support- Should we fall back on iptables?

* How to flush ipvs rules? - Is it ok to clear all ipvs rules on a kubernetes node?

I am sure this doc is easier to read and understand. Please take another look and give your opinions. Much appreciated.

contributors/design-proposals/ipvs-proxy.md
m1093782566: Updates:

As discussed in sig-network meeting, IPVS proxier will fall back on iptables when support node port type service.

contributors/design-proposals/ipvs-proxy.md
m1093782566: Update:

Since ipvs proxier will fall back on iptables when support node port type service. I should give another example.

> Note that the relationship between kubernetes service and ipvs service is 1:N。The address of ipvs service corresponding to service's access IP, such as Cluster IP, external IP and LB.ingress.IP. If a kubernetes service has more than one access IP, For example, a external IP type service has 2 access IP(ClusterIP and External IP), then ipvs proxier will create 2 ipvs serivice - one for Cluster IP and the other one for External IP.

m1093782566:
As discussed in sig-netwrok meeting.

This design proposal updated to:

* fall back on iptables when support node port type service

* clear all ipvs rules on a kubernetes node when user call kube-proxy to flush proxy rules.

haibinxie:
@thockin Could you confirm if @m1093782566 's comment above is the right thing to do?  anything else is left on closing it.  please expect me keep bothering you until it's closed :)  

If you get a chance we can have a quick phone call on reviewing and addressing issues on this, we want to commit to release the feature in 1.8.

m1093782566:
@danwinship Do you have interest in reviewing this design proposal? Any comments are welcome. :)

m1093782566:
Hi @thockin @ddysher 

I just come up with an idea about implementing nodeport type service via ipvs.

Can we take all the IP address which ADDRTYPE match dst-type LOCAL as the address of ipvs service? For example,

```
[root@100-106-179-225 ~]# ip route show table local type local
100.106.179.225 dev eth0  proto kernel  scope host  src 100.106.179.225 
127.0.0.0/8 dev lo  proto kernel  scope host  src 127.0.0.1 
127.0.0.1 dev lo  proto kernel  scope host  src 127.0.0.1 
172.16.0.0 dev flannel.1  proto kernel  scope host  src 172.16.0.0 
172.17.0.1 dev docker0  proto kernel  scope host  src 172.17.0.1 
192.168.122.1 dev virbr0  proto kernel  scope host  src 192.168.122.1 
```

Then, [100.106.179.225, 127.0.0.0/8, 127.0.0.1, 172.16.0.0, 172.17.0.1, 192.168.122.1] would be the address of ipvs service for nodeport service.

I assume KUBE-NODEPORTS chain created by iptables proxier did the same thing? For example,

```
Chain KUBE-SERVICES (2 references)
target     prot opt source               destination         
KUBE-NODEPORTS  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service nodeports; NOTE: this must be the last rule in this chain */ ADDRTYPE match dst-type LOCAL
```

Looking forward to receive your opinions.

m1093782566:
**By doing this, we can remove the design constraint that assuming node IP is the address of `eth{x}.`**

m1093782566:
@feiskyer Do you have bandwidth to take a look? Thanks :)

feiskyer:
Using a list of IP addresses for nodePort has potential problems, e.g. ip addresses may be changed or new nics may be added later. And I don't think watching the changes of ip addresses and nics is a good idea.

Maybe using iptables for nodePort services is a better choice.

m1093782566:
Glad to receive your feedback, @feiskyer 

> And I don't think watching the changes of ip addresses and nics is a good idea.

Yes, I agree. Thanks for your thoughts.

luxas:
ping @kubernetes/sig-network-feature-requests 
Any movement here lately?

haibinxie:
@luxas it's released in 1.8 https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#downloads-for-v180

haibinxie:
@luxas alpha in 1.8. we are working on beta release in 1.9

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/692)

spiffxp:
/lifecycle freeze
@cmluciano I'm keeping this open on your behalf, if this is no longer relevant to keep open please /lifecycle-cancel freeze

spiffxp:
/lifecycle frozen
@cmluciano I'm keeping this open on your behalf, if this is no longer relevant to keep open please /lifecycle-cancel frozen

cmluciano:
@m1093782566 Is there a PR that supersedes this one?

m1093782566:
@cmluciano 

NO.

This PR is the only design proposal. IPVS proxier already reached beta while this document is still pending, unfortunately.

PR #699, opened 2017-06-09
Update local storage overview

ddysher:
Update local storage overview to alpha API.  Also add some info on secondary partition.

### DIFF
--- a/contributors/design-proposals/local-storage-overview.md
+++ b/contributors/design-proposals/local-storage-overview.md
@@ -96,7 +96,7 @@ The term `Partitions` are used here to describe the main use cases for local sto
        resources:
          limits:
            storage.kubernetes.io/logs: 500Mi
-           storage.kubernetes.io/writable: 1Gi
+           storage.kubernetes.io/overlay: 1Gi
        volumeMounts:
        - name: myEmptyDir
          mountPath: /mnt/data
@@ -107,9 +107,9 @@ The term `Partitions` are used here to describe the main use cases for local sto
     ```
 
 3. Alice’s pod “foo” is Guaranteed a total of “21.5Gi” of local storage. The container “fooc” in her pod cannot consume more than 1Gi for writable layer and 500Mi for logs, and “myEmptyDir” volume cannot consume more than 20Gi.
-4. For the pod resources, `storage.kubernetes.io/logs` resource is meant for logs. `storage.kubernetes.io/writable` is meant for writable layer.
+4. For the pod resources, `storage.kubernetes.io/logs` resource is meant for logs. `storage.kubernetes.io/overlay` is meant for writable layer.
 5. `storage.kubernetes.io/logs` is satisfied by `storage.kubernetes.io/scratch`.
-6. `storage.kubernetes.io/writable` resource can be satisfied by `storage.kubernetes.io/overlay` if exposed by nodes or by `storage.kubernetes.io/scratch` otherwise. The scheduler follows this policy to find an appropriate node which can satisfy the storage resource requirements of the pod.
+6. `storage.kubernetes.io/overlay` resource can be satisfied by `storage.kubernetes.io/overlay` if exposed by nodes or by `storage.kubernetes.io/scratch` otherwise. The scheduler follows this policy to find an appropriate node which can satisfy the storage resource requirements of the pod.
 7. EmptyDir.size is both a request and limit that is satisfied by `storage.kubernetes.io/scratch`.
 8. Kubelet will rotate logs to keep scratch space usage of “fooc” under 500Mi
 9. Kubelet will track the usage of pods across logs and overlay filesystem and restart the container if it's total usage exceeds it's storage limits. If usage on `EmptyDir` volume exceeds its `limit`, then the pod will be evicted by the kubelet. By performing soft limiting, users will be able to easily identify pods that run out of storage.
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/local-storage-overview.md
```diff
@@ -101,22 +103,22 @@ Since local PVs are only accessible from specific nodes, the scheduler needs to
...
      - name: myEmptyDir
        emptyDir:
-	     sizeLimit: 20Gi
+         sizeLimit: 20Gi
     ```
 
 3. Alice’s pod “foo” is Guaranteed a total of “21.5Gi” of local storage. The container “fooc” in her pod cannot consume more than 1Gi for writable layer and 500Mi for logs, and “myEmptyDir” volume cannot consume more than 20Gi.
 4. For the pod resources, `storage.kubernetes.io/logs` resource is meant for logs. `storage.kubernetes.io/writable` is meant for writable layer.
 5. `storage.kubernetes.io/logs` is satisfied by `storage.kubernetes.io/scratch`.
 6. `storage.kubernetes.io/writable` resource can be satisfied by `storage.kubernetes.io/overlay` if exposed by nodes or by `storage.kubernetes.io/scratch` otherwise. The scheduler follows this policy to find an appropriate node which can satisfy the storage resource requirements of the pod.
```
ddysher: @jingxu97 Are we still using `storage.kubernetes.io/writable` as a resource type?  What's the relationship between `storage.kubernetes.io/overlay`?

contributors/design-proposals/local-storage-overview.md
```diff
@@ -512,7 +523,7 @@ Note: Block access will be considered as a separate feature because it can work
         requests:
           storage: 80Gi
     ```
-4. It is also possible for a PVC that requests `volumeType: file` to also use a block-based PV.  In this situation, the block device would get formatted with the filesystem type specified in the PV spec.  And when the PV gets destroyed, then the filesystem also gets destroyed to return back to the original block state.
+4. The block device would get formatted with the filesystem type specified in the PV spec.  And when the PV gets destroyed, then the filesystem also gets destroyed to return back to the original block state.
```
ddysher: @msau42  volumeType is not in alpha API, should we remove it here?

ddysher:
@jingxu97 @msau42 @vishh

contributors/design-proposals/local-storage-overview.md
msau42: No, this document is forward looking and contains ideas beyond the initial alpha phase.

I have a separate user guide document with the current state of what's implemented.

contributors/design-proposals/local-storage-overview.md
```diff
@@ -196,37 +198,46 @@ Since local PVs are only accessible from specific nodes, the scheduler needs to
...
     toplogyKey: kubernetes.io/hostname
     ```
     ```yaml
     kind: PersistentVolume
     apiVersion: v1
     metadata:
       name: local-pv-1
-      labels:
-        kubernetes.io/hostname: node-1
+      annotations:
```
msau42: Let's have this doc reflect the eventual state of the API that we want.  So let's make this a field instead of an annotation.

contributors/design-proposals/local-storage-overview.md
```diff
@@ -475,7 +487,7 @@ Since local PVs are only accessible from specific nodes, the scheduler needs to
 Note: Block access will be considered as a separate feature because it can work for both remote and local storage.  The examples here are a suggestion on how such a feature can be applied to this local storage model, but is subject to change based on the final design for block access.
 
 1. The cluster that Bob uses has nodes that contain raw block devices that have not been formatted yet.
-2. The same addon DaemonSet can also discover block devices and creates corresponding PVs for them with the `block` field.  
+2. The same addon DaemonSet can also discover block devices and creates corresponding PVs for them. `path` is overloaded here to mean both fs path and block device path.
```
msau42: The PersistentVolume object should have a volumeType field to indicate if it's a block volume or a filesystem volume.

contributors/design-proposals/local-storage-overview.md
msau42: I think actually to align with CSI, the fs type should be specified in the PVC.

contributors/design-proposals/local-storage-overview.md
jingxu97: No, we no longer user writable, it is the same as overlay

ddysher: ok, changed`storage.kubernetes.io/writable` to `storage.kubernetes.io/overlay`.

contributors/design-proposals/local-storage-overview.md
ddysher: Done

contributors/design-proposals/local-storage-overview.md
ddysher: Updated.

contributors/design-proposals/local-storage-overview.md
ddysher: Yes, user should be able to handle this.  It's not clear yet whether kubelet should format it or not.  My understanding is if fsType is given in PVC then user would expect it be formatted by kubernetes; otherwise they'll just get a raw block device.

The whole block storage is under active discussion, I'm only updating the obvious part here.

ddysher:
@msau42 @jingxu97 updated, PTAL

@jsafrane in case you are interested in block device updates

msau42:
/lgtm

Thanks for updating!

PR #705, opened 2017-06-12
make'ing the build requires a null directory CDPATH.

### LINKED ISSUE #704: Document that calling `make` requires CDPATH to include the null directory
Building github.com/kubernetes/kubernetes with a `CDPATH` set to

`.:/my/path`

yields a build error like this:

```
hack/make-rules/../../hack/lib/init.sh: line 39: /Users/treimann/Development/kubernetes/src/k8s.io/kubernetes
/Users/treimann/Development/kubernetes/src/k8s.io/kubernetes/hack/lib/util.sh: No such file or directory
make: *** [all] Error 1
```

If I unset `CDPATH` or remove the leading period (i.e., have it start with a colon), things work fine.

I asked for help on Slack and got the information that the `CDPATH` must be updated accordingly, and I was asked to submit a documentation-updating PR. IMHO, Kubernetes should be able to build without relying to `CDPATH` having some particular format (especially since the bash man-page shows an example with the current directory in the lead). However, a bit of trial-and-error on my side showed that this might not be super easy to achieve. Hence, I suppose updating the docs is in order.

PR to follow.

timoreimann:
Fixes #704.

### DIFF
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -254,13 +254,12 @@ Then edit code on the `myfeature` branch.
 
 #### Build
 
+*Note:* If you are using `CDPATH`, you must either start it with a leading colon, or unset the variable. The make rules and scripts to build require the current directory to come first on the CD search path in order to properly navigate between directories.
+
 ```sh
 cd $working_dir/kubernetes
 make
 ```
-
-(Note that this requires an unset `CDPATH` environment variable or one starting
-with the null directory (i.e., contains a leading colon).
 
 To remove the limit on the number of errors the Go compiler reports (default
 limit is 10 errors):
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -259,6 +259,9 @@ cd $working_dir/kubernetes
 make
 ```
 
+(Note that this requires an unset `CDPATH` environment variable or one starting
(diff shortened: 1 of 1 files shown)


timoreimann:
Signed the CNCF CLA.

cblecker:
/assign
Will review.

contributors/devel/development.md
```diff
@@ -259,6 +259,9 @@ cd $working_dir/kubernetes
 make
 ```
 
+(Note that this requires an unset `CDPATH` environment variable or one starting
```
cblecker:
```
*Note:* If you are using `CDPATH`, you must either start it with a leading colon, or unset the variable. The make rules and scripts to build require the current directory to come first on the CD search path.
```

Feel free to tweak it a bit more, but we want to:
- Start off with the conditional as many people will not even have `CDPATH` set in the first place, so we don't need to state it's a requirement before we establish that they even have `CDPATH` set.
- Explain why this is needed.

cblecker:
/lgtm
@timoreimann Can you double check your CLA status? It's still showing that you haven't signed it.

timoreimann:
I apparently did something wrong on my first pass, but now I received positive feedback the CLA was signed successfully.

Let me know if I need to do anything else.

PR #708, opened 2017-06-12
Proposal: Validation for CustomResources.

nikhita:
ThirdPartyResource (TPR) is deprecated and CustomResourceDefinition (CRD) is the successor which solves the fundamental issues of TPRs to form a stable base for further features.

Currently we do not provide validation for CustomResources (CR). This proposal proposes the design and describes a way to add JSON-Schema based validation for CustomResources.

cc @sttts @deads2k @enisoc @xiao-zhou @erictune @lavalamp @brendanburns @philips @fabxc @frankgreco @sdminonne

### DIFF
--- a/contributors/design-proposals/customresources-validation.md
+++ b/contributors/design-proposals/customresources-validation.md
@@ -1,0 +1,525 @@
+# Validation for CustomResources
+
+Authors: @nikhita, @sttts, some ideas integrated from @xiao-zhou’s proposal<sup id="f1">[1](#footnote1)</sup>
+
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Background](#background)
+    1. [Goals](#goals)
+    2. [Non-Goals](#non-goals)
+3. [Proposed Extension of CustomResourceDefinition](#proposed-extension-of-customresourcedefinition)
+    1. [API Types](#api-types)
+    2. [Examples](#examples)
+        1. [JSON-Schema](#json-schema)
+        2. [Error messages](#error-messages)
+4. [Validation Behavior](#validation-behavior)
+    1. [Metadata](#metadata)
+    2. [Server-Side Validation](#server-side-validation)
+    3. [Client-Side Validation](#client-side-validation)
+    4. [Comparison between server-side and client-side Validation](#comparison-between-server-side-and-client-side-Validation)
+    5. [Existing Instances and changing the Schema](#existing-instances-and-changing-the-schema)
+    6. [Outlook to Status Sub-Resources](#outlook-to-status-sub-resources)
+    7. [Outlook Admission Webhook](#outlook-admission-webhook)
+5. [Implementation Plan](#implementation-plan)
+6. [Appendix](#appendix)
+    1. [Expressiveness of JSON-Schema](#expressiveness-of-json-schema)
+    2. [JSON-Schema Validation Runtime Complexity](#json-schema-validation-runtime-complexity)
+    3. [Alternatives](#alternatives)
+        1. [Direct Embedding of the Schema into the Spec](#direct-embedding-of-the-schema-into-the-spec)
+        2. [External CustomResourceSchema Type](#external-customresourceschema-type)
+
+## Overview
+
+This document proposes the design and describes a way to add JSON-Schema based validation for Custom Resources.
+
+## Background
+
+ThirdPartyResource (TPR) is deprecated and CustomResourceDefinition (CRD) is the successor which solves the fundamental [issues](https://github.com/kubernetes/features/issues/95) of TPRs to form a stable base for further features.
+
+Currently we do not provide validation for CustomResources (CR), i.e. the CR payload is free-form JSON. However, one of the most requested [[1](https://github.com/kubernetes/features/issues/95#issuecomment-296416969)][[2](https://github.com/kubernetes/features/issues/95#issuecomment-298791881)] features is validation and this proposal seeks to add it.
+
+## Goals
+
+1. To provide validation for CustomResources using a declarative specification language for JSON data.
+2. To keep open the door to add other validation mechanisms later.<sup id="f2">[2](#footnote2)</sup>
+3. To allow server-side validation.
+4. To be able to integrate into the existing client-side validation of kubectl.
+5. To be able to define defaults in the specification (at least in a follow-up after basic validation support).
+
+## Non-Goals
+
+1. The JSON-Schema specs can be used for creating OpenAPI documentation for CRs. The format is compatible but we won’t propose an implementation for that.
+2. A turing-complete specification language is not proposed. Instead a declarative way is proposed to express the vast majority of validations.
+3. For now, CRD only allows 1 version at a time. Supporting multiple versions of CRD and/or conversion of CRD is not within the scope of this proposal.
+
+## Proposed Extension of CustomResourceDefinition
+
+We propose to add a field `validation` to the spec of a CustomResourceDefinition. As a first validation format we propose to use [JSON-Schema](http://json-schema.org/) under `CRD.Spec.Validation.JSONSchema`.
+
+JSON-Schema is a [standardized](https://tools.ietf.org/html/draft-zyp-json-schema-04) declarative specification language. Different keywords may be utilized to put constraints on the data. Thus it provides ways to make assertions about what a valid document must look like. 
+
+It is already used in Swagger/OpenAPI specs in Kubernetes and hence such a CRD specification integrates cleanly into the existing infrastructure of the API server which serves these specifications,
+* into kubectl which is able to verify YAML and JSON objects against the returned specification.
+* With the https://github.com/go-openapi/validate library, we have a powerful JSON-Schema validator which can be used client and server-side.
+
+## API Types
+
+The schema is referenced in [`CustomResourceDefinitionSpec`](https://github.com/kubernetes/kubernetes/commit/0304ef60a210758ab4ac43a468f8a5e19f39ff5a#diff-0e64a9ef2cf809a2a611b16fd44d22f8). `Validation` is of the type `CustomResourceValidation`. The JSON-Schema is stored in a field of `Validation`. This way we can make the validation generic and add other validations in the future as well.
+
+The schema types follow those of the OpenAPI library, but we decided to define them independently for the API to have full control over the serialization and versioning. Hence, it is easy to convert our types into those used for validation or to integrate them into an OpenAPI spec later.
+
+Reference http://json-schema.org is also used by OpenAPI. We propose this as there are implementations available in Go and with OpenAPI, we will also be able to serve OpenAPI specs for CustomResourceDefintions.
+
+```go
+// CustomResourceSpec describes how a user wants their resource to appear
+type CustomResourceDefinitionSpec struct {
+    Group string `json:"group" protobuf:"bytes,1,opt,name=group"`
+    Version string `json:"version" protobuf:"bytes,2,opt,name=version"`
+    Names CustomResourceDefinitionNames `json:"names" protobuf:"bytes,3,opt,name=names"`
+    Scope ResourceScope `json:"scope" protobuf:"bytes,8,opt,name=scope,casttype=ResourceScope"`
+    // Validation describes the validation methods for CustomResources
+    Validation CustomResourceValidation `json:"validation,omitempty"`
+}
+
+// CustomResourceValidation is a list of validation methods for CustomResources
+type CustomResourceValidation struct {
+    // JSONSchema is the JSON Schema to be validated against.
+    // Can add other validation methods later if needed.
+    JSONSchema *JSONSchemaProps `json:"jsonSchema,omitempty"`
+}
+
+// JSONSchemaProps is a JSON-Schema following Specification Draft 4 (http://json-schema.org/).
+type JSONSchemaProps struct {
+	ID                   string                     `json:"id,omitempty"`
+	Schema               JSONSchemaURL              `json:"-,omitempty"`
+	Ref                  JSONSchemaRef              `json:"-,omitempty"`
+	Description          string                     `json:"description,omitempty"`
+	Type                 StringOrArray              `json:"type,omitempty"`
+	Format               string                     `json:"format,omitempty"`
+	Title                string                     `json:"title,omitempty"`
+	Default              interface{}                `json:"default,omitempty"`
+	Maximum              *float64                   `json:"maximum,omitempty"`
+	ExclusiveMaximum     bool                       `json:"exclusiveMaximum,omitempty"`
+	Minimum              *float64                   `json:"minimum,omitempty"`
+	ExclusiveMinimum     bool                       `json:"exclusiveMinimum,omitempty"`
+	MaxLength            *int64                     `json:"maxLength,omitempty"`
+	MinLength            *int64                     `json:"minLength,omitempty"`
+	Pattern              string                     `json:"pattern,omitempty"`
+	MaxItems             *int64                     `json:"maxItems,omitempty"`
+	MinItems             *int64                     `json:"minItems,omitempty"`
+	// disable uniqueItems for now because it can cause the validation runtime
+	// complexity to become quadratic.
+	UniqueItems          bool                       `json:"uniqueItems,omitempty"`
+	MultipleOf           *float64                   `json:"multipleOf,omitempty"`
+	Enum                 []interface{}              `json:"enum,omitempty"`
+	MaxProperties        *int64                     `json:"maxProperties,omitempty"`
+	MinProperties        *int64                     `json:"minProperties,omitempty"`
+	Required             []string                   `json:"required,omitempty"`
+	Items                *JSONSchemaPropsOrArray    `json:"items,omitempty"`
+	AllOf                []JSONSchemaProps          `json:"allOf,omitempty"`
+	OneOf                []JSONSchemaProps          `json:"oneOf,omitempty"`
+	AnyOf                []JSONSchemaProps          `json:"anyOf,omitempty"`
+	Not                  *JSONSchemaProps           `json:"not,omitempty"`
+	Properties           map[string]JSONSchemaProps `json:"properties,omitempty"`
+	AdditionalProperties *JSONSchemaPropsOrBool     `json:"additionalProperties,omitempty"`
+	PatternProperties    map[string]JSONSchemaProps `json:"patternProperties,omitempty"`
+	Dependencies         JSONSchemaDependencies     `json:"dependencies,omitempty"`
+	AdditionalItems      *JSONSchemaPropsOrBool     `json:"additionalItems,omitempty"`
+	Definitions          JSONSchemaDefinitions      `json:"definitions,omitempty"`
+}
+
+// JSONSchemaRef represents a JSON reference that is potentially resolved.
+// It is marshaled into a string using a custom JSON marshaller.
+type JSONSchemaRef struct {
+	ReferencePointer JSONSchemaPointer
+	HasFullURL       bool
+	HasURLPathOnly   bool
+	HasFragmentOnly  bool
+	HasFileScheme    bool
+	HasFullFilePath  bool
+}
+
+// JSONSchemaPointer is the JSON pointer representation.
+type JSONSchemaPointer struct {
+	ReferenceTokens []string
+}
+
+// JSONSchemaURL represents a schema url. Defaults to JSON Schema Specification Draft 4.
+type JSONSchemaURL string
+
+const (
+	// JSONSchemaDraft4URL is the url for JSON Schema Specification Draft 4.
+	JSONSchemaDraft4URL SchemaURL = "http://json-schema.org/draft-04/schema#"
+)
+
+// StringOrArray represents a value that can either be a string or an array of strings.
+// Mainly here for serialization purposes.
+type StringOrArray []string
+
+// JSONSchemaPropsOrArray represents a value that can either be a JSONSchemaProps
+// or an array of JSONSchemaProps. Mainly here for serialization purposes.
+type JSONSchemaPropsOrArray struct {
+	Schema      *JSONSchemaProps
+	JSONSchemas []JSONSchemaProps
+}
+
+// JSONSchemaPropsOrBool represents JSONSchemaProps or a boolean value.
+// Defaults to true for the boolean property.
+type JSONSchemaPropsOrBool struct {
+	Allows bool
+	Schema *JSONSchemaProps
+}
+
+// JSONSchemaDependencies represent a dependencies property.
+type JSONSchemaDependencies map[string]JSONSchemaPropsOrStringArray
+
+// JSONSchemaPropsOrStringArray represents a JSONSchemaProps or a string array.
+type JSONSchemaPropsOrStringArray struct {
+	Schema   *JSONSchemaProps
+	Property []string
+}
+
+// JSONSchemaDefinitions contains the models explicitly defined in this spec.
+type JSONSchemaDefinitions map[string]JSONSchemaProps
+```
+
+Note: A reflective test to check for drift between the types here and the OpenAPI types for runtime usage will be added.
+
+## Examples
+
+### JSON-Schema
+
+The following example illustrates how a schema can be used in `CustomResourceDefinition`. It shows various restrictions that can be achieved for validation using JSON-Schema.
+
+```json
+{
+    "apiVersion": "apiextensions.k8s.io/v1beta1",
+    "kind": "CustomResourceDefinition",
+    "metadata": {
+        "name": "noxus.mygroup.example.com"
+    },
+    "spec": {
+        "group": "mygroup.example.com",
+        "version": "v1alpha1",
+        "scope": "Namespaced",
+        "names": {
+            "plural": "noxus",
+            "singular": "noxu",
+            "kind": "Noxu",
+            "listKind": "NoxuList"
+        },
+        "validation": {
+            "jsonSchema": {
+                "$schema": "http://json-schema.org/draft-04/schema#",
+                "type": "object",
+                "description": "Noxu is a kind of Custom Resource which has only fields that are specified",
+                "required": [
+                    "alpha",
+                    "beta",
+                    "gamma",
+                    "delta",
+                    "epsilon",
+                    "zeta"
+                ],
+                "properties": {
+                    "alpha": {
+                        "description": "Alpha is an alphanumeric string with underscores which defaults to foo_123",
+                        "type": "string",
+                        "pattern": "^[a-zA-Z0-9_]*$",
+                        "default": "foo_123"
+                    },
+                    "beta": {
+                        "description": "We need at least 10 betas. If not specified, it defaults to 10.",
+                        "type": "number",
+                        "minimum": 10,
+                        "default": 10
+                    },
+                    "gamma": {
+                        "description": "Gamma is restricted to foo, bar and baz",
+                        "type": "string",
+                        "enum": [
+                            "foo",
+                            "bar",
+                            "baz"
+                        ]
+                    },
+                    "delta": {
+                        "description": "Delta is a string with a maximum length of 5 or a number with a minimum value of 0",
+                        "anyOf": [
+                            {
+                                "type": "string",
+                                "maxLength": 5
+                            },
+                            {
+                                "type": "number",
+                                "minimum": 0
+                            }
+                        ]
+                    },
+                    "epsilon": {
+                        "description": "Epsilon is either of type one zeta or two zeta",
+                        "allOf": [
+                            {
+                                "$ref": "#/definitions/zeta"
+                            },
+                            {
+                                "properties": {
+                                    "type": {
+                                        "enum": [
+                                            "one",
+                                            "two"
+                                        ]
+                                    }
+                                },
+                                "required": [
+                                    "type"
+                                ],
+                                "additionalProperties": false
+                            }
+                        ]
+                    },
+                    "additionalProperties": false,
+                    "definitions": {
+                        "zeta": {
+                            "description": "Every zeta needs to have foo, bar and baz",
+                            "type": "object",
+                            "properties": {
+                                "foo": {
+                                    "type": "string"
+                                },
+                                "bar": {
+                                    "type": "number"
+                                },
+                                "baz": {
+                                    "type": "boolean"
+                                }
+                            },
+                            "required": [
+                                "foo",
+                                "bar",
+                                "baz"
+                            ],
+                            "additionalProperties": false
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+### Error messages
+
+The following examples illustrate the type of validation errors generated by using the go-openapi validate library.
+
+The description is not taken into account, but a better error output can be easily [added](https://github.com/go-openapi/errors/blob/master/headers.go#L23) to go-openapi.
+
+1. `data.foo in body should be at least 4 chars long`
+2. `data.foo in body should be greater than or equal to 10`
+3. `data.foo in body should be one of [bar baz]`
+4. `data.foo in body must be of type integer: "string"`
+5. `data.foo in body should match '^[a-zA-Z0-9_]*$'`
+6. `data.foo in body is required`
+7. When foo validates if it is a multiple of 3 and 5:
+```
+data.foo in body should be a multiple of 5
+data.foo in body should be a multiple of 3
+must validate all the schemas (allOf)
+```
+
+## Validation Behavior
+
+The schema will be described in the `CustomResourceDefinitionSpec`. The validation will be carried out using the [go-openapi validation library](https://github.com/go-openapi/validate).
+ 
+While creating/updating the CR, the metadata is first validated. To validate the CR against the spec in the CRD, we _must_ have server-side validation and we _can_ have client-side validation.
+
+### Metadata
+
+ObjectMeta and TypeMeta are implicitly specified. They do not have to be added to the JSON-Schema of a CRD. The validation already happens today as part of the apiextensions-apiserver REST handlers.
+
+### Server-Side Validation
+
+The server-side validation is carried out after sending the request to the apiextensions-apiserver, i.e. inside the CREATE and UPDATE handlers for CRs.
+
+We do a schema pass there using the https://github.com/go-openapi/validate validator with the provided schema in the corresponding CRD. Validation errors are returned to the caller as for native resources.
+
+JSON-Schema also allows us to reject additional fields that are not defined in the schema and only allow the fields that are specified. This can be achieved by using `"additionalProperties": false` in the schema. However, there is danger in allowing CRD authors to set `"additionalProperties": false` because it breaks version skew (new client can send new optional fields to the old server). So we should not allow CRD authors to set `"additionalProperties": false`.
+
+### Client-Side Validation
+
+The client-side validation is carried out before sending the request to the api-server, or even completely offline. This can be achieved while creating resources through the client i.e. kubectl using the --validate option.
+
+If the API type serves the JSON-Schema in the swagger spec, the existing kubectl code will already be able to also validate CRs. This will be achieved as a follow-up.
+
+### Comparison between server-side and client-side Validation
+
+The table below shows the cases when server-side and client-side validation methods are applicable.
+
+| Case                                               | Server-Side  | Client-Side   |
+|:--------------------------------------------------:|:------------:|:-------------:|
+| Kubectl create/edit/replace with validity feedback | &#x2713;     | &#x2713;      | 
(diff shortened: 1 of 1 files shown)


deads2k:
@sdminonne

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+    7. [Outlook Admission Webhook](#webhooks)
+5. [Implementation Plan](#implementation)
+6. [Appendix](#appendix)
+    1. [Expressiveness of JSON-Schema](#expressiveness)
+    2. [JSON-Schema Validation Runtime Complexity](#complexity)
+    3. [Alternatives](#alternatives)
+        1. [Direct Embedding of the Schema into the Spec](#direct-embed)
+        2. [External CustomResourceSchema Type](#external-type)
+
+## <a name="overview"></a>Overview
```
fabiand: You might want to stick to markdown only, thus remove the <a> anchor.

enisoc:
cc @mbohlool

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+	Property []string
+}
+
+// JSONSchemaDefinitions contains the models explicitly defined in this spec.
+type JSONSchemaDefinitions map[string]JSONSchemaProps
+```
+
+Note: A reflective test to check for drift between the types here and the OpenAPI types for runtime usage will be added.
+
+## JSON-Schema Examples
```
enisoc: Could you also include some examples of what validation errors would look like in this proposal? For example, does the description field allow the author to inject hints into the error message?

One trade-off between this approach (using a JSON-Schema library) and alternatives (e.g. webhooks) is that the CRD author has less control over the content of validation error messages. Of course we're not limited to just printing whatever the library spits out, but if more work will be needed to improve on what the library gives us, we should scope it out here since telling the user how to fix the problem is arguably the most important part of validation.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+| Kubectl create/edit/replace with validity feedback | &#x2713;     | &#x2713;      | 
+| Custom controller creates/updates CRs              | &#x2713;     | &#x2717;      |
+| CRs are created by an untrusted party              | &#x2713;     | &#x2717;      |
+| Not making validation for CRs a special case       | &#x2713;     | &#x2717;      |
+
+The above table is an evidence that we need server-side validation as well, next to the client-side validation we easily get, nearly for free, by serving Swagger/OpenAPI specs in apiextension-apiserver. 
+
+This is especially true in situations when CRs are used by components that are out of the control of the admin. Example: A user can create a database CR for a Database-As-A-Service. In this case, only server-side validation can give confidence that the CRs are well formed.
+
+### Existing Instances and changing the Schema
```
enisoc: In the Google Doc, there was some discussion about ways to deal with CRs that exist before the introduction of validation on a given CRD for the first time. Can you summarize the findings in this proposal?

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+This validates: 
+* `{"version": "a",  "spec": {"foo": 42}}`
+* `{"version": "b",  "spec": {"bar": 42}}`
+
+but not:
+* `{"version": "a",  "spec": {"bar": 42}}`.
+ 
+Note: this is a workaround while we do not support multiple versions and conversion for custom resources.
+
+### JSON-Schema Validation Runtime Complexity
```
enisoc: In general, it seems like this will be slower than native validation. We should consult with @kubernetes/sig-scalability-proprosals to define the performance bar this needs to meet and create regression tests for large numbers of objects.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+Following “JSON: data model, query languages and schema specification<sup id="f4">[4](#footnote4)</sup>” and “Formal Specification, Expressiveness and Complexity analysis for JSON Schema<sup id="f5">[5](#footnote5)</sup>”, JSON-Schema validation
+* without the uniqueItems operator and 
+* without recursion for the $ref operator
+has linear runtime in the size of the JSON input and the size of the schema (Th. 1 plus Prop. 7).
+ 
+If we allow uniqueItems, the runtime complexity becomes quadratic in the size of the JSON input. Hence, we might want to consider forbidding the uniqueItems operator in order to avoid DDoS attacks, at least if the schema definitions of CRDs cannot be trusted.
+
+The CRD JSON-Schema will be validated to have neither recursion, nor `uniqueItems=true` being set.
+
+### Alternatives
```
enisoc: I think many people will see imperative validation (custom code inserted in the request path somehow) as the main alternative to this proposal. You mentioned above they are not mutually exclusive, but it would be good to document here why we think it makes sense to put our effort in creating the declarative version first.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+
+ObjectMeta and TypeMeta are implicitly specified. They do not have to be added to the JSON-Schema of a CRD. The validation already happens today as part of the apiextensions-apiserver REST handlers.
+
+### <a name="server-side"></a>Server-Side Validation
+
+The server-side validation is carried out after sending the request to the apiextensions-apiserver, i.e. inside the CREATE and UPDATE handlers for CRs.
+
+We do a schema pass there using the https://github.com/go-openapi/validate validator with the provided schema in the corresponding CRD. Validation errors are returned to the caller as for native resources.
+
+JSON-Schema also allows us to reject additional fields that are not defined in the schema and only allow the fields that are specified. This can be achieved by using `"additionalProperties": false` in the schema.
```
bgrant0607: Server-side validation needs to be optional:
https://github.com/kubernetes/kubernetes/issues/5889#issuecomment-178742763

bgrant0607: That is, schema validation needs to be optional. Value validation does not.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+
+This document proposes the design and describes a way to add JSON-Schema based validation for Custom Resources.
+
+## Background
+
+ThirdPartyResource (TPR) is deprecated and CustomResourceDefinition (CRD) is the successor which solves the fundamental issues of TPRs to form a stable base for further features.
+
+Currently we do not provide validation for CustomResources (CR), i.e. the CR payload is free-form JSON. However, one of the most requested [[1](https://github.com/kubernetes/features/issues/95#issuecomment-296416969)][[2](https://github.com/kubernetes/features/issues/95#issuecomment-298791881)] features is validation and this proposal seeks to add it.
+
+## Goals
```
enisoc: I just realized there are examples of giving fields default values below. Is it a goal of this proposal to support defaulting? Either way, defaulting should be added to either Goals or Non-Goals to make it clear.

contributors/design-proposals/customresources-validation.md
enisoc: To elaborate, my interpretation of the chain of comments @bgrant0607 linked is: Yes, there is danger in allowing CRD authors to set `additionalProperties: false` because it breaks version skew (new client sends new optional field to old server). At most, the client should be able to opt-in to "reject unrecognized fields" on a per-request basis. However, there is currently no such standard query param (AFAIK), so we should not allow CRD authors to set `additionalProperties: false`.

bgrant0607:
Original description of this GSOC project:
https://summerofcode.withgoogle.com/organizations/6540924424290304/#5982049109278720

bgrant0607:
Ref https://github.com/kubernetes/kubernetes/issues/38117

contributors/design-proposals/customresources-validation.md
bgrant0607: Schema validation performed by kubectl is driven by Swagger/OpenAPI models. An equivalent capability needs to move to the server (https://github.com/kubernetes/kubernetes/issues/5889) as part of the general effort to move functionality to the server in order to improve extensibility and to simplify clients (https://github.com/kubernetes/kubernetes/issues/12143).

I would like to see us move towards declarative validation generally for the API, as much as possible:
https://github.com/kubernetes/kubernetes/issues/25460

@mbohlool may have already started on a proposal for that.

Also, ideally, I'd like the same spec to be usable to serve an OpenAPI spec for the CRD APIs, which would mean it would have to be an OpenAPI-compatible flavor of JSON schema. I really don't want multiple different schema languages in the system.

As for a hook-based approach, (potentially multiple) special-purpose hooks per resource would be harder to understand, as well as being more fragile, slower, etc. I'd prefer to find a way to use admission-control extension for that use case, so as not to add yet another hook mechanism:

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/admission_control_extension.md

bgrant0607:
@kubernetes/sig-api-machinery-proposals

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+
+1. To provide validation for CustomResources using a declarative specification language for JSON data.
+2. To keep open the door to add other validation mechanisms later.<sup id="f2">[2](#footnote2)</sup>
```
fabiand: You might want to drop the `<sup>` tag

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+
+## Non-Goals
+
+1. The JSON-Schema specs can be used for creating OpenAPI documentation for CRs. The format is compatible but we won’t propose an implementation for that.
+2. A turing-complete specification language is not proposed. Instead a declarative way is proposed to express 90% of validations.
+3. For now, CRD only allows 1 version at a time. Supporting multiple versions of CRD and/or conversion of CRD is not within the scope of this proposal.
+
+## Proposed Extension of CustomResourceDefinition
+
+We propose to add a field `validation` to the spec of a CustomResourceDefinition. As a first validation format we propose to use [JSON-Schema](http://json-schema.org/) under `CRD.Spec.Validation.JSONSchema`.
```
fabiand: `validation` sounds pretty generic, solely from the term this could also refer to webhook links.

What about using the field name `scheme` or `schema` instead?

contributors/design-proposals/customresources-validation.md
sttts: This is certainly slower than hand crafted validation. The main point of this paragraph is not that's it might be slower by a factor of 10 (a guess), but that it cannot be used to bring down an API server because the validation algorithm is quadratic of even exponential in the input size.

But of course, having a performance bar with tests would be great.

contributors/design-proposals/customresources-validation.md
sttts: No need for another hook mechanism. We don't want that. I verified that CRDs play nicely with initializers in 1.7, same should be true for admission webhooks (not verified).

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,493 @@
...
+
+The webhook admission prototyped at https://github.com/kubernetes/kubernetes/pull/46316 is expected to work with CRs out of the box. Hence, for more advanced validation the webhook admission is an option as well (when it is merged).
+
+Clearly, providing such a webhook is more complex and therefore cannot replace a built-in JSON-Schema validation and its use-cases.
+
+## Implementation Plan
+
+The implementation is planned in the following steps:
+
+1. Add the proposed types to the v1beta1<sup id="f3">[3](#footnote3)</sup> version of the CRD type.
```
fabiand: You probably want to drop `<sup>` here as well.

contributors/design-proposals/customresources-validation.md
nikhita: Added examples of validation errors.

contributors/design-proposals/customresources-validation.md
nikhita: Updated to not allow CRD authors to set `additionalProperties: false`.

contributors/design-proposals/customresources-validation.md
nikhita: Added.

contributors/design-proposals/customresources-validation.md
nikhita: Added a note about intializers.

contributors/design-proposals/customresources-validation.md
sttts: I went into go-openapi this morning and prototyped defaulting support: https://github.com/go-openapi/validate/pull/27.

So this is not available currently upstream, but is feasible. The only support that exists is that errors of required fields are surpressed if a default value is given. My PR adds support to actually take that default and write it into the field.

sttts: @nikhita has added a link below to that prototype below.

contributors/design-proposals/customresources-validation.md
nikhita: Github Flavored Markdown does not support footnotes without the `<sup>` tags.

Please let me know if this is indeed possible and I'm missing something here!

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+	ExclusiveMaximum     bool                       `json:"exclusiveMaximum,omitempty"`
+	Minimum              *float64                   `json:"minimum,omitempty"`
+	ExclusiveMinimum     bool                       `json:"exclusiveMinimum,omitempty"`
+	MaxLength            *int64                     `json:"maxLength,omitempty"`
+	MinLength            *int64                     `json:"minLength,omitempty"`
+	Pattern              string                     `json:"pattern,omitempty"`
+	MaxItems             *int64                     `json:"maxItems,omitempty"`
+	MinItems             *int64                     `json:"minItems,omitempty"`
+	// disable uniqueItems for now because it can cause the validation runtime
+	// complexity to become quadratic.
```
deads2k: CRDs are inherently a cluster-admin resource since they affect the resources for the entire cluster.  I don't feel bad about allowing a cluster-admin to make an expensive resource.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+
+We do a schema pass there using the https://github.com/go-openapi/validate validator with the provided schema in the corresponding CRD. Validation errors are returned to the caller as for native resources.
+
+JSON-Schema also allows us to reject additional fields that are not defined in the schema and only allow the fields that are specified. This can be achieved by using `"additionalProperties": false` in the schema. However, there is danger in allowing CRD authors to set `"additionalProperties": false` because it breaks version skew (new client can send new optional fields to the old server). So we should not allow CRD authors to set `"additionalProperties": false`.
+
+### Client-Side Validation
+
+The client-side validation is carried out before sending the request to the api-server, or even completely offline. This can be achieved while creating resources through the client i.e. kubectl using the --validate option.
+
+If the API type serves the JSON-Schema in the swagger spec, the existing kubectl code will already be able to also validate CRs. This will be achieved as a follow-up.
```
deads2k: @mbohlool this is going to require changes to the openapi aggregation triggers.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+
+If the schema is made stricter later, the existing CustomResources might no longer comply with the spec. This will make them unchangeable and essentially read-only.
+
+To avoid this, it is the responsibility of the user to make sure that any changes made to the schema are such that the existing CustomResources remain validated. 
+ 
+Note:
+
+1. This is the same behavior that we require for native resources. Validation cannot be made stricter in later Kubernetes versions without breaking compatibility.
+
+2. For migration of CRDs with no validation to CRDs with validation, we can create a controller that will validate and annotate invalid CRs once the spec changes, so that the custom controller can choose to delete them (this is also essentially the status condition of the CRD). This can be achieved, but it is not part of the proposal.
```
deads2k: I'm not really a fan of this.  It seems like it should mirror kube-apis: it's the responsibility of the APi author and fixups are performed client-side.

contributors/design-proposals/customresources-validation.md
sttts: Will cluster-admins understand complexity?

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+It is [verified](https://github.com/kubernetes/kubernetes/pull/47252) that CRDs work well with initializers. It is also expected that webhook admission prototyped at https://github.com/kubernetes/kubernetes/pull/46316 will work with CRs out of the box. Hence, for more advanced validation webhook admission is an option as well (when it is merged).
+
+Clearly, providing such a webhook is more complex and therefore cannot replace a built-in JSON-Schema validation and its use-cases.
+
+## Implementation Plan
+
+The implementation is planned in the following steps:
+
+1. Add the proposed types to the v1beta1<sup id="f3">[3](#footnote3)</sup> version of the CRD type.
+2. Add a validation step to the CREATE and UPDATE REST handlers of the apiextensions-apiserver.
```
deads2k: I think you actually just want to update the strategy.

sttts: yes, that's actually meant here.

deads2k:
minor comments.  This lgtm.

contributors/design-proposals/customresources-validation.md
deads2k: > Will cluster-admins understand complexity?

Will they care?  I suspect that first line of defense would simply be to scale out the masters.

contributors/design-proposals/customresources-validation.md
nikhita: Yes, will be passing the CRD to the strategy.

sdminonne:
I liked the declarative approach defining validation but I would love to have a way to reuse definitions similar to [swagger multiple files](https://github.com/OAI/OpenAPI-Specification/blob/master/guidelines/REUSE.md)
Even better would be to reuse swagger as already reported by @bgrant0607-nocc 

The same mechanism should be used to reference _native resources_ like PodSpec. 
To be more clear: If my CDR contains a `v1.PodSpec`  I would like references the swagger definition of the `io.k8s.kubernetes.pkg.api.v1.PodSpec` avoiding cut & paste of the swagger.json.
 
According to @sttts this mechanism may be added in a second step (brief discussion on slack). So I'm in general OK with this approach.

sttts:
Adding to @sdminonne's comment:

The languages of native kube object specs in swagger and those proposed here are the same. Hence, a schema reference could be used to link a CRD to a native object spec, using a JSON-Schema reference with a certain URL.

The main conceptual problem to solve is to define this URL namespace. Note that we are coupling non-kube-CRDs with the kube type universe, maybe even with a number of user provided API servers. One could use the service names of those API servers to reference their specs, e.g. https://kubernetes/api/v1/pods/schema.json and https://kubernetes/apis/you.group.com/your-custom-resources/schema.json. But this needs deep thought as we are coupling independent components, after splitting them apart into single API servers with a lot of effort.

Hence, I see this as a follow-up to this proposal. Right now, we will keep schema references local and the schemas self-contained.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+```
+Then we can validate the status against the sub-schema easily. Hence, this proposal will be compatible with a later sub-resource extension.
+
+### Outlook Admission Webhook
+
+Apiextensions-apiserver uses the normal REST endpoint implementation and only customizes the registry and the codecs. The admission plugins are inherited from the kube-apiserver (when running inside of it via apiserver delegation) and therefore they are supposed to apply to CRs as well.
+
+It is [verified](https://github.com/kubernetes/kubernetes/pull/47252) that CRDs work well with initializers. It is also expected that webhook admission prototyped at https://github.com/kubernetes/kubernetes/pull/46316 will work with CRs out of the box. Hence, for more advanced validation webhook admission is an option as well (when it is merged).
+
+Clearly, providing such a webhook is more complex and therefore cannot replace a built-in JSON-Schema validation and its use-cases.
```
brendandburns: I don't think this is true. In terms of both cognitive complexity and lines of code, implementing a webhook that gets a JSON payload and validates soem fields is _way_ more straightforward than learning to write JSON-Schema...

Take for example wanting to validate that a field ('account') is present and non-negative, here's the webhook:

```javascript
var validateHandler = function(context) {
   if (!context.body.account) {
       return error('account field is required');
   }
   if (context.body.account < 0) {
       return error('account field must be non-negative')
    }
}
```

I personally believe we should support webhook validation in the CRD definition at the start, but its not a requirement.

contributors/design-proposals/customresources-validation.md
brendandburns: Commented above. The commentary assumes/implies that webhooks are harder and less useful than json-schema. I think exactly the opposite is true.

Personally, I would argue for webhook validation in the CRD resource because conceptually it is part of the resource. While it is possible to do it in an extension admission controller, the user experience is way worse, because suddenly I have to create/manage/synchronize two different objects to implement my desired experience.

Instead, I think we should add webhooks to the CRD, and then implement a controller that reflects that webhook into an admission control extension.  (similar to how `ReplicaSet` creates `Pod` objects, etc)

IMHO we shouldn't mix our desires around implementation complexity (only one kind of webhook) with our designs around API complexity (everything relating to a particular object should be in that object)

philips:
cc @ericchiang @lpabon @xiang90

contributors/design-proposals/customresources-validation.md
frankgreco: Also, webhook validation allows you to dynamically validate specs where as I'm assuming that JSON-Schema route is more static.

deads2k: The way this is structured, the JSON-schema does not preclude other validation from coming later, but json schema does provide

1. useful validation
2. no external call latency in the API server
3. a standard structure that doesn't require setting up an external service
4. no "fail open/fail closed" decision
5. a structure converting to our resource schema description like every other resource

This is an important step regardless of future action.

brendandburns: Agree that this is useful and not trying to block the implementation. However, I _do_ want to make sure the language in the proposal is accurate, and that it suggests the right path forward for people who come along later and think about webhook based validation.

It's also important to note that for most validation that people actually want to do this is going to be insufficient, since its going to be things like "field foo has to be a dns name" or "field bar is only present if field baz is not"

Again, perfectly happy to see the work here get done, just want to make sure we paint the right picture for what is to come in the future.

nikhita: @brendanburns Thanks, I agree with what you said - and webhook based validation can definitely be implemented at a later stage. Updated the proposal to reflect this. PTAL.

enisoc: > field foo has to be a dns name

JSON Schema might be able to do this with either a [hostname format](http://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3.3) or a regexp [pattern](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.8), assuming the library we use supports those, and assuming it's possible to match the k8s definition of a valid name.

> field bar is only present if field baz is not

I think is is possible with [oneOf](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.28) or with [dependencies](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.21).

I do agree it's more cognitive load to think in these terms (a win for imperative/webhook), but it doesn't seem quite true that JSON Schema is going to be "insufficient" for what people will want to do.

brendandburns: (nb: I'm not a swagger/openapi expert)

Most real world validation beyond simple syntactic tests (does it parse, are all fields valid) that I've ever seen has been a programmatic.

Which actually points to a bigger point, you may want to have _both_ JSON-Schema and webhook, we should clarify that it should be possible to have multiple validation methods specified.

brendandburns:
I commented above, but a thing that just occurred to me is that we will want to support multiple different validations all running for the same CRD. That's because Schematic and Programattic validation are both valuable and users will likely want to combine them.

frankgreco:
With the declarative/static approach to validation, how would dynamic fields be checked. An example of this is what the `Service` spec does to check the uniquess of hard-coded `NodePort`s. I use dynamic validation all the time for TPRs.

deads2k:
> With the declarative/static approach to validation, how would dynamic fields be checked.

This proposal explicitly notes that multiple validation approaches are a design goal https://github.com/kubernetes/community/pull/708/files#diff-9679c445e428c00f0907fa6fe5c6d8e4R46, but intentionally does not include that implementation.  The CRD schema is structured to make adding that in the future easy/reasonable to do, but it can come later at a different time.

brendanburns:
But it doesn't (as far as I could find) describe that multiple different
validations could be active at the same time. The API object should
probably be structured as an array of validation objects, not a single
object...

deads2k:
> But it doesn't (as far as I could find) describe that multiple different
> validations could be active at the same time. The API object should
> probably be structured as an array of validation objects, not a single
> object...

The the API has a validation stanza, which currently has one element for JSONSchema and could easily be existed to have another element called "webhook" or some such.

brendanburns:
Right, but my point is that I may actually want _multiple_ web hooks, or a
JSON schema and a webhook.  I think an array captures this more elegantly
than a single top-level struct.

deads2k:
> Right, but my point is that I may actually want _multiple_ web hooks, or a
> JSON schema and a webhook.  I think an array captures this more elegantly
> than a single top-level struct.

The current struct does not preclude introducing a slice of webhooks.  Your example also highlights a deficiency of an array of validators, since some combinations like multiple JSON-schemas don't make sense.  Rather than enforce that in resource validation, I would make the struct represent the valid combinations.

agonzalezro:
If multiple validations are added I suppose it's going to be a new struct field here: `CRD.Spec.Validation.*`.

Is the order of this validations important? For example, somebody would like to run JSON Schema validations before calling a webhook.

In case it's important what do you think about something like this (please, take this as pseudo code 😬):

```
type CustomResourceDefinitionSpec struct {
   ...
   Validations []CustomResourceValidation `json:"validation,omitempty"`
}

type CustomResourceValidation interface{
  Validate() []err
}
```

This would allow us to even repeat validations which can be good and bad at the same time:

- good: we could validate with JSON Schema, webhook A (microservice from dev team), webhook B (microservice from finance team -> billing).
- bad: you could as well add several JSON Schemas validation to the same CRD that could end up fighting between them (one returning true and the other false).

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+1. To provide validation for CustomResources using a declarative specification language for JSON data.
+2. To keep open the door to add other validation mechanisms later.<sup id="f2">[2](#footnote2)</sup>
+3. To allow server-side validation.
+4. To be able to integrate into the existing client-side validation of kubectl.
+5. To be able to define defaults in the specification (at least in a follow-up after basic validation support).
+
+## Non-Goals
+
+1. The JSON-Schema specs can be used for creating OpenAPI documentation for CRs. The format is compatible but we won’t propose an implementation for that.
+2. A turing-complete specification language is not proposed. Instead a declarative way is proposed to express 90% of validations.
```
jamiehannaford: Is 90% literal here, or just an example?

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+        1. [Direct Embedding of the Schema into the Spec](#direct-embedding-of-the-schema-into-the-spec)
+        2. [External CustomResourceSchema Type](#external-customresourceschema-type)
+
+## Overview
+
+This document proposes the design and describes a way to add JSON-Schema based validation for Custom Resources.
+
+## Background
+
+ThirdPartyResource (TPR) is deprecated and CustomResourceDefinition (CRD) is the successor which solves the fundamental issues of TPRs to form a stable base for further features.
```
jamiehannaford: Is it outside the scope of this doc to briefly mention what those issues are? Perhaps just as links

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+
+JSON-Schema is a [standardized](https://tools.ietf.org/html/draft-zyp-json-schema-04) declarative specification language. Different keywords may be utilized to put constraints on the data. Thus it provides ways to make assertions about what a valid document must look like. 
+
+It is already used in Swagger/OpenAPI specs in Kubernetes and hence such a CRD specification integrates cleanly into the existing infrastructure of the API server which serves these specifications,
+* into kubectl which is able to verify YAML and JSON objects against the returned specification.
+* With the https://github.com/go-openapi/validate library, we have a powerful JSON-Schema validator which can be used client and server-side.
+
+## API Types
+
+The schema is referenced in [`CustomResourceDefinitionSpec`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1/types.go#L22). `Validation` is of the type `CustomResourceValidation`. The JSON-Schema is stored in a field of `Validation`. This way we can make the validation generic and add other validations in the future as well.
```
jamiehannaford: Nit: maybe we can pin to a specific commit in the URL? I've lost track of the amount of times I've followed master links and the lines are outdated 😄

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+
+// CustomResourceValidation is a list of validation methods for CustomResources
+type CustomResourceValidation struct {
+    // JSONSchema is the JSON Schema to be validated against.
+    // Can add other validation methods later if needed.
+    JSONSchema *JSONSchemaProps `json:"jsonSchema,omitempty"`
+}
+
+// JSONSchemaProps is a JSON-Schema following Specification Draft 4 (http://json-schema.org/).
+type JSONSchemaProps struct {
```
jamiehannaford: Could we use a json schema lib like [https://github.com/xeipuuv/gojsonschema](https://github.com/xeipuuv/gojsonschema) instead of defining this ourselves? I know folks are wary of more imports but just thought I'd float the idea

contributors/design-proposals/customresources-validation.md
nikhita: Will add them. :+1:

contributors/design-proposals/customresources-validation.md
nikhita: Just an example.

contributors/design-proposals/customresources-validation.md
nikhita: good point. :)
will do.

contributors/design-proposals/customresources-validation.md
nikhita: I believe we don't really want to import here and end up making our types depend on external types.

Had considered using https://github.com/xeipuuv/gojsonschema but instead decided to go with go-openapi because we want to be able to express the schema via openapi docs.

nikhita:
@agonzalezro That looks like a nice approach but what I am concerned about is that we can end up with multiple JSON schemas and allow invalid combinations. We need the JSON schema to only have one entry point (not multiple).

contributors/design-proposals/customresources-validation.md
jamiehannaford: Ah okay, can we change to `express the vast majority of validations` or something? Just for folks who might take this literally

contributors/design-proposals/customresources-validation.md
jamiehannaford: Yeah, it's also defined [here](https://github.com/go-openapi/spec/blob/master/schema.go#L202). Just wondering if we can avoid duplication and the risk of feature drift.

@liggitt should we be avoiding external types here?

liggitt: external types don't have the protobuf annotations needed by our API types. tying our API to an external API is not something we want to do.

contributors/design-proposals/customresources-validation.md
```diff
@@ -0,0 +1,525 @@
...
+
+JSON-Schema is a [standardized](https://tools.ietf.org/html/draft-zyp-json-schema-04) declarative specification language. Different keywords may be utilized to put constraints on the data. Thus it provides ways to make assertions about what a valid document must look like. 
+
+It is already used in Swagger/OpenAPI specs in Kubernetes and hence such a CRD specification integrates cleanly into the existing infrastructure of the API server which serves these specifications,
+* into kubectl which is able to verify YAML and JSON objects against the returned specification.
+* With the https://github.com/go-openapi/validate library, we have a powerful JSON-Schema validator which can be used client and server-side.
+
+## API Types
+
+The schema is referenced in [`CustomResourceDefinitionSpec`](https://github.com/kubernetes/kubernetes/commit/0304ef60a210758ab4ac43a468f8a5e19f39ff5a#diff-0e64a9ef2cf809a2a611b16fd44d22f8). `Validation` is of the type `CustomResourceValidation`. The JSON-Schema is stored in a field of `Validation`. This way we can make the validation generic and add other validations in the future as well.
```
sttts: To address @brendanburns's comment https://github.com/kubernetes/community/pull/708#discussion_r122353129:
```
This way we can make the validation generic and add other validation options in the future. Even multiple validation options could be used at the same time within one CustomResourceDefinition.
```

sttts:
I can see value in @agonzalezro's example with multiple webhooks. Also the possibility to rely on static JSON-schema validation before calling a webhook sounds reasonable. 

Though I am not so sure about introducing any more structure, i.e. a whole slice of validation mechanisms. As the next step we would want to define the parallalism, i.e. sequential or in parallel, eventually leading to complete pipeline semantics. I agree with @deads2k that having a `struct { JSONSchema, []WebHook }` is a good line to draw. If some day we have some embedded interpreter of some sort, we could have `struct { JSONSchema, Script, []WebHook }`.

agonzalezro:
I see `struct { JSONSchema, []WebHook }` reasonable. In the case that it's EXTREMELY important for somebody to to run the webhooks before the schema they could implement the schema validation in their latest webhook. Of course, this case would be like the 0.1% of the people :)

Looks pretty good, good job you all.

deads2k:
lgtm

frankgreco:
I am very interested in contributing to this change. Is this the right issue to be following for that. I assume at some point this thread will turn into more of implementation discussion once the final design is finished.

enisoc:
@frankgreco There is work in progress here: https://github.com/kubernetes/kubernetes/pull/47263

deads2k:
> I am very interested in contributing to this change. Is this the right issue to be following for that. I assume at some point this thread will turn into more of implementation discussion once the final design is finished.

I figured I'd give a day for closing arguments and merge this tomorrow.

deads2k:
> I figured I'd give a day for closing arguments and merge this tomorrow.

Ok.  I'll merge this as a starting point.  Specific issues are welcome as are comments on the WIP https://github.com/kubernetes/kubernetes/pull/47263

contributors/design-proposals/customresources-validation.md
mbohlool: Sorry that I see this late. I have two points here:

1. If we register CRDs with aggregation server (which I think we should), this will be aggregated automatically.
2. OpenAPI supports a subset of json schema. I think instead of asking users to provide JSON Schema we should ask them to provide OpenAPI spec that has a Definition model in it (that is a subset of JSON Schema).

sttts: about 2) why should we restrict ourselfes to openapi? The openapi spec is a side-product of this validation and I don't want to loose expressivity because openapi hasn't updated to the JSON schema draft 4 yet. 

Moreover, some day openapi v3 is finalized, we will support v3 and v2 for backward compatibility. Do you suggest that we stay at the common denominator, i.e. to that JSON schema subset supported by v2? In other words, at some point we will have to diverge anyway.

Is there anything that stops us from stripping the given spec from unsupported features for consumation by the openapi aggregator? As far as I see all JSON Schema fields are monotonic in the sense that stripping fields makes the spec weaker. That's what we need for easy openapi consumation.

mbohlool:
OpenAPI v3 is release and we are going to support it but even v3 Schema is
a subset of JSON schema. I don't think OpenAPI supports complete JSON
schema any time soon if ever.

If we provide JSON schema for CRDs there is no path forward for converting
it to OpenAPI (hence at least the proposal's assertion that it is
compatible with OpenAPI is false). If we are thinking of also asking users
to provide OpenAPI spec that would be a bad design in my opinion. Two very
similar things. I don't think OpenAPI's Schema is that restrictive and I do
believe we should use that as it widely accepted (not that JSON schema is
not but for the puropose of describing API objects, OpenAPI designed for
that).

sttts:
@mbohlool let's continue in the issue https://github.com/kubernetes/kubernetes/issues/49879

PR #695, opened 2017-06-07
Device Plugin Design Proposal

RenaudWasTaken:
Notes for reviewers

First proposal submitted to the community repo, please advise if something's not right with the format or procedure, etc.

cc @vishh  @derekwaynecarr

PR #713, opened 2017-06-13
Proposal: Host-side Volume Discovery and Management Library

leonwanghui:
As we all know, Kubernetes can be deployed anywhere, including various cloud platforms and bare metals. But as for storage resources, you have to choose specific storage backend according to the environment Kubernetes be deployed. 

So we want to create a standalone project in Kubernetes that acts a library providing volume discovery and local management, aiming to break the limitation between storage systems and Kubernetes deployment environment, and to provide any storage backend regardless of deployment environment(cloud platforms and bare metals).

### DIFF
--- a/contributors/design-proposals/hostside-volume-discovery-and-management-library.md
+++ b/contributors/design-proposals/hostside-volume-discovery-and-management-library.md
@@ -49,7 +49,7 @@ and even any other storage systems written in Go.
 As we know, there are a lot of storage protocols, such as iscsi, rbd, fc,
 smbfs and so forth, and some of them are implemented in different ways
 according to different system types(x86, s390, ppc64) and os types
-(linux, windows), so it is a quite complicated work if we add these device
+(linux, windows), so it is quite a complicated work if we add these device
 drivers directly into volume plugins. But what we can do is to create a
 library to communicate with kernel and expose a unified interface to
 volume plugins.
@@ -98,7 +98,7 @@ const (
 // Connector is an interface indicating what outside world can do with this
 // library, notice that it is at very early stage right now.
 type Connector interface {
-	GetConnectorProperties(multiPath string, doLocalAttach bool) (*ConnectorProperties, error)
+	GetConnectorProperties(multiPath, doLocalAttach bool) (*ConnectorProperties, error)
 	
 	ConnectVolume(conn *ConnectionInfo) (string, error)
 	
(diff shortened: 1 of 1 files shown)


leonwanghui:
Got my account verified now.

cmluciano:
@kubernetes/sig-node-feature-requests @kubernetes/sig-storage-feature-requests

leonwanghui:
@gnufied @saad-ali @jingxu97 @cantbewong

cmluciano:
Adding https://github.com/kubernetes/community/pull/695 as related to some degree. This proposal is about mostly about pluggable hardware devices, but some of the ideas of discovery are similar.

leonwanghui:
@cmluciano Hi, very thank you for your precious suggestion.

I had a look at this proposal and found that there are some differences between these two proposals:
1. Device plugin can and should be directly charged by kubelet, but volume discovery library is called and monitored by volume plugins (such as EBS, Cinder, iSCSI).
2. As for GPU, FPGA these stuffs, most of these hardware devices should be install locally on kubelet nodes, but storage devices can be attached remotely, which means you can deploy a storage array on a standalone node.

Plz speak freely if you have any thought.

jsafrane:
@leonwanghui, there is a Container Storage Interface (CSI) being standardized right now at https://docs.google.com/document/d/1JMNVNP-ZHz8cGlnqckOnpJmHF-DNY7IYP-Di7iuVhQI/ and https://github.com/container-storage-interface/spec and which should address pluggable storage for Kubernetes, Mesos and Docker. Perhaps you should get involved there to make sure your use cases are included there.

leonwanghui:
@jsafrane Hi, thank you for your suggestion.

Here are my thoughts of the connection between this proposal and CSI:
1. It has no doubt that CSI plugin will be the major caller of this library, otherwise we won't expect it as a  library just like [utils](https://github.com/kubernetes/utils).
2. We find that some in-tree volume drivers don't support bare-metal scenario, and under this condition we want this proposal to cover the shortage and help these volume drivers provide bare-metal storage service as an optional solution.

Just my 2 cents : )

contributors/design-proposals/hostside-volume-discovery-and-management-library.md
```diff
@@ -0,0 +1,156 @@
...
+attaching volume to host is quite complicated. Can we shield the complex
+implementation and provide a simple interface for their secondary development?
+
+All in all, our motivation is to break the limitation between storage systems
+and Kubernetes deployment environment, and to provide any storage backend
+regardless of deployment environment(cloud platforms and bare metals).
+
+## Goal
+
+To slove the problem, we plan to create a standalone project in Kubernetes
```
vikaschoudhary16: s/slove/solve

PR #716, opened 2017-06-13
Update PodPreset modification annotation format

jpeeler:
The new format simply adds a prefix of "podpreset-" to the current
annotation that records what presets have acted on a pod. The new naming
makes it such that there is no chance of collision with the newly
introduced opt-out annotation (or future ones yet to be added).

Opt-out annotation PR:
kubernetes/kubernetes#44965


derekwaynecarr:
LGTM

derekwaynecarr:
@jpeeler -- need a change that describes the annotation you can stick on the pod to opt-out.

PR #721, opened 2017-06-14
Basic proposal for restricted nodes, including dedicated nodes

Q-Lee:
This is a proposal to implement restricted nodes with Taints, Tolerations, and PodSecurityPolicies. A restricted node with a single workload is a dedicated node.

Specifically, PodSecurityPolicies will allow certain Tolerations. This will permit placement on restricted nodes.

Unfortunately, this is insufficient, as one purpose for dedicated nodes is to permit privileged workloads. Granting a PodSecurityPolicy with privileged execution compromises the whole cluster, because the privileged workload could land anywhere.

The Taint/Toleration relationship must be reversed, so that nodes can tolerate pod taints. The recommended implementation is a new type of symmetric Taint/Toleration. The PodSecurityPolicy would then need a list of *required* Tolerations.

### DIFF
--- a/contributors/design-proposals/restricted-nodes.md
+++ b/contributors/design-proposals/restricted-nodes.md
@@ -1,0 +1,85 @@
+Kubernetes should have a mechanism to restrict pod exection on nodes. The
+difficulty is two-fold. First, nodes aren't namespaced, so it's difficult to
+give selective access to only a subset of nodes. Second, placement of a pod on a
+node is equivalent to write access on the pod/node resource.
+
+#Use cases for sole-tenancy
+
+There are many use cases for dedicated nodes, so a non-comprehensive list is
+provided.
+
+1. Many applications require elevated privileges, which could compromise other
+applications on the same node.
+1. Many applications cannot trust a shared platform.
+1. Many applications are designed to be 1:1 with a node (e.g., storage servers).
+1. Many applications may want access to special hardware.
+1. Many applications are critical, and cannot tolerate interference from
+   neighbors.
+1. Many applications may need isolation for compliance.
+
+#Requirements for restricted nodes
+
+1. Nodes must be capable of rejecting non-conforming workloads.
+1. Nodes must allow conforming workloads.
+
+#Requirements for dedicated workloads
+
+1. Placement must be guaranteeable.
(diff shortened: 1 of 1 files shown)


Q-Lee:
@kubernetes/sig-auth-misc

liggitt:
Limiting to a specific set of nodes is what node selectors are for... whitelisting tolerations allowed on the pod *and* requiring certain nodeSelector components on the pod both make sense

Q-Lee:
@liggitt that would also work. I just wanted to try and limit the number of different features required to establish dedicated nodes.

bgrant0607:
Related:
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/taint-toleration-dedicated.md
https://github.com/kubernetes/kubernetes/issues/17190
https://github.com/kubernetes/kubernetes/issues/17151
https://github.com/kubernetes/kubernetes/issues/14573

cc @davidopp

bgrant0607:
@Q-Lee 

You want to divide nodes into nodes that can execute privileged containers and those that can't? Apply taints to all nodes: privileged=true or privileged=false, and then use an initializer to set tolerations based on whether the pods contain privileged containers or not.

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/admission_control_extension.md

I can't foresee adding more features in this area at the current time. The focus is shifting to finishing existing APIs and project health:
https://groups.google.com/forum/#!topic/kubernetes-dev/PpgLgkffr3o

davidopp:
Unfortunately I haven't had time to read this proposal yet.

We *do* need a way to limit who can add dedicated node group tolerations to pods, otherwise dedicated node groups are impossible to do securely, which makes them useless.

RBAC doesn't have that granularity, so it's not applicable.

A custom admission controller is the most obvious solution that comes to mind. For the feature to be usable, we need an admission controller to apply the tolerations anyway, and it could also check to make sure that there are no dedicated node tolerations already set on the pods it is receiving (or could enforce any other policy about who can set such tolerations). There's no standard taint name for dedicated node groups so something custom (or at least configurable) seems necessary anyway.

I don't want to discount @Q-Lee's proposal since I haven't read it yet, but I did want to at least write down an alternative to using PSP.

PR #723, opened 2017-06-14
[Proposal] Allow a Pod Security Policy to managing access to the Flexvolumes

php-coder:
This PR proposes to add the `AllowedFlexVolumes` to a PSP to control pod's access to the different Flexvolume drivers.

PTAL @smarterclayton @pweil- @mfojtik

### DIFF
--- a/contributors/design-proposals/flex-volumes-drivers-psp.md
+++ b/contributors/design-proposals/flex-volumes-drivers-psp.md
@@ -1,0 +1,94 @@
+# Allow Pod Security Policy to manage access to the Flexvolumes
+
+## Current state
+
+Cluster admins can control the usage of specific volume types by using Pod
+Security Policy (PSP). Admins can allow the use of Flexvolumes by listing the
+`flexVolume` type in the `volumes` field. The only thing that can be managed is
+allowance or disallowance of Flexvolumes.
+
+Technically, Flexvolumes are implemented as vendor drivers. They are executable
+files that must be placed on every node at
+`/usr/libexec/kubernetes/kubelet-plugins/volume/exec/<vendor~driver>/<driver>`.
+In most cases they are scripts. Limiting driver access means not only limiting
+an access to the volumes that this driver can provide, but also managing access
+to executing a driver’s code (that is arbitrary, in fact).
+
+It is possible to have many flex drivers for the different storage types. In
+essence, Flexvolumes represent not a single volume type, but the different
+types that allow usage of various vendor volumes.
+
+## Desired state
+
+In order to further improve security and to provide more granular control for
+the usage of the different Flexvolumes, we need to enhance PSP. When such a
+change takes place, cluster admins will be able to grant access to any
+Flexvolumes of a particular driver (in contrast to any volume of all drivers).
+
+For example, if we have two drivers for Flexvolumes (`cifs` and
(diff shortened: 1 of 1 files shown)


liggitt:
@Q-Lee @kubernetes/sig-auth-proposals

adelton:
Do we really want to go with separate `AllowedFlexDrivers`? Would it be easier for users to list the allowed vendor/drivers as part of the volumes list, something like `flexVolume/redhat.com/kvm`, `flexVolume/example.com/keytab`?

In a similar manner, there could then be `hostPath/var/lib/datax` and similar ways to restrict scope of other volume types.

pweil-:
> Do we really want to go with separate AllowedFlexDrivers? Would it be easier for users to list the allowed vendor/drivers as part of the volumes list, something like flexVolume/redhat.com/kvm, flexVolume/example.com/keytab?

We currently validate that we're covering volume sources to prevent drift between allowed and available volumes so that would have to be changed if we decided to do this (just need to keep that in mind)

liggitt:
each volume source type is likely to have different options. for hostpath, it is the local filesystem path. for flexvolumes, it's driver (and possibly option-level control in the future). I'd rather keep different things different.

adelton:
In general, the proposal LGTM.

php-coder:
`AllowedFlexDrivers` was renamed to `AllowedFlexvolumes` and type was changed from `[]string` to `[]AllowedFlexvolume`. This lets us expand options in the future (suggested by @liggitt).

contributors/design-proposals/flex-volumes-drivers-psp.md
```diff
@@ -0,0 +1,97 @@
...
+
+```go
+// PodSecurityPolicySpec defines the policy enforced.
+type PodSecurityPolicySpec struct {
+    ...
+    // AllowedFlexvolumes is a whitelist of allowed Flexvolumes.  Empty or nil indicates that all
+    // Flexvolumes may be used.  This parameter is effective only when the usage of the Flexvolumes
+    // is allowed in the "Volumes" field.
+    // +optional
+    AllowedFlexvolumes []AllowedFlexvolume
```
liggitt: nit, AllowedFlexVolumes (match the case in the field)

php-coder: Thanks, fixed.

liggitt:
cc @kubernetes/sig-storage-api-reviews @kubernetes/sig-auth-api-reviews

contributors/design-proposals/flex-volumes-drivers-psp.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Admission controller modifications
+
+Admission controller should be updated accordingly to inspect a Pod's volumes.
+If it finds a `flexVolume`, it should ensure that its driver is allowed to be
+used.
+
+### Validation rules
+
+No validation is expected for Flexvolume driver names. API server should allow
```
liggitt: nit: driver name should be non-empty

liggitt:
this LGTM and parallels fine-grained control over hostpaths

php-coder:
Commits were squashed without changes.

PTAL.

childsb:
This LGTM

saad-ali:
Seems reasonable to me.

PR #727, opened 2017-06-15
Add retainKeys and setElementOrder to SMP doc

mengqiy:
Update doc

### DIFF
--- a/contributors/devel/strategic-merge-patch.md
+++ b/contributors/devel/strategic-merge-patch.md
@@ -210,6 +210,127 @@ Note: When merging two set, the primitives are first deduplicated and then merge
 In an erroneous case, the set may be created with duplicates. Deleting an
 item that has duplicates will delete all matching items.
 
+## `setElementOrder` Directive
+
+### Purpose
+
+`setElementOrder` directive provides a way to specify the order of a list.
+The relative order specified in this directive will be retained.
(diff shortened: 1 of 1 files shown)


contributors/devel/strategic-merge-patch.md
```diff
@@ -210,6 +210,127 @@ Note: When merging two set, the primitives are first deduplicated and then merge
...
+    ...
+  - name: b
+    ...
+  - name: c
+    ...
+```
+
+To reorder the elements order in the list, we can send a patch:
+```yaml
+# each map in the list should only include the mergeKey
```
alexandercampbell: >each map in the list should only include the mergeKey

Will the program abort if there are more fields than just the `mergeKey`?

contributors/devel/strategic-merge-patch.md
```diff
@@ -210,6 +210,127 @@ Note: When merging two set, the primitives are first deduplicated and then merge
...
+  bar: c
+```
+
+After applying this patch, we get:
+```
+union:
+  # Field foo and other have been cleared w/o explicitly set them to null.
+  another: d
+  bar: c
+```
```
alexandercampbell: Could you add an example where one field from the original union remains?

alexandercampbell:
haha, nevermind :)

PR #738, opened 2017-06-20
Fix indentation for nested numerated list

loburm:
In case of using two spaces github doesn't treat nested list as a separate numerated list.


bgrant0607:
Thanks

PR #741, opened 2017-06-20
Metrics server design proposal

piosz:
This was reviewed and approved in https://docs.google.com/document/d/1w6-ZfnA18aKYLJ8DCLBFKlv_1umm74x0I2X4V188kvU/edit#

### DIFF
--- a/contributors/design-proposals/metrics-server.md
+++ b/contributors/design-proposals/metrics-server.md
@@ -1,0 +1,89 @@
+Metrics Server
+==============
+
+Resource Metrics API is an effort to provide a first-class Kubernetes API
+(stable, versioned, discoverable, available through apiserver and with client support)
+that serves resource usage metrics for pods and nodes. The use cases were discussed
+and the API was proposed a while ago in
+[another proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-metrics-api.md).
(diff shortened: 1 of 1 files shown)


piosz:
As I wrote this was discussed, reviewed and approved in the mentioned doc, so I'm merging it right now, but feel free to comment. Since this is an alpha feature I'm open to revisit the design in a follow up PR.

pigmej:
@piosz FYI this document has restricted access.

zen:
I just wanted to note here that there is a very good practice not to merge own PRs. Also this merge is very, very fast, I don't think anyone was able to even read the document, not to mention any discussion, since document is restricted.

piosz:
@pigmej this is shared with `kubernetes-sig-instrumentation@` and `kubernetes-sig-api-machinery@`. I'm also happy to share it upon request. Unfortunatelly there is not an option to make it public. I'll share this also with `kubernetes-dev@`.

@zen apologies for that. I find it much, much, much easier to review the proposal in a doc rather than on github and as I mentioned we went through the review process with approvers in the doc, but you are totally right and I've just realized that I haven't reached everyone who might be interested.

Anyway all feedback is welcome. Feel free to comment and I'm happy to apply them in a follow up PR.

PR #748, opened 2017-06-21
add notes from presentations - leadership summit

nyener:


bgrant0607:
Thanks @nyener !

Merging.

PR #758, opened 2017-06-22
Remove timestamp from sig file generation

cblecker:
Timestamps in the sig docs generation create more potential for merge conflicts, as well as touch/modify files that would otherwise not need to be modified when running `make all`.

cc: @jamiehannaford


jamiehannaford:
👍

cblecker:
Self LGTM based on @jamiehannaford's above comment.

PR #761, opened 2017-06-23
Reword build cop duties for batch merges.

spxtr:
There was some confusion about this in slack, so I've changed the wording to make it clear that the build cop only needs to make sure that they're running, not that they're flake-free.

### DIFF
--- a/contributors/devel/on-call-build-cop.md
+++ b/contributors/devel/on-call-build-cop.md
@@ -31,7 +31,8 @@ longer a part of the mandate.
 ## Workflow
 
 1. Check the Prow batch dashboard: [https://prow.k8s.io/?type=batch](https://prow.k8s.io/?type=batch) 
-to ensure that merges are occurring regularly.
+to ensure that batch jobs are running regularly. It's okay to see occasional flakes. Do not worry
+about manually re-running individual tests, since Prow will rerun them.
 2. If there are post-submit blocking jobs (see [link](https://submit-queue.k8s.io/#/e2e)), ensure 
 that those builds are green and allowing merges to occur.
 3. If several batch merges are failing, file an issue for that job and describe the possible 


contributors/devel/on-call-build-cop.md
```diff
@@ -31,7 +31,7 @@ longer a part of the mandate.
 ## Workflow
 
 1. Check the Prow batch dashboard: [https://prow.k8s.io/?type=batch](https://prow.k8s.io/?type=batch) 
-to ensure that merges are occurring regularly.
+to ensure that batch jobs are running regularly. It's okay to see occasional flakes.
```
mml: I would keep the original text.  The buildcop actually cares about the merge rate, not that the batch jobs are running (seems like that would be a test-infra-oncall problem, no?).

contributors/devel/on-call-build-cop.md
```diff
@@ -31,7 +31,7 @@ longer a part of the mandate.
...
 2. If there are post-submit blocking jobs (see [link](https://submit-queue.k8s.io/#/e2e)), ensure 
 that those builds are green and allowing merges to occur.
 3. If several batch merges are failing, file an issue for that job and describe the possible
```
mml: Maybe move the guidance here?  In the next line, `s/for the failure. /for the failure. Do not worry about manually re-running individual tests.  Prow manages that itself. `

contributors/devel/on-call-build-cop.md
spxtr: Sure, but checking the prow dashboard doesn't tell you that merges are happening.

contributors/devel/on-call-build-cop.md
spxtr: Done.

PR #336, opened 2017-02-07
lister-gen documentation

MHBauer:
documentation based on reverse engineering of usage.

PR #769, opened 2017-06-28
Create code-generation subsection

MHBauer:
- move protobufs and codecgen into new subsection
 - subsection details on code generation in the build
 - details on lister-gen code generator

per #336 with some additional info and organization.

/cc @pmorie @pwittrock @caesarxuchao


pwittrock:
@MHBauer Thanks this is really useful.  WDYT of adding instructions on how to use each of the generators?  The required flags are surprisingly not-obvious.

contributors/devel/api_changes.md
```diff
@@ -462,8 +462,26 @@ generator to create it from scratch.
...
+run. They have some dependencies on each other so they must be run in
+a certain order.
+
+Approximately:
+ - `defaulter-gen`
+ - `deepcopy-gen`
+ - `conversion-gen`
+ - `client-gen`
+ - `lister-gen` which depends upon the output of `client-gen`.
+ - `informer-gen` which depends upon the output of `lister-gen` and `client-gen`.
```
pwittrock: openapi-gen?

contributors/devel/api_changes.md
```diff
@@ -462,8 +462,26 @@ generator to create it from scratch.
...
+ - `codecgen` for fast json serialization with a codec.
+
+Many of the generators are based on
```
pwittrock: I don't think the flags are treated that same for each of the generators in all cases.  :(

contributors/devel/api_changes.md
MHBauer: Sure, this is why we have reviews.

contributors/devel/api_changes.md
MHBauer: Should I end up with an 'example' run for each? To show the flags?

pwittrock: I think that would be ideal

contributors/devel/api_changes.md
```diff
@@ -479,7 +508,30 @@ use our own equivalents for JSON serialization. The `pkg/api/serialization_test.
...
+### Generate Clientset
+
+See [this document](generating-clientset.md) specific to generating client sets.
+
+### Generate Listers
+
+`lister-gen` is a tool to generate listers for a client. It requires a
+client generated by `client-gen`.
+
+`lister-gen` requires the `// +genclient=true` annotation on each
```
pwittrock: I *think* this has to be a line above the struct with a space in-between.  Can you confirm?  If so, this seems a rather important detail.  e.g. 

```go
// +genclient=true

type MyType struct {
```

instead of 

```go
// +genclient=true
type MyType struct {
```

pwittrock:
/assign @pwittrock

This lgtm to me.  I suggest we merge and iterate.  @ncdc @derekwaynecarr Is this ok with you?

contributors/devel/api_changes.md
pmorie: I am pretty sure @pwittrock is right about this.

pmorie: Global resources also need `// +nonNamespaced=true` as a directive.

Also, these annotations are required for client-gen as well, and the linked document doesn't describe them.  I'm starting to wonder whether the generating-clientset.md file should just be generating-code.md and have doc for all the things there.

contributors/devel/api_changes.md
```diff
@@ -479,7 +508,30 @@ use our own equivalents for JSON serialization. The `pkg/api/serialization_test.
...
+exported type in both the unversioned base `types.go` as well as each
+specifically versioned `types.go`.
+
+`lister-gen` requires the `// +groupName=` annotation on the `doc.go` in
```
pmorie: There are a couple more annotations:

```go
// +k8s:defaulter-gen=TypeMeta
// +k8s:deepcopy-gen=package,register
```

That are not doc'd anywhere that I'm aware of - will you doc those as well?

pmorie:
I'm totally fine with merge and iterate.

MHBauer:
I can do follow ups. Or someone else. I want to try and write this down before I forget.

PR #321, opened 2017-02-01
Add proposal for mount options

gnufied:
Adding initial proposal for mount options. 

CC @saad-ali @matchstick @rootfs

PR #771, opened 2017-06-29
Proposal for taking mount option to GA

gnufied:
Also add support for specifying mount options
in storage classes.

cc @kubernetes/sig-storage-api-reviews @liggitt

### DIFF
--- a/contributors/design-proposals/mount-options.md
+++ b/contributors/design-proposals/mount-options.md
@@ -76,8 +76,9 @@ The mount option specified in Storage Class will be used while provisioning pers
 and added as a field to PVs.
 
 If admin has configured mount option for a storage type that does not support mount options,
-then `mountOptions` parameter will be simply ignored for such volume types.  Also, if configured
-mount option is invalid then corresponding mount time failure error will be added to pod object.
+then a "provisioning failed" event will be added to PVC and PVC will stay in pending state.
+
+Also, if configured mount option is invalid then corresponding mount time failure error will be added to pod object.
 
 
 ## Preventing users from specifying mount options in inline volume specs of Pod
--- a/contributors/design-proposals/mount-options.md
+++ b/contributors/design-proposals/mount-options.md
@@ -13,9 +13,31 @@ Mount time options that are operationally important and have no security implica
 
 ## Design
 
-We are going to add support for mount options in PVs as a beta feature to begin with.
+### Mount option support in Persistent Volume Objects
 
-Mount options can be specified as `mount-options` annotations in PV. For example:
+Mount options can be specified as a field on PVs. For example:
+
+``` yaml
+ apiVersion: v1
+  kind: PersistentVolume
+  metadata:
+    name: pv0003
+  spec:
+    capacity:
+      storage: 5Gi
+    accessModes:
+      - ReadWriteOnce
+    persistentVolumeReclaimPolicy: Recycle
+    mountOptions: "hard,nolock,nfsvers=3"
+    nfs:
+      path: /tmp
+      server: 172.17.0.2
+```
+
+
+Beta support for mount options introduced via `mount-options` annotation will be supported for near future
+and deprecated in future.
+
 
 ``` yaml
  apiVersion: v1
@@ -35,6 +57,28 @@ Mount options can be specified as `mount-options` annotations in PV. For example
       server: 172.17.0.2
 ```
 
+### Mount option support in Storage Classes
+
+Kubernetes admin can also specify mount option as a parameter in storage class.
+
+```yaml
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: slow
+provisioner: kubernetes.io/glusterfs
+parameters:
+  type: gp2
+  mountOptions: "auto_mount"
+```
+
+The mount option specified in Storage Class will be used while provisioning persistent volumes
+and added as a field to PVs.
+
+If admin has configured mount option for a storage type that does not support mount options,
+then `mountOptions` parameter will be simply ignored for such volume types.  Also, if configured
+mount option is invalid then corresponding mount time failure error will be added to pod object.
+
 
 ## Preventing users from specifying mount options in inline volume specs of Pod
 
@@ -42,7 +86,7 @@ While mount options enable more flexibility in how volumes are mounted, it can r
 in user specifying options that are not supported or are known to be problematic when
 using inline volume specs.
 
-After much deliberation it was decided that - `mount-options` as an API parameter will not be supported
+After much deliberation it was decided that - `mountOptions` as an API parameter will not be supported
 for inline volume specs.
 
 ### Error handling and plugins that don't support mount option


contributors/design-proposals/mount-options.md
```diff
@@ -35,14 +57,36 @@ Mount options can be specified as `mount-options` annotations in PV. For example
...
+parameters:
+  type: gp2
+  mountOptions: "auto_mount"
+```
+
+The mount option specified in Storage Class will be used while provisioning persistent volumes
+and added as a field to PVs.
+
+If admin has configured mount option for a storage type that does not support mount options,
+then `mountOptions` parameter will be simply ignored for such volume types.  Also, if configured
```
wongma7: we have always returned a "provisioning failed" error when the storageclass has an invalid/unrecognized param

contributors/design-proposals/mount-options.md
```diff
@@ -35,14 +57,36 @@ Mount options can be specified as `mount-options` annotations in PV. For example
...
+
 
 ## Preventing users from specifying mount options in inline volume specs of Pod
 
 While mount options enable more flexibility in how volumes are mounted, it can result
 in user specifying options that are not supported or are known to be problematic when
 using inline volume specs.
 
-After much deliberation it was decided that - `mount-options` as an API parameter will not be supported
+After much deliberation it was decided that - `mountOptions` as an API parameter will not be supported
```
liggitt: will the volume source API structs used within a PV be cloned to add this new field?

gnufied: No, the `mountOptions` is not a field of volume source APIs. It is a field of PV itself as proposed above

contributors/design-proposals/mount-options.md
gnufied: That seems correct, updating the proposal to reflect this.

gnufied:
@liggitt @saad-ali ping on this one again.

nicorevin:
Imagine I want to mount /tmp as an EmptyDir of medium: Memory.
The recommendation is to mount /tmp with `nosuid,nodev,noexec` options. And this is what I see when running `docker run --tmpfs /tmp`:
```
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,noexec,relatime)
```
But in kubernetes I observe:
```
tmpfs on /tmp type tmpfs (rw,relatime)
```
The proposal mentions PersistentVolumes and StorageClasses but how could one pass proper options to the EmptyDir?

nicorevin:
err. it looks impossible atm https://github.com/kubernetes/kubernetes/blob/74f19437742abcff81510dc347f2cc70ec190e99/pkg/volume/empty_dir/empty_dir.go#L258
And it turns out that the emptydir case is beyond the proposal...

contributors/design-proposals/mount-options.md
```diff
@@ -35,14 +57,37 @@ Mount options can be specified as `mount-options` annotations in PV. For example
...
+
+```yaml
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: slow
+provisioner: kubernetes.io/glusterfs
+parameters:
+  type: gp2
+  mountOptions: "auto_mount"
```
gnufied: To continue on discussion from sig meeting today. @saad-ali I think having this in parameters frees us from having to perform validation of storageclass itself and such decision can be deferred to individual provisioner.

contributors/design-proposals/mount-options.md
```diff
@@ -13,9 +13,31 @@ Mount time options that are operationally important and have no security implica
...
+      - ReadWriteOnce
+    persistentVolumeReclaimPolicy: Recycle
+    mountOptions: "hard,nolock,nfsvers=3"
+    nfs:
+      path: /tmp
+      server: 172.17.0.2
+```
+
+
+Beta support for mount options introduced via `mount-options` annotation will be supported for near future
```
saad-ali: Beta -> Alpha?

contributors/design-proposals/mount-options.md
saad-ali: I think `mountOptions` should be a first class field in the `StorageClass` because it is controlling a first class field in the PV object (not specific to a single volume plugin). As far as validation is concerned, for in-tree plugins you can call `SupportsMountOption(...)` on the provisioner. The tricky bit will be how to pass this through for out-of-tree volume provisioners, but that problem needs to be solved for reclaimPolicy (https://github.com/kubernetes/kubernetes/pull/47987) as well.

contributors/design-proposals/mount-options.md
```diff
@@ -35,14 +57,37 @@ Mount options can be specified as `mount-options` annotations in PV. For example
...
 
 ## Preventing users from specifying mount options in inline volume specs of Pod
 
 While mount options enable more flexibility in how volumes are mounted, it can result
 in user specifying options that are not supported or are known to be problematic when
 using inline volume specs.
 
-After much deliberation it was decided that - `mount-options` as an API parameter will not be supported
+After much deliberation it was decided that - `mountOptions` as an API parameter will not be supported
 for inline volume specs.
```
rata: So, if I understand correctly, this won't be possible to specify on pod.spec.volumes, right? I mean, as in a pod, that you can specify a volume and fsType, like in the example here: https://kubernetes.io/docs/concepts/storage/volumes/#awselasticblockstore

Can you please elaborate on the much deliberation?

Don't get me wrong, I'm not saying I'm against it, I just want to understand a little more. It seems weird that you can have the fsType there but not `mountOptions`. But I understand that pvc are preferred now, also.

So, if you have some time, can I ask what was taken into account on this deliberation, the trade offs of different approaches considered, etc.? Something high level is fine. But I understand it's a busy time on the release, so no worries if you can't :)

contributors/design-proposals/mount-options.md
```diff
@@ -35,14 +57,37 @@ Mount options can be specified as `mount-options` annotations in PV. For example
...
+then a "provisioning failed" event will be added to PVC and PVC will stay in pending state.
+
+Also, if configured mount option is invalid then corresponding mount time failure error will be added to pod object.
+
 
 ## Preventing users from specifying mount options in inline volume specs of Pod
 
 While mount options enable more flexibility in how volumes are mounted, it can result
 in user specifying options that are not supported or are known to be problematic when
 using inline volume specs.
```
rata: `When using inline volumes`? What is more problematic about an in-line volume? Coordinate the `fsType` and the mount options? If that guess is on track, then it's more problematic for file-system dependent mount options (there are file-system independent options too) and when not specifyng in-line the `fsType`?

contributors/design-proposals/mount-options.md
gnufied: No it was beta - https://kubernetes.io/docs/concepts/storage/persistent-volumes/#mount-options

contributors/design-proposals/mount-options.md
gnufied: I think bulk of discussion is here - https://github.com/kubernetes/community/pull/321#pullrequestreview-20780606

The tldr; of the discussion is - allowing normal users to specify mountOptions can enable users to compromise the host operation system itself. Users can disable selinux or can cause kernel panic which would crash the host operating system. That is why mount option should be carefully controlled.

contributors/design-proposals/mount-options.md
gnufied: responded above.

contributors/design-proposals/mount-options.md
rata: Thanks a lot for the link! I guess I'm with saad-ali, but all makes sense and it's nice to see the background. Thanks :)

contributors/design-proposals/mount-options.md
rata: Thanks again! :)

contributors/design-proposals/mount-options.md
wongma7: agree it should be 1st class for that reason.

I think the reason gnufied is hesitant to do create-time validation is we actually can't call SupportsMountOption due to, iirc, the cost of importing all the volume plugin code into the validation code. But that was resolved, even if it was a bit ugly..

It's true if we do create-time validation of storageClass.mountOptions for in-tree plugins there will be some asymmetry between in- & out-. But I don't think solving that is worth it. While admins deploying out-of-tree provisioners may appreciate the upfront warning that they've configured their storageClass with an unsupported mountOptions field, realistically they care more about mistakes in the parameters field which won't show up until run-time anyway... So I think they'll settle for run-time validation of storageClass.mountOptions.

Plus, this problem shouldn't exist for reclaimPolicy because every plugin already supports `Retain,` they can't opt-out of that like they can mount options.

gnufied: I am fine making it first class field on StorageClass. The minor catch is - we can't really do creation time validation of storageClass for verifying if it supports mount options without hardcoding name of volume plugins that will support it (like literally hardcoding plugin name strings).

But I think that is relatively minor issue, because in my quick survey most volume types that support dynamic provisioning also support mount options (i.e at plugin level in current master) and hence there is no practical benefit of validation the SC for mount options during creation. So, what we can do is:

1. Make mountOption first class field for StorageClass.
2. Don't do any validation of mount option when storageClass is created.
3. Perform provisioning time `SupportsMountOption` validation and check if mount option is supported. This is really just a safety net. As I said above, all volume types that support dynamic provisioning also support mount options - so for most part this check will always be successful. Also, even if we didn't check during provisioning the PV creation will be anyways rejected if volume type doesn't support mount options.

saad-ali: > 1. Make mountOption first class field for StorageClass.
> 2. Don't do any validation of mount option when storageClass is created.
> 3. Perform provisioning time SupportsMountOption validation and check if mount option is supported. This is really just a safety net. As I said above, all volume types that support dynamic provisioning also support mount options - so for most part this check will always be successful. Also, even if we didn't check during provisioning the PV creation will be anyways rejected if volume type doesn't support mount options.


SGTM!

PR #772, opened 2017-07-01
Proposal for how labels should be applied to PRs to populate release …

pwittrock:
…notes

### DIFF
--- a/contributors/devel/release/releasenotes.md
+++ b/contributors/devel/release/releasenotes.md
@@ -1,0 +1,94 @@
+# Creating PRs for release notes
+
+This document describes how to create PRs in Kubernetes repos so that the changes can be tracked with the
+Kubernetes release and published in release notes.
+
+## Curated release notes
+
+The release notes for each Kubernetes release are divided into several sections.
+
+All sections are grouped into subsections by component.
+
+#### Action required
+
+The action required section outlines any steps the user must perform before upgrading their cluster.
+The may include actions such as migrating APIs or commands to new versions, upgrading /
+installing tools, or validating the cluster environment.
+
+#### Deprecations
+
+The deprecation section outlines any APIs, tools, commands or flags that are being deprecated.
+They will continue to be available for this release, but will be removed in a subsequent release.
+
+#### Feature highlights
+
+The feature highlights section is manually populated from the [features repo] and contains
+a list of the feature highlights for the release. 
+
+#### Notable changes
+
+The notable changes section is generated from the PR release notes, and contains a list
+of features and bug fixes going into the release.
+
+## Release note labels
+
+All  PRs with user visible changes must have a release note attached.  This will automatically
+
+populate them in the release notes.
+
+Release notes are added applying on of the `release-note*` labels by adding a section
+to the PR description containing the test that will appear in the release notes:
+
+>\```release-note*
+>
+> User visible change description
+>
+> \```
+
+The change description will be directly added to the release notes, and should include the visible
+impact to the end user.  Individual SIGs should consider developing guidelines around how to write
+release notes for the components they own.  The release-note labels follow:
+
+- `release-note-action-required`
+
+The `release-note-action-required` label will cause the PR to be added to the Action Required section
+of the release notes.  This label should be added for any PRs that result in the user
+needing to take action before, during or after upgrading their cluster to the new release.  The release
+note should outline what action the user must perform.  e.g.
+
+> kubectl --foo has been removed and replaced with --bar.  All references to --foo must be updated.
+
+- `release-note-deprecation`
+
+The `release-note-deprecation` label will cause the PR to be added to the Deprecation section
+of the release notes.  This label should be added for any PRs that mark APIs, tools, commands or
+flags as deprecated.
+
+- `release-note`
+
+The `release-note` label will cause the PR to be added to the Notable changes section.
+
+- `release-note-none`
+
+The `release-note-none` should be applied to any changes that are not visible to the end user. e.g.
+refactorings, test changes, and disabled code.
+
+## Additional required labels
+
+- `sig/*` e.g. `sig/cli`
+
+The `sig` label will be used by SIGs to process and edit release notes before the release is published.
+Release notes often must be ordered and reworded so they are consistent.  The SIG labels will not
+appear in the final release notes.
+
+- `component/*`
+
+The component label will be used to further group the release notes by component.  This is useful for
+SIGs with multiple components - e.g. SIG node with kubelet, kubeproxy, cri
+
+- `kind/bug`, `kind/feature`
+
+The `kind` label will be used to sort the release notes as either bug fixes or new features.
+
+[kubernetes repo]: (https://github.com/kubernetes/kubernetes)
+[features repo]: (https://github.com/kubernetes/features)


pwittrock:
cc @kubernetes/sig-contributor-experience-proposals 

@marun @calebamiles @dchen1107 @spiffxp

grodrigues3:
Should we delete this section then: https://github.com/kubernetes/community/blob/master/contributors/devel/pull-requests.md#write-release-notes-if-needed

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
+# Creating PRs for release notes
+
+This document describes how to create PRs in Kubernetes repos so that the changes can be tracked with the
+Kubernetes release and published in release notes.
+
```
grodrigues3: Any thoughts on adding/linking to a few examples?  

1. An example of a user-visible change (something that might want a rn) and what a good rn looks like
1. An example a bugfix that the user likely won't care about and probably doesn't need a release note.

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+All sections are grouped into subsections by component.
+
+#### Action required
+
+The action required section outlines any steps the user must perform before upgrading their cluster.
+The may include actions such as migrating APIs or commands to new versions, upgrading /
+installing tools, or validating the cluster environment.
+
+#### Deprecations
```
luxas: What about removals? I think they deserve their own section as well

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Feature highlights
+
+The feature highlights section is manually populated from the [features repo] and contains
+a list of the feature highlights for the release. 
+
+#### Notable changes
+
+The notable changes section is generated from the PR release notes, and contains a list
+of features and bug fixes going into the release.
```
luxas: of _notable_ features and bug fixes...

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Release note labels
+
+All  PRs with user visible changes must have a release note attached.  This will automatically
+
```
luxas: dead line

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+The `release-note` label will cause the PR to be added to the Notable changes section.
+
+- `release-note-none`
+
+The `release-note-none` should be applied to any changes that are not visible to the end user. e.g.
+refactorings, test changes, and disabled code.
+
+## Additional required labels
+
+- `sig/*` e.g. `sig/cli`
```
luxas: Is the new policy to enforce sig labels on all `!release-note-none` PRs?

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
+# Creating PRs for release notes
+
+This document describes how to create PRs in Kubernetes repos so that the changes can be tracked with the
+Kubernetes release and published in release notes.
```
luxas: For every unstable release as well (e.g. `v1.8.0-alpha.1`) or is this only for `.0` minor releases like `v1.8.0`?

luxas: FWIW, I think it would make sense to group things better in alpha/beta/rcs as well

contributors/devel/release/releasenotes.md
pwittrock: yes

contributors/devel/release/releasenotes.md
pwittrock: In the past I think those have gone in action required before upgrading.

contributors/devel/release/releasenotes.md
pwittrock: This is just for minor releases.  The release notes for all unstable releases are aggregated and reformatted for minor releases.

contributors/devel/release/releasenotes.md
luxas: SGTM to do that in the future as well, may be nice to point out as an example

contributors/devel/release/releasenotes.md
luxas: Please clarify that in the doc

contributors/devel/release/releasenotes.md
luxas: Ok, can you modify the title to be `Additional required labels when "release-note" is set`

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Curated release notes
+
+The release notes for each Kubernetes release are divided into several sections.
+
+All sections are grouped into subsections by component.
```
spiffxp: Can we define what a component is?

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+#### Action required
+
+The action required section outlines any steps the user must perform before upgrading their cluster.
+The may include actions such as migrating APIs or commands to new versions, upgrading /
+installing tools, or validating the cluster environment.
+
+#### Deprecations
+
+The deprecation section outlines any APIs, tools, commands or flags that are being deprecated.
+They will continue to be available for this release, but will be removed in a subsequent release.
```
spiffxp: As a cluster operator or app developer, it would be helpful if deprecation notices included:
- what's being deprecated, and what I should use instead
- the schedule for when it's being removed
- a link to an issue that I can follow to track the removal of the deprecated functionality

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+The `sig` label will be used by SIGs to process and edit release notes before the release is published.
+Release notes often must be ordered and reworded so they are consistent.  The SIG labels will not
+appear in the final release notes.
+
+- `component/*`
+
+The component label will be used to further group the release notes by component.  This is useful for
+SIGs with multiple components - e.g. SIG node with kubelet, kubeproxy, cri
+
+- `kind/bug`, `kind/feature`
```
spiffxp: We have many other `kind/foo` labels, would it be worth adding more of those to this list and/or paring down the kind labels that we have? eg: `kind/enhancement` might be useful to an end-user, `kind/technical-debt` less so, `kind/api-change` might signify something that's worth a deprecation notice, etc.

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+- `sig/*` e.g. `sig/cli`
+
+The `sig` label will be used by SIGs to process and edit release notes before the release is published.
+Release notes often must be ordered and reworded so they are consistent.  The SIG labels will not
+appear in the final release notes.
+
+- `component/*`
+
+The component label will be used to further group the release notes by component.  This is useful for
+SIGs with multiple components - e.g. SIG node with kubelet, kubeproxy, cri
```
spiffxp: I would prefer to see more clarity on what a component is... somewhere.  Maybe this isn't the proposal to hash it out, but beyond "some set of binaries for some definition of user-facing" I'm not sure what a component is

eg: there are no current component/foo labels in kubernetes/kubernetes, I think we have N components spread across M repos

One of [an explicitly defined list]? A binary that is contained within the kubernetes.tar.gz? A binary that is produced by one of the kubernetes org repos? A project that is tracked by the sig-release team? Well-known addons? etc

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
+# Creating PRs for release notes
+
+This document describes how to create PRs in Kubernetes repos so that the changes can be tracked with the
+Kubernetes release and published in release notes.
+
+## Curated release notes
+
+The release notes for each Kubernetes release are divided into several sections.
```
spiffxp: There has in the past been a Known Issues section, though it's been collated (by a human) via an accumulator issue.  Does it muddy the water to describe that mechanism here given that everything is PR and label based?

contributors/devel/release/releasenotes.md
```diff
@@ -0,0 +1,94 @@
...
+The notable changes section is generated from the PR release notes, and contains a list
+of features and bug fixes going into the release.
+
+## Release note labels
+
+All  PRs with user visible changes must have a release note attached.  This will automatically
+
+populate them in the release notes.
+
+Release notes are added applying on of the `release-note*` labels by adding a section
```
spiffxp: These labels should be applicable via community members who do not have labeling privileges on the repo in question, eg: via a [bot command](https://github.com/kubernetes/test-infra/blob/master/commands.md) in the comments

(I'd to explicitly avoid spelling out the syntax here lest it change in the future)

Reminder that new labels like `-deprecation` should be added to the commands if they're not presently supported

Reminder that the labels should be consistently applied to other repos in the kubernetes org beyond kubernetes/kubernetes for this to take effect

contributors/devel/release/releasenotes.md
pwittrock: Would you provide a definition and I will put it in?

spiffxp: @pwittrock the problem is I don't really know what a component is, I took various stabs in a further review comment, but I don't know if any of those hit the mark?

pwittrock: Or do you know if we have a definition for component somewhere?

PR #774, opened 2017-07-01
Proposal for publishing and communicating status of patch releases

pwittrock:


contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,78 @@
+# Building Kubernetes patch releases
+
+This document describes the process for creating Kubernetes patch releases and communicating publishing
+status regarding known issues impacting the current release.
+
+## TL;DR
+
+The patch release czar will summarize all issues discovered since the last patch release weekly
```
idvoretskyi: Can we avoid mentioning "czar" naming? It seems to be not used anymore for the non-patch releases since 1.6.

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,78 @@
...
+at the SIG release meeting and publish the summary to kubernetes-sig-release@googlegroups.com.
+
+## Process
+
+After a Kubernetes minor release, patch releases are managed by the Kubenretes patch release czar.
```
idvoretskyi: s/Kubenretes patch/Kubernetes patch/

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,78 @@
...
+The patch release czar works with SIGs to coordinate merging cherrypicks into the release branch,
+then cutting and publishing a new patch release.
+
+## Communicating critical fixes to the community
+
+In order to provide better transparency with the community on the patch release status and process,
+the patch release czar provides a weekly summary of all issues that have been discovered since the
+*last patch release* (for the most recent minor release only).
+
+The summary will be presented at the weekly sig-release meeting as the first agenda item (recorded
```
idvoretskyi: We have patch release slot at the weekly community meeting. Will this report be shared across the whole community?

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,78 @@
...
+in the release notes), and then mailed to kubernetes-sig-release@googlegroups.com.  During / after
+the summary at the SIG meeting, the patch release czar can answer questions from the community.
+
+The email update includes the state and a short summary for each item using the following template.
```
enisoc: This sounds like a lot of manual work. In the span between 1.6.0 and 1.7.0, I triaged 150+ cherrypick PRs.

I would rather see a specification for how we tag issues to encode all this info (if it isn't there already), and then a dashboard or script that automatically pulls it out.

contributors/devel/release/patch_release.md
enisoc: We should only have one required meeting per week for patch release managers, to keep the load down so people will actually volunteer. The sig-release meeting makes the most sense to me.

If we do detailed per-minor-release updates at the sig-release meeting, perhaps we can replace the community meeting slot with a high-level summary across all minor releases, given by a sig-release lead.

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,78 @@
...
+- Impact: foo can cause bar resulting in baz (Impact of the issue to the user or cluster)
+- PRs: #23456 (List of PR fixes)
+
+# 6789
+...
+```
+
+## Persistent record of issues discovered in each patch release
+
+The weekly summaries will be copied to the `release1.X/patch_history.md` document so it is easy
```
enisoc: If we tag issues with all the above info, can we just find the history with a GitHub search query (and/or a dashboard)? Until a given patch release goes out, the info is subject to change anyway, and once the release happens the info is reflected in the release notes.

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,78 @@
...
+
+## 1.X.(Y-1)
+
+...
+
+```
+
+## Communicating security fixes to the community
+
+Due to the sensitive nature of security fixes, they may be omitted from the weekly summary
```
enisoc: The patch release manager should probably mention in the sig-release meeting that they are working with the [product security team](https://github.com/kubernetes/community/blob/master/contributors/devel/security-release-process.md) on an issue, even if they can't say what the issue is.

contributors/devel/release/patch_release.md
pwittrock: After going through the 1.7 generated release notes, I am skeptical that we are at a state where we can trust auto-tagging of anything to produce something consumable, but am open to exploring anything.  Cherry pick PRs don't map 1-1 with issues, and the release note descriptions + PR labels are frequently confusing or misleading.

The end state we need is a reliable summary of all issues (fixed and newly discovered) for each patch release that folks can use to make a decisions that depend on the stability of the release.  Suggestions on how to produce this with less overhead?

contributors/devel/release/patch_release.md
pwittrock: Maybe?  Would such information be sufficient for the patch release manager to make decisions and have full confidence in the patch release state without reaching out to the PR authors?  IMHO, we need any conversations with the patch release manager to be summarized somewhere that is published publicly.

contributors/devel/release/patch_release.md
pwittrock: +1

contributors/devel/release/patch_release.md
enisoc: I'm not sure if it's what you meant by "auto-tagging", but I meant that we would define a set of tags that we expect humans to apply, like `status/under-investigation`. It's the same amount of human work as putting it into the template, except that the GitHub issue remains the single source of truth and the info is searchable. If there is no issue or the release note is confusing, I would rather correct those things in GitHub's database than write down a better copy in a static file.

contributors/devel/release/patch_release.md
enisoc: Whatever would have been written in this history doc, we could add as a comment on the GitHub issue, possibly with a magic string so a script could pull it out to produce a summary. I just want the issue to be the source of truth for everything known about that issue.

contributors/devel/release/patch_release.md
pwittrock: I think that could work.  We would need to support tagging issues to specific patch releases that they are resolved in.  Thoughts?

contributors/devel/release/patch_release.md
pwittrock: What would be the workflow for a user to quickly find all the issues in 1.7.1?  GH comment threads are difficult to quickly parse a status summary out of, and current state of an issue is often unclear.  Putting the template in the issue description might work, but who would be responsible for keeping it up to date?  Thoughts?

contributors/devel/release/patch_release.md
pwittrock: Agreed it is too much to have the patch manager attend both meetings and give summaries in both.  Is the community meeting still the correct place to give a release summary?  Should we just move it to the sig-release meeting and send out an update to kubernetes-dev?

idvoretskyi: @pwittrock I'd prefer having a brief regular update at the community meeting - to keep the whole community updated, and more detailed (ad-hoc) at the sig-release meeting.

idvoretskyi:
@pwittrock @kubernetes/sig-release-members any progress on this?

pwittrock:
@idvoretskyi There are 2 open issues we don't have a resolution for:

- Community and sig-release updates
Where are updates given?  Who gives the updates at each meeting?

- How are changes recorded so they are communicated
Issues?  Release notes?  Email summary?  How is status across issues aggregated and summarized?

pwittrock:
@idvoretskyi Updated the proposal to use GH issues as the canonical source and defined a structure for them.  Lets punt on changing how we update the community via meetings and address that separately.

pwittrock:
PTAL

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,87 @@
...
+
+GitHub issues targeted at patch releases must contain the following items:
+
+### Release label
+
+The release label contains information about which releases are impacted by the issue, and
+which release the issue is resolved in.
+
+- `release/introduced-in/1.7.0`
+- `release/resolved-in/1.7.2`
```
spiffxp: I get the intent and I can't think of a better way, but it does feel like opening the door to drowning in labels.  This implies as followup:
- `release/introduced-in/x.y.z` label created as part of each release
- `release/resolved-in/x.y.z` created as part of each release
- a `/release` bot command to allow non-members to add these labels?
- verification that each issue to which each cherry-pick PR is linked has these labels... I'm guessing this is a small enough set that release managers can verify this by hand?

the semver bikeshed nerd in me is wondering if these should be `release/foo-in/v1.2.3`

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,87 @@
+# Building Kubernetes patch releases
+
+This document describes the process for creating Kubernetes patch releases and communicating publishing
+status regarding known issues impacting the current release.
+
+## Process
+
+After a Kubernetes minor release, patch releases are managed by the Kubenretes patch release czar.
```
luxas: nit: Do we still use the `czar` term? I thought we had switched to `manager`...?
Not very important though

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,87 @@
...
+- `release/introduced-in/1.7.0`
+- `release/resolved-in/1.7.2`
+
+### Priority label
+
+The priority label describes the urgency and severity of the issue.
+
+- `priority/critical-urgent`: the issue is severe enough that it requires an immediate patch release
+  (storage or network issues that can cause data corruption or outages for instance)
+- `priority/important-soon`: the issue should be fixed in a patch does not need to be fixed immediately
```
luxas: This seems to be equal to no label at all (pretty much current process). Why add this label with no extra value?

contributors/devel/release/patch_release.md
luxas: Is the purpose here to be able to filter or is this just metadata?
If metadata is the important thing, we can do this like the release note block

Something like this:
If this is a cherrypick, comment out this field:
<!- -
\```
release/introduced-in=vX.Y.Z
\```
-->

Also we need a `/` command for adding `cherrypick-candidate`. Any org member should be able to do that,
but obvs the patch manager is the one that's adding the `cherrypick-approved` label.

IMO the bot could fill in `release/resolved-in/x.y.z` metadata since it can track which commits are in which releases...

contributors/devel/release/patch_release.md
```diff
@@ -0,0 +1,87 @@
...
+
+The issue description should be kept up-to-date by the issue owner through out the resolution process.  Items
+discovered as part of triage should be reflected in the issue description.
+
+While much of this information may exist as discussion comments on the issue,
+providing the information in an easily understandable format and location
+makes it much easier to quickly understand the state of upcoming patch release.
+
+```
+## Symptoms
```
luxas: +1 for having this special template for cherrypicks

contributors/devel/release/patch_release.md
pwittrock: Good point @luxas about the bot being able to apply the label.  I like that.

pwittrock: I think we want to filter to be able to ask - which issues were resolved in patch-release-X.  This would allow us to generate the doc you are suggesting.

pwittrock: @spiffxp I agree drowning in labels isn't great.  Maybe the introduced-in and resolved-in could be part of the template, and the bot could automatically apply the labels.  I think this is pretty critical information.  If a major regression is introduced, a user should be able to easily understand which releases are impacted.

contributors/devel/release/patch_release.md
pwittrock: To avoid ambiguity.  Implicitly defaulting the severity means we can't differentiate between "someone forgot to add a label to a critical issue" and "this can wait on this"

contributors/devel/release/patch_release.md
pwittrock: done

pwittrock:
PTAL

pwittrock:
Having a bot set the resolved-in makes sense.  A human will have to do it until the bot can do it for them.

pwittrock:
@spiffxp @idvoretskyi @luxas @enisoc 

Thanks for all the feedback.  Sorry for taking a while to respond.

pwittrock:
Plan on merging this next week if there are no additional comments.

PR #777, opened 2017-07-03
Update outdated links from kubernetes.io

zhangxiaoyu-zidif:
Update outdated links from  kubernetes.io


MHBauer:
I didn't check them all, but it looks like it changes the lings from redirect pages to the direct targets of those redirect pages.

PR #778, opened 2017-07-04
Proposal for service policy

xilabao:
ServicePolicy is the intent to manage the services.

### DIFF
--- a/contributors/design-proposals/service-policy.md
+++ b/contributors/design-proposals/service-policy.md
@@ -1,0 +1,68 @@
+# Service Policy Proposal
+
+## Abstract
+
+ServicePolicy allows cluster administrators to control the creation and validation of a service.
+The intent of ServicePolicy is to manage the services.
+
+## Motivation
+
+As we know, field level control over allowed fields in services does not exist yet. we can not limit
+the ip range for different users. and if we don't want people use the node port directly, we can not
+implement it although the podsecuitypolicy can limit the "hostPort".
+
+## Goals
+
+1.  Associate [service accounts](../design-proposals/service_accounts.md), groups, and users with
+a set of constraints that dictate how a security context is established for a service.
+1.  Provide the ability for users and infrastructure components to use services with elevated privileges
+on behalf of another user or within a namespace where privileges are more restrictive.
+1.  Secure the ability to reference elevated permissions or to change the constraints under which
+a user runs.
+
+## Design
+
+### Model
+
+ServicePolicy objects exist in the root scope, outside of a namespace. The ServicePolicy will reference
+users and groups that are allowed to operate under the constraints.
+
+```go
+// ServicePolicy governs the ability to make requests that affect the SecurityContext
+// that will be applied to a pod and container.
+type ServicePolicy struct {
+	unversioned.TypeMeta `json:",inline"`
+	api.ObjectMeta       `json:"metadata,omitempty"`
+
+	// Spec defines the policy enforced.
+	Spec ServicePolicySpec `json:"spec,omitempty"`
+}
+
+// ServicePolicySpec defines the policy enforced.
+type ServicePolicySpec struct {
+	// Type is the type that can be visit the service
+	Type []AllowedType `json:"type,omitempty"`
+	// NodePorts determines which node port ranges are allowed to be exposed.
+	NodePorts []NodePortRange `json:"nodePorts,omitempty"`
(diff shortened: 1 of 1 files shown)


xiangpengzhao:

cmluciano:
Is this similar to egress networkpolicy, but extended to also include a service selector?

cc @kubernetes/sig-network-feature-requests

mikedanese:
@kubernetes/sig-auth-feature-requests

xilabao:
@liggitt @smarterclayton @pweil-

contributors/design-proposals/service-policy.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Abstract
+
+ServicePolicy allows cluster administrators to control the creation and validation of a service.
+The intent of ServicePolicy is to manage the services.
+
+## Motivation
+
+As we know, field level control over allowed fields in services does not exist yet. we can not limit
+the ip range for different users. and if we don't want people use the node port directly, we can not
```
caseydavenport: Could you add some detail here into the use-cases you're considering for limiting the IP range and nodePort range for different users?

NodePorts are in my mind primarily intended to be a resource where you don't _care_ about the specific port that has been allocated since they will usually be behind a loadbalancer of some sort anyway.

ClusterIPs are a bit different, but similarly it's best not to make too many assumptions about the range or value you might be assigned else you break application portability across clusters.

contributors/design-proposals/service-policy.md
```diff
@@ -0,0 +1,68 @@
...
+implement it although the podsecuitypolicy can limit the "hostPort".
+
+## Goals
+
+1.  Associate [service accounts](../design-proposals/service_accounts.md), groups, and users with
```
caseydavenport: How do you imagine tying a service account to one of these ServicePolicy objects? I don't currently see that in this proposal.

Presumably there's another "binding" type resource that's missing?

caseydavenport:
> Is this similar to egress networkpolicy, but extended to also include a service selector?

@cmluciano my read of this is different from NetworkPolicy. NP is control over traffic, whereas this is meant to add field-level access control of the Service object.

contributors/design-proposals/service-policy.md
xilabao: > Could you add some detail here into the use-cases you're considering for limiting the IP range and nodePort range for different users?
NodePorts are in my mind primarily intended to be a resource where you don't care about the specific port that has been allocated since they will usually be behind a loadbalancer of some sort anyway.
ClusterIPs are a bit different, but similarly it's best not to make too many assumptions about the range or value you might be assigned else you break application portability across clusters.

Since loadbanlancer is suggested to use, I think we could limit the use of NodePorts (e.g. disable  nodeport).
If we can cut clusterIPs into slices,  may be can reduce the collisions?

contributors/design-proposals/service-policy.md
xilabao: > How do you imagine tying a service account to one of these ServicePolicy objects? I don't currently see that in this proposal.
Presumably there's another "binding" type resource that's missing?

Reference to PodSecurityPolicy, ServicePolicy may be need "binding" type resource too.

wojtek-t:
@bowei @dnardo - FYI

coresolve:
Can this be resolved with resource quota? 
https://kubernetes.io/docs/concepts/policy/resource-quotas/#object-count-quota

You can limit the number of services of type loadbalancer or type node port using that.

contributors/design-proposals/service-policy.md
smarterclayton: We can already limit nodeports with quota.

contributors/design-proposals/service-policy.md
```diff
@@ -0,0 +1,68 @@
...
+a set of constraints that dictate how a security context is established for a service.
+1.  Provide the ability for users and infrastructure components to use services with elevated privileges
+on behalf of another user or within a namespace where privileges are more restrictive.
+1.  Secure the ability to reference elevated permissions or to change the constraints under which
```
smarterclayton: Why would you not use pod secure policy for this?  It’s not clear what you intend.

contributors/design-proposals/service-policy.md
smarterclayton: Reducing collisions of service ips doesn’t sound very useful.  Explain more, please.

m1093782566:
xref: https://github.com/kubernetes/community/pull/1551

FYI.

PR #779, opened 2017-07-04
Fix link to point to English docs instead of Chinese

nikhita:
Earlier the link redirected to docs in Chinese because of the redirects present here: https://github.com/kubernetes/kubernetes.github.io/blob/master/cn/docs/tutorials/object-management-kubectl/object-management.md.


cblecker:
This change is okay because it eliminates the need for the redirect at all, but the fact that a previously en URL is redirecting to cn docs in the first place is concerning.

cc: @kubernetes/sig-docs-maintainers @chenopis

PR #782, opened 2017-07-06
Add New Resource API proposal

vikaschoudhary16:
Notes for reviewers

First proposal submitted to the community repo, please advise if something's not right with the format or procedure, etc.

cc @aveshagarwal @jeremyeder @derekwaynecarr @vishh @jiayingz 

Signed-off-by: vikaschoudhary16 <vichoudh@redhat.com>

### DIFF
--- a/contributors/design-proposals/resource-class.md
+++ b/contributors/design-proposals/resource-class.md
@@ -1,0 +1,317 @@
+# New Resource API Proposal
+
+
+- [Abstract](#abstract)
+- [Background](#background)
+- [Use Stories](#user-stories)
+  - [As a cluster operator](#as-a-cluster-operator)
+  - [As a developer](#as-a-developer)
+  - [As a vendor](#as-a-vendor) 
+- [Objectives](#objectives)
+- [Non Objectives](#non-objectives)
+- [Components](#components)
+  - [ComputeResource API](#computeResource-api)
+  - [ResourceClass API](#resourceclass-api)
+- [Kubelet Extension](#kubelet-extension)
+- [Scheduler Extension](#scheduler-extension)
+- [Quota Extension](#quota-extension)
+- [Roadmap](#roadmap)
+
+_Authors:_
+
+* @vikaschoudhary16 - Vikas Choudhary &lt;vichoudh@redhat.com&gt;
+* @jiayingz - Jiaying Zhang &lt;jiayingz@google.com&gt;
+
+## Abstract
+In this document we will describe a new resource API model to better support non-native compute resources on Kubernetes.
+
+## Background
+We are seeing increasing needs to better support non-native compute resources on Kubernetes that cover a wide range of resources such as GPUs, High-performance NICs, Infiniband and FPGAs. Such resources often require vendor specific setup, and have rich sets of different properties even across devices of the same type. This brings new requirements for Kubernetes to better support non-native compute resources to allow vendor specific setup, dynamic resource exporting, flexible resource configuration, and portable resource specification.
+
+The device plugin support added in Kubernetes 1.8 makes it easy for vendors to dynamically export their resources through a plugin API without changing Kubernetes core code. Taking a further step, this document proposes a new resource API model that can be used to describe, manage, and consume such vendor specific and metadata rich resources in simple and portable ways. This model introduces two API objects: ComputeResource API object that represents a type of physical compute resource with a unique set of common properties, and ResourceClass API object that maps from the physical ComputeResources matching certain specified property constraints to a portable name, with which cluster admins can further define the associated resource policies such as quota and limitRange.
+
+## Use Stories
+### As a cluster operator:
+- Nodes in my cluster has GPU HW from different generations. I want to classify GPU nodes into one of the three categories, silver , gold and platinum depending upon the launch timeline of the GPU family eg: Kepler K20, K80, Pascal P40, P100, Volta V100. I want to charge each of the three categories differently. I want to offer my clients 3 GPU rates/classes to choose from.<br/>
+**Motivation:** As time progresses in a cluster lifecycle, new advanced, high performance, expensive variants of GPUs gets added to the cluster nodes. At the same time older variants also co-exist. There are workloads which strictly wants latest GPUs and also there are workloads which are fine with older GPUs. But since there is a wide range of types, it will be hard to manage and confusing at the same time to have granularity at each GPU type. Grouping into few broad categories will be convenient to manage.<br/>
+**Can this be solved without resource classes:** A unique taint can be used to represent a category like silver. Nodes can be tainted accordingly depending upon the type of GPUs availability. User pods can use tolerations to steer workloads to the appropriate nodes. But problem is how to restrict a user pod from not using the toleration that it should not be using?<br/>
+**How Resource classes can solve this:** I, operator/admin, creates three resource classes: GPU-Platinum, GPU-Gold, GPU-Silver. Now since resource classes are quota controlled, end-user will be able to request resource classes only if quota is allocated.
+
+- I want a mechanism where it is possible to offer a group of devices, which are co-located on a single node and shares a common property, as a single resource that can be requested in pod container spec. Example, N GPU units interconnected by NVLink or N cpu cores on same NUMA node.<br/>
+**Motivation:** Increased performance because of local reference. Local reference also helps better use of cache<br/>
+**How Resource classes can solve this:**  Property/attribute which forms the grouping can be advertised in the device attributes and then a resource can be created to form a grouped super-resource based on that property.<br/>
+**Can this be solved without resource classes:** No
+
+- I want to have quota control on the devices at the granularity of device properties. For example, I want to have a separate quota for ECC enabled GPUs. I want a specific user to not let use more than ‘N’ number of ECC enabled GPUs overall at namespace level.<br/>
+**Motivation:**  This will make it possible to charge user per specialized hw consumption. Since special HW is costly, as an Operator i want to have this capability.<br/>
+**How Resource classes can solve this:** Quota will be supported on resource class objects and by allowing resource request in user pods via resource class, charging policy can be linked with resource consumption.<br/>
+**Can this be solved without resource classes:** No
+
+- In my cluster, i have many different classes(different capabilities) of a device type (ex: NICs). End user’s expectations are met as long as device has a very small subset of these capabilities. I want a mechanism where end user can request devices which satisfies their minimum expectation.
+Few nodes are connected to data network over 40 Gig NICs and others are connected over normal 1 Gig NICs. I want end user pods to be able to request:
+Data network connectivity with high network performance 
+In default case, data network connectivity via normal 1 Gbps NICs<br/>
+**Motivation:** If some workloads demand higher network bandwidth, it should be possible to run these workloads on selected nodes.<br/>
+**Can this be solved without resource classes:** Taints and tolerations can help in steering pods but the problem in that there is no way today to have access control over use of tolerations and therefore if multiple users are there, it is not possible to have control on allowed tolerations.<br/>
+**How Resource classes can solve this:** I can define a ResourceClass for the high-performance NIC with minimum bandwidth requirements, and makes sure only users with proper quota can use such resources.
+
+- I want to be able to utilize different 'types' of a HW resource while not losing workload portability when moving from one cluster to another. There can be Nvidia GPUs on one cluster and AMD GPUs on another cluster. This is example of different ‘types’ of a HW resource(GPU). I want to offer GPUs to be consumed under a same portable name, as long as their capabilities are almost same. If pods are consuming these GPUs with a generic resource class name, workload can be migrated from one cluster to another transparently.<br/>
+**Motivation:**  less downtime, optimal use of resources<br/>
+**How Resource classes can solve this:** Explained above<br/>
+**Can this be solved without resource classes:** No
+
+- I want to be able to query or list extended resources in my cluster through the standard Kubernetes API or CLI.<br/>
+**Motivation:**  Improves introspectability of non-primary compute resources.<br/>
+**How the new ComputeResource API can solve this:** By introducing ComputeResource API objects, we allow users with proper access to list all compute resources across the cluster to have a cluster level overview or get a list of ComputeResource objects with certain properties through standard Kubernetes API operations. This allows easy introspection of compute resources.
+  
+### As a developer:
+- I want the ability to be able to request devices which have specific capabilities. Eg: GPUs that are Volta or newer.<br/>
+  **Motivation:** I want minimum guaranteed compute performance<br/>
+  **Can this be solved without resource classes:**<br/>
+  - Yes, using node labels and NodeLabelSelectors.
+    Problem: Same problem of lack of access control on using labelselectors at user level as with the use of tolerations. 
+  - OR, Instead of using resource class, provide flexibility to query resource properties directly in pod container resource requests.
+    Problem: In a large cluster, computing operators like “greater than”, “less than” at pod creation can be a very slow operation and is not scalable. 
+
+  **How Resource classes can solve this:**
+The Kubernetes scheduler is the central place to map container resource requests expressed through ResourceClass names to the underlying qualified physical resources, which automatically supports metadata aware resource scheduling.
+
+- As a data scientist, I want my workloads to use advanced compute resources available in the running clusters without understanding the underlying hardware configuration details. I want the same workload to run on either on-prem Kubernetes clusters or on cloud, without changing its pod spec. When a new hardware driver comes out, I hope all the required resource configurations are handled properly by my cluster operators and things will just continue to work for any of my existing workloads.<br/>
+**Motivation:** Separation of concerns between cluster operators and cluster users.<br/>
+**Can this be solved without resource classes:**<br/>
+Without the additional abstraction layer, consuming the non-standard, metadata-rich compute resources would be fragmented. More likely, we would see cluster providers implement their own solutions to address their user pains, and it would be hard to provide a consistent user experience for consuming extended resources in the future.
+
+### As a vendor:
+- I want an easy and extensible mechanism to export my resource to Kubernetes. I want to be able to roll out new hardware features to the users who require those features without breaking users who are using old versions of hardware.<br/>
+**Motivation:** enables more compute resources and their advanced features on Kubernetes<br/>
+**Can this be solved without resource classes:**<br/>
+Yes, Using node labels and NodeLabelSelectors.<br/>
+Problem: Lack of access control and lack of the ability to differentiate between hardware properties on the same node. E.g., if on the same node, some GPU devices are connected through nvlink while others are connected through PCI-e, vendors don’t have ways to export such resource properties that can have very different performance impacts.<br/>
+**How Resource classes can solve this:**<br/>
+Vendors can use DevicePlugin API to propagate new hardware features, and provide best-practice ResourceClass spec to consume their new hardware or new hardware features on Kubernetes. Vendors don’t need to worry supporting this new hardware would break existing use cases on old hardware because the Kubernetes scheduler takes the resource metadata into account during pod scheduling, and so only pods that explicitly request this new hardware through the corresponding ResourceClass name will be allocated with such resources.
+
+## Objectives
+By introducing this new API model, we are hoping to achieve the following objectives:
+- **Allows workloads to request compute resources with wide range of properties in simple and standard ways.** By introducing ComputeResource API object, Kubelet can encapsulate the resource metadata information into the corresponding ComputeResource object, and propagate this information to the scheduler. With this information, the Kubernetes scheduler can determine the fitness of a node for a container resource request expressed through ResourceClass name by evaluating whether the node has enough capacity of a ComputeResource matching the property constraints specified in the ResourceClass. This allows the Kubernetes scheduler to take resource property into account to schedule pods on the right node whose hardware configuration meets the specific resource metadata constraints.
+- **Allows cluster admins to configure and manage different kinds of non-native compute resources in flexible and simple ways.** A cluster admin creates a ResourceClass object that specifies a portable ResourceClass name (e.g., `fast-nic-gold`), and list of property matching constraints (e.g., `resourceName in (solarflare.com/fast-nic, intel.com/fast-nic)`, or `type=XtremeScale-8000`, or `bandwidth=100G`, or `zone in (us-west1-b, us-west1-c)`). The property matching constraints follow the generic LabelSelector format, which allows us to cover a wide range of resource specific properties. The cluster admin can then define and manage resource quota with the created ResourceClass object.
+- **Allows vendors to export their resources on Kubernetes more easily.** The device plugin that a vendor needs to implement to make their resource available on Kubernetes lives outside Kubernetes core repository. The device plugin API will be extended to pass device properties from device plugin to Kubelet. Kubelet will propagate this information to the scheduler through ComputeResource and the scheduler will match a ComputeResource with certain properties to the best matching ResourceClass, and support resource requests expressed through ResourceClass name. The device plugin only needs to retrieve device properties through some device-specific API or tool, without needing to watch or understand either ComputeResource objects or ResourceClass objects.
+- **Provides a unified interface to interpret compute resources across various system components such as Quota and CLI.** By introducing ComputeResource API objects, we allow users to list all compute resources across the cluster to have a cluster level overview or get a list of ComputeResource objects with certain properties through standard Kubernetes API operations. This allows easy introspection of compute resources.
+- **Supports node-level resources as well as cluster-level resources.** Certain types of compute resources are tied to single nodes and are only accessible locally on those nodes. On the other hand, some types of compute resources such as network attached resources, can be dynamically bound to a chosen node that doesn’t have the resource available till the binding finishes. For such resources, the metadata constraints specified through ResourceClass can be consumed by a standalone controller or a scheduler extender during their resource provisioning or scheduler filtering so that the resource can be provisioned properly to meet the specified metadata constraints.
+- **Supports cluster auto scaling for extended resources.** We have seen challenges on how to make cluster autoscaler work seamlessly with dynamically exported extended resources. In particular, for node level extended resources that are exported by a device plugin, cluster autoscaler needs to know what resources will be exported on a newly created node, how much of such resources will be exported, and how long it will take for the resource to be exported on the node. Otherwise, it would keep creating new nodes for the pending pod during this time gap. For cluster level extended resources, their resource provisionings are generally performed dynamically by a separate controller. Cluster autoscaler needs to be taught to filter out the resource requests for such resources for the pending pod so that it can create right type of node based on node level resource requests. Note that Kubelet and the scheduler have the similar need to ignore such resource requests during their PodFitsResources evaluation. By exporting this information through the ResourceClass API, we can ensure a consistent resource evaluation policy among these components.
+- **Defines an easy and seamless migration path** for clusters to adopt ResourceClass even if they have existing pods requesting compute resources through raw resource name. In particular, suppose a cluster is running some workloads that have non native compute resources, such as `nvidia.com/gpu`, in their pod resource requirements. Such workloads should still be scheduled properly when the cluster admin creates a ResourceClass that matches the gpu resources in the cluster.
+
+
+## Non Objectives
+- Extends the current resource requirement API of the container spec. The current resource requirement API is basically a “name:value” list. A commonly arising question is whether we should extend this API to support resource metadata requirements. We can consider this as a possible extension orthogonal to the ComputeResource and ResourceClass proposal. A primary reason we propose to introduce ComputeResource and ResourceClass APIs first is because non-native compute resources usually lack standard resource properties. Although there are benefits to allow users to directly express their resource metadata requirements in their container spec, it may also compromise workload portability if not used carefully. By introducing ResourceClass as an additional resource abstraction layer, users can express their special resource requirements through a high-level portable name, and cluster admins can configure compute resources properly on different environments to meet such requirements. We feel this helps promote portability and separation of concerns, while still maintains API compatibility.
+- Unifies with the StorageClass API. Although ResourceClass shares many similar motivations and requirements as the existing StorageClass API, they focus on different kinds of resources. StorageClass is used to represent storage resources that are stateful and contains special storage semantics. ResourceClass, on the other hand, focuses on stateless compute resources, whose usage is bound to container lifecycle and can’t be shared across multiple nodes at the same time. For these reasons, we don’t plan to unify the two APIs.
+- Resource overcommitment, fractional resource requirements, native compute resource (i.e., cpu and memory) with special metadata requirements, and group compute resources. They are out of our current scope.
+
+## Components
+### ComputeResource API
+We propose to introduce a new ComputeResource API object that represents a type of physical compute resource with unique sets of properties. We can start with the following simple spec:
+
+```golang
+struct ComputeResource {
+  metav1.ObjectMeta
+  Spec ComputeResourceSpec
+  Status ComputeResourceStatus
+}
+struct ComputeResourceSpec {
+     // For node-level resource, this is the name of the Node that owns this ComputeResource.
+    // For cluster-level resource, this field is empty.
+    NodeName string
+     // Raw resource name. E.g.: nvidia.com/gpu
+    ResourceName string
+    // gpuType: k80
+    // zone: us-west1-b
+   // Note Kubelet adds a special property corresponding to the above ResourceName field.
+   // This will allow a single ResourceClass (e.g., “gpu”) to match multiple types of 
+   // resources (e.g., nvidia.com/gpu and amd.com/gpu) through general set selector.
+   Properties map[string]string
+}
+struct ComputeResourceStatus {
+   Capacity resource.Quantity
+}
+```
+
+ComputeResource objects are similar to PV objects on storage. It is tied to the physical resource that can have a wide range of vendor specific properties. Kubelet will create or update ComputeResource objects upon any resource availability or property change for node-level resources. Once a node is configured to support ComputeResource API and the underlying resource is exported as a ComputeResource, its quantity should NOT be included in the conventional NodeStatus Capacity/Allocatable fields to avoid resource multiple counting. ComputeResource API can be included into NodeStatus to facilitate resource introspection.
+For cluster level resources, a special controller or a scheduler extender can create a ComputeResource and dynamically bind that to a node during or after scheduling.
+
+### ResourceClass API
+ComputeResource mostly serves as an internal API to communicate resource states between different Kubernetes components, like Kubelet and the Scheduler. We will introduce ResourceClass as an additional abstraction layer for better portability and easy management, similar to why StorageClass was introduced to address similar requirements on storage. The scheduler will need to track both ResourceClass and ComputeResource objects to support resource requests expressed through ResourceClass names. 
+
+During the initial phase, we propose to start with the following ResourceClass API spec that defines the basic ResourceClass name to the underlying compute resource plus metadata matching. We can extend this API to include auto provisioning configurations and support group resources in the following phases of development. However, this document will mostly focus on the design to support initial phase requirements.
+
+```golang
+// +nonNamespaced=true
+// +genclient=true
+
+type ResourceClass struct {
+        metav1.TypeMeta
+        metav1.ObjectMeta
+        Spec ResourceClassSpec
+        // +optional
+}
+
+type ResourceClassSpec struct {
+        // defines general resource property matching constraints.
+        // e.g.: zone in { us-west1-b, us-west1-c }; type: k80
+        MetadataRequirements metav1.LabelSelector
+        // used to compare preference of two matching ResourceClasses
+        // The purpose to introduce this field is explained more later
+        Priority int
+}
+```
+
+YAML example 1:
+```yaml
+kind: ResourceClass
+metadata:
+  name: nvidia.high.mem
+spec:
+  labelSelector:
+    - matchExpressions:
+        - key: "Kind"
+          operator: "In"
+          values:
+            - "nvidia-gpu"
+        - key: "memory"
+          operator: "GtEq"
+          values:
+            - "30G"
+```
+Above resource class will select all the nvidia-gpus which have memory greater
+than and equal to 30 GB.
+
+YAML example 2:
+```yaml
+kind: ResourceClass
+metadata:
+  name: fast.nic
+spec:
+  labelSelector:
+    - matchExpressions:
+        - key: "Kind"
+          operator: "In"
+          values:
+            - "nic"
+        - key: "speed"
+          operator: "GtEq"
+          values:
+            - "40GBPS"
+```
+Above resource class will select all the NICs with speed greater than equal to
+40 GBPS.
+
+### Kubelet Extension
+Kubelet needs to be extended to create or update ComputeResource objects upon any resource availability or property change for node-level resources. During the initial phase, we plan to start with exporting extended resources through the ComputeResource API but leaves primary resources in its current exporting model. We can extend the ComputeResource model to support primary resources later after getting more experience through the initial phase.
+
+On node level, extended resources can be exported automatically by a third-party plugin through the Device Plugin API. We will extend the current Device Plugin API to allow device plugins to send to the Kubelet per-device properties during device listing. Exporting device properties at per-device level instead of per-resource level allows a device plugin to manage devices with heterogeneous properties.
+
+After receiving device availability and property information from a device plugin, Kubelet will create a ComputeResource object to represent the device resource and the associated resource properties. Once a node is configured to support ComputeResource API and the underlying resource is exported as a ComputeResource, its quantity should NOT be included in the conventional NodeStatus Capacity/Allocatable fields to avoid resource multiple counting. ComputeResource API can be included into NodeStatus to facilitate resource introspection.
+
+For cluster level resources, a special controller or a scheduler extender can create a ComputeResource and dynamically bind that to a node during or after scheduling.
+
+The ComputeResource name needs to be unique and deterministically generated. We propose to use “nodeName-resourceName-propertyHash” naming convention for node level resources, where propertyHash is generated by calculating a hash over all resource properties. The properties of a physical resource may change, e.g., through a driver or node upgrade. When this happens, Kubelet will need to create a new ComputeResource object and delete the old one. However, some pods may already be scheduled on the node and allocated with the old ComputeResource, and it is hard for the scheduler to do its bookkeeping when such pods are still running on the node while the associated ComputeResource is removed. We have a few options to handle this situation.
+- First, we may document that to change or remove a device resource, the node has to be drained. This requirement however will complicate device plugin upgrade process.
+- Another option is that Kubelet can evict the pods that are allocated with an unexisting ComputeResource. Although simple, this approach may disturb long-running workloads during device plugin upgrade.
+- To support a less disruptive model, upon resource property change, Kubelet can still export capacity at old ComputeResource name for the devices used by active pods, and exports capacity at new matching ComputeResource name for devices not in use. Only when those pods finish running, that particular node finishes its transition. This approach avoids resource multiple counting and simplifies the scheduler resource accounting. One potential downside is that the transition may take quite long process if there are long running pods using the resource on the nodes. In that case, cluster admins can still drain the node at convenient time to speed up the transition. Note that this approach does add certain code complexity on Kubelet DeviceManager component.
+
+### Scheduler Extension
+The scheduler needs to watch for ComputeResource objects and ResourceClass objects, and caches the binding information between the ResourceClass and the matching ComputeResource objects to serve container resource request expressed through ResourceClass names.
+
+A natural question is how we should define the matching behavior. Suppose there are two ResourceClass objects. ResourceClass RC1 has metadata matching constraint “property1 = value1”, and ResourceClass RC2 has metadata matching constraint “property2 = value2”. Suppose a ComputeResource has both “property1: value1” and “property2: value2” properties, and so match both ResourceClasses. Now should the scheduler consider this ComputeResource as qualified for both ResourceClasses RC1 and RC2, or only one of them?
+There are a few possible ways to solve multiple matching problem:
+- One way to deal with this problem is to disallow ambiguous matching on ResourceClasses. i.e., two ResourceClasses can’t define non overlapping property constraints. This way, we can order ResourceClasses based on their number of constraints and match a ComputeResource to the ResourceClass with most constraints. Although we can enforce this policy during ResourceClass creation, we feel this requirement may complicate the use of ResourceClass, given that the underlying resources can have wide range of properties across different generations of products.
+- Another option is to allow a ComputeResource to match any ResourceClasses as long as it qualifies their metadata constraints. Although this approach sounds natural, it may create deployment complexities. Suppose a cluster already has certain existing physical resources and the corresponding ResourceClasses matching these physical resources. Now suppose the cluster is provisioned with a new type of physical resources that can be distinguished through a new property but they also have other properties that match certain existing ResourceClasses. To prevent this new type of resources from being scheduled by existing workloads accidentally, the cluster admin would need to modify the existing ResourceClasses to not match the new property explicitly.
+- Proposed solution: introduce a special Priority field in the ResourceClass spec. When a ComputeResource matches multiple ResourceClasses, the scheduler will choose the one with the highest Priority. With this approach, even though we allow cluster admins to create multiple ResourceClasses that may potentially match the same physical resource, the underlying resource can only be requested through a single ResourceClass name (we do allow the underlying resource to be requested through the raw Resource name for backward compatibility that we will discuss in more detail later). A valid question is whether users may desire the flexibility to request the same physical resource through multiple ResourceClass names. From our past experience, non-native computing resources are generally more expensive and users usually prefer to reserving such resources for workloads that really need them. I.e., ambiguous resource matching is generally depreferred. Therefore, we favor this approach for its simplicity, more deterministic scheduling behavior, and the model is relatively easy to understand by users. If multiple matching is indeed desired, users have the option to define multiple ResourceClasses with the same priority, in which case the scheduler will consider a ComputeResource to match all these ResourceClasses.
+
+Because a ComputeResource object can match multiple ResourceClasses, Scheduler and Kubelet need to ensure a consistent view on ComputeResource to ResourceClass request binding. Let us consider an example to illustrate this problem. Suppose a node has two ComputeResources, CR1 and CR2, that have the same raw resource name but different sets of properties. Suppose they both satisfy the property constraints of ResourceClass RC1, but only CR2 satisfies the property constraints of another ResourceClass RC2. Suppose a Pod requesting RC1 is scheduled first. Because the RC1 resource request can be satisfied by either CR1 or CR2, it is important for the scheduler to record the binding information and propagate it to Kubelet, and Kubelet should honor this binding instead of making its own binding decision. This way, when another Pod comes in that requests RC2, the scheduler can determine whether Pod can fit on the node or not, depending on whether the previous RC1 request is bound to CR1 or CR2.
+
+To maintain and propagate ComputeResource to ResourceClass binding information, the scheduler will need to record this information in a newly introduced PodSpec field, similar to the existing NodeName field, and Kubelet will need to consume this information. Note that if this field is pre-specified, scheduler should honor this binding instead of overwriting it, similar to how it handles pre-specified NodeName.
+
+As mentioned earlier, we want to ensure that existing pods requesting resources through raw resource name can continue to be scheduled properly when administrators add ResourceClass in a cluster. To guarantee this, the scheduler may consider raw resource as a special ResourceClass that matches the ComputeResources of that raw resource.
+
+Cluster admins may want to modify an existing ResourceClass, e.g., adding or removing some metadata constraints or changing its priority, or deleting some ResourceClasses that are still requested by some pods. By making the scheduler the single central place to make the ResourceClass to ComputeResource binding decision, and record the binding information in a PodSpec field, ResourceClass updates should have little impact on the pods already being scheduled, and the scheduler can make further scheduling decisions based on its current ResourceClass state and ComputeResource allocation state.
+
+### Quota Extension
+As mentioned earlier, an important goal of introducing ResourceClass is to allow cluster admins to manage non-native compute resource quota more easily. With the current resource quota system, admins can define the hard limit on the overall quantity of pod resource requests through raw resource name. They don’t have a mechanism to express finer-granularity resource request constraints that map to resources with certain metadata constraints. Following the previous GPU example, admins may want to constraint P100 and K80 GPU requests in a given namespace, other than the total GPU resource requests. By allowing admins to express resource quota constraints at ResourceClass level, we can now support more flexible resource quota specification.
+
+This flexibility also brings an interesting question on how we can enable this benefit while also maintain backward compatibility. We have previously discussed the importance to maintain backward compatibility and how we may extend the scheduler so that pods requesting extended resources through raw resource name can still be scheduled properly. Now a natural question is whether ResourceClass quota constraints should also limit how pods’ raw resource requests are bound to available ResourceClass resources. 
+
+Note that resource quota enforcement and resource request fitting validation are performed by different components. The former enforcement is implemented in the Quota admission controller, while the later task is performed by the scheduler. Enforcing quota constraints at scheduling time would require significant amount of change. We hope to avoid this complexity at the initial stage. Instead, we propose to keep resource quota at its current meaning, i.e., a resource quota limits how much resource can be requested by all of the pods in their container specs in a given namespace. We feel this still allows cluster admins to gradually take advantage of ResourceClass through a manageable migration path. Again, let us walk through an example migration scenario to illustrate how cluster admins can gradually use ResourceClass to express finer-granularity quota constraints:
+
+Step 1: Suppose a cluster already has the following quota constraint to limit total number of gpu devices that pods in a given namespace may request:
+
+```yaml
+kind: ResourceQuota
+metadata:
+  name: gpu-quota-example
+spec:
+  Hard:
+    nvidia.com/gpu: “100”
+```
+Step 2: Now suppose the cluster admin want to enforce finer-granularity resource constraints through ResourceClass. Following the previous example, the admin defines two resource classes, gpu-silver that maps to K80 gpu devices, and gpu-gold that maps to P100 gpu devices. The cluster admin then migrates a small percent, say 10%, of workloads in the namespace to request resources through the created ResourceClass names in e.g., 3:2 ratio. The admin can then modify the existing quota spec as follows to express the current resource constraints:
+
+```yaml
+kind: ResourceQuota
+metadata:
+  name: gpu-quota-example
+spec:
+  Hard:
+    nvidia.com/gpu: “90”
+    gpu-silver: "6"
+    gpu-gold: “4”
+```
+Step 3: Now suppose the experiment of using ResourceClass is successful, and the cluster admin have converted all the workloads running in the namespace to request resources through ResourClass name, the admin can then enforce the finer granularity resource quota across all workloads by modifying the quota spec as follows:
+
+```yaml
+kind: ResourceQuota
+metadata:
+  name: gpu-quota-example
+spec:
+  Hard:
+    nvidia.com/gpu: “0”
+    gpu-silver: "60"
+    gpu-gold: “40”
+```
+
+It is easy to notice that we propose to avoid introducing any changes to the existing quota tracking system. It is possible that in the future, we may need to extend the scheduler to enforce certain quota constraints to support more advanced use cases like better batch job scheduling. When that time comes, the ResourceClass API can be extended to directly express quota constraints, but we will leave that discussion outside the current design.
+
+## Roadmap
+### Phase 1: Support ResourceClass based scheduling for node resources
+Defines the ComputeResource API, the ResourceClass API, and extends Kubelet, scheduler, and device plugin API to support ResourceClass based resource requirements.
+
+### Phase 2: Support auto provisioning and cluster resources
+As we are offering more flexibility on introducing new types of compute resources into Kubernetes, we are also seeing more challenges on how to make auto provisioning work seamlessly with such wide range, dynamically exported, and probably dynamically bound resources. Here is the list of problems we have seen in the past related to auto scaling and auto provisioning in the presence of extended resources.
+
+- For node level extended resources that are exported by a device plugin, cluster autoscaler needs to know what resources will be exported on a newly created node, how much of such resources will be exported, and how long it will take for the resource to be exported on the node. Otherwise, it would keep creating new nodes for the pending pod during this time gap.
+- For cluster level extended resources, their resource provisionings are generally performed dynamically by a separate controller. Cluster autoscaler and auto provisioner need to be taught to filter out the resource requests for such resources for the pending pod so that it can create right type of node based on node level resource requests. Note that Kubelet and the scheduler have the similar need to ignore such resource requests during their PodFitsResources evaluation. By exporting this information through the ResourceClass API, we can ensure a consistent resource evaluation policy among these components.
+
+During the second phase, we propose to extend the ResourceClass API with the following fields that can convey the resource provisioning related information.
+
+```golang
+ProvisionConfig {
+	// CLUSTERLEVEL or NODELEVEL.
+    // CLUSTERLEVEL resource can be ignored by scheduler, Kubelet, and
+    // Cluster Autoscaler during their node resource fitness evaluation.
+	ResourceType string
+	// Resource specific node label keys attached on a node with the resource. The
+	// value of such a node label indicates the expected quantity of the resource on the node.
+	// It may contain multiple node label keys if the given ResourceClass matches multiple
+	// kinds of node resources. E.g., we may have different node labels for different types
+	// of GPU. A resource class matching all such GPUs should list all their node label keys.
+	// Cluster autoscaler can use this field in combination with resource allocatable
+    // to determine node readiness for the given resource, and the expected quantity
+    // of the resource on the node.
+	CapacityNodeLabelKeys []string
+	// Represents latency upper bound for the resource to be exported on a new node.
+	// Cluster Autoscaler can wait up to this bound before it considers the node not viable
+	// for the resource.
+	ProvisionSeconds int64
+}
+```
+### Phase 3: Support group resources
+i.e., a ResourceClass can represent a group of resources. E.g., a gpu-super may include two gpu devices with high affinity, an infiniband-super may include a high-performance nic plus 1G memory, etc.
+
+


vikaschoudhary16:
cc @aveshagarwal

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// A resource selector requirement is a selector that contains values, a key, and an operator
+// that relates the key and values
+type ResourceSelectorRequirement struct {
+        // The label key that the selector applies to
+        // +patchMergeKey=key
+        // +patchStrategy=merge
+        Key string
+        // Example 0.1, intel etc
```
jeremyeder: What is this?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,209 @@
...
+}
+type ResourceSelectorOperator string
+
+const (
+        ResourceSelectorOpIn           ResourceSelectorOperator = "In"
+        ResourceSelectorOpNotIn        ResourceSelectorOperator = "NotIn"
+        ResourceSelectorOpExists       ResourceSelectorOperator = "Exists"
+        ResourceSelectorOpDoesNotExist ResourceSelectorOperator = "DoesNotExist"
+        ResourceSelectorOpGt           ResourceSelectorOperator = "Gt"
+        ResourceSelectorOpLt           ResourceSelectorOperator = "Lt"
```
jeremyeder: You mentioned ```Equal To``` on line 43 but no ```Eq``` operator here.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,209 @@
...
+3. Node removal from the cluster.
+4. Pod creation if pod requests a resource class.
+5. Pod deletion if pod was consuming resource class.
+
+`ResourceClassStatus` serves the following two purposes:
+* Scheduler predicates evaluation while pod creation. For details, please refer further sections
+* User can view the current usage/availability details about the resource class using kubectl.
+
+### User story
+Admin knows what all devices are present on nodes and have deployed corresponding device plugins. Device plugins will make devices appear in node status. Next admin creates resource classes that have generic/portable names and metadata which can select available devices.
```
jeremyeder: Slightly rephrased this paragraph.

The administrator as deployed device plugins to support hardware present in the cluster.  The Kubelet will update its status indicating the presence of this hardware via those device plugins.  To offer this hardware to applications deployed on kubernetes in a portable way, the administrators creates a number of resource classes to represent that hardware.  These resource classes will include metadata about the device and selection criteria.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,209 @@
...
+1. A user submits a pod spec requesting 'X' resource classes.
+2. The scheduler filters the nodes which do not match the resource requests.
+3. scheduler selects a device for each resource class requested and annotates the pod object with device selection info.
+4. Kubelet reads the device request from pod annotation and calls `Allocate` on the matching Device Plugins.
+5. The user deletes the pod or the pod terminates
+6. Kubelet reads pod object annotation for devices consumed and calls `Deallocate` on the matching Device Plugins
+
+The scheduler is incharge of both, selecting a node and also selecting a device for requested resource classes.
```
jeremyeder: In addition to node selection, the scheduler is also responsible for selecting a device that matches the resource class requested by the user.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,209 @@
...
+### Reason for not preferring device selection at kubelet
+Kubelet does not maintain any cache. Therefore to know the availability of a device, will have to iterate over all the existing pods running on the node. This is already done today while running predicates for each new incoming pod at kubelet. Even if we assume that scheduler cache and consumption state that is created at runtime for each pod, are exactly same, current api interfaces does not allow to pass selected device to container runtime (where actually device plugin will be invoked). This problem occurs because devices are determined internally from resource classes while other resource requests can be determined from pod object directly. To summarize, device selection at the kubelet can be done in one of the following two ways:
+* Select device at pod admission while applying predicates and change all api interfaces that are required to pass selected device to container runtime manager.
+* Create resource consumption state again at container runtime manager and select device.
+
+None of the above approach seems cleaner than doing device selection at scheduler.
```
jeremyeder: Is it worth mentioning here that this decision helps to retain a clean abstraction between the container runtime and kubernetes?  ISTR that was a portion of the discussion.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,209 @@
...
+1. Iterate over the all the devices on the node that this pod was scheduled to and find out the devices being used by pod.
+2. For each device consumed by pod, update availability state of Resource classes which can select this device in the cache.
+3. Patch `ResourceClassStatus` with new availability state.
+
+## Kubelet Changes
+Update logic at container runtime manager to look for device annotations, prefixed by 'scheduler.alpha.kubernetes.io/resClass' and call matching device plugins.
+
+## Future Scope
+* RBAC: It can further be explored that how to tie resource classes with RBAC like any other existing API resource objects.
+* Nested Resource Classes: In future device plugins and resource classes can be extended to support the nested resource class functionality where one resource class could be comprised of a group of sub-resource classes. For example 'numa-node' resource class comprised of sub-resource classes, 'single-core'.
```
jeremyeder: We need to communicate the plans/decisions on how ResourceClasses relate to Opaque Integer Resources.

@ConnorDoyle

contributors/design-proposals/resource-class.md
vikaschoudhary16: these are example of `Values`. For example if `Key` is 'version', `Values` can have '0.1'  OR `Key` is 'vendor', `Values` can have 'intel'. I will make it more clear.

contributors/design-proposals/resource-class.md
vikaschoudhary16: will correct at line 43 to use "In" in place of "Equal to"

contributors/design-proposals/resource-class.md
vikaschoudhary16: Thanks, will update!

contributors/design-proposals/resource-class.md
vikaschoudhary16: whatever existing resource discovery tools are available which update node objects with OIR, those will adapt to update node status `devAllocatable` with devices instead. Will add more details.

RenaudWasTaken:
@vikaschoudhary16 I don't see any mention of overlapping resources is that something you plan to address in another PR ?

vikaschoudhary16:
@RenaudWasTaken It is explained implicitly by explaining how resource class can select device with k-v metadata and how a resource class can select different devices. Overlapping is essentially the functionality which is providing portability.
I will add more details to make it more visible and explicit.
Thanks!

RenaudWasTaken:
> @RenaudWasTaken It is explained implicitly by explaining how resource class can select device with k-v metadata

I'm also interested in how you solved selecting multiple overlapping resource class because it is not a trivial problem. 
An example would be:
- node has 2 GPUs:
    - 1 GPU with 4G
    - 1 GPU with 8G
- Cluster has 2 resource classes:
   - GPU with memory > 2 (lowMemGPU)
   - GPU with memory > 4 (highMemGPU)
- User submits pod with 2 containers:
  - The first one requests 1 lowMemGPU
  - The second one requests 1 highMemGPU

In this example if your selection algorithm is a first fit then it is very possible that you won't be able to satisfy the request because you might give the gpu with 8G to the first container.

It seems to me that your only solution is to generate all the possible permutations but that doesn't scale well...
Another solution I thought about was to have multiple algorithms and give the option to either the end user or the cluster admin to select which one he wanted to use.
But it feels like this edge case doesn't have any good solutions...

vikaschoudhary16:
@RenaudWasTaken 
> Another solution I thought about was to have multiple algorithms and give the option to either the end user or the cluster admin to select which one he wanted to use.


Thanks i missed to add these details though had in mind. This proposal is implementing a first fit selection process. For better resource usage and scalability, in future, selection algorithm will be optimized on the lines you mentioned.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Abstract
+In this document we will describe *resource classes* which is a new model to
+represent compute resources in Kubernetes. This document should be seen as a
+successor to [device plugin proposal](https://github.com/kubernetes/community/pull/695/files)
+and has a dependency on the same.
+
+## Motivation
+Compute resources in Kubernetes are represented as a key-value map with the key
+being a string and the value being a 'Quantity' which can (optionally) be
```
vikaschoudhary16: @fabiand yes.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+all the (operator,key,value) are ANDed together to evaluate the result.
+
+YAML example 1:
+```yaml
+kind: ResourceClass
+metadata:
+  name: nvidia.high.mem
+spec:
+  resourceSelector:
+    -
```
vikaschoudhary16: @fabiand this is yaml syntax for nested sequences, https://learn.getgrav.org/advanced/yaml#sequences

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+      matchExpressions:
+        -
```
vikaschoudhary16: @fabiand same as above.

contributors/design-proposals/resource-class.md
fabiand: Oh obviously …
Would pulling `matchExpressions` into line 89 look saner`

```yaml
    - matchExpressions:
```

vikaschoudhary16: Ah, sure, will update. Thanks!

saad-ali:
A couple of drive-by comments:

1. What happens when a `ResourceClass` is deleted before the pod referencing it? Particularly if kubelet has to enforce the limits.
2. Also consider the case where kubelet and/or scheduler crash and lose in-memory state (and the ResourceClass object is deleted).
3. Define which, if any, of the fields of `ResourceClass` are immutable, and if they are mutable what the expected behavior is.
4. Maybe worth calling out more explicitly that this will be non-namespaced and expected to be created by cluster admins.

vikaschoudhary16:
@saad-ali 
Thanks for taking a look. Please find my responses as follows:
> 1. What happens when a ...

This proposal's scope is bounded by device plugin proposal's scope, which is to cover devices only and not resources such as CPU and memory.
Since device for the pod is being selected by scheduler, so in case resource class gets deleted, either pod will fail the predicate at scheduler or pod will have the device info.  

> 2. Also consider the case where kubelet and/or scheduler crash ...

In scheduler crash case, for any unscheduled pods which request the deleted resource class, scheduling will fail at predicate validation. Similarly [predicate will also fail at kubelet because for any new pod  kubelet creates resource consumption state from the beginning](https://github.com/vikaschoudhary16/kubernetes/pull/3/files#diff-c83f39f5599c8d22b6fd1b0e34e649e4R102).

3. Once created, resource class object is immutable, only status will be updated by scheduler later on.
4.  sure, will mention explicitly.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+   devices as well.
+
+From the events perspective, handling for the following events will be added/updated:
+
+### Resource Class Creation
+1. Init and add resource class info into local cache
+2. Iterate over all existing nodes in cache to figure out if there are devices
+   on these nodes which are selectable by resource class. If found, update the
+   resource class availability status in local cache.
+3. Patch the status of resource class api object with availability state in locyy
```
balajismaniam: s/locyy/local

contributors/design-proposals/resource-class.md
balajismaniam: I think using `In` for `Eq` causes confusion.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+   cache
+
+### Resource Class Deletion
+Delete the resource class info from the cache.
+
+### Node Addition
+Scheduler already caches `NodeInfo`. Now additionally update device state:
```
balajismaniam: Would `NodeInfo` contain `requestedResources` of `ResourceClass` similar to other resources?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+        ResourceSelectorOpDoesNotExist ResourceSelectorOperator = "DoesNotExist"
+        ResourceSelectorOpGt           ResourceSelectorOperator = "Gt"
+        ResourceSelectorOpLt           ResourceSelectorOperator = "Lt"
+)
+```
+### ResourceClassStatus
+```golang
+type ResourceClassStatus struct {
+        Allocatable resources.Quantity
+        Request     resources.Quantity
```
balajismaniam: s/Request/Requested. It is similar to `requestedResources` in `nodeInfo` of the `schedulercache`.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+4. Kubelet reads the device request from pod annotation and calls `Allocate` on
+   the matching Device Plugins.
+5. The user deletes the pod or the pod terminates
+6. Kubelet reads pod object annotation for devices consumed and calls `Deallocate`
+   on the matching Device Plugins
+
+In addition to node selection, the scheduler is also responsible for selecting a
+device that matches the resource class requested by the user.
+
+### Reason for not preferring device selection at kubelet
```
balajismaniam: May be change this to `Preferring device selection in the scheduler (instead of the Kubelet)`.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+Kubelet does not maintain any cache. Therefore to know the availability of a device,
```
balajismaniam: s/to know the availability of a device/to know the quantity of the device available for scheduling/

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+will have to calculate current total consumption by iterating over all the admitted
+pods running on the node. This is already done today while running predicates for
```
balajismaniam: May be rephrase as `the quantity of the device already consumed should be calculated by iterating`.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+each new incoming pod at kubelet. Even if we assume that scheduler cache and
+consumption state that is created at runtime for each pod, are exactly same,
+current api interfaces does not allow to pass selected device to container manager
+(where actually device plugin will be invoked from). This problem occurs because
+devices are determined internally from resource classes while other resource
```
balajismaniam: Is this not opposite (i.e., resource classes are determined from devices)?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+Scheduler already listens and maintains state in the cache for any changes in
+node or pod objects. We will enhance the logic:
+1. To listen and maintain the state in cache for user created *Resource Class* objects.
+2. To look for device related details in node objects and maintain accounting for
+   devices as well.
+
+From the events perspective, handling for the following events will be added/updated:
+
+### Resource Class Creation
+1. Init and add resource class info into local cache
```
balajismaniam: Nit: s/Init/Initialize

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Kubelet Changes
+Update logic at container runtime manager to look for device annotations,
+prefixed by 'scheduler.alpha.kubernetes.io/resClass' and call matching device
+plugins.
+
+## Opaque Integer Resources
+This API will supercede the [Opaque Integer Resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#opaque-integer-resources-alpha-feature)
+(OIR). External agents can continue to attach additional 'opaque' resources to
+nodes, but the special naming scheme that is part of the current OIR approach
```
balajismaniam: s/but the special naming scheme that is part of the current OIR approach will no longer be necessary/using device plugins

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+   which can select this device in the cache.
+3. Patch `ResourceClassStatus` with new availability state.
+
+## Kubelet Changes
+Update logic at container runtime manager to look for device annotations,
+prefixed by 'scheduler.alpha.kubernetes.io/resClass' and call matching device
+plugins.
+
+## Opaque Integer Resources
+This API will supercede the [Opaque Integer Resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#opaque-integer-resources-alpha-feature)
```
balajismaniam: Nit: s/supercede/supersede

contributors/design-proposals/resource-class.md
ConnorDoyle: This section was a little confusing to read (what exactly does device selection mean?) -- maybe a concrete example would help?

Was the option considered where the scheduler is responsible for mapping resource classes from the container spec to resource names from the node capacityV2, but assigning specific devices is left to the Kubelet? IIRC a reason to delay device binding to Kubelet was to avoid publishing hardware topology to the API server.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+### Node Deletion
+If node has devices which are selectable by existing resource classes:
+1. Adjust resource class state in local cache.
+2. Update resource class status by patching api object.
+
+### Pod Creation
+1. Get the requested resource class name and quantity from pod spec.
+2. Select nodes by applying predicates according to requested quantity and Resource
+   class's state present in the cache.
+3. On the selected node, select a Device from the stored devices info in cache
```
ConnorDoyle: Would it be better to say something like "concrete resource" here instead of "Device"? There's no limitation that resource classes can only represent devices right?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+
+1. Define and add support in the API for a new type, *Resource Class*.
+2. Add support for *Resource Class* in the scheduler.
+
+## Non Objectives
+1. Discovery, advertisement, allocation/deallocation of devices is expected to
+   be addressed by [device plugin proposal](https://github.com/kubernetes/community/pull/695/files)
+
+## Resource Class
+*Resource Class* is a new type, objects of which provides abstraction over
```
ConnorDoyle: s/provides/provide

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,333 @@
...
+The administrator has deployed device plugins to support hardware present in the
+cluster. Device plugins, running on nodes, will update node status indicating
+the presence of this hardware. To offer this hardware to applications deployed
+on kubernetes in a portable way, the administrator creates a number of resource
+classes to represent that hardware. These resource classes will include metadata
+about the devices as selection criteria.
+
+1. A user submits a pod spec requesting 'X' resource classes.
+2. The scheduler filters the nodes which do not match the resource requests.
+3. scheduler selects a device for each resource class requested and annotates
```
ConnorDoyle: > annotates the pod object

Do you have an example of what the annotation might look like?

contributors/design-proposals/resource-class.md
vikaschoudhary16: @balajismaniam 
[`NodeInfo` will have dict of `ResourceClassInfo`s](https://github.com/vikaschoudhary16/kubernetes/pull/3/files#diff-558cb8bde14dca10a3151bfc222a3aaeR84)
`ResourceClassInfo` structure will have `Requested`

contributors/design-proposals/resource-class.md
vikaschoudhary16: @balajismaniam 
not necessarily device plugins, OIRs could also be used same as it is used today(without device plugins). Using OIR with device plugins will be case where plugin has not adapted device advertisement as per [device plugin proposal](https://github.com/kubernetes/community/pull/695) and use OIRs to advertise resources.

contributors/design-proposals/resource-class.md
vikaschoudhary16: @ConnorDoyle 
I have updated several initial sections of the doc to make it more clear. Device selection means selecting a device using the resource class details which includes applying different operators like  explained in 'Resource Class' section of this document. 
> scheduler is responsible for mapping resource classes from the container spec to resource names from the node capacityV2

Yes. and this section notes down challenges in that approach. I have updated this section to make it more clear. Hope it is more understandable now.
 

> Kubelet was to avoid publishing hardware topology to the API server.

This proposal assumes that device details are updated in node status by vendor device plugins, as proposed in device plugin proposal.

contributors/design-proposals/resource-class.md
vikaschoudhary16: yes, there is a limitation. Resource class would be able to represent only the devices which are advertised by device plugins in [Device structure format](https://github.com/RenaudWasTaken/community/blob/806b924ff7222a2f68c9a1caf79daba064ef0100/contributors/design-proposals/device-plugin.md#device) in the node status.

contributors/design-proposals/resource-class.md
vikaschoudhary16: @ConnorDoyle 
sure, will update. Thanks!

vikaschoudhary16:
Thanks @ConnorDoyle @saad-ali @balajismaniam @RenaudWasTaken @fabiand for the review comments.
I have updated the doc to address those. PTAL!

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+
+
+## Future Scope
+* RBAC: It can further be explored that how to tie resource classes with RBAC
+  like any other existing API resource objects.
+* Nested Resource Classes: In future device plugins and resource classes can be 
+  extended to support the nested resource class functionality where one resource
+  class could be comprised of a group of sub-resource classes. For example 'numa-node'
+  resource class comprised of sub-resource classes, 'single-core'.
+* Multiple device selection algorithms, each with a different selection strategy,
```
RenaudWasTaken: I think this should be thoroughly discussed and not just a "side note" on the bottom of the design doc.
The decision we take regarding the way we address this problem will have lasting impact on
GPU providers (e.g: GKE).
But then I'm not against selling more GPUs because we have poor scheduling algorithms :D

vikaschoudhary16: @RenaudWasTaken totally agree. Scheduler section is also updated about "first fit" approach.
Example use case you quoted can be handled by "best fit" approach, which I thought to cover in follow-up proposal. And meanwhile if a user dont want to use "first fit", he/she can request devices using OIRs and bypassing resource classes. Thats how i am thinking. Anyways if community thinks differently, happy to discuss and adapt the proposal accordingly.

On a funny note,
Please note that importance of a point is not directly propotional to the height of the point in the design doc. :D

contributors/design-proposals/resource-class.md
balajismaniam: Got it. Thanks.

contributors/design-proposals/resource-class.md
balajismaniam: Thanks for the PoC link. Will take a look.

contributors/design-proposals/resource-class.md
vikaschoudhary16: /sig scheduling

derekwaynecarr:
i will prioritize reviewing this further when device plugins are agreed upon.

bgrant0607:
cc @davidopp @vishh @bsalamat 
@kubernetes/sig-node-proposals @kubernetes/sig-scheduling-proposals

bgrant0607:
I haven't had time to look at this in detail, but I wrote some text on the topic of the FooClass pattern of extension registration APIs in the architecture roadmap doc:

https://docs.google.com/document/d/1VEzP20JvIKckPrGSs_R-D4vlwFzMAaqJd-9Lvfyi3jw/edit#heading=h.u6da5oouw3la

The map in StorageClass is underspecified. We need at least descriptions in addition to parameter names. Somewhere (I'll try to find it) I posted an example using a subset of json schema.

See also comments here:
https://github.com/kubernetes/kubernetes/issues/50658#issuecomment-326174398

vishh:
There are a few other features that are not tackled in this proposal like cluster level resources, being able to list available capacity easily, etc.
We do not have bandwidth to review this feature as of now. I'm hoping we can pick this up after v1.9.
@bgrant0607 I'm working closely with @vikaschoudhary16 and when we start making progress on this, we will engage with you.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+  further sections
+* User can view the current usage/availability details about the resource class
+  using kubectl.
+
+### User story
+The administrator has deployed device plugins to support hardware present in the
+cluster. Device plugins, running on nodes, will update node status indicating
+the presence of this hardware. To offer this hardware to applications deployed
+on kubernetes in a portable way, the administrator creates a number of resource
+classes to represent that hardware. These resource classes will include metadata
```
ScorpioCPH: > the administrator creates a number of resource classes to represent that hardware

Hi, I am just wondering why we create more than one resource class to represent one hardware, should it be one by one matching? 

In other word, one hardware need and only need one reource class to represent it (with multi selectors of course).

Please let me know if i missing something about this proposal, thanks~

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+  resourceSelector:
+    - matchExpressions:
+        - key: "Kind"
+          operator: "In"
+          values:
+            - "nvidia-gpu"
+        - key: "memory"
+          operator: "GtEq"
+          values:
+            - "30G"
```
weiwei04: In the [device-plugin design](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-management/device-plugin.md)
>```go
>// E.g:
>// struct Device {
>//    ID: "GPU-fef8089b-4820-abfc-e83e-94318197576e",
>//    State: "Healthy",
>//}
>message Device {
>	string ID = 2;
>	string health = 3;
>}
>```

it seems the device resource are just expressed as number of devices, not including information like `gpu cores on a device`, `gnu memory`, from where the ResourceClass can get extra info like`memory`?

fabiand: We must not expose device specififc properties, as we - at least - can imagine to use the resource class to expose virt functionality of the node, and there you would not have gpu cores or gpu memory

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+kind: ResourceClass
+metadata:
+  name: hugepages-1gig
+spec:
+  resourceSelector:
+    - matchExpressions:
+        - key: "Kind"
+          operator: "In"
+          values:
+            - "huge-pages"
```
ddysher: nit: this should be `hugepages-1Gi` or `hugepages-2Mi`.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+application pod may perform well only with resources which have a certain
+capability, for example GPUs greater than version 'V'. OIR does not allow such,
+metadata based, selection/filtering of resources.
+The current model requires identity mapping between available resources and requested
+resources. Identity mapping means there must be a one-to-one mapping between the
+resource reference used in the spec to request the resource and resource reference
+used to advertise the resource availablity.
+
+Since 'CPU' and 'Memory' are resources that are available across all kubernetes
+deployments and need no metadata to describe the resource type any further,
```
ddysher: We've seen use case where people want kubernetes to schedule pods based on different types of CPUs, e.g. run critical application on newer/better version of CPU. Is this something within the scope of resource class?  I read comment below, it seems resource claas only cares about resources advertised by device plugin.  We could of course write one, but i'm not quite sure how it interacts with the standard integer based CPU scheduling.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+_Authors:_
+
+* @vikaschoudhary16 - Vikas Choudhary &lt;vichoudh@redhat.com&gt;
+* @aveshagarwal - Avesh Agarwal &lt;avagarwa@redhat.com&gt;
+
+## Abstract
+In this document we will describe *resource classes* which is a new model to
+represent compute resources in Kubernetes. This document should be seen as a
+successor to [device plugin proposal](https://github.com/kubernetes/community/pull/695/files)
+and has a dependency on the same.
```
ddysher: imho, the proposal is orthogonal to device plugin; device plugin provides more 'primitives' for resource scheduling, while resource claas provides better resource representation.  Resource class is less useful without device plugin so it makes sense to implement it after device plugin, but i'm not quite sure about the dependency here.  Is it a hard dependency?

contributors/design-proposals/resource-class.md
fabiand: Right, that's a good hint. Please mark that in the release notes.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+    'a set of accepted values' are supported on the compute resource metadata.
+
+## Objectives
+
+1. Define and add support in the API for a new type, *Resource Class*.
+2. Add support for *Resource Class* in the scheduler.
+
+## Non Objectives
+1. Discovery, advertisement, allocation/deallocation of devices is expected to
+   be addressed by [device plugin proposal](https://github.com/kubernetes/community/pull/695/files)
```
tengqm: Resource class is a very generic term, so I was expecting a more general model for ALL resource types we can think of. In other words, the Resource Class model should not depend on the device plugins. This "non objectives" item is a little bit misleading. It is somehow implying that the underlying provisioning of resource "instances" are based on device plugins.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+
+## Resource Class
+*Resource Class* is a new type, objects of which provide abstraction over
+[Devices](https://github.com/RenaudWasTaken/community/blob/a7762d8fa80b9a805dbaa7deb510e95128905148/contributors/design-proposals/device-plugin.md#resourcetype).
+A *Resource Class* object selects devices using `matchExpressions`, a list of
+(operator, key, value). A *Resource Class* object selects a device if at least
+one of the `matchExpressions` matches with device details. Within a matchExpression,
+all the (operator,key,value) are ANDed together to evaluate the result.
```
tengqm: Then ... there is an OR logic between individual matchExpressions?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+        Allocatable resources.Quantity
+        Requested   resources.Quantity
+}
+```
+ResourceClass status is updated by the scheduler at:
+1. New *Resource Class* object creation.
+2. Node addition to the cluster.
+3. Node removal from the cluster.
+4. Pod creation if pod requests a resource class.
+5. Pod deletion if pod was consuming resource class.
```
tengqm: Pod update?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+2. For each device consumed by pod, update availability state of Resource classes
+   which can select this device in the cache.
+3. Patch `ResourceClassStatus` with new availability state.
+
+## Kubelet Changes
+Update logic at container runtime manager to look for device annotations,
+prefixed by 'scheduler.alpha.kubernetes.io/resClass' and call matching device
+plugins.
+
+## Opaque Integer Resources
```
tengqm: Note that OIR is deprecated by extended resource as well ...

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/782)

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+used to advertise the resource availablity.
+
+Since 'CPU' and 'Memory' are resources that are available across all kubernetes
+deployments and need no metadata to describe the resource type any further,
+the current user facing API (Pod Specification) remains portable as long as pod
+requests only CPU and Memory. However the current model cannot support
+complex resources like GPUs, ASICs, NICs, local storage, etc.
+To support heterogeneity, portability and management at scale, such resources
+must be represented(advertised) in a form which allows metadata inclusion and a
+metadata based resource selection mechanism.
```
sameo: @vikaschoudhary16 Ok, does that mean we need to add a property map to the device plugin `Device` structure for this PR to be relevant? I'm probably going to state the obvious but such map would be a requirement for supporting heterogeneous nodes.

resouer: +1, offline talked with sameo and seems GPU topology, FPGAs are all cases to consume these properties.

ns-sundar:
The notion of the scheduler picking a device has its merits. However, as proposed, it would still not handle all device scenarios, such as for FPGAs. That would require a scheduler extender.

contributors/design-proposals/resource-class.md
RenaudWasTaken: That was what was in the initial design document.
@vishh @jiayingz were concerned that since this was not supported at the scheduler level it was useless

resouer: @RenaudWasTaken OK, I see the problem. Em, what if we put this proposal under the umbrella of topology design? 

Otherwise, I think a simple scheduler extender also make sense here. We can include a example at least.

resouer:
I would love to see this been reactive :)  Or we are planning to start a new one? @vikaschoudhary16 @sameo

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+In this document we will describe *resource classes* which is a new model to
+represent compute resources in Kubernetes. This document should be seen as a
+successor to [device plugin proposal](https://github.com/kubernetes/community/pull/695/files)
+and has a dependency on the same.
+
+## Motivation
+Kubernetes system knows only two resource types, 'CPU' and 'Memory'. Any other
+resource can be requested by pod using opaque-integer-resource(OIR) mechanism.
+OIR is a key-value pair with the key being a string and the value being a 'Quantity'
+which can (optionally) be fractional. The current model is great for supporting
```
jaypipes: If an OIR quantity can be optionally fractional, how is it an "Integer" resource?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+But there is a problem in representing resources like GPUs, ASICs, local storage
+etc in the form of OIRs. Each of such resource type generally has a rich
+metadata like version, capabilities etc. to describe the resource. A particular
+application pod may perform well only with resources which have a certain
+capability, for example GPUs greater than version 'V'. OIR does not allow such,
+metadata based, selection/filtering of resources.
+The current model requires identity mapping between available resources and requested
+resources. Identity mapping means there must be a one-to-one mapping between the
+resource reference used in the spec to request the resource and resource reference
+used to advertise the resource availablity.
```
jaypipes: I would highly recommend not mixing quantitative and qualitative concepts. Resources are consumable quantitative things. Capabilities/metadata are not consumable. In other words, you don't "consume an OpenCL capability". Instead, you consume one or more GPUs (or vGPUs) where the thing that is providing the GPU/vGPU has OpenCL/CUDA driver support. The distinction is subtle but important. If you combine the quantitative and qualitative things together, you will end up needing to have many "resource classes" that are only different with regards to the various permutations of capabilities that the provider of the quantitative resources exposes.

For example, you will end up having a bunch of resource classes that read like a vendor product list:

 nvidia.com/gpu/tesla.k80
 nvidia.com/gpu/tesla.k40
 nvidia.com/gpu/tesla.c2075
 nvidia.com/gpu/tesla.c2070
 nvidia.com/gpu/tesla.c2050
 ... lots more for Intel, AMD, etc

Disadvantages of doing things this way include:

* what happens when the inevitable product line rebranding occurs
* you lose the abstraction that should be possible for a cloud to provide with regards to vendor neutrality. Instead of describing capabilities and quantities of like resources, instead the API is exposing vendor-specific stuff
* a lack of flexibility in how users request a quantity of some resource (vGPU or display head, for example) and capabilities (OpenCL support, for example)

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+  * I want to have a compute resource type, 'Resource Class',
+    which can be created with meaningful and portable names. This compute
+    resource can hold additional metadata as well, for example:
+    * `nvidia.gpu.high.mem` is the name and metadata is memory greater than 'X' GB.
+    * `fast.nic` is the name and associated metadata is bandwidth greater than
+      'B' gbps.
+  * If I request a resource `nvidia.gpu.high.mem` for my pod, any 'nvidia-gpu'
+    type device which has memory greater than or equal to 'X' GB, should be able
+    to satisfy this request, independent of other device capabilities such as
+    'version' or 'nvlink locality' etc.
```
jaypipes: "high" is subjective, though, right? Anything that is subjective is guaranteed to be non-portable across Kubernetes deployments. One k8s deployment might consider 2G to be high. Another might consider 8G to be high. This is the problem with using a qualitative string like "nvidia.gpu.high.mem" to describe a quantitative concept. Instead, I'd recommend segregating the quantitative concepts from the qualitative things.

For instance, consider making the resource classes strictly quantitative:
* `vgpu`
* `vgpu_display_head`
* `gpu_memory_bytes`

etc. And then have a separate set of qualitative strings/metadata that would allow the provider of those quantitative resources to be further described:

* `gpu.opencl.1_0`
* `gpu.opencl.1_1`
* `gpu.opencl.1_2`
* `gpu.opencl.2_0`
* etc..

That way the user can request 2 vGPUs and 4 display heads from something that provides OpenCL version 2.0 support. The separation of quantitative things and qualitative things is clean.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+  * Similarly, if I request a resource `fast.nic`, any nic device with speed
+    greater than 'B' gbps should be able to meet the request.
+  * I want a rich metadata selection interface where operators like 'Eq' for
+    'equals to', 'Lt' for 'less than', 'LtEq' for 'less than equal to', 'Gt' for
+    'greater than', 'GtEq' for 'greater than and equal to' and 'In' for
+    'a set of accepted values' are supported on the compute resource metadata.
```
jaypipes: I'd recommend starting with boolean capabilities and only adding the Lt/Gt rich stuff if the need arises.

contributors/design-proposals/resource-class.md
fabiand: Thanks for this nice example to highlight the qualitative vs quantitative aspects.

contributors/design-proposals/resource-class.md
vikaschoudhary16: Let say there are two NICs: NIC1 40 GB and NIC2 40GB + vlan offloading
Now two classes can be created:
1. Class offering only 40G speed NICs
2. Class offering vlan offloading capable NICs.

You see both the classes can select NIC1. This is what it meant by `number of resource classes to represent ...` . Does that make sense?

PS: sorry for late response.

contributors/design-proposals/resource-class.md
vikaschoudhary16: > from where the ResourceClass can get extra info like memory?
@weiwei04 This proposal suggests adding `Device` object to the api and also enhancing the device structure to have a map of device properties:
https://github.com/RenaudWasTaken/community/blob/a7762d8fa80b9a805dbaa7deb510e95128905148/contributors/design-proposals/device-plugin.md#resourcetype

When I submitted this PR, device plugin design was underway. I need to update this proposal according to current state of device plugin.

contributors/design-proposals/resource-class.md
vikaschoudhary16: By dependency, i meant the `Device` structure. This proposal needs `Device` to be an API object that is created by device plugins. Also `Device` api object should have a map of device properties.
Now that we a minimum device plugin functionality, i will update this proposal now with the enhancements that are needed in the existing device plugin functionality.

contributors/design-proposals/resource-class.md
vikaschoudhary16: > Is this something within the scope of resource class? 

Yes, but not in the first phase. Plan is to go incrementally and first handling only plugin advertised resources.

contributors/design-proposals/resource-class.md
vikaschoudhary16: No. In this example, `huge-pages` is the type of the resource and to indicate the flavor/characteristic like 1Gi or 2Mi , that is reflected in the name of the resource class. Name of resource class in this example is `hugepages-1gig`. Another example resource class for `kind` `huge-pages` could be:

```
kind: ResourceClass
metadata:
  name: hugepages-2Mi
spec:
  resourceSelector:
    - matchExpressions:
        - key: "Kind"
          operator: "In"
          values:
            - "huge-pages"
        - key: "size"
          operator: "GtEq"
          values:
            - "2Mi"

`huge-pages` is the resource for which many resource classes can exist.

File: contributors/design-proposals/resource-class.md [hunk 38]
```diff
@@ -0,0 +1,363 @@
...
+
```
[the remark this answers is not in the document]
vikaschoudhary16: > Is it possible to request two CPUs that share the same L2 cache, i.e. two cores from the same processor?

In first phase, only device plugin advertised resources are supported. Once CPU is supported by resource classes, then yes, it will be possible to query such resource requests. 
 
> Is is possible to request a FPGA card using PCIe with a prebuilt gzip logic on it?
>Is it possible to request a NIC that is connected to a different switch for security, availability's purpose?

Yes. Thats the whole point of resource classes i.e to be able to create custom selectors for devices. Any special characteristic of the Device will be updated in the device `properties` by the device plugin. For example in FPGA case, there can be two properties in the device of `kind` `FPGA`:
properties: ['PCI': 'true', 'prebuilt_gzip': 'true']
Then Admin would create, for example two resource classes:
1. 'FPGA': A resource which cares that node should have FPGA.
2. 'FPGA_Initialized': Resource class which will have selectors 'PCI' = 'true' and 'prebuild_gzip' = 'true' in the resource class object.

contributors/design-proposals/resource-class.md
vikaschoudhary16: To handle scheduler part, I propose to add Device object in the API. And then build the schedulercache reading the `Device` objects for properties and node status for availability. In the PoC, i did make these changes. Will update the PR to bring clarity on this.

vikaschoudhary16:
@ns-sundar 
> The notion of the scheduler picking a device has its merits. However, as proposed, it would still not handle all device scenarios, such as for FPGAs.That would require a scheduler extender.

This solution proposed Devices to be API objects and created by device plugins. And then scheduler will have everything it needs to handle even FPGAs.

> This proposal is layered on top of the device plugin proposal. If we provide a mechanism for devices to express resources

Device plugin will create `Device` objects at k8s API.

 > and pod spec to ask for resources,

Pod Spec will ask for resource class

 many aspects of this proposal would get addressed.

contributors/design-proposals/resource-class.md
vikaschoudhary16: >"high" is subjective, though, right? Anything that is subjective is guaranteed to be non-portable across Kubernetes deployments. One k8s deployment might consider 2G to be high. Another might consider 8G to be high. 

@jaypipes  `nvidia.gpu.high.mem` is just a reference to the resource class object, ex:
```
kind: ResourceClass
metadata:
  name: nvidia.high.mem
spec:
  resourceSelector:
    - matchExpressions:
        - key: "Kind"
          operator: "In"
          values:
            - "nvidia-gpu"
        - key: "memory"
          operator: "GtEq"
          values:
            - "30G"
```

"high" is not a selection parameter rather properties are, and properties are quantitative. Using above res class example, this is a resource class of GPUs which have memory greater than 30 GB. As long as resource class objects are same across clusters, there is no problem in app/pod portability across these clusters. WDYT?

contributors/design-proposals/resource-class.md
vikaschoudhary16: Lt/Gt is necessary to tackle the example scenario you suggested above:
>nvidia.com/gpu/tesla.k80
nvidia.com/gpu/tesla.k40
nvidia.com/gpu/tesla.c2075
nvidia.com/gpu/tesla.c2070
nvidia.com/gpu/tesla.c2050

For such cases, we can have a fewer resource classes classified using Lt/Gt/IN on the, for example, model number. one resource class `tesla-k-series` : 
```
kind: ResourceClass
metadata:
  name: tesla-k-series
spec:
  resourceSelector:
    - matchExpressions:
        - key: "Kind"
          operator: "In"
          values:
            - "nvidia-gpu"
        - key: "model"
          operator: "In"
          values:
            - "k80", "k40"
```
cluster admin must have knowledge that how device plugin is going to represent properties of device the `Device` api object and correspondingly should create abstracted resource classes, to be offered to users for consuption.

contributors/design-proposals/resource-class.md
jaypipes: @vikaschoudhary16 My point was that the resource class object is mixing quantitative and qualitative things. It's mixing quantitative things like amount of memory with qualitative things like "Kind: nvidia". By doing so, I suspect you'll end up in a situation where you have a whole bunch of resource classes that only differ slightly on some minor qualitative things (or minor quantitative things). For example, you may end up with separate resource class objects for various permutations of vendor, pGPU type, OpenCL/CUDA software version support, display heads, etc.

Is the idea to have lots of these resource class objects exposed to the user that they must search through to determine the exact combination of qualitative and quantitative things that the resource class object represents? Or is the idea to allow the user to specify the more granular resource quantitative and qualitative parts of the request in-line in the ContainerSpec's ResourceRequirements.requests object specification?

I'm trying to determine whether these resource classes will end up being a special snowflake way of requesting *some* kinds of resources instead of using the existing ResourceRequirements.resource/limits ways of specifying "normal" resource amounts.

Best,
-jay

contributors/design-proposals/resource-class.md
jaypipes: The In operator is fine -- it's just an OR operator which works fine when the qualitative strings are considered boolean values -- i.e. the provider of the resource either has the capability or it doesn't have the capability.

The example scenario I put above with nvidia.com/gpu/tesla.k80, etc was something I was recommending NOT to get into :)

I would not recommend using vendor product lines as the things that you match against because:

* The product/brand line from a vendor inevitably gets changed as marketing departments change :)
* The product/vendor/brand isn't necessarily what the *end user* is interested in. The end user wants a GPU (or # of vGPUs) that support X number of display heads and supports some specific software/hardware firmware version (e.g. supports OpenCL 2.0). The user doesn't care which vendor product is called. The user just wants a specific set of capabilities and requested resource amounts.

contributors/design-proposals/resource-class.md
vikaschoudhary16: @jaypipes My idea is that Resource classes will be a way to offer different flavors of devices. Just to give an analogy like ec2 instance types. End user will just request resource class in the pod. resource classes will be RBAC enabled. End user will not be able to modify/update resource class. 

> I'm trying to determine whether these resource classes will end up being a special snowflake way of requesting some kinds of resources instead of using the existing ResourceRequirements.resource/limits ways of specifying "normal" resource amounts.

resource class will be requested using the existing ResourceRequirements.resource/limits way and can be seen as an abstracted resource.

Thoughts?

contributors/design-proposals/resource-class.md
vikaschoudhary16: Sure.
That was just an example. It could be any property. Proposal provides approach to use any arbitrary key-value map of device properties. This proposal provides infra and it would depend on device plugin and cluster admin that how best/worse, it is used.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+  resourceSelector:
+    - matchExpressions:
+        - key: "Kind"
+          operator: "In"
+          values:
+            - "nic"
+        - key: "speed"
+          operator: "GtEq"
+          values:
+            - "40GBPS"
```
dashpole: What is the plan for comparing values?  Will all suffixes needed to be recognized by [quantity](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/suffix.go#L108)?  Or would someone specify `40G` instead of `40GBPS`, with the "per-second" implied by the key.
How would we know that `v0.28.3` comes after `v1.0.0`?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+        Status ResourceClassStatus
+}
+// Spec defines resources required
+type ResourceClassSpec struct {
+        // Resource Selector selects resources
+        ResourceSelector []ResourcePropertySelector
+}
+
+// A null or empty selector matches no resources
+type ResourcePropertySelector struct {
```
dashpole: What is the purpose of having two levels of lists in `ResourceClassSpec`?  At the end of the day they are all ANDed together.  If a use case is a good way to illustrate the reason for this, I would consider adding one.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+about the devices as selection criteria.
+
+1. A user submits a pod spec requesting 'X' resource classes.
+2. The scheduler filters the nodes which do not match the resource requests.
+3. scheduler selects a device for each resource class requested and annotates
+   the pod object with device selection info. eg:
+   `scheduler.alpha.kubernetes.io/resClass_test-res-class_nvidia-tesla-gpu=4`
+   where `scheduler.alpha.kubernetes.io/resClass` is the common prefix for all the
+   device annotations, `tes-res-class` is resource class name,
+   `nvidia-tesla-gpu` is the selected device name and `4` is the quantity requested.
```
jaypipes: @vikaschoudhary16 OK, I see here is how the user would request an amount of one of these resources. I still think that combining the *device* with the resource class in the selector is going to be problematic, though.

Let me ask something... do you envision the user knowing about the devices on nodes that are providing a particular resource class? Put another way... would (should?) the proposed end user that is constructing the pod spec and specifying the resource requirement selector `scheduler.alpha.kubernetes.io/resClass_test-res-class_nvidia-tesla-gpu=4` actually know that the "test-res-class" resources were being provided by the "nvidia-tesla-gpu" device(s)?

My guess is that, often, the end user won't know about the underlying device that is providing some amount of abstracted resources. The deployer of the cloud infrastructure knows that information, of course. But the end user that is consuming some of those resources won't necessarily know the exact vendor/device information and deployers may not actually *want* the end user to know the device/vendor/model specifics :-)

To respond to your comment from above that we should think of resource classes in the same vein as EC2 instance types, I'd just comment that Amazon has complete and total control over the information it gives users with regards to the amount of resources and the types of capabilities that its instance types comprise. Amazon can change (and has changed in the past) its mind about the quantities of resources and quality/vendor models associated with a particular instance type. If Kubernetes continues to be agnostic to cloud infrastructure providers, I think two things are necessary:

1) The way that Kubernetes advertises resources and capabilities *to end users* should hide underlying device implementation details as much as possible. This includes hiding vendor information as much as possible.
2) All concepts that provide a grouping/coupling mechanism between qualitative and quantitative things should be possible to specify as their individual quantitative and qualitative components.

The second point probably deserves a little more explanation. What I mean is that if you are going to expose a concept like a ResourceClass object -- something that allows the deployer to describe a collection of consumed resource amounts as well as capabilities that describe one or more providers of those consumable resources -- then the end user should be able to request a pod consumes those coupled resources and lands on a node with those capabilities without using the coupled object.

In other words, if you have a ResourceClass that looks like this:

```yaml
kind: ResourceClass
metadata:
  name: gpu.high.mem
spec:
  resourceSelector:
    - matchExpressions:
        - key: "gpu.vendor"
          operator: "In"
          values:
            - "nvidia"
            - "intel"
        - key: "gpu.memory"
          operator: "GtEq"
          values:
            - "4G"
```

the end user should be able to request a pod where container in the pod need 1 `gpu.high.mem` OR the end user should be able to request a pod where containers in the pod need 4G of resource type `gpu.memory` and the "gpu.vendor" annotation/selector is "intel" instead of either "intel" OR "nvidia". That's what I mean about breaking the coupling down into its finest-grained representation and allowing the end user to specify that fine-grained request.

Hope that makes sense! I recognize that sometimes, the terminology I use is overlapping and confusing, so I apologize in advance about that. I'm trying to bridge the terminology differences between the OpenStack infrastructure representation of these things with the proposed Kubernetes representation of similar ideas.

Best,
-jay

vikaschoudhary16: @jaypipes 
> The way that Kubernetes advertises resources and capabilities to end users should hide underlying device implementation details as much as possible. This includes hiding vendor information as much as possible.

Thats upto the deployer. Deployer is free to not use any keys in the resource class which he thinks, user should not know about. Proposal does not make it mandatory to create resource classes with any vendor specific details. 

> All concepts that provide a grouping/coupling mechanism between qualitative and quantitative things should be possible to specify as their individual quantitative and qualitative components.

I think i understood what your point is. Problem with this approach is that it will become identity mapping. Resource classes are aimed at creating broader abstractions where arbitrary ranges for resource properties is also able to be supported. There are two main problems:
* Portability gone: With current approach, resource class is an allocatable unit. So though its providing a broader abstraction but still its consumed, capacity and remaining units are countable. By this, admin knows, what is the quota which is being offered. If user is left free to choose the range of resource property, like gpu.memory gt 30, one cluster may support but other may not. But if resource classes are treated as a single allocated unit with non-mutated properties, it is easier for admins to control the availability of it as a resource across clusters.
* We cant expect end user to know that much about the device properties. There will be hell more chances of mis-configuration.

Thoughts?

contributors/design-proposals/resource-class.md
vikaschoudhary16: @dashpole plan is to be recognized all by `quantity`.

contributors/design-proposals/resource-class.md
vikaschoudhary16: @dashpole L# 93:
 >Resource Class object selects devices using matchExpressions, a list of (operator, key, value). A Resource Class object selects a device if at least one of the matchExpressions matches with device details. Within a matchExpression, all the (operator,key,value) are ANDed together to evaluate the result. 

Example resource class could be a smart-nic which either (speed 40G && melanox) OR ( speed 30 && Solarflare). Might not be the best example, but i hope should be enough to give a sense.

contributors/design-proposals/resource-class.md
ScorpioCPH: @vikaschoudhary16 But devices may have the same name (device-id) across nodes:

- Node1 have dev-1 which satisfied ResourceClassA
- Node2 have dev-1 which also satisfied ResourceClassA

How do we cache the devices info in ResourceClassA? use this pattern: `Node1-dev-1` and `Node1-dev-1`?

contributors/design-proposals/resource-class.md
ScorpioCPH: Thanks for your explanation.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+* User can view the current usage/availability details about the resource class
+  using kubectl.
+
+### User story
+The administrator has deployed device plugins to support hardware present in the
+cluster. Device plugins, running on nodes, will update node status indicating
+the presence of this hardware. To offer this hardware to applications deployed
+on kubernetes in a portable way, the administrator creates a number of resource
+classes to represent that hardware. These resource classes will include metadata
+about the devices as selection criteria.
```
ScorpioCPH: Can you give an example how resource classes know which devices are selectable? AFAIK, as mentioned above, the administrator creates a resource class without any devices info or node info, who is responsible to do the mapping (ResourceClass --> Devices) work?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+2. To look for device related details in node objects and maintain accounting for
+   devices as well.
+
+From the events perspective, handling for the following events will be added/updated:
+
+### Resource Class Creation
+1. Initialize and add resource class info into local cache
+2. Iterate over all existing nodes in cache to figure out if there are devices
+   on these nodes which are selectable by resource class. If found, update the
+   resource class availability status in local cache.
```
ScorpioCPH: @vikaschoudhary16 I'm very interested in the details about this? As commented above, how does Resource Class know which devices is selectable?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+   `nvidia-tesla-gpu` is the selected device name and `4` is the quantity requested.
+
+4. Kubelet reads the device request from pod annotation and calls `Allocate` on
+   the matching Device Plugins.
+5. The user deletes the pod or the pod terminates
+6. Kubelet reads pod object annotation for devices consumed and calls `Deallocate`
+   on the matching Device Plugins
+
+In addition to node selection, the scheduler is also responsible for selecting a
+device that matches the resource class requested by the user.
```
ScorpioCPH: I'm wondering how scheduler know which device matches the resource class?

contributors/design-proposals/resource-class.md
vikaschoudhary16: A deviceinfo structure will be maintained in schedulercache. So each device will have a list of what all resource class it is satisfying.
Take a look: https://github.com/vikaschoudhary16/kubernetes/pull/3/files#diff-558cb8bde14dca10a3151bfc222a3aae

contributors/design-proposals/resource-class.md
vikaschoudhary16: A deviceinfo structure will be maintained in schedulercache. So each device will have a list of what all resource class it is satisfying.
Take a look: https://github.com/vikaschoudhary16/kubernetes/pull/3/files#diff-558cb8bde14dca10a3151bfc222a3aae

contributors/design-proposals/resource-class.md
vikaschoudhary16: yes.

contributors/design-proposals/resource-class.md
vikaschoudhary16: Scheduler will watch apiserver for resource class objects creation and then scheduler already has nodeinfo in the cache. For each kind of device, a deviceinfo structure will be instantiated and list of device info will reside in nodeinfo. At each resource class creation, scheduler will iterate over the deviceinfo list and if resourceclass matches the device, deviceinfo's list of resourceclass references is updated.

For more detalis take a look at PoC: https://github.com/vikaschoudhary16/kubernetes/pull/3/files#diff-558cb8bde14dca10a3151bfc222a3aae

contributors/design-proposals/resource-class.md
vikaschoudhary16: Sorry, now that device plugin has got merged, i need to update this proposal for remaining enhancements of device plugin for resource classes.
In the ListWatch() response, device plugin will send device properties also in the arbitrary key-value map. Using this, a Device api object will be created by device manager. Scheduler will keep watching api server for any new Device object creation and will keep synching its device info in its cache.

ScorpioCPH: Thanks, sound like the it depends on device properties which exposed by device plugin. So we can't support any `matchExpressions` that can not be understand by device properties, am i right?

vikaschoudhary16: right. matchExpressions must have a subset of properities exposed by device plugin.

contributors/design-proposals/resource-class.md
vikaschoudhary16: Does above reply answer this question?

ScorpioCPH: Yes, thanks!

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+
+## Abstract
+In this document we will describe *resource classes* which is a new model to
+represent compute resources in Kubernetes. This document should be seen as a
+successor to [device plugin proposal](https://github.com/kubernetes/community/pull/695/files)
+and has a dependency on the same.
+
+## Motivation
+Kubernetes system knows only two resource types, 'CPU' and 'Memory'. Any other
+resource can be requested by pod using opaque-integer-resource(OIR) mechanism.
```
resouer: Maybe we should say ExtendedResource now?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+            - "1G"
+```
+Above resource class will select all the hugepages with size greater than and
+equal to 1 GB.
+
+YAML example 3:
+```yaml
+kind: ResourceClass
+metadata:
+  name: fast.nic
```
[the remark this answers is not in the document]
jeremyeder: This is a key cross-sig deliverable.  We have discussed it at length in the RMWG and it was deferred because we wanted it to be led by sig-network (or more likely the new Network Plumbing WG).

contributors/design-proposals/resource-class.md
vikaschoudhary16: right. As we discussed on slack, will iterate.

contributors/design-proposals/resource-class.md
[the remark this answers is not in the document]
jiayingz: Is it possible that we start with a simple model to support high-performance nic with a combination of CNI plugin and device plugin? The CNI plugin takes care of network interface setup and management, and can be portable across different container orchestration systems. The device plugin can run as a sidecar container and take care of device initialization, health monitoring, and resource advertising. I know having device plugin act like a middleman between CNI and Kubelet may pose certain limitations. We can discuss them and see whether we may solve them by enriching the information passed between device plugin and Kubelet. However, at least for now, I hope the information passed between device plugin and Kubelet can stay at resource level (such as resource name and properties) and device level (such as device runtime configuration). This way, it is clear at the API level that Kubelet is the central place in charge of resource allocation and container runtime setup, which I feel perhaps will be easier to be extended to support future features like cross-resource affinity allocation and etc.

contributors/design-proposals/resource-class.md
dashpole: I see, thanks.

contributors/design-proposals/resource-class.md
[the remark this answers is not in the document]
resouer: I don't think the sig have enough resource to implement this case, I would like to see this design keep as simple as possible and hopefully get merged in 1.10 cycle.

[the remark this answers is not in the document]
fabiand: Ack, thanks for the info.

contributors/design-proposals/resource-class.md
jaypipes: This was actually my point above about separating quantitative and qualitative things. If users were able to ask for an amount of 1 of resource class of `nic`, an amount of 40 (or >= 40) of resource class `nic.speed_bits_per_second` and a separate qualitative string of `nic.offload.vlan`, then device plugins can advertise resource inventories (amounts) separately from capabilities (like offloads). As this proposal stands right now, you will end up having to create quite a few resource classes, with each of the resource classes varying only slightly based on capabilities.

contributors/design-proposals/resource-class.md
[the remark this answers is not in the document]
fabiand: @jiayingz technically it can also be done closer to CNI. To me it just looks like it will be a much cleaner interface to provide the network connectivity via the device manager. This will effectively formalize how components can add additional NICs to a pod. Today this is unspecified and everybody is pretty much using CNI in some way to achieve this.
The problem is that there are so many different assumptions by different projects. Eventually it even implies changing the CNI config of the host which is not necessarily desireable.

OTOH if we could use the device manager, then we know a way how to distribute such a plugin and how it would operate, which could eventually lead to more cooperation, and to less wild west.

contributors/design-proposals/resource-class.md
timothysc: Getting too selective on non-fungible resources can be a dangerous game vs. channeling a lowest common denominator.  

For a decade on grid systems we allow arbitrary matching on any attribute via an expressive configuration language, and it was eventually highly abused by its users to hoard the prized resources.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+4. Kubelet reads the device request from pod annotation and calls `Allocate` on
+   the matching Device Plugins.
+5. The user deletes the pod or the pod terminates
+6. Kubelet reads pod object annotation for devices consumed and calls `Deallocate`
+   on the matching Device Plugins
+
+In addition to node selection, the scheduler is also responsible for selecting a
+device that matches the resource class requested by the user.
+
+### Reason for preferring device selection at the scheduler and not at the kubelet
```
lichuqiang: What if a user requests ER in pods that not handled by scheduler(daemonset pods or static pods)?

resouer:
@vikaschoudhary16 knock knock :) As @timothysc suggested, maybe we can arrange a PoC? Any free bandwidth?

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+1. Get the requested resource class name and quantity from pod spec.
+2. Select nodes by applying predicates according to requested quantity and Resource
+   class's state present in the cache.
+3. On the selected node, select a Device from the stored devices info in cache
+   after matching key,value from requested resource class.
+4. After device selection, update(decrease) 'Requested' for all the resource
+   classes which could select this device in the cache.
+5. Patch the resource class objects with new 'Requested' in the `ResourceClassStatus`.
+6. Add the pod reference in local DeviceToPod mapping structure in the cache.
+7. Patch the pod object with selected device annotation with prefix 'scheduler.alpha.kubernetes.io/resClass'
```
RenaudWasTaken: Would like to know the opinion of some sig-scheduling folks on the best way to achieve that.
Sending a patch request during host selection for a pod doesn't seem to be the right model.

Maybe @bsalamat or @timothysc ?

RenaudWasTaken:
Do you mind adding a pod spec example to your design document? I'm fairly certain Resource classes are supposed to be requested in the resources field but It's probably better to have explicit confirmation :)

After discussing it a bit internally and since this is going to be discussed at the Face 2 Face, we think it might be a good idea to discuss device sharing ideas in this proposal as sharing might have a significant impact on the technical implementation.
And in general if we agree that the sharing block should be implemented with this model, it would be a pretty good argument that this design document is a good step forward.

There are two kinds of device sharing:
- "simple", simple because you only need to express the sharing notion between the containers
- "complex", complex because it requires expressing more than just the fact that devices are shared, it also needs to express an action by the underlying infrastructure for the service to properly be used (mainly the device plugin).
 e.g: for GPUs [MPS](https://docs.nvidia.com/deploy/pdf/CUDA_Multi_Process_Service_Overview.pdf) needs a daemon to run

**simple sharing**

Seem to be something that might be expressed as a construct on top of the ResourceClass API and built into the podSpec exactly like other sharing APIs such as [volumes](https://kubernetes.io/docs/concepts/storage/volumes/).

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cuda-vector-add
spec:
  restartPolicy: OnFailure
  initContainers:
    - name: myInitContainer1
      image: "nvidia/cuda"
      resources:
        limits:
          devices: ["nvidia-gpu"]
  containers:
    - name: myInferenceContainer1
      image: "nvidia/cuda"
      resources:
        limits:
          devices: ["nvidia-gpu"]
    - name: myInferenceContainer2
      image: "nvidia/cuda"
      resources:
        limits:
          devices: ["nvidia-gpu"]
  devices:
    - name: "nvidia-gpu"
      resources:
        nvidia.high.mem: 1
--- 
kind: ResourceClass
metadata:
  name: nvidia.high.mem
spec:
  resourceSelector:
    - matchExpressions:
        - key: "Kind"
          operator: "In"
          values:
            - "nvidia-gpu"
        - key: "memory"
          operator: "GtEq"
          values:
            - "30G"
```

**Simple sharing**

Coud be expressed by adding labels to the `devices` field such as the following (which could be advertised by the device plugin):
```yaml
  devices:
    - name: "nvidia-gpu"
      labels: ["nvidia.com/MPS"]
      resources:
        nvidia.high.mem: 1
```

contributors/design-proposals/resource-class.md
resouer: Sending a patch request directly here is not right.

But we can probably update the annotation of `assumedPod`, and update Pod api object during `bind()`, which is async.

@bsalamat Make sense?

bsalamat: I generally do not like the idea of scheduler knowing about mapping of devices to resource classes. This does not fit well in scheduler's logic and does not seem to be something that scheduler should be aware of.
It is also not clear why scheduler needs to update the pod with the selected device annotation.

contributors/design-proposals/resource-class.md
```diff
@@ -0,0 +1,363 @@
...
+## Objectives
+
+1. Define and add support in the API for a new type, *Resource Class*.
+2. Add support for *Resource Class* in the scheduler.
+
+## Non Objectives
+1. Discovery, advertisement, allocation/deallocation of devices is expected to
+   be addressed by [device plugin proposal](https://github.com/kubernetes/community/pull/695/files)
+
+## Resource Class
```
bsalamat: I skimmed over the doc. I think any resource class proposal should cover both node resources and cluster resources. This proposal does not cover the latter.

ns-sundar:
I have been lurking for a while to see how this shapes up. This proposal seems aimed at only at selecting devices based on metadata. There is certainly a need to model resources in K8s but the current proposal needs to be enhanced in many ways to reflect the variety of
devices, usage models and use cases. 

Let us start by stating what we would want from a resource model:
- Many types of resources cannot be shared: once assigned to a pod/
  container, they cannot be assigned to another until the first one
  releases it (or is preempted). E.g. SR-IOV VF. So, we would want
  a way to track the inventory and usage of resources.
- Some devices such as FPGAs can offer multiple 'regions', which can be
  programmed with different accelerators. If we represent each accelerator
  as a resource, a simpe model of 'devices with resource classes' is
  not enough. We instead need a way to express that a region nested
  inside an FPGA contains an accelerator. This calls for a hierarchy.
- FPGAs (and GPUs) often contain local memory, which is a different kind
  of resource. A user may want, say an ipsec accelerator with 2 GB of
  local memory, both of which need to come from the same device.
  Further, depending on the implementation, some of the memory may be
  dedicated to one region, while others can be shared across regions
  within the device. 

To crystallize the ideas above, consider [this FPGA card](http://www.nallatech.com/store/fpga-accelerated-computing/pcie-accelerator-cards/nallatech-510t-fpga-computing-acceleration-card/) currently in the market [*]. It can be abstracted as below:
```
      dedicated <-> region <-> common <-> region <-> dedicated
        memory        A        memory       B         memory
```
      
How would such a device be represented and handled in this proposal? I think the scope needs to be broadened considerably. 

[*] This is only an example. I am not affiliated with, or own stock in, the company selling this product.

contributors/design-proposals/resource-class.md
resouer: @bsalamat The problem I can see here is: device plugin is not expected to have any scheduling logic (e.g. select devices) by design, and if scheduler also does not want to have it. The only place to select devices will be kubelet. But what this proposal is describing requires awareness of whole picture of all devices info in the cluster to make decision, this is not what one kubelet is capable of (it only has device info of this node). 

I am willing to pushing this design as it do help to enable GPU topology in Kubernetes a lot, and also, fixes blockers for other devices like FPGA. So would be great to know your ideas for scheduling.

XiaoningDing:
@RenaudWasTaken  In your example of complex sharing, do users have to be aware of MPS? Is it possible that device plugin talks to MPS daemon to setup per-container GPU resource limit, without users knowing about MPS?

vikaschoudhary16:
Created a KEP, https://github.com/kubernetes/community/pull/2265,  for this, so closing this one.

PR #786, opened 2017-07-06
Updating FlexVolume doc with capabilities addition.

verult:
The following PR introduces a FlexVolume API change, so the documentation needs to be updated.
https://github.com/kubernetes/kubernetes/pull/47503

### DIFF
--- a/contributors/devel/flexvolume.md
+++ b/contributors/devel/flexvolume.md
@@ -19,8 +19,10 @@ Call-outs are invoked from Controller-manager only when "--enable-controller-att
 ### Driver invocation model:
 
 #### Init:
-Initializes the driver. Called during Kubelet & Controller manager initialization.
-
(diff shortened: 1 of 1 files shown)


contributors/devel/flexvolume.md
```diff
@@ -19,7 +19,8 @@ Call-outs are invoked from Controller-manager only when "--enable-controller-att
 ### Driver invocation model:
 
 #### Init:
-Initializes the driver. Called during Kubelet & Controller manager initialization.
+Initializes the driver. Called during Kubelet & Controller manager initialization. On success, a capabilities map is returned.
+Currently the only capability supported is `attach`, indicating whether the driver requires attach and detach operations. This field is *required*, although for backward-compatibility the default value is set to `true`, i.e. requires attach and detach.
```
jingxu97: Can user set it to false? and what does it mean?

verult: Updated.

PR #788, opened 2017-07-07
update api_changes.md to reflect the latest status of codegen tools and repo structure

caesarxuchao:
cc @MHBauer @pwittrock who updated/reviewed these steps last time.

cc @bsalamat @droot @janetkuo who need to add/change API recently.

I'll update the adding-an-APIGroup.md next. The content of adding-an-APIGroup.md largely overlaps with this doc and is more stale.

Fix https://github.com/kubernetes/kubernetes/issues/49023


contributors/devel/api_changes.md
```diff
@@ -442,43 +447,39 @@ function converting type X in pkg a to type Y in pkg b, should be named:
...
 generator to create it from scratch.
 
-Unsurprisingly, adding manually written conversion also requires you to add
-tests to `pkg/api/<version>/conversion_test.go`.
-
 ## Generate Code 
 
-When any `types.go` change is made, the generated code must be updated
-by running the generators. There are many small generators that
-run. They have some dependencies on each other so they must be run in
```
caesarxuchao: I removed the words on "dependencies" because it's handled by hack/update-codegen.sh"

bsalamat: Is there any reason we don't suggest running "hack/update-all.sh"? If so, you may want to mention the reasons in the doc.

caesarxuchao: hack/update-all.sh runs more scripts than needed here.

caesarxuchao:
https://github.com/kubernetes/kubernetes/pull/48724 could be an example for adding new version to an existing API group. I'll add it to the doc when the it's merged.

contributors/devel/api_changes.md
```diff
@@ -347,21 +347,25 @@ before starting "all the rest".
 
 ### Edit types.go
```
janetkuo: Somewhere in this doc:

If you add a new APi version, add that new version to `KUBE_AVAILABLE_GROUP_VERSIONS` in hack/lib/init.sh

pwittrock:
lgtm, reassigned to @janetkuo since she has some comments

janetkuo:
I tried to follow this in https://github.com/kubernetes/kubernetes/pull/48746. However, some of the details aren't covered by this doc (such as enable version by default, ..., see my first commit in that PR.)

> Is there any reason we don't suggest running "hack/update-all.sh"? If so, you may want to mention the reasons in the doc.

I have the same question. Most of the people who make major API changes need to run hack/update-all.sh anyways. Not mentioning `hack/update-all.sh` doesn't avoid the need to run it, and I end up running more scripts. 

Here's what I did:
```sh
# Followed the doc in this PR...
make clean && make generated_files
hack/update-codegen.sh
# Before submitting my changes, I ran this to avoid verification failure:
hack/update-all.sh
# Surprisingly, verify-all failed, and I had to fix these two:
hack/update-api-reference-docs.sh
hack/update-golint.sh
```

contributors/devel/api_changes.md
MHBauer: It may be handled by the script in the repo, but it is not otherwise documented for those reusing the generators. 

Where is the appropriate place to discuss and document these dependencies?

contributors/devel/api_changes.md
```diff
@@ -442,43 +447,39 @@ function converting type X in pkg a to type Y in pkg b, should be named:
 Also note that you can (and for efficiency reasons should) use auto-generated
 conversion functions when writing your conversion functions.
 
+Unsurprisingly, adding manually written conversion also requires you to add
```
MHBauer: Drop or replace "unsurprisingly".

contributors/devel/api_changes.md
caesarxuchao: Done.

contributors/devel/api_changes.md
caesarxuchao: Done.

caesarxuchao:
@janetkuo update-api-reference-docs is included in [update-all.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/update-all.sh#L71). I think you encountered a bug.

contributors/devel/api_changes.md
caesarxuchao: I think probably in the lister-gen's doc.go, or maybe we can expand the https://github.com/kubernetes/community/blob/master/contributors/devel/generating-clientset.md.

caesarxuchao:
Ok, i'll recommend hack/update-all.sh.

cblecker:
`make update` is preferred over `hack/update-all.sh`.. at the moment one just runs the other, but the makefile is preferred if that functionality was to change in the future.

caesarxuchao:
@cblecker thanks. Updated.

caesarxuchao:
@janetkuo comments addressed, PTAL. Thanks.

I also removed the adding-an-APIGroup.md, as its contents are outdated and are mostly covered in this doc.

caesarxuchao:
Do we document how to update the storage/registry anywhere? I guess that's worth another doc.

MHBauer:
I'm confused by the location work should be done in. `staging/k8s.io/api` vs `pkg/apis`. Instructions do not seem consistent anymore.

caesarxuchao:
@MHBauer I mentioned staging/src/k8s.io/api a few times in the doc. Perhaps I still missed some use cases, could you elaborate on your confusion?

contributors/devel/api_changes.md
```diff
@@ -397,22 +404,22 @@ used.
 ### Edit types.go
 
 Similar to the versioned APIs, the definitions for the internal structs are in
-`pkg/api/types.go`. Edit those files to reflect the change you want to make.
-Keep in mind that the internal structs must be able to express *all* of the
-versioned APIs.
+`pkg/apis/<group>/types.go`. Edit those files to reflect the change you want to
+make.  Keep in mind that the internal structs must be able to express *all* of
```
ironcladlou: Extra space after the period

contributors/devel/api_changes.md
```diff
@@ -397,22 +404,22 @@ used.
...
-opportunities we have to make a great user experience - good error messages and
-thorough validation help ensure that users are giving you what you expect and,
-when they don't, that they know why and how to fix it. Think hard about the
-contents of `string` fields, the bounds of `int` fields and the
+`pkg/apis/<group>/validation/validation.go`. This validation is the one of the
+first opportunities we have to make a great user experience - good error
+messages and thorough validation help ensure that users are giving you what you
+expect and, when they don't, that they know why and how to fix it. Think hard
+about the contents of `string` fields, the bounds of `int` fields and the
 requiredness/optionalness of fields.
```
ironcladlou: Could requiredness/optionalness be replaced by "optionality" (a dictionary term)?

contributors/devel/api_changes.md
```diff
@@ -442,43 +449,39 @@ function converting type X in pkg a to type Y in pkg b, should be named:
...
 ```sh
-hack/update-codegen.sh
+make clean && make generated_files
 ```
 
-As part of the build, kubernetes will also generate code to handle deep copy of
-your versioned api objects. The deep copy code resides with each versioned API:
-   - `<path_to_versioned_api>/zz_generated.deepcopy.go` containing auto-generated copy functions
+`make clean` is important, otherwise the generated files might be stale, because
+the build system uses custom cache.
```
ironcladlou: "`make clean` is necessary to avoid using stale files from the build system's cache"?

contributors/devel/api_changes.md
```diff
@@ -492,10 +495,16 @@ top of the generated file and should be checked with the
 [`repo-infra/verify/verify-boilerplane.sh`](https://github.com/kubernetes/repo-infra/blob/master/verify/verify-boilerplate.sh)
 script at a later stage of the build.
 
+To invoke these generators, you can run [make generated], which runs a bunch of
```
ironcladlou: [make generated] => `make generated`?

contributors/devel/api_changes.md
```diff
@@ -538,30 +565,56 @@ of api objects - this is to improve the overall system performance.
...
-to a bash array.  See [docs on adding an API group](adding-an-APIGroup.md) for
-more.
-
-Adding API groups outside of the `pkg/apis/` directory is not currently
-supported, but is clearly desirable. The deep copy & conversion generators need
-to work by parsing go files instead of by reflection; then they will be easy to
-point at arbitrary directories: see issue [#13775](http://issue.k8s.io/13775).
+If you are adding a new API version to an existing group, you can copy the
+structure of the existing `pkg/apis/<group>/<existing-version>` and
+`staging/src/k8s.io/api/<group>/<existing-version>` folders.
```
ironcladlou: s/folders/directories

contributors/devel/adding-an-APIGroup.md
```diff
@@ -1,117 +0,0 @@
-Adding an API Group
```
janetkuo: You probably should redirect instead of removing this file

contributors/devel/api_changes.md
janetkuo: Did you plan to add a link here?

contributors/devel/api_changes.md
```diff
@@ -347,21 +348,25 @@ before starting "all the rest".
...
 ### Edit types.go
 
-The struct definitions for each API are in `pkg/api/<version>/types.go`. Edit
-those files to reflect the change you want to make. Note that all types and
-non-inline fields in versioned APIs must be preceded by descriptive comments -
-these are used to generate documentation. Comments for types should not contain
-the type name; API documentation is generated from these comments and end-users
-should not be exposed to golang type names.
+The struct definitions for each API are in
+`staging/src/k8s.io/api/<group>/<version>/types.go`. Edit those files to reflect
```
janetkuo: https://github.com/kubernetes/kubernetes/pull/48746 made some changes which I believe require this line to be added to go struct of runtime Objects to generate `DeepCopyObject` functions in `<path_to_versioned_api>/zz_generated.deepcopy.go`:

```go
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```

@sttts

sttts: :+1:

contributors/devel/api_changes.md
```diff
@@ -380,14 +385,16 @@ all in a different order (i.e. you started with the internal structs), then you
...
 
 Note that the conversion machinery doesn't generically handle conversion of
 values, such as various kinds of field references and API constants. [The client
-library](../../pkg/client/restclient/request.go) has custom conversion code for
-field references. You also need to add a call to
-api.Scheme.AddFieldLabelConversionFunc with a mapping function that understands
-supported translations.
+library](https://github.com/kubernetes/client-go/blob/v4.0.0-beta.0/rest/request.go#L352)
+has custom conversion code for field references. You also need to add a call to
+AddFieldLabelConversionFunc of your scheme with a mapping function that
```
sttts: nit: missing quotes around AddFieldLabelConversionFunc

contributors/devel/api_changes.md
sttts: This used to work well without the clean, but recent turned out to be not that reliable. The dependency logic in  `Makefile.generated_files` is non-trivial. So a clean is never wrong if you know you changed something.

contributors/devel/api_changes.md
```diff
@@ -442,43 +449,39 @@ function converting type X in pkg a to type Y in pkg b, should be named:
...
+`make all` will invoke `make generated_files` as well.
+
+The `make generated_files` will also regenerate the `zz_generated.deepcopy.go`,
+`zz_generated.defaults.go`, and `api/openapi-spec/openapi-spec`.
 
 If regeneration is somehow not possible due to compile errors, the easiest
 workaround is to comment out the code causing errors and let the script to
 regenerate it. If the auto-generated conversion methods are not used by the
 manually-written ones, it's fine to just remove the whole file and let the
 generator to create it from scratch.
```
sttts: most of our generators (all? maybe ugorji the last which is not) are now gengo based and do not need that the code compiles. So usually removing the files is fine.

contributors/devel/api_changes.md
```diff
@@ -538,30 +565,56 @@ of api objects - this is to improve the overall system performance.
...
+
+Due to the fast changing nature of the project, the following content is probably out-dated:
+* You can control if the version is enabled by default by update
+[pkg/master/master.go](https://github.com/kubernetes/kubernetes/blob/v1.8.0-alpha.2/pkg/master/master.go#L381).
+* You must add your new version to
+  [pkg/apis/<group>/install/install.go](https://github.com/kubernetes/kubernetes/blob/v1.8.0-alpha.2/pkg/apis/apps/install/install.go).
+* You must add the new version to
+  [hack/lib/init.sh#KUBE_AVAILABLE_GROUP_VERSIONS](https://github.com/kubernetes/kubernetes/blob/v1.8.0-alpha.2/hack/lib/init.sh#L53).
+* You must add the new version  to
+  [cmd/libs/go2idl/go-to-protobuf/protobuf/cmd.go](https://github.com/kubernetes/kubernetes/blob/v1.8.0-alpha.2/cmd/libs/go2idl/go-to-protobuf/protobuf/cmd.go#L64)
```
sttts: this will soon be `staging/src/k8s.io/kube-gen/cmd/go-to-protobuf/protobuf/cmd.go`. Clearly, we don't want to keep the apigroup list in there medium term.

contributors/devel/api_changes.md
```diff
@@ -538,30 +565,56 @@ of api objects - this is to improve the overall system performance.
...
+  to generate protobuf IDL and marshallers.
+* You must add the new version  to
+  [cmd/kube-apiserver/app#apiVersionPriorities](https://github.com/kubernetes/kubernetes/blob/v1.8.0-alpha.2/cmd/kube-apiserver/app/aggregator.go#L172)
+  to let the aggregator list it. This list will be removed before release 1.8.
```
sttts: why removed?

contributors/devel/api_changes.md
sttts: In fact, I would like to see a list + some docs for all of our tags somewhere in this document.

sttts:
Overall, very good doc update 👍

contributors/devel/api_changes.md
caesarxuchao: Added description on `+k8s:deepcopy-gen`.

Only people who add new API group or version need to know the [tags in doc.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/doc.go#L17-L20). In the "adding new API" sections, the instruction is copy the existing group and replace the "group" and "version" strings, that should cover all such tags. I think people won't appreciate long explanations on each of those tags. Also IMHO most tags should be moved to the build system and handled implicitly.

For the tags on the types, i.e., `genclient` and `deepcopy-gen`, the doc explained them.

contributors/devel/api_changes.md
caesarxuchao: Absolutely agree.

contributors/devel/api_changes.md
caesarxuchao: The comment says: "That ripples out every bit as far as you'd expect, so for 1.7 we'll include the list here instead of being built up during storage."

We don't want to keep this hard-coded list forever, right?

contributors/devel/api_changes.md
caesarxuchao: Yes. I verified that `make generated_files` worked even if I removed pkg/api/v1/zz_generated.conversion.go.

contributors/devel/api_changes.md
caesarxuchao: Done

contributors/devel/api_changes.md
caesarxuchao: Done.

contributors/devel/api_changes.md
caesarxuchao: Done.

contributors/devel/api_changes.md
caesarxuchao: Done.

contributors/devel/adding-an-APIGroup.md
caesarxuchao: Done.

contributors/devel/api_changes.md
caesarxuchao: Done.

caesarxuchao:
I have addressed most comments. May I request that we merge the PR asap and address any left issues in a followup PR? I had to send a few developers to this PR in the past days.

contributors/devel/adding-an-APIGroup.md
```diff
@@ -1,115 +1,7 @@
...
-
-1. Add your group in pkg/api/testapi/testapi.go, then you can access the group
-in tests through testapi.`<group>`;
-
-2. Add your "group/version" to `KUBE_TEST_API_VERSIONS` in
-   hack/make-rules/test.sh and hack/make-rules/test-integration.sh
-
-TODO: Add a troubleshooting section.
-
+Please refer to [api_changes.md#making-a-new-api-group].
```
janetkuo: You should do `[text](link)`

contributors/devel/api_changes.md
```diff
@@ -347,21 +348,34 @@ before starting "all the rest".
...
+The struct definitions for each API are in
+`staging/src/k8s.io/api/<group>/<version>/types.go`. Edit those files to reflect
+the change you want to make. Note that all types and non-inline fields in
+versioned APIs must be preceded by descriptive comments - these are used to
+generate documentation. Comments for types should not contain the type name; API
+documentation is generated from these comments and end-users should not be
+exposed to golang type names.
+
+For types that need the generated
+[DeepCopyObject`](https://github.com/kubernetes/kubernetes/commit/8dd0989b395b29b872e1f5e06934721863e4a210#diff-6318847735efb6fae447e7dbf198c8b2R3767)
```
janetkuo: Remove `

contributors/devel/api_changes.md
```diff
@@ -347,21 +348,34 @@ before starting "all the rest".
...
-should not be exposed to golang type names.
+The struct definitions for each API are in
+`staging/src/k8s.io/api/<group>/<version>/types.go`. Edit those files to reflect
+the change you want to make. Note that all types and non-inline fields in
+versioned APIs must be preceded by descriptive comments - these are used to
+generate documentation. Comments for types should not contain the type name; API
+documentation is generated from these comments and end-users should not be
+exposed to golang type names.
+
+For types that need the generated
```
janetkuo: Only runtime.Object needs `DeepCopyObject`

contributors/devel/api_changes.md
```diff
@@ -538,30 +574,58 @@ of api objects - this is to improve the overall system performance.
...
+You need to regenerate the generated code as instructed in the sections above.
+
+## Making a new API Group
+
+You'll have to make a new directory under `pkg/apis/` and
+`staging/src/k8s.io/api`; copy the directory structure of an existing API group,
+e.g. `pkg/apis/authentication` and `staging/src/k8s.io/api/authentication`;
+replace "authentication" with your group name and replace versions with your
+versions; replace the API kinds in register.go
+([versioned](https://github.com/kubernetes/kubernetes/blob/v1.8.0-alpha.2/staging/src/k8s.io/api/authentication/v1/register.go#L47)
```
janetkuo: Remove the first `(`

contributors/devel/api_changes.md
caesarxuchao: Reworded

caesarxuchao:
@janetkuo, thanks for the reviews, all addressed.

caesarxuchao:
Thanks for the reviews. Reviewers, please keep commenting in this PR if there are more issues, i'll address them in followup PRs.

contributors/devel/api_changes.md
sttts: Do we have some todo for that? @deads2k

deads2k: > Do we have some todo for that? @deads2k

created https://github.com/kubernetes/kubernetes/issues/49391

PR #793, opened 2017-07-09
Propose to add federated resource quota functionality

weijinxu:
In a single Kubernetes cluster there is a concept of Resource Quota that manages the resources by namespace so that user usages can be limited to their fair share of the resources. We want to expand this concept to the federation level so that users can treat Federation Control Plane as a “Single Pane of Glass” for seamlessly managing the Resource Quota  information across all the underlying federated clusters.

### DIFF
--- a/contributors/design-proposals/federated-resource-quota.md
+++ b/contributors/design-proposals/federated-resource-quota.md
@@ -1,0 +1,53 @@
+# Federated ResourceQuotaRequirements & Design
+@weijinxu, @IrfanUrRehman, @JianhuiZ
+reviewed by  @quinton-hoole, @deepak-vijSummary
+
+In a single Kubernetes cluster there is a concept of Resource Quota that manages the resources by namespace so that user usages can be limited to their fair share of the resources. We want to expand this concept to the federation level so that users can treat Federation Control Plane as a “Single Pane of Glass” for seamlessly managing the Resource Quota  information across all the underlying federated clusters.
+
+## User scenario
+
+As a federation user and an admin of clusters, I want to be able to define an overall resource quota for some specific or all resources for a given federated namespace. The overall resource quota limit should be applicable across clusters and preferably enforced at federation control plane level.
+
+For example, let us say that we’ve specified 100 pods in the resource quota spec in the federation namespace, with multiple underlying federated clusters. Let us assume that 50 pods have been deployed already across the federation in a previous deployment run. In a subsequent deployment run, if a user requests additional 50+ more pods to be deployed, the deployment will be rejected as a whole by the federation control plane before deployment process actually starts. If user requests less than 50 pods, the deployment should succeed as per the deployment preferences.
+
+## Implementation details
+### Create Local Resource Quota in underlying clusters
+
+A straightforward thought is to divide and propagate the federation quota value to the underlying clusters. This kind of distribution may cause divided quota gets too small share which hardly meet users requests, and may end up in a state that a user may have quite a few resources available but not in the cluster he/she preferred. What makes this worse is that the user wants to deploy a big workload but none of the clusters have enough resource quota value left anymore. This means that we may need to somehow adjust the allocated quota of the underlying clusters accordingly. Similarly, this is also true for clusters joining and leaving the federation.
+Our proposal is that to not split the quota value but instead give every underlying cluster the entire federation quota. When a user creates Federation Resource Quota, Federation Resource Quota controller will propagate the federation resource quota value to each federated cluster. At the same time, the Federation Resource Quota controller then keeps monitoring underlying Local Resource Quota information and updates the Federation Resource Quota status with already used/hard resource information to be subsequently used by the Federation Resource Quota Admission Controller.
+### Additional Quota Controller Functionality
+Above proposal allows users to consume resource at their will freely. But user can go to underlying cluster directly and as each cluster has reserved quota amount same as Federation Resource Quota, they can be consumed more than quota limits if multiple deployments are executed against more than one underlying cluster.
+The proposal is that to give every underlying cluster the federation quota initially, and then ensure every cluster’s quota to `LRQ.Used + FRQ.Remaining` on each reconciliation. Which means that given any time, any cluster has a remaining quota the same as the federation remaining quota. This also simplifies cluster joining (assign it federation remain quota) and cluster leaving (nothing needed). And, when the user bypass the federation and deploy some workload to the underlying clusters, the federation quota controller should update the quotas of other clusters accordingly.
+#### NOTE: This method can prevent over consumption for users accessing underlying local clusters directly, if they only work against local cluster individually. This can’t prevent the over consumption if they deploy to multiple local clusters parallelly.
+### Consider this example:
+1. Having 3 clusters A, B and C
+2. Assigned user (namespace) a quota of 10 pods
+3. Federation quota controller set quota of cluster A to 10, B to 10 and C to 10
+4. User deploys a replicaset of 5 where 3 go to cluster A and 2 go to cluster B; user now has a remaining quota of 5 pods
+5. Federation quota controller set quota of cluster A, B and C to:<br>
+    A: 8 (3+5); B: 7 (2+5);  C: 5 (0+5)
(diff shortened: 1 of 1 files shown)


liggitt:
@deads2k @derekwaynecarr for cluster resource quota experience and feedback

contributors/design-proposals/federated-resource-quota.md
```diff
@@ -0,0 +1,53 @@
...
+
+## User scenario
+
+As a federation user and an admin of clusters, I want to be able to define an overall resource quota for some specific or all resources for a given federated namespace. The overall resource quota limit should be applicable across clusters and preferably enforced at federation control plane level.
+
+For example, let us say that we’ve specified 100 pods in the resource quota spec in the federation namespace, with multiple underlying federated clusters. Let us assume that 50 pods have been deployed already across the federation in a previous deployment run. In a subsequent deployment run, if a user requests additional 50+ more pods to be deployed, the deployment will be rejected as a whole by the federation control plane before deployment process actually starts. If user requests less than 50 pods, the deployment should succeed as per the deployment preferences.
```
derekwaynecarr: quota today is not enforced on the controller replica count, but on the literal pod creation request.

this text implies to me that we will try to enforce quota about pods with knowledge about controller replica count.  i don't think we should do that at this time.  i think have general object count quota for federated api types is fine, but since pods are not in the control plane itself, this seemed problematic.

contributors/design-proposals/federated-resource-quota.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Implementation details
+### Create Local Resource Quota in underlying clusters
```
derekwaynecarr: i would prefer this document separate discussion of quota in two parts:

1. how to handle generic object count quota (i.e. # of secrets, config maps, services, etc.)
2. how to handle compute or storage quota (i.e # of pods, cpu, memory, gpus, storage, etc.)

contributors/design-proposals/federated-resource-quota.md
```diff
@@ -0,0 +1,53 @@
...
+    A: 6 (4+2); C: 4 (2+2); D: 4 (2+2)
+16. User bypasses federation and deploy 1 pods to cluster C
+17. Federation quota controller catches this event and updates quotas to:<br>
+    A: 5 (4+1); C: 4 (3+1); D: 3 (2+1)
+18. User bypasses federation and deploy 1 pod to cluster C, and 1 pod to cluster D simultaneously. Since both clusters have 1 remaining quota which satisfy the request, both pods are deployed successfully.
+19. Federation quota controller catches this event and updates quotas to:<br>
+    A: 4 (4+0); C: 4 (4+0); D: 3 (3+0)
+20. The total used quota count is 11, which is 1 more than the original quota created in Federation Resource Quota object. This is the result of step 18 when user sent deployment request to different clusters parallelly. But since the remaining quota has been trimmed to 0, no more deployment can be executed either from federation level or through local cluster directly.
+
+### Checking admit policy on Federation Control Plane (FCP)
```
derekwaynecarr: i dont think the FCP should become controller replica count aware.

contributors/design-proposals/federated-resource-quota.md
irfanurrehman: @derekwaynecarr thanks for having a look at this.This text actually implies (as later explained in the doc) that the fcp tries to put a check on the object count quota at federation using admission control (just like what a k8s cluster APIserver would do). Actually the example  using pods is misleading, because the pod access as of now is unavailable at fcp. If pods as an api object is exposed from fcp, this example will stand valid. 

@weijinxu I suggest you to update the example with some object for which the api is exposed from fcp (eg replicaset or configmap) and ensure that the quota type as count is explicitly specified somewhere in this text. 

On the other hand it is correct that for the  other controllers which work on replica count (viz deployment, rs), the check cannot happen at admission. As per this design, that check is delegated to federated clusters by creating cluster local resource quota objects (with some reduced limit, which is continually updated based on the quota status observed from all clusters). The cluster local runtime quota enforcement and update of status on those quota objects is done only by local check quota controller(s).  Please let us know if that sounds like a bad idea.

contributors/design-proposals/federated-resource-quota.md
irfanurrehman: Yes makes sense. Actually we did try to put up a lot more detail in the design earlier, but it did seem to become quite long and unreadable. (Somehow our problem that we could not flush out the details in a concise manner).
Actually our intention was to handle this design in two portions indeed, which to me is encompasses ur suggestion.
- quota enforcement at admission in fcp (handles object count as in ur point 1, except the embedded objects, viz pods in rs and rs in deployment).
- quota enforcement run time (as in your point 2 above).

@weijinxu I see that you have a small section at the end (Checking admit policy on Federation Control Plane (FCP)), but probably the rest of the doc is still confusing wrt to the two enforcement mechanisms. I suggest you make these 2 mechanisms in 2 sections explicit and modify this doc and examples accordingly.

contributors/design-proposals/federated-resource-quota.md
irfanurrehman: You are right, and FCP is not aware about the replica counts as per this design proposal, however admission controller makes a lot of sense for the object count check at creation in fcp.

contributors/design-proposals/federated-resource-quota.md
```diff
@@ -0,0 +1,53 @@
...
+Our proposal is that to not split the quota value but instead give every underlying cluster the entire federation quota. When a user creates Federation Resource Quota, Federation Resource Quota controller will propagate the federation resource quota value to each federated cluster. At the same time, the Federation Resource Quota controller then keeps monitoring underlying Local Resource Quota information and updates the Federation Resource Quota status with already used/hard resource information to be subsequently used by the Federation Resource Quota Admission Controller.
```
irfanurrehman: @weijinxu, This detail of this proposal should be explicit with some as follows:

The fcp will create and monitor the ***resource quota objects*** locally into each federated clusters, initially with the same quota value as submitted to fcp. The ***status*** (actual run time usage) of the quota will be updated against these local quota object by the cluster local quota controller(s) in one or more clusters, which fcp will use to update the quota value (spec) against individual quota objects across the federated clusters.

You have used ***resource quota value*** at most of the places, which makes the method of enforcement unclear. Please follow this scheme of usage across the doc including the details in the examples.

irfanurrehman:
> can we separate the discussion in two parts.
> 
> simple object count quota
> distributed quota allocation per cluster
> i worry the requirements for item 2 are not well-defined enough at this time. for example, i could see a policy being defined about how quota is apportioned to each member cluster rather than the propagation model presented here.

@derekwaynecarr We understand your concerns. I have tried to clarify some details. Please feel free to provide more suggestions. Your feedback is really important to us.
@weijinxu Please update the document as per suggestions in the comments.

contributors/design-proposals/federated-resource-quota.md
weijinxu: There is a parallel design which addresses the ability to get pods information. Please see https://github.com/kubernetes/features/issues/71

weijinxu:
Thanks @derekwaynecarr and @irfanurrehman for the comments/suggestions. I will update the doc to reflect them. However I want to point out that maintaining number of pods as one of the distributed resources may be a desirable requirement from customers and there is a parallel project which supporting this: https://github.com/kubernetes/features/issues/71 . Please have a look at that and provide feedback if this is supportable.

irfanurrehman:
> Thanks @derekwaynecarr and @irfanurrehman for the comments/suggestions. I will update the doc to reflect them. However I want to point out that maintaining number of pods as one of the distributed resources may be a desirable requirement from customers and there is a parallel project which supporting this: kubernetes/features#71 . Please have a look at that and provide feedback if this is supportable.

@weijinxu kubernetes/features#71 is only for ***read access*** for all objects from underlying clusters. That is useful mainly to avoid users maintaining direct access to all clusters, separately. I don't think there are any plans for a pod to be made a first class federated object anytime in near future.

weijinxu:
We are still internally investigate a design related issue before publishing the new draft.

PR #794, opened 2017-07-10
Document scalability validation process for release

shyamjvs:
This outlines details of a new/improved scalability validation process for the release.

cc @kubernetes/sig-release-misc  @kubernetes/sig-scalability-misc @dchen1107 @thockin @wojtek-t @gmarek

### DIFF
--- a/contributors/devel/release/scalability-validation.md
+++ b/contributors/devel/release/scalability-validation.md
@@ -1,0 +1,125 @@
+# Scalability validation of the release
+
+Scalability is an integral part of k8s. Numerous issues were identified during release 1.7 while running tests on large clusters (2k-5k nodes). A majority of these manifested only in large clusters - more info under umbrella issue [#47344]. The issues ranged from large cluster setup (both cloud-provider specific and independent ones) to test-infra problems to previously uncaught regressions with performance and resource usage and so on.
+
+We started [supporting] 5000-node clusters from k8s 1.6 and need to make sure this scale is supported in future releases too. We did it this time by running performance and correctness tests under different configurations manually. But this testing procedure needs to be automated, concrete & well-maintained.
+
+## Goals
+
+In this document, we address the following process-related problems wrt scale testing:
+
+- Automate large cluster tests with a reasonable frequency
+- Concretely define the testing configuration used for the release
+- Make scalability validation a release prerequisite
+- Clearly lay down responsibilities for scale tests
+
+## Non-Goals
+
+We do not intend to:
+
+- Define the set of tests that comprise scalability and correctness suite
+- Define SLIs/SLOs (that’s discussed [here]) and thresholds for the tests
+- Discuss particular performance issues we’re facing wrt different releases
+
+While these are interesting from scalability perspective, they are not testing process issues but details of the tests themselves. And can change without requiring much changes in the process.
+
+## Proposed design
+
+For each of the problems above we propose solutions and discuss some caveats. This is an open discussion and better solutions may evolve in future.
+
+### Automate large cluster tests
+
+Currently, there are 2 kinds of tests needed to validate the claim “kubernetes supports X-node clusters”. These are:
+
+- Correctness tests - e2e tests verifying expected system behaviours
+- Performance tests - e2e tests for stress-testing the system perf ([Feature:Performance])
+
+We need to run them on 5k-node clusters, but they’re:
+
+- Time-consuming (12-24 hrs)
+- Expensive (tens of thousands of core hours per run)
+- Blocking other large tests (quota limitations + only one large test project available viz. 'kubernetes-scale')
+
+So we don’t want to run them too frequently. On the other hand, running them too infrequently means late identification and piling up of regressions. So we choose the following middleground:  
+(B = release-blocking, NB = not release-blocking)
+
+- Performance tests on 2k-node cluster in GCE/GKE alternatingly each week (NB)
+  - would catch most scale-related regressions
+  - would catch bugs in both GCE and GKE
+- Performance tests on 5k-node cluster in GCE each week (B)
+  - would catch scale-related bugs missed by 2k-node test (should be rare)
+- Correctness tests on 2k-node cluster in GCE/GKE alternatingly each week (NB)
+  - correctness tests failing on just large clusters is rare, so weekly is enough
+  - would catch bugs in both GCE and GKE
+- Correctness tests on 5k-node cluster in GCE/GKE alternatingly each week (B for GCE)
+  - would catch bugs left out by 2k-node (should be rare)
+  - would verify 5k-node clusters on GKE but without blocking the release
+
+Here's the proposed schedule (may be fine-tuned later based on test health / release schedule):
+
+| | Mon | Tue | Wed | Thu | Fri | Sat | Sun
+| ---- | :----: | :----: | :----:| :----: | :----: | :----: | :----:
+| Week 2k | GCE 5k perf (B) | | | GCE 5k corr (B) | | GKE 2k perf+corr (NB) |
+| Week 2k+1 | GCE 5k perf (B) | | | GKE 5k corr (NB) | | GCE 2k perf+corr (NB) |
+
+Why this schedule?
+
+- 5k tests might need special attention in case of failures so they should run on weekdays
+- Running 5k perf test on Mon gives the whole week for debugging in case of failures
+- Running 5k corr test on Thu gives Tue-Wed to debug any setup issues found on Mon
+- Running 5k corr tests alternatingly on GKE and GCE validates both setups
+- 2k perf and corr tests can be run in parallel to save time
+- Running 2k tests on weekend will free up a weekday for the project to run other tests
+
+Why GKE tests?
+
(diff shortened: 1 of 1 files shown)


shyamjvs:
@smarterclayton Could you also PTAL?

@dchen1107 Could you PTAL specially the part about the impact this has towards the release?

lxpollitt:
cc @caseydavenport

contributors/devel/release/scalability-validation.md
```diff
@@ -0,0 +1,123 @@
...
+- 2k perf and corr tests can be run in parallel to save time
+- Running 2k tests on weekend will free up a weekday for the project to run other tests
+
+Why GKE tests?
+
+Google is currently using a single project for scalability testing, on both GCE and GKE. As a result we need to schedule them together. There's a plan for CNCF becoming responsible for funding k8s testing, and GCE/GKE tests would be separated to different projects when that happens, with only GCE being funded by them. This ensures fairness across all cloud providers.
+
+### Concretely define test configuration
+
+This is a relatively minor issue but it is important that we clearly define the test configuration we use for the release. E.g. there was a confusion this time around testing k8s services, machine-type and no. of the nodes we used (we tested 4k instead of 5k due to a CIDR-setup problem). For ref - [#47344] [#47865]. To solve this, we need to document it under the scalability section of the release-notes in features repo, with the following template:
```
spiffxp: I agree with documenting the configuration used to validate large cluster performance.  I am, however, against release-note bloat.  In other words, I don't think this belongs in CHANGELOG.md.  

Would prefer to see this as a separately linked document, either within kubernetes/community/sig-scalability, or colocated with other release-related docs (eg: kubernetes/features/release-N/scalability-validation-report.md)

contributors/devel/release/scalability-validation.md
```diff
@@ -0,0 +1,123 @@
...
+
+```
+Validated large cluster performance under the following configuration:
+- Cloud-provider - [GCE / GKE / ..]
+- No. of nodes - [5k (desired) / 4k / ..]
+- Node size, OS, disk size/type
+- Master size, OS, disk size/type
+- Any non-default config used - (monitoring with stackdriver, logging with elasticsearch, etc)
+- Any important test details - (services disabled in load test, pods increased in density test, etc)
```
spiffxp: external dependency info, eg:
- container engine used, version
- etcd version used

contributors/devel/release/scalability-validation.md
```diff
@@ -0,0 +1,123 @@
...
+- <job-name, run#> of the validating run (to know other specific details from the logs)
+
+Validated large cluster correctness under the following configuration:
+- <similar to above>
+
+```
+
+### Make scalability validation a release prerequisite
+
+The model we followed this time was to create an umbrella issue ([#47344]) for scalability testing and labelling it as a release-blocker. While it helped block the release, it didn’t receive enough traction from individual SIGs as scale tests were not part of the release-blocking suite. As a result, the onus for fixing issues fell almost entirely on sig-scalability due to time constraints. Thus the obvious requirement here is to make the 5k-node tests release blockers. This along with test automation ensures failures are identified quickly and get traction from relevant SIGs.
```
spiffxp: nit: s/labelling/labeling

contributors/devel/release/scalability-validation.md
```diff
@@ -0,0 +1,123 @@
...
+| | Mon | Tue | Wed | Thu | Fri | Sat | Sun
+| ---- | :----: | :----: | :----:| :----: | :----: | :----: | :----:
+| Week 2k | GCE 5k perf (B) | | | GCE 5k corr (B) | | GKE 2k perf+corr (NB) |
+| Week 2k+1 | GCE 5k perf (B) | | | GKE 5k corr (NB) | | GCE 2k perf+corr (NB) |
+
+Why this schedule?
+
+- 5k tests might need special attention in case of failures so they should run on weekdays
+- Running 5k perf test on Mon gives the whole week for debugging in case of failures
+- Running 5k corr test on Thu gives Tue-Wed to debug any setup issues found on Mon
```
spiffxp: fair, but OTOH this schedule leaves zero days between the 2k/5k-perf runs, which would make me wonder if we'd end up with late-starting monday (or tuesday) runs... guess we'll just find out

contributors/devel/release/scalability-validation.md
spiffxp: in the context of wanting to ensure all release-blocking tests are passing prior to cutting alphas or betas, would we want to maintain a stance of always-5k, or relax to 2k for pre-release builds?

contributors/devel/release/scalability-validation.md
shyamjvs: SG. However it's preferable to only mention non-default configuration used for testing, for brevity. Default configuration is assumed for others (same as in the release), like the ones you've mentioned.

contributors/devel/release/scalability-validation.md
shyamjvs: Makes sense. It's a trade-off between having more days to fix 5k-perf bugs vs. freeing up a weekday by running 2k-node on weekend. We can tune this later as apt.

contributors/devel/release/scalability-validation.md
```diff
@@ -0,0 +1,123 @@
...
+Scalability is an integral part of k8s. Numerous issues were identified during release 1.7 while running tests on large clusters (2k-5k nodes). A majority of these manifested only in large clusters - more info under umbrella issue [#47344]. The issues ranged from large cluster setup (both cloud-provider specific and independent ones) to test-infra problems to previously uncaught regressions with performance and resource usage and so on.
+
+We started [supporting] 5000-node clusters from k8s 1.6 and need to make sure this scale is supported in future releases too. We did it this time by running performance and correctness tests under different configurations manually. But this testing procedure needs to be automated, concrete & well-maintained.
+
+## Goals
+
+In this document, we address the following process-related problems wrt scale testing:
+
+- Automate large cluster tests with a reasonable frequency
+- Concretely define the testing configuration used for the release
```
jeremyeder: This scale out reference architecture has been an area that sig-scale has not been able to agree upon, despite quite a few discussions over the last year.  Will a highly available control plane be part of this reference architecture?

Hmm, seems that your goals and non-goals are in conflict with each other.  How can you concretely define a configuration without defining (or at least knowing) what tests, workloads and SLOs you are subject to?

contributors/devel/release/scalability-validation.md
shyamjvs: Good point! Incorporated your suggestion in the doc.

contributors/devel/release/scalability-validation.md
shyamjvs: > nit: s/labelling/labeling

Done

> would we want to maintain a stance of always-5k, or relax to 2k for pre-release builds?

This is matter of debate and I think the release team (cc @dchen1107 ) might be having a good answer. But IMO we should enforce 5k for alphas and betas too, otherwise we risk ending up in a situation where performance regressions get piled up while we continue adding features. The whole purpose of tying closely dev and testing is that we don't end up catching all bugs towards the end of the release (like this time). Just my 2 cents :)

contributors/devel/release/scalability-validation.md
countspongebob: @jeremyeder Update... we did agree on the scale out reference architecture:
https://github.com/kubernetes/community/blob/master/sig-scalability/provider-configs.md

shyamjvs: > Hmm, seems that your goals and non-goals are in conflict with each other. How can you concretely define a configuration without defining (or at least knowing) what tests, workloads and SLOs you are subject to?

Our workloads, SLOs and other parameters are internals of the e2e tests we run. For defining the process of scale-testing it doesn't matter what exactly those tests are or what they do. All we care is that there is a suite of performance tests defined by "[Feature:Performance]" (what ever that may include and it can change over time) and we just want to ensure it is validated for the release.

> This scale out reference architecture has been an area that sig-scale has not been able to agree upon, despite quite a few discussions over the last year. Will a highly available control plane be part of this reference architecture?

I agree that this is a subject of debate. Different people would want to test different parameters (our current standard is api latency and pod-startup latency SLOs and there's [this discussion](https://docs.google.com/document/d/15rD6XBtKyvXXifkRAsAVFBqEGApQxDRWM3H1bZSBsKQ) to extend/modify it). And like I said above these questions are orthogonal to the process itself and they existed even before :)

shyamjvs: @countspongebob Didn't know we had such a doc. That's great!

wojtek-t: @jeremyeder - yes, HA setup should also be supported here eventually. Though this requires some work (but I really believe then it should be). And yes we already agreed on something during the SIG and the doc @countspongebob is summarizing it.

wojtek-t: > Hmm, seems that your goals and non-goals are in conflict with each other. How can you concretely define a configuration without defining (or at least knowing) what tests, workloads and SLOs you are subject to?

@jeremyeder - I completely don't agree here. We know that we (will) have some set of performance tests we believe are what we are trying to set. And *all* of those are hidden under this "performance tests* term. Defining those tests is unrelated to the effort of saying how important they are (which is more-or-less this doc).

contributors/devel/release/scalability-validation.md
wojtek-t: Good point, but that's some kind of initial proposal. We can tune it later.

spiffxp:
discussed one last time at the sig-scalability meeting, merging

PR #801, opened 2017-07-12
take Phillip Wittrock out of contrib-x sig-leads

pwittrock:


grodrigues3:
can you remove from [here](https://github.com/kubernetes/community/tree/master/sig-contributor-experience) too

cblecker:
Please remove from sigs.yaml and run a `make` instead

pwittrock:
updated PTAL

grodrigues3:
Docs removal was on purpose too right?  If so,

/lgtm

PR #808, opened 2017-07-13
Change the scalability SIG lead from Google to wojtek-t

wojtek-t:


sig-scalability/README.md
```diff
@@ -18,7 +18,7 @@ For more details about our objectives please review our [Scaling And Performance
 Meeting notes and Agenda can be found [here](https://docs.google.com/a/bobsplanet.com/document/d/1hEpf25qifVWztaeZPFmjNiJvPo-5JX1z0LSvvVY5G2g/edit?usp=drive_web).
 
 ## Leads
-* [Daniel Smith](https://github.com/lavalamp), Google
+* [Wojciech Tyczynski](https://github.com/wojtek-t), Google
```
jamiehannaford: Would you mind editing [these lines](https://github.com/kubernetes/community/blob/master/sigs.yaml#L496-L498) instead? All docs are generated from that file using `make gen-docs`.

See [here](https://github.com/kubernetes/community/tree/master/generator) for more info.

wojtek-t: Thanks for the pointer. Done.

PR #809, opened 2017-07-13
Add a proposal for high level volume metrics

gnufied:
This document adds a proposal for gathering metrics
at operation level for volume operations. This ensures
metrics can be captured regardless of individual volume
plugin implementation.

xref https://github.com/kubernetes/features/issues/349

### DIFF
--- a/contributors/design-proposals/volume-metrics.md
+++ b/contributors/design-proposals/volume-metrics.md
@@ -1,0 +1,141 @@
+# Volume operation metrics
+
+## Goal
+
+Capture high level metrics for various volume operations in Kubernetes.
+
+## Motivation
+
+Currently we don't have high level metrics that captures time taken
+and success/failures rates of various volume operations.
+
+This proposal aims to implement capturing of these metrics at a level
+higher than individual volume plugins.
+
+## Implementation
+
+### Metric format and collection
+
+Volume metrics emitted will fall under category of service metrics
+as defined in [Kubernetes Monitoring Architecture](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/monitoring_architecture.md).
+
+
+The metrics will be emitted using [Prometheus format](https://prometheus.io/docs/instrumenting/exposition_formats/) and available for collection
+from `/metrics` HTTP endpoint of kubelet and controller-manager.
+
+
+Any collector which can parse Prometheus metric format should be able to collect
+metrics from these endpoints.
+
+A more detailed description of monitoring pipeline can be found in [Monitoring architecture](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/monitoring_architecture.md#monitoring-pipeline) document.
+
+### Metric Types
+
+Since we are interested in count(or rate) and time it takes to perform certain volume operation - we will use [Histogram](https://prometheus.io/docs/practices/histograms/) type for
+emitting these metrics.
+
+We will be using `HistogramVec` type so as we can attach dimensions at runtime. All
+the volume operation metrics will be named `storage_operation_duration_seconds`.
+Name of operation and volume plugin's name will be emitted as dimensions. If for some reason
+volume plugin's name is not available when operation is performed - label's value can be set
+to `<n/a>`.
+
+
+We are also interested in count of volume operation failures and hence a metric of type `NewCounterVec`
+will be used for keeping track of errors. The error metric will be similarly named `storage_operation_errors_total`.
+
+Following is a sample of metrics (not exhaustive) that will be added by this proposal:
+
+
+```
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "volume_attach" }
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "volume_detach" }
+storage_operation_duration_seconds { volume_plugin = "glusterfs", operation_name = "provision" }
+storage_operation_duration_seconds { volume_plugin = "gce-pd", operation_name = "volume_delete" }
+storage_operation_duration_seconds { volume_plugin = "vsphere", operation_name = "volume_mount" }
+storage_operation_duration_seconds { volume_plugin = "iscsi" , operation_name = "volume_unmount" }
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "mount_device" }
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "unmount_device" }
+storage_operation_duration_seconds { volume_plugin = "cinder" , operation_name = "verify_volume" }
+```
+
+Similarly errors will be named:
+
+```
+storage_operation_errors_total { volume_plugin = "aws-ebs", operation_name = "volume_attach" }
+storage_operation_errors_total { volume_plugin = "aws-ebs", operation_name = "volume_detach" }
+storage_operation_errors_total { volume_plugin = "glusterfs", operation_name = "provision" }
+storage_operation_errors_total { volume_plugin = "gce-pd", operation_name = "volume_delete" }
+storage_operation_errors_total { volume_plugin = "vsphere", operation_name = "volume_mount" }
+storage_operation_errors_total { volume_plugin = "iscsi" , operation_name = "volume_unmount" }
+storage_operation_errors_total { volume_plugin = "aws-ebs", operation_name = "mount_device" }
+storage_operation_errors_total { volume_plugin = "aws-ebs", operation_name = "unmount_device" }
+storage_operation_errors_total { volume_plugin = "cinder" , operation_name = "verify_volume" }
+```
+
+### Implementation Detail
+
+We propose following changes as part of implementation details.
+
+1. All volume operations are executed via `goroutinemap.Run` or `nestedpendingoperations.Run`.
+`Run` function interface of these two types can be changed to include a `operationComplete` callback argument.
+
+   For example:
+
+   ```go
+   // nestedpendingoperations.go
+   Run(v1.UniqueVolumeName, types.UniquePodName, func() error, opComplete func(error)) error
+   // goroutinemap
+   Run(string, func() error, opComplete func(error)) error
+   ```
+
+   This will enable us to know when a volume operation is complete.
+
+2. All `GenXXX` functions in `operation_generator.go` should return plugin name in addition to function and error.
+
+   for example:
+
+   ```go
+   GenerateMountVolumeFunc(waitForAttachTimeout time.Duration,
+       volumeToMount VolumeToMount,
+       actualStateOfWorldMounterUpdater
+       ActualStateOfWorldMounterUpdater, isRemount bool) (func() error, pluginName string, err error)
+   ```
+
+   Similarly `pv_controller.scheduleOperation` will take plugin name as additional parameter:
+
+   ```go
+   func (ctrl *PersistentVolumeController) scheduleOperation(
+           operationName string,
+           pluginName string,
+           operation func() error)
+   ```
+
+3. Above changes will enable us to gather required metrics in `operation_executor` or when scheduling a operation in
+pv controller.
+
+   For example, metrics for time it takes to attach Volume can be captured via:
+
+   ```go
+   func operationExecutorHook(plugin, operationName string) func(error) {
+       requestTime := time.Now()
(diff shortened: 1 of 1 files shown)


gnufied:
cc @saad-ali @childsb @piosz

piosz:
cc @kubernetes/sig-instrumentation-misc

contributors/design-proposals/volume-metrics.md
```diff
@@ -0,0 +1,138 @@
...
+storage_volume_unmount_seconds { plugin = "iscsi" }
+storage_volume_mount_device_seconds { plugin = "aws-ebs"}
+storage_volume_unmount_device_seconds { plugin = "aws-ebs"}
+storage_verify_volume_attach_seconds { plugin = "cinder" }
+```
+
+Similarly errors will be named:
+
+```
+storage_volume_attach_errors { plugin = "aws-ebs" }
```
brancz: The suffix should be `_errors_total`.

fabiand:
What about "pro-active" monitoring, i.e. performing reads on a volume while it is attatched, to identify issues at runtime?

Is this in general interesting, and should this be a spearate issue?

gnufied:
@fabiand can you elaborate more? What sort of metrics we are talking about?

But it does sound like something out of scope for this proposal since this proposal is more about metric collection at controller level.

fabiand:
Ah yes, you can capture the events at the controller level.

I was thinking of i.e. doing active monitoring of attached storage, i.e. general connectivity checks, or read/write/seek latency checks while storage is attached, to ensure that the storage is not malfunctioning.

contributors/design-proposals/volume-metrics.md
```diff
@@ -0,0 +1,138 @@
...
+## Implementation
+
+### Metric format and collection
+
+Volume metrics emitted will fall under category of service metrics
+as defined in [Kubernetes Monitoring Architecture](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/monitoring_architecture.md).
+
+
+The metrics will be emitted using [Prometheus format](https://prometheus.io/docs/instrumenting/exposition_formats/) and available for collection
+from `/metrics` HTTP endpoint of kubelet, controller etc. All Kubernetes core components already emit
```
piosz: Can you please clarify which components will expose those metrics?
Will this be implemented in a common library?

contributors/design-proposals/volume-metrics.md
```diff
@@ -0,0 +1,138 @@
...
+metrics on `/metrics` HTTP endpoint.
+
+
+Any collector which can parse Prometheus metric format should be able to collect
+metrics from these endpoints.
+
+A more detailed description of monitoring pipeline can be found in [Monitoring architecture] (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/monitoring_architecture.md#monitoring-pipeline) document.
```
piosz: Broken link. Remove space between `]` and `(`.

contributors/design-proposals/volume-metrics.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Metric Types
+
+Since we are interested in count(or rate) and time it takes to perform certain volume operation - we will use [Histogram](https://prometheus.io/docs/practices/histograms/) type for
+emitting these metrics.
+
+We will be using `HistogramVec` type so as we can attach dimensions at runtime. Name of operation will become
+part of metric name and at minimum name of volume plugin will be emitted as a dimension.
```
piosz: Why not having operation as label as well? @brancz @fabxc wdyt?

```
minimum name of volume plugin will be emitted as a dimension
```
Sounds weird to me, do you mean that all metrics will be labeled with `plugin`? What is the definition of plugin?

brancz: looking at the examples, maybe `provider` would be clearer than `plugin`?

contributors/design-proposals/volume-metrics.md
gnufied: This would be available from both kubelet and controller-manager depending on component where particular operation was performed.  This isn't implemented as a common library, more like hook into place where volume operations are executed.

contributors/design-proposals/volume-metrics.md
gnufied: > Why not having operation as label as well? @brancz @fabxc wdyt?

Because name of operation is already in the metric name.

Also a volume plugin is typically a third party service that provides actual volume services. Such as - EBS or GCE-PD or Openstack-Cinder etc.  The idea behind labeling the metrics with plugin name is - typically in a cluster user may have more than one volume plugin configured. Using plugin name as a dimension allows them to isolate operation timing from one plugin to another.

brancz: I wasn't questioning including it, rather the naming :slightly_smiling_face: .

gnufied: @brancz yeah I was answering to @piosz's question above. I think naming the label to `provider` is fine.

gnufied: I have renamed `plugin` label to `volume_plugin` - hopefully this would make it clearer.

brancz: `volume_plugin` sounds good to me as well

saad-ali:
+@bowei @msau42

saad-ali:
CC @kubernetes/sig-storage-feature-requests @kubernetes/sig-instrumentation-feature-requests

contributors/design-proposals/volume-metrics.md
```diff
@@ -0,0 +1,138 @@
...
+
+We are also interested in count of volume operation failures and hence a metric of type `NewCounterVec`
+will be used for keeping track of errors.
+
+Following is a list of metrics (not exhaustive) that will be added by this proposal:
+
+
+```
+storage_volume_attach_seconds { volume_plugin = "aws-ebs" }
```
msau42: Can we have operation as a parameter instead of in the metric name?  That's more similar to what we did for cloud provider metrics.

contributors/design-proposals/volume-metrics.md
```diff
@@ -0,0 +1,138 @@
...
+
+   This will enable us to know when a volume operation is complete.
+
+2. All `GenXXX` functions in `operation_generator.go` should return plugin name in addition to function and error.
+
+   for example:
+
+   ```go
+   GenerateMountVolumeFunc(waitForAttachTimeout time.Duration,
+       volumeToMount VolumeToMount,
```
msau42: Can the plugin name be returned in VolumeToMount instead?

contributors/design-proposals/volume-metrics.md
gnufied: What advantage we are looking to gain from doing that? The reason we chose same name in cloudprovider metrics is because - we were interested in metrics such as, how many cloudprovider API calls kubernetes makes per minute in total. That metric is useful because it gives the user a good idea of whether he is within API quota or not. 

Are we looking to do similar aggregation for these metrics? I think not and hence different metric name might be better. Is aggregating say volume_attach and mount_device metric useful in any sense?

msau42: It makes it easier to add the metrics to any querying/display system, especially if we want to add more volume operations in the future.  Then all the consumers of these metrics don't need to update every time.  My understanding of the Prometheus format, is that you can filter by the labels, so you could implement a display by iterating through all the ops, instead of hardcoding every op.

gnufied: right, but moving the operation name to label introduces more cardinality to same metric. It can go either way tbh. There is another advantage of moving operation name to dimension - it makes code maintainence bit easier since each new metric (for an operation) has to be registered separately whereas using label means - we need to register just one metric.

By default if a metric has too many dimensions, some dimensions are elided in dashboards until you apply filters.  lets ask what @brancz and @piosz think on this one.

contributors/design-proposals/volume-metrics.md
gnufied: `VolumeToMount` contains volume spec and at that point plugin name is kind of unknown.  The resolution of which plugin will perform mounting or attaching or some other volume operation - usually happens inside `GenXXX` functions of `operationGenerator` module - https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/operationexecutor/operation_generator.go#L360

So returning plugin that was chosen to perform certain operation avoid having to modify all these internal structs.

contributors/design-proposals/volume-metrics.md
piosz: So please update this in the text.
>from `/metrics` HTTP endpoint of kubelet, controller etc.

suggests that there is more

contributors/design-proposals/volume-metrics.md
piosz: @gnufied I mean having one metric, where operation is specified through label (not as a part of the name). This would for example allow you to see metrics regarding all operations easier (with the current approach you need to sum up across multiple metrics).

cc @loburm

contributors/design-proposals/volume-metrics.md
piosz: This is what I suggested in https://github.com/kubernetes/community/pull/809#discussion_r127915637

brancz: As this is rather controlled information, I think this is generally fine to do. Where it's important to look out for these problems is when the label values can be completely arbitrary, let's say if you gave a request an ID, that would make the time-series created explode and naturally put high time-series churn on any tsdb. If I understand correctly the "operation" is controlled by us implementing said operations, so that generally sounds sane to me.

gnufied: ack, I will move the operation name to metric. Wondering, what will be a good generic name for the metric itself then - "storage_operation_duration_seconds` is what I am thinking.

contributors/design-proposals/volume-metrics.md
gnufied: ack

contributors/design-proposals/volume-metrics.md
brancz: `storage_operation_duration_seconds` sounds good to me

piosz: +1 but I'll let the other sig-storage folks to decide.

gnufied:
@msau42 @brancz @piosz addressed most of the concerns on this document. ptal.

piosz:
/lgtm
please squash commits before merge

piosz:
@bowei @saad-ali do you want to review it?

brancz:
from metric design and Prometheus perspective this looks good to me, but can't comment on the implementation details

contributors/design-proposals/volume-metrics.md
```diff
@@ -0,0 +1,139 @@
...
+```
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "volume_attach" }
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "volume_detach" }
+storage_operation_duration_seconds { volume_plugin = "glusterfs", operation_name = "provision" }
+storage_operation_duration_seconds { volume_plugin = "gce-pd", operation_name = "volume_delete" }
+storage_operation_duration_seconds { volume_plugin = "vsphere", operation_name = "volume_mount" }
+storage_operation_duration_seconds { volume_plugin = "iscsi" , operation_name = "volume_unmount" }
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "mount_device" }
+storage_operation_duration_seconds { volume_plugin = "aws-ebs", operation_name = "unmount_device" }
+storage_operation_duration_seconds { volume_plugin = "cinder" , operation_name = "verify_volume" }
```
wongma7: there may be some troubles implementing metrics for VerifyVolumesAreAttached using the proposed method because it's done on a per-node basis, not per-plugin. Unless the plugin supports bulk verification, then it's fine.

gnufied: if plugin name is not available we can emit the volume_plugin dimension as `<n/a>` if applicable. Usually, people don't mix the plugins so users would know which volume plugin is that, but we will at least have some metrics in that case. I will update the proposal

gnufied:
@saad-ali @bowei please have a look when you get a chance. This design is just waiting for approval from one of #sig-storage members.

bowei:
lgtm

PR #811, opened 2017-07-14
Initial version of scalability thresholds

wojtek-t:
@kubernetes/sig-scalability-pr-reviews @kubernetes/sig-scalability-api-reviews 
@smarterclayton @jeremyeder @timothysc @countspongebob @gmarek @shyamjvs

### DIFF
--- a/sig-scalability/thresholds.md
+++ b/sig-scalability/thresholds.md
@@ -1,0 +1,100 @@
+# Kubernetes Scalability thresholds
+
+## Background
+
+Since 1.6 release Kubernetes officially supports 5000-node clusters. However,
+the question is what that actually means. As of early Q3 2017 we are in the
+process of defining set of performance-related SLIs ([Service Level Indicators])
+and SLOs ([Service Level Objectives]).
+
+However, no matter what SLIs and SLOs we have, there will always be some users
+coming and saying that their cluster is not meeting the SLOs. And in most cases
+it appears that the reason behind is that we (as developers) have silently
+assumed something (e.g. there will be no more than 10000 services in the
+cluster) and users were not aware of that.
+
+This document is trying to explicitly summarize limits for the number of objects
+in the system that we are aware of and state if we will try to relax them in the
+future or not.
+
+## Kubernetes thresholds
+
+We start with explicit definition of quantities and thresholds we assume are
+satisfied in the cluster. This is followed by an explanations for some of those.
+Important notes about the numbers:
+1. In most cases, exceeding these thresholds doesn’t mean that the cluster
+   fails over - it just means that its overall performance degrades.
+1. **Some thresholds below (e.g. total number of all objects, or total number of
+   pods or namespaces) are given for the largest possible cluster. For smaller
+   clusters, the limits are proportionally lower.**
+1. The thresholds obviously differ between different Kubernetes releases
+   (hopefully each of them is non-decreasing). The numbers we present are for
+   the current release (Kubernetes 1.7 release).
+1. There are a lot of factors that influence the thresholds, e.g. etcd version
+   or storage data format. For each of those we assume the default from the
+   release to avoid providing numbers for huge number of combinations of those.
+1. The “Head threshold” is representing the status of Kubernetes head. This
+   column should be snapshotted at every release to produce per-release
+   thresholds (and dedicated column for each release should then be added).
+
+| Quantity                            | Head threshold | 1.8 release | Long term goal |
+|-------------------------------------|----------------|-------------|----------------|
+| Total number of all objects         | 250000         |             | 1000000        |
+| Number of nodes                     | 5000           |             | 5000           |
+| Number of pods                      | 150000         |             | 500000         |
+| Number of pods per node<sup>1</sup> | 100            |             | 100            |
+| Number of pods per core<sup>1</sup> | 10             |             | 10             |
+| Number of namespaces (ns)           | 10000          |             | 100000         |
+| Number of pods per ns               | 15000          |             | 50000          |
+| Number of services                  | 10000          |             | 100000         |
+| Number of all services backends     | TBD            |             | 500000         |
+| Number of backends per service      | 5000           |             | 5000           |
+| Number of deployments per ns        | 20000          |             | 10000          |
+| Number of pods per deployment       | TBD            |             | 10000          |
+| Number of jobs per ns               | TBD            |             | 1000           |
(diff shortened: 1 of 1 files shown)


shyamjvs:
Thanks for the doc. Gives much needed guidance for scale testing concretely.

wojtek-t:
@gmarek @jeremyeder - I moved the google doc version to markdown.

PTAL

sig-scalability/thresholds.md
```diff
@@ -0,0 +1,96 @@
...
+   (hopefully each of them is non-decreasing). The numbers we present are for
+   the current release (Kubernetes 1.7 release).
+1. There are a lot of factors that influence the thresholds, e.g. etcd version
+   or storage data format. For each of those we assume the default from the
+   release to avoid providing numbers for huge number of combinations of those.
+1. The “Current threshold” is representing the status of Kubernetes head. This
+   column should be snapshotted at every release to produce per-release
+   thresholds (and dedicated column for each release should then be added).
+
+| Quantity                            | Current threshold | Long term goal |
```
jeremyeder: The last point above talks about the meaning of "Current".  Why not just create all possible table columns in this initial PR?

Otherwise I immediately wonder what Current means and have to go searching around (in a doc like this, my eyes go straight to the table).

```Quantity  |  Kubernetes 1.7 threshold  | Master threshold | Long term goal``` ?

File: sig-scalability/thresholds.md [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+| Number of pods                      | 150000            | 500000         |
+| Number of pods per node<sup>1</sup> | 100               | 100            |
+| Number of pods per core<sup>1</sup> | 10                | 10             |
+| Number of namespaces (ns)           | 10000             | 100000         |
+| Number of pods per ns               | 15000             | 50000          |
+| Number of services                  | 10000             | 100000         |
+| Number of all services backends     | ?                 | 500000         |
+| Number of backends per service      | 5000              | 5000           |
+| Number of deployments per ns        | 20000             | 10000          |
+| Number of pods per deployment       | ?                 | 10000          |
```
jeremyeder: You will want to very clearly define what the Question marks mean.  Maybe even replace them with "TBD"?

wojtek-t: For me they exactly mean - there is a limit, but we are not really sure what it is (which is exactly the same a TBD for me). But sure - replaced with TBD.

sig-scalability/thresholds.md
wojtek-t: I'm not sure doing it now for 1.7 branch makes sense. But I added an empty column for 1.8 release.

wojtek-t:
@jeremyeder - PTAL

sig-scalability/thresholds.md
```diff
@@ -0,0 +1,96 @@
...
+process of defining set of performance-related SLIs ([Service Level Indicators])
+and SLOs ([Service Level Objectives]).
+
+However, no matter what SLIs and SLOs we have, there will always be some users
+coming and saying that their cluster is not meeting the SLOs. And in most cases
+it appears that the reason behind is that we (as developers) have silently
+assumed something (e.g. there will be no more than 10000 services in the
+cluster) and users were not aware of that.
+
+This document is trying to explicitly summarize all the limitations that we are
```
gmarek: limits for number of objects in the system (there's a 'churn' limit that we don't touch here)

sig-scalability/thresholds.md
```diff
@@ -0,0 +1,96 @@
...
+1. There are a lot of factors that influence the thresholds, e.g. etcd version
+   or storage data format. For each of those we assume the default from the
+   release to avoid providing numbers for huge number of combinations of those.
+1. The “Current threshold” is representing the status of Kubernetes head. This
+   column should be snapshotted at every release to produce per-release
+   thresholds (and dedicated column for each release should then be added).
+
+| Quantity                            | Master threshold | 1.8 release | Long term goal |
+|-------------------------------------|------------------|-------------|----------------|
+| Total number of all objects         | 250000           |             | 1000000        |
```
gmarek: We discussed it briefly at some point - those numbers are for 5k Node cluster. We should explicitly say that they are proportional to number of Nodes/master size/something.

sig-scalability/thresholds.md
```diff
@@ -0,0 +1,96 @@
...
+
+The rationale for some of those numbers:
+1. Total number of objects <br/>
+There is a limitation on the total number of objects on the system, as this
+affects among others etcd and its resource consumption.
+1. Number of nodes <br/>
+We believe that having clusters with more than 5000 nodes is not the best
+option and users should consider splitting into multiple clusters. However,
+we may consider bumping the long term goal at some time in the future.
+1. Number of services and endpoints <br/>
```
gmarek: Are iptables used in all configurations with all networking providers? (I really don't know)

sig-scalability/thresholds.md
wojtek-t: done

sig-scalability/thresholds.md
```diff
@@ -0,0 +1,96 @@
...
+   (hopefully each of them is non-decreasing). The numbers we present are for
+   the current release (Kubernetes 1.7 release).
+1. There are a lot of factors that influence the thresholds, e.g. etcd version
+   or storage data format. For each of those we assume the default from the
+   release to avoid providing numbers for huge number of combinations of those.
+1. The “Current threshold” is representing the status of Kubernetes head. This
+   column should be snapshotted at every release to produce per-release
+   thresholds (and dedicated column for each release should then be added).
+
+| Quantity                            | Master threshold | 1.8 release | Long term goal |
```
wojtek-t: I meant "Master branch" (which is Head). I can change to Head, but that doesn't really matter, I think.

sig-scalability/thresholds.md
wojtek-t: KubeProxy supports also "in-memory proxier, but that's way less efficient and I don't know if anyone is using it.

sig-scalability/thresholds.md
wojtek-t: Added point above describing it (not all numbers are for largest clusters, e.g. #objects / namespace is not really for largest clusters).

jeremyeder:
Thanks for updating it.  Looks good to me now.

wojtek-t:
@jeremyeder - thanks!

@gmarek confirmed offline that he is also fine with it. 

Merging - it's just a starting point, we should iterate on this doc later.

PR #816, opened 2017-07-17
Add missing update staging godeps command

ash2k:
I'm not sure whether it should be before or after `update-staging-client-go.sh` is invoked.

For the record: found while working on https://github.com/kubernetes/kubernetes/pull/48384

### DIFF
--- a/contributors/devel/godep.md
+++ b/contributors/devel/godep.md
@@ -90,6 +90,9 @@ rm -rf vendor
 ./hack/update-bazel.sh
 ./hack/update-godep-licenses.sh
 ./hack/update-staging-client-go.sh
+# If you haven't followed this doc step-by-step and haven't created a dedicated GOPATH,
+# make sure there is no client-go or other staging repo in $GOPATH before running the next command.
+./hack/update-staging-godeps.sh
 git checkout -- $(git status -s | grep "^ D" | awk '{print $2}' | grep ^Godeps)
 ```
 


ericchiang:
cc @caesarxuchao

contributors/devel/godep.md
```diff
@@ -90,6 +90,7 @@ rm -rf vendor
 ./hack/update-bazel.sh
 ./hack/update-godep-licenses.sh
 ./hack/update-staging-client-go.sh
+./hack/update-staging-godeps.sh
```
caesarxuchao: The order is correct. Also it's worth mentioning one needs to make sure before running this script, there should be no client-go or other staging repo in the GOPATH.

ash2k: @caesarxuchao well, I think if someone is following this document then that is always the case? If someone is not precisely following the document then there are probably lots of things that can go wrong and I don't think we have to document each of them? WDYT? I can certainly add a comment, just don't see the point.

cblecker: Many people make assumptions about what should work, or what works for them in other environments. If we have explicit requirements, it's never a bad idea to call them out in comments if they are a common tripping point.

+1 to adding a comment.

ash2k: Done!

caesarxuchao: Ah, i didn't read the entire doc. As you said, if one follows the doc and creates a dedicated GOPATH, then it's not a problem.

Maybe the comment could say "If you haven't followed this doc step-by-step and haven't created a dedicated GOPATH, make sure there is no client-go or other staging repo in $GOPATH before running the next command."

caesarxuchao: Could update the comment again? Sorry for the trouble.

caesarxuchao:
And please squash.

ash2k:
Done

caesarxuchao:
/lgtm

Thanks.

PR #821, opened 2017-07-19
Add Patch Release Manager Playbook.

enisoc:


cblecker:
Out of curiosity: is there any part of the branch manager's role/tools that still requires this person to be a googler? If not, 🎉 ! If so, would it be worth documenting what those are?

enisoc:
@cblecker Yes, it currently still needs to be a Googler. The mechanical reasons (assumptions in tooling) are being tracked in https://github.com/kubernetes/release/issues/268 and should be straightforward to overcome.

However, the main blocker is controlling access to push bits to the official release buckets. As it is, raw access to those buckets is already too broad (all Goobernauts). Rather than widen access to include non-Googler release managers, we want to restrict it to an automated build system. Then we would only give release managers access to command the build system to push what it built. This is being tracked in https://github.com/kubernetes/release/issues/31.

PR #825, opened 2017-07-19
Add kubeadm upgrades proposal

luxas:
This proposal has so far been developed in this Google doc: https://docs.google.com/document/d/1PRrC2tvB-p7sotIA5rnHy5WAOGdJJOIXPPv23hUFGrY/edit

Features issue: https://github.com/kubernetes/features/issues/296

@kubernetes/sig-cluster-lifecycle-proposals  
@kubernetes/sig-onprem-proposals 
for general review/approval

@kubernetes/sig-architecture-proposals 
for review of Kubernetes upgrades with no external dependencies.
The aim is to provide an easy way to do upgrades against any cluster given some basic requirements long-term

@kubernetes/sig-api-machinery-proposals a heads up on the expected retry loop while trying to bind to a port

@kubernetes/sig-apps-proposals for being able to upgrade DaemonSets using an "add first, then delete" strategy

@kubernetes/sig-node-proposals for the expected checkpointing functionality

I'm not sure if the markdown converter I used could preserve styling well, if not, I'll update that in the coming days.

### DIFF
--- a/contributors/design-proposals/kubeadm-upgrades.md
+++ b/contributors/design-proposals/kubeadm-upgrades.md
@@ -1,0 +1,452 @@
+kubeadm upgrades proposal
+
+Lucas Käldström, Luke Marsden & the SIG Cluster Lifecycle team
+July 2017
+
+See also: [sig-cluster-lifecycle 1.7 Planning](https://docs.google.com/document/d/1xpjz9373KQ56gr7812eo85wB7G6n39LtUIn9Lq74t0Q/edit) and [sig-cluster-lifecycle 1.8 Planning](https://docs.google.com/document/d/17C2D1Ghgv40pWByevo0ih8T-18Q1GlxbR4gu1_ZKlPE/edit).
+
+# Abstract
+
+This proposal describes how kubeadm will support upgrades via self-hosting, and includes enough design for HA to ensure that we're not painting ourselves into an architectural corner.
+
+This document is somewhat based on the [notes](https://docs.google.com/document/d/19GzaaqqW7Q6IdfJa6gYy2JsCgnmChTmol4V0kHI50pc/edit) from the face-to-face SIG-cluster-lifecycle meeting during KubeCon 2017 in Berlin and the "self-hosting, upgrades & HA"-special effort SIG Cluster Lifecycle meetings late June and early July 2017.
+
+Timescale for implementation of this proposal (working self-hosting & upgrades) is Kubernetes 1.8. This means that [features issues must be in by Aug 1](https://github.com/kubernetes/features/pull/305) and [code must be in by Sept 1](https://github.com/kubernetes/features/pull/305).
+
+# Self-hosting
+
+Self-hosting will be the default way of deploying the control plane, but it will be optional. There will be a `kubeadm init`-time feature gate to disable it.
+
+Only *control plane components* will be in-scope for self-hosting for v1.8.
+
+Notably, not etcd or kubelet for the moment.
+
+Self-hosting will be a [phase](https://github.com/kubernetes/kubeadm/blob/master/docs/design/design.md) in the kubeadm perspective. As there are Static Pods on disk from earlier in the `kubeadm init` process, kubeadm can pretty easily parse that file and extract the PodSpec from the file. This PodSpec will be modified a little to fit the self-hosting purposes and be injected into a DaemonSet which will be created for all control plane components (API Server, Scheduler and Controller Manager). kubeadm will wait for the self-hosted control plane Pods to be running and then destroy the Static Pod-hosted control plane.
+
+What actually happens in the Static Pod -> Self-hosted transition?
+
+First the PodSpec is modified. For instance, instead of hard-coding an IP for the API server to listen on it is dynamically fetched from the Downward API. When it’s ok to proceed, the self-hosted control plane DaemonSets are posted to the Static Pod API Server. The Static Pod API Server starts the self-hosted Pods normally. The Pods get in the Running state, but fails to bind to the port on the host and backs off internally. During that time, kubeadm notices the Running state of the self-hosted Pod, and deletes the Static Pod file which leads to that the kubelet stops the Static Pod-hosted component immediately. On the next run, the self-hosted component will try to bind to the port, which now is free, and will succeed. A self-hosted control plane has been created!
+
+Status as of v1.7: self hosted config file option exists; not split out into phase. The code is fragile and uses Deployments for the scheduler and controller-manager, which unfortunately [leads to a deadlock](https://github.com/kubernetes/kubernetes/issues/45717) at `kubeadm init` time. It does not leverage checkpointing so the entire cluster burns to the ground if you reboot your computer(s). Long story short; self-hosting is not production-ready in 1.7.
+
+The rest of the document assumes a self-hosted cluster. It will not be possible to use 'kubeadm upgrade' a cluster unless it's self-hosted. The sig-cluster-lifecycle team will still provide some manual, documented steps that an user can follow to upgrade a non-self-hosted cluster.
+
+Self-hosting implementation for v1.8
+
+The API Server, controller manager and scheduler will all be deployed in DaemonSets with node affinity (using `nodeSelector`) to masters. The current way DaemonSet upgrades currently operate is "remove an old Pod, then add a new Pod". Since supporting “single masters” is a definite requirement, we have to either workaround removing the only replica of the scheduler/controller-manager by duplicating the existing DaemonSets (e.g. a `temp-self-hosted-kube-apiserver` DS will be created as a copy of the normal `self-hosted-kube-apiserver` DS during the upgrade) or ask for a [new upgrade strategy](https://github.com/kubernetes/kubernetes/issues/48841) (“add first, then delete”, which is generally useful) from sig-apps.
+
+Checkpointing
+
+In order to be able to reboot a self-hosted cluster (e.g. a single, self-hosted master), there has to be some kind of checkpointing mechanism. Basically the kubelet will write some state to disk for Pods that opt-in to checkpointing ("I’m running an API Server, let’s write that down so I remember it", etc.). Then if the kubelet reboots (in the single-master case for example), it will check the state store for Pods it was running before the reboot. It discovers that it ran an API Server, Controller Manager and Scheduler and starts those Pods now as well.
+
+This solves the chicken-and-egg problem that otherwise would occur when the kubelet comes back up, where the kubelet tries to connect to the API server, but the API server hasn’t been started yet, it should be running as a Pod on that kubelet, although it isn’t aware of that.
+
+We’re aiming to build this functionality into the kubelet behind an alpha feature flag.
+
+# Upgrading policies
+
+Defining decent upgrading and version skew policies is important before implementing.
+
+Definitions
+
+Minor version upgrade = This is an upgrade from a version (vX.Y.Z1) with the minor component equal to `Y` to a version (vX.Y+1.Z2) with the minor component of `Y+1`. For instance, a minor version upgrade can look like `v1.8.3` to `v1.9.0`.
+
+Patch version upgrade = This is an upgrade from a version (vX.Y.Z) with the patch component equal to `Z` to a version (vX.Y.Z+n) with the patch component of `Z+n`. `n` can be freely chosen by the user. For instance, a patch version upgrade can look like `v1.8.3` to `v1.8.6`.
+
+1. The user must upgrade the kubeadm CLI before the control plane
+
+    1. Upgrading to a higher patch release than the kubeadm CLI version will
+*be possible*, but yield a warning and require the `--force` flag to `kubeadm upgrade apply` to continue
+
+    2. Upgrading to a higher minor release than the kubeadm CLI version will **not** be supported.
+
+    3. Example: kubeadm v1.8.3 can upgrade your v1.8.3 cluster to v1.8.6 if you specify `--force` at the time of the upgrade, but kubeadm can never upgrade your v1.8.3 cluster to v1.9.0
+
+2. The kubeadm CLI can create clusters with the latest two minor releases.
+
+    4. Example: kubeadm v1.8.2 supports *creating a new cluster* for Kubernetes versions v1.7.x and v1.8.x
+
+3. The control plane must be upgraded before the kubelets in the cluster
+
+    5. For kubeadm; the maximum amount of skew between the control plane and the kubelets is *one minor release*.
+
+    6. Example: running `kubeadm upgrade apply --version v1.9.0` against a v1.8.2 control plane will error out if the nodes are still on v1.7.x
+
+4. This means that there are possibly two kinds of upgrades kubeadm can do:
+
+    7. A patch release upgrade (from v1.8.1 to v1.8.3 for instance)
+
+        1. This is very much about bumping the version number. It is very uncommon that things change in a breaking manner between these versions
+
+    8. A minor release upgrade (from v1.7.5 to v1.8.3 for instance)
+
+        2. This kind of upgrade is much harder as lots of things tend to change with every minor version. Common minor release upgrade (from v1.7.5 to v1.8.3 for instance) tasks kubeadm must handle:
+
+            1. Old/deprecated flags that should be removed
+
+                1. Example: the --insecure-experimental-approve-all-node-csrs flag for the controller-manager
+
+            2. Arguments that have to be changed
+
+                2. Example: the list of admission controllers grew from v1.6 to v1.7; two new admission controllers (Initializers and NodeRestriction) had to be inserted in the right place of the chain
+
+            3. New flags that should be added
+
+                3. In order to enable new and required features, kubeadm should automatically append these config options
+
+            4. Resources that have to be modified
+
+                4. Example: The `system:nodes` ClusterRoleBinding had to lose its binding to the `system:nodes` Group when upgrading to v1.7; otherwise the Node Authorizer wouldn’t have had any effect.
+
+5. Using kubeadm, you must upgrade the control plane atomically.
+
+    9. It’s out of scope for now to upgrade only a part of the control plane. All  control plane components must be upgraded at the same time for it to be a supported operation from kubeadm’s point of view.
+
+    10. If you upgrade your cluster, the apiserver, controller-manager, scheduler, proxy and dns will get upgraded at the same time, and necessary resources (like RBAC rules) are modified in-cluster.
+
+        3. Kube-dns will be upgraded to the latest available, branched, manifest inside of the kubeadm source code.
+
+6. `kubeadm upgrade plan` will fetch the latest stable versions from `dl.k8s.io/release/stable-1.X.txt` where X is the minor version for the given release branch.
+
+    11. If the user can’t connect to the internet, `kubeadm upgrade plan` won’t show the user which versions he/she can upgrade to, but `kubeadm upgrade apply --version` will work just normally. 
+
+    12. The only thing that changes is that the user has to know which version to upgrade to, but that is a good requirement, since the user is the one that knows which images has been pre-pulled anyway.
+
+7. Upgrading etcd might or might not be supported, depending on the complexity of the upgrade. 
+
+    13. Major version bumps with schema changes require a proposal on its own
+
+    14. If the user opts-in to it, and kubeadm knows that a higher Kubernetes version than what the user have now requires or recommends a minor/patch version bump of etcd, kubeadm may be able to upgrade etcd via a Job (see the related section below)
+
+8. An upgrade *could* imply other things than a version number bump, for example 
+
+# The `kubeadm upgrade` command
+
+There will exist a `kubeadm upgrade` command that takes care of all the hard parts involved in upgrading a Kubernetes cluster.
+
+It will have two subcommands:
+
+* `plan`: Usually the first command the user will run (the recommended).
+
+    * Validates that the cluster is upgradable
+
+    * Outputs available versions to upgrade to (if any)
+
+    * In an air-gapped environment, it won’t show the available versions to upgrade to.
+
+* `apply`: Will do the actual upgrade
+
+    * `--dry-run`: Computes the changes/actions that would have to be taken and outputs the manifests that will be applied, resources that will be modified, etc.
+
+UX when upgrading a patch version:
+
+$ kubeadm upgrade plan
+
+Inspecting current cluster state:
+
+ -> Self-hosting active ✓
+
+ -> Configuration exists in-cluster ✓
+
+ --> Validating configuration ✓
+
+ -> Fetching cluster versions ✓
+
+ --> Control plane version: v1.8.1
+
+ --> Latest stable patch version: v1.8.3
+
+ --> Latest stable minor version: v1.8.3
+
+ --> Kubeadm CLI version: v1.8.3
+
+ ---> Good! Kubeadm is up-to-date!
+
+ Upgrade to the latest patch version of your release branch:
+
+			Current version	Upgrade available
+
+ API server:		v1.8.1		v1.8.3
+
+ Controller-manager:	v1.8.1		v1.8.3
+
+ Scheduler:		v1.8.1		v1.8.3
+
+ Kube-proxy:		v1.8.1		v1.8.3
+
+ Kube-dns:			v1.15.4		v1.15.5
+
+ Etcd				v3.1.10		v3.1.10
+
+ Upgrades available!
+
+ You can now apply the upgrade by executing the following command:
+
+	kubeadm upgrade apply --version v1.8.3
+
+ Components that must be upgraded manually:
+
+ Kubelet			2 x v1.8.0		v1.8.3
+
+				3 x v1.8.1		v1.8.3
+
+To upgrade the kubelets in the cluster, please upgrade them from the source you installed them from (most often using your package manager).
+
+$ kubeadm upgrade apply --version v1.8.3
+
+Pulling the required images... ✓
+
+* gcr.io/google_containers/kube-apiserver: 2/2 pulled
+
+* gcr.io/google_containers/kube-controller-manager: 2/2 pulled
+
+* gcr.io/google_containers/kube-scheduler: 2/2 pulled
+
+* gcr.io/google_containers/kube-proxy: 5/5 pulled
+
+Upgrading the API Server DaemonSet... 
+
+* Applying the new version of the DaemonSet... ✓
+
+* Verifying that it comes up cleanly... ✓
+
+Upgrading the Controller-manager DaemonSet...
+
+* Applying the new version of the DaemonSet... ✓
+
+* Verifying that it comes up cleanly... ✓
+
+Upgrading the Scheduler DaemonSet...
+
+* Applying the new version of the DaemonSet... ✓
+
+* Verifying that it comes up cleanly... ✓
+
+Upgrading the kube-proxy DaemonSet... ✓
+
+Upgrading the kube-dns Deployment... ✓
+
+Success! Your cluster was upgraded to v1.8.3
+
+UX when upgrading a minor version:
+
+$ kubeadm upgrade plan
+
+Inspecting current cluster state:
+
+ -> Self-hosting active ✓
+
+ -> Configuration exists in-cluster ✓
+
+ --> Validating configuration ✓
+
+ -> Fetching cluster versions ✓
+
+ --> Control plane version: v1.7.5
+
+ --> Latest stable patch version: v1.7.8
+
+ --> Latest stable minor version: v1.8.3
+
+ --> Kubeadm CLI version: v1.8.1
+
+ ---> Note! The kubeadm CLI should be upgraded
+      to v1.8.3 if you want to upgrade your
+      Cluster to v1.8.3
+
+ Both minor and patch release upgrades available!
+
+ Upgrade to the latest patch version of your branch:
+
+			Current version	Upgrade available
+
+ API server:		v1.7.5		v1.7.8
+
+ Controller-manager:	v1.7.5		v1.7.8
+
+ Scheduler:		v1.7.5		v1.7.8
+
+ Kube-proxy:		v1.7.5		v1.7.8
+
+ Kube-dns:			v1.14.2		v1.14.4
+
+ Etcd:			v3.0.17		v3.1.10
+
+ You can now apply the upgrade by executing the following command:
+
+	kubeadm upgrade apply --version v1.7.8
+
+Upgrade to the latest stable minor release:
+
+			Current version	Upgrade available
+
+ API server:		v1.7.5		v1.8.3
+
+ Controller-manager:	v1.7.5		v1.8.3
+
+ Scheduler:		v1.7.5		v1.8.3
+
+ Kube-proxy:		v1.7.5		v1.8.3
+
+ Kube-dns:			v1.14.2		v1.15.5
+
+ You can now apply the upgrade by executing the following command:
+
+	kubeadm upgrade apply --version v1.8.3
+
+ Components that must be upgraded manually:
+
+ Kubelet			2 x v1.8.0		v1.8.3
+
+				3 x v1.8.1		v1.8.3
+
+# Implementation options of the upgrade
+
+High-level approaches:
+
+1. Build the upgrade logic into kubeadm
+
+2. Build an upgrading Operator that does the upgrading for us
+
+    * Would consume a TPR with details how to do the upgrade
+
+The historical preference was to build a reusable Operator that does the upgrading for us. *Either way we would deliver the same 'kubeadm upgrade' UX.* In the latter case, that would involve kubeadm deploying an upgrading controller on behalf of the user, and tracking its lifecycle through the kubernetes API.
+
+There are pros and cons for each method:
+
+1. Build it into the `kubeadm` CLI command
+
+    1. Pros:
+
+        1. *Simplicity*; Very straightforward to implement; basically doing the same steps as `kubeadm init` would, only slightly modified to fit the upgrading purpose. We will construct a `kubeadm upgrade` command in any case
+
+        2. The kubeadm v1.6 -> v1.7 upgrade used this method this successfully
+
+        3. Being able to easily have an interactive flow where the user can inspect the changes that are about to be applied and accept/deny for instance.
+
+        4. Doesn’t have to deal with any version skew between the Operator image tag/version and the CLI version
+
+        5. Doesn’t have to define an API (CRD or the like) between the client and the Operator
+
+    2. Cons:
+
+        6. Maybe not as reusable as an Operator inside the cluster would be
+
+2. Build an Operator
+
+    3. Pros:
+
+        7. The idea is that it would be generic and could be re-used by others
+
+        8. Other operators are already handling complex upgrades successfully, and we could even standardise patterns
+
+    4. Cons:
+
+        9. Has to deal with any version skew between the controller and the CLI
+
+        10. Has to deal with pre-pulling the image to a specific node in the case of no internet connection
+
+        11. Has to define an API (CRD or the like) between the client and the Operator
+
+**Decision**: Keep the logic inside of the kubeadm CLI (option 1) for the implementation in v1.8.0.
+
+We *might* build an upgrade operator in future kubeadm versions.
+
+Upgrade a Static Pod-hosted cluster to a self-hosted one
+
+The first thing `kubeadm upgrade` would have to do when upgrading from v1.7 to v1.8 would be to make the static-pod-hosted v1.7 cluster self-hosted. This can easily be done thanks to the recent refactoring of self-hosting into an atomically invoked phase. 
+`kubeadm upgrade plan` will inform the user about the Static Pod => self-hosted switch in the case of an v1.7.x to v1.8.x upgrade. It will also be possible to convert a Static Pod-hosted control plane to a self-hosted one with another command like `kubeadm phase selfhosting apply`.
+
+Keeping user customizations between releases
+
+One of the hardest parts with implementing the upgrade will be to respect customizations made by the user at `kubeadm init` time. The proposed solution would be to store the kubeadm configuration given at `init`-time in the API as a ConfigMap, and then retrieve that configuration at upgrade time, parse it using the API machinery and use it for the upgrade.
+
+This highlights a very important point: **We have to get the kubeadm configuration API group to Beta (v1beta1) in time for v1.8.**
+
+By declaring Beta status on kubeadm’s configuration, we can be sure that kubeadm CLI v1.9 can read it and be able to upgrade from v1.8 to v1.9.
+
+### Alternative considered
+
+Instead of generating new manifests from a versioned configuration object, we could try to add "filters" to the existing manifests and apply different filters depending on what the upgrade looks like. This approach, to modify existing manifests in various ways depending on the version bump has some pros (modularity, simple mutating functions), but the matrix of functions and different policies would grow just too big for this kind of system, so we voted against this alternative in favor for the solution above.
+
+"Upgrading" without bumping the version number
+
+One interesting use-case for `kubeadm upgrade` will be to modify other configuration in the cluster. If we assume the configuration file options from `kubeadm init` will be fetchable from a ConfigMap inside of the cluster at any time, it would be possible to fetch the current state, compare it to the desired state (from `kubeadm apply --config`) and build a patch of it. Kubeadm would validate the fields changed in that patch and allow/deny "upgrading" those values.
+
+This is a stretch goal for v1.8, but not a strictly necessary feature.
+
+Pre-pulling of images/ensuring the upgrade doesn’t take too long
+
+One tricky part of the upgrade flow is that it will require new images to be run on a given node; and that image might not exist locally. When upgrading, the kubelet will start to pull the image (which is fine), but it might take a while. We don’t want the scenario of a poor internet connection when everything is proceeding fine but slow and `kubeadm upgrade` times out before the image is pulled.
+
+We plan to solve this by having a separate `prepull` task in the beginning of the upgrade to minimize service outage. What it will basically do is to create a DaemonSet for all the control plane components, but instead of running the component, it executes `sleep 3600` or a similar command.
+
+Upgrading etcd
+
+In v1.7 and v1.8, etcd runs in a Static Pod on the master. In v1.7, the default etcd version for Kubernetes is v3.0.17 and in k8s v1.8 the recommended version will be something like v3.1.10. In the v1.7->v1.8 upgrade path, we could offer upgrading etcd as well as an opt-in*. *This only applies to minor versions and 100% backwards-compatible upgrades like v3.0.x->v3.1.y, not backwards-incompatible upgrades like etcdv2 -> etcdv3.
+
+The easiest way of achieving an etcd upgrade is probably to create a Job (with `.replicas=<masters>`, Pod Anti-Affinity and `.spec.parallellism=1`) of some kind that would upgrade the etcd Static Pod manifest on disk by writing a new manifest, waiting for it to restart cleanly or rollback, etc.
+
+Verbose upgrading mode
+
+Currently it is not possible to specify a `--v` flag to `kubeadm init`. In time for v1.8 that should be possible, as it is needed for `kubeadm upgrade` as well. Upgrading a cluster is a task that has to be handled carefully, and often the user wants to know what happens under the hood in order to trust the component that automates that work away.
+
+Network Plugins
+
+As CNI network providers are crucial to the health of the cluster, we have to somehow ensure that the network plugin continues to work after a minor release upgrade.
+
+Since we can’t assume that the same network provider manifest that worked in v1.7 will continue to work in v1.8 we might have to treat it in a special manner.
+
+The "real" solution to this problem would be to implement an “[Addon Manager](https://github.com/kubernetes/features/issues/18)” in core Kubernetes as an API Group. Then CNI providers would just hook up themselves to that API and the addon manager would take care of using the right manifest for the right version.
+
+One proposed method of dealing with this is to have a `--cni-network-manifests` flag to `kubeadm upgrade apply`. The flag would be optional for patch upgrades but semi-required with minor upgrades (skippable with `--force` or similar)
+
+Interactive vs non-interactive mode
+
+By default, kubeadm should tell the user what it is doing in a way that is verbose enough to satisfy the most users (but doesn’t print out a wall of text). Also see the verbosity section.
+
+`kubeadm upgrade apply` implementation flow
+
+1. Validate the version given and configuration. Fail fast if the configuration is missing or invalid. Fail fast if the user is trying to upgrade to a version that is not supported.
+
+2. Make sure the cluster is healthy
+
+    1. Make sure the API Server’s `/healthz` endpoint returns `ok`
+
+    2. Makes sure all Nodes return `Ready` status
+
+    3. Makes sure all desired self-hosting Pods for the control plane are Running and Ready
+
+3. (Upgrade the cluster to use self-hosting if static pods are used currently)
+
+    4. Might require some extra flags, etc
+
+4. Parse the control plane arguments into `map[string]string`, traverse them and create a new alternative.
+
+5. Upgrade the API Server, Controller-manager and Scheduler in turn
+
+    5. This will result in a very short API Server outage (will post here with information about how many milliseconds) if you have a single master (which is the case with v1.7)
+
+    6. If any of them don’t come up cleanly, kubeadm will rollback.
+
+6. Modify resources in the API as needed (e.g. RBAC rules)
+
+7. Upgrade kube-proxy and kube-dns
+
+8. Apply the network manifest if specified
+
+9. Update the ConfigMap with the new config file that was used
+
+Possible failures, timeouts/deadlocks and rollback policy
+
+By default, kubeadm should have a (configurable) timeout value; if a given task takes longer than; kubeadm will automatically try to rollback.
+
+# HA design
+
+Note: HA being fully implemented and on by default for v1.8 is a *non-goal*.
+
+Design docs are available:
+
+* [Kubeadm (HA) - @timothysc](https://docs.google.com/document/d/1lH9OKkFZMSqXCApmSXemEDuy9qlINdm5MfWWGrK3JYc/edit#)
+
+* [kubeadm HA implementation plan - @luxas](https://docs.google.com/document/d/1ff70as-CXWeRov8MCUO7UwT-MwIQ_2A0gNNDKpF39U4/edit#)


contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+What actually happens in the Static Pod -> Self-hosted transition?
+
+First the PodSpec is modified. For instance, instead of hard-coding an IP for the API server to listen on it is dynamically fetched from the Downward API. When it’s ok to proceed, the self-hosted control plane DaemonSets are posted to the Static Pod API Server. The Static Pod API Server starts the self-hosted Pods normally. The Pods get in the Running state, but fails to bind to the port on the host and backs off internally. During that time, kubeadm notices the Running state of the self-hosted Pod, and deletes the Static Pod file which leads to that the kubelet stops the Static Pod-hosted component immediately. On the next run, the self-hosted component will try to bind to the port, which now is free, and will succeed. A self-hosted control plane has been created!
```
kargakis: Can you break all single-line comments into separate lines? Will be better for reviews.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+# Self-hosting
+
+Self-hosting will be the default way of deploying the control plane, but it will be optional. There will be a `kubeadm init`-time feature gate to disable it.
+
+Only *control plane components* will be in-scope for self-hosting for v1.8.
+
+Notably, not etcd or kubelet for the moment.
+
+Self-hosting will be a [phase](https://github.com/kubernetes/kubeadm/blob/master/docs/design/design.md) in the kubeadm perspective. As there are Static Pods on disk from earlier in the `kubeadm init` process, kubeadm can pretty easily parse that file and extract the PodSpec from the file. This PodSpec will be modified a little to fit the self-hosting purposes and be injected into a DaemonSet which will be created for all control plane components (API Server, Scheduler and Controller Manager). kubeadm will wait for the self-hosted control plane Pods to be running and then destroy the Static Pod-hosted control plane.
```
kargakis: Assuming the DaemonSet controller continues to schedule DS pods, isn't it better for the controller manager to be a Deployment as opposed to a DaemonSet? If all control plane components run as DaemonSets, isn't the controller manager a single point of failure?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+# Implementation options of the upgrade
+
+High-level approaches:
+
+1. Build the upgrade logic into kubeadm
+
+2. Build an upgrading Operator that does the upgrading for us
+
+    * Would consume a TPR with details how to do the upgrade
```
kargakis: CRD instead of TPR

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+We plan to solve this by having a separate `prepull` task in the beginning of the upgrade to minimize service outage. What it will basically do is to create a DaemonSet for all the control plane components, but instead of running the component, it executes `sleep 3600` or a similar command.
+
+Upgrading etcd
+
+In v1.7 and v1.8, etcd runs in a Static Pod on the master. In v1.7, the default etcd version for Kubernetes is v3.0.17 and in k8s v1.8 the recommended version will be something like v3.1.10. In the v1.7->v1.8 upgrade path, we could offer upgrading etcd as well as an opt-in*. *This only applies to minor versions and 100% backwards-compatible upgrades like v3.0.x->v3.1.y, not backwards-incompatible upgrades like etcdv2 -> etcdv3.
+
+The easiest way of achieving an etcd upgrade is probably to create a Job (with `.replicas=<masters>`, Pod Anti-Affinity and `.spec.parallellism=1`) of some kind that would upgrade the etcd Static Pod manifest on disk by writing a new manifest, waiting for it to restart cleanly or rollback, etc.
```
kargakis: Would be nice to add an example in the kubeadm repo.

contributors/design-proposals/kubeadm-upgrades.md
roberthbailey: The kubelet will keep the controller manager running. You could get into a state where the controller manager is unrunnable due to a configuration or coding error, but I don't see how that would be any better with deployments.

contributors/design-proposals/kubeadm-upgrades.md
roberthbailey: > The Pods get in the Running state...

I believe that the behavior described is only true for the apiserver. The KCM and scheduler should just run fine since they aren't trying to bind to host ports.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+The historical preference was to build a reusable Operator that does the upgrading for us. *Either way we would deliver the same 'kubeadm upgrade' UX.* In the latter case, that would involve kubeadm deploying an upgrading controller on behalf of the user, and tracking its lifecycle through the kubernetes API.
+
+There are pros and cons for each method:
+
+1. Build it into the `kubeadm` CLI command
+
+    1. Pros:
+
+        1. *Simplicity*; Very straightforward to implement; basically doing the same steps as `kubeadm init` would, only slightly modified to fit the upgrading purpose. We will construct a `kubeadm upgrade` command in any case
```
roberthbailey: You should make these bullets instead of numbers so that the alphabet doesn't keep increasing across top level items.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+    4. Cons:
+
+        9. Has to deal with any version skew between the controller and the CLI
+
+        10. Has to deal with pre-pulling the image to a specific node in the case of no internet connection
+
+        11. Has to define an API (CRD or the like) between the client and the Operator
+
+**Decision**: Keep the logic inside of the kubeadm CLI (option 1) for the implementation in v1.8.0.
```
roberthbailey: This makes sense. I don't think it would be difficult to move the logic later (at least not more difficult than using an operator now).

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+Upgrade a Static Pod-hosted cluster to a self-hosted one
+
+The first thing `kubeadm upgrade` would have to do when upgrading from v1.7 to v1.8 would be to make the static-pod-hosted v1.7 cluster self-hosted. This can easily be done thanks to the recent refactoring of self-hosting into an atomically invoked phase. 
+`kubeadm upgrade plan` will inform the user about the Static Pod => self-hosted switch in the case of an v1.7.x to v1.8.x upgrade. It will also be possible to convert a Static Pod-hosted control plane to a self-hosted one with another command like `kubeadm phase selfhosting apply`.
+
+Keeping user customizations between releases
+
+One of the hardest parts with implementing the upgrade will be to respect customizations made by the user at `kubeadm init` time. The proposed solution would be to store the kubeadm configuration given at `init`-time in the API as a ConfigMap, and then retrieve that configuration at upgrade time, parse it using the API machinery and use it for the upgrade.
+
+This highlights a very important point: **We have to get the kubeadm configuration API group to Beta (v1beta1) in time for v1.8.**
```
roberthbailey: Is this on track?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+Instead of generating new manifests from a versioned configuration object, we could try to add "filters" to the existing manifests and apply different filters depending on what the upgrade looks like. This approach, to modify existing manifests in various ways depending on the version bump has some pros (modularity, simple mutating functions), but the matrix of functions and different policies would grow just too big for this kind of system, so we voted against this alternative in favor for the solution above.
+
+"Upgrading" without bumping the version number
+
+One interesting use-case for `kubeadm upgrade` will be to modify other configuration in the cluster. If we assume the configuration file options from `kubeadm init` will be fetchable from a ConfigMap inside of the cluster at any time, it would be possible to fetch the current state, compare it to the desired state (from `kubeadm apply --config`) and build a patch of it. Kubeadm would validate the fields changed in that patch and allow/deny "upgrading" those values.
+
+This is a stretch goal for v1.8, but not a strictly necessary feature.
+
+Pre-pulling of images/ensuring the upgrade doesn’t take too long
```
roberthbailey: This doesn't seem like a subsection of alternatives considered.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+By default, kubeadm should tell the user what it is doing in a way that is verbose enough to satisfy the most users (but doesn’t print out a wall of text). Also see the verbosity section.
+
+`kubeadm upgrade apply` implementation flow
+
+1. Validate the version given and configuration. Fail fast if the configuration is missing or invalid. Fail fast if the user is trying to upgrade to a version that is not supported.
+
+2. Make sure the cluster is healthy
+
+    1. Make sure the API Server’s `/healthz` endpoint returns `ok`
```
roberthbailey: Should we look at /componentstatuses too?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+    2. Makes sure all Nodes return `Ready` status
```
roberthbailey: Do you also check node conditions?

contributors/design-proposals/kubeadm-upgrades.md
liggitt: if you want kubeadm to be a building block for widely varying clusters, be cautious about requiring a custom resource be part of the base install path.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+Checkpointing
+
+In order to be able to reboot a self-hosted cluster (e.g. a single, self-hosted master), there has to be some kind of checkpointing mechanism. Basically the kubelet will write some state to disk for Pods that opt-in to checkpointing ("I’m running an API Server, let’s write that down so I remember it", etc.). Then if the kubelet reboots (in the single-master case for example), it will check the state store for Pods it was running before the reboot. It discovers that it ran an API Server, Controller Manager and Scheduler and starts those Pods now as well.
+
+This solves the chicken-and-egg problem that otherwise would occur when the kubelet comes back up, where the kubelet tries to connect to the API server, but the API server hasn’t been started yet, it should be running as a Pod on that kubelet, although it isn’t aware of that.
```
yujuhong: nit: Break up the two sentences: " API server hasn’t been started yet, it should be running

s/it should be/The API server
s/on that kubelet/on that node
s/it isn't aware/the kubelet isn't aware

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+Checkpointing
+
+In order to be able to reboot a self-hosted cluster (e.g. a single, self-hosted master), there has to be some kind of checkpointing mechanism. Basically the kubelet will write some state to disk for Pods that opt-in to checkpointing ("I’m running an API Server, let’s write that down so I remember it", etc.). Then if the kubelet reboots (in the single-master case for example), it will check the state store for Pods it was running before the reboot. It discovers that it ran an API Server, Controller Manager and Scheduler and starts those Pods now as well.
```
yujuhong: What's the downside of checkpointng all pods vs. selective, opt-in pods?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+This solves the chicken-and-egg problem that otherwise would occur when the kubelet comes back up, where the kubelet tries to connect to the API server, but the API server hasn’t been started yet, it should be running as a Pod on that kubelet, although it isn’t aware of that.
+
+We’re aiming to build this functionality into the kubelet behind an alpha feature flag.
+
+# Upgrading policies
+
+Defining decent upgrading and version skew policies is important before implementing.
+
+Definitions
```
yujuhong: Make it a header for readability, e.g., `##Definitions`

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+    4. Example: kubeadm v1.8.2 supports *creating a new cluster* for Kubernetes versions v1.7.x and v1.8.x
+
+3. The control plane must be upgraded before the kubelets in the cluster
+
+    5. For kubeadm; the maximum amount of skew between the control plane and the kubelets is *one minor release*.
+
+    6. Example: running `kubeadm upgrade apply --version v1.9.0` against a v1.8.2 control plane will error out if the nodes are still on v1.7.x
+
+4. This means that there are possibly two kinds of upgrades kubeadm can do:
```
yujuhong: Is this part of the upgrade policy, or what's derived from it? I feel like this section is mixed with policy, what `kubeadm` can do, and implementation specifics. I'd suggest splitting them if possible.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+            3. New flags that should be added
+
+                3. In order to enable new and required features, kubeadm should automatically append these config options
+
+            4. Resources that have to be modified
+
+                4. Example: The `system:nodes` ClusterRoleBinding had to lose its binding to the `system:nodes` Group when upgrading to v1.7; otherwise the Node Authorizer wouldn’t have had any effect.
+
+5. Using kubeadm, you must upgrade the control plane atomically.
```
yujuhong: If the upgrade failed, would kubeadm roll back the changes?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+Timescale for implementation of this proposal (working self-hosting & upgrades) is Kubernetes 1.8. This means that [features issues must be in by Aug 1](https://github.com/kubernetes/features/pull/305) and [code must be in by Sept 1](https://github.com/kubernetes/features/pull/305).
+
+# Self-hosting
+
+Self-hosting will be the default way of deploying the control plane, but it will be optional. There will be a `kubeadm init`-time feature gate to disable it.
+
+Only *control plane components* will be in-scope for self-hosting for v1.8.
+
+Notably, not etcd or kubelet for the moment.
```
yujuhong: How do users of `kubeadm` upgrade kubelets today?

contributors/design-proposals/kubeadm-upgrades.md
yujuhong: What's TPR? Couldn't find it in this proposal...

roberthbailey:
Thanks for the comments @yujuhong!

contributors/design-proposals/kubeadm-upgrades.md
aaronlevy: Are there any requirements for KCM and scheduler to use host networking? We've been successful running them both on the pod network & using service-account for apiserver access.

I believe the health check wouldn't come into play in the case of apiserver because it is just reaching out to the same host:port regardless if it's static manifest or the self-hosted "replacement". It's a bit disingenuous in that the healthcheck is really only testing one of them.

One problematic piece here is that the apiserver can't actually bind on the port, so it will be in a restart loop. If it does this enough then the backoff period can make it seem like something went wrong (we start the pivot by removing static pod, then the backoff period on the replacement causes it to not be restarted for a while).

As long as the kubelet itself doesn't die during that process, it will recover. It's just somewhat less than ideal.

contributors/design-proposals/kubeadm-upgrades.md
timothysc: It's done via package manager.  yum/apt.

contributors/design-proposals/kubeadm-upgrades.md
timothysc: > On the other hand, if this is not true, there'd be two copies of KCM and scheduler running. That may cause unexpected consequences.

It shouldn't, they are setup to lock on endpoints by default.

mtaufen: It would be nice to have these steps in a numbered list

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+The rest of the document assumes a self-hosted cluster. It will not be possible to use 'kubeadm upgrade' a cluster unless it's self-hosted. The sig-cluster-lifecycle team will still provide some manual, documented steps that an user can follow to upgrade a non-self-hosted cluster.
+
+Self-hosting implementation for v1.8
+
+The API Server, controller manager and scheduler will all be deployed in DaemonSets with node affinity (using `nodeSelector`) to masters. The current way DaemonSet upgrades currently operate is "remove an old Pod, then add a new Pod". Since supporting “single masters” is a definite requirement, we have to either workaround removing the only replica of the scheduler/controller-manager by duplicating the existing DaemonSets (e.g. a `temp-self-hosted-kube-apiserver` DS will be created as a copy of the normal `self-hosted-kube-apiserver` DS during the upgrade) or ask for a [new upgrade strategy](https://github.com/kubernetes/kubernetes/issues/48841) (“add first, then delete”, which is generally useful) from sig-apps.
```
mtaufen: +1 add first, then delete. That sounds much nicer than hacking around it with temporary duplicates.

contributors/design-proposals/kubeadm-upgrades.md
mtaufen: Make sure that if the Pods are still running, the Kubelet doesn't create duplicates on restart. There are many reasons the Kubelet could restart without a full node reboot: OOM kill, dynamic config, etc.

mtaufen: Yes, why wouldn't we just checkpoint all pods by default? How much disk space would this consume? Would it be small enough that we could require opt-out instead of opt-in?

contributors/design-proposals/kubeadm-upgrades.md
mtaufen: Just to be clear, you're saying that checkpointing solves the problem of the Kubelet hosting itself in a Pod if it can't contact the API server and learn that it should?

Side question: If we have checkpoints, can we start provisioning initial checkpoints instead of static pods?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+Patch version upgrade = This is an upgrade from a version (vX.Y.Z) with the patch component equal to `Z` to a version (vX.Y.Z+n) with the patch component of `Z+n`. `n` can be freely chosen by the user. For instance, a patch version upgrade can look like `v1.8.3` to `v1.8.6`.
+
+1. The user must upgrade the kubeadm CLI before the control plane
+
+    1. Upgrading to a higher patch release than the kubeadm CLI version will
+*be possible*, but yield a warning and require the `--force` flag to `kubeadm upgrade apply` to continue
+
+    2. Upgrading to a higher minor release than the kubeadm CLI version will **not** be supported.
+
+    3. Example: kubeadm v1.8.3 can upgrade your v1.8.3 cluster to v1.8.6 if you specify `--force` at the time of the upgrade, but kubeadm can never upgrade your v1.8.3 cluster to v1.9.0
```
mtaufen: nit: be a little more specific - s/but kubeadm can never/but kubeadm v1.8.3 can never...

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+2. The kubeadm CLI can create clusters with the latest two minor releases.
+
+    4. Example: kubeadm v1.8.2 supports *creating a new cluster* for Kubernetes versions v1.7.x and v1.8.x
+
+3. The control plane must be upgraded before the kubelets in the cluster
+
+    5. For kubeadm; the maximum amount of skew between the control plane and the kubelets is *one minor release*.
```
mtaufen: I think we should support two trailing versions for clients. API changes are supposed to be backwards-compatible, so this shouldn't be a problem for kubeadm. In theory.
I'm not sure whether we should try to support skipping a minor version during an upgrade though. I think people typically do this one version at a time.

contributors/design-proposals/kubeadm-upgrades.md
timothysc: The plan is to opt in with annotations or label.

contributors/design-proposals/kubeadm-upgrades.md
timothysc: It's CRD.

roberthbailey:
@luxas - you mentioned on Friday that you were going to wait until Tuesday to collect feedback and then update the proposal. 

Friendly ping to let us know when the PR is ready to be re-reviewed.

contributors/design-proposals/kubeadm-upgrades.md
jessicaochen: Perhaps it would be worth linking to the expected node upgrade process or having a brief description. I was wondering how the nodes get upgraded as well.

luxas:
@roberthbailey I changed my mind ;)
I found it more valuable to actually focus my hours on working on the code in this phase of the cycle and get dependent PRs merged.
The WIP is here: https://github.com/kubernetes/kubernetes/pull/48899

Feel free to take it for a spin.
I'm gonna do the cleanup of this doc later when more of the actual dependent code is merged, otherwise it would be too tight. This proposal as-is had consensus in the SIG, and the remaining comments are minor, wording or clarifications, so it has lower priority for me than actually shipping the code.

I hope I can get to this by the end of next week at least.

Finally, thank you everyone for commenting -- I'm sorry I'm overloaded with other things at the moment plus the upgrades impl., I'll answer your questions as soon as I can.

contributors/design-proposals/kubeadm-upgrades.md
luxas: I'll do that

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+Status as of v1.7: self hosted config file option exists; not split out into phase. The code is fragile and uses Deployments for the scheduler and controller-manager, which unfortunately [leads to a deadlock](https://github.com/kubernetes/kubernetes/issues/45717) at `kubeadm init` time. It does not leverage checkpointing so the entire cluster burns to the ground if you reboot your computer(s). Long story short; self-hosting is not production-ready in 1.7.
+
+The rest of the document assumes a self-hosted cluster. It will not be possible to use 'kubeadm upgrade' a cluster unless it's self-hosted. The sig-cluster-lifecycle team will still provide some manual, documented steps that an user can follow to upgrade a non-self-hosted cluster.
```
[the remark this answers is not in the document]
luxas: Yeah, I think that breaking out the self-hosting parts is a good idea.
Also self-hosting didn't come out as important in this process as we had thought

contributors/design-proposals/kubeadm-upgrades.md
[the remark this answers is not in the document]
luxas: clarified that we're using the nodeSelector feature right now, not the "real" node affinity one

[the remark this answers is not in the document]
luxas: fixed

contributors/design-proposals/kubeadm-upgrades.md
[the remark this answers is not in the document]
luxas: fixed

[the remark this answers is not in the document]
luxas: fixed

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,452 @@
...
+
+In order to be able to reboot a self-hosted cluster (e.g. a single, self-hosted master), there has to be some kind of checkpointing mechanism. Basically the kubelet will write some state to disk for Pods that opt-in to checkpointing ("I’m running an API Server, let’s write that down so I remember it", etc.). Then if the kubelet reboots (in the single-master case for example), it will check the state store for Pods it was running before the reboot. It discovers that it ran an API Server, Controller Manager and Scheduler and starts those Pods now as well.
+
+This solves the chicken-and-egg problem that otherwise would occur when the kubelet comes back up, where the kubelet tries to connect to the API server, but the API server hasn’t been started yet, it should be running as a Pod on that kubelet, although it isn’t aware of that.
+
+We’re aiming to build this functionality into the kubelet behind an alpha feature flag.
+
+# Upgrading policies
+
+Defining decent upgrading and version skew policies is important before implementing.
```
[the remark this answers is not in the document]
luxas: fixed

contributors/design-proposals/kubeadm-upgrades.md
luxas: fixed

contributors/design-proposals/kubeadm-upgrades.md
luxas: @timothysc see comment on checkpointing ^

contributors/design-proposals/kubeadm-upgrades.md
luxas: I implemented this the hacky way for now, as the UpdateStrategy didn't make v1.8
However, we definitely want add first, then delete longer-term

contributors/design-proposals/kubeadm-upgrades.md
luxas: Will do

roberthbailey:
@luxas - should we clean this up and get it merged (now that kubeadm supports upgrade)?

luxas:
Yeah, sorry, it's on my ever-growing backlog. I'll try to get to this in a week or two or so, lots of other tasks to complete still :(

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/825)

luxas:
@timothysc @roberthbailey Please read this doc up to the "Various other notes"; I need to finish the last sections there yet. Thanks!

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+
+ - The kubeadm CLI version must be higher or equal to the upgrade target version
+   - **Required policy**: The kubeadm CLI minor version must be equal to or higher than the upgrade target minor version.
+     - Example: kubeadm CLI version `v1.8.3` can upgrade a cluster to `v1.7.4` and `v1.8.5`, but not `v1.9.0`.
+   - **Optional policy**: The kubeadm CLI patch version should be equal to or higher than the upgrade target patch version.
+     - Example: kubeadm CLI version `v1.8.3` can upgrade a cluster to `v1.8.3` without an explicit grant, but upgrading to `v1.8.5` requires forcing. 
+ - The kubeadm CLI of version `vX.Y.Z` can create clusters with minor versions `Y` and `Y-1`.
+   - This is a **required policy**, and is not specific to upgrades. This policy is also is also enforced when creating a cluster.
+     - Example: kubeadm CLI version `v1.8.2` can create clusters of (and upgrade clusters to) version `v1.8.x` and `v1.7.x`.
+ - The control plane version must always be higher than or equal to the version of all the kubelets, and the maximum allowed skew is one minor versions.
```
roberthbailey: This assumes that the entire control plane has a consistent version (which we may need to relax a bit for HA). Would it be better to phrase as "all control plane components must have a version higher than or equal to the version of all kubelets"?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+   - **Required policy**: The control plane version must be higher than or equal to the kubelet version. This is a generic Kubernetes policy.
+     - Example: For a control plane of version `v1.8.4`; a kubelet version may be `v1.8.2`, but not `v1.9.3`
+   - **Skippable policy**: The maximum allowed minor skew between the control plane and the kubelets is one minor version.
+     - Example: For a control plane of version `v1.7.2` being upgraded to `v1.8.5`; the minimum allowed kubelet version is `v1.7.0`.
+     - Note: We might relax this policy to allow two minor versions before going to GA.
```
roberthbailey: above in the ga in 1.10 section you say "Supported Control Plane <-> kubelet minor skew: two minor versions." so it sounds like you are proposing relaxing this constraint.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+ - The version change must be positive, i.e. only upgrades are supported; not downgrades or runtime reconfiguration.
```
roberthbailey: since this is skippable, can you change must to should in this sentence?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+   - This is a **skippable policy** in v1.8, and will be removed in future versions before going to GA.
+     - Example: Downgrading from `v1.8.3` to `v1.8.2` or reconfiguring a cluster on version `v1.8.1` can only be done (initially) along with the `--force` flag.
+ - The maximum allowed upgrade minor skew is one minor version.
+   - This is a **required policy**. You can upgrade from `vX.Y.Z1` to `vX.Y+1.Z2`, but not to `vX.Y+2.Z3`.
+     - Example: Upgrading from `v1.6.2` to `v1.8.4` is not supported.
+     - Note: We might relax this policy to allow two minor versions before going to GA.
```
roberthbailey: this seems reasonable for the control plane. once we start talking about kubelet upgrades i think we will want to support upgrading multiple minor versions at a time.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+ - The target upgrade version must not be an alpha, beta version unless the user explicitely allows it with `--allow-experimental-upgrades`
+   - This is a **skippable policy**, in order to have an extra safety check but still allow it for users who may want it. Such upgrades are not supported.
+     - Example: Upgrading from `v1.7.1` to `v1.8.0-beta.3` can only be done with at least one of the `--allow-experimental-upgrades` or `--force` flags.
+ - The target upgrade version must not be a release candidate version unless the user explicitely allows it with `--allow-release-candidate-upgrades` (or `--allow-experimental-upgrades`)
+   - This is a **skippable policy**, in order to have an extra safety check but still allow users to easily test our RCs. These upgrades "should work".
```
roberthbailey: does "should work" mean that we do support these (as opposed to upgrades to alpha / beta releases)?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+     - Example: Upgrading from `v1.7.1` to `v1.8.0-rc.1` can only be done with at least one of the `--allow-release-candidate-upgrades`, `--allow-experimental-upgrades` or `--force` flags.
+
+## Initial scope of `kubeadm upgrade`
+
+The initial scope of kubeadm upgrades in v1.8 is to upgrade the _control plane_ along with kube-proxy and kube-dns that are hosted on top of Kubernetes.
+The control plane upgrade is atomic; either all or nocontrol plane components get upgraded. If a post-upgrade step fails though (like upgrading Bootstrap Tokens or applying a new kube-dns), kubeadm upgrade just returns a non-zero code instead of reverting the successful control plane upgrade. We might change the latter behavior in the future.
```
roberthbailey: s/nocontrol/no control

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+
+In the first iteration, upgrading the kubelets and/or etcd in the cluster is a manual task (see the sections below).
```
roberthbailey: replace and/or with and

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+We will address these limitations in future versions before graduating this functionality to GA.
```
roberthbailey: this makes me cautious about promising ga with 1.10, since it isn't clear how long this will take to implement and get ready for ga.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+     - `--config`: A path to a configuration file to use as the desired state for the cluster component configuration. If omitted, configuration is read from a ConfigMap in the cluster.
+     - `--kubeconfig`: The path to a configuration file to use when talking to the cluster. Default: `/etc/kubernetes/admin.conf`.
+     - `--allow-experimental-upgrades`: Allow the user to upgrade to alpha, beta and release candidate versions.
+     - `--allow-release-candidate-upgrades`: Allow the user to upgrade to release candidate versions.
+     - `--print-config`: Print out the configuration read from the file or ConfigMap in the cluster.
+     - `--skip-preflight-checks`: Skip the preflight checks that are run before execution.
+ - `kubeadm upgrade plan`: Serves as a way to show users:
+   - If the cluster is in an upgradeable state.
+   - What versions the user can upgrade the cluster to.
+ - `kubeadm upgrade apply [version]`:
```
roberthbailey: per the sample output below this seems like it should be `kubeadm upgrade apply [--version VERSON]`

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+[kubeadm upgrade apply output]
+```
+
+## Upgrade planning
+
+In order for the user to know whether the cluster may be upgraded or not; the `kubeadm upgrade plan` command was written.
+The command:
+
+1. Checks if the cluster is healthy
+  - Does the API server `/healthz` endpoint return `ok`?
```
roberthbailey: this didn't render properly in the markdown (the bullets aren't nested under the numbers and the count resets each time there are bullets). please fix.

roberthbailey: we don't check the other master components as well? i know that the componentstatuses endpoint is a terrible hack but it does tell you whether the scheduler and controller manager (and etcd if it's at localhost) are also healthy.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+  - (If Static Pods are used) Do all required Static Pod manifests exist on disk?
+  - (If cluster is self-hosted) Are all control plane DaemonSets healthy?
+1. Builds an internal configuration object to apply to the cluster
+  - The `--config` file takes precedence; otherwise the current configuration stored in a ConfigMap in the cluster is used.
+  - The `KubernetesVersion` field is set to the desired version.
+  - The built configuration object is optionally printed to STDOUT if `--print-config` is `true`.
+1. Information about what releases are available are downloaded from the Kubernetes Release & CI GCS buckets.
+  - The `dl.k8s.io/release/stable.txt` endpoint (and optionally `.../stable-1.X.txt`) is queried for the latest stable version to display.
+  - The API Server, kubeadm CLI, and all kubelets' versions are taken into account when computing the possible upgrades.
+  - The command may show zero to four upgrade possibilities.
```
roberthbailey: is there a reason that four is the upper bound?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+  - (If Static Pods are used) Do all required Static Pod manifests exist on disk?
+  - (If cluster is self-hosted) Are all control plane DaemonSets healthy?
+1. Builds an internal configuration object to apply to the cluster
+  - The `--config` file takes precedence; otherwise the current configuration stored in a ConfigMap in the cluster is used.
+  - The `KubernetesVersion` field is set to the desired version.
+  - The built configuration object is optionally printed to STDOUT if `--print-config` is `true`.
+1. Enforces the version policies highlighted above in [Pre-upgrade policies](#pre-upgrade-policies)
+1. Asks the user for a `[Y|n]` answer whether to proceed with the upgrade or not (if the session is interactive)
+1. Performs a method of ensuring that all necessary control plane images are available locally to the master(s), aka. "pre-pulling"
+  - Due to lack of an API to the kubelet that would make it possible to say "make sure this image X is available locally"; we have to
```
roberthbailey: Have we filed a feature request against sig-node to add such an API?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+   - Would have to deal with pre-pulling the image to a specific node in the case of no internet connection
+   - Would have to define an API (CRD or the like) between the client and the Operator (more time-consuming to get right)
+
+**Decision**: Keep the logic inside of the kubeadm CLI (option 1) for the implementation in v1.8.0.
+
+We *might* revisit this and build an upgrade operator in future, either before GA of this functionality or after, in v2.
+
+**Rewrite manifests completely from config object or mutate existing manifest:**
+
+Instead of generating new manifests from a versioned configuration object, we could try to add "filters" to the existing manifests and apply different filters depending on what the upgrade looks like. This approach, to modify existing manifests in various ways depending on the version bump has some pros (modularity, simple mutating functions), but the matrix of functions and different policies would grow just too big for this kind of system, so we voted against this alternative in favor for the solution above.
```
roberthbailey: did we actually vote? maybe say decided instead?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+A control plane can be set up in a couple of ways:
+
+ - Running the control plane components standalone; as systemd services for instance.
+   - kubeadm **does not** support handling any such cluster
+ - Running the control plane components in containers with kubelet as the babysitter; using Static Pods
+   - [Static Pods](#TODO) is a way to run standalone Pods on a node by dropping Pod manifests in a specific path.
+   - In kubeadm v1.4 to v1.8, this is the default way of setting up the control plane
+ - Running the control plane in Kubernetes-hosted containers as DaemonSets; aka. [Self-Hosting](#TODO)
+   - When creating a self-hosted cluster; kubeadm first creates a Static Pod-hosted cluster and then pivots to the self-hosted control plane.
+   - This is the default way to deploy the control plane since v1.9; but the user can opt-out if it and stick with the Static Pod-hosted cluster.
```
roberthbailey: s/since/beginning with/

also replace ; with ,

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+
+When the upgrade functionality is introduced in v1.8; the default way to set up the control plane is by using Static Pods.
+
+When the control plane is hosted using Static Pods, the actual upgrade flow looks like this:
+ - Gets the Mirror Pod manifests from the API Server (`kubectl -n kube-system get pod ${component}-${nodename} -ojson`), and sha256-hashes the JSON content.
+ - Writes new Static Pod manifests to a **temporary directory** (with the prefix `/etc/kubernetes/tmp/kubeadm-upgraded-manifests*`) based on the new configuration object built internally earlier.
+ - Creates a **backup directory** with the prefix `/etc/kubernetes/tmp/kubeadm-backup-manifests*`.
+   - Q: Why `/etc/kubernetes/tmp`?
+   - A: Possibly not very likely, but we concluded that there may be an attack area for computers where `/tmp` is shared and writable by all users.
+     We wouldn't want anyone to mock with the new Static Pod manifests being applied to the clusters. Hence we chose `/etc/kubernetes/tmp`, which is
```
roberthbailey: replace mock with muck or mess

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+     root-only owned.
+ - In a loop for all control plane components:
```
roberthbailey: does the order of the control plane components matter?

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+ - In order to not leak directories on the host, kubeadm deletes the **backup** and **temporary directory**.
+   - If there are user requests, we might make it possible to opt-out of the "always delete old manifests"-policy.
+
+##### Rollbacks
+
+In case any of these operations fail, kubeadm will rollback the earlier manifests to the previous state. Hence the backup directory is important.
+For instance, if the scheduler doesn't come up cleanly; kubeadm will rollback the previously (successfully upgraded) API server, controller manager
+manifests as well as the scheduler manifest.
+
+#### Self-hosted control plane
```
roberthbailey: it looks like this section isn't finished so i'm going to stop reviewing here.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
+# kubeadm upgrades proposal
+
+Authors: Lucas Käldström & the SIG Cluster Lifecycle team
+Last updated: October 2017
```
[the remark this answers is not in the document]
luxas: fixed

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+
+## Abstract
+
+This proposal describes how kubeadm will support a upgrading clusters in an user-friendly and automated way using different techniques.
```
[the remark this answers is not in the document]
luxas: oops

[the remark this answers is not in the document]
luxas: removed the ambiguity; I mean that kubeadm can actually perform different tasks for different clusters under the hood (self-hosted vs static pod hosted), but that isn't relevant here

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+## Graduation requirements
+
+ - Beta in v1.8:
+   - Support for the `kubeadm upgrade plan` and `kubeadm upgrade apply` commands.
+   - Support for dry-running upgrades.
+   - Supports upgrading clusters, but not 100% support for downgrades or "runtime reconfiguration"
+   - Support for upgrading the API Server, the Controller Manager, the Scheduler, kube-dns & kube-proxy
+   - Support for performing necessary post-upgrade steps like upgrading Bootstrap Tokens that were alpha in v1.6 & v1.7 to beta ones in v1.8
+   - Automated e2e tests running.
+ - GA in v1.10:
```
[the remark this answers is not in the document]
luxas: yup, I left this as `in a future version` now instead

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+   - Support for downgrades and runtime reconfiguration.
+   - Support for upgrading etcd.
+   - The upgrading functionality is available from the `kubeadm phase` command.
+   - Support for upgrading kubelets.
+   - Supported Control Plane <-> kubelet minor skew: two minor versions.
+   - Support for fully remote-executable upgrade operations against the cluster
+   - Support for machine-readable `kubeadm plan` output
+   - Support for upgrading HA/multi-master clusters
```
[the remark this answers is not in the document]
luxas: Yeah, I'll leave this here as-is now and fixup later when we merge all HA considerations and GA plans into this doc later.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+The initial task that must be performed when considering to add functionality to seamlessly upgrade clusters, is to define policies for which kinds of upgrades
+and version skews should be supported.
+
+### Definitions
+
+ - **Minor version upgrade**: An upgrade from a version (`vX.Y.Z1`) with the minor component equal to `Y` to a version (`vX.Y+1.Z2`) with the minor component of `Y+1`. Upgrading from `v1.8.3` to `v1.9.0` is a minor version upgrade.
+ - **Patch version upgrade**: An upgrade from a version (`vX.Y.Z`) with the patch component equal to `Z` to a version (`vX.Y.Z+n`) with the patch component of `Z+n`. `n` can be freely chosen by the user. Upgrading from `v1.8.3` to `v1.8.6` is a patch version upgrade.
+ - **Required upgrade policy**: An upgrade policy that must be satisfied; a precondition the user can't ignore.
+ - **Skippable upgrade policy**: An upgrade policy that should be satisfied, but is skippable by passing the `--force` flag.
+ - **Runtime reconfiguration**: An user invokes the `kubeadm upgrade apply` command with the current version, but with a different configuration.
```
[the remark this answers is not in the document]
luxas: The actual flow to be implemented is the same, right?
It definitely is more like an update (vs upgrade).

Do you think we should strongly distinguish these UX-wise or implementation-wise?
I mostly meant the impl. is kind of identical.

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+
+### Pre-upgrade policies
+
+ - The kubeadm CLI version must be higher or equal to the upgrade target version
```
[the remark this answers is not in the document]
luxas: Anything I should add there or?

contributors/design-proposals/kubeadm-upgrades.md
luxas: fixed

contributors/design-proposals/kubeadm-upgrades.md
```diff
@@ -0,0 +1,350 @@
...
+   - **Required policy**: The kubeadm CLI minor version must be equal to or higher than the upgrade target minor version.
+     - Example: kubeadm CLI version `v1.8.3` can upgrade a cluster to `v1.7.4` and `v1.8.5`, but not `v1.9.0`.
+   - **Optional policy**: The kubeadm CLI patch version should be equal to or higher than the upgrade target patch version.
+     - Example: kubeadm CLI version `v1.8.3` can upgrade a cluster to `v1.8.3` without an explicit grant, but upgrading to `v1.8.5` requires forcing. 
+ - The kubeadm CLI of version `vX.Y.Z` can create clusters with minor versions `Y` and `Y-1`.
+   - This is a **required policy**, and is not specific to upgrades. This policy is also is also enforced when creating a cluster.
+     - Example: kubeadm CLI version `v1.8.2` can create clusters of (and upgrade clusters to) version `v1.8.x` and `v1.7.x`.
```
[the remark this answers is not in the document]
luxas: This explains the minor version policy for downwards; i.e. what older k8s versions are supported

contributors/design-proposals/kubeadm-upgrades.md
luxas: Yeah; does that shine through clearly enough?

contributors/design-proposals/kubeadm-upgrades.md
luxas: done

contributors/design-proposals/kubeadm-upgrades.md
luxas: yeah, this focuses on the control plane for now. Let's see what we do with the kubelets here later

contributors/design-proposals/kubeadm-upgrades.md
luxas: Well I mean that the risk should be low, but we can't support these cases (alpha is alpha, right?)

contributors/design-proposals/kubeadm-upgrades.md
luxas: fixed

contributors/design-proposals/kubeadm-upgrades.md
luxas: done

contributors/design-proposals/kubeadm-upgrades.md
luxas: now renamed to not state v1.10, and just state what requirements there are

contributors/design-proposals/kubeadm-upgrades.md
luxas: I'll paste some real output here in a minute. It should be `kubeadm upgrade apply [version]` as people thought an extra `--version` arg was unnecessary

contributors/design-proposals/kubeadm-upgrades.md
luxas: Thanks, I think this should be fixed now

luxas: We don't right now, as I don't know how reliable that actually is.

contributors/design-proposals/kubeadm-upgrades.md
luxas: added a rationale

contributors/design-proposals/kubeadm-upgrades.md
luxas: somewhere there is such an issue..., but we haven't escalated it more broadly IIRC

contributors/design-proposals/kubeadm-upgrades.md
luxas: right, we didn't actually vote. Thanks for the reword suggestion there.

contributors/design-proposals/kubeadm-upgrades.md
luxas: done

contributors/design-proposals/kubeadm-upgrades.md
luxas: hehe, typo ;)

contributors/design-proposals/kubeadm-upgrades.md
luxas: added a comment that order _shouldn't_ matter, but we do apiserver, ctlr-mgr, sched

contributors/design-proposals/kubeadm-upgrades.md
luxas: Yeah, thanks for the review so far!

contributors/design-proposals/kubeadm-upgrades.md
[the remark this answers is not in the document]
roberthbailey: I agree that the implementation is the same, but I was commenting on how the user sees it. I'm still not sure how important it is to distinguish update vs. upgrade but they definitely have different risk profiles.

contributors/design-proposals/kubeadm-upgrades.md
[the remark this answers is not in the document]
roberthbailey: No, just wanted to add that thought when it occurred to me.

contributors/design-proposals/kubeadm-upgrades.md
roberthbailey: in that case make it explicit (as above) that these upgrades are generally expected to work but are also officially unsupported.

contributors/design-proposals/kubeadm-upgrades.md
roberthbailey: componentstatuses is pretty reliable the way we have the master configured in gke. i don't know if that translates to it being reliable in general though.

roberthbailey:
Please poke when the remainder of the doc is finished so I can hopefully just do one more pass.

kargakis:
@luxas @roberthbailey should this doc move inside https://github.com/kubernetes/community/tree/master/contributors/design-proposals/cluster-lifecycle?

luxas:
@kargakis yes indeed. I initially filed this PR before we did that, will update on my next round here.

PR #832, opened 2017-07-22
add proposal for subresourced secrets

ericchiang:
cc @kubernetes/sig-auth-misc 

This is a proposal to require accessing secret data through a subresource, with list and watch operations not returning the data for secrets that opt-in to the feature. This feature been mulling around in my head ever since @liggitt brought it up in sig-auth. Wanted to write up and formalize the thought.

This notably isn't part of the existing secrets roadmap.[1] Would like to get some review from the people working on that.

[1] https://docs.google.com/document/d/1JAwPuZg47UhfRVlof-lMw08OJztunW8pvTNxDK3rCF8/edit#heading=h.vw9xyk1ib8nn

### DIFF
--- a/contributors/design-proposals/subresourced-secrets.md
+++ b/contributors/design-proposals/subresourced-secrets.md
@@ -1,0 +1,81 @@
+# Subresourced Secret Data
+
+Author: Eric Chiang (eric.chiang@coreos.com)
+
+Note: This is a write up and formalization of an idea @liggitt (Red Hat) had during a SIG-auth meeting.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


liggitt:
After thinking through it more, I don't really think it would work well. Confidential fields would not play well with apply (which copies the entire object into annotations), and subresource aren't addressable by object definitions handled by thinks like `kubectl update`, etc

ericchiang:
>  subresource aren't addressable by object definitions handled by thinks like `kubectl update`, etc

Can you elaborate on that one more? My thought was that write actions such as "update" wouldn't be required to go through the subresource.

> Confidential fields would not play well with apply

As in `kubectl apply`? That seems like something we could special case and work through over an alpha period, though I can see creating new behavior like this is might break assumptions of existing tools.

liggitt:
> My thought was that write actions such as "update" wouldn't be required to go through the subresource.

That gets really weird... should a put with no data remove the secret data? Should a put with one key set remove other unset keys or just override the one set key?

> I can see creating new behavior like this is might break assumptions of existing tools.

Yeah, it's clever, but it runs counter to the rest of the API. I would foresee problems with most things built to interact with the k8s API.

erictune:
@destijl @saad-ali

PR #833, opened 2017-07-22
Dynamic Flexvolume plugin discovery proposal.

verult:
This lays out the design for the improved deployment story of Flexvolume, making it easier to work with out-of-tree volume plugins before CSI comes to fruition.

### DIFF
--- a/contributors/design-proposals/flexvolume-deployment.md
+++ b/contributors/design-proposals/flexvolume-deployment.md
@@ -1,0 +1,148 @@
+# **Automated Flexvolume Deployment**
+
+## **Objective**
+
+Automate the deployment of Flexvolume drivers with the following goals:
+*   Drivers must be deployed on nodes (and master when attach is required) without having to manually access any machine instance.
+*   Kubelet and controller-manager do not need to be restarted manually in order for the new plugin to be recognized.
+
+## **Background**
+
+Beginning in version 1.8, the Kubernetes Storage SIG is putting a stop to accepting in-tree volume plugins and advises all storage providers to implement out-of-tree plugins. Currently, there are two recommended implementations: Container Storage Interface (CSI) and Flexvolume.
+
+[CSI](https://github.com/container-storage-interface/spec/blob/master/spec.md) provides a single interface that storage vendors can implement in order for their storage solutions to work across many different container orchestrators, and volume plugins are out-of-tree by design. This is a large effort, the full implementation of CSI is several quarters away, and there is a need for an immediate solution for storage vendors to continue adding volume plugins.
+
+[Flexvolume](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md) is an in-tree plugin that has the ability to run any storage solution by executing volume commands against a user-provided driver on the Kubernetes host, and this currently exists today. However, the process of setting up Flexvolume is very manual, pushing it out of consideration for many users. Problems include having to copy the driver to a specific location in each node, manually restarting kubelet, and user's limited access to machines.
+
+
+## **Overview**
+
+
+### User Story
+
+Driver Installation:
+
+*   Alice is a storage plugin author and would like to deploy a Flexvolume driver on all node instances. She
+    1. prepares her Flexvolume driver directory, with driver names in `[vendor~]driver/driver` format (e.g. `flexvolume/k8s~nfs/nfs`, see [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)).
+    2. creates an image by copying her driver to the Flexvolume deployment base image at `/flexvolume`.
+    3. makes her image available Bob, a cluster admin.
+*   Bob modifies the existing deployment DaemonSet spec with the name of the given image, and creates the DaemonSet.
+*   Charlie, an end user, creates volumes using the installed plugin.
+
+The user story for driver update is similar: Alice creates a new image with her new drivers, and Bob deploys it using the DaemonSet spec.
+
+Note that the `/flexvolume` directory must look exactly like what is desired in the Flexvolume directory on the host (as described in the [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)). The deployment will replace the existing driver directory on the host with contents in `/flexvolume`. Thus, in order to add a new driver without removing existing ones, existing drivers must also appear in `/flexvolume`.
+
+
+### High Level Design
+
+The DaemonSet mounts a hostpath volume exposing the host's Flexvolume driver directory onto every pod. The base deployment image contains a script that copies drivers in the image to the hostpath. A notification is then sent to the filesystem watch from kubelet or controller manager. During volume creation, if there is a signal from the watch, kubelet or controller manager probes the driver directory and loads currently installed drivers as volume plugins.
+
+
+## **Detailed Design**
+
+### Public Deployment Base Image
+Composed of the busybox image and the deployment script described in [Driver Installation Script](#driver-installation-script).
+
+### Copying Driver File to Image
+
+Using the deployment base image, the plugin author copies the Flexvolume driver directory (e.g. `flexvolume/k8s~nfs/nfs`) to `/flexvolume` and makes the image available to the cluster admin.
+
+
+### Driver Installation Script
+
+The script will copy the existing content of `/flexvolume` on the host to a location in `/tmp`, and then attempt to copy user-provided drivers to that directory. If the copy fails, the original drivers are restored. This script will not perform any driver validation.
+
+### Deployment DaemonSet
+``` yaml
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: flex-set
+spec:
+  template:
+    metadata:
+      name: flex-deploy
+      labels:
+        app: flex-deploy
+    spec:
+      containers:
+        - image: <deployment_image>
+          name: flex-deploy
+          volumeMounts:
+            - mountPath: /flexmnt
+              name: flexvolume-mount
+      volumes:
+        - name: flexvolume-mount
+          hostPath:
+            path: <host_driver_directory>
+```
+
+### Dynamic Plugin Discovery
+
+In the volume plugin code, introduce a `PluginStub` interface containing a single method `Init()`, and have `VolumePlugin` extend it. Create a `PluginProber` type which extends `PluginStub` and includes methods `Init()` and `Probe()`.
+
+`Init()` initializes fsnotify, creates a watch on the driver directory as well as its subdirectories (if any), and spawn a goroutine listening to the signal. When the goroutine receives signal that a new directory is created, create a watch for the directory so that driver changes can be seen.
+
+`Probe()` scans the driver directory only when the goroutine sets a flag. If the flag is set, return true (indicating that new plugins are available) and the list of plugins. Otherwise, return false and nil. After the scan, the watch is refreshed to include the new list of subdirectories. The goroutine should only record a signal if there has been a 1-second delay since the last signal (see [Security Considerations](#security-considerations)). Because inotify (used by fsnotify) can only be used to watch an existing directory, the goroutine needs to maintain the invariant that the driver directory always exists.
+
+Inside `InitPlugins()` from `volume/plugins.go`, if the `PluginStub` is an instance of `PluginProber`, only call its `Init()` and nothing else. Add an additional field, `flexVolumePluginList`, in `VolumePluginMgr` as a cache. For every iteration of the plugin list, call `Probe()` and update `flexVolumePluginList` if true is returned, and iterate through the new plugin list. If the return value is false, iterate through the existing `flexVolumePluginList`.
+
+Because Flexvolume has two separate plugin instantiations (attachable and non-attachable), it's worth considering the case when a driver that implements attach/detach is replaced with a driver that does not, or vice versa. This does not cause an issue because plugins are recreated every time the driver directory is changed.
+
+There is a possibility that a probe occurs at the same time the DaemonSet updates the driver, so the prober's view of drivers is inconsistent. However, this is very rare and when it does occur, the next `Probe()`call, which occurs shortly after, will be consistent.
+
+
+## **Alternative Designs**
+
+1) Make `PluginProber` a separate component, and pass it around as a dependency.
+
+Pros: Avoids the common `PluginStub` interface. There isn't much shared functionality between `VolumePlugin` and `PluginProber`. The only purpose this shared abstraction serves is for `PluginProber` to reuse the existing machinery of plugins list.
+
+Cons: Would have to increase dependency surface area, notably `KubeletDeps`.
+
+I'm currently undecided whether to use this design or the `PluginStub` design.
+
+2) Use a polling model instead of a watch for probing for driver changes.
+
+Pros: Simpler to implement.
+
+Cons: Kubelet or controller manager iterates through the plugin list many times, so Probe() is called very frequently. Using this model would increase unnecessary disk usage. This issue is mitigated if we guarantee that `PluginProber` is the last `PluginStub` in the iteration, and only `Probe()` if no other plugin is matched, but this logic adds additional complexity.
+
+3) Use a polling model + cache. Poll every x seconds/minutes.
+
+Pros: Mostly mitigates issues with the previous approach.
+
+Cons: Depending on the polling period, either it's needlessly frequent, or it's too infrequent to pick up driver updates quickly.
+
+4) Using Jobs instead of DaemonSets to deploy.
+
+Pros: Designed for containers that eventually terminate. No need to have the container go into an infinite loop.
+
+Cons: Does not guarantee every node has a pod running. Pod anti-affinity can be used to ensure no more than one pod runs on the same node, but since the Job spec requests a constant number of pods to run to completion, Jobs cannot ensure that pods are scheduled on new nodes.
+
+5) Have the `flexVolumePluginList` cache live in `PluginProber` instead of `VolumePluginMgr`.
+
+Pros: `VolumePluginMgr` doesn't need to treat Flexvolume plugins any differently from other plugins.
+
+Cons: `PluginProber` doesn't have the function to validate a plugin. This function lives in `VolumePluginMgr`. Alternatively, the function can be passed into `PluginProber`.
+
+
+## **Security Considerations**
+
+The Flexvolume driver directory can be continuously modified (accidentally or maliciously), making every` Probe()` call trigger a disk read, and `Probe()` calls could happen every couple of milliseconds and in bursts (i.e. lots of calls at first and then silence for some time). This may decrease kubelet's or controller manager's disk IO usage, impacting the performance of other system operations.
+
+As a safety measure, add a 1-second minimum delay between the processing of filesystem watch signals.
+
+
+## **Testing Plan**
+
+Add new unit tests in `plugin_tests.go` to cover new probing functionality and the heterogeneous plugin types in the plugins list.
+
+Add e2e tests that follow the user story. Write one for initial driver installation, one for an update for the same driver, one for adding another driver, and one for removing a driver.
+
+## **Open Questions**
+
+* How does this system work with containerized kubelet?
+* What if drivers are updated while Flexvolume plugin is executing commands?
+* If DaemonSet deployment fails, how are errors shown to the user?
--- a/contributors/design-proposals/flexvolume-deployment.md
+++ b/contributors/design-proposals/flexvolume-deployment.md
@@ -1,10 +1,8 @@
-# **Automated Flexvolume Deployment**
+# **Dynamic Flexvolume Plugin Discovery**
 
 ## **Objective**
 
-Automate the deployment of Flexvolume drivers with the following goals:
-*   Drivers must be deployed on nodes (and master when attach is required) without having to manually access any machine instance.
-*   Kubelet and controller-manager do not need to be restarted manually in order for the new plugin to be recognized.
+Kubelet and controller-manager do not need to be restarted manually in order for new Flexvolume plugins to be recognized.
 
 ## **Background**
 
@@ -14,79 +12,23 @@ Beginning in version 1.8, the Kubernetes Storage SIG is putting a stop to accept
 
 [Flexvolume](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md) is an in-tree plugin that has the ability to run any storage solution by executing volume commands against a user-provided driver on the Kubernetes host, and this currently exists today. However, the process of setting up Flexvolume is very manual, pushing it out of consideration for many users. Problems include having to copy the driver to a specific location in each node, manually restarting kubelet, and user's limited access to machines.
 
+An automated deployment technique is discussed in [Recommended Deployment Method](#recommended-driver-deployment-method). The crucial change required to enable this method is allowing kubelet and controller manager to dynamically discover plugin changes.
+
 
 ## **Overview**
 
-
-### User Story
-
-Driver Installation:
-
-*   Alice is a storage plugin author and would like to deploy a Flexvolume driver on all node instances. She
-    1. prepares her Flexvolume driver directory, with driver names in `[vendor~]driver/driver` format (e.g. `flexvolume/k8s~nfs/nfs`, see [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)).
-    2. creates an image by copying her driver to the Flexvolume deployment base image at `/flexvolume`.
-    3. makes her image available Bob, a cluster admin.
-*   Bob modifies the existing deployment DaemonSet spec with the name of the given image, and creates the DaemonSet.
-*   Charlie, an end user, creates volumes using the installed plugin.
-
-The user story for driver update is similar: Alice creates a new image with her new drivers, and Bob deploys it using the DaemonSet spec.
-
-Note that the `/flexvolume` directory must look exactly like what is desired in the Flexvolume directory on the host (as described in the [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)). The deployment will replace the existing driver directory on the host with contents in `/flexvolume`. Thus, in order to add a new driver without removing existing ones, existing drivers must also appear in `/flexvolume`.
-
-
-### High Level Design
-
-The DaemonSet mounts a hostpath volume exposing the host's Flexvolume driver directory onto every pod. The base deployment image contains a script that copies drivers in the image to the hostpath. A notification is then sent to the filesystem watch from kubelet or controller manager. During volume creation, if there is a signal from the watch, kubelet or controller manager probes the driver directory and loads currently installed drivers as volume plugins.
+When there is a modification of the driver directory, a notification is sent to the filesystem watch from kubelet or controller manager. When kubelet or controller-manager searches for plugins (such as when a volume needs to be mounted), if there is a signal from the watch, it probes the driver directory and loads currently installed drivers as volume plugins.
 
 
 ## **Detailed Design**
 
-### Public Deployment Base Image
-Composed of the busybox image and the deployment script described in [Driver Installation Script](#driver-installation-script).
-
-### Copying Driver File to Image
-
-Using the deployment base image, the plugin author copies the Flexvolume driver directory (e.g. `flexvolume/k8s~nfs/nfs`) to `/flexvolume` and makes the image available to the cluster admin.
-
-
-### Driver Installation Script
-
-The script will copy the existing content of `/flexvolume` on the host to a location in `/tmp`, and then attempt to copy user-provided drivers to that directory. If the copy fails, the original drivers are restored. This script will not perform any driver validation.
-
-### Deployment DaemonSet
-``` yaml
-apiVersion: extensions/v1beta1
-kind: DaemonSet
-metadata:
-  name: flex-set
-spec:
-  template:
-    metadata:
-      name: flex-deploy
-      labels:
-        app: flex-deploy
-    spec:
-      containers:
-        - image: <deployment_image>
-          name: flex-deploy
-          volumeMounts:
-            - mountPath: /flexmnt
-              name: flexvolume-mount
-      volumes:
-        - name: flexvolume-mount
-          hostPath:
-            path: <host_driver_directory>
-```
-
-### Dynamic Plugin Discovery
-
-In the volume plugin code, introduce a `PluginStub` interface containing a single method `Init()`, and have `VolumePlugin` extend it. Create a `PluginProber` type which extends `PluginStub` and includes methods `Init()` and `Probe()`.
+In the volume plugin code, introduce a `PluginStub` interface containing a single method `Init()`, and have `VolumePlugin` extend it. Create a `PluginProber` type which extends `PluginStub` and includes methods `Init()` and `Probe()`. Change the type of plugins inside the volume plugin manager's plugin list to `PluginStub`.
 
 `Init()` initializes fsnotify, creates a watch on the driver directory as well as its subdirectories (if any), and spawn a goroutine listening to the signal. When the goroutine receives signal that a new directory is created, create a watch for the directory so that driver changes can be seen.
 
 `Probe()` scans the driver directory only when the goroutine sets a flag. If the flag is set, return true (indicating that new plugins are available) and the list of plugins. Otherwise, return false and nil. After the scan, the watch is refreshed to include the new list of subdirectories. The goroutine should only record a signal if there has been a 1-second delay since the last signal (see [Security Considerations](#security-considerations)). Because inotify (used by fsnotify) can only be used to watch an existing directory, the goroutine needs to maintain the invariant that the driver directory always exists.
 
-Inside `InitPlugins()` from `volume/plugins.go`, if the `PluginStub` is an instance of `PluginProber`, only call its `Init()` and nothing else. Add an additional field, `flexVolumePluginList`, in `VolumePluginMgr` as a cache. For every iteration of the plugin list, call `Probe()` and update `flexVolumePluginList` if true is returned, and iterate through the new plugin list. If the return value is false, iterate through the existing `flexVolumePluginList`.
+Iterating through the list of plugins inside `InitPlugins()` from `volume/plugins.go`, if the plugin is an instance of `PluginProber`, only call its `Init()` and nothing else. Add an additional field, `flexVolumePluginList`, in `VolumePluginMgr` as a cache. For every iteration of the plugin list, call `Probe()` and update `flexVolumePluginList` if true is returned, and iterate through the new plugin list. If the return value is false, iterate through the existing `flexVolumePluginList`.
 
 Because Flexvolume has two separate plugin instantiations (attachable and non-attachable), it's worth considering the case when a driver that implements attach/detach is replaced with a driver that does not, or vice versa. This does not cause an issue because plugins are recreated every time the driver directory is changed.
 
@@ -115,13 +57,7 @@ Pros: Mostly mitigates issues with the previous approach.
 
 Cons: Depending on the polling period, either it's needlessly frequent, or it's too infrequent to pick up driver updates quickly.
 
-4) Using Jobs instead of DaemonSets to deploy.
-
-Pros: Designed for containers that eventually terminate. No need to have the container go into an infinite loop.
-
-Cons: Does not guarantee every node has a pod running. Pod anti-affinity can be used to ensure no more than one pod runs on the same node, but since the Job spec requests a constant number of pods to run to completion, Jobs cannot ensure that pods are scheduled on new nodes.
-
-5) Have the `flexVolumePluginList` cache live in `PluginProber` instead of `VolumePluginMgr`.
+4) Have the `flexVolumePluginList` cache live in `PluginProber` instead of `VolumePluginMgr`.
 
 Pros: `VolumePluginMgr` doesn't need to treat Flexvolume plugins any differently from other plugins.
 
@@ -141,8 +77,64 @@ Add new unit tests in `plugin_tests.go` to cover new probing functionality and t
 
 Add e2e tests that follow the user story. Write one for initial driver installation, one for an update for the same driver, one for adding another driver, and one for removing a driver.
 
+## **Recommended Driver Deployment Method**
+
+This section describes one possible method to automatically deploy Flexvolume drivers. The goal is that drivers must be deployed on nodes (and master when attach is required) without having to manually access any machine instance.
+
+Driver Installation:
+
+*   Alice is a storage plugin author and would like to deploy a Flexvolume driver on all node instances. She
+    1. prepares her Flexvolume driver directory, with driver names in `[vendor~]driver/driver` format (e.g. `k8s~nfs/nfs`, see [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)).
+    2. creates an image by copying her driver and the [deployment script](#driver-deployment-script) to a busybox base image.
+    3. makes her image available Bob, a cluster admin.
+*   Bob modifies the existing deployment DaemonSet spec with the name of the given image, and creates the DaemonSet.
+*   Charlie, an end user, creates volumes using the installed plugin.
+
+The user story for driver update is similar: Alice creates a new image with her new drivers, and Bob deploys it using the DaemonSet spec.
+
+Note that the `/flexvolume` directory must look exactly like what is desired in the Flexvolume directory on the host (as described in the [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)). The deployment will replace the existing driver directory on the host with contents in `/flexvolume`. Thus, in order to add a new driver without removing existing ones, existing drivers must also appear in `/flexvolume`.
+
+### Driver Deployment Script
+
+The script will copy the existing content of `/flexvolume` on the host to a location in `/tmp`, and then attempt to copy user-provided drivers to that directory. If the copy fails, the original drivers are restored. This script will not perform any driver validation.
+
+### Deployment DaemonSet
+``` yaml
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: flex-set
+spec:
+  template:
+    metadata:
+      name: flex-deploy
+      labels:
+        app: flex-deploy
+    spec:
+      containers:
+        - image: <deployment_image>
+          name: flex-deploy
+          securityContext:
+              privileged: true
+          volumeMounts:
+            - mountPath: /flexmnt
+              name: flexvolume-mount
+      volumes:
+        - name: flexvolume-mount
+          hostPath:
+            path: <host_driver_directory>
+```
+
+### Alternatives
+
+* Using Jobs instead of DaemonSets to deploy.
+
+Pros: Designed for containers that eventually terminate. No need to have the container go into an infinite loop.
+
+Cons: Does not guarantee every node has a pod running. Pod anti-affinity can be used to ensure no more than one pod runs on the same node, but since the Job spec requests a constant number of pods to run to completion, Jobs cannot ensure that pods are scheduled on new nodes.
+
 ## **Open Questions**
 
 * How does this system work with containerized kubelet?
-* What if drivers are updated while Flexvolume plugin is executing commands?
 * If DaemonSet deployment fails, how are errors shown to the user?
+* Are there any SELinux implications?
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+### Dynamic Plugin Discovery
+
+Create a `PluginStub` interface containing a single method `Init()`, and have `VolumePlugin` extend it. Create a `PluginProber` type which extends `PluginStub` and includes methods `Init()` and `Probe()`. `Init()` initializes fsnotify, creates a watch on the driver directory as well as its subdirectories (if any), and spawn a goroutine listening to the signal. When the goroutine receives signal that a new directory is created, create a watch for the directory so that driver changes can be seen. `Probe()` scans the driver directory only when the goroutine sets a flag. If the flag is set, return true (indicating that new plugins are available) and the list of plugins. Otherwise, return false and nil. After the scan, the watch is refreshed to include the new list of subdirectories. The goroutine should only record a signal if there has been a 1-second delay since the last signal (see Security Considerations). Because inotify (used by fsnotify) can only be used to watch an existing directory, the goroutine needs to maintain the invariant that the driver directory always exists.
```
msau42: Who creates this interface?

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,111 @@
...
+
+The DaemonSet mounts a hostpath volume exposing the host's Flexvolume driver directory onto every pod. The base deployment image contains a script that copies drivers in the image to the hostpath. A notification is then sent to the filesystem watch from kubelet or controller manager. During volume creation, if there is a signal from the watch, kubelet or controller manager probes the driver directory and loads currently installed drivers as volume plugins.
+
+
+## **Detailed Design**
+
+
+### Copying Driver File to Image
+
+Using a publicly available deployment image containing the script as the base image, the plugin author copies the Flexvolume driver directory (e.g. `flexvolume/k8s~nfs/nfs`) to `/flexvolume` and makes the image available to the cluster admin.
```
msau42: Maybe also have a section for the public deployment base image.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,111 @@
...
+
+Inside `InitPlugins()` from `volume/plugins.go`, if the `PluginStub` is an instance of `PluginProber`, only call its `Init()` and nothing else. Add an additional field, `flexVolumePluginList`, in `VolumePluginMgr` as a cache. For every iteration of the plugin list, call `Probe()` and update `flexVolumePluginList` if true is returned, and iterate through the new plugin list. If the return value is false, iterate through the existing `flexVolumePluginList`.
+
+Because Flexvolume has two separate plugin instantiations (attachable and non-attachable), it's worth considering the case when a driver that implements attach/detach is replaced with a driver that does not, or vice versa. This does not cause an issue because plugins are recreated every time the driver directory is changed.
+
+There is a possibility that a probe occurs at the same time the DaemonSet updates the driver, so that the prober's view of drivers is inconsistent. However, this is very rare and when it does occur, the next `Probe()`call, which occurs shortly after, will be consistent.
```
msau42: There are some parts of the volume code that iterate through all the volume plugins.  This will also need to deal with synchronization with those calls.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+### Copying Driver File to Image
+
+Using a publicly available deployment image containing the script as the base image, the plugin author copies the Flexvolume driver directory (e.g. `flexvolume/k8s~nfs/nfs`) to `/flexvolume` and makes the image available to the cluster admin.
+
+
+### Driver Installation Script
+
+The script will copy the existing content of `/flexvolume` on the host to a location in `/tmp`, and then attempt to copy user-provided drivers to that directory. If the copy fails, the original drivers are restored. This script will not perform any driver validation.
```
verult: Not sure what's the best way. I'll add that to Open Questions.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,124 @@
...
+
+Inside `InitPlugins()` from `volume/plugins.go`, if the `PluginStub` is an instance of `PluginProber`, only call its `Init()` and nothing else. Add an additional field, `flexVolumePluginList`, in `VolumePluginMgr` as a cache. For every iteration of the plugin list, call `Probe()` and update `flexVolumePluginList` if true is returned, and iterate through the new plugin list. If the return value is false, iterate through the existing `flexVolumePluginList`.
+
+Because Flexvolume has two separate plugin instantiations (attachable and non-attachable), it's worth considering the case when a driver that implements attach/detach is replaced with a driver that does not, or vice versa. This does not cause an issue because plugins are recreated every time the driver directory is changed.
```
kokhang: If the flexvolume implements attach/detach interface, how are you going to extend the controller-manager (assuming you are using a daemonset)?

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,124 @@
...
+
+Pros: Mostly mitigates issues with the previous approach.
+
+Cons: Depending on the polling period, either it's needlessly frequent, or it's too infrequent to pick up driver updates quickly.
+
+4) Using Jobs instead of DaemonSets to deploy.
+
+Pros: Designed for containers that eventually terminate. No need to have the container go into an infinite loop.
+
+Cons: Does not guarantee every node has a pod running.
```
kokhang: Can we use affinity/placement to target all nodes? Although you would still have issues if new nodes are added to the k8s cluster.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,124 @@
...
+
+Cons: Would have to increase dependency surface area, notably `KubeletDeps`.
+
+I'm currently undecided whether to use this design or the `PluginStub` design.
+
+2) Use a polling model instead of a watch for probing for driver changes.
+
+Pros: Simpler to implement.
+
+Cons: Kubelet or controller manager iterates through the plugin list many times, so Probe() is called very frequently. Using this model would increase unnecessary disk usage. This issue is mitigated if we guarantee that `PluginProber` is the last `PluginStub` in the iteration, and only `Probe()` if no other plugin is matched, but this logic adds additional complexity.
```
kokhang: I agree. Flexvolume drivers are not something that will be added or modified often.

contributors/design-proposals/flexvolume-deployment.md
verult: What do you mean by "extend"?

If you are asking how the drivers get deployed through DaemonSet to a master: DaemonSet will add a pod to the master node regardless of whether the node is set to schedulable.

If you are asking how the controller-manager picks up the newly added attach/detach procedures when the driver is replaced: during plugin probe a FlexVolumeAttachablePlugin is created, which replaces the previous plugin. The attachable plugin will enable attach/detach calls from AttachDetachController.

contributors/design-proposals/flexvolume-deployment.md
verult: That's a good point, I'll add it to the doc. Yes like you said, with Jobs you have to specify how many pods you'd like to see running to completion, which is a problem when new nodes come up.

verult:
Just to clarify: the containerized mounts feature doesn't install mount utilities on the host directly; rather, it keeps them in a container, and volume plugins need to do a container exec to access them.

The deployment script in this design can be composed with the mount utility container, and the DaemonSet spec here can be combined with the containerized mounts DaemonSet spec. The Flexvolume drivers will have to be able to perform the container execs to access containerized mount utilities, but otherwise it should just work.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+*   Charlie, an end user, creates volumes using the installed plugin.
+
+The user story for driver update is similar: Alice creates a new image with her new drivers, and Bob deploys it using the DaemonSet spec.
+
+Note that the `/flexvolume` directory must look exactly like what is desired in the Flexvolume directory on the host (as described in the [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)). The deployment will replace the existing driver directory on the host with contents in `/flexvolume`. Thus, in order to add a new driver without removing existing ones, existing drivers must also appear in `/flexvolume`.
+
+
+### High Level Design
+
+The DaemonSet mounts a hostpath volume exposing the host's Flexvolume driver directory onto every pod. The base deployment image contains a script that copies drivers in the image to the hostpath. A notification is then sent to the filesystem watch from kubelet or controller manager. During volume creation, if there is a signal from the watch, kubelet or controller manager probes the driver directory and loads currently installed drivers as volume plugins.
```
gnufied: Why does the flexvolume content have to be copied to *every* pod? Assuming non-containarized kubelet, shouldn't this just be on host?

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+
+`Probe()` scans the driver directory only when the goroutine sets a flag. If the flag is set, return true (indicating that new plugins are available) and the list of plugins. Otherwise, return false and nil. After the scan, the watch is refreshed to include the new list of subdirectories. The goroutine should only record a signal if there has been a 1-second delay since the last signal (see [Security Considerations](#security-considerations)). Because inotify (used by fsnotify) can only be used to watch an existing directory, the goroutine needs to maintain the invariant that the driver directory always exists.
+
+Inside `InitPlugins()` from `volume/plugins.go`, if the `PluginStub` is an instance of `PluginProber`, only call its `Init()` and nothing else. Add an additional field, `flexVolumePluginList`, in `VolumePluginMgr` as a cache. For every iteration of the plugin list, call `Probe()` and update `flexVolumePluginList` if true is returned, and iterate through the new plugin list. If the return value is false, iterate through the existing `flexVolumePluginList`.
```
gnufied: > if the `PluginStub` is an instance of `PluginProber`

I thought both `PluginStub` and `PluginProber` were interfaces. Do you mean if plugin in `InitPlugins` loop is an instance of `PluginProber` ? I assume all Volume plugins will be instance of `PluginStub` since it would be the type that would be extended by `VolumePlugin`.

gnufied: okay, so the idea is, each time any function on `VolumePluginMgr` is executed, we call `Probe` on plugin and if that returns true we rescan the directories and return with updated plugin list.  I am wondering, if we have can simply rescan the directory always whenever a call to `VolumePluginMgr` is made, that is keep the polling on-demand rather than periodic and we won't need inotify etc hook.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+## **Overview**
+
+
+### User Story
+
+Driver Installation:
+
+*   Alice is a storage plugin author and would like to deploy a Flexvolume driver on all node instances. She
+    1. prepares her Flexvolume driver directory, with driver names in `[vendor~]driver/driver` format (e.g. `flexvolume/k8s~nfs/nfs`, see [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)).
+    2. creates an image by copying her driver to the Flexvolume deployment base image at `/flexvolume`.
```
liggitt: it might simplify this proposal to focus it on dynamic discovery of new plugins, and update of existing plugins. deployment via daemonset could be one mechanism, but shouldn't be prescriptive.

contributors/design-proposals/flexvolume-deployment.md
chakri-nelluri: Daemonset has to be privileged to copy the driver.

contributors/design-proposals/flexvolume-deployment.md
chakri-nelluri: Can we use the exit status from Daemonset to notify the error?

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+    3. makes her image available Bob, a cluster admin.
+*   Bob modifies the existing deployment DaemonSet spec with the name of the given image, and creates the DaemonSet.
+*   Charlie, an end user, creates volumes using the installed plugin.
+
+The user story for driver update is similar: Alice creates a new image with her new drivers, and Bob deploys it using the DaemonSet spec.
+
+Note that the `/flexvolume` directory must look exactly like what is desired in the Flexvolume directory on the host (as described in the [Flexvolume documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites)). The deployment will replace the existing driver directory on the host with contents in `/flexvolume`. Thus, in order to add a new driver without removing existing ones, existing drivers must also appear in `/flexvolume`.
```
liggitt: > in order to add a new driver without removing existing ones, existing drivers must also appear in `/flexvolume`

that is unexpected. I'd expect a flexvolume deployment to be focused on installing a single driver, e.g. populating one driver under `/usr/libexec/kubernetes/kubelet-plugins/volume/exec/<vendor>~<driver>/<driver>`

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+
+Cons: Kubelet or controller manager iterates through the plugin list many times, so Probe() is called very frequently. Using this model would increase unnecessary disk usage. This issue is mitigated if we guarantee that `PluginProber` is the last `PluginStub` in the iteration, and only `Probe()` if no other plugin is matched, but this logic adds additional complexity.
+
+3) Use a polling model + cache. Poll every x seconds/minutes.
+
+Pros: Mostly mitigates issues with the previous approach.
+
+Cons: Depending on the polling period, either it's needlessly frequent, or it's too infrequent to pick up driver updates quickly.
+
+4) Using Jobs instead of DaemonSets to deploy.
```
liggitt: deployment method should be a separated out of this proposal, IMO. The core need is for kubelet and controller manager to  detect and use added flex plugins without restart. If that requirement is met, you can use all sorts of methods for installing/maintaining the drivers.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+
+Inside `InitPlugins()` from `volume/plugins.go`, if the `PluginStub` is an instance of `PluginProber`, only call its `Init()` and nothing else. Add an additional field, `flexVolumePluginList`, in `VolumePluginMgr` as a cache. For every iteration of the plugin list, call `Probe()` and update `flexVolumePluginList` if true is returned, and iterate through the new plugin list. If the return value is false, iterate through the existing `flexVolumePluginList`.
+
+Because Flexvolume has two separate plugin instantiations (attachable and non-attachable), it's worth considering the case when a driver that implements attach/detach is replaced with a driver that does not, or vice versa. This does not cause an issue because plugins are recreated every time the driver directory is changed.
+
+There is a possibility that a probe occurs at the same time the DaemonSet updates the driver, so the prober's view of drivers is inconsistent. However, this is very rare and when it does occur, the next `Probe()`call, which occurs shortly after, will be consistent.
```
chakri-nelluri: +1 for this approach.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+Composed of the busybox image and the deployment script described in [Driver Installation Script](#driver-installation-script).
+
+### Copying Driver File to Image
+
+Using the deployment base image, the plugin author copies the Flexvolume driver directory (e.g. `flexvolume/k8s~nfs/nfs`) to `/flexvolume` and makes the image available to the cluster admin.
+
+
+### Driver Installation Script
+
+The script will copy the existing content of `/flexvolume` on the host to a location in `/tmp`, and then attempt to copy user-provided drivers to that directory. If the copy fails, the original drivers are restored. This script will not perform any driver validation.
```
chakri-nelluri: Also, what happens if the probe fails or the plugin fails to initialize?
We might have to document how to revert to an working older version.

chakri-nelluri: Please add this is a recommended way. Users are free to innovate :)

msau42: We're planning to provide a driver installation mechanism as part of this proposal, but like @liggitt mentioned, it can be separated out.

contributors/design-proposals/flexvolume-deployment.md
thockin: I don't understand the "are restored" ?  A driver should never ever touch anything that isn't it's own.

The install process, for most drivers, should be:

```
while true; do 
  if [ -f /flexvolume/myvendor~mydriver/mydriver ]; then
    sleep 60
  else
    mkdir /flexvolume/myvendor~mydriver
    cp -a /mydriver /flexvolume/myvendor~mydriver
  fi
done
```

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+        - name: flexvolume-mount
+          hostPath:
+            path: <host_driver_directory>
+```
+
+### Dynamic Plugin Discovery
+
+In the volume plugin code, introduce a `PluginStub` interface containing a single method `Init()`, and have `VolumePlugin` extend it. Create a `PluginProber` type which extends `PluginStub` and includes methods `Init()` and `Probe()`.
+
+`Init()` initializes fsnotify, creates a watch on the driver directory as well as its subdirectories (if any), and spawn a goroutine listening to the signal. When the goroutine receives signal that a new directory is created, create a watch for the directory so that driver changes can be seen.
```
thockin: why use notify, if you are just going to cache the results?

contributors/design-proposals/flexvolume-deployment.md
thockin: I don't think anything related to flex should leak outside the flex directory.  It should just be part of the initial handshake to figure out if each thing is a plugin or a meta-plugin.  Meta-plugins are probed later.

contributors/design-proposals/flexvolume-deployment.md
thockin: Why is probe called so often?

contributors/design-proposals/flexvolume-deployment.md
thockin: Agree.  We should provide an example. That's all.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+
+`Probe()` scans the driver directory only when the goroutine sets a flag. If the flag is set, return true (indicating that new plugins are available) and the list of plugins. Otherwise, return false and nil. After the scan, the watch is refreshed to include the new list of subdirectories. The goroutine should only record a signal if there has been a 1-second delay since the last signal (see [Security Considerations](#security-considerations)). Because inotify (used by fsnotify) can only be used to watch an existing directory, the goroutine needs to maintain the invariant that the driver directory always exists.
```
jsafrane: So, mere presence of a new directory means that there is a new flex driver there? File copy is not an atomic operation and it may happen that only part of the driver script was copied there. You should wait until whole driver has been copied there. And big question is how do you recognize that...

IMO, installation of a new driver (or driver update) should be atomic operation on the fs, e.g. link(2).

contributors/design-proposals/flexvolume-deployment.md
jsafrane: > DaemonSet will add a pod to the master node

Not all masters run kubelet. E.g. GKE.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+## **Testing Plan**
+
+Add new unit tests in `plugin_tests.go` to cover new probing functionality and the heterogeneous plugin types in the plugins list.
+
+Add e2e tests that follow the user story. Write one for initial driver installation, one for an update for the same driver, one for adding another driver, and one for removing a driver.
+
+## **Open Questions**
+
+* How does this system work with containerized kubelet?
+* What if drivers are updated while Flexvolume plugin is executing commands?
```
jsafrane: This is very important question and I'd like to see some answer. Especially kubelet must not ever execute a half-copied driver - running partly copied shell script as root sounds scary to me.

contributors/design-proposals/flexvolume-deployment.md
thockin: Jan, excellent point.  The only atomic file op is rename, so we should be sure that flex EXPLICITLY ignores files that start with a `.`.  The installer must copy to `flex/.../.mydriver` and then rename that to `mydriver`.  Let's make this as explicit as possible.

contributors/design-proposals/flexvolume-deployment.md
thockin: Rename, as described above, should not suffer this problem, I think.

contributors/design-proposals/flexvolume-deployment.md
jsafrane: That assumes that the driver is a single file. The installer must rename multiple files one by one (and `rename` does not work well with directories).

I am ok with requiring the driver to be a single file, with no helper scripts around, but it must be clearly defined somewhere in flex documentation and a release note.

contributors/design-proposals/flexvolume-deployment.md
verult: Other than the scenario Jan mentioned, there are two other possible bad scenarios:
1) When a driver command is called, part of the driver command is loaded into memory and is being executed. The driver gets modified, and the remaining part of the command (which is new) gets executed. I'm not sure if this could happen.
2) The driver gets modified while the command is read into memory, before execution even occurs. This could definitely happen.

I don't think renaming solves either of these issues. A write should never happen while an execution is underway. Ideally the solution fixes both the issues above and Jan's scenario.

My first thought is a RWLock implementation, but I'd like to avoid locking if possible.

contributors/design-proposals/flexvolume-deployment.md
verult: It's called for every FindPluginBy*(), because currently it iterates through *all* plugins and errors out when multiple plugins are found. As for why FindPluginBy*() is called often, I don't know. I can only tell that from logs.

contributors/design-proposals/flexvolume-deployment.md
verult: The model in the current design is the entire Flexvolume directory containing all drivers get replaced, so when a copy fails, the backup of the entire directory is restored.

After moving to the single-driver install model, this will be changed to backing up a single driver instead.

thockin:
Rename does work with symlinks, though.  It's ugly but doable.  mydriver ->
.mydriver/mydriver

contributors/design-proposals/flexvolume-deployment.md
thockin: I think the hard line is at "you should never ever touch something you don't own".

contributors/design-proposals/flexvolume-deployment.md
thockin: we should track that backwards - it is surprising to me.

contributors/design-proposals/flexvolume-deployment.md
thockin: If probe() were called a reasonable amount of times per second, would you reconsider this point?  Just trying to bound complexity.

contributors/design-proposals/flexvolume-deployment.md
thockin: 1) should get an ETXTBUSY for the person trying to modify the binary

I don't understand 2?

In any case, rename is atomic.  As soon as I open the file descriptor, if anyone renames another file over this file, I still get the old file.  It is pinned until I close that fd.  Happy to explain more - it's a good trick to have in your toolbox.

contributors/design-proposals/flexvolume-deployment.md
php-coder: @chakri-nelluri Could you clarify which exactly operations require privileged mode?

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,148 @@
...
+Cons: Does not guarantee every node has a pod running. Pod anti-affinity can be used to ensure no more than one pod runs on the same node, but since the Job spec requests a constant number of pods to run to completion, Jobs cannot ensure that pods are scheduled on new nodes.
+
+5) Have the `flexVolumePluginList` cache live in `PluginProber` instead of `VolumePluginMgr`.
+
+Pros: `VolumePluginMgr` doesn't need to treat Flexvolume plugins any differently from other plugins.
+
+Cons: `PluginProber` doesn't have the function to validate a plugin. This function lives in `VolumePluginMgr`. Alternatively, the function can be passed into `PluginProber`.
+
+
+## **Security Considerations**
```
php-coder: Could we also mention how it's supposed to work with [Pod Security Policy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/)? Which actions will be required? Do we need to create a special role maybe?

liggitt: Not sure that level of detail is required... you'd need whatever permissions are required to mount and write files to a directory in the kubelet's flexvolume driver dir. Again, I think this proposal should focus on the kubelet/controllermanager aspects, and the mechanism for ensuring atomic load of drivers (write to dot-prefixed dir or file, rename or symlink+rename, etc), over delivery mechanisms.

php-coder: > you'd need whatever permissions are required to mount and write files to a directory in the kubelet's flexvolume driver dir

That's what I wanted to see here. If it would be explicitly mentioned, it will be good.

contributors/design-proposals/flexvolume-deployment.md
verult: That's really neat. In that case yes renaming will solve both of these scenarios.

What I tried to describe in (2) is: before the driver gets executed, part of its commands are read into memory, then the remaining commands are modified, and lastly the modified commands are read in.

contributors/design-proposals/flexvolume-deployment.md
verult: Yes definitely. Unfortunately I was seeing bursts of tens of milliseconds between calls. If we change the Find*() logic so that if there's already a match then don't check Flex, then we don't need a watch.

verult:
Changed focus of the document to dynamic plugin discovery, and moved DaemonSet deployment to a separate section.

Will incorporate DaemoSet deployment comments in another commit.

verult:
@thockin Not sure if this is what you are suggesting, but when the script installs a driver, it creates a symlink mydriver -> .<some-random-string>_mydriver/mydriver. If mydriver already exists, it'll just replace the symlink.

I would hope replacing a link is atomic?

jsafrane:
> creates a symlink `mydriver` -> `.[some-random-string]_mydriver/mydriver`

`symlink(2)` syscall can't overwrite existing symlink, you need to `unlink` it first. And thus it's not really atomic.

What is atomic is to create a symlink `.[some-random-string]-aspiring-symlink` -> `.[some-random-string]_mydriver/mydriver` and then rename `.[some-random-string]-aspiring-symlink` -> `mydriver/mydriver`. Rename is atomic.

I am a bit scared by these random directories popping up in `/usr/libexec`, I hope somebody will clear them... But that's just an implementation detail.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,140 @@
...
+
+An automated deployment technique is discussed in [Recommended Deployment Method](#recommended-driver-deployment-method). The crucial change required to enable this method is allowing kubelet and controller manager to dynamically discover plugin changes.
+
+
+## **Overview**
+
+When there is a modification of the driver directory, a notification is sent to the filesystem watch from kubelet or controller manager. When kubelet or controller-manager searches for plugins (such as when a volume needs to be mounted), if there is a signal from the watch, it probes the driver directory and loads currently installed drivers as volume plugins.
+
+
+## **Detailed Design**
```
jsafrane: (a random point in the proposal to start a new separate thread so the discussion is organized)

# Mount propagation

I am preparing mount propagation for release 1.8 + possibility to run mount utilities (`mount.glusterfs`, `/usr/bin/rbd`, flex drivers, ...) in *pods* instead of running them on the *host*. Initially, I planned not to support flex volumes in the first release and get the implementation solid first, but now that there is someone else preparing dynamic probing of flex plugin it should not be hard to extend it.

See https://github.com/kubernetes/community/pull/589 for full details.

tl;dr, the proposal expects that pod with mount utilities will put a unix domain socket into `/var/lib/kubelet/plugin-sockets/<plugin name>`, e.g. `/var/lib/kubelet/plugin-sockets/kubernetes.io/glusterfs`. That can be easily extended to flex volumes:

* Flex volume author puts all necessary utilities, usual `/usr/libexec/kubernetes/kubelet-plugins/volume/exec/<vendor>~<driver>/<driver>` and a `volume-exec` daemon (shipped by Kubernetes) into an image.
* System admin runs the image as a DaemonSet with privileged pods that have shared mount-propagation on `/var/lib/kubelet` from the pod to the host. Pods in the daemon set run `volume-exec` daemon with proper parameters (namely the volume plugin name).
  * `volume-exec` daemon puts an unix socket into `/var/lib/kubelet/plugin-sockets/<vendor>~<driver>/<driver>` on the host.
* Probe in kubelet scans `plugin-sockets` and registers a new flex volume plugin for every socket it finds there. This is the same as discovery of new drivers as designed in this proposal.
* When kubelet wants to call the driver, it checks if `plugin-sockets/<vendor>~<driver>/<driver>` socket exists.
  * If if the socket does not exist, it uses plain old `os,.Exec` to execute the driver.
  * If the socket exists, it uses a gRPC API provided by `volume-exec` to execute stuff in the pod that runs `volume-exec`. As result, **`/usr/libexec/kubernetes/kubelet-plugins/volume/exec/<vendor>~<driver>/<driver>` is executed in the pod**. Due to shared mount propagation, the driver can run in the pod and kubelet will all mounts it creates.
  * (all this is already part of #589, there will be very little changes to flex volume implementation).

Installation of a driver is quite simple, no need to copy the driver from the pod to the host, creating one socket is enough. Upgrade of the daemon set is more complicated though, as any fuse daemons now run in the pod  and not on the host. If the pod is killed during the update all volumes that it served are unmounted. See #589 for details.

To sum it up, if the probe as suggested in this proposal is implemented, it should be fairly easy to extend it to scan also for the sockets. That's the only necessary change and we get completely containerized flex volumes. Question is if we then need to mess up with copying drivers from pods to the host and such. On the other way, #589 is just a proposal and it may be changed and who knows if it catches 1.8 and in what shape.

contributors/design-proposals/flexvolume-deployment.md
verult: I added the necessary security policy in the example Deployment DaemonSet spec.

contributors/design-proposals/flexvolume-deployment.md
verult: Seems like there are (at least) two reasons:

1) In every DSW populator loop, it's called for volumes of every existing pod, in order to re-associate each pod with its required volumes.

2) Certain volume types are set to constantly remount, triggering a Probe every time the remount occurs.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,161 @@
...
+
+while : ; do
+  sleep 3600
+done
+```
+
+### Deployment DaemonSet
+``` yaml
+apiVersion: extensions/v1beta1
+kind: DaemonSet
```
bassam: will this daemonset run on master nodes too by default?

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,161 @@
...
+          name: flex-deploy
+          securityContext:
+              privileged: true
+          volumeMounts:
+            - mountPath: /flexmnt
+              name: flexvolume-mount
+      volumes:
+        - name: flexvolume-mount
+          hostPath:
+            path: <host_driver_directory>
```
bassam: It would be good if the author of this daemonset does not need to make assumptions about the host environment. For example, if `--volume-plugin-dir` varies for the nodes in the cluster it would still be good to author a single daemonset that install the flex volume correctly.

Could we make `--volume-plugin-dir` available in the downstream api so that it can be used as an env variable in the daemonset spec?

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+### Atomic Driver Installation
+Regular file copy is not an atomic file operation, so if it were used to install the driver, it's possible that kubelet or controller manager executes the driver when it's partially installed, or the driver gets modified while it's being executed. Care must be taken to ensure the installation operation is atomic.
+
+The deployment script provided above uses renaming, which is atomic, to ensure that from the perspective of kubelet or controller manager, the driver file is completely written to disk in a single operation.
+
+### Alternatives
```
bassam: I wonder if init containers could be used too? For rook we plan on deploying a daemonset for a rook agent on each node, could we use an init container to deploy the flex volume? this would avoid the infinite loop.

kokhang: I havent used init-containers. Are init-containers spawned and run on the same host as the parent daemonset?

contributors/design-proposals/flexvolume-deployment.md
bassam: We should consider the versioning the flex volume drivers. It's likely that the flex volume interface would change in future versions, or that the flex volume author has different implementations. Is it possible to get the version of K8S running through downward api? This could be passed to the flex volume deploy script for it to install the correct version.

Similarly when kubernetes is upgraded, I'm not sure I understand how flex volumes are upgraded. Consider the case where we have a 1.8 cluster with a flex volume designed for 1.8. If in 1.9, K8S adds a new method to flex, and a cluster is upgraded from 1.8-->1.9 how does the new flex volume get deployed?

bassam:
When the kubelet execs code in the flex volume, will the flex volume have a security context to call the K8S API if it needs to like in the case of the rook volume plugin wanting to create CRDs through flex?

contributors/design-proposals/flexvolume-deployment.md
saad-ali: Not unless the master node is registered as schedulable.

contributors/design-proposals/flexvolume-deployment.md
saad-ali: Could be a useful improvement for a subsequent iteration.

saad-ali: +@chakri-nelluri

contributors/design-proposals/flexvolume-deployment.md
saad-ali: Ya, theoretically that should be possible!

saad-ali:
> When the kubelet execs code in the flex volume, will the flex volume have a security context to call the K8S API if it needs to like in the case of the rook volume plugin wanting to create CRDs through flex?

I believe so. Since it is running on the host, you should be able to use the same security context as the kubelet. Not pretty, but it should work. Alternatively, (and possibly a cleaner approach) would be to have the credentials shared during deployment: have the user create a secret with the user/token for accessing API server, pass that secret in to the DeamonSet (or InitContainer) that you use for deploying the Flexvolume, and have the deployment script inject the credentials for the script to use.

bassam:
> Since it is running on the host, you should be able to use the same security context as the kubelet.

How can the flex volume impersonate the kubelet? Is there a service account env variable set? That would be ideal. otherwise we're back to credential management and pushing this problem on to users.

contributors/design-proposals/flexvolume-deployment.md
bassam: without this I'm worried that the flex volume will end up in a different place, and there will be little in terms of diagnosis for what actually happened. Note also that this could help the kubelet in container problem.

contributors/design-proposals/flexvolume-deployment.md
msau42: I'm not sure if this is actually a bug, but I noticed that my local storage daemonset does attempt to get scheduled on the master node, even though it is unschedulable.  And it fails to run on master because there are not enough resources.

kokhang: @verult can you confirm? I think this PR made daemonset respect taint. https://github.com/kubernetes/kubernetes/pull/41172

saad-ali:
> How can the flex volume impersonate the kubelet? Is there a service account env variable set? That would be ideal. otherwise we're back to credential management and pushing this problem on to users.

Use the kubelet `/var/lib/kubelet/kubeconfig` files

kokhang:
> Use the kubelet /var/lib/kubelet/kubeconfig files
@saad-ali, Is this consistent across all Kubernetes deployments? I ask because I have been testing with `minikube` and `./hack/local-up-cluster.sh` and they dont have a `/var/lib/kubelet/kubeconfig` file.

contributors/design-proposals/flexvolume-deployment.md
verult: Tested the DaemonSet with a master node having more memory than usual, and yes it does run on master by default, even though it's marked as SchedulingDisabled.

liggitt:
> Since it is running on the host, you should be able to use the same security context as the kubelet. Not pretty, but it should work.

You'd have to make assumptions about (or allow configuring) where on disk the kubelet credentials were. Kubelets wouldn't have permission to make API requests to the custom API resources by default anyway.

> Alternatively, (and possibly a cleaner approach) would be to have the credentials shared during deployment: have the user create a secret with the user/token for accessing API server, pass that secret in to the DeamonSet (or InitContainer) that you use for deploying the Flexvolume, and have the deployment script inject the credentials for the script to use.

If you wanted to distribute API credentials as part of the deployment, a pod doing a deployment of a flex driver already has service account API credentials available to it. Leaving those accessible to the flex driver would probably be more predictable than trying to access the kubelet credentials.

bassam:
> If you wanted to distribute API credentials as part of the deployment, a pod doing a deployment of a flex driver already has service account API credentials available to it. Leaving those accessible to the flex driver would probably be more predictable than trying to access the kubelet credentials.

so essentially save the service account credentials to a .dot file (or subdir) and have the flex volume driver load them from there. Are there cases where the service account credentials are updated/rotated that we'd need to deal with?

kokhang:
> so essentially save the service account credentials to a .dot file (or subdir) and have the flex volume driver load them from there. Are there cases where the service account credentials are updated/rotated that we'd need to deal with?

I tried doing this but i got an error because the K8S API IP was a in-network service IP and not routable by the host.

verult:
If we go with the route of having a separate credentials file, then like @jsafrane and @thockin mentioned before, we need to make sure that driver command execution doesn't interleave with driver installation. The deployment script in this proposal assumes there is a single file inside the driver directory.

This problem only comes up if the crendentials file needs to be changed. One solution is to install each file individually (copy to .dot file & rename) rather than installing the entire directory all at once.

By the way, the credentials file doesn't have to have a .dot prefix. The Flexvolume plugin uses the driver directory name to compute the name of the driver file.

contributors/design-proposals/flexvolume-deployment.md
verult: Do you have a specific example in mind where --volume-plugin-dir has to be set differently between nodes?

bassam: @verult no specific example, just pointing out that if it is set then the daemonset would break. I'm hoping we can arrive at a "universal" daemonset that can install the flex volume independent of how the host is configured.

verult: Yes that's a good point. For most OS distros it's fine to use the default plugin path. `--volume-plugin-dir` is usually set to a non-default value due to filesystem limitations. I think this scenario is rare enough that if it does occur, the cluster admin could resort to deploying individual pods, or pods + Daemonset with taints.

AFAIK downward API only exposes pod and container attributes, not kubelet options.

verult:
@kokhang Have you tried using the master IP directly?

MrHohn:
> I tried doing this but i got an error because the K8S API IP was a in-network service IP and not routable by the host.

Kubernetes service VIP should be routable on the host as well, assuming you have kube-proxy running on the same host?

kokhang:
@verult @MrHohn  How do i know what the master IP is?  
I have been testing on Minikube. What i tried was constructing the kubeconfig file using information from the daemonset that contains the flex driver. I tried getting the serviceaccount certs, the kubernetes host and port by doing something similar to https://github.com/kubernetes/client-go/blob/master/rest/config.go#L310.

However on minikube, the IP i get from the KUBERNETES_SERVICE_HOST env is not routable.

MrHohn:
> How do i know what the master IP is?

@kokhang For Minikube, probably try `minikube ip`.

liggitt:
>> How do i know what the master IP is?
>
> @kokhang For Minikube, probably try minikube ip.

The point is that flex plugins need a reliable way to know how to reach the API server. If they have to try to figure out what type of kube deployment they are on, they'll do it inconsistently at best and fail to handle some kube deployments at worst.

In-process volume plugins are given an API client (via the host interface) as part of their `Init()` call. Passing information about the API server to the flexplugin's init() call seems like a parallel to that. I could see the following possibilities:
* no api server info (kubelet is running in standalone mode)
* full path to a kubeconfig file containing api server address, CA, credentials, etc.

for now, the kubeconfig file could be the same as the one the kubelet uses, but if a need for a separate identity came up in the future, a different kubeconfig could be fed to the plugin

saad-ali:
We had a productive meeting today.

We identified three problems for the Rook Flex Driver:
1. Discovery of k8s API endpoint by Rook Flex Drive
2. Credentials Rook Flex Driver should use against k8s API endpoint.
3. Discovery of Flex directory that deployment DeamonSet pod should drop the Rook Flex Driver in (since this directory is configurable).

Agreed upon solution:
* Rook has a "proxy" pod deployed as a DeamonSet that proxies calls between the Flex volume driver and the k8s API server. The proxy exposes a Unix Domain Socket via HostPath to the host that the Flex Driver talks to.
  * With this approach, inside the pod, the Rook driver can do DNS lookup (kube DNS) to discover k8s API endpoint in a reliable manner, and it can the pod can run with a service account that has the correct permission to create CRDs.
* Rook will investigate using the node API to read kubelet `configz` and extract the Flex driver path for kubelets.

Once @verult's dynamic discovery of Flex drivers work is merged, we should add E2E test for Flex to prevent regressions (accidental changes to the Flex API).

verult:
A PR for e2e Flexvolume test was recently merged and is in the process of being added to the testgrid: https://github.com/kubernetes/kubernetes/pull/48366

Currently it uses kubelet restart. Once dynamic discovery is finished, this can be changed to make the test non-disruptive.

contributors/design-proposals/flexvolume-deployment.md
chakri-nelluri: @bassam Versioning came up in new flex-volume proposal which never materialized. I will look into adding support to negotiate versioning with the existing API.

contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -0,0 +1,161 @@
...
+
+Because Flexvolume has two separate plugin instantiations (attachable and non-attachable), it's worth considering the case when a driver that implements attach/detach is replaced with a driver that does not, or vice versa. This does not cause an issue because plugins are recreated every time the driver directory is changed.
+
+There is a possibility that a Flexvolume command execution occurs at the same time as the DaemonSet updates the driver, which leads to a bad execution. This cannot be solved within the Kubernetes system without an overhaul. Instead, this is discussed in [Atomic Driver Installation](#atomic-driver-installation) as part of the deployment mechanism. As part of the solution, the Prober will ignore all files that begins with "." in the driver directory.
```
adelton: This is the first time DaemonSet was mentioned and it caught me by surprise that it is being referred to as **the** DaemonSet.

Could the proposal be extended to describe that DaemonSet is expected to be the mechanism to run on every node something which will then copy scripts to the host, before discussing the race condition?

liggitt: Or avoid discussing specific deployment mechanisms here and just say "…possibility that a Flexvolume command execution occurs at the same time as the driver is updated…"

contributors/design-proposals/flexvolume-deployment.md
saad-ali: This would be cool. Can consider it for 1.9

verult:
I think @jsafrane 's proposal would be a natural evolution of Flex. The dynamic plugin discover work here will build towards that goal, offering the ability to discover unix domain socket on the fly.

contributors/design-proposals/flexvolume-deployment.md
verult: Updated. Thanks for catching this!

saad-ali:
Merging this.

contributors/design-proposals/flexvolume-deployment.md
verult: @php-coder Kubelet executes Flexvolume drivers outside the container's namespace, so driver files have to be copied into the hostpath volume in privileged mode.

PR #836, opened 2017-07-23
update .1 to 100m of core resource

zhangxiaoyu-zidif:
Add 0


mikedanese:
Why?

derekwaynecarr:
either form is correct, not sure i understand, unless the idea is to have them all use consistent notation in example.

zhangxiaoyu-zidif:
I think it is a result from linux terminal. According to current show, I only find 100 mili ,not .1
So I think it need modify. If I am wrong, please close this PR at any time. Thanks =) 
@derekparker @mikedanese

PR #837, opened 2017-07-24
HugePages proposal

derekwaynecarr:
A pod may request a number of huge pages.  The `scheduler` is able to place the pod on a node that can satisfy that request.  The `kubelet` advertises an allocatable number of huge pages to support scheduling decisions. A pod may consume hugepages via `hugetlbfs` or `shmget`.  

Planned as Alpha for Kubernetes 1.8 release.

See feature: https://github.com/kubernetes/features/issues/275

@kubernetes/sig-scheduling-feature-requests
@kubernetes/sig-scheduling-misc
@kubernetes/api-approvers
@kubernetes/api-reviewers

### DIFF
--- a/contributors/design-proposals/hugepages.md
+++ b/contributors/design-proposals/hugepages.md
@@ -1,0 +1,308 @@
+# HugePages support in Kubernetes
+
+**Authors**
+* Derek Carr (@derekwaynecarr)
+* Seth Jennings (@sjenning)
+* Piotr Prokop (@PiotrProkop)
+
+**Status**: In progress
+
+## Abstract
+
+A proposal to enable applications running in a Kubernetes cluster to use huge
+pages.
+
+A pod may request a number of huge pages.  The `scheduler` is able to place the
+pod on a node that can satisfy that request.  The `kubelet` advertises an
+allocatable number of huge pages to support scheduling decisions. A pod may
+consume hugepages via `hugetlbfs` or `shmget`.  Huge pages are not
+overcommitted.
+
+## Motivation
+
+Memory is managed in blocks known as pages.  On most systems, a page is 4Ki. 1Mi
+of memory is equal to 256 pages; 1Gi of memory is 256,000 pages, etc. CPUs have
+a built-in memory management unit that manages a list of these pages in
+hardware. The Translation Lookaside Buffer (TLB) is a small hardware cache of
+virtual-to-physical page mappings.  If the virtual address passed in a hardware
+instruction can be found in the TLB, the mapping can be determined quickly.  If
+not, a TLB miss occurs, and the system falls back to slower, software based
+address translation.  This results in performance issues.  Since the size of the
+TLB is fixed, the only way to reduce the chance of a TLB miss is to increase the
+page size.
+
+A huge page is a memory page that is larger than 4Ki.  On x86_64 architectures,
+there are two common huge page sizes: 2Mi and 1Gi.  Sizes vary on other
+architectures, but the idea is the same.  In order to use huge pages,
+application must write code that is aware of them.  Transparent huge pages (THP)
+attempts to automate the management of huge pages without application knowledge,
+but they have limitations.  In particular, they are limited to 2Mi page sizes.
+THP might lead to performance degradation on nodes with high memory utilization
+or fragmentation due to defragmenting efforts of THP, which can lock memory
+pages. For this reason, some applications may be designed to (or recommend)
+usage of pre-allocated huge pages instead of THP.
+
+Managing memory is hard, and unfortunately, there is no one-size fits all
+solution for all applications.
+
+## Scope
+
+This proposal only includes pre-allocated huge pages configured on the node by
+the administrator at boot time or by manual dynamic allocation.  It does not
+discuss how the cluster could dynamically attempt to allocate huge pages in an
+attempt to find a fit for a pod pending scheduling.  It is anticipated that
+operators may use a variety of strategies to allocate huge pages, but we do not
+anticipate the kubelet itself doing the allocation.  Allocation of huge pages
+ideally happens soon after boot time.
+
+This proposal defers issues relating to NUMA.
+
+## Use Cases
+
+The class of applications that benefit from huge pages typically have
+- A large memory working set
+- A sensitivity to memory access latency
+
+Example applications include:
+- database management systems (MySQL, PostgreSQL, MongoDB, Oracle, etc.)
+- Java applications can back the heap with huge pages using the
+  `-XX:+UseLargePages` and `-XX:LagePageSizeInBytes` options.
+- packet processing systems (DPDK)
+
+Applications can generally use huge pages by calling
+- `mmap()` with `MAP_ANONYMOUS | MAP_HUGETLB` and use it as anonymous memory
+- `mmap()` a file backed by `hugetlbfs`
+- `shmget()` with `SHM_HUGETLB` and use it as a shared memory segment (see Known
+  Issues).
+
+1. A pod can use huge pages with any of the prior described methods.
+1. A pod can request huge pages.
+1. A scheduler can bind pods to nodes that have available huge pages.
+1. A quota may limit usage of huge pages.
+1. A limit range may constrain min and max huge page requests.
+
+## Feature Gate
+
+The proposal introduces huge pages as an Alpha feature.
+
+It must be enabled via the `--feature-gates=HugePages=true` flag on pertinent
+components pending graduation to Beta.
+
+## Node Specfication
+
+Huge pages cannot be overcommitted on a node.
+
+A system may support multiple huge page sizes.  It is assumed that most nodes
+will be configured to primarily use the default huge page size as returned via
+`grep Hugepagesize /proc/meminfo`.  This defaults to 2Mi on most Linux systems
+unless overriden by `default_hugepagesz=1g` in kernel boot parameters.  
+
+For each supported huge page size, the node will advertise a resource of the
+form `hugepages-<hugepagesize>`.  On Linux, supported huge page sizes are
+determined by parsing the `/sys/kernel/mm/hugepages/hugepages-{size}kB`
+directory on the host. Kubernetes will expose a `hugepages-<hugepagesize>`
+resource using binary notation form. It will convert `<hugepagesize>` into the
+most compact binary notation using integer values.  For example, if a node
+supports `hugepages-2048kB`, a resource `hugepages-2Mi` will be shown in node
+capacity and allocatable values. Operators may set aside pre-allocated huge
+pages that are not available for user pods similar to normal memory via the
+`--system-reserved` flag.
+
+There are a variety of huge page sizes supported across different hardware
+architectures.  It is preferred to have a resource per size in order to better
+support quota.  For example, 1 huge page with size 2Mi is orders of magnitude
+different than 1 huge page with size 1Gi.  We assume gigantic pages are even
+more precious resources than huge pages.
+
+Pre-allocated huge pages reduce the amount of allocatable memory on a node. The
+node will treat pre-allocated huge pages similar to other system reservations
+and reduce the amount of `memory` it reports using the following formula:
+
+```
+[Allocatable] = [Node Capacity] - 
+ [Kube-Reserved] - 
+ [System-Reserved] - 
+ [Pre-Allocated-HugePages * HugePageSize] -
+ [Hard-Eviction-Threshold]
+```
+
+The following represents a machine with 10Gi of memory.  1Gi of memory has been
+reserved as 512 pre-allocated huge pages sized 2Mi.  As you can see, the
+allocatable memory has been reduced to account for the amount of huge pages
+reserved.
+
+```
+apiVersion: v1
+kind: Node
+metadata:
+  name: node1
+...
+status:
+  capacity:
+    memory: 10Gi
+    hugepages-2Mi: 1Gi
+  allocatable:
+    memory: 9Gi
+    hugepages-2Mi: 1Gi
+...  
+```
+
+## Pod Specification
+
+A pod must make a request to consume pre-allocated huge pages using the resource
+`hugepages-<hugepagesize>` whose quantity is a positive amount of memory in
+bytes.  The specified amount must align with the `<hugepagesize>`; otherwise,
+the pod will fail validation.  For example, it would be valid to request
+`hugepages-2Mi: 4Mi`, but invalid to request `hugepages-2Mi: 3Mi`.
+
+The request and limit for `hugepages-<hugepagesize>` must match.  Similar to
+memory, an application that requests `hugepages-<hugepagesize>` resource is at
+minimum in the `Burstable` QoS class.
+
+If a pod consumes huge pages via `shmget`, it must run with a supplemental group
+that matches `/proc/sys/vm/hugetlb_shm_group` on the node.  Configuration of
+this group is outside the scope of this specification.
+
+Initially, a pod may not consume multiple huge page sizes in a single pod spec.
+Attempting to use `hugepages-2Mi` and `hugepages-1Gi` in the same pod spec will
+fail validation.  We believe it is rare for applications to attempt to use
+multiple huge page sizes. This restriction may be lifted in the future with
+community presented use cases.  Introducing the feature with this restriction
+limits the exposure of API changes needed when consuming huge pages via volumes.
+
+In order to consume huge pages backed by the `hugetlbfs` filesystem inside the
+specified container in the pod, it is helpful to understand the set of mount
+options used with `hugetlbfs`.  For more details, see "Using Huge Pages" here:
+https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt
+
+```
+mount -t hugetlbfs \
+	-o uid=<value>,gid=<value>,mode=<value>,pagesize=<value>,size=<value>,\
+	min_size=<value>,nr_inodes=<value> none /mnt/huge
+```
+
+The proposal recommends extending the existing `EmptyDirVolumeSource` to satisfy
+this use case.  A new `medium=HugePages` option would be supported.  To write
+into this volume, the pod must make a request for huge pages. The `pagesize`
+argument is inferred from the `hugepages-<hugepagesize>` from the resource
+request.  If in the future, multiple huge page sizes are supported in a single
+pod spec, we may modify the `EmptyDirVolumeSource` to provide an optional page
+size.  The existing `sizeLimit` option for `emptyDir` would restrict usage to
+the minimum value specified between `sizeLimit` and the sum of huge page limits
+of all containers in a pod. This keeps the behavior consistent with memory
+backed `emptyDir` volumes whose usage is ultimately constrained by the pod
+cgroup sandbox memory settings.  The `min_size` option is omitted as its not
+necessary.  The `nr_inodes` mount option is omitted at this time in the same
+manner it is omitted with `medium=Memory` when using `tmpfs`.
+
+The following is a sample pod that is limited to 1Gi huge pages of size 2Mi. It
+can consume those pages using `shmget()` or via `mmap()` with the specified
+volume.
+
+```
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example
+spec:
+  containers:
+...
+    volumeMounts:
+    - mountPath: /hugepages
+      name: hugepage
+    resources:
+      requests:
+        hugepages-2Mi: 1Gi
+      limits:
+        hugepages-2Mi: 1Gi
+  volumes:
+  - name: hugepage
+    emptyDir:
+      medium: HugePages
+```
+
+## CRI Updates
+
+The `LinuxContainerResources` message should be extended to support specifying
+huge page limits per size.  The specification for huge pages should align with
+opencontainers/runtime-spec.
+
+see:
+https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#huge-page-limits
+
+The CRI changes are required before promoting this feature to Beta.
+
+## Cgroup Enforcement
+
+To use this feature, the `--cgroups-per-qos` must be enabled.  In addition, the
+`hugetlb` cgroup must be mounted.
+
+The `kubepods` cgroup is bounded by the `Allocatable` value.
+
+The QoS level cgroups are left unbounded across all huge page pool sizes.
+
+The pod level cgroup sandbox is configured as follows, where `hugepagesize` is
+the system supported huge page size(s).  If no request is made for huge pages of
+a particular size, the limit is set to 0 for all supported types on the node.
+
+```
+pod<UID>/hugetlb.<hugepagesize>.limit_in_bytes = sum(pod.spec.containers.resources.limits[hugepages-<hugepagesize>])
+```
+
+If the container runtime supports specification of huge page limits, the
+container cgroup sandbox will be configured with the specified limit.
+
+The `kubelet` will ensure the `hugetlb` has no usage charged to the pod level
+cgroup sandbox prior to deleting the pod to ensure all resources are reclaimed.
+
+## Limits and Quota
+
+The `ResourceQuota` resource will be extended to support accounting for
+`hugepages-<hugepagesize>` similar to `cpu` and `memory`.  The `LimitRange`
+resource will be extended to define min and max constraints for `hugepages`
+similar to `cpu` and `memory`.
+
+## Scheduler changes
+
+The scheduler will need to ensure any huge page request defined in the pod spec
+can be fulfilled by a candidate node.
+
+## cAdvisor changes
+
+cAdvisor will need to be modified to return the number of pre-allocated huge
+pages per page size on the node.  It will be used to determine capacity and
+calculate allocatable values on the node.
+
+## Roadmap
+
+### Version 1.8
+
+Initial alpha support for huge pages usage by pods.
+
+### Version 1.9
+
+Resource Quota support. Limit Range support. Beta support for huge pages
+(pending community feedback)
+
+## Known Issues
+
+### Huge pages as shared memory
+
+For the Java use case, the JVM maps the huge pages as a shared memory segment
+and memlocks them to prevent the system from moving or swapping them out.
+
+There are several issues here:
+- The user running the Java app must be a member of the gid set in the
+  `vm.huge_tlb_shm_group` sysctl
+- sysctl `kernel.shmmax` must allow the size of the shared memory segment
+- The user's memlock ulimits must allow the size of the shared memory segment
+- `vm.huge_tlb_shm_group` is not namespaced.
+
+### NUMA
+
+NUMA is complicated.  To support NUMA, the node must support cpu pinning,
+devices, and memory locality.  Extending that requirement to huge pages is not
+much different.  It is anticipated that the `kubelet` will provide future NUMA
+locality guarantees as a feature of QoS.  In particular, pods in the
+`Guaranteed` QoS class are expected to have NUMA locality preferences.
+


contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,278 @@
...
+
+Pre-allocated huge pages reduce the amount of allocatable memory on a node. The
+node will treat pre-allocated huge pages similar to other system reservations
+and reduce the amount of `memory` it reports using the following formula:
+
+```
+[Allocatale] = [Node Capacity] - 
+ [Kube-Reserved] - 
+ [System-Reserved] - 
+ [Pre-Allocated-HugePages * HugePageSize]
```
davidopp: I think there's a minus missing at the end of this line

derekwaynecarr:
I may choose to use explicit huge page sizes after more reflection. Want to experiment with a prototype some today.  Difference between 2MB and 1Gi pages is significant enough that I would want to quota them separately.  The empty dir topic would then become a separate volume type with pageSize option.

jeremyeder:
Two questions:

* eviction-hard only looks at the non-hugepages allocatable remaining, right?  IOW if there are plenty of free hugepages but the non-hugepages amount is getting low...eviction would be triggered despite free hugepages.  Assume that's the case, but just checking.
* The kubelet will set the node label automatically according to what it discovers?
* Will any hugepage info show up in cadvisor?

derekwaynecarr:
@jeremyeder 

* eviction-hard - if there are plenty of free hugepages, but normal memory is low, eviction is triggered.
* node label - yes, the kubelet woudl set it, but the more i think on this, the more i think we should just have the huge page size explicit in the resource name
* cadvisor will report hugepages allocated per size; i will add a section for monitoring, i suspect we would want to see hugepage usage per container/pod cgroup.

jeremyeder:
* node label - explicit name...makes sense since there are so few possibilities and even if the list grows it will still be manageable.
* cadvisor -- in addition to per-container/cgroup, perhaps it can also report the hugepages allocatable/capacity for each node.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+## Motivation
+
+Memory is managed in blocks known as pages.  On most systems, a page is 4KB. 1MB
+of memory is equal to 256 pages; 1GB of memory is 256,000 pages, etc.  CPUs have
+a built-in memory management unit that manages a list of these pages in
+hardware. The Translation Lookaside Buffer (TLB) is a small hardware cache of
+virtual-to-physical page mappings.  If the virtual address passed in a hardware
+instruction can be found in the TLB, the mapping can be determined quickly.  If
+not, a TLB miss occurs, and the system falls back to slower, software based
+memory management.  This results in performance issues.  Since the hardware is
```
sjenning: nit: s/memory management/address translation/

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+fixed, the only way to alleviate the performance impact of a TLB miss is to
```
sjenning: s/alleviate the performance impact/reduce the chance/

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+increase the page size.
+
+A huge page is a memory page that is larger than 4KB.  On x86_64 architectures,
+there are two huge page sizes: 2MB and 1GB.  Sizes vary on other architectures,
+but the idea is the same.  In order to use huge pages, application must write
+code that is aware of them.  Transparent huge pages (THP) attempt to automate
```
sjenning: nit: s/attempt/attempts/

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+the management of huge pages without application knowledge, but they have
+limitations.  In particular, they are limited to 2MB page sizes.
```
sjenning: Also, dynamic splitting of THPs into standard pages during certain conditions can result in undesirable application stalls.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+components pending graduation to Beta.
+
+## Node Specfication
+
+Huge pages cannot be overcommitted on a node.
+
+A system may support multiple huge page sizes.  It is assumed that most
+nodes will be configured to primarily use the default huge page size as
+returned via `grep Hugepagesize /proc/meminfo`.  This defaults to 2MB on
+most Linux systems unless overriden by `default_hugepagesz=1g` in kernel
```
sjenning: nit s/overriden/overridden/

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+boot parameters.  This design does not limit the ability to use multiple
+huge page sizes at the same time, but we expect this to not be the normal
+setup.  For each supported huge page size, the node will advertise a
```
sjenning: nit: not sure if we need this sentence (starting with "This design..")  i found it confusing i.e. you can, but maybe you shouldn't because... reasons?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+  allocatable:
+    memory: 9Gi
+    hugepages.2MB: 512
+...  
+```
+
+## Pod Specification
+
+A pod must make a request to consume pre-allocated huge pages using the resource
+`hugepages.<hugepagesize>` whose quantity is a non-negative number of pages. The
```
sjenning: s/non-negative/positive integer/

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+request and limit for `hugepages.<hugepagesize>` must match until such time the
+node provides more reliability in handling overcommit (potentially via
+eviction).  An application that consumes `hugepages.<hugepagesize>` is not in
+the `BestEffort` QoS class.
```
sjenning: not sure if we need this sentence.  If we do need it, i'd go with

An application that consumes `hugepages.<hugepagesize>` is, by definition, not in the `BestEffort` QoS class as it specifies a request for one or more resources.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+mount -t hugetlbfs \
+	-o uid=<value>,gid=<value>,mode=<value>,pagesize=<value>,size=<value>,\
+	min_size=<value>,nr_inodes=<value> none /mnt/huge
+```
+
+The proposal recommends a new volume type `hugePages`.  It will support a single
+`pageSize` option that must be provided by users.  The `pageSize` option aligns
+with the `pagesize` argument in the above mount.  It is used to specify the huge
+page size and pool that is used for accounting.  To write into this volume, the
+pod must make a request for huge pages that match this `pagesize`, otherwise its
```
sjenning: s/pod/container/ ?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#huge-page-limits
+
+The CRI changes are required before promoting this feature to Beta.
+
+## Cgroup Enforcement
+
+To use this feature, the `--cgroups-per-qos` must be enabled.  In addition, the
+`hugetlb` cgroup must be mounted.
+
+The QoS level cgroups are left unbounded across all huge page pool sizes.
```
sjenning: do we want to bound at the kubepods (node) level in case system services are using hugepages?

ConnorDoyle:
cc @PiotrProkop

contributors/design-proposals/hugepages.md
ConnorDoyle: also nit: s/the hardware is fixed/the size of the TLB is fixed/

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+A system may support multiple huge page sizes.  It is assumed that most
+nodes will be configured to primarily use the default huge page size as
+returned via `grep Hugepagesize /proc/meminfo`.  This defaults to 2MB on
+most Linux systems unless overriden by `default_hugepagesz=1g` in kernel
+boot parameters.  This design does not limit the ability to use multiple
+huge page sizes at the same time, but we expect this to not be the normal
+setup.  For each supported huge page size, the node will advertise a
+resource of the form `hugepages.<hugepagesize>`.  For example, if a node
+supports 2MB huge pages, a resource `hugepages.2MB` will be shown in node
+capacity and allocatable values.
```
ConnorDoyle: Does allocatable == capacity in all cases for hugepages resources, or are node-allocatable enhancements also in scope? The text describes how allocatable memory is affected by hugepages, but not how the operator could set aside pre-allocated hugepages that are not available for user pods.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+better support quota.  For example, 1 huge page with size 2MB is an order
+of magnitude different than 1 huge page with size 1GB.  We assume gigantic
+pages are even more precious resources than huge pages.
+
+Pre-allocated huge pages reduce the amount of allocatable memory on a node. The
+node will treat pre-allocated huge pages similar to other system reservations
+and reduce the amount of `memory` it reports using the following formula:
+
+```
+[Allocatale] = [Node Capacity] -
```
ConnorDoyle: s/Allocatale/Allocatable

contributors/design-proposals/hugepages.md
ConnorDoyle: +1 for rewording

contributors/design-proposals/hugepages.md
ConnorDoyle: volumes are per-pod

ConnorDoyle:
Do proposals usually link to the feature ([here](https://github.com/kubernetes/features/issues/275))?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+There are several issues here:
+- The user running the Java app must be a member of the gid set in the
+  `vm.huge_tlb_shm_group` sysctl
+- sysctl `kernel.shmmax` must allow the size of the shared memory segment
+- The user's memlock ulimits must allow the size of the shared memory segment
+- `vm.huge_tlb_shm_group` is not namespaced.
+
+### NUMA
+
+NUMA is complicated.  To support NUMA, the node must support cpu pinning,
```
ConnorDoyle: Agree with waiting to solve NUMA. Could we add a plan for how this will work as a pre-requisite for beta? I have some concerns related to non-fungibility of pages from separate NUMA nodes.

derekwaynecarr: how will the plan differ from normal memory?

ConnorDoyle: ¯\_(ツ)_/¯

contributors/design-proposals/hugepages.md
derekwaynecarr: I have been told that mixed use of 2MB and 1GB pages is possible on a node.

contributors/design-proposals/hugepages.md
derekwaynecarr: Good point.  I will add text.  I see no reason why a reservation would be a problem at first thought.

contributors/design-proposals/hugepages.md
derekwaynecarr: Idea at time is not all resources impact QoS (ie OIR) and placement in cgroup hierarchy, but this would.  Will reason if it was confusing.

contributors/design-proposals/hugepages.md
jeremyeder: Don't think I've seen it.  Mostly because the use-cases for 1GB pages make co-locating other apps that use 2MB pages on the same hardware architecturally undesirable (from a deployment standpoint).

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,281 @@
...
+page size and pool that is used for accounting.  To write into this volume, the
+pod must make a request for huge pages that match this `pagesize`, otherwise its
+usage will be limited to 0 bytes.  No other option from the mount is carried
+over to the `hugePages` volume type.  The `size` option is defaulted to sum of
+`hugepages.<hugepagesize>` requests made by each container for the specified
+`pagesize`.  This keeps the behavior consistent with memory backed `emptyDir`
+volumes whose usage is ultimately constrained by the pod cgroup sandbox memory
+settings.  The `min_size` option is omitted as its not necessary. It's possible
+we could have extended `emptyDir` to support an additional option for
+`medium=Memory` but extend it to support a `pageSize` paramter, but this
```
tanshanshan: s/paramter/parameter

contributors/design-proposals/hugepages.md
sjenning: i guess i was thinking requests where per container.  i can read it both ways now.

contributors/design-proposals/hugepages.md
sjenning: For the record, yes mixing is possible for the three way to explicitly use huge pages.
https://gist.github.com/sjenning/b6bed5bf029c9fd6f078f76b37f0a73f#consuming-hugepages

hugetlbfs - pagesize mount option
mmap() - MAP_HUGE_2MB|MAP_HUGE_1GB
shmget() - SHM_HUGE_2MB, SHM_HUGE_1GB

derekwaynecarr: removed to avoid confusion.

contributors/design-proposals/hugepages.md
derekwaynecarr: fixed

contributors/design-proposals/hugepages.md
derekwaynecarr: added.

derekwaynecarr:
Updated based on feedback.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,282 @@
...
+support quota.  For example, 1 huge page with size 2MB is an order of magnitude
+different than 1 huge page with size 1GB.  We assume gigantic pages are even
+more precious resources than huge pages.
+
+Pre-allocated huge pages reduce the amount of allocatable memory on a node. The
+node will treat pre-allocated huge pages similar to other system reservations
+and reduce the amount of `memory` it reports using the following formula:
+
+```
+[Allocatable] = [Node Capacity] -
```
vishh: Why is this necessary? Doesn't node capacity already get reduced whenever static huge pages are created? I thought `/proc/meminfo` reduces total memory available in the presence of static huge pages.

derekwaynecarr: it does not.  `MemTotal` is fixed in `/proc/meminfo` even after allocating hugepages.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,282 @@
...
+unless overriden by `default_hugepagesz=1g` in kernel boot parameters.  For each
+supported huge page size, the node will advertise a resource of the form
+`hugepages.<hugepagesize>`.  For example, if a node supports 2MB huge pages, a
+resource `hugepages.2MB` will be shown in node capacity and allocatable values.
+Operators may set aside pre-allocated huge pages that are not available for user
+pods similar to normal memory via the `--system-reserved` flag.
+
+There are a variety of huge page sizes supported across different hardware
+architectures.  It is preferred to have a resource per size in order to better
+support quota.  For example, 1 huge page with size 2MB is an order of magnitude
```
thockin: 3 orders of magnitude, actually

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,282 @@
...
+```
+apiVersion: v1
+kind: Node
+metadata:
+  name: node1
+...
+status:
+  capacity:
+    memory: 10Gi
+    hugepages.2MB: 512
```
vishh: Can we avoid exposing size in the API for now? In the future, we do plan on supporting attributes for resources. Size can be an attribute then. Until attributes are available, would node labels suffice?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,282 @@
...
+...  
+```
+
+## Pod Specification
+
+A pod must make a request to consume pre-allocated huge pages using the resource
+`hugepages.<hugepagesize>` whose quantity is a positive number of pages. The
+request and limit for `hugepages.<hugepagesize>` must match until such time the
+node provides more reliability in handling overcommit (potentially via
+eviction).  An application that consumes `hugepages.<hugepagesize>` is not in
```
thockin: WRT class, are you saying that using hugepages gets me an automatic priority boost?  Why?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,282 @@
...
+options used with `hugetlbfs`.  For more details, see "Using Huge Pages" here:
+https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt
+
+```
+mount -t hugetlbfs \
+	-o uid=<value>,gid=<value>,mode=<value>,pagesize=<value>,size=<value>,\
+	min_size=<value>,nr_inodes=<value> none /mnt/huge
+```
+
+The proposal recommends a new volume type `hugePages`.  It will support a single
```
thockin: better as a new volume or as a twist on emptyDir, `medium=HugePages, hugePagesConfig: { pageSize: 2Mi }` ?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,282 @@
...
+`pageSize` option that must be provided by users.  The `pageSize` option aligns
```
thockin: does it matter that resource.Quantity is spelled `2Mi` but the resource name above is `2MB` ?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,282 @@
...
+pod must make a request for huge pages that match this `pagesize`, otherwise its
+usage will be limited to 0 bytes.  No other option from the mount is carried
+over to the `hugePages` volume type.  The `size` option is defaulted to sum of
+`hugepages.<hugepagesize>` requests made by each container for the specified
+`pagesize`.  This keeps the behavior consistent with memory backed `emptyDir`
+volumes whose usage is ultimately constrained by the pod cgroup sandbox memory
+settings.  The `min_size` option is omitted as its not necessary. It's possible
+we could have extended `emptyDir` to support an additional option for
+`medium=Memory` but extend it to support a `pageSize` parameter, but this
+proposal prefers to create an additional volume type instead.
```
thockin: Care to explain why?  I am open to discussion, but I want to know the rationale here.

@saad-ali

contributors/design-proposals/hugepages.md
vishh: +1 for a twist of EmptyDir. We already deal with local storage that is split between EmptyDir (volumes abstraction), container logs and overlay filesystems.

vishh:
I added a few comments. @derekwaynecarr explained offline that hugepages are supported across a few public clouds in addition to on-prem data centers. That alleviates my portability concerns around APIs.

I do not have sufficient bandwidth and @dchen1107 does. She will help review this feature.

contributors/design-proposals/hugepages.md
sjenning: I think this is more: by definition, an application that consumes hugepages in not in the BestEffort QoS class because it must have a resource request for hugepages, making it at least Burstable (Guaranteed if all resources are specified and requests == limits)

contributors/design-proposals/hugepages.md
ConnorDoyle: What would be the semantics of the request if size were not in the name? It seems to me there are two ways to do that:

1. Use `bytes` as the resource quantity unit instead of `pages`. This may require requests to be quantized to a multiple of available page size. What if the page sizes are heterogeneous across the set of nodes?
1. Continue to use `pages` as the resource quantity unit, and you just get the node's default page size whatever that may be (if you care, influence the scheduler with node label affinity.)

(2) above seems better than (1) to me. However neither option significantly reduces changes that pod authors would have to make once resource attributes are available. _Caveat, this is without having reviewed a concrete proposal for resource attributes._

contributors/design-proposals/hugepages.md
ConnorDoyle: From a user perspective, probably no material difference. If EmptyDir is easier for the project to manage then :+1:

contributors/design-proposals/hugepages.md
derekwaynecarr: (2) doesn't satisfy the quota concerns.  It makes quota on pages useless.   This is why I settled on the name proposed.

contributors/design-proposals/hugepages.md
derekwaynecarr: I am fine with emptyDir.  The rationale at time was code complexity, but it's probably simpler API to keep with emptyDir

contributors/design-proposals/hugepages.md
derekwaynecarr: If no objection, I am happy to extend emptyDir as described in earlier comment.  The life cycle requirements were same as memory backed emptyDir

contributors/design-proposals/hugepages.md
vishh: A better compute resource API can handle quota as well. Can the quota requirement be deferred?

contributors/design-proposals/hugepages.md
PiotrProkop: +1 for EmptyDir as the only memory backed volume in k8s.

contributors/design-proposals/hugepages.md
derekwaynecarr: basically what seth said.

contributors/design-proposals/hugepages.md
derekwaynecarr: @thockin - at time of writing, i had mapped pageSize to align with the OCI runtime-spec pageSize field for huge page limits, so it didn't need to be a literal resource.Quantity.

see: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#huge-page-limits

it seems more common to use decimal rather than binary units when describing the pageSize, but i think it should be fine if we want to support binary units only.

derekwaynecarr: looking at this closer in runc, supported hugepage sizes are determined here:
https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/utils.go#L408
https://github.com/opencontainers/runc/blob/master/vendor/github.com/docker/go-units/size.go#L73

it appears to ultimately be parsing the representation returned from the `sys/kernel/mm/hugepages/hugepages-{size}kB` in kibibytes, so binary notation in kubernetes representation should be fine...

derekwaynecarr: actually, this is useful reference:
https://en.wikipedia.org/wiki/Page_%28computer_memory%29#Page_size_trade-off

the actual page sizes are binary notation.  even reading 2MByte on intel chipset was 2^21 and 4MByte was 2^22 as expected.  binary format is actually preferred.  (see pg 109 https://software.intel.com/sites/default/files/managed/a4/60/325384-sdm-vol-3abcd.pdf for sample reference.)

derekwaynecarr:
@thockin @saad-ali @vishh @dchen1107 -- updated to respond to feedback.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+1. A pod can request huge pages.
+1. A scheduler can bind pods to nodes that have available huge pages.
+1. A quota may limit usage of huge pages.
+1. A limit range may constrain min and max huge page requests.
+
+## Feature Gate
+
+The proposal introduces huge pages as an Alpha feature.
+
+It must be enabled via the `--feature-gates=HugePages=true` flag on pertient
```
zhouhaibing089: pertinent?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+    resources:
+      requests:
+        hugepages.2Mi: "512"
+      limits:
+        hugepages.2Mi: "512"
+  volumes:
+  - name: hugepage
+    emptyDir:
+      medium: HugePages
+      pageSize: "2Mi"
```
vishh: If we treat each hugepage size as a separate resource, that's one thing. But parsing the resource name seems hacky.
Can we instead have emptydir volumes uses whatever hugepage resource was specified as part of the container specification and also require all containers to specify the same type of hugepage resource?
For the latter it may help having a different resource name like `hugepages.kubernetes.io/2Mi`.
Or the other option is to have `medium` be the hugepage resource type itself (`hugepages.2Mi`). `memory` medium has a corresponding `memory` compute resource.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+container cgroup sandbox will be configured with the specified limit.
+
+The `kubelet` will ensure the `hugetlb` has no usage charged to the pod level
+cgroup sandbox prior to deleting the pod to ensure all resources are reclaimed.
+
+## Limits and Quota
+
+The `ResourceQuota` resource will be extended to support accounting for
+`hugepages.<hugepagesize>` similar to `cpu` and `memory`.  The `LimitRange`
+resource will be extended to define min and max constraints for `hugepages`
```
vishh: Burst ratio wouldn't apply to integer resources like hugepages right?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+
+For the Java use case, the JVM maps the huge pages as a shared memory segment
+and memlocks them to prevent the system from moving or swapping them out.
+
+There are several issues here:
+- The user running the Java app must be a member of the gid set in the
+  `vm.huge_tlb_shm_group` sysctl
+- sysctl `kernel.shmmax` must allow the size of the shared memory segment
+- The user's memlock ulimits must allow the size of the shared memory segment
+- `vm.huge_tlb_shm_group` is not namespaced.
```
vishh: How are these issues going to be simplified for users?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+
+The class of applications that benefit from huge pages typically have
+- A large memory working set
+- A sensitivity to memory access latency
+
+Example applications include:
+- database management systems (MySQL, PostgreSQL, MongoDB, Oracle, etc.)
+- Java applications can back the heap with huge pages using the
+  `-XX:+UseLargePages` and `-XX:LagePageSizeInBytes` options.
+- packet processing systems (DPDK)
```
fabiand: Virtual Machines (qemu/kvm/[libvirt](https://libvirt.org/formatdomain.html#elementsMemoryBacking)) can leverage huge pages

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+## Node Specfication
+
+Huge pages cannot be overcommitted on a node.
+
+A system may support multiple huge page sizes.  It is assumed that most nodes
+will be configured to primarily use the default huge page size as returned via
+`grep Hugepagesize /proc/meminfo`.  This defaults to 2MB on most Linux systems
+unless overriden by `default_hugepagesz=1g` in kernel boot parameters.  For each
+supported huge page size, the node will advertise a resource of the form
+`hugepages.<hugepagesize>`.
```
fabiand: Just to be clear - Resources as in kubernetes/community#782 ? Might be worth a link then.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+
+Supported huge page sizes are determined by parsing the
+`sys/kernel/mm/hugepages/hugepages-{size}kB` directory on the host.  Kubernetes
+will expose the `hugepages.<hugepagesize>` resource using binary notation form.
+For example, if a node supports `hugepages-2048kB`, a resource `hugepages.2Mi`
```
fabiand: Previously you said the pages will be advertised by using  `hugepages.<hugepagesize>` but you did not define the unit of the size - in this line however you implicitly convert `kB` to `Mi` - which is okay, but it's probably worth noting the default unit, to be able to specify the right resources whne creating pods.

IOW The proposal should define the uni to use, otherwise it's unclear if you have to request `2048kB` or `2Mi` units.

contributors/design-proposals/hugepages.md
fabiand: Yes - It's probably good to tie the `EmptyDir` to the request hugepages. Otherwise there is also the option of abusing the mechanism, by requesting one size, but then using a different `pageSize`.

Quite unlikely, but still possible is that different hugepage sizes are requested, thus the volume needs a reference to the hugepage size. Just saying that a volume can not infer the page size automatically if there is more than one given. This in turn would mean that the volume should reference the resource (thus `hugepages.2Mi`in the example above)

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,288 @@
...
+## Scheduler changes
+
+The scheduler will need to ensure any huge page request defined in the pod spec
+can be fulfilled by a candidate node.
+
+## cAdvisor changes
+
+cAdvisor will need to be modified to return the number of pre-allocated huge
+pages per page size on the node.  It will be used to determine capacity and
+calculate allocatable values on the node.
```
fabiand: This reporting is just informative, and not for scheduling, right?

As I'd expect that the resource class mechanism is used to advertise available hugepages.

contributors/design-proposals/hugepages.md
fabiand: IIUIC then the smallest granularity level of the scheduler so far is the node level - Do I recall this correctly?
For NUMA however, the granularity level would be intra-node locality - thus the locality of NUMA nodes and memory.

contributors/design-proposals/hugepages.md
derekwaynecarr: @fabiand - this is unrelated to resource class.

contributors/design-proposals/hugepages.md
derekwaynecarr: i will add text describing the size.

contributors/design-proposals/hugepages.md
fabiand: Right - so will it then be advertised as part of the node info?

contributors/design-proposals/hugepages.md
derekwaynecarr: @fabiand - for the initial alpha version of this feature, I am going to update the proposal to restrict the pod from consuming more than one huge page pool size.  The volume will then infer the pagesize to use based on the requests.

contributors/design-proposals/hugepages.md
derekwaynecarr: correct.

contributors/design-proposals/hugepages.md
derekwaynecarr: i want to evaluate the associated burden based on usage of the feature rather than pre-design.

contributors/design-proposals/hugepages.md
derekwaynecarr: the reporting is for scheduling.  the resource class mechanism is later.

derekwaynecarr:
@vishh @thockin @dchen1107 @saad-ali - updated based on last round of feedback.

notable changes:

1. a pod is limited to a single huge page size pool for the initial iteration of this feature.  this limits the exposure of changes to emptyDir to only require an alternate medium than an API schema change.  feedback at this point is that a pod consuming multiple page size pools is rare.
2. a pod requests huge pages like memory rather than integral values.  for example, `hugepages-2Mi: 1Gi` rather than `hugepages-2Mi: 512`.  This aligns more naturally to how users think about the memory, and aligns with how i would anticipate monitoring tools to report usage back via hugetlb cgroup (i.e. in actual bytes used instead of pages).

contributors/design-proposals/hugepages.md
derekwaynecarr: @fabiand - we anticipate the scheduler to only maintain visibility at the node level.  we anticipate the kubelet to take responsibility for intra-node locality, and that will be a factor of QoS.

contributors/design-proposals/hugepages.md
derekwaynecarr: @vishh - i think adjusting [Node Capacity] as described in your edited comment introduces far more confusion when you layer in other monitoring tools.  The machines memory capacity is [Node Capacity], the operator has reserved a portion of that capacity for use in huge pages.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,308 @@
...
+support quota.  For example, 1 huge page with size 2Mi is orders of magnitude
+different than 1 huge page with size 1Gi.  We assume gigantic pages are even
+more precious resources than huge pages.
+
+Pre-allocated huge pages reduce the amount of allocatable memory on a node. The
+node will treat pre-allocated huge pages similar to other system reservations
+and reduce the amount of `memory` it reports using the following formula:
+
+```
+[Allocatable] = [Node Capacity] -
```
dchen1107: Based on  formula and the example you listed below, looks like you change the semantics of NodeCapacity and NodeAllocatable which both are declared as ResourceList. Taking your example given below, it is confused to the user that the node has total memory 10Gi = 9Gi (memory allocatable) + 1Gi (hugepage-2Mi allocatable) or 11Gi  = 10Gi (memory capacity) + 1Gi (hugepage-2Mi)

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,308 @@
...
+
+A pod must make a request to consume pre-allocated huge pages using the resource
+`hugepages-<hugepagesize>` whose quantity is a positive amount of memory in
+bytes.  The specified amount must align with the `<hugepagesize>`; otherwise,
+the pod will fail validation.  For example, it would be valid to request
+`hugepages-2Mi: 4Mi`, but invalid to request `hugepages-2Mi: 3Mi`.
+
+The request and limit for `hugepages-<hugepagesize>` must match.  Similar to
+memory, an application that requests `hugepages-<hugepagesize>` resource is at
+minimum in the `Burstable` QoS class.
```
dchen1107: I thought we don't want to overcommit hugepage, at least for 1.8 release. If that is true, shouldn't we only support Guaranteed QoS class for hugepage? I understand that we changed QoS crossing all resources, and all containers, which makes any pod requiring hugepage is at least at Burstable class, but I want to make sure that is what you are referring here.

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,308 @@
...
+
+If a pod consumes huge pages via `shmget`, it must run with a supplemental group
+that matches `/proc/sys/vm/hugetlb_shm_group` on the node.  Configuration of
+this group is outside the scope of this specification.
```
dchen1107: Agreed that how to configure such supplemental group is outside the scope, but can you have an example on how a user pod to request huge pages via shmget?

contributors/design-proposals/hugepages.md
```diff
@@ -0,0 +1,308 @@
...
+
+Initially, a pod may not consume multiple huge page sizes in a single pod spec.
+Attempting to use `hugepages-2Mi` and `hugepages-1Gi` in the same pod spec will
+fail validation.  We believe it is rare for applications to attempt to use
+multiple huge page sizes. This restriction may be lifted in the future with
+community presented use cases.  Introducing the feature with this restriction
+limits the exposure of API changes needed when consuming huge pages via volumes.
```
dchen1107: Can we make the node specification with multiple huge page sizes invalid at the initial node capacity discovery level?

contributors/design-proposals/hugepages.md
ConnorDoyle: Isn't this the first instance of a cross-resource data dependency in node status? @derekwaynecarr could you expand on how the monitoring tools are affected by this decision?

contributors/design-proposals/hugepages.md
derekwaynecarr: we do not want to overcommit hugepages.  the request must always match the limit for hugepages.  i should be able have the request != limit for cpu and memory resource requirements on any pod that consumes hugepages.  i was trying to assert that hugepage requests impact qos (similar to cpu and memory).

contributors/design-proposals/hugepages.md
derekwaynecarr: i will add an example of a pod that runs a jvm backed by huge pages.

contributors/design-proposals/hugepages.md
derekwaynecarr: i guess that is fine.  the kubelet could fail startup if it detects multiple huge page sizes have been pre-allocated.

contributors/design-proposals/hugepages.md
derekwaynecarr: this was my thought process at the time:

1. most users associate capacity[memory] with `/proc/meminfo` `MemTotal` (i.e. total physical RAM)
1. pre-allocating huge pages does not change `MemTotal`, it just reduces `MemAvailable`
1. pre-allocating huge pages is an implicit reservation of `memory` for use in a different mode against memory capacity
1. i suspect users would be confused if the capacity value for `memory` as reported back by the kubelet differed from what is reported in cAdvisor (or other monitoring tools) and could bubble up through Prometheus, etc.  i really feel people will be confused as most monitoring tools will read `MemTotal`.

i don't think users should expect to add numbers with two different units when evaluating capacity.

if we wanted the operator to explicitly account for pre-allocated hugepages in either system-reserved or kube-reserved, i suspect that causes more trouble when we actually enforce-node-allocatable for system and kube cgroups as that memory is not actually available (its been reserved for hugepage usage only).

dchen1107: Ok, I agreed with you on this.

contributors/design-proposals/hugepages.md
dchen1107: sgtm

contributors/design-proposals/hugepages.md
dchen1107: Unfortunately not all kernel has MemAvailable in /proc/meminfo.

dchen1107:
/lgtm

We can address several comments in separate prs. Thanks!

PR #840, opened 2017-07-24
Initial draft of SIG architecture charter

bgrant0607:
Write up charter as discussed so far.

cc @kubernetes/sig-architecture-proposals @kubernetes/sig-architecture-pr-reviews

### DIFF
--- a/sig-architecture/README.md
+++ b/sig-architecture/README.md
@@ -8,7 +8,7 @@ To understand how this file is generated, see generator/README.md.
 -->
 # Architecture SIG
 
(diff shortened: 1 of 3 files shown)


jbeda:
This LGTM.  

My only concern is that I think there is a lot of detail missing in terms of the process of how decisions are made.  I'd love a TODO recognizing that this is the appropriate place to define/reference that.

bgrant0607:
@jbeda I added TODOs

bgrant0607:
Squashed. Merging. Will iterate more after tackling some other SIG arch issues.

PR #844, opened 2017-07-24
Adding a proposal for hardware accelerators

vishh:
This proposal captures the current state of affairs in the community around support for Hardware Accelerators.
This proposal is meant to help set expectations across the community.

The workloads aspect of hardware accelerators is intentionally pending. I intend to extend this proposal in the near future with workload specific user journeys.

@jiayingz @mindprince @thockin @derekwaynecarr @davidopp

### DIFF
--- a/contributors/design-proposals/hardware-accelerators.md
+++ b/contributors/design-proposals/hardware-accelerators.md
@@ -36,14 +36,13 @@ Some of them are general purpose, but many of them are purpose built for specifi
 Exposing all hardware accelerators as well known (first class) Compute Resource Types will bloat the API and compromise portability.
 For this reason, Hardware Accelerators are expected to be handled as “Extended Compute Resources”.
 
-Kubernetes nucleus will recommend and document a general purpose resource name for each family of accelerators - examples include `nvidia-gpu`, `amd-gpu`, `google-tpu`, etc., with a standard prefix `extensions.kubernetes.io`. This naming scheme partially mimics PCI ID - `<Vendor Name>-<Device Type>`.
-Any further differentiation amongst hardware accelerators using the resource name will not be considered “portable” across Kubernetes clusters.
-It is expected that accelerator hardware vendors will define and manage Resource Types.
+Kubernetes nucleus will recommend and document a general purpose resource name for each family of accelerators - examples include `nvidia.com/gpu`, `amd.com/gpu`, `google.com/tpu`, etc., with a standard domain name that unambiguously identifies the vendor of a hardware followed by the hardware type `<hardware-vendor-domain>/<hardware-type>`.
+It is expected that the hardware vendors will work with the kubernetes community to keep their resource names consistent across kubernetes clusters.
 
 Nodes are expected to be homogenous and any attributes specific to hardware accelerators are expected to be exposed as node labels in Kubernetes to begin with.
 Users can expose “extended resources” with other names and consume them in their own clusters.
-The admission logic will be extended to allow any resource with an `extensions.kubernetes.io` prefix.
-The scheduler will be extended to treat any resource with that prefix as an integer resource to begin with.
+The admission logic will be extended to allow any resource with a non empty non-default (not `kubernetes.io`) domain name.
+The scheduler will be extended to treat such extended resources as an integer resource to begin with.
 
 GPU workloads may wish to express soft and hard preferences for specific accelerator sub-types within a family.
 For example, a CUDA app may benefit from running on the latest and greatest Nvidia GPU, but has a minimum GPU version requirement.
@@ -116,7 +115,7 @@ Support for Nvidia GPUs has been in alpha since `v1.6`.
 
 ### Timelines
 
-* Current target is `v1.9`
+* Current target is `v1.10`
 
 ## General Availability
 
@@ -130,4 +129,4 @@ Support for Nvidia GPUs has been in alpha since `v1.6`.
 
 ### Timelines
 
-* Current target is `v1.11`
+* Current target is `v1.12`
--- a/contributors/design-proposals/hardware-accelerators.md
+++ b/contributors/design-proposals/hardware-accelerators.md
@@ -1,0 +1,133 @@
+# Hardware Accelerators
+
+Author: Vishnu Kannan (vishh@)
+
+## Introduction
+
+Hardware Accelerators are becoming a widely used commodity across various industries.
+Accelerators can bring down computing latency and/or costs significantly.
+Many of these accelerators present unique technical constraints.
+Kubernetes as an established Application Management Platform can play a crucial role in getting workloads that require hardware acceleration to be portable and function at scale.
+Most of the large public clouds already provide various types of Hardware Accelerators.
+There are quite a few Kubernetes users who are already managing Accelerators at scale on their own datacenters.
+
+The remainder of this document provides a technical overview of how hardware accelerators will be integrated into Kubernetes.
+
+## Goals
+
+* Make Kubernetes Machine Learning and Deep Learning friendly.
+* Extensible support for various types of hardware accelerators Nvidia GPUs, AMD GPUs and Google Tensor Processing Units
+* Portability across Kubernetes clusters
+* Limit feature creep to Kubernetes nucleus
+* Provide baseline application performance guarantees
+
+## Non Goals
+* Support for Cloud Gaming, Simulations, Remote Desktops and other workloads
+  * Support for these workloads will be tackled once support for Machine Learning matures
+
+## System Design
+
+The following sections highlight some of the critical design points for supporting Hardware Accelerators
+
+### API
+
+A plethora of Hardware accelerators exist in the world.
+Some of them are general purpose, but many of them are purpose built for specific use cases.
+Exposing all hardware accelerators as well known (first class) Compute Resource Types will bloat the API and compromise portability.
+For this reason, Hardware Accelerators are expected to be handled as “Extended Compute Resources”.
+
+Kubernetes nucleus will recommend and document a general purpose resource name for each family of accelerators - examples include `nvidia-gpu`, `amd-gpu`, `google-tpu`, etc., with a standard prefix `extensions.kubernetes.io`. This naming scheme partially mimics PCI ID - `<Vendor Name>-<Device Type>`.
+Any further differentiation amongst hardware accelerators using the resource name will not be considered “portable” across Kubernetes clusters.
+It is expected that accelerator hardware vendors will define and manage Resource Types.
+
+Nodes are expected to be homogenous and any attributes specific to hardware accelerators are expected to be exposed as node labels in Kubernetes to begin with.
+Users can expose “extended resources” with other names and consume them in their own clusters.
+The admission logic will be extended to allow any resource with an `extensions.kubernetes.io` prefix.
+The scheduler will be extended to treat any resource with that prefix as an integer resource to begin with.
+
+GPU workloads may wish to express soft and hard preferences for specific accelerator sub-types within a family.
+For example, a CUDA app may benefit from running on the latest and greatest Nvidia GPU, but has a minimum GPU version requirement.
+In addition to this, [feedback from Nvidia](https://docs.google.com/document/d/1lSwVh2ZfJ2FeLXIeyyiNqN_hKPYpahJiwN5X5cszjOk/edit) has indicated that newer version of GPUs are placed alongside older generations on the same machine (heterogenous nodes). 
+To support a combination of these two use cases, new scheduling features have been proposed and discussed in the community [here](https://docs.google.com/document/d/1666PPUs4Lz56TqKygcy6mXkNazde-vwA7q4e5H92sUc/edit). 
+Kubernetes will support heterogenous nodes in the future, but the initial plan is to support homogenous nodes with standard node labels prior to tackling heterogeneous nodes.
+Node labels satisfy the former "soft" and "hard" preferences use case on homogeneous nodes.
+Homogeneous nodes are recommended over heteregenous nodes since the latter may cause resource fragmentation at scale.
+
+### SW Infrastructure for Accelerators
+
+Hardware Accelerators often need vendor provided kernel and user space software.
+These software at times introduce tight coupling between the host and applications.
+Nvidia GPUs for example are consumed via higher level APIs like CUDA, CUVID, etc.
+These APIs are available via user space libraries.
+The libraries themselves are tied to the host image (kernel and Nvidia kernel driver versions primarily).
+These APIs break the abstraction of containers where the general assumption is that applications inside a container bring all their libraries as part of the container image.
+
+#### Extensibility
+
+Instead of building a special solution for Nvidia GPUs in Kubernetes, a standard extension pipeline called "Hardware Device Plugin" [has been proposed](https://docs.google.com/a/google.com/document/d/1LHeTPx_fWA1PdZkHuALPzYxR0AYXUiiXdo3S0g2VSlo/edit?usp=drive_web) to support arbitrary hardware (and virtual) devices without requiring device specific changes to Kubernetes nucleus.
+SW for hardware accelerators are expected to be shipped via standard containers. These containers are expected to be deployed on every node with accelerators. These containers are expected to install necessary SW for initializing hardware accelerators, register themselves with the Kubelet via standard device plugin APIs and exposing accelerators as consumable compute resources via Kubernetes APIs.
+Kubelet will handle allocation of hardware accelerators to pods and containers.
+Kubelet will communicate with the plugins to ensure that the necessary environment (SW, devices, env variables, etc.) to access hardware accelerators assigned to a pod/container are made accessible within the pod/container sandbox.
+
+Kubernetes will not provide any primitives to manage lifecycle of SW for hardware accelerators that are handled outside of Hardware Device Plugins. 
+
+### Monitoring
+
+Hardware Accelerators are expensive and typically have unique hardware architectures.
+Programming against these accelerators, improving performance and utilization is non-trivial.
+Certain generic metrics like `utilization` and `usage_time`, and vendor specific metrics are expected to be exposed via cAdvisor and made available to monitoring solutions.
+These metrics will not be available as part of the core Metrics APIs since Kubernetes nucleus isn't providing any functionality based on these metrics.
+
+### Predictable performance
+
+Accelerators are preferred over CPUs mainly for performance reasons.
+Accelerators typically have extreme requirements at the hardware level in terms of power, hardware interconnect bandwidth, latency, etc.
+These high performance devices require careful placement of user workloads on specific CPUs, Memory banks and Accelerator devices to reduce latency and guarantee application performance.
+Kubernetes will support support performance isolation for these hardware accelerators, by allowing hardware device plugins to expose a hardware topology graph where each edge represents latency to access one or more CPUs.
+Kubelet will combine graphs from multiple plugins along with the node’s NUMA topology to handle hardware device assignment.
+Performance guarantees are expected to be tackled once basic support for Hardware Accelerators mature.
+
+## Implementation Plan
+
+The following implementation plan is a proposal. Actual delivery dates may change.
+
+## Alpha
+
+### Requirements
+* Opt-in with no support
+* No patch fixes or backwards compatibility guarantees.
+
+### Timeline
+
+Support for Nvidia GPUs has been in alpha since `v1.6`.
+
+## Beta
+
+### Requirements 
+
+* Backwards compatibile Resource APIs
+* Portability across Kubernetes clusters
+* Adequate documentation on API including vendor specific attributes
+* End to end tests that exercise the APIs
+
+### Dependencies
+
+* Hardware Device Plugin feature set and functionalities mature with backwards compatible APIs
+
+### Timelines
+
+* Current target is `v1.9`
+
+## General Availability
+
+### Requirements
+
+* Minimal performance guarantees
+
+### Dependencies
+
+* Hardware topology aware scheduling in Kubelet along with support for topology in Hardware Device Plugin APIs
+
+### Timelines
+
+* Current target is `v1.11`


contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,133 @@
+# Hardware Accelerators
+
+Author: Vishnu Kannan (vishh@)
+
+## Introduction
+
+Hardware Accelerators are becoming a widely used commodity across various industries.
+Accelerators have bring down computing latency and/or costs significantly.
```
derekwaynecarr: nit: Accelerators can bring* ?

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,133 @@
...
+
+* Make Kubernetes Machine Learning and Deep Learning friendly.
+* Extensible support for various types of hardware accelerators Nvidia GPUs, AMD GPUs and Google Tensor Processing Units
+* Portability across Kubernetes clusters
+* Limit feature creep to Kubernetes nucleus
+* Provide baseline application performance guarantees
+
+## Non Goals
+* Support for Cloud Gaming, Simulations, Remote Desktops and other workloads
+  * Support for these workloads will be tackled once support for ML and DL matures
```
derekwaynecarr: maybe link to what differs about these workloads for those unfamiliar

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,133 @@
...
+A plethora of Hardware accelerators exist in the world.
+Some of them are general purpose, but many of them are purpose built for specific use cases.
+Exposing all hardware accelerators as well known (first class) Compute Resource Types will bloat the API and compromise portability.
+For this reason, Hardware Accelerators are expected to be handled as “Extended Compute Resources”.
+
+Kubernetes nucleus will recommend and document a general purpose resource name for each family of accelerators - examples include `nvidia-gpu`, `amd-gpu`, `google-tpu`, etc., with a standard prefix `extensions.kubernetes.io`. This naming scheme partially mimics PCI ID - `<Vendor Name>-<Device Type>`.
+Any further differentiation amongst hardware accelerators using the resource name will not be considered “portable” across Kubernetes clusters.
+It is expected that accelerator hardware vendors will define and manage Resource Types.
+
+Nodes are expected to be homogenous and any attributes specific to hardware accelerators are expected to be exposed as node labels in Kubernetes to begin with.
```
derekwaynecarr: i thought we wanted to support heterogeneous devices via resource class in the future?  am i misinterpreting the homogeneous expectation?

derekwaynecarr: ok, i see clarification below.  i would prefer we phrase this differently.  we wish to support heterogenous devices in the future, but initially we will tackle the homegenous case first.

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,133 @@
...
+* Adequate documentation on API including vendor specific attributes
+* End to end tests that exercise the APIs
+
+### Dependencies
+
+* Hardware Device Plugin feature set and functionalities mature with backwards compatible APIs
+
+### Timelines
+
+* Current target is `v1.9`
```
derekwaynecarr: this assumes device api is alpha in 1.8, correct?

derekwaynecarr:
a few questions, but generally, looks good to me and is consistent with what we have discussed in community.

contributors/design-proposals/hardware-accelerators.md
vishh: Oops. Fixing it.

contributors/design-proposals/hardware-accelerators.md
vishh: The difference doesn't matter much from K8s perspective. Hence removing it.

contributors/design-proposals/hardware-accelerators.md
vishh: Ack

contributors/design-proposals/hardware-accelerators.md
vishh: Yes.

vishh:
@derekwaynecarr can I get an LGTM?

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,133 @@
...
+The following sections highlight some of the critical design points for supporting Hardware Accelerators
+
+### API
+
+A plethora of Hardware accelerators exist in the world.
+Some of them are general purpose, but many of them are purpose built for specific use cases.
+Exposing all hardware accelerators as well known (first class) Compute Resource Types will bloat the API and compromise portability.
+For this reason, Hardware Accelerators are expected to be handled as “Extended Compute Resources”.
+
+Kubernetes nucleus will recommend and document a general purpose resource name for each family of accelerators - examples include `nvidia-gpu`, `amd-gpu`, `google-tpu`, etc., with a standard prefix `extensions.kubernetes.io`. This naming scheme partially mimics PCI ID - `<Vendor Name>-<Device Type>`.
```
thockin: If you want to go down this road, someone has to manage a registry of vendor names (e.g. "nvidia" and never "nVidia" or "NVidia" or "n-vidia"), and do some sort of verification of ownership and non-infringement, and so on.  The normal "use your own domain as the prefix" is a bit more self-service.

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Extensibility
+
+Instead of building a special solution for Nvidia GPUs in Kubernetes, a standard extension pipeline called "Hardware Device Plugin" [has been proposed](https://docs.google.com/a/google.com/document/d/1LHeTPx_fWA1PdZkHuALPzYxR0AYXUiiXdo3S0g2VSlo/edit?usp=drive_web) to support arbitrary hardware (and virtual) devices without requiring device specific changes to Kubernetes nucleus.
+SW for hardware accelerators are expected to be shipped via standard containers. These containers are expected to be deployed on every node with accelerators. These containers are expected to install necessary SW for initializing hardware accelerators, register themselves with the Kubelet via standard device plugin APIs and exposing accelerators as consumable compute resources via Kubernetes APIs.
+Kubelet will handle allocation of hardware accelerators to pods and containers.
+Kubelet will communicate with the plugins to ensure that the necessary environment (SW, devices, env variables, etc.) to access hardware accelerators assigned to a pod/container are made accessible within the pod/container sandbox.
```
thockin: Where is the spec for kubelet <-> plugin?  Is this another form of CSI-style plugin?  I'd like to converge these models if we can.

contributors/design-proposals/hardware-accelerators.md
vishh: One of the goals is to provide guidelines similar to the PCI ID schema in Linux. 
Custom domains may not be consistent across different drivers (plugins) for the same hardware. WDYT?

contributors/design-proposals/hardware-accelerators.md
vishh: It is being designed via this proposal - https://github.com/RenaudWasTaken/community/blob/f1b462b12353df8c4467aec2ace251c7424c9078/contributors/design-proposals/device-plugin.md

There are several open issues with the Spec and so I don't think it's ready for a thorough review. It will provide a good overview though.

vishh: IIUC, CSI APIs are very closely tied to filesystem APIs. The main overlap between CSI and the hardware device plugin API that I can see if with local storage devices. Since Storage resource APIs are completely different from other resources and they need a notion of Identity, I felt these two APIs are catering different needs.
If you can think of a means to converge to a single API, I'm all ears @thockin

thockin: What I really meant was convergence in transport and feel, rather than API.  If they were all grpc with the same basic assumptions about auth and discovery, with the same basic concepts, it would help a lot.

That said, there's a proposal on the table to change storage stuff to a much more declaratoive model, and it has some compelling aspects, too.

contributors/design-proposals/hardware-accelerators.md
thockin: My concern is the implication that we will manage a database of abstract names in perpetuity.

contributors/design-proposals/hardware-accelerators.md
vishh: @thockin Got it. We have 3 more months to graduate the resource/device plugin API to beta. If the CSI API infra get's finalized by then, we can definitely re-use it.

contributors/design-proposals/hardware-accelerators.md
vishh: @thockin As discussed offline, can you take a look at the language and suggest appropriate changeS?

derekwaynecarr: the alternative is to have a name like the following:

`<vendor-name>.extensions.kubernetes.io/<device-type>`

which seems to satisfy @thockin concern.

vishh: @derekwaynecarr in your proposal I assume you want the Pod Spec to exclude the `<vendor-name>`. I find that a bit odd. Even in that case, we still need to curate the portable resource name, `amd-gpu` for example.

derekwaynecarr: @vishh - i did not assume the pod spec exclude the vendor-name.

`amd.extensions.kubernetes.io/gpu`

would have appeared in the pod spec.

vishh: Originally, the plan was to just special case `extensions.kubernetes.io` domain similar to OIRs.
Now though, we are thinking of treating any resources that's not under `kubernetes.io` or `default` domain as being an integer resource. So your proposal should also work @derekwaynecarr 

@thockin WDYT about including the vendor name in the prefix?
TBH, I feel the `extensions.kubernetes.io` would be redundant once we go down that path.

jiayingz: @vishh any thoughts on how to determine whether an integer resource allows overcommits or not? Right now, we have no-overcommit limitation for ResourceNvidiaGPU but not for the general OIR. Given that we want to deprecate ResourceNvidiaGPU in the future, how can we know a particular vendor resource allows overcommit or not?

ConnorDoyle: @jiayingz  there are at least a few options available there:

1. Encode whether the resource can be overcommitted somehow in the resource name.
1. Maintain a list of either overcommit or no-overcommit resource names and use them in admission control.
1. Don't allow overcommit for now, except for first-class resources. A future API for resources could allow enabling overcommit on a case-by-case basis.
    1. Do this in a way that soft-breaks OIR (currently they can be overcommitted)
    1. Special-case OIR to preserve their overcommit behavior.

(1) seems likely to require a future breaking change.
(2) potentially couples admission to extended resource names.
(3.a) might be OK but we don't have a good way to determine the extent of the breakage across the ecosystem so I would vote not to do this without a deprecation cycle.
(3.b) could be a decent stopgap, made more tolerable if we deprecate OIR at the same time. The breakage due to feature removal would be that `pod.alpha.kubernetes.io/opaque-int-<foo>` resources can no longer be overcommited.

jiayingz:
On Thu, Aug 10, 2017 at 5:22 PM, Connor Doyle <notifications@github.com>
wrote:

> *@ConnorDoyle* commented on this pull request.
> ------------------------------
>
> In contributors/design-proposals/hardware-accelerators.md
> <https://github.com/kubernetes/community/pull/844#discussion_r132599666>:
>
> > +## Non Goals
> +* Support for Cloud Gaming, Simulations, Remote Desktops and other workloads
> +  * Support for these workloads will be tackled once support for Machine Learning matures
> +
> +## System Design
> +
> +The following sections highlight some of the critical design points for supporting Hardware Accelerators
> +
> +### API
> +
> +A plethora of Hardware accelerators exist in the world.
> +Some of them are general purpose, but many of them are purpose built for specific use cases.
> +Exposing all hardware accelerators as well known (first class) Compute Resource Types will bloat the API and compromise portability.
> +For this reason, Hardware Accelerators are expected to be handled as “Extended Compute Resources”.
> +
> +Kubernetes nucleus will recommend and document a general purpose resource name for each family of accelerators - examples include `nvidia-gpu`, `amd-gpu`, `google-tpu`, etc., with a standard prefix `extensions.kubernetes.io`. This naming scheme partially mimics PCI ID - `<Vendor Name>-<Device Type>`.
>
> @jiayingz <https://github.com/jiayingz> there are at least a few options
> available there:
>
>    1. Encode whether the resource can be overcommitted somehow in the
>    resource name.
>    2. Maintain a list of either overcommit or no-overcommit resource
>    names and use them in admission control.
>    3. Don't allow overcommit for now, except for first-class resources. A
>    future API for resources could allow enabling overcommit on a case-by-case
>    basis.
>       1. Do this in a way that soft-breaks OIR (currently they can be
>       overcommitted)
>       2. Special-case OIR to preserve their overcommit behavior.
>
> (1) seems likely to require a future breaking change.
> (2) potentially couples admission to extended resource names.
> (3.a) might be OK but we don't have a good way to determine the extent of
> the breakage across the ecosystem so I would vote not to do this without a
> deprecation cycle.
> (3.b) could be a decent stopgap, made more tolerable if we deprecate OIR
> at the same time. The breakage due to feature removal would be that
> pod.alpha.kubernetes.io/opaque-int-<foo> resources can no longer be
> overcommited.
>
Agree 3.b seems a good stopgap. Do we have a time plan to deprecate OIR? We
can remove this
OIR special-case at the time OIR is deprecated.

Jiaying

contributors/design-proposals/hardware-accelerators.md
vishh: Extended or Opaque resources do not need overcommit. 
I will update this PR to reflect the fact that a curated domain name is no longer necessary for plugins.

thockin: Is it legitimate for two different authors to produce drivers for the same resource?  Does it matter here if an NV-provided driver and an OSS driver both claim to provide ".../nvidia-gpu" ?  Or do we want the resource name to reflect the driver provider as well as the device?

contributors/design-proposals/hardware-accelerators.md
thockin: I don't think it's optional.  We have to be driving plugins towards convergence.  No "if" here.  Or we have to have a good reason why we can't and never intend to do so.

thockin:
@mindprince you get the lgtm :)

contributors/design-proposals/hardware-accelerators.md
vishh: From a user perspective, the resource name is an API that gives them access to a set of resources (character device files, user space libraries, kernel APIs, etc.) Identifying and documenting that API will be key to the project. That would then make it possible to have alternate implementations of device plugins for the same hardware.
There may be legal issues that vendors may potentially raise, but I do not have a good handle on that. 
Thoughts?

vishh:
@thockin updated the naming scheme to reflect recent conversations. PTAL if you have some cycles. 
@derekwaynecarr can you re-review this patch?

cc @jiayingz @mindprince

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,132 @@
...
+
+A plethora of Hardware accelerators exist in the world.
+Some of them are general purpose, but many of them are purpose built for specific use cases.
+Exposing all hardware accelerators as well known (first class) Compute Resource Types will bloat the API and compromise portability.
+For this reason, Hardware Accelerators are expected to be handled as “Extended Compute Resources”.
+
+Kubernetes nucleus will recommend and document a general purpose resource name for each family of accelerators - examples include `nvidia.com/gpu`, `amd.com/gpu`, `google.com/tpu`, etc., with a standard domain name that unambiguously identifies the vendor of a hardware followed by the hardware type `<hardware-vendor-domain>/<hardware-type>`.
+It is expected that the hardware vendors will work with the kubernetes community to keep their resource names consistent across kubernetes clusters.
+
+Nodes are expected to be homogenous and any attributes specific to hardware accelerators are expected to be exposed as node labels in Kubernetes to begin with.
```
derekwaynecarr: The expectation stated here for homogenous nodes contradicts later sections in same paragraph.   Possibly rephrase to state priority order?

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,132 @@
...
+Users can expose “extended resources” with other names and consume them in their own clusters.
+The admission logic will be extended to allow any resource with a non empty non-default (not `kubernetes.io`) domain name.
+The scheduler will be extended to treat such extended resources as an integer resource to begin with.
```
derekwaynecarr: Note we will not support overcommiting initially?

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,132 @@
...
+Kubelet will handle allocation of hardware accelerators to pods and containers.
+Kubelet will communicate with the plugins to ensure that the necessary environment (SW, devices, env variables, etc.) to access hardware accelerators assigned to a pod/container are made accessible within the pod/container sandbox.
+
+Kubernetes will not provide any primitives to manage lifecycle of SW for hardware accelerators that are handled outside of Hardware Device Plugins. 
+
+### Monitoring
+
+Hardware Accelerators are expensive and typically have unique hardware architectures.
+Programming against these accelerators, improving performance and utilization is non-trivial.
+Certain generic metrics like `utilization` and `usage_time`, and vendor specific metrics are expected to be exposed via cAdvisor and made available to monitoring solutions.
```
derekwaynecarr: What is being done here explicitly?

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,132 @@
...
+* Opt-in with no support
+* No patch fixes or backwards compatibility guarantees.
+
+### Timeline
+
+Support for Nvidia GPUs has been in alpha since `v1.6`.
+
+## Beta
+
+### Requirements
```
derekwaynecarr: Is basic monitoring implied here?

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,132 @@
...
+Programming against these accelerators, improving performance and utilization is non-trivial.
+Certain generic metrics like `utilization` and `usage_time`, and vendor specific metrics are expected to be exposed via cAdvisor and made available to monitoring solutions.
+These metrics will not be available as part of the core Metrics APIs since Kubernetes nucleus isn't providing any functionality based on these metrics.
+
+### Predictable performance
+
+Accelerators are preferred over CPUs mainly for performance reasons.
+Accelerators typically have extreme requirements at the hardware level in terms of power, hardware interconnect bandwidth, latency, etc.
+These high performance devices require careful placement of user workloads on specific CPUs, Memory banks and Accelerator devices to reduce latency and guarantee application performance.
+Kubernetes will support support performance isolation for these hardware accelerators, by allowing hardware device plugins to expose a hardware topology graph where each edge represents latency to access one or more CPUs.
```
ScorpioCPH: typo? `support`

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/844)

contributors/design-proposals/hardware-accelerators.md
ddysher: @vishh For discovery in particular, device plugin proposal assumes plugins call kubelet for registration; whereas in csi, if implemented in kubernetes, operator negotiate plugin unix socket path between kubelet and plugin, and kubelet is responsible to watch for new plugin.  Wondering what's the reason for this difference?

do you recall the location of the proposal :) @thockin 
> That said, there's a proposal on the table to change storage stuff to a much more declaratoive model, and it has some compelling aspects, too.

jiayingz:
On Mon, Oct 16, 2017 at 7:58 PM, Deyuan Deng <notifications@github.com>
wrote:

> *@ddysher* commented on this pull request.
> ------------------------------
>
> In contributors/design-proposals/hardware-accelerators.md
> <https://github.com/kubernetes/community/pull/844#discussion_r145017477>:
>
> > +
> +### SW Infrastructure for Accelerators
> +
> +Hardware Accelerators often need vendor provided kernel and user space software.
> +These software at times introduce tight coupling between the host and applications.
> +Nvidia GPUs for example are consumed via higher level APIs like CUDA, CUVID, etc.
> +These APIs are available via user space libraries.
> +The libraries themselves are tied to the host image (kernel and Nvidia kernel driver versions primarily).
> +These APIs break the abstraction of containers where the general assumption is that applications inside a container bring all their libraries as part of the container image.
> +
> +#### Extensibility
> +
> +Instead of building a special solution for Nvidia GPUs in Kubernetes, a standard extension pipeline called "Hardware Device Plugin" [has been proposed](https://docs.google.com/a/google.com/document/d/1LHeTPx_fWA1PdZkHuALPzYxR0AYXUiiXdo3S0g2VSlo/edit?usp=drive_web) to support arbitrary hardware (and virtual) devices without requiring device specific changes to Kubernetes nucleus.
> +SW for hardware accelerators are expected to be shipped via standard containers. These containers are expected to be deployed on every node with accelerators. These containers are expected to install necessary SW for initializing hardware accelerators, register themselves with the Kubelet via standard device plugin APIs and exposing accelerators as consumable compute resources via Kubernetes APIs.
> +Kubelet will handle allocation of hardware accelerators to pods and containers.
> +Kubelet will communicate with the plugins to ensure that the necessary environment (SW, devices, env variables, etc.) to access hardware accelerators assigned to a pod/container are made accessible within the pod/container sandbox.
>
> @vishh <https://github.com/vishh> For discovery in particular, device
> plugin proposal assumes plugins call kubelet for registration; whereas in
> csi, if implemented in kubernetes, operator negotiate plugin unix socket
> path between kubelet and plugin, and kubelet is responsible to watch for
> new plugin. Wondering what's the reason for this difference?
>
@ddysher Thanks a lot for the information! Could you explain a bit more on
what "operator negotiate plugin unix socket path between kubelet and
plugin, and kubelet is responsible to watch for new plugin" means? I
haven't followed the CSI design closely, although it is on our agenda to
compare different plugin systems including CSI, CNI, and DevicePlugin and
makes sure their communication models stay consistent, as @thockin
suggested :).

Jiaying

ddysher:
@jiayingz  I mean operator (human) is responsible to pass csi endpoint to both server (e.g. kubelet) and plugin.  plugin will create and listen on the endpoint (e.g. socket), while server monitors the endpoint.  If server observes that endpoint exists, it will establish the connection.

I didn't involve in csi discussion, this is my understanding after going through the [spec](https://github.com/container-storage-interface/spec/blob/master/spec.md)

contributors/design-proposals/hardware-accelerators.md
```diff
@@ -0,0 +1,132 @@
...
+Kubelet will combine graphs from multiple plugins along with the node’s NUMA topology to handle hardware device assignment.
```
jaypipes: I was under the impression that the scheduler was the component which placed Pods on a target node and allocated resources from that target node to the Pod. If that is the case, this seems to be carving out some special types of resources that will *not* be allocated by the scheduler but instead will be allocated/scheduled by the Kubelet. Is my understanding of that correct?

ddysher: Here is a comment found in the design doc
https://docs.google.com/document/d/1LHeTPx_fWA1PdZkHuALPzYxR0AYXUiiXdo3S0g2VSlo/edit?disco=AAAABKV195w

sisosiso:
wireless network monitoring

PR #849, opened 2017-07-26
Remove extra whitespace from sig-arch

cblecker:
This just removed an extra whitespace that snuck in there.

/cc @xiangpengzhao


xiangpengzhao:
Thanks @cblecker ! 
I verify this locally, it works as expected (not touching `sig-architecture/README.md` accidentally).
This LGTM, though I can't label `label` :)

cblecker:
Thanks! Self-merging because of the single character change :)

PR #862, opened 2017-07-29
Remove a not exist update-staging-client-go.sh

FengyunPan:
The update-staging-client-go.sh has been removed, and there is no
need to run it.


cblecker:
/lgtm
https://github.com/kubernetes/kubernetes/pull/49346 removed this script/step :)
FYI @caesarxuchao

PR #866, opened 2017-07-31
Specify --provider option on local e2e test

oomichi:
e2e test finds a suitable util.sh from "--provider" option.
However the e2e test way of community repo didn't show the setting of
the option and the gce's util.sh was selected as the default.

This patch fixes it.


xiangpengzhao:
Seems like `KUBERNETES_PROVIDER` is deprecated. See: https://github.com/kubernetes/community/commit/07c6da22fe9f97ab44a02e67a7c574b49998605d

/cc @crimsonfaith91@janetkuo

oomichi:
Thanks for your comment, xiangpengzhao

I confirmed "--provider=local" works fine instead of specifying the deprecated env.
So I will update this PR with the option.

PR #869, opened 2017-08-01
Update alpha field guidance

liggitt:
Summarizing outcome of discussions with @kubernetes/sig-api-machinery-misc 

ref:
* https://docs.google.com/document/d/1wuoSqHkeT51mQQ7dIFhUKrdi3-1wbKrNWeIL4cKb9zU/edit#
* https://github.com/kubernetes/kubernetes/issues/34508#issuecomment-319412303
* https://www.youtube.com/watch?v=bzSK00TxEx0&list=PL69nYSiGNLP21oW3hbLyjjj4XhrwKxH2R&index=4&t=2580 (43:00-59:00)
* https://github.com/kubernetes/kubernetes/issues/30819

### DIFF
--- a/contributors/devel/api_changes.md
+++ b/contributors/devel/api_changes.md
@@ -788,8 +788,10 @@ For example, consider the following object:
 ```go
 // API v6.
 type Frobber struct {
-  Height int    `json:"height"`
-  Param  string `json:"param"`
+  // height ...
+  Height *int32 `json:"height" protobuf:"varint,1,opt,name=height"`
+  // param ...
+  Param  string `json:"param" protobuf:"bytes,2,opt,name=param"`
 }
 ```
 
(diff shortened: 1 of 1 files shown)


liggitt:
cc @lavalamp @smarterclayton @bgrant0607 @kubernetes/sig-api-machinery-pr-reviews @kubernetes/api-reviewers

contributors/devel/api_changes.md
```diff
@@ -812,45 +817,104 @@ However, the new feature is not stable enough to be used in a stable version
...
+      // param ...
+      Param  string `json:"param" protobuf:"bytes,2,opt,name=param"`
+      // width indicates how wide the object is.
+      // This field is alpha-level and should only be sent to servers that enable the Frobber2D feature.
+      // +optional
+      Width  *int32 `json:"width,omitempty" protobuf:"varint,3,opt,name=width"`
+    }
+    ```
+
+3. In validation, ensure that the new field is rejected if the feature gate is disabled:
```
deads2k: Doing this would make turning on the feature-gate a one way operation since future updates (even updates that attempt to delete the thing from GC) would fail.

contributors/devel/api_changes.md
```diff
@@ -812,45 +817,104 @@ However, the new feature is not stable enough to be used in a stable version
...
+  // param ...
+  Param  string `json:"param" protobuf:"bytes,2,opt,name=param"`
+  // width ...
+  Width  *int32 `json:"width,omitempty" protobuf:"varint,3,opt,name=width"`
 }
 ```
 
 The latter requires that all objects in the same API group as `Frobber` to be
-replicated in the new version, `v6alpha2`. This also requires user to use a new
+replicated in the new version, `v7alpha1`. This also requires user to use a new
```
MHBauer: I think this particular example offers a strong amount of guidance.

before it seemed to suggest taking an api 'back to alpha', chronologically, like so:
```
v6alpha1
v6beta1
v6
v6alpha2
```

and now what makes more sense to me to suggest a main version bump, again chronologically
```
v6alpha1
v6beta1
v6
v7alpha1
```

contributors/devel/api_changes.md
```diff
@@ -812,45 +817,104 @@ However, the new feature is not stable enough to be used in a stable version
...
-A preferred option is to first make a release where the new value (`Width` in
-this example) is specified via an annotation, like this:
-
-```go
-kind: frobber
-version: v6
-metadata:
-  name: myfrobber
-  annotations:
-    frobbing.alpha.kubernetes.io/width: 2
```
MHBauer: Is this annotation process no longer advocated? @cmluciano !!

liggitt: it is not advocated. see the links in this PR's description for the problems. the alternatives to the approach described in this PR are:

The alternatives are:

* use annotations - exposes you to time-bomb data, no way for an admin to control use, if the field graduates to beta/stable it requires breaking clusters using alpha fields or trying to migrate/surface the options in both an annotation and field at the same time (which doesn't work)
* use fields with "alpha" in the name - if the field graduates to beta/stable it requires breaking clusters using alpha fields or trying to migrate/surface the options in two fields at the same time (which doesn't work)
* use fields without "alpha" in the name and allow changing them in incompatible ways - breaks old servers and/or clients if the schema changes

liggitt:
> Is there a use of this feature gate process I can look at?

it was used for the priority fields added to the pod spec in https://github.com/kubernetes/kubernetes/pull/45610/commits/91f893eebe6d9d58ebc1b89dd1f0ac328c8e19bc

contributors/devel/api_changes.md
liggitt: agree, starting v6alpha2 after v6 exists wouldn't make sense.

contributors/devel/api_changes.md
```diff
@@ -798,9 +800,12 @@ A developer is considering adding a new `Width` parameter, like this:
...
 type Frobber struct {
-  Height int    `json:"height"`
-  Width  int    `json:"height"`
-  Param  string `json:"param"`
+  // height ...
+  Height *int32 `json:"height" protobuf:"varint,1,opt,name=height"`
+  // param ...
+  Param  string `json:"param" protobuf:"bytes,2,opt,name=param"`
+  // width ...
+  Width  *int32 `json:"width,omitempty" protobuf:"varint,3,opt,name=width"`
```
seh: Do you prescribe adding the new field to the end of the Go struct, with the intention of allowing existing "unkeyed" [composite literals](https://golang.org/ref/spec#Composite_literals) to remain valid when recompiled, or do you mandate use keyed elements so that existing code can't impose assumptions about field order?

liggitt: > Do you prescribe adding the new field to the end of the Go struct, with the intention of allowing existing "unkeyed" composite literals to remain valid when recompiled

No, unkeyed literal compatibility is impossible when adding new fields.

Adding new fields at the end just makes it easier to see the incrementing protobuf tag.

seh: Ah, yes, a trip to the Go Playground confirmed what I had forgotten. I was thinking that that an unkeyed composite literal could get away with specifying a prefix of the fields, but then this is where that trick of using a field called "_"  comes in, defending against callers using unkeyed literals.

contributors/devel/api_changes.md
```diff
@@ -812,45 +817,104 @@ However, the new feature is not stable enough to be used in a stable version
...
+    var defaultKubernetesFeatureGates = map[utilfeature.Feature]utilfeature.FeatureSpec{
+      ...
+      Frobber2D: {Default: false, PreRelease: utilfeature.Alpha},
+    }
+    ```
+
+2. Add the field to the API type:
+
+    * mark the field `omitempty` and optional
+    * ensure the field is entirely absent from the API when empty (might require making the field a pointer)
```
thockin: all optionals should be pointers or otherwise nil-able

contributors/devel/api_changes.md
thockin: gates are for alphas, and we have no strong guarantees for alpha anyway?

liggitt:
> If this is the new guidance ( I thought we had larger concerns with this approach when we discussed it last year) we need to spread the word FAR and WIDE.

Agree. This is the start. The alternatives (annotations, field with alpha-name, field without alpha-name that allows changing the schema) all have major downsides.

> Are we comfortable with the idea that an alpha feature can burn a field name basically forever?

I don't think we have much of an option... if you name the field specially, there is no path to beta/stable that doesn't summarily break everyone using the alpha field. If you don't name it specially, and just change the schema, you break decoding of new beta data by old clients and servers with inert alpha fields in their internal structs.

liggitt:
> gates are for alphas, and we have no strong guarantees for alpha anyway?

gates ensure the server doesn't accept alpha data without opt-in, since it can drive unwanted or unstable behavior

contributors/devel/api_changes.md
```diff
@@ -812,45 +817,104 @@ However, the new feature is not stable enough to be used in a stable version
...
+        allErrs = append(allErrs, field.Forbidden(fldPath.Child("width"), "disabled by feature-gate"))
+      }
+      ...
+    }
+    ```
+
+In future versions:
+
+* if the feature progresses to beta or stable status, the feature gate can simply be enabled by default.
+* if the schema of the alpha field must change in an incompatible way, a new field name must be used.
```
DirectXMan12: so, if we're working with a new alpha field `Quuxer`, and, while still in alpha, we want to change the value type from `int` to `resource.Quantity`, we have to abandon the name `Quuxer` forever?  That seems to partially defeat the purpose of alpha...

liggitt: until we generate servers and clients with alpha fields completely absent from the go structs, changing the schema breaks old clients and servers:

* an old client with the inert alpha field `quxxer int` in its go struct would fail to decode a server sending data including the beta field `"quxxer":"1m"`, which breaks forward client compatibility
* an old server with the inert alpha field `quxxer int` in its go struct would fail to decode a client sending data including the beta field `"quxxer":"1m"`, which breaks backward client compatibility

contributors/devel/api_changes.md
deads2k: > gates are for alphas, and we have no strong guarantees for alpha anyway?

If you had an alpha field on a Deployment, then you wanted to turn it off, after you turned it off, you would not be able to touch the deployment again and might not even be able to update status on the objects.  In fact, because of the way validation and GC works, you may not even be able to *delete* the object after turning on the alpha field.

The only "fix" is to rewrite the objects and hope no one races you in.  This becomes really problematic for alpha fields that have reflections in status which are controller maintained.  In such a case, I don't see how you turn alpha back off with this validation requirement.

I'm not saying we can't do it, but the warnings around this need to be explicit.  I doubt that its an obvious repercussion of the approach for many/most people.

liggitt: There is a similar issue with enabling an alpha version... a cluster admin has to be aware that they might have to delete all alpha data before upgrading or they can get stuck with undecodeable content in etcd (if the alpha version was removed in the upgraded version, which is allowed).

I'll update the text to be explicit, but enabling an alpha version or an alpha feature puts more burdens on a cluster admin, and I'm ok with that.

bgrant0607:
Also ref https://github.com/kubernetes/kubernetes/issues/4855

contributors/devel/api_changes.md
bgrant0607: I think we need to mention the old approach (and that it is no longer recommended/allowed for new features), since it will still exist in a number of places for some time to come.

contributors/devel/api_changes.md
```diff
@@ -812,45 +817,104 @@ However, the new feature is not stable enough to be used in a stable version
...
-```
-
-This format allows users to specify the new field, but makes it clear that they
-are using a Alpha feature when they do, since the word `alpha` is in the
-annotation key.
+#### Alpha field in existing API version
+
+The preferred approach adds an alpha field to the existing object, and ensures it is disabled by default.
+
+1. Add a feature gate to the API server to control enablement of the new field (and associated function):
```
bgrant0607: A link to the file or at least a string to search for would be useful.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/util/feature/feature_gate.go

contributors/devel/api_changes.md
bgrant0607: Link to the documentation on optional fields:

https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required

contributors/devel/api_changes.md
```diff
@@ -812,45 +817,104 @@ However, the new feature is not stable enough to be used in a stable version
...
+    
+    ```go
+    // API v6.
+    type Frobber struct {
+      // height ...
+      Height int32  `json:"height" protobuf:"varint,1,opt,name=height"`
+      // param ...
+      Param  string `json:"param" protobuf:"bytes,2,opt,name=param"`
+      // width indicates how wide the object is.
+      // This field is alpha-level and should only be sent to servers that enable the Frobber2D feature.
```
bgrant0607: @mbohlool Would it be possible for us to strip disabled fields out of the OpenAPI spec?

contributors/devel/api_changes.md
bgrant0607: Rejection when specifying an unknown field is incompatible with expected behavior.

Strong schema validation needs to be optional, and I don't see a reason to treat disabled alpha fields differently. Discussed further here:
https://github.com/kubernetes/kubernetes/issues/5889

More consistent behavior would be to ensure that the contents were dropped when the field is disabled.

contributors/devel/api_changes.md
bgrant0607: This is admittedly a compromise, but the alternative discussed was to always prefix alpha fields with `alpha`, which would always incur breakage when transitioning out of alpha, as it is not possible to support multiple representations of the same attribute in the same API version.

https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#backward-compatibility-gotchas

contributors/devel/api_changes.md
bgrant0607: It was recently pointed out to me that we don't really document this API versioning convention anywhere.

bgrant0607:
Thanks, @liggitt. I think the only major issue is what to do when a value is specified for a disabled field. IMO, we need to drop the contents, silently by default, as with entirely unrecognized fields.

contributors/devel/api_changes.md
liggitt: > Rejection when specifying an unknown field is incompatible with expected behavior.

> More consistent behavior would be to ensure that the contents were dropped when the field is disabled.

good point, will update, though I'll have to think about where we'd recommend the clearing be done, especially for fields in reused types (like PodSpec). There's not a great single place to clear the field based on a feature gate:
* PrepareForCreate/PrepareForUpdate would have to be set in every resthandler that uses the type, and it'd be tough to ensure we didn't miss any
* Validation functions shouldn't mutate
* defaulting or conversion would affect components other than the API server

contributors/devel/api_changes.md
mbohlool: Do we want to remove them if the Frobber2D feature is disabled? conditional removal of the fields is possible but it needs a design. What comes to my mind right now is to provide a set of feature flags in OpenAPI generators config (e.g. `map[string]bool{"Fordber2D": true, ...}`) and add comment tags to these fields (e.g. `// +openapi_gen:enableOn(Forbber2D)`).

contributors/devel/api_changes.md
liggitt: > More consistent behavior would be to ensure that the contents were dropped when the field is disabled.

There are three types of alpha fields:
1. new additive fields an old server would ignore while still accepting the rest of the object.
2. new fields in a union type (e.g. local persistent volume source) that would cause an old server to reject the object (since it did not have a recognized field in the union type)
3. new values for an enumerated field an old server would reject.

I think only the first type should be dropped. The other two should fail validation.

liggitt: Opened https://github.com/kubernetes/kubernetes/pull/50924 to drop contents of disabled alpha fields

liggitt:
Updated with links to other docs, updated guidance to clear disabled alpha fields before storing

bgrant0607:
Thanks.
/lgtm
/approve

PR #871, opened 2017-08-02
Updated interface sketches to match implementation

ConnorDoyle:
- Added a note about the CPUPressure node condition.
- Added a note about when pod evictions are necessary.

### DIFF
--- a/contributors/design-proposals/cpu-manager.md
+++ b/contributors/design-proposals/cpu-manager.md
@@ -65,7 +65,7 @@ This new method is invoked from two places in the CPU manager: during each
 call to `RegisterContainer` and also periodically from a separate
 reconciliation loop.
 
-![cpu-manager-block-diagram](https://user-images.githubusercontent.com/379372/28443427-bf1b2972-6d6a-11e7-8acb-6cbe9013ac28.png)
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/cpu-manager.md
```diff
@@ -329,6 +331,16 @@ func (p *dynamicPolicy) UnregisterContainer(s State, containerID string) error {
    directly from the shared pool, is too simplistic.
     1. Mitigation: defer supporting this until a new policy tailored for
        use with `isolcpus` can be added.
+1. CPU exhaustion. If all available CPUs are allocated exclusively,
+   additional steps must be taken to remove any best-effort and
+   zero-CPU burstable pods from the node. In addition, the system must
```
sjenning: might just reduce

> best-effort and zero-CPU burstable pods

to

> pods with no CPU request, referred to here as 'no-CPU pods'"

Then you can use the no-CPU term in place of zero-CPU.  no-CPU is more accurate in my mind.

sjenning:
Other than that one comment, lgtm :+1: thanks!

contributors/design-proposals/cpu-manager.md
ConnorDoyle: To be more accurate, it's pods with at least one no-CPU container. Will update.

PR #880, opened 2017-08-07
Add wg-container-identity

smarterclayton:
Needs generated contents


sigs.yaml
```diff
@@ -661,3 +661,23 @@ workinggroups:
...
+      company: Google
+    meetings:
+    - day: Tuesday
+      utc: "15:00"
+      frequency: bi-weekly (On demand)
+    meeting_url: TBD
+    meeting_archive_url: https://docs.google.com/document/d/1bCK-1_Zy2WfsrMBJkdaV72d2hidaxZBhS5YQHAgscPI/edit
+    contact:
+      slack: sig-auth
+      mailing_list: https://groups.google.com/forum/#!forum/kubernetes-sig-auth
```
jbeda: Working groups should have their own mailing lists and slack channels. They cut across SIGs.

smarterclayton: OK, these are placeholders until they are created.

smarterclayton:
Updated with generated code and wg specific lists.

PR #881, opened 2017-08-07
API-machinery performance SLIs/SLOs

wojtek-t:
@kubernetes/sig-scalability-api-reviews @kubernetes/sig-api-machinery-api-reviews  @lavalamp


resouer:
Great work, I did collected some feedback from big companies about performance of CRD, controllers etc.

sig-scalability/slis/apimachinery_slis.md
```diff
@@ -0,0 +1,199 @@
...
+
+Below we introduce all SLIs and SLOs we would like to have in the api-machinery
+area. A bunch of those are not easy to understand for users, as they are
+designed for developers or performance tracking of higher level
+user-understandable SLOs. The user-oriented one (which we want to publicly
+announce) are additionally highlighted with bold.
+
+### Prerequisite
+
+Kubernetes cluster is available and serving.
```
resouer: The first questions is: do we need to define the target cluster size first? For example, below SLO applies to 2k nodes?

wojtek-t: > The first questions is: do we need to define the target cluster size first? For example, below SLO applies to 2k nodes?

It's not only cluster size. There are a bunch of things that needs to be bounded. This is the doc that is initial attempt to address it:
https://github.com/kubernetes/community/blob/master/sig-scalability/thresholds.md

We are missing a bit more background for this doc, which is here: https://docs.google.com/document/d/15rD6XBtKyvXXifkRAsAVFBqEGApQxDRWM3H1bZSBsKQ/edit?pli=1#heading=h.ahomqwpyjhso
I will try to move parts of that doc to .md to provide more context some time in the future.

> And also, I am afraid using cloud provider VS baremetal also matters here. We may also need to clarify the cluster details first?

Cloud-provider VS baremetal doesn't matter for this particular thing at all. Though obviously how the cluster is set up matters (we won't satisfy SLO if e.g. master components will not have enough CPU). 
However, those aspects are being addressed in this doc:
https://github.com/kubernetes/community/blob/master/sig-scalability/provider-configs.md

There obviously might be missing informations there, but setup definitions is not in scope for this doc.


So to summarize - you have valid questions, but both are supposed to be addressed in separate docs (their initial versions already exist). We definitely can't put everything into a single doc, because then this doc won't be managable (I think we can easily get to 100 pages relatively soon).

resouer: SGTM

resouer:
Update with some detail, one of the users' case is: a TPR controller (CRD in future), has a high write OP frequency to apiserver, with qps like 50-100. The cluster is 1k~2k nodes.

I think this is related to watch SLI/SLO, how we can define something to tell user like "yes, your controller is safe", or "sorry, you may need to restrict your write frequency or merge write Ops"?

I also believe in controller model, object data size may also affect the performance (i.e. JSON marshal cost), will it be put into consideration?

I will input more details next week (after talk more with eng team) here but please don't see this as a blocker, just kind of brain storm.

gmarek:
You can drop 'verb' from SLI for List. LGTM otherwise.

sig-scalability/slis/apimachinery_slis.md
```diff
@@ -0,0 +1,199 @@
+# API-machinery SLIs and SLOs
+
+The document was converted from [Google Doc]. Please refer to the original for
+extended commentary and discussion.
+
+***TODO(wojtek-t,deads2k,lavalamp): Get to consensus if we want to define latency
```
deads2k: I still think that for non-long-running-requests last byte matter more to clients/users than first byte.

@smarterclayton last byte is what our clients/controllers seem sensitive to today, right?

sig-scalability/slis/apimachinery_slis.md
```diff
@@ -0,0 +1,199 @@
...
+__***SLI1: Non-streaming API calls for multiple objects (LIST) latency for
+every (resource, verb) pair, measure as 99th percentile over last 5 minutes***__
+
+__***SLI2: 99th percentile for (resource, verb) pairs [excluding virtual and
+aggregated resources and Custom Resource Definitions] combined***__
+
+__***SLO1: In default Kubernetes installation, 99th percentile of SLI2 per
+cluster-day***__
+- __***is <= 1s if total number of objects of the same type as resource in the
+system <= X***__
```
deads2k: Do you have guesses for these or will we shoot for finding them experimentally?

wojtek-t: I have guesses, but I would prefer to experiment more with that.
I would like to avoid providing them now and then finding that they are e.g. too agreesive and changing them.

Does it make sense?

sig-scalability/slis/apimachinery_slis.md
```diff
@@ -0,0 +1,199 @@
...
+### Latency<sup>[1](#footnote1)</sup> of API calls for single objects
+
+__***SLI1: Non-streaming API calls for single objects (POST, PUT, PATCH, DELETE,
+GET) latency for every (resource, verb) pair, measured as 99th percentile over
+last 5 minutes***__
+
+__***SLI2: 99th percentile for (resource, verb) pairs \[excluding virtual and
+aggregated resources and Custom Resource Definitions\] combined***__
+
+__***SLO: In default Kubernetes installation, 99th percentile of SLI2
```
deads2k: We enable RBAC in new installation by default now (at least @cjcullen was trying again as I recall).  I'm fine with gathering this with RBAC on, but it might actually affect the result.

sig-scalability/slis/apimachinery_slis.md
deads2k: > I would like to avoid providing them now and then finding that they are e.g. too agreesive and changing them.
> 
> Does it make sense?

Yes. I just wanted to make sure it was intentional.

deads2k:
I had a couple questions, but overall it looks good.  I do think that last byte latency matters more to clients and users than first byte latency.

sig-scalability/slis/apimachinery_slis.md
wojtek-t: Yes - I know we are enabling RBAC. And I think that should be included (as this is what we recommend).
That said, I don't think this impact of it should be significant (there will be some, but it shouldn't make things that much slower so we won't be able to satisfty the target).

Or you don't agree with that?

sig-scalability/slis/apimachinery_slis.md
smarterclayton: Last-byte is all that matters for any client receiving lists of things, because the client can't start any work for an object until the last byte arrives.  That'll be true even with paging.

sig-scalability/slis/apimachinery_slis.md
deads2k: > That said, I don't think this impact of it should be significant (there will be some, but it shouldn't make things that much slower so we won't be able to satisfty the target).
> 
> Or you don't agree with that?

I agree.  I just wanted to know what to expect.

sig-scalability/slis/apimachinery_slis.md
wojtek-t: OK - so I removed this TODO.

wojtek-t:
@deads2k - I think all comments are answered/addressed now.

PTAL

PR #882, opened 2017-08-07
api-conventions: status subresource updates metadata

nikhita:
Currently, the docs say that the `/status` sub-resource only updates the status of the resource. However, it also updates the metadata.

Examples:
1. [ReplicationController](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/replicationcontroller/strategy.go#L173)
2. [StatefulSet](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/apps/statefulset/strategy.go#L140)
3. [Deployment]( https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/extensions/deployment/strategy.go#L111)
4. [Job](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/batch/job/strategy.go#L157)

### DIFF
--- a/contributors/devel/api-conventions.md
+++ b/contributors/devel/api-conventions.md
@@ -1,7 +1,7 @@
 API Conventions
 ===============
 
-Updated: 3/7/2017
+Updated: 8/7/2017
 
 *This document is oriented at users who want a deeper understanding of the
 Kubernetes API structure, and developers wanting to extend the Kubernetes API.
@@ -272,7 +272,7 @@ desired state.
 
 The PUT and POST verbs on objects MUST ignore the "status" values, to avoid
 accidentally overwriting the status in read-modify-write scenarios. A `/status`
-subresource MUST be provided to enable system components to update statuses of
+subresource MUST be provided to enable system components to update statuses and metadata of
 resources they manage.
 
 Otherwise, PUT expects the whole object to be specified. Therefore, if a field


nikhita:
cc @sttts

contributors/devel/api-conventions.md
```diff
@@ -272,7 +272,7 @@ desired state.
 
 The PUT and POST verbs on objects MUST ignore the "status" values, to avoid
 accidentally overwriting the status in read-modify-write scenarios. A `/status`
-subresource MUST be provided to enable system components to update statuses of
+subresource MUST be provided to enable system components to update statuses and metadata of
```
liggitt: there is debate over whether this is desired, especially as the amount of data in `metadata` is expanded (see https://github.com/kubernetes/kubernetes/pull/45552). I don't think status subresources should be required to expose all of metadata to modification

nikhita: @liggitt I also noticed https://github.com/kubernetes/kubernetes/issues/45539.

If I were to create a new `/status` subresource (this is with respect to CustomResources), do you think it should touch metadata then?

cc @deads2k

deads2k: > If I were to create a new /status subresource (this is with respect to CustomResources), do you think it should touch metadata then?
> 
> cc @deads2k

No.  I don't think `/status` should touch objectmeta.  See https://github.com/kubernetes/kubernetes/pull/45552 as a possible means of preventing it.

idvoretskyi:
cc @kubernetes/api-reviewers @kubernetes/sig-api-machinery-api-reviews

smarterclayton:
At the current time, we have to support status update of object meta for existing apis.  There is some discussion about future changes that would not make that mandatory.  Since it's not clear we want to make it mandatory (although consistency in this case is likely to outweigh the arguments against), I'm going to close this for now.  Sorry it languished.

PR #885, opened 2017-08-08
Update Example conventions

tcharding:
Currently code in `kubectl` for Example strings is not uniform and often does not follow the conventions. This PR updates the conventions to be more descriptive. If this PR is merged I will audit all Example strings in `kubernetes/pkg/kubectl/cmd/*.go` and bring them into line with the conventions. Trivial I know ;)

### DIFF
--- a/contributors/devel/kubectl-conventions.md
+++ b/contributors/devel/kubectl-conventions.md
@@ -255,8 +255,10 @@ input, output, commonly used flags, etc.
 
 
   * Example should contain examples
+    * A comment should precede each example command. Comment should start with
+      an uppercase letter and terminate with a period
     * Start commands with `$`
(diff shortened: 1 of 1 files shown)


cblecker:
@kubernetes/sig-cli-pr-reviews

fabianofranz:
LGTM, thanks @tcharding!

TIP: note that [in this file](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/util/sanity/cmd_sanity.go) we have a few checks for conventions in the CLI. Those run on `make verify`, meaning that if people try to add/change CLI commands that don't pass these tests, they don't get merged. So when you do the auditing, you might want to add a check there for this convention.  ;)

PR #887, opened 2017-08-09
rules review API: what can I do?

ericchiang:
xref https://github.com/kubernetes/features/issues/380

cc @kubernetes/sig-auth-proposals @kubernetes/api-reviewers @xilabao

### DIFF
--- a/contributors/design-proposals/rules-review-api.md
+++ b/contributors/design-proposals/rules-review-api.md
@@ -1,0 +1,128 @@
+# "What can I do?" API
+
+Author: Eric Chiang (eric.chiang@coreos.com)
+
+## Overview
+
+Currently, to determine if a user is authorized to perform a set of actions, that user has to query each action individually through a `SelfSubjectAccessReview`.
+
+Beyond making the authorization layer hard to reason about, it means web interfaces such as the OpenShift Web Console, Tectonic Console, and Kubernetes Dashboard, have to perform individual calls for _every resource_ a page displays. There's no way for a user, or an application acting on behalf of a user, to ask for all the permissions a user can make in bulk. This makes its hard to build pages that are proactive about what's displayed or grayed out based on the end user's permissions. UIs can only handle 403 responses after a user has already performed a forbidden action.
+
+This is a proposal to add authorization APIs that allow a client to determine what actions they can make within a namespace. We expect this API to be used by UIs to show/hide actions, or to quickly let an end user reason about their permissions. This API should NOT be used by external systems to drive their own authorization decisions, as this raises confused deputy, cache lifetime/revocation, and correctness concerns. The `*AccessReview` APIs remain the correct way to defer authorization decisions to the API server.
+
+OpenShift adopted a [`RulesReview` API][openshift-rules-review] to accomplish this same goal, and this proposal is largely a port of that implementation.
+
+[kubernetes/kubernetes#48051](https://github.com/kubernetes/kubernetes/pull/48051) implements most of this proposal.
+
+## API additions
+
+Add a top level type to the `authorization.k8s.io` API group called `SelfSubjectRulesReview`. This mirrors the existing `SelfSubjectAccessReview`.
+
+```
+type SelfSubjectRulesReview struct {
+	metav1.TypeMeta
+
+	Spec SelfSubjectRulesReviewSpec
+
+	// Status is filled in by the server and represents the set of actions a user can perform.
+	Status SubjectRulesReviewStatus
+}
+
+type SelfSubjectRulesReviewSpec struct {
+	// Namespace to evaluate rules for. Required.
+	Namespace string
+}
+
+type SubjectRulesReviewStatus struct {
+	// ResourceRules is the list of actions the subject is allowed to perform on resources.
+	// The list ordering isn't significant, may contain duplicates, and possibly be incomplete.
+	ResourceRules []ResourceRule
+	// NonResourceRules is the list of actions the subject is allowed to perform on non-resources.
+	// The list ordering isn't significant, may contain duplicates, and possibly be incomplete.
+	NonResourceRules []NonResourceRule
+	// EvaluationError can appear in combination with Rules. It indicates an error occurred during
+	// rule evaluation, such as an authorizer that doesn't support rule evaluation, and that
+	// ResourceRules and/or NonResourceRules may be incomplete.
+	EvaluationError string
+    // Incomplete indicates that the returned list is known to be incomplete.
+    Incomplete bool
+}
+```
+
+The `ResourceRules` and `NonResourceRules` rules are similar to the types use by RBAC and the internal authorization system.
+
+```
+# docstrings omitted for brevity. 
+type ResourceRule struct {
+	Verbs []string
+	APIGroups []string
+	Resources []string
+	ResourceNames []string
+}
+
+type NonResourceRule struct {
+	Verbs []string
+	NonResourceURLs []string
+}
+```
+
+All of these fields can include the string `*` to indicate all values are allowed.
+
+### Differences from OpenShift: user extras vs. scopes
+
+OpenShift `SelfSubjectRulesReviewSpec` takes a set of [`Scopes`][openshift-scopes]. This lets OpenShift clients use the API for queries such as _"what could I do if I provide this scope to limit my credentials?"_
+
+ In core kube, scopes are replaced by "user extras" field, a map of opaque strings that can be used for implementation specific user data. Unlike OpenShift, where scopes are always used to restrict credential powers, user extras are commonly used to expand powers. For example, the proposed [Keystone authentiator][keystone-authn] used them to include additional roles and project fields.
+
+Since user extras can be used to expand the power of users, instead of only restricting, this proposal argues that `SelfSubjectRulesReview` shouldn't let a client specify them like `Scopes`. It wouldn't be within the spirit of a `SelfSubject` resource to let a user determine information about other projects or roles.
+
+This could hopefully be solved by introducing a `SubjectRulesReview` API to query the rules for any user. An aggregated API server could use the `SubjectRulesReview` to back an API resource that let a user provide restrictive user extras, such as scopes.
+
+## Webhook authorizers
+
+Some authorizers live external to Kubernetes through an API server webhook and wouldn't immediately support a rules review query.
+
+To communicate with external authorizers, the following types will be defined to query the rules for an arbitrary user. This proposal does NOT propose adding these types to the API immediately, since clients can use user impersonation and a `SelfSubjectRulesReview` to accomplish something similar.
+
+```
+type SubjectRulesReview struct {
+	metav1.TypeMeta
+
+	Spec SubjectRulesReviewSpec
+
+	// Status is filled in by the server and indicates the set of actions a user can perform.
+	Status SubjectRulesReviewStatus
+}
+
+type SubjectRulesReviewSpec struct {
+	// Namespace to evalue rules for. Required.
+	Namespace string
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/rules-review-api.md
```diff
@@ -0,0 +1,126 @@
...
+
+ In core kube, scopes are replaced by "user extras" field, a map of opaque strings that can be used for implementation specific user data. Unlike OpenShift, where scopes are always used to restrict credential powers, user extras are commonly used to expand powers. For example, the proposed [Keystone authentiator][keystone-authn] used them to include additional roles and project fields.
+
+Since user extras can be used to expand the power of users, instead of only restricting, this proposal argues that `SelfSubjectRulesReview` shouldn't let a client specify them like `Scopes`. It wouldn't be within the spirit of a `SelfSubject` resource to let a user determine information about other projects or roles.
+
+This could hopefully be solved by introducing a `SubjectRulesReview` API to query the rules for any user. An aggregated API server could use the `SubjectRulesReview` to back an API resource that let a user provide restrictive user extras, such as scopes.
+
+## Webhook authorizers
```
ericchiang: @cjcullen @dims for proposed webhook integration.

contributors/design-proposals/rules-review-api.md
```diff
@@ -0,0 +1,126 @@
...
+
+type NonResourceRule struct {
+	Verbs []string
+	NonResourceURLs []string
+}
+```
+
+All of these fields can include the string `*` to indicate all values are allowed.
+
+### Differences from OpenShift: user extras vs. scopes
```
ericchiang: @deads2k tired to articulate my concerns about making user extras configurable. ptal

contributors/design-proposals/rules-review-api.md
```diff
@@ -0,0 +1,126 @@
...
+
+OpenShift `SelfSubjectRulesReviewSpec` takes a set of [`Scopes`][openshift-scopes]. This lets OpenShift clients use the API for queries such as _"what could I do if I provide this scope to limit my credentials?"_
+
+ In core kube, scopes are replaced by "user extras" field, a map of opaque strings that can be used for implementation specific user data. Unlike OpenShift, where scopes are always used to restrict credential powers, user extras are commonly used to expand powers. For example, the proposed [Keystone authentiator][keystone-authn] used them to include additional roles and project fields.
+
+Since user extras can be used to expand the power of users, instead of only restricting, this proposal argues that `SelfSubjectRulesReview` shouldn't let a client specify them like `Scopes`. It wouldn't be within the spirit of a `SelfSubject` resource to let a user determine information about other projects or roles.
```
deads2k: Unless there's an authorizer that actually conditionally allows access to particularly named resources based one their actual, I don't think this allows you to go fishing in other namespaces.  I don't care enough to try to stop the good bits of the API, but is there any other way?

erictune:
There is not enough discussion, for my taste, of how a client (e.g. dashboard) would use the results of this API.  Is it just so that it can hide/inactivate UI options that are not possible, or would it actually make its own authorization decisions and then act as a more powerful user?  What client libraries will be provided to help the client make correct decisions from the cached results of the rules review?  How long should clients cache the rules review?  Is this just a glorified form of caching, and why not just add more pure caching of AccessReview decisions?

bryk:
CC @kubernetes/dashboard-maintainers

liggitt:
In Openshift, we use it to show/hide actions in a UI based on the user's permissions:

Example "can I" function:
https://github.com/openshift/origin-web-common/blob/master/src/services/authorizationService.js#L102-L131

Example use:
https://github.com/openshift/origin-web-console/blob/2fb62238b922bebd311ccb43bc00c39500452828/app/scripts/directives/overview/serviceInstanceRow.js#L59-L73

In a dashboard with many different actions, some of which involve multiple API calls, determing those permissions via individual SAR checks was far more expensive (both in terms of API requests and in terms of server-side work)

cc @spadgett @jwforres

erictune:
When you say "server side" do you mean the UI server?  Are you saying that
the rules get loaded into javascript in the users browser to show, and
cause actions to show/hide in the browser.

Then, when the user takes an action, is the final decision still made by
the UI-server or by the apiserver?

liggitt:
> When you say "server side" do you mean the UI server? 

No, the apiserver that would have evaluated those dozens of subjectaccessreview requests.

> Are you saying that the rules get loaded into javascript in the users browser to show, and cause actions to show/hide in the browser.

Yes

> Then, when the user takes an action, is the final decision still made by the UI-server or by the apiserver?

The final decision is the API servers. This lets the UI avoid showing the user actions and wizards and pages that would just result in 403 errors from the API

erictune:
Cool.  I like it.  Would you consider adding these sentences or something
like them to the doc:

We do recommend is that the rules API be used for UIs to show/hide actions.
We don't recommend that you have a highly privileged server which uses the
output of the rules API to make its own authz decisions, as this raises
confused deputy, cache lifetime/revokation, and correctness concerns.
Preferred pattern is that the client (e.g. javascript in browser) still
makes calls directly to the apiserver and is authorized by apiserver.

contributors/design-proposals/rules-review-api.md
ericchiang: > Unless there's an authorizer that actually conditionally allows access to particularly named resources based one their actual

From what I can see @dims' keystone authorizer does that https://github.com/dims/k8s-keystone-auth

@cjcullen any other insight into how/if GKE user extra?

Maybe we could re-word user-extra to try to emphasis its usage as part of this change?

> Extra fields should NOT be used to grant access to new namespaces. Only to modify the
> access of a user within a namespace.

deads2k: > Extra fields should NOT be used to grant access to new namespaces. Only to modify the
> access of a user within a namespace.

Granting access to other namespaces does not leak information, since if I lie about my extra and see "yes you have access to namespace X", I'd get that result regardless of whether the namespace exists.

ericchiang: > Granting access to other namespaces does not leak information, since if I lie about my extra and see "yes you have access to namespace X", I'd get that result regardless of whether the namespace exists.

You're still leaking information about what users have been given access to what though, which could correlate to what exists. I could find out that project `team-foo` doesn't have any powers in namespace `secret-project-bar` but does have powers in namespace `secret-project-spam`.

It's probably not the biggest concern, but it feels weird to allow a user to control any fields that extend their access.

Can we let clients provide something like an additive "Impersonate-Extra-*" header that would only add extras to the authenticated user info? That way an admin could declare what extras are okay to freely add to a user:

```
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  name: scopes-adder
- apiGroups: ["authentication.k8s.io"]
  resources: ["userextras/scopes"]
  verbs: ["impersonate:additive"] # this is a bad name for the verb
```

This would let them use both `SelfSubjectRulesReview` with whatever extras an admin deemed okay, and would extend those capabilities to other commands.

ericchiang:
Thanks @erictune. Expanded the bit about how UIs might use this and tried to make it clear that this API shouldn't be used to drive external authz.

contributors/design-proposals/rules-review-api.md
```diff
@@ -0,0 +1,126 @@
...
+
+	Spec SelfSubjectRulesReviewSpec
+
+	// Status is filled in by the server and represents the set of actions a user can perform.
+	Status SubjectRulesReviewStatus
+}
+
+type SelfSubjectRulesReviewSpec struct {
+	// Namespace to evaluate rules for. Required.
+	Namespace string
```
rf232: Having this all be dependent on the user knowing which Namespaces they have some form of access to might be quite inconvenient for a UI. 
e.g. the kubernetes/dashboard currently (1) queries for the available namespaces and then (2) allows viewing the resources in each of those namespaces. If I understand the proposal correctly this would support the (2) part of this flow, but no way to support for the (1) part of this flow.
Is there room for a non namespaced version of this Spec, or a "Which namespaces have I more than zero access to" api? This would be a helpful part of this API for the dashboard

ericchiang: @rf232 for some background querying resources across all namespaces was deemed too expensive.

I know the actual namespace object is a bit special. I think users can "GET" the namespace object resource by name if they have any access in it already (couldn't find the code path that allows this). CoreOS has internally built apps that do something like:

* privileged service account lists all namespace
* looks up if a user can get each specific namespace
* builds list to display user from that

Agree that probably doesn't work at scale, but a "what namespaces can I see" might be a different API. I think openshift just implemented namespace filtering, so there's no prior art there.

liggitt: > Is there room for a non namespaced version of this Spec

Yes, but it wouldn't do what you describe, it would tell you the things you could do across *all* namespaces (e.g. `kubectl get pods --all-namespaces`)

> or a "Which namespaces have I more than zero access to" api? This would be a helpful part of this API for the dashboard

Agree it is useful (we built something close to that API in openshift), but it's a different API than this one and requires more capabilities from the authorization interface to implement performantly

> I think users can "GET" the namespace object resource by name if they have any access in it already

Not quite, the admin/edit/view roles simply grant "get" "namespace", so when those roles are bound in a particular namespace, the ability to get that namespace is granted.

rf232: > * privileged service account lists all namespace
> * looks up if a user can get each specific namespace
> * builds list to display user from that
This is a possibility, but recently at the dashboard we were looking at reducing the privileges the dashboard itself needs to lower the risk of privilege escalations with the dashboard.

lpabon:
LGTM. Suggestion: Is it possible to add a client-server workflow example?

contributors/design-proposals/rules-review-api.md
```diff
@@ -0,0 +1,126 @@
+# "What can I do?" API
+
+Author: Eric Chiang (eric.chiang@coreos.com)
+
+## Overview
+
+Currently, to determine if a user is authorized to perform a set of actions, that user has to query each action individually thorough a `SelfSubjectAccessReview`.
```
liggitt: typo: through

contributors/design-proposals/rules-review-api.md
```diff
@@ -0,0 +1,126 @@
...
+
+ In core kube, scopes are replaced by "user extras" field, a map of opaque strings that can be used for implementation specific user data. Unlike OpenShift, where scopes are always used to restrict credential powers, user extras are commonly used to expand powers. For example, the proposed [Keystone authentiator][keystone-authn] used them to include additional roles and project fields.
+
+Since user extras can be used to expand the power of users, instead of only restricting, this proposal argues that `SelfSubjectRulesReview` shouldn't let a client specify them like `Scopes`. It wouldn't be within the spirit of a `SelfSubject` resource to let a user determine information about other projects or roles.
+
+This could hopefully be solved by introducing a `SubjectRulesReview` API to query the rules for any user. An aggregated API server could use the `SubjectRulesReview` to back an API resource that let a user provide restrictive user extras, such as scopes.
+
+## Webhook authorizers
+
+Some authorizers are live external to Kubernetes through an API server webhook and wouldn't immediately support a rules review query.
```
liggitt: "Some authorizers live external"

ericchiang:
@kubernetes/sig-auth-proposals I'd like to merge this. SelfSubjectRulesReview was introduced in 1.8 and the webhook support has been reviewed https://github.com/kubernetes/kubernetes/pull/53922 :)

deads2k:
This represents the current state.  I think updates may be required in light of https://github.com/kubernetes/kubernetes/pull/53922, but we can start here.

/lgtm

PR #888, opened 2017-08-09
Vaultproviderspec (#1)

kksriram:
* Initial version of Vault KMS Provider
* Renamed image file to be consistent with spec name
* Brought spec inline with EnvelopeTransformer interfaces in PR 49350
* Wrap at 80


kksriram:
check CLA again

kksriram:
Proposal for Issue https://github.com/kubernetes/kubernetes/issues/49817

deads2k:
I'm very interested in having a second example of a KMS provider.

@kubernetes/sig-api-machinery-feature-requests @kubernetes/sig-auth-feature-requests

kksriram:
This is the same content that was socialized via the [google doc](https://docs.google.com/document/d/15-baW4i7qws1yxxIYjHXqKpk259ebauQbECQCpPD308/edit#heading=h.67bgmqyjswzf) at sig-auth 7/27, with a few updates to bring it inline with @sakshamsharma 's changes.

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,302 @@
...
+Here is a sample configuration file with the vault provider configured:
+
+	kind: EncryptionConfig
+	apiVersion: v1
+	resources:
+	  - resources:
+	    - secrets
+	    providers:
+	    - kms:
+	      name: vault
```
sakshamsharma: Indent the properties of KMS (name, cache size, config file) one more level.

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,302 @@
...
+thereafter will be encrypted by vault provider.
+
+## Performance
+1. KMS provider framework uses LRU cache to minimize the requests to KMS for
+encryption and decryption of DEKs.
+2. Note that there will be a request to KMS for every cache miss causing a
+performance impact. Hence, depending on the cache size, there will be a
+performance impact.
+3. Response time.
+    4. will depend on choice of encryption algorithm and strength.
```
sakshamsharma: Nit: extra tab?

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,302 @@
...
+
+Vault is widely used for Data encryption and securely storing secrets.
+Externalizing encryption/decryption of kubernetes secrets to vault provides
+various benefits
+
+* Choice of industry standard encryption algorithms and strengths without having
+to implement specific providers for each (in K8S).
+* Reduced risk of encryption key compromise.
+    * encryption key is stored and managed in Vault.
+    * encryption key does not need to leave the Vault.
```
sakshamsharma: Nit: extra tab?

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,302 @@
...
+rotated key is identified by key version which is prefix to ciphertext.
+
+A new key can be added in the list. Encryption will be done using the first key
+in the list. Decryption can happen using any of the keys in the list based on
+the prefix to the encrypted DEK stored in etcd
+
+#### Authentication Configuration
+##### Vault Server Authentication
+
+For the Kubernetes cluster to authenticate the vault server, if TLS is enabled :
```
dims: "if TLS is enabled" why optional? let's mandate it!

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,302 @@
...
+	      		key-names:
+	           		- kube-secret-enc-key
+	      		ca-cert:/var/run/kubernetes/ssl/vault.crt
+	      		token: 8dad1053-4a4e-f359-2eab-d57968eb277f
+
+###### Vault AppRole based authentication
+1. ``role-id`` : RoleID of the AppRole
+2. ``secret-id`` : secret Id only if associated with the appRole.
+
+Here's a sample configuration file with Vault AppRole
```
dims: This sample configuration did not render well when i clicked on "view"

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,302 @@
...
+Backend.
+
+### Diagram illustrating interfaces and implementations
+
+![Image of interfaces](vault-based-kms-class-diagram.png)
+
+### Pseudocode
+#### Prefix Metadata
+Every encrypted secret will have the following metadata prefixed.
+``k8s:enc:kms:<api-version>:vault:len(<KEK-key-name>:<KEK-key-version>:<DEK
```
dims: Should all this just be json or yaml?

contributors/design-proposals/vault-based-kms-provider.md
kksriram: Fixed.

contributors/design-proposals/vault-based-kms-provider.md
kksriram: Makes sense. Fixed.

contributors/design-proposals/vault-based-kms-provider.md
kksriram: Fixed.

contributors/design-proposals/vault-based-kms-provider.md
kksriram: Fixed.

contributors/design-proposals/vault-based-kms-provider.md
kksriram: Fixed.

contributors/design-proposals/vault-based-kms-provider.md
kksriram: Fixed.

contributors/design-proposals/vault-based-kms-provider.md
sakshamsharma: Still not fixed. If you look at the file preview, the bullet points are inline and not nested as you wanted.

kksriram:
@destijl You had a question at sig-auth today whether supporting additional Vault Auth Backends will need code changes. 

Right now, of the supported [Auth Backends](https://www.vaultproject.io/docs/auth/index.html), typically its just the three in the current spec that would be used by system processes. The rest would typically be used by humans (which I think doesn't matter in this particular case). 

At first glance, I don't see how we can get away with not changing code when adding other auth backends. 

Talking over with @vineet-garg we will need code changes if supporting new auth back-ends beyond what's in the proposal. I would prefer to focus on a small set of auth backends (the ones in the proposal) and get those in first. Based on interest in the community, we can consider supporting others.

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,276 @@
...
+key(s) locally in a server configuration file. The provider encrypts and
+decrypts secrets in-process. Building upon these, a KMS provider framework with
+an option to support different KMS providers like google cloud KMS is being
+added via PRs [48574](https://github.com/kubernetes/kubernetes/pull/48575) and
+[49350](https://github.com/kubernetes/kubernetes/pull/49350). The new KMS
+provider framework uses an envelope encryption scheme.
+
+
+This proposal adopts the KMS provider framework and adds a new KMS provider that
+uses Hashicorp Vault with a transit backend, to encrypt and decrypt the DEK
```
hongchaodeng: > to encrypt and decrypt the DEK stored in encrypted form in etcd3 along with encrypted secrets.

Is it only dealing with DEK? Or also the secrets? This line is confusing.

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,276 @@
...
+
+Kubernetes, starting with the release 1.7, adds Alpha support ( via PRs
+[41939](https://github.com/kubernetes/kubernetes/pull/41939) and
+[46460](https://github.com/kubernetes/kubernetes/pull/46460)) to encrypt secrets
+and resources in etcd3 via a configured Provider. This release supports three
+providers viz. aesgcm, aescbc, secretbox. These providers store the encryption
+key(s) locally in a server configuration file. The provider encrypts and
+decrypts secrets in-process. Building upon these, a KMS provider framework with
+an option to support different KMS providers like google cloud KMS is being
+added via PRs [48574](https://github.com/kubernetes/kubernetes/pull/48575) and
```
hongchaodeng: > 48574  

Do you mean https://github.com/kubernetes/kubernetes/issues/48522 ?

vineet-garg: the proposal refers to pull request#(48574), what you posted is corresponding issue#(48522).

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,276 @@
...
+Here's a sample configuration file using a Vault AppRole for authentication.
+```
+  key-names:
+    - kube-secret-enc-key
+  addr: https://localhost:8200
+  ca-cert: /var/run/kubernetes/ssl/vault.crt
+  role-id: db02de05-fa39-4855-059b-67221c5c2f63
+```
+
+## Key Generation and rotation
```
hongchaodeng: How is DEK generated?

vineet-garg: That is in envelope.generateKey which is part of existing envelope transformer.

jcbsmpsn:
This [comment](https://github.com/kubernetes/kubernetes/pull/48574\#issuecomment-325417525) on the Google KMS review will affect your implementation. Basically, the expected direction is to implement these integrations out of the Kubernetes repo. The mechanism to accomplish that does not exist yet.

liggitt:
I still think gaining experience with a *small* number of encryption providers in-tree is valuable in designing an out of process plugin mechanism.

kksriram:
@jcbsmpsn @liggitt I haven't caught up on the in-tree v out-of-process discussion over on @sakshamsharma PR 48574. FWIW, my colleague @vineet-garg has an implementation for this spec, that should be coming upstream later this week. We could use that to review what the impact is with a second provider. 

More after I catch up on the discussion on 48574.

deads2k:
> @destijl You had a question at sig-auth today whether supporting additional Vault Auth Backends will need code changes.

@destijl the response https://github.com/kubernetes/community/pull/888#issuecomment-324433021 looked pretty reasonable as a starting point.  Do you have further questions?

kksriram:
I meant to update earlier but slipped my mind. I did catch up on the in-tree vs out-of-tree discussion and [the proposal](https://docs.google.com/document/d/19SbqLrSC_MpmxbLCxasubRJLaZxgqfw1KeIwg3fHHLU/edit#) . 

I agree with @liggitt [earlier comment](https://github.com/kubernetes/community/pull/888#issuecomment-325519221) that gaining experience with a second encryption provider in-tree would help guide the design of the proposal for out-of-tree. 

As such, my preference would be to get the Vault code merged after review, as an in-tree provider and then refactor based on what we learn with the out-of-tree proposal.

deads2k:
This lgtm.  I'm going to see where we'll put it in API machinery, but the KMS plugin structure is unopinionated about whether it is or isn't a cloud provider, so that won't cause any trouble.

lavalamp:
What new (compile-time) dependencies would be required?

This is not really a sustainable way of developing addons/plugins. I would much rather have a process boundary here. I am slightly persuaded by the argument that we should have two concrete uses before generalizing, but only slightly. I would definitely want an agreement that we would force the next person to want to add one of these to build a cross-process API. We can talk about it at tomorrow's sig meeting.

deads2k:
> I would definitely want an agreement that we would force the next person to want to add one of these to build a cross-process API.

Assuming this one goes well and doesn't expose fundamental concerns, I'm ok with that.  Though I may be inclined to force the existing impls out of tree a release or two later to maintain equal footing.

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,276 @@
...
+The KEK is generated in Vault and rotated using direct API call or CLI to Vault
+itself. The Key never leaves the vault.
+
+Note that when a key is rotated, Vault does not allow to choose a different
```
ericchiang: s/Vault does not allow to choose/Vault does not allow choosing/g

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,276 @@
...
+
+#### Minimal required Configuration
+The Vault based Provider needs the following configuration elements, at a
+minimum:
+
+1. ``addr`` Vault service base endpoint eg.  https://example.com:8200
+2. ``key-names`` list of names of the keys in Vault to be used. eg: key-name:
+kube-secret-enc-key.
+
+Note :  key name does not need to be changed if key is rotated in Vault, the
```
ericchiang: s/if key is/if the key is/

contributors/design-proposals/vault-based-kms-provider.md
```diff
@@ -0,0 +1,276 @@
...
+	        name: vault
+	        cachesize: 10
+	        configfile: /home/myvault/vault-config.yaml
+
+#### Minimal required Configuration
+The Vault based Provider needs the following configuration elements, at a
+minimum:
+
+1. ``addr`` Vault service base endpoint eg.  https://example.com:8200
+2. ``key-names`` list of names of the keys in Vault to be used. eg: key-name:
```
ericchiang: Do you need an API version field for the config here to inform the `<api-version>` of the metadata listed above?

contributors/design-proposals/vault-based-kms-provider.md
ericchiang: Worth adding that note to the proposal?

contributors/design-proposals/vault-based-kms-provider.md
kksriram: Probably do. Missed that. Thanks for catching it.

vineet-garg: It is the existing EnvelopeTransformer that is reading the property from configuration file and adding it to metadata, Not the vaultEnvelopeService as mentioned in above : "Of the above metadata..."

vineet-garg:
@lavalamp Compile time dependencies:
github.com/ghodss/yaml
github.com/hashicorp/vault/api

PR #896, opened 2017-08-11
Design for consistent API chunking in Kubernetes

smarterclayton:
In order to reduce the memory allocation impact of very large LIST operations against the Kubernetes apiserver, it should be possible to receive a single large resource list as many individual page requests to the server.

Part of kubernetes/features#365. Taken from kubernetes/kubernetes#49338. Original discussion in kubernetes/kubernetes#2349

### DIFF
--- a/contributors/design-proposals/api-chunking.md
+++ b/contributors/design-proposals/api-chunking.md
@@ -1,0 +1,182 @@
+# Allow clients to retrieve consistent API lists in chunks
+
+On large clusters, performing API queries that return all of the objects of a given resource type (GET /api/v1/pods, GET /api/v1/secrets) can lead to significant variations in peak memory use on the server and contribute substantially to long tail request latency.
+
+When loading very large sets of objects -- some clusters are now reaching 100k pods or equivalent numbers of supporting resources -- the system must:
+
+* Construct the full range description in etcd in memory and serialize it as protobuf in the client
+  * Some clusters have reported over 500MB being stored in a single object type
+  * This data is read from the underlying datastore and converted to a protobuf response
+  * Large reads to etcd can block writes to the same range (https://github.com/coreos/etcd/issues/7719)
+* The data from etcd has to be transferred to the apiserver in one large chunk
+* The `kube-apiserver` also has to deserialize that response into a single object, and then re-serialize it back to the client
+  * Much of the decoded etcd memory is copied into the struct used to serialize to the client
+* An API client like `kubectl get` will then decode the response from JSON or protobuf
+  * An API client with a slow connection may not be able to receive the entire response body within the default 60s timeout
+    * This may cause other failures downstream of that API client with their own timeouts
+  * The recently introduced client compression feature can assist
+  * The large response will also be loaded entirely into memory
+
+The standard solution for reducing the impact of large reads is to allow them to be broken into smaller reads via a technique commonly referred to as paging or chunking. By efficiently splitting large list ranges from etcd to clients into many smaller list ranges, we can reduce the peak memory allocation on etcd and the apiserver, without losing the consistent read invariant our clients depend on.
+
+This proposal does not cover general purpose ranging or paging for arbitrary clients, such as allowing web user interfaces to offer paged output, but does define some parameters for future extension. To that end, this proposal uses the phrase "chunking" to describe retrieving a consistent snapshot range read from the API server in distinct pieces.
+
+Our primary consistent store etcd3 offers support for efficient chunking with minimal overhead, and mechanisms exist for other potential future stores such as SQL databases or Consul to also implement a simple form of consistent chunking.
+
+Relevant issues:
+
+* https://github.com/kubernetes/kubernetes/issues/2349
+
+## Terminology
+
+**Consistent list** - A snapshot of all resources at a particular moment in time that has a single `resourceVersion` that clients can begin watching from to receive updates. All Kubernetes controllers depend on this semantic. Allows a controller to refresh its internal state, and then receive a stream of changes from the initial state.
+
+**API paging** - API parameters designed to allow a human to view results in a series of "pages".
+
+**API chunking** - API parameters designed to allow a client to break one large request into multiple smaller requests without changing the semantics of the original request.
+
+
+## Proposed change:
+
+Expose a simple chunking mechanism to allow large API responses to be broken into consistent partial responses. Clients would indicate a tolerance for chunking (opt-in) by specifying a desired maximum number of results to return in a `LIST` call. The server would return up to that amount of objects, and if more exist it would return a `continue` parameter that the client could pass to receive the next set of results.  The server would be allowed to ignore the limit if it does not implement limiting (backward compatible), but it is not allowed to support limiting without supporting a way to continue the query past the limit (may not implement `limit` without `continue`).
+
+```
+GET /api/v1/pods?limit=500
+{
+  "metadata": {"continue": "ABC...", "resourceVersion": "147"},
+  "items": [
+     // no more than 500 items
+   ]
+}
+GET /api/v1/pods?limit=500&continue=ABC...
+{
+  "metadata": {"continue": "DEF...", "resourceVersion": "147"},
+  "items": [
+     // no more than 500 items
+   ]
+}
+GET /api/v1/pods?limit=500&continue=DEF...
+{
+  "metadata": {"resourceVersion": "147"},
+  "items": [
+     // no more than 500 items
+   ]
+}
+```
+
+The token returned by the server for `continue` would be an opaque serialized string that would contain a simple serialization of a version identifier (to allow future extension), and any additional data needed by the server storage to identify where to start the next range. 
+
+The continue token is not required to encode other filtering parameters present on the initial request, and clients may alter their filter parameters on subsequent chunk reads. However, the server implementation **may** reject such changes with a `400 Bad Request` error, and clients should consider this behavior undefined and left to future clarification. Chunking is intended to return consistent lists, and clients **should not** alter their filter parameters on subsequent chunk reads.
+
+If the resource version parameter specified on the request is inconsistent with the `continue` token, the server **must** reject the request with a `400 Bad Request` error.
+
+The schema of the continue token is chosen by the storage layer and is not guaranteed to remain consistent for clients - clients **must** consider the continue token as opaque. Server implementations **should** ensure that continue tokens can persist across server restarts and across upgrades.
+
+Servers **may** return fewer results than `limit` if server side filtering returns no results such as when a `label` or `field` selector is used. If the entire result set is filtered, the server **may** return zero results with a valid `continue` token. A client **must** use the presence of a `continue` token in the response to determine whether more results are available, regardless of the number of results returned. A server that supports limits **must not** return more results than `limit` if a `continue` token is also returned. If the server does not return a `continue` token, the server **must** return all remaining results. The server **may** return zero results with no `continue` token on the last call.
+
+The server **may** limit the amount of time a continue token is valid for. Clients **should** assume continue tokens last only a few minutes.
+
+The server **must** support `continue` tokens that are valid across multiple API servers. The server **must** support a mechanism for rolling restart such that continue tokens are valid after one or all API servers have been restarted.
+
+
+### Proposed Implementations
+
+etcd3 is the primary Kubernetes store and has been designed to support consistent range reads in chunks for this use case. The etcd3 store is an ordered map of keys to values, and Kubernetes places all keys within a resource type under a common prefix, with namespaces being a further prefix of those keys. A read of all keys within a resource type is an in-order scan of the etcd3 map, and therefore we can retrieve in chunks by defining a start key for the next chunk that skips the last key read.
+
+etcd2 will not be supported as it has no option to perform a consistent read and is on track to be deprecated in Kubernetes.  Other databases that might back Kubernetes could either choose to not implement limiting, or leverage their own transactional characteristics to return a consistent list. In the near term our primary store remains etcd3 which can provide this capability at low complexity.
+
+Implementations that cannot offer consistent ranging (returning a set of results that are logically equivalent to receiving all results in one response) must not allow continuation, because consistent listing is a requirement of the Kubernetes API list and watch pattern.
+
+#### etcd3
+
+For etcd3 the continue token would contain a resource version (the snapshot that we are reading that is consistent across the entire LIST) and the start key for the next set of results. Upon receiving a valid continue token the apiserver would instruct etcd3 to retrieve the set of results at a given resource version, beginning at the provided start key, limited by the maximum number of requests provided by the continue token (or optionally, by a different limit specified by the client). If more results remain after reading up to the limit, the  storage should calculate a continue token that would begin at the next possible key, and the continue token set on the returned list.
+
+The storage layer in the apiserver must apply consistency checking to the provided continue token to ensure that malicious users cannot trick the server into serving results outside of its range. The storage layer must perform defensive checking on the provided value, check for path traversal attacks, and have stable versioning for the continue token.
+
+#### Possible SQL database implementation
+
+A SQL database backing a Kubernetes server would need to implement a consistent snapshot read of an entire resource type, plus support changefeed style updates in order to implement the WATCH primitive. A likely implementation in SQL would be a table that stores multiple versions of each object, ordered by key and version, and filters out all historical versions of an object. A consistent paged list over such a table might be similar to:
+
+    SELECT * FROM resource_type WHERE resourceVersion < ? AND deleted = false AND namespace > ? AND name > ? LIMIT ? ORDER BY namespace, name ASC
+
+where `namespace` and `name` are part of the continuation token and an index exists over `(namespace, name, resourceVersion, deleted)` that makes the range query performant. The highest returned resource version row for each `(namespace, name)` tuple would be returned.
+
+
+### Security implications of returning last or next key in the continue token
+
+If the continue token encodes the next key in the range, that key may expose info that is considered security sensitive, whether simply the name or namespace of resources not under the current tenant's control, or more seriously the name of a resource which is also a shared secret (for example, an access token stored as a kubernetes resource). There are a number of approaches to mitigating this impact:
+
+1. Disable chunking on specific resources
+2. Disable chunking when the user does not have permission to view all resources within a range
+3. Encrypt the next key or the continue token using a shared secret across all API servers
+4. When chunking, continue reading until the next visible start key is located after filtering, so that start keys are always keys the user has access to.
+
+In the short term we have no supported subset filtering (i.e. a user who can LIST can also LIST ?fields= and vice versa), so 1 is sufficient to address the sensitive key name issue. Because clients are required to proceed as if limiting is not possible, the server is always free to ignore a chunked request for other reasons. In the future, 4 may be the best option because we assume that most users starting a consistent read intend to finish it, unlike more general user interface paging where only a small fraction of requests continue to the next page.
+
+
+### Handling expired resource versions
+
+If the required data to perform a consistent list is no longer available in the storage backend (by default, old versions of objects in etcd3 are removed after 5 minutes), the server **must** return a `410 Gone ResourceExpired` status reponse (the same as for watch), which means clients must start from the beginning.
+
+```
+# resourceVersion is expired
+GET /api/v1/pods?limit=500&continue=DEF...
+{
+  "kind": "Status",
+  "code": 410,
+  "reason": "ResourceExpired"
+}
+```
+
+Some clients may wish to follow a failed paged list with a full list attempt.
+
+The 5 minute default compaction interval for etcd3 bounds how long a list can run.  Since clients may wish to perform processing over very large sets, increasing that timeout may make sense for large clusters. It should be possible to alter the interval at which compaction runs to accomodate larger clusters.
+
+
+#### Types of clients and impact
+
+Some clients such as controllers, receiving a 410 error, may instead wish to perform a full LIST without chunking.
+
+* Controllers with full caches
+  * Any controller with a full in-memory cache of one or more resources almost certainly depends on having a consistent view of resources, and so will either need to perform a full list or a paged list, without dropping results
+* `kubectl get`
+  * Most administrators would probably prefer to see a very large set with some inconsistency rather than no results (due to a timeout under load).  They would likely be ok with handling `410 ResourceExpired` as "continue from the last key I processed"
+* Migration style commands
+  * Assuming a migration command has to run on the full data set (to upgrade a resource from json to protobuf, or to check a large set of resources for errors) and is performing some expensive calculation on each, very large sets may not complete over the server expiration window.
+
+For clients that do not care about consistency, the server **may** return a `continue` value on the `ResourceExpired` error that allows the client to restart from the same prefix key, but using the latest resource version.  This would allow clients that do not require a fully consistent LIST to opt in to partially consistent LISTs but still be able to scan the entire working set. It is likely this could be a sub field (opaque data) of the `Status` response under `statusDetails`.
+
+
+### Rate limiting
+
+Since the goal is to reduce spikiness of load, the standard API rate limiter might prefer to rate limit page requests differently from global lists, allowing full LISTs only slowly while smaller pages can proceed more quickly.
+
+
+### Chunk by default?
+
+On a very large data set, chunking trades total memory allocated in etcd, the apiserver, and the client for higher overhead per request (request/response processing, authentication, authorization).  Picking a sufficiently high chunk value like 500 or 1000 would not impact smaller clusters, but would reduce the peak memory load of a very large cluster (10k resources and up).  In testing, no significant overhead was shown in etcd3 for a paged historical query which is expected since the etcd3 store is an MVCC store and must always filter some values to serve a list.
+
+For clients that must perform sequential processing of lists (kubectl get, migration commands) this change dramatically improves initial latency - clients got their first chunk of data in milliseconds, rather than seconds for the full set. It also improves user experience for web consoles that may be accessed by administrators with access to large parts of the system.
+
+It is recommended that most clients attempt to page by default at a large page size (500 or 1000) and gracefully degrade to not chunking.
+
+
+### Other solutions
+
+Compression from the apiserver and between the apiserver and etcd can reduce total network bandwidth, but cannot reduce the peak CPU and memory used inside the client, apiserver, or etcd processes. 
+
+Various optimizations exist that can and should be applied to minimizing the amount of data that is transferred from etcd to the client or number of allocations made in each location, but do not how response size scales with number of entries.
+
+
+## Plan
+
+The initial chunking implementation would focus on consistent listing on server and client as well as measuring the impact of chunking on total system load, since chunking will slightly increase the cost to view large data sets because of the additional per page processing. The initial implementation should make the fewest assumptions possible in constraining future backend storage.
+
+For the initial alpha release, chunking would be behind a feature flag and attempts to provide the `continue` or `limit` flags should be ignored. While disabled, a `continue` token should never be returned by the server as part of a list.
+
+Future work might offer more options for clients to page in an inconsistent fashion, or allow clients to directly specify the parts of the namespace / name keyspace they wish to range over (paging).
+
+
+<!-- BEGIN MUNGE: GENERATED_ANALYTICS -->
+[![Analytics](https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/docs/proposals/server-get.md?pixel)]()
+<!-- END MUNGE: GENERATED_ANALYTICS -->


smarterclayton:
@kubernetes/sig-api-machinery-api-reviews @kubernetes/sig-scalability-api-reviews

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+     // no more than 500 items
+   ]
+}
+```
+
+The token returned by the server for `continue` would be an opaque serialized string that would contain a simple serialization of the initially requested limit, a version identifier (to allow future extension), and any additional data needed by the server storage to identify where to start the next range. 
+
+The schema of the continue token is chosen by the storage layer and is not guaranteed to remain consistent for clients - clients *must* consider the continue token as opaque. Server implementations *should* ensure that continue tokens can persist across server restarts and across upgrades.
+
+Servers *may* return fewer results than `limit` if server side filtering returns no results such as when a `label` or `field` selector is used. If the entire result set is filtered, the server *may* return zero results with a valid `continue` token.
```
liggitt: can you clarify the interaction of the `continue` parameter with the other parameters (resourceVersion, labelSelector/fieldSelector) passed to `list` calls? Is the expectation that all parameters present on the initial request are also passed in subsequent requests with identical values? Is the server responsible for ensuring that is the case? Paged list results with resourceVersion=0 stands out as having potentially odd implications

smarterclayton: So etcd3 does not require that any of the parameters stay the same - it'll work regardless.  I'm slightly inclined to describe changing field/label selectors as undefined behavior, and that the server may return a 400 bad request if it is not allowed.  I think in general the storage structure required to do paging in a useful way to clients is going to be based on being able to range over an ordered list with snapshot isolation of versions, and if people are implementing paging with things like "holding open database cursors for 5 minutes" they're going to come to tears anyway.

Specifying resourceVersion inconsistent with continue should be a 400 bad request.  Will comment.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+
+Implementations that cannot offer consistent ranging (returning a set of results that are logically equivalent to receiving all results in one response) must not allow continuation, because consistent listing is a requirement of the Kubernetes API list and watch pattern.
+
+#### etcd3
+
+For etcd3 the continue token would contain a resource version (the snapshot that we are reading that is consistent across the entire LIST) and the start key for the next set of results. Upon receiving a valid continue token the apiserver would instruct etcd3 to retrieve the set of results at a given resource version, beginning at the provided start key, limited by the maximum number of requests provided by the continue token (or optionally, by a different limit specified by the client). If more results remain after reading up to the limit, the  storage should calculate a continue token that would begin at the next possible key, and the continue token set on the returned list.
```
liggitt: > and the start key for the next set of results

this has the potential to return information in the continue token that the user would not have received in a filtered list API call

if we gain the ability to authorize clients to only do LIST calls filtered to specific label selectors (c.f. https://github.com/kubernetes/kubernetes/issues/40403, https://github.com/kubernetes/kubernetes/issues/44703), this could expose keys the user doesn't actually have permission to know about

smarterclayton: Yeah. Namespace names and object names (for those reading at home).  We could encrypt and sign the continuation token but that's another value we have to distribute.  Would also be vulnerable to repeat attacks.  Hash is out because it has to be reversible.

liggitt: > limited by the maximum number of requests provided by the continue token (or optionally, by a different limit specified by the client)

When filtering (via labelSelector/fieldSelector) is done after obtaining the results, there is no guarantee the number of items matching the filter will equal the number retrieved from etcd. the number of matching results returned to the client may be much smaller (or in pathological cases, even 0). Would you expect the server to continue paging internally until it had obtained enough filtered results to satisfy the client's requested limit?

smarterclayton: I think for a first approach no, but leave it up to the impl to be improved.

smarterclayton: This concern could be handled per resource type with the following possibilities:

1. disallow paging on certain resources (a resource whose name was a secret)
2. disallow paging when performing certain filtering queries that would otherwise expose info the user could not see
3. continue paging until at least one visible key is found, and return that as the next key to continue on (i.e. users only ever see keys they would see normally)
4. encrypt the key name using a pre shared secret across masters

I think all of these are possible options and easy to implement.  Will add this to the proposal.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+When loading very large sets of objects -- some clusters are now reaching 100k pods or equivalent numbers of supporting resources -- the system must:
+
+* Construct the full range description in etcd in memory and serialize it as protobuf in the client
+  * Some clusters have reported over 500MB being stored in a single object type
+  * This data is read from the underlying datastore and converted to a protobuf response
+  * Large reads to etcd can block writes to the same range (https://github.com/coreos/etcd/issues/7719)
+* The data from etcd has to be transferred to the apiserver in one large chunk
+* The `kube-apiserver` also has to deserialize that response into a single object, and then re-serialize it back to the client
+  * Much of the decoded etcd memory is copied into the struct used to serialize to the client
+* A client like `kubectl get` will then decode the response from JSON or protobuf
```
sttts: "etcd client", "kube client" would make it easier to read. Many clients involved here.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+  * A client with a slow connection may not be able to receive the entire response body within the default 60s timeout
```
sttts: and those connections with >60s pile up due to retries and non-adaptive back preasure, easily leading to congestion of an apiserver exceeding max-inflight  all the time.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+  * The recently introduced client compression feature can assist
+  * The large response will also be loaded entirely into memory
+
+The standard solution for reducing the impact of large reads is to allow them to be broken into smaller reads via a technique commonly referred to as paging. By efficiently splitting large list ranges from clients to etcd into many smaller list ranges, we can reduce the maximum memory allocation on etcd and the apiserver.
+
+Our primary consistent store etcd3 offers support for efficient paging with minimal overhead, and mechanisms exist for other potential future stores such as SQL databases or Consul to also implement a simple form of paging.
```
sttts: missing the word "consistent" page mechanism, i.e. transactional view

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+
+Relevant issues:
+
+* https://github.com/kubernetes/kubernetes/issues/2349
+
+## Proposed change:
+
+Expose a simple paging mechanism to allow large API responses to be broken into consistent partial responses. Clients would indicate a tolerance for paging (opt-in) by specifying a desired maximum number of results to return in a `LIST` call. The server would return up to that amount of objects, and if more exist it would return a `continue` parameter that the client could pass to receive the next set of results.  The server would be allowed to ignore the limit if it does not implement limiting (backward compatible), but it is not allowed to support limiting without supporting a way to continue the query past the limit (may not implement `limit` without `continue`).
```
sttts: is there a life-time of such a continue parameter? Can I come back tomorrow and continue a paged list from today?

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+GET /api/v1/pods?limit=500&continue=DEF...
+{
+  "metadata": {"resourceVersion": "147"},
+  "items": [
+     // no more than 500 items
+   ]
+}
+```
+
+The token returned by the server for `continue` would be an opaque serialized string that would contain a simple serialization of the initially requested limit, a version identifier (to allow future extension), and any additional data needed by the server storage to identify where to start the next range.
```
sttts: why is the `initially requested limit` important? Sounds like an implementation detail if at all.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+  "metadata": {"resourceVersion": "147"},
+  "items": [
+     // no more than 500 items
+   ]
+}
+```
+
+The token returned by the server for `continue` would be an opaque serialized string that would contain a simple serialization of the initially requested limit, a version identifier (to allow future extension), and any additional data needed by the server storage to identify where to start the next range. 
+
+The schema of the continue token is chosen by the storage layer and is not guaranteed to remain consistent for clients - clients *must* consider the continue token as opaque. Server implementations *should* ensure that continue tokens can persist across server restarts and across upgrades.
```
sttts: and be valid in an HA environment on different servers.

contributors/design-proposals/api-paging.md
sttts: in other words: either the result is empty, we get a full bucket with continue param, or we get the remaining items without continue param. But non-empty, len != limit plus continue is forbidden.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+
+Servers *may* return fewer results than `limit` if server side filtering returns no results such as when a `label` or `field` selector is used. If the entire result set is filtered, the server *may* return zero results with a valid `continue` token.
+
+
+### Specific Implementations
+
+Etcd2 would not be supported as it has no options to perform a consistent read and is on track to be deprecated in Kubernetes.  Other databases that might back Kubernetes could either chose to not implement limiting, or leverage their own transactional characteristics to return a consistent list. In the near term our primary store remains etcd3 which can provide this capability at low complexity.
```
sttts: s/chose/choose/

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+
+Implementations that cannot offer consistent ranging (returning a set of results that are logically equivalent to receiving all results in one response) must not allow continuation, because consistent listing is a requirement of the Kubernetes API list and watch pattern.
```
sttts: missing some words about discovery of this feature above. Do we want this being part of discovery info?

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+
+```
+GET /api/v1/pods?limit=500
+{
+  "metadata": {"continue": "ABC...", "resourceVersion": "147"},
+  "items": [
+     // no more than 500 items
+   ]
+}
+GET /api/v1/pods?limit=500&continue=ABC...
```
sttts: from the "token encodes limit" below I conclude that we can also continue without passing the limit value again?

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### etcd3
+
+For etcd3 the continue token would contain a resource version (the snapshot that we are reading that is consistent across the entire LIST) and the start key for the next set of results. Upon receiving a valid continue token the apiserver would instruct etcd3 to retrieve the set of results at a given resource version, beginning at the provided start key, limited by the maximum number of requests provided by the continue token (or optionally, by a different limit specified by the client). If more results remain after reading up to the limit, the  storage should calculate a continue token that would begin at the next possible key, and the continue token set on the returned list.
+
+The storage layer in the apiserver must apply consistency checking to the provided continue token to ensure that malicious users cannot trick the server into serving results outside of its range. The storage layer must perform defensive checking on the provided value, check for path traversal attacks, and have stable versioning for the continue token.
```
sttts: does this mean we encode the token cryptographically (e.g. via JSON token standard) ?

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+{
+  "kind": "Status",
+  "code": 410,
+  "reason": "ResourceExpired"
+}
+```
+
+Some clients may wish to follow a failed paged list with a full list attempt.
+
+The 5 minute default compaction interval for etcd3 bounds how long a list can run.  Since clients may wish to perform processing over very large sets, increasing that timeout may make sense for large clusters. It should be possible to alter the interval at which compaction runs to accomodate larger clusters.
```
sttts: would a non-consistent continuation make sense? Are there clients which might be happy with this? Would it lessen the burden of the apiserver to always return consistent views?

sttts: discussed below....

contributors/design-proposals/api-paging.md
deads2k: Why would it include the initially requested limit?  That doesn't sound very opaque if we're getting into the details of what's in it.

contributors/design-proposals/api-paging.md
deads2k: > and be valid in an HA environment on different servers.

Seems like this may actually be a must.  Otherwise the feature isn't as useful.

contributors/design-proposals/api-paging.md
deads2k: If the user requests 20 items and the server finds a list to return with 50 items, can the server choose to only send 10 items?

deads2k: How does the server indicate that the `continue` token is no longer valid?

contributors/design-proposals/api-paging.md
deads2k: > continue paging until at least one visible key is found, and return that as the next key to continue on (i.e. users only ever see keys they would see normally)

This seems like the most reasonable.

contributors/design-proposals/api-paging.md
deads2k: > does this mean we encode the token cryptographically (e.g. via JSON token standard) ?

I think you could just validate it.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+#### Types of clients and impact
+
+Some clients such as controllers, receiving a 410 error, may instead wish to perform a full LIST without paging.
+
+* Controllers with full caches
+  * Any controller with a full in-memory cache of one or more resources almost certainly depends on having a consistent view of resources, and so will either need to perform a full list or a paged list, without dropping results
+* `kubectl get`
+  * Most administrators would probably prefer to see a very large set with some inconsistency rather than no results (due to a timeout under load).  They would likely be ok with handling `410 ResourceExpired` as "continue from the last key I processed"
```
deads2k: How would such a client indicate that?

contributors/design-proposals/api-paging.md
duglin: If the client asked for just 5 because it can't handle more, I'm not sure the server should be allowed to ignore the request and flood them with a massive response. I'd prefer an error that tells the client that the server doesn't support paging and then let the client decide what to do next.

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
...
+* Migration style commands
+  * Assuming a migration command has to run on the full data set (to upgrade a resource from json to protobuf, or to check a large set of resources for errors) and is performing some expensive calculation on each, very large sets may not complete over the server expiration window.
+
+For clients that do not care about consistency, the storage *may* return a `continue` value with the `ResourceExpired` error that allows the client to restart from the same prefix key, but using the latest resource version.  This would allow clients that do not require a fully consistent LIST to opt in to partially consistent LISTs but still be able to scan the entire working set.
```
deads2k: may, not must?  Seems generous, but ok.  Move this up closer to where you describe the behavior.

contributors/design-proposals/api-paging.md
duglin: while I agree that to the client its opaque, as part of this design I think it would be good to know how we expect kube servers expect to populate it.  For example, do we expect different data in there based on the resource type?

contributors/design-proposals/api-paging.md
duglin: Is there any guarantee w.r.t. state of the list?  For example, while we assume that the "continute" token tells us where keep going, is there any agreement that the rest of the list is fixed based on when the initial query was done or can it change?  This is an important detail since at some point we may add a 'size' field to the initial response and I don't think we want to change the semantics of the overall query in a future release.  Or, as another example, two queries at the same time but one client is slower to get all chunks - should the results look the same or will the slower one possible get different data because the system changed after the faster one finished.  We just need to be clear on what clients should see.

contributors/design-proposals/api-paging.md
duglin: It might be better/clearer to talk about this outside the scope of filtering because I don't think filtering should impact these points. In particular, I think what's trying to be said is:
- server MAY return less than "limit" on each response, including zero items
- server MUST include a "continue" on each response if it does not consider the paging to be completed
- server MUST NOT include a "continue" on the final response, to indicate the end of the paging

Is that correct?

contributors/design-proposals/api-paging.md
duglin: related to my comment above... can you elaborate on "consistent ranging" ? Does the result set have to be static for the duration of the paging?

contributors/design-proposals/api-paging.md
duglin: I'm wondering whether setting "continue" to the "next key" is the best approach.  Not only because of a possible security implication, but if we get to the point where we're paging over data that is not immediately available to us right away, and the server wants to return "what it has as of now", then it may not know the "next key". I think it would be better to return some "token" that's really only known to the paging implementation of that resource so it can figure out how best to represent "where to continue".

contributors/design-proposals/api-paging.md
duglin: using something other than "the next key" would solve this issue too.

contributors/design-proposals/api-paging.md
smarterclayton: > is there a life-time of such a continue parameter? 

It's up to the backing store to have the historical data available.  So for our default config is 5m for etcd3.

> If the client asked for just 5 because it can't handle more, I'm not sure the server should be allowed to ignore the request and flood them with a massive response. I'd prefer an error that tells the client that the server doesn't support paging and then let the client decide what to do next.

That is not backwards compatible - old servers wouldn't return an error.  We generally err on not introducing new incompatibilities.  Paging is a way to get less results in a chunk, not a way to slowly browse through the data store.  So naive web clients that want to display a chunk of results at a time needs something more involved than simple continuations and would be an additional set of parameters.

contributors/design-proposals/api-paging.md
smarterclayton: It's not actually important, and hadn't planned on implementing that.  You still have to specify limit on each call.

contributors/design-proposals/api-paging.md
smarterclayton:
```
Continue AND 0 < x < limit = allowed
Continue AND x == 0 = allowed
Continue AND x == limit = allowed
No Continue AND x > limit = allowed (backwards compatibility)
Continue AND x > limit = not allowed
```

* Continue being non empty means there are more results
* The server may not return more than Limit if it supports Limit.  
* The server may always return less than or equal to Limit

contributors/design-proposals/api-paging.md
smarterclayton: This version of paging is for processes that expected to get a full list but want it in chunks.  I think the design is biased towards that being an optional implementation.  A future improvement to paging might make it usable for end users who want to walk the tree (supporting `?fromKey=`) but that is not in scope for this design.

contributors/design-proposals/api-paging.md
smarterclayton: You can continue without passing limit, but the intent is that limit is applied per call.  If you continue without specifying limit you'll get the rest.  Limit should be able to be changed so that a client can potentially perform a slow start (reserving the right to do so).  The text about token encodes limit is wrong and will be removed.

contributors/design-proposals/api-paging.md
smarterclayton: Doesn't need the limit, that was part of an earlier draft.  Will remove.

contributors/design-proposals/api-paging.md
smarterclayton: The server may return any number of results `<= limit` at any time, as long as a continue token is also returned.  If the continue token is not set, the server MUST return all remaining results.

smarterclayton: 410 Resource Expired

contributors/design-proposals/api-paging.md
smarterclayton: Will clarify, would probably be a `statusDetails.` sub field.  We probably want to let opaque data be returned on status.

contributors/design-proposals/api-paging.md
smarterclayton: This is about consistent reads, so if you start a LIST and get a continue token, the server MUST return a snapshot that is consistent with the resourceVersion in the initial LIST call.  LIST with paging must be equivalent to LIST without paging from a consistency point of view.  Will clarify earlier.

contributors/design-proposals/api-paging.md
smarterclayton: Correct.

contributors/design-proposals/api-paging.md
smarterclayton: Yes, as per our existing LIST behavior - LIST with paging is an optimization on LIST to avoid space, not a new arbitrary query tool for users.  The mechanisms we use to do LIST with paging in the future could be expanded to allow additional flexibility such as a `?fromKey` field.  Will clarify earlier.

contributors/design-proposals/api-paging.md
liggitt: that's fine, as long as the documentation is clear that when a continue token is returned, it *must* be used to ensure obtaining a complete list (even if fewer results were requested, even possibly 0)

contributors/design-proposals/api-paging.md
hongchaodeng: Shedding some thoughts here :)

Aren't most paging-list things are immutable? For example, names? If use cases are well understood, there are good alternative rather than listing non-consistently.

In a resource, there are frequently updated parts (status) and pretty much unchanged parts (spec, metadata). If they are decoupled, and keep different versions, then it is easy to list pages of names without losing consistency. Right?

smarterclayton:
The lists here care about consistency.  These are for the core controller
watch loops and other systems that must LIST all, but want to do it in
chunks.  So consistent read is required for the core use case.
Inconsistent read is probably something others will want to do later, but
in that case key ranging is sufficient for them.

contributors/design-proposals/api-paging.md
xiang90: Etcd2 -> etcd2

contributors/design-proposals/api-paging.md
xiang90: we could set it as the last key, then increase it for the next range call.

smarterclayton: There's no guarantee the user should be able to see that last key, and if large chunks of entries are being filtered out there's no longer a guarantee that the paging can make progress.

> where we're paging over data that is not immediately available to us right away

That would explicitly not be supported - this is about returning the consistent read in multiple steps, so there is no "data we don't know about yet".

bgrant0607:
cc @cheftako

contributors/design-proposals/api-paging.md
```diff
@@ -0,0 +1,148 @@
+# Reduce the amount of client and server memory allocated during large API lists
+
+On large clusters, performing API queries that return all of the objects of a given resource type (GET /api/v1/pods, GET /api/v1/secrets) can lead to significant variations in peak memory use on the server and contribute substantially to long tail request latency.
+
+When loading very large sets of objects -- some clusters are now reaching 100k pods or equivalent numbers of supporting resources -- the system must:
+
+* Construct the full range description in etcd in memory and serialize it as protobuf in the client
+  * Some clusters have reported over 500MB being stored in a single object type
```
lpabon: Quick question. One of the issues is the memory consumption at the server side, this paper calls for limits to be done at the client. I understand the reason is to be compatible, but that leaves the issue at the server still, if I am correct, because the client may choose not to send the `limit` query. If this is true, I would suggest a model which the server automatically limits the response. The servers know how large the payload would be and could predict how much memory it would consume, optimizing accordingly. I think leaving this to the client may solve only a part of the issue if the client decides to not send the `limit` query.

I do understand that we need to be backwards compatible, but I would suggest that the limit is _also_ done at the server including limit from the client. The server would then use the minimum limit value.

liggitt: >  I think leaving this to the client may solve only a part of the issue if the client decides to not send the limit query.

A client that does not send a limit query may be incapable of understanding partial responses with continue tokens. We cannot break compatibility with such clients.

lpabon: @liggitt Correct. I am just pointing out that clients that do not send `limit` will still cause issues. We should come up with a way that helps resolve the memory issue at the server.

For example: We could truncate results returned to clients which do not set the `limit` if the request is too large. Or we could return an `HTTP status 413` for large requests forcing the client to update and to use `limit`.

lpabon:
One other possible solution which allows backward compatibility (although truncated) is use [RFC 5988](https://tools.ietf.org/html/rfc5988) which is used by GitHub's linking/paging API: https://developer.github.com/v3/#pagination

bgrant0607:
For reference/comparison: https://cloud.google.com/apis/design/design_patterns#list_pagination

contributors/design-proposals/api-paging.md
smarterclayton: Clients have to opt in.  If users want to fix the memory problem, they have to upgrade.  We can't break the semantics of the existing APIs for existing clients, so everything follows from that.

smarterclayton:
Truncation is not backwards compatible and doesn't meet the design goals of the paging (which is to allow clients to opt in to more efficient retrieval, not to force clients to use more efficient retrieval).

lpabon:
@smarterclayton Understood, but then the issue of memory consumption on the server will still be there. Maybe the paper should point this out as the expected result when clients do not use paging.

smarterclayton:
To avoid confusion I am renaming this as "API chunking" since "API paging" is more broad and can cover use cases that we do not intend to support in the near term.  Chunking and paging would reuse similar mechanisms, but chunking is more specific.  All review comments addressed.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+  "metadata": {"resourceVersion": "147"},
+  "items": [
+     // no more than 500 items
+   ]
+}
+```
+
+The token returned by the server for `continue` would be an opaque serialized string that would contain a simple serialization of a version identifier (to allow future extension), and any additional data needed by the server storage to identify where to start the next range. 
+
+The continue token is not required to encode other filtering parameters present on the initial request, and clients may alter their filter parameters on subsequent chunk reads. However, the server implementation **may** reject such changes with a `400 Bad Request` error, and clients should consider this behavior undefined and left to future clarification. Chunking is intended to return consistent lists.
```
liggitt: if the behavior is undefined and the server may return a bad request, I'd characterize this as "clients should not alter their filter parameters on subsequent chunk reads"

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+If the resource version parameter specified on the request is inconsistent with the `continue` token, the server **must** reject the request with a `400 Bad Request` error.
```
liggitt: is limit/continue supported with resourceVersion=0?

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+The schema of the continue token is chosen by the storage layer and is not guaranteed to remain consistent for clients - clients **must** consider the continue token as opaque. Server implementations **should** ensure that continue tokens can persist across server restarts and across upgrades.
+
+Servers **may** return fewer results than `limit` if server side filtering returns no results such as when a `label` or `field` selector is used. If the entire result set is filtered, the server **may** return zero results with a valid `continue` token. A client **must** use the presence of a `continue` token in the response to determine whether more results are available, regardless of the number of results returned. A server that supports limits **must not** return more results than `limit` if a `continue` token is also returned. If the server does not return a `continue` token, the server **must** return all remaining results.
```
liggitt: > If the entire result set is filtered, the server **may** return zero results with a valid `continue` token.

On the final set, the server may also return zero results with no `continue` token

contributors/design-proposals/api-chunking.md
smarterclayton: Continue + resourceVersion=0 should probably return a bad request.  Continue token should be the only field required to get a consistent paging.  resourceVersion=!0 should probably be accepted with continuation.  resourceVersion=0... probably should as well, since chunking should be orthogonal.  We might have to say that it's unsupported in the alpha release though if it involves mucking with the watch cache.

smarterclayton:
@liggitt addressed your final comments.  Any other comments from anyone?

lpabon:
@smarterclayton LGTM, just one quick question. If the apiserver notices it has OOM or close to it, what is the expected result at the client?

liggitt:
LGTM as a first step we can iterate on. The exposed API surface is minimal, largely opaque, and aligns well with existing chunking mechanisms in other APIs.

liggitt:
> If the apiserver notices it has OOM or close to it, what is the expected result at the client?

Unchanged by this proposal. API server OOM during request handling is considered abnormal, and if it occurs, the connection to the client is terminated.

smarterclayton:
Squashing

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+### Security implications of returning last or next key in the continue token
+
+If the continue token encodes the next key in the range, that key may expose info that is considered security sensitive, whether simply the name or namespace of resources not under the current tenant's control, or more seriously the name of a resource which is also a shared secret (for example, an access token stored as a kubernetes resource). There are a number of approaches to mitigating this impact:
+
+1. Disable chunking on specific resources
+2. Disable chunking when the user does not have permission to view all resources within a range
+3. Encrypt the next key or the continue token using a shared secret across all API servers
+4. When chunking, continue reading until the next visible start key is located after filtering, so that start keys are always keys the user has access to.
```
lavalamp: If you're worried only about construction and not about making it opaque, you could sign or HMAC the key to prevent users from constructing them.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+#### Types of clients and impact
+
+Some clients such as controllers, receiving a 410 error, may instead wish to perform a full LIST without chunking.
+
+* Controllers with full caches
+  * Any controller with a full in-memory cache of one or more resources almost certainly depends on having a consistent view of resources, and so will either need to perform a full list or a paged list, without dropping results
+* `kubectl get`
+  * Most administrators would probably prefer to see a very large set with some inconsistency rather than no results (due to a timeout under load).  They would likely be ok with handling `410 ResourceExpired` as "continue from the last key I processed"
```
lavalamp: Doesn't etcd's ordering potentially change with each update? I would expect lots of duplicates / omissions in this case.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+* Migration style commands
+  * Assuming a migration command has to run on the full data set (to upgrade a resource from json to protobuf, or to check a large set of resources for errors) and is performing some expensive calculation on each, very large sets may not complete over the server expiration window.
+
+For clients that do not care about consistency, the server **may** return a `continue` value on the `ResourceExpired` error that allows the client to restart from the same prefix key, but using the latest resource version.  This would allow clients that do not require a fully consistent LIST to opt in to partially consistent LISTs but still be able to scan the entire working set. It is likely this could be a sub field (opaque data) of the `Status` response under `statusDetails`.
```
lavalamp: Are etcd responses sorted by key? For some reason I thought it was by position in the etcd db file. @jpbetz

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+For clients that must perform sequential processing of lists (kubectl get, migration commands) this change dramatically improves initial latency - clients got their first chunk of data in milliseconds, rather than seconds for the full set. It also improves user experience for web consoles that may be accessed by administrators with access to large parts of the system.
+
+It is recommended that most clients attempt to page by default at a large page size (500 or 1000) and gracefully degrade to not chunking.
+
+
+### Other solutions
+
+Compression from the apiserver and between the apiserver and etcd can reduce total network bandwidth, but cannot reduce the peak CPU and memory used inside the client, apiserver, or etcd processes.
```
lavalamp: Why can't apiserver do stream processing?

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+If the resource version parameter specified on the request is inconsistent with the `continue` token, the server **must** reject the request with a `400 Bad Request` error.
+
+The schema of the continue token is chosen by the storage layer and is not guaranteed to remain consistent for clients - clients **must** consider the continue token as opaque. Server implementations **should** ensure that continue tokens can persist across server restarts and across upgrades.
+
+Servers **may** return fewer results than `limit` if server side filtering returns no results such as when a `label` or `field` selector is used. If the entire result set is filtered, the server **may** return zero results with a valid `continue` token. A client **must** use the presence of a `continue` token in the response to determine whether more results are available, regardless of the number of results returned. A server that supports limits **must not** return more results than `limit` if a `continue` token is also returned. If the server does not return a `continue` token, the server **must** return all remaining results. The server **may** return zero results with no `continue` token on the last call.
```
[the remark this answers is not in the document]
smarterclayton: No it doesn't.  In fact, filtering at the DB level should be opaque and possible - the continue token is generated by storage and surfaced all the way up, and filtering at the DB would simply identify the appropriate next key.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+The server **may** limit the amount of time a continue token is valid for. Clients **should** assume continue tokens last only a few minutes.
```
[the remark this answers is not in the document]
smarterclayton: They do, 410 Resource Expired.   

For that case the default pager impl falls back to not paged.  On our largest pathological data sets (1.2 gb at rest, 20gb in memory) i was able to read 100k secrets (500M of data) in chunks of 500 in under 4 seconds.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+The storage layer in the apiserver must apply consistency checking to the provided continue token to ensure that malicious users cannot trick the server into serving results outside of its range. The storage layer must perform defensive checking on the provided value, check for path traversal attacks, and have stable versioning for the continue token.
+
+#### Possible SQL database implementation
+
+A SQL database backing a Kubernetes server would need to implement a consistent snapshot read of an entire resource type, plus support changefeed style updates in order to implement the WATCH primitive. A likely implementation in SQL would be a table that stores multiple versions of each object, ordered by key and version, and filters out all historical versions of an object. A consistent paged list over such a table might be similar to:
```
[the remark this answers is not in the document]
smarterclayton: I just want to be really clear on the equivalence between an MVCC store like etcd and someone who really believes they want to have an alternate backend.  At our scales it doesn't look like it's necessary anymore, but I want to be able to point back to a clear consideration.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+    SELECT * FROM resource_type WHERE resourceVersion < ? AND deleted = false AND namespace > ? AND name > ? LIMIT ? ORDER BY namespace, name ASC
+
+where `namespace` and `name` are part of the continuation token and an index exists over `(namespace, name, resourceVersion, deleted)` that makes the range query performant. The highest returned resource version row for each `(namespace, name)` tuple would be returned.
+
+
+### Security implications of returning last or next key in the continue token
+
+If the continue token encodes the next key in the range, that key may expose info that is considered security sensitive, whether simply the name or namespace of resources not under the current tenant's control, or more seriously the name of a resource which is also a shared secret (for example, an access token stored as a kubernetes resource). There are a number of approaches to mitigating this impact:
```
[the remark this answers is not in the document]
smarterclayton: If we filter by ACL on a subset (user A can see namespaces 1, 3, and 10) and the last page of 3 extends into 4, if we don't continue reading until the next visible key we would return a startKey of `4/somename` which would expose both the presence of 4 and `sometime` to the user who shouldn't know that 4 exists.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+1. Disable chunking on specific resources
+2. Disable chunking when the user does not have permission to view all resources within a range
+3. Encrypt the next key or the continue token using a shared secret across all API servers
```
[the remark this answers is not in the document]
smarterclayton: Wouldn't work across api servers unless persisted to etcd.

contributors/design-proposals/api-chunking.md
smarterclayton: It does not.  Etcd guarantees key order.

contributors/design-proposals/api-chunking.md
smarterclayton: Etcd is a sorted map in lexicographic key order.

contributors/design-proposals/api-chunking.md
smarterclayton: Depends on how you define a stream.  It can.  Would require a dramatic change to the API shape of LIST.

bgrant0607:
Speaking of dramatic changes to APIs, we need to address the Endpoints API eventually.

contributors/design-proposals/api-chunking.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### etcd3
+
+For etcd3 the continue token would contain a resource version (the snapshot that we are reading that is consistent across the entire LIST) and the start key for the next set of results. Upon receiving a valid continue token the apiserver would instruct etcd3 to retrieve the set of results at a given resource version, beginning at the provided start key, limited by the maximum number of requests provided by the continue token (or optionally, by a different limit specified by the client). If more results remain after reading up to the limit, the  storage should calculate a continue token that would begin at the next possible key, and the continue token set on the returned list.
+
+The storage layer in the apiserver must apply consistency checking to the provided continue token to ensure that malicious users cannot trick the server into serving results outside of its range. The storage layer must perform defensive checking on the provided value, check for path traversal attacks, and have stable versioning for the continue token.
```
serathius, line 94: This cannot happen as etcd doesn't have idea of path. Key is just a blob, it's Kubernetes that puts paths in key.

PR #901, opened 2017-08-11
create owners files for community repo

grodrigues3:
In order to enable the mungebot on the community repo, we need OWNERS files.

As a first pass, I used coreos/findowners tool to create owners files for community.  These are just a start based on commits and likely need to be changed/updated.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,18 @@
+reviewers:
+	- sarahnovotny
+	- idvoretskyi
+	- calebamiles
+	- grodrigues3
+approvers:
+	- sarahnovotny
+	- idvoretskyi
+	- calebamiles
+	- grodrigues3
+	- brendandburns
+	- dchen1107
+	- jbeda
+	- lavalamp
+	- smarterclayton
+	- thockin
+	- wojtek-t
+
--- a/contributors/devel/OWNERS
+++ b/contributors/devel/OWNERS
@@ -1,0 +1,14 @@
+reviewers:
+	- grodrigues3
+	- Phillels
+	- idvoretskyi
+	- calebamiles
+	- cblecker
+	- spiffxp
+approvers:
+	- grodrigues3
+	- Phillels
+	- idvoretskyi
+	- calebamiles
+	- cblecker
+	- spiffxp
--- a/sig-api-machinery/OWNERS
+++ b/sig-api-machinery/OWNERS
@@ -1,0 +1,10 @@
+reviewers:
+	- lavalamp
+	- deads2k
+	- mbohlool
+	- bgrant0607
+approvers:
+	- deads2k
+	- lavalamp
+	- mbohlool
+	- bgrant0607
--- a/sig-auth/OWNERS
+++ b/sig-auth/OWNERS
@@ -1,0 +1,8 @@
+reviewers:
+	- ericchiang
+	- liggitt
+	- deads2k
+approvers:
+	- ericchiang
+	- liggitt
+	- deads2k
--- a/sig-product-management/OWNERS
+++ b/sig-product-management/OWNERS
@@ -1,0 +1,8 @@
+reviewers:
+	- apsinha
+	- idvoretskyi
+	- calebamiles
+approvers:
+	- apsinha
+	- idvoretskyi
+	- calebamiles
--- a/committee-steering/OWNERS
+++ b/committee-steering/OWNERS
@@ -1,0 +1,16 @@
+reviewers:
+	- bgrant0607
+	- philips
+	- thockin
+	- sarahnovotny
+	- smarterclayton
+	- jbeda
+	- brendanburns
+approvers:
+	- bgrant0607
+	- philips
+	- thockin
+	- sarahnovotny
+	- smarterclayton
+	- jbeda
+	- brendanburns
--- a/community/OWNERS
+++ b/community/OWNERS
@@ -1,0 +1,12 @@
+reviewers:
+	- czahedi
+	- nyener
+	- zehicle
+	- sarahnovotny
+	- jberkus
+approvers:
+	- czahedi
+	- nyener
+	- zehicle
+	- sarahnovotny
+	- jberkus
--- a/contributors/design-proposals/clustering/OWNERS
+++ b/contributors/design-proposals/clustering/OWNERS
@@ -1,0 +1,4 @@
+reviewers:
+	- michelleN
+approvers:
+	- michelleN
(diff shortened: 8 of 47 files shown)


grodrigues3:
@sarahnovotny @bgrant0607 @cjwagner

grodrigues3:
Once this is merged, we can merge https://github.com/kubernetes/test-infra/pull/3664

cjwagner:
/lgtm
Since you specified a `reviewers` section in all the OWNERS files I'll enable the `use-reviewers` option to make blunderbuss use the `reviewers` sections to assign OWNERS instead of the `approvers` sections.

grodrigues3:
I'll wait for one more set of eyes before merging

@calebamiles @sarahnovotny @idvoretskyi @castrojo

cblecker:
This works as a start. I can see a few things that would be good to update (things like putting sig leads as the reviewers/approvers for the sig folders), but that can be iterated on after the files are created.

contributors/design-proposals/OWNERS
```diff
@@ -0,0 +1,8 @@
+reviewers:
```
smarterclayton: Should be sig-architecture

contributors/devel/OWNERS
```diff
@@ -0,0 +1,8 @@
+reviewers:
```
smarterclayton: Should probably be sig contribx except for a few things like api which are sig architecture

sig-api-machinery/OWNERS
```diff
@@ -0,0 +1,16 @@
+reviewers:
+	- czahedi
+	- mbohlool
+	- cblecker
+	- jamiehannaford
+	- deads2k
+	- bgrant0607
+	- pmorie
+approvers:
```
smarterclayton: Hrm, should be the leads

sig-auth/OWNERS
```diff
@@ -0,0 +1,12 @@
+reviewers:
+	- jamiehannaford
+	- idvoretskyi
+	- pmorie
+	- zigmax
+	- cblecker
+approvers:
+	- jamiehannaford
```
smarterclayton: Not the right group for sig-auth

smarterclayton:
The sig leads look wrong.  Seems like that should be populated from the sigs yaml in the future

liggitt:
most of the author info was lost in the move from the kubernetes/kubernetes repo to here. I actually don't think the output of that tool is a good starting point for these folders

zhangxiaoyu-zidif:
It's a wired name list.

sig-product-management/OWNERS
```diff
@@ -0,0 +1,6 @@
+reviewers:
```
idvoretskyi: Should be SIG-PM leads

grodrigues3:
Alright, here's going to be my approach.  I'll get the top level ones right (based on consensus).  I'll also make sure sig leads are in their respective owners files.  

I'll delete the others, and create an issue pinging people/sigs to create/update them.

Sound reasonable?

cblecker:
For the sig folders, yes. sig leads (and their delegates) should absolutely have control over those. Trickier are things like `community/` and `contributors/` if the top level pool is small.

idvoretskyi:
> I'll delete the others, and create an issue pinging people/sigs to create/update them.

@grodrigues3 SGTM

grodrigues3:
@smarterclayton @cblecker @idvoretskyi 

The sig leads folders now match the sigs.yaml file. 

As cblecker mentioned, the community, contributor and top level OWNERS files are trickier.  Since sig-architecture (for contributor/design-proposals) is just jdumars and bgrant, I imagine they'd get overloaded quickly.  Any suggestions for creating that OWNERS file?  Should we do that in a separate PR where we define an aliases file and a `sig-architecture` alias?

For contributor/devel, I added contribX leads and others involved in contribX work.

Suggestions?

contributors/devel/OWNERS
grodrigues3: added ContribX people.  we don't have an Aliases file yet.  I'll work on that in a separate PR.

OWNERS
```diff
@@ -0,0 +1,18 @@
...
+	- calebamiles
+	- grodrigues3
+	- brendandburns
+	- dchen1107
+	- jbeda
+	- lavalamp
+	- smarterclayton
+	- thockin
+	- wojtek-t
+
```
cblecker: nit: blank line

contributors/devel/OWNERS
cblecker: Should this file perhaps be at the `contributors/` level, and all the OWNER files in subfolders (like the ones in `contributors/design-proposals/`) be removed?

OWNERS
grodrigues3: I'm the worst at that.  Fixing

cblecker:
The only other concern I have with this: Updating the sig files requires updating sigs.yaml and sig-list.md. These are at the root of the repo, so a top-level approver would be required for any changes to those files. Not sure how best to address that though

contributors/devel/OWNERS
grodrigues3: Unfortunately no.  There's a distinction between the ContribX work and design proposals work.  Ideally, the design proposals will get organized into sig subdirs and owners can be tied to sig aliases (that's for a different PR).

cblecker: My concern right now is the stuff in design-proposals doesn't have very many approvers in it because the history was lost.

grodrigues3:
@cblecker that's a good point.  I don't really have an answer either.  I'm really just trying to get the bot on.  For a repo that I thought was quiet, we somehow have acquired 100+ open PRs.

contributors/devel/OWNERS
grodrigues3: maybe we focus on reviewers and turn on the bot without the approvers mechanism.  This way PRs are auto-assigned via blunderbuss, but we don't have a bottleneck.  Anyone can still lgtm like before.

grodrigues3:
Summary 
All sigs should have the sig leads.
Steering committee has the seven steering committee members
Contributors/devel has ContribX folks
Contributors/design-proposals **still needs work**.  I'm inclined to merge without this and have it updated it elsewhere.

The real  goal is to get the PRs auto-assigned.  I added the approvers section, but it may make more sense to turn on the mungebot without approvers at all, requiring only lgtm to merge and using OWNERS files for their blunderbuss integration (i.e. PRs are auto-assigned to reviewers to look over).

cjwagner:
If more people need to be able to edit `sigs.yaml` and `sig-list.md` couldn't they be moved to a sub folder and given their own OWNERS file specifying a broader set of approvers?

PR #905, opened 2017-08-13
Update golang version

zhangxiaoyu-zidif:
Update golang version


zhangxiaoyu-zidif:
I can not build K8s, if my go version is below 1.8.3. So I want to send this PR

PR #908, opened 2017-08-14
Fixed link to auth_test.go in coding-conventions.md

ianchakeres:


contributors/devel/coding-conventions.md
```diff
@@ -75,7 +75,7 @@ following Go conventions - `stateLock`, `mapLock` etc.
 tests
 
   - Table-driven tests are preferred for testing multiple scenarios/inputs; for
-example, see [TestNamespaceAuthorization](../../test/integration/auth/auth_test.go)
+example, see [TestNamespaceAuthorization](https://github.com/kubernetes/kubernetes/blob/master/test/integration/auth/auth_test.go)
```
cblecker: Can we use "https://git.k8s.io/kubernetes/test/integration/auth/auth_test.go" for this?

ianchakeres: @cblecker - I'm fine with using the path that you mention, it resolves to the URL I specified. Why do you prefer this path?

Note: the path I've used is referenced in a few other spots in the community repo.

cblecker: In the (unlikely) event that the target for that document changes, git.k8s.io is something we control [here](https://github.com/kubernetes/k8s.io/blob/fb20b6aa18d8d6fab882df36ee534bd8208d9900/k8s.io/configmap-nginx.yaml#L205-L215). Not critical to go at the moment and update everything, but if we're fixing broken links, we might as well use it :)

ianchakeres: @cblecker - ok - I updated the three references to auth_test.go.

PR #911, opened 2017-08-14
design: reduce scope of node on node object

mikedanese:
Super mini design doc about centralizing reporting of some sensitive kubelet attributes.

@kubernetes/sig-auth-misc 
@roberthbailey 
@kubernetes/sig-cluster-lifecycle-misc as it relates to registration


contributors/design-proposals/limit-node-object-self-control.md
```diff
@@ -0,0 +1,90 @@
...
+these attributes from the kubelet.
+
+We can scope down the initial Node object creation by moving to a centralized
+controller model. In many deployment environments, the sensitive attributes of a
+Node object discussed above ("labels", "taints", "addresses") are discoverable
+by consulting a machine database (e.g. the GCE API). Using the
+[initializer](admission_control_extension.md) mechanism, a centralized
+controller can register an initializer for the node object and build the
+sensitive fields by consulting the machine database. The
+`cloud-controller-manager` is an obvious candidate to house such a controller.
```
smarterclayton: Since those fields change over time I'm not even sure that initializers are required for anything except strong exclusion rules.

smarterclayton:
Has come up a lot in security auditing.

contributors/design-proposals/limit-node-object-self-control.md
mikedanese: I think you are right. I think we want strong exclusion of sensitive labels, taints and addresses though. I will keep the initializer and add a stanza that allows the central controller reconcile objects as well.

contributors/design-proposals/limit-node-object-self-control.md
```diff
@@ -0,0 +1,90 @@
...
+### Capturing Dedicated Workloads
+
+Suppose company `foo` needs to run an application that deals with PII on
+dedicated nodes to comply with government regulation. A common mechanism for
+implementing dedicated nodes in Kubernetes today is to set a label or taint
+(e.g. `foo/dedicated=customer-info-app`) on the node and to select these
+dedicated nodes in the workload controller running `customer-info-app`.
+
+Since the nodes self reports labels upon registration, an intruder can easily
+register a compromised node with label `foo/dedicated=customer-info-app`. The
```
ConnorDoyle: The label part seems minor in comparison to the problem of compromised nodes registering themselves at all.

ConnorDoyle: Which is another way to say, if we can't trust a node to set appropriate labels then why are we trusting it at all?

liggitt: labels allow steering workloads. we need to be able to set up labeled topologies of nodes to keep classes of workloads separate, and be able to isolate a compromised node by tainting it, unlabeling it, and know that it didn't steer workloads to itself by adding to its own labels.

contributors/design-proposals/limit-node-object-self-control.md
```diff
@@ -0,0 +1,90 @@
...
+scheduler will then bind `customer-info-app` to the compromised node potentially
+giving the intruder easy access to the PII.
```
liggitt: the same applies to allowing a node to remove taints (or delete its own Node API object while tainted)

liggitt:
cc @kubernetes/sig-node-proposals

mikedanese:
This was discussed in sig-auth and the proposal needs some further consideration:
* We need to dig into the label/taint whitelisting mechanism. We need to make it easy for kubelet to set some of these on itself. We need to make it clear that kubelet self-set labels cannot be used to implement strong exclusion.
* We need to clarify what the expectations are for a deployment running without a cloud provider.
* We need to decide whether a node should be able to delete itself, thereby potentially removing taints from itself.
* We may need to consider how we might need to adjust kubelet starts up.

contributors/design-proposals/limit-node-object-self-control.md
liggitt: "sensitive labels" is going to be *really* tough to pin down. Examples of node self-labeling I've seen just in the past two days are cpu policy and kernel version, both of which *could* be used for capability steering (a workload requiring a specific CPU policy or kernel version to run) or for security purposes ("find nodes running a kernel with a known vulnerability", "schedule pods to a known good kernel")

liggitt:
this also came up in the GCE cloud provider trying to determine what zones exist by looking at the node API objects and trusting whatever zone they reported they were in (https://github.com/kubernetes/kubernetes/pull/52322#issuecomment-331519659)

roberthbailey:
@mikedanese - are you aiming to get this merged and something implemented for 1.9? Or is this a longer term proposal?

liggitt:
> are you aiming to get this merged and something implemented for 1.9? Or is this a longer term proposal?

I'd like to see the labeling/tainting approach agreed on and implemented for 1.9

The node addresses are more involved and have more ties to cloud providers and variance between cloud/non-cloud environments.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/911)

luxas:
ping @mikedanese @kubernetes/sig-auth-feature-requests any progress for v1.9 for this feature?

mikedanese:
This is a longer term proposal. We need someone to work on this and it isn't highest priority yet.

roberthbailey:
Do we expect to finish and merge the proposal soon even if it won't be implemented right away?

mikedanese:
I'm incorporating feedback this week. I would like to get the design merged by 1.9.

roberthbailey:
@mikedanese - is this ready for review again?

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,90 @@
...
+We can scope down subsequent updates of the Node object by moving control of
+sensitive fields into the NodeRestriction admission controller. For backwards
+compatibility we can begin by zero'ing updates to sensitive fields and after the
+kubelet compatibility window has expired, begin to return 403s. This also has
+the nice property of giving us fine grained control over field/values of
+updates. For example, in this model we could easily allow a kubelet to update
+it's `OutOfMemory` taint and disallow the kubelet from updating its `dedicated`
+taint.
+
+In this design we assume that the infrastructure has already validated caller ID
```
tamalsaha: Hi, how is it guaranteed that caller id is trusted. 
`nodeclient - a request by a user for a client certificate with O=system:nodes and CN=system:node:(node name)`
The `nodeclient` provides the CSR request and that request is coming from kubelet also. Who is creating this client cert? kubeadm?

mikedanese: It's not guaranteed. I'm saying for the purposes of this design, we should assume that the node authentication credentials are trusted. Hardening that is a separate topic and out of scope of this proposal.

ericchiang:
Just to add some thoughts for the 1.10 timeline.

Is central reporting a blocker for this work? There are still aspects that only the node will know. For example, on Container Linux, which version of CL the node auto-updated too.

The problem seems to be that:
* Admins partition nodes on labels that are self-reported.
* Nodes have the ability to un-taint themselves. So tainting an owned node doesn't have the desired effect.
* Addresses are self-reported.

Address being self-reported also seems like it could be solved by proper node attestation (https://github.com/kubernetes/kubernetes/issues/57164). 

Can we scope this work to solve the label and tainting problem? e.g. nodes can't untaint themselves, and we only let them set certain white-listed labels?

smarterclayton:
I think we should scope this down to labelling and tainting.

Today, nodes are effectively "best effort self-managed".  That leads to us giving nodes higher rights than necessary.  We have a pretty strong requirement to expose autoscaling with minimal security credential injection, which means we want to delegate enough permission to join and reasonably configure node placement without leaving the door open for impersonating a master.

Not being able to un-taint, and white listing labels nodes can set, would be a pretty powerful combo.  I don't know that we need the other bit right now.

mikedanese:
I would be okay with that down scoping.

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Proposed Solution
+
+In many environments, we can improve the situation by centralizing reporting of
+sensitive node attributes to a more trusted source and disallowing reporting of
+these attributes from the kubelet.
+
+### Label And Taint Restriction
+
+An operator will configure a whitelist of taints and labels that nodes are
```
mikedanese: Should the list support key prefixes?

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+  --register-with-taints
+  --node-labels
+```
+
+This whitelist is passed as a command line flag to the apiserver. Admission
+control (either NodeRestriction or in a new admission control) will then prevent
+setting and modification by nodes of all taints and labels with keys not in the
+whitelist.
```
tallclair: This means that if an administrator wants to add a new node pool with a new taint or label, they need to restart the APIserver. Is there an argument against dynamic config?

EDIT: I guess the work around is to just manually (or via an external component) label the nodes after they're registered?

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,127 @@
...
+when it's external name (inferred through the cloud provider) changes. This code
+path will never be executated on the majority of cloud providers and this
+capability undermines the usage of taints as a strong exclusion primitive.
+
+For example, suppose an operator sets a taint `compromised` on a node that they
+believe has been compromised. Currently, the compromised node could delete and
+recreate itself thereby removing the `compromised` taint.
+
+To prevent this, we will add a configuration option to the apiserver that
+indicates whether the NodeAuthorizer should allow self deletion.
```
tallclair: :+1:

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,127 @@
...
+An operator will configure a whitelist of taints and labels that nodes are
+allowed to set on themselves. This list should include the taints and labels
+that the kubelet is already setting on itself.
+
+Well known taint keys:
+```
+node.kubernetes.io/not-ready
+node.alpha.kubernetes.io/notReady
+node.kubernetes.io/unschedulable
+node.kubernetes.io/out-of-disk
```
k82cn: Those taints are set by NodeController instead of node. Do you mean we need to change that?

liggitt: no, it is strongly preferred for the NodeController to set/clear these taints. what taints do kubelets set on themselves?

contributors/design-proposals/node/limit-node-object-self-control.md
liggitt: > I guess the work around is to just manually (or via an external component) label the nodes after they're registered?

I'd actually consider that the main recommendation, not a workaround

k82cn: maybe a configure file or confmap is better for dynamic update :)

contributors/design-proposals/node/limit-node-object-self-control.md
liggitt: ExternalID has been marked as deprecated since 1.1

https://github.com/kubernetes/kubernetes/blob/release-1.1/pkg/api/v1/types.go#L1745-L1747

can we evaluate dropping it and the deletion cases associated with it altogether instead?

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,127 @@
...
+  --register-with-taints
+  --node-labels
+```
+
+This whitelist is passed as a command line flag to the apiserver. Admission
+control (either NodeRestriction or in a new admission control) will then prevent
+setting and modification by nodes of all taints and labels with keys not in the
+whitelist.
+
+Proposed flags to add to the apiserver to support this:
```
liggitt: I'd expect these enforcements in the NodeRestriction admission controller, and the config for those in the config file for that admission controller. That controller is what enforces boundaries between nodes today, and knows whether the incoming request is from a node or not.

contributors/design-proposals/node/limit-node-object-self-control.md
k82cn: Only `node.cloudprovider.kubernetes.io/uninitialized` is set by `kubelet` and removed by cloud-controller-manager, the others are set by `NodeController`.

**TaintBasedEviction (alpha)**:
node.kubernetes.io/not-ready
node.alpha.kubernetes.io/notReady

node.kubernetes.io/unreachable <===== seems missed ?
node.alpha.kubernetes.io/unreachable 

**TaintNodeByCondtion (alpha)**:
node.kubernetes.io/unschedulable
node.kubernetes.io/out-of-disk
node.kubernetes.io/memory-pressure
node.kubernetes.io/disk-pressure
node.kubernetes.io/network-unavailable
node.kubernetes.io/pid-pressure

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,127 @@
...
+## Proposed Solution
+
+In many environments, we can improve the situation by centralizing reporting of
+sensitive node attributes to a more trusted source and disallowing reporting of
+these attributes from the kubelet.
+
+### Label And Taint Restriction
+
+An operator will configure a whitelist of taints and labels that nodes are
+allowed to set on themselves. This list should include the taints and labels
```
k82cn: It should work together with Node Authorization and RBAC.

liggitt: labels/taints are enforced in admission, not authorization (the authorization layer does not have access to the content of the posted node object). I'd expect them to be enforced in the NodeRestriction admission plugin, along with other content-related restrictions on nodes.

k82cn: only authenticated nodes can request update; admission enforces labels/taints :)

contributors/design-proposals/node/limit-node-object-self-control.md
mikedanese: Thanks, I'll update the list.

contributors/design-proposals/node/limit-node-object-self-control.md
mikedanese: Removing it would be preferable.

contributors/design-proposals/node/limit-node-object-self-control.md
mikedanese: It's state below that this will be enforced in NodeRestriction

contributors/design-proposals/node/limit-node-object-self-control.md
mikedanese: It states below that this will be enforced in NodeRestriction. I'll make adjustments to configure thies in the admission controller config file.

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,113 @@
...
+the same way and can now use the node certificate to read the company's serving
+certificate key.
+
+## Proposed Solution
+
+In many environments, we can improve the situation by centralizing reporting of
+sensitive node attributes to a more trusted source and disallowing reporting of
+these attributes from the kubelet.
+
+### Label And Taint Restriction
```
mtaufen: Add a note to think through how to represent self-labeling/tainting in kubelet.config.k8s.io/v1beta1 API.

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,113 @@
...
+beta.kubernetes.io/instance-type
+beta.kubernetes.io/os
+beta.kubernetes.io/arch
+```
+
+As well as any taints and labels that the operator is setting using:
+
+```
+  --register-with-taints
+  --node-labels
```
mikedanese: @mtaufen wants to know whether these flags should be included in kubelet config.

contributors/design-proposals/node/limit-node-object-self-control.md
mtaufen: https://github.com/kubernetes/kubernetes/issues/61659 and https://github.com/kubernetes/kubernetes/issues/61655

mikedanese:
@liggitt added config structs.

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,128 @@
...
+controller with the name `noderestriction.admission.k8s.io`. It will contain one
+config object:
+
+```golang
+type Configuration struct {
+  // RestrictLabels specifies whether the admission controller should restrict
+  // node self set labels to the whitelist.
+  RestrictLabels bool
+  // LabelWhitelist is a list of labels a node is allowed to set on itself.
+  LabelWhitelist []string
```
mtaufen: consider AllowedLabels/Taints?

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,128 @@
...
+
+A new configuration API group will be created for the NodeRestriction admission
+controller with the name `noderestriction.admission.k8s.io`. It will contain one
+config object:
+
+```golang
+type Configuration struct {
+  // RestrictLabels specifies whether the admission controller should restrict
+  // node self set labels to the whitelist.
+  RestrictLabels bool
```
mtaufen: What's the tradeoff between a bool and allowing `*` in the whitelist?

saad-ali:
On the storage side of things, we are planning on allowing CSI drivers to be able to specify node labels for storage availability topology labeling. As mentioned in https://github.com/kubernetes/kubernetes/pull/61866#issuecomment-379649244 this labeling will be controlled by the kubelet, which would prefix each label key (@verult can provide details about this), so we would be able to take advantage of the node label restriction proposed here.

verult:
One possible label format for CSI topology is something like this:
`"csi.kubernetes.io.<driver-name>/<topology-name>":"<topology-value>"`

Example:
`"csi.kubernetes.io.csi-driver.example.com/rack": "rack1"`

Is it possible to whitelist a prefix (in this case "csi.kubernetes.io") instead of the entire label name?

contributors/design-proposals/node/limit-node-object-self-control.md
ericchiang: Also, any way for us to make this not default to "allow all"?

ericchiang:
> Is it possible to whitelist a prefix (in this case "csi.kubernetes.io") instead of the entire label name?

Not as it stands but @mikedanese seems open to it https://github.com/kubernetes/community/pull/911/files/cc0d9c65e7e45b2e85aa5ca5e7bec69040e6d7c1#r176191881

contributors/design-proposals/node/limit-node-object-self-control.md
mikedanese: @ericchiang how do we start restricting without breaking compatibility with the API?

contributors/design-proposals/node/limit-node-object-self-control.md
mikedanese: Done.

mikedanese:
I updated the doc to include support for key prefixes, incorporate some other comments on the structure, and add example configs.

contributors/design-proposals/node/limit-node-object-self-control.md
ericchiang: @mikedanese by specifying the config you opt into the behavior of the `NodeRestriction` component? Tune it to get the old, less secure behavior back?

contributors/design-proposals/node/limit-node-object-self-control.md
mikedanese: I'd like to just delete those knobs, but I suspect we will have to move them into kubeletconfig.

contributors/design-proposals/node/limit-node-object-self-control.md
ericchiang: I guess I'm confused how this config is being passed to kube. Is it not a flag on the API server or something?

mikedanese: Ya, I'm fine with that approach. The config is passed in via the --admission-control-config-file flag on the apiserver:

https://kubernetes.io/docs/admin/admission-controllers/#configuration-file-format

contributors/design-proposals/node/limit-node-object-self-control.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Label And Taint Restriction
+
+An operator will configure a whitelist of taints and labels that nodes are
+allowed to set on themselves. This list should include the taints and labels
+that the kubelet is already setting on itself.
+
+Well known taint keys:
+```
+node.cloudprovider.kubernetes.io/uninitialized
```
liggitt: @kubernetes/sig-scheduling-proposals do kubelets ever modify/remove taints from their node objects, or add taints beyond the initial registration step (based on `--register-with-taints`)? it sounds like they don't.

if not, it seems like we could simply allow nodes to register with whatever taints they want (they're allowed to register with no taints today, and opting into more restrictions should be ok), and simply disallow changes during update. That wouldn't disrupting any existing behavior, and shouldn't require any admission configuration.

tallclair: Can we take the same approach with labels?

liggitt: > Can we take the same approach with labels?

We could possibly do something similar, though we can't just let anything through on create. Adding arbitrary labels on registration lets nodes steer workloads to themselves.

One idea I had was to allow nodes to manage whitelisted labels (the well-known label keys mentioned below), and move any other labels a kubelet attempts to set/clear on itself into an annotation. That does several things:
* preserves the label information the kubelet was trying to set
* makes that information observable via a central controller with higher permissions that could promote those to be real labels on the Node object if desired
* avoids giving kubelets direct control over steering workloads to themselves
* avoids disrupting normal registration/heartbeating activities of kubelets
* avoids needing to configure the NodeRestriction admission plugin

opened https://github.com/kubernetes/kubernetes/pull/64055 as a POC

roberthbailey:
@mikedanese - it's been a few months since this PR was updated; what is the current status?

mikedanese:
@roberthbailey we've been making progress slowly. We locked down taints in 1.11. @liggitt is working on something similar for labels in 1.13.

liggitt:
> From sig-storage perpective the proposal for "suffixing the prefix" such that a CSI label like acme.com/zone: foo will be stored in Kubernetes as acme.com.topology.kubernetes.io/zone: foo is not ideal, because it reduces readability.
>
> Ideally we would prefer a solution where the node authorizer (or wherever this logic would live) would allow dynamically changing the whitelist of allowed label keys (so a new CSI driver, as part of registration, could specify what keys it will require), and retaining the original key as presented by the CSI driver without alteration.

I added a section discussing alternatives considered, like a dynamic API label whitelist mechanism. The primary motivator wasn't complexity, but consistency across clusters and removing escalation paths. That said, I agree it would be better for usability if the CSI topology matched the Node labels. One way to accomplish this within the proposed approach would be to have the CSI topology keys use the allowed topology label prefix.

ddebroy:
I guess this is a basic question: Does labels have to be of the form `[*.]/[*.]` ? Is it possible to have labels of the form `[*.]/[*.]/[*./]` that may potentially better accommodate something along the lines of: `[vendor.com]/[topology.kubernetes.io]/[label]` and keep it readable?

liggitt:
> I guess this is a basic question: Do labels have to be of the form `[*.]/[*.]` ? Is it possible to have labels of the form `[*.]/[*.]/[*./]` that may potentially better accommodate something along the lines of: `vendor.com/topology.kubernetes.io/label` and might be more readable than `vendor.com.topology.kubernetes.io/label`?

unfortunately, kubernetes labels are dual-segment only. I agree the multi-level namespacing would read better.

saad-ali:
We could possibly shrink Kubernetes to k8s, e.g. `acme.com.topology.k8s.io/zone`?

mikedanese:
> unfortunately, kubernetes labels are dual-segment only. I agree the multi-level namespacing would read better.

That's a relaxable constraint....

> We could possibly shrink Kubernetes to k8s, e.g. `acme.com.topology.k8s.io/zone`?

+1

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+* The kubelet transforms the topology info to a `{"topology.kubernetes.io/zone":"A"}` label and adds it to the Node object
+* The storage class lists allowed topology selectors of `{"key":"topology.kubernetes.io/zone","values":["A"]}`
+* The CSI provisioner transforms the topology key of `"zone"` to a label name of `"topology.kubernetes.io/zone"` in [getTopologyFromNode](https://github.com/kubernetes-csi/external-provisioner/blob/v0.5.0-alpha.0/pkg/controller/topology.go#L295-L305)
+
+## Implementation Timeline
+
+v1.13:
+
+* Kubelet deprecates setting labels other than the ones described above via `--node-labels`, and warns when invoked with labels outside that set
+* NodeRestriction admission prevents kubelets from setting labels other than the ones described above on Node *update* (not on Node create)
```
enj: And this validation only runs if the labels were changed?

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+
+v1.15:
+
+* Kubelet removes the ability to set labels other than the ones described above via `--node-labels` (deprecation period of 6 months for CLI elements of admin-facing components is complete)
+
+v1.17:
+
+* NodeRestriction admission prevents kubelets from setting labels other than the ones describe above on Node create (oldest supported kubelet running against a v1.17 apiserver is v1.15)
```
enj: Same question as above.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
liggitt: correct.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
liggitt: all labels are "changed" on create from non-existent to present, so all labels on create would be subject to validation

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+facing nginx on dedicated nodes to reduce exposure to the company's publicly
+trusted server certificates. They use the secret mechanism to distribute the
+serving certificate key. An intruder captures the dedicated nginx workload in
+the same way and can now use the node certificate to read the company's serving
+certificate key.
+
+## Proposal
+
+The NodeRestriction admission plugin will be updated to limit the labels a kubelet
+can set on its Node object. Kubelets will only be allowed to set the following labels.
```
szuecs: I think we use the node taint to make sure application pods can not run before all base node containers are running to fix for example kube2iam not running before an application that needs aws api access granted by the role fetched by kube2iam.
I think somehow we need a way how to mark a node to be not ready for application containers to run to help our teams to have no weird problems occurring.

@mikkeloscar can you check if our use case of bootstrapping nodes are violated by this change?

mikkeloscar: We don't currently have this in place as it didn't work with the autoscaler.

However we use the label feature a lot to identify different node pools in our clusters and this we do by passing the `--node-labels` flag. While I can understand the security concern in certain environments I don't see the reason for removing a useful feature rather than just make it possible to restrict it if needed. Simply make it configurable on the apiserver what labels are allowed and let the cluster admins decide whether they want this protection or not. I'm fine with making it a default as long as we can disable it.

Assuming this would be in place and `--node-labels` would be dropped, how is it intended that we label nodes in an easy way?

Also the argument about making management difficult is a bit far fetched IMO. You don't configure `N` kubelet command lines, you have the configuration somewhere and reuse it for all nodes that are supposed to have the same configuration. [This is how we do it](https://github.com/zalando-incubator/kubernetes-on-aws/blob/bc59ef967e9d78358ac6bfb80c75525cf51b64a3/cluster/node-pools/worker-default/userdata.clc.yaml#L162). We manage 100's of nodes like this, it's not an issue :)

liggitt: > @szuecs:
> I think we use the node taint to make sure application pods can not run before all base node containers are running to fix for example kube2iam not running before an application that needs aws api access granted by the role fetched by kube2iam. I think somehow we need a way how to mark a node to be not ready for application containers to run to help our teams to have no weird problems occurring.

This is not removing the ability of nodes to add taints on creation.

> @mikkeloscar:
> While I can understand the security concern in certain environments I don't see the reason for removing a useful feature rather than just make it possible to restrict it if needed. Simply make it configurable on the apiserver what labels are allowed and let the cluster admins decide whether they want this protection or not. I'm fine with making it a default as long as we can disable it.

On the apiserver side, all of the proposed restrictions are within the [NodeRestriction](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction) admission plugin. The job of this plugin is to restrict nodes to the workloads that are scheduled to run on them. `kube-apiserver` does *not enable it by default*. If you chose to run without that admission plugin enabled, your nodes could continue to label themselves arbitrarily.

That said, it is important for the kubelet to work well with recommended apiserver configurations, which is why this is proposing limiting the labels `--node-labels` accepts to ones the NodeRestriction plugin would allow.

> Assuming this would be in place and --node-labels would be dropped, how is it intended that we label nodes in an easy way?

This is not dropping `--node-labels`, it is limiting the set of labels to ones that can be distinguished as kubelet-set labels, like `kubelet.kubernetes.io/foo`. Over the next two releases, kubelet invocations would need to move to labels with those prefixes.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+
+This list includes all of the labels that kubelets already set on their Node objects.
+
+```
+kubernetes.io/hostname
+failure-domain.beta.kubernetes.io/zone
+failure-domain.beta.kubernetes.io/region
+beta.kubernetes.io/instance-type
+beta.kubernetes.io/os
+beta.kubernetes.io/arch
```
krmayankk: so if i understand this correctly , the fix is to NOT allow arbitrary labels from kubelet but only well defines ones ? Currently we specify some pool labels for nodes using kubelet, where will that go now ?

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
krmayankk: So we are saying that the comprised node, can only try to update the labels for its own node, not add new labels ? What if the comprised node can register as a new node ?

liggitt: > So we are saying that the comprised node, can only try to update the labels for its own node, not add new labels ?

Preventing labels outside a known set can be done on update immediately because kubelets do not update arbitrary labels on their node objects today.

> What if the comprised node can register as a new node?

Letting nodes register as arbitrary new/different nodes is not recommended.

krmayankk:
Is this the main reason for gke, its recommended to not run different untrusted workloads on the same gke cluster because a comproprised node can take over the entire cluster @liggitt  @tallclair  ?

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+creation-date: 2017-08-14
+last-updated: 2018-10-25
+status: provisional
+---
+
+# Bounding Self-Labeling Kubelets
+
+## Motivation
+
+Today the node client has total authority over its own Node labels.
```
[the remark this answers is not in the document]
liggitt: Yes

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
liggitt: > Currently we specify some pool labels for nodes using kubelet, where will that go now?

You would choose labels with a `kubelet.kubernetes.io/` or `topology.kubernetes.io/` prefix to represent your node pool labels you wanted the kubelet to set. See the second and third sections for examples.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
krmayankk: i dont fully understand this . If a node gets compromised today, it can modify the topology label of that node and so attract arbitrary workloads which were meant for other nodes. How does this feature prevent that. The node is still allowed to update the topology label(one of the allowed ones). What we want is the node only allows a fix set of labels and they can never be updated for a given node. 
 
Also my other doubt (mainly because i dont understand node authorizer properly ) is can a  compromised node ever register as a new node ?

liggitt: > If a node gets compromised today, it can modify the topology label of that node and so attract arbitrary workloads which were meant for other nodes. How does this feature prevent that.

The preferred way of labeling node objects is from a central authoritative source. Bounding the labels kubelets can self-apply means that centrally-set labels outside those prefixes can be trusted to not be settable by compromised nodes.

For example, labels that isolate nodes by tenant, by security policy, by trust level, by workload sensitivity, etc, are all things that are orthogonal to topology, and should be set top-down and be unchangeable by kubelets.

> can a compromised node ever register as a new node 

The node authorizer and noderestriction admission plugin limit kubelets to registering a node object that matches their identity.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+
+CSI topology labels could be transformed to kubernetes labels within the permitted topology label namespace:
+
+* CSI topology labels that did not have a prefix would have a prefix of `topology.kubernetes.io/` prepended
+* CSI topology labels with a prefix not ending in `topology.kubernetes.io` would have that string added to their prefix
+
+Component behavior:
+
+* The kubelet would transform CSI topology to kubernetes labels before adding them to the `Node` API object.
+* The CSI provisioner would transform CSI topology labels to kubernetes labels when [retrieving topology info from a given node](https://github.com/kubernetes-csi/external-provisioner/blob/v0.5.0-alpha.0/pkg/controller/topology.go#L295-L305).
```
verult: The other way around? i.e. from Kubernetes labels to CSI topology labels in order to generate the right CSI accessibility requirements

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+* The CSI provisioner would transform kubernetes labels into CSI topology labels when [building required topology from a storage class AllowedTopologies](https://github.com/kubernetes-csi/external-provisioner/blob/v0.5.0-alpha.0/pkg/controller/topology.go#L75-L91).
+* The associated storage class would continue to specify topology in terms of kubernetes label selectors.
+
```
verult: nit: for the sake of completeness, the provisioner also translates CSI topology from the provisioned volume to volume node affinity, which uses k8s labels.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+Example 1:
+
+* CSI driver returns topology labels `{"acme.com/zone":"A"}`
+* The kubelet records `["acme.com/zone"]` in CSINodeInfo.CSIDriverInfo.TopologyKeys
```
verult: Related comment: https://github.com/kubernetes-csi/external-provisioner/pull/160#issuecomment-433576451

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,163 @@
...
+* The kubelet transforms the topology info to a `{"acme.com.topology.kubernetes.io/zone":"A"}` label and adds it to the Node object
+* The storage class lists allowed topology selectors of `{"key":"acme.com.topology.kubernetes.io/zone","values":["A"]}`
+* The CSI provisioner transforms the topology key of `"acme.com/zone"` to a label name of `"acme.com.topology.kubernetes.io/zone"` in [getTopologyFromNode](https://github.com/kubernetes-csi/external-provisioner/blob/v0.5.0-alpha.0/pkg/controller/topology.go#L295-L305)
```
verult: Same as above - the other way around

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
liggitt: I think this is correct… the input is topology keys, and it needs to look up the values from the node labels, so it translates from CSI topology keys to label keys

verult: right, but then the output should be converted from k8s label format to CSI topology format so it could be processed as required/preferred topology

liggitt: The translated key is only used to look up the label value from the node object: https://github.com/kubernetes-csi/external-provisioner/pull/160/commits/0f111a2c50aa4a58a139f9d5314fecad3f7af9bf#diff-94bad98230836caa3e44ca0c9b4281c3L294

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
szuecs: Thanks for your clarification

mikkeloscar: @liggitt
> Over the next two releases, kubelet invocations using --node-labels would need to move to labels with those prefixes.

Would this still only be the case if `NodeRestriction` plugin is enabled, or would it be enforced on the `kubelet` side? If it's enforced by the kubelet it will be a bit annoying.

liggitt:
I know Tim has been involved in a lot of the various topology discussions. Would like his thoughts on this as well

/assign @thockin

thockin:
Trying to catch up on this after Saad asked my thoughts.  I have just skimmed the comments, will read the content later.

First reaction: recoil in horror.

We've thus far made a point to accept that topology is arbitrary and that any label COULD represent topology (e.g. kubernetes.io/hostname counts).  Manipulating automatic labels is gross and awkward, IMO.

We've been assuming that topology is carried in labels - what if it wasn't?  What if we made it more of a first-class field that was designated JUSt for topology information?  @johnbelamaric

liggitt:
> First reaction: recoil in horror.
> 
> We've thus far made a point to accept that topology is arbitrary and that any label COULD represent topology (e.g. kubernetes.io/hostname counts). Manipulating automatic labels is gross and awkward, IMO.

No disagreement.

> We've been assuming that topology is carried in labels - what if it wasn't? What if we made it more of a first-class field that was designated JUSt for topology information?

Unifying selection on labels (across storageclass, persistentvolume, scheduler, podspec, etc) is really simple and convenient. I'd hate to give that up.

What if we limited the restriction on self-labeling to k8s.io and kubernetes.io labels (which we can more reasonably claim ownership of).

1. Kubelets can only self-set the following labels under k8s.io and kubernetes.io namespaces:

    ```
    kubernetes.io/hostname
    failure-domain.[beta.]kubernetes.io/zone
    failure-domain.[beta.]kubernetes.io/region
    [beta.]kubernetes.io/instance-type
    [beta.]kubernetes.io/os
    [beta.]kubernetes.io/arch
    [*.]kubelet.kubernetes.io/*
    [*.]topology.kubernetes.io/*
    ```

2. Define a label prefix people who wanted to label nodes for isolation purposes can use without fear of future conflict:

    > The `node-isolation.kubernetes.io/*` label namespace is reserved for cluster deployers to isolate node pools.

That would accomplish the following:

- continue allowing people to use arbitrary labels under their own namespaces any way they wished
- provide a spot under the kubernetes.io namespace for node isolation labeling
- provide a spot under the kubernetes.io namespace for kubelets to self-label with kubelet-specific and topology-specific labels
- remove the need for any label translation

johnbelamaric:
@thockin it doesn't eliminate the ability to have arbitrary topology, I don't think it is horrible - especially with @liggitt's tweaks. At least not appreciably more horrible than what we have already.

My main concern would be that we are building semantics into Kubernetes that are based on top of the labels, which should be the operators and end users sandbox. I know it's not the first time but it starts to blur the line for when something should be a label versus a field. I wouldn't want labels to become a big dumping ground just because they are easy to add and are indexed.

@liggitt Your suggestion makes sense to me as it allows users to still have freedom. I might call it `node-restriction` not `node-isolation`, so it matches the admission controller plugin name. It makes it clear that that part of the namespace is owned by that plugin.

liggitt:
updated with changes in https://github.com/kubernetes/community/pull/911#issuecomment-434757411

smarterclayton:
This seems like a reasonable compromise.

I also think these are similar rules that I would apply to things like namespace labels or annotations (and I know we've discussed similar outcomes) where we want to:

1. allow reasonable flexibility for end users
2. give admins control over a specific space
3. take full ownership of a set of values

Given that we've already shipped the unrestricted path (anyone can set anything) taking control of our namespaces seems like the next best thing.

I'm in favor, and barring a significant failure mode we haven't thought of would approve this as is.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,131 @@
...
+facing nginx on dedicated nodes to reduce exposure to the company's publicly
+trusted server certificates. They use the secret mechanism to distribute the
+serving certificate key. An intruder captures the dedicated nginx workload in
+the same way and can now use the node certificate to read the company's serving
+certificate key.
+
+## Proposal
+
+1. Modify the `NodeRestriction` admission plugin to prevent Kubelets from self-setting labels
+within the k8s.io and kubernetes.io namespaces, *except for these specifically allowed labels/prefixes*:
```
enj: formatting nit: use backticks around the reserved namespaces

saad-ali:
/lgtm
New proposal SGTM from CSI/storage perspective. Kubelet can continue to apply arbitrary CSI without any translation. Won't require any further work on CSI side to handle.

saad-ali:
/hold
Until all approvals are in

thockin:
@liggitt

> Unifying selection on labels (across storageclass, persistentvolume, scheduler, podspec, etc) is really simple and convenient. I'd hate to give that up.

Agree

> 1. The NodeRestriction admission plugin would prevent Kubelets from self-setting labels within the k8s.io and kubernetes.io namespaces except for these specific labels/prefixes:
>
> kubernetes.io/hostname

OK

> failure-domain.[beta.]kubernetes.io/zone
> failure-domain.[beta.]kubernetes.io/region

Unfortunate, but OK

> [beta.]kubernetes.io/instance-type
> [beta.]kubernetes.io/os
> [beta.]kubernetes.io/arch

sure but something smells

> [*.]kubelet.kubernetes.io/*

Do we use this today?

> [*.]topology.kubernetes.io/*

Why?  Is anyone putting labels under there today?  I'd argue that `**.kubernetes.io` are *all* reserved and are not an extension point.

> Since this will incentivize deployers choosing labels for node isolation to use labels under kubernetes.io

I don't know if I follow?  Is the idea that since kubernetes.io is banned-for-self, they can rely on it to not be modifiable by an attack on a node (defending against east-west escalation)?

I'm not 100% against specific carve-outs (despite what I said above), but it smells like we're missing a policy somewhere else.  We really want the user to designate their own suffix that is also banned-for-self, right?  Can we do that in a simpler way than "run your own admission control" or general-purpose label-setting-policy scheme?

seh:
With this proposal, it sounds like we would lose the ability to have the kubelet set role-related labels like "node-role.kubernetes.io/master" (more generally, "node-role.kubernetes.io/\<role name\>"). Is the idea that we should apply a label like that to a worker node after running _kubeadm join_, since we couldn't ask the kubelet to label its representative node itself like that when it starts?

liggitt:
>> `[*.]kubelet.kubernetes.io/*`
>
> Do we use this today?

That was to provide a place for self-set labels in kubernetes-owned bringup scripts to migrate to over the next two releases:

https://github.com/kubernetes/kubernetes/blob/023892af3c64075473b325fbf4fa906ad4065f00/cluster/gce/config-default.sh#L46

https://github.com/kubernetes/kubernetes/blob/023892af3c64075473b325fbf4fa906ad4065f00/cluster/gce/util.sh#L539


>> `[*.]topology.kubernetes.io/*`
>
> Why? Is anyone putting labels under there today? I'd argue that **.kubernetes.io are all reserved and are not an extension point.

That was the one I was least sure about, and am fine with dropping it.


> I'm not 100% against specific carve-outs (despite what I said above), but it smells like we're missing a policy somewhere else. We really want the user to designate their own suffix that is also banned-for-self, right? Can we do that in a simpler way than "run your own admission control" or general-purpose label-setting-policy scheme?

A solution to configurable policy that we were happy with was not forthcoming (see "alternatives considered" section). I think providing a safe label prefix for this use with zero config is very valuable. If we wanted to let the cluster admin specify additional protected labels/prefixes in the future, and came up with a mechanism we were happy with, that could easily be added as a second iteration.

> With this proposal, it sounds like we would lose the ability to have the kubelet set role-related labels like "node-role.kubernetes.io/master" (more generally, "node-role.kubernetes.io/<role name>"). Is the idea that we should apply a label like that to a worker node after running _kubeadm join_, since we couldn't ask the kubelet to label its representative node itself like that when it starts?

Yes. A node should not be able to unilaterally self-state it is a master, and that master-credential-conferring daemonsets should be scheduled to it.

liggitt:
> Is the idea that since kubernetes.io is banned-for-self, they can rely on it to not be modifiable by an attack on a node (defending against east-west escalation)?

Exactly. Out of the box, there is no label prefix someone can use to isolate their nodes that is invulnerable to a compromised node. This aims to provide that.

thockin:
>>> [*.]kubelet.kubernetes.io/*

>> Do we use this today?

> That was to provide a place for self-set labels in kubernetes-owned bringup scripts to migrate to over the next two releases.

Blech.  Those are an abomination.  Why should they be under `kubelet` rather 
than `node` or something?

>>> [*.]topology.kubernetes.io/*

> That was the one I was least sure about, and am fine with dropping it.

I think we should.

> I think providing a safe label prefix for this use with zero config is very valuable.

Perhaps we'd do better to grant users a specific prefix-suffix -- e.g. make
`*.my.kubernetes.io` reserved for end-users and disallowed for 
self-modification.  The name itself could be debated, but the point is 
a for-the-user carveout rather than a for-the-system carevout.

If we're going to encourage users to use something, let's be precise?

Or maybe both.  Kubelet is allowed to write X, Y, and Z.  We explicitly grant 
users A, B, and C.  Anything else is reserved.  Not sure how we would evolve 
cleanly

enj:
>  `*.my.kubernetes.io` reserved for end-users and disallowed for self-modification

Does not the proposed `node-restriction.kubernetes.io/*` handle this?  The name is debatable but I believe the idea is the same.

thockin:
I guess so.

liggitt:
> Blech. Those are an abomination. Why should they be under `kubelet` rather
> than `node` or something?

I'd be just as happy removing those uses of the self-labeling mechanism rather than migrating them to a kubelet-specific label prefix.

I do think we are guaranteed to have use cases for kubelets to add labels in the future, and defining where they will be allowed to do that under `kubernetes.io` is important. I'm ambivalent whether that is under `kubelet.kubernetes.io` or `node.kubernetes.io` (or both).

>> [_.]topology.kubernetes.io/_
>> That was the one I was least sure about, and am fine with dropping it.
> 
> I think we should.

Dropped.

> Or maybe both. Kubelet is allowed to write X, Y, and Z (under kubernetes.io). We explicitly grant
> users A, B, and C (.kubernetes.io). Anything else .kubernetes.io is reserved. Anything outside kubernetes.io is allowed.

That's exactly what this is proposing:
* The kubelet can only write the labels in the allowed list under `kubernetes.io` / `k8s.io`
* Users are explicitly granted `node-restriction.kubernetes.io/*`
* Anything else under `kubernetes.io` / `k8s.io` is reserved
* Anything outside `kubernetes.io` / `k8s.io` is allowed

As long as we are clear about what prefix users are granted and should use for node isolation labels, I think we have flexibility on how to evolve this in the future. Can we close on this approach so this first iteration can be delivered for 1.13?

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Capturing Dedicated Workloads
+
+Suppose company `foo` needs to run an application that deals with PII on
+dedicated nodes to comply with government regulation. A common mechanism for
+implementing dedicated nodes in Kubernetes today is to set a label or taint
+(e.g. `foo/dedicated=customer-info-app`) on the node and to select these
+dedicated nodes in the workload controller running `customer-info-app`.
+
+Since the nodes self reports labels upon registration, an intruder can easily
```
thockin: I'm all for belts and suspenders, but is "an intruder registers a node into our cluster" a high-prio attack vector?

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,131 @@
...
+certificate key.
+
+## Proposal
+
+1. Modify the `NodeRestriction` admission plugin to prevent Kubelets from self-setting labels
+within the `k8s.io` and `kubernetes.io` namespaces, *except for these specifically allowed labels/prefixes*:
+
+    ```
+    kubernetes.io/hostname
+    failure-domain.[beta.]kubernetes.io/zone
```
thockin: Maybe simpler to say:

```
kubernetes.io/hostname
kubernetes.io/os
kubernetes.io/arch
kubernetes.io/instance-type
[*.]beta.kubernetes.io/* (deprecated)
failure-domain.kubernetes.io/*
[*.]kubelet.kubernetes.io/*
￼[*.]node.kubernetes.io/*
```

Could we maybe argue to allow all of naked `kubernetes.io` ?

Or at least make it clear that this list may change in the future.  Concretely, we might add more top-level things, we might enable new prefixes, and we might even provide policy rules to users.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
```diff
@@ -0,0 +1,131 @@
...
+    failure-domain.[beta.]kubernetes.io/region
+    [beta.]kubernetes.io/instance-type
+    [beta.]kubernetes.io/os
+    [beta.]kubernetes.io/arch
+    [*.]kubelet.kubernetes.io/*
+    [*.]node.kubernetes.io/*
+    ```
+
+2. Reserve/recommend the `node-restriction.kubernetes.io/*` label prefix for users
```
thockin: Is this name already codified somewhere?  I don't like the length or specificity of it.

spitballing: 

protected.kubernetes.io
user.kubernetes.io
admin.kubernetes.io
local.kubernetes.io
site.kubernetes.io
my.kubernetes.io
x.kubernetes.io

Or maybe a distinct TLD?

[*.]local/*
[*.].k8s/*

Think through whether this prefix will be applicable in any other context (one advantage of node-restriction is that it is pretty clearly node-related).

Naming is hard, but I am OK with the rest of this proposal

thockin:
I'll LGTM for merge now, but would appreciate just a BIT more thought on naming and scope.

/lgtm
/approve

liggitt:
I am, will update the PR today after thinking through Tim's last comments

thockin:
See https://github.com/kubernetes/kubernetes/pull/70555/files for another
proposed use of a label to control daemonset addons.

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
liggitt: > Could we maybe argue to allow all of naked `kubernetes.io` ?

I'd prefer to start with this specific set, and document the allowed set may grow or shrink in the future.

> Or at least make it clear that this list may change in the future. Concretely, we might add more top-level things, we might enable new prefixes, and we might even provide policy rules to users.

+1

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
liggitt: > Is this name already codified somewhere? I don't like the length or specificity of it.

No, it's new to this proposal. I agree on the length, but I like the specificity.

> Think through whether this prefix will be applicable in any other context (one advantage of node-restriction is that it is pretty clearly node-related).

having stewed on it for a few days, I actually like this prefix for this use:

* it is clearly node-related
* it only reserves a single specific label prefix
* it connects the label to the admission plugin that enforces it
* it connotes both a restriction on the node itself, and reads ok as a node selector (`node-restriction.kubernetes.io/fips=true`, `node-restriction.kubernetes.io/pci-dss=true`, etc)

keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md
smarterclayton: It turns "can launch a VM inside a given infrastructure account" into "can root the entire infrastructure account" when you take into account that the masters need certain privileges on the infrastructure account.  So I would say yes.

PR #913, opened 2017-08-15
Proposal: SubResources for CustomResources

nikhita:
[CustomResourceDefinitions](https://github.com/kubernetes/community/pull/524) (CRDs) were introduced in 1.7. The objects defined by CRDs are called CustomResources (CRs). Currently, we do not provide subresources for CRs.

However, it is one of the [most requested features](https://github.com/kubernetes/kubernetes/issues/38113) and this proposal seeks to add  `/status` and `/scale` subresources for CustomResources.

cc @sttts @deads2k @enisoc @bgrant0607 @erictune @lavalamp @brendandburns @philips @liggitt @mbohlool @fabxc @adohe @munnerz

### DIFF
--- a/contributors/design-proposals/api-machinery/customresources-subresources.md
+++ b/contributors/design-proposals/api-machinery/customresources-subresources.md
@@ -1,0 +1,203 @@
+# Subresources for CustomResources
+
+Authors: @nikhita, @sttts
+
+## Table of Contents
+
+1. [Abstract](#abstract)
+2. [Goals](#goals)
+3. [Non-Goals](#non-goals)
+4. [Proposed Extension of CustomResourceDefinition](#proposed-extension-of-customresourcedefinition)
+    1. [API Types](#api-types)
+    2. [Feature Gate](#feature-gate)
+5. [Semantics](#semantics)
+    1. [Validation Behavior](#validation-behavior)
+        1. [Status](#status)
+        2. [Scale](#scale)
+    2. [Status Behavior](#status-behavior)
+    3. [Scale Behavior](#scale-behavior)
+        1. [Status Replicas Behavior](#status-replicas-behavior)
+        2. [Selector Behavior](#selector-behavior)
+4. [Implementation Plan](#implementation-plan)
+5. [Alternatives](#alternatives)
+    1. [Scope](#scope)
+
+## Abstract
+
+[CustomResourceDefinitions](https://github.com/kubernetes/community/pull/524) (CRDs) were introduced in 1.7. The objects defined by CRDs are called CustomResources (CRs). Currently, we do not provide subresources for CRs.
+
+However, it is one of the [most requested features](https://github.com/kubernetes/kubernetes/issues/38113) and this proposal seeks to add  `/status` and `/scale` subresources for CustomResources.
+
+## Goals
+
+1. Support status/spec split for CustomResources:
+    1. Status changes are ignored on the main resource endpoint.
+    2. Support a `/status` subresource HTTP path for status changes.
+    3. `metadata.Generation` is increased only on spec changes.
+2. Support a `/scale` subresource for CustomResources.
+3. Maintain backward compatibility by allowing CRDs to opt-in to enable subresources.
+4. If a CustomResource is already structured using spec/status, allow it to easily transition to use the `/status` and `/scale` endpoint.
+5. Work seamlessly with [JSON Schema validation](https://github.com/kubernetes/community/pull/708).
+
+## Non-Goals
+
+1. Allow defining arbitrary subresources i.e. subresources except `/status` and `/scale`.
+
+## Proposed Extension of CustomResourceDefinition
+
+### API Types
+
+The addition of the following external types in `apiextensions.k8s.io/v1beta1` is proposed:
+
+```go
+type CustomResourceDefinitionSpec struct {
+    ...
+    // SubResources describes the subresources for CustomResources
+    // This field is alpha-level and should only be sent to servers that enable
+    // subresources via the CurstomResourceSubResources feature gate.
+    // +optional
+    SubResources *CustomResourceSubResources `json:"subResources,omitempty"`
+}
+
+// CustomResourceSubResources defines the status and scale subresources for CustomResources.
+type CustomResourceSubResources struct {
+    // Status denotes the status subresource for CustomResources
+    Status *CustomResourceSubResourceStatus `json:"status,omitempty"`
+    // Scale denotes the scale subresource for CustomResources
+    Scale  *CustomResourceSubResourceScale  `json:"scale,omitempty"`
+}
+
+// CustomResourceSubResourceStatus defines how to serve the HTTP path <CR Name>/status.
+type CustomResourceSubResourceStatus struct {
+}
+
+// CustomResourceSubResourceScale defines how to serve the HTTP path <CR name>/scale.
+type CustomResourceSubResourceScale struct {
+    // required, e.g. “.spec.replicas”. Must be under `.spec`.
+    // Only JSON paths without the array notation are allowed.
+    SpecReplicasPath string `json:"specReplicasPath"`
+    // optional, e.g. “.status.replicas”. Must be under `.status`.
+    // Only JSON paths without the array notation are allowed.
+    StatusReplicasPath string `json:"statusReplicasPath,omitempty"`
+    // optional, e.g. “.spec.labelSelector”. Must be under `.spec`.
+    // Only JSON paths without the array notation are allowed.
+    LabelSelectorPath string `json:"labelSelectorPath,omitempty"`
+    // ScaleGroupVersion denotes the GroupVersion of the Scale
+    // object sent as the payload for /scale. It allows transition
+    // to future versions easily.
+    // Today only autoscaling/v1 is allowed.
+    ScaleGroupVersion schema.GroupVersion `json:"groupVersion"`
+}
+```
+
+### Feature Gate
+
+The `SubResources` field in `CustomResourceDefinitionSpec` will be gated under the `CustomResourceSubResources` alpha feature gate.
+If the gate is not open, the value of the new field within `CustomResourceDefinitionSpec` is dropped on creation and updates of CRDs.
+
+### Scale type
+
+The `Scale` object is the payload sent over the wire for `/scale`. The [polymorphic `Scale` type](https://github.com/kubernetes/kubernetes/pull/53743) i.e. `autoscaling/v1.Scale` is used for the `Scale` object.
+
+Since the GroupVersion of the `Scale` object is specified in `CustomResourceSubResourceScale`, transition to future versions (eg `autoscaling/v2.Scale`) can be done easily.
+
+Note: If `autoscaling/v1.Scale` is deprecated, then it would be deprecated here as well.
+
+## Semantics
+
+### Validation Behavior
+
+#### Status
+
+The status endpoint of a CustomResource receives a full CR object. Changes outside of the `.status` subpath are ignored.
+For validation, the JSON Schema present in the CRD is validated only against the `.status` subpath.
+
+To validate only against the schema for the `.status` subpath, `oneOf` and `anyOf` constructs are not allowed within the root of the schema, but only under a properties sub-schema (with this restriction, we can project a schema to a sub-path). The following is forbidden in the CRD spec:
+
+```yaml
+validation:
+    openAPIV3Schema:
+        oneOf:
+            ...
+```
+
+**Note**: The restriction for `oneOf` and `anyOf` allows us to write a projection function `ProjectJSONSchema(schema *JSONSchemaProps, path []string) (*JSONSchemaProps, error)` that can be used to apply a given schema for the whole object to only the sub-path `.status` or `.spec`.
+
+#### Scale
+
+Moreover, if the scale subresource is enabled:
+
+On update, we copy the values from the `Scale` object into the specified paths in the CustomResource, if the path is set (`StatusReplicasPath` and `LabelSelectorPath` are optional).
+If `StatusReplicasPath` or `LabelSelectorPath` is not set, we validate that the value in `Scale` is also not specified and return an error otherwise.
+
+On `get` and on `update` (after copying the values into the CustomResource as described above), we verify that:
+
+- The value at the specified JSON Path `SpecReplicasPath` (e.g. `.spec.replicas`) is a non-negative integer value and is not empty.
+
+- The value at the optional JSON Path `StatusReplicasPath` (e.g. `.status.replicas`) is an integer value if it exists (i.e. this can be empty).
+
+- The value at the optional JSON Path `LabelSelectorPath` (e.g. `.spec.labelSelector`) is a valid label selector if it exists (i.e. this can be empty).
+
+**Note**: The values at the JSON Paths specified by `SpecReplicasPath`, `LabelSelectorPath` and `StatusReplicasPath` are also validated with the same rules when the whole object or, in case the `/status` subresource is enabled, the `.status` sub-object is updated.
+
+### Status Behavior
+
+If the `/status` subresource is enabled, the following behaviors change:
+
+- The main resource endpoint will ignore all changes in the status subpath.
+(note: it will **not** reject requests which try to change the status, following the existing semantics of other resources).
+
+- The `.metadata.generation` field is updated if and only if the value at the `.spec` subpath changes.
+Additionally, if the spec does not change, `.metadata.generation` is not updated.
+
+- The `/status` subresource receives a full resource object, but only considers the value at the `.status` subpath for the update.
+The value at the `.metadata` subpath is **not** considered for update as decided in https://github.com/kubernetes/kubernetes/issues/45539.
+
+Both the status and the spec (and everything else if there is anything) of the object share the same key in the storage layer, i.e. the value at  `.metadata.resourceVersion` is increased for any kind of change. There is no split of status and spec in the storage layer.
+
+The `/status` endpoint supports both `get` and `update` verbs.
+
+### Scale Behavior
+
+The number of CustomResources can be easily scaled up or down depending on the replicas field present in the `.spec` subpath.
+
+Only `ScaleSpec.Replicas` can be written. All other values are read-only and changes will be ignored. i.e. upon updating the scale subresource, two fields are modified:
+
+1. The replicas field is copied back from the `Scale` object to the main resource as specified by `SpecReplicasPath` in the CRD, e.g.  `.spec.replicas = scale.Spec.Replicas`.
+
+2. The resource version is copied back from the `Scale` object to the main resource before writing to the storage: `.metadata.resourceVersion = scale.ResourceVersion`.
+In other words, the scale and the CustomResource share the resource version used for optimistic concurrency.
+Updates with outdated resource versions are rejected with a conflict error, read requests will return the resource version of the CustomResource.
+
+The `/scale` endpoint supports both `get` and `update` verbs.
+
+#### Status Replicas Behavior
+
+As only the `scale.Spec.Replicas` field is to be written to by the CR user, the user-provided controller (not any generic CRD controller) counts its children and then updates the controlled object by writing to the `/status` subresource, i.e. the `scale.Status.Replicas` field is read-only.
+
+#### Selector Behavior
+
+`CustomResourceSubResourceScale.LabelSelectorPath` is the label selector over CustomResources that should match the replicas count.
+The value in the `Scale` object is one-to-one the value from the CustomResource if the label selector is non-empty.
+Intentionally we do not default it to another value from the CustomResource (e.g. `.spec.template.metadata.labels`) as this turned out to cause trouble (e.g. in `kubectl apply`) and it is generally seen as a wrong approach with existing resources.
+
+## Implementation Plan
+
+The `/scale` and `/status` subresources are mostly distinct. It is proposed to do the implementation in two phases (the order does not matter much):
+
+1. `/status` subresource
+2. `/scale` subresource
+
+## Alternatives
+
+### Scope
+
+In this proposal we opted for an opinionated concept of subresources i.e. we restrict the subresource spec to the two very specific subresources: `/status` and `/scale`.
+We do not aim for a more generic subresource concept. In Kubernetes there are a number of other subresources like `/log`, `/exec`, `/bind`. But their semantics is much more special than `/status` and `/scale`.
+Hence, we decided to leave those other subresources to the domain of User provided API Server (UAS) instead of inventing a more complex subresource concept for CustomResourceDefinitions.
+
+**Note**: The types do not make the addition of other subresources impossible in the future.
+
+We also restrict the JSON path for the status and the spec within the CustomResource.
+We could make them definable by the user and the proposed types actually allow us to open this up in the future.
+For the time being we decided to be opinionated as all status and spec subobjects in existing types live under `.status` and `.spec`. Keeping this pattern imposes consistency on user provided CustomResources as well.


nikhita:
Ref: https://github.com/kubernetes/kubernetes/issues/38113.

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+    3. `metadata.Generation` is increased only on spec changes.
+2. Support a `/scale` subresource for CustomResources.
+3. Maintain backward compatibility by allowing CRDs to opt-in to enable subresources.
+4. If a CustomResource is already structured using spec/status, allow it to easily transition to use the `/status` and `/scale` endpoint.
+5. Work seamlessly with [JSON Schema validation](https://github.com/kubernetes/community/pull/708).
+
+## Non-Goals
+
+1. Allow defining arbitrary subresources i.e. subresources except `/status` and `/scale`.
+2. Unify the many `Scale` types: as long as there is no generic `Scale` object in Kubernetes, we will propose to introduce yet another `Scale` type in the `apiextensions.k8s.io` api group. If Kubernetes switches to a generic `Scale` object, `apiextensions.k8s.io` will follow.
```
liggitt: Making scale operate with a different API type will prevent it from working with HPA. I'd expect it to use the same API type

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// CustomResourceSubResourceStatus defines how to serve the HTTP path <CR Name>/status.
+type CustomResourceSubResourceStatus struct {
+    // The JSON path (e.g. “.status”) of the status of a CustomResource.
+    // The whole object is transferred over the wire, but only the status can be mutated via the /status subresource.
+    // StatusPath is restricted to be “.status” and defaults to “.status”.
+    StatusPath string `json:“statusPath,omitempty”`
+    // The JSON path (e.g. “.spec”) of the spec of a CustomResource.
+    // SpecPath is restricted to be “.spec” and defaults to “.spec”.
```
liggitt: Is there a reason to make spec/status paths be specified if they are locked to a single path?

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+type CustomResourceSubResourceScale struct {
+    // required, e.g. “.spec.replicas”.
+    // Only JSON paths without the array notation are allowed.
+    SpecReplicasPath string `json:“specReplicasPath,omitempty”`
+    // optional, e.g. “.status.replicas”.
+    // Only JSON paths without the array notation are allowed.
+    StatusReplicasPath string `json:“statusReplicasPath,omitempty”`
+    // optional, e.g. “.spec.labelSelector”.
+    // Only JSON paths without the array notation are allowed.
+    LabelSelectorPath string `json:“labelSelectorPath,omitempty”`
```
liggitt: Define behavior if the specified paths do not exist or do not contain valid data (non-integers for scale paths, non-selector data for the selector path, etc)

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+    // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
+    // +optional
+    Selector *metav1.LabelSelector `json:"selector"`
+}
+```
+
+### Feature Gate
+
+The `SubResources` field in `CustomResourceDefinitionSpec` will be gated under the `CustomResourceSubResources` alpha feature gate.
+If the gate is not open, the value of the new field within `CustomResourceDefinitionSpec` is rejected on creation and updates of CRDs.
```
liggitt: Additionally, if the gate is off, the alpha fields should be dropped from incoming requests (see https://github.com/kubernetes/community/pull/869#discussion_r132290125)

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+    // label query over pods that should match the replicas count.
+    // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
+    // +optional
+    Selector *metav1.LabelSelector `json:"selector"`
+}
+```
+
+### Feature Gate
+
+The `SubResources` field in `CustomResourceDefinitionSpec` will be gated under the `CustomResourceSubResources` alpha feature gate.
```
liggitt: nit: I think I'd expect this to be "Subresources", since we generally say "subresources", not "sub resources" or "sub-resources"

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+- The value at the optional JSON Path `LabelSelectorPath` (e.g. `.spec.labelSelector`) is validated to be a valid label selector if it exists (i.e. this can be empty). 
+
+### Status Behavior
+
+If the `/status` subresource is enabled, the following behaviors change:
+
+- The main resource endpoint will ignore all changes in the specified status subpath.
+(note: it will **not** reject requests which try to change the status, following the existing semantics of other resources).
+
+- The `.metadata.generation` field is updated if and only if the value at the specified `SpecPath` (e.g. `.spec`)  changes.
```
liggitt: Does it prevent changing generation if spec does not change?

contributors/design-proposals/customresources-subresources.md
sttts: Same in which sense? Same GroupVersion? Extensionally equally?

contributors/design-proposals/customresources-subresources.md
sttts: We want the possibility to add fields later (i.e. no bool, but a struct) and we want to avoid an empty struct. Any better idea?

contributors/design-proposals/customresources-subresources.md
sttts: "do not exist" is described below. Invalid data after the path has been changed in a CRD (can it be changed?) must be added.

contributors/design-proposals/customresources-subresources.md
sttts: This was not clear from existing implementation and the api convention text. Thanks for clarifying. This matches my understanding.

contributors/design-proposals/customresources-subresources.md
sttts: The understanding is that the update endpoint increases the generation iff it sees a difference in the spec. What is the expected behaviour about posting an object with modified Generation? Currently the proposal says that this value if overridden by the strategy.

contributors/design-proposals/customresources-subresources.md
enisoc: I thought so too, until I found that we use `SubResource` in both client-go and as a field in `AdmissionReviewSpec`. Should we be consistent with English, or with the rest of k8s?

contributors/design-proposals/customresources-subresources.md
enisoc: @foxish Is the plan to move everything to `autoscaling.Scale`?

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+
+// ScaleStatus represents the current status of a scale subresource.
+type ScaleStatus struct {
+    // actual number of observed instances of the scaled object.
+    Replicas int32 `json:"replicas"`
+
+    // label query over pods that should match the replicas count.
+    // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
+    // +optional
+    Selector *metav1.LabelSelector `json:"selector"`
```
enisoc: In the external (versioned) structs, all of `apps.ScaleStatus`, `extensions.ScaleStatus`, and `autoscaling.ScaleStatus` use a string (encoded as a label query) for the selector. The internal structs vary: `extensions.ScaleStatus` uses `*metav1.LabelSelector` internally, but `autoscaling.ScaleStatus` keeps it as a string.

We should make this consistent, although I don't know with what.

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+If the gate is not open, the value of the new field within `CustomResourceDefinitionSpec` is rejected on creation and updates of CRDs.
+
+## Semantics
+
+### Validation Behavior
+
+#### Status
+
+The status endpoint of a CustomResource receives a full CR object. Changes outside of the status path are ignored.
+For validation everything outside of the JSON Path `StatusPath` can be reset to the existing values of the CustomResource. Then the JSON Schema presented in the CRD is validated against the whole object.
```
enisoc: It will still pass along the ResourceVersion from the update request to be verified, right? We need to be careful not to overwrite it with the old value.

nikhita: Yes, it will use the new ResourceVersion value. :+1:

sttts: For the validation step the resource version is not important. It will play a role when sending the update object to the storage layer. For that the described semantics are applied, i.e. the status ResourceVersion is used in order to be consistent with optimistic consistency of the whole object.

contributors/design-proposals/customresources-subresources.md
sttts: Right. We did not really specify whether we describe the internal or external types. But the way to go for the selector in the external types is to use a string. Good catch. @nikhita can you update that?

nikhita: Updated. Thanks, @enisoc.

contributors/design-proposals/customresources-subresources.md
sdminonne: /sub

contributors/design-proposals/customresources-subresources.md
kargakis: > What is the expected behaviour about posting an object with modified Generation?

For core objects, we ignore the change.

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Proposed Extension of CustomResourceDefinition
+
+### API Types
+
+The addition of the following external types in `apiextensions.k8s.io/v1beta1` is proposed:
+
+```go
+type CustomResourceDefinitionSpec struct {
+    Group string
```
kargakis: Can you omit all existing fields?

contributors/design-proposals/customresources-subresources.md
sttts: @kargakis thanks for clarification. So this means: ignore = we reset it to the old value if spec is unchanged.

In other words, metadata.Generation is completely uncontrollable for the controllers. This is probably a good thing for consistency.

contributors/design-proposals/customresources-subresources.md
sttts: Just some `...`, yes.

contributors/design-proposals/customresources-subresources.md
nikhita: > For core objects, we ignore the change.

@kargakis Maybe missing something here but I can't find any place where we reset metadata.Generation to the old value if spec is unchanged. One example: https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/replicationcontroller/strategy.go#L70.

kargakis: We restore to the old generation one level prior to calling the resource-specific rest handlers in https://github.com/kubernetes/apiserver/blob/149fc2228647cea28b0670c240ec582e985e8eda/pkg/registry/rest/update.go#L100

contributors/design-proposals/customresources-subresources.md
nikhita: Added.

contributors/design-proposals/customresources-subresources.md
nikhita: Thanks, @kargakis. Updated.

contributors/design-proposals/customresources-subresources.md
nikhita: Done.

contributors/design-proposals/customresources-subresources.md
nikhita: Updated.

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+## Semantics
+
+### Validation Behavior
+
+#### Status
+
+The status endpoint of a CustomResource receives a full CR object. Changes outside of the status path are ignored.
+For validation everything outside of the JSON Path `StatusPath` can be reset to the existing values of the CustomResource. Then the JSON Schema presented in the CRD is validated against the whole object.
+
+Note: one could extract the status suboject, filter the JSON schema by rules applied to the status and then validate the status subobject only.
```
ash2k: Typo in `suboject`

contributors/design-proposals/customresources-subresources.md
deads2k: HPA should be able to speak multiple versions of the scale subresource exposed via discovery: https://github.com/kubernetes/kubernetes/pull/49971 .  I think I'd rather have autoscaling depend on apiextensions than the other way around.

contributors/design-proposals/customresources-subresources.md
deads2k: > We want the possibility to add fields later (i.e. no bool, but a struct) and we want to avoid an empty struct. Any better idea?

If you make it a pointer (like it is), doesn't the presence or absence of the field trigger you?

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// CustomResourceSubResourceScale defines how to serve the HTTP path <CR name>/scale.
+type CustomResourceSubResourceScale struct {
+    // required, e.g. “.spec.replicas”.
+    // Only JSON paths without the array notation are allowed.
+    SpecReplicasPath string `json:“specReplicasPath,omitempty”`
+    // optional, e.g. “.status.replicas”.
+    // Only JSON paths without the array notation are allowed.
+    StatusReplicasPath string `json:“statusReplicasPath,omitempty”`
```
deads2k: Is the belief that we are too late to be prescriptive?

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+    // optional, e.g. “.spec.labelSelector”.
+    // Only JSON paths without the array notation are allowed.
+    LabelSelectorPath string `json:“labelSelectorPath,omitempty”`
+}
+
+// The following is the payload to send over the wire for /scale. It happens
+// to be defined here in apiextensions.k8s.io because we don’t have a global
```
deads2k: And I don't think that we should.  If polymorphic endpoints are to succeed, they need to be able to exist outside of meta.k8s.io.

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+If the gate is not open, the value of the new field within `CustomResourceDefinitionSpec` is dropped on creation and updates of CRDs.
+
+## Semantics
+
+### Validation Behavior
+
+#### Status
+
+The status endpoint of a CustomResource receives a full CR object. Changes outside of the status path are ignored.
+For validation everything outside of the JSON Path `StatusPath` can be reset to the existing values of the CustomResource. Then the JSON Schema presented in the CRD is validated against the whole object.
```
deads2k: I'd recommend skipping validation on the spec.  If you tighten validation, you don't generally want to reject callers which are writing "this is the state of the world".

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+#### Scale
+
+Moreover, if the scale subresource is enabled:
+
+- The value at the specified JSON Path `SpecReplicasPath` (e.g. `.spec.replicas`) is validated to ensure that it contains non-negative integer value and is not empty.
+
+- The value at the optional JSON Path `StatusReplicasPath` (e.g. `.status.replicas`) is validated to be a an integer number if it exists (i.e. this can be empty).
+
+- The value at the optional JSON Path `LabelSelectorPath` (e.g. `.spec.labelSelector`) is validated to be a valid label selector if it exists (i.e. this can be empty).
+If it is invalid, an empty `LabelSelector` is used.
```
deads2k: Why does this field allow invalid values?

contributors/design-proposals/customresources-subresources.md
sttts: we can choose to have fixed paths or what we have for  `EmptyDir` today: `StatusSubResource: {}`. 

Which way to we like more? I don't like the `{}` notation, but that probably the way to go in YAML, as ugly as it is.

contributors/design-proposals/customresources-subresources.md
sttts: Not a decission we have to make here now. I hope until implementation this topic is solved.

contributors/design-proposals/customresources-subresources.md
sttts: That's the effect. It's just an implementation detail with the same semantics as "checking only the status".

contributors/design-proposals/customresources-subresources.md
sttts: Use-case: opt-in to subresources later-on.

contributors/design-proposals/customresources-subresources.md
sttts: @enisoc ^^, you had the argument that we have plenty of paths already today in existing resources. Moreover, there might be a number of existing TPR/CRD users who have chosen their schema. By being prescriptive we make it impossible for them to opt-in.

enisoc: My argument was that the purpose of the /scale subresource, as far as I know, is to abstract away the knowledge of which field a given object considers to represent its "scale". If we wanted to go the route of a convention for field names, like it must be `.status.replicas`, then there would be no need for a subresource. You could just look directly at `.status.replicas`.

This is in contrast to the `/status` subresource, which I argued ought to be restricted to the convention of `.spec` and `.status` field names. The `/status` endpoint returns the whole object, not a sub-field or a new type like `/scale` does, and so the only way the caller knows which part is the status is by agreeing on the conventional field name. Similarly, the caller needs to agree that `.spec` is the field name that triggers `.metadata.generation` to be incremented and so on.

Because `/scale` instead returns a separate, abstracted type, there is no need to agree on the underlying field names.

contributors/design-proposals/customresources-subresources.md
enisoc: I think @deads2k has a point. Consider:

1) Create object.
2) Tighten validation such that the existing object Spec no longer passes.
3) Controller tries to update object Status, perhaps to indicate the object is invalid.

I'm not sure if this is worth fixing though, if all it would do is allow Status to be updated when the Spec is invalid. Users could still hit the same problem by tightening validation on Status itself.

contributors/design-proposals/customresources-subresources.md
deads2k: > Which way to we like more? I don't like the {} notation, but that probably the way to go in YAML, as ugly as it is.

I prefer the `{}` to fields that require specific values as a marker.

contributors/design-proposals/customresources-subresources.md
deads2k: > I'm not sure if this is worth fixing though, if all it would do is allow Status to be updated when the Spec is invalid. Users could still hit the same problem by tightening validation on Status itself.

It's been a big deal for first-class resources during upgrades. It's worth fixing.

contributors/design-proposals/customresources-subresources.md
sttts: We discussed that before (in the proposal I think) and concluded that the user is responsible to ensure that old objects validate. There is not much we can do. I don't see this case here any different.

contributors/design-proposals/customresources-subresources.md
sttts: @deads2k looks like your comment is placed in the wrong thread. Can you elaborate what you mean worth fixing?

contributors/design-proposals/customresources-subresources.md
deads2k: > We discussed that before (in the proposal I think) and concluded that the user is responsible to ensure that old objects validate. There is not much we can do. I don't see this case here any different.

There's no way to identify all the validation rules under `status.` in the jsonpath description and only run those?

sttts: At the very least for `OneOf` of JSON schema it's not that obvious how to cut down a schema in general. If I don't miss anything, the other JSON schema constructs are at least monotonic, i.e. if we remove rules from the schema it will become weaker. But still it needs some deep thought to implement such a projection to a JSON path. My current gut feeling is that with the exception of `OneOf` it's doable. My argument above was that it's not worth it to investigate or implementation this if we can test the whole object after restoring the spec and metadata.

lpabon: I think to avoid confusion, the paper should provide an example use case work flow, which shows how the data will be interpreted.

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,221 @@
...
+
+1. The replicas field is copied back from the `Scale` object to the main resource as specified by `SpecReplicasPath` in the CRD, e.g.  `.spec.replicas = scale.Spec.Replicas`.
+
+2. The resource version is copied back from the `Scale` object to the main resource before writing to the storage: `.metadata.resourceVersion = scale.ResourceVersion`.
+In other words, the scale and the CustomResource share the resource version used for optimistic concurrency.
+Updates with outdated resource versions are rejected with a conflict error, read requests will return the resource version of the CustomResource.
+
+#### Status Replicas Behavior
+
+As only the `scale.Spec.Replicas` field is to be written to by the CR user, the user-provided controller (not any generic CRD controller) counts its children and then updates the controlled object by writing to the `/status` subresource, i.e. the `scale.Status.Replicas` field is read-only.
```
lpabon: @nikhita I'm new to this part. Why is it that `.spec.replicas` is read-only?

nikhita: > @nikhita I'm new to this part. Why is it that .spec.replicas is read-only?

we cannot write to the spec or the status of the CR using the `/scale` subresource. Additionally,

- `.spec.replicas` can be written to by the CR user by updating the object.
- `.status.replicas` can be written to by the CR controller using `/status` subresource.

liggitt: typically, spec.replicas is mutable via the scale subresource

sttts: To clarify and sum up:
- values in `/ .spec.replicas` (the main resource) and `/scale .spec.replicas` are the same value. Both can be written if RBAC rules allow that.
- values `/status .status.replicas` and `/scale .spec.status` are the same value. The first **can** be written if RBAC rules allow that, the second can't.
   
In practice an autoscaling controller might only write `/scale .spec.replicas` (and RBAC will be restrictive to only allow that). A user might do `kubectl edit` and write to `/ .spec.replicas`, or he uses `kubectl scale` which writes to `/scale .spec.replicas`. A controller will write to `/status .status.replicas`, but not to `/ .spec.replicas` or `/scale .spec.replicas`.

rjeberhard:
We are working on an operator and a CRD.  With this proposed change and our creating the necessary SubResource, would we then be able to give administrators the option of scaling with HPA?

enisoc:
@rjeberhard wrote:
> With this proposed change and our creating the necessary SubResource, would we then be able to give administrators the option of scaling with HPA?

Yup, that's the plan. There is another piece in addition to this that also needs to fall into place, though. Currently HPA can only work with things in the `extensions` API group, but they are planning to make it fully generic so it can work with any object that implements the Scale subresource according to the convention that we're proposing to support here.

contributors/design-proposals/customresources-subresources.md
sttts: As there are CRDs out in the wild which have chosen a certain replicas path, I still think we should support abritrary JSON paths for the replicas fields. @deads2k is this something you can live with? 

In contrast, I agree that for `.spec` and `.status` we should prescribe those paths. I would expect that 99% have users have chosen those anyway.

deads2k: > @deads2k is this something you can live with?

I can live with arbitrary json paths for the replicas fields.

contributors/design-proposals/customresources-subresources.md
deads2k: I still feel pretty strongly about this.   Looks like one of my previous comments landed in the wrong thread:

> I'm not sure if this is worth fixing though, if all it would do is allow Status to be updated when the Spec is invalid. Users could still hit the same problem by tightening validation on Status itself.

It's been a big deal for first-class resources during upgrades. It's worth fixing.



> My argument above was that it's not worth it to investigate or implementation this if we can test the whole object after restoring the spec and metadata.

That doesn't do the same thing.  If validation is tightened on spec, you can still fail status updates on it.  I don't even mind the validation running as long we never report or fail on bad spec.

contributors/design-proposals/customresources-subresources.md
bgrant0607: An aside:

The main advantage of subresources is that they enable distinct authorization policies for the subresource endpoints.

To enable just polymorphism of spec and status, I want to make duck typing work, more like our metadata convention. That would be both a simpler mechanism and would be compatible with client-based tools. `kubectl set replicas` should be able to set the replicas field of any API that has a replicas field.

contributors/design-proposals/customresources-subresources.md
bgrant0607: Add fields to what? Going back to @liggitt's original question: I do not think the spec and status paths should be configurable.

contributors/design-proposals/customresources-subresources.md
bgrant0607: I am not ok with arbitrary paths.

bgrant0607:
@enisoc What is our current plan for this feature? Is anyone working on it in 1.9?

nikhita:
@bgrant0607 I am working on this feature. It is currently blocked on the decision for polymorphic scale: https://github.com/kubernetes/kubernetes/issues/49504#issuecomment-328449194.

contributors/design-proposals/customresources-subresources.md
sttts: @bgrant0607 I don't care much for new CRDs. Forcing new users to use a certain replicas field path is fine. But with dropping the ability of custom paths IMO we fail on goal 4: transition of existing CRDs. If we accept that, fixed paths are fine.

bgrant0607:
@nikhita Thanks. I'll see what I can do to unblock it.

contributors/design-proposals/customresources-subresources.md
nikhita: Update the proposal. Now, `status` and `spec` paths are hardcoded to `.status` and `.spec` respectively and are no longer configurable.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/913)

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,213 @@
...
+
+If the `/status` subresource is enabled, the following behaviors change:
+
+- The main resource endpoint will ignore all changes in the status subpath.
+(note: it will **not** reject requests which try to change the status, following the existing semantics of other resources).
+
+- The `.metadata.generation` field is updated if and only if the value at the `.spec` subpath changes.
+Additionally, if the spec does not change, `.metadata.generation` is not updated.
+
+- The `/status` subresource receives a full resource object, but only considers the value at the `.status` subpath for the update.
```
hongchaodeng: The following is confusing.

> only considers the value at the `.status` subpath for the update.

This already said only `.status` would be considered.

> +The value at the `.metadata` subpath is **not** considered for update

Isn't this unnecessary? Because `.metadata` is obviously not part of `.status`?

liggitt: For legacy reasons, many resources allow updating metadata via the status subresource. That was primarily an accident, but dependent behavior was built around it.

For new resources and CRDs, it is valuable to define status updates more strictly

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,213 @@
...
+    2. Support a `/status` subresource HTTP path for status changes.
+    3. `metadata.Generation` is increased only on spec changes.
+2. Support a `/scale` subresource for CustomResources.
+3. Maintain backward compatibility by allowing CRDs to opt-in to enable subresources.
+4. If a CustomResource is already structured using spec/status, allow it to easily transition to use the `/status` and `/scale` endpoint.
+5. Work seamlessly with [JSON Schema validation](https://github.com/kubernetes/community/pull/708).
+
+## Non-Goals
+
+1. Allow defining arbitrary subresources i.e. subresources except `/status` and `/scale`.
```
fabiand: Is there a separate proposal for arbitrary subresources?

KubeVirt is considering this to support graphical console access, i.e. using `virtualmachines/testvm/spice`

liggitt: Not for CRD objects. Arbitrary subresources with connect-type behavior must be implemented by actual add-on API servers.

fabiand: Right, but can an add-on API server be registered to provide a sub-resource for a CRD?

liggitt: No, a resource and its subresources are owned by a single server. In the case of CRDs, that is the kube apiserver

fabiand: Right, thanks for clarification. That was my understanding, your previous comment just confused me a little.

contributors/design-proposals/customresources-subresources.md
nikhita: @deads2k Updated the proposal. Now we validate only against the status.

contributors/design-proposals/customresources-subresources.md
sttts: If nobody speaks up, I propose to use what we have here now. @liggitt last chance to articulate a strong opinion.

sttts:
Does anybody have objections to proceed with this proposal? As far as I see, all outstanding comments are addressed. We would like to go forward with the alpha implementation for 1.9.

contributors/design-proposals/customresources-subresources.md
liggitt: no strong opinion

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,184 @@
...
+    // Scale denotes the scale subresource for CustomResources
+    Scale  *CustomResourceSubResourceScale  `json:"scale,omitempty"`
+}
+
+// CustomResourceSubResourceStatus defines how to serve the HTTP path <CR Name>/status.
+type CustomResourceSubResourceStatus struct {
+}
+
+// CustomResourceSubResourceScale defines how to serve the HTTP path <CR name>/scale.
+type CustomResourceSubResourceScale struct {
```
deads2k: How are you going to handle versioning of the scale object you're returning?  Right now, there's only one reasonable choice (`Scale.v1.autoscaling`), but to handle a future change that won't break API compatibility for the CustomResource, don't you have to make that explicit?

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+### Feature Gate
+
+The `SubResources` field in `CustomResourceDefinitionSpec` will be gated under the `CustomResourceSubResources` alpha feature gate.
+If the gate is not open, the value of the new field within `CustomResourceDefinitionSpec` is dropped on creation and updates of CRDs.
+
+### Scale type
+
+The `Scale` object is the payload sent over the wire for `/scale`. The [polymorphic `Scale` type](https://github.com/kubernetes/kubernetes/pull/53743) i.e. `autoscaling/v1.Scale` is used for the `Scale` object.
```
deads2k: Need a discussion of how you'll handle a transition to `autoscaling/v2.Scale`

contributors/design-proposals/customresources-subresources.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Semantics
+
+### Validation Behavior
+
+#### Status
+
+The status endpoint of a CustomResource receives a full CR object. Changes outside of the `.status` subpath are ignored.
```
deads2k: Which verbs are you supporting on this endpoint?

contributors/design-proposals/customresources-subresources.md
deads2k: > By changing (or setting for the first time) LabelSelectorPath in an existing CRD. Then we have no chance to validate this post-creation in all CRs.

In such a case, why shouldn't we return an error from the API until the data is correct?  The data isn't doing what the user creating it thought it should.  An empty label selector doesn't necessarily mean the same thing.

contributors/design-proposals/customresources-subresources.md
sttts: Good point. It's about this mapping server side for the CR scale object: https://github.com/kubernetes/kubernetes/pull/53743/commits/d61a2d90372c301dd11088df8941acf2bb01c38c#diff-49992b38a536f6f81a4100abe29c1334R38

What are our options? We can put a GVK here, defaulting and validating to autoscaling/v1.Scale today. When v2 comes out we allow both and when v1 is deprecated, we deprecate the v1 value here as well. Technically, we don't have to add this GVK fields today, but should do so reasonably early before a v1 deprecation comes to the table. But having a plan and implementing that today doesn't harm either.

We can accept v1 and v2 scale types on the endpoint without a problem. But we can only send one of them to the client and could control that with the GVK field. This would be a hard switch as soon as the CRD is updated. To make this smooth on the (go-)client side one could make the polymorphic discovery client to handle both types gracefully.

@deads2k What are you thoughts about this?

contributors/design-proposals/customresources-subresources.md
sttts: "If it is invalid, an empty `LabelSelector` is used." means that a label selector pointing to something not-existing in the CR maps to the default value of an empty label selector.

In general, we don't want to synchronously check all CRs in etcd of a given GV against changes in these fields. I.e. we don't want to implement a validation which takes a snapshot of all CR of a CRD and verifies them. First, there can be many, second to make that consistent we have to disable updates on CRs during that time. So the choice here is to have a good enough fall-back that it does not break the CRD user's neck, but give him the responsibility to define consistent things for his CRs. Defining a CRD with subresources is nothing you do every day or even ad-hoc.

contributors/design-proposals/customresources-subresources.md
deads2k: > @deads2k What are you thoughts about this?

I agree with your idea of baking it in early. It mirrors my thoughts.

contributors/design-proposals/customresources-subresources.md
sttts: We have to clarify:
- a given path is empty. This maps to the default value (0 for the replicas and empty for the label selector)
- a given path exists, but it does not validate (no integer or no label selector). In that case we can return an error.

The error code, as this is in fact a server caused error, should be in the 5xx range. But also 422 (and strict following the spec also 403) are good candidates.

contributors/design-proposals/customresources-subresources.md
nikhita: get and update for both `/status` and `/scale`.

deads2k:
Needs a few updates, but nothing fundamental.  Go ahead and update the features repo to list this for 1.9.  Update the doc for 
 1. verbs on subresources
 2. GVK for scale
 3. clarify invalid labelselector handling.

nikhita:
@deads2k Updated the proposal. If this lgty, I'll squash the commits.

Also, the proposal needs to be moved to `contributors/design-proposals/api-machinery` since all the proposals were re-organized recently (this will not show the diffs/comments anymore, so waiting for final lgtm).

contributors/design-proposals/customresources-subresources.md
```diff
@@ -75,13 +75,18 @@ type CustomResourceSubResourceStatus struct {
...
     // Only JSON paths without the array notation are allowed.
     StatusReplicasPath string `json:"statusReplicasPath,omitempty"`
     // optional, e.g. “.spec.labelSelector”. Must be under `.spec`.
     // Only JSON paths without the array notation are allowed.
     LabelSelectorPath string `json:"labelSelectorPath,omitempty"`
+    // ScaleGroupVersion denotes the GroupVersion of the Scale
+    // object sent as the payload for /scale. It allows transition
+    // to future versions easily.
+    // Today only autoscaling/v1 is allowed.
+    ScaleGroupVersion schema.GroupVersion `json:"groupVersion"`
```
deads2k: this is a logical type, but don't actually embed the type here

sttts: The external type will have a string, the internal the logical one.

deads2k: > The external type will have a string, the internal the logical one.

Ok.

deads2k:
lgtm. squash for merge.

nikhita:
@deads2k Done.

abbi-gaurav:
Is there any related issue for the `status` implementation. I can see one for the `scale` though.

contributors/design-proposals/api-machinery/customresources-subresources.md
```diff
@@ -0,0 +1,203 @@
...
+    // Only JSON paths without the array notation are allowed.
+    StatusReplicasPath string `json:"statusReplicasPath,omitempty"`
+    // optional, e.g. “.spec.labelSelector”. Must be under `.spec`.
+    // Only JSON paths without the array notation are allowed.
+    LabelSelectorPath string `json:"labelSelectorPath,omitempty"`
+    // ScaleGroupVersion denotes the GroupVersion of the Scale
+    // object sent as the payload for /scale. It allows transition
+    // to future versions easily.
+    // Today only autoscaling/v1 is allowed.
+    ScaleGroupVersion schema.GroupVersion `json:"groupVersion"`
```
lavalamp: I thought we agreed to remove this field in another thread? Did I miss some followup?

nikhita: > I thought we agreed to remove this field in another thread? Did I miss some followup?

It is removed. :+1: 

The proposal included this field. It was later decided in the implementation PR to remove it and so it was removed there.

Only the proposal contains it. It needs to updated (will do).

nikhita:
> Is there any related issue for the status implementation. I can see one for the scale though.

@abbi-gaurav https://github.com/kubernetes/kubernetes/pull/55168 contains the implementation for scale as well as status. You can find docs on how to use status here: https://deploy-preview-7439--kubernetes-io-vnext-staging.netlify.com/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#status-subresource.

PR #917, opened 2017-08-15
Update SIG Release info

calebamiles:
We moved into the top level `kubernetes/sig-release` but need to
include ourselves in the canonical list


cblecker:
Can you run `make` on this and commit the changes?

calebamiles:
No @cblecker I can't, running on master from a clean checkout

```
docker build -q -t kube-communitydocs -f generator/Dockerfile generator
sha256:dd005cf512b074016d7cd0146c18872bf5eed75d55079071e7e52e2fbcc77815
docker run --rm -e WG -e SIG -v /home/caleb/workspace/community:/go/src/app/generated kube-communitydocs app
2017/08/16 00:04:00 open generated/sigs.yaml: permission denied
make: *** [Makefile:13: generate] Error 1
```

SIG release is now a [top level repository](https://github.com/kubernetes/sig-release) so the `sig-release` directory is empty except for a forwarding link to the new location. The proposed changes were produced by running

```
docker run 6da0415bdb00 go-wrapper run app.go
```

by hand and then copying the pertinent changes (only affecting `sig-list.md`) back into the tree. I believe the [README](https://github.com/kubernetes/community/blob/master/generator/README.md) needs to be updated because trying to update the docs for a single SIG produces

```
make: *** No rule to make target 'gen-docs'.  Stop.
```

I have also made sure to do the ["remove all containers and images"](https://techoverflow.net/2013/10/22/docker-remove-all-images-and-containers/) song and dance.

pwittrock:
The changes LGTM

jbeda:
Can we get an updated `sig-release/README.md` too?  

(We discussed on slack making the script work with SELinux with a `:Z` option on the docker mount in the script)

jbeda:
@cblecker It looks like the submit queue is barfing on this (because `sig-release/README.md` hasn't been update, I think) but there is no way to tell what is going on from the "Details" link above.

jbeda:
Canceled as there *was* a presubmit check for this stuff but it looks to be disabled now.  Was going to rely on that. I'll wait until README.md is updated to LGTM.

jbeda:
/lgtm

Thanks @calebamiles!  Hopefully we can make this more automatic and more correct over time.

PR #927, opened 2017-08-16
Update godep instructions to recreate vendor/BUILD file

ixdy:
If we don't recreate the `vendor/BUILD` file after deleting it, then the `all-srcs` rule in the root `BUILD.bazel` file will get updated for all changes to `vendor/`, which we definitely do not want.

/assign @nicksardo


nicksardo:
/lgtm

Thanks for adding this!

PR #930, opened 2017-08-17
Proposal for adding PVC info to VolumeStats

vkamra:
Flushes out details for part 1 of the changes described in
[#855](https://github.com/kubernetes/community/pull/855)

Feature: [#363](https://github.com/kubernetes/features/issues/363)

### DIFF
--- a/contributors/design-proposals/volume_stats_pvc_ref.md
+++ b/contributors/design-proposals/volume_stats_pvc_ref.md
@@ -41,7 +41,7 @@ type VolumeStats struct {
 - The limitation of this approach is that we're limited to reporting only what is available in the pod spec (Pod namespace and PVC claimname)
 
 ### Option 2
-- Modify the ```volumemanager::GetMountedVolumesForPod()``` (or add a new function) to return additional volume information from the ASOW/DSOW caches
+- Modify the ```volumemanager::GetMountedVolumesForPod()``` (or add a new function) to return additional volume information from the actual/desired state-of-world caches
     - Use this to populate PVCRef in VolumeStats
 
 - This allows us to get information not available in the Pod spec such as the PV name/UID which can be used to label metrics - enables exposing/querying volume metrics by PV name
--- a/contributors/design-proposals/volume_stats_pvc_ref.md
+++ b/contributors/design-proposals/volume_stats_pvc_ref.md
@@ -48,6 +48,9 @@ type VolumeStats struct {
 - It's unclear whether this is a use case we need to/should support:
   * Volume metrics are only refreshed for mounted volumes which implies a bound/available PVC
   * We expect most user-storage interactions to be via the PVC
+- Admins monitoring PVs (and not PVC's) so that they know when their users are running out of space or are over-provisioning would be a use case supporting adding PV information to
+  metrics
+
 
 
 
--- a/contributors/design-proposals/volume_stats_pvc_ref.md
+++ b/contributors/design-proposals/volume_stats_pvc_ref.md
@@ -1,0 +1,54 @@
+# Add PVC reference in Volume Stats
+
+## Background
+Pod volume stats tracked by kubelet do not currently include any information about the PVC (if the pod volume was referenced via a PVC)
+
+This prevents exposing (and querying) volume metrics labeled by PVC name which is preferable for users, given that PVC is a top-level API object.
+
+## Proposal
+
+Modify ```VolumeStats``` tracked in Kubelet and populate with PVC info:
+
+```
+// VolumeStats contains data about Volume filesystem usage.
+type VolumeStats struct {
+	// Embedded FsStats
+	FsStats
+	// Name is the name given to the Volume
+	// +optional
+	Name string `json:"name,omitempty"`
++	// PVCRef is a reference to the measured PVC.
++	// +optional
++	PVCRef PVCReference `json:"pvcRef"`
+}
+
++// PVCReference contains enough information to describe the referenced PVC.
++type PVCReference struct {
++	Name      string `json:"name"`
++	Namespace string `json:"namespace"`
++}
+```
+
+## Implementation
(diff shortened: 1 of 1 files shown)


vkamra:
@jingxu97 please review - thanks!

contributors/design-proposals/volume_stats_pvc_ref.md
```diff
@@ -0,0 +1,54 @@
...
+### Option 1
+- Modify ```kubelet::server::stats::calcAndStoreStats()```
+    - If the pod volume is referenced via a PVC, populate ```PVCRef``` in VolumeStats using the Pod spec
+
+    - The Pod spec is already available in this method, so the changes are contained to this function.
+
+- The limitation of this approach is that we're limited to reporting only what is available in the pod spec (Pod namespace and PVC claimname)
+
+### Option 2
+- Modify the ```volumemanager::GetMountedVolumesForPod()``` (or add a new function) to return additional volume information from the ASOW/DSOW caches
```
jingxu97: Could you use actual and desired state instead of ASOW/DSOW?

vkamra: @jingxu97 - addressed.

contributors/design-proposals/volume_stats_pvc_ref.md
```diff
@@ -0,0 +1,54 @@
...
+
++// PVCReference contains enough information to describe the referenced PVC.
++type PVCReference struct {
++	Name      string `json:"name"`
++	Namespace string `json:"namespace"`
++}
+```
+
+## Implementation
+2 options are described below. Option 1 supports current requirements/requested use cases. Option 2 supports an additional use case that was being discussed and is called out for completeness/discussion/feedback.
```
wongma7: Agree use case of getting metrics by PV name is iffy, it could be left as the responsibility of the higher-level tool (heapster?) to do that, since it should be trivial for them to find the PVC to which PV is bound.

If not option 2 sounds good to me, the PV name is right there in volume spec, it just needs to be plumbed through GetMountedVolumesForPod.

wongma7:
Do we care about exposing these metrics via prometheus? Is there any other case where we expose the same metrics via kubelet summary API *and* via prometheus directly, or is that unwanted duplication? I am not too familiar with the metrics stack.

contributors/design-proposals/volume_stats_pvc_ref.md
```diff
@@ -0,0 +1,54 @@
...
+- The limitation of this approach is that we're limited to reporting only what is available in the pod spec (Pod namespace and PVC claimname)
+
+### Option 2
+- Modify the ```volumemanager::GetMountedVolumesForPod()``` (or add a new function) to return additional volume information from the actual/desired state-of-world caches
+    - Use this to populate PVCRef in VolumeStats
+
+- This allows us to get information not available in the Pod spec such as the PV name/UID which can be used to label metrics - enables exposing/querying volume metrics by PV name
+- It's unclear whether this is a use case we need to/should support:
+  * Volume metrics are only refreshed for mounted volumes which implies a bound/available PVC
+  * We expect most user-storage interactions to be via the PVC
```
wongma7: One use-case to call out is admins monitoring PVs so that they know when their users are running out of space and they can immediately find the PV's backing storage to resize

contributors/design-proposals/volume_stats_pvc_ref.md
jingxu97: PV name is not available in volume spec, I think. We have to retrieve through API server to get the PVC object first to get PV name, I think?

wongma7: from asw.attachedVolumes we can check attachedVolume.spec.PersistentVolume (attachedVolume.spec is volume.Spec https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/plugins.go#L277), so option #2 is possible w/o API call

so I think we have "option #1 + API `get`" or "option #2" if we want metrics indexed by PV name. I prefer option #2

vkamra: Agreed - given that we're already populating these caches via API calls - option #2 would be preferred rather than adding API calls in the volume stats collector.

vkamra:
@wongma7 - we should expose these metrics via prometheus (/metrics API) as well. @jingxu97 details that in https://github.com/kubernetes/community/pull/855. 

This proposal (and corresponding implementation PR) just gets PVC information into the volume stats.

wongma7:
I see, thanks. I would like to see that for 1.8, do you plan to implement it too? Else I can have a try after your PR merges, though I'm not sure where to register those prometheus metrics tbh :)

vkamra:
I've looked at it but haven't had a chance to work on it for 1.8

Prometheus metrics are registered in https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/metrics/metrics.go. After registering them there, we could observe/update those in volume_stat_calculator::calcAndStoreStats().

contributors/design-proposals/volume_stats_pvc_ref.md
jingxu97: you are right, I got confused. @vkamra I think we could add a separate function to get a list of PV name indexed by outerVolumeSpecName similarly to ListVolumesForPod. Then we can add the PV name information too.

vkamra: 👍 - that's what I have in option 2. We should do that incrementally though if we need to return PV. For now, PVC seems good enough.

childsb:
This is for 1.8, correct?

@vkamra i believe the only outstanding change is @wongma7 use case mentioned then we are good to merge.

vkamra:
@childsb - yes, for 1.8. Updated with the use case - should be good to go. The corresponding implementation of this is in PR [51448](https://github.com/kubernetes/kubernetes/pull/51448) also ready to be merged

PR #824, opened 2017-07-19
sig-docs meets weekly now

zoidbergwill:
In the minutes from the meetings: https://docs.google.com/document/d/1Ds87eRiNZeXwRBEbFr6Z7ukjbTow5RQcNZLaSvWWQsE/edit#heading=h.k7ir3bpr9sl6

PR #931, opened 2017-08-17
Update leads and meeting time of the SIG-UI

bryk:


floreks:
LGTM

xiangpengzhao:
Hi @bryk, FYI: 

It'd be better only modify `sigs.yaml` manually and then run `make` to update associated docs automatically. See: https://github.com/kubernetes/community/pull/824#issuecomment-317223125

e.g. If we modify `sigs.yaml` and then run `make`, the docs `sig-foo/README.md` and `sig-list.md` will be updated.

PR #934, opened 2017-08-17
Update sig-autoscaling meeting link

mwielgus:


sigs.yaml
```diff
@@ -112,7 +112,7 @@ sigs:
     - day: Thursday
       utc: "15:30"
       frequency: biweekly/triweekly
-    meeting_url: https://plus.google.com/hangouts/_/google.com/k8s-autoscaling
+    meeting_url: https://zoom.us/meeting/176352399
```
DirectXMan12: whoops, this needs to be `/j/` instead of `/meeting/`

PR #391, opened 2017-02-22
Describe API groups and spec/status for simple kinds

smarterclayton:
Clarifies a few points folks have asked.

PR #937, opened 2017-08-17
Fix sig-ui updates

cblecker:
Sig UI updates from https://github.com/kubernetes/community/pull/931 weren't done in the right place, so autogeneration reverted them.


grodrigues3:
I can confirm that the updates to sigs.yaml is fine and you ran make correctly.

Someone from sig-ui (which doesn't appear to have gh teams set up) should confirm leadership change and lgtm

grodrigues3:
@danielromlein

cblecker:
The leadership change was in #391, but it got reverted because it was done in the wrong place

grodrigues3:
Think you meant #931.

/lgtm

cblecker:
yes, sorry :) thanks @grodrigues3 !

PR #938, opened 2017-08-17
Update SIG Apps leads

mattfarina:
There were two changes:
1. Adnan has joined as a co-lead
2. Matt is now working at Samsung SDS


mattfarina:
cc @prydonius @michelleN

grodrigues3:
@spiffxp OWNERS is just for assigning PRs for now.  LGTM can come from anyone in the org currently (same as other PRs).  

/lgtm

spiffxp:
and that's why I didn't /lgtm, figured it would be more kosher to let the SIG decide

PR #939, opened 2017-08-17
add apiserver-count fix proposal

rphillips:
This is a proposal to fix the apiserver-count issue at https://github.com/kubernetes/kubernetes/issues/22609. I would appreciate a review on the proposal.

- [x] Add ConfigMap for configurable options
- [ ] Find out dependencies on the Endpoints API and add them to the proposal

### DIFF
--- a/contributors/design-proposals/apiserver-count-fix.md
+++ b/contributors/design-proposals/apiserver-count-fix.md
@@ -22,6 +22,7 @@ stale API Endpoints. The issue is when the number of kube-apiserver instances
 gets below or above the masterCount. If the below case happens, the stale
 instances within the Endpoints does not get cleaned up, or in the latter case
 the endpoints start to flap.
+
 ## Known Issues
 
 Each apiserver’s reconciler only cleans up for it's own IP. If a new
@@ -34,11 +35,11 @@ to the masterCount. For example:
 * The IP for endpoint ‘B’ is not
 removed from the Endpoints list
 
-There is logic within the [MasterCountEndpointReconciler](https://github.com/kubernetes/kubernetes/blob/68814c0203c4b8abe59812b1093844a1f9bdac05/pkg/master/controller.go#L293) to attempt to make
-the Endpoints eventually consistent, but the code relies on the
-Endpoints count becoming equal to or greater than masterCount. When the
-apiservers become greater than the masterCount the Endpoints tend to
-flap.
+There is logic within the
+[MasterCountEndpointReconciler](https://github.com/kubernetes/kubernetes/blob/68814c0203c4b8abe59812b1093844a1f9bdac05/pkg/master/controller.go#L293)
+to attempt to make the Endpoints eventually consistent, but the code relies on
+the Endpoints count becoming equal to or greater than masterCount. When the
+apiservers become greater than the masterCount the Endpoints tend to flap.
 
 If the number endpoints were scaled down from automation, then the
 Endpoints would never become consistent.
@@ -49,75 +50,27 @@ Endpoints would never become consistent.
 
 | Kubernetes Release  | Quality | Description |
 | ------------- | ------------- | ----------- |
-| 1.9           | alpha         | <ul><li>Add a new reconciler</li><li>Add a command-line type `--alpha-apiserver-endpoint-reconciler-type`<ul><li>configmap</li><li>default</li></ul></li></ul>
-| 1.10          | beta          | <ul><li>Turn on the `configmap` type by default</li></ul>
+| 1.9           | alpha         | <ul><li>Add a new reconciler</li><li>Add a command-line type `--alpha-apiserver-endpoint-reconciler-type`<ul><li>storage</li><li>default</li></ul></li></ul>
+| 1.10          | beta          | <ul><li>Turn on the `storage` type by default</li></ul>
 | 1.11          | stable        | <ul><li>Remove code for old reconciler</li><li>Remove --apiserver-count</li></ul>
 
-The MasterCountEndpointReconciler does not meet the current needs for
-durability of API Endpoint creation, deletion, or failure cases.
+The MasterCountEndpointReconciler does not meet the current needs for durability
+of API Endpoint creation, deletion, or failure cases.
 
-Create a new `MasterEndpointConfigMapReconciler` within
-master/controller.go.
+Custom Resource Definitions and ConfigMaps were proposed, but since they are
+watched globally, liveliness updates could be overly chatty.
 
-Add a `kube-apiserver-config` ConfigMap in the `kube-system`
-namespace. The duration found within the Coordination map would be configurable by
-admins without a recompile. The ConfigMap would include the following:
-
-```go
-KubeApiServerConfigMap{
-  "expiration-duration":        "1m",
-  ... [other flags potentially]
-}
-```
-
-The Coordination ConfigMap would be formed such that:
-
-```go
-CoordinationConfigMap{
-  "[namespace]/[name]": "[KubeAPIServerEndpoint]",
-}
-```
-
-*Key Format*: ip-[IP String Formatted]-[port]
-
-```go
-KubeAPIServerEndpointConfigMap{
-  "master-count":                "0",
-	"ip-2001-4860-4860--8888-443": "time.Time",
-	"ip-192-168-0-3-443":          "time.Time",
-}
-```
-
-***TODO: should it a serialized struct, or the raw time value?***
-
-The reconcile loop will expire endpoints that do not meet the duration.
-On each reconcile loop (the loop runs every 10 seconds currently,
-but interval will be changed to 80% of the `expiration-duration`):
-
-1. GET `kube-apiserver-endpoints` ConfigMap (as endpointMap)
-1. Update the timestamp for the currently running API server
-1. Remove all endpoints where the timestamp is greater than `expiration-duration` from the configMap.
-1. Do a GET on the `kube-apiserver-endpoints`
-1. Construct a PATCH operation with the deleted endpoints
-
-configmap.yml:
-
-```yaml
-kind: ConfigMap
-apiVersion: v1
-metadata:
-  name: kube-apiserver-endpoints
-  namespace: default
-```
+By porting OpenShift's
+[LeaseEndpointReconciler](https://github.com/openshift/origin/blob/master/pkg/cmd/server/election/lease_endpoint_reconciler.go)
+to Kubernetes we can use use the Storage API directly to store Endpoints
+dynamically within the system.
 
 ### Alternate Proposals
 
-#### Custom Resource Definitions
+#### Custom Resource Definitions and ConfigMaps
 
-CRD's were considered for this proposal, but were not proposed due to
-constraints of having CRDs within core has not been clearly defined.
-Layering could also be an issue, so the proposal defined ConfigMaps as
-the current path forward.
+CRD's and ConfigMaps were considered for this proposal. They were not adopted
+for this proposal by the community due to tecnical issues explained earlier.
 
 #### Refactor Old Reconciler
 
--- a/contributors/design-proposals/apiserver-count-fix.md
+++ b/contributors/design-proposals/apiserver-count-fix.md
@@ -13,33 +13,34 @@ Authors: @rphillips
 
 Proposal to fix Issue [#22609](https://github.com/kubernetes/kubernetes/issues/22609)
 
-`kube-apiserver` currently has a command-line argument `--apiserver-count`
-specifying the number of api masters. This masterCount is used in the
-MasterCountEndpointReconciler on a 10 second interval to potentially cleanup
-stale API Endpoints. The issue is when the number of kube-apiserver instances
-gets below masterCount. If this case happens, the stale instances within the
-Endpoints does not get cleaned up.
+`kube-apiserver` currently has a command-line argument
+`--apiserver-count` specifying the number of api masters. This
+masterCount is used in the MasterCountEndpointReconciler on a 10 second
+interval to potentially cleanup stale API Endpoints. The issue is when
+the number of kube-apiserver instances gets below masterCount. If this
+case happens, the stale instances within the Endpoints does not get
+cleaned up.
 
 ## Known Issues
 
-Each apiserver’s reconciler only cleans up for it's own IP. If a new server
-is spun up at a new IP, then the old IP in the Endpoints list is only
-reclaimed if the number of apiservers becomes greater-than or equal to the
-masterCount. For example:
+Each apiserver’s reconciler only cleans up for it's own IP. If a new
+server is spun up at a new IP, then the old IP in the Endpoints list is
+only reclaimed if the number of apiservers becomes greater-than or equal
+to the masterCount. For example:
 
-* If the masterCount = 3, and there are 3 API servers running (named: A, B, and
-C) 
+* If the masterCount = 3, and there are 3 API servers running (named: A, B, and C) 
 * ‘B’ API server is terminated for any reason
 * The IP for endpoint ‘B’ is not
 removed from the Endpoints list
 
 There is logic within the [MasterCountEndpointReconciler](https://github.com/kubernetes/kubernetes/blob/68814c0203c4b8abe59812b1093844a1f9bdac05/pkg/master/controller.go#L293) to attempt to make
-the Endpoints eventually consistent, but the code relies on the Endpoints
-count becoming equal to or greater than masterCount. When the apiservers
-become greater than the masterCount the Endpoints tend to flap.
+the Endpoints eventually consistent, but the code relies on the
+Endpoints count becoming equal to or greater than masterCount. When the
+apiservers become greater than the masterCount the Endpoints tend to
+flap.
 
-If the number endpoints were scaled down from automation, then the Endpoints
-would never become consistent. 
+If the number endpoints were scaled down from automation, then the
+Endpoints would never become consistent.
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


hongchaodeng:
Shouldn't the commit msg and title be "add apiserver-count fix proposal"? Mind the "-".

ericchiang:
cc @kubernetes/sig-api-machinery-proposals

luxas:
cc @kubernetes/sig-cluster-lifecycle-feature-requests

lpabon:
Hi @rphillips thanks for putting this together. Could you add the following to the documentation:

* Configurable values for timeouts or other magic numbers
* Explanation in the document on how clients consume this information

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,98 @@
...
+the Endpoints eventually consistent, but the code relies on the Endpoints
+count becoming equal to or greater than masterCount. When the apiservers
+become greater than the masterCount the Endpoints tend to flap.
+
+If the number endpoints were scaled down from automation, then the Endpoints
+would never become consistent. 
+
+## Proposal
+
+### Create New Reconciler
```
mikedanese: This design says "create new reconciler" and has a struct definition but doesn't say what the reconciler does or how the struct is used.

rphillips: Thank you @mikedanese. Added a stanza in commit 81b6c58

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+
+| Kubernetes Release  | Quality | Description |
+| ------------- | ------------- | ----------- |
+| 1.9           | alpha         | <ul><li>Add a new reconciler</li><li>Add a command-line switch --new-reconciler</li><li>Add a command-line switch --old-reconciler</li></ul>
```
smarterclayton: We generally don't add flags like this, it would usually be a type flag with an experimental prefix for alpha.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+
+Add a `kube-apiserver-endpoints-config` ConfigMap in the `default` namespace. The duration found within the map would be configurable by admins without a recompile. The ConfigMap would include the following:
+
+```go
+ConfigMap{
+	"expire-duration": "1m", // golang duration (ns,us,ms,s,m,h)
+}
+```
+
+Add a standard `kube-apiserver-endpoints` ConfigMap in the `default` namespace. The ConfigMap would be formed such that:
```
smarterclayton: The last proposed fix discussed adding an extension resource via crd instead of using config map.  Add a section discussing tradeoffs for that here

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+| ------------- | ------------- | ----------- |
+| 1.9           | alpha         | <ul><li>Add a new reconciler</li><li>Add a command-line switch --new-reconciler</li><li>Add a command-line switch --old-reconciler</li></ul>
+| 1.10          | beta          | <ul><li>Turn on the new reconciler by default |</li></ul>
+| 1.11          | stable        | <ul><li>Remove code for old reconciler</li><li>Remove --old-reconciler</li><li>Remove --new-reconciler</li><li>Remove --apiserver-count</li></ul>
+
+The MasterCountEndpointReconciler does not meet the current needs for durability of API Endpoint creation, deletion, or failure cases.
+
+Create a new `MasterEndpointReconciler` within master/controller.go.
+
+Add a `kube-apiserver-endpoints-config` ConfigMap in the `default` namespace. The duration found within the map would be configurable by admins without a recompile. The ConfigMap would include the following:
```
lavalamp: The namespace and configmap name should probably be flags :/

I would at least default to the `kube-system` namespace.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+}
+```
+
+Add a standard `kube-apiserver-endpoints` ConfigMap in the `default` namespace. The ConfigMap would be formed such that: 
+
+*Key Format*: ip-[IP String Formatted]-[port]
+
+```go
+ConfigMap{
+	"ip-2001-4860-4860--8888-443": "serialized JSON ControllerEndpointData",
```
lavalamp: Does anything other than a time need to be in the value? I'm not sure what benefit we get from cramming all the other stuff in there?

contributors/design-proposals/apiserver-count-fix.md
lavalamp: Actually the name of this should probably be `kube-apiserver-config`; it should grow to include everything that is currently a flag. Endpoints don't need a special config thingy.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+| 1.11          | stable        | <ul><li>Remove code for old reconciler</li><li>Remove --old-reconciler</li><li>Remove --new-reconciler</li><li>Remove --apiserver-count</li></ul>
+
+The MasterCountEndpointReconciler does not meet the current needs for durability of API Endpoint creation, deletion, or failure cases.
+
+Create a new `MasterEndpointReconciler` within master/controller.go.
+
+Add a `kube-apiserver-endpoints-config` ConfigMap in the `default` namespace. The duration found within the map would be configurable by admins without a recompile. The ConfigMap would include the following:
+
+```go
+ConfigMap{
```
lavalamp: The name and namespace of the endpoint coordination config map should be stored in this configmap rather than be hard-coded.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+     api.EndpointPort
+     CreationTimestamp Time `json:"creationTimestamp,omitempty" protobuf:"bytes,8,opt,name=creationTimestamp"`
+     UpdateTimestamp   Time `json:"creationTimestamp,omitempty" protobuf:"bytes,8,opt,name=updateTimestamp"`
+}
+```
+
+The reconcile loop will expire endpoints that do not meet the duration. On
+each reconcile loop (the loop runs every 10 seconds currently):
+
+1. Retrieve `kube-apiserver-endpoints-config` ConfigMap (as configMap)
```
lavalamp: The config probably doesn't need to be gotten all that frequently, although it doesn't hurt anything, either.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+1. Retrieve `kube-apiserver-endpoints` ConfigMap (as endpointMap)
+1. Update the `UpdateTimestamp` for the currently running API server
```
lavalamp: Can this be done via a PATCH?

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+1. Remove all endpoints where the UpdateTimestamp is greater than `expire-duration` from the configMap.
```
lavalamp: I would do this in two requests rather than a single one, to avoid conflicts when multiple apiservers delete a record. Actually I'm not 100% sure it's a conflict if two requests delete the same thing, would a PATCH like that apply cleanly or give an error?

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+     metav1.TypeMeta   `json:",inline"`
+     metav1.ObjectMeta `json:",inline"`
+     api.EndpointPort
+     CreationTimestamp Time `json:"creationTimestamp,omitempty" protobuf:"bytes,8,opt,name=creationTimestamp"`
+     UpdateTimestamp   Time `json:"creationTimestamp,omitempty" protobuf:"bytes,8,opt,name=updateTimestamp"`
+}
+```
+
+The reconcile loop will expire endpoints that do not meet the duration. On
+each reconcile loop (the loop runs every 10 seconds currently):
```
lavalamp: Instead of a fixed 10s, this should probably be something like 50% to 80% of the expiration time, whatever that is.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+  namespace: default
+```
+
+### Refactor Old Reconciler
+
+| Kubernetes Release | Quality | Description |
+| ------------- | ------------- | ----------- |
+| 1.9           | stable  | <ul><li>Change the logic in the current reconciler</li></ul>
+
+We could potentially reuse the old reconciler, but ignore the masterCount and change the logic to use the proposal from the previous section.
```
lavalamp: A first step might just be to count the number of entries in the config map and change the masterCount variable to that every time through the above loop (probably need to add a lock). That would probably be just a few line change.

contributors/design-proposals/apiserver-count-fix.md
lavalamp: I think someone may have been working on moving flags to a config map, you may want to coordinate with them. @mikedanese do you recall who was working on that?

rphillips:
Thank you all for the feedback! I'll add everything in.

fisherxu:
Have little doubts, why can‘t we add label on kube-apiserver's pod and through the selector in service sync the endpoint...

lavalamp:
> Have little doubts, why can‘t we add label on kube-apiserver's pod and through the selector in service sync the endpoint...

Yeah, this would definitely be the way to go if all installations of Kubernetes were self-hosted. Unfortunately that is not the case.

contributors/design-proposals/apiserver-count-fix.md
luxas: Wasn't that for the full-blown APIServer status registration or do I recall incorrectly?
Anyway, I think that a ConfigMap is good here.
We've discussed layer violations of this kind before and it seems like that would be one.

contributors/design-proposals/apiserver-count-fix.md
luxas: Kind of agree; but at the same time we don't know what the future holds for us :)

contributors/design-proposals/apiserver-count-fix.md
luxas: Is this GET on every reconcile loop or WATCH? Retrieve could mean both I guess

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+
+```yaml
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: kube-apiserver-endpoints
+  namespace: default
+```
+
+### Refactor Old Reconciler
```
luxas: nit: Can you make it more obvious that these are the two current viable ways to do this? (and add one for the CRD thingy??)
like: `Possible solution 1: foo`, `Possible solution 2: bar`

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
...
+
+| Kubernetes Release | Quality | Description |
+| ------------- | ------------- | ----------- |
+| 1.9           | stable  | <ul><li>Change the logic in the current reconciler</li></ul>
+
+We could potentially reuse the old reconciler, but ignore the masterCount and change the logic to use the proposal from the previous section.
+
+## Prior Art
+
+[Security Labeller](https://github.com/coreos-inc/security-labeller/issues/18#issuecomment-320791878)
```
luxas: 404 for me

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,114 @@
+# apiserver-count fix proposal
+
+Authors: @rphillips
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Known Issues](#known-issues)
+3. [Proposal](#proposal)
```
luxas: Add sub-section for different strategies?

rphillips:
I've incorporated all the changes thus far.

There is still a pending question on what the payload is for  `KubeAPIServerEndpointConfigMap`, whether it should a string or a serialized structure. I have changed the field to be a serialization of time.Time, but I could swayed either way.

contributors/design-proposals/apiserver-count-fix.md
smarterclayton: I don't think it would be a layer violation, because this is the core api server (it's allowed to depend on apis).  We've already discussed it several times at an arch level.

lavalamp:
I think some sort of config has to be in the nucleus. The nucleus needs to
be configured.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+### Alternate Proposals
+
+#### Custom Resource Definitions
+
+CRD's were considered for this proposal, but were not proposed due to
+constraints of having CRDs within core has not been clearly defined.
+Layering could also be an issue, so the proposal defined ConfigMaps as
+the current path forward.
```
smarterclayton: @bgrant0607 when we discussed this before on whether we should be abusing configmaps for election we discussed a CRD as being acceptable for use within the api server.  

I think my preference at this point is not to abuse configmaps for high volume write traffic that is not related directly to the purpose of delivering configuration into a pod.  If we do not want to use the previously proposed mechanism of using storage with TTL (which I still think is the correct approach since it is how everyone else using etcd deals with leases) then we need to make a case that it is ok for a single component (the core kube-apiservers) to constantly write to a configmap, but that other components should not do so / are not affected by it.

lavalamp:
I almost suggested directly using etcd from apiserver, but I didn't because
I would like us to solve this problem in the way any user could. I don't
like having multiple mechanisms for solving the same problem.

Also etcd3's TTL is not as convenient as etcd2's was.

Do we really consider writing configmaps frequently as abuse? We recommend
using them for leader election, I thought. I don't see why we wouldn't
allow / want this use of a configmap.

​

smarterclayton:
I don't think we *want* users using configmap for election.  ConfigMap as a high write volume target that some components may ListAll/WatchAll gives me concern (garbage collector today, config map triggers in the future).  I'd rather have a resource that *no* one ever watches.  ConfigMap could easily be O(applications), which means any dense cluster SLO we have will potentially have to take into account O(applications) = 10k.  That makes configmap somewhat fraught.

The reason Brian and I discussed a CRD or new resources for this is because

1. CRD has to work
2. CRD is in core-isa
3. CRD can be arbitrarily sliced and ACLd

I don't want my apps able to write their own configmap (for security reasons).  I don't want to apply specific ACL magic rules to service accounts (all pods can update config maps that are linked in their pod), but we *can* do that for a CRD or core resource.

I agree with the sentiment of setting up a general mechanism for users to do this.  I do not agree with doing it on a general purpose like ConfigMap with more thought.  I prefer using something under the api layer (if this is api server specific) or something purpose built for frequent updates from pods with ACL implications taken into account (if this is not api server specific)

smarterclayton:
@kubernetes/sig-auth-feature-requests re: "pods that are able to write to config maps without ACL getting ugly" - we may be able to do this with RBAC, but if my pod can edit config maps I'd probably say by default that config map **should not** have other config in it.  At that point, it's not really a config map, it's "pod scratch coordination space".

lavalamp:
I agree re: configmaps for coordination should NOT also have other config
in them. This proposal has two config maps.

I will think about the other stuff you said and see if I can come up with
another alternative.

xiang90:
@lavalamp 

> I almost suggested directly using etcd from apiserver, but I didn't because
I would like us to solve this problem in the way any user could. I don't
like having multiple mechanisms for solving the same problem.

user already could do this today for any applications running inside k8s as pods. k8s controller will update endpoints according to readiness.

we should not mix this proposal with leader election, since it is really not. it is a time-based liveness status for the "out-of-cluster" components to update endpoint. 


@smarterclayton 

>  I prefer using something under the api layer (if this is api server specific) or something purpose built for frequent updates from pods with ACL implications taken into account (if this is not api server specific)

Are you suggesting to enable this directly in the storage layer?

smarterclayton:
The storage layer can already be used for this -
https://github.com/openshift/origin/blob/master/pkg/cmd/server/election/lease_endpoint_reconciler.go

In talking with Xiang we also brought up another point - if we don't know
*exactly* the use case we want to solve for end users, we shouldn't half
create an API pattern here.  We should either design the pattern that
solves this generically, or specifically solve it under the api server.  I
lean towards us not having enough time to solve this generically, which is
why I still think that the kube-apiserver maintaining the minimal lease
based reconciler above is the right solution for the right now.

smarterclayton:
(also, the code linked above has been used in production for about a year on hundreds of clusters and we haven't had a problem yet, which makes me want to avoid solving a new problem)

contributors/design-proposals/apiserver-count-fix.md
bgrant0607: +1 to not abusing ConfigMap.

Extension mechanisms, such as CRD, will be part of the Nucleus.

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -49,75 +50,27 @@ Endpoints would never become consistent.
...
-1. GET `kube-apiserver-endpoints` ConfigMap (as endpointMap)
-1. Update the timestamp for the currently running API server
-1. Remove all endpoints where the timestamp is greater than `expiration-duration` from the configMap.
-1. Do a GET on the `kube-apiserver-endpoints`
-1. Construct a PATCH operation with the deleted endpoints
+The MasterCountEndpointReconciler does not meet the current needs for durability
+of API Endpoint creation, deletion, or failure cases.
 
-configmap.yml:
+Custom Resource Definitions and ConfigMaps were proposed, but since they are
```
xiang90: I can understand ConfigMap has the watching issue. But CRD should not be watched globally. The problem with CRD , as far as I can tell,  is the not super clean layering, and potentially we might want something different that can handle both liveness and locking as a k8s native resource. And we are not sure yet.

rphillips:
@smarterclayton @lavalamp updated

I have the LeaseEndpointReconciler partially ported (clean build), but more work to do on it.

rphillips:
Is fixing kube-proxy is scope of this issue and eventual PR? or should it be a separate effort?

contributors/design-proposals/apiserver-count-fix.md
```diff
@@ -0,0 +1,86 @@
...
+server is spun up at a new IP, then the old IP in the Endpoints list is
+only reclaimed if the number of apiservers becomes greater-than or equal
+to the masterCount. For example:
+
+* If the masterCount = 3, and there are 3 API servers running (named: A, B, and C) 
+* ‘B’ API server is terminated for any reason
+* The IP for endpoint ‘B’ is not
+removed from the Endpoints list
+
+There is logic within the
```
bgrant0607: I assume aggregated apiservers won't do something similar?

liggitt: aggregated servers can either delegate to service dns names (and therefore kube-proxy) or route to an IP selected from the endpoints for the service (and therefore take advantage of whatever mechanism removes unhealthy IP addresses from those endpoints)

lavalamp:
Aggregated API servers use the service concept to present themselves to the
system, so there's no need for them to do anything special here. (Yes,
perhaps technically it should just be endpoints, but the service concept
works and is easy.)

(We should make the main apiserver work the same way eventually, IMO)

rphillips:
I have submitted a PR for this issue.

klausenbusk:
> Yeah, this would definitely be the way to go if all installations of Kubernetes were self-hosted. Unfortunately that is not the case.

Just a question. Is there anyway to disable the default reconciler? So I can try with `selector` already now.

lavalamp:
That's a great point, we should make the default behavior flag-disableable
so that it's possible to experiment with fully self hosted configurations.

rphillips:
@lavalamp @klausenbusk  I added a 'none' reconciler to the PR to effectively be a noop.

lavalamp:
/approve
/lgtm

At the SIG meeting yesterday we collectively decided that using the storage interface directly was the least bad way of implementing this.

Note that other apiservers won't have this problem (they can use services & endpoints), it is unique to kube-apiserver.

luxas:
@lavalamp what "storage interface" are you regerring to in this context?

This will still use Endpoints, but in an other way, right?

lavalamp:
The storage.Interface go interface is what apiserver uses to talk to etcd.

PR #945, opened 2017-08-18
Admission control to limit events accepted by an API Server

staebler:
This is a proposal for adding an admission control that allows the user to limit the rate of events accepted by an API Server. The limits can be set per-server, per-namespace, per-user, and per-source+object.

### DIFF
--- a/contributors/design-proposals/admission_control_event_rate_limit.md
+++ b/contributors/design-proposals/admission_control_event_rate_limit.md
@@ -1,0 +1,141 @@
+# Admission control plugin: EventRateLimit
+
+## Background
+
+This document proposes a system for using admission control to enforce a limit
+on the number of event requests that the API Server will accept in a given time
+slice. In a large multi-tenant cluster, there may be a small percentage of
+tenants that are always in some type of error state, in which each tenant is
+producing a steady stream of error event requests. Each individual tenant may not
+be producing an absurd amount of event requests on its own, but taken collectively
+the errors from this small percentage of tenants can have a significant impact on
+the performance of the cluster overall. 
+
+## Use cases
+
+1. Ability to protect the API Server from being flooded by event requests.
+2. Ability to protect the API Server from being flooded by event requests for
+   a particular namespace.
+3. Ability to protect the API Server from being flooded by event requests for
+   a particular user.
+4. Ability to protect the API Server from being flooded by event requests from
+   a particular source+object.
+
+## Data Model
+
+### Configuration
+
+```go
+// Configuration provides configuration for the EventRateLimit admission controller.
+type Configuration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Limits to place on events received.
+	// Limits can be placed on events received server-wide, per namespace, and
+	// per source+object.
+	// At least one limit is required.
+	Limits []Limit `json:"limits"`
+}
+
+// Limit provides configuration for a single type of limit
+type Limit struct {
+	// Type of limit.
+	// The following are valid values.
+	// "server": limits are maintained against all events received by the server
+	// "namespace": limits are maintained against events from each namespace
+	// "source+object": limits are maintained against events from each source+object
+	Type string `json:"type"`
+
+	// Maximum QPS of events for this limit
+	QPS float32 `json:"qps"`
+
+	// Maximum burst for throttle of events for this limit
+	Burst int ` json:"burst"`
+
+	// Maximum number of limits to maintain. If room is needed in the cache for a
+	// new limit, then the least-recently used limit is evicted, resetting the
+	// stats for that subset of the universe.
+	//
+	// For example, if the type of limit is "namespace" and the limit for
+	// namespace "A" is evicted, then the next event received from namespace "A"
+	// will use reset stats, enabling events from namespace "A" as though no
+	// events from namespace "A" have yet been received.
+	//
+	// If the type of limit is "server", then CacheSize is ignored and can be
+	// omitted.
+	CacheSize int `json:"cacheSize"`
+}
+```
+
+### Validation
+
+Validation of a **Configuration** enforces that the following rules apply:
+
+* There is at least one item in **Limits**.
+* Each item in **Limits** has a unique **Type**.
+
+Validation of a **Limit** enforces that the following rules apply:
+
+* **Type** is one of "server", "namespace", "user", and "source+object".
+* **QPS** is positive.
+* **Burst** is positive.
+* If **Type** is not "server", then **CacheSize** is positive.
+
+### Default Value Behavior
+
+If there is no item in **Limits** for a particular limit type, then no limits
+will be enforced for that type of limit.
+
+## AdmissionControl plugin: EventRateLimit
+
+The **EventRateLimit** plug-in introspects all incoming event requests and
+determines whether the event fits within the rate limits configured.
+
+To enable the plug-in and support for EventRateLimit, the kube-apiserver must
+be configured as follows:
+
+```console
+$ kube-apiserver --admission-control=EventRateLimit --admission-control-config-file=$ADMISSION_CONTROL_CONFIG_FILE
+```
+
+## Example
+
+An example EventRateLimit configuration:
+
+| Type | RequestBurst | RequestRefillRate | CacheSize |
+| ---- | ------------ | ----------------- | --------- |
+| Server | 1000 | 100 | |
+| Namespace | 100 | 10 | 50 |
+
+The API Server starts with an allowance to accept 1000 event requests. Each
+event request received counts against that allowance. The API Server refills
+the allowance at a rate of 100 per second, up to a maximum allowance of 1000.
+If the allowance is exhausted, then the API Server will respond to subsequent
+event requests with 429 Too Many Requests, until the API Server adds more to
+its allowance.
+
+For example, let us say that at time t the API Server has a full allowance to
+accept 1000 event requests. At time t, the API Server receives 1500 event
+requests. The first 1000 to be handled are accepted. The last 500 are rejected
+with a 429 response. At time t + 1 second, the API Server has refilled its
+allowance with 100 tokens. At time t + 1 second, the API Server receives
+another 500 event requests. The first 100 to be handled are accepted. The last
+400 are rejected.
+
+The API Server also starts with an allowance to accept 100 event requests from
+each namespace. This allowance works in conjunction with the server-wide
+allowance. An accepted event request will count against both the server-side
+allowance and the per-namespace allowance. The API Server keeps an allowance
+for at most 50 namespaces. The API Server will evict the allowance for the
+least-recently used nameapce if more than event requests from more than 50
+namespaces are received. If an event request for namespace N is received after
+the API Server has evicted the allowance for namespace N, then a new, full
+allowance will be created for namespace N.
+
+In this example, the API Server will not maintain any allowances for the
+user nor source+object of the event in an event request because both the
+user and the source+object details have been omitted from the configuration.
+The allowance and eviction for per-user and per-source+object rate limiting
+works identically to the per-namespace rate limiting, with the exception that
+the former consider the user or source+object of the event and the latter
+considers the namespace of the event request.
--- a/contributors/design-proposals/admission_control_event_rate_limit.md
+++ b/contributors/design-proposals/admission_control_event_rate_limit.md
@@ -139,3 +139,11 @@ The allowance and eviction for per-user and per-source+object rate limiting
 works identically to the per-namespace rate limiting, with the exception that
 the former consider the user or source+object of the event and the latter
 considers the namespace of the event request.
+
+## Client Behavior
+
+Currently, the Client event recorder treats a 429 response as http transport
+type of error, that warrants retrying the event request. Instead, the event
+recorder should abandon the event. Additionally, the event recorder should
+abandon all future events for the period of time specified in the
+retryAfterSeconds field of the 429 response.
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
@gmarek @bgrant0607 @jbeda -- this an example of a more granular panic button.

@staebler -- please describe what the event recorder will do today in response to a 429 for clarity w/ future readers.

davidopp:
I very much think we need this in general (it has been on my TODO-some-day list for a while); here are a few comments
* Are we sure that admission control is "early" enough in the API server pipeline to do effective rate limiting?
* I suggest not calling these things "events," to avoid confusion with the Event API resource
* You say "the limits can be set per-server, per-namespace, per-user, and per-source+object" but unless I'm missing something (which is entirely possible), the object you propose to specify it does not allow you to specify the last two?

staebler:
@davidopp 
> I suggest not calling these things "events," to avoid confusion with the Event API resource
The term "events" is used because the proposed admission control only deals with events from the Event API.
> Are we sure that admission control is "early" enough in the API server pipeline to do effective rate limiting?
In the real-world scenario that we have experienced, the system broke down because the etcd was struggling with the quantity of event patches that were being received. The proposed admission control would give the user the ability to limit that quantity. The API Server would still pay some of the processing costs, but this event processing does not significantly impact the performance of the API Server under our scenario.
> You say "the limits can be set per-server, per-namespace, per-user, and per-source+object" but unless I'm missing something (which is entirely possible), the object you propose to specify it does not allow you to specify the last two?
Per-user and per-source+object can be specified by adding a Limit to Configuration.Limits with a Type of "user" and "source+object", respectively. The example at the end of the only included a limit for server and namespaces. User and source+object filtering would work similarly to namespace filtering.

davidopp:
> The term "events" is used because the proposed admission control only deals with events from the Event API.

Oops, thanks for the clarification, I was confused. 

> Per-user and per-source+object can be specified by adding a Limit to Configuration.Limits with a Type of "user" and "source+object", respectively

Ah sorry, I thought you meant "per specific user(s) that is named in the config" and "per specific source(s)+object(s) that are named in the config" but you meant "per every user" and "per every source+object"

So now I understand this isn't the thing that has been on my TODO list. :) I was thinking of something that could rate limit users to enforce fair sharing of the control plane resources among users, where "fair" doesn't necessarily mean "each one is entitled to 1/N when there are N tenants trying to use the control plane" but means "according to some specified policy." I think such a system would subsume what you're doing here (but would be more complicated), so my one concern is that we might eventually replace your admission controller with something else. But since this is an admission controller, that shouldn't be hard to do, so I don't have any objection to this

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,149 @@
+# Admission control plugin: EventRateLimit
+
+## Background
+
+This document proposes a system for using admission control to enforce a limit
+on the number of event requests that the API Server will accept in a given time
+slice. In a large multi-tenant cluster, there may be a small percentage of
+tenants that are always in some type of error state, in which each tenant is
```
deads2k: Use something other than tenant please.  Do you mean namespace, controller, kubelet, client-identity, something else?

gmarek: I believe 'tenant' here is fine, as we already use 'multi-tenancy' phrase.

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,149 @@
...
+the errors from this small percentage of tenants can have a significant impact on
+the performance of the cluster overall. 
+
+## Use cases
+
+1. Ability to protect the API Server from being flooded by event requests.
+2. Ability to protect the API Server from being flooded by event requests for
+   a particular namespace.
+3. Ability to protect the API Server from being flooded by event requests for
+   a particular user.
```
gmarek: AFAIK user is not a kubernetes concept. cc @deads2k

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,149 @@
...
+	Limits []Limit `json:"limits"`
+}
+
+type Limit struct {
+	// Type of limit.
+	// The following are valid values.
+	// "server": limits are maintained against all events received by the server
+	// "namespace": limits are maintained against events from each namespace
+	// "user": limits are maintained against events from each user
+	// "source+object": limits are maintained against events from each source+object
```
gmarek: I'd do `LimitType` type being a string alias and define constants. This would make the code more self-describing.

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,149 @@
...
+	// stats for that subset of the universe.
+	//
+	// For example, if the type of limit is "namespace" and the limit for
+	// namespace "A" is evicted, then the next event received from namespace "A"
+	// will use reset stats, enabling events from namespace "A" as though no
+	// events from namespace "A" have yet been received.
+	//
+	// If the type of limit is "server", then CacheSize is ignored and can be
+	// omitted.
+	CacheSize int `json:"cacheSize"`
```
gmarek: Comment is hard to parse for me (though I think I finally understood it)

Is this really needed? What kind of cache it is? LRU? Generally it makes the semantics rather strange. E.g. when you have limit, say 5, and 7 namespaces in the system and you get Events from all namespaces regularly you'll end up with small number of used limit for all namespaces. I think if we want to implement this, we need to store information about all "buckets". Source+object is most problematic in this case, but this can be handled client-side, so it should be fine to remove it.

contributors/design-proposals/admission_control_event_rate_limit.md
deads2k: > AFAIK user is not a kubernetes concept. cc @deads2k

Users are a concept, post-authentication in the admission and REST handling chain, so I think this is correct.

contributors/design-proposals/admission_control_event_rate_limit.md
deads2k: > I think if we want to implement this, we need to store information about all "buckets". Source+object is most problematic in this case, but this can be handled client-side, so it should be fine to remove it.

Turns out you don't really need to store it to have a big impact.  Since it ramps up fairly quickly, keeping the information locally, per-apiserver, is enough to prevent crushing an apiserver.

gmarek: Sorry, I wrote it in a confusing way. I don't mean storing in etcd or anything, I mean by "storing" in cache. I.e. limited number of entries in cache is problematic in this case IMO.

staebler: The cache is used to place an upper-bound on the memory usage of the admission controller. The cache size can be set large enough to fit accommodate all of the "buckets" that exist when the system is in a steady state.

For example, if your system has 7 namespaces, you would use a cache size for "namespace" that is at least 7. If you had the admission controller configured poorly, so that the cache size for "namespace" is only 5, the admission controller is still likely to serve its purpose, assuming that you don't have more than 5 namespaces behaving poorly, where poorly behaving means sending too many events to the API Server. The poorly behaving namespaces are least likely to be evicted from the cache, by the very nature of their poor behavior.

If we did not use a cache, then a bad actor could bring down the API Server over time being simply creating and deleting namespaces. The downside of having unbounded memory usage seems to me to be far greater than occasionally having a "bucket" be granted a greater number of event requests than it should.

gmarek: But we can watch namespaces and delete non-existing ones ourselves, right? If you expect to have a bucket per "class" then I think there's even less reason to limit size of the cache (as you assume it's practically unbounded). This would add a place when cluster admin can configure something wrong, and given the number of kubernetes clusters out there someone will and this is not something I'd want to debug.

staebler: Pardon my ignorance as I am somewhat new to kubernetes. What does "class" mean here? I am having a hard time following the logic of why assuming that something is practically unbounded would make you less inclined to set up a cache for it. I would think that the expectation that something is unbounded would be exactly the reason why you would want/need to place limits on it with something like a cache. Are you suggesting that we could watch and delete buckets for source+object and user as well, and that that would be a practically way of limiting those?

I am also struggling to understand what you expect the catastrophe to be of a poorly configured cache size. If the cache size is configured too large, then that is no worse than having no cache. If the cache size if configured too small, then the admission control may allow some more events through than it should. Is there some other thing that may happen that I am not seeing? This admission control is a last effort to protect the API Server from being overwhelmed by events, but is not the only protection in place. There is already event rate limiting client-side, assuming that the client is using the kubernetes event recorder.

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,155 @@
...
+allowance with 100 tokens. At time t + 1 second, the API Server receives
+another 500 event requests. The first 100 to be handled are accepted. The last
+400 are rejected.
+
+The API Server also starts with an allowance to accept 100 event requests from
+each namespace. This allowance works in conjunction with the server-wide
+allowance. An accepted event request will count against both the server-side
+allowance and the per-namespace allowance. The API Server keeps an allowance
+for at most 50 namespaces. The API Server will evict the allowance for the
+least-recently used nameapce if more than event requests from more than 50
```
eparis: s/nameapce/namespace/

I'm not sure what `...if more than event requests from more than...` means. Did it mean `...if more than 1 request from more than...`

contributors/design-proposals/admission_control_event_rate_limit.md
gmarek: By class I meant abstraction class (e.g. all events from the same namespace). I wasn't clear what I meant. Cache itself is a good idea. I don't like it being bounded (hence the need for this field).

As I wrote I believe source+object limiting can be done client side (as source is defined by client), so I think we can just drop it here.

Catastrophe is a big word - I expect people to be confused. They set up admission controller to rate limit their Event influx, and probably use default values for cache sizes because that's what you do when you start. Then they forget about this setting and everything works until it doesn't. They'll see that more Events getting into the API server that it should and they'll try to figure out why, which I don't expect to be easy.

Generally I think that bounding the cache is causing more harm than good.

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,155 @@
...
+	// At least one limit is required.
+	Limits []Limit `json:"limits"`
+}
+
+type Limit struct {
+	// Type of limit
+	Type LimitType `json:"type"`
+
+	// Maximum QPS of events for this limit
+	QPS float32 `json:"qps"`
```
bgrant0607: No floats, please.

bgrant0607: I realize this is configuration, but it should follow the conventions:
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#primitive-types

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,155 @@
...
+	// NamespaceLimitType limits are maintained against events from each namespace
+	NamespaceLimitType LimitType = "namespace"
+	// UserLimitType limits are maintained against events from each user
+	UserLimitType LimitType = "user"
+	// SourceObjectLimitType limits are maintained against events from each source+object
+	SourceObjectLimitType LimitType = "source+object"
+)
+
+// Configuration provides configuration for the EventRateLimit admission controller.
+type Configuration struct {
```
bgrant0607: More specific type names, such as EventRateLimiter, would be useful.

contributors/design-proposals/admission_control_event_rate_limit.md
bgrant0607: https://kubernetes.io/docs/admin/authentication/#users-in-kubernetes

bgrant0607:
Related to the outstanding proposal:
https://docs.google.com/document/d/13BeJlrEcJhSKgsHOHWmHdJGqXrjSm_o9XxOtwcN6yNg/edit?ts=596f77f1#heading=h.ry2bhg2nw1t1

contributors/design-proposals/admission_control_event_rate_limit.md
davidopp: Do not use tenant. It is not specific enough in this context. There is no formal tenant concept in Kubernetes today.

bgrant0607:
I like this better than a global on/off switch (aka panic button).

If Events really are the source of the most problems, then something like this may be a reasonable bandaid until we have a more general mechanism that could work for arbitrary resource types and verbs, adapt to actual capacity and load, and employ other load-management techniques in addition to load shedding.

contributors/design-proposals/admission_control_event_rate_limit.md
staebler: The other API Server admission controllers with configuration--podtolerationrestriction and resroucequota--both use the name Configuration, so I assumed that was the preferred name for the configuration type. If that is not the case, then I am more than happy to change to a more specific name.

deads2k:
lgtm.

I think this ready for a squash.

derekwaynecarr:
LGTM

nice job on this @staebler

contributors/design-proposals/admission_control_event_rate_limit.md
```diff
@@ -0,0 +1,176 @@
...
+the user of the event request or source+object of the event and the latter
+considers the namespace of the event request.
+
+## Client Behavior
+
+Currently, the Client event recorder treats a 429 response as an http transport
+type of error, which warrants retrying the event request. Instead, the event
+recorder should abandon the event. Additionally, the event recorder should
+abandon all future events for the period of time specified in the
+Retry-After header of the 429 response.
```
jfbai: Hi, is there any plan to make event recorder to abandon events when receiving 429 ? Thanks a lot.

PR #955, opened 2017-08-22
Update writing-good-e2e-tests.md

wackxu:
fix bad url


xiangpengzhao:
checked. 
/lgtm

BTW, it'd be good to put such similar patches (#953, #954, #955, #956, #957, #959, #960, #961, #962, #963, #964, #965) into one or two larger PRs in the future :)

wackxu:
@xiangpengzhao I am so sorry for that and I will pay attention to this problem

PR #957, opened 2017-08-22
fix bad url in how-to-doc.md

wackxu:
fix bad url

### DIFF
--- a/contributors/devel/how-to-doc.md
+++ b/contributors/devel/how-to-doc.md
@@ -56,7 +56,7 @@ Use inline links instead of urls at all times. When you add internal links to
 `http://releases.k8s.io/HEAD/<path/to/link>`. For example, avoid using:
 
 ```
-[GCE](https://github.com/kubernetes/kubernetes/blob/master/docs/getting-started-guides/gce.md)  # note that it's under docs/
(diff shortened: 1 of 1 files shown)


contributors/devel/how-to-doc.md
```diff
@@ -56,7 +56,7 @@ Use inline links instead of urls at all times. When you add internal links to
 `http://releases.k8s.io/HEAD/<path/to/link>`. For example, avoid using:
 
 ```
-[GCE](https://github.com/kubernetes/kubernetes/blob/master/docs/getting-started-guides/gce.md)  # note that it's under docs/
+[GCE](https://github.com/kubernetes/kubernetes.github.io/blob/master/docs/getting-started-guides/gce.md)  # note that it's under docs/
```
xiangpengzhao: We may need to edit the example instead of just the link.

cblecker: Link should be https://kubernetes.io/docs/getting-started-guides/gce/

cblecker: ping @wackxu

wackxu: @cblecker Thanks for your kindly suggest.And I have read the file again.I agree with the opinion of @xiangpengzhao ,We should also edit the example and I have fix it.

cblecker:
please squash :)

wackxu:
@cblecker I have fix it, thanks!

PR #959, opened 2017-08-22
Update running-locally.md

wackxu:
fix bad url

### DIFF
--- a/contributors/devel/running-locally.md
+++ b/contributors/devel/running-locally.md
@@ -25,7 +25,7 @@ Getting started locally
 
 #### Linux
 
-Not running Linux? Consider running [Minikube](http://kubernetes.io/docs/getting-started-guides/minikube/), or on a cloud provider like [Google Compute Engine](../getting-started-guides/gce.md).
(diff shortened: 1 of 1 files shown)


contributors/devel/running-locally.md
```diff
@@ -25,7 +25,7 @@ Getting started locally
 
 #### Linux
 
-Not running Linux? Consider running [Minikube](http://kubernetes.io/docs/getting-started-guides/minikube/), or on a cloud provider like [Google Compute Engine](../getting-started-guides/gce.md).
+Not running Linux? Consider running [Minikube](http://kubernetes.io/docs/getting-started-guides/minikube/), or on a cloud provider like [Google Compute Engine](https://github.com/kubernetes/kubernetes.github.io/blob/master/docs/getting-started-guides/gce.md).
```
nikhiljindal: https://github.com/kubernetes/kubernetes.github.io/blob/master can be replaced by kubernetes.io, so the link becomes:
kubernetes.io/docs/getting-started-guides/gce.md

nikhiljindal: Change it to https://kubernetes.io/docs/getting-started-guides/gce/ in line with the minikube link above

wackxu: That's ok and I will fix it

PR #966, opened 2017-08-22
add guidelines for merge commits

lavalamp:
Well, against merge commits, to be more accurate.

### DIFF
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -329,6 +329,12 @@ git fetch upstream
 git rebase upstream/master
 ```
 
+Please don't use `git pull` instead of the above `fetch` / `rebase`. `git pull`
+does a merge, which leaves merge commits. These make the commit history messy
+and violate the principle that commits ought to be individually understandable
+and useful (see below). You can also consider changing your `.git/config` file via
+`git config branch.autoSetupRebase always` to change the behavior of `git pull`.
+
 ### 5 Commit
 
 Commit your changes.
@@ -384,11 +390,23 @@ Upon merge (by either you or your reviewer), all commits left on the review
 branch should represent meaningful milestones or units of work.  Use commits to
 add clarity to the development and review process.
 
-Before merging a PR, squash any _fix review feedback_, _typo_, and _rebased_
-sorts of commits.
+Before merging a PR, squash any _fix review feedback_, _typo_, _merged_, and
+_rebased_ sorts of commits.
 
 It is not imperative that every commit in a PR compile and pass tests
 independently, but it is worth striving for.
+
+In particular, if you happened to have used `git merge` and have merge
+commits, please squash those away: they do not meet the above test.
+
+A nifty way to manage the commits in your PR is to do an [interactive
+rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History),
+which will let you tell git what to do with every commit:
+
+```sh
+git fetch upstream
+git rebase -i upstream/master
+```
 
 For mass automated fixups (e.g. automated doc formatting), use one or more
 commits for the changes to tooling and a final commit to apply the fixup en


caesarxuchao:
cc @sttts

contributors/devel/development.md
```diff
@@ -329,6 +329,11 @@ git fetch upstream
 git rebase upstream/master
 ```
 
+Please don't use `git merge` instead of the above `fetch` / `rebase`. Merge
```
apelisse: And/or `git pull` (or use `git pull --rebase`). I think most of the time it happens because people use `pull`.

contributors/devel/development.md
```diff
@@ -384,12 +389,15 @@ Upon merge (by either you or your reviewer), all commits left on the review
...
 
-Before merging a PR, squash any _fix review feedback_, _typo_, and _rebased_
-sorts of commits.
+Before merging a PR, squash any _fix review feedback_, _typo_, _merged_, and
+_rebased_ sorts of commits.
 
 It is not imperative that every commit in a PR compile and pass tests
 independently, but it is worth striving for.
 
+In particular, if you happened to have used `git merge` and have non-empty merge
```
apelisse: is there any good reason to have merge commit at all?

contributors/devel/development.md
ncdc: Maybe recommend `git config branch.autoSetupRebase always`?

contributors/devel/development.md
```diff
@@ -384,12 +389,15 @@ Upon merge (by either you or your reviewer), all commits left on the review
 branch should represent meaningful milestones or units of work.  Use commits to
 add clarity to the development and review process.
 
-Before merging a PR, squash any _fix review feedback_, _typo_, and _rebased_
-sorts of commits.
+Before merging a PR, squash any _fix review feedback_, _typo_, _merged_, and
```
grodrigues3: I know this is something contributors can just google for, but if we're going to explain commands for getting the branch updated, can we also explain how to squash.

contributors/devel/development.md
lavalamp: Argh, I meant to write git pull. will fix.

eparis:
I'll take my usual "there is nothing wrong with merge commits" stance. Merge commits are what make git better than CVS. Merge commits are the thing that show that when 2 developers develop at the same time on non-overlapping code but functionally overlapping areas that it wasn't dev A or dev B who was wrong but it was the merger of A and B's work that caused a failure. Good CI may catch the problem. It may not. But a rebase is a lie. B didn't develop and test after A. He did it at the same time, and the code history should show that.

I'll also accept the recommendation that not using merge commits is a best practice for kube because they can be used poorly and many people end up with merges that shouldn't exist at all.

Any tooling that fails with a non-empty merge commit needs fixed. It is tooling that is incompatible with git itself. There was a [reasonable article](http://blog.ffwll.ch/2017/08/github-why-cant-host-the-kernel.html) I read the other day talking about the kernel as a monotree rather than a monorepo. The foundation of that type of scalability lies in merges. If we do ever decide to try to re-architect our scalability direction (obviously not happening any time soon!!) wholesale banning of merges would make it impossible.

Anyway, I'm good with this PR as written, just want to keep saying that merges are are good.

lavalamp:
@eparis I basically agree with your second paragraph. Indeed, we are not using git as it is meant to be used. I am OK with that given that we need to coordinate thousands of developers.

contributors/devel/development.md
lavalamp: Added below.

contributors/devel/development.md
lavalamp: fixed

apelisse:
Looks good to me too, thanks Daniel

PR #970, opened 2017-08-23
node label for policy and condition for CPUPressure

sjenning:
converted from https://github.com/ConnorDoyle/community/pull/2

This is to address feedback from the RWMG discussion about a month ago.

Adds information on node label advertising cpu-policy and CPUPressure node condition.

@derekwaynecarr @ConnorDoyle

### DIFF
--- a/contributors/design-proposals/cpu-manager.md
+++ b/contributors/design-proposals/cpu-manager.md
@@ -145,6 +145,10 @@ configuration. The three policies are **none**, **static** and **dynamic**.
 
 The active CPU manager policy is set through a new Kubelet
 configuration value `--cpu-manager-policy`. The default value is `none`.
+This policy will be reported in the Node resource as a label with key
+`alpha.kubernetes.io/cpu-policy`.  This allows users with with a policy
+preference to use a node selector to ensure the pod lands on a node with a
+particular cpu manager policy enabled.
 
 The number of CPUs that pods may run on can be implicitly controlled using the
 existing node-allocatable configuration settings. See the [node allocatable
@@ -266,14 +270,22 @@ func (p *staticPolicy) UnregisterContainer(s State, containerID string) error {
        cpuset for all containers running in the shared pool.
 
 1. _The shared pool becomes empty._
-    1. The CPU manager adds a node condition with effect NoSchedule,
-       NoExecute that prevents BestEffort and Burstable QoS class pods from
-       running on the node. BestEffort and Burstable QoS class pods are
-       evicted from the node.
+    1. The CPU manager sets a CPUPressure node condition to true that prevents
+       BestEffort and Burstable QoS class pods with no CPU resource request
+       from being scheduled on the node. Already running BestEffort and Burstable QoS
+       class pods without a CPU resource request are evicted from the node.
+
+       NOTE: The decision to allow the shared pool to completely empty is
+       driven by a desire to keep the scheduler accounting simple. If a
+       --cpu-policy-static-min-shared-cpus flag were to exist, there is no simple way
+       to convey that information to the scheduler. The scheduler would then need to
+       know that some portion of the CPU Allocatable is "special", for use only by
+       non-exlusive containers.
 
 1. _The shared pool becomes nonempty._
-    1. The CPU manager removes the node condition with effect NoSchedule,
-       NoExecute for BestEffort and Burstable QoS class pods.
+    1. The CPU manager sets a CPUPressure node condition to false that allows
+       BestEffort and Burstable QoS class pods with no CPU resource request to
+       be scheduled on the node.
 
 #### Policy 3: "dynamic" cpuset control
 


contributors/design-proposals/cpu-manager.md
```diff
@@ -145,6 +145,10 @@ configuration. The three policies are **none**, **static** and **dynamic**.
 
 The active CPU manager policy is set through a new Kubelet
 configuration value `--cpu-manager-policy`. The default value is `none`.
+This policy will be reported in the Node resource as a label with key
+`alpha.kubernetes.io/cpu-policy`.  This allows users with with a policy
+preference to use a node selector to ensure the pod lands on a node with a
+particular cpu manager policy enabled.
```
liggitt: there are efforts in progress (https://github.com/kubernetes/community/pull/911) to remove or restrict the ability of nodes to label themselves, since that allows them to steer workloads to themselves.

contributors/design-proposals/cpu-manager.md
```diff
@@ -266,14 +270,22 @@ func (p *staticPolicy) UnregisterContainer(s State, containerID string) error {
...
-       running on the node. BestEffort and Burstable QoS class pods are
-       evicted from the node.
+    1. The CPU manager sets a CPUPressure node condition to true that prevents
+       BestEffort and Burstable QoS class pods with no CPU resource request
+       from being scheduled on the node. Already running BestEffort and Burstable QoS
+       class pods without a CPU resource request are evicted from the node.
+
+       NOTE: The decision to allow the shared pool to completely empty is
+       driven by a desire to keep the scheduler accounting simple. If a
+       --cpu-policy-static-min-shared-cpus flag were to exist, there is no simple way
```
ConnorDoyle: nit: add backticks to flag name to avoid soft-wrapping.

contributors/design-proposals/cpu-manager.md
```diff
@@ -266,14 +270,22 @@ func (p *staticPolicy) UnregisterContainer(s State, containerID string) error {
...
+       to convey that information to the scheduler. The scheduler would then need to
+       know that some portion of the CPU Allocatable is "special", for use only by
+       non-exlusive containers.
```
ConnorDoyle: Possibly obvious, but consider also mentioning what would happen with a shared pool low-water-mark but without scheduler enhancements -- Kubelet would need to reject some G pods after scheduling, with high potential to confuse users.

contributors/design-proposals/cpu-manager.md
sjenning: Wow ok.  Didn't know about that.

I can either 1) remove this section completely or 2) s/will/can/ if trusted node label mechanism configures it independently rather then the kubelet knowing it's own config and reporting it through a label.

What will happen to all the node labels in `kubernetes.io` that the node self-labels now?

liggitt: we'll have to whitelist them (which we may do for things like os/arch/instance-type) or silently drop them until we can remove them from the kubelet and get past the version skew window for kubelets that try to set them

liggitt: using labels for this could still work, I would just expect them to be applied to nodes, rather than having nodes self-report them

derekwaynecarr: I understand the spirit of the design, but it does mean that some fields the kubelet sets today based on its own introspection of the machine need to move to fields.   Are we ok with the node controller then converting those fields to labels?  Can cpu manager policy be included in that initial whitelist?

derekwaynecarr: @vishh - have you been tracking removing the ability for a node to self label?  did you have other ideas for converting them to fields and then node controller assigning them as labels/taints?

vishh: We rely on Node labels heavily to address scheduling constraints. Fields in node object cannot be used for scheduling purposes as of now.
Having users create labels that map kubelet configuration is not a great user experience. I can see the security aspect of having nodes introduce arbitrary labels. 
Perhaps a whitelist of label keys is a way to get around the security issue? Ideally such a whitelist would be configurable (Configmaps maybe?) which would give cluster admins the ability to choose what labels they'd like to expose and also extend kubernetes easily.

vishh: @derekwaynecarr https://github.com/kubernetes/community/pull/911 is news to me. I wonder why that proposal wasn't discussed in SIG-node yet.

contributors/design-proposals/cpu-manager.md
```diff
@@ -266,14 +270,22 @@ func (p *staticPolicy) UnregisterContainer(s State, containerID string) error {
...
 
 1. _The shared pool becomes nonempty._
-    1. The CPU manager removes the node condition with effect NoSchedule,
-       NoExecute for BestEffort and Burstable QoS class pods.
+    1. The CPU manager sets a CPUPressure node condition to false that allows
```
vishh: As I mentioned in the PR, this node condition isn't necessary if we reserve a core by default which I feel will be user's expectation.

ConnorDoyle: xref [#49186 (comment)](https://github.com/kubernetes/kubernetes/pull/49186#discussion_r134932892)

derekwaynecarr:
I spoke with @ConnorDoyle and agree we can drop the CPU pressure condition.

@sjenning -- can you remove the CPU pressure condition parts of this PR.

sjenning:
@derekwaynecarr with the "no node self labeling" and dropping of the CPU Pressure condition, I'm not sure if any parts of this PR are needed anymore.  Should I just close?

vishh:
This PR can probably be closed, but we need to open a separate issue for https://github.com/kubernetes/community/pull/970#discussion_r134910980

PR #972, opened 2017-08-23
Adding new SIG BigData Lead

foxish:
Added new SIG Lead:
Announcement: https://groups.google.com/forum/#!topic/kubernetes-sig-big-data/CdJFcLVux4U


grodrigues3:
Can you add it to the sigs.yaml file in the root dir and then run make.

PR #976, opened 2017-08-23
Display github handle for sig leads

cblecker:
This changes the template format of the sig readmes so that the github handle for sig leads is displayed as well as linked to.


grodrigues3:
That's a nice little improvement.
Thanks @cblecker 

/lgtm

PR #124, opened 2016-12-01
Proposal: Pod Safety Guarantees

smarterclayton:
Changes to the system to better improve how pod safety is handled and
set a foundation for reliable cluster changes in stateful sets and
others.

From kubernetes/kubernetes#34160

PR #977, opened 2017-08-24
Add "DaemonSet create-first rolling update" design proposal.

diegs:
This document captures the results of the discussion here:

https://github.com/kubernetes/kubernetes/issues/48841

Relevant feature:

https://github.com/kubernetes/features/issues/373

cc @kow3ns @erictune @roberthbailey @janetkuo @luxas @lpabon @aaronlevy @kubernetes/sig-cluster-lifecycle-feature-requests @kubernetes/sig-apps-feature-requests

### DIFF
--- a/contributors/design-proposals/daemonset-create-first-rolling-update.md
+++ b/contributors/design-proposals/daemonset-create-first-rolling-update.md
@@ -1,0 +1,82 @@
+## DaemonSet create-first rolling update strategy
+
+DaemonSets currently support a "delete first, then add" rolling update
+strategy.
+
+This document presents a design for a new "add first, then delete" rolling
+update strategy, called `SurgingRollingUpdate`. This strategy will create new
+pods first, cleaning up the old pods once the new ones are running. As with
+Deployments, users will use a `maxSurge` parameter to control how many nodes
+are allowed to have >1 pod running during a rolling update.
+
+### Example use case
+
+For self-hosted Kubernetes DaemonSets are a natural fit for running the
+scheduler and controller-manager on master nodes. However, since DaemonSets are
+managed by the controller-manager, if `maxUnavailable` >= the number of masters
+then a rolling update of the controller-manager would delete all the old pods
+before the new pods were scheduled, leading to a control plane outage.
+
+This feature is a blocker for supporting self-hosted Kubernetes in `kubeadm`.
+
+### Implementation plan
+
+A new `SurgingRollingUpdate` strategy will be added to the `DaemonSet` API
+object. Then an additional hook for this strategy will be added to the part of
+the reconciliation loop that runs when pod updates are needed. That hook will:
+
+1. Determine how many new pods can be created, according to the number of extra
+   pods currently running and the `maxSurge` parameter.
+2. Determine how many previous-generation pods can be retired once their
+   replacement pods are scheduled and running.
+3. Instruct the controller-manager to add and delete pods based on the results
+   of (1) and (2).
+
+This is mechanically almost identical to the current `RollingUpdate` strategy,
+with the key differences being the direct creation of new pods (the
+`RollingUpdate` strategy only deletes pods up to the `maxUnavailable` threshold
+and then waits for the main reconciler create new pods to replace them).
+
+One minor complication is that the controller-manager's reconciliation loop
+actively enforces the one-pod-per-node invariant. This invariant will need to
+be relaxed when the `SurgingRollingUpdate` strategy temporarily creates extra
+pods.
+
+### Alternatives considered
+
+The `maxSurge` parameter could have been added to the existing `RollingUpdate`
+strategy (as is the case for Deployments). However, this would break backwards
+compatibility, since the `maxUnavailable` parameter currently requires
+`maxUnavailable` to be > 0, but this is not required with
+`SurgingRollingUpdate`.
+
+There are plans to create a `v1beta2` iteration of the extensions API, so it
+may be possible to add the `maxSurge` strategy to `RollingUpdate` and deprecate
+`SurgingRollingUpdate` then.
+
+For more background see https://github.com/kubernetes/kubernetes/issues/48841.
+
+### Considerations / questions
+
+1. How are `hostPort`s handled? 
+
+They are not handled as part of this proposal. We can either:
+
+  1. Attempt to determine a mechanism to handle this (unclear)
+  2. Note in the documentation that this may cause port conflicts that prevent
+    new pods from scheduling successfully (and therefore updates from completing
+    successfully)
+  3. Add a validation check that rejects the combination of `hostPort` and the
+    `SurgingRollingUpdate` strategy.
+
+2. How will the scheduler handle hostPort collisions?
+
+TODO(diegs): investigate this.
+
+3. How are other shared resources (e.g. local devices, GPUs, specific core
+   types) handled?
+
+This is a generalization of the `hostPort` question. I think that in this case
+(especially for now) we should make a note in the API documentation at best
+that running two pods that require scarce resources at once on a node may
+(naturally) cause conflicts.
--- a/contributors/design-proposals/daemonset-create-first-rolling-update.md
+++ b/contributors/design-proposals/daemonset-create-first-rolling-update.md
@@ -28,15 +28,15 @@ follows:
 
 ```go
 type DaemonSetUpdateStrategy struct {
-  // Type of daemon set update. Can be "RollingUpdate", "SurgingUpdate", or "OnDelete".
-  // Default is OnDelete.
-  // +optional
-  Type DaemonSetUpdateStrategyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
-
-  // Rolling update config params. Present only if type = "RollingUpdate".
-  //---
-  // TODO: Update this to follow our convention for oneOf, whatever we decide it
-  // to be. Same as Deployment `strategy.rollingUpdate`.
+	// Type of daemon set update. Can be "RollingUpdate", "SurgingUpdate", or "OnDelete".
+	// Default is OnDelete.
+	// +optional
+	Type DaemonSetUpdateStrategyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
+	
+	// Rolling update config params. Present only if type = "RollingUpdate".
+	//---
+	// TODO: Update this to follow our convention for oneOf, whatever we decide it
+	// to be. Same as Deployment `strategy.rollingUpdate`.
 	// See https://github.com/kubernetes/kubernetes/issues/35345
 	// +optional
 	RollingUpdate *RollingUpdateDaemonSet `json:"rollingUpdate,omitempty" protobuf:"bytes,2,opt,name=rollingUpdate"`
@@ -53,16 +53,16 @@ type DaemonSetUpdateStrategy struct {
 ...
 
 const (
- 	// Replace the old daemons by new ones using rolling update i.e replace them on each node one after the other,
- 	// killing the old pod before starting the new one.
- 	RollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "RollingUpdate"
+	// Replace the old daemons by new ones using rolling update i.e replace them on each node one after the other,
+	// killing the old pod before starting the new one.
+	RollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "RollingUpdate"
  
- 	// Replace the old daemons by new ones using rolling update i.e replace them on each node one
- 	// after the other, creating the new pod and then killing the old one.
- 	SurgingRollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "SurgingRollingUpdate"
+	// Replace the old daemons by new ones using rolling update i.e replace them on each node one
+	// after the other, creating the new pod and then killing the old one.
+	SurgingRollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "SurgingRollingUpdate"
  
- 	// Replace the old daemons only when it's killed
- 	OnDeleteDaemonSetStrategyType DaemonSetUpdateStrategyType = "OnDelete"
+	// Replace the old daemons only when it's killed
+	OnDeleteDaemonSetStrategyType DaemonSetUpdateStrategyType = "OnDelete"
 )
 
 ...
@@ -87,15 +87,16 @@ type SurgingRollingUpdateDaemonSet struct {
 Then an additional hook for this strategy will be added to the part of the
 reconciliation loop that runs when pod updates are needed. That hook will:
 
-1. Determine how many new pods can be created, according to the number of extra
-   pods currently running and the `maxSurge` parameter.
-2. Determine how many previous-generation pods can be retired once their
-   replacement pods are scheduled and running.
-3. Instruct the controller-manager to add and delete pods based on the results
-   of (1) and (2).
+1. Determine how many nodes can have new pods scheduled, according to the
+   number of extra nodes currently running more than one pod and the `maxSurge`
+   parameter.
+2. Determine how many nodes can have their previous-generation pods retired
+   once the replacement pods are scheduled and running.
+3. Instruct the controller-manager to add and delete pods on nodes based on the
+   results of (1) and (2).
 
 This is mechanically almost identical to the current `RollingUpdate` strategy,
-with the key differences being the direct creation of new pods (the
+with the key differences being the direct creation of new pods on nodes (the
 `RollingUpdate` strategy only deletes pods up to the `maxUnavailable` threshold
 and then waits for the main reconciler create new pods to replace them).
 
--- a/contributors/design-proposals/daemonset-create-first-rolling-update.md
+++ b/contributors/design-proposals/daemonset-create-first-rolling-update.md
@@ -32,7 +32,7 @@ type DaemonSetUpdateStrategy struct {
 	// Default is OnDelete.
 	// +optional
 	Type DaemonSetUpdateStrategyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
-	
+
 	// Rolling update config params. Present only if type = "RollingUpdate".
 	//---
 	// TODO: Update this to follow our convention for oneOf, whatever we decide it
@@ -56,11 +56,11 @@ const (
 	// Replace the old daemons by new ones using rolling update i.e replace them on each node one after the other,
 	// killing the old pod before starting the new one.
 	RollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "RollingUpdate"
- 
+
 	// Replace the old daemons by new ones using rolling update i.e replace them on each node one
 	// after the other, creating the new pod and then killing the old one.
 	SurgingRollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "SurgingRollingUpdate"
- 
+
 	// Replace the old daemons only when it's killed
 	OnDeleteDaemonSetStrategyType DaemonSetUpdateStrategyType = "OnDelete"
 )
@@ -125,17 +125,17 @@ For more background see https://github.com/kubernetes/kubernetes/issues/48841.
 
 ### Considerations / questions
 
-1. How are `hostPort`s handled? 
+1. How are `hostPort`s handled?
 
 They are not handled as part of this proposal. We can either:
 
-  1. Attempt to determine a mechanism to handle this (unclear, perhaps best
-     left as future work, i.e. prior to GA)
-  2. Note in the documentation that this may cause port conflicts that prevent
-     new pods from scheduling successfully (and therefore updates from
-     completing successfully)
-  3. Add a validation check that rejects the combination of `hostPort` and the
-     `SurgingRollingUpdate` strategy.
+   1. Attempt to determine a mechanism to handle this (unclear, perhaps best
+      left as future work, i.e. prior to GA)
+   2. Note in the documentation that this may cause port conflicts that prevent
+      new pods from scheduling successfully (and therefore updates from
+      completing successfully)
+   3. Add a validation check that rejects the combination of `hostPort` and the
+      `SurgingRollingUpdate` strategy.
 
 2. How will the scheduler handle hostPort collisions?
 
@@ -144,7 +144,11 @@ can get stuck in pending). To verify, create a new Deployment that uses a
 hostPort, has replicas = the number of workers (or use a selector), and set
 `maxUnavailable` to 0. If you attempt to update this Deployment the update will
 get stuck since the scheduler cannot find any nodes upon which to place the
-pods. 
+pods.
+
+It's noted that DaemonSets use `hostPort`s more frequently than Deployments,
+and Deployment replica counts are ideally << the number of nodes, so this
+issue will likely affect DaemonSets much more than Deployments.
 
 3. How are other shared resources (e.g. local devices, GPUs, specific core
    types) handled?
--- a/contributors/design-proposals/daemonset-create-first-rolling-update.md
+++ b/contributors/design-proposals/daemonset-create-first-rolling-update.md
@@ -17,13 +17,75 @@ managed by the controller-manager, if `maxUnavailable` >= the number of masters
 then a rolling update of the controller-manager would delete all the old pods
 before the new pods were scheduled, leading to a control plane outage.
 
-This feature is a blocker for supporting self-hosted Kubernetes in `kubeadm`.
+This feature is a blocker for single-master self-hosted Kubernetes, e.g. in
+`kubeadm`, and we also believe it would be a generally useful feature.
 
 ### Implementation plan
 
 A new `SurgingRollingUpdate` strategy will be added to the `DaemonSet` API
-object. Then an additional hook for this strategy will be added to the part of
-the reconciliation loop that runs when pod updates are needed. That hook will:
+object. The data structures in `extensions/v1beta1` will be modified as
+follows:
+
+```go
+type DaemonSetUpdateStrategy struct {
+  // Type of daemon set update. Can be "RollingUpdate", "SurgingUpdate", or "OnDelete".
+  // Default is OnDelete.
+  // +optional
+  Type DaemonSetUpdateStrategyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
+
+  // Rolling update config params. Present only if type = "RollingUpdate".
+  //---
+  // TODO: Update this to follow our convention for oneOf, whatever we decide it
+  // to be. Same as Deployment `strategy.rollingUpdate`.
+	// See https://github.com/kubernetes/kubernetes/issues/35345
+	// +optional
+	RollingUpdate *RollingUpdateDaemonSet `json:"rollingUpdate,omitempty" protobuf:"bytes,2,opt,name=rollingUpdate"`
+
+	// Surging rolling update config params. Present only if type = "SurgingRollingUpdate".
+	//---
+	// TODO: Update this to follow our convention for oneOf, whatever we decide it
+	// to be. Same as Deployment `strategy.rollingUpdate`.
+	// See https://github.com/kubernetes/kubernetes/issues/35345
+	// +optional
+	SurgingRollingUpdate *SurgingRollingUpdateDaemonSet `json:"surgingRollingUpdate,omitempty" protobuf:"bytes,3,opt,name=surgingRollingUpdate"`
+}
+
+...
+
+const (
+ 	// Replace the old daemons by new ones using rolling update i.e replace them on each node one after the other,
+ 	// killing the old pod before starting the new one.
+ 	RollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "RollingUpdate"
+ 
+ 	// Replace the old daemons by new ones using rolling update i.e replace them on each node one
+ 	// after the other, creating the new pod and then killing the old one.
+ 	SurgingRollingUpdateDaemonSetStrategyType DaemonSetUpdateStrategyType = "SurgingRollingUpdate"
+ 
+ 	// Replace the old daemons only when it's killed
+ 	OnDeleteDaemonSetStrategyType DaemonSetUpdateStrategyType = "OnDelete"
+)
+
+...
+
+// Spec to control the desired behavior of a daemon set surging rolling update.
+type SurgingRollingUpdateDaemonSet struct {
+	// The maximum number of DaemonSet pods that can be scheduled above the desired number of pods
+	// during the update. Value can be an absolute number (ex: 5) or a percentage of the total number
+	// of DaemonSet pods at the start of the update (ex: 10%). The absolute number is calculated from
+	// the percentage by rounding up. This cannot be 0. The default value is 1. Example: when this is
+	// set to 30%, at most 30% of the total number of nodes that should be running the daemon pod
+	// (i.e. status.desiredNumberScheduled) can have 2 pods running at any given time. The update
+	// starts by starting replacements for at most 30% of those DaemonSet pods. Once the new pods are
+	// available it then stops the existing pods before proceeding onto other DaemonSet pods, thus
+	// ensuring that at most 130% of the desired final number of DaemonSet  pods are running at all
+	// times during the update.
+	// +optional
+	MaxSurge *intstr.IntOrString `json:"maxSurge,omitempty" protobuf:"bytes,2,opt,name=maxSurge"`
+}
+```
+
+Then an additional hook for this strategy will be added to the part of the
+reconciliation loop that runs when pod updates are needed. That hook will:
 
 1. Determine how many new pods can be created, according to the number of extra
    pods currently running and the `maxSurge` parameter.
@@ -38,21 +100,25 @@ with the key differences being the direct creation of new pods (the
 and then waits for the main reconciler create new pods to replace them).
 
 One minor complication is that the controller-manager's reconciliation loop
-actively enforces the one-pod-per-node invariant. This invariant will need to
-be relaxed when the `SurgingRollingUpdate` strategy temporarily creates extra
-pods.
+actively tries to enforce the one-pod-per-node invariant, and work is underway
+to fix cases in which the invariant is not held
+[kubernetes/kubernetes#50477](https://github.com/kubernetes/kubernetes/issues/50477).
+However, this invariant will need to be relaxed when the `SurgingRollingUpdate`
+strategy temporarily creates extra pods.
 
 ### Alternatives considered
 
 The `maxSurge` parameter could have been added to the existing `RollingUpdate`
-strategy (as is the case for Deployments). However, this would break backwards
-compatibility, since the `maxUnavailable` parameter currently requires
-`maxUnavailable` to be > 0, but this is not required with
+strategy (as is the case for Deployments). However, this would break validation
+backwards compatibility, since the `maxUnavailable` parameter currently
+requires `maxUnavailable` to be > 0, but this is not required with
 `SurgingRollingUpdate`.
 
-There are plans to create a `v1beta2` iteration of the extensions API, so it
-may be possible to add the `maxSurge` strategy to `RollingUpdate` and deprecate
-`SurgingRollingUpdate` then.
+There are plans to create a `v1beta2` iteration of the apps API (enabled by
+default in 1.9+), so it may be possible to add the `maxSurge` strategy to
+`RollingUpdate` and deprecate `SurgingRollingUpdate` then, or make the change
+now since the `v1beta2` API is incoming. However, to be conservative it was
+decided to implement this as a separate strategy.
 
 For more background see https://github.com/kubernetes/kubernetes/issues/48841.
 
@@ -62,16 +128,22 @@ For more background see https://github.com/kubernetes/kubernetes/issues/48841.
 
 They are not handled as part of this proposal. We can either:
 
-  1. Attempt to determine a mechanism to handle this (unclear)
+  1. Attempt to determine a mechanism to handle this (unclear, perhaps best
+     left as future work, i.e. prior to GA)
   2. Note in the documentation that this may cause port conflicts that prevent
-    new pods from scheduling successfully (and therefore updates from completing
-    successfully)
+     new pods from scheduling successfully (and therefore updates from
+     completing successfully)
   3. Add a validation check that rejects the combination of `hostPort` and the
-    `SurgingRollingUpdate` strategy.
+     `SurgingRollingUpdate` strategy.
 
 2. How will the scheduler handle hostPort collisions?
 
-TODO(diegs): investigate this.
+The pods will not be scheduled. This is also true of Deployments (the new pods
+can get stuck in pending). To verify, create a new Deployment that uses a
+hostPort, has replicas = the number of workers (or use a selector), and set
+`maxUnavailable` to 0. If you attempt to update this Deployment the update will
+get stuck since the scheduler cannot find any nodes upon which to place the
+pods. 
 
 3. How are other shared resources (e.g. local devices, GPUs, specific core
    types) handled?
@@ -80,3 +152,26 @@ This is a generalization of the `hostPort` question. I think that in this case
 (especially for now) we should make a note in the API documentation at best
 that running two pods that require scarce resources at once on a node may
 (naturally) cause conflicts.
+
+4. When can this strategy be used?
+
+This strategy will be selectable by users at both DaemonSet creation time
+and modification time. If a rolling update is occurring while the strategy
+is changed the new strategy will take effect and reconciliation will occur.
+
+In the case of `RollingUpdate` => `SurgingRollingUpdate`:
+
+- New pods will be created up to `maxSurge` (see below).
+- The controller will stop deleting the old pods *before* new pods are running,
+  instead waiting for the surging pods to become ready.
+
+In the case of `SurgingRollingUpdate` => `RollingUpdate`:
+
+- Surging pods will be killed by the controller (the current behavior in the
+  controller is to retain the oldest pod per node when multiple are present).
+- The controller will begin deleting the old pods up to `maxUnavailable` and
+  then let the controller reschedule them.
+
+5. Is this strategy enabled by default? Feature gated? Alpha? Beta?
+
+#TODO(luxas)


contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Example use case
+
+For self-hosted Kubernetes DaemonSets are a natural fit for running the
+scheduler and controller-manager on master nodes. However, since DaemonSets are
+managed by the controller-manager, if `maxUnavailable` >= the number of masters
+then a rolling update of the controller-manager would delete all the old pods
+before the new pods were scheduled, leading to a control plane outage.
+
+This feature is a blocker for supporting self-hosted Kubernetes in `kubeadm`.
```
liggitt: is it a blocker with a HA master node setup with maxUnavailable set to 1?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+There are plans to create a `v1beta2` iteration of the extensions API, so it
+may be possible to add the `maxSurge` strategy to `RollingUpdate` and deprecate
+`SurgingRollingUpdate` then.
+
+For more background see https://github.com/kubernetes/kubernetes/issues/48841.
+
+### Considerations / questions
+
+1. How are `hostPort`s handled?
```
liggitt: aren't hostPorts likely to be common in self-hosted scenarios (one of the driving use cases for this proposal)?

thockin: This proposal is simply incompatible with hostPorts, barring some really clever work that isn't on any roadmap.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+may be possible to add the `maxSurge` strategy to `RollingUpdate` and deprecate
+`SurgingRollingUpdate` then.
+
+For more background see https://github.com/kubernetes/kubernetes/issues/48841.
+
+### Considerations / questions
+
+1. How are `hostPort`s handled? 
+
+They are not handled as part of this proposal. We can either:
```
lukaszo: Allowing non working configuration is a bad idea.  I think option 3 should be implemented.

contributors/design-proposals/daemonset-create-first-rolling-update.md
luxas: This doesn't affect HA deployments given all components talk to a loadbalanced endpoint of any kind

luxas: @diegs We could point out that this is for the one-master case only, but that it is a _generally useful feature_, not a kubeadm specific thing.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+One minor complication is that the controller-manager's reconciliation loop
+actively enforces the one-pod-per-node invariant. This invariant will need to
+be relaxed when the `SurgingRollingUpdate` strategy temporarily creates extra
+pods.
+
+### Alternatives considered
+
+The `maxSurge` parameter could have been added to the existing `RollingUpdate`
+strategy (as is the case for Deployments). However, this would break backwards
```
luxas: break _validation_ backwards compability

Please clarify that while all this would have been technically possible due to the shift to v1beta2; it was discarded just to be on the safe side.

contributors/design-proposals/daemonset-create-first-rolling-update.md
luxas: @liggitt In kubeadm's case; no. We're using DaemonSets with hostNetworking (due to an other scheduling bug)

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+For more background see https://github.com/kubernetes/kubernetes/issues/48841.
+
+### Considerations / questions
+
+1. How are `hostPort`s handled? 
+
+They are not handled as part of this proposal. We can either:
+
+  1. Attempt to determine a mechanism to handle this (unclear)
```
luxas: point out that this may be a future improvement, but will not be taken into consideration right now.
Possibly for GA?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+  2. Note in the documentation that this may cause port conflicts that prevent
+    new pods from scheduling successfully (and therefore updates from completing
+    successfully)
+  3. Add a validation check that rejects the combination of `hostPort` and the
```
luxas: :+1: please point out as the accepted proposal

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+    `SurgingRollingUpdate` strategy.
+
+2. How will the scheduler handle hostPort collisions?
+
+TODO(diegs): investigate this.
+
+3. How are other shared resources (e.g. local devices, GPUs, specific core
```
luxas: How do deployments handle this?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+strategy (as is the case for Deployments). However, this would break backwards
+compatibility, since the `maxUnavailable` parameter currently requires
+`maxUnavailable` to be > 0, but this is not required with
+`SurgingRollingUpdate`.
+
+There are plans to create a `v1beta2` iteration of the extensions API, so it
+may be possible to add the `maxSurge` strategy to `RollingUpdate` and deprecate
+`SurgingRollingUpdate` then.
+
+For more background see https://github.com/kubernetes/kubernetes/issues/48841.
```
kow3ns: Why is your proposal a better alternative than creating a new DaemonSet and only deleting the current DaemonSet after the first has saturated the desired number of Nodes? Can you discuss this more here?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Considerations / questions
+
+1. How are `hostPort`s handled? 
+
```
kow3ns: I don't think its acceptable to completely punt on hostPort conflicts. They are the most common way of communicating with applications in a DaemonSet. Using validation to block the creation of DaemonSets with this updateStrategy (as you suggest below) is one option, but it really limits the usefulness of the updateStrategy.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+They are not handled as part of this proposal. We can either:
+
+  1. Attempt to determine a mechanism to handle this (unclear)
+  2. Note in the documentation that this may cause port conflicts that prevent
+    new pods from scheduling successfully (and therefore updates from completing
+    successfully)
+  3. Add a validation check that rejects the combination of `hostPort` and the
+    `SurgingRollingUpdate` strategy.
+
+2. How will the scheduler handle hostPort collisions?
```
kow3ns: Unless application designers implement clever coordination, it will attempt to schedule both Pods, and one of them will continuously fail.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+   replacement pods are scheduled and running.
+3. Instruct the controller-manager to add and delete pods based on the results
+   of (1) and (2).
+
+This is mechanically almost identical to the current `RollingUpdate` strategy,
+with the key differences being the direct creation of new pods (the
+`RollingUpdate` strategy only deletes pods up to the `maxUnavailable` threshold
+and then waits for the main reconciler create new pods to replace them).
+
+One minor complication is that the controller-manager's reconciliation loop
```
kow3ns: This isn't actually true. One Pod per node is best effort. In contrast to this proposal, which would relax this constraint, the community has request exactly one Pod per node to provide better guarantees with respect to the overlap of lifecycles of the Pods in their DeamonSets. Right now it is possible for DaemonSet to schedule a new Pod while another Pod is still terminating. We consider this to be a bug (https://github.com/kubernetes/kubernetes/issues/50477).

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+    `SurgingRollingUpdate` strategy.
+
+2. How will the scheduler handle hostPort collisions?
+
+TODO(diegs): investigate this.
+
+3. How are other shared resources (e.g. local devices, GPUs, specific core
+   types) handled?
+
+This is a generalization of the `hostPort` question. I think that in this case
```
kow3ns: I think that consuming GPU and specific core types can be considered out of scope. It is not a common use case for DaemonSets. For hostPath volumes, you could perhaps advise using filesystem locks to prevent data corruption when applications use this update strategy. If you combine filesystem locks, a burden on application implementors (at least initially), with a robust solution to hostPort conflicts, you have something that most applications can consume.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+For self-hosted Kubernetes DaemonSets are a natural fit for running the
+scheduler and controller-manager on master nodes. However, since DaemonSets are
+managed by the controller-manager, if `maxUnavailable` >= the number of masters
+then a rolling update of the controller-manager would delete all the old pods
+before the new pods were scheduled, leading to a control plane outage.
+
+This feature is a blocker for supporting self-hosted Kubernetes in `kubeadm`.
+
+### Implementation plan
```
kow3ns: There are a few things which aren't clear to me.

1. When can I select this strategy? Does it have to be set on creation? Can I change the updateStrategy from RollingUpdate to SurgeRollingUpdate (currently updateStrategy is mutable)?
1. How will the controller handle the case where the updateStrategy is mutated to SurgeRollingUpdate when RollingUpdate was previously selected and an update is in progress.
1. How will the controller handle the case where the updateStrategy is mutated to RollingUpdate and SurgeRollingUpdate was previously selected and an update is in progress.
1. Is this on or off by default? Do you intend to use feature gate flags make this an alpha feature, or will all clusters immediately get this feature?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Alternatives considered
+
+The `maxSurge` parameter could have been added to the existing `RollingUpdate`
+strategy (as is the case for Deployments). However, this would break backwards
+compatibility, since the `maxUnavailable` parameter currently requires
+`maxUnavailable` to be > 0, but this is not required with
+`SurgingRollingUpdate`.
+
+There are plans to create a `v1beta2` iteration of the extensions API, so it
```
liggitt: > There are plans to create a `v1beta2` iteration of the extensions API

do you mean v1beta2 of the apps API group? The goal is to eliminate the extensions API group, not perpetuate it

contributors/design-proposals/daemonset-create-first-rolling-update.md
liggitt: host ports are only one possible scarce resource (albeit the most likely). validating just that still allows for impossible-to-satisfy updates involving scarce resources, which is worse for daemonsets than for deployments because there is no possibility of finding another node with the available resources.

contributors/design-proposals/daemonset-create-first-rolling-update.md
liggitt: they find other nodes with available resources

contributors/design-proposals/daemonset-create-first-rolling-update.md
liggitt: > We're using DaemonSets with hostNetworking

if you are using host networking, won't your new containers still be unable to bind to the same ports (since they'd still be held by the old container), and never go healthy, and fail upgrade?

contributors/design-proposals/daemonset-create-first-rolling-update.md
lukaszo: As far as I remember DaemonSet controller checks only  for HostPorts conflicts. It will not allow new pod to run on a node and it will cause dead lock.

contributors/design-proposals/daemonset-create-first-rolling-update.md
lukaszo: It will not schedule the new pod. https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/daemon/daemon_controller.go#L1253

kow3ns: Is that true for the default scheduler as well (Assuming me move to it one day)?

lukaszo: I think that for default scheduler pod will be created and will be in Pending state

contributors/design-proposals/daemonset-create-first-rolling-update.md
thockin: SO_REUSEPORT ?

What I alluded to with "really clever work" would be to emulate that in the port-forwarding  path, but that sounds disastrous.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+
+DaemonSets currently support a "delete first, then add" rolling update
+strategy.
+
+This document presents a design for a new "add first, then delete" rolling
+update strategy, called `SurgingRollingUpdate`. This strategy will create new
+pods first, cleaning up the old pods once the new ones are running. As with
+Deployments, users will use a `maxSurge` parameter to control how many nodes
+are allowed to have >1 pod running during a rolling update.
+
```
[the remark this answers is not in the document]
diegs: Done

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Done

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Added answers in the "Considerations / questions" section.

diegs: Regarding alpha/beta/feature gate, I defer to @luxas and others.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Clarified that the invariant is not 100% enforced yet but it is a goal.

This strategy is incompatible with the invariant, so either we need to relax it during updates as mentioned, or not implement this.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Done

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Fixed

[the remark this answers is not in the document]
diegs: Noted

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Can you elaborate on this strategy? Do you mean having the user create a new DaemonSet with a different name, and then deleting this one?

If so I can think of many reasons (retaining history, minimizing clutter, actually being able to introspect the status of the update, etc.) but I want to make sure I understand which you mean.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: I would argue that special cases for hostPorts, hostNetworking, and other scarce resources like GPUs should not be a goal here. In particular, the behavior you describe is also true of Deployments.

For example (and I tested this), create a deployment where replicas = worker nodes, and use a hostport. Set `maxUnavailable` to 0, and then update your deployment. The new pods will be stuck in pending since the scheduler can't place the new pods anywhere.

This behavior is also true for DaemonSets as you point out (updated in the doc). And yes, I can see how DaemonSets may more often use hostPorts / hostNetworking in practice, but I think the scheduler's job is to place pods subject to the intersection of their resource constraints, not do resource planning for you.

We should absolutely call out this risk in the documentation. But I don't think that it should be this code's job to try to foresee potential resource conflicts.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: See above.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: As I mentioned above, you can reach this state with Deployments too. I'll defer on picking an answer to this question until we resolve that discussion.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: That is the behavior I observed.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Agreed, and if you think of DaemonSets as a special case of a Deployment with numReplicas = numNodes and antiaffinity then the scheduling will degenerate in the same way if you are using scarce resources.

It would require a lot of special-case engineering to work around all possible issues, and I'm sure there will always be more.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,177 @@
...
+  // Type of daemon set update. Can be "RollingUpdate", "SurgingUpdate", or "OnDelete".
+  // Default is OnDelete.
+  // +optional
+  Type DaemonSetUpdateStrategyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
+
+  // Rolling update config params. Present only if type = "RollingUpdate".
+  //---
+  // TODO: Update this to follow our convention for oneOf, whatever we decide it
+  // to be. Same as Deployment `strategy.rollingUpdate`.
+	// See https://github.com/kubernetes/kubernetes/issues/35345
```
xiang90: fix indentation?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+A new `SurgingRollingUpdate` strategy will be added to the `DaemonSet` API
+object. Then an additional hook for this strategy will be added to the part of
+the reconciliation loop that runs when pod updates are needed. That hook will:
+
+1. Determine how many new pods can be created, according to the number of extra
+   pods currently running and the `maxSurge` parameter.
+2. Determine how many previous-generation pods can be retired once their
+   replacement pods are scheduled and running.
+3. Instruct the controller-manager to add and delete pods based on the results
+   of (1) and (2).
```
janetkuo: `SurgingRollingUpdate` will create a new DaemonSet pod being created before the old one gets terminated **on the same node**, correct? 

For example, let's say you have 3 nodes, maxSurge=1, maxUavailable=0. Node 1 has 2 pods, and each of the other 2 nodes has 1 pod (4 pods in total). Then you can only delete pods on node 1, but not the other two. 

If this is the requirement, this is not just about pods anymore, you need to mention nodes here.

contributors/design-proposals/daemonset-create-first-rolling-update.md
janetkuo: That bug is unexpected and was recently introduced by https://github.com/kubernetes/kubernetes/pull/43077 (1.8). This behavior (don't create new pods until the old one is terminated) has been supported by DaemonSet controller for a long time.

contributors/design-proposals/daemonset-create-first-rolling-update.md
janetkuo: Yes it should be `apps/v1beta2`. Since `apps/v1beta2` is introduced in 1.8, this validation change can be made in `apps/v1beta2` only if it's implemented in 1.8.

contributors/design-proposals/daemonset-create-first-rolling-update.md
janetkuo: Agree. Multiple DaemonSets should be listed as an alternative here, too.

contributors/design-proposals/daemonset-create-first-rolling-update.md
thockin: I basically agree.  I don't think it is tractable to try to fix this special case.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
...
+be relaxed when the `SurgingRollingUpdate` strategy temporarily creates extra
+pods.
+
+### Alternatives considered
+
+The `maxSurge` parameter could have been added to the existing `RollingUpdate`
+strategy (as is the case for Deployments). However, this would break backwards
+compatibility, since the `maxUnavailable` parameter currently requires
+`maxUnavailable` to be > 0, but this is not required with
+`SurgingRollingUpdate`.
```
janetkuo: Also, most DaemonSet applications can't tolerate two pods running at the same time. That's why `RollingUpdate` is implemented that way (delete first, then create).

contributors/design-proposals/daemonset-create-first-rolling-update.md
janetkuo: Agree, hostPort conflicts need to be handled for `SurgingRollingUpdate` to be useful.

diegs: @janetkuo handled in what sense? Can you comment on the discussion immediately above this one? We could block it in validation (since I guess the code already has a special case to block it from a scheduling perspective), but besides that I'm not sure what else we can do.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Correct. Clarified.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Done

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,82 @@
+## DaemonSet create-first rolling update strategy
+
+DaemonSets currently support a "delete first, then add" rolling update
+strategy.
+
+This document presents a design for a new "add first, then delete" rolling
+update strategy, called `SurgingRollingUpdate`. This strategy will create new
```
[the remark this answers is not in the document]
diegs: This is noted in **Alternatives Considered** below. It was decided in https://github.com/kubernetes/kubernetes/issues/48841 to add a new strategy so validation would remain backwards-compatible.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,177 @@
...
+pods first, cleaning up the old pods once the new ones are running. As with
+Deployments, users will use a `maxSurge` parameter to control how many nodes
```
lpabon: What happens if maxSuge is zero? Would that interrupt the algorithm in this paper? Also what is the default value for maxSuge?

contributors/design-proposals/daemonset-create-first-rolling-update.md
lpabon: This should be documented and highlighted.m  The documentation can point the user to use the "delete before add" model. 

Also, it is possible to test for a pod having a hostPath and refusing to do a "add before delete" option? (updated: it is noted in the document on line 136)

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,177 @@
...
+1. How are `hostPort`s handled? 
+
+They are not handled as part of this proposal. We can either:
+
+  1. Attempt to determine a mechanism to handle this (unclear, perhaps best
+     left as future work, i.e. prior to GA)
+  2. Note in the documentation that this may cause port conflicts that prevent
+     new pods from scheduling successfully (and therefore updates from
+     completing successfully)
+  3. Add a validation check that rejects the combination of `hostPort` and the
```
lpabon: :+1:

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,178 @@
...
+     `SurgingRollingUpdate` strategy.
+
+2. How will the scheduler handle hostPort collisions?
+
+The pods will not be scheduled. This is also true of Deployments (the new pods
+can get stuck in pending). To verify, create a new Deployment that uses a
+hostPort, has replicas = the number of workers (or use a selector), and set
+`maxUnavailable` to 0. If you attempt to update this Deployment the update will
+get stuck since the scheduler cannot find any nodes upon which to place the
+pods.
```
janetkuo: It's common for a DaemonSet to specify `hostPort` in its pod template, but not for Deployments.

janetkuo:
Would like to see other concrete example applications that will use this new strategy. Also, since `hostPort` is the most common way of communicating with DaemonSets, it makes sense for old and new DaemonSet pods to transfer control of the port (if they're using the same port) when in this strategy.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,178 @@
...
+
+For more background see https://github.com/kubernetes/kubernetes/issues/48841.
+
+### Considerations / questions
+
+1. How are `hostPort`s handled? 
+
+They are not handled as part of this proposal. We can either:
+
+  1. Attempt to determine a mechanism to handle this (unclear, perhaps best
```
janetkuo: Minor comment on the format: the ordered list items are not rendered as expected. Try putting 3 spaces before sub-list?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,178 @@
...
+This document presents a design for a new "add first, then delete" rolling
+update strategy, called `SurgingRollingUpdate`. This strategy will create new
+pods first, cleaning up the old pods once the new ones are running. As with
+Deployments, users will use a `maxSurge` parameter to control how many nodes
+are allowed to have >1 pod running during a rolling update.
+
+### Example use case
+
+For self-hosted Kubernetes DaemonSets are a natural fit for running the
+scheduler and controller-manager on master nodes. However, since DaemonSets are
```
liggitt: will the proposed approach even work for scheduler and controller manager the way kubeadm wants it to? running with hostNetwork, they will try to bind their ports for health checks and be blocked by the existing pods, right?

lpabon: Sounds like any host resources (like hostPath) will have conflicts.

liggitt: so, it doesn't sound like implementing this will unblock kubeadm's use of daemonsets for single master self-hosting

eparis: We could bind the second time with SO_REUSEADDR to get around the hostPort issue.

HostPath will conflict but wouldn't likely complain and could easily result in some form of data corruption.

erictune:
Can you use [socket passing](https://gist.github.com/majek/2500291) to an unrelated Pod?  Would be a way to hand off the port.

Health checks could be exec based, and would need to return healthy even when the port was not "held".

eparis:
I think that socket passing could also solve the problem but would take a bit of a re-arch. It would require some parent daemon to start which opens the socket and which passed it to the real daemon. That passing would require a shared filesystem/mount. Probably a shared hostPath. It probably would be possible to make the `parent` actually be the 'active' daemon and have it be the one who passes the socket to the next daemon. But that's a lot of work...

Making them both listen at the same time is what haproxy does. It can/does result in the possibility of some dropped connections if they come in at just the right time, but all of our tools would just try again and be just fine...

In any case hostPort is solvable, hostPath less so..

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: In `bootkube` we don't run the scheduler or controller-manager with host networking. The apiserver does, but we already run that as a daemonset and use the standard kill-then-start strategy (which the checkpointer helps facilitate) during upgrades.

I'm not sure what the plan is for `kubeadm` however. cc @luxas

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Added a note to that effect for full honesty.

contributors/design-proposals/daemonset-create-first-rolling-update.md
diegs: Attempted to fix...

diegs:
@janetkuo and others, to summarize the remaining items:

(1) It sounds like the main blocker here is that we want to find out some kind of way to coordinate handing `hostPorts` over, as opposed to failing a validation check if the DaemonSet contains a `hostPort`?

If this is the case, it seems like there are solutions, but none are straightforward, and they may require some cooperation from the daemon processes themselves? Please confirm if this is where we want to go and I can investigate this further.

(2) Regarding more use cases, I can't think of any other concrete ones; this feature was admittedly born out of the specific needs of self-hosted control planes. I imagine a general use case would be that for daemonsets that *don't* have resource conflicts this may be a preferable strategy to minimize downtime of whatever service they provide. For example, for some sort of "node healthy" agent, having 2 agents report that the same node is healthy for a short while may be preferable to no agent reporting any status for a short while.

Thanks!

janetkuo:
Yes, the main blocker is to propose a way to hand off the port. It'd be great to find another concrete use case, to demonstrate that this feature needs to be implemented in Kubernetes core.

luxas:
@janetkuo Any highly-available node daemon that always needs to have at least 1 replica running will need this. That daemon then uses leader election to hand over responsibility internally.
This can be an ingress controller or whatever that should not have to suffer from much downtime.

contributors/design-proposals/daemonset-create-first-rolling-update.md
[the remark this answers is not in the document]
krmayankk: I think its a great idea to converge on strategies across all controllers. We could add MaxSurge and MaxUnavailability and keep the default as MaxSurge=0 and MaxUnavailability as 1 to keep it backward compatible. @diegs how does this break validation ?

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,182 @@
...
+However, this invariant will need to be relaxed when the `SurgingRollingUpdate`
+strategy temporarily creates extra pods.
+
+### Alternatives considered
+
+The `maxSurge` parameter could have been added to the existing `RollingUpdate`
+strategy (as is the case for Deployments). However, this would break validation
+backwards compatibility, since the `maxUnavailable` parameter currently
+requires `maxUnavailable` to be > 0, but this is not required with
+`SurgingRollingUpdate`.
```
krmayankk: the amount of confusion a new type of RollingUpdate will create and all the documentation required to explain minor and subtle differences between these new rolling updates will add to users confusion. Was there a way to ask the community what they thought, breaking the validation seems like a not big issue as long as we keep the default behavior backward compatible

rphillips:
@janetkuo wouldn't the application within the daemonset need to support socket passing to handoff the port correctly? Seems like it would be a burden on the developer, or is there another way?

bgrant0607:
cc @kubernetes/sig-scheduling-proposals

bgrant0607:
cc @bsalamat @k82cn

ant31:
I don't an issue with the proposal of failing the validation if a `hostPort` is present:
1. The other solution is way more complex and still 'fragile' (depend on the app & network)
2. There was no demand for this from users (in this issue) to implement such logic
3. Handling hostPort could be implemented later as a second step (1.20/1?)

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,182 @@
...
+
+   1. Attempt to determine a mechanism to handle this (unclear, perhaps best
+      left as future work, i.e. prior to GA)
+   2. Note in the documentation that this may cause port conflicts that prevent
+      new pods from scheduling successfully (and therefore updates from
+      completing successfully)
+   3. Add a validation check that rejects the combination of `hostPort` and the
+      `SurgingRollingUpdate` strategy.
+
+2. How will the scheduler handle hostPort collisions?
```
k82cn: The pod will not be scheduled because of `PodFitsHostPorts` predicate; and kubelet will also reuse this predicates to handle multi-scheduler case.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,182 @@
...
+update strategy, called `SurgingRollingUpdate`. This strategy will create new
+pods first, cleaning up the old pods once the new ones are running. As with
+Deployments, users will use a `maxSurge` parameter to control how many nodes
+are allowed to have >1 pod running during a rolling update.
+
+### Example use case
+
+For self-hosted Kubernetes DaemonSets are a natural fit for running the
+scheduler and controller-manager on master nodes. However, since DaemonSets are
+managed by the controller-manager, if `maxUnavailable` >= the number of masters
```
k82cn: For this case, can we add validation to avoid such case? or make `maxUnavailable` to be percentage, e.g. 10% of master; if `maxUnavailable` can not meet, raise an event to identify that the upgrade failed.

k82cn:
> It'd be great to find another concrete use case, to demonstrate that this feature needs to be implemented in Kubernetes core.

+1 :)

> if `maxUnavailable` >= the number of masters then a rolling update of the controller-manager would delete all the old pods before the new pods were scheduled, leading to a control plane outage.

As I comments, can we make `maxUnavailabe` be percentage of master node? and if `maxUnavailable` is 0, raise an event that upgrade is failed.

I'm also thinking how we will handle `maxSurge` if Quota is full ? rolling upgrade will fail?

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/977)

kargakis:
> I'm also thinking how we will handle maxSurge if Quota is full ? rolling upgrade will fail?

If you are constrained by quota you need to use maxUnavailable, otherwise you are blocked.

kargakis:
> Yes, the main blocker is to propose a way to hand off the port. It'd be great to find another concrete use case, to demonstrate that this feature needs to be implemented in Kubernetes core.

kubeadm is not using host ports so it sounds to me that this proposal does not need to handle host ports and that can be handled in a separate proposal.

bgrant0607:
Our top priority in the workloads area is to advance those APIs to stable/GA in 1.9.
https://github.com/kubernetes/features/issues/353

This seems complicated. It's arguable whether we should ever do it, but we should not do it in 1.9.

timothysc:
Why couldn't this be an alpha feature-gated annotation?

erictune:
We need concrete use cases to prioritize and to focus design discussion.

I see janetkuo [requested them in August](https://github.com/kubernetes/community/pull/977#issuecomment-325851925).

Suggest tabling further design and priority discussion until we have them.

timothysc:
The primary motivation was from a sig-cluster-lifecycle user story for a single self-hosted master upgrade scenario.  We needed a means to ensure the new pod was there before the old was torn down.

kargakis:
> The primary motivation was from a sig-cluster-lifecycle user story for a single self-hosted master upgrade scenario. We needed a means to ensure the new pod was there before the old was torn down.

Also, copying text from the proposal:
> For self-hosted Kubernetes DaemonSets are a natural fit for running the
> scheduler and controller-manager on master nodes. However, since DaemonSets are
> managed by the controller-manager, if `maxUnavailable` >= the number of masters
> then a rolling update of the controller-manager would delete all the old pods
> before the new pods were scheduled, leading to a control plane outage.
>
> This feature is a blocker for single-master self-hosted Kubernetes, e.g. in
> `kubeadm`

anguslees:
> The primary motivation was from a sig-cluster-lifecycle user story for a single self-hosted master upgrade scenario. We needed a means to ensure the new pod was there before the old was torn down.

Here's my dumb newbie question for today: Why do the self-hosted kubeadm jobs use DaemonSets and not Deployments?

If they were Deployments, we'd already have maxSurge, and for things like controller-manager, it doesn't even make sense to be strictly 1:1 with master nodes (afaics?).  The apiserver is a bit special because it also has to manage the serving ip/port(s), but that seems like an unrelated challenge - indeed conceptually the apiserver is a regular stateless web server, and could even be hooked up to an HPA (ie: obvious Deployment use-case).

diegs:
> Here's my dumb newbie question for today: Why do the self-hosted kubeadm jobs use DaemonSets and not Deployments?

> If they were Deployments, we'd already have maxSurge, and for things like controller-manager, it doesn't even make sense to be strictly 1:1 with master nodes (afaics?). The apiserver is a bit special because it also has to manage the serving ip/port(s), but that seems like an unrelated challenge - indeed conceptually the apiserver is a regular stateless web server, and could even be hooked up to an HPA (ie: obvious Deployment use-case).

We currently do use Deployments in [bootkube](https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/asset/internal/templates.go#L550) and it mostly works. There are a few drawbacks for HA setups, however:

- If you don't specify anti-affinity then all of your pods may get scheduled on the same node. For normal deployments this is not necessarily a problem, but if all your schedulers or controller-managers are on one node and that node goes down then your control plane dies and won't come back up.
- If you do add pod anti-affinity to deployments to ensure that the pods spread across your master nodes then you need to micro-manage that during upgrades by bumping the anti-affinity labels every time.
- You have to manually manage the number of replicas, i.e. to match the number of master nodes you have (something like [cluster-proportional-autoscaler](https://github.com/kubernetes-incubator/cluster-proportional-autoscaler) could help but currently it doesn't support node selectors).

Using DaemonSets would alleviate a lot of these problems. Currently we believe the safest HA setup is to have the apiserver, scheduler, and controller manager running on every master node in your cluster.

smarterclayton:
If i understand the case correctly, my recommendation is:

1. remove the port declaration from the daemonset pod
2. have the service explicitly identify the port by name
3. have the pod have a pre-stop hook that waits until a new pod is detected or some timeout
4. lengthen the grace period
5. use daemonset as is today

Daemonset should create new pod before old is deleted (while pre-stop is still running).  New pod can bind 0.0.0.0:HOSTPORT with SO_REUSEPORT.  Old pod detects and pre-stop exits, and then is torn down.

diegs:
@smarterclayton a few comments:

1. The hostPort discussion is orthogonal to our use case. The scheduler and controller-manager don't need hostPorts. It's a discussion point for the strategy design and implementation, but one of the solutions I proposed (failing validation) works fine for the self-hosted use case (and I have to assume many others).
2. Regardless, I can see how your workaround could somewhat serve our needs regardless of whether a hostPort is reqiured or not. However, I don't think hacking the semantics of the existing strategy by relying on timeouts and coordination between the binaries inside the new and old pod is not a robust solution. You run into the classic problem of never being able to set a timeout long enough to guarantee this handoff always works and running into a situation where it's exceeded, and then your control plane and cluster will go down. (As an analogy, it's the difference between relying on sleep()/timing in a multithreaded test versus explicit coordination via semaphores--the former is all but guaranteed to flake at some point).

I believe that for self-hosting / kubeadm we would much rather go with something that explicitly provides the correct semantics.

smarterclayton:
> However, I don't think hacking the semantics of the existing strategy by relying on timeouts and coordination between the binaries inside the new and old pod is not a robust solution

You misunderstand - this is supported.  It's not a hack - it's what we tell everyone who has a similar problem to do (for deployments, stateful sets, etc).  Timing will always be an issue - but nodes can fail at any arbitrary point, and we don't promise that the controller holds locks on nodes in general.  The tradeoff with the timeout is that you have to make a choice between forward progress and avoiding disrupting users.  That's something totally under the control of the pod today, without the daemonset being involved.

The general reason why the sig pushed back on this is that we have a general pattern that you can use today to solve the problem, and so rushing a new strategy change in to 1.9 isn't necessary.  It's desirable that other approaches be introduced (i think it's a reasonable addition), but there is no urgency to it that requires a 1.9 change.

diegs:
@smarterclayton I was not aware of that. Is it still valid now that #50477 has been resolved?

smarterclayton:
kubernetes/kubernetes#50477 closes the door for the use case you describe - but the exclusion principle that PR enforced should probably be a tunable on the DaemonSet, and is really orthogonal to update strategy.  The existing example of this is StatefulSet - we have a PodManagementPolicy that is orthogonal to the update policy for this reason:

```
	// podManagementPolicy controls how pods are created during initial scale up,
	// when replacing pods on nodes, or when scaling down. The default policy is
	// `OrderedReady`, where pods are created in increasing order (pod-0, then
	// pod-1, etc) and the controller will wait until each pod is ready before
	// continuing. When scaling down, the pods are removed in the opposite order.
	// The alternative policy is `Parallel` which will create pods in parallel
	// to match the desired scale without waiting, and on scale down will delete
	// all pods at once.
```

I think it's a reasonable ask for the daemonset to have an analogue to StatefulSets's pod management policy (in fact, it's arguable that it should be required because both stateful set and daemon set need to be correct with regards to exclusion).

The daemonset replacement policy should respect the pod management policy when making decisions about when to place pods on nodes - at most one, at least one, at most one running, etc.  That's a minimal change that would probably be more tailored to the immediate need.

I might argue that a strategy should take its cues from such a setting, not the other way around.

smarterclayton:
To expand a bit: every pod has a phase change on deletion:

1. running
2. graceful termination
3. terminated
4. removed from etcd

ReplicaSets create pods at 2 always because they are replicas (we provide "at-least 1" semantics).  StatefulSets create pods at step 4 because they are strongly tied to pod identity (we provide "at-most 1" semantics).  DaemonSets are in the middle - they are not replicas, but they are not pod identity.  Instead, they are tied to node-identity.  The real decision that a user has to make is whether they can tolerate at-most-1 or at-least-1.  The issue triggered by kubernetes/kubernetes#50477 was a user who cared about at-most-1.  This issue was triggered by a user who needs at-least-1.  Both are valid use cases.

The *behavior* in a distributed system that is required to guarantee at-most-1 is pretty aggressive - I don't think we currently guarantee that with daemonsets (for instance, just run two controller managers at the same time and they will race to create multiple pods on the node).  So the user in 50477 is almost certainly in for a rude awakening at some point when they *depend* on at most one and we don't provide it.  In a sense, that's a violation of #124 and we should fix that as soon as we can.

I think we probably need to either commit to giving the users a choice, or to EXPLICITLY always be at-least-1 so that users aren't surprised when we pretend to be at-most-1 but then fail to do so.

The upgrade strategy is subordinate to this choice - a rolling update strategy has to respect at most 1 or at least 1, but is probably free to do the right thing as long as the guarantee isn't violated.  Any workload that supports at-least-1 probably wants the surge behavior on rolling update, no matter what.

jbeda:
I'm not totally read in to this stuff but the `at-most-1` vs. `at-least-1` seems to be a useful way to break this down.

For `at-least-1` we'd want to have some patterns well documented on how to do a graceful handoff from the old thing to the new thing.  That could be just a graceful shutdown or it could also be a migration of data to the new thing to keep state around and alive.  This is a bit of a "live migration" type of situation.

diegs:
Yes, I was under the assumption that no part of k8s guaranteed `at-most-1`, and was surprised by  kubernetes/kubernetes#50477. I completely agree that it will probably be a difficult constraint to truly guarantee in perpetuity.

I also agree that it's orthogonal to the strategy; `maxUnavailable > 0` with `maxSurge = 0` will terminate pods before scheduling new ones, and `maxUnavailable = 0` with `maxSurge > 0` will to schedule new pods before terminating old ones; whether there are actually 0, 1, or 2 pods running in either scenario is not guaranteed.

Going back to this proposal, the `maxSurge` parameter in the strategy as I implemented it was not affected by kubernetes/kubernetes#50477. But it goes back to what I said about preferring explicitly implementing the semantics we want instead of relying on implementation details that may change. In the surging strategy we will always *attempt* to schedule a new pod before setting the old one to terminating, so that gives a higher chance of success than waiting for the old pod to be in a terminating state before scheduling the new one, as is the behavior with the existing strategy.

smarterclayton:
If we decide that daemonsets should be at-least-1, then there is no real need for the existing update strategy to be doing what it is doing today.  In which case surge should be part of the existing strategy like ReplicaSets and we don't need or want a new strategy, we just want rolling update to do the right thing.

If we are going to support at-most-1 we need to decide on that before we go to v1.

I would argue anyone can implement at-most-one on a node by using a shared filesystem lock.  While that requires a shared hostPath volume, it's a common daemonset use case.  Therefore, that's an argument that we should be at-least-one because we support a workaround.

What's clear is that this discussion blocks taking daemonset to GA.  We need to follow up and ensure we reach closure on something this fundamental before going GA.

diegs:
@smarterclayton agree that this could be implemented by adding `maxSurge` to the existing strategy; it was not to retain backwards-compatibility for validation.

The underlying motivation here was to achieve parity with the Deployment rolling update strategy, which has both params.

smarterclayton:
Two clarifications:

1. Stateful sets guarantee at most one pod per identity and do so safely.  We guarantee at most one pod with a given name is running on the cluster at any one time.
2. Safety rules and ensuring user data is safe trumps backwards compatibility.  If we determine our current semantics are confusing users that they get at most 1, it's within our remit to break users in order to make them safe.

The best thing now is for interested parties to make a case for either at-most-1 as a feature or as a supported workaround as Joe noted.  That will clarify next steps - reverting / altering 50477, an API change to DS to let a choice be possible, or delaying DS GA until we can make at-most-1 correct.

janetkuo:
FWIW https://github.com/kubernetes/kubernetes/issues/50477 is a regression (caused by https://github.com/kubernetes/kubernetes/pull/43077). DaemonSet controller had `at-most-1` logic implemented long before https://github.com/kubernetes/kubernetes/issues/50477. DaemonSet controller will not create DaemonSet pod on a node if the previous one is not gone. 

StatefulSet uses pod name to guarantee at-most-1 pod per index. A possible way to guarantee at-most-1 per node for DaemonSet is to use node name as part of DaemonSet pod's name, but the pod name might become too long.

kow3ns:
I don't think this decision should gate v1. DaemonSet controller was always designed to attempt at most one semantics, but I believe we could loosen or tighten this constraint as a feature, post GA, without breaking backwards compatibility.

For at most one:
1.  I'm not entirely convinced that trying to tighten the constraint (e.g. by incorporating the Node name in the generated Pod name) would be a breaking change if we turned this on in the general case.
    1. The general expectation is that the controller is already trying to achieve at most one.
    1. Name generation for DaemonSet Pods is random, so if we use the Node name to make the Pods Node unique it doesn't violate users' expectations (for DaemonSets the name of a Pod is an implementation detail, not a conformance requirement).

1. If we think it is a breaking change, than we could make it opt in via the addition of a flag on the DaemonSet Spec (e.g. `spec.nameGeneration=Random|Node`) where the default uses random name generation.

1. As pointed out above, applications can use advisory file system locks for tighter guarantees of uniqueness to achieve stricter semantics where necessary, so I don't think we have to provide stronger guarantees as a prerequisite for promotion.

For at least once
1. We can modify the update strategy to use maxSurge, or add a new UpdateStrategy. In either case the new behavior can be implemented without breaking backward compatibility. 

Also, We should not revert [#50477](https://github.com/kubernetes/kubernetes/issues/50477). This reverted a change that introduced a bug where DaemonSet controller would launch Pods while a current Pod was still terminating. Reverting this will not implement maxSurge behavior.

smarterclayton:
> DaemonSet controller was always designed to attempt at most one semantics, but I believe we could loosen or tighten this constraint as a feature, post GA, without breaking backwards compatibility.

Looking at the code we'd need to tighten at least a few things to guarantee that (as you mentioned with node name).  But I am a -1 on declaring something GA that has even the potential to violate a user assumption or violate pod safety guarantees, and that's where we are today.  My working assumption was that daemonset is at-least-one because we are using generateName, and you can't use generateName and get at-most-one.

> We can modify the update strategy to use maxSurge, or add a new UpdateStrategy

at-most-1 or at-least-1 is not about update strategy.  It's a spec property independent of update.  In the absence of an update during a drain or eviction the daemonset should follow the same property.  It's fundamental to daemonset.

> Also, We should not revert #50477

Agree, but not-quite at-most-1 is also not the right thing to ship.  By revert I meant, "don't almost give at-most-1 as the user who requested #50477 raised".

> The general expectation is that the controller is already trying to achieve at most one.

I think cluster-lifecycle's position is that it's actually not what they need, and other issues (cursory search includes https://github.com/kubernetes/kubernetes/issues/51974 and https://github.com/kubernetes/kubernetes/issues/48307) makes it less than clear.  Looking at my own use cases, I have a small number that want at-most-one and a much larger number that want at-least-one for rolling update.

> As pointed out above, applications can use advisory file system locks for tighter guarantees of uniqueness to achieve stricter semantics where necessary, so I don't think we have to provide stronger guarantees as a prerequisite for promotion.

I don't think this is the right mindset for providing an API to users based on our previous history here on statefulsets and deployments.  Providing an ALMOST correct guarantee creates a scenario where users expect it to work except when it doesn't.  That means it fails at the worst time and either eats data or blows up a system.  Either we implement at-most-1, or we implement at-least-1 and tell people how to get at-most-1.  We've never shied away from doing the right thing for workload APIs before, not sure why we should do it here.

I'd probably be ok with a v1 that did at-least-1 that could be turned to almost-at-most-1 with caveats and warnings, where in the future we'd implement correct at-most-1.  I can't reconcile shipping almost-at-most-1 only given #124 and history with statefulset.

bgrant0607:
Quick comment: Any application that really needs at most one on a node (e.g., because multiple instances would otherwise trash persistent data on local disk) absolutely has to use a file lock or other OS locking mechanism. We should document that. K8s cannot and should not provide application-level exclusion/handshaking.

kow3ns:
> I'd probably be ok with a v1 that did at-least-1 that could be turned to almost-at-most-1 with caveats and warnings, where in the future we'd implement correct at-most-1. I can't reconcile shipping almost-at-most-1 only given #124 and history with statefulset.

1. What almost at most one means right now is that DaemonSet respects the termination grace period of the Pods that it created. Currently the DaemonSet controller will not launch a new Pod until any other Pod, that it has created on the same node, is completely deleted. A failure to correctly implement this was the root cause of [#50477](https://github.com/kubernetes/kubernetes/issues/50477). 

1. As @smarterclayton points out above, this behavior is important for DaemonSets that use scarce resources as a new Pod may interfere with the termination lifecycle of the current Pod, and this can occur during unplanned termination (e.g. eviction due to taints). I don't think that it ever makes sense to modify this behavior for DaemonSet Pods during unplanned disruptions.

1. For this proposal, and for [#51974](https://github.com/kubernetes/kubernetes/issues/51974) ([#48307](https://github.com/kubernetes/kubernetes/issues/48307) is about the interaction between Deployments and PDBs and I don't see why it's relevant here), the best way to deal with port conflicts would be to implement something like [Node Local Services](https://github.com/kubernetes/kubernetes/issues/28610) first. If we don't do this, I don't see how any modification to the existing strategy, or any new strategy that attempts to launch more than one Pod per Node, will provide value in the general case (i.e. It might work for self hosting, but it will not solve [#51974](https://github.com/kubernetes/kubernetes/issues/51974)).

1. It sounds like we have broad agreement that advisory filesystem locking, or some other locking method, must be used to achieve strong guarantees about at most one semantics. If we want to give users the option of having DaemonSet controller generate a Node unique name to preclude the controller from creating two Pods on the same Node, we can do so using an addition to the spec as discussed above, but if at most one is critical, implementing that assurance via local, application level coordination will always provide a more robust solution.

1. The DaemonSet [feature](https://github.com/kubernetes/kubernetes/issues/1518) was opened in Sept 2014, though it has evolved, its general design constraints haven't changed significantly since then. Given the number of [addons](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons) that are currently using DaemonSet, and that we are even considering what needs to be done for self hosting, I think we have a strong signal, from our own usage, that the current implementation is mature enough for promotion. We can do (3), (4), or (3) and (4) after GA promotion without breaking backward compatibility.

smarterclayton:
> What almost at most one means right now is that DaemonSet respects the termination grace period of the Pods that it created

This is not at-most-one (to be clear for other participants in this thread) because any interruption or race in the controller can and will spawn two containers.  This is "best effort at most one".  

> this behavior is important for DaemonSets that use scarce resources as a new Pod may interfere with the termination lifecycle of the current Pod, and this can occur during unplanned termination (e.g. eviction due to taints

I don't follow.  Nodes don't release resources until a pod is stopped, so a new pod launched by the daemonset to replace an evicted pod will be held up by resource release in the kubelet (or should, since this is a condition only the node can validate).

> #48307 specifically comments from a user:

> Another thing is a planned maintenance on a node: it should be fairly simple to make sure extra pods are started elsewhere before we shut down this node for maintenance

There is a reasonable expectation among a wide range of kubernetes users I have interacted with that kubernetes should work to avoid application downtime by being aggressive about scale up.  Users have the same expectation with daemon sets.

> Port conflicts ... If we don't do this, I don't see how any modification to the existing strategy, or any new strategy that attempts to launch more than one Pod per Node, will provide value in the general case

Port conflicts are irrelevant since they can be avoided by the user omitting the port definition on the pod.  A user today can remove the port conflict and do port handoff themselves with SO_REUSEPORT, but the current behavior of the daemonset prevents that.  I don't know of any other conflict that impacts normal users, and omitting the pod's port definition is not burdensome to users if they can achieve availability.  So the blocker is still the daemonset controller imposing at-most-one-ish.

> At most one through locking ... more robust

The point though is that giving the appearance of a robust solution deters users from being actually robust which leads to application failure.  We have documented and agreed on not violating user expectation previously.

> Given the number of addons that are currently using DaemonSet, and that we are even considering what needs to be done for self hosting...

Since the self-hosting folks are saying there's a gap, this seems a weak argument.  V1 is EXACTLY when we should be confident it works.  A perceived pressure to declare workload APIs v1 means very little to me if the semantics will lead to failure.  I don't care whether workload v1 hits in 1.9 or 1.10, I care very much whether my production users are justified in their believe that we have their best interests in mind.

kow3ns:
> This is not at-most-one (to be clear for other participants in this thread) because any interruption or race in the controller can and will spawn two containers. This is "best effort at most one".

Correct, to be clear, as stated above, its just about respecting termination grace. 

> I don't follow. Nodes don't release resources until a pod is stopped, so a new pod launched by the daemonset to replace an evicted pod will be held up by resource release in the kubelet (or should, since this is a condition only the node can validate).

If we create another Pod on the same node while the current Pod is modifying some host resource, say the file system, as part of its termination, while the new Pod is attempting to utilize that resource it is surprising to users as in (#50477)[https://github.com/kubernetes/kubernetes/issues/50477]

>  Port conflicts are irrelevant since they can be avoided by the user omitting the port definition on the pod. A user today can remove the port conflict and do port handoff themselves with SO_REUSEPORT,

The issue is that when users run a DaemonSet that provides a shared resource on the node that is utilized by many Pods (e.g. log shipping, distributed file system, metrics collection), the current [recommend](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#communicating-with-daemon-pods) way of communicating with the Pod is via a well known host port. Many applications (e.g. [HDFS](https://issues.apache.org/jira/browse/HDFS-9861) ) don't allow the user to configure SO_REUSEPORT. This means that updating the Pods in the DaemonSet is inherently disruptive to all of its client Pods. Local services would provide a layer of abstraction that allows a surge update strategy to be applied in such a way that it is not disruptive to the node local client Pods. 

> The point though is that giving the appearance of a robust solution deters users from being actually robust which leads to application failure. We have documented and agreed on not violating user expectation previously.

Agree that we need to clearly document the behavior and limitations, and if that documentation is presently lacking, no matter what else we do, we need to correct it for users of the current implementation, but I don't see how promoting what is currently implemented, documented, and in use on production clusters can violate the users expectations.

> V1 is EXACTLY when we should be confident it works. 

I strongly agree with this. Which is why I think we should not modify the existing semantics that have been in place since the original design, or ship something drastically different than the DaemonSet that we use for our addons and that users have come to rely on in their production systems. 

>  Since the self-hosting folks are saying there's a gap, this seems a weak argument.

After talking to the self hosting folks, they indicated they can achieve their objective by creating a new DaemonSet prior to deleting the previous one. This is the way that many users achieve blue-green for DaemonSets and StatefulSets today. Given that those users have a path forward, and that we can add this feature later to further improve their experience, I think we should prioritize providing our users, and our own cluster addons, stability over new features.

smarterclayton:
> If we create another Pod on the same node while the current Pod is modifying some host resource, say the file system, as part of its termination, while the new Pod is attempting to utilize that resource it is surprising to users as in #50477.

I agree that user surprise should be avoided, and I might even agree that the correct default for DS is at-most-one.  On the other hand, we often slap people in Kubernetes hard the first time they use a particular controller specifically so that they know what they can rely on (the hard stop on graceful termination, the crash loop, the liveness check, replica scale out) up front without finding out later.  And still, the number of users that assumed that replica set with scale = 1 means "at most one" is terrifying, but all of them were at least guided to the correct knowledge when their databases didn't scale up because the old pod was tearing down, or they saw two pods taking traffic.

If we had a really clear path to at-most-one correctly in the near term that wouldn't break backward compatibility and we were positive wouldn't require API changes to achieve, I'm less worried.  I'd like to spend our time trying to figure out how that can be provided.

> Local services would provide a layer of abstraction that allows a surge update strategy to be applied in such a way that it is not disruptive to the node local client Pods.

Yeah, I'm just mostly thinking of things using host port networking or are the ones providing services (i.e. kube-proxy, dns, sdn) that can't rely on that.

> I don't see how promoting what is currently implemented, documented, and in use on production clusters can violate the users expectations.

If the pod watch is delayed and a resync is fired, it looks like the daemonset controller can create two pods on a node (from a cursory look and previous controller experience).  Is that incorrect?  Delayed pod watches have occurred in the last several etcd versions due to various bugs, and we have stomped several bugs in caches where they can also occur (beyond just normal distributed systems).  If I have a pod that I have written that assumes that only one pod runs on a node at a time, nothing in Kube today pushes me to use filesystem locking because I'm trusting kube.

> indicated they can achieve their objective by creating a new DaemonSet prior to deleting the previous one

I'm fine with that.  I'm still concerned about the previous point.

smarterclayton:
It looks like all the protects the daemonset from creating multiple pods on a node is the expectation system, which is single process based.

1. new node is added at RV=9
2. controller creates pod at RV=10
3. controller crashes
4. new controller pod cache syncs to pods at RV=9 (this happens all the time for multiple reasons in the api server)
5. controller assumes pod caches are up to date, has no expectations, creates a second pod at RV=11

mattfarina:
@smarterclayton we discussed this in SIG Apps. The advice at this point was to take it to SIG Architecture in office hours this week. Would be nice to get some unblocking decisions there.

smarterclayton:
Spawned the last part (daemonset not being resistant to controller restart and slow caches) as https://github.com/kubernetes/kubernetes/issues/54096 and summarized the discussion here.  

Otherwise, this is on hold for 1.9 given sig-cluster-lifecycle having an alternative (correct?).

diegs:
@smarterclayton yes, we discussed in sig-cluster-lifecycle today and the plan is to move forward with an alternative.

Personally, I'm in agreement that coming to a conclusion on the supported semantics of daemonsets should precede this discussion; once that is done then it would be nice to see if there is room for a `maxSurge` parameter in the rolling update strategy. For example, if `maxUnavailable > 0 && maxSurge == 0` then you get at-most-one semantics, if `maxUnavailable == 0 && maxSurge > 0` you get at-least-one semantics.

Once that is settled other issues brought up here (scarce resources such as hostPorts, etc) can be addressed.

contributors/design-proposals/daemonset-create-first-rolling-update.md
```diff
@@ -0,0 +1,182 @@
...
+default in 1.9+), so it may be possible to add the `maxSurge` strategy to
+`RollingUpdate` and deprecate `SurgingRollingUpdate` then, or make the change
+now since the `v1beta2` API is incoming. However, to be conservative it was
+decided to implement this as a separate strategy.
+
+For more background see https://github.com/kubernetes/kubernetes/issues/48841.
+
+### Considerations / questions
+
+1. How are `hostPort`s handled?
```
discordianfish: Since such use case brought me here I'm wondering if my use case and similar onces could be solved by using a special form of NodePort that doesn't redirect to any pod but only pods on the same node. This would avoid having to use hostPort in many cases where all you want is one pod per node that is reachable on a well known port.

I know this is beyond the scope of this, but knowing how a solution to this hostPort scenario likely would look like in the future might influence this design here.

zoidbergwill: In [1.5 and newer](https://v1-6.docs.kubernetes.io/docs/tutorials/services/source-ip/) you can use the `service.beta.kubernetes.io/external-traffic=OnlyLocal` annotation, and it seems like in [newer versions](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-typenodeport) you can do `kubectl patch svc nodeport -p '{"spec":{"externalTrafficPolicy":"Local"}}'` to only allow local traffic.

diegs:
Closing this because I'm no longer working on it and I'm not aware of what the expected future will be. We can re-open or re-visit at a later time if needed.

krmayankk:
Why is this being abandoned ? seems like adding maxSurge would be ideal and at parity with Deployments

curlup:
I would like to know what should be done to put that suggestion back on the table?
Can we reopen it or something?


As for concrete user examples I would like to suggest this.
A monitoring agent on a node — checking health of all the hardware or/and services as well. It seems that DaemonSet is the right choice for that.
But having (for a short period of an update) 2 agents on a node, checking and reporting, is much better than, no agent reporting for a period. 

In a matter of HA for such agent, it event might be desirable to have 2 agents on a node constantly, in order to mitigate that agent unavailability or crash.

captn3m0:
We have another usecase of running ingress controller as a daemonset on a separate set of nodes. Version Updates should not result in a scenario where a node does not have a single pod running, but it currently does.

PR #989, opened 2017-08-26
1.7 Local Persistent Volume design doc

msau42:
Migrating the design doc in google docs to Markdown, and updating it to reflect the implementation we did for 1.7.

Same sections:
* Background
* Use Cases
* Environments
* Goals/non-goals
* Plugin design
* Initial volume config

Updated sections:
* Plugin API changes
* Provisioner
* Test Cases
* Feature Plan

New sections:
* Node affinity
* Provisioner boostrapper
* 1.7 Limitations
* 1.9 Outline


contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+
+The name of the PV needs to be unique across the cluster.  The provisioner will hash the node name,
+StorageClass name, and base file name in the volume path to generate a unique name.
+
+##### Packaging
+
+The provisioner is packaged as a container image and will run on each node in the cluster as part of
+a DaemonSet.  It needs to be run with a user or service account with the following permissions:
+
+* Create/delete/list/get PersistentVolumes - Can use the `system:persistentvolumeprovisioner` ClusterRoleBinding
```
liggitt: This gives every node write access to all PVs, which is not intended

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+
+##### API Changes
+
+For the initial alpha phase, node affinity is expressed as an optional
+annotation in the PersistentVolume object.
+
+```
+// AlphaStorageNodeAffinityAnnotation defines node affinity policies for a PersistentVolume.
+// Value is a string of the json representation of type NodeAffinity
+AlphaStorageNodeAffinityAnnotation = "volume.alpha.kubernetes.io/node-affinity"
```
liggitt: is the plan to migrate the annotation to a field on the PV spec or drop the alpha annotation when the field is added?

zhouhaibing089:
cc @chengyli @wenlxie

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+Local PVs are statically created and not dynamically provisioned for the first phase.
+To mitigate the amount of time an administrator has to spend managing Local volumes,
+a Local static provisioner application will be provided to handle common scenarios.  For
+uncommon scenarios, a specialized provisioner can be written.
+
+The Local static provisioner will be developed in the
+[kubernetes-incubator/external-storage](https://github.com/kubernetes-incubator)
+repository, and will loosely follow the external provisioner design, with a few differences:
+
+* A provisioner instance needs to run on each node and only manage the local storage on its node.
```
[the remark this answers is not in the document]
msau42: The permissions are handled through ClusterRoleBinding.  But I am unaware of a way to have a different ClusterRoleBinding for each instance of a DaemonSet so that it only can see the PVs for that node.

Each provisioner instance will filter the PVs it handles in its code, but the PV informer itself is not filtered.  Each instance gives itself a name of "local-provisioner-<node uid>", and marks every PV It creates with that name in the "provisioned-by" annotation (part of the general [external provisioner](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/volume-provisioning.md)  design).  And it filters the PVs on that annotation.

If we want to filter at the informer level, I could also label the PVs with the provisioner name and filter on labels.  But that's still just relying on the implementation to correctly filter itself.  I would be open to any suggestions you may have.

contributors/design-proposals/local-storage-pv.md
msau42: This is true of any external-provisioner in general.  The external provisioner design relies on the provisioner implementation to filter the PVs and only operate on PVs that it created by marking and checking the "provisioned-by" annotation.  Perhaps that design needs to be revisited so that you could create a rolebinding on a reduced set.

liggitt: Correct, but typically an external provisioner would not be required to run on all nodes, so you could avoid running it on a node containing untrusted workloads

contributors/design-proposals/local-storage-pv.md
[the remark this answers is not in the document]
msau42: Corrected the above comment to add nodeUID to the name.

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+* Allow pods to access local volumes without needing to undestand the storage
+layout on every node.
+* Persist local volumes and provide data gravity for pods.  Any pod
+using the local volume will be scheduled to the same node that the local volume
+is on.
+* Allow tolerations on volumes to be specified to release the volume during
+failure conditions, such as node, storage or scheduling failures.
+* Allow pods to specify local storage as part of a Deployment or StatefulSet.
+* Allow Administrators to set up and configure local volumes with simple methods.
+* Do not require qdministrators to manage the local volumes once provisioned
```
xiang90: qdministrators -> administrators

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+* Allow pods to mount any local block or filesystem based volume.
+* Allow pods to mount dedicated local disks, or channeled partitions as volumes for
+IOPS isolation.
+* Allow pods do access local volumes without root privileges.
+* Allow pods to access local volumes without needing to undestand the storage
+layout on every node.
+* Persist local volumes and provide data gravity for pods.  Any pod
+using the local volume will be scheduled to the same node that the local volume
+is on.
+* Allow tolerations on volumes to be specified to release the volume during
```
xiang90: not sure i understand this. can we explain this more clear?

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+persistent volumes where the data can persist beyond the lifetime of the pod.
+
+There is increasing demand for using local storage with persistent volumes,
+especially for distributed filesystems and databases such as GlusterFS and
+Cassandra.  The main motivations for using local storage include:
+
+* Performance:  Local SSDs achieve higher IOPS and throughput than a cloud
+provider’s remote storage.
+
+* Cost: High-performance networked storage is expensive to setup and maintain.
```
xiang90: if the upper layer application already does replication (and is not file/block level replication aware), it usually makes sense to just use local disk and let the application take over replication instead of using replicated block device.

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+* Allow pods do access local volumes without root privileges.
+* Allow pods to access local volumes without needing to undestand the storage
+layout on every node.
+* Persist local volumes and provide data gravity for pods.  Any pod
+using the local volume will be scheduled to the same node that the local volume
+is on.
+* Allow tolerations on volumes to be specified to release the volume during
+failure conditions, such as node, storage or scheduling failures.
+* Allow pods to specify local storage as part of a Deployment or StatefulSet.
+* Allow Administrators to set up and configure local volumes with simple methods.
```
zhangxiaoyu-zidif: Administrators  -> administrators

contributors/design-proposals/local-storage-pv.md
msau42: The default behavior with local storage is that if the node goes down, your pod cannot be rescheduled because you're bound to that local volume.  If your application uses replication, it can potentially be rescheduled, on a brand new node, with a new volume and then resync the replica.  So the idea of tolerations here is similar to the pod tolerations feature, except at a volume level.  The tolerations setting will determine how long we wait on specific error conditions (like node unavailable) before we unbind the PVC to find a new one.

contributors/design-proposals/local-storage-pv.md
msau42: Good point.  I think I mentioned it already in the distributed db and fs use case, but will add it here too.

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+
+### Phase 2
+
+* Block devices and raw partitions
+* Smarter PV binding to consider local storage and pod scheduling constraints
+
+### Phase 3
+
+* Support common partitioning patterns
+* Volume forgiveness
```
xiang90: what is this? needs more details or link to another doc?

msau42: This is the tolerations idea I mentioned above.  I'm leaving the design details to a future release when we actually have time to work on it.  There's too many features needed for local volumes to design it all in one go.

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,631 @@
...
+In the above example, the generated `mountDir` is `/mnt/local-storage/mnt ~ssds` and
+`/mnt/local-storage/mnt~hdds`, respectively.
+
+
+#### Limitations
+
+This initial phase for Local PV support will provide the following:
+
+* Local directories to be specified as Local PVs with node affinity
+* Pod using a PVC that is bound to a Local PV will always be scheduled to that node
```
xiang90: what is the alternatives? why this is an limitation?

contributors/design-proposals/local-storage-pv.md
xiang90: ok. makes sense. basically pod just wants a local pv, but can forgive its state. maybe we can make this clear in the doc.

contributors/design-proposals/local-storage-pv.md
xiang90: thanks.

contributors/design-proposals/local-storage-pv.md
msau42: Sorry i need to organize this section better.  I wanted to start off with the use cases that are supported in 1.7, and then follow up with the limitations in 1.7.  So this first section is describing what a user can do.

contributors/design-proposals/local-storage-pv.md
msau42: What about something like this: Have a single master provisioner that manages and operates on the PV objects.  Have it talk to worker pods on each node.  Each worker pod is only allowed to:

1. Report what local volumes it discovered.
2. Handle cleanup of its local volumes when requested by the master.
3. In the future, provision a new volume from a shared pool when requested by the master.

This way all the PV objects are managed by a single pod that can run on a trusted node, and each worker pod can only touch the local volumes on it's node.  The worker pods don't even need to look at or operate on PV objects.

xiang90:
@msau42 Thanks for the doc! Looks pretty solid to me in general.

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,632 @@
...
+dedicated disks; others may only need small partitions and are ok with sharing disks.
+Instead of forcing a partitioning scheme on storage administrators, the Local volume
+is represented by a path, and lets the administrators partition their storage however they
+like, with a few minimum requirements:
+
+* The paths to the mount points are always consistent, even across reboots or when storage
+is added or removed.
+* The paths are backed by a filesystem (block devices or raw partitions are not supported for
+the first phase)
+* The storage is mounted according to the provisioner discovery scheme defined below
```
ddysher: Is this a minimal requirement?  My understanding is that admin can use local volume without provisioner so doesn't have to conform to provisioner discovery scheme.

contributors/design-proposals/local-storage-pv.md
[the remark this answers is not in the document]
ddysher: Is it possible to extend [node authorization](https://kubernetes.io/docs/admin/authorization/node/)?

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,632 @@
...
+To mitigate the amount of time an administrator has to spend managing Local volumes,
+a Local static provisioner application will be provided to handle common scenarios.  For
+uncommon scenarios, a specialized provisioner can be written.
+
+The Local static provisioner will be developed in the
+[kubernetes-incubator/external-storage](https://github.com/kubernetes-incubator)
+repository, and will loosely follow the external provisioner design, with a few differences:
+
+* A provisioner instance needs to run on each node and only manage the local storage on its node.
+* For this phase, it does not handle dynamic provisioning.  Instead, it performs static provisioning
```
ddysher: s/this phase/initial phase

The doc will keep updating, so `this phase` can be ambiguous :)

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,632 @@
...
+
+PV deletion will operate on the Update event.  If the PV it provisioned changes to the “Released”
+state, and if the reclaim policy is Delete, then it will cleanup the volume and then delete the PV,
+removing it from the cache.
+
+PV creation does not operate on any informer events.  Instead, it periodically monitors the discovery
+directories, and will create a new PV for each path in the directory that is not in the PV cache.  It
+sets the "pv.kubernetes.io/provisioned-by" annotation so that it can distinguish which PVs it created.
+
+For this phase, the allowed discovery file types are directories and mount points.  The PV capacity
```
ddysher: ditto

contributors/design-proposals/local-storage-pv.md
liggitt: central management of PV resources coordinated with local management of local volumes, where the central manager only pays attention to what the local worker reports for volumes that are actually bound to the node seems better

contributors/design-proposals/local-storage-pv.md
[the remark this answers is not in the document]
liggitt: to do that, several things would be required:
* the instance on each node would have to run with the node's credentials (not as a daemonset, etc)
* there would have to be strong ties from a PV to a particular node (not just a label selector or annotation)

at first glance, a division of responsibilities like https://github.com/kubernetes/community/pull/989#discussion_r135700184 seems like it might work better

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,632 @@
...
+for a node.
+
+## Non-Goals
+
+* Provide data availability for a local volume beyond its local node.
+* Support the use of HostPath volumes and Local PVs on the same volume.
+
+## Background
+
+In Kubernetes, there are two main types of storage: local and remote.
```
ianchakeres: How about switching these around and a few other edits...
"""
In Kubernetes, there are two main types of storage: remote and local.

Remote storage is typically used with persistent volumes where the data persists beyond the lifetime of the pod.

Local storage may desirable for ephemeral use where the data persists only during the lifetime of the pod. Local storage is also desirable for distributed storage systems, such as GlusterFS and Cassandra.

The main motivations for using local storage include:

Performance:  Local SSDs can achieve higher IOPS and throughput than remote storage.

Cost: Operational costs may be reduced by leveraging existing local storage, especially in bare metal environments. Network storage can be expensive to setup and maintain, and it may not be necessary for certain applications. 

"""

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,632 @@
...
+
+## Feature Plan
+
+A detailed implementation plan can be found in the
+[Storage SIG planning spreadhseet](https://docs.google.com/spreadsheets/d/1t4z5DYKjX2ZDlkTpCnp18icRAQqOE85C1T1r2gqJVck/view#gid=1566770776).
+The following is a high level summary of the goals in each phase.
+
+### Phase 1
+
+* Support Pod, Deployment, and StatefulSet requesting local storage
```
ianchakeres: + for a single local volume.

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,632 @@
...
+* Support pre-configured, statically partitioned, filesystem-based local storage
+
+### Phase 2
+
+* Block devices and raw partitions
+* Smarter PV binding to consider local storage and pod scheduling constraints
```
ianchakeres: add: e.g. multiple local volumes per pod.

contributors/design-proposals/local-storage-pv.md
msau42: done

contributors/design-proposals/local-storage-pv.md
msau42: done

contributors/design-proposals/local-storage-pv.md
msau42: done

contributors/design-proposals/local-storage-pv.md
msau42: Yes, thanks for catching, will remove.

contributors/design-proposals/local-storage-pv.md
msau42: done

contributors/design-proposals/local-storage-pv.md
msau42: done

contributors/design-proposals/local-storage-pv.md
msau42: Split up this section into deliverables and limitations.

contributors/design-proposals/local-storage-pv.md
msau42: I updated this section to mention that it will be addressed in a future release, and added a high level overview at the bottom.

contributors/design-proposals/local-storage-pv.md
msau42: Agree, I tried to reword it to make it sound scarier.

contributors/design-proposals/local-storage-pv.md
```diff
@@ -0,0 +1,659 @@
...
+The basic design of the provisioner will have two separate handlers: one for PV deletion and
+cleanup, and the other for static PV creation.  A PersistentVolume informer will be created
+and its cache will be used by both handlers.
+
+PV deletion will operate on the Update event.  If the PV it provisioned changes to the “Released”
+state, and if the reclaim policy is Delete, then it will cleanup the volume and then delete the PV,
+removing it from the cache.
+
+PV creation does not operate on any informer events.  Instead, it periodically monitors the discovery
+directories, and will create a new PV for each path in the directory that is not in the PV cache.  It
```
hongchaodeng: What happens if no sub-dir exists?

msau42: nothing happens as we don't detect any volumes.

contributors/design-proposals/local-storage-pv.md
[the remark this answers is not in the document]
ddysher: hmm...  do you think this will become a general requirement where daemonset pod can be  restricted to only access local resources, not just for local volume use case?

msau42:
Can folks try to review this week?  I would like to get this in, so that we can start extending it for 1.9 features.

PR #991, opened 2017-08-29
Local ephemeral storage resource management

jingxu97:
This feature describes the support for local ephemeral storage resource
isolation


contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,580 @@
...
+    local-pv-1 100Gi      Bound     www-local-pvc-1 node-1
+    local-pv-2 10Gi       Bound     log-local-pvc-1 node-1
+    local-pv-1 100Gi      Bound     www-local-pvc-2 node-2
+    local-pv-2 10Gi       Bound     log-local-pvc-2 node-2
+    local-pv-1 100Gi      Bound     www-local-pvc-3 node-3
+    local-pv-2 10Gi       Bound     log-local-pvc-3 node-3
+    ```
+
+8. If a pod dies and is replaced by a new one that reuses existing PVCs, the pod will be placed on the same node where the corresponding PVs exist. Stateful Pods are expected to have a high enough priority which will result in such pods preempting other low priority pods if necessary to run on a specific node.
+9. Forgiveness policies can be specified as tolerations in the pod spec for each failure scenario.  No toleration specified means that the failure is not tolerated.  In that case, the PVC will immediately be unbound, and the pod will be rescheduled to obtain a new PV.  If a toleration is set, by default, it will be tolerated forever.  `tolerationSeconds` can be specified to allow for a timeout period before the PVC gets unbound.
```
hongchaodeng: Does this mean if a local PV fail, the pod will be "migrated" automatically to another node with another working local PV?

If so, I'm curious what is the reasoning for this? Why not just mark the pod as failure instead?

jingxu97: I am not sure your question, but here the case is pod failed, and we have the options to continue to run on the previous node with the same PV, or schedule to a different node with different PV.

hongchaodeng: > or schedule to a different node with different PV.

Does this happen automatically without human interaction?

jingxu97: yes, statefulset will check if a pod dies, it restart a new pod automatically.

hongchaodeng: What happens if a local PV fail? What would the pod become without any external action (e.g. stateful controller)?

jingxu97: what do you mean by local PV fail?

hongchaodeng: My fault. Let me clarify my thoughts and questions in details.

> Forgiveness policies can be specified as tolerations in the pod spec for each failure scenario.

I think this is not actually related to volume tolerations. At this moment, it can only tolerate node unready conditions. Note that node failure is different from volume failure. And this doc is about volume not about node.

Instead, I think this paragraph could be improved to talk about how to tolerate volume failure.

So on volume failure, my thought is that if the underlying PV has failed, it will have taint "storage.kubernetes.io/pvUnhealthy". Right? What happens happen? Will the related PVC change any status? Will the related pod change any status?

jingxu97: Currently we don't have a health check for local disk. The plan is to add this function and if disk fails, PV is marked as "taint". And depending on the policy, the PVC will become unbound, then kubelet should check the status and evict the pod. Statefulset controller will create a new pod which will use a different PV. 

You could check the feature on local persistent volume https://github.com/kubernetes/community/pull/989

hongchaodeng: That's great. Now I understand it.

> The plan is to add this function and if disk fails, PV is marked as "taint". And depending on the policy, the PVC will become unbound, then kubelet should check the status and evict the pod.

I think this information is important but missing in the docs. So it would be helpful to add this into the paragraph for folks reading this later.

contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,399 @@
...
+  containers and pods
+* Provide ephemeral storage requirement and garantee for containers and pods
+* Add quota management for ephemeral storage for shared root parition
+* Add limitrange support for ephemeral storage
+
+# Non Goals
+* Provide storage usage isolation for non-shared partitions.
+* Support for I/O isolation using CFS & blkio cgroups.
+  * IOPS isn't safe to be a schedulable resource. IOPS on rotational media is very limited compared to other resources like CPU and Memory. This leads to severe resource stranding.
+  * Blkio cgroup + CFS based I/O isolation doesn't provide deterministic behavior compared to memory and cpu cgroups. Years of experience at Google with Borg has taught that relying on blkio or I/O scheduler isn't suitable for multi-tenancy.
```
gnufied: daft question  - what does CFS stands for? can we perhaps expand it?

contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,399 @@
...
+* EmptyDir SizeLimit: Because empryDir volume is a pod-level resource which is managed separately, we also add a sizeLimit for emptyDir Volume for additional storage isolation. If this limit is set for emptyDir volume (default medium), eviction manager will validate this limit with emptyDir usage too in additional to the above pod-level resource constraints. 
+
+## Eviction Policy and Scheduler Predicates
+CPU and memory use cgroup for limiting the resource usage. However, cgroup for disk usage is not available. Our current design is to evict pods when exceeding the limit set for local storage. The eviction policy is listed as follows.
+* If the container writable layer (overlay) usage exceeds its limit, pod gets evicted.
+* If the emptyDir volume usage exceeds the sizeLimit, pod gets evicted.
+* If the sum of the usage from emptyDir and all contains exceeds the sum of the container’s local storage limits (pod-level limit), pod gets evicted.
+
+When scheduler admits pods, it sums up the storage requests from each container and pod can be scheduled only if the sum is smaller than the allocatable of the local storage space.
```
ddysher: Do we consider emptyDir.SizeLimit as well, not just sum of container requests?

contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Resource Quota
+Add two more resource quotas for storage. The request and limit set constraints on the total requests/limits of all containers’ in a namespace 
+ResourceRequestsStorageScratch
```
ddysher: we changed from scratch to ephemeral?
s/ResourceRequestsStorageScratch/ResourceRequestsEphemeralStorage
s/ResourceLimitStorageScratch/ResourceLimitsEphemeralStorage

Also, adding newline in markdown won't render new lines, you can change to use bullet list.

contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,399 @@
...
+    apiVersion: v1
+    kind: pod
+    metadata:
+     name: foo
+    spec:
+     containers:
+     - name: fooc
+       resources:
+         limits:
+           ephemeral-storage: 2Gi
```
ddysher: nit: can we add `volumes` section for completeness?

contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,399 @@
...
+       requests:
+         storage.kubernetes.io/logs: 500Mi
+         storage.kubernetes.io/overlay: 500Mi
+     volumeMounts:
+     - name: myEmptyDir
+       mountPath: /mnt/data
+   volumes:
+   - name: myEmptyDir
+     emptyDir:
+       sizeLimit: 2Gi
```
ddysher: Use a larger value?

contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,399 @@
...
+  apiVersion: v1
+  kind: pod
+  metadata:
+   name: foo
+  spec:
+   containers:
+   - name: fooc
+     resources:
+       requests:
+         storage.kubernetes.io/logs: 500Mi
```
ddysher: This should be removed.

contributors/design-proposals/local-ephemeral-storage.md
```diff
@@ -0,0 +1,399 @@
...
+       mountPath: /mnt/data
+   volumes:
+   - name: myEmptyDir
+     emptyDir:
+       sizeLimit: 2Gi
+  ```
+
+6. It is recommended to require `limits` to be specified for `storage` in all pods. `storage` will not affect the `QoS` Class of a pod since no SLA is intended to be provided for storage capacity isolation. It is recommended to use Persistent Volumes as much as possible and avoid primary partitions.
+
+### Alice manages a Database which needs access to “durable” and fast scratch space
```
ddysher: Should this be moved local persistent volume ?

contributors/design-proposals/local-ephemeral-storage.md
ddysher: I think this is out of scope for this design doc.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/991)

contributors/design-proposals/local-ephemeral-storage.md
jingxu97: Completely Fair Scheduler, added it

contributors/design-proposals/local-ephemeral-storage.md
jingxu97: no, we don't consider emptyDir sizeLimit. That is a separate limit we want to enforce emptyDir volume explicitly. .

contributors/design-proposals/local-ephemeral-storage.md
jingxu97: fixed

contributors/design-proposals/local-ephemeral-storage.md
jingxu97: remove this section since it is not very related

contributors/design-proposals/local-ephemeral-storage.md
jingxu97: fixed

contributors/design-proposals/local-ephemeral-storage.md
jingxu97: fixed

adohe:
@jingxu97 so this is the ephemeral storage design proposal?

jingxu97:
@adohe Yes. This feature is alpha now, plan to promote to beta in 1. 10.

PR #997, opened 2017-08-30
Flexvolume Dynamic Plugin Discovery: Added comments about possible driver installation actions

verult:
Made it explicit that driver uninstalls are allowed, and cautioned against uninstalling while a volume depends on the driver.

/sig storage
/assign @chakri-nelluri @saad-ali

### DIFF
--- a/contributors/design-proposals/flexvolume-deployment.md
+++ b/contributors/design-proposals/flexvolume-deployment.md
@@ -19,6 +19,7 @@ An automated deployment technique is discussed in [Recommended Driver Deployment
 
 When there is a modification of the driver directory, a notification is sent to the filesystem watch from kubelet or controller manager. When kubelet or controller-manager searches for plugins (such as when a volume needs to be mounted), if there is a signal from the watch, it probes the driver directory and loads currently installed drivers as volume plugins.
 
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/flexvolume-deployment.md
```diff
@@ -19,6 +19,7 @@ An automated deployment technique is discussed in [Recommended Driver Deployment
 
 When there is a modification of the driver directory, a notification is sent to the filesystem watch from kubelet or controller manager. When kubelet or controller-manager searches for plugins (such as when a volume needs to be mounted), if there is a signal from the watch, it probes the driver directory and loads currently installed drivers as volume plugins.
 
+The modification can be a driver addition (install), update (upgrade/downgrade), or deletion (uninstall). If a volume depends on an existing driver, this driver can be *updated* but not *deleted*.
```
chakri-nelluri: Minor suggestions:
I would use install/upgrade/downgrade/deletion as primary and move addition/update/deletion to brackets.
Also, change the last sentence to "It can be updated but not deleted".

verult: ACK

verult:
Addressed comments

PR #998, opened 2017-08-30
Update the docs around alpha fields & deprecating

thockin:
Fix some wording.

Remove the demand for validation to fail.  It also says to clearthose
fields, which more closely models non-existent fields.

Change the tombstone (proposed - nobody is using this yet, obviously).
This format matches what gengo uses for other tags.

### DIFF
--- a/contributors/devel/api_changes.md
+++ b/contributors/devel/api_changes.md
@@ -852,7 +852,7 @@ The preferred approach adds an alpha field to the existing object, and ensures i
     * ensure the field is [optional](api-conventions.md#optional-vs-required)
         * add the `omitempty` struct tag
         * add the `// +optional` comment tag
-        * ensure the field is entirely absent from API responses when empty (if it is a struct type, it must be a pointer)
+        * ensure the field is entirely absent from API responses when empty (optional fields should be pointers, anyway)
     * include details about the alpha-level in the field description
     
     ```go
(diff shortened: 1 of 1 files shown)


davidopp:
@bsalamat

bgrant0607:
@liggitt Some people found this section confusing.

When disabled, the behavior should be as if the field were not present. The current convention is to silently drop unknown fields. Since the field should be cleared on creation and update, it wasn't clear why the validation would still be needed and what the expected behavior was.

Some day we will hopefully have warnings and optional server-side schema validation, which will provide better ways for users to understand such cases.

contributors/devel/api_changes.md
```diff
@@ -852,7 +852,7 @@ The preferred approach adds an alpha field to the existing object, and ensures i
     * ensure the field is [optional](api-conventions.md#optional-vs-required)
         * add the `omitempty` struct tag
         * add the `// +optional` comment tag
-        * ensure the field is entirely absent from API responses when empty (if it is a struct type, it must be a pointer)
+        * ensure the field is entirely absent from API responses when empty (optional fields should be pointers, anyway)
```
liggitt: Is this true even for primitive types where omitempty/zero-values work as desired?

contributors/devel/api_changes.md
```diff
@@ -885,32 +885,30 @@ One possible place to do this is in the REST storage strategy's PrepareForCreate
...
-4. In validation, ensure the alpha field is not set if the feature gate is disabled:
+4. In validation, ignore the alpha field if the feature gate is disabled:
 
     ```go
     func ValidateFrobber(f *api.Frobber, fldPath *field.Path) field.ErrorList {
       ...
       if utilfeature.DefaultFeatureGate.Enabled(features.Frobber2D) {
         ... normal validation of width field ...
-      } else if f.Width != nil {
-        allErrs = append(allErrs, field.Forbidden(fldPath.Child("width"), "disabled by feature-gate"))
```
liggitt: This was sort of belt and suspenders to make sure alpha data isn't persisted, especially since the clearing of the field is not as automatic as we'd want yet. For fields in reused types (like PodTemplateSpec), we're currently counting on manual sweeping to catch all the handlers where we need to clear

liggitt: for example, this would help with the concern noted in https://github.com/kubernetes/kubernetes/pull/50924#issuecomment-325045360

thockin:
I don't have a string feeling, and if you think it matters, I will leave
it, but both steps are manual - why is one more likely to be done than the
other?  And why do we suggest PrepareFor... rather than defaulting logic?
Defaulting is at the type level, regardless of resource.

liggitt:
> why is one more likely to be done than the other?

For fields in reused types (like PodTemplateSpec and descendants) the common validation is called more consistently

> And why do we suggest PrepareFor... rather than defaulting logic?

Defaulting is still built into clients, where we don't have good feature flag support yet.

thockin:
Defaulting also happens on server-side, so even if they default it to a
value client-side, the server-side gate will kick in and can clear the
field?  Do we want it to behave as if the field doesn't exist or as an
error?  I assumed the former..

liggitt:
> Defaulting also happens on server-side, so even if they default it to a value client-side, the server-side gate will kick in and can clear the field? Do we want it to behave as if the field doesn't exist or as an error? I assumed the former..

The issue with defaulting is that if we clear there, that also happens client side in some cases, and you'd have a hard time sending alpha fields to a server that had the alpha feature enabled.

A lot of this boils down to not having the auto-clearing in the generated codecs that can be selectively enabled server-side, so we're trying to find the least-bad places to put it in the meantime.

thockin:
Ahh, good point wrt defaulting.  I can leave the guidance here to force
validation, if you prefer.  I'll just add words about why.  Your call.

liggitt:
Explaining the double checking in validation would be good ("this *should* have already been cleared, but in case it wasn't or the type was used in another object whose strategy didn't correctly clear alpha data...")

contributors/devel/api_changes.md
thockin: yes.  Conventions doc says so, and the consistency is valuable.

thockin:
Updated, PTAL

liggitt:
/lgtm
thanks

bgrant0607:
Thanks.
/approve

PR #1017, opened 2017-09-04
kubectl: Dynamic Resource/Shortcut Fecthing

shiywang:
@kubernetes/sig-cli-maintainers @juanvallejo @tcharding @liggitt @deads2k PTAL

### DIFF
--- a/contributors/design-proposals/sig-cli/dynamic_resource_shortcut_fecth.md
+++ b/contributors/design-proposals/sig-cli/dynamic_resource_shortcut_fecth.md
@@ -6,15 +6,16 @@ Implementation Owner: @shiywang
 
 ## Motivation
 
-Previously, when a new resources added to server, developer should add those resources/shortcuts name into two places of kubectl: here and there. then kubectl get/describe/explain... -h will show the new added hard code resources/shortcuts, But some times people always forget to add those things, and we have to submit new pr to fix that everytime.
+Previously, when a new resource added to server, developer should add those resources/shortcuts name into two places of kubectl: [here](https://github.com/kubernetes/kubernetes/blob/357db0c39c7203809b369516c0ec93831ead8649/pkg/kubectl/cmd/cmd.go#L201) and [there](https://github.com/kubernetes/kubernetes/blob/f9913c4948038b9b51f2342134d546c6bb74e7a3/pkg/kubectl/kubectl.go#L46). then `kubectl get/describe/explain... -h` will show the new added hard code resources/shortcuts, But some times people always forget to add those things, and we have to submit new pr to fix that everytime.
 
 ## Proposal
 
-We should eliminate those every time add new hardcoded resources in kubectl, improve development experiences, but for how to prompt those information proply, there's two solutions here: 
+We should eliminate those every time add new hardcoded resources in kubectl, improve development experiences, but for how to prompt those information properly, there's two solutions here I've come up with: 
 
 ### solution 1
 
-Keep current help messages, and add an ring1Factory function use clientDiscovery fetch resources/shortcuts from server then concatenate those valid resources string and print every time when user execute command like this:
+Keep current help messages, and add a ring1Factory function use clientDiscovery fetch resources/shortcuts from server then concatenate those valid resources string and print every time when user execute command like this:
+
 ```
 ➜  ~ kubectl get -h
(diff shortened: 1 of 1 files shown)


dixudx:
Since we've already got a cached discovery file name `servergroups.json` in folder `~/.kube/cache/discovery/<server-name>/servergroups.json`, why not utilize this file. This will quite improve the user experience(maybe they just want to get the help message, there is no need to perform requests to any clusters) and bring a lot of convenience. And this file will be updated automatically when out-of-dated by other current kubectl subcommands.

@shiywang @liggitt WDYT? Can we add this as an alternative.

tcharding:
What is the exact problem we are trying to solve?

1. Fix the developer experience.
2. Get a more accurate valid resource list.
3. Some other problem.

If it is #1 we could solve this by simply refactoring the code to have a single source of truth.

liggitt:
> If the file ~/.kube/cache/discovery/<server-name>/servergroups.json does not exist, we could perform solution1 to fetch those valid resources.

The full command tree (including the help text for every command) is built on every invocation of kubectl. We cannot make server calls and should not make expensive calls (like loading lots of files from disk) in order to do that

liggitt:
> What is the exact problem we are trying to solve?
>
> 1. Fix the developer experience.
> 2. Get a more accurate valid resource list.
> 3. Some other problem.

We're trying to fix the experience by giving an accurate resource list. I think the best way to do that without adversely affecting kubectl is to have a dedicated command to list valid resources and to reference that from help text.

contributors/design-proposals/sig-cli/dynamic_resource_shortcut_fecth.md
```diff
@@ -0,0 +1,107 @@
...
+Previously, when a new resource added to server, developer should add those resources/shortcuts name into two places of kubectl: [here](https://github.com/kubernetes/kubernetes/blob/357db0c39c7203809b369516c0ec93831ead8649/pkg/kubectl/cmd/cmd.go#L201) and [there](https://github.com/kubernetes/kubernetes/blob/f9913c4948038b9b51f2342134d546c6bb74e7a3/pkg/kubectl/kubectl.go#L46). then `kubectl get/describe/explain... -h` will show the new added hard code resources/shortcuts, But some times people always forget to add those things, and we have to submit new pr to fix that everytime.
+
+## Proposal
+
+We should eliminate those every time add new hardcoded resources in kubectl, improve development experiences, but for how to prompt those information properly, there's two solutions here I've come up with: 
+
+### solution 1
+
+Keep current help messages, and add a ring1Factory function use clientDiscovery fetch resources/shortcuts from server then concatenate those valid resources string and print every time when user execute command like this:
+
```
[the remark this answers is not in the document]
shiywang: yes, this is situation happened in soltuion1, if no available cluster, we will fall back to hardcoded resources, see last time sig-cli mtg: https://www.youtube.com/watch?v=ykUDPVizdG8&t=1983s

contributors/design-proposals/sig-cli/dynamic_resource_shortcut_fecth.md
```diff
@@ -0,0 +1,107 @@
...
+## User Experience
+
+### Use Cases
+
+`kubectl get/describe/explain..   -h`
+
+
+### Shortage
+
+solution 1 may have some shortage in first time prompt help messages.
```
apelisse: is this why you have two solutions?

contributors/design-proposals/sig-cli/dynamic_resource_shortcut_fecth.md
```diff
@@ -0,0 +1,107 @@
...
+This is a simple one, only change is remove hardcoded list in help messages of `kubectl get/explain/describe ...` and implement a new command `kubectl api-resources`.
+
+
+## Client/Server Backwards/Forwards compatibility
+
+Depend on which solution we choose.
+
+## Other things need to discuss
+
+1. Use binary cache from OpenAPI instead of cache from `.kube/cache/discovery` ? I think this is only happend in we chose solution one, and I think openapi is more suit for validation, and the information we need here is only resources name and shortcut.
```
apelisse: @liggitt Can we actually get that data from OpenAPI? I don't think there is any "shortcut" in OpenAPI.

PR #1020, opened 2017-09-06
pwittrock bio for voter guide

pwittrock:


committee-steering/pwittrock_bio.md
```diff
@@ -0,0 +1,53 @@
...
+the products that we build:
+
+- address their current challenges and pain points
+- include the resources they need to resolve issues they encounter on their own
+- grow with them
+
+## Roles held
+
+Since joining the community in 2015, my contributions have been broad and shallow in several roles
+across the project.  Areas I have been contributed to in some forminclude:
```
thockin: s/forminclude/form include/

pwittrock:
@thockin SGTM.  Got a few words left before I hit the 300 word count.

pwittrock:
Add it to the end.  If there is a structure you would suggest for the purposes of consistency, I will update my bio to conform.

derekwaynecarr:
@smarterclayton @thockin - i am using this bio as a template.  is an indication of employer consistent with what is written here (that it should be a brand free activity)?

https://github.com/kubernetes/steering/blob/master/charter.md#limiting-corporate-campaigning

eparis:
I would argue that stating you corp affiliation is ok, but campaigning meaningfully on that affiliation instead of your own merits is bad.

smarterclayton:
That sounds reasonable.

derekwaynecarr:
@eparis - makes sense to me, i just wanted to make sure we are consistent.  its not hard to find anyone's employer.

pwittrock:
@derekwaynecarr Cool link, I didn't realize that.  I don't have a strong opinion on the matter.  @thockin Suggested including it since there are a max number of seats per-corp and voters might find it useful in this context.

I'll leave it until directed otherwise.  It does eat a few words on my word limit that I would accept back.

xiangpengzhao:
@pwittrock FYI: please move your bio to dir `community/elections/2017`. See: #1030 

cc @calebamiles

pwittrock:
@xiangpengzhao Done

PR #1024, opened 2017-09-07
Fix 'Upgrading Go' outdated description

billy2180:
1. dockerized-e2e-runner.sh has been deleted， FYI [#3702](https://github.com/kubernetes/test-infra/pull/3702)
2. The  [e2e-image](https://github.com/kubernetes/test-infra/tree/master/jenkins/e2e-image) and [build/common.sh](https://github.com/kubernetes/kubernetes/blob/master/build/common.sh) do not need to be updated

### DIFF
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -144,10 +144,7 @@ images.
 
 - The image for cross compiling in [build/build-image/cross].
   The `VERSION` file and `Dockerfile`.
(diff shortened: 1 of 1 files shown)


cblecker:
A wholesale rewrite is actually needed here. I've added a note to https://github.com/kubernetes/kubernetes/issues/49484 so that we circle back around to this, but updating that one image isn't the only thing that needs to be done.

PR #1028, opened 2017-09-08
Proposal for re-architecting apply

pwittrock:


cblecker:
@kubernetes/sig-cli-proposals 

Also cc: @kubernetes/sig-apps-proposals and @kubernetes/helm-maintainers .. I know there is some disparity between the way a `helm upgrade` works and the way a `kubectl apply` does when modifying existing manifests. I wonder if this is an opportunity to more closely align them within the kubernetes ecosystem :). If not, I apologize for the spam.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,91 @@
...
+  and diffs it against the local configuration.  Building a patch containing special merge
+  directives is a more complicated approach than modifying the object read from the server
+  and updating it with a PUT.
+- Patches provide weaker optimistic locking guarantees than PUT.  When encountering an optimistic
+  lock failure in apply, it is easy to re-read the object and try to update it again.  If the
+  apply command tries to modify a field that was modified since apply read the object, it will fail
+  and optionally be able to show this to a user, whereas this conflict may not be detect with a PATCH
+  operation.
+
+#### New approach
```
smarterclayton: Why not just move apply to the server via a formal API?

pwittrock: @bgrant0607 asked a similar question.

My hope as that the steps in this proposal will make moving the logic to an API easier by reducing the complexity of the existing solution and making it a modular package that does not depend on cobra or other kubectl libraries to diff & update objects.

I'd be open to exposing the re-architected libraries as an API.

smarterclayton:
Yeah library components are good.  I just wanted to see it mentioned in the
doc so when the "apply on server" proposal gets thrown together I can point
back to that line.

bgrant0607:
Server-side apply issue is #17333

bgrant0607:
Also relevant are uses of both apply and patch in the declarative app mgmt doc: https://goo.gl/T66ZcD

Umbrella apply issue: https://github.com/kubernetes/kubernetes/issues/35234

Original apply proposal: https://github.com/kubernetes/kubernetes/issues/1702

bgrant0607:
Original original issue: https://github.com/kubernetes/kubernetes/issues/1178

bgrant0607:
Previous umbrella issue with design description: https://github.com/kubernetes/kubernetes/issues/15894

contributors/design-proposals/sig-cli/apply_v2.md
anguslees: Because that's how we got strategic merge :P   In the general case, the server _can't_ know whether particular operations should be "replace" or "merge", and trying to capture all that flexibility in the API has created exactly the sort of complexity that motivated this doc.  The alternative is just to do it client-side, where naturally there is more context available to be able to implement the exact semantics intended by the client.

anguslees:
+100.  I've been slowly building up evidence to justify why I need to implement a new client-side "patch" as part of [kubecfg](http://github.com/ksonnet/kubecfg).  I'm happy to see that other people are sharing my pain :)

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,91 @@
...
+
+- Read the live object
+- Compare the live object to last-recorded and local files
```
liggitt: do you still plan to record last-applied as an annotation?
discussion of the versioning issues (apply v1alpha1, time passes, want to apply v1beta2) would be good as well

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,91 @@
...
+- Update the fields on the live object that was read
```
liggitt: using what merge strategies?

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: Yes, this proposal does not change what is stored for any of the objects, and works exactly the same with the exception that it uses PUT instead of PATCH to perform the update.  It might best to solve the issues around versioning in a server-side API.  I imagine the server side API would type convert all 3 objects (recorded from annotation, local from request and remote from live-lookup) to the same version before performing the diff-merge-update.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: The merge strategy / behavior would be exactly the same as the current version of apply, though more easily extended due to code structuring.

pwittrock:
@smarterclayton Good point.  I will add a section about that in the doc and the opportunity to integrate from the server-side on the future.

bgrant0607:
Also relevant: discriminated unions: https://github.com/kubernetes/kubernetes/issues/35345

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+
+- Patch API is not fully featured enough to represent all of the operations required
+  to update an object and frequently has to be expanded with new side-directives describing
+  how the server should apply the patch.  Updates are done in the form of adding new
+  merge directives to the patch format on the server, and invoking them from the client.
+- Having to update both client and server to fix bugs introduces challenges to supporting
+  version skew, as the client will depend on the server supporting a patch semantic.  There
+  is currently no versioning of the patch semantics, so the client cannot reliably know what
+  patch semantics the server supports.
+- Patch syntax is not human readable, making it hard for humans to evaluate the end state
```
bgrant0607: Could you expand on the readability? Is this discussing strategic merge patch? Is there special syntax in the typical case other than for deletions? Where has the deletion syntax been made unreadable?

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+
+Issues with updating the object by calculating and sending a patch to the server
+
+- Patch API is not fully featured enough to represent all of the operations required
+  to update an object and frequently has to be expanded with new side-directives describing
+  how the server should apply the patch.  Updates are done in the form of adding new
+  merge directives to the patch format on the server, and invoking them from the client.
+- Having to update both client and server to fix bugs introduces challenges to supporting
+  version skew, as the client will depend on the server supporting a patch semantic.  There
+  is currently no versioning of the patch semantics, so the client cannot reliably know what
```
bgrant0607: We would version it by adding new content types, which I believe we also do with proto encodings.

But discoverability of supported versions is a good point.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+  patch semantics the server supports.
+- Patch syntax is not human readable, making it hard for humans to evaluate the end state
+  that will result from the patch.  `--dry-run` is nearly useless for understanding the
```
bgrant0607: Why is dry run useless? Because it is also buggy and incomplete? Because it is implemented in the client rather than in the server?

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Why
+
+Issues with updating the object by calculating and sending a patch to the server
+
+- Patch API is not fully featured enough to represent all of the operations required
+  to update an object and frequently has to be expanded with new side-directives describing
+  how the server should apply the patch.  Updates are done in the form of adding new
+  merge directives to the patch format on the server, and invoking them from the client.
+- Having to update both client and server to fix bugs introduces challenges to supporting
```
bgrant0607: Links to exemplary bugs?

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+  is currently no versioning of the patch semantics, so the client cannot reliably know what
+  patch semantics the server supports.
+- Patch syntax is not human readable, making it hard for humans to evaluate the end state
+  that will result from the patch.  `--dry-run` is nearly useless for understanding the
+  end state that will result from apply, and implementing a diff operation has been exceedingly
+  copmlex.
+- Generating a patch to make a change is more complicated than simply making the change to
+  the object read from the server.  The apply command reads the live object from the server
+  and diffs it against the local configuration.  Building a patch containing special merge
+  directives is a more complicated approach than modifying the object read from the server
```
bgrant0607: True.

However:
* My impression is that most of the complexity is not in the directives but in the API schema and defaulting behaviors, as I mentioned in the [server-side apply proposal](https://docs.google.com/document/d/1G4L4Kdyfg80DDuVOXHY0UbFuqDO29kECnOacVlseiFw/edit#).
* The schema needs to be accurately communicated to the client from the server (e.g., via extended OpenAPI) for client-side patch operations in order to support API aggregation and CRD, which re-introduces versioning and skew issues until the syntax stabilizes, but actually would improve skew behavior once it does.
* We need to support these patch behaviors independently of apply.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+  and updating it with a PUT.
+- Patches provide weaker optimistic locking guarantees than PUT.  When encountering an optimistic
```
bgrant0607: I'm not sure that's the case. I thought resourceVersion could be provided in patches.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+  and optionally be able to show this to a user, whereas this conflict may not be detect with a PATCH
+  operation.
+
+#### New approach
+
+- Read the live object
+- Compare the live object to last-recorded and local files
+- Update the fields on the live object that was read
+- Send a PUT to update the modified object
+- If encountering optimistic lock failure, retry after re-reading the object
```
bgrant0607: Specifically, GET, merge, and then try to PUT again.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+- Larger request sizes to the server than when sending a patch
+
+### Restructure code into modular components
+
+In the current implementation of apply - parsing and traversing the object trees, diffing the
+contents and generating the patch are logically entangled.  This creates maintenance and
+testing challenges.  We should instead modularize discrete responsibilities - collating the
+object values and updating the target object.
+
+#### Phase 1: Parse last-recorded, local, live objects and collate
```
bgrant0607: last-applied-configuration?

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,94 @@
...
+
+Good:
+
+- Simpler code to maintain
+- Behaves correctly in more instances
+- Easy to look at the output and determine the final state
+
+Bad:
+
+- Larger request sizes to the server than when sending a patch
```
bgrant0607: We had also discussed sending the full configured resource decorated with deletion directives, which would have been slightly larger.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: Any examples in the API schema where it shouldn't be obvious? In the few ordered lists we have (e.g., command, args), those need to be replaced. All other cases should be merged. In the case of no conflict between the configured state and the live state, a merge and a replacement should be the same.

bgrant0607:
Draft of server-side proposal: https://docs.google.com/document/d/1G4L4Kdyfg80DDuVOXHY0UbFuqDO29kECnOacVlseiFw/edit#

mengqiy:
> Using compiled-in API types in kubectl

This should be solved by https://github.com/kubernetes/kubernetes/pull/51321. But it didn't make it in 1.8, I will reopen it after code freeze.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: Re. discoverability: It's possible to specify supported content types for each verb X resource via OpenAPI

mengqiy:
Link the related proposal of sending mergekeys in the patch body as a directive: https://github.com/kubernetes/community/pull/476
The proposal above is trying to solve the problem of the mergekeys between client and serve are not synced in the case of version skew. This problem will be resolved if apply (diff+merge) is done completely at server-side (or client-side).

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: It doesn't show the user what they really want to know - what will be the new object in the cluster after apply is run.  It somewhat shows the changes being made, but without the context of what they are being made on.

- I would expect apply users care less about the patch format itself, and care more about the object in the cluster that will result from the patch.
- The patch output obfuscates validation errors caused by the values set by patch conflicting with other values set on the live object missing from the config (e.g. rolloutStrategy for Deployment)
- The patch itself is buggy and broken cases are hard to see just by looking at a patch (e.g. container ports merge key being incorrect)
- The server may apply the patch differently than the client, so running the patch locally against the file may result in something different than what the server does (e.g. the server doesn't support optional directives because it is skewed)

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: > My impression is that most of the complexity is not in the directives but in the API schema and defaulting behaviors, as I mentioned in the server-side apply proposal.

Constructing a set of directives to order and delete items from a list results in more complicated code than just deleting the objects from the list.

> The schema needs to be accurately communicated to the client from the server (e.g., via extended OpenAPI) for client-side patch operations in order to support API aggregation and CRD, which re-introduces versioning and skew issues until the syntax stabilizes, but actually would improve skew behavior once it does.

Agreed.  This proposal should also work independently of whether it is done on the client or server.  The focus of this proposal is on rewriting the 3-way diff code to be maintainable.  The inputs will be the 3 versions of the object and the output will be the new object.  

> We need to support these patch behaviors independently of apply.

Yes, but why do we need to use them in apply?

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: Yes this is strategic merge patch.  I think there are 2 main problems, there are several different directives now (delete, delete-primitive, ordering), and how the patch will be applied is communicated out of band.  There are still a number of cases that are not supported by strategic patch that may require additional directives to support if done client side (e.g. how do we change the merge key for a field)

Directives:
- deletion of complex objects from lists
- deletion of primitive objects from lists (looks different)
- absolute ordering of items in a list

Additional issues with readability:
- merge keys for lists are not contained in the patch itself - unclear from looking at a patch whether it is adding an item to a list that looks similar to another item, or if it is updating the existing item in a list
- merge strategy not clear - are items in a list being replace, or are they being merged, (or using retain keys)?
- server silently drops directives that it doesn't recognize, so they may or may-not be used by the server - e.g. ordering may be randomized even with a directive specifying the ordering

pwittrock:
> Using compiled-in API types in kubectl

@mengqiy already has a PR for this that has been lgtm'ed and approved.  Should go once the code freeze is lifted.

> Client-side apply

This doesn't directly address this, but I think we will want to rewrite / restructure the code in addtion to moving it to the server.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: FWIW, I don't understand what apply's current patch shows me when I do `--dry-run -o json`.  It looks like it is just showing me the local object without applying the patch.

pwittrock:
> I don't have a strong opinion about PATCH vs. PUT, but it seems to me that this doesn't solve what I see as the 2 biggest problems:

The biggest immediate problem I face with apply is that the code to generate a PATCH is complex and difficult to change / review.  (A contributor spent 1-2 releases trying to implement `diff` and the code was so complex we had to drop 2 separate attempts).  Moving it to the server will not fix this unless we rewrite it in the process.

bgrant0607:
@pwittrock Could you please reference some specific issues and/or PRs that illustrate the general problems that need to be solved?

php-coder:
This proposal should be places into `community/contributors/design-proposals/cli` directory (because `sig-cli` was renamed recently).

pwittrock:
@bgrant0607 Yes, good suggestion.  I have a few in mind that I think will do a good job.  I will add a section to the doc that performs an analysis of past PRs, how they were written, and how they would have been written under the new architecture.

pwittrock:
@bgrant0607 Updated with a better explanation of why the patch isn't human readable, and several PRs that show the complexity of making changes to the diff logic + accompanying analysis of where the complexity comes from.

pwittrock:
@php-coder thanks.  will move once all comments have been addressed - don't want to risk losing them.

contributors/design-proposals/sig-cli/apply_v2.md
mengqiy: `retainKeys` is also a directive that you didn't mentioned above.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+  - are spread across 1000+ lines of existing code
+- Changes that add new directives require updates in multiple locations - create patch + apply patch
+
+### PRs
+
+[38665](https://github.com/kubernetes/kubernetes/pull/38665/files)
+- Support deletion of primitives from lists
+- Lines (non-test): ~200
+- ~6 weeks
+[44597](https://github.com/kubernetes/kubernetes/pull/44597/files)
```
bgrant0607: This was the wrong solution to this problem, IMO.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+- Support deleting fields not listed in the patch
+- Lines (non-test): ~250
+- ~6 weeks
+[45980](https://github.com/kubernetes/kubernetes/pull/45980/files#diff-101008d96c4444a5813f7cb6b54aaff6)
+- Keep ordering of items when merging lists
+- Lines (non-test): ~650
+[46161](https://github.com/kubernetes/kubernetes/pull/46161/files#diff-101008d96c4444a5813f7cb6b54aaff6)
+- Support using multiple fields for a merge key
```
bgrant0607: This seems like it will still be a problem.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+- Status: Closed - too hard for maintainers to understand impact and correctness of changes
+[49174](https://github.com/kubernetes/kubernetes/pull/49174/files)
+- Support diff apply (2nd attempt)
+- Status: Deferred indefinitely - too hard for maintainers to understand impact and correctness of changes
+- Maintainer reviews: 3
+
+
+### Analysis - causes of complexity
+
+Apply is implemented by diffing the 3 sources (last-applied, local, remote) as 2 2-way diffs and then
```
bgrant0607: Alternative implementation that has been proposed in the past:
1. Diff last-applied and local, and generate a patch by supplementing local with list and map deletion directives
2. Diff last-applied and remote/live to determine whether the live state was changed in a conflicting manner

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+are only compared indirectly when the 2 diff results are merged.  Information that is redundant or
+should be checked for consistency across all 3 sources (e.g. checking for conflicts) is spread across
+3 logic locations - the first 2-way diff, the second 2-way diff and the merge of the 2 diffs.
+
+That the diffs each may produce multiple patch directives + results that constitute an update to a single
+field compounds the complexity of that comparing a single field occurs across 3 locations.
+
+Additionally, the logic to apply the patch must be updated both on the the server side
+and on the client, but cannot be done so transactionally.
+Because there there is no way for the client to know what directives a server supports (may be old),
```
bgrant0607: Regardless what we do for apply, it sounds like we need a discovery mechanism for this.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+new directives must always be present but the patch must continue to function if they are ignored
+by the server.
+Because the server has no guarantee the client will send all directives it expects (may be old), it must
+not require any new directives to apply a patch "correctly".
```
bgrant0607: It sounds like the behavior before the new directives were implemented was also incorrect?

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+  - Humans must mentally apply directives (delete, retain keys, ordering) in order to understand
+    the result of the patch.
+  - Merge keys are implicit (not shown in the patch), but change the result
+  - Merge strategy is implicit (not show in the patch), but change the result
+  - Merge keys and strategy are incorrect and buggy for several object fields (e.g. Container.Ports)
+    resulting in counter-intuitive results (and there is no diff command to help users see this
+    beforehand)
+  - Patch hides values on the remote object, which may have hidden interactions with the patch values
+    (e.g. Deployment.Spec.RolloutStrategy)
+- Generating a patch to make a change is more complicated than simply making the change to
```
bgrant0607: It seems to me that the client vs. server discussion isn't the main point. 

The main point is whether the merge is accomplished via separate diff and patch passes, or with a custom merge implementation.

bgrant0607: And the new implementation should be mostly reusable for a server-side apply API.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+
+- PRs frequently closed or deferred because maintainers / reviewers cannot reason about the impact or
+  correctness of the changes
+- Relatively simple changes
+  - are 200+ lines of code
+  - modify dozens of existing locations in the code
+  - are spread across 1000+ lines of existing code
+- Changes that add new directives require updates in multiple locations - create patch + apply patch
+
+### PRs
```
bgrant0607: I want overlays to work, so I think all of these problems still need to be solved for strategic merge patch.

bgrant0607:
I'm not opposed to implementing a custom merge operation for apply if that's simpler than diff and patch, but I disagree that it would address most of the difficulties discussed in the motivation.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+
+This is a text-book example of what the [Visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) 
+was designed to address.
+
+- Encapsulate logic in *Element*s and *Visitor*s
+- Introduce logic for new a field type by adding a new *Element* type
+- Introduce logic for new a merge strategy by defining a new *Visitor* implementation
+- Introduce logic on structuring of a field by updating the parsing function for that field type
+
+If the apply diff logic was redesigned, most of the preceding PRs could be implemented by
```
bgrant0607: Note that this would still change user-visible behavior. While technically bug fixes, we should consider how to communicate such changes to users, whether we need to retain prior behavior, and how we should transition to implementations with different behavior.

contributors/design-proposals/sig-cli/apply_v2.md
mengqiy: I want to understand why.

contributors/design-proposals/sig-cli/apply_v2.md
mengqiy: Maybe we can introduce multi-fields merge key as a new tag for apply. It may be the right chance to fix the wrong merge key issue, e.g. container.ports.
Server-side apply in the future can reuse this new merge key.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: We want the correct behavior regardless whether POST, PUT, PATCH (any flavor), or apply is used. Therefore, the proper fix would be in the API. My proposal in https://github.com/kubernetes/kubernetes/issues/35345 was to add discriminators. It was incorrectly concluded that "We should set discriminators by default", and was dismissed as unsafe without further explanation.

Regardless, unions/OneOf are a property of the API schema. A patch directive should not have been needed. The patch version could be specified via Content-Type, which is how we currently distinguish among the flavors of patches:
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#patch-operations

bgrant0607: And it's not clear to me that changing default API behavior is any more or less a breaking change than changing default kubectl behavior with no way to preserve the previous behavior.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: If we use content types for specifying the patch version, that can be represented in OpenAPI.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: And what would break, actually? A patch to another key in a union should result in validation failure in the case that the previously set key were not cleared. Making such a request succeed would not be a breaking change -- all requests that previously worked would still work the same way.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: Yes, but the directives may only be needed for some cases.  For example, apply works for merging `Container.Ports` when 2 Ports don't share the same port number with different protocols.  This should continue to work with the new directives.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: Yes exactly.  I can change the name of the doc if that would help.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: Agreed.  We should talk offline about this, since the requirements for overlays may be a subset of what is need for apply to work.  The semantics of apply are a bit different (though there is a lot of overlap) since apply is merging 3 modified versions of the same object to produce an updated object.  Whereas overlays do work in this context.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: Yes this makes sense.

In some cases the fixes purely fix something that was broken or unspecified (e.g. maintaining ordering for lists when merging is strictly more correct than randomizing the ordering).  In other cases I think it is pretty clearly an improvement, but the change should be communicated (e.g. deleting an item from a primitive lists now deletes it).  

The cases more related to your point would be fixing an unintended / broken merge-key or merge-strategy.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: Yes that sounds like a good direction.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: It will still be a problem for PATCH, but provides a simpler solution for apply.

We won't need to come up with a strategy for communicating the merge keys within the PATCH (either explicitly or implicitly through versioning).  A relatively simple solution would be communicate the updated merge-keys to the client in openAPI, and then the client can decide if they want to use these keys or not.

Solving this for PATCH is a larger issue.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: I would be happy to see Union get implemented.  It will need a champion for that to happen.  As things stand today, don't see my team tackling this in the next year.

pwittrock:
@bgrant0607 

> I'm not opposed to implementing a custom merge operation for apply if that's simpler than diff and patch, but I disagree that it would address most of the difficulties discussed in the motivation.

The motivation then didn't do a sufficient job of arguing that the current code structure is now unmaintainable and brittle.  This is due largely to how the needs have evolved, and the original structure does not support extending the logic.

There seems to be consensus with the @apelisse and @mengqiy  that the [new architecture](https://github.com/kubernetes/kubernetes/pull/52349) will be simpler to maintain.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: No directives should be needed for that case.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: Even without discriminated unions, I don't see why directives were necessary.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: Patches shouldn't need to communicate merge keys. That should be a property of the schema, not a property of the patch.

pwittrock:
@bgrant0607 @smarterclayton I've tried to address your comments.  LMK if there is anything else.

pwittrock:
@apelisse LMK if there is anything I should change since you are reviewing the code.

bgrant0607:
@pwittrock Did you intend to push new updates? I don't see any commits newer than about a week old

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: The client needs to communicate what it thinks the merge keys are - either implicitly through some versioning, or explicitly as some directive - for us to change the merge keys for things.

contributors/design-proposals/sig-cli/apply_v2.md
pwittrock: If it is not explicitly part of the patch, then we need to introduce versioning semantics so both the client and server agree on the implicit merge keys since they may change.

pwittrock:
@bgrant0607 I not sure what changes to this proposal you are requesting before it is merged.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
+# Apply v2
+
+## Background
+
+`kubectl apply` reads a file or set of files, and updates the cluster state based off the file contents.
+It does a couple things:
+
+1. Create / Update / (Delete) the resource based on the resources
```
bgrant0607: the live resources based on the file contents

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+2. Only set / delete fields defined in the file either currently or in during last update from the file
```
bgrant0607: Update currently and previously configured fields, without clobbering fields set by other means, such as imperative kubectl commands, other deployment and management tools, admission controllers, initializers, horizontal and vertical autoscalers, operators, and other controllers.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+  - Retain updates to fields made by controllers that were never managed in the file
+3. Merge fields based on the field name, merge lists based on a field contained within the list elements
```
bgrant0607: I would move this below. It's a detail of field updates.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+
+The apply logic has been difficult to maintain from a contributor perspective, and difficult
+to understand from a user perspective.  Much of this pain has comes from the `apply` command out
```
bgrant0607: Much of this pain has come from trying to paper over API schema complexity in an ad hoc incremental fashion using imperative patch directives.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+growing its implementation.  In order to continue to support apply and provide better user visibility
+into what it is doing, the command must be re-architected to break client dependencies on the server-side
```
bgrant0607: I thought we agreed that client-server was not the main issue.

bgrant0607: In order to make apply sufficiently maintainable and extensible to new API types, as well as to make its behavior more intuitive for users, the merge behavior, including how it is specified in the API schema, must be systematically redesigned and more thoroughly tested.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: For me: links to bugs are below.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+
+Implementation of proposed changes under review in PR [52349](https://github.com/kubernetes/kubernetes/pull/52349)
+
+### Use read-update instead of patch
+
+#### Why
+
+Issues with updating the object by calculating and sending a patch to the server
+
+- Patch API is not fully featured enough to represent all of the operations required
```
bgrant0607: has been frequently expanded

It doesn't have to be this way.

contributors/design-proposals/sig-cli/apply_v2.md
```diff
@@ -0,0 +1,201 @@
...
+  to update an object and frequently has to be expanded with new side-directives describing
+  how the server should apply the patch.  Updates are done in the form of adding new
+  merge directives to the patch format on the server, and invoking them from the client.
```
bgrant0607: Again, please rephrase. The schema MUST come from the server, in order to properly work with version skew, API aggregation, CRDs, etc.

contributors/design-proposals/sig-cli/apply_v2.md
bgrant0607: Please add something about versioning the patch API in the future.

pwittrock:
@bgrant0607 I've updated the doc to be focussed only on the code refactoring.  I added some bugs and called out the need to fix issues with the schema semantics as a separate item.

contributors/design-proposals/cli/apply_refactor.md
```diff
@@ -0,0 +1,184 @@
...
+- Using PATCH semantics is more work to implement a diff of the changes as
+  PATCH must be separately merged on the remote object for to display the diff.
+
+#### New approach
+
+- Read the live object
+- Compare the live object to last-applied and local files
+- Update the fields on the live object that was read
+- Send a PUT to update the modified object
+- If encountering optimistic lock failure, retry after re-reading the object
```
smarterclayton: retry back to step 1?

contributors/design-proposals/cli/apply_refactor.md
```diff
@@ -0,0 +1,184 @@
...
+
+- Merge will use openapi to get the schema from the server
+- Merge can be run either on the server side or the client side
+- Merge can handle 2-way or 3-way merges of objects (initially will not support PATCH directives)
+
+## Out of scope of this doc
+
+In order to make apply sufficiently maintainable and extensible to new API types, as well as to make its
+behavior more intuitive for users, the merge behavior, including how it is specified in the API schema,
+must be systematically redesigned and more thoroughly tested.
```
smarterclayton: +1

smarterclayton:
No further comments from me

contributors/design-proposals/cli/apply_refactor.md
pwittrock: That was Brian :)

bgrant0607:
Thanks.
/lgtm

PR #1037, opened 2017-09-11
Integrate voting guide into the existing structure

castrojo:
This integrates @parispittman 's voter's guide into the existing structure, and then I hyperlinked in the first two submissions.


xiangpengzhao:
+1

PR #1048, opened 2017-09-12
Update Docker Shared PID proposal for per-pod configuration

verb:
We will need to support configuring shared PID namespace in docker per-pod.

### DIFF
--- a/contributors/design-proposals/node/pod-pid-namespace.md
+++ b/contributors/design-proposals/node/pod-pid-namespace.md
@@ -1,72 +1,319 @@
 # Shared PID Namespace
 
-Pods share namespaces where possible, but a requirement for sharing the PID
-namespace has not been defined due to lack of support in Docker. Docker began
-supporting a shared PID namespace in 1.12, and other Kubernetes runtimes (rkt,
-cri-o, hyper) have already implemented a shared PID namespace.
-
-This proposal defines a shared PID namespace as a requirement of the Container
-Runtime Interface and links its rollout in Docker to that of the CRI.
+*   Status: Pending
+*   Version: Alpha
+*   Implementation Owner: [@verb](https://github.com/verb)
 
 ## Motivation
 
-Sharing a PID namespace between containers in a pod is discussed in
-[#1615](https://issues.k8s.io/1615), and enables:
+Pods share namespaces where possible, but support for sharing the PID namespace
+had not been defined due to lack of support in Docker. This created an implicit
+API on which certain container images now rely. This document proposes adding
+support for sharing a process namespace between containers in a pod while
+maintaining backwards compatibility with the existing implicit API.
 
-  1. signaling between containers, which is useful for side cars (e.g. for
-     signaling a daemon process after rotating logs).
-  2. easier troubleshooting of pods.
-  3. addressing [Docker's zombie problem][1] by reaping orphaned zombies in the
-     infra container.
+## Proposal
 
-## Goals and Non-Goals
+### Goals and Non-Goals
 
 Goals include:
-  - Changing default behavior in the Docker runtime as implemented by the CRI
-  - Making Docker behavior compatible with the other Kubernetes runtimes
+
+*   Backwards compatibility with container images expecting `pid == 1` semantics
+*   Per-pod configuration of PID namespace sharing
+*   Ability to change default sharing behavior in `v2.Pod`
 
 Non-goals include:
-  - Creating an init solution that works for all runtimes
-  - Supporting isolated PID namespace indefinitely
 
-## Modification to the Docker Runtime
+*   Creating a general purpose container init solution
+*   Multiple shared PID namespaces per pod
+*   Per-container configuration of PID namespace sharing
 
-We will modify the Docker implementation of the CRI to use a shared PID
-namespace when running with a version of Docker >= 1.12. The legacy
-`dockertools` implementation will not be changed.
+### Summary
 
-Linking this change to the CRI means that Kubernetes users who care to test such
-changes can test the combined changes at once. Users who do not care to test
-such changes will be insulated by Kubernetes not recommending Docker >= 1.12
-until after switching to the CRI.
+We will add support for configuring pod-shared process namespaces by adding a
+new boolean field `ShareProcessNamespace` to the pod spec. The default to false
+means that each container will have a separate process namespace. When set to
+true, all containers in the pod will share a single process namespace.
 
-Other changes that must be made to support this change:
+The Container Runtime Interface (CRI) will be updated to support three namespace
+modes: Container, Pod & Node. The Runtime Manager will translate the pod spec
+into one of these modes as follows:
 
-1. Add a test to verify all containers restart if the infra container
-   responsible for the PodSandbox dies. (Note: With Docker 1.12 if the source
-   of the PID namespace dies all containers sharing that namespace are killed
-   as well.)
-2. Modify the Infra container used by the Docker runtime to reap orphaned
-   zombies ([#36853](https://pr.k8s.io/36853)).
+Pod `shareProcessNamespace` | Pod `hostPID` | CRI PID Mode
+--------------------------- | ------------- | ------------
+false                       | false         | Container
+false                       | true          | Node
+true                        | false         | Pod
+true                        | true          | *Error*
 
-## Rollout Plan
+If a runtime does not implement a particular PID mode, it must return an error.
+For reference, Docker will support all three modes when using version >= 1.13.1.
 
-SIG Node is planning to switch to the CRI as a default in 1.6, at which point
-users with Docker >= 1.12 will receive a shared PID namespace by default.
-Cluster administrators will be able to disable this behavior by providing a flag
-to the kubelet which will cause the dockershim to revert to previous behavior.
+The shared PID functionality will be hidden behind a new feature gate in both
+the API server and the kubelet, and the existing `--docker-disable-shared-pid`
+flag will be removed from the kubelet, subject to [deprecation
+policy](https://kubernetes.io/docs/reference/deprecation-policy/).
 
-The ability to disable shared PID namespaces is intended as a way to roll back
-to prior behavior in the event of unforeseen problems. It won't be possible to
-configure the behavior per-pod. We believe this is acceptable because:
+## User Experience
 
-* We have not identified a concrete use case requiring isolated PID namespaces.
-* Making PID namespace configurable requires changing the CRI, which we would
-  like to avoid since there are no use cases.
+### Use Cases
 
-In a future release, SIG Node will recommend docker >= 1.12. Unless a compelling
-use case for isolated PID namespaces is discovered, we will remove the ability
-to disable the shared PID namespace in the subsequent release.
+Sharing a PID namespace between containers in a pod is discussed in
+[#1615](https://issues.k8s.io/1615) and enables:
 
+1.  signaling between containers, which is useful for side cars (e.g. for
+    signaling a daemon process after rotating logs).
+1.  easier troubleshooting of pods.
+1.  addressing [Docker's zombie
+    problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/)
+    by reaping orphaned zombies in the infra container.
 
-[1]: https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/
+### Behavioral Changes
+
+Sharing a process namespace fits well with Kubernetes' pod abstraction, but it's
+a significant departure from the traditional behavior of Docker. This may break
+container images and development patterns that have come to rely on process
+isolation. Notably:
+
+1.  **The main container process no longer has PID 1**. It cannot be signalled
+    using `kill 1`, and attempting to do so will instead signal the
+    infrastructure container and potentially restart the pod. Containers
+    shipping an init system like systemd may [require additional
+    flags](https://github.com/kubernetes/kubernetes/issues/48937#issuecomment-321243669).
+1.  **Processes are visible to other containers in the pod**. This includes all
+    information visible in `/proc`, such as passwords as arguments or
+    environment variables, and process signalling. This can be somewhat
+    mitigated by running processes as separate, non-root users.
+1.  **Container filesystems are visible to other containers in the pod through
+    the <code>/proc/$pid/root</code> magic symlink**. This makes debugging
+    easier, but it also means that secrets are protected only by standard
+    filesystem permissions.
+
+## Implementation
+
+### Kubernetes API Changes
+
+`v1.PodSpec` gains a new field named `ShareProcessNamespace`:
+
+```
+// PodSpec is a description of a pod.
+type PodSpec struct {
+    ...
+       // Use the host's pid namespace.
+       // Note that HostPID and ShareProcessNamespace cannot both be set.
+       // Optional: Default to false.
+       // +k8s:conversion-gen=false
+       // +optional
+       HostPID bool `json:"hostPID,omitempty" protobuf:"varint,12,opt,name=hostPID"`
+       // Share a single process namespace between all of the containers in a pod.
+       // Note that HostPID and ShareProcessNamespace cannot both be set.
+       // Optional: Default to false.
+       // +k8s:conversion-gen=false
+       // +optional
+       ShareProcessNamespace *bool `json:"shareProcessNamespace,omitempty" protobuf:"varint,XX,opt,name=shareProcessNamespace"`
+       ...
+```
+
+The field name deviates from that of HostPID in an attempt to [better signal the
+consequences](https://github.com/kubernetes/community/pull/1048/files#r159146536)
+of setting the option. Setting both `ShareProcessNamespace` and `HostPID` will
+cause a validation error.
+
+### Container Runtime Interface Changes
+
+Namespace options in the CRI are currently specified for both `PodSandbox` and
+`Container` creation requests via booleans in `NamespaceOption`:
+
+```
+message NamespaceOption {
+    // If set, use the host's network namespace.
+     bool host_network = 1;
+     // If set, use the host's PID namespace.
+     bool host_pid = 2;
+     // If set, use the host's IPC namespace.
+     bool host_ipc = 3;
+}
+```
+
+We will change `NamespaceOption` to use a `NamespaceMode` enumeration for the
+existing namespace options:
+
+```
+enum NamespaceMode {
+    POD       = 0;
+    CONTAINER = 1;
+    NODE      = 2;
+}
+
+// NamespaceOption provides options for Linux namespaces.
+message NamespaceOption {
+    // Network namespace for this container/sandbox.
+    // Runtimes must support: POD, NODE
+    NamespaceMode network = 1;
+    // PID namespace for this container/sandbox.
+    // Note: The CRI default is POD, but the v1.PodSpec default is CONTAINER.
+    // The kubelet's runtime manager will set this to CONTAINER explicitly for v1 pods.
+    // Runtimes must support: POD, CONTAINER, NODE
+    NamespaceMode pid = 2;
+    // IPC namespace for this container/sandbox.
+    // Runtimes must support: POD, NODE
+    NamespaceMode ipc = 3;
+}
+```
+
+Note that this breaks backwards compatibility in the CRI, which is still in
+alpha.
+
+The protocol default for a namespace is `POD` because that's the default for
+network and IPC, and we will consider making it the default for PID in `v2.Pod`.
+The kubelet will explicitly set `pid` to `CONTAINER` for `v1.Pod` by default so
+that the default behavior of `v1.Pod` does not change.
+
+This CRI design allows different namespace configuration for each of the
+containers in the pod and the sandbox, but currently we have no plans to support
+this in the Kubernetes API. The kubelet will translate namespace booleans from
+v1.PodSpec into a single `NamespaceMode` to be used for the sandbox and all
+regular and init containers in a pod.
+
+#### Targeting a Specific Container's Namespace
+
+Though we don't intend to support this in general pod configuration, there is a
+use case for mixed process namespaces within a single pod. [Troubleshooting
+Running Pods](troubleshooting-running-pods.md) allows inserting an ephemeral
+Debug Container in an existing, running pod. In order for this to be useful we
+want to share, within the pod, a process namespace between the new container
+performing the debugging and its existing target container.
+
+This is done with the additional `NamespaceMode` `TARGET` and field `target_id`:
+
+```
+enum NamespaceMode {
+    POD       = 0;
+    CONTAINER = 1;
+    NODE      = 2;
+    TARGET    = 3;
+}
+
+// NamespaceOption provides options for Linux namespaces.
+message NamespaceOption {
+    // Network namespace for this container/sandbox.
+    // Runtimes must support: POD, NODE
+    NamespaceMode network = 1;
+    // PID namespace for this container/sandbox.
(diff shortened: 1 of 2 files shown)


thockin:

contributors/design-proposals/pod-pid-namespace.md
```diff
@@ -60,18 +95,11 @@ The ability to disable shared PID namespaces is intended as a way to roll back
 to prior behavior in the event of unforeseen problems. It won't be possible to
 configure the behavior per-pod. We believe this is acceptable because:
```
smarterclayton: Should probably change this section.

contributors/design-proposals/pod-pid-namespace.md
```diff
@@ -11,23 +11,67 @@ Runtime Interface and links its rollout in Docker to that of the CRI.
...
+should be implemented by all runtimes where possible.
+
+### Option 2: Migration to Shared-only Namespaces
+
+In order to avoid breaking compatibility with existing images the default could
+not change until v2.Pod and wouldn't be completed until v1.Pod is deprecated,
+but choosing not to support isolated PID namespaces indefinitely would influence
+design choices. In this scenario we would avoid modifying the CRI or
+implementing isolated namespaces in other runtimes. Enabling the shared
+namespace in v1.Pod could happen in a new field or in an annotation.
```
yujuhong: I am having problems parsing the two options. What's the difference? Is it just the default?

verb:
Please stand by for a full rewrite based on our discussion in sig node.

>

verb:
@smarterclayton @yujuhong @derekwaynecarr This is now updated, internally consistent (I hope) and ready for review. It includes what we discussed previously in the SIG node meetup, which was adding a configuration option to both the Pod & CRI APIs.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,146 @@
...
+1.  easier troubleshooting of pods.
+1.  addressing [Docker's zombie
+    problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/)
+    by reaping orphaned zombies in the infra container.
+
+## Implementation
+
+### Kubernetes API Changes
+
+`v1.PodSecurityContext` gains a new field resembling the existing `hostPID`
```
thockin: Do we need this?  is shared PID somewhow a security escalation that we might want to deny users?

contributors/design-proposals/pod-pid-namespace.md
```diff
@@ -60,18 +95,11 @@ The ability to disable shared PID namespaces is intended as a way to roll back
 to prior behavior in the event of unforeseen problems. It won't be possible to
 configure the behavior per-pod. We believe this is acceptable because:
 
-* We have not identified a concrete use case requiring isolated PID namespaces.
-* Making PID namespace configurable requires changing the CRI, which we would
-  like to avoid since there are no use cases.
+*   We have not identified a concrete use case requiring isolated PID
```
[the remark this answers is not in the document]
verb: @thockin I'm not sure how this happened, but some of your comments are for an outdated revision and some are current. I suspect you started a review and I updated the doc in the middle of it. Sorry about that.

The latest version of the proposal dropped this section and lists supporting pid == 1 in the goals.

contributors/design-proposals/node/pod-pid-namespace.md
verb: From the discussion in https://github.com/kubernetes/kubernetes/issues/48937 and https://github.com/kubernetes/kubernetes/issues/1615, people want to be able to use PID namespace separation as a security feature.

derekwaynecarr:
@mrunalp - please review as well.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,150 @@
...
+docker versions less than 1.13.1, the runtime will respond with an error and the
+pod will not start.
 
-In a future release, SIG Node will recommend docker >= 1.12. Unless a compelling
-use case for isolated PID namespaces is discovered, we will remove the ability
-to disable the shared PID namespace in the subsequent release.
+The Container Runtime Interface (CRI) will be updated to explicitly support
+shared PID namespaces. When `sharedPID` is true, runtimes MUST share a PID
+namespace for the pod or return an error. When `sharedPID` is false, runtimes
+MAY implement isolated PID namespaces. That is, omitting this option defaults to
```
mrunalp: Wondering why this is left as a MAY. Can't it be MUST?

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,150 @@
...
+pre-existing runtime behavior. For Docker the default remains isolated PID
```
derekwaynecarr: nit: the default is not really runtime specific.  we need the api default to remain isolated across any runtime at this moment.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,150 @@
...
+The Docker runtime implements the pod sandbox as a container running the pause
+container image. The `shared_pid` of `LinuxSandboxSecurityContext` will be
+ignored by this runtime as the sandbox container will become the namespace
+that's shared with the containers in the pod.
+
+The dockershim will translate a `CreateContainerRequest` with `shared_pid` set
+to true in `LinuxContainerSecurityContext` to a call to Docker's create
+container request with `PidMode` set to the container ID of the sandbox
+container. All containers in the pod will then start in the same PID namespace
+as the sandbox container, with the `/pause` binary as PID 1.
```
mrunalp: Does it make sense to allow users to use their own custom sandbox image with pid 1 capabilities?

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,150 @@
...
+    HostPID bool
+    // Use a single shared PID namespace for the pod.
+    // Optional: Default to false.
+    // +k8s:conversion-gen=false
+    // +optional
+    SharedPID bool
+       ...
+```
+
+`PodSecurityContext` was chosen because it's where namespace configuration
```
yujuhong: I believe this is the internal API object..... The v1.PodSecurityContext does not contain the PID namespace fields...Could you update the text/code here to reflect that?

contributors/design-proposals/node/pod-pid-namespace.md
yujuhong: There's nothing preventing users to configure their runtime/shim to use a different image, if the runtime/shim supports it. I don't see we support that in kube/CRI in the near future though.

contributors/design-proposals/node/pod-pid-namespace.md
verb: Oh, I thought the rkt & hypervisor runtimes defaulted to shared PID so making isolated the default could break existing config. I'm happy to define the default as isolated, I was just trying to be careful about backwards compatibility.

contributors/design-proposals/node/pod-pid-namespace.md
verb: MAY is the status quo of no explicit default. I left it MAY for two reasons: 

1. Do isolated PID namespaces exist in all runtimes? What do hypervisor runtimes do? What about rkt? I'm not sure since I only use Docker.
2. Do any current runtimes default to shared PID? Do we care about being backwards compatible with them? If yes, then we can't retroactively define a default behavior for v1.Pod, I think.

contributors/design-proposals/node/pod-pid-namespace.md
verb: It'd be great to support such a feature, but I think it would be complicated. That's more towards the "general purpose init" to which I alluded in Non-goals.

contributors/design-proposals/node/pod-pid-namespace.md
verb: @yujuhong What do you think about an enum for the CRI? Something like: https://github.com/verb/kubernetes/commit/db76a1bc05d582a56f53cb56df3a65e6b236cd13

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,150 @@
...
-configure the behavior per-pod. We believe this is acceptable because:
+### Configurable PID Namespace Sharing
 
-* We have not identified a concrete use case requiring isolated PID namespaces.
-* Making PID namespace configurable requires changing the CRI, which we would
-  like to avoid since there are no use cases.
+We will add first class support for configuring isolated and shared PID
+namespaces for pods by adding a new boolean field `sharedPID` to the pod spec.
+When set to true, all containers in the pod will share a single PID namespace
+when supported by runtime. When not supported by the runtime, for example with
```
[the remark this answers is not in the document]
verb: Done

contributors/design-proposals/node/pod-pid-namespace.md
verb: I've rewritten this section to clearly define the behavior.

contributors/design-proposals/node/pod-pid-namespace.md
verb: I've rewritten this section to clearly define the behavior.

contributors/design-proposals/node/pod-pid-namespace.md
verb: Whoops, you're right. Updated.

contributors/design-proposals/node/pod-pid-namespace.md
yujuhong: > Do isolated PID namespaces exist in all runtimes? What do hypervisor runtimes do? What about rkt? I'm not sure since I only use Docker.

Hypervisor-based runtime cannot support host PID namespace, etc. We still need more changes even in Kubernetes to fully integrate with them. AFAIK, hypervisor-based runtimes that support docker container images AND Linux should be able to support shared and individual PID namespaces.

/cc @feiskyer 

> Do any current runtimes default to shared PID? Do we care about being backwards compatible with them? If yes, then we can't retroactively define a default behavior for v1.Pod, I think.

This change in the proposal is to ensure the behavior of v1.Pod in Kubernetes does not change, so I think it makes sense to enforce this in kubelet through CRI.

yujuhong: > @yujuhong What do you think about an enum for the CRI? Something like: verb/kubernetes@db76a1b

Personally I like that better than a set of booleans for toggling the behavior :-) We could name the `enum` NamespaceType to be more generic.
That is, however, an API breaking change for CRI. 

@mrunalp @Random-Liu @feiskyer thoughts?

verb: Ok, great, a default of isolated it is. Good point on the generic name.

Also I realized that we'll need to be able to target a container's namespace for debug containers, which might influence the API design. This could be done as a fourth NamespaceType: https://github.com/verb/kubernetes/commit/48fe4c260c6bf0ba5b4f8b71b71da08d12a9e4d2

dchen1107: @feiskyer for frakti? @Random-Liu for containerd @mrunalp for cri-o?

dchen1107: At sig-node, we talked about this, and the decision is introducing enum. Thanks!

Random-Liu: @yujuhong Enum does look cleaner to me. And we could easily change to use the new enum in cri-containerd.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,152 @@
...
 
-## Rollout Plan
+### Configurable PID Namespace Sharing
 
-SIG Node is planning to switch to the CRI as a default in 1.6, at which point
-users with Docker >= 1.12 will receive a shared PID namespace by default.
-Cluster administrators will be able to disable this behavior by providing a flag
-to the kubelet which will cause the dockershim to revert to previous behavior.
+We will add first class support for configuring isolated and shared PID
+namespaces for pods by adding a new boolean field `sharedPID` to the pod spec,
```
thockin: Thinking forward to a v2 API, we probably want the default to be shared, and private PID to be the exceptional case, right?  Does it make sense then to define this as `privatePID` with a default of `true` (but not `omitempty`).  Or can we imagine a plausible 3rd state (e.g. should this have been `pid: {private | shared | host}` more like Docker's native API?

This does touch on an interesting subject - defaulting in clients.  We know it's not super helpful for a client to apply defaults before sending up to the API server (since we lose track of what the user wanted vs what was defaulted).  

At some point, we opted NOT to demand Go-style "zero-value" default semantics, which means a bool with default true is legit, and a string with default non-empty is legit.  So how does a client know what the value of a field actually is?  Since optional fields are (almost) all pointers, clients that want to look at those fields can either nil-check and assume default meanings (hopefully as per our docs), or else run something we provide that applies defaults to their structs so they can use them.

But if the client understands fields that the server does not, it might default them to values and assume that the server implements them, when it doesn't.  We have some `omitempty` fields in our API - considering this (never thought about it this way) the client can't tell if a field is not implemented or just empty.

Should apiserver always be passing back fully-expanded values, such that if the client sees a nil it knows that field is not implemented on server?

I think we're sort of lost in the middle right now.

@kubernetes/sig-api-machinery-misc @lavalamp @caesarxuchao for more current thoughts on that topic.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,152 @@
...
+       // Use the host's pid namespace.
+       // Optional: Default to false.
+       // +k8s:conversion-gen=false
+       // +optional
+       HostPID bool `json:"hostPID,omitempty" protobuf:"varint,12,opt,name=hostPID"`
+    // Use a single shared PID namespace for the pod.
+       // Optional: Default to false.
+       // +k8s:conversion-gen=false
+       // +optional
+       SharedPID bool `json:"sharedPID,omitempty" protobuf:"varint,XX,opt,name=sharedPID"`
```
thockin: optional fields should all be pointers

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1048)

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,77 +1,250 @@
...
+       // Optional: Default to false.
+       // +k8s:conversion-gen=false
+       // +optional
+       SharedPID *bool `json:"sharedPID,omitempty" protobuf:"varint,XX,opt,name=sharedPID"`
+       ...
+```
+
+Setting both `SharedPID` and `HostPID` will cause a validation error.
+
+### Container Runtime Interface Changes
```
dchen1107: cc/ @Random-Liu for containerd and @mrunalp for cri-o on CRI changes.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,72 +1,251 @@
...
+### Configurable PID Namespace Sharing
 
-1. Add a test to verify all containers restart if the infra container
-   responsible for the PodSandbox dies. (Note: With Docker 1.12 if the source
-   of the PID namespace dies all containers sharing that namespace are killed
-   as well.)
-2. Modify the Infra container used by the Docker runtime to reap orphaned
-   zombies ([#36853](https://pr.k8s.io/36853)).
+We will add first class support for configuring isolated and shared PID
+namespaces for pods by adding a new boolean field `sharedPID` to the pod spec,
```
smarterclayton: The name leaves something to be desired, even if it is a mirror of `hostPID` (which itself is not a great name).

I'd probably prefer a name that covers the two important aspects of the flag:

1. that none of the containers will be PID 1
2. that all processes in each container will be able to see each other (and potentially interact)

`shareProcessNamespace` or something that captures "all containers will be part of the shared namespace" is better than `sharedPID`.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,72 +1,251 @@
...
-[1]: https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/
+Sharing a PID namespace between containers in a pod is discussed in
+[#1615](https://issues.k8s.io/1615) and enables:
+
+1.  signaling between containers, which is useful for side cars (e.g. for
+    signaling a daemon process after rotating logs).
+1.  easier troubleshooting of pods.
+1.  addressing [Docker's zombie
+    problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/)
+    by reaping orphaned zombies in the infra container.
```
smarterclayton: Please note the downsides of enabling this shared mode - sidecar containers that were previously isolated are no longer so, environment variables are now visible to all other processes, any "kill all" semantics used within the process are now broken, exec processes from other containers will now show up, etc.  This doc should clarify tradeoffs.

smarterclayton:
@mrunalp @runcom can you please comment on this with respect to CRI-O?  I'd like to see forward progress on this in 1.10.

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,72 +1,251 @@
...
+
+// NamespaceOption provides options for Linux namespaces.
+message NamespaceOption {
+    // Network namespace for this container/sandbox.
+    // Runtimes must support: POD, NODE
+    NamespaceMode network = 1;
+    // PID namespace for this container/sandbox.
+    // Note: The CRI default is POD, but the v1.PodSpec default is CONTAINER.
+    // The kubelet's runtime manager will set this to CONTAINER by default for v1 pods.
+    // Runtimes must support: POD, CONTAINER, NODE
```
yujuhong: Shouldn't this support TARGET too?

yujuhong: Follow-up question: do we want other namespaces to also support TARGET?

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,72 +1,251 @@
...
+    NamespaceMode pid = 2;
+    // IPC namespace for this container/sandbox.
+    // Runtimes must support: POD, NODE
+    NamespaceMode ipc = 3;
+    // Target Container or Sandbox ID for NamespaceMode of TARGET
```
yujuhong: Do we need to allow Sandbox IDs?

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,72 +1,251 @@
...
+this user configurable. The kubelet will translate namespace configuration from
+v1.PodSpec into a single `NamespaceOption` to be used for the sandbox and all
+regular and init containers in a pod.
+
+#### Targeting a Specific Container's Namespace
+
+Since we will support isolated PID namespaces indefinitely, we will need to be
+able to target the namespace of specific isolated containers to enable
+[Troubleshooting Running Pods](troubleshooting-running-pods.md). This is done
+with an additional field in `NamespaceOption`:
```
yujuhong: I think the previous subsection is more straightforward, but the introduction of TARGET makes the use case more complex. Could you elaborate a bit more on how the debug container would be configured in CRI?

In CRI today, you'd need to create a container in a PodSandbox, i.e., your debug container needs to join an existing PodSandbox, or you'd need to create a new PodSandbox for it. Are you going to allow pod A to join some of pod B's namespaces, or do you always require the debug container to join the PodSandbox in which the target container resides? What combination of NamespaceOptions and PodSandbox configs are valid?

contributors/design-proposals/node/pod-pid-namespace.md
```diff
@@ -1,72 +1,251 @@
...
+       // +optional
+       SharedPID *bool `json:"sharedPID,omitempty" protobuf:"varint,XX,opt,name=sharedPID"`
+       ...
+```
+
+Setting both `SharedPID` and `HostPID` will cause a validation error.
+
+### Container Runtime Interface Changes
+
+Namespace options in the CRI are currently specified for both `PodSandbox` and
```
yujuhong: The CRI changes in this subsection looks good to me. It makes the API more clear.

/cc a few CRI runtime maintainers since the change won't be backward compatible.
@Random-Liu @mrunalp @runcom

contributors/design-proposals/node/pod-pid-namespace.md
verb: Oh, no, and I don't think we want to. Containers should only be able to target their own pod. I'll remove that.

mrunalp:
We should be able to support this in CRI-O.

contributors/design-proposals/node/pod-pid-namespace.md
verb: It's a bit strange for the other namespaces since they don't support isolation. We don't need to support TARGET for things like network & ipc because POD & NODE are sufficient for the new container to be placed in the correct namespace.

contributors/design-proposals/node/pod-pid-namespace.md
verb: Added a section here. Let me know if I missed anything.

wking:
To get the infra-reaping benefits of pod mode with the isolation benefits of container-mode, I'd like to also specify a mode where the pod's worker containers have PID namespaces which are children of the infra container's PID namespace.  So:

* host PID namespace
    * infra PID namespace
        * worker PID namespace
        * worker PID namespace
        * …

That's the `pod-container` choice in the implementation I've floated in kubernetes-incubator/cri-o#1280.

verb:
@wking Interesting, what are the benefits of "pod-container" over "container"? Is it that with pod-container the infra container will inherit orphaned zombies somehow?

wking:
> Interesting, what are the benefits of "pod-container" over "container"? Is it that with pod-container the infra container will inherit orphaned zombies somehow?

Assuming there are no subreaper modifications, here's an example process tree with `pod-container`:

| PID (host) | PID (infra) | PID (container) | Reaper (host namespace) |
| -- | -- | -- | -- |
| 1 | - | - | - |
| 2 | 1 | - | 1 |
| 3 | 2 | 1 | 2 |
| 4 | 3 | 2 | 3 |

And here's the same setup with `container`:

| PID (host) | PID (infra) | PID (container) | Reaper (host namespace) |
| -- | -- | -- | -- |
| 1 | - | - | - |
| 2 | 1 | - | 1 |
| 3 | - | 1 | 1 |
| 4 | - | 2 | 3 |

And here's the same setup with `pod`:

| PID (host) | PID (infra) | Reaper (host namespace) |
| -- | -- | -- |
| 1 | - | - | - |
| 2 | 1 | 1 |
| 3 | 2 | 2 |
| 4 | 3 | 2 |

But I think the real benefit is that `pod-container` makes it easy for the infra process(es) to monitor processes in the worker containers (like the `pod` approach) while making it hard for those worker processes to mess with the infra processes (like the `container` approach).

yujuhong:
> That's the pod-container choice in the implementation I've floated in kubernetes-incubator/cri-o#1280.

I think this falls under the purview of the runtime given that we don't restrict how podsandbox must be implemented in CRI.

wking:
On Tue, Jan 23, 2018 at 10:30:50PM +0000, Yu-Ju Hong wrote:
> > That's the pod-container choice in the implementation I've floated
> > in kubernetes-incubator/cri-o#1280.
>
> I think this falls under the purview of the runtime given that we
> don't restrict how podsandbox must be implemented in CRI.

Are you saying that both my ‘container’ and ‘pod-container’ approaches
would be valid implementations of this proposal's ‘container’ choice,
and that there would be no standard way for a user to select between
them?

yujuhong:
@wking yes. As of now, CRI does not dictate how runtime implements the podsandbox.

PR #1050, opened 2017-09-12
correct Michael Rubin github handle

sjenning:


cblecker:
FYI @matchstick 
/lgtm

PR #1054, opened 2017-09-13
Volume topology aware scheduling design

msau42:
Proposal for a smarter scheduler that influences PV binding.

Part of kubernetes/features#121

/sig storage
/sig scheduling
/cc @kubernetes/sig-storage-proposals @kubernetes/sig-scheduling-proposals

### DIFF
--- a/contributors/design-proposals/sig-storage/volume-topology-scheduling.md
+++ b/contributors/design-proposals/sig-storage/volume-topology-scheduling.md
@@ -1,0 +1,580 @@
+# Volume Topology-aware Scheduling
+
+Authors: @msau42
+
+This document presents a detailed design for making the default Kubernetes
+scheduler aware of volume topology constraints, and making the
+PersistentVolumeClaim (PVC) binding aware of scheduling decisions.
+
+
+## Goals
+* Allow a Pod to request one or more topology-constrained Persistent
+Volumes (PV) that are compatible with the Pod's other scheduling
+constraints, such as resource requirements and affinity/anti-affinity
+policies.
+* Support arbitrary PV topology constraints (i.e. node,
+rack, zone, foo, bar).
+* Support topology constraints for statically created PVs and dynamically
+provisioned PVs.
+* No scheduling latency performance regression for Pods that do not use
+topology-constrained PVs.
+
+
+## Non Goals
+* Fitting a pod after the initial PVC binding has been completed.
+    * The more constraints you add to your pod, the less flexible it becomes
+in terms of placement.  Because of this, tightly constrained storage, such as
+local storage, is only recommended for specific use cases, and the pods should
+have high priority.
+    * Pod priorities could handle this scenario.  PV taints/tolerations could
+also handle this case.
+* Binding decision considering scheduling constraints from two or more pods
+sharing the same PVC.
+    * The scheduler itself only handles one pod at a time.  It’s possible the
+two pods may not run at the same time either, so there’s no guarantee that you
+will know both pod’s requirements at once.
+    * For two+ pods simultaneously sharing a PVC, this scenario may require an
+operator to schedule them together.
+    * For two+ pods non-simultaneously sharing a PVC, this scenario could be
+handled by pod priorities.
+
+
+## Problem
+Volumes can have topology constraints that restrict the set of nodes that the
+volume is accessible from.  For example, a GCE PD can only be accessed from a
+single zone, and a local disk can only be accessible from a single node.  In the
+future, there could be other topology constraints, such as rack or region.
+
+A pod that uses such a volume must be scheduled to a node that fits within the
+volume’s topology constraints.  In addition, a pod can have further constraints
+and limitations, such as the pod’s resource requests (cpu, memory, etc), and
+pod/node affinity and anti-affinity policies.
+
+Currently, the process of binding and provisioning volumes are done before a pod
+is scheduled.  Therefore, it cannot take into account any of the pod’s other
+scheduling constraints.  This makes it possible for the PV controller to bind a
+PVC to a PV or provision a PV with constraints that can make a pod unschedulable.
+
+### Examples
+* In multizone clusters, the PV controller has a hardcoded heuristic to provision
+PVCs for StatefulSets spread across zones.  If that zone does not have enough
+cpu/memory capacity to fit the pod, then the pod is stuck in pending state because
+its volume is bound to that zone.
+* Local storage exasperates this issue.  The chance of a node not having enough
+cpu/memory is higher than the chance of a zone not having enough cpu/memory.
+* Local storage PVC binding does not have any node spreading logic.  So local PV
+binding will very likely conflict with any pod anti-affinity policies if there is
+more than one local PV on a node.
+* A pod may need multiple PVCs.  As an example, one PVC can point to a local SSD for
+fast data access, and another PVC can point to a local HDD for logging.  Since PVC
+binding happens without considering if multiple PVCs are related, it is very likely
+for the two PVCs to be bound to local disks on different nodes, making the pod
+unschedulable.
+* For multizone clusters and deployments requesting multiple dynamically provisioned
+zonal PVs, each PVC Is provisioned independently, and is likely to provision each PV
+In different zones, making the pod unschedulable.
+
+To solve the issue of initial volume binding and provisioning causing an impossible
+pod placement, volume binding and provisioning should be more tightly coupled with
+pod scheduling.
+
+
+## Background
+In 1.7, we added alpha support for local PVs (TODO: link) with node affinity.
+You can specify a PV object with node affinity, and if a pod is using such a PV,
+the scheduler will evaluate the PV node affinity in addition to the other
+scheduling predicates.  So far, the PV node affinity only influences pod
+scheduling once the PVC is already bound.  The initial PVC binding decision was
+unchanged.  This proposal addresses the initial PVC binding decision.
+
+
+## Design
+The design can be broken up into a few areas:
+* User-facing API to invoke new behavior
+* Integrating PV binding with pod scheduling
+* Binding multiple PVCs as a single transaction
+* Recovery from kubelet rejection of pod
+* Making dynamic provisioning topology-aware
+
+For the alpha phase, only the user-facing API and PV binding and scheduler
+integration are necessary.  The remaining areas can be handled in beta and GA
+phases.
+
+### User-facing API
+This new binding behavior will apply to all unbound PVCs and can be controlled
+through install-time configuration passed to the scheduler and controller-manager.
+
+In alpha, this configuration can be controlled by a feature gate,
+`VolumeTopologyScheduling`.
+
+#### Alternative
+An alternative approach is to only trigger the new behavior only for volumes that
+have topology constraints.  A new annotation can be added to the StorageClass to
+indicate that its volumes have constraints and will need to use the new
+topology-aware scheduling logic.
+
+```
+// Value is “true” or “false”.  Default is false.
+const AlphaVolumeTopologySchedulingAnnotation = “volume.alpha.kubernetes.io/topology-scheduling”
+```
+
+While this approach can let us introduce the new behavior gradually, it has a
+few downsides:
+* StorageClass will be required to use this new logic, even if dynamic
+provisioning is not used (in the case of local storage).
+* We have to maintain two different paths for volume binding.
+* We will be depending on the storage admin to correctly configure the
+StorageClasses.
+
+For those reasons, we prefer to switch the behavior for all unbound PVCs and
+clearly communicate these changes to the community before changing the behavior
+by default in GA.
+
+### Integrating binding with scheduling
+For the alpha phase, the focus is on static provisioning of PVs to support
+persistent local storage.
+
+TODO: flow chart
+
+The proposed new workflow for volume binding is:
+1. Admin statically creates PVs and/or StorageClasses.
+2. User creates unbound PVC and there are no prebound PVs for it.
+3. **NEW:** PVC binding and provisioning is delayed until a pod is created that
+references it.
+4. User creates a pod that uses the PVC.
+5. Pod starts to get processed by the scheduler.
+6. **NEW:** A new predicate function, called MatchUnboundPVCs, will look at all of
+a Pod’s unbound PVCs, and try to find matching PVs for that node based on the
+PV topology.  If there are no matching PVs, then it checks if dynamic
+provisioning is possible for that node.
+7. **NEW:** The scheduler continues to evaluate priorities.  A new priority
+function, called PrioritizeUnboundPVCs, will get the PV matches per PVC per
+node, and compute a priority score based on various factors.
+8. **NEW:** After evaluating all the existing predicates and priorities, the
+scheduler will pick a node, and call a new assume function, AssumePVCs,
+passing in the Node.  The assume function will check if any binding or
+provisioning operations need to be done.  If so, it will update the PV cache to
+mark the PVs with the chosen PVCs.
+9. **NEW:** If PVC binding or provisioning is required, we do NOT AssumePod.
+Instead, a new bind function, BindPVCs, will be called asynchronously, passing
+in the selected node.  The bind function will prebind the PV to the PVC, or
+trigger dynamic provisioning, and then wait until the PVCs are bound
+successfully or encounter failure.  Then, it always sends the Pod through the
+scheduler again for reasons explained later.
+10. When a Pod makes a successful scheduler pass once all PVCs are bound, the
+scheduler assumes and binds the Pod to a Node.
+11. Kubelet starts the Pod.
+
+This new workflow will have the scheduler handle unbound PVCs by choosing PVs
+and prebinding them to the PVCs.  The PV controller completes the binding
+transaction, handling it as a prebound PV scenario.
+
+Prebound PVCs and PVs will still immediately be bound by the PV controller.
+
+One important point to note is that for the alpha phase, manual recovery is
+required in following error conditions:
+* A Pod has multiple PVCs, and only a subset of them successfully bind.
+* The scheduler chose a PV and prebound it, but the PVC could not be bound.
+
+In both of these scenarios, the primary cause for these errors is if a user
+prebinds other PVCs to a PV that the scheduler chose.  Some workarounds to
+avoid these error conditions are to:
+* Prebind the PV instead.
+* Separate out volumes that the user prebinds from the volumes that are
+available for the system to choose from by StorageClass.
+
+In addition, this new design will cause functional regression for the
+following use cases:
+* While a Pod is running, a user deletes the PVC and PV, and then recreates them.
+    * With the current behavior, the PVC will rebind to the same or different PV.
+    * With this new behavior, the PVC will not rebind because binding would only
+be triggered when the Pod has to go through scheduling.  This could cause
+problems with Pod volume cleanup because the PVC will no longer be bound.
+    * But deleting the PVC and PV (especially both) while a Pod is still using it
+is also dangerous and should be prevented in the first place.  There could be
+serious problems if you recreate the PVC and it rebinds to a different PV.  If
+the PV had a reclaim policy of Delete, then deleting the PVC could cause the
+underlying disk to get deleted while the Pod is still using it.
+
+
+#### PV Controller Changes
+When the feature gate is enabled, the PV controller needs to skip binding and
+provisioning all unbound PVCs that don’t have prebound PVs, and let it come
+through the scheduler path.  This is the block in `syncUnboundClaim` where
+`claim.Spec.VolumeName == “”` and there is no PV that is prebound to the PVC.
+
+No other state machine changes are required.  The PV controller continues to
+handle the remaining scenarios without any change.
+
+The methods to find matching PVs for a claim, prebind PVs, and to invoke and
+handle dynamic provisioning need to be refactored for use by the new scheduler
+functions.
+
+#### Scheduler Changes
+
+##### Predicate
+A new predicate function checks all of a Pod's unbound PVCs can be satisfied
+by existing PVs or dynamically provisioned PVs that are
+topologically-constrained to the Node.
+```
+MatchUnboundPVCs(pod *v1.Pod, node *v1.Node) (canBeBound bool, err error)
+```
+1. If all the Pod’s PVCs are bound, return true.
+2. Otherwise try to find matching PVs for all of the unbound PVCs in order of
+decreasing requested capacity.
+3. Walk through all the PVs.
+4. Find best matching PV for the PVC where PV topology is satisfied by the Node.
+5. Temporarily cache this PV in the PVC object, keyed by Node, for fast
+processing later in the priority and bind functions.
+6. Return true if all PVCs are matched.
+7. If there are still unmatched PVCs, check if dynamic provisioning is possible.
+For this alpha phase, the provisioner is not topology aware, so the predicate
+will just return true if there is a provisioner specified (internal or external).
+8. Otherwise return false.
+
+TODO: caching format and details
+
+##### Priority
+After all the predicates run, there is a reduced set of Nodes that can fit a
+Pod. A new priority function will rank the remaining nodes based on the
+unbound PVCs and their matching PVs.
+```
+PrioritizeUnboundPVCs(pod *v1.Pod, filteredNodes HostPriorityList) (rankedNodes HostPriorityList, err error) 
+```
+1. For each Node, get the cached PV matches for the Pod’s PVCs.
+2. Compute a priority score for the Node using the following factors:
+    1. How close the PVC’s requested capacity and PV’s capacity are.
+    2. Matching static PVs is preferred over dynamic provisioning.
+
+TODO (beta): figure out weights and exact calculation
+
+##### Assume
+Once all the predicates and priorities have run, then the scheduler picks a
+Node.  Then we can bind or provision PVCs for that Node.  For better scheduler
+performance, we’ll assume that the binding will likely succeed, and update the
+PV cache first.  Then the actual binding API update will be made
+asynchronously, and the scheduler can continue processing other Pods.
+
+For the alpha phase, the AssumePVCs function will be directly called by the
+scheduler.  We’ll consider creating a generic scheduler interface in a
+subsequent phase.
+
+```
+AssumePVCs(pod *v1.Pod, node *v1.Node) (pvcBindingRequired bool, err error)
+```
+1. If all the Pod’s PVCs are bound, return false.
+2. For static PV binding:
+    1. Get the cached matching PVs for the PVCs on that Node.
+    2. Validate the actual PV state.
+    3. Mark PV.ClaimRef in the PV cache.
+3. For in-tree and external dynamic provisioning:
+    1. Nothing.
+4. Return true.
+
+
+##### Bind
+If AssumePVCs returns pvcBindingRequired, then the BindPVCs function is called
+as a go routine.  Otherwise, we can continue with assuming and binding the Pod
+to the Node.
+
+For the alpha phase, the BindUnboundPVCs function will be directly called by the
+scheduler.  We’ll consider creating a generic scheduler interface in a subsequent
+phase.
+
+```
+BindUnboundPVCs(pod *v1.Pod, node *v1.Node) (err error)
+```
+1. For static PV binding:
+    1. Prebind the PV by updating the `PersistentVolume.ClaimRef` field.
+    2. If the prebind fails, revert the cache updates.
+    3. Otherwise, wait for the PVCs to be bound, PVC/PV object is deleted, or
+PV.ClaimRef field is cleared
+2. For in-tree dynamic provisioning:
+    1. Make the provision call, which will create a new PV object that is
+prebound to the PVC.
+    2. If provisioning is successful, wait for the PVCs to be bound, PVC/PV
+object is deleted, or PV.ClaimRef field is cleared
+3. For external provisioning:
+    1. Set the annotation for external provisioners.
+4. Send Pod back through scheduling, regardless of success or failure.
+    1. In the case of success, we need one more pass through the scheduler in
+order to evaluate other volume predicates that require the PVC to be bound, as
+described below.
+    2. In the case of failure, we want to retry binding/provisioning.
+
+Note that for external provisioning, we do not wait for the PVCs to be bound, so
+the Pod will be sent through scheduling repeatedly until the PVCs are bound.
+This is because there is no function call for external provisioning, so if we
+did wait, we could be waiting forever for the PVC to bind.  It’s possible that
+in the meantime, a user could create a PV that satisfies the PVC and doesn’t
+need the external provisioner anymore.
+
+TODO: pv controller has a high resync frequency, do we need something similar
+for the scheduler too
+
+##### Pod preemption considerations
+The MatchUnboundPVs predicate does not need to be re-evaluated for pod
+preemption.  Preempting a pod that uses a PV will not free up capacity on that
+node because the PV lifecycle is independent of the Pod’s lifecycle.
+
+##### Other scheduler predicates
+Currently, there are a few existing scheduler predicates that require the PVC
+to be bound.  The bound assumption needs to be changed in order to work with
+this new workflow.
+
+TODO: how to handle race condition of PVCs becoming bound in the middle of
+running predicates?  Do we need to impose ordering on the predicates?
+
+###### Max PD Volume Count Predicate
+This predicate checks the maximum number of PDs per node is not exceeded.  It
+needs to be integrated into the binding decision so that we don’t bind or
+provision a PV if it’s going to cause the node to exceed the max PD limit.  But
+until it is integrated, we need to make one more pass in the scheduler after all
+the PVCs are bound.  The current copy of the predicate in the default scheduler
+has to remain to account for the already-bound volumes.
+
+###### Volume Zone Predicate
+This predicate makes sure that the zone label on a PV matches the zone label of
+the node.  If the volume is not bound, this predicate can be ignored, as the
+binding logic will take into account zone constraints on the PV.
+
+However, this assumes that zonal PVs like GCE PDs and AWS EBS have been updated
+to use the new PV topology specification, which is not the case as of 1.8.  So
+until those plugins are updated, the binding and provisioning decisions will be
+topology-unaware, and we need to make one more pass in the scheduler after all
+the PVCs are bound.
+
+This predicate needs to remain in the default scheduler to handle the
+already-bound volumes using the old zonal labeling.  It can be removed once that
+mechanism is deprecated and unsupported.
+
+###### Volume Node Predicate
+This is a new predicate added in 1.7 to handle the new PV node affinity.  It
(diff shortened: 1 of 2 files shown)


contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+
+TODO: pv controller has a high resync frequency, do we need something similar
+for the scheduler too
+
+##### Pod preemption considerations
+The MatchUnboundPVs predicate does not need to be re-evaluated for pod
+preemption.  Preempting a pod that uses a PV will not free up capacity on that
+node because the PV lifecycle is independent of the Pod’s lifecycle.
+
+##### Other scheduler predicates
```
resouer: Could you please also evaluation if this new predicate need special care in `factory.go` according to: https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/scheduler/algorithmprovider/defaults/defaults.go#L70-L76

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+is scheduled.  Therefore, it cannot take into account any of the pod’s other
+scheduling constraints.  This makes it possible for the PV controller to bind a
+PVC to a PV or provision a PV with constraints that can make a pod unschedulable.
+
+### Examples
+* In multizone clusters, the PV controller has a hardcoded heuristic to provision
+PVCs for StatefulSets spread across zones.  If that zone does not have enough
+cpu/memory capacity to fit the pod, then the pod is stuck in pending state because
+its volume is bound to that zone.
+* Local storage exasperates this issue.  The chance of a node not having enough
```
gnufied: exacerbate ?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+scheduler again for reasons explained later.
+10. When a Pod makes a successful scheduler pass once all PVCs are bound, the
+scheduler assumes and binds the Pod to a Node.
+11. Kubelet starts the Pod.
+
+This new workflow will have the scheduler handle unbound PVCs by choosing PVs
+and prebinding them to the PVCs.  The PV controller completes the binding
+transaction, handling it as a prebound PV scenario.
+
+Prebound PVCs and PVs will still immediately be bound by the PV controller.
```
gnufied: What you mean by this? if they were prebound - why do they need binding?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+avoid these error conditions are to:
+* Prebind the PV instead.
+* Separate out volumes that the user prebinds from the volumes that are
+available for the system to choose from by StorageClass.
+
+In addition, this new design will cause functional regression for the
+following use cases:
+* While a Pod is running, a user deletes the PVC and PV, and then recreates them.
+    * With the current behavior, the PVC will rebind to the same or different PV.
+    * With this new behavior, the PVC will not rebind because binding would only
```
gnufied: On EBS at least, since you can't delete a volume that is mounted and in use. Deleting a dynamically bound PVC results in PV being "Released" (I think).  So when user creates a new PVC with same name (because user does not want to change pod's yaml), a new PV is created. So we get 2 PVs (one released and one bound) mapped to same PVC.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+MatchUnboundPVCs(pod *v1.Pod, node *v1.Node) (canBeBound bool, err error)
+```
+1. If all the Pod’s PVCs are bound, return true.
+2. Otherwise try to find matching PVs for all of the unbound PVCs in order of
+decreasing requested capacity.
+3. Walk through all the PVs.
+4. Find best matching PV for the PVC where PV topology is satisfied by the Node.
+5. Temporarily cache this PV in the PVC object, keyed by Node, for fast
+processing later in the priority and bind functions.
+6. Return true if all PVCs are matched.
```
gnufied: What happens if matched PVs are deleted or PVC is resized since scheduler started matching? I am just trying to think, there appears to be a delay between when PVs/node is matched and when binding happens. Does the pod go through `MatchUnbounPVCs` again if something changed since match began?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+as a go routine.  Otherwise, we can continue with assuming and binding the Pod
+to the Node.
+
+For the alpha phase, the BindUnboundPVCs function will be directly called by the
+scheduler.  We’ll consider creating a generic scheduler interface in a subsequent
+phase.
+
+```
+BindUnboundPVCs(pod *v1.Pod, node *v1.Node) (err error)
+```
```
gnufied: Do we check again if PVC constraints are still met? Can we clarify how this function accesses PV cache - because from what I see it only get pod and hence it must only have pvc name.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: Not all storage providers may be able to detect if a volume is in use before deleting. But assuming it can detect and fail the deletion, let's break this up into 2 scenarios actually: 

1. Delete PVC, PV is released, but not deleted.  Recreate PVC.
If the container restarts, the old volume is still used in the new container.  If the Pod restarts, then it goes back through scheduling, and we rebind the PVC at that time.  Pod termination works because we only need the PV, not the PVC object still exists.  So no regression here.

2. Delete both PVC and PV objects.  Recreate PVC and PV objects.
Pod termination still works because PV object is still in the actual state.  If kubelet restarts, volume reconstruction will recreate the PV object.  No regression here.

So actually, I made a wrong assumption that we needed the PVC object to be bound for Pod teardown.  kubelet volume manager only needs the PV object, and it can reconstruct it too if kubelet crashes.  Of course, this also assumes that the volume plugin implements reconstruction correctly.

I'll remove this section unless we can think of another scenario that won't work with the new design.  I'll leave the discussion of if we want to prevent PVC and PV deletion when it's still in use to a separate issue.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
liggitt: > if the user set PVC.volumeName, then the PV controller will set PV.ClaimRef.

That should not be the case... a PVC user cannot target a specific PV by name.

msau42: I just tried it and it works.  I created 2 PVs of different sizes, and then created a PVC that requests the size of the smaller PV, but specifies the name of the larger PV.  And the volume got bound to the specified PV name.  From my understanding, it's been this way since the beginning.

liggitt: > From my understanding, it's been this way since the beginning.

I don't think so. It looks like that was added in 1.3 in https://github.com/kubernetes/kubernetes/pull/24331/files#diff-3aaafc34911c234537e0076b5ed40b37R202. I would not expect a namespaced user to be able to target a particular PV.

msau42: Regardless, it's been like this for many releases now.  If we want to change it to not allow prebinding PVCs, then it can be considered separately from this proposal.  This proposal is not changing the current prebinding behavior.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
resouer: Cool, thanks!

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
jsafrane: @liggitt, I thought it was common practice - a Pod author can set `pod.spec.nodeName` to bypass scheduling. This is the same thing.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+through install-time configuration passed to the scheduler and controller-manager.
+
+In alpha, this configuration can be controlled by a feature gate,
+`VolumeTopologyScheduling`.
+
+#### Alternative
+An alternative approach is to only trigger the new behavior only for volumes that
+have topology constraints.  A new annotation can be added to the StorageClass to
+indicate that its volumes have constraints and will need to use the new
+topology-aware scheduling logic.
```
NickrenREN: That is to say: when pv controller synchronized unbound pvc, it will always have to check storageclass's annotation to see which volume binding path it should go?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+For the alpha phase, the BindUnboundPVCs function will be directly called by the
+scheduler.  We’ll consider creating a generic scheduler interface in a subsequent
+phase.
+
+```
+BindUnboundPVCs(pod *v1.Pod, node *v1.Node) (err error)
+```
+1. For static PV binding:
+    1. Prebind the PV by updating the `PersistentVolume.ClaimRef` field.
+    2. If the prebind fails, revert the cache updates.
```
jsafrane: Is `pod.spec.nodeName` already set? Will it be reverted? It probably should.

jsafrane: If I understand it correctly, at least in alpha it won't roll back any previously saved PVs, right? Does it then revert this binding from cache? It should probably stay there.

liggitt: I can't tell if you mean the pod's node name is set in the API. If so, that's irreversible

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+2. Otherwise try to find matching PVs for all of the unbound PVCs in order of
+decreasing requested capacity.
+3. Walk through all the PVs.
+4. Find best matching PV for the PVC where PV topology is satisfied by the Node.
+5. Temporarily cache this PV in the PVC object, keyed by Node, for fast
+processing later in the priority and bind functions.
+6. Return true if all PVCs are matched.
+7. If there are still unmatched PVCs, check if dynamic provisioning is possible.
+For this alpha phase, the provisioner is not topology aware, so the predicate
+will just return true if there is a provisioner specified (internal or external).
```
NickrenREN: What does this mean ?
> if there is a provisioner specified

Do you mean the pvc provisioner annotation is set ?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: Yes the alternative is to have the trigger be in the storageclass, so you need to check for the annotation every time you handle unbound volumes to determine if you should delay binding or not.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+follow the new workflow closely with some modifications.
+* The FindUnboundPVCs predicate function needs to get provisionable capacity per
+topology dimension from the provisioner somehow.
+* The PrioritizeUnboundPVCs priority function can add a new priority score factor
+based on available capacity per node.
+* The BindUnboundPVCs bind function needs to pass in the node to the provisioner.
+The internal and external provisioning APIs need to be updated to take in a node
+parameter.
+
+
```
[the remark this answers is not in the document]
msau42: This is definitely going to make it harder to run your own scheduler, and I think this is also one of the reasons why the scheduling sig added the scheduler extender interface, so that you can still use the default scheduler + your custom extensions.  In general, the scheduling team has been adding many advanced scheduling features to the default scheduler that it's becoming harder to run your own already.

If you don't use the default scheduler, and you use persistent volumes, then you'll have to pull this code and integrate it into your custom scheduler.  Of course, we will have to give notice well in advance.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+### Binding multiple PVCs in one transaction
+For the alpha phase, this is not required.  Since the scheduler is serialized, a
+partial binding failure should be a rare occurrence.  Manual recovery will need
+to be done for those rare failures.
+
+### Recovering from kubelet rejection of pod
+Again, for alpha phase, manual intervention will be required.  For beta, we’ll
+have to consider some sort of rollback mechanism for more robustness.  Perhaps
+this rollback mechanism can also be used for binding multiple PVCs in one
```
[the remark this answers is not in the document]
jsafrane: I think that recycler (and deleter) is safe here, PV controller should be robust enough already.

On the other hand, any rollback in recycler may conflict with binding in PV controller. Scheduler sets pv.spec.claimRef and leaves the rest to PV controller, right? PV controller will save pvc.spec.volumeName, update status and some annotations. In case the scheduler wants to revert the binding, clearing pv.spec.claimRef won't be enough, it needs to clear everything. And it will race with the PV controller trying to fix the binding, it tries to fix the status and binding if something overwrites it.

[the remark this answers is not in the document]
msau42: Yah we will need to add some new state so that the PV controller will not try to fix the binding on a rollback.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
jsafrane: I thought more about it, if two pods use the same unbound PVC and two different schedulers bind this PVC to a different PV (i.e. one saves `pv1.spec.claimRef = pvc` and the second one `pv2.spec.claimRef = pvc`), PV controller is smart enough to recover from this conflict by un-binding one of these PVs, i.e. it clears say `pv1.spec.claimRef=nil` (and deletes it when necessary).

If both schedulers implement step `"9. **NEW:** If PVC binding or provisioning is required, we do NOT AssumePod ... then **wait until the PVCs are bound successfully** or encounter failure"` then the PVC gets eventually bound to one of these PVs. Not necessarily the one that a particular scheduler has chosen, but that should not be a problem, right?

So, we're fine as long as both schedulers implement the rules you specified. They may generate some noise by provisioning an unnecessary PV or do extra binding, but PV controller should clean up this mess.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
jsafrane: edit: rollback in recycler -> rollback in scheduler

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+Instead, a new bind function, BindPVCs, will be called asynchronously, passing
+in the selected node.  The bind function will prebind the PV to the PVC, or
+trigger dynamic provisioning, and then wait until the PVCs are bound
+successfully or encounter failure.  Then, it always sends the Pod through the
+scheduler again for reasons explained later.
+10. When a Pod makes a successful scheduler pass once all PVCs are bound, the
+scheduler assumes and binds the Pod to a Node.
+11. Kubelet starts the Pod.
+
+This new workflow will have the scheduler handle unbound PVCs by choosing PVs
```
liggitt: moving the binding responsibility from the binder controller into the scheduler means you no longer have a single writer, and that all external schedulers must also implement PVC binding, right?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
msau42: Thanks for the analysis.  It should be fine, but the only potential problem is if the 2nd scheduler bound the PVC in a way that cannot satisfy the all the predicates in the 1st scheduler.  In general, we should only recommend that one scheduler handle PVC binding.

Can @kubernetes/sig-scheduling-pr-reviews comment on the practice of running two schedulers against the same Pod?  Is that a supported configuration, or do we only support a model where:
- You run your own scheduler by itself
- You run default scheduler + your custom extensions

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: Yes, if you use a custom scheduler, and not the default scheduler, then you will have to pull in the PV matching and binding logic into your own scheduler.  We can try to modularize the library calls as much as possible to make it easier to import the methods.  Note that we don't pull in the entire binding state machine into the scheduler.  The scheduler's responsibilities are to:
1. Find an appropriate PV for unbound PVCs
2. Initiate (but not complete) the binding by setting the PV.ClaimRef.

liggitt: > The scheduler's responsibilities are to:
>
> Find an appropriate PV for unbound PVCs
> Initiate (but not complete) the binding by setting the PV.ClaimRef.

the split of binding responsibilities definitely makes the PV binder more difficult to reason about...

msau42: How about this?

* The scheduler handles choosing an appropriate PV for the PVC based on all the pod's requirements.  It's like a user manually picking their volumes, except automated.
* The PV controller handles the mechanics of binding the PVC and PV together, and the cleanup lifecycle.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+8. **NEW:** After evaluating all the existing predicates and priorities, the
+scheduler will pick a node, and call a new assume function, AssumePVCs,
+passing in the Node.  The assume function will check if any binding or
+provisioning operations need to be done.  If so, it will update the PV cache to
+mark the PVs with the chosen PVCs.
+9. **NEW:** If PVC binding or provisioning is required, we do NOT AssumePod.
+Instead, a new bind function, BindPVCs, will be called asynchronously, passing
+in the selected node.  The bind function will prebind the PV to the PVC, or
+trigger dynamic provisioning, and then wait until the PVCs are bound
+successfully or encounter failure.  Then, it always sends the Pod through the
```
bsalamat: IIUC, scheduler revisits scheduling of the pod to the node in a second round. What if the pod is no longer schedulable on the node, because, for example, a different pod is already scheduled there and the node no longer has enough resources. Doesn't cause PV stranding and starvation of the pod?

msau42: Yes that is possible this design (and also is the current behavior).  One possible fix that we're thinking of for beta, is to implement a PVC binding "rollback" if the volumes have not been mounted by any pods yet.

Also in general for local storage once it's bound, since your node choices get constrained to 1, we are recommending that those pods have high priority to avoid starvation.

bsalamat: Yeah, I think having a way of "unbinding" volumes is necessary, but even then the logic would be a bit complex. It will be something like this:
1. Try to find a node for the pod.
2. Bind the volumes.
3. Try to schedule the pod again (probably by recording which node was picked at step 1 and try that node)
4. If step 3 fails, unbind the volumes and go to 1 (after wiping the previously chosen node)

Step 3 is not an orthodox way of scheduling. It probably needs some changes to the guts of scheduler.

msau42: We also need to remember though, that rolling back a volume binding is a lot harder and should be less common than rolling back pod bindings.  Once a pod starts using the volume we cannot unbind it, even if scheduling fails.  

For zonal storage, failing scheduling after binding is more rare because you have more nodes to choose from.  For local storage, it's more likely since there's only one node, so your pods should be higher priority to force preemption

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+* No scheduling latency performance regression for Pods that do not use
+topology-constrained PVs.
+
+
+## Non Goals
+* Fitting a pod after the initial PVC binding has been completed.
+    * The more constraints you add to your pod, the less flexible it becomes
+in terms of placement.  Because of this, tightly constrained storage, such as
+local storage, is only recommended for specific use cases, and the pods should
+have high priority.
```
ianchakeres: /high priority./higher scheduling priority than unconstrained pods./

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+will know both pod’s requirements at once.
+    * For two+ pods simultaneously sharing a PVC, this scenario may require an
+operator to schedule them together.
+    * For two+ pods non-simultaneously sharing a PVC, this scenario could be
+handled by pod priorities.
+
+
+## Problem
+Volumes can have topology constraints that restrict the set of nodes that the
+volume is accessible from.  For example, a GCE PD can only be accessed from a
```
ianchakeres: /the volume is accessible from./can mount them./

alternatively /can access them./

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+single zone, and a local disk can only be accessible from a single node.  In the
```
ianchakeres: /accessible/accessed/

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+unbound PVCs and their matching PVs.
+```
+PrioritizeUnboundPVCs(pod *v1.Pod, filteredNodes HostPriorityList) (rankedNodes HostPriorityList, err error) 
+```
+1. For each Node, get the cached PV matches for the Pod’s PVCs.
+2. Compute a priority score for the Node using the following factors:
+    1. How close the PVC’s requested capacity and PV’s capacity are.
+    2. Matching static PVs is preferred over dynamic provisioning.
+
+TODO (beta): figure out weights and exact calculation
```
ianchakeres: Assuming that multiple node fit the constraints and are all equal in weight, is the ranked order of these nodes random or deterministically based on some identifier?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+After all the predicates run, there is a reduced set of Nodes that can fit a
+Pod. A new priority function will rank the remaining nodes based on the
+unbound PVCs and their matching PVs.
+```
+PrioritizeUnboundPVCs(pod *v1.Pod, filteredNodes HostPriorityList) (rankedNodes HostPriorityList, err error) 
+```
+1. For each Node, get the cached PV matches for the Pod’s PVCs.
+2. Compute a priority score for the Node using the following factors:
+    1. How close the PVC’s requested capacity and PV’s capacity are.
+    2. Matching static PVs is preferred over dynamic provisioning.
```
ianchakeres: Why are static PVs preferred over dynamic?

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+In both of these scenarios, the primary cause for these errors is if a user
+prebinds other PVCs to a PV that the scheduler chose.  Some workarounds to
+avoid these error conditions are to:
+* Prebind the PV instead.
+* Separate out volumes that the user prebinds from the volumes that are
+available for the system to choose from by StorageClass.
+
+In addition, this new design will cause functional regression for the
+following use cases:
+* While a Pod is running, a user deletes the PVC and PV, and then recreates them.
```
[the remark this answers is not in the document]
kow3ns: So you unmount the Pods storage and it doesn't die? What application can tolerate having a volume unmounted from under it without killing the underlying application.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+clearly communicate these changes to the community before changing the behavior
+by default in GA.
+
+### Integrating binding with scheduling
+For the alpha phase, the focus is on static provisioning of PVs to support
+persistent local storage.
+
+TODO: flow chart
+
+The proposed new workflow for volume binding is:
```
kow3ns: Does this assume that workload spreading across failure domains will be achieved by anti-affinity predicates?

msau42: Yes

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
msau42: The volume does not get unmounted while the Pod is still running, even if the PVC/PV are deleted.  Unmount only happens when the Pod is terminated.  

However, if PVC is deleted, and reclaim policy is Delete, then PV controller will immediately try to delete the volume.  Luckily, nothing bad happens because it depends on the storage provider to prevent volume deletion while the disk is still attached to a node.  

But this is potentially a problem for local storage, which doesn't have that protection.  I need to think about if there's a reliable way to detect and prevent it (in a separate issue).

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: It's assuming that you had specifically statically created the PVs for your application.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: Today, the scheduler does a round robin selection of nodes with equal priority by incrementing an index.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+local storage, is only recommended for specific use cases, and the pods should
+have high priority.
+    * Pod priorities could handle this scenario.  PV taints/tolerations could
+also handle this case.
+* Binding decision considering scheduling constraints from two or more pods
+sharing the same PVC.
+    * The scheduler itself only handles one pod at a time.  It’s possible the
+two pods may not run at the same time either, so there’s no guarantee that you
+will know both pod’s requirements at once.
+    * For two+ pods simultaneously sharing a PVC, this scenario may require an
```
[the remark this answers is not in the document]
msau42: Yes, I'm mainly referring to local volumes, and I agree, the simplest scenario is to combine all your containers into one Pod so that they will be scheduled together.  But for whatever reason, if a user may want two different Pods (maybe they're not running simultaneously), then this is trying to say that we can't guarantee scheduling of the second Pod if its scheduling constraints are not compatible with the first Pod.

We definitely want to support the use case of specifying pod anti-affinity/affinity along with local storage, and having the volume binding take that into account (spread my pods and local storage across nodes/racks/zones).

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+The MatchUnboundPVs predicate does not need to be re-evaluated for pod
+preemption.  Preempting a pod that uses a PV will not free up capacity on that
+node because the PV lifecycle is independent of the Pod’s lifecycle.
+
+##### Other scheduler predicates
+Currently, there are a few existing scheduler predicates that require the PVC
+to be bound.  The bound assumption needs to be changed in order to work with
+this new workflow.
+
+TODO: how to handle race condition of PVCs becoming bound in the middle of
```
bsalamat: As you mentioned, running PVC binding in parallel with the scheduler will introduce a race condition and ordering of the predicates will not address the problem. IIUC, synchronization will be needed and synchronization could significantly impact scheduler's performance.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
liggitt: > a Pod author can set pod.spec.nodeName to bypass scheduling

that is prevented in some clusters

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+##### One scheduler function
+An alternative design considered was to do the predicate, priority and bind
+functions all in one function at the end right before Pod binding, in order to
+reduce the number of passes we have to make over all the PVs.  However, this
+design does not work well with pod preemption.  Pod preemption needs to be able
+to evaluate if evicting a lower priority Pod will make a higher priority Pod
+schedulable, and it does this by re-evaluating predicates without the lower
+priority Pod.
+
+If we had put the MatchUnboundPVCs predicate at the end, then pod preemption
```
bsalamat: This can possibly be addressed by registering extra functions that preemption logic should call to filter nodes further. This might be an easier change compared to the current proposal.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: One possible way to deal with this is to mark the Pod at the very beginning of scheduling if any PVCs are unbound.  Then at the end, we can check for this to determine if we need to send the Pod through a second pass.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: Yes I think that would eliminate the need for some of the caching from the current design.  But it will still require a second pass afterwards.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
```diff
@@ -0,0 +1,580 @@
...
+provisioning is not used (in the case of local storage).
+* We have to maintain two different paths for volume binding.
+* We will be depending on the storage admin to correctly configure the
+StorageClasses.
+
+For those reasons, we prefer to switch the behavior for all unbound PVCs and
+clearly communicate these changes to the community before changing the behavior
+by default in GA.
+
+### Integrating binding with scheduling
```
bsalamat: From software engineering point of view, mixing the responsibilities of modules and removing the isolation of responsibilities is not a great practice. With this design, the scheduler process would perform volume binding which is not one of the usual responsibilities of a cluster scheduler. Other than removing the modularity, this could introduce runtime problems as well. For example, a null dereference in volume binding logic will cause a scheduler crash. High memory or CPU consumption in volume binding logic could cause scheduler to run out of memory or starve.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
bsalamat: That's right. The second pass will still be needed.

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
bsalamat: Let me clarify that I am not saying that we shouldn't move forward with this design, but when we consider pros and cons of various approaches, this should be considered as a negative point of this design.

msau42: I think we do need a tighter coupling (somehow) between scheduler and volume binding in order to address the problems I outlined in the beginning.  At a high level:
* Scheduler needs to be aware of a volume's topology wrt the node when it is scheduling a pod
* The volume binding decision needs to respect the pod's other scheduling constraints

This design proposes a very tight coupling by putting the volume binding directly into the scheduler.  You are correct in that it's expanding the scheduler's responsibilities to more than just Pod->Node binding.  The major risk I see with this current approach is that it's trying to add in the volume binding using the existing scheduling mechanics, which may be designed with the assumption that Pod->Node binding is the only thing the scheduler does.  So this approach breaks when you consider the scenario where volumes are unbound, but the user has already manually scheduled the Pod to a Node.  Fixing that I think is going to require some fundamental changes in the scheduler, which could greatly increase the scope of this design.  But it may be something still worth considering, especially if more objects also need to integrate their binding with scheduling.  I know @vishh and @yguo0905 are also looking into extending the scheduler for binding other resources (although may not be as complex as volume binding), and they may also run into the same issues as me.

I did also consider an alternate approach with loose coupling, where it tries to keep scheduler and PV controller separate.  It's similar to this [alternative discussion](https://github.com/kubernetes/community/pull/1054/files#diff-fb55f38843406f82a21ecd36f75efd72R456).  The basic idea is the PV controller waits for the Pod.Spec.NodeName to be set before it does the binding.  However, even though the binding process is decoupled from the scheduler, the binding decision is much more prone to split brain problems.  We still need a new scheduler predicate so that the scheduler can pick an appropriate node, but because the binding happens completely asynchronously in the PV controller, it's possible that the actual binding will pick a different volume, or fail completely.  Then you need a way to tell the scheduler to revert its PV cache update, and also a way for the PV controller to tell kubelet to reject the Pod so that it can be sent back through scheduling.  So while this approach keeps the binding process and code separate, I found that it:
* is more prone to split brain failures due to scheduler and PV controller trying to make the same decisions, but operating on different and out-of-sync caches.
* required very tight, complex, and delicate coordination between scheduler, PV controller, and kubelet processes to handle recovering from these split brain failures.

I am open to revisiting this alternative if you think it looks more promising than the current approach, but from my initial investigation, it seemed much more complex and fragile.

jsafrane: > With this design, the scheduler process would perform volume binding which is not one of the usual responsibilities of a cluster scheduler.

I am not sure what "cluster scheduler" is, however with this PR we're moving from Pod scheduler to Pod and PV scheduler. And I am sure more resources will come in future.

E.g. setting `pod.spec.nodeName` should not bypass the scheduler as it should schedule the storage too.

msau42: I think we could benefit from extending the scheduler to support multiple object scheduling/binding in a modular manner.  Each module would need to:
* Given a Pod object, evaluate if it needs to run through the scheduler
* Register assume/bind functions
* Register rollback function

Of course, there is a ton of detail that needs to be worked out

contributors/design-proposals/sig-storage/volume-topology-scheduling.md
msau42: Another hole is DaemonSets, which does its own scheduling.

msau42:
Updated to add downsides of proposed approach, more details about the alternative approach, and an idea on how to handle PV binding rollback.

bsalamat:
As discussed offline and in order to unblock SIG Storage, I LGTM this PR. We will hopefully have a better mechanism in scheduler that allows an easier and cleaner way of performing post scheduling tasks, such as volume or TPU binding.

/lgtm

msau42:
Ah sorry, there were still some TODO items that I still need to sort out, especially regarding how to handle scheduler bypass scenarios.  Please continue commenting/discussing.

bsalamat:
Yeah, I was not expecting it to be auto-merged upon my lgtm. Anyway, you may want to send another PR to address the remaining items.

PR #1058, opened 2017-09-14
Make bio files follow name convention

xiangpengzhao:
In community/elections/2017/README.md, wrote `If you don't see a link, try here under first+lastname.` However, some bio files don't follow this convention. This PR fixes this.

/cc @castrojo


xiangpengzhao:
TBH, I'm not sure if this name convention is necessary :)

justinsb:
I think we should remove the suggestion; people can update the links.

Setting rules on people's names does not end well :-)

justinsb:
Sent https://github.com/kubernetes/community/pull/1064 to avoid the naming policy

xiangpengzhao:
@justinsb agreed. makes a lot sense to me. Thanks!

/close

PR #1061, opened 2017-09-14
Removing prom-push-gateway flag

nielsole:
The flag does not work anymore as per https://github.com/kubernetes/kubernetes/issues/45947


grodrigues3:
@ixdy adding you since you have more test infra expertise.  results no longer pushed to prometheus or is this flag no longer relevant?

nielsole:
The results are already no longer pushed to prometheus(https://github.com/kubernetes/kubernetes/pull/34597 from 2016). The linked pending Pull Request (https://github.com/kubernetes/kubernetes/issues/45947) additionally removes the flags that were still left-over.

PR #1062, opened 2017-09-14
Election guide: add justinsb

justinsb:

PR #1059, opened 2017-09-14
add sebgoa

sebgoa:
Adds sebastien goasguen's bio to the ballot table :)

PR #1064, opened 2017-09-14
Election: Amend bio links statement

justinsb:
* Don't impose a naming policy on people

* Candidates were not asked at the same time to provide bios, so avoid
statements that could give the wrong impression.

### DIFF
--- a/community/elections/2017/README.md
+++ b/community/elections/2017/README.md
@@ -36,7 +36,7 @@ Immediately following the meeting, the raw voting results and winners will be pu
 For more information, definitions, and/or detailed election process, see full [steering committee charter](https://github.com/kubernetes/steering/blob/master/charter.md).
 
 # NOMINEES
-Don’t know someone on this list? We asked the nominees to provide a <300 word bio about their k8s experience that we'll link to their name in this table. If you don't see a link, try [here](https://github.com/kubernetes/community/tree/master/community/elections/2017) under first+lastname.  
+Don’t know someone on this list? We asked the nominees to provide a <300 word bio about their k8s experience that we'll link to their name in this table.
 
 Name | Organization/Company | GitHub 
 --- | --- | --- 


castrojo:
This looks fine but we'd like to keep the 300 word limit.

xiangpengzhao:
@castrojo I'm so sorry that I reviewed and lgtm'd https://github.com/kubernetes/community/pull/1059 and https://github.com/kubernetes/community/pull/1062 which exceed the word limit. Just wanted the bios can be merged quickly :(

lxpollitt:
@justinsb Can you comment more here about the motivation for this change?  

If we expect voters to read a lot of the bios (which would be a good thing, because we want voters to be making informed choices) then the 300 word limit makes this much more likely.

castrojo:
@xiangpengzhao I think filing an issue and assigning it to each candidate to revise their statements is probably the best way to do this.

justinsb:
Well I don't know where the 300 word limit came from, TBH.

I think in general written communications are our primary means of communication, and will be so for the steering committee.  I think some voters will be looking for conciseness, and other voters will be looking for a more expository style.  From that point of view, I see a limit as detrimental in facilitating an informed electorate.  But I'm also unaware of the benefits to the limit, so the benefits may well outweigh this.

From my point of view, the bio thing came as a surprise to me, so I included my "candidate statement" in my leadership nomination.  It was about 600 words, because that is more my style.  I think editing it would actually create more work because now someone that wants to read the statements would have to read both!

How about we rephrase the limit to be a "target length of 300 words".

parispittman:
@justinsb the 300 word limit came from an email that I sent to all nominees.

justinsb:
@lxpollitt I realize I didn't address your concerns.

So I removed: 

>Don’t know someone on this list? We asked the nominees to provide a <300 word bio about their k8s experience that we'll link to their name in this table. If you don't see a link, try [here](https://github.com/kubernetes/community/tree/master/community/elections/2017) under first+lastname.  

The core issue was the suggestion that people start guessing URLs, but I think we're fully capable of making working links, and the idea of enforcing real-world names was surprising.  So I would hope removing that would be non-contenious:

> Don’t know someone on this list? We asked the nominees to provide a <300 word bio about their k8s experience that we'll link to their name in this table.

"That we'll link to their name in this table" is now obvious both to readers and contributors, now that we have many links in place.

>Don’t know someone on this list? We asked the nominees to provide a <300 word bio about their k8s experience.

Then we shouldn't be telling candidates which values they think are most important - that is up to electorate - so scratch "about their k8s experience".

> Don’t know someone on this list? We asked the nominees to provide a <300 word bio.

So I have issues with the word limit, but most importantly it doesn't appear relevant to the electorate:

> Don’t know someone on this list? We asked the nominees to provide a bio.

Then the bio thing was a little late in the game, and some people were notified before others, so I felt it was giving false signal to suggest that candidates whose bios were not linked were somehow non-conformant.

Then frankly, I editorialized to the more positive suggestion in this PR "Candidates are invited to provide a statement which will be linked below:", because it felt happier and more fun to me, while communicating the same thing :-)

lxpollitt:
I think the intention is that by the time of the election every nominee has a 300 word bio that is linked from this guide.  I believe that @parispittman emailed the nominees on Monday with instructions.  Sounds like maybe some nominees didn't get this correctly, in which case we should fix ASAP to avoid frustrations.

(I personally think removing the 300 word limit at this point is not a good idea.  I believe it is well motivated limit to make it more likely that voters read more of the bios, which I think is a good thing.  If we changed it now then many of the candidates who have already completed or drafted their bios will likely have to re-work them in order to be equally represented in the guide.  This is my personal view point though - and others may disagree.)

justinsb:
Some bios predate the email.

As we've reached an impasse here: Did the 300 word limit come from the bootstrap committee?  They worked very hard to come up with this whole process.  If so, we obviously should use that - I'll edit this PR and my bio stat.  If not, I propose we make it a "target" or "around 300 words".

And bring on the steering committee for an end to these sorts of time sinks!

parispittman:
It came from the bootstrap committee; I'll let one of them weigh in.

sarahnovotny:
Yes.  Bootstrap committee defined the limit.  Please be kind to the messengers.

justinsb:
Thanks - I'll amend!

bgrant0607:
@justinsb I wouldn't have a problem with 301 words, but it is helpful for bios to be concise and to the point.

Your bio is the longest currently. However, more concerning with yours is whether you're clear on the role. Please take a look at the backlog:
https://github.com/kubernetes/steering/blob/master/backlog.md

In particular the following topics aren't the focus of the SC:
* Our releases should be more focused on the end-user experience
* We should re-examine our processes
* We should allow our many Kubernetes projects the freedom to experiment with alternative processes

And nothing prevents you from making headway on those outside the SC, in SIG Release, SIG Contributor Experience, SIG PM, SIG Architecture, etc. Mostly we lack people with time and interest in working on processes and seeing through their execution, measurement, and improvement.

justinsb:
Amended to only remove the try firstname lastname suggestion.

justinsb:
@bgrant0607 I'm going to respectfully disagree, and say that I believe in our community's ability to contribute with great abundance if given the right framework, and that I believe the role of the SC is to ensure that framework.

cblecker:
LGTM on the specifics of this PR.
/lgtm

Issue #841, opened 2017-07-24
SIGs list uses UTC, but the SIGs generally don't


grodrigues3:
https://github.com/kubernetes/community/blob/master/sigs.yaml
lists times in UTC but approximately every SIG actually meets on local time, so the times listed here are wrong half of the year in locales that have Daylight Savings Time

Created issue for tracking/triage, reported by @davidopp

xiangpengzhao:
I remember that the times once were listed on local time in https://github.com/kubernetes/community/blob/master/sig-list.md. They were mapped to my local time on google calender. That's convenient·

xiangpengzhao:
See also: https://github.com/kubernetes/community/issues/724#issuecomment-308751868

garethr:
This just hit me this week for SIG Apps and for the App Def working group. I'll use SIG Apps as an example but I don't think this the only one.

The SIG page on this site uses UTC, as do all of the SIG pages I had a quick look at: https://github.com/kubernetes/community/tree/master/sig-apps

However the agenda and minutes at https://docs.google.com/document/d/1LZLBGW2wRDwAfdBNHJjFfk9CFoyZPcIYGWU7R1PQ3ng/edit# state:

> Meetings are held via Zoom at 9am PST on Mondays,

@mattfarina is going to change the SIG Apps page to make it match reality, the SIG actually always meets at 9am PST, which means the time varies for non-US folks like me.

Note that the time varying is not, in my view, a problem. Due to the vagaries of timezones and daylight saving no time avoids this. The problem here is as stated in the title of this issue - and appears one of policy (the community repo template uses UTC) and reality (most of the SIGs don't).

cblecker:
Yeah, this is confusing. Unless anyone has any objections, I'd be fine with moving this to use PST as the base time zone. I'm not sure what the best way to represent DST is though.. we'd have to clarify with the sigs if they'll be meeting at the same time or if it will shift for Daylight Savings. My guess is the latter.

cc: @kubernetes/sig-contributor-experience-misc-use-only-as-a-last-resort

lavalamp:
PDT = daylight adjusted PST

cblecker:
Right, but there are currently no locations that are in PDT (UTC-7). Everything has transitioned over to PST (UTC-8). There are some locations that use [different dates for transitioning](https://en.wikipedia.org/wiki/Pacific_Time_Zone#Daylight_time), so it's possible to have both PST and PDT time zones on the same day.

[...]

PR #1067, opened 2017-09-14
Fix timezone for sig-aws zoom meeting

justinsb:


sig-aws/README.md
```diff
@@ -11,7 +11,7 @@ To understand how this file is generated, see generator/README.md.
 Covers maintaining, supporting, and using Kubernetes hosted on AWS Cloud.
 
 ## Meetings
-* [Fridays at 17:00 UTC](https://zoom.us/my/k8ssigaws) (biweekly). [Convert to your timezone](http://www.thetimezoneconverter.com/?t=17:00&tz=UTC).
+* On [Zoom](https://zoom.us/my/k8ssigaws), Fridays at 9:00 Pacific, biweekly. [Convert to your timezone](http://www.thetimezoneconverter.com/?t=9:00&tz=PT).
```
xiangpengzhao: This file is generated by running `make`. We should modify what we want in `sigs.yaml` and then run `make`. But seems like the Makefile doesn't support to use `PT` now. The docker image which `make` uses should be modified, IMO.

BTW, there is an issue on timezone concern: #841 

cc @grodrigues3 @cblecker

justinsb: Thanks - for now I just set it to 16:00 UTC along with a comment linking #841.

PR #1075, opened 2017-09-15
WIP: versioning of external provisioner protocol

jsafrane:
In 1.8 we introduced new fields in `StorageClass` that old provisioner won't understand. Therefore we should introduce some generic versioning mechanism so old provisioner that don't support new features know that a newer feature is requested and not provision a PVC.

There are two new annotations introduced, one that makes the PVCs that need newer features invisible to the old provisioner, second one introduces very simple versioning based on Kubernetes release.

We missed this in 1.8 development and now it becomes a bug. Should we fix it before 1.8 is released?

@kubernetes/sig-storage-pr-reviews  @kangarlou

### DIFF
--- a/contributors/design-proposals/storage/volume-provisioning.md
+++ b/contributors/design-proposals/storage/volume-provisioning.md
@@ -171,7 +171,8 @@ Existing behavior is unchanged for claims that do not specify
   claim creation or update, it implements these steps:
 
   1. The provisioner inspects if
-     `claim.Annotations["volume.beta.kubernetes.io/storage-provisioner"] == <provisioner name>`.
+     `claim.Annotations["volume.beta.kubernetes.io/storage-provisioner"] == <provisioner name>` or
+     `claim.Annotations["volume.kubernetes.io/storage-provisioner"] == <provisioner name>`
      All other claims MUST be ignored.
 
   2. The provisioner MUST check that the claim is unbound, i.e. its
@@ -182,9 +183,10 @@ Existing behavior is unchanged for claims that do not specify
      just created by admin is discussed below.*
 
   3. It tries to find a StorageClass instance referenced by annotation
-     `claim.Annotations["volume.beta.kubernetes.io/storage-class"]`. If not
-     found, it SHOULD report an error (by sending an event to the claim) and it
-     SHOULD retry periodically with step i.
+     `claim.Annotations["volume.beta.kubernetes.io/storage-class"]` or
+     `claim.Spec.StorageClassName`. If not found, it SHOULD report an error (by
+     sending an event to the claim) and it SHOULD retry periodically with step
+     i.
 
   4. The provisioner MUST parse arguments in the `StorageClass` and
      `claim.Spec.Selector` and provisions appropriate storage asset that matches
@@ -205,7 +207,15 @@ Existing behavior is unchanged for claims that do not specify
      }
      ```
 
-  5. When the volume is provisioned, the provisioner MUST create a new PV
+  5. The provisioner MUST parse
+     `claim.Annotations["volume.kubernetes.io/external-provisioning-version"]`
+     and check that it supports at least the version of this protocol requested
+     there. It SHOULD report an error and it MUST NOT provision a volume when
+     the annotation value is higher than the version known to the provisioner.
+
+     Table of protocol versions is available below.
+
+  6. When the volume is provisioned, the provisioner MUST create a new PV
      representing  the storage asset and save it in Kubernetes. When this fails,
      it SHOULD retry creating the PV again few times. If all attempts fail, it
      MUST delete the storage asset. All errors SHOULD be sent as events to the
@@ -450,6 +460,38 @@ With the scheme outlined above the provisioner creates PVs using parameters spec
 will be extended to contain StorageClass.Parameters.
 
 The existing provisioner implementations will be modified to accept the StorageClass configuration object.
+
+### External provisioner interface changes
+
+Sometimes PersistentVolume, PersistentVolumeClaim or StorageClass objects get
+new fields and/or different behavior is expected from an external provisioner.
+Old provisioner that don't know about these new fields must not provision PVCs
+that expect the provisioner knows about them.
+
+As an example, in 1.8 we introduced `StorageClass.MountOptions` and
+`StorageClass.ReclaimPolicy`. Each provisioner must copy these fields from
+`StorageClass` to created PV during provisioning. Old provisioners do not know
+about this requirement and Kubernetes should not allow them to provision a PVC
+whose `StorageClass.MountOptions` or `ReclaimPolicy` is set.
+
+* When no extra feature (i.e. Kubernetes 1.7 and below behavior is expected), Kubernetes adds these three annotations to a PVC:
+* `volume.beta.kubernetes.io/storage-provisioner` - to keep the old provisioners running.
+  * `volume.kubernetes.io/storage-provisioner` - to keep the new provisioners running.
+  * `volume.kubernetes.io/external-provisioning-version = "1.7"` - to explicitly state that no extra features/behavior is expected.
+
+* When behavior corresponding to Kubernetes 1.8 is expected (i.e. `StorageClass.MountOptions` or `StorageClass.ReclaimPolicy` is set), Kubernetes adds only these two annotations to a PVC:
+  * `volume.kubernetes.io/storage-provisioner` - to keep the new provisioners running.
+  * `volume.kubernetes.io/external-provisioning-version = "1.8"` - to explicitly state that the provisioner must know about `StorageClass.MountOptions` and how to deal with it.
+  * As no beta annotation is set, this PVC is invisible to old provisioners that support only 1.7.
+
+In future, when we add e.g. block device support, Kubernetes will set `external-provisioning-version = "1.9"` when a block volume is requested or `"1.8"` when filesystem volume with `MountOptions` or `ReclaimPolicy` is requested or even `"1.7"` when nothing special is required.
+
+#### Table of protocol versions
+
+| Version | Required features |
+|---------|-------------------|
+| 1.7     | None, standard provisionig that was working in 1.4 - 1.7 is expected |
+| 1.8     | Provisioner must interpret `StorageClass.MountOptions` and `StorageClass.ReclaimPolicy` when it's set |
 
 ### PV Controller Changes
 
--- a/contributors/design-proposals/storage/volume-provisioning.md
+++ b/contributors/design-proposals/storage/volume-provisioning.md
@@ -60,26 +60,26 @@ We propose that:
 3.  An api object will be incubated in storage.k8s.io/v1beta1 to hold the a `StorageClass`
     API resource. Each StorageClass object contains parameters required by the provisioner to provision volumes of that class.  These parameters are opaque to the user.
 
-4.  `PersistentVolume.Spec.Class` attribute is added to volumes. This attribute
+4.  `PersistentVolume.Spec.StorageClassName` attribute is added to volumes. This attribute
     is optional and specifies which `StorageClass` instance represents
     storage characteristics of a particular PV.
 
-    During incubation, `Class` is an annotation and not
+    During incubation, `StorageClassName` is an annotation and not
     actual attribute.
 
 5.  `PersistentVolume` instances do not require labels by the provisioner.
 
-6.  `PersistentVolumeClaim.Spec.Class` attribute is added to claims. This
+6.  `PersistentVolumeClaim.Spec.StorageClassName` attribute is added to claims. This
     attribute specifies that only a volume with equal
-    `PersistentVolume.Spec.Class` value can satisfy a claim.
+    `PersistentVolume.Spec.StorageClassName` value can satisfy a claim.
 
-    During incubation, `Class` is just an annotation and not
+    During incubation, `StorageClassName` is just an annotation and not
     actual attribute.
 
 7.  The existing provisioner plugin implementations be modified to accept
     parameters as specified via `StorageClass`.
 
-8.  The persistent volume controller modified to invoke provisioners using `StorageClass` configuration and bind claims with `PersistentVolumeClaim.Spec.Class` to volumes with equivalent `PersistentVolume.Spec.Class`
+8.  The persistent volume controller modified to invoke provisioners using `StorageClass` configuration and bind claims with `PersistentVolumeClaim.Spec.StorageClassName` to volumes with equivalent `PersistentVolume.Spec.StorageClassName`
 
 9.  The existing alpha dynamic provisioning feature be phased out in the
     next release.
@@ -89,24 +89,24 @@ We propose that:
 0. Kubernetes administator can configure name of a default StorageClass. This
    StorageClass instance is then used when user requests a dynamically
    provisioned volume, but does not specify a StorageClass. In other words,
-   `claim.Spec.Class == ""`
+   `claim.Spec.StorageClassName == ""`
    (or annotation `volume.beta.kubernetes.io/storage-class == ""`).
 
 1.  When a new claim is submitted, the controller attempts to find an existing
     volume that will fulfill the claim.
 
-    1. If the claim has non-empty `claim.Spec.Class`, only PVs with the same
-        `pv.Spec.Class` are considered.
+    1. If the claim has non-empty `claim.Spec.StorageClassName`, only PVs with the same
+        `pv.Spec.StorageClassName` are considered.
 
-    2. If the claim has empty `claim.Spec.Class`, only PVs with an unset `pv.Spec.Class` are considered.
+    2. If the claim has empty `claim.Spec.StorageClassName`, only PVs with an unset `pv.Spec.StorageClassName` are considered.
 
     All "considered" volumes are evaluated and the
     smallest matching volume is bound to the claim.
 
-2.  If no volume is found for the claim and `claim.Spec.Class` is not set or is
+2.  If no volume is found for the claim and `claim.Spec.StorageClassName` is not set or is
     empty string dynamic provisioning is disabled.
 
-3.  If `claim.Spec.Class` is set the controller tries to find instance of StorageClass with this name.  If no
+3.  If `claim.Spec.StorageClassName` is set the controller tries to find instance of StorageClass with this name.  If no
     such StorageClass is found, the controller goes back to step 1. and
     periodically retries finding a matching volume or storage class again until
     a match is found. The claim is `Pending` during this period.
@@ -130,7 +130,7 @@ We propose that:
       periodically.
 
   8.  If `Provision` returns no error, the controller creates the returned
-      `api.PersistentVolume`, fills its `Class` attribute with `claim.Spec.Class`
+      `api.PersistentVolume`, fills its `StorageClassName` attribute with `claim.Spec.StorageClassName`
       and makes it already bound to the claim
 
     1.  If the create operation for the `api.PersistentVolume` fails, it is
@@ -141,7 +141,7 @@ We propose that:
         on the claim
 
 Existing behavior is unchanged for claims that do not specify
-`claim.Spec.Class`.
+`claim.Spec.StorageClassName`.
 
 * **Out of tree provisioning**
 
@@ -415,22 +415,22 @@ type StorageClass struct {
 
 ```
 
-`PersistentVolumeClaimSpec` and `PersistentVolumeSpec` both get Class attribute
+`PersistentVolumeClaimSpec` and `PersistentVolumeSpec` both get StorageClassName attribute
 (the existing annotation is used during incubation):
 
 ```go
 type PersistentVolumeClaimSpec struct {
     // Name of requested storage class. If non-empty, only PVs with this
-    // pv.Spec.Class will be considered for binding and if no such PV is
+    // pv.Spec.StorageClassName will be considered for binding and if no such PV is
     // available, StorageClass with this name will be used to dynamically
     // provision the volume.
-    Class string
+    StorageClassName string
 ...
 }
 
 type PersistentVolumeSpec struct {
     // Name of StorageClass instance that this volume belongs to.
-    Class string
+    StorageClassName string
 ...
 }
 ```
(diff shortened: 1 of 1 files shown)


liggitt:
@kubernetes/api-reviewers for versions-within-versions question

contributors/design-proposals/storage/volume-provisioning.md
```diff
@@ -451,6 +461,38 @@ will be extended to contain StorageClass.Parameters.
...
 The existing provisioner implementations will be modified to accept the StorageClass configuration object.
 
+### External provisioner interface changes
+
+Sometimes PersistentVolume, PersistentVolumeClaim or StorageClass objects get
+new fields and/or different behavior is expected from an external provisioner.
+Old provisioner that don't know about these new fields must not provision PVCs
+that expect the provisioner knows about them.
+
+As an example, in 1.8 we introduced `StorageClass.MountOptions` and
```
msau42: These two new 1.8 options are optional without defautling right?

wongma7:
I agree with removing volume.beta.kubernetes.io/storage-provisioner for 1.8 as a bugfix to avoid old(current) provisioners from doing wrong things in 1.8 clusters. If we can't get agreement on the version-within-version scheme in time, at least we could do that. Doing so will break all old(current) provisioners upon upgrade to 1.8. But the alternative scenario is maybe more scary? admin creates storageclass.reclaimpolicy=retain->external provisioner ignores it->user deletes pvc thinking it's safe->data unexpectedly gone.

anyway, I do think `volume.kubernetes.io/external-provisioning-version` is needed. The kube external provisioner interface is incidental/accidental (i.e. it came essentially free from the way the pv controller works and it's not defined in code so much as in the proposal doc), so we are currently lacking a way to define versioning of it somewhere. An alternative would be to increment storageclass api version instead of adding new fields to older versions, but I'm pretty sure that would be excessive.

contributors/design-proposals/storage/volume-provisioning.md
wongma7: storageclass.reclaimpolicy will default to Delete (exact same behaviour as before when Delete was the only option)

msau42: If the new fields are defaulted, then this scheme doesn't work because we will always have to set version to 1.8 since the new field is always set?

wongma7:
We could set the version according to whether it breaks the “API” i.e. only
if reclaimpolicy is retain

smarterclayton:
I'm confused - did we break the storage class API in an incompatible way?  That sounds like something we should revert.

wongma7:
No, it's not that bad. I'm misusing language, sorry for that. Nothing has broken.

There are two interfaces: the storageClass API, and the beta external provisioning interface. The external provisioning interface is defined here in this document and Jan is proposing we need to version it.

We've added two fields to storageClass for 1.8. Existing external provisioners will continue to function but they won't be able to take advantage of the new fields because they're ignorant of them. That's technically acceptable API-wise, we've not "broken" any API, but it's undesirable and so we think it's a bug. Jan is proposing the fix.

liggitt:
> That's technically acceptable API-wise, we've not "broken" any API, but it's undesirable and so we think it's a bug. Jan is proposing the fix.

Those two statements seem contradictory. 

How is this different from an external scheduler needing to be updated to take advantage and be aware of additional pod features that could impact scheduling? If you want external provisioners to fail when presented with unknown fields (instead of the normal approach of ignoring unknown fields) in a storage class, it seems like it would be simpler to document that behavior as required.

wongma7:
I think you're right. (My bias may be getting in the way of my englishing)

Yes, that would be simpler. I'm guessing unknown fields can be found easily
by using the dynamic client?

contributors/design-proposals/storage/volume-provisioning.md
```diff
@@ -451,6 +461,38 @@ will be extended to contain StorageClass.Parameters.
...
+`StorageClass.ReclaimPolicy`. Each provisioner must copy these fields from
+`StorageClass` to created PV during provisioning. Old provisioners do not know
+about this requirement and Kubernetes should not allow them to provision a PVC
```
kangarlou: > Kubernetes should not allow them to provision a PVC

I assume you meant provisioning a PV here, but how does Kubernetes prevent a pre-1.8 external provisioner from creating the PV? How does Kubernetes know that an external provisioner exhibit the correct behavior for a given version of Kubernetes?

I understand why you're introducing the `volume.kubernetes.io/external-provisioning-version` annotation to the PVC, but an external provisioner can easily retrieve such information by contacting the API server. For example, [Trident](https://github.com/NetApp/trident) determines the Kubernetes version and decides whether to support the default storage class feature or not. As I said, the part that I don't understand is how Kubernetes enforces external provisioners to exhibit the correct behavior based on the version of Kubernetes. 

It would be helpful to document the changes related to provisioning/storage classes introduced in different versions of k8s on this page, so that we don't have to dig the CHANGELOG to know the expected behavior for each release. Thanks for copying me on the PR!

wongma7:
o right, @liggitt the difference is that new storageclass fields don't always mean new features that must be taken into account. Like if storageClass.reclaimPolicy has defaulted to Delete, we want to allow pre-1.8 external provisioners to provision a PVC by setting `volume.kubernetes.io/external-provisioning-version:1.7`. Only when storageClass.reclaimPolicy is Retain do we want to disallow provisioning by setting version to `1.8`. Similarly we want old provisioners to function if: mountOptions is blank, resizable false, volumeType non-block(?). Explicit versioning gives us that.

Basically the choice is between: complicating kubernetes' end by introducing storageclass versions-within-versions OR complicating the external provisioner end by mandating they update every time we introduce a new storageclass field.

(So wehn I say bias I mean I would rather complicate the kuberentes end than make it harder for external provisioners.)

liggitt:
every external provisioner has some set of features they support. it seems like it is the responsibility of the storage admin to not configure storage classes that make use of features the provisioner doesn't support. this is true both for storage class parameters and for first-class fields.

contributors/design-proposals/storage/volume-provisioning.md
jsafrane: Kubernetes can always check that 1.8 feature is really needed or 1.7-ish provisioner can do the job:

```go
if sc.MountOptions == "" && sc.ReclaimPolicy == "Delete" {
   /* use 1.7 */
} else {
  /* require 1.8 */
}
```

jsafrane:
@liggitt, can we trust admins they are so careful?

In addition, in 1.9 we plan to introduce raw block storage that introduces new fields in PVC. Old provisioners will ignore this field and may provision something that *user* did not want. Admin would have no control about it.

liggitt:
> in 1.9 we plan to introduce raw block storage that introduces new fields in PVC. Old provisioners will ignore this field and may provision something that user did not want. Admin would have no control about it.

that seems like a flaw in the API design for that feature... admins *must* have control over provisioners via storage class... whatever the user specifies must be within the parameters of the storage class.

jsafrane:
> that seems like a flaw in the API design for that feature... admins must have control over provisioners via storage class... whatever the user specifies must be within the parameters of the storage class.

Can you please elaborate on this? Why it is necessary? E.g. for raw block devices, I don't think that admin needs to enable it explicitly per storage class. IMO no damage can be done by accessing a block device instead of accessing a filesystem on top of it.

I think it's a wrong way to protect old providers from PVCs that need newer features. The provisioner should be able to detect it is able to provision a PVC out of the box, without any admin configuration.

liggitt:
PVC-creating users should not be able to specify options that control how the provisioner creates the volume unless the admin that created the storageclass permitted it

kangarlou:
@liggitt Admins still have control over provisioners as they can define storage classes such that all parameters are understood by the provisioner and are acted upon as expected by a particular version of k8s. However, I think that puts too much burden on the admins as they have to know what's supported by the different versions of k8s and provisioners. The main complication arises when there is no one-to-one relationship between the k8s and provisioner versions. With k8s volume plugins moving out of tree in the future, this problem would also present itself there. We shouldn't expect a new version of the external provisioner or out-of-tree volume plugin for every release of k8s.

I do agree with you that there seems to be a flaw in the API design. Jan and I had this exchange earlier: https://groups.google.com/forum/#!topic/kubernetes-sig-storage/Yc--M83BWGo
My opinion is that having first class fields in the storage class (I referred to them as keywords in the list of opaque parameters in the storage class) creates a tight coupling between k8s and provisioners/volume plugins. I think it's up to provisioners to implement capabilities like ReclaimPolicy and MountOptions however they like. For example, [Trident](https://github.com/NetApp/trident/) allows configuring these options before any of these changes were introduced in 1.8. Creating first class fields in the storage class may help with standardizing the behavior for different provisioners/plugins, but it also creates tight coupling between k8s and provisioners/volume plugins that is undesirable. 

I think the admin's job is easier if they are only concerned with the provisioner/plugin parameters. Effectively, we're replacing a 2D matrix that tracks changes across both k8s and volume plugins to a 1D matrix that includes parameters accepted by the volume plugin. For this scenario, provisioner/plugin fields still need to be versioned but again that's up to the provisioner/plugin developers to do versioning in however fashion they deem appropriate. Obviously, changes to PVC, PV, and non-opaque storage class fields would require changes to the external provisioners/plugins, but I don't think we should introduce first class parameters to the storage class and mandate changes to the provisioners and volume plugins as a result.

liggitt:
> We shouldn't expect a new version of the external provisioner or out-of-tree volume plugin for every release of k8s.

I agree, but provisioners should document what version of kubernetes (or what parts of storage classes) they support

kangarlou:
> I agree, but provisioners should document what version of kubernetes (or what parts of storage classes) they support

@liggitt I agree with you that provisioners should document which versions of k8s they support, but the version of k8s should have no bearing on what can be included in the storage class; the version of the provisioner/plugin mandates what can be specified in the storage class. I think storage class parameters should be opaque parameters that are only understood by the provisioner or volume plugin. In other words, k8s shouldn't expect any special behavior by the provisioner/plugin when a specific parameter is defined in the storage class.

liggitt:
> the version of k8s should have no bearing on what can be included in the storage class;

That's not realistic. What a provisioner can support is always an intersection of the provisioner's capabilities and the kubernetes cluster's capabilities. For example, only 1.8 supports separate namespaces for secrets referenced by azure/ceph PVs. The azure/ceph provisioners can support a "secretNamespace" parameter, but that will only work on kube 1.8.

> the version of the provisioner/plugin mandates what can be specified in the storage class.

That is already the case, whether the parameters are opaque in the parameters map or specified as storage class fields.

jsafrane:
As a though experiment, let there be a new field to StorageClass:

```go
type StorageClass struct {
    ...
    ProvisioningFeatures []ProvisioningFeature
}

type ProvisioningFeature string
const (
    ProvisioningFeatureMountOptions ProvisioningFeature = "MountOptions"
    ProvisioningFeatureRelaimPolicy ProvisioningFeature = "ReclaimPolicy"
    ProvisioningFeatureBlockStorage ProvisioningFeature = "BlockStorage"
)
```

Expected behavior would be:
* Admin must set this field to include all features that the external provisioner supports.
* Before triggering external dynamic provisioning, Kubernetes checks that features that PVC/StorageClass needs is compatible with StorageClass.ProvisioningFeatures and refuses to provision anything if it does not match.
* External provisioners don't do anything. Any misconfiguration of StorageClass is admin's fault.

While this could work, I think it's too tiresome for admins - they need to re-evaluate their StorageClasses after every update of Kubernetes or external provisioner. IMO they will be lazy to configure it and most of them will run with empty ProvisioningFeatures.

wongma7:
> every external provisioner has some set of features they support. it seems like it is the responsibility of the storage admin to not configure storage classes that make use of features the provisioner doesn't support. this is true both for storage class parameters and for first-class fields.

@liggitt right, but for first-class fields Kubernetes will default it so even if the admin hasn't configured it to use the new feature they could be surprised when their provisioner rejects a pvc because the provisioner doesn't recognize the field 

@kangarlou tbh I think it is too late to talk about whether mountOptions and reclaimPolicy should be parameters or first-class fields but... This was definitely brought up in the proposals and even up until the PRs we were still deciding. Basically IMO my priority is: end-user UX>storage admin>provisioner author. Having these things first class field makes it abundantly clear that they have first-class support, they are common among plugins, they can be validated (a no-brainer when an it's an enum like reclaimPolicy), lets admins specify things in a more pleasant way than map[string]string (e.g. mountOptions as a list of strings instead of comma-separated).

kangarlou:
> What a provisioner can support is always an intersection of the provisioner's capabilities and the kubernetes cluster's capabilities. For example, only 1.8 supports separate namespaces for secrets referenced by azure/ceph PVs.

@liggitt I agree that the version of k8s determines certain capabilities that can be supported by provisioners. This is precisely why provisioners should document what versions of Kubernetes they support or require, but if a provisioner decides not to take advantage of the separate namespace feature, upgrade to 1.8 shouldn't mandate upgrade to the provisioner. The part of the proposal that I disagree with is the following:

> As an example, in 1.8 we introduced StorageClass.MountOptions and StorageClass.ReclaimPolicy. Each provisioner must copy these fields from StorageClass to created PV during provisioning. Old provisioners do not know about this requirement and Kubernetes should not allow them to provision a PVC whose StorageClass.MountOptions or ReclaimPolicy is set.

It seems like certain parameters in the storage class impose a specific behavior on provisioners. As I mentioned in my first comment on this thread, it's unclear how k8s enforces provisioners to do the right thing for a given version of k8s. My thinking is that **all** parameters in storage class are determined by the provisioner. In other words, the version of k8s, shouldn't require support for certain parameters or certain type of behavior on the part of provisioners.

@wongma7 I understand that it might be too late to have these discussions, but I thought I share my views here as we'll have to deal with these issues with CSI and external volume plugins. Thanks!

jsafrane:
> It seems like certain parameters in the storage class impose a specific behavior on provisioners

**All** parameters in the storage class impose a specific behavior. Until now we had `provisioner` and `parameters`, now we have `mountOptions` and `reclaimPolicy`. As @wongma7 pointed out, they are common for all volume plugins (external or internal) and thus have a specific field.

It is the same as in PVCs - provisioners must understand all fields in PVCs. There will be no magic dictionary there and when we add `volumeMode` field, all provisioners must understand it.

This PR tries to establish a way how old provisioners do not need to understand all the fields if the expected behavior is the same as in Kubernetes version they conform to.

kangarlou:
@jsafrane All parameters in the storage class do impose some specific behavior. However, I'm of the opinion that such behavior should be defined by the provisioner/plugin, not by k8s. In other words, I'm not in favor of the first class parameters in the storage class, which are imposed by k8s. The reason I'm against the first class parameters is that we're effectively extending the k8s API to the opaque parameters in the storage class. These opaque parameters are just like annotations. It's hard to do any validation on them for the different versions of k8s. If the first class parameters are applicable across all volume plugins, why shouldn't we make them proper fields in the storage class? With this design, we have a clear boundary between what's defined/expected by a particular version of k8s (i.e., everything not in the parameters section) and what's defined and expected by the provisioner and volume plugin (everything in the parameters section). The k8s portion and the provisioner/plugin portion of the storage class each can be versioned independently.

This thread captures a lot of good discussions. I share @wongma7's view that our priority should be the following: "end-user UX>storage admin>provisioner author"

If you think it's useful and if I can find some time this week, I'm going to go over the proposal and the discussion here in order to draw a few conclusions. Thanks!

childsb:
I'm starting to feel that almost no new fields are added to Storage Class, ever... Both mount options and reclaim policy imposes behavior on the provisioner and leads to confusing UX for user and admin..  As example, a storageclass may support multiple retention policies.. or the provisioner may want to add its own mount options to the PV.. In both of those cases, any value for in the SC is ambiguous and confuses user and admin.  Even if a field is 'common to all plugins', that doesn't mean the use of the PVs common fields via provisioner is common.

I propose these guidelines:

1) Allow new fields on Storage Class only if it changes how a PV is used by the POD.. (ie. block)
2) Any other meta characteristic for PVs are via Properties.. That leaves behavior and impl up to the provisioner.

kangarlou:
@childsb I agree with you. In the case of Trident, we have limited storage class parameters to the properties of storage pools (e.g., attributes like media type, snapshot support, IOPS range). Therefore, storage class X can lead to the creation of PVs with different reclaim policies. We're using PVC annotations as a mechanism for users to set the reclaim policy on the PVs. I think it's a good idea to have a proper field in the PVC to set the reclaim policy for the PV. Regarding mount options, we're leaning towards setting mount options in the backend configuration files that we already use to make the provisioner aware of various storage backends. This is how we do things currently, it may change in the future, and I'm certainly not suggesting others should use the storage classes the same way we do. In fact, all my comments on this thread are about keeping the storage class parameters free-form, so that provisioners/plugins can use these parameters however they like. I agree standardizing parameter names help with user experience, but nothing should break if a provisioner/plugin isn't conforming to the standardized (or first class) parameters. Otherwise, we're effectively extending the Kubernetes API to these opaque parameters and admins have to keep track of what version of k8s and what version of provisioner support certain fields or not.

Can you provide an example of a field that changes how a PV is used by a pod and is a good candidate for being included as a new field in the storage class?

childsb:
@kangarlou 
> Can you provide an example of a field that changes how a PV is used by a pod and is a good candidate for being included as a new field in the storage class?

Block vs File PVs is the only example I have right now.

wongma7:
> Both mount options and reclaim policy imposes behavior on the provisioner and leads to confusing UX for user and admin.. As example, a storageclass may support multiple retention policies.. or the provisioner may want to add its own mount options to the PV.. In both of those cases, any value for in the SC is ambiguous and confuses user and admin. Even if a field is 'common to all plugins', that doesn't mean the use of the PVs common fields via provisioner is common.

I don't understand, this sounds like an argument *for* making mountoptions and reclaimpolicy SC fields. If we want to let provisioners do anything they want, how is it better UX for the admin if every provisioner can be different: e.g. if A does comma-separated mountoptions and B space-separated. Similarly, how is it better UX if we allow people to set *anything* for reclaimpolicy and to interpret it any way they want, when reclaimpolicy has a strict meaning encoded in API, and validation of it can only improve UX. For a real example of the downside of letting authors have control over reclaimpolicy, see a contributor's well-intentioned idea for "`Delete` but not really" in which the PV is deleted but not the storage asset: https://github.com/kubernetes-incubator/external-storage/pull/266#discussion_r131438984

Anyway I feel this discussion has strayed too far from the actual proposal. I want badly to avoid a flexvolume-like situation where authors have to struggle to decide what updates, if any, they need to make to their plugin every release.

msau42:
Another need for versioning in the future is for topology-aware dynamic provisioning.

kangarlou:
It seems like we have three classes of storage class parameters (assuming we're going forward with first class parameters):

1. Provisioner- or plugin-specific parameters: These parameters solely depend on the storage backend the provisioner/plugin supports. These parameters can change with new versions of the provisioners/plugins (or the storage backends that they support) and are completely decoupled from Kubernetes.
2. Loosely-coupled Kubernetes parameters:  These parameters are introduced as first class parameters in some version of Kubernetes; however, they don't depend on a specific functionality from a certain version of Kubernetes. I put `MountOptions` and `ReclaimPolicy` in this category as introducing these parameters didn't require any changes to the Kubernetes core (e.g., the persistent volume controller didn't have to change). If a provisioner/plugin supports these parameters, it can potentially support these parameters for earlier versions of k8s where these parameters were not considered as first class parameters.
3. Tightly-coupled Kubernetes parameters: These parameters depend on certain functionality introduced in some version of Kubernetes. Separate namespaces for secrets, local or block storage parameters, etc. are examples of fields that require the provisioner to support a minimum or a specific version of Kubernetes.

It seems that storage class versioning at the level of provisioner/plugin handles the first two class of parameters, and the third class requires a matrix for that captures both provisioner and k8s versions. Given that we can't assume a new version of provisioner/plugin for every release of Kubernetes, I don't see a way where we can avoid storage class versioning (say via a field called `Version` in the storage class parameters). I think it's up to the provisioner/plugin providers to document the 2D matrix that captures the supported parameters across both provisioner and k8s versions.

> For a real example of the downside of letting authors have control over reclaimpolicy, see a contributor's well-intentioned idea for "Delete but not really" in which the PV is deleted but not the storage asset: kubernetes-incubator/external-storage#266 (comment)

@wongma7 I agree that if a `ReclaimPolicy` is set to `Delete` or `Retain` that has certain meaning in the context of Kubernetes. If a provisioner/plugin writer deviates from that intended behavior (like the example you linked), the author is creating a usability problem but, with or without first class parameters, there is no way we can enforce the correct behavior. I don't think we should be too concerned here as everything an external provisioner/plugin does can potentially deviate from the expected behavior for a provisioner. If a provisioner author misinterprets the intended behavior or creates inconsistencies with other provisioners/plugins, I think that should be the provider's concern, not ours as Storage SIG.

While we're on the topic of `ReclaimPolicy`, I'm not sure what to make of the Retain reclaim policy. Deleting a PVC with ReclaimPolicy set to Retain, leads to the corresponding PV being in the Released state; however, there is no way for a user to reuse the PV. In absence of manual intervention by an admin, where a new PV is created with the same properties as the released PV, and somehow a new PVC gets bound to the newly created available PV, I don't see a way for users to access a volume once a PVC with reclaim policy set to `Delete` is deleted.

liggitt:
I still don't really see a difference between this and any other external kubernetes API integration:
* ingress adds a new field, existing external ingress controllers may or may not honor it
* pods add a new field, existing external schedulers may not be aware of it and honor it
* endpoints add a new field, existing external DNS providers may not be aware of it and honor it

kangarlou:
@liggitt Yes, they're mostly similar but one important difference is that the proposal is not introducing new fields for the first class parameters. Rather, it's using the existing map[string]string parameters in the storage class, effectively extending Kubernetes API to these opaque parameters. These opaque parameters are just like annotations, so it's hard to do any validation on them for different versions of k8s, if that's something we care about to reduce the burden on admins.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1075)

jsafrane:
I asked @kubernetes/sig-api-machinery-feature-requests at their [mailing list](https://groups.google.com/forum/#!msg/kubernetes-sig-api-machinery/HTjJnrhCmps/2KvmpfOvAgAJ), and I got this recommendation:

> There have been a fair amount of discussions on this - I think the latest was having a way on decode to identify fields that we don't recognize and then give the controller the option of communicating that back to the user.

(controller = external provisioner in this context).

But I got no reply if it's is implemented anywhere or even if it could be implemented. Generated client throws away unknown fields now and controller does now know about them.

So the recommendation for now is that external provisioner checks version of API server it talks to and refuses to work with a version it was not tested with.

wongma7:
@jsafrane until that is implemented, we can also tell people to use dynamic client and provide some code snippet, WDYT? just tried it, it preserves all fields in a map[string] that can then be compared with the known list of fields like &{map[kind:StorageClass apiVersion:storage.k8s.io/v1 metadata:map[creationTimestamp:2017-11-01T00:06:59Z labels:map[addonmanager.kubernetes.io/mode:Reconcile] annotations:map[storageclass.beta.kubernetes.io/is-default-class:true] name:standard selfLink:/apis/storage.k8s.io/v1/storageclasses/standard uid:94b6af0a-be98-11e7-9edf-00c2c6e2e2cb resourceVersion:133] provisioner:kubernetes.io/host-path reclaimPolicy:Delete]}.

jsafrane:
Got more info from API machinery, they're working on https://docs.google.com/document/d/1wuoSqHkeT51mQQ7dIFhUKrdi3-1wbKrNWeIL4cKb9zU, no deadline yet.

@wongma7, dynamic client is the best good way to go until we have something better in API.

The recommendation still stays the same, admin must ensure that all external provisioners are compatible with the version of Kubernetes where they run.

**There will be no explicit versioning of external provisioning protocol.**

PR #1078, opened 2017-09-16
Adjust the value of the current and long-term goal.

warmchang:
Now, the number of max pods per machine is 110: https://github.com/kubernetes/kubernetes.github.io/blob/master/docs/admin/kubelet.md

The goal of max pods per machine is 500: https://github.com/kubernetes/community/blob/master/sig-scalability/goals.md

### DIFF
--- a/sig-scalability/thresholds.md
+++ b/sig-scalability/thresholds.md
@@ -42,7 +42,7 @@ Important notes about the numbers:
 | Total number of all objects         | 250000         |             | 1000000        |
 | Number of nodes                     | 5000           |             | 5000           |
 | Number of pods                      | 150000         |             | 500000         |
-| Number of pods per node<sup>1</sup> | 100            |             | 100            |
+| Number of pods per node<sup>1</sup> | 110            |             | 500            |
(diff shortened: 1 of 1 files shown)


sig-scalability/thresholds.md
```diff
@@ -42,7 +42,7 @@ Important notes about the numbers:
 | Total number of all objects         | 250000         |             | 1000000        |
 | Number of nodes                     | 5000           |             | 5000           |
 | Number of pods                      | 150000         |             | 500000         |
-| Number of pods per node<sup>1</sup> | 100            |             | 100            |
+| Number of pods per node<sup>1</sup> | 110            |             | 500            |
```
wojtek-t: I agree with 110.

I have heard any commitment that we will ever want to get to 500 and make it official.

warmchang: Yes, RedHat is dealing with the 500 goal, and now the baseline is 110 pods/node.

warmchang:
@wojtek-t @jbeda @zhangxiaoyu-zidif  If it's ok for commit, pls tag a "lgtm" label, thanks. 😄

PR #1096, opened 2017-09-19
Fix sigs.yaml from #1093

cblecker:


cblecker:
Self LGTM as this just fixes sigs.yaml with the changes from #1093

PR #1097, opened 2017-09-19
add bio for steering committee elections(michelle)

michelleN:


eparis:
You might want to link from: https://github.com/kubernetes/community/blob/master/community/elections/2017/README.md

Issue #588, opened 2017-04-28
Create kubernetes cluster without public hosted domain with working k8 dashboard on AWS


vivekaundkar:
I am trying to create a k8 cluster without public hosted domain. I just created a private route 53 zone entry and using this entry and an S3 bucket as inputs, with help of kops able to bring up a kubernetes cluster.
I am able to see load balancer entries, route tables, 3 master and 2 nodes, a bastion server created. It seems everything is in place. Since I used private topology option while creating cluster all instances got private i/p's only bastion instance has public ip. I am able to SSH onto bastion server and to K8 master instance from there. Somehow I am able to make kubectl get nodes command running on k8 master.
Question is where and how to make k8 dashboard up an running?
I am assuming inside my VPC I can start a new instance and run k8 dashboard from there or use bastion server to run k8 dashboard. But how exactly to do that in case I am on right track.

Please suggest if there is an alternate way to create a k8 cluster without public hosted domain with working dashboard and yaml deployment possible from any instance inside same VPC.

I am going to be using VPN as next step so if solution is considering VPN that will be great.

jessfraz:
Hi @vivekaundkar I believe you opened this issue in the wrong repo, would you mind moving it to kubernetes/kubernetes or perhaps joining #sig-aws on slack :)

jessfraz:
also please see @kris-nova's response on https://github.com/kubernetes/community/issues/573#issuecomment-297066002

PR #1100, opened 2017-09-19
Add coredns proposal

johnbelamaric:
Add coredns proposal for https://github.com/kubernetes/features/issues/427

### DIFF
--- a/contributors/design-proposals/network/coredns.md
+++ b/contributors/design-proposals/network/coredns.md
@@ -38,9 +38,70 @@ intent to make it the default in the future.
  * Limit PTR replies to the cluster CIDR [#125](https://github.com/kubernetes/dns/issues/125)
  * Serve DNS for selected namespaces [#132](https://github.com/kubernetes/dns/issues/132)
  * Serve DNS based on a label selector
+ * Support for wildcard queries (e.g., `*.namespace.svc.cluster.local` returns all services in `namespace`)
 
 By default, the user experience would be unchanged. For more advanced uses, existing users would need to modify the
 ConfigMap that contains the CoreDNS configuration file.
+
+### Configuring CoreDNS
+
+The CoreDNS configuration file is called a `Corefile` and syntactically is the same as a [Caddyfile]
+(https://caddyserver.com/docs/caddyfile). The file consists of multiple stanzas called _server blocks_.
+Each of these represents a set of zones for which that server block should respond, along with the list
+of plugins to apply to a given request. More details on this can be found in the 
+[Corefile Explained](https://coredns.io/2017/07/23/corefile-explained/) and
+[How Queries Are Processed](https://coredns.io/2017/06/08/how-queries-are-processed-in-coredns/) blog
+entries.
+
+### Configuration for Standard Kubernetes DNS
+
+The intent is to make configuration as simple as possible. The following Corefile will behave according
+to the spec, except that it will not respond to Pod queries. It assumes the cluster domain is `cluster.local`
+and the cluster CIDRs are all within 10.0.0.0/8.
+
+```
+. {
+  errors
+  log
+  cache 30
+  health
+  prometheus
+  kubernetes 10.0.0.0/8 cluster.local
+  proxy . /etc/resolv.conf
+}
+
+```
+
+The `.` means that queries for the root zone (`.`) and below should be handled by this server block. Each
+of the lines within `{ }` represent individual plugins:
+
+  * `errors` enables [error logging](https://coredns.io/plugins/errors)
+  * `log` enables [query logging](https://coredns.io/plugins/log/)
+  * `cache 30` enables [caching](https://coredns.io/plugins/cache/) of positive and negative responses for 30 seconds
+  * `health` opens an HTTP port to allow [health checks](https://coredns.io/plugins/health) from Kubernetes
+  * `prometheus` enables Prometheus [metrics](https://coredns.io/plugins/metrics)
+  * `kubernetes 10.0.0.0/8 cluster.local` connects to the Kubernetes API and serves records for the `cluster.local` domain and reverse DNS for 10.0.0.0/8.
+  * `proxy . /etc/resolv.conf` forwards any queries not handled by other plugins (the `.` means the root domain) to the nameservers configured in `/etc/resolv.conf`
+
+### Configuring Stub Domains
+
+To configure stub domains, you add additional server blocks for those domains:
+
+```
+example.com {
+  proxy example.com 8.8.8.8:53
+}
+
+. {
+  errors
+  log
+  cache 30
+  health
+  prometheus
+  kubernetes 10.0.0.0/8 cluster.local
+  proxy . /etc/resolv.conf
+}
+```
 
 ## Implementation
 
--- a/contributors/design-proposals/network/coredns.md
+++ b/contributors/design-proposals/network/coredns.md
@@ -45,8 +45,8 @@ ConfigMap that contains the CoreDNS configuration file.
 
 ### Configuring CoreDNS
 
-The CoreDNS configuration file is called a `Corefile` and syntactically is the same as a [Caddyfile]
-(https://caddyserver.com/docs/caddyfile). The file consists of multiple stanzas called _server blocks_.
+The CoreDNS configuration file is called a `Corefile` and syntactically is the same as a
+[Caddyfile](https://caddyserver.com/docs/caddyfile). The file consists of multiple stanzas called _server blocks_.
 Each of these represents a set of zones for which that server block should respond, along with the list
 of plugins to apply to a given request. More details on this can be found in the 
 [Corefile Explained](https://coredns.io/2017/07/23/corefile-explained/) and
@@ -80,8 +80,8 @@ of the lines within `{ }` represent individual plugins:
   * `cache 30` enables [caching](https://coredns.io/plugins/cache/) of positive and negative responses for 30 seconds
   * `health` opens an HTTP port to allow [health checks](https://coredns.io/plugins/health) from Kubernetes
   * `prometheus` enables Prometheus [metrics](https://coredns.io/plugins/metrics)
-  * `kubernetes 10.0.0.0/8 cluster.local` connects to the Kubernetes API and serves records for the `cluster.local` domain and reverse DNS for 10.0.0.0/8.
-  * `proxy . /etc/resolv.conf` forwards any queries not handled by other plugins (the `.` means the root domain) to the nameservers configured in `/etc/resolv.conf`
+  * `kubernetes 10.0.0.0/8 cluster.local` connects to the Kubernetes API and [serves records](https://coredns.io/plugins/kubernetes/) for the `cluster.local` domain and reverse DNS for 10.0.0.0/8 per the [spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md)
+  * `proxy . /etc/resolv.conf` [forwards](https://coredns.io/plugins/proxy) any queries not handled by other plugins (the `.` means the root domain) to the nameservers configured in `/etc/resolv.conf`
 
 ### Configuring Stub Domains
 
@@ -102,6 +102,57 @@ example.com {
   proxy . /etc/resolv.conf
 }
 ```
+
+### Configuring Federation
+
+Federation is implemented as a separate plugin. You simply list the federation names and
+their corresponding domains.
+
+```
+. {
+  errors
+  log
+  cache 30
+  health
+  prometheus
+  kubernetes 10.0.0.0/8 cluster.local
+  federation cluster.local {
+    east east.example.com
+    west west.example.com
+  }
+  proxy . /etc/resolv.conf
+}
+```
+
+### Deployment and Operations
+
+Typically when deployed for cluster DNS, CoreDNS is managed by a Deployment. The
+CoreDNS pod only contains a single container, as opposed to kube-dns which requires three
+containers. This simplifies troubleshooting.
+
+The Kubernetes integration is stateless and so multiple pods may be run. Each pod will have its
+own connection to the API server. If you (like OpenShift) run a DNS pod for each node, you should not enable
+`pods verified` as that could put a high load on the API server. Instead, if you wish to support
+that functionality, you can run another central deployment and configure the per-node
+instances to proxy `pod.cluster.local` to the central deployment.
+
+All logging is to standard out, and may be disabled if
+desired. In very high queries-per-second environments, it is advisable to disable query logging to
+avoid I/O for every query.
+
+CoreDNS can be configured to provide an HTTP health check endpoint, so that it can be monitored
+by a standard Kubernetes HTTP health check. Readiness checks are not currently supported but
+are in the works (see [#588](https://github.com/coredns/coredns/issues/588)). For Kubernetes, a
+CoreDNS instance will be considered ready when it has finished syncing with the API.
+
+CoreDNS performance metrics can be published for Prometheus.
+
+When a change is made to the Corefile, you can send each CoreDNS instance a SIGUSR1, which will
+trigger a graceful reload of the Corefile.
+
+### Performance and Resource Load
+
+Evaluation of performance and resource load (memory, CPU) is still underway. Initial results show memory consumption is generally less than 40MB without `pods verified`.
 
 ## Implementation
 
--- a/contributors/design-proposals/network/coredns.md
+++ b/contributors/design-proposals/network/coredns.md
@@ -1,0 +1,72 @@
+# Add CoreDNS for DNS-based Service Discovery
+
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: @johnbelamaric
+
+## Motivation
+
+CoreDNS is another CNCF project and is the successor to SkyDNS, which kube-dns is based on. It is a flexible, extensible
+authoritative DNS server and directly integrates with the Kubernetes API. It can serve as cluster DNS,
+complying with the [dns spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md). 
+
+CoreDNS has fewer moving parts than kube-dns, since it is a single executable and single process. It is written in Go so
+it is memory-safe (kube-dns includes dnsmasq which is not). It supports a number of use cases that kube-dns does not
+(see below). As a general-purpose authoritative DNS server it has a lot of functionality that kube-dns could not reasonably
+be expected to add. See, for example, the [intro](https://docs.google.com/presentation/d/1v6Coq1JRlqZ8rQ6bv0Tg0usSictmnN9U80g8WKxiOjQ/edit#slide=id.g249092e088_0_181) or [coredns.io](https://coredns.io) or the [CNCF webinar](https://youtu.be/dz9S7R8r5gw).
+
+## Proposal
+
+The proposed solution is to enable the selection of CoreDNS as an alternate to Kube-DNS during cluster deployment, with the
+intent to make it the default in the future.
+
+## User Experience
+
+### Use Cases
+
+ * Standard DNS-based service discovery
+ * Federation records
+ * Stub domain support
+ * Adding custom DNS entries
+   * Making an alias for an external name [#39792](https://github.com/kubernetes/kubernetes/issues/39792)
+   * Dynamically adding services to another domain, without running another server [#55](https://github.com/kubernetes/dns/issues/55)
+   * Adding an arbitrary entry inside the cluster domain (for example TXT entries [#38](https://github.com/kubernetes/dns/issues/38))
+ * Verified pod DNS entries (ensure pod exists in specified namespace)
+ * Experimental server-side search path to address latency issues [#33554](https://github.com/kubernetes/kubernetes/issues/33554)
+ * Limit PTR replies to the cluster CIDR [#125](https://github.com/kubernetes/dns/issues/125)
+ * Serve DNS for selected namespaces [#132](https://github.com/kubernetes/dns/issues/132)
+ * Serve DNS based on a label selector
+
+By default, the user experience would be unchanged. For more advanced uses, existing users would need to modify the
+ConfigMap that contains the CoreDNS configuration file.
+
+## Implementation
+
+Each distribution project (kubeadm, minikube, kubespray, and others) will implement CoreDNS as an optional
+add-on as appropriate for that project.
+
+### Client/Server Backwards/Forwards compatibility
+
+No changes to other components are needed.
+
+The method for configuring the DNS server will change. Thus, in cases where users have customized
+the DNS configuration, they will need to modify their configuration if they move to CoreDNS.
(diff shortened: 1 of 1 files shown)


cblecker:

thockin:
I'm OK with this.  Any net change in footprint - CPU or memory required, performance or scale characteristics?

smarterclayton:
Seems reasonable to me as well.  Also want to see some perf diff info - tail latency on representative workloads.

smarterclayton:
Also add to the proposal security implications (I see some interesting things in the current codebase) for @kubernetes/sig-auth-api-reviews

smarterclayton:
On a 6.5k service / 10k pod cluster i was seeing about 18-35mb in use, and on a 15k service / 18k pod cluster I was seeing slightly higher.  No obvious allocations or slow paths on a range of queries.  It seems to fare a bit better than most projects that integrate with kube :)

Would be nice if it upgraded to protobuf

contributors/design-proposals/network/coredns.md
```diff
@@ -0,0 +1,72 @@
...
+ * Standard DNS-based service discovery
+ * Federation records
+ * Stub domain support
+ * Adding custom DNS entries
+   * Making an alias for an external name [#39792](https://github.com/kubernetes/kubernetes/issues/39792)
+   * Dynamically adding services to another domain, without running another server [#55](https://github.com/kubernetes/dns/issues/55)
+   * Adding an arbitrary entry inside the cluster domain (for example TXT entries [#38](https://github.com/kubernetes/dns/issues/38))
+ * Verified pod DNS entries (ensure pod exists in specified namespace)
+ * Experimental server-side search path to address latency issues [#33554](https://github.com/kubernetes/kubernetes/issues/33554)
+ * Limit PTR replies to the cluster CIDR [#125](https://github.com/kubernetes/dns/issues/125)
```
smarterclayton: Are PTR records for services implemented?  I saw https://github.com/coredns/coredns/issues/1074

contributors/design-proposals/network/coredns.md
```diff
@@ -0,0 +1,72 @@
...
+ * Serve DNS for selected namespaces [#132](https://github.com/kubernetes/dns/issues/132)
+ * Serve DNS based on a label selector
+
+By default, the user experience would be unchanged. For more advanced uses, existing users would need to modify the
+ConfigMap that contains the CoreDNS configuration file.
+
```
smarterclayton: Discuss operational characteristics of coredns against existing kube-dns.

Have the existing e2e tests been run against a cluster configured with coredns?

contributors/design-proposals/network/coredns.md
johnbelamaric: Yes, they are. You have to configure the reverse zone to make it work. That means knowing the service CIDR and configuring that ahead of time (would love to have https://github.com/kubernetes/kubernetes/issues/25533 implemented).

Since reverse DNS zones are on classful boundaries, if you have a classless CIDR for your service CIDR (say, a /12), then you have widen that to the containing classful network. That leaves a subset of that network open to the spoofing described in https://github.com/kubernetes/dns/issues/125, and so the issue you reference is to fix that.

We still have that issue (PTR hijacking) with CoreDNS, but we are in a position to fix it if the operator is willing to enable `pods verified`.

contributors/design-proposals/network/coredns.md
johnbelamaric: Ok, I will add some to the proposal regarding the operational characteristics.

Existing e2e tests have not been run yet, see https://github.com/coredns/coredns/issues/993 - we plan on this for sure.

johnbelamaric:
@smarterclayton your comment "would be nice if it upgraded to protobuf" - can you explain further?

We actually can answer gRPC queries, which means protobuf over http/2 and TLS. But of course that requires the client to use that. Right now that is a simple `Query` method that just wraps an ordinary DNS packet. What we'd really like to do in the future is extend the gRPC API to be bi-directional and allow push-based updates. This is especially useful for headless services where the pods may come and go - clients would be informed immediately of changes, rather than having to re-resolve the SRV queries.

contributors/design-proposals/network/coredns.md
smarterclayton: We generally recommend production user disable pod IP DNS for this reason as well.  I would prefer to let pod DNS get deprecated out since it was intentionally a stop gap.

Thanks for the clarity on the reverse CIDR.  I think one part of this that would be good would be to include a sample Corefile in this proposal that implements conformance with the Kube DNS spec.  To someone new to the core file syntax but deeply familiar with the kube dns spec I had to dig through the code to know what I had to set up.  A core file would go a long way to assisting in understanding the implications.

smarterclayton:
No, I mean that you should use a version of the kubernetes client with the protobuf serialization.  That reduces server load significantly.  Not a requirement for this, but in the performance / scale section it's worth mentioning.  I know some people run kube DNS on many many nodes - openshift runs it on *every* node and so the impact of performance on the cluster is key

contributors/design-proposals/network/coredns.md
johnbelamaric: Ok, I have added that. Let me know if I need any more examples (e.g., federation).

johnbelamaric:
@luxas Even with what we get as a CNCF project from Packet, I don't have the resources to do 1000+ node clusters. Are there some other resources available for this sort of test?

johnbelamaric:
Oh, I see. kubemark simulates clusters. That's new to me. We'll look into it.

smarterclayton:
Examples are great.  I'll have someone poke at coredns against our super dense clusters to get a rough idea of the outcome.

johnbelamaric:
Excellent, thanks.

contributors/design-proposals/network/coredns.md
```diff
@@ -0,0 +1,184 @@
...
+The proposed solution is to enable the selection of CoreDNS as an alternate to Kube-DNS during cluster deployment, with the
+intent to make it the default in the future.
+
+## User Experience
+
+### Use Cases
+
+ * Standard DNS-based service discovery
+ * Federation records
+ * Stub domain support
```
bowei: Is the stub domain support interoperable with the current kube-dns configmap?

contributors/design-proposals/network/coredns.md
```diff
@@ -0,0 +1,184 @@
...
+ * Experimental server-side search path to address latency issues [#33554](https://github.com/kubernetes/kubernetes/issues/33554)
+ * Limit PTR replies to the cluster CIDR [#125](https://github.com/kubernetes/dns/issues/125)
+ * Serve DNS for selected namespaces [#132](https://github.com/kubernetes/dns/issues/132)
+ * Serve DNS based on a label selector
+ * Support for wildcard queries (e.g., `*.namespace.svc.cluster.local` returns all services in `namespace`)
+
+By default, the user experience would be unchanged. For more advanced uses, existing users would need to modify the
+ConfigMap that contains the CoreDNS configuration file.
+
+### Configuring CoreDNS
```
bowei: Ideally when bundled with Kubernetes, it should come with a standard configuration file that makes the CoreDNS installation behave exactly the same as the current kube-dns. Additional configuration should be done via the existing kube-system:kube-dns configmap

contributors/design-proposals/network/coredns.md
```diff
@@ -0,0 +1,184 @@
...
+CoreDNS instance will be considered ready when it has finished syncing with the API.
+
+CoreDNS performance metrics can be published for Prometheus.
+
+When a change is made to the Corefile, you can send each CoreDNS instance a SIGUSR1, which will
+trigger a graceful reload of the Corefile.
+
+### Performance and Resource Load
+
+Evaluation of performance and resource load (memory, CPU) is still underway. Initial results show memory consumption is generally less than 40MB without `pods verified`.
```
bowei: Should probably expressed as a function of number of {pods,services}/

contributors/design-proposals/network/coredns.md
```diff
@@ -0,0 +1,184 @@
...
+## Implementation
+
+Each distribution project (kubeadm, minikube, kubespray, and others) will implement CoreDNS as an optional
+add-on as appropriate for that project.
+
+### Client/Server Backwards/Forwards compatibility
+
+No changes to other components are needed.
+
+The method for configuring the DNS server will change. Thus, in cases where users have customized
```
bowei: Is there a good reason to break compatibility here?

johnbelamaric: Since CoreDNS isn't solely a K8s product, it's configuration of these options is more general purpose and so falls outside of our Kubernetes plugin. Internally that would mean that the Kubernetes plugin has to reconfigure the overall server. I don't think that could be done in a deterministic way. That is, we don't know everything a user might configure, so we can't easily modify the overall configuration based on the additional ConfigMaps.

What we *can* do, is provide a tool to convert the existing kube-dns configuration on a cluster into a CoreDNS configuration. So, it would be a one-time operation when moving from kube-dns to CoreDNS.

(I think this answers your other config-related questions above too, let me know if not).

We *may* want to keep the service name `kube-dns`. If we do that, then the switch can be done without loss of service in a production environment. You just deploy CoreDNS, then modify the kube-dns service to point to it. Otherwise you have to delete that service and recreate one with the same IP. In the interim you will have DNS lookup failures.

bowei: The ideal would be to be able to seamlessly drop the CoreDNS Pod in as the replacement for the existing kube-dns. By CoreDNS pod, I mean the CoreDNS + the appropriate configuration for a Kubernetes cluster.

CoreDNS itself (the generic piece) does not have to change, but the ideal scenario would be to have the pod setup in such a way (i.e. via a very small shim sidecar) that preserve compatibility with the original kube-dns, especially wrt the parameters it consumes from the ConfigMap. This way we can switch to using coredns transparently with minimal user disruption. We would probably still need to maintain at least a two release deprecation of the existing ConfigMap and provide upgrade/downgrade automation.

contributors/design-proposals/network/coredns.md
smarterclayton: I'm not sure it's appropriate to reuse the same config map.  Users who have customized their config need to learn a new config.  I don't think replacing ecosystem components will require every component to emulate the other implementations config.

contributors/design-proposals/network/coredns.md
smarterclayton: I'm somewhat skeptical of this approach - I understand the reasoning, but I also don't know that I consider it a requirement for a replacement implementation to emulate the previous config.  I think that is something that might be desirable, but I also don't think I consider it a requirement for someone to use CoreDNS in a conformant way.

szuecs:
First, thanks for the work!

I would like to see some key metrics that we can use to monitor in this proposal.
In my opinion we have a lot of metrics everywhere, but sometimes it's hard to find what you should really use and this should be fixed as early as possible.

johnbelamaric:
@szuecs Yes, that's a good point. Using CPU often isn't really the right metric, because you may not be using all the CPU but could still be dropping packets due to overflowing the socket receive buffer, if queries are coming in really fast.

We can do some application-level metrics like QPS. Although the right values for that based on scaling is going to vary as well. I will look into getting together the right recommendations for this.

contributors/design-proposals/network/coredns.md
bowei: At the very least we need to support existing users of kube-dns to migrate to CoreDNS if CoreDNS is to replace kube-dns as the default provider. I know for sure that we have lots of users that utilize the stub domains via kube-system:kube-dns config map.

miekg: @bowei what are you asking? If CoreDNS support stubdomains? (It does via the *proxy* plugin). Or that you still think we need to automatically convert the config?

bowei: Yes, we would need to have an upgrade/downgrade path for existing users. Otherwise changing coredns to be the default will break people's clusters...

bowei: Alternatively, we could clean up kube-system:kube-dns to be more generic, say be portable across different DNS providers. E.g. make it part of the API for k8s cluster, rather than viewing it as a entirely bound to the existing implementation.

liggitt: I'd expect the DNS API to be defined in terms of the DNS schema that is served, and the pod/service/endpoint inputs, not the config of the DNS server itself

bowei: @liggitt I can see this specific config being out of scope for k8s -- but we still need to support our users across versions in a transparent way. 

If we make coredns the default "kube-dns" in say 1.10, it can't be the case that the users configs are not carried through. The proposal is replace kube-dns.

liggitt: Sure, those deployments that auto installed kube-dns would be responsible for adapting/shimming/converting the kube-dns config to cause the equivalent behavior via coredns if they wanted to switch on upgrade

contributors/design-proposals/network/coredns.md
```diff
@@ -0,0 +1,201 @@
...
+avoid I/O for every query.
+
+CoreDNS can be configured to provide an HTTP health check endpoint, so that it can be monitored
+by a standard Kubernetes HTTP health check. Readiness checks are not currently supported but
+are in the works (see [#588](https://github.com/coredns/coredns/issues/588)). For Kubernetes, a
+CoreDNS instance will be considered ready when it has finished syncing with the API.
+
+CoreDNS performance metrics can be published for Prometheus.
+
+When a change is made to the Corefile, you can send each CoreDNS instance a SIGUSR1, which will
```
bowei: Does CoreDNS support watching a config file and auto-updating when it changes? Would make it easy to consume updated config via ConfigMap.

bowei:
@johnbelamaric would be great to talk briefly about this proposal at the network SIG as a 1.9 feature, I'm sure other people are also interested.

luxas:
Looping in @kubernetes/sig-autoscaling-feature-requests as well -- could we use the metrics server (aka. core metrics pipeline) & a HPA to automatically scale the CoreDNS deployment based on cluster size?
That is, doing it a "native k8s way" instead of a dedicated core-dns/kube-dns only scaler application in the cluster.

miekg:
[ Quoting <notifications@github.com> in "Re: [kubernetes/community] Add core..." ]
>bowei commented on this pull request.
>
>
>
>> +`pods verified` as that could put a high load on the API server. Instead, if you wish to support
>+that functionality, you can run another central deployment and configure the per-node
>+instances to proxy `pod.cluster.local` to the central deployment.
>+
>+All logging is to standard out, and may be disabled if
>+desired. In very high queries-per-second environments, it is advisable to disable query logging to
>+avoid I/O for every query.
>+
>+CoreDNS can be configured to provide an HTTP health check endpoint, so that it can be monitored
>+by a standard Kubernetes HTTP health check. Readiness checks are not currently supported but
>+are in the works (see [#588](https://github.com/coredns/coredns/issues/588)). For Kubernetes, a
>+CoreDNS instance will be considered ready when it has finished syncing with the API.
>+
>+CoreDNS performance metrics can be published for Prometheus.
>+
>+When a change is made to the Corefile, you can send each CoreDNS instance a SIGUSR1, which will
>
>Does CoreDNS support watching a config file and auto-updating when it changes? Would make it easy to consume updated config via ConfigMap.

Nope, there is no support for that currently.


--
Miek Gieben

contributors/design-proposals/network/coredns.md
johnbelamaric: As Miek said, not currently, but certainly it's feasible to add.

johnbelamaric:
@bowei Ok, next meeting is 10/5 it looks like. I will be there.

johnbelamaric:
An update here. @rajansandeep has been looking into kubemark, but I am thinking it's not the right tool. From what he has seen, it's focused around API server performance when spinning up lots of pods - does it also create services? The real driver for resource usage here is services and endpoints (and pods if `pods verified` is on), not nodes. And of course actual queries being run.

He is still looking into it, but in parallel we're going to spin up a smaller cluster and create services and manual endpoints. That way we should be able to push it to a much larger number of services than normal (no actual backing pods). Thoughts on that?

bowei:
There is some setup in the scale tests where lots of fake nodes (i.e. kubelets)/services are created, this would make it possible to simulate large clusters without actually running one. @wojtek-t

wojtek-t:
> An update here. @rajansandeep has been looking into kubemark, but I am thinking it's not the right tool. From what he has seen, it's focused around API server performance when spinning up lots of pods - does it also create services? The real driver for resource usage here is services and endpoints (and pods if pods verified is on), not nodes. And of course actual queries being run.

That's not entirely true. In kubemark you can easily create services (and some of our tests are doing that).
The only caveat is that the underlying pods are not really running, so those services can't really serve anything. 

Sorry for not reading this whole proposal and comments above - I'm don't have full context, but if you are only interested in understanding how fast you coredns can process new services and/or their endpoints and you don't really need to use those services, kubemark should be perfect tool for doing that.

Let me know if I can somehow help with that - I will try to go over the proposal and comments to get more context and will get back to this early next week (I'm out of the city Fri-Sun).

johnbelamaric:
@wojtek-t thanks. CoreDNS will have a watch on services and endpoints, so yeah, we need to measure the impact on memory and CPU with different levels of services and endpoints, as well as different rates of change in those (along with different QPS levels). Any pointers you have on getting kubemark to do this would be helpful.

thockin:
Assign back to me once it is LGTM'ed and ready for another look.

cblecker:
@thockin There is no approve yet in community, so an LGTM would merge it. Is that your intent?

wojtek-t:
> @wojtek-t thanks. CoreDNS will have a watch on services and endpoints, so yeah, we need to measure the impact on memory and CPU with different levels of services and endpoints, as well as different rates of change in those (along with different QPS levels). Any pointers you have on getting kubemark to do this would be helpful.

Kubemark is generally setting cluster control plane (apiserver, etcd, controller manager, ...) + a number of "fake nodes" (those nodes pretend to be real nodes, but they are not starting real containers, mount volumes, etc. - they just fake those operations). But from apiserver perspective they behave as real nodes (e.g. they confirm that a pod is running, or has been killed etc.)
See https://github.com/kubernetes/community/blob/master/contributors/devel/kubemark-guide.md for more information.

So once you have kubemark setup, you can run any test or workflow against it. In particular we are running some of load tests against it:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/load.go
which has an option to create services:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/load.go#L184

With 5000-node kubemark cluster, our main test is creating ~16k services, with 150.000 pods (each of them being part of some service). But you can run any test you want against the kubemark control plane.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1100)

rajansandeep:
### Performance-test for CoreDNS in Kubernetes
The performance test was done in GCE with the following components:
* CoreDNS system with machine type : n1-standard-1 ( 1 CPU, 2.3 GHz Intel Xeon E5 v3 (Haswell))
* Client system with machine type: n1-standard-1 ( 1 CPU, 2.3 GHz Intel Xeon E5 v3 (Haswell))
* Kubemark Cluster with 5000 nodes

CoreDNS and client are running out-of-cluster (due to it being a Kubemark cluster)

The following is the summary of the performance of CoreDNS.
CoreDNS cache was disabled.

Services (with 1% change per minute*) | Max QPS** | Latency (Median) | CoreDNS memory (at max QPS) | CoreDNS CPU (at max QPS) |
------------ | ------------- | -------------- | --------------------- | ----------------- | 
1,000 | 18,000 | 0.1 ms | 38 MB | 95 % |
5,000 | 16,000 | 0.1 ms | 73 MB | 93 % |
10,000 | 10,000 | 0.1 ms | 115 MB | 78 % |

*We simulated service change load by creating and destroying 1% of services per minute.
** Max QPS with < 1 % packet loss

johnbelamaric:
FYI, latency is so low and the same across all, we are double checking to make sure the test is valid.

luxas:
For reference, what are these numbers for kube-dns? That might be a valuable comparision to perform, possibly encouraging the broader community to switch over sooner rather than later...

johnbelamaric:
We're getting real responses with that low latency. Probably the two VMs (coredns and client) are on the same physical host or something. Although I imagine latency between GCE hosts is extremely low. Anyway, it can be taken to mean that:

1) Number of services does not affect latency; and
2) Latency inside CoreDNS is very low

As an aside, before we merged https://github.com/coredns/coredns/pull/1149, the latency on the 10k services was really high (like 1 second).

johnbelamaric:
@luxas we can do that. The numbers above are uncached. So, we would need to configure `dnsmasq` to not cache anything (forward all queries to `kube-dns`) for an equivalent test.

For general `cache` tests, I would expect at least 30k QPS per core from CoreDNS, based on past experience. I also would not be surprised if the `dnsmasq` cache were faster than that, based on other testing we have done versus C-based servers. I have no way of estimating the cache hit ratio in a "typical" cluster, so it's hard to evaluate this in a general way. Nonetheless, with the numbers above I think the overall solution with CoreDNS will be better even if the cache performance isn't quite as good.

johnbelamaric:
Ok, I think this latest commit covers all the open questions, except running the equivalent kube-dns numbers (which I don't think necessarily belongs here, but we can add it if others wish).

rajansandeep:
### Performance-test for Kubedns + dnsmasq  in Kubernetes
The performance test was done in GCE with the following components:
* Kubedns + dnsmasq system with machine type : n1-standard-1 ( 1 CPU, 2.3 GHz Intel Xeon E5 v3 (Haswell))
* Client system with machine type: n1-standard-1 ( 1 CPU, 2.3 GHz Intel Xeon E5 v3 (Haswell))
* Kubemark Cluster with 5000 nodes

Kubedns + dnsmasq and client are running out-of-cluster (due to it being a Kubemark cluster)

The following is the summary of the performance of Kubedns + dnsmasq.
Cache was disabled.

Services (with 1% change per minute*) | Max QPS** | Latency (Median) | Kubedns + dnsmasq memory (at max QPS) | Kubedns + dnsmasq CPU (at max QPS) |
------------ | ------------- | -------------- | --------------------- | ----------------- | 
1,000 | 8,000 | 0.2 ms | 45 MB | 85 % |
5,000 | 7,000 | 0.2 ms | 97 MB | 89 % |
10,000 | 6,000 | 0.2 ms | 191 MB | 81 % |

*We simulated service change load by creating and destroying 1% of services per minute.
** Max QPS with < 1 % packet loss

smarterclayton:
Really encouraging numbers - I'm seeing similar numbers from large clusters
that I am testing against.

bowei:
/lgtm for alpha

For beta, we would want to have migration story for existing users of kube-dns.

warmchang:
👍

PR #1103, opened 2017-09-20
Update the pointer to a new doc and create a link to archived document

vaikas-google:


cblecker:
This needs to be updated via sigs.yaml, and a `make` run:
https://github.com/kubernetes/community/blob/c83b695e15d630aad4e291c3e91bda16c328ed2b/sigs.yaml#L605

xiangpengzhao:
@cblecker I saw several PRs trying to update SIG README in which the authors didn't know they need to run `make` to update the README. So I think we should document this. I suggest to add a tip field in sigs.yaml. When running `make`, the tip will be automatically added to the header of each of the SIG README. WDYT?

vaikas-google:
Thanks for the review, PTAL.

PR #1105, opened 2017-09-20
[WIP] Adding initial design doc for local PV distributed static provisioner.

verult:
Proposal for improving the static provisioner to achieve the following goals:

1. Limit all PV object interactions to a single master pod in order to minimize node access to the Kubernetes system.
1. Assume disk deletion is asynchronous, as required by block volumes.

/kind design
/sig storage
/cc @kubernetes/sig-storage-proposals @dhirajh
/assign @msau42

### DIFF
--- a/contributors/design-proposals/local-pv-provisioner.md
+++ b/contributors/design-proposals/local-pv-provisioner.md
@@ -1,13 +1,13 @@
 # Local PV Distributed Static Provisioner
 Authors: @verult, @msau42
 
-The current preference is [design 3](#design-3-event-driven-with-custom-resources).
+The current preference is [design 5](#design-5-worker-as-jobs-watcher-in-daemonset).
         
 ## Design Goals
 1. Limit all PV object interactions to a single master pod in order to minimize node access to the Kubernetes system. The original discussion is [here](https://github.com/kubernetes/community/pull/989#discussion_r135397619).
 1. Assume disk deletion is asynchronous, as required by block volumes.
 
-In all designs mentioned below, the provisioner will be run in a master-minion configuration. ProvisionerWorkers are deployed on every node in a DaemonSet (just as before) and ProvisionerMaster is deployed in a Deployment on a trusted node.
+In all designs mentioned below, the provisioner will be run in a master-worker configuration. ProvisionerWorkers are deployed on every node in a DaemonSet (just as before) and ProvisionerMaster is deployed in a Deployment on a trusted node.
 
 ## Related Works
 * Original provisioner design, [here](https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/provisioner#design).
@@ -36,7 +36,7 @@ The following fields are retrieved as part of the `GetWorkerStatus()` call:
 
 | Field               | Description |
 | --- | --- |
-| `LocalPVs`          | A collection of PVs mounted in the local storage directory on the node. The discoverer is triggered to acquire this information. |
+| `LocalVolumes`          | A collection of PVs mounted in the local storage directory on the node. The discoverer is triggered to acquire this information. |
 | `DeletesInProgress` | A collection of PVs currently being deleted asynchronously. This status is managed by the cleaner. |
 
 Every worker must maintain the invariant that `DeletesInProgress` is correct even after a worker restart. In the case that a user manually deletes the PV currently being deleted by a worker, master can only depend on the WorkerStatus to ensure a replacement PV isn't created.
@@ -54,7 +54,7 @@ Each WorkerInfo entry contains the following fields:
 ### Goroutine
 The goroutine will start by populating the `apiCache` using the populator logic that already exists in the previous design. Then, it will run the following loop with several seconds of delay in between executions:
 1. Call the RPC `GetWorkerStatus()`.
-1. Using `LocalPVs` from the worker status, **reconcile** with the `PVCache`. Create an additional PV if it exists in `LocalPVs` but not in `PVCache` *and no delete is in progress for this PV, i.e. it's not in `DeletesInProgress`*.
+1. Using `LocalVolumes` from the worker status, **reconcile** with the `PVCache`. Create an additional PV if it exists in `LocalVolumes` but not in `PVCache` *and no delete is in progress for this PV, i.e. it's not in `DeletesInProgress`*.
 1. For each PV in `VolumeReleased` state (ready for cleanup), if it's not in `DeletesInProgress`, call the `Delete()` RPC.
 
 Master must delete the PV object after the disk delete operation is complete. In order for master to know when this occurs, the worker must persist a "complete" state, because in the wake of worker failure, master has to differentiate between an error and a completion. Master also needs to notify the worker that the PV is deleted, so the persisted state can be removed.
@@ -122,12 +122,12 @@ To ensure correctness, master and worker never write to the same field. Specific
 * After a `DeleteOperation` is created, its `PV` field is never modified.
 * Master can only create and delete `DeleteOperation`s. It never modifies those objects.
 * Worker can only modify `DeleteStatus`. It can never create or delete `DeleteOperation`s.
-* Only a worker can write to `LocalPVs`.
+* Only a worker can write to `LocalVolumes`.
 
 ### Worker Provisioner
 Create the ProvisionerStatus object on initialization.
 
-The Discoverer periodically probes the filesystem and updates ProvisionerStatus `LocalPVs`. This can be further improved by using a filesystem watch.
+The Discoverer periodically probes the filesystem and updates ProvisionerStatus `LocalVolumes`. This can be further improved by using a filesystem watch.
 
 The Deleter uses ProvisionerStatus to keep track of states for each PV being deleted. Deletion is started when the ProvisionerStatus is updated (by the master) and some PVs are in the Requested state. Details will be included in the async volume cleaner design.
 
@@ -135,26 +135,51 @@ The Deleter uses ProvisionerStatus to keep track of states for each PV being del
 A single goroutine processes events from all informers.
 
 On ProvisionerStatus Informer update, perform the following actions:
-1. For deletion: If a PV is in the Completed state, delete the PV from the API server and remove the entry from `DeleteOperations`.
-1. For discovery: compare `LocalPVs` with existing PVs in the API server. If a PV exists in `LocalPVs` but has not been persisted, and if the disk is not about to be deleted (i.e. in Requested or InProgress state), then create the PV.
+1. For deletion: If a volume is in the Completed state, delete the corresponding PV from the API server and remove the entry from `DeleteOperations`.
+1. For discovery: compare `LocalVolumes` with existing PVs in the API server. If a volume exists in `LocalVolumes` but has not been persisted, and if the disk is not about to be deleted (i.e. in Requested or InProgress state), then create the PV.
 
 On PV Informer update: if the updated PV is in VolumeReleased state, add the PV to `DeleteOperations` with the Requested state.
 
 On Node deletion: delete the associated ProvisionerStatus object.
 
-**TODO**: Consider marking cache dirty when its corresponding API object is modified within master and ignoring Informer events if cache is dirty. This ensures master is always operating on the most up-to-date data. One possible bad scenario it could prevent:
-1. A worker discovers a volume and updates `LocalPVs`.
-1. A PV corresponding to this volume already exists, and now gets released.
-1. The PV Informer fires on master, which then adds a new entry to `DeleteOperations` to signal a delete request.
-1. The ProvisionerStatus Informer fires on master from the discovered volume. Since the cache is not updated with the new `DeleteOperations`, master proceeds to create a volume, while a delete operation is underway in reality.
+**TODO**: Consider marking cache dirty when its corresponding API object is modified within master and ignoring Informer events if cache is dirty. This ensures master is always operating on the most up-to-date data. 
+
+## Design 4: Hybrid RPC-CRD Approach
+Keep the same CRD object described in Design 3, but only master has write access to them. Workers must persist its delete state locally and report its status to master through gRPC. Master makes requests to workers by writing to CRD, and fetches worker status by polling (similar to Design 1).
+
+Several modifications to Design 1:
+* For discovery, do not rely on worker to report its identity as it can be compromised. Identify workers using the Endpoints object (pod name and node name).
+* In future versions, master goroutines should be event-driven, as constant polling is not scalable to large number of nodes. A worker pings the master when it's ready to report status, and master makes a separate call to fetch status from the worker. Worker must continuously ping master until it receives a status request to ensure master recognizes the status update. As a safety check, master should also poll workers periodically, but with very low frequency.
+  * It's not as secure to have workers push status directly because a worker can't be accurately identified from a gRPC request (the request IP can be spoofed).
+  * It's more convenient to use gRPC server streams, but it's hard to scale this to thousands of nodes as streams require TCP connections to be kept open.
+
+**TODO**: Use network policy to limit pod communication to only within the provisioner system?
+
+## Design 5: Worker as Jobs, Watcher in DaemonSet
+One of the major issues with previous approaches is worker identity - when a worker tries to communicate with master, it has to prove it's the worker it claims to be. To fix these issues, each worker must maintain different credentials.
+
+Alternatively, master can actively create worker Jobs to perform discovery and delete rather than having long-running worker pods in a DaemonSet sending data to the master. For discovery, workers can store the list of local volumes in its container STDOUT (**TODO**: decide on a form of structured output rather than plain text). For deletion, the status of the Job acts as an indication of deletion state. Master has perfect knowledge of which Job runs on which node, so if any worker is compromised, the damage is contained within the node the worker is scheduled on.
+
+To trigger discovery, we have local volume directory watcher pod running in a DaemonSet. It keeps an in-memory cache of local volumes and periodically checks the volume directory for changes. If there is a change, it notifies master through either gRPC or a CRD notification object, and master can then spawn a discovery Job.
+
+To ensure the provisioner system has up-to-date state on initialization, the following needs to be done:
+* When the master initializes, it spawns discovery jobs on every node to make sure it knows about all local volumes in the cluster.
+* When a watcher initializes, it should notify master so a discovery Job can be spawned to fetch the latest list of local volumes.
+
+Master can be made completely event driven using Job informers.
 
 ## Conclusion
 One main downside of design 2 is the possibility of duplicate discover and delete operations, as a result of the cache not being updated immediately. Also, because the logic of PV updates is divided over the network, it's easy for PVs to end up in an inconsistent state.
 
-Design 3 seems like the best approach because:
+Design 3 has the following advantages:
 * Master has perfect knowledge of when to issue PV delete (based on the Complete DeleteStatus) without requiring multiple acknowledgment messages between master and worker.
 * Because the API server persists state reliably, we can simplify our reasoning with failure modes.
 * Provisioner discovery is no longer necessary.
 * Retries and backoffs are no longer necessary.
 * Provisioner state is more visible to admins.
 * Almost entirely event-driven (the only exception being the filesystem probe).
+But one major concern with this approach is each worker has access to all ProvisionerStatus objects. If a worker is compromised, it could disrupt the entire provisioner system instead of just its node.
+
+Design 4 is a compromise, taking advantage of benefits from other designs. However, it would not work if the network can be damaged to the point that a compromised worker can intercept all traffic to the IP of another normal worker.
+
+Design 5 removes the need to maintain worker identity, and thus is the simplest design that protects against our security concerns. Each pod in the DaemonSet is also smaller in size and actual workers are spawned only as needed.
--- a/contributors/design-proposals/local-pv-provisioner.md
+++ b/contributors/design-proposals/local-pv-provisioner.md
@@ -1,0 +1,160 @@
+# Local PV Distributed Static Provisioner
+Authors: @verult, @msau42
+
+The current preference is [design 3](#design-3-event-driven-with-custom-resources).
+        
+## Design Goals
+1. Limit all PV object interactions to a single master pod in order to minimize node access to the Kubernetes system. The original discussion is [here](https://github.com/kubernetes/community/pull/989#discussion_r135397619).
+1. Assume disk deletion is asynchronous, as required by block volumes.
+
+In all designs mentioned below, the provisioner will be run in a master-minion configuration. ProvisionerWorkers are deployed on every node in a DaemonSet (just as before) and ProvisionerMaster is deployed in a Deployment on a trusted node.
+
+## Related Works
+* Original provisioner design, [here](https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/provisioner#design).
+* Disk cleaner (deleter) implementation for block volumes, [here](https://github.com/kubernetes-incubator/external-storage/pull/312).
+
+## Design 1: Master-Heavy
+
+### High Level Overview
+The static provisioner will be split into a single master (ideally placed on the master node) and many workers (one per node). The master contains the bulk of provisioner logic, including syncing PV objects with the API server and triggering discovery and deletion. Workers are left to perform the physical discovery and deletion of the local volumes on its node. gRPC is used to communicate between master and workers, with the master as the client and each worker as a separate server.
+
+The master keeps track of `WorkerInfo`, a collection containing one entry for each worker keeping track of relevant information. The master also runs a group of goroutines, one for each worker, which controls and interacts with the workers.
+
+### Worker Provisioner
+The worker provisioner exposes the following RPCs:
+
+| RPC                           | Description |
+| --- | --- |
+| `GetWorkerName()`             | Returns the worker provisioner's name. |
+| `GetProvisionerStatus()`      | Returns the worker status, as described in the next section. |
+| `Delete(PersistentVolume)`    | Starts the asynchronous delete of the given PV. |
+
+In addition, the Discoverer and Deleter from the previous design are kept here, but they only perform filesystem operations. All other logic is moved to the master.
+
+### ProvisionerStatus
+The following fields are retrieved as part of the `GetWorkerStatus()` call:
+
+| Field               | Description |
+| --- | --- |
+| `LocalPVs`          | A collection of PVs mounted in the local storage directory on the node. The discoverer is triggered to acquire this information. |
+| `DeletesInProgress` | A collection of PVs currently being deleted asynchronously. This status is managed by the cleaner. |
+
+Every worker must maintain the invariant that `DeletesInProgress` is correct even after a worker restart. In the case that a user manually deletes the PV currently being deleted by a worker, master can only depend on the WorkerStatus to ensure a replacement PV isn't created.
+
+### WorkerInfo
+Each WorkerInfo entry contains the following fields:
+
+| Field                 | Description |
+| --- | --- |
+| `Name`                | Name of the worker provisioner. |
+| `Endpoint`            | The worker's host address. |
+| `PVCache`             | Cache of API PV objects managed by this worker. |
+| `StopCh`              | Used to terminate a goroutine. |
+
+### Goroutine
+The goroutine will start by populating the `apiCache` using the populator logic that already exists in the previous design. Then, it will run the following loop with several seconds of delay in between executions:
+1. Call the RPC `GetWorkerStatus()`.
+1. Using `LocalPVs` from the worker status, **reconcile** with the `PVCache`. Create an additional PV if it exists in `LocalPVs` but not in `PVCache` *and no delete is in progress for this PV, i.e. it's not in `DeletesInProgress`*.
+1. For each PV in `VolumeReleased` state (ready for cleanup), if it's not in `DeletesInProgress`, call the `Delete()` RPC.
+
+Master must delete the PV object after the disk delete operation is complete. In order for master to know when this occurs, the worker must persist a "complete" state, because in the wake of worker failure, master has to differentiate between an error and a completion. Master also needs to notify the worker that the PV is deleted, so the persisted state can be removed.
+
+**TODO**: gRPC backoff strategy in the case of network partition or worker failure.
+
+### Worker Discovery
+The master needs to have the host address of every worker in order to communicate with them directly. To do this, workers expose a single Service, which is associated with an Endpoints object containing all worker addresses.
+
+To listen to worker membership changes, master uses an Endpoints informer. An informer update is triggered for every endpoint addition and removal. For each informer update, master sends a `GetWorkerName()` call to every endpoint to obtain their identity. If master is aware of the worker, then it updates the worker's `Endpoint`. Otherwise, master adds a new entry to `WorkerInfo` and spawns a new goroutine. If a worker no longer exists, master sends a signal through the `StopCh` of the goroutine handling the worker associated with this endpoint, then deletes the associated entry. **TODO**: Master must wait until goroutine finishes before `WorkerInfo` entry is removed, but this wait could delay the update of other Endpoints. This is also related to gRPC backoff strategy. Need to handle this.
+
+At initialization, master iterates through fetched endpoints and adds them appropriately to the system as described above. 
+
+It's possible for a new pod to reuse the IP of a deleted pod, so even if the endpoint stays the same over time, the associated worker might be different. To fix this issue, master can pass the target worker's name in every RPC call, and have each worker verify the name. Eventually, worker discovery will assign the Endpoint to the correct worker.
+
+In the case of network partition or worker failure during the `GetWorkerStatus()` call, retry a few times, then skip this worker. If the Endpoint is unreachable, the Informer will eventually be triggered, which will correct this scenario.
+
+## Design 2: Worker-Heavy
+
+### High Level Overview
+Another possible design is to leave most of provisioner logic inside workers, only moving PV API interactions to the master. The worker still maintains a PV cache, which is populated by the master, but the difference in this design is it's solely updated by the master. Without this difference, managing the correctness of cache data becomes more difficult, especially when the master fails. When a worker needs to update a PV object, it puts the request in a queue, which is periodically polled by the master.
+
+### Worker Provisioner
+The worker provisioner retains most key components from the previous design, including the Cache, Discoverer, and Deleter. The Populator is moved to master because it interacts with the API server. In order to interact with the master, it exposes the following gRPC calls:
+
+| RPC                      | Description |
+| --- | --- |
+| `Add/Update/RemovePV(PersistentVolume)` | Updates the Cache with the given PV. Called by master when PV informer is triggered. |
+| `PushCurrentPVs([]PersistentVolume)`    | Merges given list of PVs with the Cache. Used by the Populator. |
+| `GetPVRequest()`                        | Peeks at the next unhandled request in the queue. |
+| `CompletePVRequest(PVRequest)`          | Signals that the last request is completed and can be removed from the queue. |
+
+In the case of master failure while a request is being processed, the worker needs to keep record of the request so that a new master can process it again. Thus it's necessary for master to signal request completion.
+
+### Master Provisioner
+The populator is initialized so that it calls the appropriate gRPC for every PV Informer update. Then it populates each worker's cache by calling `PushCurrentPVs()`, passing the collection of PVs to be controlled by this worker.
+
+The master then performs worker discovery as described in the [Worker Discovery](#worker-discovery) section of design 1, and spawns a goroutine for each worker. Each goroutine periodically calls `GetPVRequest()`, performing that request if it receives one. Then it calls `CompletePVRequest()` to signal the request completion.
+
+## Design 3: Event-Driven with Custom Resources
+
+### High Level Overview
+The previous two designs are complicated by the asynchronous nature of delete and the presence of failures. 
+As recommended by @dhirajh, another design is to persist ProvisionerStatus as a Custom Resource, and to drive all operations with PV and ProvisionerStatus Informer events.
+
+To ensure no malicious pods can affect provisioner execution, the CustomResourceDefinition and all pods must be in a trusted admin namespace.
+
+In the following description, we assume all Informer updates are cached and events buffered. On initialization, all caches are populated, and every Informer update event handler is executed once to ensure the entire provisioner system is in a consistent state.
+
+### ProvisionerStatus CRD
+The CRD contains the following fields (similar to design 1):
+
+* `LocalVolumes` - A collection of volumes mounted in the local storage directory on the node. The discoverer is triggered to acquire this information.
+* `DeleteOperations` - A list keeping track of information related to each delete operation. This status is managed by the cleaner. It contains the following fields:
+  * PV - The PV to be deleted on disk.
+  * DeleteStatus - takes one of the following values:
+    * Requested
+    * InProgress
+    * Complete
+    * Error
+    
+There exists one ProvisionerStatus object per node.
+
+To ensure correctness, master and worker never write to the same field. Specifically:
+* After a `DeleteOperation` is created, its `PV` field is never modified.
+* Master can only create and delete `DeleteOperation`s. It never modifies those objects.
+* Worker can only modify `DeleteStatus`. It can never create or delete `DeleteOperation`s.
+* Only a worker can write to `LocalPVs`.
+
+### Worker Provisioner
+Create the ProvisionerStatus object on initialization.
+
+The Discoverer periodically probes the filesystem and updates ProvisionerStatus `LocalPVs`. This can be further improved by using a filesystem watch.
+
+The Deleter uses ProvisionerStatus to keep track of states for each PV being deleted. Deletion is started when the ProvisionerStatus is updated (by the master) and some PVs are in the Requested state. Details will be included in the async volume cleaner design.
+
+### Master Provisioner
+A single goroutine processes events from all informers.
+
+On ProvisionerStatus Informer update, perform the following actions:
+1. For deletion: If a PV is in the Completed state, delete the PV from the API server and remove the entry from `DeleteOperations`.
+1. For discovery: compare `LocalPVs` with existing PVs in the API server. If a PV exists in `LocalPVs` but has not been persisted, and if the disk is not about to be deleted (i.e. in Requested or InProgress state), then create the PV.
+
+On PV Informer update: if the updated PV is in VolumeReleased state, add the PV to `DeleteOperations` with the Requested state.
+
+On Node deletion: delete the associated ProvisionerStatus object.
+
+**TODO**: Consider marking cache dirty when its corresponding API object is modified within master and ignoring Informer events if cache is dirty. This ensures master is always operating on the most up-to-date data. One possible bad scenario it could prevent:
+1. A worker discovers a volume and updates `LocalPVs`.
+1. A PV corresponding to this volume already exists, and now gets released.
+1. The PV Informer fires on master, which then adds a new entry to `DeleteOperations` to signal a delete request.
+1. The ProvisionerStatus Informer fires on master from the discovered volume. Since the cache is not updated with the new `DeleteOperations`, master proceeds to create a volume, while a delete operation is underway in reality.
+
+## Conclusion
+One main downside of design 2 is the possibility of duplicate discover and delete operations, as a result of the cache not being updated immediately. Also, because the logic of PV updates is divided over the network, it's easy for PVs to end up in an inconsistent state.
+
+Design 3 seems like the best approach because:
+* Master has perfect knowledge of when to issue PV delete (based on the Complete DeleteStatus) without requiring multiple acknowledgment messages between master and worker.
+* Because the API server persists state reliably, we can simplify our reasoning with failure modes.
+* Provisioner discovery is no longer necessary.
+* Retries and backoffs are no longer necessary.
+* Provisioner state is more visible to admins.
+* Almost entirely event-driven (the only exception being the filesystem probe).
--- a/contributors/design-proposals/local-pv-provisioner.md
+++ b/contributors/design-proposals/local-pv-provisioner.md
@@ -43,6 +43,7 @@ The following fields are retrieved as part of the `GetWorkerStatus()` call:
 
 ### Master
 The master provisioner exposes the following RPC:
+
 | RPC                           | Description |
 | --- | --- |
 | `ProvisionerStatusReady()`             | Notifies master that there is a status update. Also called on worker initialization. |
@@ -183,6 +184,7 @@ If there is a network failure, the gRPC call will fail and the watcher retries w
 ### Master
 Master exposes the same RPC calls as Design 1 to enable discovery notifications.
 The master provisioner exposes the following RPC:
+
 | RPC                           | Description |
 | --- | --- |
 | `LocalVolumeReady()`             | Notifies master that there are new local volumes. Also called on watcher initialization. |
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,92 @@
...
+| RPC                           | Description |
+| --- | --- |
+| `GetMinionName()`             | Returns the minion provisioner's name. |
+| `GetMinionStatus()`           | Returns the minion status, as described in the next section. |
+| `Delete(PersistentVolume)`    | Starts the asynchronous delete of the given PV. |
+
+In addition, the Discoverer and Deleter from the previous design are kept here, but they only perform filesystem operations. All other logic is moved to the master.
+
+### MinionStatus
+The following fields are retrieved as part of the `GetMinionStatus()` call:
```
jsafrane: How often is `GetMinionStatus` polled by master? Would it make sense to add `Watch()` call that would return `stream Events`? It would either poke the master that it should call `GetMinionStatus` or the event could even contain what has changed so the master does not need to call `GetMinionStatus` over and over again.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Minion Provisioner
+The minion provisioner retains most key components from the previous design, including the Cache, Discoverer, and Deleter. The Populator is moved to master because it interacts with the API server. In order to interact with the master, it exposes the following gRPC calls:
+
+| RPC                      | Description |
+| --- | --- |
+| `Add/Update/RemovePV(PersistentVolume)` | Updates the Cache with the given PV. Called by master when PV informer is triggered. |
+| `PushCurrentPVs([]PersistentVolume)`    | Merges given list of PVs with the Cache. Used by the Populator. |
+| `GetPVRequest()`                        | Peeks at the next unhandled request in the queue. |
```
jsafrane: Again, this could return `stream PVRequest` so minion could return new PVs as they are created/mounted without excessive polling

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,92 @@
...
+
+The master keeps track of `MinionInfo`, a collection containing one entry for each minion keeping track of relevant information. The master also runs a group of goroutines, one for each minion, which controls and interacts with the minions.
+
+### Minion Provisioner
+The minion provisioner exposes the following RPCs:
+
+| RPC                           | Description |
+| --- | --- |
+| `GetMinionName()`             | Returns the minion provisioner's name. |
+| `GetMinionStatus()`           | Returns the minion status, as described in the next section. |
```
jsafrane: What about `ListAndWatch()` that would return `stream of VolumeEvent` that returns all existing directories + waits for event for every created / deleted directory?

Using golang syntax instead of protobuf:

```go
type VolumeEvent struct {
    Volume Volume
    EventType string
}

type Volume struct {
    DirectoryName string
    StorageClassName string
    Capacity ...
    ...
}

type EventType string
const EventTypeAdded = "added" /* new directory found */
const EventTypeAdded = "deleted" /* something deleted existing directory, master should mark it as unusable somehow */
```

jsafrane: Replying to myself, ListAndWatch needs to be smarter. Master starts, reads existing PVs from API server, runs ListAndWatch and needs to mark all directories that were deleted while the master was offline so user knows that the PV is unusable. It won't get "deleted" event, directory was removed before the master started -> its PV is still marked as available -> troubles.

This is implementation detail though, there are many ways out of this. I still recommend something like `ListAndWatch`.

jsafrane:
If I understand it correctly the main driver for this approach is security - we don't want a pod on a node to mess up with all PersistentVolumes in the cluster, right?

In both designs we assume that service EndPoints are secure enough and nobody can can put there an evil pod that would pretend to be minion provisioner. Attacker would need to be able to put a pod to the same namespace as the other minions and that seems to be hard enough to me. @liggit, WDYT?

Regarding the detailed design, I don't like constant polling, some event-based interface would be better to me. Anyway, all this belongs to external-storage and we may polish it there in smaller steps and with schedule that's not related to Kubernetes, right?

My personal preference would be design 1), it has nicer API to me.

dhirajh:
As @jsafrane asked, is the driving goal to avoid modifying PVs (and any special privs needed to do that)? Is it ok to use and manage some other resource type like a custom resource type? Here is the reason I am asking:

I find this direct communication via gRPC to be not quite Kubernetes way of doing things. In many other such use cases in Kubernetes things generally work with a declarative approach that leverages the API server to store desired state and current state. What if we defined a custom resource  say PVMinionInfo or PVDeleteStatus? Each minion would create an object of this resource type for each Volume that it has discovered and managed. The resource would track the cleaning (a.k.a Delete) status of the volume (minions would update that status). The states could be "NeedsDeletion", "DeletesInProgress" and  "DeleteCompleted". This state machine achieves the following

- It accurately lets the master know (that watches these objects) when it is safe to make the privileged call to delete a PV (i.e. when DeleteCompleted state is reached)
- Minions only update the status of these resources and don't have to communicate directly with master, nor do they need to discover each others location.
- It allows master and minions to freely reboot and restart, nothing impacts the state information as all that is persisted in the API server/etcd. 
- Admins can get great insight into status of deletion (cleaning) by just running existing kubectl CLI or API calls in k8s to list these objects.
- You can easily use the ListAndWatch() libraries k8s provides (as jsafrane mentioned in another context) in master code to observe and react.
- Finally, this approach is in my mind a very kubernetes-like declarative approach for dealing with state transitions and for cooperation between different controllers (in this case master and minions).

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,92 @@
...
+### Goroutine
+The goroutine will start by populating the `apiCache` using the populator logic that already exists in the previous design. Then, it will run the following loop with several seconds of delay in between executions:
+1. Call the RPC `GetMinionStatus()`.
+1. Using `LocalPVs` from the minion status, **reconcile** with the `APICache`. Create an additional PV if it exists in `LocalPVs` but not in `APICache` *and no delete is in progress for this PV, i.e. it's not in `DeletesInProgress`*.
+1. For each PV in `VolumeReleased` state (ready for cleanup), if it's not in `DeletesInProgress`, call the `Delete()` RPC. **TODO**:  How to delete PV object after disk delete operation is complete? Depends on cleaner implementation.
+
+**TODO**: gRPC backoff strategy in the case of network partition or minion failure.
+
+### Minion Discovery
+The master needs to have the host address of every minion in order to communicate with them directly. To do this, minions expose a single Service, which is associated with an Endpoints object containing all minion addresses.
```
liggitt: how is this different from the node API object, which contains addresses in status?

liggitt:
> If I understand it correctly the main driver for this approach is security - we don't want a pod on a node to mess up with all PersistentVolumes in the cluster, right?
> 
> In both designs we assume that service EndPoints are secure enough and nobody can can put there an evil pod that would pretend to be minion provisioner. Attacker would need to be able to put a pod to the same namespace as the other minions and that seems to be hard enough to me. @liggit, WDYT?

Some context about what this is trying to replace would be helpful. Is this moving a controller out of the controller manager binary? Or is it replacing something else? Or is it a net new component?

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,92 @@
...
+## Related Works
+* Original provisioner design, [here](https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/provisioner#design).
+* Disk cleaner (deleter) implementation for block volumes, [here](https://github.com/kubernetes-incubator/external-storage/pull/312).
+
+## Design 1: Master-Heavy
+
+### High Level Overview
+The static provisioner will be split into a single master (ideally placed on the master node) and many minions (one per node). The master contains the bulk of provisioner logic, including syncing PV objects with the API server and triggering discovery and deletion. Minions are left to perform the physical discovery and deletion of the local volumes on its node. gRPC is used to communicate between master and minions, with the master as the client and each minion as a separate server.
+
+The master keeps track of `MinionInfo`, a collection containing one entry for each minion keeping track of relevant information. The master also runs a group of goroutines, one for each minion, which controls and interacts with the minions.
```
smarterclayton: Minion terminology is problematic for historical reasons because we have been trying to be consistent.  I assume you're referring to an agent that runs on the node that is tracking volume state?  Why not call it "node volume detector" and be consistent with "node problem detector"?

I don't like using minion because it's too generic and it overlaps too heavily with our historical work around node, and it's not clear that this is a very specific agent.

verult:
> If I understand it correctly the main driver for this approach is security - we don't want a pod on a node to mess up with all PersistentVolumes in the cluster, right?

Yes that's correct. This is an improve on the current external provisioner that stemmed from this discussion: https://github.com/kubernetes/community/pull/989#discussion_r135397619

verult:
Updated design doc with CRD approach, and made the entire system mostly event-driven.

If the reliability of Informers is a concern (i.e. occasionally missing events), it can be augmented with a sync loop.

Also added a link to the original discussion of the distributed provisioner model.

contributors/design-proposals/local-pv-provisioner.md
verult: ACK, will update in a later iteration.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,156 @@
...
+* `LocalPVs` - A collection of PVs mounted in the local storage directory on the node. The discoverer is triggered to acquire this information.
+* `DeleteOperations` - A list keeping track of information related to each delete operation. This status is managed by the cleaner. It contains the following fields:
+  * PV - The PV to be deleted on disk.
+  * DeleteStatus - takes one of the following values:
+    * Requested
+    * InProgress
+    * Complete
+    * Error
+    
+There exists one ProvisionerStatus object per node.
```
msau42: Will this approach address the security issue of a compromised Pod on a node modifying another node's object?  You could modify another node's DeleteStatus, which could cause the PV to get deleted and recreated prematurely.  You could also modify another node's LocalPVs, to return fake volumes or remove some existing volumes.

Is there a way we can restrict a single Pod in a DaemonSet to have permissions to only modify its object based on the Node its running on?

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,156 @@
...
+| `Delete(PersistentVolume)`    | Starts the asynchronous delete of the given PV. |
+
+In addition, the Discoverer and Deleter from the previous design are kept here, but they only perform filesystem operations. All other logic is moved to the master.
+
+### MinionStatus
+The following fields are retrieved as part of the `GetMinionStatus()` call:
+
+| Field               | Description |
+| --- | --- |
+| `LocalPVs`          | A collection of PVs mounted in the local storage directory on the node. The discoverer is triggered to acquire this information. |
```
msau42: I think the minions should not need to be aware of the concept of "PVs".  I like the `type Volume` that @jsafrane suggested above.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,156 @@
...
+| `DeletesInProgress` | A collection of PVs currently being deleted asynchronously. This status is managed by the cleaner. |
```
msau42: Can we merge `DeletesInProgress` with the `Volume` information?  Maybe just have a `Status` field in the `Volume` with some values like: `Available, DeleteInProgress, DeleteError`, etc.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,156 @@
...
+1. Assume disk deletion is asynchronous, as required by block volumes.
+
+## Related Works
+* Original provisioner design, [here](https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/provisioner#design).
+* Disk cleaner (deleter) implementation for block volumes, [here](https://github.com/kubernetes-incubator/external-storage/pull/312).
+
+## Design 1: Master-Heavy
+
+### High Level Overview
+The static provisioner will be split into a single master (ideally placed on the master node) and many minions (one per node). The master contains the bulk of provisioner logic, including syncing PV objects with the API server and triggering discovery and deletion. Minions are left to perform the physical discovery and deletion of the local volumes on its node. gRPC is used to communicate between master and minions, with the master as the client and each minion as a separate server.
```
msau42: Mention that minions are run as a DaemonSet

contributors/design-proposals/local-pv-provisioner.md
msau42: In the master's initial sync with the minion, it can use the lack of a `Volume` to indicate deletion.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,156 @@
...
+Every minion must maintain the invariant that `DeletesInProgress` is correct even after a minion restart. In the case that a user manually deletes the PV currently being deleted by a minion, master can only depend on the MinionStatus to ensure a replacement PV isn't created.
+
+### MinionInfo
+Each MinionInfo entry contains the following fields:
+
+| RPC                   | Description |
+| --- | --- |
+| `Name`                | Name of the minion provisioner. |
+| `Endpoint`            | The minion's host address. |
+| `APICache`            | Cache of API PV objects managed by this minion. |
```
msau42: I prefer `PVCache` as the name

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,156 @@
...
+| `StopCh`              | Used to terminate a goroutine. |
+
+### Goroutine
+The goroutine will start by populating the `apiCache` using the populator logic that already exists in the previous design. Then, it will run the following loop with several seconds of delay in between executions:
+1. Call the RPC `GetMinionStatus()`.
+1. Using `LocalPVs` from the minion status, **reconcile** with the `APICache`. Create an additional PV if it exists in `LocalPVs` but not in `APICache` *and no delete is in progress for this PV, i.e. it's not in `DeletesInProgress`*.
+1. For each PV in `VolumeReleased` state (ready for cleanup), if it's not in `DeletesInProgress`, call the `Delete()` RPC.
+
+Master must delete the PV object after the disk delete operation is complete. In order for master to know when this occurs, the minion must persist a "complete" state, because in the wake of minion failure, master has to differentiate between an error and a completion. Master also needs to notify the minion that the PV is deleted, so the persisted state can be removed.
```
msau42: Can we reuse "Available" state to indicate that deletion is done?  Then the master does not need to notify the minion that the PV object is deleted.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,156 @@
...
+| --- | --- |
+| `LocalPVs`          | A collection of PVs mounted in the local storage directory on the node. The discoverer is triggered to acquire this information. |
+| `DeletesInProgress` | A collection of PVs currently being deleted asynchronously. This status is managed by the cleaner. |
+
+Every minion must maintain the invariant that `DeletesInProgress` is correct even after a minion restart. In the case that a user manually deletes the PV currently being deleted by a minion, master can only depend on the MinionStatus to ensure a replacement PV isn't created.
+
+### MinionInfo
+Each MinionInfo entry contains the following fields:
+
+| RPC                   | Description |
```
msau42: This is a field, not rpc

contributors/design-proposals/local-pv-provisioner.md
msau42: The minions run as Pods, so will have IPs different from the Node

contributors/design-proposals/local-pv-provisioner.md
verult: ACK

verult:
Addressed the ACK'd comments, leaving some parts out as implementation detail.

contributors/design-proposals/local-pv-provisioner.md
dhirajh: @msau42 One possible way security can be maintained here is by having the provisioner also store the digital signature of some critical fields (DeleteOperations?) of the CRD in the CRD itself. The digital signature can be created using a secret key that is generated (one time activity) by the provisioner and stored locally in the node. The provisioner can "watch" its own CRD object. If it ever takes on a value that does not match the signature, then it can set itself into a "compromised" state (perhaps by writing a local file) where-in it will no longer cleans (deletes) volumes.  The only way this state can be reset is when an  admin manually deletes the file.  This admin intervention should never be needed unless there was a real security compromise or if there was a bug in provisioner code.

verult: Had a conversation with @liggitt offline. One possible solution is to give each pod its own service account, with write access only to its provisioner object's status. This wouldn't work with a DaemonSet and requires each pod run as something like a Deployment.

Alternatively we can have a hybrid approach with both CRD and gRPC: Only the master can write to CRDs, and minion provisioner communicates with master through gRPC to provide status updates.

If we use gRPC, we'll need to have a notion of container identity + authentication. Otherwise a malicious pod can claim to be another provisioner reporting false LocalVolumes, and there's no way for master to differentiate between a real provisioner and an impostor.

@dhirajh A malicious pod can still disrupt the entire local provisioner system, making local storage unusable. Admin can set an alert on provisioner objects becoming "compromised", but ideally we should make it impossible to have any disruption.

liggitt: > If we use gRPC, we'll need to have a notion of container identity + authentication. Otherwise a malicious pod can claim to be another provisioner reporting false LocalVolumes, and there's no way for master to differentiate between a real provisioner and an impostor.

only if you go with a push model. if the master pulls from worker pods, it can take status from the worker on node *n* and update the correct object

dhirajh: The service account approach seemed interesting. I know that typically pods declare their service account in their manifest and then it gets used automatically in all their API requests. This is why I guess a daemonset cannot be used. Is there no way to allow a Pod to come up and then pick a service account after API access?

verult: If workers expose a Service in headless mode, then the Endpoints object is a reliable source of identification since it contains an IP address and its corresponding pod & node name.

It can be used for both push and pull models. For the push model, master extracts the source IP from the request and finds the matching node name in the Endpoints object.

verult: Do we need to assume IPs can be spoofed, or does Kubernetes already protect against that somehow?

contributors/design-proposals/local-pv-provisioner.md
verult: Revisiting gRPC designs since a hybrid approach (gRPC + CRD) looks more promising.

@jsafrane If the number of nodes is large, would there be a scalability problem with a large number of streams?

msau42: You may hit the default process limit of 1024 file descriptors.

msau42: In order to scale with # nodes, we will not be able to keep long running connections open.  We may need to open a new connection with every request, which is not great if requests are frequent.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,160 @@
...
+
+To ensure no malicious pods can affect provisioner execution, the CustomResourceDefinition and all pods must be in a trusted admin namespace.
+
+In the following description, we assume all Informer updates are cached and events buffered. On initialization, all caches are populated, and every Informer update event handler is executed once to ensure the entire provisioner system is in a consistent state.
+
+### ProvisionerStatus CRD
+The CRD contains the following fields (similar to design 1):
+
+* `LocalVolumes` - A collection of volumes mounted in the local storage directory on the node. The discoverer is triggered to acquire this information.
+* `DeleteOperations` - A list keeping track of information related to each delete operation. This status is managed by the cleaner. It contains the following fields:
```
jsafrane: I may be missing something obvious, but why we need to keep track of DeleteOperations at all? Provisioner/minion can get the signal to clean a local volume by watching PVs. Or is reading/watching all PVs on API server so dangerous?

Status of the cleanup is IMO internal state of the minion, do we need it to be exposed? The cleanup is implicitly "in progress" when PV is missing for a local volume. When it's complete, corresponding PV is created.

msau42: The problem is that the user can prematurely delete the PV object before it's done cleaning.  If the minion restarts, it loses the in memory state of the cleaning status.

However if we assume PV deletion issue is going to be fixed, then that would simplify this design and remove the need for a crd.

jsafrane: The minion can have some local storage to keep track of cleanups in progress.

verult: If the minion has to keep track of state anyway, one advantage of keeping state in the CRD is visibility to admins.

But in any case, giving minions write access to CRDs is no longer an option because of CRD isolation concerns, so each minion has to store local state anyway.

contributors/design-proposals/local-pv-provisioner.md
dhirajh: Do we need long running connections here? I think it makes the most sense for the minion to initiate the chatter by poking the server (as jsafrane suggested). It will make sure interactions are only driven by actual events and not regular polling. So the process could look like this:

1) Minion has some update to report, connects & requests a callback from master and disconnects.
2) This request gets queued up in the master for processing. 
3) Master calls back to get the information and updates CRD and disconnects.

The number of go routines pulling things off the work queue can be limited to any file descriptor limits. In general I believe it should be easy for a process to set the ulimit within a container to whatever fd limit it wants, so this limit could be high. But in general, I don't think we need long running connections for infrequent state changes in volume's cleaning and recycling status.

contributors/design-proposals/local-pv-provisioner.md
dhirajh: I actually think the minion need not maintain any state locally. If we follow the pattern mentioned above of pinging the master when a state needs to be updated (and having the master call back/authenticate the minon), then the CRD effectively acts as the persistent store for the minion. I certainly like the ability CRD provides for fetching status of cleaning for all disks via kubectl. The other option of logging into the host and poking around local files and making sense of state in them seems less appealing.

contributors/design-proposals/local-pv-provisioner.md
msau42: That's fine, we just can't use rpc streams then because that requires the connection to stay open.  If minion initiates the connection, then that brings up the issue of how to verify minion identity in case minion tries to spoof another node.

contributors/design-proposals/local-pv-provisioner.md
msau42: Can someone just query the PVs to get the deletion state?  We're going to log an event in the PV when deletion is started and if deletion failed.

contributors/design-proposals/local-pv-provisioner.md
verult: Is it correct to say even if a faulty minion spoofs an IP, if master sends a request to that IP, it'll still be routed to the legitimate minion?

@dhirajh In addition, minions should keep poking master until it receives a status request, to make sure master gets a status update. It can also be less aggressive and only retry if there was a network failure. And as a safety measure, master should still periodically poll for status, but very infrequently.

contributors/design-proposals/local-pv-provisioner.md
verult: @dhirajh CRD becomes a less reliable persistent store due to an unreliable network and possibility of master crashes. In this case we're probably OK as long as the delete request is persisted.

contributors/design-proposals/local-pv-provisioner.md
dhirajh: @verult @msau42 No. The above measures only safeguards against a random client with arbitrary network identity claiming to be the POD. If the network communication layer has been compromised to the extent that packets meant for the POD are re-routed to a fake POD then these measures won't work. To guard against such an attack, one would need minion authentication (just like PKI in Kubelets). Using any authentication mechanism which certifies each minion uniquely will probably require a non-DaemonSet approach as each POD will need unique secret. 

How about a different approach to this problem? 
### Job Based Approach
1. We still have a master and minions in each node.
2. The minions have one and only one role to play - They detect new disks and they poke the master.
3. The master acts (and trusts) purely through "Jobs" (https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/).

#### Disk Discovery:
When it gets pinged by a minion about new disk, the master launches a Job which runs on the node (by setting nodeSelector to that hostname). That job could have a secrets volume mounted to effectively authenticate itself with the master as it reports back the newly discovered disk via gRPC or REST. The lifespan of that secrets volume can even be limited to that job run if we want to be very secure.

#### Disk Cleaning/Deleting:
When the master concludes (based on PV state) that disk needs cleaning, it launches a cleaning Job on the node. This Job doesn't even need to report back to the master on state changes.

 - If the Job is still running => Cleaning is still in progress
- If the Job fails => Cleaning failed
- If the Job completes => Cleaning completed.

Note that the minion is not involved in the cleaning process. The master uses the status of the Job to determine if the PV recycling is completed or not. All status is maintained in Kubernetes resources (PVs and Jobs), which is highly desirable. Jobs could be named in a manner that makes it easy to identify which node and disk they are targeting (to improve administrative insight).

verult: The Job might not even need credentials. It could print the discovery result to STDOUT, and master can then grab its logs.

msau42: The Jobs approach looks promising, thanks so much for the idea!  If possible, can we try to eliminate grpc completely from the design?  It would make the design and implementation much simpler if we can stick to one consistent communication mechanism.

At the simplest, you could just use Jobs if the discovery was polling-based.  If we want to support event-based discovery, could a crd work?  If the only field is to trigger discovery, then the worst that could happen we run discovery jobs more frequently than necessary.

msau42: If the discovery is polling based, then you can eliminate the DaemonSet entirely.

dhirajh: As @verult suggested the master could read STDOUT for results (and avoid grpc). STDOUT tends to be brittle because any extraneous message can break parsing. However, given that these are simple jobs that will do one very specific thing and exit, the chances of this happening are very low and they will only happen due to genuine errors in the primary task being carried out.

Regarding polling, I shied away from polling using jobs as the cost at higher frequency was unclear to me. However, if its ok to poll at a low frequency (once every few hours, or day, configurable?) then its low impact. An admin could force polling (because they just made a change to the host) by checking a field - perhaps a label?. We won't even need a CRD for that, the label could be set on the node. The master anyway needs to "watch" all nodes to discover new nodes as they come online.  The protocol could be as follows:
The master sets the time it last polled a node in a node label.
- New nodes won't have that label, so they will get immediately polled. 
- The master uses the label value to decide when next to poll (has it been 2 hrs yet?).
- An Admin can force the polling on a node by deleting the label or setting it to zero.

msau42: I meant, even if you can read the discovery output from stdout, how does the master know when to launch the job in an event-based model?  I don't think we should modify the node object labels or annotations.  Only kubelet should.

Could stdout potentially get truncated if the output is too long?  I'm not familiar with how the logging mechanism works, but mixing logs with data may not be great.  Can we pass data with a secret?

ianchakeres: Using job logs is another nice potential avenue for simplifying the overall design. 

If the logs are used, we should leverage some type of structured logging (e.g. https://github.com/sirupsen/logrus and json). Instead of simple less structured log lines (e.g. traditional glog).

dhirajh: For achieving event based discovery, one of these two seem possible:

1. Discovery is run as a long running job or regular Pod explicitly by the master.  This Job/Pod can easily be supplied with unique serviceAccount information to update CRDs or use stdout or grpc, because the manifest for them is crafted by the master.  The downside is the need to manually start this Job/pod on every node and be aware of new nodes as they come online, deal with node failure, etc. This seems complicated to me as the master gets into the business of being a Pod controller.

2. Use the DaemonSet approach of deploying Pods. A DaemonSet takes a lot of complexity out  of the equation for the master regarding discovery of new nodes and other issues mentioned in option 1. To your point, yes, we could use CRD for communication here. Because its a DaemonSet, it would use the same serviceAccount on all Pods to update some "notification" CRD (instead of stdout or grpc/REST), but this is just for notification and as you pointed out the ill effect of a security compromise is a more frequent discovery request. Once the master sees the notification, it can launch a job, which could use a unique service account specified in its manifest to update the CRD (instead of stdout or grpc/REST with unique tokens).  The main question I have is how complicated it is to maintain unique service accounts for each node? Would that involve so much complexity that one of the other mechanisms seem more preferable?

msau42: Agree we don't want to over complicate this.  Adding/removing local disks from a node should be an uncommon occurrence.  One more thing to keep in mind: if it's uncommon enough, is it worth it to have a long running pod always running on every node?  That's more system overhead we are adding in order to handle a rare event.

verult: IMO using CRDs only as notifications is an abuse of its original intent. It's meant as a mechanism for a declarative API, not just a way to share state. Between gRPC/REST and CRD for notifications I would prefer gRPC.

dhirajh: @verult gRPC would be fine. One thing you might want to iron out is how a notification will be either persisted by the master or retried by the minion (if the master loses it). Keep in mind that the master will need to be capable of being rescheduled on another node if the node it is running on has failed, so it may not want to use local storage for persistence.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1105)

contributors/design-proposals/local-pv-provisioner.md
msau42: @jsafrane @childsb can you guys review this Jobs approach considering the issues you guys encountered with recycling?

verult:
Aggregated comments and new designs into the doc. Changed terminology from "minion" to "worker".

Amidst the security concerns, the Jobs approach (design 5) currently seems the most promising.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+* Master can only create and delete `DeleteOperation`s. It never modifies those objects.
+* Worker can only modify `DeleteStatus`. It can never create or delete `DeleteOperation`s.
+* Only a worker can write to `LocalVolumes`.
+
+### Worker Provisioner
+Create the ProvisionerStatus object on initialization.
+
+The Discoverer periodically probes the filesystem and updates ProvisionerStatus `LocalVolumes`. This can be further improved by using a filesystem watch.
+
+The Deleter uses ProvisionerStatus to keep track of states for each PV being deleted. Deletion is started when the ProvisionerStatus is updated (by the master) and some PVs are in the Requested state. Details will be included in the async volume cleaner design.
```
NickrenREN: Sorry, `some PVs are in the Requested state` means: `DeleteStatus` value is `Requested`, right ?

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+
+The master keeps track of `WorkerInfo`, a collection containing one entry for each worker keeping track of relevant information. The master also runs a group of goroutines, one for each worker, which controls and interacts with the workers.
+
+### Worker Provisioner
+The worker provisioner exposes the following RPCs:
+
+| RPC                           | Description |
+| --- | --- |
+| `GetWorkerName()`             | Returns the worker provisioner's name. |
+| `GetProvisionerStatus()`      | Returns the worker status, as described in the next section. |
```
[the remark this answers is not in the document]
verult: Yes, thanks for catching that!

contributors/design-proposals/local-pv-provisioner.md
verult: Yes

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+
+To trigger discovery, we have local volume directory watcher pod running in a DaemonSet. It keeps an in-memory cache of local volumes and periodically checks the volume directory for changes. If there is a change, it notifies master through either gRPC or a CRD notification object, and master can then spawn a discovery Job.
+
+To ensure the provisioner system has up-to-date state on initialization, the following needs to be done:
+* When the master initializes, it spawns discovery jobs on every node to make sure it knows about all local volumes in the cluster.
+* When a watcher initializes, it should notify master so a discovery Job can be spawned to fetch the latest list of local volumes.
+
+Master can be made completely event driven using Job informers.
+
```
tallclair: Here's a 6th design, which I'm hesitant to suggest because it's a huge hack, but maybe more performant / reliable than 5:

Basically, handle master to worker communication via the exec/attach/portforward API. This interface is handled directly by the kubelet via the CRI (kubelet <-> pod communication over a unix domain socket), and the kubelet <-> master (apiserver) communication with mutual TLS. There would be bunch of proxy overhead though, so I don't think you would want to keep the stream open always. If performance was an issue the master provisioner could authenticate to the kubelet directly, but it would be much simpler to just communicate via the apiserver.

tallclair:
Maybe we're overthinking this? Why isn't there a discussion of using just using a HostPort?

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+On Node deletion: delete the associated ProvisionerStatus object.
+
+**TODO**: Consider marking cache dirty when its corresponding API object is modified within master and ignoring Informer events if cache is dirty. This ensures master is always operating on the most up-to-date data. 
+
+## Design 4: Hybrid RPC-CRD Approach
+Keep the same CRD object described in Design 3, but only master has write access to them. Workers must persist its delete state locally and report its status to master through gRPC. Master makes requests to workers by writing to CRD, and fetches worker status by polling (similar to Design 1).
+
+Several modifications to Design 1:
+* For discovery, do not rely on worker to report its identity as it can be compromised. Identify workers using the Endpoints object (pod name and node name).
```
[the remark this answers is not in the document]
tallclair: Chatted with @thockin - something like this might be possible by providing a link-local kubelet address, which might have other uses as well. However, I've convinced myself that this specific use case (node attestation) is not a good one. (I.e. we should opt for something like mutual TLS or HostPort instead).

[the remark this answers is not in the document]
mikedanese: Ideally the pod identity doc has sufficiently trustworthy provenance and asserts sufficiently informative claims for a recipient to such an authorization decision. If pod proof of identity is not strong enough, we should invest in making it stronger instead of building other forms of authentication.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+
+### ProvisionerStatus
+The following fields are retrieved as part of the `GetWorkerStatus()` call:
+
+| Field               | Description |
+| --- | --- |
+| `LocalVolumes`          | A collection of PVs mounted in the local storage directory on the node. The discoverer is triggered to acquire this information. |
+| `DeletesInProgress` | A collection of PVs currently being deleted asynchronously. This status is managed by the cleaner. |
+
+Every worker must maintain the invariant that `DeletesInProgress` is correct even after a worker restart. In the case that a user manually deletes the PV currently being deleted by a worker, master can only depend on the WorkerStatus to ensure a replacement PV isn't created.
```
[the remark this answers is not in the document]
verult: A delete could potentially take hours or days. If multiple deletes are underway on different nodes, it seems like an overhead to keep many connections open for so long.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+1. For discovery: compare `LocalVolumes` with existing PVs in the API server. If a volume exists in `LocalVolumes` but has not been persisted, and if the disk is not about to be deleted (i.e. in Requested or InProgress state), then create the PV.
+
+On PV Informer update: if the updated PV is in VolumeReleased state, add the PV to `DeleteOperations` with the Requested state.
+
+On Node deletion: delete the associated ProvisionerStatus object.
+
+**TODO**: Consider marking cache dirty when its corresponding API object is modified within master and ignoring Informer events if cache is dirty. This ensures master is always operating on the most up-to-date data. 
+
+## Design 4: Hybrid RPC-CRD Approach
+Keep the same CRD object described in Design 3, but only master has write access to them. Workers must persist its delete state locally and report its status to master through gRPC. Master makes requests to workers by writing to CRD, and fetches worker status by polling (similar to Design 1).
```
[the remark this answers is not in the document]
verult: Yeah that's another way to look at it. The major difference here is the gRPC surface area is smaller, which can be considered as an implementation detail

tallclair:
I'm not sure whether I commented on master -> worker authentication, but that's definitely something you need to worry about as well (you don't want just anyone to be able to issue a request to delete a volume).

verult:
Using HostPort sounds like a reasonable approach. It can be applied to design 1, and worker discovery is simplified to listening for node changes. Also removes the security problems with design 1. One downside is users must be aware of potential port collisions.

In the current design doc, the most secure approach was Design 5 with Jobs. One of its benefits is workers don't need to persist state on the host. Its biggest drawback is log parsing.

Between these two designs I prefer design 1 + HostPort. Design 1 can be further simplified by the status notification mechanism in design 5 so that long polling is not necessary.

contributors/design-proposals/local-pv-provisioner.md
verult: Regarding securing CRDs: if we make the resource non-namespaced instead, wouldn't it be difficult enough to meddle with the resource since most users/service accounts shouldn't be allowed to access non-namespaced resources anyway?

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+One main downside of design 2 is the possibility of duplicate discover and delete operations, as a result of the cache not being updated immediately. Also, because the logic of PV updates is divided over the network, it's easy for PVs to end up in an inconsistent state.
+
+Design 3 has the following advantages:
+* Master has perfect knowledge of when to issue PV delete (based on the Complete DeleteStatus) without requiring multiple acknowledgment messages between master and worker.
+* Because the API server persists state reliably, we can simplify our reasoning with failure modes.
+* Provisioner discovery is no longer necessary.
+* Retries and backoffs are no longer necessary.
+* Provisioner state is more visible to admins.
+* Almost entirely event-driven (the only exception being the filesystem probe).
+But one major concern with this approach is each worker has access to all ProvisionerStatus objects. If a worker is compromised, it could disrupt the entire provisioner system instead of just its node.
```
mikedanese: Could an admission controller help mitigate this?

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,185 @@
...
+To ensure the provisioner system has up-to-date state on initialization, the following needs to be done:
+* When the master initializes, it spawns discovery jobs on every node to make sure it knows about all local volumes in the cluster.
+* When a watcher initializes, it should notify master so a discovery Job can be spawned to fetch the latest list of local volumes.
+
+Master can be made completely event driven using Job informers.
+
+## Conclusion
+One main downside of design 2 is the possibility of duplicate discover and delete operations, as a result of the cache not being updated immediately. Also, because the logic of PV updates is divided over the network, it's easy for PVs to end up in an inconsistent state.
+
+Design 3 has the following advantages:
```
mikedanese: One huge benefit not mentioned: bidirectional authentication to the apiserver is already solved.

mikedanese:
Why in most of these designs does the master establish connections to the nodes? This is inverted from what we usually do in kubernetes.

liggitt:
> Why in most of these designs does the master establish connections to the nodes? This is inverted from what we usually do in kubernetes.

This isn't piggy-backing on existing node authentication, so it doesn't have a great way to run a workload on every node with an identity distinct to that node.

dhirajh:
@mikedanese @liggitt I will say that the Jobs approach (design 5) leverages the existing trust between API server/controllers and kubelets to implement the provisioner functionality, so master and nodes need not attempt to establish identity through other means and the master also does not need to poll the agents.

verult:
Updated designs 1, 2, 4, and 5 to reflect the latest discussions of the design.

With the designs involving gRPC, it's possible to achieve the security requirements with TLS and client authentication, but it's quite involved to set them up today. Design 5 (the Jobs approach) avoids the security pains by communicating through the API server, among many other benefits. So I'd like to propose Design 5 as the final design.

PTAL

verult:
PTAL @msau42 @dhirajh @liggitt @tallclair

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Design 5: Worker as Jobs, Watcher in DaemonSet
+
+### Overview
+One of the major issues with previous approaches is worker identity - when a worker tries to communicate with master, it has to prove it's the worker it claims to be. To fix these issues, each worker must maintain different credentials.
+
+Alternatively, master can actively create worker Jobs to perform discovery and delete rather than having long-running worker pods in a DaemonSet sending data to the master. To trigger discovery, we also have local volume directory watcher pods running in a DaemonSet.
+
+### Worker
+Workers are spawned as Jobs by master on demand. In discovery mode, a worker writes the list of local volumes in JSON format to a file inside an emptyDir. A sidecar container then reads the file and writes to its STDOUT. In deletion mode, the worker executes async delete and uses the status of the Job as an indication of deletion state.
```
dhirajh: It looks like the implementation will be supporting gRPC. Is it not possible for the worker to report back the JSON via gRPC? The job could be launched with a secret provided to it (via environment variable or file mounting) and that secret could be used to prove the authenticity of the gRPC. The secrets could be "one time use", with the secrets being created and destroyed along with the creation and destruction of jobs.

Just something to consider if stdout parsing turns out to be flakey.

contributors/design-proposals/local-pv-provisioner.md
```diff
@@ -0,0 +1,204 @@
...
+If there is a network failure, the gRPC call will fail and the watcher retries with exponential backoff. If master fails after receiving the notification, once it comes back up it will spawn a discovery worker on every node (as explained in the Master section).
+
+### Master
+Master exposes the same RPC calls as Design 1 to enable discovery notifications.
+The master provisioner exposes the following RPC:
+| RPC                           | Description |
+| --- | --- |
+| `LocalVolumeReady()`             | Notifies master that there are new local volumes. Also called on watcher initialization. |
+
+Master can be made completely event driven using Job informers and watcher notification. Master deploys a discovery worker when it gets a notification from a watcher. When the Job is complete, master parses the output of the sidecar container as the list of discovered volumes. If the output is empty (may occur if container logs have been rotated) or invalid, it must re-deploy the worker. Master deploys a deletion worker when a PVC is released and uses the Job's status as deletion state. When the Job is complete, it removes the corresponding PV's finalizer and the PV resource itself.
```
dhirajh: I would suggest having a nomenclature for the worker Jobs. For cleanup it could be
<operation name>_nodename_pvname
For example: cleanup_minion1.acme.com_lv-1234
For discover: discover_minion1.acme.com

The advantage of this naming is that kubernetes will automatically provide some kind of single threaded behavior here. You will not be able to create a second job of the same name, until the first job has been deleted.  I see some value in this single threaded behavior for cleanup workers (having more than one active for the same PV is undesirable). For discovery, this is less important. The con is that one has to delete terminated jobs, not sure if there is anything undesirable about doing that.

contributors/design-proposals/local-pv-provisioner.md
msau42: According to the documentation, secrets can be read by any pod in the same namespace.

msau42: Summarizing a bit, this design is basically providing semantics of an API object that can only be updated by a single node.  We are basically implementing a specialized API server, doing our own marshaling and data transport.

How painful would this alternative be?  Turn the master into an operator that creates a namespace + CRD per node.  Would it be annoying for users to see one namespace per node when they list namespaces?  The namespaces could be labeled so at least they can be filtered out.  This alternative means implementing a specialized DaemonSet controller, but you get to communicate through the normal API methods.

verult: My mistake for not including this option in the doc. My concern is in the presence of master and network failures, the job has to stay alive and keep on retrying, until it receives an ACK from master. This adds some additional complexity, and we also lose the benefit of relying on Job status as task completion status because the job won't be Complete until master acks. In comparison, I think the stdout parsing approach is simpler.

msau42: You also mentioned the grpc alternative for parsing discovery data further down in the doc.  Another complication is you have to add authentication if you do that.

contributors/design-proposals/local-pv-provisioner.md
verult: If a job has an error, we might want to keep it around so the container logs aren't evicted

contributors/design-proposals/local-pv-provisioner.md
verult: @msau42 wouldn't each worker still need different credentials even if each CRD is in its own namespace?

msau42: Each worker would need its own namespace, and the rolebindings to modify the CRD in their namespace.

verult: Right, and a different service account for each rolebinding. Credentials injection would be painful

contributors/design-proposals/local-pv-provisioner.md
dhirajh: Ah OK, in that case I guess we need to keep the terminated jobs.

contributors/design-proposals/local-pv-provisioner.md
smarterclayton: Namespace per node isn’t going to scale very well on large dense clusters.

We have a long term direction for containers to give them unique identities that are node specific.  We will be doing that for daemonsets specifically to address “per node identity”.  The remaining problem is scoping a nodes actions to a specific subset of resources, which seems like something sig-auth should help with.

msau42: Yes, pod identity per node + restricting a subset of resources per pod is exactly what we need. 

But it sounds like we won't have solutions in the near future.  

So I guess the question is should we wait until those solutions exist, or proceed with the current proposal and potentially change again later

verult: Given that we'd like to have an implementation this quarter, I would proceed with the current proposal and change it later. I'll keep in mind while implementing that the design could potentially change.

contributors/design-proposals/local-pv-provisioner.md
verult: The nomenclature can be augmented to have an integer prefix that increments when error jobs exist. For example, if two jobs in error state exist named `cleanup_0_minion1.acme.com_lv-1234` and `cleanup_1_minion1.acme.com_lv-1234`, the next created job would be `cleanup_2_minion1.acme.com_lv-1234`. In the common case (no errors) we still get the single-threaded benefit.

contributors/design-proposals/local-pv-provisioner.md
msau42: I would like to avoid drastically changing the design now, and then to have to drastically change it again later.  So a design that can align well with future plans would be ideal, but it also depends on how far away "future" is.

If the future plan is to be able to have per node identity + restrict resource scope for DaemonSets, can we actually just keep our current DaemonSet implementation?  We would need to understand the details of these future features to evaluate if it can address our use case.

Assuming that the API scope issues are solved, do we still gain other benefits by moving to a master/worker model?

dhirajh: Just trying see if we see the same gains here:
1)  If node identity issues can be solved in the long term, then one could actually stay with the current daemonset approach. The only functionality that needs to move is the API server calls to some central proxy. The central proxy can now trust the integrity of the commands/messages originating from the daemonset pods.

2) If resource ACLs could be refined to limit access to specific nodes, then we wouldn't need central proxy at all.

Of the two, I think (1) would have the most impact on design as the daemonset implementation stays largely the same.  On the other hand (2) would be a nice to have addition which would eliminate a central proxy completely and move that functionality of API calls (small in terms of code change) back to the daemonset pods. I would not consider changes due to (2) coming into existence to be a drastic change.

dhirajh: I am also assuming above that (1) is something we know we can get much sooner than (2). I should probably clarify if thats true.

verult: One of the reasons why we leaned towards a master-worker approach rather than a proxy approach is because, in the future when we support dynamic provisioning, we want to have a single leader deciding which node to provision to. @msau42 is this still true, now that most of the PV node selection logic is in the scheduler?

@smarterclayton @mikedanese is there a timeline for per-node identity?

msau42: It's my AI this quarter to propose a design for scheduling and dynamic provisioning.  I'm leaning towards having the scheduler give direct input to the PVC on which node a volume should be provisioned on. 

@verult can you make a list of requirements/use-cases we would need from per-node identity + resource scope restriction features?

verult: @msau42 ACK

verult: @dhirajh regarding the proxy approach, another potential issue is it's not immune to MITM attacks as it stands. Every worker should have a unique encryption key and encrypt its messages to the proxy.

verult:
I added a section on specific security requirements [here](https://github.com/verult/community/blob/39f84fa319144d5856220a08eea540057798cbd6/contributors/design-proposals/local-pv-provisioner.md#security-requirements). As I understand it, per-node identity satisfies all of requirement (2) and parts of (1) and (3).

Are there existing solutions that satisfy some of these, or are there plans to develop solutions that fit these requirements?

verult:
@msau42, @tallclair and I had a chat offline. Once per-node identity for DaemonSets is achieved, in order to get resource restrictions, an admission controller can be added on top of the existing static provisioner solution to only accept PV requests if the PV's node label matches the identity (possibly in the form of a service account). Because admission controllers need to be compiled with API server, [external admission webhooks](https://kubernetes.io/docs/admin/extensible-admission-controllers/#external-admission-webhooks) can be used for the same purpose but as an out-of-tree option. This design has an extra benefit that users can choose to keep the provisioner lean if they don't require the security guarantees.

verult:
Sync'd up with the container identity WG regarding having per-pod identity with DaemonSets. Given that now there are many use cases for per-pod identity with various workloads (including this case), the WG will further evaluate this feature (https://github.com/kubernetes/kubernetes/issues/59670) and work toward a solution. Rather than implementing an approach to get around this limitation and drastically change the current design, I would prefer to wait for per-pod identity.

/close

PR #1112, opened 2017-09-21
Rename security-context-constraints.md to pod-security-policy.md

php-coder:
It's counter intuitive that proposal about PSP has name `security-context-constraints.md`.

@pweil- Do you have objections against renaming a file?

CC @simo5

### DIFF
--- a/contributors/design-proposals/auth/apparmor.md
+++ b/contributors/design-proposals/auth/apparmor.md
@@ -85,7 +85,7 @@ annotation. If a profile is specified, the Kubelet will verify that the node mee
 the container, and will not run the container if the profile cannot be applied. If the requirements
 are met, the container runtime will configure the appropriate options to apply the profile. Profile
 requirements and defaults can be specified on the
-[PodSecurityPolicy](security-context-constraints.md).
+[PodSecurityPolicy](pod-security-policy.md).
 
 ## Prerequisites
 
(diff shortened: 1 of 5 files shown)


php-coder:
This doc also has a link for Google Analytics that should be renamed too but I don't know which format it should have:
```
<!-- BEGIN MUNGE: GENERATED_ANALYTICS -->
[![Analytics](https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/docs/proposals/security-context-constraints.md?pixel)]()
<!-- END MUNGE: GENERATED_ANALYTICS -->
```

eparis:
Didn't we give up on GENERATED_ANALYTICS in the repos?  I know we did in core cube. I think all of the CI verifiers are gone. I think you can just delete it and everyone will be ok...

php-coder:
> Didn't we give up on GENERATED_ANALYTICS in the repos?

I don't know, probably we didn't for this repo:
```console
$ fgrep -rc 'BEGIN MUNGE' contributors/ | awk -F: '{sum+=$2}END{print sum}'
204
```

bgrant0607:
I don't think analytics matter for these docs, and we solve the problem a different way for kubernetes.io docs. I don't even know if anyone still has the credentials for the analytics account.

bgrant0607:
I think that renaming the doc is a good idea, btw.

php-coder:
PR has been updated to remove Google Analytics link.

PR #1119, opened 2017-09-22
cpu-manager: Fix static policy examples

sameo:
For both A and B pods to be part of the shared cpuset both cpu
limits should be lower than 1.

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>


sameo:
Or does the pod land in the shareable cpuset because its request is _not_ an integer ?

sameo:
This is specified:


> The "static" policy allocates exclusive CPUs for containers if they are included in a pod of "Guaranteed" QoS class and the container's resource limit for the CPU resource is an **integer** greater than or equal to one.

So this PR is invalid.

PR #1122, opened 2017-09-22
Update controller example code to Kubernetes 1.8

sttts:


contributors/devel/controllers.md
```diff
@@ -103,6 +103,30 @@ func (c *Controller) Run(threadiness int, stopCh chan struct{}) {
 		go wait.Until(c.runWorker, time.Second, stopCh)
 	}
 
+	// register event handlers to fill the queue with pod creations, updates and deletions
+	c.pods.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
liggitt: this should happen during construction, not in Run(). adding the event handler here (after start) would be ignored

jhorwit2: @liggitt it works after Run as been called. See [here](https://github.com/kubernetes/client-go/blob/db8228460e2de17f5d3a9a453f61dde0ba86545a/tools/cache/shared_informer.go#L318).

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1122)

spiffxp:
/retest
/lgtm
Just in time for 1.9

/hold
Is this still relevant given that we have https://github.com/kubernetes/sample-controller?  If you think so you're free to /hold cancel

jhorwit2:
@spiffxp That's an example for CRD operators. The docs in this PR are for controllers. I think it's still relevant. I use this example all the time when writing new non-crd controllers.

sttts:
> Is this still relevant given that we have https://github.com/kubernetes/sample-controller?

These are docs, sample-controller is an e2e example. Two purposes.

PR #1126, opened 2017-09-25
Updating the generator docs to match current functionality

mattfarina:
It appears the Makefile had changed at some point. This updates the docs to reflect the change.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -33,12 +33,15 @@ make all
 To build docs for one SIG, run these commands:
 
 ```bash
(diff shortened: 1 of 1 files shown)


cblecker:
Hey @mattfarina.. thanks for pointing this out. That's confusing an unintended. I've put in a fix here (https://github.com/kubernetes/community/pull/1127) that should fix up the functionality, and the docs too.

mattfarina:
I prefer the fix in #1127. So, closing.

PR #1127, opened 2017-09-25
Update generator to use WHAT variable

cblecker:
Having to set two different variables is confusing. Changed this to look at the `WHAT` envvar and match on a suffix.

Fixes #1126.

cc: @jamiehannaford


mattfarina:
I like this approach. lgtm

mattfarina:
Note, I did manually test this and verified that it works for me.

cblecker:
Self-LGTM based on comments :)

PR #1116, opened 2017-09-21
Fix broken links after moving proposals to subdirs

php-coder:
Fix broken links after kubernetes/community#1010
Because I fixed them semi-automatically it's possible that patch has some mistakes (but I hope there is none).

Fixes #1109

CC @grodrigues3 @bgrant0607 @simo5

Issue #1109, opened 2017-09-21
Design proposals have lots of broken links to other design proposals


mindprince:
https://github.com/kubernetes/community/pull/1010 moved design proposals to SIG specific directories but didn't update the internal links from one design proposal to another. So, all proposals that refer to other proposals have broken links now which makes them harder to read.

/cc @grodrigues3 @bgrant0607

bgrant0607:
@mindprince Could you please send a PR to fix them?

php-coder:
I'll fix them.

xiangpengzhao:
@php-coder awesome! @zhangxiaoyu-zidif had similar idea in https://github.com/kubernetes/kubernetes.github.io/pull/5530#issuecomment-331099373

Can the tools you used in your PR automatically check the broken links referenced in other repo?

php-coder:
> Can the tools you used in your PR automatically check the broken links referenced in other repo?

I didn't use any tools this time. It was one-liners on shell :)

xiangpengzhao:
@php-coder thanks! If you try and find some tools workable, feel free to ping @zhangxiaoyu-zidif :)

duglin:
sig svc-cat uses this:  https://github.com/duglin/vlinker

PR #1010, opened 2017-08-31
Organize Design Proposals Dir by SIG

grodrigues3:
Organize design proposals directory into sigs.  This will make it easier to assign reviewers for future proposals.

Fixes #918

PR #1128, opened 2017-09-25
Fix links after design proposals move

nikhita:
Some links broke when design proposals were organized in https://github.com/kubernetes/community/pull/1010. They were mostly fixed in https://github.com/kubernetes/community/pull/1116 but a few seem to still be pointing to the old ones.

Takes another stab at https://github.com/kubernetes/community/issues/1109.

/cc @grodrigues3

### DIFF
--- a/contributors/design-proposals/api-machinery/apiserver-build-in-admission-plugins.md
+++ b/contributors/design-proposals/api-machinery/apiserver-build-in-admission-plugins.md
@@ -4,7 +4,7 @@
 
 | Topic | Link |
 | ----- | ---- |
-| Admission Control | https://github.com/kubernetes/community/blob/master/contributors/design-proposals/admission_control.md |
+| Admission Control | https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/admission_control.md |
 
 ## Introduction
 
--- a/contributors/design-proposals/api-machinery/dynamic-admission-control-configuration.md
+++ b/contributors/design-proposals/api-machinery/dynamic-admission-control-configuration.md
@@ -2,13 +2,13 @@
 ## Background
 
 The extensible admission control
-[proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/admission_control_extension.md)
+[proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/admission_control_extension.md)
 proposed making admission control extensible. In the proposal, the `initializer
 admission controller` and the `generic webhook admission controller` are the two
 controllers that set default initializers and external admission hooks for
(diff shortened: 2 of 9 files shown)


nikhita:
I am not sure about this but noticed that some proposals ([example](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/kubelet-hypercontainer-runtime.md)) have the Anayltics link as:

```
<!-- BEGIN MUNGE: GENERATED_ANALYTICS -->
[![Analytics](https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/docs/proposals/kubelet-hypercontainer-runtime.md?pixel)]()
<!-- END MUNGE: GENERATED_ANALYTICS -->
```

Should this link also be updated?

spiffxp:
AFAICT, the code that generated that munge has been deleted.  I suspect we might even be able to get rid of the links entirely?

Does someone from @kubernetes/sig-docs-maintainers know the state of our google analytics usage? I'm not sure it even applies to .md files in github repos.  A glance through our kubernetes/* and kubernetes-incubator/* shows many outdated links (eg: the code-of-conduct.md links)

zacharysarah:
@spiffxp 👋 I can't speak to site analytics _per se_, and it may be unrelated to the matter at hand, but here's what the search console shows for 404s from links with `*.md` in the past 30 days:

screen shot 2017-10-05 at 3 12 26 pm

steveperry-53:
I'm checking into this. I'll reply later this afternoon.

steveperry-53:
Signed in as stevepe@google.com, at https://analytics.google.com, I can see
the Kubernetes Google Analytics account. I have limited access (Read and
Analyze).

Nate Kartchner, at the Linux Foundation, is a manager for the Kubernetes GA
account. If you would like to be a member of the Kubernetes Google
Analytics account, send mail directly to nkartchner@linuxfoundation.org.

Account: Kubernetes

Property: Kubernetes

Views:

   PROD: Kuberbetes Github

   PROD: Kubernetes.io (This is the one we use for the doc set at
kubernetes.io.)

   PROD: RAW: Kubernetes


Here are my notes from the past, but they don't seem to be true any more.
So at this point I don't understand what we get from PROD: Kubernetes
Github and PROD: RAW: Kubernetes.

PROD: RAW: Kubernetes has no filters. It counts both internal and external
traffic. It counts visits to both kubernetes.io and github.com/kubernetes.

PROD: Kubernetes.io filters out visits from internal IP addresses. And it
counts only visits to kubernetes.io.

PROD: Kubernetes Github filters out visits from internal IP addresses. And
it counts only visits to github.com/kubernetes

mindprince:
/lgtm

The current changes look good. We can fix the analytics links in another PR.

PR #1129, opened 2017-09-26
The ../user-guide links do not seem to work anymore.

adelton:


castrojo:
I see PRs and issues being filed in the docs repo wrt. redirects and broken links recently, so asking for some guidance here: should we be direct linking to the canonical final URL with the domain and everything or is there another preferred method?

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1129)

heckj:
There is a significant amount of work going on as part of the transition to refactor the documentation to make it more accessible. The `/docs/user-guide` and `/docs/admin` structure are still there as a part of this migration, but effectively deprecated as the documentation has been transforming to a base organization layered on one of 5 groups: `setup`, `tasks`, `concepts`, `tutorials`, and `reference`. As docs are updated, the redirects on the site are also being updated to preserve links but there can still be misses. There are also monthly reviews of 404 reports to try and resolve and bolster these redirects to catch any misses.

I'm hardly the only voice, but I'd recommend direct linking, especially from markdown that's now in the community space (or might be migrating about) rather than relying on relative links to nearby directories that may not stay in the same repository. This also allows for flexibilty on publishing as we are finishing the transition out from github pages to hosting under netlify.

/cc zacharysarah

adelton:
Rebased on master -> ea22ffe05c208d68a9b133b52290a5284c9370c9.

PR #1132, opened 2017-09-26
Providing external secrets to Pods in Kubernetes.

adelton:


adelton:
I was noted offline that from the document it's not clear that

   Different sets of container will be able to mount different flexVolume types which can use different credentials or configuration for different vault. And each pod will get separate FUSE process. Or rather, each FUSE-d flexVolume volume in that pod will get one FUSE process.

enj:
@kubernetes/sig-auth-proposals @kubernetes/sig-node-proposals @kubernetes/sig-storage-proposals

contributors/design-proposals/auth/external-secrets.md
```diff
@@ -0,0 +1,218 @@
+# Providing external secrets to Pods in Kubernetes
+
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: openshift/sig-security
```
smarterclayton: Should be sig-auth

contributors/design-proposals/auth/external-secrets.md
```diff
@@ -0,0 +1,218 @@
...
+For that reason we are looking for mechanism of exposing the secrets
+from external sources to Pods and their containers via filesystem
+interface. With that approach, the individual secrets would be seen as
+files in filesystem structure, while backed by active mechanisms
+of external secrets providers or vaults, and would only request the
+secret at the time it is actually accessed by the container.
+
+## Proposal
+
+We are proposing flexVolume plugin which would utilize FUSE. FUSE
```
smarterclayton: You need to establish why someone consuming this proposal cares about FUSE (which is an implementation detail).  Someone could just as easily create a flex volume on tmpfs and inject it into the container - you need to articulate why FUSE is special here.

contributors/design-proposals/auth/external-secrets.md
```diff
@@ -0,0 +1,218 @@
...
+}
+```
+
+are passed to the flexVolume plugin, so the flexVolume plugin can
+parse the parameters and get the `kubernetes.io/pod.uid` value
+and pass it to the FUSE filesystem via `mount` option. That provides
+the FUSE filesystem with the identity of the clients that will be
+accessing the secrets which can be used for access authorization.
+
+The FUSE filesystem runs as root so it can have access to all needed
```
smarterclayton: The container identity WG is currently proposing several options, including a private key created on the pods behalf that it may or may not have access to.  It would be good to frame this proposal in terms of the container identity WG stated goals and how this offers a near term mechanism to expose external secrets that is not dependent on the container identity.

contributors/design-proposals/auth/external-secrets.md
```diff
@@ -0,0 +1,218 @@
...
+checking that the on-Node identity which makes the request for
+secrets is only making the request on behalf of the Pods that are
+currently running on the Node. That way, if a Node is compromised,
+only (secrets of) Pods running on that single Node are compromised.
+The Custodia server on that separate host will then proxy the requests
+to the actual vault or secrets provider.
+
+## Implementation
+
+Implementation of FUSE filesystem using libfuse 2.9 will show example
```
smarterclayton: This is pretty low level given the proposal above - would expect you to mention that you have a particular implementation proposed, but the mechanism whereby you are implementing this allows other approaches (and list and describe them here, including drawbacks / advantages).

contributors/design-proposals/auth/external-secrets.md
```diff
@@ -0,0 +1,218 @@
...
+The SELinux boolean `virt_sandbox_use_fusefs` has to be set to true to
+allow Kubernetes containers to access the filesystem.
+
+### Client/Server Backwards/Forwards compatibility
+
+Not affected.
+
+## Alternatives considered
+
+Container Storage Interface approach might be the long-run preferred
```
smarterclayton: If this proposal is "flex volume can be used to inject secrets" (which is good to document) there needs to be more discussion of different ways and tradeoffs that flex volumes can be used (you talk a bit about a daemon socket, but don't discuss injecting static tmpfs volumes or other solutions in use in the space).  You want to make the case that flex volume *should* be used for this use case until a later point which may include a new storage mechanism.  I would also call out that evolution of this use case should inform the CSI group and so that there would be a request for sig-storage to review the proposal and the case you are making so that they can understand and comment as to the direction of CSI.

adelton:
A more generic proposal with full example of FUSE flexVolume is now in https://github.com/kubernetes/community/pull/1147.

contributors/design-proposals/auth/external-secrets.md
adelton: Would the motivation/proposal in https://github.com/kubernetes/community/pull/1147 be the wording that we might be looking for?

contributors/design-proposals/auth/external-secrets.md
adelton: Do you have links to the latest proposals so that we could reference it here?

I'm not sure I understand the "external secrets that is not dependent on the container identity" -- is this "not dependent on container identity as opposed to pod identity", or "not dependent on container nor pod identity"? My understanding here is that the information / secrets that are made available are dependent on the pod identity.

contributors/design-proposals/auth/external-secrets.md
adelton: Do you mean different versions of `libfuse` and different FUSE client modules for different languages?

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1132)

adelton:
@smarterclayton, based on offline discussion with Simo I've got some questions about assumptions we can / cannot make.

In the FUSE code, can we assume that the FUSE mountpoint is only used and accessed by the Pod for which the filesystem was created, not shared among Pods?

Or should we recommend (and perhaps in the example in https://github.com/kubernetes/community/pull/1147 demonstrate) doing some sort of validation that the request to `getattr`/`open`/`read` comes from a process which is running in a container matching the Pod whose `pod.uid` we got during the `mount` operation?

smarterclayton:
I think in practice that's a property of the flex volume.  It's probably
better for a flex volume daemon like this to be more restrictive rather
than less when dealing with an API exposed into the pod.  If you're just
mounting a filesystem, there are less (although not zero) attack angles for
a pod to escape its mount namespace so I don't know that I would be as
concerned.

adelton:
> I think in practice that's a property of the flex volume.

I'm sorry but to avoid reading it incorrectly -- the property here being that the FUSE mountpoint is only used by a single Pod, the correct one? Or the contrary, that the mountpoint may become accessible by other Pods?

adelton:
Presumably, I cannot add the 'keep-open' label, so hopefully this comment will do the trick.

simo5:
@adelton the flexvolume should guarntee only that pod can access the volume as it is only mounted in the pods namespace, however it is safer to *not* make this assumption in the FUSE driver itself. This way we have defense in depth. The fuse driver would be more robust if it were able to track pod access and refuse access to anything else (I would prevent root from accessing it if it were not doing so with a process that is part of the pod namespace for example).

adelton:
Understood. So now the question is, from the FUSE driver during mount time, how can we obtain (via call to Kubernetes API?) the SELinux context of the (processes in the) pod, to then verify it against the context of the process making the `open`/`getattr`/`read` calls. Shoud `http://localhost:8080/api/v1/namespaces/<namespace>/pods/<pod-name>` give that information?

adelton:
I have now added the complete implementation of the FUSE filesystem which invokes external helper programs to get the secrets. The setup was tested with HashiCorp's `vault read` and Custodia's `custodia-cli get` operations.

Rebased on master.

->1c62e3683f6d1dbfdf325cb72dabc82d321882b0

adelton:
The latest work is an extension of the `pod-info-fuse` example from https://github.com/kubernetes/community/pull/1147.

adelton:
/assign @dchen1107

per bot's recommendation.

adelton:
I've copied the content of this pull request to https://github.com/adelton/kubernetes-flexvolume-fuse for easier consumption. New README was also added there.

PR #1137, opened 2017-09-28
start extensible admission to beta

deads2k:
@liggitt @ironcladlou @sttts @smarterclayton initial comments please.


sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,362 @@
...
+it less error-prone, easier to configure, and more flexible for admission plugin
+developers.
+
+## Requirements for extensible admission
+Ordering the admission chain is critical.  One mutating admission plugin can invalidate
+the results of other admission plugins.  For instance, the `LimitVerifier` would
+invalidate results for the `ResourceQuota` plugin.  More generally, object mutations
+must happen before object validations.  In addition, compiled validating admission plugins
+must run after mutating admission webhooks.  Splitting the chain into mutators and validators
+will resolve the coarse grained ordering problem and would require two different webhook APIs
```
liggitt: > would require two different webhook APIs

The differences don't seem significant enough to warrant two separate resources. I'd expect mutating webhooks to indicate that in their spec, and the portions of the status related to mutation simply be required to be empty in responses from non-mutating webhooks.

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,362 @@
...
+since the `Status` would be different for both now and we can reasonably expect different
+evolutions.  It would also force changes in the registration APIs since ordering matters
+for individual mutators, but ordering does not matter for validators.  Having fewer items to
```
liggitt: > ordering does not matter for validators

it could matter for remote systems, if the webhook is non-mutating from kube's perspective of the object being admitted, but has side effects outside of kube (like quota)

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,362 @@
...
+must run after mutating admission webhooks.  Splitting the chain into mutators and validators
+will resolve the coarse grained ordering problem and would require two different webhook APIs
+since the `Status` would be different for both now and we can reasonably expect different
+evolutions.  It would also force changes in the registration APIs since ordering matters
+for individual mutators, but ordering does not matter for validators.  Having fewer items to 
+order makes the registration API easier to reason about for maintainability.
+
+Initializers as implemented today attempt to push new API validation requirements into the 
+strategy and new requirements onto the admission plugins handling updates and creates.  The
+former is unecessary and the latter doesn't work.
```
liggitt: typo: unecessary

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,362 @@
...
+
+API validation on uninitialized objects can reasonably follow the rules for validation on create.
```
liggitt: modulo `ValidateObjectMetaUpdate`

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,362 @@
...
+Since the object hasn't been "ready", there is no reason to add the validation rules required on
+updates since restrictions like field immutability don't logically apply and would restrict the power
+of admission extensions the point of unusability.  Using the results for creation validation on
```
liggitt: `to the point`

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,362 @@
...
+uninitialized objects and objects transitioning from uninitialized to initialized lets the
+strategy author focus on their API needs.
+
+Admission validation is also needed on resources being initialized.  During initialization, there
+may be some admission validation required, but it isn't normal API update admission.  The update
+admission mutators shouldn't run and since the subject (user.Info) on the admission attributes is
+NOT the original user, normal update admission mutators and validators are likely to make improper
+choices.  When the object transitions from uninitialized to initialized state, the original object
+creator should be artificially injected into the admission attributes and the create admission 
+validators should run.  This allows mechanisms like PSP to function properly during their validation
```
liggitt: > When the object transitions from uninitialized to initialized state, the original object creator should be artificially injected into the admission attributes and the create admission validators should run.

I can see use cases both ways. Being able to write an initializer that injects a privileged container

sig-api-machinery/extensible-admission-to-beta.md
deads2k: > The differences don't seem significant enough to warrant two separate resources. I'd expect mutating webhooks to indicate that in their spec, and the portions of the status related to mutation simply be required to be empty in responses from non-mutating webhooks.

I really like different things to be different.  You just described different validation requirements.  They are definitely similar, but the semantics are definitely different.

sig-api-machinery/extensible-admission-to-beta.md
deads2k: > it could matter for remote systems, if the webhook is non-mutating from kube's perspective of the object being admitted, but has side effects outside of kube (like quota)

Providing order on validators would have significant performance implications.  Even in your quota example, it's not important for quota to be ordered.  It's a performance optimization if something validates "no".  Quota (and things like it), still have to have reconciliation loops since failures could happen down in storage.

liggitt: > Providing order on validators would have significant performance implications

Not providing order could also have significant performance implications, if the external check has an expensive side effect.

> Even in your quota example, it's not important for quota to be ordered. It's a performance optimization if something validates "no".

Performance optimizations can be important.

bgrant0607:
cc @cheftako

sig-api-machinery/extensible-admission-to-beta.md
deads2k: > modulo ValidateObjectMetaUpdate

agreed

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,363 @@
...
+4. connecter.Connect().ServeHTTP()
+
+#### Proposed behavior
+
+1. connecter.NewConnectOptions
+2. decode request connect options
+1. admit.Validate
+2. connecter.Connect().ServeHTTP()
+
+## Ability to run out-of-tree plugins
```
enisoc: Are you talking about a separate plugin mechanism here, or do you mean this will plug in through admission webhooks?

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,363 @@
...
+unreachable ("fail closed")? I think this is what we must do, for the reason
+stated above. Assuming this is what we choose, how do we restore the cluster to
+a healthy state, as the default "fail closed" behavior won't allow any new pods,
+and we need at least one new openshift-apiserver pod to handle the external
+admission plugin invocations. We need some way to a) indicate that a certain set
+of changes to the system are being made to restore the external admission
+plugin, and b) ensure that users can't "sneak in" anything while admission isn't
+fully operational. Some possible options include:
+
+- Specify via flag/configuration the identity of a user or group whose requests
```
enisoc: What if we register the namespace that the admission plugin lives in, and say that any given plugin does not apply to objects in its own namespace? Then you use regular RBAC, e.g., to control who has access to create things in that namespace.

sig-api-machinery/extensible-admission-to-beta.md
deads2k: uh, old doc.  Let me scrub that section out

deads2k:
@enisoc Sorry, I pulled in an old doc that described some of the updates to chain to illustrate how things could be done.  it had some sections that no longer apply.  I updated the pull to remove them.

deads2k:
@kubernetes/sig-api-machinery-api-reviews 
@ncdc lifted some sections from https://github.com/kubernetes/kubernetes/pull/34131.  Didn't figure you'd mind.

ncdc:
@deads2k nope, thanks

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,363 @@
...
+  1. e.Storage.Create
+  1. e.AfterCreate
+  1. e.Decorator
+
+#### Proposed behavior
+
+This is the proposed new behavior:
+
+1. decode request
+1. admit.ApplyMutations
```
caesarxuchao: At this point the object will be of the internal version. We should send the external version objects to the webhooks. I think it should be required for beta, otherwise webhooks will be hard to maintain.

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,363 @@
...
+may be some admission validation required, but it isn't normal API update admission.  The update
+admission mutators shouldn't run and since the subject (user.Info) on the admission attributes is
+NOT the original user, normal update admission mutators and validators are likely to make improper
+choices.  When the object transitions from uninitialized to initialized state, the original object
+creator should be artificially injected into the admission attributes and the create admission 
+validators should run.  This allows mechanisms like PSP to function properly during their validation
+phase.  Such injection requires persisting the original user.Info (probably encrypted) into the 
+object.
+
+## How much is needed for beta
```
caesarxuchao: API server is sending internal version objects to the webhooks.  I think it's required for beta to send external versions of the objects, otherwise webhooks will be hard to maintain.

liggitt: > I think it's required for beta to send external versions of the objects

no question, that's absolutely required

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,363 @@
...
+
+Admission validation is also needed on resources being initialized.  During initialization, there
+may be some admission validation required, but it isn't normal API update admission.  The update
+admission mutators shouldn't run and since the subject (user.Info) on the admission attributes is
+NOT the original user, normal update admission mutators and validators are likely to make improper
+choices.  When the object transitions from uninitialized to initialized state, the original object
+creator should be artificially injected into the admission attributes and the create admission 
+validators should run.  This allows mechanisms like PSP to function properly during their validation
+phase.  Such injection requires persisting the original user.Info (probably encrypted) into the 
+object.
```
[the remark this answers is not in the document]
liggitt: > How does PSP handle normal UPDATE initiated by controllers? Does it have the same problem?

Normal updates don't get to mutate the important bits of the pod spec.

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,363 @@
...
+strategy and new requirements onto the admission plugins handling updates and creates.  The
+former is unnecessary and the latter doesn't work.
+
+API validation on uninitialized objects can reasonably follow the rules for validation on create
+plus `ValidateObjectMetaUpdate` (to prevent name and UID changes and the like).
+Since the object hasn't been "ready", there is no reason to add the validation rules required on
+updates since restrictions like field immutability don't logically apply and would restrict the power
+of admission extensions to the point of unusability.  Using the results for creation validation on 
+uninitialized objects and objects transitioning from uninitialized to initialized lets the
+strategy author focus on their API needs.
```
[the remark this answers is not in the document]
liggitt: > PodStatus shouldn't be updated when the pod is not initialized

that's not enforced by validation... that's done in PrepareForUpdate

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
caesarxuchao: > Normal updates don't get to mutate the important bits of the pod spec.

Yes, I understand that. Let me ask the question more specifically: the PSP AC [handles](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/security/podsecuritypolicy/admission.go#L91) UPDATE. And in the beginning of its `Admit()` method, it [finds](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/security/podsecuritypolicy/admission.go#L134) the matched psp based on the user info. If a controller updates the object, does that mean the PSP controller will find the policy for the controller rather than for the original creator?

Also, assuming an initializer adds a side-car container to the pod, should that container subject to the original creator's psp, or the initializer's psp? I think it should be the later.

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
caesarxuchao: Do you mean disallow any UPDATE to subresources when the object is not initialized? That sounds right.

@deads2k it sounds like you don't want to pursue the "uninitialized objects can be invalid" argument anymore?

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
liggitt: I take that back, the image is able to be updated, which is why PSP cares about update at all. Also, I think there are some pod annotations being used for features like App armor that are mutable by virtue of being  annotations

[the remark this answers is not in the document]
deads2k: > Are there examples other than the PSP?

The fact that we needed it so quickly indicates that the use-case has a lot of value.


> Also, assuming an initializer adds a side-car container to the pod, should that container subject to the original creator's psp, or the initializer's psp? I think it should be the later.

Post-validation mutation (which is what this is describing) would be possible to describe, but it seems really risky.  If it were expressed in our current admission chain, the second mutation (adding a side-car container than the requestor couldn't request), would not be validated at all.

I don't think it would be a good idea to add that behavior in a rush.  Do we have any current admission plugins in our chain trying to do this?  I don't know of any.

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
deads2k: > @deads2k it sounds like you don't want to pursue the "uninitialized objects can be invalid" argument anymore?

Oh, I'd still like that and I still think it would be correct to allow it, but that decision isn't required to make the API beta

sig-api-machinery/extensible-admission-to-beta.md
erictune: Clarify: Are you loosely describing initializers as "webhooks", or are you introducing a new extension mechanism, which is a webhook that can also mutate?

deads2k: > Clarify: Are you loosely describing initializers as "webhooks", or are you introducing a new extension mechanism, which is a webhook that can also mutate?

Webhooks already exist.  They are one of the admission features that we're trying to move to beta.  Right now they aren't mutating, but to have enough expressablity to support our current chain, they need to add it.

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
caesarxuchao: Not any in-tree admission plugins. But out-of-tree, adding a side-car container is a common use case of initializers, I know a couple of google projects that use initializers this way.

That aside, I don't think recording the original user info is required. Pods are created in two ways: 1. by controllers (e.g., by the replicaset controller) 2. manually by end-user.
For case 1, like what's suggested in the PSP [doc](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#working-with-rbac), PSP should be bound to service account presented in the podTemplate. The original creator's identity is not important. Also, users shouldn't give initializers superuser permissions, like they shouldn't grant such permissions to controllers.
For case 2, I think first we don't encourage user to create pod manually, and second users can always use the service account to workaround.

cc @tallclair

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
caesarxuchao: @deads2k, agree it's not required for beta. One way to allow "invalid" uninitialized objects is introducing the `ValidateUninitialized` method, so that strategy author can further relax the validation. Does that sound reasonable?

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
tallclair: > For case 2, I think first we don't encourage user to create pod manually, and second users can always use the service account to workaround.

Can we just deprecate this approach? I think it makes the PSP usage harder to reason about, and in production users shouldn't really be creating pods directly (outside of debugging, maybe). In the long term, maybe we can come up with a mechanism to attribute pod creation & update back to the original user (although I'm not entirely sure that's desireable).

> Also, I think there are some pod annotations being used for features like App armor that are mutable by virtue of being annotations

Yeah, although we're trying to move away from that.

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,292 @@
...
+NOT the original user, normal update admission mutators and validators are likely to make improper
+choices.  When the object transitions from uninitialized to initialized state, the original object
+creator should be artificially injected into the admission attributes and the create admission 
+validators should run.  This allows mechanisms like PSP to function properly during their validation
+phase.  Such injection requires persisting the original user.Info (probably encrypted) into the 
+object.
+
+## How much is needed for beta
+Basically all of it.  After considering the needs of initializers, the admission chain needs to
+be split into mutators and validators.  That has a direct impact on the webhook admission registration
```
tallclair: Would mutators also be allowed to validate, or would it be completely separate? I'm an advocate for breaking out the defaulting (mutation) from the PSP, but as it currently functions it's incompatible with this split.

deads2k: > Would mutators also be allowed to validate, or would it be completely separate?

If it isn't completely separate, then such a plugin cannot observe the final object.  Other mutators in the direct chain and all initializers will happen after it has validated.

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
deads2k: Knowing the original user is a key piece of admission used by PSP, NodeRestriction, GC permission checks, and eventratelimiting.  Eliminating the power for dynamic admission hooks makes it impossible to build important classes of admission plugins.  The workaround of SAs only works for one of those checks.

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
deads2k: > @deads2k, agree it's not required for beta. One way to allow "invalid" uninitialized objects is introducing the ValidateUninitialized method, so that strategy author can further relax the validation. Does that sound reasonable?

As long as the normal flow runs the standard create validation (not update validation), then I'm ok having an opt-in method.  The vast majority of resources won't care about specific initialization logic if run the create validation for them instead of the update validation.

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
caesarxuchao: I think recording the original creator will be useful. Actually its usefulness goes beyond the dynamic admission. I'm trying to argue that it's not an urgent goal for beta, because dynamic admission mechanisms can support existing plugins without this feature.

My assumption regarding when the validation webhooks are run:
* Currently they are exercised to validate the original CREATE, and the intermediate initialization UPDATEs.
* In the future, for performance sake, webhooks can skip the intermediate initialization UPDATEs. They are exercised to validate the original CREATE, and the last UPDATE that completes the initialization.

If so, then for the admission plugins in question:
* NodeRestriction: it only handles mirror pods, which are exempted from initialization, so no initializer will mutate the pods that NodeRestrction cares about.

* GC permission checks: the plugin cares about if the modifier of the ownerReference has "delete" permission of the dependent object, or the "finalize" permission of the owner object. If an initializer modifies the ownerReference, I think it's reasonable to check if the initializer controller has the permissions, rather than checking the permissions of the original creator.

* eventratelimiting: the original CREATE will be validated, and will be rejected if the user creates too many events. Further, I think we should ban initializing events, otherwise there will be performance issue given the huge number of events.

* PSP: should use ServiceAccount instead ([details](https://github.com/kubernetes/community/pull/1137#discussion_r142241079))

To make these plugins dynamic, they can be adapted as webhooks. The apiserver sends the `User.Info()` to the webhooks.

Taking a step back, recording the original user is backwards compatible, we can add that after beta.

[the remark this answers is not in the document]
deads2k: > I think recording the original creator will be useful. Actually its usefulness goes beyond the dynamic admission. I'm trying to argue that it's not an urgent goal for beta, because dynamic admission mechanisms can support existing plugins without this feature.

If you change PSP you will break use-cases like an infrastructure component that wants to escalate a user by using a privileged identity for a specifically built pod (we know this use-case exists since we use it openshift).  

Attempting to run the current chain with PSP, without having this information can result in security exposure.  That breaks the current admission chain and that cannot be allowed for beta.

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,292 @@
...
+  ApplyMutations(a Attributes) error
+}
+
+type Validator interface {
+  // Validate is executed after the appropriate `rest.Storage` handler's
+  // validation occurs. This gives the plugin the opportunity to validate the
+  // object after the handler has validated it, but before it is persisted. If
+  // an error is returned, the request is marked as failed and no changes are
+  // persisted.
+  Validate(a Attributes) error
```
derekwaynecarr: so limit range defaulting moves to apply mutation, but limit range enforcement moves to validate.

resource quota enforcement moves to validate only, and we stop having the problem of invalid objects bumping usage.

:100:

sig-api-machinery/extensible-admission-to-beta.md
```diff
@@ -0,0 +1,292 @@
...
+We'd like to break this into multiple interfaces, one per phase, called at the appropriate times
+within the request lifecycle.
+
+```go
+type Mutator interface {
+  // ApplyMutations is executed after the request body is deserialized into an
+  // object for create and update requests.
+  // Plugins are free to modify the object as they see fit. If an error is
+  // returned, the request is marked as failed and no changes are persisted.
+  ApplyMutations(a Attributes) error
```
derekwaynecarr: each mutator is free to leave the object in an invalid state, correct?

derekwaynecarr:
the proposed call sequence would simplify a lot of real world problems we have encountered.

sig-api-machinery/extensible-admission-to-beta.md
deads2k: > resource quota enforcement moves to validate only, and we stop having the problem of invalid objects bumping usage.

Correct.

sig-api-machinery/extensible-admission-to-beta.md
deads2k: > each mutator is free to leave the object in an invalid state, correct?

That is an ongoing debate.  Effectively today, yes.  I'm not sure how strict @liggitt  or @caesarxuchao want to be on synchronous webhook admission plugins.

sig-api-machinery/extensible-admission-to-beta.md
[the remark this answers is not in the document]
caesarxuchao: > If you change PSP you will break use-cases like an infrastructure component that wants to escalate a user by using a privileged identity for a specifically built pod (we know this use-case exists since we use it openshift).

For such specific pod, can it just skip initialization?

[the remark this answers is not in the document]
deads2k: > For such a specific pod, can it just skip initialization?

The controller itself is still subject to PSP and should still be subject to PSP.  It is simply more powerful than the user, not all powerful.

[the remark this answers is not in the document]
caesarxuchao: Do you mean it should be subject to initializers? Is it possible to point me to the code? It might be easier to understand the problem.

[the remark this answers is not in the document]
tallclair: > Knowing the original user is a key piece of admission used by PSP, NodeRestriction, GC permission checks, and eventratelimiting. Eliminating the power for dynamic admission hooks makes it impossible to build important classes of admission plugins. The workaround of SAs only works for one of those checks.

Are you proposing that updates be checked against PSP's accessible by the original creator's identity? I think that opens up an attack pattern of swapping out the image on an existing privileged pod...

I think the PSP needs to be checked with the updater's identity (+pod SA identity), i.e. no changes to how it works today. I'm not sure I see any way around it, but would love to be proven wrong.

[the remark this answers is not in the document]
caesarxuchao: Maybe the PSP plugin can whitelist the fields it doesn't care, and allow initializers to change them w/o reevaluating the entire pod. And for the sensitive fields, initializers will need proper security permission to update them.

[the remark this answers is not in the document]
caesarxuchao: Just thought of a similar problem: today the garbage collector controller updates pods to change ownerRefs . How does the PSP plugin handle that? Does it give the GC controller super permission?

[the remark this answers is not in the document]
liggitt: > today the garbage collector controller updates pods to change ownerRefs . How does the PSP plugin handle that? 

Funny you should ask: https://github.com/kubernetes/kubernetes/commit/32e5147af94f5becd359a71e3e1f421e9b8a721c

> Does it give the GC controller super permission?

Definitely not.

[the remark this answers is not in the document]
caesarxuchao: That looks good. @liggitt can we do similar things for initializers? Essentially what I proposed in the earlier [comment](https://github.com/kubernetes/community/pull/1137#discussion_r143064604).

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1137)

deads2k:
I think we got most of this tidied up.

PR #1140, opened 2017-10-02
Add  provisioner changes to support block dev

dhirajh:
Moving the doc https://docs.google.com/document/d/1hRRzZpWtbHyJC1lotE8rqp2h1NEXuFv-rAW-vLuQa94/edit?usp=sharing into this block pv document.

cc @msau42

### DIFF
--- a/contributors/design-proposals/storage/local-storage-pv.md
+++ b/contributors/design-proposals/storage/local-storage-pv.md
@@ -432,16 +432,13 @@ metadata:
   name: local-volume-config
   namespace: kube-system
 data:
-  "local-fast": |
-    {
-      "hostDir": "/mnt/ssds",
-      "mountDir": "/local-ssds"
-    }
-  "local-slow": |
-    {
-      "hostDir": "/mnt/hdds",
-      "mountDir": "/local-hdds"
-    }
+  storageClassMap: |
+    local-fast:
+      hostDir: "/mnt/ssds"
+      mountDir: "/local-ssds"
+    local-slow:
+      hostDir: "/mnt/hdds"
+      mountDir: "/local-hdds"
 ```
 
 The `hostDir` is the discovery path on the host, and the `mountDir` is the path it is mounted to in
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -634,10 +634,116 @@ limitations outlined above wil be fixed.
...
-Design TBD
+
+##### Discovery
+
+Much like the current file based PVs, the local volume provisioner will look for block devices
+under designated directories that have been mounted on the provisioner container. Currently, for
+each storage class, the provisioner has a configmap entry that looks like this:
+
+```
+"local-fast":   {
```
msau42: We changed the configmap to use yaml instead of json

dhirajh: Fixed the doc to use yaml format.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -634,10 +628,120 @@ limitations outlined above wil be fixed.
...
+within the directory managed by the provisioner, to prevent destructive operations on arbitrary
+paths.
+
+The provisioner logic currently does each volume’s cleanup as a synchronous serial activity.
+However, with cleanup now potentially being a multi hour activity, the processes will have to
+be asynchronous and capable of being executed in parallel. The provisioner will ensure that all
+current asynchronous cleanup processes are tracked. Special care needs to be taken to ensure that
+when a disk has only been partially cleaned. This scenario can happen if some impatient user
+manually deletes a PV and the provisioner ends up re-creating pv ready for use (but only partially
+cleaned). This issue will be addressed in the re-design of the provisioner (details will be provided
```
msau42: This is still in discussion, whether keeping some local state is good enough or we use a crd.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1140)

contributors/design-proposals/storage/local-storage-pv.md
msau42: While the redesign is still in discussion, should we just store some local state to handle this case?

dhirajh: @msau42 I think there are couple of other options that might be worth considering
1. Since block is still not enabled in provisioner and API, and the concerns about impatient admin only seem relevant there, we could go with the current code without any signifcant risks.
2. We could actually go with the CRD approach for book keeping here as it seems the least controversial and is likely to be used even in the new design.

msau42: Ok, let's leave the design open, but let's check in the code (disabled by default).

msau42:
/lgtm

It's been a month and no-one else has commented.  I'll assume that there are no objections.  We can update the doc with any changes in the future.

PR #1143, opened 2017-10-04
Don't retire kube-state-metrics. Retire md-check.

mindprince:
kube-state-metrics is active once again (it became active soon after the first commit to this file when it was proposed to be retired). It seems to be the only popular tool that does what it does.

Decision to retire md-check was taken in https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-dev/0FcCGNtnxE0/JmQfETD0AwAJ


bgrant0607:
thanks
/lgtm

PR #1145, opened 2017-10-04
Adding ONCALL.md with details about our oncall rotation

nikhiljindal:
All this info was in our oncall calendar invite.
This will make it easily accessible to everyone.

cc @quinton-hoole @csbell

### DIFF
--- a/sig-federation/ONCALL.md
+++ b/sig-federation/ONCALL.md
@@ -1,0 +1,50 @@
+# Overview
+
+We have an oncall rotation in the SIG. The role description is as follows:
+
+* Ensure that the testgrid (https://k8s-testgrid.appspot.com/sig-federation) is green. This person will be the point of contact if testgrid turns red. Will identify the problem and fix it (most common scenarios: find culprit PR and revert it or free quota by deleting leaked resources).
+Will also report most common failure scenarios and suggest improvements. Its up to the sig or individuals to prioritize and take up those tasks.
+
+Oncall playbook: https://github.com/kubernetes/community/blob/master/contributors/devel/on-call-federation-build-cop.md
+
+# Joining the rotation
+
+Add your name at the end of the current rotation schedule.
+
+# Oncall schedule
+
+```
+25 September - 1 October: Madhu (https://github.com/madhusudancs)
+2 October - 8 October: Shashidhara (https://github.com/shashidharatd)
+9 October - 15 October: Christian (https://github.com/csbell)
+16 October - 22 October: Nikhil Jindal (https://github.com/nikhiljindal)
+23 October - 29 October: Irfan (https://github.com/irfanurrehman)
+30 October - 5 November: Maru (https://github.com/marun)
+6 November - 12 November: Jonathan (https://github.com/perotinus)
+(Madhu to be removed from next rotation)
+
+(Adding Irfan)
+7 August - 13 August: Nikhil Jindal (https://github.com/nikhiljindal)
+14 August - 20 August: Shashidhara (https://github.com/shashidharatd)
+21 August - 27 August: Christian (https://github.com/csbell)
+28 August - 3 September: Madhu (https://github.com/madhusudancs)
+4 September - 10 September: Irfan (https://github.com/irfanurrehman)
+11 September - 17 September: Maru (https://github.com/marun)
+18 September - 24 September: Jonathan (https://github.com/perotinus)
+
+
+(Adding Jonathan)
+26 June - 2 July: Nikhil Jindal (https://github.com/nikhiljindal)
+3 July - 9 July: Shashidhara (https://github.com/shashidharatd)
+10 July - 16 July: Christian (https://github.com/csbell)
+17 July - 23 July: Madhu (https://github.com/madhusudancs)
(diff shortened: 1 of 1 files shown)


sig-federation/ONCALL.md
```diff
@@ -0,0 +1,50 @@
...
+* Ensure that the testgrid (https://k8s-testgrid.appspot.com/sig-federation) is green. This person will be the point of contact if testgrid turns red. Will identify the problem and fix it (most common scenarios: find culprit PR and revert it or free quota by deleting leaked resources).
+Will also report most common failure scenarios and suggest improvements. Its up to the sig or individuals to prioritize and take up those tasks.
+
+Oncall playbook: https://github.com/kubernetes/community/blob/master/contributors/devel/on-call-federation-build-cop.md
+
+# Joining the rotation
+
+Add your name at the end of the current rotation schedule.
+
+# Oncall schedule
```
quinton-hoole: We're going to need a process to ensure that this get refreshed periodically, and that when people are unable to do their oncall duty (vacation, illness etc), there's a well-defined process for updating this.

Here's a strawman.  If you I agree, I suggest you add it to the on-call playbook, with a link from here.

1. The rotation schedule stretches one quarter into the future (so that everyone has due warning, and can plan accordingly).
2. The current oncall is responsible for checking that, and extending it as soon as we have less than 1 quarter's worth of runway (when they come oncall, they need to check who was on last, who's on next etc anyway, so submitting a short PR to extend the cycle seems easy enough).
3. If anyone is unable to do their oncall duty, they are responsible for finding someone to swap with (by sending a PR, approved by that person).  Swapping one week for another is usually relatively uncontentious.

Thoughts?

sig-federation/ONCALL.md
```diff
@@ -0,0 +1,50 @@
+# Overview
+
+We have an oncall rotation in the SIG. The role description is as follows:
+
+* Ensure that the testgrid (https://k8s-testgrid.appspot.com/sig-federation) is green. This person will be the point of contact if testgrid turns red. Will identify the problem and fix it (most common scenarios: find culprit PR and revert it or free quota by deleting leaked resources).
```
quinton-hoole: I think that we might want to extend this to monitoring the incoming queue of bug reports, and prioritizing and/or assigning the critical ones.  But we can debate that in the sig and do it in a followup PR as necessary.

sig-federation/ONCALL.md
```diff
@@ -0,0 +1,50 @@
...
+25 September - 1 October: Madhu (https://github.com/madhusudancs)
+2 October - 8 October: Shashidhara (https://github.com/shashidharatd)
+9 October - 15 October: Christian (https://github.com/csbell)
+16 October - 22 October: Nikhil Jindal (https://github.com/nikhiljindal)
+23 October - 29 October: Irfan (https://github.com/irfanurrehman)
+30 October - 5 November: Maru (https://github.com/marun)
+6 November - 12 November: Jonathan (https://github.com/perotinus)
+(Madhu to be removed from next rotation)
+
+(Adding Irfan)
```
quinton-hoole: Can we get rid of (or gray out) the history?  I assume it's here so that we have a record and can extend the rotation fairly?  It's a bit distracting for people who are only looking for who's oncall today.

quinton-hoole:
Minor comments.  Otherwise looks good.

sig-federation/ONCALL.md
nikhiljindal: a quarter seems too much. Will we not add or remove members till end of quarter?
If we do allow changing then people cant really plan.

Agree on 3rd thats what we have been doing. Will add it to the doc.

sig-federation/ONCALL.md
nikhiljindal: ack

sig-federation/ONCALL.md
nikhiljindal: Have put it in a different Past cycles section.

nikhiljindal:
Thanks for the quick review @quinton-hoole 
Updated the PR as per comments. PTAL.

sig-federation/ONCALL.md
```diff
@@ -0,0 +1,62 @@
...
+We have an oncall rotation in the SIG. The role description is as follows:
+
+* Ensure that the testgrid (https://k8s-testgrid.appspot.com/sig-federation) is green. This person will be the point of contact if testgrid turns red. Will identify the problem and fix it (most common scenarios: find culprit PR and revert it or free quota by deleting leaked resources).
+Will also report most common failure scenarios and suggest improvements. Its up to the sig or individuals to prioritize and take up those tasks.
+
+Oncall playbook: https://github.com/kubernetes/community/blob/master/contributors/devel/on-call-federation-build-cop.md
+
+# Joining the rotation
+
+Add your name at the end of the current rotation schedule.
```
cblecker: Are there requirements to join? Permissions? Do you have to be a googler? I'm guessing these are important pieces for any on-call rotation.

nikhiljindal: Thanks for pointing that out. Added information to the doc. PTAL

PR #1148, opened 2017-10-05
Clearer generator docs

### LINKED ISSUE #1121: Document how to update SIG-foo READMEs
We now should modify `sigs.yaml` and then run `make` to update `sig-list.md` and `sig-foo README`. But recently I saw there were several PRs whose authors didn't know this.

https://github.com/kubernetes/community/pull/931#issuecomment-323051821
https://github.com/kubernetes/community/pull/1089#issuecomment-330575222
https://github.com/kubernetes/community/pull/1120#discussion_r140520897

We should document this. I suggested to add some description into `sigs.yaml` and then `make` to add the description to the doc header of each sig README.

cc @jamiehannaford @cblecker

jamiehannaford:
There still seems to be a few PRs that are unaware of how to generate SIG docs, so this PR hopes to plug that knowledge gap by:

- adding a reference to the README
- adding a PR template that contributors see every time they send a changeset

We might not need the checkbox in the PR template, so I can remove if we feel it should go.

Fixes #1121.

/cc @xiangpengzhao @cblecker

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -3,13 +3,5 @@
 - If you are editing SIG information, please follow these instructions: https://github.com/kubernetes/community/tree/master/generator
   You will need to follow these steps:
   1. Edit sigs.yaml with your change 
-  2. Generate docs with `make all`. To build docs for one sig, run `make WHAT=sig-apps all`
+  2. Generate docs with `make generate`. To build docs for one sig, run `make WHAT=sig-apps generate`
 -->
-
-**What this PR does**:
-
-**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #
-
-**SIG changes**
-
-- [] If changing SIG information, I have read [these instructions](https://github.com/kubernetes/community/tree/master/generator) and re-generated the docs
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,0 +1,15 @@
+<!--  Thanks for sending a pull request!  Here are some tips for you:
+- If this is your first contribution, read our Getting Started guide https://github.com/kubernetes/community#your-first-contribution
+- If you are editing SIG information, please follow these instructions: https://github.com/kubernetes/community/tree/master/generator
+  You will need to follow these steps:
+  1. Edit sigs.yaml with your change 
+  2. Generate docs with `make all`. To build docs for one sig, run `make WHAT=sig-apps all`
+-->
+
+**What this PR does**:
+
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,15 @@
+<!--  Thanks for sending a pull request!  Here are some tips for you:
+- If this is your first contribution, read our Getting Started guide https://github.com/kubernetes/community#your-first-contribution
+- If you are editing SIG information, please follow these instructions: https://github.com/kubernetes/community/tree/master/generator
+  You will need to follow these steps:
+  1. Edit sigs.yaml with your change 
+  2. Generate docs with `make all`. To build docs for one sig, run `make WHAT=sig-apps all`
```
cblecker: The keyword is "generate" not "all"

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,15 @@
...
+-->
+
+**What this PR does**:
+
+**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #
+
+**SIG changes**
+
+- [] If changing SIG information, I have read [these instructions](https://github.com/kubernetes/community/tree/master/generator) and re-generated the docs
```
cblecker: I don't think the stuff outside of the commented out section is needed. Community is more flexible, and many PRs here are proposals and such

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,15 @@
+<!--  Thanks for sending a pull request!  Here are some tips for you:
```
xiangpengzhao: When contributors are at this step (sending a PR), they may modify the docs in the wrong way already. I suggest to make them see the tips as early as possible. What I expected in #1121 is adding some tips in each README of the sigs. We can add a field `tip` in `sigs.yaml` and auto generate it to all the READMEs. Contributors can see what they should do when they want to make their change on a README.

jamiehannaford: We already do this with comments at the top of every README (see [here](https://raw.githubusercontent.com/kubernetes/community/master/sig-api-machinery/README.md)), so folks will know by the time they come to edit the file. 

Do you mean make these tips visible on each SIG README too? I'm wondering how that adds more value than just embedding stuff in the comment section, since they'll see it.

.github/PULL_REQUEST_TEMPLATE
jamiehannaford: Yeah, good point. I'll remove this

.github/PULL_REQUEST_TEMPLATE
xiangpengzhao: Yes I meant this. But since the tips are already in the comments (sorry I didn't notice that :), it's not necessary to make them visible.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1148)

PR #1149, opened 2017-10-05
Updating Scalability page to reflect shift to biweekly schedule.

countspongebob:
Updating Scalability page to reflect shift to biweekly schedule.

PR #1150, opened 2017-10-05
Fix sig-scalability

cblecker:
#1149 updated the README for sig-scalability without modifying sigs.yaml. This fixes that, as well as moves the remaining meeting dates into the custom content bit.


countspongebob:
@cblecker Did I make a change in the wrong way? Point me to the proper procedure if so.

cblecker:
@countspongebob We're working to improve the docs (https://github.com/kubernetes/community/blob/master/generator/README.md), but the change would be made in the sigs.yaml file in the root of the repo, and then a `make` run to generate the README and the sigs-list.md. A bit of standardization to ensure all the sig docs look the same, and that dates/times are in all the right places

PR #1152, opened 2017-10-05
adding predicate ordering design proposal

yastij:
/cc @gmarek @k82cn @bsalamat

### DIFF
--- a/contributors/design-proposals/predicates-ordering.md
+++ b/contributors/design-proposals/predicates-ordering.md
@@ -1,0 +1,93 @@
+# predicates ordering
+
+
+
+Status: proposal
+
+Author: yastij
+Approvers: 
+* gmarek
+* bsalamat
+* k82cn
+
+
+
+
+## Abstract
+
+This document describes how and why reordering predicates helps to achieve performance for the kubernetes scheduler.
+We will expose the motivations behind this proposal, The two steps/solution we see to tackle this problem and the timeline decided to implement these.
+
+
+## Motivation
+
+While working on a [Pull request](https://github.com/kubernetes/kubernetes/pull/50185) related to a proposal, we saw that the order of running predicates isn’t defined. 
+
+This makes the scheduler perform extra-computation that isn’t needed, As an example we [outlined](https://github.com/kubernetes/kubernetes/pull/50185) that the kubernetes scheduler runs predicates against nodes even if marked “unschedulable”.
+
+Reordering predicates allows us to avoid this problem, by computing the most restrictive predicates first. To do so, we propose two reordering types.
+
+
+
+## Static ordering
+
+This ordering will be the default ordering. If a policy config is provided with a subset of predicates, only those predicates will be invoked using the static ordering. 
+
+
+
+
+|Position                  | Predicate                        | comments (note, justification...)              |
+ ----------------- | ---------------------------- | ------------------
+| 1 | `CheckNodeConditionPredicate`  | we really don’t want to check predicates against unschedulable nodes. |
+| 2           | `PodFitsHost`            | we check the pod.spec.nodeName. |
+| 3           | `PodFitsHostPorts` | we check ports asked on the spec. |
+| 4 | `PodMatchNodeSelector`            | check node label after narrowing search. |
+| 5           | `PodFitsResources `            | this one comes here since it’s not restrictive enough as we do not try to match values but ranges. |
+| 6           | `NoDiskConflict` | Following the resource predicate, we check disk |
+| 7 | `PodToleratesNodeTaints '`            | check toleration here, as node might have toleration |
+| 8          | `PodToleratesNodeNoExecuteTaints`            | check toleration here, as node might have toleration |
+| 9           | `CheckNodeLabelPresence ` | labels are easy to check, so this one goes before |
+| 10 | `checkServiceAffinity `            | - |
+| 11           | `MaxPDVolumeCountPredicate `            | - |
+| 12           | `VolumeNodePredicate ` | - |
+| 13 | `VolumeZonePredicate `            | - |
+| 14           | `CheckNodeMemoryPressurePredicate`            | doesn’t happen often |
+| 15           | `CheckNodeDiskPressurePredicate` | doesn’t happen often |
+| 16 | `InterPodAffinityMatches`            | Most expensive predicate to compute |
+
+
+## End-user ordering
+
+Using scheduling policy file, the cluster admin can override the default static ordering. This gives administrator the maximum flexibility regarding scheduler behaviour and enables scheduler to adapt to cluster usage. 
+Please note that the order must be a positive integer, also, when providing equal ordering for many predicates, scheduler will determine the order and won't guarantee that the order will remain the same between them.
+Finally updating the scheduling policy file will require a scheduler restart.
+
+as an example the following is scheduler policy file using an end-user ordering:
+
+``` json
+{
+"kind" : "Policy",
+"apiVersion" : "v1",
+"predicates" : [
+	{"name" : "PodFitsHostPorts", "order": 2},
+	{"name" : "PodFitsResources", "order": 3},
+	{"name" : "NoDiskConflict", "order": 5},
+	{"name" : "PodToleratesNodeTaints", "order": 4},
+	{"name" : "MatchNodeSelector", "order": 6},
+	{"name" : "PodFitsHost", "order": 1}
+	],
+"priorities" : [
+	{"name" : "LeastRequestedPriority", "weight" : 1},
+	{"name" : "BalancedResourceAllocation", "weight" : 1},
+	{"name" : "ServiceSpreadingPriority", "weight" : 1},
+	{"name" : "EqualPriority", "weight" : 1}
+	],
+"hardPodAffinitySymmetricWeight" : 10
+}
+```
+
+
+## Timeline
+
+* static ordering: GA in 1.9
+* dynamic ordering: TBD based on customer feedback


contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+## Static ordering
+
+This ordering will be the default ordering. If a policy config is provided with a subset of predicates, only those predicates will be invoked using the static ordering. 
+
+
+
+
+|Position                  | Predicate                        | comments (notes, justifications...)              |
+ ----------------- | ---------------------------- | ------------------
+| 1 | `CheckNodeConditionPredicate`  | we really don’t wanna check predicates against unschedulable nodes. |
```
bsalamat: s/wanna/want to/

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+| 13 | `VolumeZonePredicate `            | - |
+| 14           | `CheckNodeMemoryPressurePredicate`            | doesn’t happen often |
+| 15           | `CheckNodeDiskPressurePredicate` | doesn’t happen often |
+| 16 | `InterPodAffinityMatches`            | Most expensive predicate to compute |
+
+
+## End-user ordering
+
+using scheduling policy file, the cluster admin can override the default static ordering. This gives administrator the maximum flexibility regarding scheduler behaviour and enables scheduler to adapt to cluster usage. as an example the following is scheduler policy file using an end-user ordering:
+
```
bsalamat: Is `order` a positive integer?

Could you please also add that when the provided `order` is the same for two or more predicates, scheduler will determine the order and there is no guarantee that the order remains the same over time.

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
+# predicates ordering
+
```
bsalamat: This proposal should go in the `scheduling` directory.

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+## End-user ordering
+
+using scheduling policy file, the cluster admin can override the default static ordering. This gives administrator the maximum flexibility regarding scheduler behaviour and enables scheduler to adapt to cluster usage. as an example the following is scheduler policy file using an end-user ordering:
+
+``` json
+{
+"kind" : "Policy",
+"apiVersion" : "v1",
+"predicates" : [
+	{"name" : "PodFitsHostPorts", "order": 2},
```
k82cn: Can we just follow the order in configuration? so "order" is not necessary.

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
+# predicates ordering
+
+
+
+status: proposal
+
+Author: Yassine TIJANI.
```
k82cn: please also share your github handle :).

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
+# predicates ordering
+
+
+
+status: proposal
```
k82cn: s/status/Status

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+* gmarek
+* bsalamat
+* k82cn
+
+
+
+
+## Abstract
+
+This document describes how and why reordering predicates helps to achieve performance for the kubernetes scheduler. We will expose the motivations behind this proposal, The two steps/solution we see to tackle this problem and the timeline decided to implement these.
```
k82cn: There's general suggestion that: build paragraph multiple lines, so reviewers can comments on each line here.

contributors/design-proposals/predicates-ordering.md
k82cn: @yastij , as we talked on Slack, prefer to also add something like `O(n), n= # of node`. The user can evaluate the complexity of predicates based on their environment; so the lowest complexity with  most restrictive come first.

contributors/design-proposals/predicates-ordering.md
yastij: @k82cn - I'd prefer having a clear ordering here, as using an implicit one might confuse users.

contributors/design-proposals/predicates-ordering.md
yastij: @k82cn - I thought that time complexity might be more relevant in the user docs, as most of end-user will refer to the documentation when trying to check time complexity, you think we should add it here instead ?

contributors/design-proposals/predicates-ordering.md
yastij: @bsalamat - indeed you're right

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1152)

contributors/design-proposals/predicates-ordering.md
bsalamat: I also like the explicit order, but we should be backward compatible. So, absence of "order" should be treated as `order: 0`. We should mention this in the design.

yastij: @bsalamat - also maybe add that if there's a non-logical ordering (e.g. order:1, order:2, order:16) default to static ordering with a warning ?

Also I agree that absence of order should be treated as order:0 (i.e. executed at the end & randomly right after the ranked predicates)

bsalamat: What is wrong with 1, 2, 16? It shows a clear ordering and should be treated a valid ordering.

Yes for the second part. Order:0 means executed at no particular order and after the ordered predicates. If none of the predicates has an order, then should run with the default order.

yastij: Meant -16

bsalamat: Yes, order must be a positive integer. In case of an invalid value (such as a negative number or a string) we should probably fail to start. Users may miss the warning and not notice the issue in their config. Failing to start is in-line with how scheduler reacts to other config issues.

yastij: SGTM

k82cn:
*[contributors/design-proposals/predicates-ordering.md, line 40 at r1](https://reviewable.kubernetes.io:443/reviews/kubernetes/community/1152#-Kw5zTufqEA2fjije68p-r1-40:-Kw95PULJB3MHVlimhDR:b7urz8s) ([raw file](https://github.com/kubernetes/community/blob/1a4e1a05a22ca253dcee9765e90991dca2c81afa/contributors/design-proposals/predicates-ordering.md#L40)):*
<details><summary><i>Previously, yastij (Yassine TIJANI) wrote…</i></summary><blockquote>

@k82cn - I thought that time complexity might be more relevant in the user docs, as most of end-user will refer to the documentation when trying to check time complexity, you think we should add it here instead ?
</blockquote></details>

I'd suggest to add it; so when anyone draft the user doc, just copy it. And the time complexity will also help us to work out default order of predicates :).

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/community/1152)*

k82cn:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions.

---

*[contributors/design-proposals/predicates-ordering.md, line 67 at r1](https://reviewable.kubernetes.io:443/reviews/kubernetes/community/1152#-Kw5zTufqEA2fjije68p-r1-67:-Kw965oHr3CIeaHjtayz:b-ivlld) ([raw file](https://github.com/kubernetes/community/blob/1a4e1a05a22ca253dcee9765e90991dca2c81afa/contributors/design-proposals/predicates-ordering.md#L67)):*
<details><summary><i>Previously, yastij (Yassine TIJANI) wrote…</i></summary><blockquote>

SGTM
</blockquote></details>

I'm ok with explicit order; please add enough validation for it, e.g. two predicates with same order.

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/community/1152)*

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+	{"name" : "EqualPriority", "weight" : 1}
+	],
+"hardPodAffinitySymmetricWeight" : 10
+}
+```
+
+
+## Timeline
+
+both types of ordering will be delivered as an alpha version, as for the timeline:
```
gmarek: s/b/B/

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+| 12           | `VolumeNodePredicate ` | - |
+| 13 | `VolumeZonePredicate `            | - |
+| 14           | `CheckNodeMemoryPressurePredicate`            | doesn’t happen often |
+| 15           | `CheckNodeDiskPressurePredicate` | doesn’t happen often |
+| 16 | `InterPodAffinityMatches`            | Most expensive predicate to compute |
+
+
+## End-user ordering
+
+using scheduling policy file, the cluster admin can override the default static ordering. This gives administrator the maximum flexibility regarding scheduler behaviour and enables scheduler to adapt to cluster usage. as an example the following is scheduler policy file using an end-user ordering:
```
gmarek: s/u/U/

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+"hardPodAffinitySymmetricWeight" : 10
+}
+```
+
+
+## Timeline
+
+both types of ordering will be delivered as an alpha version, as for the timeline:
+
+* static ordering: available in 1.9
```
gmarek: Static ordering is fine to do straight away, as it doesn't change how users use clusters, but I think we probably should discuss putting dynamic ordering initially as beta. I'm not sure if it's worth it - we need to discuss it with SIG PM (whatever they handle is).

yastij:
@gmarek - I thought that we'd wait for metrics from the community to start working on the dynamic ordering ?

contributors/design-proposals/predicates-ordering.md
yastij: @gmarek - I thought that we'd wait for metrics from the community to start working on the dynamic ordering ?

bsalamat: I think it is a good idea to add static ordering first and wait for feedback from users before adding dynamic ordering. If users asked for dynamic ordering, we will add it.

yastij: @gmarek @bsalamat - I'm a bit confused here, dynamic ordering isn't part of this design proposal, since we need data on predicates restrictiveness.

I'm fine with just having for static ordering and the end-user one for the moment.

bsalamat: I was referring to configurable ordering by end-users as "dynamic ordering". Do we already know customers who need the configurable ordering?

bsalamat: I still would like to hear who are the customers who needed it and had to write their own custom scheduler only to reorder predicates and why a smarter ordering (that we are proposing in static ordering of this proposal) is not enough for them.

yastij: If you are using very specific set of labels, executing checkNodeLabelPresence first is better. 

@gmarek - I'll try to join the next SIG PM meeting to discuss this.

gmarek: Oh, sorry, if the 'static' means 'at startup' and dynamic means 'changes according to the cluster state' then you're right, I understood it differently. It's probably not worth joining the meeting, as they most likely have full agenda, email should work better.

@apsinha @timothysc - do you know if we should use normal alpha-beta-GA (most likely skipping alpha) lifecycle for things like this (allowing user to change the order of predicate computation in scheduler - setting good one improves scheduler's performance, setting terrible one pretty much kills it). It's small enough change from the user perspective, that maybe we can skip this?

contributors/design-proposals/predicates-ordering.md
yastij: After thinking a bit more about time complexity, it seems that giving an estimation using the number of nodes might be not very useful, since the vast majority of predicates are computed against every node using the parallelized workqueue with a checkNode function.

@bsalamat @k82cn  - to your knowledge is there any common input to predicates that can gives us insights on time complexity ?

bsalamat: I don't think there is any way to find runtime complexity of predicates other than examining their algorithms, but the autoscaler team recently ran some tests to find execution times of various predicates. You can find the results here: https://github.com/kubernetes/autoscaler/issues/257. It gives you some idea about the runtime complexity of the predicates.

yastij: @bsalamat - thank you, indeed it should be enough

bsalamat:
We discussed the configurable ordering of predicates in the SIG scheduling meeting yesterday. We believe that the configurable ordering shouldn't be implemented at this point. We can go ahead and build the static ordering and leave the configurable ordering util we get more feedback from our customers that they need it.

yastij:
@bsalamat - SGTM

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,90 @@
...
+* dynamic ordering: available in 1.10
```
bsalamat: So, could you please change this from 1.10 to "TBD based on customer feedback"?

yastij:
@gmarek @bsalamat @k82cn - sorry guys i was off, I just came back, I updated the design proposal.

Also waiting for feedback from @timothysc and @apsinha on wether static ordering should be delivered as alpha or directly available

contributors/design-proposals/predicates-ordering.md
```diff
@@ -84,7 +88,7 @@ using scheduling policy file, the cluster admin can override the default static
 
 ## Timeline
 
-both types of ordering will be delivered as an alpha version, as for the timeline:
+Both types of ordering will be delivered as an alpha version, as for the timeline:
```
bsalamat: Static ordering of predicates shouldn't have noticeable user impact and we don't expect change of behavior. It is more of an optimization. So, I don't think we need to make it alpha. It can go to production directly.

bsalamat:
LGTM
Let's wait for others to chime in.

yastij:
@gmarek @k82cn - LGTY ?

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,92 @@
...
+Author: yastij
+Approvers: 
+* gmarek
+* bsalamat
+* k82cn
+
+
+
+
+##Abstract
```
gmarek: ## Abstract

gmarek: Ech... it does md in comments as well. Add space between # and A.

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,92 @@
...
+
+This document describes how and why reordering predicates helps to achieve performance for the kubernetes scheduler.
+We will expose the motivations behind this proposal, The two steps/solution we see to tackle this problem and the timeline decided to implement these.
+
+
+##Motivation
```
gmarek: Add space.

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,92 @@
...
+
+While working on a [Pull request](https://github.com/kubernetes/kubernetes/pull/50185) related to a proposal, we saw that the order of running predicates isn’t defined. 
+
+This makes the scheduler perform extra-computation that isn’t needed, As an example we [outlined](https://github.com/kubernetes/kubernetes/pull/50185) that the kubernetes scheduler runs predicates against nodes even if marked “unschedulable”.
+
+Reordering predicates allows us to avoid this problem, by computing the most restrictive predicates first. To do so, we propose two reordering types.
+
+
+
+##Static ordering
```
gmarek: space.

contributors/design-proposals/predicates-ordering.md
```diff
@@ -0,0 +1,92 @@
...
+| 13 | `VolumeZonePredicate `            | - |
+| 14           | `CheckNodeMemoryPressurePredicate`            | doesn’t happen often |
+| 15           | `CheckNodeDiskPressurePredicate` | doesn’t happen often |
+| 16 | `InterPodAffinityMatches`            | Most expensive predicate to compute |
+
+
+## End-user ordering
+
+Using scheduling policy file, the cluster admin can override the default static ordering. This gives administrator the maximum flexibility regarding scheduler behaviour and enables scheduler to adapt to cluster usage. 
+Please note that the order must be a positive integer, also, when providing equal ordering for many predicates, scheduler will determine the order and won't guarantee that the order will remain the same between them.
```
gmarek: I'm not sure if it's clear for everyone so we should mention that updating the Policy requires scheduler restart.

yastij:
@gmarek @bsalamat @k82cn - commits squashed and ready for approval

contributors/design-proposals/predicates-ordering.md
gmarek: Ping this comment.

yastij: @gmarek @bsalamat - it's already the case when changing predicates on the policy file ?

bsalamat: I think it is worth noting that point again in this doc, just in case someone is not fully aware of that point.

yastij: SGTM

yastij: updated

bgrant0607:
@yastij @bsalamat Please move this doc to the design-proposals/scheduling subdir. We categorized all the design docs/proposals by SIG over a month ago.

yastij:
@bgrant0607 - Indeed forgot to do it, I'll move it asap, thanks !

PR #1154, opened 2017-10-05
Initial commit of the election process

castrojo:
This adds the process we used to run the steering committee election to the community directory so we can refer to it for the election next year.


bgrant0607:
Good start. Thanks. Merge and iterate.
/lgtm

PR #1163, opened 2017-10-06
Add a proposal for ConfigMap/Secret garbage collection

janetkuo:
Original doc: https://docs.google.com/document/d/1RkB1HEz8hWSmHQp8_spDFHj498eARBz4EjPvS2zVfAg/edit (join https://groups.google.com/forum/#!forum/kubernetes-sig-apps for access)

ref https://github.com/kubernetes/kubernetes/issues/22368

@kubernetes/sig-apps-api-reviews @erictune

### DIFF
--- a/contributors/design-proposals/apps/configmap-secret-gc.md
+++ b/contributors/design-proposals/apps/configmap-secret-gc.md
@@ -1,0 +1,271 @@
+# Garbage Collection of ConfigMaps and Secrets
+
+**Author**: Janet Kuo (@janetkuo)
+
+**Status**: Proposal
+
+- [Abstract](#abstract)
+- [Requirements](#requirements)
+- [API Changes](#api-changes)
+- [Controllers](#controllers)
+  * [Core Controllers that Support ConfigMaps/Secrets Garbage Collection](#core-controllers-that-support-configmapssecrets-garbage-collection)
+  * [Update ConfigMap/Secret's Owner List](#update-configmapsecrets-owner-list)
+    + [On Controller API Resource Creation](#on-controller-api-resource-creation)
+    + [On Controller API Resource Template Updates](#on-controller-api-resource-template-updates)
+- [Alternatives](#alternatives)
+  * [Embed ConfigMaps/Secrets in Controller Spec](#embed-configmapssecrets-in-controller-spec)
+  * [Specify ConfigMaps/Secrets Garbage Collection in Controller Spec](#specify-configmapssecrets-garbage-collection-in-controller-spec)
+  * [Specify ConfigMaps/Secrets garbage collection in PodSpec](#specify-configmapssecrets-garbage-collection-in-podspec)
+  * [Automatic ConfigMap/Secret References Updates](#automatic-configmapsecret-references-updates)
+- [Caveats](#caveats)
+- [Testing Plan](#testing-plan)
+
+
+## Abstract
+
+ConfigMaps allow users to decouple configuration artifacts from image content to
+keep containerized applications portable. The data stored in ConfigMaps can be
+consumed in pods or provide the configurations for system components. Secrets
+are similar to ConfigMaps, but are intended to hold sensitive information. 
+
+To do a rolling update of a ConfigMap/Secret, users need to create a new
+ConfigMap, update its references in workloads resources (such as Deployment),
+and delete the old ConfigMap. To facilitate ConfigMap/Secrets rollouts and
+management ([#22368](https://github.com/kubernetes/kubernetes/issues/22368)), we
+propose a mechanism that supports garbage collection of unused ConfigMaps and
+Secrets. 
+
+
+## Requirements
+
+1.  Garbage collection of a ConfigMap/Secret should be easy to configure,
+    manage, and understand. 
+1.  If a user wants a ConfigMap/Secret to be garbage collected, the intent
+    should be easily discoverable by other users who consume the same
+    ConfigMap/Secret. No surprises. 
+1.  The design of garbage collection of ConfigMaps/Secrets should cover
+    Deployments, StatefulSets, DaemonSets, ReplicaSets, and
+    ReplicationControllers. 
+1.  For controllers that support history and rollback feature:
+    ConfigMaps/Secrets should not be garbage collected until the referencing
+    controller history object is removed, so that rollback still works. 
+1.  Garbage collection of ConfigMaps/Secrets should work after a controller's
+    pod template is updated.
+1.  Users can only enable or disable the garbage collection of a
+    ConfigMap/Secret when creating the ConfigMap/Secret.
+
+
+## API Changes
+
+A new *immutable* API field is introduced to ConfigMap and Secret API. 
+
+
+```go
+type ConfigMap struct {
+  // If true, the supported controllers (Deployments, StatefulSets, DaemonSets,
+  // ReplicaSets, and ReplicationControllers) that reference this ConfigMap will
+  // become its owner. 
+  // Note that garbage collector will delete this ConfigMap once its owners are 
+  // gone, even if this ConfigMap is still referenced by other API resources
+  // (such as Jobs).
+  // This can only be specified at creation time.
+  // Default to false.
+  ControllersAsOwners *bool
+}
+```
+
+
+```go
+type Secret struct {
+  // If true, the supported controllers (Deployments, StatefulSets, DaemonSets,
+  // ReplicaSets, and ReplicationControllers) that reference this Secret will
+  // become its owner. 
+  // Note that garbage collector will delete this ConfigMap once its owners are 
+  // gone, even if this ConfigMap is still referenced by other API resources
+  // (such as Jobs).
+  // This can only be specified at creation time.
+  // Default to false.
+  ControllersAsOwners *bool
+}
+```
+
+ConfigMap and Secret garbage collection are disabled by default for backward
+compatibility. 
+
+
+## Controllers
+
+This section is presented as a generalization of how an arbitrary controller can
+use this new API field to garbage collect the ConfigMaps/Secrets it references. 
+
+A controller can have the ConfigMaps/Secrets it references garbage collected by
+declaring ownership of those ConfigMaps/Secrets, i.e. set the ConfigMaps/Secrets
+(`.metadata.ownerReferences`).
+The [garbage collector](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/)
+will then delete the ConfigMaps/Secrets when the owners are all gone. Note that
+garbage collector only deletes objects that once have owners but no longer do.
+ConfigMaps/Secrets won't be garbage collected if they never have any owner. 
+
+Most workload controllers support rolling update, revision history and rollback
+features. To make sure rollback still works, for a controller with revision
+history feature, the ConfigMaps/Secrets ownership should belong to the
+controller's history instead of the controller itself. For example, a Deployment
+should not become the owner of any referenced ConfigMaps/Secrets, but the
+Deployment's history objects (ReplicaSets) will own those ConfigMaps/Secrets.
+
+If a controller does not support rolling update, revision history, or rollback
+(such as ReplicaSets and ReplicationControllers), it should mark itself as the
+owner of the ConfigMaps/Secrets it references in order to have them garbage
+collected when the controller is gone. 
+
+
+### Core Controllers that Support ConfigMaps/Secrets Garbage Collection
+
+*   Controllers with history: Deployments, StatefulSets, DaemonSets
+*   Controllers without history: ReplicaSets, ReplicationControllers 
+
+
+### Update ConfigMap/Secret's Owner List
+
+There are 2 possible scenarios when a referenced ConfigMap/Secret's owner list
+(`.metadata.ownerReferences`) needs to be updated:
+
+1.  After a controller API resource is created.
+1.  After a controller API resource's pod template is updated, and: 
+    1.  The controller's current history object is created, or 
+    1.  The ConfigMap/Secret references in the pod template are changed. 
+
+Note that Deployment is a special case here. Because a Deployment's revision
+history (ReplicaSets) can declare ConfigMaps/Secrets ownership directly, we
+don't need to implement it in Deployment controller code. Deployment controller
+delegates the work to ReplicaSet controller. 
+
+#### On Controller API Resource Creation
+
+When a controller API object (with or without history) exists, the controller
+will do the following:
+
+1.  Find all referenced ConfigMaps/Secrets of the following:
+    1.  Controller with history (e.g. StatefulSets and DaemonSets): find the
+        ConfigMaps/Secrets referenced in the pod templates of current revision
+        history. 
+    1.  Controller without history (e.g. ReplicaSets and ReplicationControlelrs):
+        find the ConfigMaps/Secrets referenced in its own pod template.
+1.  Check the value of `.controllersAsOwner` field of each such ConfigMap/Secret.
+1.  If the field is set to `true`, add the following to the ConfigMap/Secret's
+    owner list `.metadata.ownerReferences` (if it's not already in the list). If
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/apps/configmap-secret-gc.md
```diff
@@ -0,0 +1,271 @@
...
+type ConfigMap struct {
+  // If true, the supported controllers (Deployments, StatefulSets, DaemonSets,
+  // ReplicaSets, and ReplicationControllers) that reference this ConfigMap will
+  // become its owner. 
+  // Note that garbage collector will delete this ConfigMap once its owners are 
+  // gone, even if this ConfigMap is still referenced by other API resources
+  // (such as Jobs).
+  // This can only be specified at creation time.
+  // Default to false.
+  ControllersAsOwners *bool
```
smarterclayton: This feels yucky and introduces undesirable coupling in the system.  How will you deal with extension resources?  This won't work for extension controllers, which is a red flag.

i don't really like the idea of controllers being special on config maps.

contributors/design-proposals/apps/configmap-secret-gc.md
```diff
@@ -0,0 +1,271 @@
...
+history (ReplicaSets) can declare ConfigMaps/Secrets ownership directly, we
+don't need to implement it in Deployment controller code. Deployment controller
+delegates the work to ReplicaSet controller. 
+
+#### On Controller API Resource Creation
+
+When a controller API object (with or without history) exists, the controller
+will do the following:
+
+1.  Find all referenced ConfigMaps/Secrets of the following:
```
smarterclayton: Somewhat of a red flag in that it gives those controllers access to all secrets in the system.  

It would be better to have a controller that copies config maps and then makes mutations to controllers, without controllers being aware of it (for that particular problem)

contributors/design-proposals/apps/configmap-secret-gc.md
```diff
@@ -0,0 +1,271 @@
...
+relevant to the garbage collection design. 
+
+**Pros:**
+
+*   ConfigMap/Secret references are updated automatically.
+
+**Cons:**
+
+*   It becomes tricky for controllers (e.g. Deployments) to compare its pod
+    template against its history (e.g. ReplicaSets) to find current history
```
smarterclayton: Have something mutate the deployment - which avoids this problem.  Alternatively, make the deployment tolerate refinement of its ReplicaSets by admission and initializers (this is inevitable - image resolution is one concrete example we'll get asked to do).

contributors/design-proposals/apps/configmap-secret-gc.md
```diff
@@ -0,0 +1,271 @@
...
+    (ReplicaSet). 
+*   Rollbacks trigger updates to original ConfigMaps. All consumers of a
+    ConfigMap/Secret are then forced to do rollbacks together.
```
smarterclayton: I think this section is the right approach - lift the responsibility for cloning CM/S up into a top level controller and have it selectively mutate top level controllers as requested.  Define a linkage from a controller object back to a cm template, and make the new controller able to work with many different API types (possibly by using unstructured json path on annotations).

contributors/design-proposals/apps/configmap-secret-gc.md
soltysh: I agree with Clayton here. If we add field to object that means every time we'll be wanting to extend this mechanism with new resources that do not have any controller-controlled history we'll be changing APIs of those. Why not just having annotation, which would make an object eligible for GC. This will work fine with every type of object, including CRDs, etc.

soltysh: I was even thinking this could be a nice way for users to get their resources garbage collected if we combine this annotation with owerRef on each object. All we'd need to do is to enable users setting them easily. I doubt there's such a way currently. Obviously, where possible the controllers should handle those situation automatically. Then adding new resources should be pretty straightforward.

contributors/design-proposals/apps/configmap-secret-gc.md
enisoc: One of the goals of CM rollout is that it should become possible to roll back a Deployment, and have it revert to the exact, known-good combination of CM revision and Pod Template.

Once a Deployment is rolled back, how will the top-level "CM update" controller know it shouldn't bump it back to the newest CM revision?

contributors/design-proposals/apps/configmap-secret-gc.md
enisoc: If something outside mutates the Deployment, will we need a way to prevent the Deployment controller from seeing the original CM ref before it's resolved to point to a copy?

contributors/design-proposals/apps/configmap-secret-gc.md
enisoc: I like the idea to keep CM/Secret mutation outside the workload controllers if possible. It would be nice, for example, if we could have a controller that does nothing but watch ControllerRevision objects and add ownerRefs to the CMs/Secrets that they reference (if they are opted in to CM GC). However, since ControllerRevision currently only contains opaque data, we would need to modify each controller (DS/StS) to put in a list of CMs/Secrets that each ControllerRevision references.

We're also planning to consult sig-auth to see how concerned they are about such permissions, given that these controllers can already create Pods, which transitively gives them access to (at least read) all Secrets.

contributors/design-proposals/apps/configmap-secret-gc.md
enisoc: > Why not just having annotation, which would make an object eligible for GC.

That sounds interesting. In the past, we've regretted starting things out as annotations if they eventually were destined to be fields. But maybe this is a case where it will always make sense as an annotation, like the ones that Deployment puts on ReplicaSet.

After all, this flag is not really about affecting the ConfigMap itself. It's just a signal to other entities, affecting how they treat the ConfigMap. That sounds more like an annotation than a field to me.

@janetkuo What do you think?

janetkuo: @smarterclayton Extension controllers would need to implement the same logic mentioned in the "controller" section in order to support this feature. The assumption here is that users would keep creating new CMs and then update references in controller pod template. This proposal tries to help users garbage collect those kinds of CMs.

@soltysh @enisoc What does such annotation look like? If it's just a boolean value, it does not affect the design much. We need to allow users to opt in and out of CM GC, and setting a field seems more straightforward than an annotation. A field can be protected by validation as well (you cannot make an annotation immutable). 

By "combining this annotation with owerRef on each object", did you mean switching this field to an annotation, or having the garbage collector check both the annotation and ownerRef to decide if this CM should be GC'd or not? If the former, the design is the same; if the latter, we need to change the behavior of the garbage collector.

contributors/design-proposals/apps/configmap-secret-gc.md
janetkuo: If it's not allowed to give controllers access to secrets in the system, we probably cannot use ownerRefs for CM/S GC anymore.

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1163)

janetkuo:
From [Declarative application management in Kubernetes#ConfigMap and Secret updates](https://docs.google.com/document/d/1cLPGweVEYrVqQvBLJg6sxV-TrE5Rm2MNOBA_cxZP2WU/edit#heading=h.jg1k001mhu7x), [Issues related to Declarative App Management#Facilitate ConfigMap rollouts / management](https://docs.google.com/document/d/1vZCauHDM1g3iVzmM7rSeJ11lqEDdcIOvdzIvNVDPG0w/edit#), and previous discussions in https://github.com/kubernetes/kubernetes/issues/22368, going forward we plan to facilitate configmap and secret rollouts by treating CM/S as immutable, and have client-side tooling to generate unique CM/S names (implemented in https://github.com/kubernetes/kubernetes/pull/49961), update references in the controller pod templates to use the new CM/S, and then trigger a rolling update on the deployment. The missing piece for now is the GC of those CM/S. 

That said, auto updates of CM/S is not something we plan to support.

contributors/design-proposals/apps/configmap-secret-gc.md
smarterclayton: You need to turn off CM update if you want to roll back.  You were going to have to do that anyway, because the controller for cm update could bump you at any time resulting in a new deployment.  If you want to roll the state back to a previous version, you either revert the config (in which case this approach works the same as before) or you disable trigger of the configmap and then rollback.

Another advantage of this approach - the deployment is exactly exportable (to a known good state) whereas in the other approach the user would have to export the deployment and the replica set.

contributors/design-proposals/apps/configmap-secret-gc.md
smarterclayton: I think the trigger controller (which is what I would call the thing acting on the resource) is the one who is resolving refs.  That means the none of the existing controllers change.

I partially am advocating for this approach because I have direct familiarity with it from several years of working on the openshift image triggering mechanism.  The mechanism is:

1. a resource exposes a "latest state value" (an image stream tag, which points to a concrete image by digest)
2. a controller that reconciles the latest state value with the status of a deployment

https://github.com/openshift/origin/blob/master/pkg/image/controller/trigger/image_trigger_controller.go
https://github.com/openshift/origin/blob/master/pkg/image/trigger/annotations/annotations.go
https://github.com/openshift/origin/blob/master/pkg/oc/cli/cmd/set/triggers.go#L809

The user sets an annotation on any object (deployment, stateful set, daemon set, replica set) pointing to an image stream tag (roughly `foo/bar:latest` which resolves to `sha256:abc`).  The controller is level driven and ensures that all resources with that annotation have the correct resolved value in their config.  The mechanism can be paused by updating the annotation.  Otherwise, any watched resource can be mutated.

The key advantages are that this can be extended to any object (it could be used for an etcd operator, a new workload controller, etc) just by adding the annotation.  The downside is that the object is under automatic control.  However, this is similar to the autoscaler - a controller is programming down onto the object and the object is unaware of its programming.

A similar model for secret and config maps would be:

1. support an annotation on the deployment that indicates it wishes to receive the latest copy of a configmap, and give a volume name (or jsonpath, or other target) to replace
2. the controller watches the configmap and ensures
    1. a versioned configmap exists for every targeted configmap
    2. every resource referencing that versioned configmap is pointing to the latest copy of the configmap

Rollback is accomplished by reverting the old configmap, pausing the trigger on a per resource basis, or removing the annotation.  Same advantage exists here - no changes to existing controllers is required, all logic is centralized, and all extension / CRD objects can benefit from this.

The per controller approach has merits, but I think it's biggest weakness is that it requires a large amount of duplicate code that can be better centralized.  Annotations are annoying - but if necessary we could introduce future changes that make this a formal field on an object (it's unlikely a common schema exists), or we could place it on a config map.

smarterclayton: The controller model for this is roughly what is described by @janetkuo down below, just automated.  I believe it's a natural next step to automate (we had been planning on doing just that for other reasons), which makes me lean towards GC mechanisms that align with that model.

contributors/design-proposals/apps/configmap-secret-gc.md
kow3ns: From the [docs](https://kubernetes.io/docs/concepts/configuration/secret/#risks), if you can create a Pod that uses the secret you can access the secret. This seems transitive through  controllers today. If I can create a Deployment that creates Pods that can access the secret I can access the secret. There is nothing to stop a user from creating a Deployment that reads a secret and serves it via http. Giving a controller the ability to mutate the secret adds an additional risk to integrity (of the secret itself not the data its protecting), but you could (in some cases) address this by creating a secret per Deployment. We probably don't want to give controllers delete permissions. That defeats the purpose of the GC, and is an unnecessary risk to availability.

contributors/design-proposals/apps/configmap-secret-gc.md
kow3ns: How does GC work in this case? The solution @janetkuo is proposing ensures that the owner references of the config objects point to the objects that implement history for a controller. As long as the config is within the revision history limit you can be assured that it still exists and that you can roll back to it. Also, how do you ensure that the config injection controller updates the configuration of the ReplicaSet prior to the ReplicaSet controller materializing Pods that point to a nil configuration.

contributors/design-proposals/apps/configmap-secret-gc.md
kow3ns: The benefit of the model that @janetkuo proposes is that the entire configuration is encapsulated in the config map itself. For any of the controllers, if you point them to a previous config map or secret, they will detect the identical previous state, and perform a minimally disruptive rollback. If you export the Deployment, with a reference to the config map, and reapply it, you will get the same behavior. Also, both approaches can be used in conjunction with kubectl rollback. I think the proposed approach is actually more flexible.

janetkuo:
Current design can be implemented in a backward compatible way. Therefore, this should not block workloads controller GA (1.9).  To finalize configmap/secret GC design, we also need more concrete Declarative App Management design.

contributors/design-proposals/apps/configmap-secret-gc.md
bgrant0607: This case is mostly needed for ConfigMap and Secret, which are "consumed" by other resources.

I don't like the name of the field, but the idea is that referrers should take ownership -- the resource should be GCed when no longer needed. It's actually a hint to consumers. It could be an annotation, but that wouldn't be very discoverable.

If we had split out the PodTemplate from controllers and used it for controller history universally, this would be easier, and could work OOTB for extensions.

As it is, this does work for extensions, but requires extension implementors to do some work.

bgrant0607: BTW, I believe this hint would be useful even without the history issue.

contributors/design-proposals/apps/configmap-secret-gc.md
bgrant0607: We deliberately don't enforce hard referential integrity.

kubectl currently doesn't create/update resources in any particular order, and controllers may observe changes out of order even if it did. 

However, I would like controllers, such as Deployment and ReplicaSet, to behave gracefully in the absence of dependencies, such as ConfigMap and Secret. For example, possibly they shouldn't create lots of pods that can't function.

If a resource can't properly function due to a nonexistent dependent resource, I recommend reporting that as the reason the resource isn't fully functional in status. 

This would require at least being able to confirm the resource's existence or lack thereof.

bgrant0607: If we're concerned about permissions, we could create a subresource for updating ownerReferences, finalizers, and possibly other metadata in order to facilitate lifecycle management without privilege escalation.

contributors/design-proposals/apps/configmap-secret-gc.md
enisoc: Since it's only a hint (true means it's ok to do it, but doesn't guarantee anything), and the thing it allows is for the object to be deleted when it's no longer referenced, how about `AllowRefCount`?

contributors/design-proposals/apps/configmap-secret-gc.md
enisoc: That sounds interesting... filed #54269.

contributors/design-proposals/apps/configmap-secret-gc.md
bgrant0607: I agree with @kow3ns. Whether the references are updated by a controller or by a non-continuously running client, the GC implementation is orthogonal and the same.

bgrant0607: Updating the image field to point at the digest in the deployment pod template is analogous to updating references to ConfigMaps and Secrets.

yliaog:
cc @yliaog

liggitt:
if this is still active, it should be converted to a KEP in the enhancements repo

/close

PR #1165, opened 2017-10-09
add draft exit code to kubectl-conventions

shiywang:
@fabianofranz @smarterclayton @pwittrock @kubernetes/sig-cli-maintainers

### DIFF
--- a/contributors/devel/kubectl-conventions.md
+++ b/contributors/devel/kubectl-conventions.md
@@ -15,6 +15,7 @@ Updated: 3/23/2017
   - [Documentation conventions](#documentation-conventions)
   - [kubectl Factory conventions](#kubectl-Factory-conventions)
   - [Command implementation conventions](#command-implementation-conventions)
+  - [Exit code conventions](#exit-code-conventions)
   - [Generators](#generators)
 
 <!-- END MUNGE: GENERATED_TOC -->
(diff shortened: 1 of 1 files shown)


contributors/devel/kubectl-conventions.md
```diff
@@ -375,6 +376,21 @@ this is not a mandatory structure and not every command is implemented this way,
 but this is a nice convention so try to be compliant with it. As an example,
 have a look at how [kubectl logs](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/logs.go) is implemented.
 
+## Exit code conventions
+
+Generally, for all the commands exit code, result of `zero` means succeed and `non-zero` means error.
+
+For some declarative commands like `apply` may have a situation which command exited success but without actual changes, we provide a flag `--error-unchanged` to specify the exit code to 2 if necessary.
```
liggitt: For idempotent operations, requiring no changes *is* success (apply is a "make it so" command… if it is already so, a non-zero exit code makes sense)

contributors/devel/kubectl-conventions.md
```diff
@@ -266,7 +267,7 @@ rather than "RESOURCE" or "KIND"
 
 * Use "NAME" for resource names
 
-## kubectl Factory conventions
+## Kubectl Factory conventions
```
smarterclayton: This is unnecessary, it's acceptable to use `kubectl` lower case at the beginning of a title when referring to a CLI command.

contributors/devel/kubectl-conventions.md
smarterclayton: Yeah, please rephrase as Jordan mentioned.  Idempotent ("make-it-so") commands should return 0 when successful (even if no changes were provided).  For commands that are not explicitly expected to be idempotent, consumers can request treating "already-so" as "make-it-so" via flag.

contributors/devel/kubectl-conventions.md
```diff
@@ -375,6 +376,21 @@ this is not a mandatory structure and not every command is implemented this way,
...
+For some declarative commands like `apply` may have a situation which command exited success but without actual changes, we provide a flag `--error-unchanged` to specify the exit code to 2 if necessary.
+
+For other commands like `create` that return error when unchanged, we have the flag `--ignore-unchanged` to make it return 0 and no errors.
+
+
+| Exit Code Number | Meaning | Enable |
+| :---             | :---    | :---    |
+| 0                | Command exited success  | By default, By flag `--ignore-unchanged` |
+| 1                | Command exited for general errors  | By default |
+| 2                | Command exited success but without changes | By flag `--error-unchanged`|
```
smarterclayton: Probably rephrase this to `Command was successful, but the user requested a distinct exit code when no change was made`.

Also, we shouldn't be using 2 which is reserved as a bash shell error.  We should use 3 or higher.  See http://www.tldp.org/LDP/abs/html/exitcodes.html

contributors/devel/kubectl-conventions.md
shiywang: combine two comments above then add a little redundancy and parallelism, ptal

shiywang:
combine two comments above then add a little redundancy and parallelism, @liggitt @smarterclayton @fabianofranz ptal

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1165)

mengqiy:
@shiywang Have you implement this?

shiywang:
@mengqiy here is my pr: https://github.com/kubernetes/kubernetes/pull/52647

PR #1166, opened 2017-10-09
Proposal: Allowing context to be used with client-go

maleck13:
This is a first pass at a set of changes aimed at allowing consumers of client-go to leverage functionality offered by the ```context```  package in the std lib. It is in response to the following issue:
https://github.com/kubernetes/kubernetes/issues/46503

**To the reviewers:**
This is the first proposal I have done for Kubernetes so any feedback and guidance would be appreciated, I am not really sure of the process after creating the proposal.

### DIFF
--- a/contributors/design-proposals/api-machinery/client-go-context.md
+++ b/contributors/design-proposals/api-machinery/client-go-context.md
@@ -1,0 +1,153 @@
+# Using context.Context with client-go
+
+* Authors: @maleck13
+
+## Abstract
+
+This proposal aims to outline how to allow using the [context package](https://golang.org/pkg/context/) with [client-go](https://github.com/kubernetes/client-go) to allow propagation of cancellation and timeouts 
+and also to allow exposing context based http tracing [go blog on context based httptrace](https://blog.golang.org/http-tracing) to consumers of the client.
+
+## Motivation and Goals
+
+When using client-go, external calls to the Kubernetes API are made in order to manage resources. The initiator of these calls may find due to an external event, 
+that they no longer need the resource they had requested, but they have no means by which to inform the client of this change. 
+
+When using client-go, having the ability to add context based http tracing would be valuable for issues around performance, observability and debugging. 
+
+- Allow consumers of the client to indicate, in an idiomatic way, that the action that caused them to invoke the client has been cancelled.
+- Allow consumers of the client to instrument and gain insight via context based http tracing.
+
+## Non Goals
+
+- Introduce a transactional system for write operations whereby upon cancellation the client would somehow rollback any current write operations.
+
+## API
+
+There are two approaches outlined. The first is a non breaking change that appends to the existing API for resources but perhaps puts more burden on the client consumer. 
+The second is a breaking change to the client-go api, that is potentially more idiomatic and clear.
+ 
+Common to both is the fact that the request type in client-go  already exposes a ```Context```  function: [request context set on client-go rest.Request](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L393). 
+If set, this context is passed all the way through to the underlying ```http.Request```: [request context passed to http request code](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L484). 
+So to achieve the goals, we need only modify the client API to allow passing in an external context that is then used to set the existing context on the request type.
+
+## Chosen Implemenation
+
+### Breaking change
+
+The chosen option is to pass the context with each action on a resource. This may be more idiomatic and explicit: 
+
+```go 
+func (c *pods) Get(ctx context.Context, name string, options meta_v1.GetOptions) (result *v1.Pod, err error) {
+	result = &v1.Pod{}
+	err = c.client.Get().
+		Namespace(c.ns).
+		Resource("pods").
+		Name(name).
+		VersionedParams(&options, scheme.ParameterCodec).
+		Context(c.ctx).
+		Do().
+		Into(result)
+	return
+}
+```
+
+These change would impact on every API. Although the changes required would be many, the simplest approach to maintaining current behavior would be to pass ```context.TODO()```.
+
+For the ```listers``` and ```informers``` the change outlined above (adding new constructors) would still allow a context to be passed through to client functions doing the work.
+
+**Concerns**
+
+This is very clearly a breaking change that would likely require extensive changes to both tests and implementation code throughout the impacted code bases.
+
+**Initial prototype PR**
+https://github.com/kubernetes/kubernetes/pull/58677
+
+
+## Alternatives considered and rejected
+
+- mailing list discussion https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-apps/xe_ii1iC1r0/oeK0g3ABAwAJ
+
+- twitter poll https://twitter.com/the_sttts/status/965540104470188032
+
+### Non breaking change
+
+The client resources (Pods, Secrets etc...) each expose a resource specific interface ```PodInterface``` for example. This is returned from a ```Getter``` interface:
+
+```go
+type PodsGetter interface {
+	Pods(namespace string) PodInterface
+}
+``` 
+
+If we were to modify these resource interfaces and their underlying concrete types ```pod``` in this case, to add a ```WithContext``` method, it would allow consumers of the client to set context while maintaining backwards compatibility for all other consumers of the API.
+
+```go
+// PodInterface has methods to work with Pod resources.
+type PodInterface interface {
+  	WithContext(ctx context.Context) PodInterface
+    ...
+}
+
+type pods struct {
+	client rest.Interface
+	ns     string
+	ctx    context.Context
+}
+
+// WithContext allows you to set a context that will be used by the underlying http request
+func (c *pods) WithContext(ctx context.Context) PodInterface {
+  	c.ctx = ctx
+  	return c
+}
+
+```
+
+To pass through this context, it would be necessary to change the underlying client calls to accept the context. Example:
+
+```go 
+func (c *pods) Get(name string, options meta_v1.GetOptions) (result *v1.Pod, err error) {
+	result = &v1.Pod{}
+	err = c.client.Get().
+		Namespace(c.ns).
+		Resource("pods").
+		Name(name).
+		VersionedParams(&options, scheme.ParameterCodec).
+		Context(ctx).
+		Do().
+		Into(result)
+	return
+}
+```
+
+This would result in an API that would be interacted with like so if the consumer wished to pass a context:
+```go  
+   ctx := req.Context()
+   pod, err := k8client.CoreV1().Pods(namespace).WithContext(ctx).Get(podName, ...)
+```
+
+
+HTTP Tracing and cancellation are also likely useful when using ```informers``` and ```listers```, as these also use the client interfaces backwards compatibility can be maintained by adding a new set of constructors for adding a
+specific context object:
+
+```go
+func NewConfigMapInformerWithContext(client kubernetes.Interface, namespace string, resyncPeriod time.Duration, indexers cache.Indexers, ctx context.Context) cache.SharedIndexInformer {}
+
+func (f *configMapInformer) InformerWithContext(ctx context.Context) cache.SharedIndexInformer {}
+
+func defaultConfigMapInformerWithContext(client kubernetes.Interface, resyncPeriod time.Duration, ctx context.Context) cache.SharedIndexInformer {
+	return NewConfigMapInformerWithContext(client, meta_v1.NamespaceAll, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},ctx)
+}
+
+func (f *configMapInformer) ListerWithContext(ctx context.Context) v1.ConfigMapLister {
+	return v1.NewConfigMapLister(f.InformerWithContext(ctx).GetIndexer())
+}
+```
+
+**Concerns / Notes**
+- The idea behind using a context with a request is it is only meant to live as long as the initial request that created it. If we expose an API like the one outlined, the consumer would need to know that they needed to call
+```WithContext(ctx context.Context)``` with each request context.
+
+- The fakes would need to be updated and new tests added to cover the new APIs.
+
+- The number of entry points and maintenance points is increased. 
+


jbeda:
It would be best for someone from API Machinery to comment here.

@kubernetes/sig-api-machinery-pr-reviews

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1166)

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,141 @@
...
+
+type pods struct {
+	client rest.Interface
+	ns     string
+	ctx    context.Context
+}
+
+// WithContext allows you to set a context that will be used by the underlying http request
+func (c *pods) WithContext(ctx context.Context) PodInterface {
+  	c.ctx = ctx
```
sttts: `WithContext` should not mutate, but has to create a shallow copy first.

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+
+These change would impact on every API. Although the changes required would be many, the simplest approach to maintaining current behaviour would be to pass in nil or ```context.ToDo()```.
+
+For the ```listers``` and ```informers``` the change outlined above (adding new constructors) would still allow a context to be passed through to client functions doing the work.
+
+**Concerns**
+
+This is very clearly a breaking change that would likely require extensive changes to both tests and implementation code throughout the impacted code bases.
```
sttts: I don't care much about everything in `k8s.io/kubernetes`. It's a one-time cleanup to adapt using `context.ToDo()`. 

Passing a context through where it makes sense would be a bigger change, but that's not dependent on which of your solution we choose.

It's also a breaking change for our published client-go. We actually had similar changes in 1.6 with GetOptions and ListOptions. It might be tedious for our 3rdparty users to fix their code. But at least the breakage and fix is obvious and easy to repair.

contributors/design-proposals/api-machinery/client-go-context.md
maleck13: Could you explain a little more, do you mean before the assignment on line 60 happens? My understanding is ctx is being passed by value into this method and so is a shallow copy or am I misunderstanding?

contributors/design-proposals/api-machinery/client-go-context.md
maleck13: Agree that is it is relatively easy to fix. Also clearly announces context support in the client. With the none breaking change, a consumer would need to discover that context support was now available.

ggaaooppeenngg:
Any further work?

maleck13:
@ggaaooppeenngg I have not put any additional work in yet. Not entirely sure what the next steps would be

smarterclayton:
This seems pretty clear.  I think the breaking change is the correct long
term behavior (consistent with the intended use of the context package).
However, it will impact a wide swath of the ecosystem, so we would need to
be very careful about communicating the changes and their impact, as well
as ensure that concerns raised by the community are addressed.

Are other clients in the Go ecosystem actively moving towards context in
its "intended" form, or is it a slower process?  I admit to not seeing a
large uptake, but haven't done a real survey.

maleck13:
@sttts You mentioned a similar change when adding the ```GetOptions``` and ```ListOptions``` are you aware of how this change was communicated?

@smarterclayton pkgs in the stdlib are making use of the context pkg, but I am also unfamiliar with the wider uptake in community pkgs.

sttts:
> You mentioned a similar change when adding the GetOptions and ListOptions are you aware of how this change was communicated?

As most of our breaking changes not very well https://github.com/kubernetes/client-go/blob/master/CHANGELOG.md#v300. But at least it's pretty obvious for every 3rdparty what changed and how to fix it. I worry more about changes which have a less clear local fix. Contexts will belong to the former.

maleck13:
Ok so is the next step here then to do an actual implementation or is there more that needs doing on the proposal?

contributors/design-proposals/api-machinery/client-go-context.md
smarterclayton: On the other hand, this approach would guarantee that we update all call sites.  The other we could silently miss them.

maleck13:
I am hoping to start on an implementation of this early in the new year after the holiday period.

contributors/design-proposals/api-machinery/client-go-context.md
sttts: I think I meant a shallow copy of `c *pods`.

contributors/design-proposals/api-machinery/client-go-context.md
liggitt: silently missing call site updates seems bad

maleck13:
@sttts should we consider adding ctx support to the client-gen tool also?

sttts:
> should we consider adding ctx support to the client-gen tool also?

We have to, don't we?

maleck13:
@sttts Yes of course we do, ignore the previous comment

maleck13:
started looking at the implementation for this, new to the client generator so it may take a little while

lavalamp:
cc @yliaog

sttts:
> I will create the WIP pr in the morning.

Sure, assign it to me.

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,141 @@
...
+		Name(name).
+		VersionedParams(&options, scheme.ParameterCodec).
+		Context(c.ctx).
+		Do().
+		Into(result)
+	return
+}
+```
+
+These change would impact on every API. Although the changes required would be many, the simplest approach to maintaining current behaviour would be to pass in nil or ```context.ToDo()```.
```
enisoc: FWIW, passing a `nil` Context is an explicit anti-pattern. Only `context.TODO()` should be used for this purpose.

https://golang.org/pkg/context/

lavalamp: +1

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,141 @@
...
+## API
+
+There are two approaches outlined. The first is a non breaking change that appends to the existing API for resources but perhaps puts more burden on the client consumer. 
+The second is a breaking change to the client-go api, that is potentially more idiomatic and clear.
+ 
+Common to both is the fact that the request type in client-go  already exposes a ```Context```  function: [request context set on client-go rest.Request](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L393). 
+If set, this context is passed all the way through to the underlying ```http.Request```: [request context passed to http request code](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L484). 
+So to achieve the goals, we need only modify the client API to allow passing in an external context that is then used to set the existing context on the request type.
+
+### Non breaking change
```
deads2k: move this section to an "alternatives considered an rejected.  Link to the mailing list discussion.

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,141 @@
...
+- The idea behind using a context with a request is it is only meant to live as long as the initial request that created it. If we expose an API like the one outlined, the consumer would need to know that they needed to call
+```WithContext(ctx context.Context)``` with each request context.
+
+- The fakes would need to be updated and new tests added to cover the new APIs.
+
+- The number of entry points and maintenance points is increased. 
+
+### Breaking change
+
+The second option is to pass the context with each action on a resource. This may be more idiomatic and explicit:
```
deads2k: This isthe option we chose

deads2k:
Reorganize a bit, link to the google group discussion, and I think this is good.

maleck13:
thanks will update shortly

sttts:
And our highly scientific twitter poll: https://twitter.com/the_sttts/status/965540104470188032, 70% vote for breaking change.

maleck13:
updated

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,149 @@
...
+## API
+
+There are two approaches outlined. The first is a non breaking change that appends to the existing API for resources but perhaps puts more burden on the client consumer. 
+The second is a breaking change to the client-go api, that is potentially more idiomatic and clear.
+ 
+Common to both is the fact that the request type in client-go  already exposes a ```Context```  function: [request context set on client-go rest.Request](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L393). 
+If set, this context is passed all the way through to the underlying ```http.Request```: [request context passed to http request code](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L484). 
+So to achieve the goals, we need only modify the client API to allow passing in an external context that is then used to set the existing context on the request type.
+
+## Alternatives considered and rejected
```
sttts: nit: I would move the chosen design above this section.

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,149 @@
...
+		Name(name).
+		VersionedParams(&options, scheme.ParameterCodec).
+		Context(c.ctx).
+		Do().
+		Into(result)
+	return
+}
+```
+
+These change would impact on every API. Although the changes required would be many, the simplest approach to maintaining current behavior would be to pass ```context.ToDo()```.
```
sttts: nit: TODO()

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,149 @@
...
+
+For the ```listers``` and ```informers``` the change outlined above (adding new constructors) would still allow a context to be passed through to client functions doing the work.
```
sttts: Should we add a link to the prototype PR?

maleck13: sounds reasonable

cceckman:
Gentle ping. Is this waiting on anything? (Attention from reviewers?)

sttts:
We decided in the sig-api-machinery meeting to go forward with a hard one-time signature change. PRs are in the works right now.

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+This would result in an API that would be interacted with like so if the consumer wished to pass a context:
+```go  
+   ctx := req.Context()
+   pod, err := k8client.CoreV1().Pods(namespace).WithContext(ctx).Get(podName, ...)
+```
+
+
+HTTP Tracing and cancellation are also likely useful when using ```informers``` and ```listers```, as these also use the client interfaces backwards compatibility can be maintained by adding a new set of constructors for adding a
```
liggitt: cancellation via deadline or timeout in a context would not work well with a single context provided to a long-lived lister/informer. similarly, trace ids are typically per request, not long-lived. I don't think a single context object for a lister/informer is what we want.

liggitt:
I'm in favor of this, other than the lister/informer bits

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,153 @@
...
+
+```go 
+func (c *pods) Get(name string, options meta_v1.GetOptions) (result *v1.Pod, err error) {
+	result = &v1.Pod{}
+	err = c.client.Get().
+		Namespace(c.ns).
+		Resource("pods").
+		Name(name).
+		VersionedParams(&options, scheme.ParameterCodec).
+		Context(ctx).
```
rlguarino: is this line meant to be: `Context(c.ctx)` ?

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,153 @@
...
+
+```go 
+func (c *pods) Get(ctx context.Context, name string, options meta_v1.GetOptions) (result *v1.Pod, err error) {
+	result = &v1.Pod{}
+	err = c.client.Get().
+		Namespace(c.ns).
+		Resource("pods").
+		Name(name).
+		VersionedParams(&options, scheme.ParameterCodec).
+		Context(c.ctx).
```
rlguarino: Is this line meant to be: `Context(ctx)`?

contributors/design-proposals/api-machinery/client-go-context.md
```diff
@@ -0,0 +1,153 @@
...
+		Do().
+		Into(result)
+	return
+}
+```
+
+These change would impact on every API. Although the changes required would be many, the simplest approach to maintaining current behavior would be to pass ```context.TODO()```.
+
+For the ```listers``` and ```informers``` the change outlined above (adding new constructors) would still allow a context to be passed through to client functions doing the work.
```
rlguarino: Would it make more sense in the context of `listers` and `informers` to specify a context constructor function?

Some sample use cases for context like tracing and deadlines, won't make much sense to set once per `informer` or `lister`, but would make more sense to create a new one every time its needed.

Trace id are commonly set per request and the only thing a deadline set at the construction of the `informer` or `lister` would do is limit the time your program can get resources from the api.

contributors/design-proposals/api-machinery/client-go-context.md
rlguarino: I'd be more concerned with third-parties scrambling to fix their build and misusing context without. If we silent'y failed back to `context.ToDo()` people could add context on their own initiative.

aanm:
> We decided in the sig-api-machinery meeting to go forward with a hard one-time signature change. PRs are in the works right now.

@sttts Any news with regarding this?

hypnoce:
This proposal would be a great help for my use case. Currently writing a small service that calls the kubernetes API server. To workaround the problem, I have to create a new kubernetes client with a custom transport every time there is an incoming request.
While it's possible to optimize memory allocation by only creating the client for the needed api group (corev1, extensionv1...), the code still allocates a bunch of resource.

liggitt:
This should move to https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery in [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/#kubernetes-enhancement-proposals-keps) form to make progress.

gkarthiks:
hi @maleck13 Since we have a KEP process, can we please merge this soon or close this and create a KEP?

liggitt:
Moved to https://github.com/kubernetes/enhancements/pull/1503

PR #1167, opened 2017-10-09
update links for container identity working group

ericchiang:
Add a link to the agenda to the working group README.

cc @destijl @smarterclayton @liggitt

PR #1169, opened 2017-10-09
Fix wg-container-identity/README.md

cblecker:
#1167 updated `wg-container-identity/README.md` instead of updating sigs.yaml. This fixes that and regenerates the doc.


ericchiang:
@cblecker ah my bad.

/lgtm

PR #1171, opened 2017-10-10
Proposal for Lifecycle Hooks

tnozicka:
cc: @mfojtik @kargakis @smarterclayton 

/xref
RFC Issue: https://github.com/kubernetes/kubernetes/issues/14512
Previous proposal: https://github.com/kubernetes/kubernetes/pull/33545

### DIFF
--- a/contributors/design-proposals/lifecycle-hooks.md
+++ b/contributors/design-proposals/lifecycle-hooks.md
@@ -1,0 +1,164 @@
+# Lifecycle Hooks
+
+**Author**: @tnozicka
+
+**Status**: Proposal
+
+**RFC Issue**: https://github.com/kubernetes/kubernetes/issues/14512
+
+## Abstract
+The intent of this proposal it to support lifecycle hooks in all objects having rolling/recreate update; currently those are Deployments, StatefulSets and DaemonSets. 
+Lifecycle hooks should allow users to better customize updates (rolling/recreate) without having to write their own controllers. 
+
+## Lifecycle Hooks
+Lifecycle hook is a Job that will get triggered by the update reaching certain progress point (like 50%). 
+The progress point is determined by the ratio of new.availableReplicas to the declared replicas or by explicitly stating the number of new.availableReplicas.
+
+### Previous Implementations
+Lifecycle hooks are already implemented in OpenShift but we want to enhance them and implement them natively in Kubernetes.
+ - [OpenShift proposal](https://github.com/openshift/origin/blob/master/docs/proposals/post-deployment-hooks.md)
+ - [OpenShift docs](https://docs.openshift.org/latest/dev_guide/deployments/deployment_strategies.html#lifecycle-hooks)
+
+### Previous Kubernetes Proposals
+ - https://github.com/kubernetes/kubernetes/pull/33545
+
+### Use Cases
+The most common use case for lifecycle hooks is to have pre, mid and/or post hook. It is mostly used to run some kind of acceptance check in the middle and/or at the end to fully verify the update is working as expected and rollback if it isn't. 
+The acceptance check may be time consuming and more thorough than what readiness and liveness probes are intended for. 
+You can also notify external services from them, migrate database in the middle of an update, send messages to IRC channel or do anything else.
+
+[A short demo](https://youtu.be/GVNTm_K43iI) simulating lifecycle hooks using auto-pausing that has been presented at SIG-Apps meeting on August 21, 2017.
+
+#### Reusability
+If you are a big shop and you are running several instances of e.g. your database you want to reuse definition of lifecycle hooks e.g. for several instances of your database. 
+This is reflected in the design bellow by having separate object to define lifecycle hooks and reference it from the objects.
+
+If you would be worried about having shared definitions so e.g. your mistakes won't spread too much you can always choose not to share those definitions and reference unique lifecycle hook objects from every instance. 
+
+## API Objects
+### New Objects
+```go
+type LifecycleTemplate struct {
+    TypeMeta
+    ObjectMeta
+    Spec LifecycleTemplateSpec
+}
+
+type LifecycleTemplateSpec struct {
+    // "Always" - keep all Jobs 
+    // "OnFailure" - keep only failed Jobs
+    // "Never" - delete Jobs immediately 
+    RetainPolicy string
+    
+    // After reaching this limit the Job history will be pruned.
+    RevisionHistoryLimit *int32
+    
+    // List of lifecycle hooks
+    // Can have multiple hooks at the same ProgressPoint; order independent
+    Hooks []LifecycleHook
+}
+
+type LifecycleHook struct {
+    // Unique name
+    Name string
+    
+    // ProgressPoint specifies the point during update when the hook should be triggered.
+    // Accepts both the number of new pods available or a percentage value representing the ratio
+    // of new.availableReplicas to the declared replicas. In case of getting a percentage
+    // it will try to reach the exact or the closest possible point right after it,
+    // trigger the job, wait for it to complete and then continue. 
+    // If such situation shall occur that two different ProgressPoints should be reached at
+    // the same time, all the hooks for an earlier ProgressPoint will be ran (and finished)
+    // before any later one.
+    ProgressPoint intstr.IntOrString
+    
+    // "Abort" - Failure for this hook is fatal, deployment is considered failed and should be rollbacked.
+    // "Ignore" - Even if the hook fails deployment continues rolling out new version.
+    FailurePolicy string
+    
+    JobSpec v1.JobSpec
+}
+
+type LifecycleHookStatus struct {
+    // Name of the hook
+    Name string
+    
+    ProgressPoint intstr.IntOrString
+    
+    // States: "Running", "Succeeded", "Failed" 
+    State string
+    
+    // Reference to locate the Job created when executing the hook
+    JobRef LocalObjectReference
+}
+
+type LifecycleHookRevisionStatus struct {
+    // Revision of the object that this hook was run for
+    Revision string
+    
+    LifecycleHookStatuses []LifecycleHookStatus
+}
+```
+
+### Affected Objects
+As of now the lifecycle hooks are relevant for Deployments, StatefulSets and DaemonSets. 
+They will require new optional field to be able to reference LifecycleTemplate and extending its status by []LifecycleHookRevisionStatus. 
+Also the controller will need to be slightly adjusted to scale in appropriate chunks and trigger the hooks.
+
+```go
+type DeploymentSpec struct {
+    // ...
+    // Addition
+    LifecycleTemplate *ObjectReference
+}
+
+type DeploymentStatus struct {
+    // ...
+    // Addition
+    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
+}
+```
+
+```go
+type StatefulSetSpec struct {
+    // ...
+    // Addition
+    LifecycleTemplate *ObjectReference
+}
+
+type StatefulSetStatus struct {
+    // ...
+    // Addition
+    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
+}
+
+```
+
+```go
+type DaemonSetSpec struct {
+    // ...
+    // Addition
+    LifecycleTemplate *ObjectReference
+}
+
+type DaemonSetStatus struct {
+    // ...
+    // Addition
+    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
+}
+```
+
+## Algorithm
+If there is a LifecycleTemplate referenced from an object:
+
+1. Calculate next partition point to reach the closest lifecycle hook progress point and scale replicas in update appropriately. If there is no hook remaining, GOTO 5.
+2. When partition point is reached, run the hook by creating a Job using LifecycleHook.JobSpec
+3.
+    1. If the hook failed and FailurePolicy is Abort - emit event, fail the update and initiate rollback (GOTO 5.) 
+    2. If the hook failed and FailurePolicy is Ignore - only emit event
+4. GOTO 1.
+5. Finish
+
+(In case of rollover cancel any hooks running and don't execute new ones.)
+
+This can be applied to all the existing update strategies as that's essentially dependent only on the ability to have a certain ratio of new.availableReplicas to the declared replicas and running the hook between changing that ratio.
--- a/contributors/design-proposals/lifecycle-hooks.md
+++ b/contributors/design-proposals/lifecycle-hooks.md
@@ -7,12 +7,13 @@
 **RFC Issue**: https://github.com/kubernetes/kubernetes/issues/14512
 
 ## Abstract
-The intent of this proposal it to support lifecycle hooks in all objects having rolling/recreate update; currently those are Deployments, StatefulSets and DaemonSets. 
+The intent of this proposal is to support lifecycle hooks in all objects having rolling/recreate update; currently those are Deployments, StatefulSets and DaemonSets. 
 Lifecycle hooks should allow users to better customize updates (rolling/recreate) without having to write their own controllers. 
 
 ## Lifecycle Hooks
 Lifecycle hook is a Job that will get triggered by the update reaching certain progress point (like 50%). 
-The progress point is determined by the ratio of new.availableReplicas to the declared replicas or by explicitly stating the number of new.availableReplicas.
+The progress point is determined by the ratio of new.availableReplicas to the declared replicas or by explicitly stating the number of new.availableReplicas. 
+The API is open for other types of triggering conditions in the future. 
 
 ### Previous Implementations
 Lifecycle hooks are already implemented in OpenShift but we want to enhance them and implement them natively in Kubernetes.
@@ -23,45 +24,24 @@ Lifecycle hooks are already implemented in OpenShift but we want to enhance them
  - https://github.com/kubernetes/kubernetes/pull/33545
 
 ### Use Cases
-The most common use case for lifecycle hooks is to have pre, mid and/or post hook. It is mostly used to run some kind of acceptance check in the middle and/or at the end to fully verify the update is working as expected and rollback if it isn't. 
+The most common use case for lifecycle hooks is to have pre, mid and/or post hook. 
+It is mostly used to run some kind of acceptance check in the middle and/or at the end to fully verify the update is working as expected and rollback if it isn't. 
 The acceptance check may be time consuming and more thorough than what readiness and liveness probes are intended for. 
 You can also notify external services from them, migrate database in the middle of an update, send messages to IRC channel or do anything else.
 
 [A short demo](https://youtu.be/GVNTm_K43iI) simulating lifecycle hooks using auto-pausing that has been presented at SIG-Apps meeting on August 21, 2017.
 
-#### Reusability
-If you are a big shop and you are running several instances of e.g. your database you want to reuse definition of lifecycle hooks e.g. for several instances of your database. 
-This is reflected in the design bellow by having separate object to define lifecycle hooks and reference it from the objects.
-
-If you would be worried about having shared definitions so e.g. your mistakes won't spread too much you can always choose not to share those definitions and reference unique lifecycle hook objects from every instance. 
 
 ## API Objects
 ### New Objects
 ```go
-type LifecycleTemplate struct {
-    TypeMeta
-    ObjectMeta
-    Spec LifecycleTemplateSpec
-}
+type HookTriggerType string
 
-type LifecycleTemplateSpec struct {
-    // "Always" - keep all Jobs 
-    // "OnFailure" - keep only failed Jobs
-    // "Never" - delete Jobs immediately 
-    RetainPolicy string
-    
-    // After reaching this limit the Job history will be pruned.
-    RevisionHistoryLimit *int32
-    
-    // List of lifecycle hooks
-    // Can have multiple hooks at the same ProgressPoint; order independent
-    Hooks []LifecycleHook
-}
+const (
+	ProgressAvailableHookTriggerType HookTriggerType = "ProgressAvailable"
+)
 
-type LifecycleHook struct {
-    // Unique name
-    Name string
-    
+type ProgressAvailableParams struct {
     // ProgressPoint specifies the point during update when the hook should be triggered.
     // Accepts both the number of new pods available or a percentage value representing the ratio
     // of new.availableReplicas to the declared replicas. In case of getting a percentage
@@ -71,51 +51,49 @@ type LifecycleHook struct {
     // the same time, all the hooks for an earlier ProgressPoint will be ran (and finished)
     // before any later one.
     ProgressPoint intstr.IntOrString
-    
-    // "Abort" - Failure for this hook is fatal, deployment is considered failed and should be rollbacked.
-    // "Ignore" - Even if the hook fails deployment continues rolling out new version.
-    FailurePolicy string
-    
-    JobSpec v1.JobSpec
 }
 
-type LifecycleHookStatus struct {
+type HookTrigger struct {
+    // Type of hook trigger. Can be ProgressAvailable. Default is ProgressAvailable.
+    Type HookTriggerType
+    
+    // ProgressAvailable config params. Present only if type = "ProgressAvailable".
+    ProgressAvailableParams *ProgressAvailableParams
+}
+
+type LifecycleHook struct {
     // Name of the hook
     Name string
     
-    ProgressPoint intstr.IntOrString
+    // Triggering config
+    Trigger HookTrigger
     
-    // States: "Running", "Succeeded", "Failed" 
-    State string
+    // "Abort" - Failure for this hook is fatal, deployment becomes perma-failed
+    // "Ignore" - Even if the hook fails deployment continues rolling out new version.
+    FailurePolicy string
     
-    // Reference to locate the Job created when executing the hook
-    JobRef LocalObjectReference
-}
-
-type LifecycleHookRevisionStatus struct {
-    // Revision of the object that this hook was run for
-    Revision string
+    // "Always" - always keep this jobs 
+    // "OnFailure" - keep this Job only if it fails
+    // "Never" - delete this job immediately after it is done 
+    RetainPolicy string
     
-    LifecycleHookStatuses []LifecycleHookStatus
+    JobSpec v1.JobSpec
 }
 ```
 
 ### Affected Objects
-As of now the lifecycle hooks are relevant for Deployments, StatefulSets and DaemonSets. 
-They will require new optional field to be able to reference LifecycleTemplate and extending its status by []LifecycleHookRevisionStatus. 
-Also the controller will need to be slightly adjusted to scale in appropriate chunks and trigger the hooks.
+As of now the lifecycle hooks are relevant for Deployments, StatefulSets and DaemonSets.
+New optional field named `LifecycleHooks` will be added to provide configuration for lifecycle hooks. 
+The related state will be reported using Conditions and details will be available on the created Job objects.
+Also the controllers will need to be adjusted to scale in appropriate chunks and trigger the hooks.
 
 ```go
 type DeploymentSpec struct {
     // ...
     // Addition
-    LifecycleTemplate *ObjectReference
-}
-
-type DeploymentStatus struct {
-    // ...
-    // Addition
-    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
+    // List of lifecycle hooks
+    // Can have multiple hooks at the same ProgressPoint; order independent
+    LifecycleHooks []LifecycleHook
 }
 ```
 
@@ -123,38 +101,30 @@ type DeploymentStatus struct {
 type StatefulSetSpec struct {
     // ...
     // Addition
-    LifecycleTemplate *ObjectReference
+    // List of lifecycle hooks
+    // Can have multiple hooks at the same ProgressPoint; order independent
+    LifecycleHooks []LifecycleHook
 }
-
-type StatefulSetStatus struct {
-    // ...
-    // Addition
-    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
-}
-
 ```
 
 ```go
 type DaemonSetSpec struct {
     // ...
     // Addition
-    LifecycleTemplate *ObjectReference
-}
-
-type DaemonSetStatus struct {
-    // ...
-    // Addition
-    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
+    // List of lifecycle hooks
+    // Can have multiple hooks at the same ProgressPoint; order independent
+    LifecycleHooks []LifecycleHook
 }
 ```
 
 ## Algorithm
-If there is a LifecycleTemplate referenced from an object:
+If there are LifecycleHooks present in an object:
 
 1. Calculate next partition point to reach the closest lifecycle hook progress point and scale replicas in update appropriately. If there is no hook remaining, GOTO 5.
-2. When partition point is reached, run the hook by creating a Job using LifecycleHook.JobSpec
+2. When partition point is reached, run the hook by creating a Job using LifecycleHook.JobSpec. 
+(Controller is allowed to run more hooks for the same progress point in parallel but has to wait for all of them to finish before it can move to the next progress point.)
 3.
-    1. If the hook failed and FailurePolicy is Abort - emit event, fail the update and initiate rollback (GOTO 5.) 
+    1. If the hook failed and FailurePolicy is Abort - emit event, permanently fail the update (GOTO 5.) 
     2. If the hook failed and FailurePolicy is Ignore - only emit event
 4. GOTO 1.
 5. Finish
@@ -162,3 +132,21 @@ If there is a LifecycleTemplate referenced from an object:
 (In case of rollover cancel any hooks running and don't execute new ones.)
 
 This can be applied to all the existing update strategies as that's essentially dependent only on the ability to have a certain ratio of new.availableReplicas to the declared replicas and running the hook between changing that ratio.
+
+Hooks (created Jobs) are tracked using ownerRefs and by being labeled with a revision they were created for. By having an ownerRef to particular controllerRevision (or RS) pruning will propagate there.
+
+## Injecting deployment information into hooks
+To help scripts make more informed decisions following environment variables will be injected into hook jobs.
+
+`K8S_ROLLOUT_FROM_OBJECT` and `K8S_ROLLOUT_TO_OBJECT` will allow scripts to use kubectl or API to get more details.
+```bash
+K8S_ROLLOUT_FROM_OBJECT="namespace/name"
+K8S_ROLLOUT_TO_OBJECT="namesapce/name"
+```
+
+`K8S_ROLLOUT_TYPE` will help scripts to easily detect a rollback.
+```
+K8S_ROLLOUT_TYPE=[ROLLOUT,ROLLBACK]
+```
+
+In case any of these environment variables is already present on the object, it won't be set by the controller.


castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1171)

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,157 @@
...
+
+type DaemonSetStatus struct {
+    // ...
+    // Addition
+    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
+}
+```
+
+## Algorithm
+If there is a LifecycleTemplate referenced from an object:
```
kargakis: Is this an owner reference?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,157 @@
+# Lifecycle Hooks
+
+**Author**: @tnozicka
+
+**Status**: Proposal
+
+**RFC Issue**: https://github.com/kubernetes/kubernetes/issues/14512
+
+## Abstract
+The intent of this proposal it to support lifecycle hooks in all objects having rolling/recreate update; currently those are Deployments, StatefulSets and DaemonSets. Lifecycle hooks should allow users to better customize updates (rolling/recreate) without having to write their own controllers.
```
kargakis: Can you please split these into separate lines? Otherwise, it's difficult to review.

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,157 @@
...
+type LifecycleHookRevisionStatus struct {
+    // Revision of the object that this hook was run for
+    Revision string
+    
+    LifecycleHookStatuses []LifecycleHookStatus
+}
+```
+
+### Affected Objects
+As of now the lifecycle hooks are relevant for Deployments, StatefulSets and DaemonSets. They will require new optional field to be able to reference LifecycleTemplate and extending its status by []LifecycleHookRevisionStatus. Also the controller will need to be slightly adjusted to scale in appropriate chunks and trigger the hooks.
```
kargakis: So this proposal is about extending the existing controllers to run the hooks?

kargakis:
@kubernetes/sig-apps-proposals

castrojo:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/community/1171)

contributors/design-proposals/lifecycle-hooks.md
tnozicka: there is an explicit reference from DeploymentSpec, StatefulSetSpec and DaemonSetSpec - `LifecycleTemplate *ObjectReference`

contributors/design-proposals/lifecycle-hooks.md
tnozicka: correct, (controllers + API objects)

contributors/design-proposals/lifecycle-hooks.md
tnozicka: done

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+    
+    LifecycleHookStatuses []LifecycleHookStatus
+}
+```
+
+### Affected Objects
+As of now the lifecycle hooks are relevant for Deployments, StatefulSets and DaemonSets. 
+They will require new optional field to be able to reference LifecycleTemplate and extending its status by []LifecycleHookRevisionStatus. 
+Also the controller will need to be slightly adjusted to scale in appropriate chunks and trigger the hooks.
+
```
kow3ns: Do you think that we might start with Deployment (its the most utilized and generally its used for stateless non-critical components), promote a stable implementation there, and then adapt that implementation to StatefulSet and DaemonSet?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+1. Calculate next partition point to reach the closest lifecycle hook progress point and scale replicas in update appropriately. If there is no hook remaining, GOTO 5.
+2. When partition point is reached, run the hook by creating a Job using LifecycleHook.JobSpec
+3.
+    1. If the hook failed and FailurePolicy is Abort - emit event, fail the update and initiate rollback (GOTO 5.) 
+    2. If the hook failed and FailurePolicy is Ignore - only emit event
+4. GOTO 1.
+5. Finish
+
+(In case of rollover cancel any hooks running and don't execute new ones.)
```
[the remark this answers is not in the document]
kargakis: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#rollover-aka-multiple-updates-in-flight

contributors/design-proposals/lifecycle-hooks.md
tnozicka: Yes, my plan is to start with Deployments.

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+### Affected Objects
+As of now the lifecycle hooks are relevant for Deployments, StatefulSets and DaemonSets. 
+They will require new optional field to be able to reference LifecycleTemplate and extending its status by []LifecycleHookRevisionStatus. 
+Also the controller will need to be slightly adjusted to scale in appropriate chunks and trigger the hooks.
+
+```go
+type DeploymentSpec struct {
+    // ...
+    // Addition
+    LifecycleTemplate *ObjectReference
```
smarterclayton: What if I want to control the rollout process?  Isn't that a job itself?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+    // ...
+    // Addition
+    LifecycleHookRevisionStatuses []LifecycleHookRevisionStatus
+}
+```
+
+## Algorithm
+If there is a LifecycleTemplate referenced from an object:
+
+1. Calculate next partition point to reach the closest lifecycle hook progress point and scale replicas in update appropriately. If there is no hook remaining, GOTO 5.
```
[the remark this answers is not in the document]
smarterclayton: Some concrete examples for 3:

1. running a database migration - should be reentrant
2. clearing a cache - should be reentrant
3. notifying an external system of our current location - should be reentrant

I think at least once is probably ok, but I don't know that all users would expect at-least-once.  Similar to daemonset discussion - will users naturally assume these are at-most-once, and if so, will they experience catastrophic failure due to it?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+    // If such situation shall occur that two different ProgressPoints should be reached at
+    // the same time, all the hooks for an earlier ProgressPoint will be ran (and finished)
+    // before any later one.
+    ProgressPoint intstr.IntOrString
+    
+    // "Abort" - Failure for this hook is fatal, deployment is considered failed and should be rollbacked.
+    // "Ignore" - Even if the hook fails deployment continues rolling out new version.
+    FailurePolicy string
+    
+    JobSpec v1.JobSpec
```
smarterclayton: How do I parameterize the job spec based on data in my deployment?  Or would I have to create one hook per deployment in that case?

contributors/design-proposals/lifecycle-hooks.md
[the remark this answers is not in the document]
smarterclayton: Of the examples from the previous comment, I think all of them imply the controller MUST block until the job completes.  Not running a database migration could be catastrophic.

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+}
+```
+
+## Algorithm
+If there is a LifecycleTemplate referenced from an object:
+
+1. Calculate next partition point to reach the closest lifecycle hook progress point and scale replicas in update appropriately. If there is no hook remaining, GOTO 5.
+2. When partition point is reached, run the hook by creating a Job using LifecycleHook.JobSpec
+3.
+    1. If the hook failed and FailurePolicy is Abort - emit event, fail the update and initiate rollback (GOTO 5.)
```
smarterclayton: Define rollback here.  I'm not sure whether you mean "give up on the new deployment and mutate the deployment back to the old revision" or "go into very long backoff".  I.e. perm-fail vs not-perm fail

contributors/design-proposals/lifecycle-hooks.md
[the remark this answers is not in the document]
tnozicka: > 1. Does the controller block until the Job completes?
Yes. (Although there is always possibility of adding an option to customize it.) 

> 3. Do we require the Jobs to be idempotent for safety. I think it will be difficult to ensure at most once execution. We could probably ensure at least once.
User will get whatever guarantees Kubernetes Job gives him. (But it should be idempotent.)

> 4. If we are using available replicas as the trigger, what happens if the number of available replicas decreases during execution of the Job?
Nothing, once we reach certain progress point we trigger the hook and never go back.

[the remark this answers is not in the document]
tnozicka: need to think about details of rollback in case of 2. more but it should be based on revision

contributors/design-proposals/lifecycle-hooks.md
tnozicka: The plan was to inject some environment variables inside (like image) but I failed to mention it in the proposal

contributors/design-proposals/lifecycle-hooks.md
tnozicka: I mean fail this rollout for good and rollback to previous (working) revision

contributors/design-proposals/lifecycle-hooks.md
kargakis: Should be a separate strategy that forces the controller to ignore the deployment. Wouldn't that process needs to implement the hooks api? Sounds like a separate design proposal.

contributors/design-proposals/lifecycle-hooks.md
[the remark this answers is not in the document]
ironcladlou: >How do you determine if a Job exists for a particular execution of the rolling update? That is, it seems to me that the execution of a Job is tied to the roll out of a specific revision, how do we track the lifecycle of the an individual Job for each roll out and rollback of a particular revision?

Would this be the same mechanism we use to determine whether a replicaset exists for a given deployment? That is, a deterministically named job based on the pod template spec hash.

contributors/design-proposals/lifecycle-hooks.md
ironcladlou: Is there any way for downward API to help here? The platform choosing a static and arbitrary set of fields to expose as (arbitrary) env vars is definitely useful, but allowing the users to select what data to expose and as mounted files seems very interesting to me. Crazy?

tnozicka: I like the idea of extending downward API to cover this use case where the objects are separate. That would allow you to pass any deployment property and we could inject just limited and necessary number of envs like the previous image or similar.

contributors/design-proposals/lifecycle-hooks.md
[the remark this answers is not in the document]
tnozicka: I think we could associate all the hooks (Jobs) by setting controllerRef to particular RS. If we decide to run hooks for rollbacks, those could be distinguished by label. 

We would have probably done it either way for purposes of adoption and garbage collection.

mfojtik:
@ironcladlou

contributors/design-proposals/lifecycle-hooks.md
liggitt: > I like the idea of extending downward API to cover this use case where the objects are separate. That would allow you to pass any deployment property

Expanding downward API to be cross-object is a pretty big change... is that desired? If so, we'd need to be really careful:
* puts a burden on the kubelet to watch more objects
* could get the kubelet to fetch and inject info you shouldn't be able to see
* can run into ordering problems when multiple unsynchronized data sources update (current downward API info all comes from the pod object which has a linearizable resourceVersion)

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+    RevisionHistoryLimit *int32
+    
+    // List of lifecycle hooks
+    // Can have multiple hooks at the same ProgressPoint; order independent
+    Hooks []LifecycleHook
+}
+
+type LifecycleHook struct {
+    // Unique name
+    Name string
```
ironcladlou: Unique in what scope?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type LifecycleTemplateSpec struct {
+    // "Always" - keep all Jobs 
+    // "OnFailure" - keep only failed Jobs
+    // "Never" - delete Jobs immediately 
+    RetainPolicy string
+    
+    // After reaching this limit the Job history will be pruned.
+    RevisionHistoryLimit *int32
```
ironcladlou: This seems a bit inflexible given the 1:1 relationship between a LifecycleHook and Job, especially given the size of Hooks is variable. Does it make any sense for the history limit to be per hook?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+    TypeMeta
+    ObjectMeta
+    Spec LifecycleTemplateSpec
+}
+
+type LifecycleTemplateSpec struct {
+    // "Always" - keep all Jobs 
+    // "OnFailure" - keep only failed Jobs
+    // "Never" - delete Jobs immediately 
+    RetainPolicy string
```
ironcladlou: Is there ever a case where this policy might differ between hooks within the same template?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,164 @@
...
+    
+    // ProgressPoint specifies the point during update when the hook should be triggered.
+    // Accepts both the number of new pods available or a percentage value representing the ratio
+    // of new.availableReplicas to the declared replicas. In case of getting a percentage
+    // it will try to reach the exact or the closest possible point right after it,
+    // trigger the job, wait for it to complete and then continue. 
+    // If such situation shall occur that two different ProgressPoints should be reached at
+    // the same time, all the hooks for an earlier ProgressPoint will be ran (and finished)
+    // before any later one.
+    ProgressPoint intstr.IntOrString
```
ironcladlou: Has anybody yet envisioned another triggering criterion? If we think we might come up with more, a single field coupled to availableReplicas will become messy and we might want to talk about a new type here.

johngmyers:
My use case for a deployment lifecycle hook is as follows:

I will have a Helm lifecycle hook which will talk to an authentication service to create a new authentication credential for the deployment version. I would want a lifecycle hook to run when that deployment version is completely rolled off of (deleted, rolled back, or replaced with new verison) in order to disable that credential. If the deployment is rolled back to that version, a lifecycle hook would have to run in order to re-enable that credential prior to starting any PODs.

I don't think this proposal handles that use case.

contributors/design-proposals/lifecycle-hooks.md
tnozicka: In the scope of `Hooks []LifecycleHook`; likely better to mention the restriction there

contributors/design-proposals/lifecycle-hooks.md
tnozicka: Both options make sense to me; limit hooks history per rollout and per hook. 

Other option is to start with no limiting because with owner references it will be automatically limited by the RevisionHistoryLimit from the deployment

contributors/design-proposals/lifecycle-hooks.md
tnozicka: Different hooks can have different priority (say notifying IRC vs. migrating a database); this way you can choose to retain only some of them.

contributors/design-proposals/lifecycle-hooks.md
tnozicka: I don't have a use case not covered by `ProgressPoint` but interested to hear ideas.

contributors/design-proposals/lifecycle-hooks.md
tnozicka: Reusing existing downward API could be a good argument for having the PolicyTemplate embedded in the object (Deployment, ...) at the expense of re-usability.

tnozicka:
@johngmyers this proposal covers rollout and we mention rollback in comments but deletion is out of scope for it. Looks like a job for external controller.

johngmyers:
Perhaps this should be called a rollout hook as it doesn't cover the complete lifecycle?

contributors/design-proposals/lifecycle-hooks.md
ironcladlou: >Reusing existing downward API could be a good argument for having the PolicyTemplate embedded in the object (Deployment, ...) at the expense of re-usability.

Talked some with @tnozicka about this offline and I'm now convinced that the lifecycle hook makes more sense embedded into the deployment spec itself rather than as a reference. Being able to use the downward API for the job template seems more generally useful than reusing hook templates across deployments. Deployments themselves (and any hooks defined within) could be templated in other ways.

contributors/design-proposals/lifecycle-hooks.md
tnozicka: Moving this to per hook then ;)

contributors/design-proposals/lifecycle-hooks.md
tnozicka: I am fine starting small without any explicit pruning.

PI-Victor:
i see the proposal discussion hit a stalemate, any updates/further discussions on it?

tnozicka:
@PI-Victor I need to update the proposal but there was KubeCon follow by me taking most of this year's PTO for personal reason so I haven't got to this yet, but it's on my TODO list. We are committed to moving this forward, the only limitation is time ;)

PI-Victor:
@tnozicka sounds good, thank you for the update! i actually need this, so if there's anything i can do to help, i'd be happy to. 

even if i take into consideration that by the time our k8s vendor updates to the version that will use this a lot of things might change :)

i'm currently exploring to see if i can write a custom controller that can at least partially implement this.

tnozicka:
@PI-Victor 

>  i actually need this, so if there's anything i can do to help, i'd be happy to.

Could you describe your use case here, just to strengthen the need for this?

Reviewing the proposal and code would certainly be helpful :)

> even if i take into consideration that by the time our k8s vendor updates to the version that will use this a lot of things might change :)
> i'm currently exploring to see if i can write a custom controller that can at least partially implement this.

This is already implemented in OpenShift for DeploymentConfigs (predecessor of Deployments), if that would help you in some way. I don't think you can implement this with external controller unless you have external strategy or auto-pausing because the controller needs to stop scaling while you execute your hook at the progressPoint. I did a PoC at Sig Apps (linked in the proposal) for external controller with StatefulSets which are the only ones supporting autopausing.

tnozicka:
I'd hope to get to update this proposal next week, when I put out some fires ;)

PI-Victor:
i've written a 7 paragraphs answer and realised that it's turning to be out of scope for this github PR. so, i'll just say that my particular use case is a niche one and doesn't add a lot of value to this, it's mostly derived from a finicky k8s-CI/CD enterprise setup.
Thanks for the heads up regarding the custom controller, seems that i'll have to come up with some sort of other solution :)

tnozicka:
I have tried to resurrect the proposal in lines of simplifying it. Notable changes:
 - embedded the hooks configuration into objects instead of referencing it
 - dropped any notion of "Downward API like" style injections, now that it's embedded it's not that needed
 - opened a path in API for another triggering criterions
 - 3 env vars to inject into hooks' jobs

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
...
+```go
+type HookTriggerType string
+
+const (
+	ProgressAvailableHookTriggerType HookTriggerType = "ProgressAvailable"
+)
+
+type ProgressAvailableParams struct {
+    // ProgressPoint specifies the point during update when the hook should be triggered.
+    // Accepts both the number of new pods available or a percentage value representing the ratio
```
kargakis: If we are going to implement something like this then isn't autopausing an easier thing to implement and handle arbitrary workflow execution (start a job or really anything else) outside of the core controllers?

tnozicka: > If we are going to implement something like this then isn't autopausing an easier thing to implement

From the scaling point of view, the amount of work will be slightly less as this will be like setting a partition, only internally, without exposing it in the API (yet). We can as well expose the partition in API (auto-pausing) but I didn't want to mix it into this proposal.

> handle arbitrary workflow execution (start a job or really anything else) outside of the core controllers?

I'd prefer to see it in apps. I think this would be beneficial and helpful for the users - I don't want to decouple this and have it only in some distribution just because we can.

Raffo:
It's really great to see this moving again, it's so much a needed feature. How do you think we can get the right reviews to make this progress?

tnozicka:
friendly ping @kow3ns @janetkuo

tnozicka:
ptal @kubernetes/sig-apps-proposals

kow3ns:
A few things 

1. You might want to restructure as a KEP and move here [https://github.com/kubernetes/community/tree/master/keps/sig-apps]. imo there has been enough discussion to merge the KEP as the SIG is interested in doing something here. I point to discussion above, and the demo referenced in this proposal as evidence.

1. You might want to more strongly emphasize that you are fundamentally proposing to upstream a feature that is in production OpenShift to Kubernetes so all Kubernetes users can benefit form the feature. You do mention this, but I just want to be sure that its clear to others.

1. We should take some time in SIG Apps workloads discussion to talk about if/why we want to do this in core. I agree with an in-core implementation, but if we are going to propose it to SIG Architecture we need to have a strong case.
  .1 Snapshots is closely coupled with PVC and PV but SIG Storage is going to have to implement it as a SIG sponsored add-on.
  .1 RuntimeClass is closely coupled with Pod, kubelet, and the underlying Kubernetes dist/env and it is going to be implemented as a SIG sponsored add-on.

1. If we are unsuccessful in convincing SIG Architecture in [3], I think we should still do this a SIG sponsored add-on that implements this feature in OSS and that can be leveraged by any distribution, but, largely because of the complexity involved in doing this correctly out of core, I'd like to at least try to do [3] first.

1. Can you talk about an implementation and promotion timeline and acceptance criteria for promotion? Feedback from SIG Architecture is that we should be clear about this prior to implementation.

1.  Do you think we should do this prior to handling alpha fields in API machinery [context](https://groups.google.com/forum/#!topic/kubernetes-sig-architecture/y9FulL9Uq6A).

tnozicka:
@kow3ns thanks for the feedback!

> 1. You might want to restructure as a KEP and move here [https://github.com/kubernetes/community/tree/master/keps/sig-apps]. imo there has been enough discussion to merge the KEP as the SIG is interested in doing something here. I point to discussion above, and the demo referenced in this proposal as evidence.

This predates KEP, but it makes sense to convert it.

> 2. You might want to more strongly emphasize that you are fundamentally proposing to upstream a feature that is in production OpenShift to Kubernetes so all Kubernetes users can benefit form the feature. You do mention this, but I just want to be sure that its clear to others.

Sure, I'll emphasize that in the updated proposal.

> 3. We should take some time in SIG Apps workloads discussion to talk about if/why we want to do this in core. I agree with an in-core implementation, but if we are going to propose it to SIG Architecture we need to have a strong case.
.1 Snapshots is closely coupled with PVC and PV but SIG Storage is going to have to implement it as a SIG sponsored add-on.
.1 RuntimeClass is closely coupled with Pod, kubelet, and the underlying Kubernetes dist/env and it is going to be implemented as a SIG sponsored add-on.

I am gonna write down some points and put the discussing in the agenda.

> 4. If we are unsuccessful in convincing SIG Architecture in [3], I think we should still do this a SIG sponsored add-on that implements this feature in OSS and that can be leveraged by any distribution, but, largely because of the complexity involved in doing this correctly out of core, I'd like to at least try to do [3] first.

I hope we succeed in 3. because the complexity of not doing this in core is significantly higher as we need to interact with workload scaling which would require autopausing and/or external strategy.

> 5. Can you talk about an implementation and promotion timeline and acceptance criteria for promotion? Feedback from SIG Architecture is that we should be clear about this prior to implementation.

I'll put that into the updated proposal.

> 6. Do you think we should do this prior to handling alpha fields in API machinery context.

Are other Sig-Apps proposals waiting for that? Is there a case where this one differs?

Generally I think the alpha gate rule is in effect until a decision is made and it seems it might be just as close as enhancing the old approach.

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
...
+    // "Abort" - Failure for this hook is fatal, deployment becomes perma-failed
+    // "Ignore" - Even if the hook fails deployment continues rolling out new version.
+    FailurePolicy string
+    
+    // "Always" - always keep this jobs 
+    // "OnFailure" - keep this Job only if it fails
+    // "Never" - delete this job immediately after it is done 
+    RetainPolicy string
+    
+    JobSpec v1.JobSpec
```
bgrant0607: I don't think the apps API group should depend on the batch API group in this way. What is needed more than a pod template?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
+# Lifecycle Hooks
```
bgrant0607: All docs in the design-proposals directly belong in SIG-specific subdirectories

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
...
+    // before any later one.
+    ProgressPoint intstr.IntOrString
+}
+
+type HookTrigger struct {
+    // Type of hook trigger. Can be ProgressAvailable. Default is ProgressAvailable.
+    Type HookTriggerType
+    
+    // ProgressAvailable config params. Present only if type = "ProgressAvailable".
+    ProgressAvailableParams *ProgressAvailableParams
```
bgrant0607: @apelisse Do we have conventions for unions yet?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
...
+    // it will try to reach the exact or the closest possible point right after it,
+    // trigger the job, wait for it to complete and then continue. 
+    // If such situation shall occur that two different ProgressPoints should be reached at
+    // the same time, all the hooks for an earlier ProgressPoint will be ran (and finished)
+    // before any later one.
+    ProgressPoint intstr.IntOrString
+}
+
+type HookTrigger struct {
+    // Type of hook trigger. Can be ProgressAvailable. Default is ProgressAvailable.
```
bgrant0607: We need to explicitly document that clients should expect the set of valid values to grow.
https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#backward-compatibility-gotchas

Are we planning to validate the value?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
...
+The related state will be reported using Conditions and details will be available on the created Job objects.
+Also the controllers will need to be adjusted to scale in appropriate chunks and trigger the hooks.
+
+```go
+type DeploymentSpec struct {
+    // ...
+    // Addition
+    // List of lifecycle hooks
+    // Can have multiple hooks at the same ProgressPoint; order independent
+    LifecycleHooks []LifecycleHook
```
bgrant0607: What are the merging semantics for these lists?

cc @apelisse

bgrant0607:
@smarterclayton @kow3ns 

External solutions, such as Helm and Operators, provide lifecycle hooks around creation, updates, and deletion. Is the main value here from mid-rollout hooks? 

Is there a reason why multiple workload controllers can't be used to expose intermediate points? For instance, I'm convinced we should never implement promotable canaries (https://github.com/kubernetes/kubernetes/issues/11505). Just create a canary deployment.

Deployments, ReplicaSets, and DaemonSets should have uniform, fungible replicas. Is the main reason to provide hooks for those resources just consistency with StatefulSet, or are there important use cases for hooks for non-stateful applications?

I'm pretty unhappy with the state of lifecycle hooks in Kubernetes currently. I'd like to see a more comprehensive, systematic proposal before we add more.

* Intra-container: entrypoint.sh is common (https://github.com/kubernetes/kubernetes/issues/30716), but no documented best practice
* Containers: postStart and preStop hooks; never added preStart and postStop (https://github.com/kubernetes/kubernetes/issues/140)
* Pods: initContainers; have not added deferContainers (https://github.com/kubernetes/community/pull/483)
* Volumes: never implemented container volumes (https://github.com/kubernetes/kubernetes/issues/831)
* Sidecar containers: quite a few challenges to make them work as needed (e.g., https://github.com/kubernetes/kubernetes/issues/25908)
* ConfigMap: never implemented HUP (https://github.com/kubernetes/kubernetes/issues/24957), nor any rollout mechanism (https://github.com/kubernetes/kubernetes/issues/22368), nor post-update hooks (https://github.com/kubernetes/kubernetes/issues/1553)

kow3ns:
@bgrant0607 - Agreed that we need more consistency here. I can add some insight as to what is being done and why.

entrypoint.sh (kubernetes/kubernetes#30716): We decided we were not going to do templates for file generation. We can add best practices to the documentation, but imo this has not been a major hurdle for our users to get over.

Containers: postStart and preStop hooks; never added preStart and postStop (kubernetes/kubernetes#140): We can take a look at adding these. However, this would have to be done in coordination with SIG Node. SIG Apps could drive it though.

Pods: initContainers; have not added deferContainers (#483) : After much discussion, its not clear that this solves actual problems. API symmetry and [this](https://github.com/kubernetes/community/pull/483#issuecomment-378659539) are the best rationals I've seen for doing this. After multiple presentations at SIG-Apps it failed to gather the support or interest of the community.

Volumes: never implemented container volumes (kubernetes/kubernetes#831): I'm not sure how this relates directly to life-cycle hooks. There are implications to the CRI and CSI here. The CSI doesn't support anything like this today, but, there is work in progress in SIG Storage that might let us do this given an arbitrary CRI implementation in the near future. I don't see how it relates to this issue though.

Sidecar containers:  (e.g., kubernetes/kubernetes#25908) : See [2148](https://github.com/kubernetes/community/pull/2148), there is an open proposal and we're talking it through with SIG Node now.

ConfigMap: never implemented HUP (kubernetes/kubernetes#24957), nor any rollout mechanism (kubernetes/kubernetes#22368), nor post-update hooks (kubernetes/kubernetes#1553) : There isn't general agreement that we should do this [1163](https://github.com/kubernetes/community/pull/1163).

So, with the exception [140](kubernetes/kubernetes#140), we have at least thought through and attempted to make progress on all of these, except for [831](kubernetes/kubernetes#831), and isn't clear to me wrt to how it relates to this space. 

This proposal, along with [2148](https://github.com/kubernetes/community/pull/2148), and potentially with  [140](kubernetes/kubernetes#140) are places we feel we have enough consensus to make significant progress. There is nothing stopping us from doing all three in tandem so we can take a look at a proposal for [140](kubernetes/kubernetes#140). When [2148](https://github.com/kubernetes/community/pull/2148) is ready to take to SIG architecture for further review we will do so. OpenShift already uses this extension in DeployConfig, and the developers want to upstream it into core. The thought is to take it to SIG Architecture for a decision on whether this should be a core feature or implemented as a add on extension.

contributors/design-proposals/lifecycle-hooks.md
apelisse: Nothing formal yet, but new APIs that have "oneof" structs should try to:
- Have a struct that only contains:
  - a discriminator as a string,
  - optional (with `// +optional`), omitempty, pointer fields,
- That structure must not be inlined,
- Values for the discriminator must match corresponding field names (modulo capitalized first letter).

bgrant0607:
@kow3ns Thanks. What I am looking for is:

1. A sketch of application lifecycle and points where users need to be able to add their own functionality. I brought up entrypoint.sh, container volumes, and config maps, because they are all similar in that application config/data pulling, preprocessing (e.g., merging with runtime state, such as pod IP, or secrets), copying, updating, etc. are very common needs. initContainers don't work with updates to volumes.

2. Some consistent principles and conventions (names, behaviors, supported lifecycle points) across all of these hook points.

3. Clear execution guarantees and failure behavior, and semantics compatible with the idea of allowing continuous updates to desired state.

4. Something like a diagram that shows all the levels of abstractions supporting the hooks, with some guidance about when to use one hook or another.

bgrant0607:
Thanks.

And, as a design alternative, how would this compare to a watching the controllers via something like metacontroller, which could provide hooks, if necessary?

contributors/design-proposals/lifecycle-hooks.md
tnozicka: Yes, I'm gonna move it to KEP; this structure is year old based on what was there at the time it started.

contributors/design-proposals/lifecycle-hooks.md
tnozicka: > We need to explicitly document that clients should expect the set of valid values to grow.
https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#backward-compatibility-gotchas

Will documents it's open-ended.

> Are we planning to validate the value?

I was. Is it ok to weaken the validation in the future when it's documented the type is open-ended?

contributors/design-proposals/lifecycle-hooks.md
tnozicka: `LifecycleHook.Name` is the unique key, when merging 2 list the newer `LifecycleHook` replaces the old one where the key matches.

Not sure if there is a syntax to express it or hint apimachinery.

apelisse: Here is an [example](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L2075-L2077).

PS: We are in the process of changing the name and semantic of these "hints", but for now you should be using `patchMergeKey` and `patchStrategy`.

contributors/design-proposals/lifecycle-hooks.md
thockin: Do you really need to expose all of Job?  Wouldn't we be better to librarify the job control logic and use it from a purpose-built controller?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
...
+**Status**: Proposal
+
+**RFC Issue**: https://github.com/kubernetes/kubernetes/issues/14512
+
+## Abstract
+The intent of this proposal is to support lifecycle hooks in all objects having rolling/recreate update; currently those are Deployments, StatefulSets and DaemonSets. 
+Lifecycle hooks should allow users to better customize updates (rolling/recreate) without having to write their own controllers. 
+
+## Lifecycle Hooks
+Lifecycle hook is a Job that will get triggered by the update reaching certain progress point (like 50%).
```
thockin: Are these generic "lifecycle" hooks or just "update progress" hooks?  The naming implies they are hooks that could apply at other points in the lifecycle (scale up, down, thresholds, delete) but they are only used for updates.  Fix the naming of the usages?

contributors/design-proposals/lifecycle-hooks.md
```diff
@@ -0,0 +1,152 @@
...
+const (
+	ProgressAvailableHookTriggerType HookTriggerType = "ProgressAvailable"
+)
+
+type ProgressAvailableParams struct {
+    // ProgressPoint specifies the point during update when the hook should be triggered.
+    // Accepts both the number of new pods available or a percentage value representing the ratio
+    // of new.availableReplicas to the declared replicas. In case of getting a percentage
+    // it will try to reach the exact or the closest possible point right after it,
+    // trigger the job, wait for it to complete and then continue.
```
thockin: Document what happens with edges - 0, "0%", and "100%"

justaugustus:

justaugustus:

/close

acasademont:
Not sure where did the discussion move to, can't find any further references. Any clues?

galindro:
@tnozicka to where this proposal was moved on? I can't find it anywhere like @acasademont doesn't too...

tnozicka:
I have not moved it anywhere yet, I am temporarily out of bandwidth to move this forward, so I'll resurrect the proposal when things are back to normal.

Jefferson-Faseler:
@tnozicka / @justaugustus Is there any update to this feature? Followed a link from @ironcladlou on [issue#14512](https://github.com/kubernetes/kubernetes/issues/14512), following a link from @kargakis on [issue#36477](https://github.com/kubernetes/kubernetes/issues/36477). Looks like it has been sitting about a year now?

james-mchugh:
Something like this is exactly what I am looking for. I want to create entries in a Redis database when certain deployments are, well... deployed. I cannot use container lifecycle hooks because the deployment should be able to start with 0 pod replicas (zero-scaled).

nilebox:
> I want to create entries in a Redis database when certain deployments are, well... deployed

@james-mchugh you can use admission webhooks for that: https://kubernetes.io/blog/2019/03/21/a-guide-to-kubernetes-admission-controllers/

PR #1197, opened 2017-10-16
Fix SIG scalability's slack channel link.

porridge:
Channel #sig-scalability does not exist. There is #sig-scale though.


xiangpengzhao:
Good catch!
But I prefer to rename the slack channel to keep consistency. @sarahnovotny can slack channel be renamed?
/cc @wojtek-t

castrojo:
I can rename a channel, would prefer a confirmation from the SIG leads though.

porridge:
@wojtek-t @countspongebob @jbeda WDYT?

countspongebob:
I'm fine with renaming the channel for consistency. I'll add it to the list for our Thursday meeting just to give people a heads up.

porridge:
Closing this one in favour of #1223 then.

PR #1202, opened 2017-10-16
Clarify what "on its own" means.

porridge:
This is my best guess. I'm assuming @spzala did not mean "on its own in
the comment" because then the "at the front" would be superfluous.


spzala:
@porridge thanks for making it more readable!

PR #1210, opened 2017-10-17
Explain how "retesting" is different from "testing".

porridge:
I do not think this is actually explained anywhere, and is pretty
confusing to newcomers, that submit queue looks at the tests both *with*
and *without* the change in question.

### DIFF
--- a/contributors/devel/automation.md
+++ b/contributors/devel/automation.md
@@ -43,8 +43,11 @@ Please visit that page for more details.
 
 ### Merge process
 
+The `testsAreStable()` call of the above pseudo-code makes sure tests are
+passing on the master branch *without* the PR in question.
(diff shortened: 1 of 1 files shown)


porridge:
Hm, after going through a couple PRs I'm wondering that perhaps the initial testing is the one performed by `pull-kubernetes-*` jobs, triggered by a robot when the PR is created? But AFAICT these are not even mentioned here.

contributors/devel/automation.md
```diff
@@ -43,8 +43,11 @@ Please visit that page for more details.
 
 ### Merge process
 
+The `testsAreStable()` call of the above pseudo-code makes sure tests are
+passing on the master branch *without* the PR in question.
```
fejta: It runs tests with the PR. It just doesn't rerun them before merging.

PR #1221, opened 2017-10-18
Update release note process

xiangpengzhao:
Update to latest process. Test in https://github.com/kubernetes/kubernetes/pull/53785

ref:
https://github.com/kubernetes/test-infra/pull/4718
https://github.com/kubernetes/kubernetes/pull/54114

/cc @cjwagner @Bradamant3

### DIFF
--- a/contributors/devel/pull-requests.md
+++ b/contributors/devel/pull-requests.md
@@ -71,8 +71,6 @@ Release notes are required for any PR with user-visible changes, such as bug-fix
 
 If you don't add release notes in the PR template, the `do-not-merge/release-note-label-needed` label is added to your PR automatically after you create it. There are a few ways to update it.
 
-**Descriptions**
-
 To add a release-note section to the PR description:
 
 For PRs with a release note:
@@ -81,26 +79,23 @@ For PRs with a release note:
     Your release note here
     ```
 
-For PRs require addtional action for users, include the string "action required" (case insensitive) in the release note:
+For PRs that require additional action from users switching to the new release, include the string "action required" (case insensitive) in the release note:
 
     ```release-note
     action required: your release note here
     ```
 
-For PRs without a release note, just write "NONE" (case insensitive):
+For PRs that don't need to be mentioned at release time, just write "NONE" (case insensitive):
 
     ```release-note
     NONE
     ```
 
+The `/release-note-none` comment command can still be used as an alternative to writing "NONE" in the release-note block if it is left empty.
+
 To see how to format your release notes, view the [PR template](https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md) for a brief example. PR titles and body comments can be modified at any time prior to the release to make them friendly for release notes.
 
 Release notes apply to PRs on the master branch. For cherry-pick PRs, see the [cherry-pick instructions](cherry-picks.md). The only exception to these rules is when a PR is not a cherry-pick and is targeted directly to the non-master branch.  In this case, a `release-note-*` label is required for that non-master PR.
-
-**Comments**
-
-- If the PR does not need to be mentioned at release time, leave the `/release-note-none` comment command.
-- (_DEPRECATED_) To create a release note, leave the `/release-note` comment command. This will generate a release note using the PR title by default OR the release-note block in the PR template, if it's filled in.
 
 Now that your release notes are in shape, let's look at how the PR gets tested and merged.
 
--- a/contributors/devel/pull-requests.md
+++ b/contributors/devel/pull-requests.md
@@ -69,7 +69,7 @@ If your PR meets all of the steps above, it will enter the submit queue to be me
 
 Release notes are required for any PR with user-visible changes, such as bug-fixes, feature additions, and output format changes.
 
-If you don't add release notes in the PR template the `do-not-merge/release-note-label-needed` label is added to your PR automatically when you create it. There are a few ways to update it.
+If you don't add release notes in the PR template, the `do-not-merge/release-note-label-needed` label is added to your PR automatically after you create it. There are a few ways to update it.
 
 **Descriptions**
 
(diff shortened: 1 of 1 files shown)


contributors/devel/pull-requests.md
```diff
@@ -94,7 +100,7 @@ Release notes apply to PRs on the master branch. For cherry-pick PRs, see the [c
 **Comments**
 
 - If the PR does not need to be mentioned at release time, leave the `/release-note-none` comment command.
-- To create a release note, leave the `/release-note` comment command. This will generate a release note using the PR title by default OR the release-note block in the PR template, if it's filled in.
+- (_DEPRECATED_) To create a release note, leave the `/release-note` comment command. This will generate a release note using the PR title by default OR the release-note block in the PR template, if it's filled in.
```
cblecker: If this is deprecated, can we just remove it?

contributors/devel/pull-requests.md
```diff
@@ -81,7 +81,13 @@ For PRs with a release note:
     Your release note here
     ```
 
-For PRs without a release note:
+For PRs require addtional action for users, include the string "action required" (case insensitive) in the release note:
```
cjwagner: "For PRs that require additional action from users switching to the new release" or something like that would be a better start to this. I was previously confused about what "action required" meant, we should make it clear that it is action required of users upgrading to the new release, not action required of the PR author to improve the release note.

contributors/devel/pull-requests.md
cjwagner: I vote for removing it. If someone does try to use that command or the `/release-note-action-required` command, a comment warning that the command is deprecated is created.

contributors/devel/pull-requests.md
```diff
@@ -94,7 +100,7 @@ Release notes apply to PRs on the master branch. For cherry-pick PRs, see the [c
 **Comments**
 
 - If the PR does not need to be mentioned at release time, leave the `/release-note-none` comment command.
```
cjwagner: Since there is only one comment that is accepted you could get rid of the "Descriptions" and "Comments" sections and just say that a `/release-note-note` comment can still be used as an alternative to writing "NONE" in the release-note block if the block is left empty.

xiangpengzhao:
@cblecker @cjwagner addressed. PTAL. Thanks!

contributors/devel/pull-requests.md
```diff
@@ -81,21 +79,24 @@ For PRs with a release note:
     Your release note here
```
Bradamant3: Suggest that we add some more guidance here. We don't really have a release note template, just spaces for release notes in PR templates. Let's try at a minimum:
"Addresses issue #(link to issue, to feature description, or to any other document that describes what the PR is a part of). (Add more details about specific work of PR)

contributors/devel/pull-requests.md
```diff
@@ -81,21 +79,24 @@ For PRs with a release note:
...
     ```
 
-For PRs without a release note:
+For PRs that require additional action from users switching to the new release, include the string "action required" (case insensitive) in the release note:
+
+    ```release-note
+    action required: your release note here
```
Bradamant3: As above:
"[action required] Addresses issue #(link to issue, feature description, other document). Users upgrading to this version must (provide details about user action required)."

contributors/devel/pull-requests.md
xiangpengzhao: Good suggestion! 
One concern is, do we want to keep two copies of the release note template text both in PR template and here? They may be different as time goes.
And note that we have at least four places wrt release note stuff: prow plugin code, pr template, relnotes tool and here. It's hard to guarantee that they keep consistent when the plugin code changes. cc @cjwagner

xiangpengzhao: >One concern is, do we want to keep two copies of the release note template text both in PR template and here? They may be different as time goes. [people may forget to update the other one when they update one place]

anyway, it might be good to list the release note template in both the doc and PR template, because people may only read one of them.

xiangpengzhao: >"Addresses issue #(link to issue, to feature description, or to any other document that describes what the PR is a part of). (Add more details about specific work of PR)"

This will lead to including an issue in each release note. Do we really need this? (since we will link to the PR in our final release note)

contributors/devel/pull-requests.md
```diff
@@ -81,21 +79,24 @@ For PRs with a release note:
...
+
+For PRs that don't need to be mentioned at release time, just write "NONE" (case insensitive):
 
     ```release-note
     NONE
     ```
 
+The `/release-note-none` comment command can still be used as an alternative to writing "NONE" in the release-note block if it is left empty.
+
 To see how to format your release notes, view the [PR template](https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md) for a brief example. PR titles and body comments can be modified at any time prior to the release to make them friendly for release notes.
```
xiangpengzhao: >To see how to format your release notes, view the [PR template](https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md) for a brief example.

If we provide release note template in both places, then we don't need this xref.

xiangpengzhao: >To see how to format your release notes, view the [PR template](https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md) for a brief example.

If we provide release note template in both places, then we don't need this xref.

contributors/devel/pull-requests.md
cjwagner: Also keep in mind that the approval-handler already enforces requiring an associated issue on PRs and will tell users to add one when one is needed.

xiangpengzhao: Yeah that's one point I was thinking of. Since we will link to the associated PR for each release note in the final CHANGELOG, and the PR already has an associated issue in it's PR body, so I don't think it's so necessarily to include associated issue in release note. @Bradamant3 WDYT?

PR #1224, opened 2017-10-18
Add Cilium as another network implementation option

chadell:


chadell:
I've already signed the CLA

PR #1228, opened 2017-10-18
Add proposal to fix starvation problem, etc., in scheduler preemption

bsalamat:
This PR proposes solutions to bring pod preemption closer to Beta/GA. It particularly proposes solutions to solve the existing starvation problem and to support PodDisruptionBudget. It also explains why we have decided to not support certain features, namely, inter-pod affinity on lower priority pods, and cross node preemption.
The PR also covers changes proposed to Cluster Autoscaler to work better with preemption.

@kubernetes/sig-scheduling-feature-requests @davidopp 
@kubernetes/autoscaler-reviewers @mwielgus for the changes in Autoscaler (See `Interactions with Cluster Autoscaler` section).


bsalamat:
cc/ @wojtek-t

contributors/design-proposals/scheduling/pod-preemption.md
```diff
@@ -106,59 +107,295 @@ Now, assume everything in the above example, but the best effort pod has priorit
...
-![image](images/preemption_2.png)
+#### New Scheduler Algorithm
+
+1.  Pick the head of the pending queue (highest priority pending pod).
+1.  Try to schedule the pod.
+1.  If the pod is schedulable, assume and bind it.
+1.  If the pod is not schedulable, run preemption for the pod.
+1.  Move the pod to the list of unschedulable pods.
+1.  If a node was chosen to preempt pods, set the node name as an annotation with
+ the "NomindatedNodeName" key to the pod.
```
MaciekPytel: I always thought the convention was to use [<something>].kubernetes.io/<something> for system annotations. 'NominatedNodeName' feels inconsistent with other annotations.

contributors/design-proposals/scheduling/pod-preemption.md
```diff
@@ -106,59 +107,295 @@ Now, assume everything in the above example, but the best effort pod has priorit
...
+1.  When any pod is scheduled, or terminated, a node is added/removed, or when
+labels of pods or nodes change, remove all the pods from the unschedulable pods
+list and add them to the pending queue. (Scheduler should keep its existing rate
+limiting.)
```
MaciekPytel: I see the logic here, but it looks like a single high-priority unschedulable pod can consume 50% of scheduler calculation time (as it goes back to the top of the heap every time anything else is scheduled). And a few such pods could completely block scheduler.

Perhaps it would be safer to avoid adding unschedulable pods back to pending queue anytime anything gets scheduled? Especially since most of the time scheduling a pod is not going to help other pods (one exception I can think of is when one of the pods in question uses pod affinity).

contributors/design-proposals/scheduling/pod-preemption.md
```diff
@@ -106,59 +107,295 @@ Now, assume everything in the above example, but the best effort pod has priorit
...
+result, we may not need to scale the cluster up for all pending pods. Particularly,
+scaling up the cluster may not be necessary in two scenarios:
+
+1. The pending pod has already preempted pods and is going to run on a node soon.
+1. The pending pod is very low priority and the owner of the cluster prefers to save
+money by not scaling up the cluster for such a pod.
+
+In order to address these cases:
+1. Cluster Autoscaler will not scale up the cluster for pods with
+`NominatedNodeName` annotation.
```
MaciekPytel: Technically speaking I think we need to go a step further and assume that the pod with `NominatedNodeName` is already scheduled on the nominated node, thus preventing scale-down of this node as well. Though this is more of an implementation detail in CA and doesn't really matter much in context of this proposal.

contributors/design-proposals/scheduling/pod-preemption.md
```diff
@@ -106,59 +107,295 @@ Now, assume everything in the above example, but the best effort pod has priorit
...
+1.  When choosing victims on any evaluated node, preemption logic will try to
+reprieve pods whose PDBs are violated first. (In the alpha version, pods are
+reprieved by their ascending priority and PDB is ignored.)
+1.  In scoring nodes and choosing one for preemption, the number of pods whose
+PDBs are violated will be the most significant metric. So, a node with the lowest
+number of victims whose PDBs are violated is the one chosen for preemption. In
+the alpha version, most significant metric is the highest priority of victims.
+If there are more than one node with the same smallest number of victims whose
+PDBs are violated, lowest high priority victim will be used (as in alpha) and
+the rest of the metrics remain the same as before.
```
MaciekPytel: Isn't that going to cause a sort of cascade-preemption? For example:
1. Pods A (priority=10, no-pdb) and pod B(priority=1, pdb preventing deletion) are running on a single node in cluster.
1. Pod C with higher priority shows up, but there is no more space in cluster.
1. Pod A is evicted to make place for C.
1. Pod C is scheduled.
1. Pod B is evicted to make space for A.

In the end PDB was violated anyway when preempting for pod A and pod A has been unnecessarily restarted. This example is, of course, artificial, but aren't we worried about the general pattern of evicting higher priority pods, only to see them preempt lower priority pods in a few seconds?

contributors/design-proposals/scheduling/pod-preemption.md
bsalamat: I think we `kubernetes.io/<name>` for user facing annotations, but I may be wrong. This is an internal annotation only.

contributors/design-proposals/scheduling/pod-preemption.md
bsalamat: Your point about the fact that an unschedulable high priority pod can take 50% of scheduler time is valid. I will think how we can address that, but a few such pods cannot completely block scheduler. If nothing schedules, scheduler moves on to the next lower priority pods.

As you guessed, the only reason to move unschedulable pods back to the queue after scheduling a pod, is to check inter-pod affinity. We can perhaps check if the pod has any inter-pod affinity rule and move them back to the queue only if they do have such rules.

contributors/design-proposals/scheduling/pod-preemption.md
bsalamat: I was under the impression that autoscaler does not scale down a cluster that has unschedulable pending pods. Does it actually scale down in presence of such pending pods? If so, then what you said is valid and it shouldn't scale down.

contributors/design-proposals/scheduling/pod-preemption.md
bsalamat: The case you mentioned could happen. I can think of three possible solutions for it:
1. Let it happen. It causes a bit of extra preemption, but at the end it gets to the desired state.
2. Detect such a scenario and preempt the pod that will eventually be preempted in the first place. Detecting this case is not trivial and is computationally expensive, especially when there is a long chain of such pods.
3. Do not respect PDB in preemption. I don't think this is a good idea.

Among the three options, I think the first one is probably the best one. That's what is already proposed.

contributors/design-proposals/scheduling/pod-preemption.md
MaciekPytel: Sorry, I wasn't very clear. I didn't mean 'completely block' literally, rather that they can easily take 95+% of scheduler calculation time, severely degrading the service.

contributors/design-proposals/scheduling/pod-preemption.md
MaciekPytel: Honestly, I don't know. If it's not breaking convention than I'm fine with the current name.

contributors/design-proposals/scheduling/pod-preemption.md
bsalamat: Yes, they could take a significant time. I think adding a check for affinity, as I explained in my previous comment, can help reduce the probability of such cases.

bsalamat:
ref/ #54501

bsalamat:
@davidopp This is already implemented. Could you please take a look and LGTM if it looks fine.

bsalamat:
ping @davidopp
Could you please take a look at this? This is already implemented. It would be great to have the documents updated sooner.

contributors/design-proposals/scheduling/pod-preemption.md
```diff
@@ -106,59 +107,298 @@ Now, assume everything in the above example, but the best effort pod has priorit
...
 Our assumption is that multiple schedulers cooperate with one another. If they don't, scheduler A may schedule pod A. Scheduler B preempts pod A to schedule pod B which is then preempted by scheduler A to schedule pod A and we go in a loop.
 
-## Preemption mechanics
 
-As explained above, evicting victim(s) and binding the pending pod are not transactional. Preemption victims may have "`TerminationGracePeriodSeconds`" which will create even a larger time gap between the eviction and binding points. When a victim with termination grace period receives its termination signal, it keeps running on the node until it terminates successfully or its grace period is over. In the meantime the node resources won't be available to another pod. So, the scheduler cannot bind the pending pod right away. Scheduler should mark the pending pod as assigned and move on to schedule other pods. To do so, we propose adding a new field to PodSpec called "`NominatedNodeName`". When this field is set, scheduler knows that the pod is destined to run on the given node and takes it into account when making scheduling decisions for other pods.
```
davidopp: Why did you get rid of this paragraph? Without it, there's no high-level explanation of `NominatedNodeName` -- you just jump into using the term.

bsalamat: @davidopp I haven't deleted the paragraph. I have changed it and separated the "solution" part of it into a sub-section and updated it to reflect the most recent design changes, e.g., `NominatedNodeName` is an annotation, etc.

contributors/design-proposals/scheduling/pod-preemption.md
```diff
@@ -106,59 +107,299 @@ Now, assume everything in the above example, but the best effort pod has priorit
...
+#### New Scheduler Algorithm
+
+1.  Pick the head of the pending queue (highest priority pending pod).
+1.  Try to schedule the pod.
+1.  If the pod is schedulable, assume and bind it.
+1.  If the pod is not schedulable, run preemption for the pod.
+1.  Move the pod to the list of unschedulable pods.
+1.  If a node was chosen to preempt pods, set the node name as an annotation with
+ the "scheduler.kubernetes.io/nominated-node-name" key to the pod. This key is referred to as
+ "NominatedNodeName" in this doc for brevity.
```
davidopp: Can you add "When this field is set, scheduler knows that the pod is destined to run on the given node and takes it into account when making scheduling decisions for other pods" which was from the delete paragraph. It is useful to say what the point of it is, in case the reader does not figure it out from reading the algorithm.

bsalamat: Sure. I added that back. PTAL.

PR #1241, opened 2017-10-20
Design proposal for bootstrap checkpointing for self hosting

timothysc:
This is a distillation, and jump, point from the other proposals.  It's meant to be a simplification to eliminate the potential for misinterpretation.  

/cc @kubernetes/sig-cli-feature-requests @yujuhong @dchen1107 @roberthbailey @luxas

### DIFF
--- a/contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
+++ b/contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
@@ -1,0 +1,147 @@
+# Kubernetes Bootstrap Checkpointing Proposal
+
+## Metadata
+
+```yaml
+kep-number: draft-20171029-boostrap-checkpointing.md
+title: Kubernetes Bootstrap Checkpointing Proposal
+authors:
+  - name: Timothy St. Clair
+    github: @timothysc
+    email: timothysc@gmail.com
+owning-sig: sig-cluster-lifecycle 
+participating-sigs:
+  - sig-node
+reviewers:
+  - @yujuhong 
+  - @luxas 
+  - @roberthbailey
+approvers:
+  - @yujuhong
+  - @roberthbailey
+editor:
+  name: @timothysc
+creation-date: 2017-10-20
+last-updated: 2017-10-20
+status: final-review
+```
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Objectives](#objectives)
+  * [Goals](#goals)
+  * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+  * [User Stories](#user-stories)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Unresolved Questions](#unresolved-questions)
+
+## Summary
+
+There are several methods to deploy a kubernetes cluster, one method that 
+offers some unique advantages is self hosting.  The purpose of this proposal 
+is to outline a method to checkpoint specific annotated pods, namely the 
+control plane components, for the purpose of enabling self hosting.  
+
+The details of self hosting are beyond the scope of this proposal, and are 
+outlined in the references listed below: 
+
+  - [Self Hosted Kubernetes][0]
+  - [Kubeadm Upgrades][1]
+
+Extra details on this proposal, and its history, can be found in the links 
+below: 
+
+  - [Bootstrap Checkpointing Draft 1][2]
+  - [Bootstrap Checkpointing Draft 2][3]
+  - [WIP Implementation][4]
+
+## Objectives
+
+The scope of this proposal is **bounded**, but has the potential for broader 
+reuse in the future.  The reader should be mindful of the explicitly stated 
+[Non-Goals](#non-goals) that are listed below.
+
+### Goals
+ 
+ - Provide a basic framework for recording annotated *Pods* to the filesystem. 
+ - Ensure that a restart of the kubelet checks for existence of these files 
+ and loads them on startup.
+
+### Non-Goals
+
+- This is not a generic checkpointing mechanism for arbitrary resources. 
+(e.g. Secrets)  Such changes require wider discussions.
+- This will not checkpoint internal kubelet state. 
+- This proposal does not cover self hosted kubelet(s).  It is beyond the 
+scope of this proposal, and comes with it's own unique set of challenges.
+
+## Proposal
+The enablement of this feature is gated by a single command line flag that 
+is passed to the kubelet on startup, ```--bootstrap-checkpoint-path``` , 
+and will be denoted that it is ```[Alpha]```. 
+
+### User Stories
+
+#### Pod Submission to Running
+- On submission of a Pod, via kubeadm or an operator, an annotation 
+```node.kubernetes.io/bootstrap-checkpoint=true``` is added to that Pod, which 
+indicates that it should be checkpointed by the kubelet.  When the kubelet 
+receives a notification from the apiserver that a new pod is to run, it will 
+inspect the ```--bootstrap-checkpoint-path``` flag to determine if 
+checkpointing is enabled.  Finally, the kubelet will perform an atomic 
+write of a ```Pod_UID.yaml``` file when the afore mentioned annotation exists.
+The scope of this annotation is bounded and will not be promoted to a field.  
+
+#### Pod Deletion
+- On detected deletion of a Pod, the kubelet will remove the associated 
+checkpoint from the filesystem.  Any failure to remove a pod, or file, will 
+result in an error notification in the kubelet logs.
+
+#### Cold Start
+- On a cold start, the kubelet will check the value of 
+```--bootstrap-checkpoint-path```.  If the value is specified, it will read in 
+the contents of the that directory and startup the appropriate Pod.  Lastly, 
+the kubelet will then pull the list of pods from the api-server and rectify 
+what is supposed to be running according to what is bound, and will go through
+its normal startup procedure.
+
+### Implementation Constraints
+Due to its opt-in behavior, administrators will need to take the same precautions 
+necessary in segregating master nodes, when enabling the bootstrap annotation.
+
+Please see [WIP Implementation][4] for more details.
+
+## Graduation Criteria
+
+Graduating this feature is a responsibility of sig-cluster-lifecycle and 
+sig-node to determine over the course of the 1.10 and 1.11 releases.  History 
+has taught us that initial implementations often have a tendency overlook use 
+cases and require refinement.  It is the goal of this proposal to have an 
+initial alpha implementation of of bootstrap checkpoining in the 1.9 cycle, 
+and further refinement will occur after we have validated it across several 
+deployments. 
+
+## Testing 
+Testing of this feature will occur in three parts. 
+- Unit testing of standard code behavior 
+- Simple node-e2e test to ensure restart recovery
+- (TODO) E2E test w/kubeadm self hosted master restart recovery of an apiserver. 
+
+## Implementation History
+
+- 20171020 - 1.9 draft proposal 
+- ? - accepted proposal 
+- ? - alpha implementation code complete
+
+## Unresolved Questions
+ 
+* None at this time.  
+
+[0]: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/cluster-lifecycle/self-hosted-kubernetes.md
+[1]: https://github.com/kubernetes/community/pull/825
+[2]: https://docs.google.com/document/d/1hhrCa_nv0Sg4O_zJYOnelE8a5ClieyewEsQM6c7-5-o/edit?ts=5988fba8#
+[3]: https://docs.google.com/document/d/1qmK0Iq4fqxnd8COBFZHpip27fT-qSPkOgy1x2QqjYaQ/edit?ts=599b797c#
+[4]: https://github.com/kubernetes/kubernetes/pull/50984 


contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
+# Kubernetes Bootstrap Checkpointing Proposal
+
+## Metadata
+
+```yaml
```
roberthbailey: This is neat. I assume it was auto-generated somehow.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+* [Unresolved Questions](#unresolved-questions)
+
+## Summary
+
+There are several methods to deploy a kubernetes cluster, one method that offers some unique advantages is self hosting.  The purpose of this proposal is to outline a method to checkpoint specific annotated pods, namely the control plane components, for the purpose of enabling self hosting.  
+
+The details of self hosting are beyond the scope of this proposal, and are outlined in the references listed below: 
+
+  - [Self Hosted Kubernetes][0]
+  - [Kubeadm Self Hosting][1]
```
roberthbailey: this link points to the upgrades pull request, which doesn't seem correct.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+ - Ensure that a restart of the kubelet checks for existence of these files and loads them on startup.
+
+### Non-Goals
+
+- This is not a generic checkpointing mechanism for arbitrary resources. (e.g. Secrets)  Such changes require wider discussions.
+- This will not checkpoint internal kubelet state. 
+- This proposal does not cover self hosted kubelet(s).  It is beyond the scope of this proposal, and comes with it's own unique set of challenges.
+
+## Proposal
+The enablement of this feature is gated by a single command line flag that is passed to the kubelet on startup, ```--bootstrap-checkpoint-path``` , and will be denoted that it is experimental.
```
roberthbailey: Hopefully part of dynamic kubelet config soon...

@mtaufen

roberthbailey: Is it common to put experimental into the flag name to denote it as such? Or does that just cause a headache when moving to a stable implementation?

@mtaufen - I'm also curious how we will denote experimental / alpha / whatever features flags in dynamic kubelet config.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+- This will not checkpoint internal kubelet state. 
+- This proposal does not cover self hosted kubelet(s).  It is beyond the scope of this proposal, and comes with it's own unique set of challenges.
+
+## Proposal
+The enablement of this feature is gated by a single command line flag that is passed to the kubelet on startup, ```--bootstrap-checkpoint-path``` , and will be denoted that it is experimental. 
+
+### User Stories
+
+#### Pod Submission to Running
+- On submission of a Pod, via kubeadm or an operator, an annotation ```node.kubernetes.io/checkpoint=true``` is added to that Pod, which indicates that it should be checkpointed by the kubelet.  When the kubelet receives a notification from the apiserver that a new pod is to run, it will inspect the ```--bootstrap-checkpoint-path``` variable to determine if checkpointing is enabled.  Finally, the kubelet will perform an atomic write of a ```Pod_UID.yaml``` file when the annotation exists.
```
roberthbailey: This sounds like all pods will have this annotation. Maybe say something like "... where an annotation blah has been added to that Pod, this indicates that it should be checkpointed by the kubelet."

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+Graduating this feature is a responsibility of sig-cluster-lifecycle and sig-node to determine 
+over the course of the 1.10 and 1.11 releases.  History has taught us that initial implementations often have a tendency overlook use cases and require refinement.  It is the goal of this proposal to have an initial alpha implementation of of bootstrap checkpoining in the 1.9 cycle, and further refinement will occur after we have validated it across several deployments. 
+
+## Implementation History
+
+- 20171020 - 1.9 draft proposal 
+- ? - accepted proposal 
+- ? - alpha implementation code complete
+
+## Unresolved Questions
```
roberthbailey: These was mentioned under non-goals so I don't think they are unresolved (or rather, it doesn't need to be resolved as part of this discussion).

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
lucab: How would this interact with `--containerized`? To be more explicit: as a user, should I take care of placing a bind-mount at this path, or would this path transparently get a `/rootfs` prefix?

mtaufen: If it's experimental/alpha/whatever please feature gate it (really we should just call these things alpha, experimental is an old colloquialism...).

For managing alpha fields in any componentconfig payload, see https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#alpha-field-in-existing-api-version.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
mtaufen: What does "variable" mean in this context? I'm assuming it's a flag, because it's prefixed with `--`?

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
timothysc: Nope, it's a modified kep from @calebamiles .  The original kep 's vernacular was a little weird imo and I may PR against the template.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+owning-sig: sig-cluster-lifecycle 
+participating-sigs:
+  - sig-node
+reviewers:
+  - @yujuhong 
+  - @luxas 
+  - @roberthbailey
+approvers:
+  - @yujuhong
+  - @roberbailey
```
luxas: nit: @roberthbailey :)

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+editor:
+  name: @timothysc
+creation-date: 2017-10-20
+last-updated: 2017-10-20
+status: draft
```
luxas: To be changed I suppose

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,108 @@
...
+  * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+  * [User Stories](#user-stories)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Unresolved Questions](#unresolved-questions)
+
+## Summary
+
+There are several methods to deploy a kubernetes cluster, one method that offers some unique advantages is self hosting.  The purpose of this proposal is to outline a method to checkpoint specific annotated pods, namely the control plane components, for the purpose of enabling self hosting.
```
luxas: nit: split at 80 chars

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
luxas: I'm about to write a dedicated self-hosting proposal soon-ish

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
luxas: @lucab Self-hosting the kubelet is a non-goal

luxas: nit: use one single `, not three

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
luxas: Yeah, just moving this to non-goals for alpha SGTM

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
luxas: This sentence doesn't parse well for me, can you clarify this a bit more please?

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
timothysc: It's not a field so that like doesn't apply.  I'm following the convention for *annotations which is to remove that (alpha) from it b/c it breaks across versioning.  This is consistent on other PRs.  

With regards to the flag naming, I'm happy to add (alpha, or experimental) but adding feature gate to symbol pollute other binaries where it only applies to the kubelet is also an anti-pattern now.  I'll take final resolution from @yujuhong , but I'm not going to symbol pollute other binaries.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
timothysc: I've removed.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
yujuhong: The new guideline discourages using annotations for alpha/beta features. I think one important question is that do we plan to convert the annotation (`node.kubernetes.io/bootstrap-checkpoint=true`) to a regular field in the future when it becomes GA? If so, we should follow the guideline to make it an alpha field and gate the feature properly.

> adding feature gate to symbol pollute other binaries where it only applies to the kubelet is also an anti-pattern now

If there is API change, I think feature gate is generally expected. I'd be slightly less concerned if the change is restricted to a new kubelet flag.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,138 @@
...
+indicates that it should be checkpointed by the kubelet.  When the kubelet 
+receives a notification from the apiserver that a new pod is to run, it will 
+inspect the ```--bootstrap-checkpoint-path``` flag to determine if 
+checkpointing is enabled.  Finally, the kubelet will perform an atomic 
+write of a ```Pod_UID.yaml``` file when the afore mentioned annotation exists.  
+
+#### Pod Deletion
+- On detected deletion of a Pod, the kubelet will remove the associated 
+checkpoint from the filesystem.  Any failure to remove a pod, or file, will 
+result in an error and an event notification.
```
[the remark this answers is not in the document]
timothysc: It would be the standard flow from what I see I'm not changing anything that isn't already there.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,138 @@
...
+
+- 20171020 - 1.9 draft proposal 
+- ? - accepted proposal 
+- ? - alpha implementation code complete
+
+## Unresolved Questions
+ 
+* If the mechanism can, or should, be reused by the sig-node team for other 
+features.  
+
```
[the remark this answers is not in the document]
timothysc: I can do that.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,138 @@
...
+## Proposal
+The enablement of this feature is gated by a single command line flag that 
+is passed to the kubelet on startup, ```--bootstrap-checkpoint-path``` , 
+and will be denoted that it is ```[Alpha]```. 
+
+### User Stories
+
+#### Pod Submission to Running
+- On submission of a Pod, via kubeadm or an operator, an annotation 
+```node.kubernetes.io/bootstrap-checkpoint=true``` is added to that Pod, which
```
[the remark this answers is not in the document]
timothysc: That remains to be seem and could be added as graduation question.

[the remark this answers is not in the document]
timothysc: Add a comment about narrow use case and not graduating to a field.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
```diff
@@ -0,0 +1,138 @@
...
+checkpoint from the filesystem.  Any failure to remove a pod, or file, will 
+result in an error and an event notification.
+
+#### Cold Start
+- On a cold start, the kubelet will check the value of 
+```--bootstrap-checkpoint-path```.  If the value is specified, it will read in 
+the contents of the that directory and startup the appropriate Pod.  Lastly, 
+the kubelet will then pull the list of pods from the api-server and rectify 
+what is supposed to be running according to what is bound, and will go through
+it's normal startup procedure.
```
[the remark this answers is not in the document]
timothysc: sure

roberthbailey:
LGTM although I agree that a section on testing would be a nice addition.

contributors/design-proposals/cluster-lifecycle/draft-20171020-bootstrap-checkpointing.md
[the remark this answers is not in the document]
smarterclayton: What security implications does this field have?  Do admins have to defend against end users setting this field?  (I assume the flag gate protects it, but want to be clear).  Does enabling the flag gate mean admins should consider this a "single-tenant, non-hostile user" cluster?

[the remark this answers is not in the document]
timothysc: It's flag-gated and opt in, only kubelets that host master components will enable.  Admins should use similar taint guards on master nodes, otherwise the annotation will be ignored.

timothysc:
All comments have been addressed, initial PR has been merged, and we will iterate on this KEP as the lifecycle progresses.

PR #176, opened 2016-12-14
Proposal for explicit service links

pmorie:
Reopened from kubernetes/kubernetes#37612; proposes a solution for kubernetes/kubernetes#1768 and https://github.com/kubernetes/kubernetes/issues/7365.

PR #1249, opened 2017-10-21
Proposal: optional service links

kongslund:
This proposal has its root in an [environment conflict that I encountered](https://groups.google.com/forum/#!topic/kubernetes-users/wTCXkzphTfo) while trying to get an Docker image to run on Kubernetes. @thockin was kind to provide a bit of input on the mailing list and at DockerCon EU.

I've got a proof of concept working which always disables the injection of service links. I'm new to the code base so I'd probably need a bit of assistance to make a proper PR. This is of course assuming that this proposal is accepted.

I'm looking forward to your input.

Ping Node SIG leads: @dchen1107 @derekwaynecarr

### DIFF
--- a/contributors/design-proposals/node/optional-service-links.md
+++ b/contributors/design-proposals/node/optional-service-links.md
@@ -1,0 +1,52 @@
+# Optional service links
+
+Status: Pending
+
+Version: Alpha | Beta | GA
+
+Implementation Owner: @kongslund
+
+## Motivation
+
+Today, a list of all services that were running when a pod's containers are created is automatically injected to those containers as environment variables matching the syntax of Docker links. There is no way to disable this.
+
+Docker links have long been considered as a [deprecated legacy feature](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) of Docker since the introduction of networks and DNS. Likewise, in Kubernetes, DNS is to be preferred over service links.
+
+Possible issues with injected service links are
+
+* Accidental coupling.
+* Incompatibilities with container images that no longer utilize service links and explicitly fail at startup time if certain service links are defined.
+* Performance penalty in starting up pods [for namespaces with many services](https://github.com/kubernetes/kubernetes/issues/1768#issuecomment-330778184).
+
+## Proposal
+
+Make it possible for a user to disable injection of service links into containers of a pod by adding a disable flag to the pod's spec. Make the default value false in order to stay backwards compatible with the v1 API. Make an exception for the `kubernetes` service in the master namespace so that it will always get injected.
+
+## User Experience
+
+### Use Cases
+
+* As a user, I want to be able to disable service link injection since the injected environment variables interfer with a Docker image that I am trying to run on Kubernetes
+* As a user, I want to be able to disable service link injection since I don't need it and it takes increasingly longer time to start pods as services are added to the namespace.
+
+## Implementation
+
+`PodSpec` is extended with an additional field, `disableServiceLinks` of type boolean. Default value is false.
+
+In `kubelet_pods.go`, the value of that field is passed along to the function `getServiceEnvVarMap` where it is used to decide which selector should be used for the `serviceLister`. Current behavior is `labels.Everything()`. In case `disableServiceLinks` is true then only the `kubernetes` service in the `kl.masterServiceNamespace` should be injected.
+
+```
+func (kl *Kubelet) getServiceEnvVarMap(ns string, disableServiceLinks bool) (map[string]string, error) {
+  ...decide on selector
+}
+```
+
+### Client/Server Backwards/Forwards compatibility
+
+Pods that do not have the field set will assume a value of false.
+
+## Alternatives considered
+
+An alternative is to add support for explicit service links, e.g. by applying a label selector map with a default behavior of including everything.
+
+Making service links explicit has been discussed in the [Explicit service links](https://github.com/pmorie/community/blob/6239773beb623c0dafc768be8372e5daf605aab0/contributors/design-proposals/service_links.md) design proposal without being merged in.
--- a/contributors/design-proposals/node/optional-service-links.md
+++ b/contributors/design-proposals/node/optional-service-links.md
@@ -26,24 +26,24 @@ Make it possible for a user to disable injection of service links into container
 
 ### Use Cases
 
-* As a user, I want to be able to disable service link injection since the injected environment variables interfer with a Docker image that I am trying to run on Kubernetes
+* As a user, I want to be able to disable service link injection since the injected environment variables interfere with a Docker image that I am trying to run on Kubernetes
 * As a user, I want to be able to disable service link injection since I don't need it and it takes increasingly longer time to start pods as services are added to the namespace.
 
 ## Implementation
 
-`PodSpec` is extended with an additional field, `disableServiceLinks` of type boolean. Default value is false.
+`PodSpec` is extended with an additional field, `enableServiceLinks`. The field should be a pointer to a boolean and default to true if nil.
 
-In `kubelet_pods.go`, the value of that field is passed along to the function `getServiceEnvVarMap` where it is used to decide which selector should be used for the `serviceLister`. Current behavior is `labels.Everything()`. In case `disableServiceLinks` is true then only the `kubernetes` service in the `kl.masterServiceNamespace` should be injected.
+In `kubelet_pods.go`, the value of that field is passed along to the function `getServiceEnvVarMap` where it is used to decide which selector should be used for the `serviceLister`. Current behavior is `labels.Everything()`. In case `enableServiceLinks` is false then only the `kubernetes` service in the `kl.masterServiceNamespace` should be injected. The latter is needed in order to preserve Kubernetes variables such as `KUBERNETES_SERVICE_HOST` since a lot of code depends on it.
 
 ```
-func (kl *Kubelet) getServiceEnvVarMap(ns string, disableServiceLinks bool) (map[string]string, error) {
+func (kl *Kubelet) getServiceEnvVarMap(ns string, enableServiceLinks *bool) (map[string]string, error) {
   ...decide on selector
 }
 ```
 
 ### Client/Server Backwards/Forwards compatibility
 
-Pods that do not have the field set will assume a value of false.
+Pods that do not have the field set will assume a value of true.
 
 ## Alternatives considered
 


contributors/design-proposals/node/optional-service-links.md
```diff
@@ -0,0 +1,52 @@
...
+## User Experience
+
+### Use Cases
+
+* As a user, I want to be able to disable service link injection since the injected environment variables interfer with a Docker image that I am trying to run on Kubernetes
+* As a user, I want to be able to disable service link injection since I don't need it and it takes increasingly longer time to start pods as services are added to the namespace.
+
+## Implementation
+
+`PodSpec` is extended with an additional field, `disableServiceLinks` of type boolean. Default value is false.
```
thockin: Open question:  I'd prefer this as `enableServiceEnvVars` with default true.  Objections?

contributors/design-proposals/node/optional-service-links.md
```diff
@@ -0,0 +1,52 @@
...
+
+In `kubelet_pods.go`, the value of that field is passed along to the function `getServiceEnvVarMap` where it is used to decide which selector should be used for the `serviceLister`. Current behavior is `labels.Everything()`. In case `disableServiceLinks` is true then only the `kubernetes` service in the `kl.masterServiceNamespace` should be injected.
```
thockin: Open: Should this include the kubernetes variables?  Or maybe we need a different field for that?  Anyone have thoughts?

contributors/design-proposals/node/optional-service-links.md
thockin: Also, would this make sense as defaulting from PodPreset?  @jessfraz and  ???

bgrant0607:
See also #176

bgrant0607:
Also ref https://github.com/kubernetes/kubernetes/issues/1768

contributors/design-proposals/node/optional-service-links.md
bgrant0607: enable is preferred over disable, yes.

@droot regarding PodPreset

contributors/design-proposals/node/optional-service-links.md
bgrant0607: All the service variables.

DNS, PodPreset, and potentially explicit service links should replace them all. The existence of any is a problem for scalability, creation-order dependencies, and other reasons.

thockin:
I called out the KUBERNETES_* ones because their existence is a check that
some code (e.g. some of the client setup) use as fallbacks:

staging/src/k8s.io/client-go/rest/config.go

bgrant0607:
cc @kubernetes/sig-api-machinery-api-reviews

bgrant0607:
We may need a special case for just the apiserver-related variables.

contributors/design-proposals/node/optional-service-links.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Proposal
+
+Make it possible for a user to disable injection of service links into containers of a pod by adding a disable flag to the pod's spec. Make the default value false in order to stay backwards compatible with the v1 API. Make an exception for the `kubernetes` service in the master namespace so that it will always get injected.
+
+## User Experience
+
+### Use Cases
+
+* As a user, I want to be able to disable service link injection since the injected environment variables interfer with a Docker image that I am trying to run on Kubernetes
```
ncdc: typo: interfer**e**

thockin:
The apiserver vars are the only one I want special-cased (and I think they
are by this proposal).  The question is whether we want a parallael flag
for them.  But for now, I'll argue not.

contributors/design-proposals/node/optional-service-links.md
kongslund: No objections.

kongslund:
Thanks for comments so far. Any other remarks?

contributors/design-proposals/node/optional-service-links.md
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+### Client/Server Backwards/Forwards compatibility
+
+Pods that do not have the field set will assume a value of false.
+
+## Alternatives considered
+
+An alternative is to add support for explicit service links, e.g. by applying a label selector map with a default behavior of including everything.
```
stp-ip: Starting with a way to disable the current behavior and considering extending via selector typ behavior might be nice, but definitely needs a few specific examples, where that might be necessary from a k8s perspective.

thockin:
@kongslund I think you're on-target.  You have a week before code freeze for 1.9 :)

kongslund:
Thanks for heads up, @thockin.
So does everyone think that a boolean flag is sufficient for now?

thockin:
I think that's fine.  Let's get it working so we can take a concrete look at it.  Deltas from there are not so terrible to contemplate.

contributors/design-proposals/node/optional-service-links.md
smarterclayton: The field should be a pointer to a boolean, default to true if nil, and have a positive value (enableServiceEnvVars).

contributors/design-proposals/node/optional-service-links.md
smarterclayton: If this omits KUBERNETES_SERVICE_HOST then in cluster config will break.  That impacts sidecars like istio.

It’s not clear to me that we should omit those variables.  If we do, we may have to make this a container field, not pod field.

thockin:
Is this abandoned?

mrostecki:
Does that proposal need to be merged before I continue the work on my implementation? If yes, @kongslund, are you up to finishing the work on proposal?

I'm sorry that I started the implementation, but I didn't see the proposal before and I just started working on the issue...

kongslund:
@mrostecki The proposal is a starting point and already has good feedback from the community. It does not need to be merged before you can provide an implementation but you should incorporate the contents of the proposal or start a discussion if you think something should be different.

bgrant0607:
cc @kubernetes/sig-architecture-proposals

contributors/design-proposals/node/optional-service-links.md
```diff
@@ -0,0 +1,52 @@
...
+## User Experience
+
+### Use Cases
+
+* As a user, I want to be able to disable service link injection since the injected environment variables interfere with a Docker image that I am trying to run on Kubernetes
+* As a user, I want to be able to disable service link injection since I don't need it and it takes increasingly longer time to start pods as services are added to the namespace.
+
+## Implementation
+
+`PodSpec` is extended with an additional field, `enableServiceLinks`. The field should be a pointer to a boolean and default to true if nil.
```
bgrant0607: This may need to be disableServiceLinks, so that the existing behavior would be in effect if it were the zero value.
cc @erictune

kongslund: Either is fine with me. However, there seems to be a preference for `enable` instead of `disable` cf. https://github.com/kubernetes/community/pull/1249#discussion_r147019407

smarterclayton: Brian, as a pointer zero value would be interpreted as existing behavior by older clients - can you articulate a bit more what you’d be concerned about as a bool ptr?

mrostecki: Even if we go with `enable`, we can set its default value to `true`, so I don't see any backwards compatibility issues there.

bgrant0607:
cc @kubernetes/sig-scalability-proprosals @kubernetes/sig-network-proposals

bgrant0607:
cc @kubernetes/api-approvers

contributors/design-proposals/node/optional-service-links.md
bgrant0607: Sorry for delay. I do want this to happen.

If we make it a pointer and omitempty, will both true and false values be explicitly encoded as long as the field is non-null? I want to ensure that podspecs without the field can be unambiguously interpreted.

Also, doing this will add a field to every Pod and every pod template rendered. It will change podspec hashes, but that should be ok. It will make all pods more verbose, but maybe more users will become aware of the field and use it.

If we could, it would be great to not apply the default in pod templates. In addition to not changing hashes, this would make it feasible to set the field via admission control.

smarterclayton: Now that we use versioned clients leaving fields unset on the wire has fewer downsides - a client already has to test existence, so not applying the default on the apiserver is reasonable (effectively how runAsUser behaves).  We definitely want to define the default, but we don’t have to fill it except at point of use.  The biggest risk is clients inconsistently interpreting the default, which our current “always default” strategy minimizes.  But as the apply discussion shows, it’s not necessarily an advantage for end users.

Maybe explicit defaulting is something a client should be able to request from the server, although we could only really change that behavior for existing fields until a v2.

lavalamp: I don't think that's the case, to what in the apply discussion are you referring?

Client-applied defaults are ~impossible to ever change in our system, even over version boundaries.

bgrant0607: I still think that defaults explicitly manifested by the server is the right thing. It just needs to be done where the resources are actually reified, not in templates.

mrostecki: So, do we have some agreement there? Are we OK with the field named `disableServiceLinks`, pointer, omitempty, with the default `false` value on the server side? That seems to address @bgrant0607's doubts - it will preserve the old behavior by default and would not require to define that field.

thockin:
@m1093782566 SRV DNS exists for that, if you don't know the port number.

In theory we could add yet another downward API here, but I'd need to see a serious use case for it.

contributors/design-proposals/node/optional-service-links.md
thockin: `enable` is probably a better API.  I think that's what Brian was saying - pointers make this possible.

Any serialized form of an object which has this value specified will retain that value.  Anything that has this value omitted will get `nil` which will be defaulted to whatever the versioned default is.

Right?

So if you serialize it in v1, when you read it back it will deserialize and default.  If you write it back as v2, that v1 default is still retained...

Am I missing something?

thockin:
I still very much want to see this fixed, if you want to carry it forward @m1093782566

thockin:
I feel like we sort of agreed on a single bool to enable service env which defaults to true.

It looks like @mrostecki is awake on https://github.com/kubernetes/kubernetes/pull/60206/files so maybe this is almost done already?

bgrant0607:
I would still like to get this proposal merged, to capture the discussion.

thockin:
Oh for sure, just that I might have false-alarmed on trying to find a new
owner :)

contributors/design-proposals/node/optional-service-links.md
mrostecki: OK, I agree. Let's go with `enable`. I will change that in my implementation.

bradhoekstra:
This feature has been merged in k/k: https://github.com/kubernetes/kubernetes/pull/68754

thockin:
@bradhoekstra - could you take the contents here and update them as a KEP?  We should have done that in the first place...

Then we can close this

bradhoekstra:
KEP PR created here: https://github.com/kubernetes/community/pull/2709

bradhoekstra:
The KEP for this feature is merged.

/close

PR #1267, opened 2017-10-25
Don't use docker by default for sig docs generator

cblecker:
Per https://github.com/kubernetes/test-infra/pull/5149, removing the requirement to run this inside a docker container.

Made a bunch of changes to make this work in just a regular GOPATH. Feedback welcomed, @jamiehannaford :)


jamiehannaford:
Ooo, interesting! I like this idea. Ping me when it's RFR

cblecker:
@jamiehannaford Oh, it's ready for a first pass -- I just marked it WIP because I likely made some mistakes haha. What's there seems to work though.

generator/app.go
```diff
@@ -215,15 +220,15 @@ func createGroupReadme(groups []Group, prefix string) error {
...
+		outputDir := filepath.Join(baseGeneratorDir, group.Dir)
 		if err := createDirIfNotExists(outputDir); err != nil {
 			return err
 		}
 
 		group.SetupGitHubTeams(prefix)
 
 		outputPath := filepath.Join(outputDir, indexFilename)
-		readmePath := fmt.Sprintf("%s_%s", prefix, readmeTemplate)
+		readmePath := fmt.Sprintf("%s/%s_%s", filepath.Join(baseGeneratorDir, templateDir), prefix, readmeTemplate)
```
jamiehannaford: nit: how about `filepath.Join(baseGeneratorDir, templateDir, fmt.Sprintf("%s_%s", prefix, readmeTemplate))`?

cblecker: I like that much better!

cblecker:
@jamiehannaford nit fixed.. also noticed a few other things when I gave it another pass, including making it more self-contained by vendoring the single dependency (which is already vendored by k/k too).

jamiehannaford:
@cblecker instead of vendoring everything, how do you feel about specify a Gopkg.toml? Wondering if we can somehow lower the download burden for folks who only want to contribute docs.

cblecker:
@jamiehannaford It's either ~400K added to the repo, or then we start putting requirements to have the right tools to download the dependencies on the fly. I've got a POC with the latter, but I'm leaning towards the former. Open to thoughts.

jamiehannaford:
@cblecker If we were to go down the latter route, the user would need to run`brew install dep && dep ensure`, then `make build`. Or they could run `make generate-dockerized` if they don't want to fuss around with a dep tool. Maybe that's an okay compromise to ensure that everyone doesn't have to download vendor files they won't necessarily use. I don't really have a strong opinion either way, just wanted to test out alternatives first. 

@jbeda I know you helped out with the generator, what do you think?

hack/verify.sh
```diff
@@ -1,38 +1,44 @@
 #!/bin/bash
 
-export CRTDIR=$(pwd)
-export TMPDIR=/tmp/testgendocs
-mkdir $TMPDIR
+CRT_DIR=$(pwd)
+VERIFY_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t k8s-community.XXXXXX)
```
jbeda: I seem to remember that mktemp has different behavior between BSD/mac and GNU.   Worth testing.

hack/verify.sh
```diff
@@ -1,38 +1,44 @@
...
+mkdir -p ${WORKING_DIR}
 
-cp -r sig* Makefile generator $TMPDIR
+function cleanup {
+  rm -rf "${VERIFY_TEMP}"
+}
+trap cleanup EXIT
 
-cd $TMPDIR
+cp -r sig* Makefile generator ${WORKING_DIR}
```
jbeda: We also have wg-* and we should take this into account here too.  Looks like that perhaps got lots when the wg* stuff was added.

cblecker: Good catch. Fixing.

hack/verify.sh
cblecker: I grabbed this from the ensure-temp-dir function in k/k for just that reason:
https://github.com/kubernetes/kubernetes/blob/be8f1cea8082b03e4322f9f6b474f5ed5ff8679a/hack/lib/util.sh#L79-L84

cblecker:
@jbeda Comments addressed. PTAL

PR #1269, opened 2017-10-25
Use v1.Container in Debug Containers API

verb:
This proposes an API for Debug Containers that's idiomatic for Kubernetes and uses the existing `Container` definition. This makes it easier for admission controllers to gate these updates.

The previous Debug Containers proposal avoided changing the core API as much as possible.

### DIFF
--- a/contributors/design-proposals/node/troubleshoot-running-pods.md
+++ b/contributors/design-proposals/node/troubleshoot-running-pods.md
@@ -1,6 +1,6 @@
 # Troubleshoot Running Pods
 
-*   Status: Pending
+*   Status: Pending Implementation
 *   Version: Alpha
 *   Implementation Owner: @verb
 
@@ -102,33 +102,22 @@ subsequently be used to reattach and is reported by `kubectl describe`.
 ### Kubernetes API Changes
 
 There has been much discussion about how this fits best into the Kubernetes API.
-The consensus is for an imperative "debug this pod" action that's implemented
-mostly in the kubelet. In order to avoid new dependencies in the kubelet, this
-will be implemented in the Core API. Three possible implementations follow, and
-additional implementations that were evaluated and dismissed are at the end of
-this document.
+The consensus is for an imperative "debug this pod" action whereby Kubernetes
+creates a new, temporary container in a pod on command. In order to avoid new
+dependencies in the kubelet, this will be implemented in the Core API. Three
+possible implementations follow, and additional implementations that were
+evaluated and dismissed are at the end of this document.
 
 All of the proposed solutions implement the user-level concept of a _Debug
 Container_ using the API-level concept of an _Ephemeral Container_. The API
 doesn't prescribe how an Ephemeral Container is used. It could conceivably see
 use other than Debug Containers, but we don't currently have other use cases.
 
-#### Chosen Solution: POST to an Existing Pod
+#### Chosen Solution: Subresource to Update PodStatus
 
-We're modifying an existing pod, so this fits as a subresource of the target
-pod. We will create a new top-level object that contains a `v1.Container` and
-`POST` to the subresource to create the Debug Container. Since this is a `POST`,
-we cannot upgrade the connection to streaming if we want to continue supporting
-web socket clients.
-
-Rather than using an existing subresource like `/exec` and conditionally stream
-based on `PodExecOptions`, we will create a new subresource with consistent
-streaming behavior. A new subresource has the added benefit of being able to
-entirely hide interface entirely behind a feature flag by conditionally
-registering the new subresource.
-
-A `v1.Container` by itself lacks type and object metadata, so we will create a
-new type:
+An Ephemeral Container is not part of the pod specification as it's not part of
+the intended state of the pod, but we describe it using the same primitives as
+`PodSpec`. An `EphemeralContainer` contains a Spec, a Status and a Target:
 
 ```
 // EphemeralContainer describes a container to attach to a running pod for troubleshooting.
@@ -152,26 +141,37 @@ type EphemeralContainer struct {
 }
 ```
 
+All Ephemeral Containers that have been created in a pod are listed in the pod's
+status:
+
+```
+type PodStatus struct {
+        ...
+        // List of user-initiated ephemeral containers that have been run in this pod.
+        // +optional
+        EphemeralContainers []EphemeralContainer `json:"commands,omitempty" protobuf:"bytes,11,rep,name=ephemeralContainers"`
+
+}
+```
+
+To create a new Ephemeral Container, one appends a new `EphemeralContainer` with
+the desired `v1.Container` as `Spec` in `Pod.Status` and updates the `Pod` in
+the API. This is accomplished via a new subresource, `/ephemeralcontainers`,
+which enforces the append-only semantics and authorization. This is similar to
+the `/status` subresource used by the kubelet to modify pod status.
+
 **Note that Ephemeral Containers are not regular containers and should not be
 used to build services.** They lack guarantees for resources or execution, and
 many of the fields of `v1.Container` will not be allowed for Debug Containers. A
-request will be rejected if any field is set other than the following
+pod update will fail validation if any field is set other than the following
 whitelisted fields: `Name`, `Image`, `Command`, `Args`, `WorkingDir`, `Env`,
 `EnvFrom`, `ImagePullPolicy`, `SecurityContext`. `TTY` and `Stdin` are always
 enabled for Debug Containers and will be ignored.
 
-The new `/ephemeralcontainers` subresource allows the following:
-
-1.  A `POST` of a `EphemeralContainer` to
-    `/api/v1/namespaces/$NS/pods/$POD_NAME/ephemeralcontainers` to create an
-    Ephemeral Container running in pod `$POD_NAME`.
-1.  Support for stopping an Ephemeral Container **could be supported in the
-    future** by a `DELETE` of
-    `/api/v1/namespaces/$NS/pods/$POD_NAME/ephemeralcontainers/$NAME`.
-
-Once created, it is the responsibility of the client to watch for the
-`EphemeralContainer` to appear in the `PodStatus` and then attach to the console
-of a debug container using the existing attach endpoint,
+Once the pod object is updated, the kubelet worker watching this pod will launch
+the Ephemeral Container and update its status. The client is expected to watch
+for the creation of the container status and then attach to the console of a
+debug container using the existing attach endpoint,
 `/api/v1/namespaces/$NS/pods/$POD_NAME/attach`. Note that any output of the new
 container between its creation and subsequent attach will not be replayed and
 can only be viewed using `kubectl log`.
@@ -211,7 +211,7 @@ exec`. To kill an Ephemeral Container one would `attach` and exit the process
 interactively or create a new Ephemeral Container to send a signal with
 `kill(1)` to the original process.
 
-#### Alternative 2: Declarative Configuration
+#### Alternative 2: Ephemeral Container Controller
 
 Using subresources is an imperative style API where the client instructs the
 kubelet to perform an action, but in general Kubernetes prefers declarative APIs
@@ -244,31 +244,16 @@ enforce, and SIG Node strongly prefers to minimize kubelet complexity.
 
 ### Ephemeral Container Status
 
-`EphemeralContainer` is included in a new field in `PodStatus`:
-
-```
-type PodStatus struct {
-        ...
-        // List of user-initiated ephemeral containers that have been run in this pod.
-        // +optional
-        EphemeralContainers []EphemeralContainer `json:"commands,omitempty" protobuf:"bytes,11,rep,name=ephemeralContainers"`
-
-}
-```
-
-The kubelet should be able to construct a complete `PodStatus` with no prior
-state using information stored in the container runtime.
-`EphemeralContainer.Status` introduces no new data, but the kubelet must also
-now populate `EphemeralContainer.Spec` &
-`EphemeralContainer.TargetContainerName`.
-
-The kubelet already persists container metadata as CRI
+We wish for the kubelet to be able to construct `PodStatus` without relying on
+prior state, so we will store `EphemeralContainer.Spec` &
+`EphemeralContainer.TargetContainerName` as runtime metadata. The kubelet
+currently persists container metadata as CRI
 [labels](https://github.com/kubernetes/kubernetes/blob/v1.10.0-alpha.0/pkg/kubelet/apis/cri/v1alpha1/runtime/api.proto#L606)
 and
 [annotations](https://github.com/kubernetes/kubernetes/blob/v1.10.0-alpha.0/pkg/kubelet/apis/cri/v1alpha1/runtime/api.proto#L613).
 The entire v1.Container used in the request will be serialized and stored as a
 runtime annotation. The value of `TargetContainerName` will be stored as a
-runtime label. Persisting this data in the runtime means it will survive kubelet
+runtime label. Persisting this data in the runtime means it survives kubelet
 restarts.
 
 At least for the Docker runtime, this is [an intended use of docker
@@ -279,62 +264,35 @@ size. Because the container spec may be examined in security sensitive contexts
 like admission control, we will conservatively limit the size of the spec to 32K
 and add a 32K minimum label length test to runtime qualification.
 
-`EphemeralContainer` is populated by the kubelet in the same way as regular
-container statuses. This is sent to the API server and displayed by `kubectl
-describe pod`.
+`EphemeralContainer.Status` is populated by the kubelet in the same way as
+regular container statuses. This is sent to the API server and displayed by
+`kubectl describe pod`.
 
 ### Creating Debug Containers
 
-1.  `kubectl` invokes the new API as described in the preceding section.
-1.  The API server checks for name collisions with existing running containers
-    (in both `PodSpec` and `PodStatus.EphemeralContainers`), performs
-    validation, admission control and proxies the connection to the kubelet's
-    `/ephemeralContainers/$NS/$POD_NAME` endpoint.
-    1.  Since a name collision could happen in the interval between container
-        creation and PodStatus being published to the API server, the kubelet
-        will perform an additional check for name collision.
-    1.  It is permissible to replace an exited container with one of the same
-        name.
-1.  The kubelet request handler opens an error channel and signals the pod's
-    sync worker with `UpdatePodOptions` that include the `EphemeralContainer` in
-    a new field and a callback in the existing
-    `UpdatePodOptions.OnCompleteFunc`.
-    1.  The pod sync worker runs the existing `syncPod()` with a new
-        `SyncPodType` of `SyncPodDebug`.
-    1.  The request handler blocks (with timeout) on receiving an error from
-        `syncPod()` via the callback. During this time, `syncPod()` starts the
-        ephemeral container, including fetching an image if necessary, and
-        publishes a new `PodStatus`.
-    1.  Timeout is configured by the cluster administrator and defaults to 2
-        minutes.
-1.  `syncPod()` again checks for container name collision and starts an
-    ephemeral container via the new `kuberuntime.StartEphemeralContainer()`.
-    1.  `StartEphemeralContainer()` call uses the existing `startContainer()`
-        method, which gains support for targeting the namespaces of a container
-        by name.
-    1.  `syncPod()` runs only from a dedicated pod worker, resolving any races
-        for container creation.
+1.  `kubectl` constructs `EphemeralContainer.Spec` and
+    `EphemeralContainer.TargetContainerName` based on command line arguments. It
+    `PUT`s the modified pod to the pod's `/ephemeralcontainers`.
+1.  The apiserver discards changes to fields other than
+    `Pod.Status.EphemeralContainers` and validates the pod update.
+    1.  Update validation fails if existing Ephemeral Containers are removed or
+        changed, or if the new Ephemeral Container has a non-empty status.
+    1.  Pod validation fails if container spec contains fields disallowed for
+        Ephemeral Containers, has the same name as a container in the spec, or
+        has the same name as another running Ephemeral Container. (see below)
+    1.  API resource versioning resolves update races.
+1.  The kubelet's pod watcher notices the update and triggers a `syncPod()`.
+    During the sync, the kubelet calls `kuberuntime.StartEphemeralContainer()`
+    for any Ephemeral Container with an empty status.
+    1.  `StartEphemeralContainer()` uses the existing `startContainer()` method,
+        which gains support for targeting the namespaces of a container by name.
     1.  After initial creation, future invocations of `syncPod()` will publish
         its ContainerStatus but otherwise ignore the Ephemeral Container. It
         will exist for the life of the pod sandbox or it exits and is garbage
         collected. In no event will it be restarted.
-1.  `syncPod()` then finishes a regular sync, publishing an updated PodStatus
-    (which includes the new `EphemeralContainer`) by its normal, existing means.
-    The pod worker sends its exit status to the request worker.
-1.  The request worker receives a (hopefully `nil`) `error` and returns it to
-    the client.
-    1.  `OnCompleteFunc` is not guaranteed to be called, so if the request
-        worker times out it will check the pod's `PodStatus` to see if the Debug
-        Container was started prior to returning an error. If this happens the
-        `EphemeralContainer.Spec` must be compared to verify it was the same one
-        as requested.
+1.  `syncPod()` finishes a regular sync, publishing an updated PodStatus (which
+    includes the new `EphemeralContainer`) by its normal, existing means.
 1.  The client performs an attach to the debug container's console.
-
-The apiserver detects container name collisions with both containers in the pod
-spec and other running Debug Containers by checking
-`PodStatus.EphemeralContainers`. In a race to create two Debug Containers with
-the same name, the API server will pass both requests and the kubelet will
-reject one in the synchronized pod worker.
 
 There are no limits on the number of Debug Containers that can be created in a
 pod, but exceeding a pod's resource allocation may cause the pod to be evicted.
@@ -342,9 +300,9 @@ pod, but exceeding a pod's resource allocation may cause the pod to be evicted.
 ### Restarting and Reattaching Debug Containers
 
 Debug Containers will never be restarted automatically. It is possible to
-replace a Debug Container that has exited by re-using a Debug Container name. It
-is an error to attempt to replace a Debug Container that is still running, which
-is detected by both the API server and the kubelet.
+"restart" a Debug Container by by re-using the name of a Debug Container that
+has exited. It is an error to re-use the name of a Debug Container that is still
+running, which is detected by API server validation.
 
 One can reattach to a Debug Container using `kubectl attach`. When supported by
 a runtime, multiple clients can attach to a single debug container and share the
@@ -352,7 +310,7 @@ terminal. This is supported by Docker.
 
 ### Killing Debug Containers
 
-Debug containers will not be killed automatically until the pod (specifically,
+Debug containers will not be killed automatically unless the pod (specifically,
 the pod sandbox) is destroyed. Debug Containers will stop when their command
 exits, such as exiting a shell. Unlike `kubectl exec`, processes in Debug
 Containers will not receive an EOF if their connection is interrupted.
@@ -367,23 +325,16 @@ are necessary in the kubelet:
     the pod spec.
 1.  As an exception to the above, `SyncPod()` will kill Debug Containers when
     the pod sandbox changes since a lone Debug Container in an abandoned sandbox
-    is not useful. Debug Containers are not automatically started in the new
+    is not useful. Debug Containers are not started automatically in the new
     sandbox.
 1.  `convertStatusToAPIStatus()` must sort Debug Containers status into
-    `EphemeralContainerStatuses` similar to as it does for
+    `EphemeralContainer.Status` similar to as it does for
     `InitContainerStatuses`
-1.  The kubelet must preserve `ContainerStatus` on debug containers for
-    reporting.
 1.  Debug Containers must be excluded from calculation of pod phase and
     condition
 
-It's worth noting some things that do not change:
-
-1.  `KillPod()` already operates on all running containers returned by the
-    runtime.
-1.  Containers created prior to this feature being enabled will have a
-    `containerType` of `""`. Since this does not match `"EPHEMERAL"` the special
-    handling of Debug Containers is backwards compatible.
+`KillPod()` already operates on all running containers returned by the runtime
+and requires no changes
 
 ### Security Considerations
 
@@ -392,10 +343,8 @@ Debug Containers have no additional privileges above what is available to any
 spec except that it is created on demand.
 
 Admission plugins must be updated to guard `/ephemeralcontainers`. In
-particular, they should enforce the same container image policy on the `Image`
-parameter as is enforced for regular containers. During the alpha phase we will
-additionally support a container image whitelist as a kubelet flag to allow
-cluster administrators to easily constrain debug container images.
+particular, they should enforce the same container image policy on the
+`EphemeralContainer.Spec` parameter as is enforced for regular containers.
 
 ### Additional Consideration
 
@@ -421,7 +370,7 @@ cluster administrators to easily constrain debug container images.
 
 #### Goals and Non-Goals for Alpha Release
 
-We're targeting an alpha release in Kubernetes 1.10 that includes the following
+We're targeting an alpha release in Kubernetes 1.11 that includes the following
 basic functionality:
 
 *   Support in the kubelet for creating debug containers in a running pod
@@ -429,14 +378,15 @@ basic functionality:
 *   `kubectl describe pod` will list status of debug containers running in a pod
 
 Functionality will be hidden behind an alpha feature flag and disabled by
-default. The following are explicitly out of scope for the alpha release, but
-must be resolved prior to beta release:
+default.
 
-*   There's no guarantee that exited Debug Containers won't be garbage collected
-    as regular containers, so they may disappear from the list of
-    `EphemeralContainers`.
-*   We could improve reliability of `UpdatePodOptions.OnCompleteFunc` by
-    prioritizing based on `SyncPodType`.
+Since the kubelet stores Debug Container metadata as runtime labels, it's lost
+when Debug Containers are garbage collected. For the alpha release we will rely
+on the apiserver to store the `EphemeralContainer` for garbage collected
+containers. The kubelet will preserve any `EphemeralContainer` it doesn't
+recognize when updating status. In the event that a `PodStatus` is lost and we
+need to regenerate it from scratch, `EphemeralContainers` will only contain
+Debug Containers that have not been garbage collected.
 
 #### Kubernetes API Changes
 
@@ -444,8 +394,7 @@ The following changes must be implemented in the API:
 
 1.  `v1.EphemeralContainer` will be added and `v1.PodStatus` will be extended as
     described above.
-1.  The new subresource will be added to the pods API, validation added and
-    proxied to the kubelet.
+1.  The new subresource will be added to the pods API.
 1.  The API server must check for Ephemeral Containers when validating `attach`.
 
 #### kubelet Implementation
@@ -462,19 +411,16 @@ following steps:
     be identified as a debug container.
 1.  `kuberuntimemanager` gains a new `StartEphemeralContainer()` which calls the
     existing `startContainer()`.
-1.  The kubelet gains a `RunDebugContainer()` method which accepts a
-    `v1.EphemeralContainer` and triggers a pod sync to create the debug
-    container. The existing `generateAPIPodStatus()` will be update to also
-    populate `EphemeralContainers`.
-1.  The kubelet API gains the new `/ephemeralContainers/` endpoint to create the
-    Debug Container.
+1.  `syncPod()` will call `StartEphemeralContainer()` to start the Debug
+    Container. The existing `generateAPIPodStatus()` will be updated to also
+    populate `EphemeralContainers.Status`.
 
 #### kubectl changes
 
 In anticipation of this change, [#46151](https://pr.k8s.io/46151) added a
 `kubectl alpha` command to contain alpha features. We will add `kubectl alpha
 debug` to invoke Debug Containers. `kubectl` does not use feature gates, so
-`kubectl alpha debug` will be visible by default in `kubectl` 1.10 and return an
+`kubectl alpha debug` will be visible by default in `kubectl` 1.11 and return an
 error when used on a cluster with the feature disabled.
 
 `kubectl describe pod` will report the contents of `EphemeralContainers` when
--- a/contributors/design-proposals/node/troubleshoot-running-pods.md
+++ b/contributors/design-proposals/node/troubleshoot-running-pods.md
@@ -1,6 +1,6 @@
 # Troubleshoot Running Pods
 
-*   Status: Pending Implementation
+*   Status: Implementing
 *   Version: Alpha
 *   Implementation Owner: @verb
 
@@ -45,7 +45,7 @@ A solution to troubleshoot arbitrary container images MUST:
 *   fetch troubleshooting utilities at debug time rather than at the time of pod
     creation
 *   be compatible with admission controllers and audit logging
-*   allow discovery of debugging status
+*   allow discovery of current debugging status
 *   support arbitrary runtimes via the CRI (possibly with reduced feature set)
 *   require no administrative access to the node
 *   have an excellent user experience (i.e. should be a feature of the platform
@@ -62,10 +62,9 @@ Whereas `kubectl exec` runs a _process_ in a _container_, `kubectl debug` will
 be similar but run a _container_ in a _pod_.
 
 A container created by `kubectl debug` is a _Debug Container_. Just like a
-process run by `kubectl exec`, a Debug Container is not part of the pod spec and
-has no resource stored in the API. Unlike `kubectl exec`, a Debug Container
-_does_ have status that is reported in `v1.PodStatus` and displayed by `kubectl
-describe pod`.
+process run by `kubectl exec`, a Debug Container is not part of the pod spec.
+Unlike `kubectl exec`, a Debug Container _does_ have status that is reported in
+`v1.PodStatus` and displayed by `kubectl describe pod`.
 
 For example, the following command would attach to a newly created container in
 a pod:
@@ -102,11 +101,11 @@ subsequently be used to reattach and is reported by `kubectl describe`.
 ### Kubernetes API Changes
 
 There has been much discussion about how this fits best into the Kubernetes API.
-The consensus is for an imperative "debug this pod" action whereby Kubernetes
-creates a new, temporary container in a pod on command. In order to avoid new
-dependencies in the kubelet, this will be implemented in the Core API. Three
-possible implementations follow, and additional implementations that were
-evaluated and dismissed are at the end of this document.
+The consensus is for an imperative "debug this pod" action whereby the kubelet
+creates a new, temporary container in a pod on command. SIG Node would like to
+avoid new dependencies in the kubelet, so this will be implemented in the Core
+API. Three possible implementations follow, and additional implementations that
+were evaluated and dismissed are at the end of this document.
 
 All of the proposed solutions implement the user-level concept of a _Debug
 Container_ using the API-level concept of an _Ephemeral Container_. The API
@@ -116,8 +115,8 @@ use other than Debug Containers, but we don't currently have other use cases.
 #### Chosen Solution: Subresource to Update PodStatus
 
 An Ephemeral Container is not part of the pod specification as it's not part of
-the intended state of the pod, but we describe it using the same primitives as
-`PodSpec`. An `EphemeralContainer` contains a Spec, a Status and a Target:
+the declared state of the pod, but we describe it using the same primitives as
+in `PodSpec`. An `EphemeralContainer` contains a Spec, a Status and a Target:
 
 ```
 // EphemeralContainer describes a container to attach to a running pod for troubleshooting.
@@ -141,8 +140,7 @@ type EphemeralContainer struct {
 }
 ```
 
-All Ephemeral Containers that have been created in a pod are listed in the pod's
-status:
+Ephemeral Containers for a pod are listed in the pod's status:
 
 ```
 type PodStatus struct {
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+type EphemeralContainer struct {
+        metav1.TypeMeta
+        metav1.ObjectMeta
+
+        // Spec describes the Ephemeral Container to be created.
+        Spec Container `json:"spec" ...`
+}
+```
+
+Note that Debug Containers are not regular containers and should not be used to
```
liggitt: s/debug/ephemeral/g, based on where the review of the last PR ended up?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+build services. They have no guarantees and most of the fields of `v1.Container`
+will not be allowed for Debug Containers. A request will be rejected if any
+field is set other than the following whitelisted fields: `Name`, `Image`,
+`Command`, `Args`, `WorkingDir`, `Env`, `EnvFrom`, `ImagePullPolicy`,
+`SecurityContext`. `TTY` and `Stdin` are always enabled for Debug Containers and
```
liggitt: Running an ephemeral container with non-interactive diagnostics and viewing the log would not require tty/stdin. Are there any containers where enabling tty/stdin would cause problems?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+We're modifying an existing pod, so this fits as a subresource of the target
+pod. Since we're required to use `v1.Container`, the API should use `POST` or
+`PUT` to avoid encoding this relatively large object into a URL query string. If
+we `POST` or `PUT` then we cannot make this a streaming call if we want to
+continue supporting web socket clients.
+
+We can either using an existing subresource like `/exec` and have the call
+stream or not based on `PodExecOptions`, or we can create a new subresource that
+never streams. For the sake of consistency, we will create a new subresource
+named `/debug`. This has the added benefit of being able to entirely hide
```
liggitt: A separate subresource makes a lot of sense to me. Adding a newly defined container (new image, new security settings, etc) is very different from exec, defining the API we want for it seems like a good idea.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+`Command`, `Args`, `WorkingDir`, `Env`, `EnvFrom`, `ImagePullPolicy`,
+`SecurityContext`. `TTY` and `Stdin` are always enabled for Debug Containers and
+will be ignored.
+
+The new `/debug` subresource allows the following:
+
+1.  A `POST` of a `PodDebugContainer` to
+    `/api/v1/namespaces/$NS/pods/$POD_NAME/debug` to create a Debug Container
+    running in pod `$POD_NAME`.
+1.  A `DELETE` of `/api/v1/namespaces/$NS/pods/$POD_NAME/debug/$NAME` will stop
```
liggitt: Named subresources are a thing I have wanted, but not something we currently have, which could make this tricky. I'd see this as nice-to-have, but not essential (as noted in the previous PR, things like exec sessions are not currently able to be individually terminated via the API)

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: +1

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: `/execEphemeralContainer` -- anything other than `/debug`.

A new endpoint seems good for the super-security conscious at the expense of making everything more complicated for everyone else. I don't have a strong opinion.

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: Yeah I don't see why you'd mandate those features on?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+
+Note that Debug Containers are not regular containers and should not be used to
+build services. They have no guarantees and most of the fields of `v1.Container`
+will not be allowed for Debug Containers. A request will be rejected if any
+field is set other than the following whitelisted fields: `Name`, `Image`,
+`Command`, `Args`, `WorkingDir`, `Env`, `EnvFrom`, `ImagePullPolicy`,
+`SecurityContext`. `TTY` and `Stdin` are always enabled for Debug Containers and
+will be ignored.
+
+The new `/debug` subresource allows the following:
```
lavalamp: s/`/debug`/something else/g

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: I suggest allowing only one ephemeral container at a time for the moment, and then you don't need the additional level? We don't support this at the moment.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+    `/api/v1/namespaces/$NS/pods/$POD_NAME/debug` to create a Debug Container
+    running in pod `$POD_NAME`.
+1.  A `DELETE` of `/api/v1/namespaces/$NS/pods/$POD_NAME/debug/$NAME` will stop
+    the Debug Container `$NAME` in pod `$POD_NAME`.
+
+Once created, a client would attach to the console of a debug container using
+the existing attach endpoint, `/api/v1/namespaces/$NS/pods/$POD_NAME/attach`.
+Note that any output of the new container between its creation and subsequent
+attach will not be replayed and can only be viewed using `kubectl log`.
+
```
lavalamp: Will the ephemeral container appear in the response when I GET the pod?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -441,10 +448,11 @@ The kubelet exposes the new functionality in its existing `/exec/` endpoint.
 `ServeExec()` constructs a `v1.Container` based on `PodExecOptions`, calls
 `RunDebugContainer()`, and performs the attach.
 
-##### Step 4: Reporting DebugStatus
+##### Step 4: Reporting EphemeralContainerStatus
```
lavalamp: Ah, looks like the answer is yes.

lavalamp:
This LGTM to me if you can fix the naming (consistency + debug -> ephemeral).

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: alternately, `DELETE` of `/api/v1/namespaces/$NS/pods/$POD_NAME/ephemeralcontainers` stops all ephemeral containers

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -235,37 +249,28 @@ The status of a Debug Container is reported in a new field in `v1.PodStatus`:
...
-in tracking status for traditional exec in a similar manner. Ideally we can
-report both types of user intervention into a container with a single new type.
+This status is only populated for Debug Containers, but there's interest in
+tracking status for traditional exec in a similar manner.
 
-Note that `Command` and `Args` must be tracked in the status object because
-there is no spec for Debug Containers or exec. These must either be made
+Note that `Command` and `Args` would have to be tracked in the status object
+because there is no spec for Debug Containers or exec. These must either be made
 available by the runtime or tracked by the kubelet. For Debug Containers this
```
liggitt: doesn't the whole ephemeral container have to be tracked by the kubelet?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+        // Spec describes the Ephemeral Container to be created.
+        Spec Container `json:"spec" ...`
+}
+```
+
+Note that Debug Containers are not regular containers and should not be used to
+build services. They have no guarantees and most of the fields of `v1.Container`
+will not be allowed for Debug Containers. A request will be rejected if any
+field is set other than the following whitelisted fields: `Name`, `Image`,
+`Command`, `Args`, `WorkingDir`, `Env`, `EnvFrom`, `ImagePullPolicy`,
```
liggitt: are HostNetwork/HostPID/HostIPC inherited from the pod spec?

liggitt: should also document this in a way that doesn't preclude supporting new container fields that make sense in the future

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: > /execEphemeralContainer -- anything other than /debug.

probably want lowercase resources, but yes

> A new endpoint seems good for the super-security conscious at the expense of making everything more complicated for everyone else. I don't have a strong opinion.

it's good for isolating the function for policy purposes, but also lets us support envvars, capabilities, and other useful knobs for shaping the environment of the ephemeral container in a way consistent with existing pods

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: only the status of it, not the spec

contributors/design-proposals/node/troubleshoot-running-pods.md
thockin: Agree we can live without it for now

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +107,63 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+interface behind a feature flag by conditionally registering the new
+subresource.
+
+As `v1.Container` lacks type and object metadata, we must wrap it in a new
+object:
+
+```
+// EphemeralContainer describes a container to attach to a running pod for troubleshooting.
+type EphemeralContainer struct {
+        metav1.TypeMeta
```
thockin: I assume the name has to be unique within the pod?

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: > probably want lowercase resources

er, right

> it's good for isolating the function for policy purposes, but also lets us support envvars, capabilities, and other useful knobs for shaping the environment of the ephemeral container in a way consistent with existing pods

I stand by my previous response :)

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: That's fine.

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: Yeah, I think it's a possibility that we'll want to support more of these fields as time goes on.

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: Yes, but the container name would be embedded in the ContainerSpec.

This type probably doesn't need an ObjectMeta--I would try it that way and see if any of the encoding stack gives you a problem.

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: why just the command and args? admission uses the definitions of containers to determine whether a user can do things like exec/attach to those containers ("cannot exec into a pod with containers running as root", etc). Container status alone (`EphemeralContainerStatuses []v1.ContainerStatus`) wouldn't include enough information about the ephemeral containers to make that determination.

what if the submitted ephemeral container and its status were both made available in pod status?

```
type PodStatus struct {
  ...
  EphemeralContainers []v1.EphemeralContainer
}

type EphemeralContainer struct {
  Spec v1.Container
  Status v1.ContainerStatus
}
```

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -235,37 +249,28 @@ The status of a Debug Container is reported in a new field in `v1.PodStatus`:
...
 
-Note that `Command` and `Args` must be tracked in the status object because
-there is no spec for Debug Containers or exec. These must either be made
+Note that `Command` and `Args` would have to be tracked in the status object
+because there is no spec for Debug Containers or exec. These must either be made
 available by the runtime or tracked by the kubelet. For Debug Containers this
 could be stored as runtime labels, but the kubelet currently has no method of
 storing state across restarts for exec. Solving this problem for exec is out of
 scope for Debug Containers, but we will look for a solution as we implement this
 feature.
```
liggitt: @smarterclayton was there something were you mentioning that allowed connecting data to running containers?

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Yes, it must be unique within the pod because it will be used for attach, logs, etc. Thanks for the tip on ObjectMeta, I will try it.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Actually, the namespaces will be inherited from the target container. Originally it was going to be inherited from the pod spec. This is something that changed a few weeks ago because we're now going to support isolated PID namespaces indefinitely, and it's not reflected in this API alternative.

With exec++ the target container was provided in PodExecOptions, but switching the API means we either have to provide the target in the URI, the posted EphemeralContainer or... query params? I haven't looked into this at all yet.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: iirc this came from a conversation from @thockin and @vishh about how having configurable stdin/tty has caused confusion for users and been more trouble than it was worth. I don't feel strongly.

Some containers don't require tty/stdin, but is there a downside to them being always on? This is a new feature, so there aren't backwards compatibility concerns here.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: SGTM, I will omit it for now.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: The kubelet currently has no method of keeping state of the v1.Container across restarts. After a restart, it can reconstruct a minimum v1.Container based on runtime labels, though. This could (but currently doesn't) include Command and Args, which can be queried from the run time.

Including the spec in the PodStatus would be pretty great, though, so it might be worth solving. @lavalamp what do you think about this?

lavalamp: Remind me why this got reopened for discussion? I feel like if we keep going down this route we will end up at my original position of adding a `type` of `ephemeral` to containers, and allowing users to add or remove such containers from the spec.

verb: I think sig-auth just wants to make this works well with admission controllers and has a strong preference for using v1.Container, [for many good reasons](https://github.com/kubernetes/community/pull/649/files#r142722579).

@liggitt This is an important feature to sig-node, but we've been stuck for about a year. If a v1.Container is available to admission controllers for the creation request, can you live with only ContainerStatus for attach, at least for alpha? We don't need to allow exec.

liggitt: > The kubelet currently has no method of keeping state of the v1.Container across restarts.

I expected the kubelet to:
1. update pod status with the ephemeral container definition/status as it created the container
2. update the ephemeral container status by name from observed container state during normal sync
3. remove the ephemeral container from the pod status when the ephemeral container was removed

None of that requires the kubelet remembering the ephemeral container definition beyond the initial create, right?

> If a v1.Container is available to admission controllers for the creation request, can you live with only ContainerStatus for attach, at least for alpha? We don't need to allow exec.

Two thoughts:
1. attach (which is the intended use case here) requires almost identical protections to exec ("cannot attach to a pod with containers running as root", etc). If the create of the ephemeral container and the attach to it are separate API calls, but the container definition is unavailable at attach time, I'm unsure how that would be implementable.
2. the structure of an ephemeral container field in the pod status could not be changed [in an incompatible way](https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#adding-unstable-features-to-stable-versions) (*if the schema of the alpha field must change in an incompatible way, a new field name must be used*), so if we started with `EphemeralContainerStatuses []v1.ContainerStatus`, that couldn't grow to include the container definition as alpha progressed.

smarterclayton:
I'd like to get this into the alpha slot for 1.10.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +101,76 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
-for all clients. We don't want to encode a (potentially large) `v1.Container` as
-an HTTP parameter, so we must extend `v1.PodExecOptions` with the specific
-fields required for creating a Debug Container:
+        // Spec describes the Ephemeral Container to be created.
+        Spec Container `json:"spec" ...`
+
+        // If specified, the ephemeral container will be started in the namespaces
+        // of this TargetContainerName. If omitted the container will inherit the
+        // behavior of the pod for namespaces.
+        TargetContainerName string `json:"spec" ...`
```
liggitt: copy/pasta, `json:"targetContainerName"`

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -437,14 +453,15 @@ Containers unless the sandbox is restarted.
 
 ##### Step 3: kubelet API changes
 
-The kubelet exposes the new functionality in its existing `/exec/` endpoint.
+The kubelet exposes the new functionality in a new `/podDebug/` endpoint.
```
liggitt: needs updating to indicate it is given the v1.Container and optional target container name

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -387,14 +407,10 @@ basic functionality:
 *   `kubectl describe pod` will list status of debug containers running in a pod
 
 Functionality will be hidden behind an alpha feature flag and disabled by
-default. The following are explicitly out of scope for the 1.9 alpha release:
+default. The following are explicitly out of scope for the 1.10 alpha release:
 
 *   Exited Debug Containers will be garbage collected as regular containers and
     may disappear from the list of Debug Container Statuses.
```
liggitt: > Exited Debug Containers will be garbage collected as regular containers and may disappear from the list of Debug Container Statuses.

harmonize this with the "The list of `EphemeralContainerStatuses` is never truncated." statement below

liggitt:
> I'd like to get this into the alpha slot for 1.10.

Agree. API-wise, this seems to have the info I would expect now (v1.Container in debug request, spec and status persisted to allow gating attach requests).

More detail is needed for information flow and sequence (which component persists which bits, synchronously or asynchronously,  etc) and some edge cases (what happens on kubelet restart? kubelet+docker restart?), but that can be a further iteration.

verb:
@liggitt awesome, thanks for the quick feedback. Since this looks to be heading in the right direction for the API, I'll double check it's still acceptable to sig-node and then do a more complete rewrite addressing your comments.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -192,104 +225,86 @@ type EphemeralContainer struct {
...
+runtime label. Persisting this data in the runtime means it will survive kubelet
+restarts.
+
+At least for the Docker runtime, this is [an intended use of docker
+labels](https://docs.docker.com/engine/userguide/labels-custom-metadata/#value-guidelines).
+Docker does not document the maximum length of labels in its API. Empirically,
+it supports up to the 64K constraint of the docker client's `bufio.Scanner`
+size. Because the container spec may be examined in security sensitive contexts
+like admission control, we will conservatively limit the size of the spec to 32K
+and add a 32K minimum label length test to runtime qualification.
```
verb: @mrunalp This is the section where I'd appreciate your review. I'm planning on storing the v1.Container in the runtime. This works for Docker. Do you know of any limitations for cri-o?

verb:
Swapping back in after focusing on kubernetes/features#495. I'll be creating PRs for the features described here over the next week.

I've gone through the doc to make it consistent with our recent changes, and I've added more details about how the kubelet will start the container and avoid races.

@liggitt Do these changes sound good enough (at least for alpha) that I can ask the sig-node-leads to review and approve this PR?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +101,88 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
-fields required for creating a Debug Container:
+**Note that Ephemeral Containers are not regular containers and should not be
+used to build services.** They lack guarantees for resources or execution, and
+many of the fields of `v1.Container` will not be allowed for Debug Containers. A
+request will be rejected if any field is set other than the following
+whitelisted fields: `Name`, `Image`, `Command`, `Args`, `WorkingDir`, `Env`,
+`EnvFrom`, `ImagePullPolicy`, `SecurityContext`. `TTY` and `Stdin` are always
+enabled for Debug Containers and will be ignored.
+
+The new `/add` subresource allows the following:
```
liggitt: is the subresource `/add` or `/ephemeralContainers`? being more explicit seems better (though we don't mix case in resource or subresource names)

verb: It's `/ephemeralcontainers` (I discovered the problem with case on subresources today, too). Fixed.

The kubelet API seems to still mix case, so I left that one alone. sig-node probably has a preference but I'm not sure what it is.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -192,96 +225,100 @@ type EphemeralContainer struct {
...
-    describe pod`.
-1.  The kubelet will upgrade the connection to streaming and attach to the
-    container's console.
-
-Rather than performing the implicit attach the kubelet could return success to
-the client and require the client to perform an explicit attach, but the
-implicit attach maintains consistent semantics across `/exec` rather than
-varying behavior based on parameters.
+    validation, admission control and proxies the connection to the kubelet's
+    `/ephemeralContainers/$NS/$POD_NAME` endpoint.
```
liggitt: it would be helpful to clarify at what point this call completes (e.g. does it block on the syncPod call or return immediately)

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -192,96 +225,100 @@ type EphemeralContainer struct {
...
+
+`EphemeralContainer` is populated by the kubelet in the same way as regular
+container statuses. This is sent to the API server and displayed by `kubectl
+describe pod`.
 
 ### Creating Debug Containers
 
-1.  `kubectl` invokes the exec API as described in the preceding section.
+1.  `kubectl` invokes the new API as described in the preceding section.
 1.  The API server checks for name collisions with existing containers, performs
```
liggitt: I'm assuming the kubelet also checks for name collisions with existing containers or existing debug containers, and that the API server validation prevents name collisions when the kubelet updates pod status with ephemeral container status

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: nm, I see the details below

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -192,96 +225,100 @@ type EphemeralContainer struct {
...
-varying behavior based on parameters.
+    validation, admission control and proxies the connection to the kubelet's
+    `/ephemeralContainers/$NS/$POD_NAME` endpoint.
+1.  The kubelet request handler opens an error channel and signals the pod's
+    sync worker with `UpdatePodOptions` that include the `EphemeralContainer` in
+    a new field and a callback in the existing
+    `UpdatePodOptions.OnCompleteFunc`.
+    1.  The pod sync worker runs the existing `syncPod()` with a new
+        `SyncPodType` of `SyncPodDebug`.
+    1.  The request handler blocks (with timeout) on receiving an error from
```
liggitt: what steps occur within this timeout? pulling the image? how is the timeout selected here?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -336,13 +373,13 @@ It's worth noting some things that do not change:
...
 `v1.Container`. It's the equivalent of configuring an shell container in a pod
-spec but created on demand.
+spec except that it is created on demand.
 
-Admission plugins that guard `/exec` must be updated for the new parameters. In
-particular, they should enforce the same container image policy on the `Image`
-parameter as is enforced for regular containers. During the alpha phase we will
-additionally support a container image whitelist as a kubelet flag to allow
-cluster administrators to easily constraint debug container images.
+Admission plugins must be updated to guard `/debug`. In particular, they should
```
liggitt: fix the subresource name

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -336,13 +373,13 @@ It's worth noting some things that do not change:
...
+enforce the same container image policy on the `Image` parameter as is enforced
+for regular containers. During the alpha phase we will additionally support a
+container image whitelist as a kubelet flag to allow cluster administrators to
+easily constraint debug container images.
```
liggitt: typo: constrain

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -356,38 +393,47 @@ cluster administrators to easily constraint debug container images.
...
+
+The following changes must be implemented in the API:
+
+1.  `v1.EphemeralContainer` will be added and `v1.PodStatus` will be extended as
+    described above.
+1.  The new subresource will be added to the pods api, validation added and
+    proxied to the kubelet.
+1.  The API server must check for Ephemeral Containers when validating `attach`.
+
+New fields will be prepended with `alpha` for the duration of this feature's
```
liggitt: remove this... field name prefixing is not recommended in https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#adding-unstable-features-to-stable-versions

liggitt:
thanks for the update. a couple nits, lgtm otherwise.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: That's correct. Checking for collisions in the api server provides a quick, user-friendly error while the kubelet resolves any races in its dedicated pod worker.

I've updated this section to clarify.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: fixed

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: fixed.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Fixed.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Yes, the container is started, including pulling the image if necessary. I was just going to make the timeout flag configurable. I'd prefer to improve the reliability of the `OnCompleteFunc` for `SyncPodDebug` to making the timeout more configurable. I've described this a little bit in the alpha notes. It should be easy to get to the point where the timeout is rarely used.

verb: Oh, also I updated the doc to reflect this.

liggitt: I'm trying to determine whether an API request to add an ephemeral container should be considered a long running request on the API side. Pulling the image within the lifetime of an API request is a new direction here.

Also, if pulling happens prior to the kubelet persisting the ephemeral container spec in the pod status, does that mean the kubelet gets a single shot at pulling and never reattempts or reports status on failure?

verb: Oh, interesting. Yes, by design we want a failure in starting the ephemeral container to be a permanent failure. There will never be any reattempts, but it should report the failure back to the user, and in the pod's event log.

For example, image pull failures return a ErrImagePull to the user. I'm not sure of the current kubelet behavior for a really slow image pull, but this would be the same for the Ephemeral Container since it uses the same code.

So if len(image_pull_secs) > len(api_request_timeout_secs) then the user will get error("start ephemeral container timed out") even though the ephemeral container will eventually be created (or not). If the Ephemeral Container is eventually created, it will be available for manual attach.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -356,38 +403,44 @@ cluster administrators to easily constraint debug container images.
...
+default. The following are explicitly out of scope for the alpha release, but
+must be resolved prior to beta release:
 
-*   Exited Debug Containers will be garbage collected as regular containers and
-    may disappear from the list of Debug Container Statuses.
-*   Security Context for the Debug Container is not configurable. It will always
-    be run with `CAP_SYS_PTRACE` and `CAP_SYS_ADMIN`.
-*   Image pull policy for the Debug Container is not configurable. It will
-    always be run with `PullAlways`.
+*   There's no guarantee that exited Debug Containers won't be garbage collected
```
yujuhong: What's the expected normal garbage collection behavior? Do debug containers get removed as regular containers?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -356,38 +403,44 @@ cluster administrators to easily constraint debug container images.
...
+    as regular containers, so they may disappear from the list of
```
yujuhong: Is kubelet supposed to continue reporting status for exited debug containers?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -192,96 +225,110 @@ type EphemeralContainer struct {
...
-varying behavior based on parameters.
+    `/ephemeralContainers/$NS/$POD_NAME` endpoint.
+    1.  Since a name collision could happen in the interval between container
+        creation and PodStatus being published to the API server, the kubelet
+        will perform an additional check for name collision.
+1.  The kubelet request handler opens an error channel and signals the pod's
+    sync worker with `UpdatePodOptions` that include the `EphemeralContainer` in
+    a new field and a callback in the existing
+    `UpdatePodOptions.OnCompleteFunc`.
+    1.  The pod sync worker runs the existing `syncPod()` with a new
```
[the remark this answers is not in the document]
verb: The debug containers won't be sync'd, except to update ContainerStatus. After being started it will be ignored until the pod sandbox dies (or is garbage collected). Clarified.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -192,96 +225,110 @@ type EphemeralContainer struct {
...
+        `SyncPodType` of `SyncPodDebug`.
+    1.  The request handler blocks (with timeout) on receiving an error from
+        `syncPod()` via the callback. During this time, `syncPod()` starts the
+        ephemeral container, including fetching an image if necessary, and
+        publishes a new `PodStatus`.
+    1.  Timeout is configured by the cluster administrator and defaults to 15
```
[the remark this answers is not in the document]
verb: No? I'm not sure what to expect for image pulls. Maybe we could set a long (~2 min) timeout and use a ticker to periodically check the PodStatus cache to see if the container was created but `OnCompleteFunc` dropped...

In any case this feels like something on which we can iterate, but you've brought up a really good point that we should initially err on the side of too long. Let's start with 2 minutes.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Last I checked the garbage collector kept the most recent exited container, which would be sufficient for most purposes. I plan on providing knob that cluster admins can use to exempt debug containers from garbage collection for the life of the pod as this was a point of contention for sig node, but before doing so I want to reopen the discussion to determine the exact requirement.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: Yes, it should report the most recent container by name, regardless of status.

contributors/design-proposals/node/troubleshoot-running-pods.md
liggitt: > by design we want a failure in starting the ephemeral container to be a permanent failure

that seems fine

> There will never be any reattempts, but it should report the failure back to the user

I'm having difficulty understanding how this works when combined with the image pull happening within the timeout. Would a pull failure still result in a pod status update? Would it make more sense for the subresource POST to just update pod status with the requested ephemeral container, the kubelet notice that via normal pod watch events, and update pod status with the results of attempting to start the ephemeral container?

That would allow the client side implementation to look like:
1. submit ephemeral subresource request, get back immediate success or error
2. on success, monitor pod status until ephemeral container is running
3. attach

verb:
I've updated this with a change suggested by @liggitt where the client updates the `EphemeralContainer` via the API and the ephemeral container is created during the regular kubelet `syncPod`

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -356,38 +358,44 @@ cluster administrators to easily constraint debug container images.
...
 
 Functionality will be hidden behind an alpha feature flag and disabled by
-default. The following are explicitly out of scope for the 1.9 alpha release:
+default.
+
+Since the kubelet stores Debug Container metadata as runtime labels, it's lost
+when Debug Containers are garbage collected. For the alpha release we will rely
+on the apiserver to store the `EphemeralContainer` for garbage collected
+containers. The kubelet will preserve any `EphemeralContainer` it doesn't
+recognize when updating status. In the event that a `PodStatus` is lost and we
```
yujuhong: In what case would `PodStatus` be lost...?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +101,88 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+`EnvFrom`, `ImagePullPolicy`, `SecurityContext`. `TTY` and `Stdin` are always
+enabled for Debug Containers and will be ignored.
+
+Once the pod object is updated, the kubelet worker watching this pod will launch
+the Ephemeral Container and update its status. The client is expected to watch
+for the creation of the container status and then attach to the console of a
+debug container using the existing attach endpoint,
+`/api/v1/namespaces/$NS/pods/$POD_NAME/attach`. Note that any output of the new
+container between its creation and subsequent attach will not be replayed and
+can only be viewed using `kubectl log`.
```
yujuhong: Does this mean `kubectl log` will be supported for debug containers and is it treated as a regular container? What other commands can be performed against a debug container, e.g., `exec`? I think it's worth calling out what operations will be supported for debug containers.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +100,91 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+
+        // Spec describes the Ephemeral Container to be created.
+        Spec *Container `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+
+        // Most recently observed status of the container.
+        // This data may not be up to date.
+        // Populated by the system.
+        // Read-only.
+        // +optional
+        Status *ContainerStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
```
verb: I'm second guessing my decision to embed the `ContainerStatus` into `EphemeralContainer` as I'm polishing up the kubelet implementation where things just expect to see a `[]ContainerStatus`.  I wonder if it would be better to do something like:

```
type EphemeralContainer struct {
    ...
    Spec *Container
    TargetContainerName string
}

type PodStatus struct {
   ...
   EphemeralContainers []EphemeralContainer
   EphemeralContainerStatuses []ContainerStatus
}
```

I guess that's just pushing the complexity of merging by name onto the client, though, where presumably the spec & status will frequently be operated upon in tandem. If targeting were unnecessary (or could be specified in `Container`) then we wouldn't need the `type EphemeralContainer` at all.

Do you have any suggestions @yujuhong or @liggitt ?

yujuhong: I think the originally proposed `EphemeralContainer` struct that groups everything together might be slightly better to parse.

yujuhong:
/lgtm

/assign @dchen1107 @derekwaynecarr 
for approval

@liggitt @lavalamp @thockin do you still have any concern?

liggitt:
> @liggitt @lavalamp @thockin do you still have any concern?

this describes an approach that ensures:
* the information needed for authz/admission/audit is present in the API requests from the user
* the persisted API object captures what ephemeral container(s) were run
* the ephemeral containers are described using v1.Container structs
so my concerns were addressed

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +100,91 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+#### Chosen Solution: Subresource to Update PodStatus
+
+An Ephemeral Container is not part of the pod specification as it's not part of
+the declared state of the pod, but we describe it using the same primitives as
+in `PodSpec`. An `EphemeralContainer` contains a Spec, a Status and a Target:
+
+```
+// EphemeralContainer describes a container to attach to a running pod for troubleshooting.
+type EphemeralContainer struct {
+        metav1.TypeMeta `json:",inline"`
```
thockin: This makes me think it is a top-level resource - no other sub-structs have their own `kind` fields - why this one?  Can you make sure to explain it in comments, if it is intended?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +100,91 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+        Spec *Container `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+
+        // Most recently observed status of the container.
+        // This data may not be up to date.
+        // Populated by the system.
+        // Read-only.
+        // +optional
+        Status *ContainerStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+
+        // If set, the name of the container from PodSpec that this ephemeral container targets.
```
thockin: I don't understand what "targets" means.  How can a container target a container?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +100,91 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
-We will extend `v1.Pod`'s `/exec` subresource to support "executing" container
-images. The current `/exec` endpoint must implement `GET` to support streaming
-for all clients. We don't want to encode a (potentially large) `v1.Container` as
-an HTTP parameter, so we must extend `v1.PodExecOptions` with the specific
-fields required for creating a Debug Container:
+To create a new Ephemeral Container, one appends a new `EphemeralContainer` with
+the desired `v1.Container` as `Spec` in `Pod.Status` and updates the `Pod` in
+the API. Users cannot normally modify the pod status, so we'll create a new
+subresource `/ephemeralcontainers` that allows an update of solely
+`EphemeralContainers` and enforces append-only semantics.
```
thockin: if it is append-only does that mean no cleanup is possible?  Can we say that?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +100,91 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+All of the proposed solutions implement the user-level concept of a _Debug
+Container_ using the API-level concept of an _Ephemeral Container_. The API
+doesn't prescribe how an Ephemeral Container is used. It could conceivably see
+use other than Debug Containers, but we don't currently have other use cases.
+
+#### Chosen Solution: Subresource to Update PodStatus
+
+An Ephemeral Container is not part of the pod specification as it's not part of
+the declared state of the pod, but we describe it using the same primitives as
+in `PodSpec`. An `EphemeralContainer` contains a Spec, a Status and a Target:
```
lavalamp: what verbs does the subresource support?

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: It is a subresource, like scale.

I don't think this can support anything but POST, is that the intention?

lavalamp: Should we include metadata, and force the user to copy the UID of the pod they are targeting, to prevent accidentally getting a different pod than they intended if there is a delete/recreate?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +100,91 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+```
+
+Ephemeral Containers for a pod are listed in the pod's status:
+
+```
+type PodStatus struct {
+        ...
+        // List of user-initiated ephemeral containers that have been run in this pod.
+        // +optional
+        EphemeralContainers []EphemeralContainer `json:"commands,omitempty" protobuf:"bytes,11,rep,name=ephemeralContainers"`
```
lavalamp: I still strongly disagree with listing a container spec anywhere inside pod status.

lavalamp: I also maintain that you cannot use the same type as the subresource for this in any case-- no typemeta/metadata makes sense here.

contributors/design-proposals/node/troubleshoot-running-pods.md
lavalamp: Users should not modify status, period. Adding a subresource doesn't remove that requirement. The subresource is so it can be permissioned separately. The user intent still should go in spec somewhere.

lavalamp:
Sorry, I think I already said these things on the PR. This came up in my email and I thought it was a recent update but now I think it hasn't been updated recently. Sorry for the noise.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,13 +100,91 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+
+Once the pod is updated, the kubelet worker watching this pod will launch the
+Ephemeral Container and update its status. The client is expected to watch for
+the creation of the container status and then attach to the console of a debug
+container using the existing attach endpoint,
+`/api/v1/namespaces/$NS/pods/$POD_NAME/attach`. Note that output of the new
+container occurring between its creation and attach will not be replayed, but it
+can be viewed using `kubectl log`.
+
+#### Alternative 1: "exec++"
```
vishh: @verb can you add a few lines to describe why these alternatives were not chosen at the end?

contributors/design-proposals/node/troubleshoot-running-pods.md
smarterclayton: We definitely need preconditions to be possible here.  Requiring resourceVersion and uid has been sufficient for all other apis so far.

contributors/design-proposals/node/troubleshoot-running-pods.md
verb: (and I'll update the comment to make this clear)

thockin:
I think that is a broken resolver and a broken DNS config.  You should
never have a host named localhost, and DNS clients should not try to
resolve it

kfox1111:
I agree on both accounts. :)

So, its a bug in golangs resolver then and we need to get golang to fix it.

kfox1111:
one other possibility, as the golang devs have known about the issue for a while and have ignored it, would be to have k8s create /etc/nslookup.conf by default in the container with some sane defaults like it does resolv.conf and hosts.

thockin:
The problem is that nsswitch (I assume that is what you meant) is an
artifact of glibc's DNS resolver.  Are we willing to manage that file in
perpetuity for all users? Also, this will brek any image that has a custom
nsswitch file built-in.

kfox1111:
yeah, thats what I meant. It would suck to maintain it for ever, but it is unclear if anyone else is willing to fix it. Its been a known issue for a long time apparently.

Is it easy to do it conditionally? check if the file exists, and if not inject it?

Alternately, it should be exceedingly rare I think for that file to be non default.  We could make it default to always override it, and add a flag not too.

thockin:
I'm not sure what a good answer here looks like.  It's sort of a broekn
image that depends on it but doesn't provide it.  Can we really be in the
business of fixing broken images dynamically?

Maybe we can instead raise the heat on the Go devs to get a discussion
going?

kfox1111:
the image is not broken in that is provides a libc that never uses it (not glibc). So it doesn't provide a file it isn't going to use. go is assuming it is using glibc and following glibc convention for how to behave when nsswitch is missing. the binary is statically linked though so doesn't actually use libc. The end user is expecting a statically linked binary to work properly on any linux, and is preferring alpine as its smaller and we're encouraging smaller containers. So, its hard to point at any one thing as the thing to fix, but it is unexpectedly effecting users. We can raise it with the go devs again and see what they say.

thockin:
Please feel free to loop me into a discussion with go devs.

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -45,26 +45,25 @@ A solution to troubleshoot arbitrary container images MUST:
...
+*   v1.Container must be available for inspection by admission controllers
 
 ## Feature Summary
 
 Any new debugging functionality will require training users. We can ease the
 transition by building on an existing usage pattern. We will create a new
 command, `kubectl debug`, which parallels an existing command, `kubectl exec`.
-Whereas `kubectl exec` runs a *process* in a *container*, `kubectl debug` will
-be similar but run a *container* in a *pod*.
+Whereas `kubectl exec` runs a _process_ in a _container_, `kubectl debug` will
```
krmayankk: i think this is underdocumented here https://kubernetes.io/docs/tasks/debug-application-cluster/get-shell-running-container/ that kubectl exec runs a process in an existing container

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -82,22 +81,16 @@ kubectl debug target-pod
 
 This creates an interactive shell in a pod which can examine and signal other
 processes in the pod. It has access to the same network and IPC as processes in
-the pod. It can access the filesystem of other processes by `/proc/$PID/root`.
-As is already the case with regular containers, Debug Containers can enter
-arbitrary namespaces of another container via `nsenter` when run with
-`CAP_SYS_ADMIN`.
+the pod. When [process namespace sharing](https://features.k8s.io/495) is
```
krmayankk: the top line should make it more clear that  this creates a new container and runs an interactive shell inside that container . `This creates a new container and starts an interactive shell inside that container in a pod called target-pod which can examine and signal other processes in the pod`

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,243 +99,180 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
-        // Image is an optional container image name that will be used to for the Debug
-        // Container in the specified Pod with Command as ENTRYPOINT. If omitted a
-        // default image will be used.
-        Image string `json:"image,omitempty" ...`
+type PodSpec struct {
+  ...
+  // List of user-initiated ephemeral containers to run in this pod.
+  // This field is alpha-level and is only honored by servers that enable the EphemeralContainers feature.
+  // +optional
+  EphemeralContainers []EphemeralContainer `json:"ephemeralContainers,omitempty" protobuf:"bytes,29,opt,name=ephemeralContainers"`
```
krmayankk: how is this different from regular container ?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,243 +99,180 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
+type EphemeralContainer struct {
+  // Spec describes the Ephemeral Container to be created.
+  Spec Container `json:"spec,omitempty" protobuf:"bytes,1,opt,name=spec"`
+
+  // If set, the name of the container from PodSpec that this ephemeral container targets.
+  // The ephemeral container will be run in the namespaces (IPC, PID, etc) of this container.
+  // If not set then the ephemeral container is run in whatever namespaces are shared
+  // for the pod.
+  // +optional
+  TargetContainerName string `json:"targetContainerName,omitempty" protobuf:"bytes,2,opt,name=targetContainerName"`
```
krmayankk: all containers of the pods are in same network namespace except possibly pid namespace. Was this field introduced for the pid namespace case only ?

contributors/design-proposals/node/troubleshoot-running-pods.md
```diff
@@ -106,243 +99,180 @@ subsequently be used to reattach and is reported by `kubectl describe`.
...
-scope for Debug Containers, but we will look for a solution as we implement this
-feature.
-
-`EphemeralContainerStatuses` is populated by the kubelet in the same way as
-regular and init container statuses. This is sent to the API server and
-displayed by `kubectl describe pod`.
+Much of the utility of Ephemeral Containers comes from the ability to run a
+container within the PID namespace of another container. `TargetContainerName`
+allows targeting a container that doesn't share its PID namespace with the rest
+of the pod. We must modify the CRI to enable this functionality (see below).
```
krmayankk: So `TargetContainerName` is only useful for pid namespace case only . It would be good to clarify that it has no meaning when all containers are in same network and pid namespace as pod

PR #1273, opened 2017-10-26
Fix node-allocatable doc typo

huangjiuyuan:
Fix node-allocatable doc typo


cblecker:
This isn't a typo. It's a list of kubernetes daemons, and then user pods.

PR #1275, opened 2017-10-26
Remove Disk Allocatable Evictions

dashpole:
After https://github.com/kubernetes/kubernetes/issues/52336, which uncovered a bug in disk accounting, I have struggled to see why ephemeral-storage allocatable evictions are necessary.
For CPU and Memory allocatable, enforcing allocatable through cgroups provides protection from being starved of compute resources by user pods.
However, for ephemeral-storage it seems that the node-level enforcement mechanisms are able to prevent the node from running out of disk, and thus prevent impact to system daemons.
Additionally, memory usage metrics read from cgroups or statfs are far more reliable than an aggregation of `du` calls which may be collected at different times (as https://github.com/kubernetes/kubernetes/issues/52336 demonstrated).
(As a side note, we should read from the /kubepods cgroup to determine allocatable memory usage)
We should still keep the scheduling aspects of node allocatable for ephemeral storage, as this allows for proper accounting.  But I would like to remove enforcement of node allocatable for ephemeral storage.

cc @dchen1107 @derekwaynecarr @jingxu97 @vishh @saad-ali 

### DIFF
--- a/contributors/design-proposals/node/kubelet-eviction.md
+++ b/contributors/design-proposals/node/kubelet-eviction.md
@@ -440,15 +440,6 @@ from placing new best effort pods on the node since they will be rejected by the
 On the other hand, the `DiskPressure` condition if true should dissuade the scheduler from
 placing **any** new pods on the node since they will be rejected by the `kubelet` in admission.
 
-## Enforcing Node Allocatable
-
-To enforce [Node Allocatable](./node-allocatable.md), Kubelet primarily uses cgroups.
-However `storage` cannot be enforced using cgroups.
-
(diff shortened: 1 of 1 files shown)


NickrenREN:
@dashpole  What can we do to prevent nodes from running out of disk(root partition) if we remove the ephemeral storage eviction logic ?

dashpole:
I am proposing removing the **allocatable** eviction, but not node level eviction.  Currently, we act on two thresholds:
1. sum(pod_disk_usage) < disk_capacity - disk_reserved - disk_hard_eviction
2. node_disk_usage < disk_capacity - disk_hard_eviction

I am proposing removing (1), but keeping (2)

derekwaynecarr:
I am fine with this change.

PR #1276, opened 2017-10-26
Add proposal for configurable pod resolv.conf

bowei:
cc: @thockin @MrHohn

### DIFF
--- a/contributors/design-proposals/network/pod-resolv-conf.md
+++ b/contributors/design-proposals/network/pod-resolv-conf.md
@@ -1,32 +1,57 @@
 # Custom /etc/resolv.conf
+
 * Status: pending
 * Version: alpha
-* Implementation owner: Bowei Du <[bowei@google.com](mailto:bowei@google.com)>, Zihong Zheng <[zihongz@google.com](mailto:zihongz@google.com)>
+* Implementation owner: Bowei Du <[bowei@google.com](mailto:bowei@google.com)>,
+  Zihong Zheng <[zihongz@google.com](mailto:zihongz@google.com)>
 
 # Overview
-The `/etc/resolv.conf` in a pod is managed by Kubelet and its contents are generated based on `pod.dnsPolicy`. For `dnsPolicy: Default`, `resolv.conf` is copied from the node where the pod is running. If the `dnsPolicy` is `ClusterFirst`, the contents of the resolv.conf is the hosts `resolv.conf` augmented with the following options:
 
-*   Search paths to add aliases for domain names in the same namespace and cluster suffix.
-*   `options ndots` to 5 to ensure the search paths are searched for all potential matches.
+The `/etc/resolv.conf` in a pod is managed by Kubelet and its contents are
+generated based on `pod.dnsPolicy`. For `dnsPolicy: Default`, the `search` and
+`nameserver` fields are taken from the `resolve.conf` on the node where the pod
+is running. If the `dnsPolicy` is `ClusterFirst`, the search contents of the
+resolv.conf is the hosts `resolv.conf` augmented with the following options:
 
-The configuration of both search paths and `ndots` results in query amplification of five to ten times for non-cluster internal names. This is due to the fact that each of the search path expansions must be tried before the actual result is found. This order of magnitude increase of query rate imposes a large load on the kube-dns service. At the same time, there are user applications do not need the convenience of the name aliases and do not wish to pay this performance cost.
+*   Search paths to add aliases for domain names in the same namespace and
+    cluster suffix.
+*   `options ndots` to 5 to ensure the search paths are searched for all
+    potential matches.
+
+The configuration of both search paths and `ndots` results in query
+amplification of five to ten times for non-cluster internal names. This is due
+to the fact that each of the search path expansions must be tried before the
+actual result is found. This order of magnitude increase of query rate imposes a
+large load on the kube-dns service. At the same time, there are user
+applications do not need the convenience of the name aliases and do not wish to
+pay this performance cost.
 
 
 ## Existing workarounds
 
-The current work around for this problem is to specify an FQDN for name resolution. Any domain name that ends with a period (e.g. `foo.bar.com.`) will not be search path expanded. However, use of FQDNs is not well-known practice and imposes application-level changes. Cluster operators may not have the luxury of enforcing such a change to applications that run on their infrastructure.
+The current work around for this problem is to specify an FQDN for name
+resolution. Any domain name that ends with a period (e.g. `foo.bar.com.`) will
+not be search path expanded. However, use of FQDNs is not well-known practice
+and imposes application-level changes. Cluster operators may not have the luxury
+of enforcing such a change to applications that run on their infrastructure.
 
-It is also possible for the user to insert a short shell script snippet that rewrites `resolv.conf` on container start-up. This has the same problems as the previous approach and is also awkward for the user. This also forces the container to have additional executable code such as a shell or scripting engine which increases the applications security surface area.
+It is also possible for the user to insert a short shell script snippet that
+rewrites `resolv.conf` on container start-up. This has the same problems as the
+previous approach and is also awkward for the user. This also forces the
+container to have additional executable code such as a shell or scripting engine
+which increases the applications security surface area.
 
 
 # Proposal sketch
 
-Add a new `pod.dnsPolicy: Custom` that allows for user customization of `resolv.conf`.
+Add a new `pod.dnsPolicy: Custom` that allows for user customization of
+`resolv.conf`.
 
 
 ## Pod API example
 
-In the example below, the user wishes to add the pod namespace and a custom expansion to the search path, as they do not use the other name aliases:
+In the example below, the user wishes to add the pod namespace and a custom
+expansion to the search path, as they do not use the other name aliases:
 
 ```yaml
 # Pod spec
@@ -41,30 +66,37 @@ spec:
     image: example
   dnsPolicy: Custom
   dnsParams:
-    searchPaths:
-    - $NAMESPACE.svc.$CLUSTER
-    - my.dns.search.suffix
-    - $HOST
-    options:
-    - ndots: 2
+    custom:
+      search:
+      - $(NAMESPACE).svc.$(CLUSTER)
+      - my.dns.search.suffix
+      - $(HOST)
+      options:
+      - "ndots:2"
 ```
 
 Given the following host `/etc/resolv.conf`:
 
 ```bash
 nameserver 1.2.3.4
-search foo.com
-options: ndots: 1
+search foo.com bar.com
+options ndots:1
 ```
 
 The pod will get the following `/etc/resolv.conf`:
 
 ```bash
 nameserver 10.240.0.10
-# Populated from searchPaths
-search ns1.svc.cluster.local my.dns.search.suffix foo.com
-# Populated from options
-options ndots: 2
+# Comments only for explication purposes.
+#
+#      $(NAMESPACE).svc.$CLUSTER
+#      |                     my.dns.search.suffix
+#      |                     |                    $(HOST)
+#      |                     |                    |
+#      V                     V                    V
+search ns1.svc.cluster.local my.dns.search.suffix foo.com bar.com
+# Populated from custom.options
+options ndots:2
 ```
 
 ## More examples
@@ -73,19 +105,22 @@ The following is a Pod dnsParams that only contains the host search paths:
 
 ```yaml
 dnsParams:
-  searchPaths:
-  - $HOST
+  custom:
+    searchPaths:
+    - $HOST
 ```
 
-Override `ndots` and add custom search path. Note that overriding the ndot may break the functionality of some of the search paths the
+Override `ndots` and add custom search path. Note that overriding the ndot may
+break the functionality of some of the search paths the
 
 ```yaml
 dnsParams:
-  searchPaths:
-  - my.custom.suffix
-  - $HOST
-  options:
-  - ndots: 3
+  custom:
+    searchPaths:
+    - my.custom.suffix
+    - $HOST
+    options:
+    - "ndots:3"
 ```
 
 # API changes
@@ -95,25 +130,33 @@ type PodSpec struct {
 	...
 	DNSPolicy string
 	DNSParams *PodDNSParams
+    ...
 }
 
 type PodDNSParams struct {
-	SearchPaths: []string
-	Options: []string
+ 	Custom PodDNSParamsCustom
+}
+
+type PodDNSParamsCustom struct {
+	Search []string
+ 	Options []string
 }
 
 // This will not appear in types.go but is here for explication purposes.
 type DNSParamsSubstitution string
 const (
-	DNSParamsSearchPathNamespace = "$NAMESPACE"
-	DNSParamsSearchPathClusterDomain = "$CLUSTER"
-	DNSParamsSearchPathHostPaths = "$HOST"
+	DNSParamsSearchPathNamespace = "$(NAMESPACE)"
+	DNSParamsSearchPathClusterDomain = "$(CLUSTER)"
+	DNSParamsSearchPathHostPaths = "$(HOST)"
 )
 ```
 
 ## Semantics
 ### searchPath
-If `dnsPolicy: Custom` is used, then the `search` line will be constructed from the entries listed in `dnsParams.searchPath`:
+
+If `dnsPolicy: Custom` is used, then the `search` line will be constructed from
+the entries listed in `dnsParams.custom.search`:
+
 ```go
 // Note: pseudocode does not include input validation.
 func SearchPath(params *PodDNSParams) []string {
@@ -137,18 +180,32 @@ func expand(s string) string {
 	}
 	return strings.Join(labels, ".")
 }
+
+var LabelSubstitutions map[string]string
+
+func Substitute(l label) string {
+  if s, ok := LabelSubstitutions[l]; ok {
+    return s
+  }
+  return l
+}
 ```
 
 #### Substitutions
-`Substitute` will replace labels that begin with `$` with values for the given Pod:
+
+`Substitute` will replace DNS labels that begin with `$` with values for the
+given Pod. Note: the DNS label MUST be an EXACT string match. E.g. if the pod
+namespace is `my-ns`, `abc$(NAMESPACE)1234` will NOT be expanded to
+`abcmy-ns1234`, but will be an invalid configuration.
 
 | Substitution | Description |
 | ----   | ---- |
-| `$NAMESPACE` | Namespace of the Pod |
-| `$CLUSTER` | Kubernetes cluster domain (e.g. `cluster.local`) |
-| `$HOST` | Host search paths. This is a special substitution that must appear at the end of the `searchPaths` list |
+| `$(NAMESPACE)` | Namespace of the Pod |
+| `$(CLUSTER)` | Kubernetes cluster domain (e.g. `cluster.local`) |
+| `$(HOST)` | Host search paths. This is a special substitution that MUST ONLY appear at the end of the `searchPaths` list |
 
 ### options
+
 Each element of options will be copied unmodified as an options line.
 
 ### Invalid configurations
--- a/contributors/design-proposals/network/pod-resolv-conf.md
+++ b/contributors/design-proposals/network/pod-resolv-conf.md
@@ -44,11 +44,40 @@ which increases the applications security surface area.
 
 # Proposal sketch
 
-Add a new `pod.dnsPolicy: Custom` that allows for user customization of
-`resolv.conf`.
+This proposal gives users a way to overlay tweaks into the existing
+`DnsPolicy`. A new PodSpec field `dnsParams` will contains fields that are
+merged with the settings currently selected with `DnsPolicy`.
 
+The fields of `DnsParams` are:
 
-## Pod API example
+* `nameservers` is a list of additional nameservers to use for resolution. On
+  `resolv.conf` platforms, these are entries to `nameserver`.
+* `search` is a list of additional search path subdomains. On `resolv.conf`
+  platforms, these are entries to the `search` setting. These domains will be
+  appended to the existing search path.
+* `options` that are an OS-dependent list of options. For containers that use
+  `/etc/resolv.conf` style configuration, these correspond to the parameters
+  passed to the `option` lines. Options will override if their names coincide,
+  i.e, if the `DnsPolicy` sets `ndots:5` and `ndots:1` appears in the `Spec`,
+  then the final value will be `ndots:1`.
+
+For users that want to completely customize their resolution configuration, we
+add a new `DnsPolicy: Custom` that does not define any settings. This is
+essentially an empty `resolv.conf` with no fields defined.
+
+## Pod API examples
+
+### Host `/etc/resolv.conf`
+
+Assume in the examples below that the host has the following `/etc/resolv.conf`:
+
+```bash
+nameserver 10.1.1.10
+search foo.com
+options ndots:1
+```
+
+### Override DNS server and search paths
 
 In the example below, the user wishes to use their own DNS resolver and add the
 pod namespace and a custom expansion to the search path, as they do not use the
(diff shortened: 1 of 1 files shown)


bowei:
@kubernetes/sig-network-api-reviews

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+metadata:
+  namespace: ns1
+  name: example
+spec:
+  containers:
+  - name: example
+    image: example
+  dnsPolicy: Custom
+  dnsParams:
+    searchPaths:
```
thockin: s/searchPaths/search/

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+    - $NAMESPACE.svc.$CLUSTER
```
thockin: We should be very careful with templating syntax.  This could set precedent.

I would suggest $(NAMESPACE) similar to other places we do expansion.  Needs REALLY good docs on what expansions are available and what happens if the user uses an invalid one, and how to get a literal value from the escape.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+    - my.dns.search.suffix
+    - $HOST
+    options:
+    - ndots: 2
```
thockin: unclear what the type behind this is?  Is it a map of string-> string?  I think you meant it to just be []string, but I bet this doesn't unmarshal properly (since you wrote a []struct{ndots int}).

you either need to say:

```
options:
- "ndots: 2"
```

or we need to actually make a schema (which I don't think we should do) or we need to make it a map (which we generally frown upon) or we need to make it:

```
options:
- name: ndots
  value: "2"
```

which I dislike.  Assuming no other major issues, we can discuss this as an API issue with API guardians.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+Given the following host `/etc/resolv.conf`:
+
+```bash
+nameserver 1.2.3.4
+search foo.com
```
thockin: if you make this hold 2 entries it will be clearer what $(HOST) means

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+In the example below, the user wishes to add the pod namespace and a custom expansion to the search path, as they do not use the other name aliases:
+
+```yaml
+# Pod spec
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: ns1
+  name: example
+spec:
```
thockin: no field for nameservers ?  E.g. if a user wants to not use cluster DNS at all?

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+# Populated from options
+options ndots: 2
+```
+
+## More examples
+
+The following is a Pod dnsParams that only contains the host search paths:
+
+```yaml
+dnsParams:
```
thockin: How does this map to Windows?  Do the concepts apply?

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+		if labels[0] == "$" {
+			labels[i] = Substitute(labels[i])
+		}
+	}
+	return strings.Join(labels, ".")
+}
+```
+
+#### Substitutions
+`Substitute` will replace labels that begin with `$` with values for the given Pod:
```
thockin: is $(NAMESPACE)12345 going to => foobar12345 or is it an error?

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+kind: Pod
+metadata:
+  namespace: ns1
+  name: example
+spec:
+  containers:
+  - name: example
+    image: example
+  dnsPolicy: Custom
+  dnsParams:
```
thockin: Maybe

```
dnsParams:
  custom:
    search:
```
so we leave room for other params over time?

contributors/design-proposals/network/pod-resolv-conf.md
matthiasr: Maybe you can get away without templating, at least initially? My guess is that any user who needs such an advanced feature can figure out how to template this themselves, or already have a templating layer anyway (we do).

This would uncouple the DNS configurability from all the complexity of adding templating.

If it turns out to be really indispensable, it can still be added later, but once it's in you can't get it out and you'll have to explain why it's not available everywhere.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
+# Custom /etc/resolv.conf
+* Status: pending
+* Version: alpha
+* Implementation owner: Bowei Du <[bowei@google.com](mailto:bowei@google.com)>, Zihong Zheng <[zihongz@google.com](mailto:zihongz@google.com)>
+
+# Overview
+The `/etc/resolv.conf` in a pod is managed by Kubelet and its contents are generated based on `pod.dnsPolicy`. For `dnsPolicy: Default`, `resolv.conf` is copied from the node where the pod is running. If the `dnsPolicy` is `ClusterFirst`, the contents of the resolv.conf is the hosts `resolv.conf` augmented with the following options:
```
phsiao: The statement "For `dnsPolicy: Default`, `resolv.conf` is copied from the node where the pod is running" is incorrect.  It only copies nameserver and search.   I'd love to see it actually does that but it is not, and I think the statement should be reflecting that.

contributors/design-proposals/network/pod-resolv-conf.md
phsiao: Want to raise that not all options are {key}:{value}, options like rotate and debug are just flags.  I think []string might be the best because the intention here is to capture the options, and no additional processing is assumed.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+    options:
+    - ndots: 2
+```
+
+Given the following host `/etc/resolv.conf`:
+
+```bash
+nameserver 1.2.3.4
+search foo.com
+options: ndots: 1
```
phsiao: This is not a valid resolv.conf syntax, should be options ndots:1.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+The pod will get the following `/etc/resolv.conf`:
+
+```bash
+nameserver 10.240.0.10
+# Populated from searchPaths
+search ns1.svc.cluster.local my.dns.search.suffix foo.com
+# Populated from options
+options ndots: 2
```
phsiao: Don't know if it is correct to have a space between : and the value, (e.g., ndots:" "2), but I have not seen it being used this way.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,167 @@
...
+# API changes
+
+```go
+type PodSpec struct {
+	...
+	DNSPolicy string
+	DNSParams *PodDNSParams
+}
+
+type PodDNSParams struct {
```
phsiao: while we are on this, should we also add domain and sortlist that are also supported in resolv.conf?

contributors/design-proposals/network/pod-resolv-conf.md
phsiao: I don't think dnsParams is the best key name for this.  The customization is around resolv.conf, which is used specifically by some resolver implementations.   I think it should be referencing resolv.conf customization.  This should also address the concern that how is this applicable to Windows.

contributors/design-proposals/network/pod-resolv-conf.md
bowei: I'm wary of adding in nameserver configuration due to differences between common libc (glibc and musl). glibc will try the nameservers in sequence whereas musl sends requests in parallel.

contributors/design-proposals/network/pod-resolv-conf.md
bowei: done

contributors/design-proposals/network/pod-resolv-conf.md
bowei: I added more detail to the psuedocode and exposition. The system works by substituting labels if the substitution value matches and is not intended to be a generic template mechanism.

@matthiasr we can leave out templating but suspect most people would want to be able to write YAMLs portable across namespaces without forcing use of another layer.

contributors/design-proposals/network/pod-resolv-conf.md
bowei: It is []string (see API code), I updated the YAML to make this clear (it's actually a YAML syntax error)

contributors/design-proposals/network/pod-resolv-conf.md
bowei: done

contributors/design-proposals/network/pod-resolv-conf.md
bowei: updated, the subst will be label for label.

contributors/design-proposals/network/pod-resolv-conf.md
bowei: According to the documentation online, Windows Containers do not support search path https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-networking.

Or maybe the support doc has not yet been updated: https://github.com/Microsoft/hcsshim/pull/98.

The user space kube-proxy for Windows actually captures the DNS request and implements the search path expansion there.

contributors/design-proposals/network/pod-resolv-conf.md
bowei: fixed

contributors/design-proposals/network/pod-resolv-conf.md
bowei: fixed

contributors/design-proposals/network/pod-resolv-conf.md
bowei: I am trying to be as parsimonious as possible in the initial allowed options.

bowei: It should be easy to add the other fields in the future...

contributors/design-proposals/network/pod-resolv-conf.md
bowei: I will make the text more clear.

bowei: Fixed in the text.

contributors/design-proposals/network/pod-resolv-conf.md
bowei: @thockin I updated

@phsiao It seems to be asymmetric to have:
```
dnsPolicy: Custom
dnsResolvConf: 
  ...
```
or on Windows
```
dnsPolicy: Custom
dnsWindows: 
  ...
```

The search path expansion seems to be something that all platforms should probably support in some way. As for options, by its nature, I would expect that to be non-portable.

bowei:
Updated, please take a look.

contributors/design-proposals/network/pod-resolv-conf.md
thockin: If a user REALLY wants to customize, they might choose to bypass kube-DNS entirely?

contributors/design-proposals/network/pod-resolv-conf.md
thockin: How does windows configure name resolution?  We need an answer to that before we merge this, really.

contributors/design-proposals/network/pod-resolv-conf.md
thockin: oh puke.

can anyone from @kubernetes/sig-windows-misc comment on this?  I'd like to not make the API any more non-windows-friendly, but I don't know how to express this in a windows-friendly way

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+	Search []string
+ 	Options []string
+}
+
+// This will not appear in types.go but is here for explication purposes.
+type DNSParamsSubstitution string
+const (
+	DNSParamsSearchPathNamespace = "$(NAMESPACE)"
+	DNSParamsSearchPathClusterDomain = "$(CLUSTER)"
+	DNSParamsSearchPathHostPaths = "$(HOST)"
```
thockin: $(HOST) feels wrong.  Can we find a better name?  $(HOST_SEARCHES) or something?

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+  name: example
+spec:
+  containers:
+  - name: example
+    image: example
+  dnsPolicy: Custom
+  dnsParams:
+    custom:
+      search:
+      - $(NAMESPACE).svc.$(CLUSTER)
```
smarterclayton: What's "$(CLUSTER)" here?  New magic downward API?

smarterclayton: I don't think its appropriate to overlap our existing downward API with a new source of values.  Will users be allowed to do downward API injection of env vars?  What happens when they clash?  Do we define three magic ones as part of the API (note in every other spot we require the user to define NAMESPACE as a downward envFrom for pod.namespace)?

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+
+`Substitute` will replace DNS labels that begin with `$` with values for the
+given Pod. Note: the DNS label MUST be an EXACT string match. E.g. if the pod
+namespace is `my-ns`, `abc$(NAMESPACE)1234` will NOT be expanded to
+`abcmy-ns1234`, but will be an invalid configuration.
+
+| Substitution | Description |
+| ----   | ---- |
+| `$(NAMESPACE)` | Namespace of the Pod |
+| `$(CLUSTER)` | Kubernetes cluster domain (e.g. `cluster.local`) |
```
smarterclayton: The moment we add this people will ask for it in other places.  This should be part of the downward API if we want to add it.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+| `$(HOST)` | Host search paths. This is a special substitution that MUST ONLY appear at the end of the `searchPaths` list |
```
smarterclayton: If it can only be at the end why not just have a boolean `includeHostSearchPaths`?

contributors/design-proposals/network/pod-resolv-conf.md
smarterclayton: And it needs a better name, it's not `CLUSTER` it's `CLUSTER_SUBDOMAIN` or similar.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+*   `options ndots` to 5 to ensure the search paths are searched for all
+    potential matches.
+
+The configuration of both search paths and `ndots` results in query
+amplification of five to ten times for non-cluster internal names. This is due
+to the fact that each of the search path expansions must be tried before the
+actual result is found. This order of magnitude increase of query rate imposes a
+large load on the kube-dns service. At the same time, there are user
+applications do not need the convenience of the name aliases and do not wish to
+pay this performance cost.
```
smarterclayton: Add the security implication of our current default search paths as well.  @thockin is this our chance to fix that one?

thockin: I think the schematic rejigger is orthogonal to this, mostly.  I'm pretty sure that to make that work, we'll need a new DNSPolicy that enacts the new search paths, opt-in, and the DNS server will either serve both styles or we will need 2 DNS servers.

We could make the default become admission control or PodPreset, so admins could flip it to new mode for the whole cluster eventually.  The old mode basically can't go away for a long time.

contributors/design-proposals/network/pod-resolv-conf.md
thockin: Making it part of downward API is an interesting idea.  Do we have other places where downward API pulls values from thin air?  ISTR most of them are pulled from pod fields.  Namespace could be metadata.namespace (if it isn't already), but we have no field for cluster suffix.

This proposal was doing a limited-set of substitutions outside of downward API because, as we know, the downward API requires more structure than simple templating.  These were not intended to be generic env-vars that expand, just a few magic names in this particular piece of the API. 
 It's hacky, and I figured it would draw attention, but I hoped we could keep it simple (since there's no risk of collision - we control all the variables, unlike `env`).

We could make this a generic env expansion, which would require users to touch both `env` and `dnsPolicy`.  That feels like overkill.

contributors/design-proposals/network/pod-resolv-conf.md
thockin: The point was to provide a small number of magic values without any other scaffolding.  Users can't define new magic here, nor does this magic overlap with env or anything else.  That may not sway the conversation, but it is context.

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+
+```yaml
+dnsParams:
+  custom:
+    searchPaths:
+    - $HOST
+```
+
+Override `ndots` and add custom search path. Note that overriding the ndot may
+break the functionality of some of the search paths the
```
MrHohn: The trailing "the" seems weird?

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+
+## More examples
+
+The following is a Pod dnsParams that only contains the host search paths:
+
+```yaml
+dnsParams:
+  custom:
+    searchPaths:
+    - $HOST
```
MrHohn: `$HOST` -> `$(HOST)`

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+
+Override `ndots` and add custom search path. Note that overriding the ndot may
+break the functionality of some of the search paths the
+
+```yaml
+dnsParams:
+  custom:
+    searchPaths:
+    - my.custom.suffix
+    - $HOST
```
MrHohn: `$HOST` -> `$(HOST)`

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,224 @@
...
+
+type PodDNSParamsCustom struct {
+	Search []string
+ 	Options []string
+}
+
+// This will not appear in types.go but is here for explication purposes.
+type DNSParamsSubstitution string
+const (
+	DNSParamsSearchPathNamespace = "$(NAMESPACE)"
```
MrHohn:
```go
DNSParamsSearchPathNamespace DNSParamsSubstitution = "$(NAMESPACE)"
```

contributors/design-proposals/network/pod-resolv-conf.md
bowei: Making this default to adding the host search paths to make it easier for users:
`excludeHostSearchPaths`
Thoughts?

contributors/design-proposals/network/pod-resolv-conf.md
bowei: done

contributors/design-proposals/network/pod-resolv-conf.md
bowei: done

contributors/design-proposals/network/pod-resolv-conf.md
bowei: done

contributors/design-proposals/network/pod-resolv-conf.md
bowei: done

contributors/design-proposals/network/pod-resolv-conf.md
bowei: The last time I looked at making this part of the downward API, this seems the first in terms of values being passed that are not part of a declared spec.

bowei:
@thockin regarding the "append" model, this seems to be not so useful as:
- Order matters, which means a lot of times you will be want to prepend (start from Custom)
- Many environment are already very close to exceeding the built-in glibc limits

bowei:
Minor updates pushed.

Looks like ths following need to be resolved:
* Templates-lite vs downward API
* thockin's suggestion for appending vs replacement.
* Windows compatibility? (Someone from sig-windows need to be comment here)

bowei:
There is an alternative to using template, which is to enumerate the current set of k8s constructed search paths explicitly by naming them:

| name | value |
|---- | ---- |
| Namespace | `<namespace>.SVC.<cluster subdomain>` |
| Service | `SVC.<cluster subdomain>` |
| ClusterSubdomain | `<cluster subdomain>` |

then YAML will look something like:
```
dnsParams:
  searchPaths:
  - suffix: Custom
    path: "my.custom.suffix"
  - suffix: Namespace  # adds "ns.svc.cluster.local"
  - suffix: ClusterSubdomain # adds "cluster.local"
  ndots: 3
```

bowei:
I'm also happy to entertain NOT pursuing the templating for the first cut as this has a wider impact and is mostly for ease of use and YAML portability.

contributors/design-proposals/network/pod-resolv-conf.md
bowei: Added `nameservers`.

contributors/design-proposals/network/pod-resolv-conf.md
dineshgovindasamy: DNS Search option is already supported for Windows, the document needs to be updated.

bowei: thanks @dineshgovindasamy

jayunit100:
Just wanted to note how important this can be in enterprises where standard inherited DNS can lead to all sorts of weird behaviours and race conditions .  Very happy to stumble upon this !

thockin:
re: append model, my goal was to make the customizations less of a special case, since I have heard of people wanting to tweak just ndots.  It would be a shame to force them all the way to Custom for that.  I figure, if anyone needs to prepend stuff, they can go Custom.  Most people will want to append search paths or options.

Another wrinkle with nameservers - if we do templating, we probably need a variable for cluster DNS.

contributors/design-proposals/network/pod-resolv-conf.md
smarterclayton: Yeah, I want this to be non linux specific.  I really, really, really don't want to embed resolvconf in our API design.  

I'd prefer for people who need to configure resolve.conf themselves supporting a configmap mounted at /etc/resolv.conf.  If we had single file injection we could solve this problem.

bowei:
@thockin: Regarding exposing templated kube-dns nameserver IP, it's a bad idea to mix kube-dns and other nameservers. This will results in uncertainty between how names are resolved with glibc (try nameservers in sequence) vs alpine (try nameservers in parallel). Even though that COULD be a use case, it will be extremely easy to shoot one's own foot.

@smarterclayton: this proposal is mostly platform agnostics (DNS nameservers, search paths) are support in both *NIX and Windows platforms. As to a custom file-based ConfigMap, yes that would be great to have, but we would then need to resolve injection of the ClusterDNS IP and search path via some kind of downward API.

thockin:
What about the append model with no templating, and we can add that back in if people use it and find it insufficient?  Does that satisfy?

bowei:
Append without templating would work. `Custom` would be basically a completely empty configuration.

bowei:
Updated proposal
* Removes all templating
* Changes fields to overlay existing Policy fields

bowei:
Changes options to be a mergable keylist

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,212 @@
...
+which increases the applications security surface area.
+
+
+# Proposal sketch
+
+This proposal gives users a way to overlay tweaks into the existing
+`DnsPolicy`. A new PodSpec field `dnsParams` will contains fields that are
+merged with the settings currently selected with `DnsPolicy`.
+
+The fields of `DnsParams` are:
```
thockin: We have at least a few examples of foobarConfig, none of foobarParams

contributors/design-proposals/network/pod-resolv-conf.md
```diff
@@ -0,0 +1,212 @@
...
+      value: 2
+    - name: edns0
+```
+
+The pod will get the following `/etc/resolv.conf`:
+
+```bash
+nameserver 1.2.3.4
+search ns1.svc.cluster.local my.dns.search.suffix
+options ndots:2
```
thockin: I think you have to flatten options into one line

bowei:
Yes -- the proposal is currently marked as alpha. Should I add what it looks like via annotations?

contributors/design-proposals/network/pod-resolv-conf.md
bowei: I'm not following this comment -- can you expand?

thockin: Name - "somethingParams" vs "somethingConfig"

thockin:
We're not doing alphas by annotation any more.  Now we use fields and alpha gates to hide those fields.

I'm approving this, but I think we should change the name

contributors/design-proposals/network/pod-resolv-conf.md
bowei: ack -- changing name.

PR #1277, opened 2017-10-26
document component config conventions

mikedanese:
This is an oldish doc that is just now being converted into a proposal.

https://docs.google.com/document/d/1arP4T9Qkp2SovlJZ_y790sBeiWXDO6SG10pZ_UUU-Lc/edit#heading=h.xgjl2srtytjt

### DIFF
--- a/contributors/design-proposals/config-conventions.md
+++ b/contributors/design-proposals/config-conventions.md
@@ -1,0 +1,222 @@
+# Component Configuration Convetions
+
+Author: Mike Danese (@mikedanese)
+
+# Objective
+
+This document concerns the configuration of Kubernetes system components (as
+opposed to the configuration of user workloads running on Kubernetes).
+Component configuration is a major operational burden for operators of
+Kubernetes clusters. To date, much literature has been written on and much
+effort expended to improve component configuration. Despite this, the state of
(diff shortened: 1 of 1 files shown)


bgrant0607:
@mikedanese Thanks. I'll take a look as soon as I can.

If this is current plan of record, I think the /devel directory would be a better home for the doc. That's where our other conventions docs are.

contributors/design-proposals/config-conventions.md
```diff
@@ -0,0 +1,222 @@
...
+ipTablesSyncPeriod: 2
+ipTablesConntrackHashSize: 2
+ipTablesConntrackTableSize: 2
+```
+
+Write:
+
+```yaml
+kind: KubeProxyConfiguration
+apiVersion: kubeproxy/v1beta3
```
bgrant0607: There is discussion on the SIG architecture thread about changing the API group naming convention.

We have componentconfig, kubeletconfig, and probably other variants currently.

My proposal was `<component>.config.k8s.io` (assuming we're going with .k8s.io suffixes).

bgrant0607:
@mikedanese Please move this doc to /devel.

PR #1280, opened 2017-10-26
First draft from the 2017-10-11 SIG meeting.

lavalamp:
I will send another PR making adjustments.

PR #1281, opened 2017-10-26
Webhook proposal up to date

lavalamp:
Changes to the doc since it was agreed on, builds on #1280.

### DIFF
--- a/contributors/design-proposals/api-machinery/admission-control-webhooks.md
+++ b/contributors/design-proposals/api-machinery/admission-control-webhooks.md
@@ -9,9 +9,6 @@ Thanks to: {@dbsmith,  @smarterclayton, @deads2k, @cheftako, @jpbetz, @mbohlool,
 [TOC]
 
 
-**Discussion and Decision for this Beta plan desired at 11 October SIG-API-Machinery meeting, due to short release cycle in 1.9.**
-
-
 # Summary 
 
 This document proposes a detailed plan for bringing Webhooks to Beta. Highlights include (incomplete, see rest of doc for complete list) :
@@ -38,6 +35,8 @@ This plan is compatible with the [original design doc]( https://github.com/kuber
 **Validating Webhook**: synonym for Non-Mutating Webhook
 
 **Static Admission Controller**: Compiled-in Admission Controllers, (in plugin/pkg/admission).
+
+**Webhook Host**: a process / binary hosting a webhook.
 
 # Naming 
 
@@ -71,23 +70,12 @@ Not in scope:
 
 *   Initializers remains Alpha for 1.9.  (See [Comparison of Webhooks and Initializers](#comparison-of-webhooks-and-initializers) section).  No changes to it.  Will revisit its status post-1.9.
 *   Converting static admission controllers is out of scope (but some investigation has been done, see Moving Built-in Admission Controllers section).
-*   Internal Go interface refactor (e.g. along the lines suggested  #[1137](https://github.com/kubernetes/community/pull/1137)).  Nice-to-have but not gating beta.   
 
 
 ## Work Items 
 
-
-
-*   Add API for registering mutating webhooks.  See 
-
-<p id="gdcalert1" ><span style="color: red; font-weight: bold">>>>>  GDC alert: undefined internal link (link text: "API Changes"). Did you generate a TOC? </span><br>(<a href="#">Back to top</a>)(<a href="#gdcalert2">Next alert</a>)<br><span style="color: red; font-weight: bold">>>>> </span></p>
-
-[API Changes](#heading=h.5xu31klaqsa0)
-*   Copy the non-mutating webhook admission controller code and rename it to be for mutating. (Splitting into two registration APIs make ordering clear.) Add changes to handle mutating responses.  See 
-
-<p id="gdcalert2" ><span style="color: red; font-weight: bold">>>>>  GDC alert: undefined internal link (link text: "Responses for Mutations"). Did you generate a TOC? </span><br>(<a href="#">Back to top</a>)(<a href="#gdcalert3">Next alert</a>)<br><span style="color: red; font-weight: bold">>>>> </span></p>
-
-[Responses for Mutations](#heading=h.a9rxps4ka8y3).
+*   Add API for registering mutating webhooks.  See [API Changes](#api-changes)
+*   Copy the non-mutating webhook admission controller code and rename it to be for mutating. (Splitting into two registration APIs make ordering clear.) Add changes to handle mutating responses.  See [Responses for Mutations](#responses-for-mutations).
 *   Document recommended flag order for admission plugins.  See [Order of Admission](#order-of-admission).
 *   In kube-up.sh and other installers, change flag per previous item. 
 *   Ensure able to monitor latency and rejection from webhooks. See [Monitorability](#monitorability).
@@ -96,6 +84,7 @@ Not in scope:
 *   Good Error Messages	.  See [Good Error Messages](#good-error-messages)	
 *   Conversion logic in GenericWebhook to send converted resource to webhook.  See [Conversion](#conversion) and [#49733](https://github.com/kubernetes/kubernetes/issues/49733).
 *   Schedule discussion around resiliency to down webhooks and bootstrapping
+*   Internal Go interface refactor (e.g. along the lines suggested  #[1137](https://github.com/kubernetes/community/pull/1137)).
 
 
 # Design Discussion 
@@ -112,13 +101,13 @@ We will do webhooks beta before initializers beta because:
 1.  **API Consistency**: Prefer completing one related pair of interfaces (both kinds of webhooks) at the same time.
 
 
-## Why Support Mutation for Beta
+## Why Support Mutation for Beta 
 
 Based on experience and feedback from the alpha phase of both Webhooks and Initializers, we believe Webhooks Beta should support mutation because:  
 
 
 
-1.   we have lots of use cases to inform this (both from Initializers, and Admission Controllers) to ensure we have needed features
+1.  We have lots of use cases to inform this (both from Initializers, and Admission Controllers) to ensure we have needed features
 1.  We have experience with Webhooks API already to give confidence in the API.  The registration API will be quite similar except in the responses.
 1.  There is a strong community demand for something that satisfies a mutating case.
 
@@ -134,7 +123,9 @@ After the release of 1.9, we will advise users who currently use initializers to
 
 We will continue to support Initializers as an Alpha API in 1.9.
 
-We will point out https://github.com/caesarxuchao/example-webhook-admission-controller and extend it to support mutating webhooks, and as that Initializer docs point to it (e.g. https://github.com/kelseyhightower/kubernetes-initializer-tutorial)
+We will make a user guide and extensively document these webhooks. We will update some existing examples, maybe https://github.com/caesarxuchao/example-webhook-admission-controller (since the initializer docs point to it, e.g. https://github.com/kelseyhightower/kubernetes-initializer-tutorial), or maybe https://github.com/openshift/generic-admission-server.
+
+We will clearly document the reasons for each and how users should decide which to use.
 
 
 ## Monitorability 
@@ -150,12 +141,12 @@ There should be prometheus variables to show:
     *   Overall
     *   By webhook name.
 
-Adding a webhook dynamically adds a key to a map-valued prometheus metric.
+Adding a webhook dynamically adds a key to a map-valued prometheus metric. Webhook host process authors should consider how to make their webhook host monitorable: while eventually we hope to offer a set of best practices around this, for the initial release we won't have requirements here.
 
 
 ## API Changes 
 
-GenericAdmissionWebhook Admission Controller is duplicated.
+GenericAdmissionWebhook Admission Controller is split and renamed.
 
 
 
@@ -163,7 +154,7 @@ GenericAdmissionWebhook Admission Controller is duplicated.
 *   The other is called `ValidatingAdmissionWebhook`
 *   Splitting them allows them to appear in different places in the `--admission-control` flag's order.
 
-ExternalAdmissionHookConfiguration API is duplicated.
+ExternalAdmissionHookConfiguration API is split and renamed.
 
 
 
@@ -186,16 +177,14 @@ InitializerConfiguration and ExternalAdmissionHookConfiguration.
 
 InitializerConfiguration will not join `admissionregistration.k8s.io/v1beta1` at this time.
 
-Any webhooks that register with v1alpha1 may or may not be surprised when they start getting Versioned data.  But we don't make any promises for Alpha, and detecting which API was used to register is too complex.
+Any webhooks that register with v1alpha1 may or may not be surprised when they start getting versioned data.  But we don't make any promises for Alpha, and this is a very important bug to fix.
 
 
 ## Order of Admission 
 
-The recommended order on kubernetes.io for `--admission-control` will be: `MutatingAdmissionWebhook,NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,ValidatingAdmissionWebhook,ResourceQuota`
+At kubernetes.io, we will document the ordering requirements or just recommend a particular order for `--admission-control`. A starting point might be `MutatingAdmissionWebhook,NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,ValidatingAdmissionWebhook,ResourceQuota`.
 
- This order has the following properties:
-
-
+ There might be other ordering dependencies that we will document clearly, but some important properties of a valid ordering:
 
 *   ResourceQuota comes last, so that if prior ones reject a request, it won't increment quota.
 *   All other Static ones are in the order recommended by [the docs](https://kubernetes.io/docs/admin/admission-controllers/#is-there-a-recommended-set-of-plug-ins-to-use).  (which variously do mutation and validation) Preserves the behavior when there are no webhooks.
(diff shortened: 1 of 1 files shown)


sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+
+Intent is Beta of Webhooks (**both** kinds) in 1.9.  
+
+Not in scope:
+
+
+
+*   Initializers remains Alpha for 1.9.  (See [Comparison of Webhooks and Initializers](#comparison-of-webhooks-and-initializers) section).  No changes to it.  Will revisit its status post-1.9.
+*   Converting static admission controllers is out of scope (but some investigation has been done, see Moving Built-in Admission Controllers section).
+*   Internal Go interface refactor (e.g. along the lines suggested  #[1137](https://github.com/kubernetes/community/pull/1137)).  Nice-to-have but not gating beta.
```
deads2k: Along with your other edits promote this to in-scope, we agreed to the mutating/validating split inside the golang code with at least PSP split as a proof of principle.  We're actively working it here https://github.com/kubernetes/kubernetes/pull/54484 and here https://github.com/kubernetes/kubernetes/pull/54689, so I don't expect delivery to be a problem.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+After the release of 1.9, we will advise users who currently use initializers to:
+
+
+
+*   Move to Webhooks if their use case fits that model well.
+*   Provide SIG-API-Machinery with feedback if Initializers is a better fit. 
+
+We will continue to support Initializers as an Alpha API in 1.9.
+
+We will point out https://github.com/caesarxuchao/example-webhook-admission-controller and extend it to support mutating webhooks, and as that Initializer docs point to it (e.g. https://github.com/kelseyhightower/kubernetes-initializer-tutorial)
```
deads2k: I think we should be less specific about exactly which example we'll doc.  I think https://github.com/openshift/generic-admission-server and https://github.com/openshift/kubernetes-namespace-reservation are probably more complete, but I don't think we need to argue about it here.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+
+We will clearly document the reasons for each and how users should decide which to use.
```
deads2k: @smarterclayton can you provide some detail here for a doc?

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -131,20 +124,20 @@ There should be prometheus variables to show:
     *   Overall
     *   By webhook name.
 
-Adding a webhook dynamically adds a key to a map-valued prometheus metric.
+Adding a webhook dynamically adds a key to a map-valued prometheus metric. Webhook host process authors should consider how to make their webhook host monitorable: while eventually we hope to offer a set of best practices around this, for the initial release we won't have requirements here.
```
deads2k: I agree its not a requirement for beta.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+
+The Group/Version called 
+
+`admissionregistration.k8s.io/v1alpha1` with kinds
+
+InitializerConfiguration and ExternalAdmissionHookConfiguration.
+
+InitializerConfiguration will not join `admissionregistration.k8s.io/v1beta1` at this time.
+
+Any webhooks that register with v1alpha1 may or may not be surprised when they start getting Versioned data.  But we don't make any promises for Alpha, and detecting which API was used to register is too complex.
```
deads2k: I think this should be stated more strongly: Sending internal data was a bug that must be corrected for transition to beta.  We acknowledge that it will break current alpha consumers, but the bug must be corrected for proper consumption.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+
+
+## Order of Admission 
+
+The recommended order on kubernetes.io for `--admission-control` will be: `MutatingAdmissionWebhook,NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,ValidatingAdmissionWebhook,ResourceQuota`
```
deads2k: I'd like to avoid pinning ourselves to what this will be in this doc.  I agree that we need to specify a recommended order.  I don't think we need to gate this doc on exactly what that order is.  For instance, I'm fairly certain that ServiceAccount should come before MutatingAdmissionWebhook, but I don't want to argue the niggles for this doc.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+
+ This order has the following properties:
```
deads2k: Keep this section, but re-frame this sentence as, "these should be considerations for our order"

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+
+
+
+*   ResourceQuota comes last, so that if prior ones reject a request, it won't increment quota.
+*   All other Static ones are in the order recommended by [the docs](https://kubernetes.io/docs/admin/admission-controllers/#is-there-a-recommended-set-of-plug-ins-to-use).  (which variously do mutation and validation) Preserves the behavior when there are no webhooks.
```
deads2k: I think some information needs to flow the static mutators to the dynamic mutators.  Things like the persistent volume labels provide information that is otherwise inaccessible.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+*   Ensures dynamic mutations happen before all validations.
+*   Ensures dynamic validations happen after all mutations.
+*   Users don't need to reason about the static ones, just the ones they add.
+
+Validation continues to happen after all the admission controllers (e.g. after mutating webhooks, static admission controllers, and non-mutating admission controllers.)
+
+**TODO**: we should move ResourceQuota after Validation, e.g. as described in #1137.  However, this is a longstanding bug and likely a larger change than can be done in 1.9, and should be revisited along with a larger quota redesign.  Therefore, it is not a requirement for Beta,
```
deads2k: I think we may actually pick this up as as bonus for the split.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+
+Reasons not to allow parallel are complexity of reasoning about concurrent patches, and CRD not supporting PATCH.
+
+`ValidatingAdmissionWebhook `is already parallel,  and there are no responses to merge.  Therefore, it stays parallel.
+
+`MutatingAdmissionWebhook `will run in serial, to ensure conflicts are resolved deterministically.
+
+The order is the sort order of all the WebhookConfigs, by name, and by index within the Webhooks list.  
+
+We don't plan to make mutating webhooks parallel at this time, but we will revisit the question in the future and decide before going to GA.If a strong need arises for parallel mutating webhooks, we will carefully consider adding a dependency graph.the API will allow adding a partial order later as follows:  Add a "parallelStage" field to WebhookConfig object.  Any objects for which this field is unset all run in serial first.  Then ones which have the phase set are run.  All the ones with the same value run in parallel, and different ones in the order based on the numeric (or lexicographical) order by values).
```
deads2k: Please leave this at, "We don't plan to make mutating webhooks parallel at this time, but we will revisit the question in the future and decide before going to GA".  I don't want to to get caught up trying to review a future design for something against a set of future problems.

sig-api-machinery/design-docs/admission-control-webhooks.md
```diff
@@ -0,0 +1,934 @@
...
+## Good Error Messages	 
+
+When a webhook is persistently failing to allow e.g. pods to be created, then the error message from the apiserver must show which webhook failed.
+
+When a core controller, e.g. ReplicaSet, fails to make a resources, it must send a helpful event that is visible in `kubectl describe` for the controlling resources, saying the reason create failed.
+
+
+## Conversion 
+
+The apiserver will send versioned objects to the webhook to avoid churn in the webhook code. The current registration API has an Rule.[APIVersions](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/admissionregistration/types.go#L86) field, which is used to match against the version in the endpoint url, and only matching requests will be forwarded to the webhook. The current API isn't able to specify what version the APIServer should convert the request body to before forwarding it to the webhook. So we propose two ways to modify the API.
```
deads2k: You modified this section heavily and I agree that it could use some work.  Rather than try to force it through here, we can simply say in this pull that, "The apiserver must send versioned objects to the webhook.  The design of exactly how will be a followup."

lavalamp: Yeah and I have a bunch more to say on this. I split it into two sections as there's actually two things to discuss.

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: Yeah the thing you quoted is a suggestion I left; everything else is what the suggestion was replacing. My doc -> md converter just pasted both there.

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: reworded a bit

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: I can likewise imagine writing a dynamic webhook that modifies a field used as input by a static mutator. There may not be a single correct order that works for everyone.

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: merged with above change

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: made it an example instead of gospel

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: added 30% more strength

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: I am working on a user guide outline.

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: added weasel words

sig-api-machinery/design-docs/admission-control-webhooks.md
lavalamp: done

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -211,30 +203,41 @@ When a webhook is persistently failing to allow e.g. pods to be created, then th
...
+*   Reads are ~10x more common that writes in a typical cluster, so any increase in compute cost for writes to multi-version objects is significantly mitigated.
+*   Most objects have < 3 versions; webhooks can be ordered any way at the moment, so they could be called in the manner which minimizes the number of conversions. IOW, conversions should worst-case scale with the number of versions, not the number of webhooks.
 
-It is possible that two different webhooks want to access the same stored-resource via different GVKs.  In this case there could be extra conversions.  Conversion do take some CPU.  However, we don't expect this to be a significant issue for beta because:
+If the administrator chooses to install webhooks that all use the same GVKs for a given internal type, then it will be more efficient, since the apiserver may not need to convert every time: it just needs to patch the json, and then send it back out again without converting to internal.  
 
+## Registering for all possible representations of the same object
```
lavalamp: @erictune As discussed this morning.

lavalamp:
PTAL

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -167,31 +158,33 @@ InitializerConfiguration and ExternalAdmissionHookConfiguration.
...
+Any webhooks that register with v1alpha1 may or may not be surprised when they start getting versioned data.  But we don't make any promises for Alpha, and this is a very important bug to fix.
 
 
 ## Order of Admission 
 
-The recommended order on kubernetes.io for `--admission-control` will be: `MutatingAdmissionWebhook,NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,ValidatingAdmissionWebhook,ResourceQuota`
-
- This order has the following properties:
-
+At kubernetes.io, we will document the ordering requirements or just recommend a particular order for `--admission-control`. A starting point might be `MutatingAdmissionWebhook,NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,ValidatingAdmissionWebhook,ResourceQuota`.
```
deads2k: I'm ok with this wording.  Before we open a pull setting this, you should take a look at how the generic apiserver allows you to specify a fully ordered set of all plugins with specific ones listed as "default to off".  This allows users to do the more common "I want to enforce X" without trying to own the complete list of all admission plugins.  It still allows users to override the list and say "I want exactly these in this order", but that use-case is less common.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -211,30 +203,41 @@ When a webhook is persistently failing to allow e.g. pods to be created, then th
...
+stored-resource via different GVKs.  In this case there could be extra
+conversions.  Conversion does take some CPU.  However, we don't expect this to be
+a significant issue for beta because:
 
-Plan B:
-
-The semantic of the APIVersions field is unchanged. We add a ConvertToGroup and ConvertToVersion field to express the versions the webhooks understand. A straw-man design is [here](https://docs.google.com/document/d/1CPcRQ12dIxIE1_T8BYx41V2w5WEGP0vHAWCa7HpqsVg/edit#heading=h.8emrnnxujf1k).
-
-It's TBD if we go with Plan A or B. 
+*   Pods are by far the most common core object to webhook, based on the  [Use Cases](#use-cases-detailed-descriptions), and pods do not have multiple versions.  Most or all other resources we know we plan to Webhook also do not have multiple versions.  So extra conversion is not needed in practice.
```
deads2k: I'd like to scratch this point.

I agree that we don't really care about conversion cost of beta, but I'm not persuaded by this.  I don't think we should be optimizing for pods, but rather making an API that works in general.

lavalamp: I think the point here isn't that it won't work for other cases, just that pods are likely to be a heavy target (I think you agreed at it's probably going to be at least 50% of the use) and they won't need it, therefore in the common case it's not expensive. I will change the last sentence here-- "not needed in practice" is misleading.

lavalamp:
I pushed a change-- PTAL so I can squash and we can merge.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -211,30 +203,41 @@ When a webhook is persistently failing to allow e.g. pods to be created, then th
...
+It is possible that two different webhooks want to access the same
+stored-resource via different GVKs.  In this case there could be extra
+conversions.  Conversion does take some CPU.  However, we don't expect this to be
+a significant issue for beta because:
 
-It's TBD if we go with Plan A or B. 
+*   Pods are by far the most common core object to webhook, based on the  [Use Cases](#use-cases-detailed-descriptions), and pods do not have multiple versions.  Most or all other resources we know we plan to Webhook also do not have multiple versions.  So extra conversions will likely not, in practice, be performed for the majority of webhook calls.
+*   Reads are ~10x more common that writes in a typical cluster, so any increase in compute cost for writes to multi-version objects is significantly mitigated.
+*   Most objects have < 3 versions; webhooks can be ordered any way at the moment, so they could be called in the manner which minimizes the number of conversions. IOW, conversions should worst-case scale with the number of versions, not the number of webhooks.
```
deads2k: We cannot assume the ability to reorder the mutating chain for conversion ease.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -211,30 +203,41 @@ When a webhook is persistently failing to allow e.g. pods to be created, then th
...
+active versions. There's not currently a way to easily tell which of the exposed
+resources map to the same "storage location". We will not try to solve that
+problem at the moment: if the system administrator wishes to hook all
+deployments, they must (e.g.) make sure their hook is registered for both
+deployments.v1beta1.extensions AND deployments.v1.apps.
 
-*   Pods are by far the most common core object to webhook, based on the  [Use Cases](#use-cases-detailed-descriptions), and pods do not have multiple versions.  Most or all other resources we know we plan to Webhook also do not have multiple versions.  So extra conversion is not needed in practice.
-*   Reads are ~10x more common that writes in a typical cluster, so any increase in compute cost for writes to multi-version objects is significantly mitigated.
+This is likely to be error-prone, especially over upgrades. For GA, we may
+consider mechanisms to make this easier. One idea which has been floated in
```
deads2k: I'd clip the "One idea which has been floated...." sentence.  Off the cuff, I'd do it differently and I'd rather not get bogged down on it here.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -256,7 +259,7 @@ We encourage the use of patch to avoid the "old clients dropping new fields" pro
 
 ## Bootstrapping 
 
-We will add a WebhookConfig.Webhooks.exemptNamespace (or  an WebhookConfig.Webhooks.Rule.exemptNamespace?  This can exempt  some namespaces from webhooks (See #51290) for bootstrapping.
+We will add a WebhookConfig.Webhooks.exemptNamespace (or  an WebhookConfig.Webhooks.Rule.exemptNamespace?  This can exempt some namespaces from webhooks (See #51290) for bootstrapping.
```
deads2k: Chao has a pull somewhere (one these tabs) describing his proposal.  Find that one and link it?

contributors/design-proposals/api-machinery/admission-control-webhooks.md
lavalamp: I thought I deleted that, oops

deads2k:
other than my comments, lgtm.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
lavalamp: sigh, ok

contributors/design-proposals/api-machinery/admission-control-webhooks.md
lavalamp: done

deads2k:
lgtm

lavalamp:
squashed

lavalamp:
Adding the tag

lavalamp:
rebase, adding the tag again

Issue #703, opened 2017-06-09
sig name consistency


spiffxp:
There are a number of places where sig names are ever-so-slightly inconsistent.  I'd like us to link back to this issue when we identify those areas.  Update docs, preferably adjust what we can.

For a sig foo, I expect:
- a sig-foo entry in kubernetes/community/sigs.yaml
- kubernetes/community/sig-foo/README.md
- sig-foo google group
- sig-foo-* teams in https://github.com/orgs/kubernetes/teams
- sig-foo-* google groups
- k8s-mirror-sig-foo-* github users
- #sig-foo in kubernetes.slack.com
- sig/foo labels
- sig-foo to appear in [the list of valid test job owners](https://github.com/kubernetes/test-infra/blob/master/jobs/validOwners.json)

Examples of where we've hit inconsistencies:
- https://groups.google.com/forum/#!topic/kubernetes-wg-contribex/Bo0TnJEljoA
- https://github.com/kubernetes/test-infra/pull/2846#discussion_r121193390

EDIT: Is it reasonable to expect the same for working groups as well?

spiffxp:
FYI @kubernetes/sig-contributor-experience-misc

spiffxp:
FYI @xiangpengzhao

xiangpengzhao:
+1000

It would be good to also have our sig meeting channels follow the name consistency.

In the page [SIGs and Working Groups](https://github.com/kubernetes/community/blob/master/sig-list.md), we use `hangouts` or `zoom` to hold meetings. And some meeting channels use number as id and others use name string.

@spiffxp

idvoretskyi:
big +1

castrojo:
So what I was thinking was proposing something to the SIG governance document that reads something like "The entry `dir: sig-api-machinery` in sigs.yaml becomes the machine-consumable naming convention that all your other resources inherit."

Then we make sure the slack channels, etc. align up. Renaming the slack channels is easy, we just renamed #sig-scale without problems. Not sure on how disruptive renaming google groups. I think the annoying bit will be the Zoom channels and calendars, which seems to be a mixture of numbers and shortcuts. 

However I think we can make the argument to all the leads that being able to just guess the channel name based on convention will be very useful for people, for example I would be able to guess that `zoom.us/j/sig-ui` is the right room. Calendars will be annoying but making SIGs follow this convention will lead to improved tooling and consistency.

cblecker:
[...]

PR #1283, opened 2017-10-27
Add sig labels and use aliases in OWNERS files

cblecker:
This PR does the following:
- Creates a sig-*-leads alias for all sigs and working groups
- Replaces the reviewers/approvers in the sig folders and design-proposals folders with that alias
- Adds a label field to the OWNERS files to allow for automatic sig labelling (https://github.com/kubernetes/test-infra/pull/3502)


contributors/design-proposals/cloud-provider/OWNERS
```diff
@@ -0,0 +1,6 @@
+reviewers:
+  - sig-cloud-provider-leads
+approvers:
+  - sig-cloud-provider-leads
+labels:
+  - sig/cloud-provider
```
xiangpengzhao: We don't have this label.

sig-on-premise/OWNERS
```diff
@@ -1,8 +1,6 @@
...
-  - marcoceppi
-  - dghubble
+  - sig-on-premise-leads
 approvers:
-  - zen
-  - marcoceppi
-  - dghubble
+  - sig-on-premise-leads
+labels:
+  - sig/onprem
```
xiangpengzhao: Hope the names can be consistent one day... onprem vs on-premise
ref: https://github.com/kubernetes/community/issues/703

sig-product-management/OWNERS
```diff
@@ -1,8 +1,6 @@
...
-  - idvoretskyi
-  - calebamiles
+  - sig-product-management-leads
 approvers:
-  - apsinha
-  - idvoretskyi
-  - calebamiles
+  - sig-product-management-leads
+labels:
+  - sig/pm
```
xiangpengzhao: pm vs product-management

contributors/design-proposals/cloud-provider/OWNERS
cblecker: Good catch

sig-on-premise/OWNERS
cblecker: I hope they can to. This isn't to change that though -- just capturing the current state of the world.

grodrigues3:
I like the idea in theory, but I'm concerned that this conflates sig-lead with technical lead.  Sig leads are responsible for organizing and driving the sig and if I recall correctly, there was an explicit decision to keep sig leads separate from tech leads.

That said, it seems to make sense for most of the design proposals dirs as it's a direct replacement of names with alias.

cblecker:
@grodrigues3 I figured this was a good starting place. Most of the `contributors/design-proposals/*` directories didn't have an owners file to begin with, as those directories were created by you in #1010. The sig leads would be welcome to add to the OWNERS files and delegate as needed. The other major plus of this would be auto-labelling PRs based on the files they touch.

grodrigues3:
This sounds good to me.  I'll tag contribX and see if we get any additional feedback.  otherwise, i'll LGTM this afternoon.

@kubernetes/sig-contributor-experience-pr-reviews

PR #1288, opened 2017-10-27
move webhooks design doc to design proposals dir

grodrigues3:
I'm not sure why this design doc was in the sig dir rather than the design-proposals directory.

@kubernetes/sig-api-machinery-proposals


cblecker:
cc: @lavalamp

lavalamp:
It's not obvious that design docs shouldn't go under the top-level sig directory.

It's especially not obvious (if you're starting at the root of the repo) that design docs will be under "contributors".

PR #1289, opened 2017-10-27
Add a proposal for addional volume metrics

gnufied:
This proposal adds additional volume metrics in addition to operation metrics that were added in k8s 1.8.

xref - https://github.com/kubernetes/features/issues/496


gnufied:
cc @kubernetes/sig-storage-api-reviews

PR #1291, opened 2017-10-28
Fix spacing in OWNERS_ALIASES file

cblecker:
Needed to add spaces 😞


xiangpengzhao:
This LGTM. But can you explain a bit why space is needed here? (maybe a requirement for parsing?)

cblecker:
That’s exactly it. It doesn’t seem to be parsing correctly without it.

xiangpengzhao:
@cblecker thanks for explanation! :)
/lgtm

PR #1292, opened 2017-10-28
fix miss-leading word in devel/api-conventionsOC

u2takey:
This expression `Update or create` is very miss-leading. As i know, In kubernetes （now 1.8）, `PUT` is used as `replace` other than `upsert`, which means it will not create any resource.

### DIFF
--- a/contributors/devel/api-conventions.md
+++ b/contributors/devel/api-conventions.md
@@ -531,7 +531,7 @@ duration in seconds before the object should be deleted. Individual kinds may
 declare fields which provide a default grace period, and different kinds may
 have differing kind-wide default grace periods. A user provided grace period
 overrides a default grace period, including the zero grace period ("now").
-* PUT /&lt;resourceNamePlural&gt;/&lt;name&gt; - Update or create the resource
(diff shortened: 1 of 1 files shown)


mikedanese:
This is correct as is. By convention, PUT should create the resource if it does not exist.

cblecker:
Addressed by @mikedanese. Closing.

mikedanese:
See here:

https://github.com/kubernetes/kubernetes/blob/298c42bbcd95c1536e0dc5f7a0aed48cec91eaf1/staging/src/k8s.io/apiserver/pkg/registry/rest/rest.go#L250-L253

and

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/update.go#L99-L110

Resources should allow create from PUT unless they have good reason not to.

u2takey:
@mikedanese thanks a lot, it's my misunderstanding. There's is a `AllowCreateOnUpdate` in resource's strategy.go which decide allow create or not.

PR #1301, opened 2017-10-30
Adding naming discussion details to the design document.

cheftako:
Adding the synopsis of a naming thread to the webhook's design document.


lavalamp:
lgtm; will give @smarterclayton a bit to see it before tagging

PR #1303, opened 2017-10-30
Webhook upgrade

lavalamp:
Fill in areas that were lacking.

### DIFF
--- a/contributors/design-proposals/api-machinery/admission-control-webhooks.md
+++ b/contributors/design-proposals/api-machinery/admission-control-webhooks.md
@@ -245,18 +245,19 @@ Webhooks will recieve the admission review subject in the exact version which
 the user sent it to the control plane. This may require the webhook to
 understand multiple versions of those types.
 
-All communication to webhooks will be JSON formatted. We will likely change that
-for GA.
+All communication to webhooks will be JSON formatted, with a request body of
+type admission.k8s.io/v1beta1. For GA, we will likely also allow proto, via a
+TBD mechanism.
 
-We need to make it possible to know whether an apiserver is safe to upgrade.
-`kube-apiserver` will set the X-Kubernetes-Version header when making any calls
-to a webhook. If the webhook is security-critical, it MUST reject any calls with
-a version greater (as compared by semver) than the most recent one it
-understands. Extension apiservers should also set the X-Kubernetes-Version, to
-be the version of Kubernetes libraries they were built with/from.
+We will not take any particular steps to make it possible to know whether an
+apiserver is safe to upgrade, given the webhooks it is running. System
+administrators must understand the stack of webhooks they are running, watch the
+Kubernetes release notes, and look to the webhook authors for guidance about
+whether the webhook supports Kubernetes version N. We may choose to address this
+deficency in future betas.
 
-This is sort of the minimum we could possibly do here; a design for the next
-steps is here: https://docs.google.com/document/d/1BT8mZaT42jVxtC6l14YMXpUq0vZc6V5MPf_jnzDMMcg/edit
+To follow the debate that got us to this position, you can look at this
+potential design for the next steps: https://docs.google.com/document/d/1BT8mZaT42jVxtC6l14YMXpUq0vZc6V5MPf_jnzDMMcg/edit
 
 
 ## Mutations 
@@ -289,8 +290,6 @@ vulnerabilities) and nice-to-have (set labels).
 Security critical webhooks cannot work with Kubernetes types they don't have
 built-in knowledge of, because they can't know if e.g. Kubernetes 1.11 adds a
 backwards-compatible `v1.Pod.EvilField` which will defeat their functionality.
-Therefore these webhooks are required to check the `X-Kubernetes-Version` header
-and return 400 Bad Request if they do not understand the version.
 
 They therefore need to be updated before any apiserver. It is the responsibility
 of the author of such a webhook to release new versions in response to new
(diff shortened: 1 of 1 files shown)


lavalamp:
cc @mbohlool

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -221,24 +224,6 @@ When a webhook is persistently failing to allow e.g. pods to be created, then th
 
 When a core controller, e.g. ReplicaSet, fails to make a resources, it must send a helpful event that is visible in `kubectl describe` for the controlling resources, saying the reason create failed.
 
-
-## Conversion and Versioning
```
lavalamp: note that this is completely rewritten, not moved.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -254,6 +239,24 @@ consider mechanisms to make this easier. We expect to gather user feedback
...
+
+Webhooks will recieve the admission review subject in the exact version which
+the user sent it to the control plane. This may require the webhook to
+understand multiple versions of those types.
+
+All communication to webhooks will be JSON formatted. We will likely change that
+for GA.
+
+We need to make it possible to know whether an apiserver is safe to upgrade.
+`kube-apiserver` will set the X-Kubernetes-Version header when making any calls
```
lavalamp: >The API types are typically the critical element here… e.g. "is there a field introduced in vX.Y I should be concerned with?"
>
>The versioning of the api types for an api server is orthogonal to the kubernetes machinery it was built on
>
>Both the kubernetes apimachinery (for things like objectmeta) and the apiserver (for its own types, whether it's the kube-apiserver or an extension server) versions are relevant.
>
>That said, I'm hesitant to make out-of-band content that's not in the admission review object part of the input to the admission decision. If we want schema versions (I'm not convinced we do), it should probably be part of the admission review spec

@liggitt The kubernetes release is in the binary version. We don't add fields except at release boundaries.

mbohlool:
LGTM. It is the minimum thing we can do to keep a sane upgrade process.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -169,6 +169,9 @@ The `ValidatingAdmissionWebhookConfiguration` stays the same as `ExternalAdmissi
 
 The `MutatingAdmissionWebhookConfiguration` is the same API as `ValidatingAdmissionWebhookConfiguration`.  It is only visible via the v1beta1 version.
 
+We will change from having a Kubernetes service object to just accepting a DNS
```
deads2k: ok

contributors/design-proposals/api-machinery/admission-control-webhooks.md
deads2k: If you want to set this value, seems like you'd just set it on the `AdmissionReviewSpec`.  `ServerVersion=foo`.

However, I don't think this field has value in general.  Since the originating server is often **not**, the `kube-apiserver` and doesn't know what level the `kube-apiserver` is, filling in its own value isn't likely to be the next best option.  If we need a minor version on the apiversions, I would focus on that instead of adding a product version to this API.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -254,6 +239,24 @@ consider mechanisms to make this easier. We expect to gather user feedback
...
+for GA.
+
+We need to make it possible to know whether an apiserver is safe to upgrade.
+`kube-apiserver` will set the X-Kubernetes-Version header when making any calls
+to a webhook. If the webhook is security-critical, it MUST reject any calls with
+a version greater (as compared by semver) than the most recent one it
+understands. Extension apiservers should also set the X-Kubernetes-Version, to
+be the version of Kubernetes libraries they were built with/from.
+
+This is sort of the minimum we could possibly do here; a design for the next
```
liggitt: I'd drop this and bring it in as an expanding PR

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -254,6 +239,24 @@ consider mechanisms to make this easier. We expect to gather user feedback
...
 
 
+## Conversion and Versioning
+
+Webhooks will recieve the admission review subject in the exact version which
+the user sent it to the control plane. This may require the webhook to
+understand multiple versions of those types.
+
+All communication to webhooks will be JSON formatted. We will likely change that
+for GA.
```
liggitt: json + admission.k8s.io/v1beta1 wrapper object for now
will be able to indicate desire for proto or admission.k8s.io/v1beta2, etc in the future

contributors/design-proposals/api-machinery/admission-control-webhooks.md
lavalamp: I feel like some justification for the above bare-bones system is necessary.

lavalamp:
Updated; PTAL

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -278,6 +282,34 @@ cluster can boot from a cold start) is made more difficult by having webhooks,
 which are a dependency of the control plane. This is covered in its [own design
 doc](./admission-webhook-bootstrapping.md).
 
+## Upgrading the control plane
+
+There are two categories of webhooks: security critical (e.g., scan images for
+vulnerabilities) and nice-to-have (set labels).
+
+Security critical webhooks cannot work with Kubernetes types they don't have
```
deads2k: Doesn't have to be here, but at some point we should document/explain the decode->encode->semantic-equals check that can be used inside of a webhook to fail closed on a new field that is actually used (not a pre-check).

contributors/design-proposals/api-machinery/admission-control-webhooks.md
```diff
@@ -254,6 +239,25 @@ consider mechanisms to make this easier. We expect to gather user feedback
 before designing this.
 
 
+## Conversion and Versioning
```
deads2k: I agree with this plan.

contributors/design-proposals/api-machinery/admission-control-webhooks.md
lavalamp: Yeah, not here, because I actually don't want to tell people to do that. I think it adds a lot of confusion for what amounts to only an 80% solution (e.g., doesn't detect annotation changes or semantic meaning changes; neither should happen but I'll be astonished if we live up to that ideal 100%).

lavalamp:
would someone like to give an LGTM, then? :) :)

PR #1265, opened 2017-10-24
Version 2 for Raw Block Devices

erinboyd:

PR #1307, opened 2017-10-31
add readOnly use case to raw block device proposal

erinboyd:
This PR is an addendum to the #1265 to dictate the behavior of the container runtime for read-only devices as specified in the PVCVolumeSource. The precedence already exists for filesystems thus having a consistent behavior for the user. UC9 was added to illustrate the usage.

### DIFF
--- a/contributors/design-proposals/storage/raw-block-pv.md
+++ b/contributors/design-proposals/storage/raw-block-pv.md
@@ -528,6 +528,40 @@ spec:
         name: my-nginx-data 
 	readOnly: false
 ```
+## UC9:
+
+DESCRIPTION:
+* A user wishes to read data from a read-only raw block device, an example might be a database for analytics processing. 
+
+ADMIN:
+* Admin creates a 1 block devices and with the intention consumption of read-only.
+
+USER:
+* User creates pod and specifies 'readOnly' as a parameter in the persistent volume claim to indicate they would
+like to be bound to a PV with this setting enabled.
+
+```
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx-pod-block-001
+spec:
+  containers:
+    - name: nginx-container
+      image: nginx:latest
+      ports:
+      - containerPort: 80
+      volumeDevices:
+        - name: data
+          devicePath: /dev/xvda
+  volumes:
+    - name: data
+      persistentVolumeClaim:
+        claimName: block-pvc001
+        readOnly: true #flag indicating read-only for container runtime
+```
+*** Note: the readOnly field already exists in the PersistentVolumeClaimVolumeSource above and will dictate the values set by the container runtime options ***
+
 
 # Container Runtime considerations
 It is important the values that are passed to the container runtimes are valid and support the current implementation of these various runtimes. Listed below are a table of various runtime and the mapping of their values to what is passed from the kubelet.
--- a/contributors/design-proposals/storage/raw-block-pv.md
+++ b/contributors/design-proposals/storage/raw-block-pv.md
@@ -533,9 +533,6 @@ spec:
 DESCRIPTION:
 * A user wishes to read data from a read-only raw block device, an example might be a database for analytics processing. 
 
-ADMIN:
-* Admin creates a 1 block devices and with the intention consumption of read-only.
-
 USER:
 * User creates pod and specifies 'readOnly' as a parameter in the persistent volume claim to indicate they would
 like to be bound to a PV with this setting enabled.


contributors/design-proposals/storage/raw-block-ro.md
```diff
@@ -0,0 +1,139 @@
...
+
+  
+# Specific use case of storage consumption for read-only:
+  * An admin wishes to expose a block volume for read-only consumption to an application without write permissions. Example
+ applications might be a database with data to be used for analytics processing in which the results of processing are written to another fileystem or application.  
+
+ 
+# Design Overview
+
+ Today in the container runtime, the raw block devices are defaulted to 'rwm' which provides, "read", "write" and "mknod". There exists such use cases as
```
screeley44: the actual value is ```mrw``` which corresponds to "mknod", "read" and "write"

contributors/design-proposals/storage/raw-block-ro.md
```diff
@@ -0,0 +1,139 @@
...
+outlined above in which we want to restrict the user to only access the block device as a reader. This support already exists in the container runtime and should
+be supported in Kube. 
+In addition, in the container spec today we use the 'readOnly' parameter for the same purpose when using a filesystem. The proposal is to add this same parameter to the volumeDevices definition in the API.
```
screeley44: yeah, make mention that in the existing container spec for VolumeMounts object (which is what VolumeDevices was generally based on) it has a ```readOnly bool```

liggitt: Given volume devices are only backed by pvc/pv objects, I sort of expected the readonly aspect to be controlled by the accessmode on the pvc or the readOnly field on the PV. Which field governs if the PV and Pod disagree?

contributors/design-proposals/storage/raw-block-ro.md
```diff
@@ -0,0 +1,139 @@
...
+```
+kind: PersistentVolume
+apiVersion: v1
+metadata:
+  name: local-raw-pv
+spec:
+  volumeMode: Block
+  capacity:
+    storage: 100Gi
+  local:
```
screeley44: for the Admin part in this, they only control the PV in this case and therefore we should either comment that ```If readOnly property is available for the VolumeSource it can be set to true to make the device readOnly``` Or choose a volumeSource that already has the readOnly property.  Does that make sense?

liggitt: We already have issues with PV volume sources that have a readonly field which is ignored in favor of the PV spec readonly field. This is adding a third level. What's the precedence?

mtanino:
@rootfs

contributors/design-proposals/storage/raw-block-ro.md
msau42: @liggitt, I don't see a readOnly field in the PV?

contributors/design-proposals/storage/raw-block-ro.md
```diff
@@ -0,0 +1,139 @@
+# Raw Block Read-Only Consumption in Kubernetes
```
msau42: It might be better to add this to the existing block proposal instead of creating a new one

contributors/design-proposals/storage/raw-block-ro.md
erinboyd: @liggitt today the accessMode is essentially a label used for binding and it's not enforced. Thus, configured correctly, the admin 'should' set the accessMode to RO in this case to indicate a user wanting to use it that is how the PV is configured. However...we don't enforce this today, thus having the user specify the read-only we can then change the behavior of the runtime to honor this. Does this make sense?

liggitt: you're right... it's on the PVCVolumeSource. Given that volumeDevices are only used with pvcVolumeSources, it seems like we already have a spot to indicate this in the API

contributors/design-proposals/storage/raw-block-ro.md
erinboyd: @liggitt this is following the same pattern already set forth for filesystems. When using a filesystem in the container/pod spec we have volumeMounts and here is where we specify the readOnly. We are just extending the same to block so they will look/feel the same to the user.

contributors/design-proposals/storage/raw-block-ro.md
erinboyd: @msau42 it was already merged..though you said to create a new one?

contributors/design-proposals/storage/raw-block-ro.md
liggitt: the readOnly field set on the volumeMount is ignored in favor of the PVCVolumeSource readOnly field today.

mtanino: @liggitt @screeley44 

Can I confirm that current filesystem volume's situation of readOnly parameter?
This is my current understanding of readOnly parameter.

object | effect | who set this?
------------ | ------------- | ------------- 
readOnly in VolumeSource | this parameter is valid for inline volume. If volume is PV, this parameter is ignored. | Admin
readOnly in PesistentVolumeSource | this parameter is ignored even if volume is PV | Admin
readOnly in volumeMounts | In the volume plugins, they check the field and then potentially chmod all the directories/files in the volume based on that | User
readOnly in PersistentVolumeClaimVolumeSource | this parameter is valid for PV | user

When we add readOnly parameter handling for block volume, I think we should add same behaviour to avoid admin/users confusion.

erinboyd: @liggitt there is no RO field in either the PV or the PVC, only the accessMode. the RO field today only exists in the pod spec

contributors/design-proposals/storage/raw-block-ro.md
mtanino: @msau42 @erinboyd I think you can update merged document to branch.

contributors/design-proposals/storage/raw-block-ro.md
msau42: Ok, there's 2 places today:
* plugin specific volume source (including PVC source)
* volumeMount

The readOnly flag in the plugin source overrides the volumeMount

contributors/design-proposals/storage/raw-block-ro.md
msau42: Create a new PR, but add to the existing doc.

contributors/design-proposals/storage/raw-block-ro.md
screeley44: ok, so VolumeDevice should behave the same then right?  Meaning we provide a readOnly flag in the VolumeDevice() same as VolumeMount() and if a plugin specific volume source (including PVC source) has the readOnly flag it should override the VolumeDevice  ... is this correct @msau42 @erinboyd @mtanino @liggitt ?

erinboyd:
@liggitt we decided to simply things and just trigger off the PVCVolumeSource in the pod. Please review UC9 in the design document. @msau42 @jsafrane

contributors/design-proposals/storage/raw-block-pv.md
```diff
@@ -528,6 +528,40 @@ spec:
...
+      volumeDevices:
+        - name: data
+          devicePath: /dev/xvda
+  volumes:
+    - name: data
+      persistentVolumeClaim:
+        claimName: block-pvc001
+        readOnly: true #flag indicating read-only for container runtime
+```
+*** Note: the readOnly field already exists in the Persistent Volume Source above and will dictate the values set by the container runtime options ***
```
mtanino: s/Persistent Volume Source/PersistentVolumeClaimVolumeSource/g

contributors/design-proposals/storage/raw-block-pv.md
```diff
@@ -528,6 +528,40 @@ spec:
         name: my-nginx-data 
 	readOnly: false
 ```
+## UC9:
+
+DESCRIPTION:
+* A user wishes to read data from a read-only raw block device, such as a database for analytics processing. 
+
+ADMIN:
+* Admin creates a 1 block devices and sets container runtime to read-only mode using the --device option.
```
msau42: Why does the admin do it?  Isn't Kubernetes going to pass the read only flag to the runtime?

mtanino: Admin creates a raw block device.

>> and sets container runtime to read-only mode using the --device option.
This operations are unnecessary. Admin doesn't create container, Kubernetes/kubelet will create and set --device option

erinboyd: yes...fixing

erinboyd: @msau42 yes essentially the admin will just create the PV with the accessMode set to RO so the intention it set ahead of time, but you are correct we won't get the runtime flag until we receive the readOnly bit from the PVCVolumeSource

msau42: The PV access mode is unrelated to the the readonly mode passed to the device.

msau42:
Also, a minor nit, I wouldn't appreciate this to rbd.  It could be confused with the volume plugin.

erinboyd:
@msau42 done ;)

mtanino:
@erinboyd Please use same title for this PR. "Update RawBock proposal with read-only use case"

contributors/design-proposals/storage/raw-block-pv.md
```diff
@@ -528,6 +528,41 @@ spec:
         name: my-nginx-data 
 	readOnly: false
 ```
+## UC9:
+
+DESCRIPTION:
+* A user wishes to read data from a read-only raw block device, such as a database for analytics processing. 
+
+ADMIN:
+* Admin creates a 1 block devices and with the intention consumption of read-only. Admin should label the PV
```
msau42: Not sure if we need to suggest access mode RO.  You could have a producer/consumer use case where one pod writes, and another reads.

erinboyd: @msau42 hum...are we suggesting we set different container runtime values for the same pvc/pv?

msau42: I believe the container runtime access modes only dictate the permission of the device in the container, not on the host.  But that needs to be verified.

erinboyd:
@msau42 lgtu?

msau42:
I think we don't need to suggest that access mode = RO because we don't look at access mode at all today.

mtanino:
I also want to remove suggestion of accessModes. It's just adding confusion.

contributors/design-proposals/storage/raw-block-pv.md
erinboyd: @msau42 that is accurate --- but shouldn't we request admin to use best practices to indicate if the volume is statically provisioned (as detailed in thise use case) to say it's RO ?

erinboyd: I'll just remove.. @msau42 @mtanino

mtanino:
lgtm.

PR #1308, opened 2017-10-31
Update links in sig docs

cblecker:
Does a couple of things:
- Only displays the archive/recording URL if there actually is one in sigs.yaml
- If a label exists for the sig/wg, then display a link to the open PRs/Issues for that sig.

### DIFF
--- a/generator/sig_readme.tmpl
+++ b/generator/sig_readme.tmpl
@@ -22,6 +22,9 @@ Meeting recordings can be found [here]({{.MeetingRecordingsURL}}).
 ## Contact
 * [Slack](https://kubernetes.slack.com/messages/{{.Contact.Slack}})
 * [Mailing list]({{.Contact.MailingList}})
+{{- if .Label }}
+* [Open Community Issues/PRs](https://github.com/kubernetes/community/labels/sig%2F{{.Label}})
+{{- end }}
 {{if .Contact.FullGitHubTeams}}
 ## GitHub Teams
 {{range .Contact.GithubTeamNames -}}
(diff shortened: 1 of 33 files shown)


cblecker:
well that's an unintended consequence with the path labelling.. changing the template (which touches all the sig docs) causes all the sigs to be labelled. I guess at least we know it's working!

generator/sig_readme.tmpl
```diff
@@ -7,8 +7,12 @@
 * [{{.Day}}s at {{.UTC}} UTC]({{$.MeetingURL}}) ({{.Frequency}}). [Convert to your timezone](http://www.thetimezoneconverter.com/?t={{.UTC}}&tz=UTC).
 {{- end }}
 
+{{ if .MeetingArchiveURL -}}
```
grodrigues3: are we requiring a MeetingArchiveURL (a link to notes) and is there any way to do data validation?  Most sigs seem to have it already? 

This doesn't block this PR or anything.  I'm just curious

cblecker: I think it's good practice to have it, but I don't think we're imposing it. Yes, all the sigs have it already. If however, for some reason it's missing, this prevents it from displaying a broken link.

generator/sig_readme.tmpl
```diff
@@ -18,6 +22,9 @@ Meeting recordings can be found [here]({{.MeetingRecordingsURL}}).
 ## Contact
 * [Slack](https://kubernetes.slack.com/messages/{{.Contact.Slack}})
 * [Mailing list]({{.Contact.MailingList}})
+{{- if .Label }}
+* [Open Community Issues/PRs](https://github.com/kubernetes/community/labels/sig%2F{{.Label}})
```
xiangpengzhao: sorry for delay. but I think people most possibly want to see a sig's open issues/prs in k/k, not in community. @cblecker WDYT?

cblecker: Wouldn't be bad to add both. I was more thinking about looking at open design proposals for the sig.

xiangpengzhao: +1 for adding both :)

PR #1311, opened 2017-11-01
proposal of coredump detector

CaoShuFeng:
This is a proposal which implements coredump isolation in kubernetes cluster.
It will work as an add-on which could be deployed in kubernetes cluster.

A demonstrate for this design has been implemented and put [here](https://github.com/CaoShuFeng/coredump-detector)

With this add-on deployed into kubernetes, when coredump happens in pods, users may get coredump info with `kubectl` command:
coredump-metadata

and check the quota:
coredump-quota

Partial-fix: https://github.com/kubernetes/kubernetes/issues/48787


contributors/design-proposals/add-on/coredump-daemonset.md
```diff
@@ -0,0 +1,115 @@
...
+}
+
+type QuotaStatus struct {
+        Used *resource.Quantity `json:"used"`
+        Hard *resource.Quantity `json:"hard"`
+}
+```
+
+# coredump-controller
+Now CRD in kubernetes doesn't support quota, so we deploy a controller to work as
```
derekwaynecarr: the plan is to support this soon once we can have a shared cache with garbage collection.

so in your case, we would just quota: `count/coredumps.coredump`

CaoShuFeng: Hi, I will update this according to the new feature.

Does quota for crd support resource.Quantity?
For example: we allow 2Gi coredump files in namespace A.
But not: we allow 200 coredump files in namespace A.

CaoShuFeng:
ping @vishh for suggestions.

contributors/design-proposals/add-on/coredump-daemonset.md
```diff
@@ -0,0 +1,115 @@
...
+   * save coredump metadata into api-server
+   * save coredump file into persistent volume
+2. A coredump-controller which support quota of coredump files in each namesapce
+![architecture](./architecture.png)
+
+# core_pattern
+To determine whether a core file is generated for process in a k8s container, we
+override /proc/sys/kernel/core_pattern kernel parameter in kubelet node.
+```
+|/coredump/coredump-detector -P=%P -p=%p -e=%e -t=%t -c=/coredump/config --log_dir=/coredump/
```
lucab: Which mount namespace would this `/coredump` path belongs to? In particular, are you assuming that this is going to be in the pod-app mount namespace, in the kubelet mount namespace or in the kernel-init/host one?

CaoShuFeng: kernel does not support namespace for coredump.
So this would be kernel-init/host namespace.

CaoShuFeng:
@luxas Do you have some suggestions about this?

contributors/design-proposals/add-on/coredump-daemonset.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+When coredump happens, linux kernel will call coredump-detector and give core
+dump file as standard input to coredump-detector.
+coredump-detector will:
+* access the docker api and distinguish where(which container) the core dump comes from
```
vikaschoudhary16: s/docker/runtime

CaoShuFeng: Good suggestion.
Now kubelet is supporting container runtimes, so I need to update this and support different container runtimes too.

contributors/design-proposals/add-on/coredump-daemonset.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Community Work
+
+- make core_pattern support namespace in kernel (not approved yet): https://lkml.org/lkml/2017/2/7/931
+
+## Proposed Design
+
+To implement this, we need to make the following components run in kubernetes cluster:
+1. A coredump-detector daemonset runs in each node. It will:
+   * put a new special program integrated with kuberentes’s API into /proc/sys/kernel/core_pattern
```
cgwalters, line 30: For OSes using systemd, I don't see the value in replacing the systemd default.  Instead, coredumps can be gathered from it.

PR #1315, opened 2017-11-01
Proposal for meta-generator

pwittrock:


contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,269 @@
...
+
+Looking for APIs outside the default location can be configured using
+the `--apis-dir` flag.  This flag may be provided multiple times to search multiple
+directories.
+
+```sh
+kubegen --apis-dir notpkg/apis --apis-dir pkg/notapis
+```
+
+- run all code generators against discovered APIs
```
sttts: discovered api mean `types.go` exists?

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,269 @@
...
+  - only run against internal types, do not run against external types
+- prepend license to generated files
+
+#### Specifying both internal API groups and external API versions
+
+To generate code for both manually specified internal and external types
+provide both as positional arguments.
+
+```sh
+kubegen apps apps/v1 apps/v1beta1 extensions extensions/v1beta1
```
sttts: do you have to pass internal and external groups? Will conversions be created if you only pass the external groups? What about clientsets and informers?

contributors/design-proposals/code-generator/kubegen.md
pwittrock: No, I don't think it should since folks may want to break up a monolithic types.go file.  I was thinking purely based off directory structure, and then the presence of +genclient

contributors/design-proposals/code-generator/kubegen.md
pwittrock: groups are always internal, external requires a version.

pwittrock:
PTAL

pwittrock:
@caesarxuchao @mbohlool 

Mind taking a look at this?

mikedanese:
@BenTheElder @ixdy please take a look at bazel pieces.

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,333 @@
...
+
+- "Apache 2.0"
+- "None"
+
+#### Define API groups and versions
+
+The set of APIs code generation is performed for can be explicitly defined instead of
+implicitly determined using the `--api-version` and `--api-group` flags.  When these
+flags are provided, API groups will not be implicitly discovered.  Only API versions
+for groups specified with `--api-group` will be discovered.
```
caesarxuchao: What's the difference between `--api-group` and `--api-version`?

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,333 @@
...
+#### Overriding the searched the api directories
+
+The `--apis-dir` defaults to `pkg/apis` and looks for API groups in that directory.
+
+Looking for APIs outside the default location can be configured using
+the `--apis-dir` flag.  This flag may be provided multiple times to search multiple
+directories.
+
+```sh
+kubegen versioned --apis-dir notpkg/apis --apis-dir pkg/notapis
```
caesarxuchao: If multiple `--apis-dir` and multiple `--api-version` are specified, will the tool try all combinations?

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,333 @@
...
+## Kubernetes API and directory structure conventions
+
+kubegen will run the full set of Kubernetes code generators for all APIs it finds in the project.
+kubegen looks for API groups directories defined under `pkg/apis/<group>/<version>` where *group* and *version*
+match the patterns following Kubernetes API conventions.
+
+- **group-pattern**: ^[a-z0-9\.]+$
+- **version-pattern**: ^v\\d+(alpha\\d+|beta\\d+)?$
+
+By default, kubegen will run code generators for both versioned types defined under `pkg/apis/<group>/<version>` and
```
caesarxuchao: Does this imply that we will move `pkg/api` to `pkg/apis/core`?

One twist is that the external types are actually under `vendor/k8s.io/api/<group>/version>`.

liggitt: > One twist is that the external types are actually under vendor/k8s.io/api/<group>/version>.

also, they're not under `<group>/<version>`, they're under `<something-vaguely-group-related>/<version>`. For example, the types for `storage.k8s.io/v1` are in the path `storage/v1`

sttts: > Does this imply that we will move pkg/api to pkg/apis/core?

https://github.com/kubernetes/kubernetes/pull/53651

contributors/design-proposals/code-generator/kubegen.md
sttts: > also, they're not under <group>/<version>, they're under <something-vaguely-group-related>/<version>. For example, the types for storage.k8s.io/v1 are in the path storage/v1

The assumption is that <something-vaguely-group-related> is equal for the internal and versioned types. Other than that it must be unique in various generators, i.e. we cannot create a clientset for different groups in a directory `foo` in different apis-dirs.

IMO, we don't need this flexibility of multiple  `--apis-dir`s. Instead, I would prefer to keep it simple and have something more explicit and with less magic like `--internal-apis-dir` and `--versioned-apis-dir`.

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### Define API groups and versions
+
+The set of APIs code generation is performed for can be explicitly defined instead of
+implicitly determined using the `--api-version` and `--api-group` flags.  When these
+flags are provided, API groups will not be implicitly discovered.  Only API versions
+for groups specified with `--api-group` will be discovered.
+
+```sh
+kubegen versioned --api-version apps/v1 --api-version apps/v1beta1 --api-version extensions/v1beta1
```
sttts: Why not `kubegen versioned apps/v1 apps/v1beta1 extensions/v1beta1` ?

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,333 @@
...
+kubegen(
+    name = "kubegen",
+    cmd = "versioned",
+)
+```
+
+### Running the Bazel target
+
+```sh
+bazel run //:kubegen
```
mikedanese: Are we going to check in the generated source? Are we going to regenerate automatically based on dependency changes?

sttts: As long as we want to be go-get'able, we have to check in the code. kubegen will not change that.

mikedanese: Neither supporting go get nor checking in code is discussed in this doc so the question is still open.

ericchiang: It seems like kubegen should support both use cases, though I'd vote for checking in code since tools like glide and dep don't support generation.

If it's a goal to support not checking in the generated assetes, examples would be appreciated. cluster-registry tried this with the existing generators, and found it wasn't obvious how to structure the BUILD files so external projects could use them.

https://github.com/kubernetes/cluster-registry/issues/24#issuecomment-339485125

contributors/design-proposals/code-generator/kubegen.md
pwittrock: I thought about that.  I don't really like how it interacts with subcommands.  e.g. if we want to introduce specific subcommands for versioned, this will get really ugly.

contributors/design-proposals/code-generator/kubegen.md
pwittrock: Eventually, but not for the MVP.

contributors/design-proposals/code-generator/kubegen.md
pwittrock: I think having 1 flag is a simpler interface, than separate flags for internal and versioned api directories.

contributors/design-proposals/code-generator/kubegen.md
pwittrock: > For example, the types for storage.k8s.io/v1 are in the path storage/v1

Are there any other patterns besides (group) and (group - domain)?

contributors/design-proposals/code-generator/kubegen.md
pwittrock: Updated

pwittrock:
PTAL.

contributors/design-proposals/code-generator/kubegen.md
pwittrock: @mikedanese The Bazel build target is at the root directory and is mean to be invoked with run.  Why does it need to be prescriptive with whether consumers check the output into source control?

thockin:
I have not had time to keep up with work here.  Iknow @sttts is busy trying to make life better.

Currently, we take great pains (Makefile.generated_files) to only run the generators we need when we need to run them.

Is the plan to move all of this into a single binary?  How does it operate?  Is there a high-level plan I can review?

contributors/design-proposals/code-generator/kubegen.md
sttts: Calling different things with the same name doesn't make it simpler IMO, only leads to magic behind the scenes that requires more knowledge from the user to understand than with two flags.

contributors/design-proposals/code-generator/kubegen.md
sttts: side note: kubernetes/kubernetes#53651 merged. 

> Are there any other patterns besides (group) and (group - domain)?

We have group names and package names. They are distinct without any pattern we can assume. `<package>/<version>` is identifiers each GroupVersion uniquely. I would not switch to the logical `<GroupName>/<GroupVersion>` for specifying them on the command line.

sttts:
> Is the plan to move all of this into a single binary? How does it operate? Is there a high-level plan I can review?

It addresses the use-case of CRD and UAS developers (in short: here is pkg/apis with types, do all generation!). It does not solve the Kubernetes use-case with 35 API groups and 20 times as many types and therefore the need to lazily and automatically generate code. I think it's worth to think through the later as well, somewhere along the lines of kubegen generating dependencies, e.g. as a Makefile. If we decide to keep that use-case distinct, that would be fine too, but should be based on clear arguments.

In other words, there is no high-level plan, but maybe stepping one or two steps back from command line flags to use-cases and requirements will be helpful.

pwittrock:
@thockin No, the goal is not to move the existing generators to a single binary.  Existing Makefiles and build systems will continue to function without change.

> User story: As an author of Kubernetes extensions, I want a simple way to run the code generators against my types.go files.

The goal is to provided a simplified interface for running code generators so that folks building extensions don't need to spend hours reading about each of our code generators and writing and maintaining [dozens of lines of bespoke Makefile](https://github.com/kubernetes-incubator/service-catalog/blob/master/Makefile#L128) logic.

thockin:
I see.  Devil's advocate:  If we linked all our generators into a single binary and made it single-pass we could:

* get rid of the crazy Makefile (one pass would probably be fast enough to just always do generation)
* use the same tool for ourselves as 3rd parties
* get rid of the crazy Makefile

A big part of the perf hit is parsing ~ the whole codebase multiple times.

sttts:
> A big part of the perf hit is parsing ~ the whole codebase multiple times.

This is worth a prototype. This was not in our focus at all, but if your claims holds, it would be a great improvement.

pwittrock:
@thockin I like the idea.  Though it wasn't the original focus, any area where we can simultaneously reduce both development complexity and build times seems like a good place to focus our efforts.

I suggest we prototype performance optimizations after completing the MVP described in this proposal.  I am guessing that doing so will require some restructuring of the code generator libraries to inject singleton instances of dependencies.

sttts:
> I suggest we prototype performance optimizations after completing the MVP described in this proposal. I am guessing that doing so will require some restructuring of the code generator libraries to inject singleton instances of dependencies.

The MVP exists with the shell scripts already. We need to prototype a binary which can launch the other generators. Command line is actually secondary for that.

pwittrock:
What is still required to get consensus on this?

> Command line is actually secondary for that

Do you mean compiling them into a single binary is secondary?

sttts:
> Do you mean compiling them into a single binary is secondary?

There is work to be done to make the generators work inside one binary (some preparations for that already merged, so it shouldn't be too hard). But to evaluate whether Tim's use-case has any chance to work out performance-wise (it would be fantastic for our tooling complexity), the actual command line is secondary.

But: if the use-case "generate for the whole kube/kube repo" is feasible, it might have bigger influence on our generator command line: we have more than one clientset in the repo, i.e. we would need a way to specify that on a kubegen commandline, while sharing the gengo Universe (i.e. to avoid re-parsing for every clientset).

thockin:
Even if we reduced the Makefile to a single target that figured out what needed to be generated-for, it would be a HUGE reduction in complexity.

pwittrock:
Thanks @thockin and @sttts for the feedback.  Your comments have been insightful and will help come up with an implementation that will result in better future iterations N+.

Now that this proposal has been open for discussion for nearly 3 weeks without any comments suggesting significant changes in design or goals.  Can we call this lazy consensus and assume that we will address additional things we discover as we implement V0 and then iterate?  Is there anything we need to resolve that we can't iterate on after this is merged?

> But to evaluate whether Tim's use-case has any chance to work out performance-wise (it would be fantastic for our tooling complexity), the actual command line is secondary.

I suggest we worry about performance optimizations after we have a working implementation of the desired user experience.

> But: if the use-case "generate for the whole kube/kube repo" is feasible, it might have bigger influence on our generator command line: we have more than one clientset in the repo, i.e. we would need a way to specify that on a kubegen commandline, while sharing the gengo Universe (i.e. to avoid re-parsing for every clientset).

We will want to get here eventually, but I would prefer launch something meeting the use case to reduce complexity for extensions (since they extension authors will not have already built out complex Makefiles that meet their needs).  It sounds like there is enough complexity re multi-clientsets that we don't want to block starting development on issues related to this.

thockin:
I have no further opinion on this PR as-is :)

sttts:
> We will want to get here eventually, but I would prefer launch something meeting the use case to reduce complexity for extensions (since they extension authors will not have already built out complex Makefiles that meet their needs). It sounds like there is enough complexity re multi-clientsets that we don't want to block starting development on issues related to this.

It makes a big difference in scope and the command line interface. I don't think it's a good idea to leave this question open if we don't have a plan how to extend the tool towards that use-case. I am fine with starting small, but let's flash out a sketch how this "multi clientset" use-case can look like. Until that I am against calling the proposal ready.

pwittrock:
> It makes a big difference in scope and the command line interface

I don't have a great grasp of the details around multi-clientset.  Can you elaborate?  How does it change the interface for the command?  What additional information is missing that will need to be present?

sttts:
> I don't have a great grasp of the details around multi-clientset. Can you elaborate? How does it change the interface for the command? What additional information is missing that will need to be present?

Now the command does one clientset, i.e. all given API groups are put into one clientset. For the kubernetes/kubernetes use-case we need multiple sets of API groups and API directories. I could imagine this:

```shell
kubegen \
  internal --api-dir k8s.io/api --api-dir --target pkg/client pkg/apis apps/v1 apps/v1beta1 core/v1 \
  versioned --api-dir k8s.io/api --target k8s.io/client-go/kubernetes apps/v1 core/v1 \
  ...
```

This would also include that we detect overlapping directories for deepcopy, conversions, etc. (all those file which are created inside the api dirs) and avoid double-generation.

sttts:
I want to see a very minimal proof-of-concept of the multi-clientset idea to proof that the speed increase of not-reparsing 25 times pays out enough. If it does, it's awesome for Kubernetes itself. If not, we don't need that additional complexity in the CLI implementation.

pwittrock:
The multi-clientset as shown definitely muddies up the interface and seems like it is more complicated and harder to build consensus around.  It is also trying to solve a different problem - speeding up Kubernetes builds vs improving development velocity by reducing complexity.

I am worried that focussing on performance optimizations will keep us from being able to move forward on the original stated goal - make running code generators for extensions accessible to non-kubernetes-veterans.  While I would love to reduce complexity in the main repo and support multi-clientsets, I don't have the domain expertise to drive such a thing.

Why don't we agree that multi-client set is out of scope for v1 and revisit in the future.

sttts:
> Why don't we agree that multi-client set is out of scope for v1 and revisit in the future.

If our command line is extensible towards a multi-client use-case, I am fine with that plan. But I would like to keep that door open.

IMO, using a multi-subcommand syntax the command line is not that bad. The question is whether the CLI library that we will use supports this.

pwittrock:
I think we have enough agreement on the details to green light development.  We can leave the door open to adapting this proposal based on feedback given during development.

> IMO, using a multi-subcommand syntax the command line is not that bad. The question is whether the CLI library that we will use supports this.

Lets just use one that does.  (e.g. cobra)

> If our command line is extensible towards a multi-client use-case, I am fine with that plan. But I would like to keep that door open.

Sure.  As long as trying to do this doesn't block making progress on the stated goals and keep us from making *any* improvements.

sttts:
@pwittrock and I talked on slack of how to go forward with this. There was some misunderstanding about who does what and how the bigger use-case of multi-clientsets fits into the design proposal and its implementation plan.

As far as I see, there are the following work items:

1. sketch how multi-clientsets can look like in the CLI as an extensions of the single-clientset use-case
2. refactor `main.go` of defaulter-gen, conversion-gen, client-gen, lister-gen, informer-gen, deepcopy-gen to be free of non-command-line logic such that the generators can be instantiated either in kubegen or in their standalone-binaries
3. prototype multi-clientsets without reparsing packages again and again (requires the client-gen part of 2)
4. implement this proposal (requires 2)
5. extend it to multi-clientset (requires 1, 2, 3, 4)

My plan is to do the most complicated one of (2), namely client-gen as a blueprint for the others ASAP, possibly this or early next week. I would be happy if @pwittrock or his colleagues can support with refactoring the other generators' `main.go`s following the client-get example.

When client-gen's `main.go` is done I will do the clientset-only prototype of (3) with a shared gengo universe until KubeCon.

Step 4 does not depend on (3) to be finished. For (1), here is an extension from the previous sketch that should unblock (4):

```
kubegen --dry-run --license BSD \
  internal --apis-dir k8s.io/api --apis-dir pkg/apis --client-output pkg/client --api-version apps/v1 --api-version apps/v1beta1 --apiversion core/v1 -- \
  versioned --apis-dir k8s.io/api --client-output k8s.io/client-go/kubernetes --api-group apps --api-group core \
  ...
```

In other words: we define a separator `--` between sub-commands and their flags and pre-parse the command line into multiple logical lists of flags, which we then pass to cobra. I.e. the upper command will behave as two single, independent invocations:

```
kubegen internal --dry-run --license BSD --apis-dir k8s.io/api --apis-dir pkg/apis --client-output pkg/client --api-version apps/v1 --api-version apps/v1beta1 --apiversion core/v1
kubegen versioned --dry-run --license BSD --apis-dir k8s.io/api --client-output k8s.io/client-go/kubernetes --api-group apps --api-group core
  ...
```

This way we can implement the single-clientset use-case soon'ish, with a feasible extension to multi-clientset later, without too much ugliness in the command line.

Potentially, we will add more sub-commands to kubegen, especially we know that we need `kubegen deepcopy <directories-that-are-no-apis>`. The multi-clientset sketch above naturally extends this way, i.e. we can handle real api-groups and support other non-api generation steps in parallel.

pwittrock:
SGTM

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,341 @@
...
+## Running via Bazel
+
+### Installing via Bazel
+
+Add the following to the project *WORKSPACE* file
+
+```py
+http_archive(
+    name = "io_k8s_rules_go",
+    url = "https://github.com/kubernetes/bazelbuild/releases/download/v1.8.0/rules_go-1.8.0.tar.gz",
```
ixdy: is this a new repo? or is it a clone of something?

contributors/design-proposals/code-generator/kubegen.md
ixdy: having a separate `bazel run` command means that you can't just do `bazel build` and have it do the right thing.

for MVP this is fine, but you can certainly get more benefits by turning this into something more like a `genrule`.

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,341 @@
...
+Kubernetes provides a collection of code generators to assist with developing APIs.  These code generators generate
+useful libraries such as clients, openapi definitions, copying go structs, etc.
+
+The `kubegen` uses directory structure conventions to detect Kubernetes APIs definitions and run the standard
+set of Kubernetes code generators.
+
+There are 2 methods for running kubegen
+
+- Directly through the kubegen command line by downloading the binary and running it from the project root
+- Through Bazel by adding rules to `WORKSPACE` and `BUILD.gazel`
```
ixdy: typo: `gazel` should be `bazel`?

contributors/design-proposals/code-generator/kubegen.md
```diff
@@ -0,0 +1,341 @@
...
+)
+```
+
+### Running the Bazel target
+
+```sh
+bazel run //:kubegen
+```
+
+### Bazel options
```
ixdy: by "options" I'm assuming you mean attributes of the `kubegen` rule?

contributors/design-proposals/code-generator/kubegen.md
ixdy: one other thing that might be worth noting is that calling `bazel run kubegen` may need to be followed by `bazel run gazelle` or whatever, since source files may change.

PR #1327, opened 2017-11-03
change 0000-kep-template

chenhonggc:
change 0000-kep-template.md
I think `Kubernaut ` should be `Kubernetes `


cblecker:
This is not a typo.

PR #1330, opened 2017-11-03
Outside collaborator role added

idvoretskyi:
Added and defined the "outside collaborator" role.

### DIFF
--- a/community-membership.md
+++ b/community-membership.md
@@ -8,6 +8,7 @@ scoped by these sub-areas.
 
 | Role | Responsibilities | Requirements | Defined by |
 | -----| ---------------- | ------------ | -------|
+| outside collaborator | new active contributor to the community | sponsored by 1 member.  existing contributions to the project. | Kubernetes GitHub org member. |
 | member | active contributor in the community | sponsored by 2 reviewers.  multiple contributions to the project. | Kubernetes GitHub org member. |
 | reviewer | review contribution from other members | history of review and authorship in an area | OWNERS file reviewer entry. |
 | approver | approve accepting contributions | highly experienced and active reviewer + contributor to an area | OWNERS file approver entry|


idvoretskyi:
cc @kubernetes/sig-contributor-experience-pr-reviews

community-membership.md
```diff
@@ -8,6 +8,7 @@ scoped by these sub-areas.
 
 | Role | Responsibilities | Requirements | Defined by |
 | -----| ---------------- | ------------ | -------|
+| outside collaborator | new active contributor to the community | sponsored by 1 member.  existing contributions to the project. | Kubernetes GitHub org member. |
```
spiffxp: defined by: read-only collaborator on a repository in the kubernetes org

How many people do we have that are collaborators vs. members? 

My concern with putting this in the table is it might imply to someone that this is a required step in the ladder.  I would much prefer people just jump straight to member, than think they have to go through this first.  Managing collaborators is noisier.

spiffxp:
How many people do we have that are collaborators vs. members? 

My concern with putting this in the table is it might imply to someone that this is a required step in the ladder.  I would much prefer people just jump straight to member, than think they have to go through this first.  Managing collaborators is noisier.

cblecker:
+1 to @spiffxp's comments.

What does being an outside collaborator get you? Assigned issues, sure. Do we know which [bot commands](https://github.com/kubernetes/test-infra/blob/master/commands.md) are gated by collaborators and which ones use `IsMember`? It looks like a collaborator may be able to LGTM things (https://github.com/kubernetes/test-infra/blob/7596869f74ce8cc76e5a6bcad84c3723911b29a5/prow/plugins/lgtm/lgtm.go#L91)?

spiffxp:
Collaborators can do all the things a member can do... but only for the repos they're specifically collaborators on, vs repos in the entire org.  This means you could eg: give someone /lgtm privileges on kubernetes/charts without giving them /lgtm privileges on kubernetes/kubernetes

idvoretskyi:
@spiffxp the reason why I've added this role to the table - as we already have this role defined below - https://github.com/kubernetes/community/blob/58a1ba55effe150b4f5f1e2d770f86ef3041ffd8/community-membership.md#new-contributors.

If you feel that we shouldn't define the "collaborator" role in a table, I'd remove it from the entire document/reformat the document to follow the consistency.

idvoretskyi:
@spiffxp @calebamiles any final decision on this?

I'd like to have it merged or closed, whatever will be decided (but not to keep it open forever).

cblecker:
IIRC, this was talked about briefly at the last contribex meeting, and most who commented were not in favour of introducing another level.

idvoretskyi:
Closing this since no decision for almost 60 days.

spiffxp:
Argh my bad, I meant to open an issue for this.  I made noises about removing this to steering@kubernetes.io, and didn't hear a wild no.  It's low on my list right now though.  I'd like to survey which repos are using collaborators to understand the scope of the role today.

PR #1338, opened 2017-11-05
typo

chenhonggc:
I think `Ubernetes` should be `Kubernetes`


liggitt:
No, ubernetes refers to multi-cluster setups

chenhonggc:
@liggitt  Thanks.

PR #1341, opened 2017-11-06
Update protobuf.md

YanTingLittleCute:


liggitt:
The test is actually called a fuzzer

PR #1345, opened 2017-11-06
Update protobuf.md

gin419:


liggitt:
the generated functions are named Marshal and Unmarshal

PR #1349, opened 2017-11-06
Update sig-testing's meeting to account for DST changeover

spiffxp:
Our meeting is defined in terms of US pacific wall-clock time, which
just jumped ahead an hour thanks to the wonderfully helpful and
productivity-increasing daylight savings time ritual that we follow
in the US.


spiffxp:
fyi @kubernetes/sig-testing-pr-reviews

spiffxp:
ref: https://github.com/kubernetes/community/issues/841

PR #1350, opened 2017-11-06
Allow SIGs/WGs to specify meetings in any time zone

### LINKED ISSUE #841: SIGs list uses UTC, but the SIGs generally don't

jbeda:
Fixes #841


jbeda:
Fixed up SIG-Testing too during a rebase so that is why it is included in the first commit.

sigs.yaml
```diff
@@ -144,7 +149,8 @@ sigs:
       company: Redspread
     meetings:
     - day: Friday
-      utc: "16:00" # Actually 9AM Pacific, but timezone support issue is https://github.com/kubernetes/community/issues/841
+      time: "16:00" # Actually 9AM Pacific, but timezone support issue is https://github.com/kubernetes/community/issues/841
```
cblecker: Perhaps we can fix this up, per the comment?

generator/README.md
```diff
@@ -5,6 +5,16 @@ different Special Interest Groups (SIGs) of Kubernetes. The authoritative
...
 updates must be done there.
 
+The schema for this file should be self explanatory. However, if you need to see all the options check out the generator code in `app.go`.
+
+**Time Zone gotcha**:
+Time zones make everything complicated.
+And Daylight Savings time makes it even more complicated.
+Meetings are specified with a time zone and we generate a link to http://www.thetimezoneconverter.com/ so that people can easily convert it to their local time zone.
+To make this work you need to specify the time zone in a way that that web site recognizes.
+Practically, that means US pacific time must be `PT (Pacific Time)`.
```
cblecker: The converter also understands UTC offsets (e.g. `UTC-7`). Perhaps we could use that instead?

cblecker: NVM.. we run into the same issue there then of DST.

lavalamp:
lgtm, thanks.

PR #1353, opened 2017-11-06
fix the contribx sig time

grodrigues3:
Fix the timezone/DST issue for contribx


sigs.yaml
```diff
@@ -315,8 +315,8 @@ sigs:
       company: CoreOS
     meetings:
     - day: Wednesday
-      time: "16:30"
-      tz: "UTC"
+      time: "9:30"
+      tz: "PST"
```
cblecker: This needs to be `"PT (Pacific Time)"` for the conversion link to work

grodrigues3:
Fixed!  I noticed that yesterday and forgot to push.  Thanks for the careful review.

PR #1371, opened 2017-11-07
Add staging godeps to godep instructions

thockin:
This was missed in the last rewrite.

### DIFF
--- a/contributors/devel/godep.md
+++ b/contributors/devel/godep.md
@@ -145,6 +145,19 @@ calculate which deps are needed, and rebuild the database.  It will also
 regenerate other metadata files which the project needs, such as BUILD files and
 the LICENSE database.
 
+## Saving deps in staging repos
+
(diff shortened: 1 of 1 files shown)


contributors/devel/godep.md
```diff
@@ -145,6 +145,18 @@ calculate which deps are needed, and rebuild the database.  It will also
 regenerate other metadata files which the project needs, such as BUILD files and
 the LICENSE database.
 
+## Saving deps in staging repos
+
+Kubernetes stores some code in a directory called `staging` which is handled
+specially, and is not covered by the above.  If you modified any code under
+staging, you'll also need to update deps there:
```
mindprince: I didn't modify any code under staging and still needed to run this. The reason was that the dependency I was updating `golang.org/x/sys/unix` is also a dependency in those staging Godeps.json.

thockin: added words

PR #1372, opened 2017-11-08
Change the sig-multicluster meeting name and update the meeting time.

perotinus:
This meeting ID should not have the joining problems that we've been having for the last few weeks.

/cc @csbell @quinton-hoole


perotinus:
Thank you!

PR #1386, opened 2017-11-09
Update SIG-CLI time to be PT

jbeda:


fabianofranz:
Thanks @jbeda! Does this actually affect the calendar entry for the SIG meeting?

/lgtm
/approve no-issue

jbeda:
No -- unfortunately, this isn't set up to update/populate calendar entries. This is just a simple web page. That would be cool thought!

PR #1395, opened 2017-11-14
Update api_changes.md

phinoceslee:


liggitt:
the text is correct as-is
/close

PR #1400, opened 2017-11-14
Proposal: Cluster Scoped Resources

nikhildl12:
Initial proposal for cluster scoped resources

Related: https://github.com/kubernetes/kubernetes/issues/19080 and https://groups.google.com/forum/#!topic/kubernetes-users/eUUrdlBwa7g

### DIFF
--- a/contributors/design-proposals/scheduling/cluster-scoped-resources.md
+++ b/contributors/design-proposals/scheduling/cluster-scoped-resources.md
@@ -1,0 +1,88 @@
+# Cluster Scoped Resources
+
+## Abstract
+Cluster scoped resources are consumable resources that do not belong to any specific node but instead are available across mulitple nodes in a cluster. These resources are accounted as other consumable resources and should be usable by the scheduler while deciding if a pod can actually be scheduled.
+
+
+## Motivation
+Resources in Kubernetes such as cpu and memory are available at a node level and can be consumed by pods by requesting them. However there are some resources that do not belong to a specific node, but they are consumable across all or a group of nodes in the cluster. Few such use cases are mentioned below.
+
+#### Use Cases
+1. Software Licenses that can be shared by pods across the entire cluster
+
+2. IP Addresses:
+   Nodes in a cluster can be partitioned into multiple network scopes and each network scope can have a certain number of available IPs that can be assigned to pods running on those nodes. Hence, the nodes belonging to a network scope can collectively run as many pods as the IP capacity in that scope.
+
+3. Rack storage:
+   Locally attached shared storage in a rack, which is consumable by pods on nodes within a rack
+
+4. Network Bandwidth:
+   Network bandwidth shared by pods. Depending on the network topology of a cluster, pods on multiple nodes will be sharing network bandwidth with each other. In use cases where pods need guranteed network throughput, representing bandwidth as a cluster resource is essential for scheduling such pods.
+
+## Goals
+The goal is to define mechanisms to expose and consume cluster scoped resources
+
+## Design
+
+### ClusterResource type
+```
+// pkg/api/types.go:
+
+// ClusterResourceQuantity represents quantity of a ClusterResource
+type ClusterResourceQuantity struct {
+	Quantity resource.Quantity `json:"quantity"`
+	// NodeSelector is a label query over nodes which collectively provide resource Quantity
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+}
+
+// ClusterResource represents a resource which is available at a cluster level
+type ClusterResource struct {
+	TypeMeta   `json:",inline"`
+	ObjectMeta `json:"metadata,omitempty"`
+	// +optional
+	Status ClusterResourceStatus `json:"status,omitempty"`
+}
+
+type ClusterResourceStatus struct {
+	// Capacity represents the total quantity of ClusterResource
+	// +optional
+	Capacity []ClusterResourceQuantity `json:"capacity,omitempty"`
+	// Allocatable represents the quantity of ClusterResource that is available for scheduling
+	// +optional
+	Allocatable []ClusterResourceQuantity `json:"allocatable,omitempty"`
+}
+```
+`ClusterResourceStatus` captures the capacity and allocatable quantity for a `ClusteResource` in the form of `ClusterResourceQuantity`. `ClusterResourceQuantity` represents the quantity of a `ClusterResource` which is collectively consumable on nodes selected by `NodeSelector`.
+`NodeSelector` is a label query over the nodes, which collectively provide this `ClusterResource`. This field is optional, and if not specfied, it means that the `ClusterResource` is consumable across all nodes in the cluster.
+
+
+### Consuming ClusterResources
+
+ClusterResources are consumable by pods just like CPU and memory, by specifying it in the pod request. The scheduler should take care of the resource accounting for ClusterResources so that no more than the available amount is simultaneously allocated to Pods. The prefix used to identify a ClusterResource coule be 
+```
+pod.alpha.kubernetes.io/cluster-resource-
+```
+
+### Accounting in scheduler
+
+ClusterResources should be tracked as normal consumable resources and should be considered by the scheduler when determining if a pod can actually be scheduled
+
+```
+// kubernetes/plugin/pkg/scheduler/schedulercache/cluster_info.go
+
+// ClusterInfo is cluster level aggregated information.
+type ClusterInfo struct {
+
+	clusterResources map[string]*ClusterResource
+}
+
+// kubernetes/plugin/pkg/scheduler/schedulercache/cache.go
+
+type schedulerCache struct {
+   	... 
+   	cluster   *ClusterInfo
+}
+```
+
+`clusterinfo` is added to scheduler cache to do accounting for ClusterResources consumed by pods. `clusterInfo` will be exposed to the predicate and priority functions in order to take ClusterResources into consideration while making scheduling decisions.


contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,76 @@
+# Cluster Scoped Resources
```
timothysc: please follow the KEP process outlined by @kubernetes/sig-architecture-feature-requests

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Abstract
+Cluster scoped resources are consumable resources that do not belong to any specific node but instead are available across mulitple nodes in a cluster. These resources are accounted as other consumable resources and should be usable by the scheduler while deciding if a pod can actually be scheduled.
+
+
+## Motivation
```
timothysc: Software licenses are the most common reason for such features in other systems.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,76 @@
...
+Resources in Kubernetes such as cpu and memory are available at a node level and can be consumed by pods by requesting them. However there are some resources that do not belong a specific node, but they are consumable across all or a group of nodes in the cluster. As an example, IP addresses in a pool can be shared across pods running on multiple nodes in a network scope. Another use case could be, locally attached shared storage in a rack, which is consumable across several nodes. Hence there is a need to represent such a resource at cluster level which is consumable acroass all or a group of nodes in the cluster.
```
zhouhaibing089: This is more like a node _group_ scoped resources in the examples.

davidopp:
Thanks for writing this, it's definitely a feature we have been talking about for a while.

I think a complete solution to this problem should consider how the resource allocator for the cluster-level resource fits in. I think that cluster-scoped resources are likely to have some kind of external allocator, for example the agent that hands out IP addresses or software licenses. It's important for the scheduler's view of free resources to stay in sync with that of the external allocator, which has the authoritative information, so that we can minimize the likelihood that a container starts up and finds that the resource is not actually available.

For example, with a normal resource the scheduler assumes the resources become freed when the pod terminates or is deleted. But with cluster-level resources, if we leave the allocation and deallocation of the resource up to the container, it might be possible to leak resources (container forgets to release the IP address or license, or gets killed before it can, so the resource is still allocated but the scheduler thinks it is free because the pod has terminated). So maybe the scheduler should be responsible for reserving the resource from the allocator before binding the pod, and unreserving the resource via the allocator when the pod terminates. It's probably quite complicated to ensure that a container only tries to allocate resources that have been reserved for it, so it's probably not a "secure" solution but might be good enough.

One approach is what we did for PDB and ResourceQuota, where decrementing the amount free is synchronous with requesting it (in the cluster-scoped resource case, this would mean the scheduler decrements the free) but replenishing the resource when it is no longer in use is asynchronous and done by a separate controller (could be the agent that is responsible for the cluster-level resource, when a container deallocates the resource).

contributors/design-proposals/scheduling/cluster-scoped-resources.md
smarterclayton: Please add a list of 5-8 example resources that would be tracked like this.  I’d like more validation and concrete discussion on each type to guide design.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,76 @@
...
+// ClusterResourceQuantity represents quantity of a ClusterResource
+type ClusterResourceQuantity struct {
+	Quantity resource.Quantity `json:"quantity"`
+	// NodeSelector is a label query over nodes which collectively provide resource Quantity
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+}
+
+// ClusterResource represents a resource which is available at a cluster level
+type ClusterResource struct {
```
smarterclayton: If this is a form of quota, it should be named as such - ClusterResourceNodeQuota.  It’s not actually clear how this api aligns with ResourceQuota, please comment to that effect.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,76 @@
...
+The goal is to define mechanisms to expose and consume cluster scoped resources
+
+## Design
+
+### ClusterResource type
+```
+// pkg/api/types.go:
+
+// ClusterResourceQuantity represents quantity of a ClusterResource
+type ClusterResourceQuantity struct {
```
vikaschoudhary16: How is the discovery/initialization flow?

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,76 @@
...
+
+// kubernetes/plugin/pkg/scheduler/schedulercache/cache.go
+
+type schedulerCache struct {
+   	... 
+   	cluster   *ClusterInfo
+}
+```
+
+`clusterinfo` is added to scheduler cache to do accounting for ClusterResources consumed by pods. `clusterInfo` will be exposed to the predicate and priority functions in order to take ClusterResources into consideration while making scheduling decisions.
```
vikaschoudhary16: How `clusterinfo` will be build?

contributors/design-proposals/scheduling/cluster-scoped-resources.md
vishh: Is KEP now a requirement or a recommendation? That was not clear from the contributor summit discussions.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
vishh: +1. There are many use cases for extending resource APIs and I'd like to first get a collection of use-cases before identifying possible solutions.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,76 @@
...
+```
+`ClusterResourceStatus` captures the capacity and allocatable quantity for a `ClusteResource` in the form of `ClusterResourceQuantity`. `ClusterResourceQuantity` represents the quantity of a `ClusterResource` which is collectively consumable on nodes selected by `NodeSelector`.
+`NodeSelector` is a label query over the nodes, which collectively provide this `ClusterResource`. This field is optional, and if not specfied, it means that the `ClusterResource` is consumable across all nodes in the cluster.
+
+
+### Consuming ClusterResources
+
+ClusterResources are consumable by pods just like CPU and memory, by specifying it in the pod request. The scheduler should take care of the resource accounting for ClusterResources so that no more than the available amount is simultaneously allocated to Pods. The prefix used to identify a ClusterResource coule be 
+```
+pod.alpha.kubernetes.io/cluster-resource-
```
vishh: I'm not a fan of special prefixes. I'd like to see if we can avoid overloading resource names.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
timothysc: /cc @jdumars

contributors/design-proposals/scheduling/cluster-scoped-resources.md
ConnorDoyle: +1, we just moved away from this pattern with extended resources.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: Added few use cases

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: @vishh @timothysc: is this the template that needs to be followed: https://github.com/kubernetes/community/blob/master/keps/0000-kep-template.md

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: `ClusterResource` is an api type that represents a cluster scoped resource. However it's integration with resourcequotas needs to be added, probably at later a phase such as `beta`?

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: Cluster admin or other controllers will post the `ClusterResource` objects that captures the capacity and allocatable quantities of a`ClusterResource`, which will then be used by scheduler

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: `clusterinfo` can be build similar to how we build `nodeInfo` since scheduler will be watching for ClusterResources

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: It can follow fully-qualified resource names similar to extended resources, but we need to see how will those be differentiated

nikhildl12:
@davidopp @timothysc @vishh: I would like to understand what could the next steps be for this proposal. As a first action item, I can submit this in the form of a KEP:  https://github.com/kubernetes/community/blob/master/keps/0000-kep-template.md

cblecker:
@nikhildl12 One important step is to sort out your CLA, as outlined here: https://github.com/kubernetes/community/pull/1400#issuecomment-344432810

/ok-to-test

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+### Accounting in scheduler
```
bsalamat: We have extended resources and several types of first class resources in the scheduler already. I think it would be possible to come up with a single presentation that covers all of these types. For example, I don't see much of a difference between a cluster resource and extended resource from scheduler's point of view. An extended resource with an additional "type" can present a cluster resource.

contributors/design-proposals/scheduling/cluster-scoped-resources.md
```diff
@@ -0,0 +1,88 @@
...
+
+ClusterResources should be tracked as normal consumable resources and should be considered by the scheduler when determining if a pod can actually be scheduled
```
bsalamat: Another important aspect of cluster resources which is not covered here is how to bind these resources to a chosen node during/after scheduling. A fairly complex logic is already added to scheduler to handle provisioning and binding PVs to nodes during scheduling. Similar processes may be needed for other resources, such TPUs, etc. I think that aspect should be covered by the proposal.

@vishh @jiayingz

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: The key difference between these two resources is the "scope".
Currently extended resources are exposed as a part of node status because of their nature of being tied to a node, while cluster scoped resources have to be represented outside the scope of a node. But we can surely have a comprehensive API that covers both. From the scheduler's point of view, it will need some additional logic to calculate and cache the available capacity of a cluster scoped resource across a set of nodes

contributors/design-proposals/scheduling/cluster-scoped-resources.md
nikhildl12: To cover that aspect,  I prefer the approach mentioned by @davidopp in his previous comment. The external agent/controller which exposes the available capacity of this resource can be made responsible for binding or making sure that those resources are ready to use when a pod is going to run on a node. Similarly when a pod dies, that agent needs to deallocate/unbind the corresponding resource and increment the available quantity so that it can be used for scheduling of new pods

contributors/design-proposals/scheduling/cluster-scoped-resources.md
bsalamat: Yes, the "ResourceClass" that @jiayingz is working on is an effort in that direction to provide a comprehensive API to represent various types of resources, including cluster resources.

jiayingz: Yes. As @bsalamat mentioned, we are working on a new Resource API proposal that aims to provide a comprehensive API for both node-level resources and cluster-level resources. Here is the current PR:
https://github.com/kubernetes/community/pull/782
It is still WIP and the current plan is to focus on node-level resources during the initial phase. But I think even the initial API should help solve some of the listed problems here. Please take a look and let us know if you see any missing pieces.

krmayankk:
why has this been abandoned ? The proposal seems fair ?

PR #1401, opened 2017-11-15
KEP template update

jbeda:
This is based on input from @timothysc.  Still need to reconcile this with the instructions.

### DIFF
--- a/contributors/design-proposals/architecture/0000-kep-template.md
+++ b/contributors/design-proposals/architecture/0000-kep-template.md
@@ -49,18 +49,32 @@ To get started with this template:
 
 The canonical place for the latest set of instructions (and the likely source of this file) is [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/0000-kep-template.md).
 
-## Metadata (remove this)
-
-The `Metadata` section is intended to support the creation of tooling around the KEP process.
+The `Metadata` section above is intended to support the creation of tooling around the KEP process.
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/architecture/0000-kep-template.md
```diff
@@ -1,67 +1,19 @@
...
-it "front matter".
-
-TODO(jbeda): Do we want to have a "people database" to reduce the amount of
-duplication on naming people here?  This would be a simple map of github ID to
-name and contact info.
-
-```yaml
-kep-number: draft-XXX
+---
+kep-number: draft-YYYYMMDD
```
timothysc: I know @mattfarina had comments on possibilities here.

bgrant0607:
@jbeda High-level explanation of main changes please?

jbeda:
@bgrant0607 Sure! Sorry.

* Move metadata to front so we can use it from content generation systems.  Also gets more compact rendering from github.
* Move away from the set of terminology that requires users to read links to understand it to a flow that is more familiar to design docs.  Basically, we define scope/motivation and the proposal itself.
* Suggest that authors use user stories to communicate the impact of the proposal. This keeps it focused on how the proposal will change the project from the user PoV vs. the nitty gritty details.
* Prompt for "Risks and Mitigations" to encourage authors to think about security and the impact this'll have on the project in a longitudinal way.

This was based on the feedback that @timothysc gave during the last SIG-Architecture meeting.

PR #1402, opened 2017-11-15
sig-auth: move role aggregation proposal to proposal directory

ericchiang:
https://github.com/kubernetes/community/pull/1219 added a proposal to the wrong directory.

/sig auth
/assign @deads2k


liggitt:
/lgtm
Thanks

PR #1408, opened 2017-11-15
In README, add ptr to CONTRIBUTING.

monopole:
In README, add ptr to CONTRIBUTING.
Some new folks were having trouble finding this; the
additional link might help with browse and search.

In CONTRIBUTING
 * further clarify order of events for preparing an initial PR,
 * provide some good links to find initial work,
 * Trim some verbiage.

### DIFF
--- a/sig-cli/CONTRIBUTING.md
+++ b/sig-cli/CONTRIBUTING.md
@@ -1,21 +1,14 @@
 # Contributing
 
-The process for contributing code to Kubernetes via SIG-cli [community][community page].
+_Welcome to the Kubernetes SIG-cli contributing guide._
 
-**Note**: This page is focused on helping new contributors become active
-members of the community through sustained contributions.
+This guide is meant to help people become active members of the
+SIG-cli [community][community page] through sustained contributions.
 
-## Introduction
-
-Welcome to the Kubernetes sig-cli contributing guide.  We are excited
-about the prospect of you joining our [community][community page]!
+We are excited by the prospect of you joining our
+[community][community page]!
 Mentoring and on-boarding new contributors is critical to the success
 of the project.
-
-Please be aware that all contributions to Kubernetes require time
-and commitment from project maintainers to direct and review work.
-This is done in additional to many other maintainer responsibilities, and
-direct engagement from maintainers is a finite resource.
 
 ## Pick your track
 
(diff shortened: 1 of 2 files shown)


monopole:
@mengqiy

monopole:
@pwittrock if sleepless

sig-cli/CONTRIBUTING.md
```diff
@@ -97,35 +90,24 @@ work.
...
 
 ### Adopt an issue
 
-Pick up an [issue] from the backlog by commenting on the issue that you would like to work on it.
-Be sure to mention the author of the issue as well as the SIG cli members `@seans3` and `@mengqiy`.
-
-Using the following comment will make it easier for us to search for issues folks want to have
-assigned to them:
-
-`cc @seans3 @mengqiy I would like to take this`
```
pwittrock: I think we are going to want to mention some folks directly so the issue gets past our filters.

sig-cli/CONTRIBUTING.md
```diff
@@ -97,35 +90,24 @@ work.
...
+Browse the [issues] for something you'd like to work on.
+The label "[good first issue]" has been applied to
+issues that might be more appropriate for newcomers.
 
-**Note:** Don't do this unless you will start work on the issue within a few days of being assigned.
+An alternative and interesting presentation
```
pwittrock: No sure how helpful this will be to new contributors.  Is that work regularly groomed, up to date and well thought out enough for new contributors to pick up arbitrary issues from it?

sig-cli/CONTRIBUTING.md
pwittrock: I see we do this later

sig-cli/CONTRIBUTING.md
monopole: There's a "good first issue" query linked to earlier in the text that serves that purpose.  The text implies those issues are is curated, and that the wiki is merely alternative and interesting.

I like the idea of a page that provides an issue overview / story / broad taxonomy that doesn't require one to both know all the labels and how to use them https://github.com/kubernetes/kubectl/labels
Brian mentions it frequently, but he and Janet seemed to have stopped maintaining it last year after ~100 edits. Nothing in community points to it, so i thought we could give it a rank boost given its clearly sig-cli.

Perhaps we could file an issue to replace this wiki with human readable sentences or headings that link to pre-baked issue queries with particular labels.

Then its maintenance would be no worse that whatever effort is currently being put into applying labels.

regardless, removed link

sig-cli/CONTRIBUTING.md
monopole: trying to put in better (not perfect) order

PR #1416, opened 2017-11-19
Proposal: Node fence design

bronhaim:
Propose fencing mechanism in k8s cluster that allows to manage nodes in cluster with advanced fence actions


rootfs:
related to https://github.com/kubernetes/community/pull/124

rootfs:

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
...
+
+If the pod belongs to a ReplicaSet, once the termination state is set, the ReplicaSet controller will immediately start another instance of that pod.  However when the pod is part of a StatefulSet, the StatefulSet controller won’t start new instance of that pod until the kubelet responds with the pod’s status. This is because there is a permanent association between a StatefulSet member and its storage.  Not waiting would potentially result in multiple copies of a member, all writing to the same volume (leading to corruption or worse).
+
+### Cloud deployments
+When k8s is deployed on a cloud such as AWS or GCE, the autoscaler uses the Cloud Provider APIs to remove failed nodes (Failure detection and allowing recovery of StatefulSets is more of a side effect of the autoscaler which is more focused on using the Cloud Provider APIs to monitor load and trigger scale up/down events), effectively bounding the amount of time the node will stay in the “not ready” state and how long the scheduler will need to wait before it can safely start the pod elsewhere.  However in the case of bare metal, no such mechanism exists and recovery will be blocked until an admin intervenes.
```
derekwaynecarr: not all cloud deployments of k8s use the cluster autoscaler.

smarterclayton: the node controller is actually the one who does this as well.  the node controller removes instances the cloud provided determines as dead.  the node controller cannot remove instances that have kernel soft lockups or other permanent faults that can only be resolved by a restart.

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
...
+
+This is particularly problematic because all StatefulSet scaling events (up and down) block until existing failures have been recovered.  From some perspectives, it could be considered that each node hosting a StatefulSet member has become a single point of failure for that set.
+
+### Scope of this proposal
+Managing HA entities in cluster can relate to nodes and applications. In this scope we cover “node isolation” actions while node is not responsive until it becomes responsive again. Node isolation can be done by power management action (rebooting the node’s machine), storage fencing (preventing node from accessing), and cluster-wide isolation actions.
```
smarterclayton: I don't understand the first sentence.

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
+# Node fencing
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: Yaniv Bronhaim (ybronhei@redhat.com)
+
+## Overview
+Rationale and architecture for the addition of a fencing mechanism in Kubernetes clusters.
```
smarterclayton: Please reference the pod safety design and describe how this alters it (probably in scope)

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
...
+  - version: 1.0
+data:
+  - method.properties: |
+name=nodeA_eaton_on
+template=eaton_pdu  
+plug=2
+action=on
+```
+
+Finally fence configs:
```
derekwaynecarr: i feel like the api definition should encode intent and the means of reaching that intent should be abstracted from the user api.  right now, it feels like i need intimate knowledge of the config map(s) to know what to specify in a NodeFenceConfig which seems less than ideal to me.

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
+# Node fencing
```
smarterclayton: In general, I'd expect both a proposal and a working prototype to demonstrate the concept.  I consider a fencer as a valuable add-on to most kubernetes clusters, but the proposal needs to address either why every cluster would want it.

contributors/design-proposals/node/node-fence.md
smarterclayton: However, for the cloud provider, you should at least describe whether a kube add-on is appropriate, or simply the cloud provider's native health checking (i.e. if i have a GCP auto scaling group with an integrated health check, do I need this component?)

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
...
+isolation=[method-storage-iso1]]
+power_managment=[method-pm1, method-kdump]
+recovery=[method1,method2,method3]
+execution_policy=[seq\cun]
+```
+Notes:
+- Each configmap we relate a node name - this should be unique, and in the properties we set order for stages’ names.
+- execution_policy means: In list of method we might want to run all methods until all finish, or until only one finishes successfully (useful for sequential pm actions). 
+- Isolation\power_managment\recovery are parameters with list value which represent the method to run in each step of the fencing.
+#### Fence Method Config ###
```
smarterclayton: This seems like an implementation detail and doesn't relate to the proposal.

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
...
+### Configurations required
+- How to trigger fence devices/apis - general template parameters (e.g. cluster UPS address and credentials) and overrides values per node for specific fields (e.g. ports related to node)
+- How to run a fence flow for each node (each node can be attached to several fence devices\apis and the fence flow can defer)
+
+The following is done by `ConfigMap` objects:
+
+#### Node Fence Config
+For each node that supports fencing we will configure NodeFenceConfig - This object is needed to centralize the information about “how” the node can be “fenced” from the cluster - we will have 3 steps for fence “isolation”, “power-management”, “recovery”. In each we will have a list of methods to perform.
+
+Note: to keep those fence configuration we had two options - Adding attributes in metadata section to the existing node object or create ConfigMap that describe this information for each node with fencing options.
```
smarterclayton: I see this more as a specific implementation detail of the add-on - I don't consider whatever config defined here as canonical (or an API that other must respect).

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
...
+agent_name=fence_ipmilan
+must_success=Yes\No
+template_param1_name=value1
+template_param2_name=value2
+...
+```
+
+### Implementation
+#### Fence Controller
+The fence controller is a stateless pod instance running in cluster that supports fence. The controller contains two main loop: 1) identifies unresponsive nodes. 2) following fence crds.
```
smarterclayton: Update formatting here

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,528 @@
...
+
+First controller will trigger one executor job to live in cluster.
+
+Identification: The controller will identify unresponsive node by check their apiserver object, once node becomes “not ready” a fence treatment will be triggered by posting crd for fence that includes all the information to execute fence stages by Fence Executor. 
+
+Following fence process using crd: The controller will check iteratively all fence objects and validate their execution (base on their status and state in stage). If execution is recognized as stuck or failed, the controller will handle the running job and run a new one on different resource to handle the request. Also, the controller is responsible to move to next “step” when node is still not ready, crd already exists, and the step is running for a configuration amount of time.
```
smarterclayton: Explain why you are using a CRD vs setting annotations / state on the node.

contributors/design-proposals/node/node-fence.md
bronhaim: Basically Huamin started to code the controller behavior in https://github.com/rootfs/node-fencing and I plan to move on with that to demonstrate the concepts better. In pod-safety proposal you refer to everything I write in the overview, here (after the overview) we try to add more detailed flow for executing fencing automatically in cluster. 
It intends to define how we declare different fence phases to isolate and handle node, how we add the configuration to be available in the cluster and more implementation details (as you stated below) to code the Fencing controller and agents executor that are mentioned in the pod-safety design - will add it to the overview and reference to your commit pod-safety.md. 

[1] https://github.com/kubernetes/community/blob/16f88595883a7461010b6708fb0e0bf1b046cf33/contributors/design-proposals/pod-safety.md

contributors/design-proposals/node/node-fence.md
bronhaim: The controller and execution are quite intuitive. we need to store parameters for each node devices\apis, and admin must have intimate knowledge with those configmaps, each describes fencing device or api that need to be attached to the node.. do you suggest to use external store for describing the methods' fields?

contributors/design-proposals/node/node-fence.md
bronhaim: Mainly to avoid adding more noise in the node object. previously in the crd I stored the list of methods to perform, now I changed it to hold only the "step" and the executor fetch the methods list. As it is now its reasonable to set it as additional annotations to the node object. I will test this option as well, it simplifies the solution of course.

contributors/design-proposals/node/node-fence.md
bronhaim: More as an api that admin must follow to enable fence for node. the controller will validate those maps and report if parameters are missing, and I mention here that it can be also part of the node object which sounds less reasonable to me

contributors/design-proposals/node/node-fence.md
bronhaim: I couldn't find how to configure node controller to remove instances that the cloud provider signs as unhealthy. can you direct me to this logic?
This paragraph only needs to say that current autoscaler impl doesn't treat partitioned nodes effectively, it waits some time period and then try to scale down the unresponsive nodes. With fencing addon the partitioned nodes can be treated immediately and with more options for actions.
The fence controller logic can use the cloud provider events in addition to the kubelet health check. we also integrate with node-problem-detector that can be used in baremetal to start the fence flow.

contributors/design-proposals/node/node-fence.md
bronhaim: I wanted to refer to HA applications and mixed it up during the sentence. I'll remove this first sentence

contributors/design-proposals/node/node-fence.md
bronhaim: The method config structure part? But this is the core for working with all those external devices and apis. unless you think that those values need to be stored differently

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,398 @@
...
+          agent_name=fence_brocade
+          ipaddr=192.168.1.2
+          password=brocade_password
+          username=brocade_admin
+```
+
+### Implementation
+#### Fence Controller
+- The fence controller is a stateless pod instance running in cluster that supports fence.
+- The controller will identify unresponsive node by check their apiserver object, once node becomes “not ready” a fence treatment will be triggered by posting crd for fence to initiate fence flow by Fence Executor.
```
CaoShuFeng: What will happen if controller or executor's node become unresponsive?

rootfs:
cc @nirs @aglitke

contributors/design-proposals/node/node-fence.md
rootfs: looks a leader election process

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,398 @@
...
+
+
+## Aditional implementations details
+### Cluster fence policies
+Fencing policy allows each cluster to behave differently in case of connectivity issues. One of the main motivation for that is to prevent “fencing storms” from happening, but there are others. The controller is responsible to force the fence policy
+
+Fencing storm is a situation in which fencing for a few nodes in the cluster is triggered at the same time, due to an environmental issue.
+
+Examples:
+1. Switch failure - a switch used to connect a few nodes to the environment is failing, and there is no redundancy in the network environment. In such a case, the nodes will be reported as unresponsive, while they are still alive and kicking, and perhaps providing service through other networks they are attached to. If we fence those nodes, a few services will be offline until restarted, while it might not have been necessary.
```
beekhof: Sort of, there are downsides either way (deciding whether to fence or not).  Have a read of http://blog.clusterlabs.org/blog/2018/two-node-problems

The problem is that you have no way to know if it is a switch failure.
So either:
- you assume it's not a network failure, initiate fencing and potentially kill a healthy Pod (does it really count as healthy if no-one can reach it though) so it can be recovered, or
- you assume it is a network failure, don't initiate fencing and risk that there are multiple copies of some Pods

contributors/design-proposals/node/node-fence.md
```diff
@@ -0,0 +1,398 @@
...
+1. In a Cloud Kubernetes deployment, StatefulSets without an autoscaler should not require admin intervention in order to scale up or down when a member Pod was active on a lost worker
+
+In other words, the failure of a worker node should not represent a single point of failure for StatefulSets.
+
+## Admin experience
+### Configurations required
+- How to trigger fence devices/apis - general template parameters (e.g. cluster UPS address and credentials) and overrides values per node for specific fields (e.g. ports related to node)
+- How to run a fence flow for each node (each node can be attached to several fence devices\apis and the fence flow can defer)
+
+The following is stored in `ConfigMap` objects:
```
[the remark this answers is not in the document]
bronhaim: I agree.. However, new crd means new object in cluster with version and specific services. for key-value fields I saw that using the builtin structure configmap seems to be good enough for our purposes. how would you optimize the format with crd?

thockin:
What is the status on this one?

beekhof:
@thockin I'm in the process of trying to resurrect it.

After additional consultation we're moving to a Machine API based approach.

rob-nicholson:
@beekhof I'm just wondering about your last comment 
https://github.com/kubernetes/community/pull/1416#issuecomment-420482003  Does that mean that  you are not now trying to resurrect this proposal? Is there someplace I could  look for discussion/elaboration of the 'machine API based approach' to solving this problem?

beekhof:
@rob-nicholson I resurrected it in a different form as a KEP - https://github.com/kubernetes/community/pull/2763/

Rather than preaching a a specific implementation, it tried to describe the interaction points that solutions should respect.

In the end it didn't get a lot of attention and was later closed when KEPs moved to a new repo.
If people are interested, I would be happy to resurrect it once again in the new, new location :)

We also have an implementation based on this work in progress at https://github.com/openshift/machine-api-operator/tree/master/pkg/controller/machinehealthcheck which we are actively working on.

fabiand:
I know that some people are working on this, thus reopening it

/reopen

PR #1418, opened 2017-11-20
Add design details for Custom resource definition webhook validation.

brendandburns:
@kubernetes/sig-api-machinery-api-reviews @lavalamp

### DIFF
--- a/contributors/design-proposals/api-machinery/customresources-validation.md
+++ b/contributors/design-proposals/api-machinery/customresources-validation.md
@@ -20,8 +20,8 @@ Authors: @nikhita, @sttts, some ideas integrated from @xiao-zhou’s proposal<su
     3. [Client-Side Validation](#client-side-validation)
     4. [Comparison between server-side and client-side Validation](#comparison-between-server-side-and-client-side-Validation)
     5. [Existing Instances and changing the Schema](#existing-instances-and-changing-the-schema)
-    6. [Outlook to Status Sub-Resources](#outlook-to-status-sub-resources)
-    7. [Outlook Admission Webhook](#outlook-admission-webhook)
+    6. [Admission Webhook](#admission-webhook)
+    7. [Outlook to Status Sub-Resources](#outlook-to-status-sub-resources)
 5. [Implementation Plan](#implementation-plan)
 6. [Appendix](#appendix)
     1. [Expressiveness of JSON-Schema](#expressiveness-of-json-schema)
@@ -381,6 +381,35 @@ Note:
 
 2. For migration of CRDs with no validation to CRDs with validation, we can create a controller that will validate and annotate invalid CRs once the spec changes, so that the custom controller can choose to delete them (this is also essentially the status condition of the CRD). This can be achieved, but it is not part of the proposal.
 
+### Admission Webhook
+
+Custom resource definitions use the the same REST endpoint as built-in Kubernetes API objects. Consequently standard
+[dynamic web-hook admission controllers](https://kubernetes.io/docs/admin/extensible-admission-controllers/#external-admission-webhooks) 
+can be used for validation of custom resource definitions in addition to built-in resources.
+
+However, because the creation of admission controllers is a fairly high-privilege activity,
+and in many cases the creator of a `CustomResourceDefinition` is third-party extension code
+(e.g. an instance of the _operator pattern_) which should not have the ability to
+create arbitrary admission controllers, it is necessary to have an alternate solution for
+adding web-hook validation to custom resource definitions.
+
+The solution is that the `CustomResourceDefinition` itself has two arrays of webhooks in the
+definition itself. The arrays will be of type `admissionregistration.Webhook`.
+One array will contain non-mutating admission controllers for things like
+validation, and one will have mutating admission controllers for the purposes of defaulting.
+The custom resource definition controller (a piece of trusted code), will register
+these webhooks, _exclusively_ to be triggered for instances of the specific custom resource.
+This ensures that third-party extension code can not register admission controllers for
+arbitrary API objects.
+
+When the custom resource is deleted, the admission controller(s) is likewise deleted.
+
+For now, it is expected that the user will manage the authentication certificates used
+to authenticate and encrypt traffic from the api server to the webhook. Eventually,
+a better general mechanism may be created, but because the `CustomResourceDefinition`
+uses the existing `admissionregistration.Webhook` type, it will simply inherit these
+improvements.
+
 ### Outlook to Status Sub-Resources
 
 As another most-wanted feature, a Status sub-resource might be proposed and implemented for CRDs. The JSON-Schema proposed here might as well cover the Status field of a CR. For now this is not handled or validated in a particular way.
@@ -391,14 +420,6 @@ When the Status sub-resource exists some day, the /status endpoint will receive 
 {"type":"object", "properties":{"status": ..., "a": ..., "b": ...}}
 ```
 Then we can validate the status against the sub-schema easily. Hence, this proposal will be compatible with a later sub-resource extension.
-
-### Outlook Admission Webhook
-
-Apiextensions-apiserver uses the normal REST endpoint implementation and only customizes the registry and the codecs. The admission plugins are inherited from the kube-apiserver (when running inside of it via apiserver delegation) and therefore they are supposed to apply to CRs as well.
-
-It is [verified](https://github.com/kubernetes/kubernetes/pull/47252) that CRDs work well with initializers. It is also expected that webhook admission prototyped at https://github.com/kubernetes/kubernetes/pull/46316 will work with CRs out of the box. Hence, for more advanced validation webhook admission is an option as well (when it is merged).
-
-JSON-Schema based validation does not preclude implementation of other validation methods. Hence, advanced webhook-based validation can also be implemented in the future.
 
 ## Implementation Plan
 


lavalamp:
Yeah, that was my suggestion. Just let me repeat that I'd like to get some miles on the webhooks before adding an auto-configuration layer.

contributors/design-proposals/api-machinery/customresources-validation.md
```diff
@@ -381,6 +381,28 @@ Note:
 
 2. For migration of CRDs with no validation to CRDs with validation, we can create a controller that will validate and annotate invalid CRs once the spec changes, so that the custom controller can choose to delete them (this is also essentially the status condition of the CRD). This can be achieved, but it is not part of the proposal.
 
+### Admission Webhook
+
+Custom resource definitions use the normal REST endpoint implementation and only customizes the registry and the codecs consequentaly dynamic web-hook admission controllers can be used
```
lavalamp: Reword, hard to read-- this is missing a comma or something.

brendandburns: Done.

brendandburns:
@lavalamp comment addressed, please re-check.

Thanks!

contributors/design-proposals/api-machinery/customresources-validation.md
```diff
@@ -381,6 +381,29 @@ Note:
 
 2. For migration of CRDs with no validation to CRDs with validation, we can create a controller that will validate and annotate invalid CRs once the spec changes, so that the custom controller can choose to delete them (this is also essentially the status condition of the CRD). This can be achieved, but it is not part of the proposal.
 
+### Admission Webhook
+
+Custom resource definitions use the the same REST endpoint as built-in Kubernetes API objects. Consequentaly standard
```
nikhita: s/Consequentaly/Consequently

contributors/design-proposals/api-machinery/customresources-validation.md
```diff
@@ -381,6 +381,29 @@ Note:
...
+[dynamic web-hook admission controllers](https://kubernetes.io/docs/admin/extensible-admission-controllers/#external-admission-webhooks) 
+can be used for validation of custom resource definitions in addition to built-in resources.
+
+However, because the creation of admission controllers is a fairly high-privilege activity,
+and in many cases the creator of a `CustomResourceDefinition` is third-party extension code
+(e.g. an instance of the _operator pattern_) which should not have the ability to
+create arbitrary admission controllers, it is necessary to have an alternate solution for
+adding web-hook validation to custom resource definitions.
+
+The solution is that the `CustomResourceDefinition` itself has an array of webhooks in the
```
sttts: technically two arrays: mutating, non-mutating

contributors/design-proposals/api-machinery/customresources-validation.md
```diff
@@ -381,6 +381,29 @@ Note:
...
+
+Custom resource definitions use the the same REST endpoint as built-in Kubernetes API objects. Consequentaly standard
+[dynamic web-hook admission controllers](https://kubernetes.io/docs/admin/extensible-admission-controllers/#external-admission-webhooks) 
+can be used for validation of custom resource definitions in addition to built-in resources.
+
+However, because the creation of admission controllers is a fairly high-privilege activity,
+and in many cases the creator of a `CustomResourceDefinition` is third-party extension code
+(e.g. an instance of the _operator pattern_) which should not have the ability to
+create arbitrary admission controllers, it is necessary to have an alternate solution for
+adding web-hook validation to custom resource definitions.
```
sttts: This raises the question about how to create+distribute the necessary certificates. Currently, we don't have a mechanism to auto-create and inject 

- the webhook server key+certificate (trusted by the apiextensions-apiserver)
- and the kube-apiserver CA (trusted and verified by the webhook).

We need both. I don't want to see every CRD controller using this implementing its own certificate generation code, and even worse creating something insecure. 

In OpenShift we have a controller that's listening on a custom annotation on apiserver pods and then creates a secret on-demand. This gives a smooth and secure setup with zero config.

ncdc: We need to solve the certificate problem. I had a brief chat with @deads2k a few weeks ago about extracting the service signing certificate controller from OpenShift and making it generally available. Whether we do that or find another solution, I believe we need to do something. Otherwise, I feel it's too much of a burden to ask users to figure out certificate generation, signing, etc. to be able to do things like this.

brendandburns: I think @ncdc is saying this, but I think this is a general problem with dynamic webhooks, not specific to this proposal. We need to make it much easier for people to build dynamic webhooks in general.

contributors/design-proposals/api-machinery/customresources-validation.md
brendandburns: done.

contributors/design-proposals/api-machinery/customresources-validation.md
brendandburns: done.

contributors/design-proposals/api-machinery/customresources-validation.md
brendandburns: Added some text about this.

brendandburns:
Comments addressed, please take another look.

contributors/design-proposals/api-machinery/customresources-validation.md
```diff
@@ -381,6 +381,38 @@ Note:
...
+
+The solution is that the `CustomResourceDefinition` itself has two arrays of webhooks in the
+definition itself. One array will contain non-mutating admission controllers for things like
+validation, and one will have mutating admission controllers for the purposes of defaulting.
+The custom resource controller (a piece of trusted code), will register
+these webhooks, _exclusively_ to be triggered for instances of the specific custom resource.
+This ensures that third-party extension code can not register admission controllers for
+arbitrary API objects.
+
+To achieve this, the `CustomResourceDefinition` has two arrays of `admissionregistration.Webhook`
```
sttts: nit: not much new in this paragraph. just move the type `admissionregistration.Webhook` into the previous paragraph.

sttts:
One nit, otherwise lgtm.

There are a number of details like what of `admissionregistration.Webhook` can be set by the user and which not. But I think we don't need that here.

contributors/design-proposals/api-machinery/customresources-validation.md
brendandburns: done.

brendandburns:
@sttts comments addressed, please re-check.

Thanks!

contributors/design-proposals/api-machinery/customresources-validation.md
```diff
@@ -381,6 +381,35 @@ Note:
...
+### Admission Webhook
+
+Custom resource definitions use the the same REST endpoint as built-in Kubernetes API objects. Consequently standard
+[dynamic web-hook admission controllers](https://kubernetes.io/docs/admin/extensible-admission-controllers/#external-admission-webhooks) 
+can be used for validation of custom resource definitions in addition to built-in resources.
+
+However, because the creation of admission controllers is a fairly high-privilege activity,
+and in many cases the creator of a `CustomResourceDefinition` is third-party extension code
+(e.g. an instance of the _operator pattern_) which should not have the ability to
+create arbitrary admission controllers, it is necessary to have an alternate solution for
```
sttts: Seeing operators in the wild that install CRDs and RBAC rules, I question that there is a real difference to "creation of admission controllers" today. Of course, I totally agree that there should be. But for that we need a similar in-CRD mechanism for RBAC rules.

contributors/design-proposals/api-machinery/customresources-validation.md
```diff
@@ -381,6 +381,35 @@ Note:
...
+adding web-hook validation to custom resource definitions.
+
+The solution is that the `CustomResourceDefinition` itself has two arrays of webhooks in the
+definition itself. The arrays will be of type `admissionregistration.Webhook`.
+One array will contain non-mutating admission controllers for things like
+validation, and one will have mutating admission controllers for the purposes of defaulting.
+The custom resource controller (a piece of trusted code), will register
```
sttts: custom resource definition controller

contributors/design-proposals/api-machinery/customresources-validation.md
brendandburns: That is clearly a job for a future design PR :)

contributors/design-proposals/api-machinery/customresources-validation.md
brendandburns: done.

brendandburns:
Comments addressed, please re-check.

Thanks!

contributors/design-proposals/api-machinery/customresources-validation.md
sttts: sure, just thinking out loud here ;)

sttts:
@deads2k @liggitt ptal

deads2k:
Thinking about the overall flow of creating a CRD, I see this

 1.  Create a CRD and affect global discovery information, including shadowing existing resources and stealing shortnames.  This is currently reserved to cluster admins by default.
 2. Create an RBAC cluster role to allow users to use the new custom resource.  This requires passing an access escalation check.  Since the custom resource won't exist, that means it is reserved to cluster admins too.
 3. Create an AdmissionRegistration.  This currently requires cluster-admin as well.

Given that the first two items require cluster-admin access and the RBAC cluster role implies the power to grant access to any resources, I don't see an avenue for attack that this locks down.  If I can create a useful and accessible CRD, I can already read every pod and secret in the system (or grant access to do so).

lavalamp:
I think Brendan probably wants to make things easy, not lock them down.

deads2k:
> I think Brendan probably wants to make things easy, not lock them down.

I'm not sure where this comment is going.  In a cluster that cares about security, I think my point stands and if you grant permissions creating separate resources works fine.  In a cluster that doesn't care about security, you don't have a restriction on those things and creating separate resources works fine.  Are you suggesting there's a third case?  Could you describe it?

lavalamp:
Maybe I should let Brendan say what he wants, but I think it is something
along the lines of "in a cluster that doesn't care about security I only
have to submit the CRD to the system, no other configuration objects".

This is about making the CRD developer and installer's life easy.

(I think we would do well to adopt that goal in general, but I am
personally ambivalent about whether this particular thing goes very far in
service of this goal, esp. if a CRD author generally ends up running a
controller etc. anyway.)

brendandburns:
(sorry returning to this)

@deads2k regarding your point, there are two things:

1) Regardless of security, I think this design prevents accidents. I'd rather not have CRDs accidentally creating badly-written/broken admission controllers and breaking the entire cluster.

2) My hoped for flow for a secure cluster would look like:

Cluster-admin creates a stub CRD with names/short-names but no real details.
CRD-Operator is created with permissions to update their own CRD, but no other permissions.
CRD-Operator updates the CRD to add a web-hook for validation.

In this model, you can keep CRD creation with cluster-admin, but delegate the details of the CRD definition to the CRD operator.

I don't see anything in your concerns that should block merging this design proposal so that we can start building it.

Thanks

deads2k:
> I don't see anything in your concerns that should block merging this design proposal so that we can start building it, let me know if you feel differently and I'll address those concerns.

My comments and concerns address the core concept of this pull and I think that until we have agreement in concept, we should not merge this change.



> CRD-Operator is created with permissions to update their own CRD, but no other permissions.

The power to update the `CustomResourceDefinition` allows power over global discovery.  Allowing the CRD-operator the power to change the CRD is akin to allowing a process to modify its own configuration.

The CRD-operator should not get to decide the shape of the CRD that it uses and it also not be able to control the validation associated with it.  When the CRD is created, it should contain its validation rules.  This means the responsibility lies with the CRD creator who is also creating the roles.  It looks like a cluster-admin to me.

lavalamp:
Maybe we would do well to work on having some common goal / requirements around usability + security for CRDs. I don't think we can come to agreement on a specific proposal until we agree what we're even trying to accomplish.

brendandburns:
@deads2k given that that is up to the cluster admin to decide via RBAC, I don't see the issue here.

IF the cluster admin wants to force manual install for all CRDs that's their call.

That said, I think there are lots of clusters were people want things to behave like plugins and just dynamically work. But forcing CRD developers to learn about admission controllers and to configure them correctly just adds complexity and work to the job of the CRD developer.

Could you describe the disagreement you see in the core concept?

For me the core-concept(s) are:
   * We should have webhook validation for CRDs
   * We should have an controlled and automated method to make it easier for CRD developers/operators to activate those webhook validators without writing raw Admission Controllers

Nothing in this proposal is affecting the security model one way or the other, the cluster admin is free to set RBAC however they choose.

Please clarify, thanks!

deads2k:
> We should have an controlled and automated method to make it easier for CRD developers/operators to activate those webhook validators without writing raw Admission Controllers

This piece.  It isn't obvious to me what value this adds. You're duplicating an existing API in a different object to save setting three fields.  In the use-cases we have already the actor creating a useful CRD already has full permission on cluster, so there isn't much value in duplicating the API.

You're still going to be writing the admission webhook code either way.  You're just eliminating writing a rule from the resource manifest.  Also, writing a library to make the admission plugin webhook is fairly easy.  See https://github.com/openshift/generic-admission-server as an example in golang.

brendandburns:
The most important feature of this isn't the simplification (though I think that's important too) it's preventing someone from making mistakes and accidentally registering an Admission Controller that has broader scope than just the particular CRD.

Given that your objection is based on not seeing the value (and I believe there is the value outlined above) can we agree to disagree and move forward with this proposal? I don't think it hurts anything, and I personally believe there is significant value in terms of ease of use and prevention of mistakes.

liggitt:
> Given that your objection is based on not seeing the value (and I believe there is the value outlined above) can we agree to disagree and move forward with this proposal?

I'd prefer we not expand/nest APIs if we don't have to. I'd rather see API additions justified by enabling use cases that are not currently possible.

brendandburns:
I don't think this is nesting, this is like the fact that `PodTemplate` is embedded in multiple different objects. It is re-use of a common object. We're not nesting a complete `AdmissionController`.

One of the recurring benefits of CRDs (and the reason why they are _way_ more popular than aggregated API servers) is ease of use.

If you tell someone that if they want to validate their CRD they need to either learn Swagger or learn about admission controllers, you are erecting serious barriers to entry.

I think you need to focus on the usability for the average user, not the person who spends their time reading sig-api-machinery.

PR #1424, opened 2017-11-21
members must have sponsors that span organizations

derekwaynecarr:
This was a discussion topic with @calebamiles @smarterclayton and myself.

The previous guidance was that sponsors across organizations results in gold star, but this guidance firms up that requirement.  I am in favor of this change as a way of demonstrating integration into the broader community.

Opening for others input.

### DIFF
--- a/community-membership.md
+++ b/community-membership.md
@@ -64,7 +64,7 @@ Members are expected to remain active contributors to the community.
 - Sponsored by 2 reviewers. **Note the following requirements for sponsors**:
     - Sponsors must have close interactions with the prospective member - e.g. code/design/proposal review, coordinating on issues, etc.
     - Sponsors must be reviewers or approvers in at least 1 OWNERS file (in any repo in the Kubernetes GitHub organization)
-    - Not a requirement, but having sponsorship from a reviewer from another company is encouraged (you get a gold star).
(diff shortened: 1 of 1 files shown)


cblecker:
@kubernetes/sig-contributor-experience-misc-use-only-as-a-last-resort 
/hold
Holding for input.

cblecker:
My personal opinion:
I think this is a good change in principal, because if you're contributing to the community, people will notice. It's not a difficult bar.

What about the following scenario however:
Contributor A is a non-member working for Company A
Reviewer B and C are both reviewers working for Company B

Can both Reviewer B and C (from the same company as each other, but not the same company as Contributor A) sponsor Contributor A? I personally think this scenario is fine, as long as (per the other requirements) both Reviewers are involved directly with the area Contributor A is working on.

calebamiles:
@cblecker, I think that the general spirit should be that of the three people (prospect, sponsor a, sponsor b), there should be at least one person who does not share an employer with the other two but this seems to be a step in the right direction anyway.

cblecker:
@calebamiles I agree with that spirit completely 👍

calebamiles:
Great, @cblecker we'll clean this up in another pass.

/approve

PR #1434, opened 2017-11-25
Update serverfault link tag per updated site tag

spzala:
The link with google-kubernetes tag doesn't give any search result. Updating
the tag to kubernetes from google-kubernetes to get an expected output. The changes are required in just one file in the repo as modified herewith.


spzala:
@cblecker Thanks for the quick review!!

PR #1435, opened 2017-11-26
fix misspelling

stewart-yu:
Fix misspelling.


stp-ip:
Thanks for the fix. If you are going through more documents please try to bundle fixes together in one commit and PR.

/lgtm

PR #1440, opened 2017-11-28
Fix broken links

fatsheep9146:
Fix broken links

Signed-off-by: Ziqi Zhao <zhaoziqi@qiniu.com>

### DIFF
--- a/contributors/design-proposals/node/pod-resource-management.md
+++ b/contributors/design-proposals/node/pod-resource-management.md
@@ -16,7 +16,7 @@ to Kubernetes. It outlines the implementation and associated rollout plan.
 ## Introduction
 
 Kubernetes supports container level isolation by allowing users
-to specify [compute resource requirements](resources.md) via requests and
+to specify [compute resource requirements](/contributors/design-proposals/scheduling/resources.md) via requests and
 limits on individual containers.  The `kubelet` delegates creation of a
 cgroup sandbox for each container to its associated container runtime.
 


contributors/design-proposals/node/pod-resource-management.md
```diff
@@ -16,7 +16,7 @@ to Kubernetes. It outlines the implementation and associated rollout plan.
 ## Introduction
 
 Kubernetes supports container level isolation by allowing users
-to specify [compute resource requirements](resources.md) via requests and
+to specify [compute resource requirements](../scheduling/resources.md) via requests and
```
stp-ip: This breaks when moving the file. Why not use: `/contributors/design-proposals/scheduling/resources.md`?

fatsheep9146: You are right. I will change it as you said.

stp-ip: Great thanks.
Also curious sidenote:
Same repo links (as we have here) can just use the relative path.
Cross repo links can be done using something like: "git.k8s.io/community/contributors/design-proposals/scheduling/resources.md"
And issue links can be done using something like: "http://issue.k8s.io/1"

stp-ip:
Now only thing left is do rebase and squash both commits into one. Then we can merge \o/

fatsheep9146:
OK !

stp-ip:
Awesome. Thanks.

PR #1441, opened 2017-11-28
updating sig-windows meeting time

michmike:
updating sig-windows time to EST to avoid DST issues that occur with UTC timezones


cblecker:
@michmike You need to update sigs.yaml and run make. This will update the time in a number of places to ensure it's consistent
https://github.com/kubernetes/community/blob/d099f66a5da9745556b38ef64b321d72f2b9dce9/sigs.yaml#L1094-L1107

michmike:
we will create a new PR with the update to the sigs.yaml file

PR #1442, opened 2017-11-29
hpa: fix invalid links

sak0:
hpa-v2.md:  fix invalid links

Signed-off-by: CuiHaozhi <cuihaozhi@chinacloud.com.cn>


stp-ip:
@feiskyer The link is not absolute and breaks, when this document is moved. Don't we usually use links like: `/contributors/design-proposals/instrumentation/resource-metrics-api.md`

feiskyer:
@stp-ip It's true for kubernetes websites, but we should make markdown link work here.

stp-ip:
Absolute paths (not absolute links) would work here too and remove the breaking, when a document is moved.

PR #1445, opened 2017-11-29
Updating calendar directions so they work

mattfarina:
Updating after communicating with @sarahnovotny


sarahnovotny:
wooooo! thank you @mattfarina

PR #1451, opened 2017-11-30
Propose solution to make memory cgroup events effective.

dashpole:
This is a proposal containing different possible solutions for https://github.com/kubernetes/kubernetes/issues/51745.
My intention is to converge on a single solution, and incorporate that into the documentation.

This proposal assumes https://github.com/kubernetes/kubernetes/pull/56112 is accepted.  Work would be targeted for 1.10.

Please share thoughts/concerns, or propose additional solutions I have not thought of.

cc @vishh @derekwaynecarr @sjenning @kubernetes/sig-node-proposals

### DIFF
--- a/contributors/design-proposals/node/kubelet-eviction.md
+++ b/contributors/design-proposals/node/kubelet-eviction.md
@@ -191,6 +191,48 @@ signal.  If that signal is observed as being satisfied for longer than the
 specified period, the `kubelet` will initiate eviction to attempt to
 reclaim the resource that has met its eviction threshold.
 
+### Memory CGroup Notifications
+
+When the `kubelet` is started with `--experimental-kernel-memcg-notification=true`, 
+it will use cgroup events on the memory.usage_in_bytes file in order to trigger the eviction manager.
+With the addition of on-demand metrics, this permits the `kubelet` to trigger the eviction manager,
+collect metrics, and respond with evictions much quicker than using the sync loop alone.
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+However, a current issue with this is that the cgroup notifications trigger based on memory.usage_in_bytes,
+but the eviction manager determines memory pressure based on the working set, which is (memory.usage_in_bytes - memory.total_inactive_file).
+For example:
+```
+capacity = 1Gi
+--eviction-hard=memory.available<250Mi
+assume memory.total_inactive_file=10Mi
+```
+When the cgroup event is triggered, `memory.usage_in_bytes = 750Mi`.
+The eviction manager observes
+`working_set = memory.usage_in_bytes - memory.total_inactive_file = 740Mi`
+Signal: `memory.available = capacity - working_set = 260Mi`
+Therefore, no memory pressure.  This will occur as long as memory.total_inactive_file is non-zero.
+
+### Proposed solutions:
+1. Set the cgroup event at `threshold*fraction`.
+For example, if `--eviction-hard=memory.available<200Mi`, set the cgroup event at `100Mi`.
+This way, when the eviction manager is triggered, it will likely observe memory pressure.
+This is not guaranteed to always work, but should prevent OOMs in most cases.
+
+2. Use Usage instead of Working Set to determine memory pressure
+This would mean that the eviction manager and cgroup notifications use the same metric,
+and thus the response is ideal: the eviction manager is triggered exactly when memory pressure occurs.
+However, the eviction manager may often evict unneccessarily if there are large quantities of memory
+the kernel has not yet reclaimed.
+
+3. Increase the syncloop interval after the threshold is crossed
+For example, the eviction manager could start collecting observations every second instead of every
+10 seconds after the threshold is crossed.  This means that even though the cgroup event and eviction
+manager are not completely in-sync, the threshold can help the eviction manager to respond faster than
+it otherwise would.  After a short period, it would resume the standard interval of sync loop calls.
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/node/kubelet-eviction.md
```diff
@@ -191,6 +191,45 @@ signal.  If that signal is observed as being satisfied for longer than the
...
+Signal: `memory.available = capacity - working_set = 260Mi`
+Therefore, no memory pressure.  This will occur as long as memory.total_inactive_file is non-zero.
+
+### Proposed solutions:
+1. Set the cgroup event at `threshold*fraction`.
+For example, if `--eviction-hard=memory.available<200Mi`, set the cgroup event at `100Mi`.
+This way, when the eviction manager is triggered, it will likely observe memory pressure.
+This is not guaranteed to always work, but should prevent OOMs in most cases.
+
+2. Use Usage instead of Working Set to determine memory pressure
```
dchen1107: Upon a system OOM, the kernel will trigger cgroup reclaim, then global reclaim, etc. Most likely those inactive pages are going to be reclaimed. The problem with sys OOM is everything running on the node will be affected, including daemons. This is why we introduced eviction mgr at the userspace to prevent such event, and minimize and localize the impact, while avoiding the unnecessary eviction. The issue with this approach is eviction mgr might take action too aggressively, thus eventually it might break service availability SLO if we defined for the users.

dchen1107:
@dashpole I have concerns with option 2, but ok with either 1 or 3. Maybe we start to prototype them, then do some tests?

contributors/design-proposals/node/kubelet-eviction.md
dashpole: I agree with this.  I removed it as an option.

sjenning:
Thanks for documenting this!

I must say I'm not a fan of any of the solutions :-/  The two remaining options converge on "increase our polling rate for a finite time after the threshold trigger occurs".  I'm not saying I have a better one though so... yeah.

With option 1 if we trigger at 1/2 the value, and the working set is still not enough to trigger eviction, then we've used our one shot at a quicker response if the memory load stays below the threshold.

With option 3 there is no guarantee that the working set will exceed the threshold during the time of increased polling and we could again not get a faster response.

Ideally, we'd like to get an edge trigger for `(memory.usage_in_bytes - memory.total_inactive_file)`, right?  I wondering if that can't be achieved by setting a trigger on both files and adjusting the notification value periodically...

dashpole:
> With option 1 if we trigger at 1/2 the value, and the working set is still not enough to trigger eviction, then we've used our one shot at a quicker response if the memory load stays below the threshold.

I have actually done some experimentation, and have found that option 1 seems to work pretty well, but my tests probably aren't very similar to actual workloads.  If we have a large `total_inactive_file` value, doesn't that meant that when the kernel reclaims memory, we should return below the threshold (and get a second shot)?  I guess this would only work for reasonably small memory thresholds, but memcg notifications are most useful then anyways.

> Ideally, we'd like to get an edge trigger for (memory.usage_in_bytes - memory.total_inactive_file), right? I wondering if that can't be achieved by setting a trigger on both files and adjusting the notification value periodically...

If you want to prototype that and test it, that could be an interesting idea.  I can add it to this proposal.

dashpole:
added item proposed in https://github.com/kubernetes/community/pull/1451#issuecomment-354904568 as option 4

dashpole:
I prototyped option 1 Here: https://github.com/kubernetes/kubernetes/compare/master...dashpole:fix_memcg_option1
The first commit is https://github.com/kubernetes/kubernetes/pull/56112, and the second is https://github.com/kubernetes/kubernetes/pull/57849.
The third commit implements option 1, and updates the PriorityMemoryEvictionOrdering node e2e test to test it.

dashpole:
I prototyped option 4 here: https://github.com/kubernetes/kubernetes/compare/master...dashpole:fix_memcg_option4
The first commit is kubernetes/kubernetes#56112, and the second is kubernetes/kubernetes#57849.
The third commit implements option 4, and updates tests to test it.

dashpole:
After playing around with both a bit, I think option 4 seems to work slightly better, as it is better able to predict when the threshold is crossed.
As long as adding and removing cgroup event control every 10 seconds is viable, I think it is the best option to start with.

derekwaynecarr:
@dashpole -- what workload are you running when testing?

derekwaynecarr:
option 4 is probably the best option of what is presented.

dashpole:
@derekwaynecarr I am just running the memory eviction tests, using aggressive (10 Mb / millisecond, 10,000x faster than current test) rates of memory consumption increases.
I realize this isn't super realistic, but I am hoping we can iterate on it in the future.

dashpole:
@derekwaynecarr I ran some more experiments by [adding in a couple best effort JVM pods to the eviction test](https://github.com/kubernetes/kubernetes/compare/master...dashpole:fix_memcg_option4#diff-590eb0d04c6b4987c1addfa5d2c4424aR746).  I found that Option 4 reliably triggers the threshold when we are under memory pressure.  I think the test may be cpu constrained, because I have seen flakes where the runtime fails to evict the pod within our 10 sec timeout, and we OOM shortly after.  But triggering the threshold seems to work reliably...

It turns out that just setting the threshold to `capacity - eviction_hard + inactive_file` works the best, since inactive_file is decreasing as available memory decreases on the node.  Previously, I had thought we should subtract an additional 10% of the eviction threshold.  This means the memcg notification always triggers slightly after we have memory pressure, which is exactly what we want.

dashpole:
Making the test slightly less aggressive (2 Mi /second instead of 20 Mi/second) makes the test pass consistently.  It seems to generally take 10-20ms between when the threshold is crossed, and when the container is killed.

dashpole:
I updated the proposal to choose option 4.

PR #1453, opened 2017-11-30
Add cblecker and castrojo to community approvers

cblecker:
This alpha sorts the OWNERS file and adds @castrojo and myself as approvers (Jorge helping community manage, and myself having been shepherding stuff in this repo). This would be for fall-through stuff, as there are more-specific owners files in all the sig-owned directories.

I think with the document sorting being mostly done, the bots and CI working well, and the OWNERS_ALIASES file set up with all the sig leads, we might be able to turn on the new prow approval plugin in kubernetes/community.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,18 +1,21 @@
 reviewers:
-  - sarahnovotny
-  - idvoretskyi
   - calebamiles
+  - castrojo
+  - cblecker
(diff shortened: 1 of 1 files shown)


cblecker:
/cc @castrojo 
I should probably CC Jorge if I'm volunteering him haha

PR #1455, opened 2017-12-01
Adding the rough draft of the cloud-provider KEP document.

cheftako:
Adding the rough draft of the cloud-provider KEP document.

### DIFF
--- a/wg-cloud-provider/draft-20171128-controller-manager.md
+++ b/wg-cloud-provider/draft-20171128-controller-manager.md
@@ -1,4 +1,4 @@
----
+```
 kep-number: draft-20171127
 title: Cloud Provider Controller Manager
 authors:
@@ -21,6 +21,7 @@ status: draft
 see-also:
 replaces:
   - contributors/design-proposals/cloud-provider/cloud-provider-refactoring.md
+```
 
 # Remove Cloud Provider Code From Kubernetes Core
 
--- a/wg-cloud-provider/draft-20171128-controller-manager.md
+++ b/wg-cloud-provider/draft-20171128-controller-manager.md
@@ -1,0 +1,294 @@
+---
+kep-number: draft-20171127
+title: Cloud Provider Controller Manager
+authors:
+  - "@cheftako"
+  - "@calebamiles"
+owning-sig: sig-apimachinery
+participating-sigs:
+  - sig-storage
+  - sig-apps
+  - sig-network
+reviewers:
+  - "@wlan0"
+  - "@calebamiles"
+approvers:
+  - TBD
+editor: TBD
+creation-date: 2017-11-27
+last-updated: 2017-11-27
+status: draft
+see-also:
+replaces:
+  - contributors/design-proposals/cloud-provider/cloud-provider-refactoring.md
+
+# Remove Cloud Provider Code From Kubernetes Core
+
+## Table of Contents
+
+   * [Remove Cloud Provider Code From Kubernetes Core](#remove-cloud-provider-code-from-kubernetes-core)
+      * [Table of Contents](#table-of-contents)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Intermediary Goals](#intermediary-goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [Controller Manager Changes](#controller-manager-changes)
+         * [Kubelet Changes](#kubelet-changes)
+         * [API Server Changes](#api-server-changes)
+         * [Volume Managent Changes](#volume-managent-changes)
+         * [Deployment Changes](#deployment-changes)
+         * [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
+         * [Security Considerations](#security-considerations)
+      * [Graduation Criteria](#graduation-criteria)
+      * [Implementation History](#implementation-history)
+      * [Drawbacks [optional]](#drawbacks-optional)
+      * [Alternatives [optional]](#alternatives-optional)
+
+## Summary
+
+We want to remove any cloud provider specific logic from the kubernetes/kubernetes repo. We want to restructure the code
+to make is easy for any cloud provider to extend the kubernetes core in a consistent manner for their cloud. New cloud 
+providers should look at the [Creating a Custom Cluster from Scratch](https://kubernetes.io/docs/getting-started-guides/scratch/#cloud-provider)
+and the [cloud provider interface](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/cloud.go#L31) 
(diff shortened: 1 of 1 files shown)


cblecker:
@cheftako It looks like you need to reset the author on your commit

wg-cloud-provider/draft-20171128-controller-manager.md
```diff
@@ -0,0 +1,283 @@
...
+For the cloud providers who are in repo, moving out would allow them to more quickly iterate on their solution and 
+decouple cloud provider fixes from open source releases. Moving the cloud provider code out of the open source 
+processes means that these processes do not need to load/run unnecessary code for the environment they are in. 
+We would like to abstract a core controller manager library so help standardize the behavior of the cloud
+controller managers produced by each cloud provider. We would like to minimize the number and scope of controllers 
+running in the cloud controller manager so as to minimize the surface area for per cloud provider deviation.
+
+### Goals
+
+Get to a point where we do not need to load the cloud interface for any of kubernetes core processes.
```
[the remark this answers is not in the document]
cheftako: I think implied is the key. I can imagine having a cloud controller manager which hosts all the controller loops but where the kubelet still loads the cloud provider. I would rather be explicit about this goal.

wg-cloud-provider/draft-20171128-controller-manager.md
```diff
@@ -0,0 +1,283 @@
...
+whose function is specific to cloud provider functionality. The kube controller manager would then be responsible
+for running all controllers whose function was not related to a cloud provider.
+
+In order to create a 100% cloud independent controller manager, the controller-manager will be split into multiple binaries.
+
+1. Cloud dependent controller-manager binaries
+2. Cloud independent controller-manager binaries - This is the existing `kube-controller-manager` that is being shipped 
+with kubernetes releases.
+
+The cloud dependent binaries will run those loops that rely on cloudprovider as a kubernetes system service.
```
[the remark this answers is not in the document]
cheftako: Honestly that was stolen from https://github.com/kubernetes/community/blob/master/contributors/design-proposals/cloud-provider/cloud-provider-refactoring.md. Updating to what I believe the intent was.

wg-cloud-provider/draft-20171128-controller-manager.md
```diff
@@ -0,0 +1,294 @@
...
+  - sig-apps
+  - sig-network
+reviewers:
+  - "@wlan0"
+  - "@calebamiles"
+approvers:
+  - TBD
+editor: TBD
+creation-date: 2017-11-27
+last-updated: 2017-11-27
```
lavalamp: meta note: Seems like a bad idea to keep these fields inside a doc in a source control system, they are tracked by git.

calebamiles: I made exactly that same point @lavalamp...but was shouted down

wg-cloud-provider/draft-20171128-controller-manager.md
```diff
@@ -0,0 +1,294 @@
+---
```
calebamiles: you might still want to enclose the YAML in back ticks so it renders a bit nicer

wlan0:
@karataliu @lpabon I'm moving the conversation about CSI and flex for CCM to this thread so that it is accessible to the larger community more easily.

We brought this up at the cloud provider working group meetup today. @karataliu - To answer your question, after conversations with @sig-storage, we found that flex will not add any more new features and CSI is the way to more forward.

In the meantime, before CSI becomes beta and more usable, @cheftako has added a temporary measure to use existing volume plugins without needing CSI. @cheftako Could you please link your PR for the `--external-cloud-volume-plugin` flag 

Additionaly, @ublubu and @yastij from our group are tasked with adding CSI plugins and accelerating CSI to beta. 

@karataliu - help me understand if you still have any concerns about using CSI instead of flex for our efforts.

cheftako:
--external-cloud-volume-plugin went in, in PR https://github.com/kubernetes/kubernetes/pull/52371.

karataliu:
Thanks @wlan0 @cheftako, nice elaboration.

Let me clarify some remaining thoughts as follows:

Currently the cloud provider (with volume support) implementations contain two parts:

1. Code implementing general cloud interface, which is to be called by various control loops
https://github.com/kubernetes/kubernetes/blob/2ae99cf146e5bb2310edcc109e887408201d274b/pkg/cloudprovider/cloud.go 

    e.g. `GetZone`, `EnsureLoadBalancer`,...


2. Code not implementing the cloud interface, whereas, called by corresponding cloud volume provider.

    e.g.
    https://github.com/kubernetes/kubernetes/blob/21a15c56732a03b064844aa041dcf19b65a41a75/pkg/cloudprovider/providers/aws/aws.go#L2077
    https://github.com/kubernetes/kubernetes/blob/2ae99cf146e5bb2310edcc109e887408201d274b/pkg/cloudprovider/providers/azure/azure_managedDiskController.go#L39
    https://github.com/kubernetes/kubernetes/blob/2ae99cf146e5bb2310edcc109e887408201d274b/pkg/cloudprovider/providers/gce/gce_disks.go#L60

    Logically the code belongs to the individual cloud volume providers, but they lay here since `pkg/cloudprovider` now serves as a centralized place for interacting with clouds.


Here are some questions to be discussed:
1. Shall we separate the two piece of code first?
`cloud-control-manager` will be more interested in part#1 code, thus if the volume related logic will be provided as external, it is useless to compile such part#2 into `cloud-controller-manager` binary.
The pain point is the 2 parts of code shares cloud auth logic to some extent, but that's also one thing to be solved by csi providers.

2. Will there be some period that CSI hasn't been ready, but we want to move out cloud-providers?
`--external-cloud-volume-plugin` would be useful in this case, and the story may go like:
    a. cloud-provider not moved (current)
    - Use unifrom cloud-controller-manager
    - Use '--external-cloud-volume-plugin' for cloud volumes.

    b. cloud-provider(part#1) moved, CSI not ready
    - Use cloud-specific cloud-controller-manager
    - Use '--external-cloud-volume-plugin' for cloud volumes.

    c. cloud-provider(part#1) moved, [CSI adapters](https://github.com/kubernetes/kubernetes/issues/57715) ready
    - Use cloud-specific cloud-controller-manager
    - Deperate '--external-cloud-volume-plugin', turn to CSI adapters

    d. cloud-provider(part#1) moved, individual CSI implementations ready
    - Use cloud-specific cloud-controller-manager
    - Turn to offical CSI implementations


3. Where shall the cloud provider (part#1) code be moved to?
Some previous discussions here:
https://github.com/kubernetes/features/issues/88#issuecomment-326819487 
https://github.com/kubernetes/community/blob/master/wg-cloud-provider/draft-20171128-controller-manager.md 

    Some cloud vendors have already hosted their out-of-tree repo. Do we propose a centralized location, or use vendor maintained ones?

PR #1458, opened 2017-12-01
Adds the design proposal for self-hosting authz webhook

filmil:
This proposal implements feature request:
  https://github.com/kubernetes/features/issues/516

Please refer to the feature request for administrative details.  Reviews appreciated.

### DIFF
--- a/contributors/design-proposals/auth/self-hosting-authorization-webhook.md
+++ b/contributors/design-proposals/auth/self-hosting-authorization-webhook.md
@@ -1,0 +1,416 @@
+# Self-hosting authorization webhooks
+
+Status: Pending
+
+Version: **Alpha** | Beta | GA
+
+Implementation Owner: @filmil
+
+## Motivation
+
+Make it possible to have the authorization (hereafter authz for short) webhook
+run as a pod in the cluster being authorized. This addresses part of the
+concerns raised in [kubernetes/kubernetes#52511][18].
+
+See:
+  - Feature request: [kubernetes/features#516][15]
+  - Earlier pull requests:
+    - [kubernetes/kubernetes#54733][16]
+    - [kubernetes/kubernetes#54163][17]
+
+
+This requires a change in the way the `kube-apiserver` webhook configuration is
+specified (a new configuration format), and a special initialization of the
+webhook module in `kube-apiserver` (in contrast to any other apiservers out
+there) as [outlined in an earlier discussion on this topic][9].
+
+As of this writing, the authorization webhook in Kubernetes is configured
+through the `kube-apiserver` flag `--authorization-webhook-config-file` which is
+a [kubeconfig][3]-formatted [configuration file][2].  An example taken from the
+documentation:
+
+```yaml
+# clusters refers to the remote service.
+clusters:
+  - name: name-of-remote-authz-service
+    cluster:
+      # CA for verifying the remote service.
+      certificate-authority: /path/to/ca.pem
+      # URL of remote service to query. Must use 'https'. May not include parameters.
+      server: https://authz.example.com/authorize
+
+# users refers to the API Server's webhook configuration.
+users:
+  - name: name-of-api-server
+    user:
+      client-certificate: /path/to/cert.pem # cert for the webhook plugin to use
+      client-key: /path/to/key.pem          # key matching the cert
+
+# kubeconfig files require a context. Provide one for the API Server.
+current-context: webhook
+contexts:
+- context:
+    cluster: name-of-remote-authz-service
+    user: name-of-api-server
+  name: webhook
+```
+
+The kubeconfig file format does not easily admit reference to an endpoint that
+is hosted within the cluster.  Earlier proposals to extend the syntax of the
+`clusters[*].cluster.server` field to admit a custom dialer that calls a service
+endpoint [were rejected][4] (also see [another rationale][5]).  The [suggested
+viable alternative][6] is to provide a new webhook file format that references
+a service explicitly, and build on top of that.
+
+One further concern to address is that this proposed change is only [relevant
+for the `kube-apiserver`][7], and not to any of the generic apiservers.  This
+means that the proposed changes will be set up such that only the
+`kube-apiserver` can make use of it.  This approach has [garnered support][8]
+from the community in the earlier proposal.
+
+## Proposal
+
+* Define a new configuration format, supported only by `kube-apiserver` that
+  admits a webhook endpoint within the cluster to be specified.
+
+  This new configuration file is still specified using the same flag as for the
+  original webhook configuration, `--auhtorization-webhook-config-file`.
+
+* Instantiate a custom subject access review client based on this configuration
+  that has special behavior as follows:
+
+  - If the new configuration has been specified, a custom dialer is supplied
+    which resolves a service endpoint and uses that service endpoint to dial
+    into when making a call.
+
+## User Experience
+
+### Use Cases
+
+1. As a cluster administrator, I would like to install Kubernetes-based products
+   that have self-contained deployments in a cluster and include authorization.
+
+   Kubernetes products typically come with an install script, say based on Helm
+   or `kubeadm`.  These allow custom deployments, including reconfiguring
+   `kube-apiserver`.  When such a product requires an authorizer, the authorizer
+   is added as a system component with a pod-based deployment in a cluster and
+   the apiserver is instructed to consult the endpoint based on that deployment
+   for authorization decisions.
+
+   The cluster administrator experience changes in that they are able to install
+   a wider range of Kubernetes-based products that include this feature,
+   compared to when this is not the case.
+
+## Security
+
+The authorizer webhook module today is an extension mechanism that wields
+considerable power.  A new authorizer has the opportunity to grant access to any
+resource and override decisions from any other authorizers.  So we look into
+some detail of the implications of a self-hosted authorizer.
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/auth/self-hosting-authorization-webhook.md
```diff
@@ -0,0 +1,369 @@
...
+### Configuration format
+
+Let us define a new configuration file format.  This is based on the [admission
+configuration file format][12] as well as the proposal from @frankfarzan for an
+analogous [configuration format for authz][13].
+
+An example configuration file for authz looks like this:
+
+```yaml
+apiVersion: authorizationconfig.k8s.io/v1alpha1
```
ericchiang: I know we've had concerns in the past about API groups with only one type. In the future, should this group hold other webhook configs? e.g. audit or authentication?

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
```diff
@@ -0,0 +1,369 @@
...
+kind: ExternalAuthorizationHookConfiguration
+metadata:
+  name: example-config
+externalAuthorizationHooks:
+- name: webhook-name
+  clientConfig:
+    serverCaFile: ... # the path to the CA file for the webhook server
```
ericchiang: nit: we should probably be more consistent about what we name these fields. What other config uses `serverCaFile`?

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
```diff
@@ -0,0 +1,369 @@
...
+
+Therefore, it seems reasonable to confine the behavior changes to
+`kube-apiserver` alone.
+
+### Unifying code paths for webhooks
+
+Versus: this work is out of scope.
+
+Our proposed solution makes no change to the current state of the webhook
+libraries.  Unifying code paths for webhooks is a worthwhile goal, but out of
```
ericchiang: If we're going to be adding features to the authorization webhook, it seems odd to declare "how other webhooks will us this feature it's out of scope." I'd like to see it at least considered, even if we don't implemented.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: My concern here was that a field named `serverCa`, or `server-certificate` isn't quite clearly saying whether what follows is a file path name, or the actual quoted content of the field.

`serverCaFile` is very clear about what the contents are expected to be.  Is this something that's not desirable?  Should the name err towards consistency regardless?  Do we support/endorse the use of yaml-style inline quoting?

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: What do you recommend to do here instead?

Not sure I can decide either way, not knowing the history of such concerns.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: Other webhooks won't use this feature per this proposal.  The proposal (at least IMHO) doesn't prevent reuse.

I can expand a bit on that viewpoint in the text, but anything I'd say would be along the lines of "this is not a concern in the context of this proposal".

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
ericchiang: Sorry, I mean we have a history of not being constant in our naming of TLS options https://github.com/kubernetes/kubernetes/issues/54665

Admission webhooks `caBundle` for example: https://kubernetes.io/docs/admin/extensible-admission-controllers/#configure-webhook-admission-controller-on-the-fly

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
ericchiang: This was referencing https://github.com/kubernetes/kubernetes/pull/55812#discussion_r151370163

Maybe this could be included in the existing `apiserver.config.k8s.io` group?

https://github.com/kubernetes/kubernetes/pull/55812/commits/229c4307b173842633327a4fea3f16cde73cf196#diff-77d495a77d16f1f8c941606e2e5dee42R30

filmil: My understanding from previous comments is that this addition would be a kube-apiserver feature only, and as such not suitable for generic apiservers.

See for example discussion around this comment: https://github.com/kubernetes/kubernetes/issues/52511#issuecomment-333224542

Would your proposal not mean that we're mixing in these specific concerns into all other apiservers?

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: Also, forgot to mention: Thanks a million for jumping to review this pull request so quickly!

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
ericchiang: Didn't realized deads2k had already commented on this. It's fine as is then. Thanks for the feedback.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: Yes, duly noted.  Do you have a specific recommendation, just so I don't need to go invent an approach if you can already see one?

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: Sure thing!  It's on me as a proposer to dig up prior art.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
ericchiang: Digging around the APIs there's not much precedence. I agree this is more explicit than `caBundle`. Let's just leave it as is. Sorry for the noise.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
ericchiang: I seflishly would like to see this usable by the authentication webhook too. Looking at https://github.com/kubernetes/kubernetes/pull/54733 it doesn't seem like that would be too difficult, the language here just worries me.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: SGTM. I updated the comments to be way more specific as to the meaning of the fields.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: Please take another look.  I tried to clarify.

filmil:
/assign @deads2k 

Per advice from the k8s-ci-robot.

mikedanese:
A bit stale but potentially related: https://github.com/kubernetes/kubernetes/issues/30395

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
```diff
@@ -0,0 +1,394 @@
...
+viable alternative][6] is to provide a new webhook file format that references
+a service explicitly, and build on top of that.
+
+One further concern to address is that this proposed change is only [relevant
+for the `kube-apiserver`][7], and not to any of the generic apiservers.  This
+means that the proposed changes will be set up such that only the
+`kube-apiserver` can make use of it.  This approach has [garnered support][8]
+from the community in the earlier proposal.
+
+## Proposal
```
mikedanese: Please add a section on security implications of running an authorizer on the cluster.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
mikedanese: Per https://github.com/kubernetes/community/blob/master/contributors/devel/component-config-conventions.md, component config apigroups should have the prefix: `.config.k8s.io`. I also don't see why this specific config format will not be useful for other webhooks (e.g. authenticator).

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
```diff
@@ -0,0 +1,394 @@
...
+The go data definition matching the file above is as follows.  This captures
+both the service reference and a URL.
+
+```go
+// Required well-known import statements elided for brevity.
+
+// Top-level
+type ExternalAuthorizationHookConfiguration struct {
+  meta.TypeMeta
+  meta.ObjectMeta
```
mikedanese: Config objects should not need ObjectMeta.

filmil: Done

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: Changed the prefix to conform to that convention if it helps.  I'm not generalizing since this is not the intention, and previous feedback has been that this should remain a kubeapiserver feature only.

contributors/design-proposals/auth/self-hosting-authorization-webhook.md
filmil: Added a section on security.

PR #1460, opened 2017-12-04
proposal: bound service account tokens

mikedanese:
Proposal to improve the utility of service account tokens by supporting audience, time and key binding.

@kubernetes/sig-auth-api-reviews

### DIFF
--- a/contributors/design-proposals/auth/bound-service-account-tokens.md
+++ b/contributors/design-proposals/auth/bound-service-account-tokens.md
@@ -1,0 +1,239 @@
+# Bound Service Account Tokens
+
+Author: @mikedanese
+
+# Objective
+
+This document describes an API that would allow workloads running on Kubernetes
+to request JSON Web Tokens that are audience, time and eventually key bound.
+
+# Background
+
+Kubernetes already provisions JWTs to workloads. This functionality is on by
+default and thus widely deployed. The current workload JWT system has serious
+issues:
+
+1.  Security: JWTs are not audience bound. Any recipient of a JWT can masquerade
+    as the presenter to anyone else.
+1.  Security: The current model of storing the service account token in a Secret
+    and delivering it to nodes results in a broad attack surface for the
+    Kubernetes control plane when powerful components are run - giving a service
+    account a permission means that any component that can see that service
+    account's secrets is at least as powerful as the component.
+1.  Security: JWTs are not time bound. A JWT compromised via 1 or 2, is valid
+    for as long as the service account exists. This may be mitigated with
+    service account signing key rotation but is not supported by client-go and
+    not automated by the control plane and thus is not widely deployed.
+1.  Scalability: JWTs require a Kubernetes secret per service account.
+
+# Proposal
+
+Infrastructure to support on demand token requests will be implemented in the
+core apiserver. Once this API exists, a client of the apiserver will request an
+attenuated token for it's own use. The API will enforce required attenuations,
+e.g. audience and time binding.
+
+## Token attenuations
+
+### Audience binding
+
+Tokens issued from this API will be audience bound. Audience of requested tokens
+will be bound by the `aud` claim. The `aud` claim is an array of strings
+(usually URLs) that correspond to the intended audience of the token. A
+recipient of a token is responsible for verifying that it identifies as one of
+the values in the audience claim, and should otherwise reject the token. The
+TokenReview API will support this validation.
+
+### Time binding
+
+Tokens issued from this API will be time bound. Time validity of these tokens
+will be claimed in the following fields:
+
+* `exp`: expiration time
+* `nbf`: not before
+* `iat`: issued at
+
+A recipient of a token should verify that the token is valid at the time that
+the token is presented, and should otherwise reject the token. The TokenReview
+API will support this validation.
+
+Cluster administrators will be able to configure the maximum validity duration
+for expiring tokens. During the migration off of the old service account tokens,
+clients of this API may request tokens that are valid for many years. These
+tokens will be drop in replacements for the current service account tokens.
+
+### Object binding
+
+Tokens issued from this API may be bound to a Kubernetes object in the same
+namespace as the service account. The name, group, version, kind and uid of the
+object will be embedded as claims in the issued token. A token bound to an
+object will only be valid for as long as that object exists.
+
+Only a subset of object kinds will support object binding. Initially the only
+kinds that will be supported are:
+
+* v1/Pod
+* v1/Secret
+
+The TokenRequest API will validate this binding.
+
+## API Changes
+
+### Add `tokenrequests.authentication.k8s.io`
+
+We will add an imperative API (a la TokenReview) to the
+`authentication.k8s.io` API group:
+
+```golang
+type TokenRequest struct {
+  Spec   TokenRequestSpec
+  Status TokenRequestStatus
+}
+
+type TokenRequestSpec struct {
+  // Audiences are the intendend audiences of the token. A token issued
+  // for multiple audiences may be used to authenticate against any of
+  // the audiences listed. This implies a high degree of trust between
+  // the target audiences.
+  Audiences []string
+
+  // ValidityDuration is the requested duration of validity of the request. The
+  // token issuer may return a token with a different validity duration so a
+  // client needs to check the 'expiration' field in a response.
+  ValidityDuration metav1.Duration
+
+  // BoundObjectRef is a reference to an object that the token will be bound to.
+  // The token will only be valid for as long as the bound object exists.
+  BoundObjectRef *BoundObjectReference
+}
+
+type BoundObjectReference struct {
+  // Kind of the referent. Valid kinds are 'Pod' and 'Secret'.
+  Kind string
+  // API version of the referent.
+  APIVersion string
+
+  // Name of the referent.
+  Name string
+  // UID of the referent.
+  UID types.UID
+}
+
+type TokenRequestStatus struct {
+  // Token is the token data
+  Token string
+
+  // Expiration is the time of expiration of the returned token. Empty means the
+  // token does not expire.
+  Expiration metav1.Time
+}
+
+```
+
+This API will be exposed as a subresource under a serviceacccount object. A
+requestor for a token for a specific service account will `POST` a
+`TokenRequest` to the `/token` subresource of that service account object.
+
+### Modify `tokenreviews.authentication.k8s.io`
+
+The TokenReview API will be extended to support passing an additional audience
+field which the service account authenticator will validate.
+
+```golang
+type TokenReviewSpec struct {
+  // Token is the opaque bearer token.
+  Token string
+  // Audiences is the identifier that the client identifies as.
+  Audiences []string
+}
+```
+
+### Example Flow
+
+```
+> POST /apis/v1/namespaces/default/serviceaccounts/default/token
+> {
+>   "kind": "TokenRequest",
+>   "apiVersion": "authentication.k8s.io/v1",
+>   "spec": {
+>     "audience": [
+>       "https://kubernetes.default.svc"
+>     ],
+>     "validityDuration": "99999h",
+>     "boundObjectRef": {
+>       "kind": "Pod",
+>       "apiVersion": "v1",
+>       "name": "pod-foo-346acf"
+>     }
+>   }
+> }
+{
+  "kind": "TokenRequest",
+  "apiVersion": "authentication.k8s.io/v1",
+  "spec": {
+    "audience": [
+      "https://kubernetes.default.svc"
+    ],
+    "validityDuration": "99999h",
+    "boundObjectRef": {
+      "kind": "Pod",
+      "apiVersion": "v1",
+      "name": "pod-foo-346acf"
+    }
+  },
+  "status": {
+    "token":
+    "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJz[payload omitted].EkN-[signature omitted]",
+    "expiration": "Jan 24 16:36:00 PST 3018"
+  }
+}
+```
+
+The token payload will be:
+
+```
+{
+  "iss": "https://example.com/some/path",
+  "sub": "system:serviceaccount:default:default,
+  "aud": [
+    "https://kubernetes.default.svc"
+  ],
+  "exp": 24412841114,
+  "iat": 1516841043,
+  "nbf": 1516841043,
+  "kubernetes.io": {
+    "serviceAccountUID": "c0c98eab-0168-11e8-92e5-42010af00002",
+    "boundObjectRef": {
+      "kind": "Pod",
+      "apiVersion": "v1",
+      "uid": "a4bb8aa4-0168-11e8-92e5-42010af00002",
+      "name": "pod-foo-346acf"
+    }
+  }
+}
+```
+
+## Service Account Authenticator Modification
+
+The service account token authenticator will be extended to support validation
+of time and audience binding claims.
+
+## ACLs for TokenRequest
+
+The NodeAuthorizer will allow the kubelet to use its credentials to request a
+service account token on behalf of pods running on that node. The
+NodeRestriction admission controller will require that these tokens are pod
+bound.
+
+## Footnotes
+
+* New apiserver flags:
+  * --service-account-issuer: Identifier of the issuer.
+  * --service-account-signing-key: Path to issuer private key used for signing.
+  * --service-account-api-audience: Identifier of the API. Used to validate
+    tokens authenticating to the Kubernetes API.
+* The Kubernetes apiserver will identify itself as `kubernetes.default.svc`
+  which is the DNS name of the Kubernetes apiserver. When no audience is
+  requested, the audience is defaulted the audience is defaulted to an array
+  containing only this identifier.
+


mikedanese:
@smarterclayton

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+`authentication.k8s.io` API group:
+
+```golang
+type TokenRequest struct {
+  Spec   TokenRequestSpec
+  Status TokenRequestStatus
+}
+
+type TokenRequestSpec struct {
+  // Audience is the intendend audience of the token.
```
ericchiang: Better to be explicit about what a token with multiple audiences means? E.g. 
```
A token issued for multiple audiences may be used to authenticate against any of the audiences listed. This
implies a high degree of trust between the target audiences.
```

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+
+  // KeyConfirmation contains information required by a recipient of a token
+  // to verify possession of a key.
+  //
+  // see key binding
+  KeyConfirmation KeyConfirmation
+}
+
+type TokenRequestStatus struct {
+  Token string
```
ericchiang: nit: add an error field?

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+      "use": "sig",
+      "kid": "68241231bbf0df8f9123d018cf9e601e2aa3673a",
+      "n": "rHozcxeim9flTWQxqC1ObpGP0EjpkUHVHpHNX8WGHHnMcVi63_9PaHn2cJeFuPF9qkI1dMPXeoX0m33N0tgM9-KOSmTg1oGbyJGoUYMFI-A7tdxoobb91LGjeNWJC0la0gLOGPcQ6zQLEU5RGftCZT0wElxMuwEH7FZoVBn5i8Ddvc2ADd4bFW0f_FckwFYN1rIU1uLf6coku_1xBfae3b_JiBq38QOGXPdPgxfPzmJEvIz_LB2WOIcwhl97DY32BQU7l_lNLYz6wMg9HeCKolypPIFEGNxLj1TcuOhwP5-BnSja-PvrB-1FN1JyzlL3nh__uJv8SoKPn0CoBPueWw",
+      "e": "AQAB"
+    }
+  ]
+}
+```
+
+## Service Account Authenticator Modification
```
ericchiang: @liggitt mentioned service account tokens which are explicitly tied to pods. E.g. service account tokens that never expire, but are implicitly revoked when the pod they're tied to is deleted from the API.

I don't know if this is "sub" or a custom claim, but that seems appealing as a way of phasing out the current service account tokens without forcing the client (or the kubelet) to handle rotation.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+type TokenRequest struct {
+  Spec   TokenRequestSpec
+  Status TokenRequestStatus
+}
+
+type TokenRequestSpec struct {
+  // Audience is the intendend audience of the token.
+  //
+  // see audience binding
+  Audience []string
```
ericchiang: How does a pod or node determine the audience of the API server they're talking to?

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+
+The service account token authenticator will be extended to support validation
+of time and audience binding claims.
+
+## ACLs for TokenRequest
+
+A new authorizer will grant the ability for service accounts to request tokens
+for themselves. Thus the service account will need an alternative method of
+authentication to initially authenticate to the apiserver (e.g. the current
+"super-jwt"). A later design may propose how a node agent (e.g. the kubelet) can
```
ericchiang: Isn't "a later design" just this: https://github.com/kubernetes/kubernetes/pull/55019 ?

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+
+The TokenReview API will be extended to support passing an additional audience
+field which the service account authenticator will validate.
+
+```golang
+type TokenReviewSpec struct {
+  // Token is the opaque bearer token.
+  Token string
+  // Audience is the identifier that the client identifies as.
+  Audience string
```
ericchiang: Would be easy-ish to plumb this through the oidc plugin (if aud != clientid -> reject). Don't know what this would look like for the wehbook authenticator or how we'd roll it out. If a webhook authentication backend didn't consider this field you could get in some dangerous situations, right?

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+
+1.  Security: JWTs are not audience bound. Any recipient of a JWT can masquerade
+    as the presenter to anyone else.
+1.  Security: The current model of storing the service account token in a Secret
+    and delivering it to nodes results in a broad attack surface for the
+    Kubernetes control plane when powerful components are run - giving a service
+    account a permission means that any component that can see that service
+    account's secrets is at least as powerful as the component.
+1.  Security: JWTs are not time bound. A JWT compromised via 1 or 2, is valid
+    for as long as the service account exists. This may be mitigated with
```
liggitt: non-expiring tokens can be revoked by removing the associated Secret or ServiceAccount object.

I think we'd still want some form of this in order to be able to generate legacy-compatible tokens out of the API. What about this:
* a non-expiring token must be bound to a deletable object name+uid (like a pod or secret)
* non-expiring tokens can only be for the apiserver audience (since only the apiserver can be counted on to verify the bound object still exists)
* the API server must verify the object+uid still exists before verifying the token

This gets the confidential and bulky portion of unexpiring service account tokens out of the API, and still lets them be revocable. It also lets us move to this new model for both legacy and expiring/rotating cases, which is desirable.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+## Token attenuations
+
+### Audience binding
+
+Tokens issued from this API will be audience bound. Audience of requested tokens
+will be bound by the `aud` claim. The `aud` claim is an array of strings
+(usually URLs) that correspond to the intended audience of the token. A
+recipient of a token is responsible for verifying that it identifies as one of
+the values in the audience claim, and should otherwise reject the token. The
+TokenReview API will support this validation.
```
liggitt: Would the tokenreview API only support the apiserver audience?

contributors/design-proposals/auth/bound-service-account-tokens.md
liggitt: passing a token intended for someone other than the apiserver to the apiserver for verification seems a little weird

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: Yes plus a bit more. I was going to group this authorization bit into part 2) how kubelet or other node agent hands jwts to workloads. I think it is separable from this proposal though.

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: > Don't know what this would look like for the wehbook authenticator or how we'd roll it out. If a webhook authentication backend didn't consider this field you could get in some dangerous situations, right?

Is this any different from webhook authenticator and svcacct id_tokens today?

> passing a token intended for someone other than the apiserver to the apiserver for verification seems a little weird

This is true of all usage of TokenReview API endpoint though, isn't it? In this case, it is less weird since the apiserver is the thing creating the tokens in the first place. An API to correctly and opaquely verify these tokens is valuable for integrations.

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: No, I don't think so. I hope that it will support any audience so integrators can easily validate tokens.

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: Why not just return an error? What other APIs have error fields in status?

contributors/design-proposals/auth/bound-service-account-tokens.md
liggitt: what audience would the apiserver use when making a TokenReview request to an authentication webhook?

contributors/design-proposals/auth/bound-service-account-tokens.md
ericchiang: TokenReview, but yeah that's so we can surface authentication errors. You're right. Disregard.

https://godoc.org/k8s.io/api/authentication/v1#TokenReviewStatus

contributors/design-proposals/auth/bound-service-account-tokens.md
smarterclayton: Not only desirable, required.  I think we want to completely supplant the old model.

contributors/design-proposals/auth/bound-service-account-tokens.md
smarterclayton: I think that's important - would like to see it referenced or called out here.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+
+type KeyConfirmation struct {
+  // Structure TBD.
+  // This will not be supported in the initial implementation. This proposal
+  // will be updated later on.
+}
+
+type TokenRequestStatus struct {
+  // Token is the token data
+  Token string
```
smarterclayton: Can token ever be data that isn't unicode data?  I.e. do we need to deal with binary tokens?  Or will we hardcode and assume all tokens are safe for HTTP header values and for strings?

liggitt: > Can token ever be data that isn't unicode data? I.e. do we need to deal with binary tokens?

I would not expect that. Bearer tokens traditionally have a really limited character set 
```
b64token    = 1*( ALPHA / DIGIT / "-" / "." / "_" / "~" / "+" / "/" ) *"="
credentials = "Bearer" 1*SP b64token
```

TokenReviewSpec has `Token string` as well

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+will be bound by the `aud` claim. The `aud` claim is an array of strings
+(usually URLs) that correspond to the intended audience of the token. A
+recipient of a token is responsible for verifying that it identifies as one of
+the values in the audience claim, and should otherwise reject the token. The
+TokenReview API will support this validation.
+
+### Time binding
+
+Tokens issued from this API will be time bound. Time validity of these tokens
+will be claimed in the following fields:
```
liggitt: "may be time bound"... see comments in https://github.com/kubernetes/community/pull/1460/files#r156198854

there are significant benefits to switching to this provisioning mechanism even for backwards-compatible unexpiring tokens. limiting unexpiring tokens to the apiserver audience and requiring binding them to a deletable object allows revocation and prevents new uses for them

contributors/design-proposals/auth/bound-service-account-tokens.md
liggitt: and how does an API server determine the valid audiences it should accept for itself?

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,180 @@
...
+  Token string
+}
+
+type KeyConfirmation struct {
+  // Structure TBD.
+  // This will not be supported in the initial implementation. This proposal
+  // will be updated later on.
+}
+
+type TokenRequestStatus struct {
```
lookuptable: This is a duplicate of  the struct on Line 98?

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: This makes key OIDC signing key rotation difficult or impossible. Do we need to support non-expiring forever or just for some number of releases (in accordance to our GA deprecation policy).

liggitt: > This makes key OIDC signing key rotation difficult or impossible.

How so? A non-expiring token doesn't mean it can't be revoked or made invalid, it just doesn't have an inherent expiration time.

ericchiang: It means you could never remove a public key form the JWKS_URI since it could potentially need to exist forever to verify a token.

liggitt: I'd consider that working as intended. No inherent expiration, but an admin decision to revoke or rotate signing keys could invalidate it.

mikedanese: The point is that it's not easy to maintain compatibility with the current client-go while also rotating SA signing keys. If we rotate signing keys every 4 days, how is a non-expiring token useful?

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: I think this should be either `kubernetes.default.svc` or the ClusterIP of the kubernetes service or both.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,179 @@
...
+  ValidityDuration metav1.Duration
+
+  // KeyConfirmation contains information required by a recipient of a token
+  // to verify possession of a key.
+  //
+  // see key binding
+  KeyConfirmation KeyConfirmation
+}
+
+type TokenRequestStatus struct {
```
liggitt: need to include the actual duration in the response, so the receiver knows how long the token is valid for

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,179 @@
...
+  // the audiences listed. This implies a high degree of trust between
+  // the target audiences.
+  //
+  // see audience binding
+  Audience []string
+
+  // ValidityDuration is the intended duration of validity of the request.
+  //
+  // see time binding
+  ValidityDuration metav1.Duration
```
liggitt: I'd expect the cluster admin to be able to configure:
* whether non-expiring tokens are allowed
* max validity duration for expiring tokens

Token requesters should be able to tell from the status:
* whether they were issued an expiring or non-expiring token
* for expiring tokens, how long the token is valid for

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,179 @@
...
+We will add an imperative API (a la TokenReview) to the
+`authentication.k8s.io` API group:
+
+```golang
+type TokenRequest struct {
+  Spec   TokenRequestSpec
+  Status TokenRequestStatus
+}
+
+type TokenRequestSpec struct {
```
liggitt: all the use cases I can think of for token requests benefit from including kind+name+uid of at least one object as a claim in the token:
* a token requested by node `n` for service account `sa` should be bound to a pod with `serviceAccountName=sa` and `nodeName=n` (and stop working against the apiserver when that pod goes away)
* a non-expiring token should be bound to some object in the namespace that indicates the existence of the token, and can be deleted in order to revoke the token (just like a service account token secret behaves today)

What if the TokenRequestSpec included an object reference (or list of object references) that the token should be bound to? That would let the NodeRestriction plugin limit nodes to requesting tokens bound to pods scheduled to themselves and referencing the correct serviceaccount.

mikedanese: Should this be an ObjectReference or LocalObjectReference (would have to add a few fields to LocalObjectReference) or should I create a new type for this?

mikedanese: Added a section on object binding.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,193 @@
...
+  // the target audiences.
+  Audience []string
+
+  // ValidityDuration is the requested duration of validity of the request. The
+  // token issuer may return a token with a different validity duration so a
+  // client needs to check the 'expiration' field in a response.
+  ValidityDuration metav1.Duration
+
+  // BoundObject is a reference to an object that the token will be bound to.
+  // The token will only be valid for as long as the bound objet exists.
```
ericchiang: s/objet/object/g

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,193 @@
...
+  // the audiences listed. This implies a high degree of trust between
+  // the target audiences.
+  Audience []string
+
+  // ValidityDuration is the requested duration of validity of the request. The
+  // token issuer may return a token with a different validity duration so a
+  // client needs to check the 'expiration' field in a response.
+  ValidityDuration metav1.Duration
+
+  // BoundObject is a reference to an object that the token will be bound to.
```
ericchiang: Couldn't a token be bounded to multiple objects? A token given to a node for a pod running a service account could be tied to:

* The node
* The pod
* The service account

mikedanese: A token should always be bound to a service account i.e. the service account needs to exist for the service account token to be valid. I consider these bindings to be extras (e.g. pod, node, secret). I don't think N bindings over 1 binding adds extra value to warrant the complexity. Deleting a node today should cause all pods on the node to be deleted almost instantaneously. Is there any other usecases for multiple bound objects?

mikedanese:
@liggitt PTAL

contributors/design-proposals/auth/bound-service-account-tokens.md
liggitt: >  I don't think N bindings over 1 binding adds extra value to warrant the complexity.

I tend to agree (and also don't want unbounded items to resolve at validation time)

contributors/design-proposals/auth/bound-service-account-tokens.md
liggitt: Last question is still outstanding

mikedanese: Sorry, missed this. I expect that the apiserver would populate audience with every name that it identifies itself as. Initially that would be `["https://kubernetes.default.svc"]` but we could add configuration to override or add more identifiers to  that list. This would be the same list of audiences that the apiserver would validate service account id tokens against in the service account id token authenticator.

liggitt: So tokenrequest provides a list of audiences that means the token is valid for *all* of those audiences?

And tokenreview takes a list of audiences that means the token should be valid for *at least one* of those audiences? (and the apiserver would pass all the audiences it knew for itself to the authentication webhook?)

mikedanese: Yes, that seems like the correct semantics to me. If you agree, I'll add that to this proposal

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,280 @@
...
+* `iat`: issued at
+
+A recipient of a token should verify that the token is valid at the time that
+the token is presented, and should otherwise reject the token. The TokenReview
+API will support this validation.
+
+Cluster administrators will be able to configure the maximum validity duration
+for expiring tokens. During the migration off of the old service account tokens,
+clients of this API may request tokens that are valid for many years. These
+tokens will be drop in replacements for the current service account tokens.
```
mikedanese: Note: I've opted to require time binding on all JWTs issued by this API. It eliminates the need to support to styles of token and can still support everything we want to support.

contributors/design-proposals/auth/bound-service-account-tokens.md
liggitt: Thinking through the audience implications on TokenReview for the following cases:

* kubelets verifying presented bearer tokens against kube-apiserver's TokenReview
  * old kubelets won't specify an audience... default should mean "valid against kube-apiserver" as it does today
  * what audience(s) should new kubelets specify? 

* extension API servers verifying bearer tokens against kube-apiserver's TokenReview
  * old extension servers won't specify an audience... default should mean "valid against kube-apiserver" as it does today
  * what audience(s) should new extension API servers specify?

* kube-apiserver verifying bearer tokens against the authentication webhook TokenReview endpoint
  * old webhooks won't be audience-aware and will just verify the token
  * what audience(s) should new webhooks expect/verify?

Also, the kube-apiserver doesn't have the plumbing needed to verify tokens for arbitrary audiences (which would be required for it to handle an `Audiences` field in `TokenReview`). The way kube-apiserver implements TokenReview today is to pass a constructed http request containing `Authorization: Bearer <token>` to the request authenticator:

https://github.com/kubernetes/kubernetes/blob/c17b418f89e056bea272ed46ca9f63f8b1e37589/pkg/registry/authentication/tokenreview/storage.go#L61-L69

I'm afraid adding arbitrary audience support would mean expanding the AuthenticateRequest, AuthenticateToken, and AuthenticatePassword interface signatures.

mikedanese:
Most of this is alpha in 1.10. Implementation is tracked here: https://github.com/kubernetes/kubernetes/issues/58790

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: > kubelets verifying presented bearer tokens against kube-apiserver's TokenReview
> * old kubelets won't specify an audience... default should mean "valid against kube-apiserver" as it does today
> * what audience(s) should new kubelets specify?

When configured to have individual audiences, the identifiers that they request as SANs seems reasonable.

https://github.com/kubernetes/kubernetes/blob/9847c8ee0a2c71fc4e9e1665017f05de7c9f815e/pkg/kubelet/kubelet.go#L745-L746

>
> extension API servers verifying bearer tokens against kube-apiserver's TokenReview
> * old extension servers won't specify an audience... default should mean "valid against kube-apiserver" as it does today
> * what audience(s) should new extension API servers specify?

Extension apiservers are rarely accessed directly today. In the cases where they are, a security conscious operator may want to configure a separate audience for the extension apiserver.

>
> kube-apiserver verifying bearer tokens against the authentication webhook TokenReview endpoint
> * old webhooks won't be audience-aware and will just verify the token
> * what audience(s) should new webhooks expect/verify?

I would not expect an google access token with scope "cloudplatform" to be valid against vault. I would not expect a google oidc id token (where the audience is the client_id) to be valid against vault. Moving forward, I would expect TokenReview webhooks to become aware of the new field.

> I'm afraid adding arbitrary audience support would mean expanding the AuthenticateRequest, AuthenticateToken, and AuthenticatePassword

It seems like it's this or some hacky side channel.

mikedanese: > I would expect TokenReview webhooks to become aware of the new field.

We could also not forward TokenReviews with non-apiserver audiences for some period of time.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,280 @@
...
+The service account token authenticator will be extended to support validation
+of time and audience binding claims.
+
+## ACLs for TokenRequest
+
+A new authorizer will grant the ability for service accounts to request tokens
+for themselves. Thus the service account will need an alternative method of
+authentication to initially authenticate to the apiserver (e.g. the current
+"super-jwt"). A later design may propose how a node agent (e.g. the kubelet) can
+use it's credentials to request a service account jwt on behalf of the pod.
```
jboeuf: typo: s/it's/its

contributors/design-proposals/auth/bound-service-account-tokens.md
liggitt: I think those make sense, though I'm unsure how we'd automatically transition to the "use individual audience" option over time

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,280 @@
...
+```golang
+type TokenReviewSpec struct {
+  // Token is the opaque bearer token.
+  Token string
+  // Audience is the identifier that the client identifies as.
+  Audience string
+}
+```
+
+### Add Service Account signing key retrieval API
```
liggitt: let's pull this bit into a follow-up PR so the proposal can be merged

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+## Service Account Authenticator Modification
+
+The service account token authenticator will be extended to support validation
+of time and audience binding claims.
+
+## ACLs for TokenRequest
+
+A new authorizer will grant the ability for service accounts to request tokens
```
liggitt: can update this to describe the node authorizer bits, and defer the self-token-request bit for a follow up

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,280 @@
...
+for themselves. Thus the service account will need an alternative method of
+authentication to initially authenticate to the apiserver (e.g. the current
+"super-jwt"). A later design may propose how a node agent (e.g. the kubelet) can
+use it's credentials to request a service account jwt on behalf of the pod.
+
+## Footnotes
+
+* New apiserver flags:
+  * --id-token-issuer-id: Identifier of the issuer.
```
liggitt: harmonize with the initial alpha impl

mikedanese:
@liggitt @jboeuf addressed comments.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -0,0 +1,239 @@
...
+  }
+}
+```
+
+The token payload will be:
+
+```
+{
+  "iss": "https://example.com/some/path",
+  "sub": "system:serviceaccount:default:default,
```
yue9944882: Hi @mikedanese , do you mean `Subject` here?  Why is it removed from API definition now?😅

liggitt: this is a claim in the generated service account JWT. it is still present.

PR #1461, opened 2017-12-05
Update the schedule

castrojo:
updating the schedule for tomorrow's event


castrojo:
/lgtm

I know this bad form, but alas, I am out of time.

PR #1465, opened 2017-12-06
Update SIG-UI meeting time

floreks:


cblecker:
@floreks This should be updated here: https://github.com/kubernetes/community/blob/3d960a351c44594deda6b8c28b16b78ee76ec7f5/sigs.yaml#L1083-L1087

And then run `make`. You can also update the time zone if it would help keep it up to date with DST (many meetings are tracking in PT `tz: "PT (Pacific Time)"`)

floreks:
Thanks for information. I'll do that.

floreks:
Closing in favour of #1487

PR #1470, opened 2017-12-08
Update README

parispittman:
added inspiration and thanks, pilot info, and reformatted


stp-ip:
Awesome. No suggestions.
New Approvers check won't let me approve, but "lgtm" from my side.

PR #1479, opened 2017-12-08
Fix feature-roadmap-2018.jpg

cblecker:
Looks like an HTML file was uploaded instead of the jpg. Can you verify this is the correct file, @castrojo ?


castrojo:
/lgtm

Thanks for the catch!

PR #1484, opened 2017-12-11
PV monitoring proposal

NickrenREN:
Add a proposal for PV monitoring.
We may focus on local storage PV monitoring at the first phase.

/assign @msau42 
/cc @ddysher  @jingxu97 
/sig storage

### DIFF
--- a/contributors/design-proposals/storage/pv-monitoring-proposal.md
+++ b/contributors/design-proposals/storage/pv-monitoring-proposal.md
@@ -1,0 +1,190 @@
+# PV monitoring proposal
+
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: NickrenREN@ 
+
+## Motivation
+
+For now, kubernetes has no way to monitor the PVs, which may cause serious problems. 
+For example: if volumes are unhealthy, and pods do not know that and still try to get and write data, 
+which will lead to data loss and unavailability of services. 
+So it is necessary to have a mechanism for monitoring PVs and react when PVs have problems.
+
+## Proposal
+
+We can separate the proposal into two parts:
+
+* monitoring PVs and marking them if they have problems
+* reacting to the unhealthy PVs
+
+For monitoring, we may create a controller for it, and each volume plugin should have its own function to check volume health. 
+Controller can call them periodically. The controller also needs to watch node events because local PVs will be unreachable if nodes break down.
+
+For reacting, different kinds of apps may have different methods,we can also create a controller for it. 
+
+At first phase, we can focus on local storage PVs monitoring.
+
+## User Experience
+### Use Cases
+
+* If the local PV path is deleted, users should know that and the local PV should be marked and deleted;
+* If the local PV path is not a mountpoint any more, the local PV should be marked and deleted;
+* If nodes which have local PVs are breaking down, the local PVs should be marked and deleted (the application has data backup and can restore it or can tolerate data loss and the PV protection feature may help);
+* For local PVs, we need to make sure that PV capacity must not be greater than device capacity and PV used bytes must not be greater than PV capacity;
+* For network storage, if the storage driver volume is deleted, the PV object in kubernetes should be marked and deleted too;
+* If we can not get access to the PV volume for a certain time (network or some other problems), we need to mark and delete the PV;
+* PV fsType checking ? bad blocks checking ?
+
+## Implementation
+
+As mentioned above, we can split this into two parts and put them in the external repo at first.
+
+### Monitoring controller: 
+
+Like PV controller, monitoring controller should check PVs’ health condition periodically and taint them if PVs are unhealthy.
+
+Health checking implementation should be per plugin. Each volume plugin needs to have its own methods to check its volumes.
+
+At the first stage, we can focus on local storage PVs, and then extend to other network storage PVs.
+#### For local storage:
+
+The local storage PV monitor consists of two parts
+
+* create a daemonset on every node, which is responsible for monitoring local PVs in that specific node, no matter the PVs are created manually or by provisioner;
+* create a monitor controller, which is responsible for watching PVs and Nodes events. PVs may be updated if they are unhealthy and we also need to react to node failure event.
+
+At the first phase, we can support local storage monitoring first.
+
+Take local storage as an example, detailed checking method may be like this:
+
+```
+// checkStatus checks local pv health condition
+func (monitor *LocalPVMonitor) checkStatus(pv *v1.PersistentVolume) {
+    // check if PV is local storage
+	if pv.Spec.Local == nil {
+		glog.Infof("PV: %s is not local storage", pv.Name)
+		return
+	}
+	// check node and pv affinity
+	fit, err := CheckNodeAffinity(pv, monitor.Node.Labels)
+	if err != nil {
+		glog.Errorf("check node affinity error: %v", err)
+		return
+	}
+	if !fit {
+		glog.Errorf("pv: %s does not belong to this node: %s", pv.Name, monitor.Node.Name)
+		return
+	}
+
+	// check if host dir still exists
+	mountPath, continueThisCheck := monitor.checkHostDir(pv)
+	if !continueThisCheck {
+		glog.Errorf("Host dir is modified, PV should be marked")
+		return
+	}
+
+	// check if it is still a mount point
+	continueThisCheck = monitor.checkMountPoint(mountPath, pv)
+	if !continueThisCheck {
+		glog.Errorf("Retrieving mount points error or %s is not a mount point any more", mountPath)
+		return
+	}
+
+	// check PV size: PV capacity must not be greater than device capacity and PV used bytes must not be greater that PV capacity
+	if pv.Spec.VolumeMode != nil && *pv.Spec.VolumeMode == v1.PersistentVolumeBlock {
+		monitor.checkPVAndBlockSize(mountPath, pv)
+	} else {
+		monitor.checkPVAndFSSize(mountPath, pv)
+	}
+
+    // other checks ...
+}
+```
+If monitor finds that one PV is unhealthy, it will mark the PV by adding annotations including timestamp. 
+The reaction controller then can react to this PV depending on the annotations and timestamp.
+
+When we first mark a PV, we will add another annotation which key is `FirstMarkTime`. 
+And if local PV is unhealthy, annotation keys may be like: `HostPathNotExist`, `MisMatchedVolSize`, and `NotMountPoint`...
+
+A marked local PV looks like:
+```
+Name:            example-local-pv-1
+Labels:          <none>
+Annotations:     FirstMarkTime=2018-04-17 07:31:02.388570492 +0000 UTC m=+600.033905921
+                 HostPathNotExist=yes
+                 NotMountPoint=yes
+                 volume.alpha.kubernetes.io/node-affinity={ "requiredDuringSchedulingIgnoredDuringExecution": { "nodeSelectorTerms": [ { "matchExpressions": [ { "key": "kubernetes.io/hostname", "operator": "In", "valu...
+Finalizers:      [kubernetes.io/pv-protection]
+StorageClass:    local-disks
+Status:          Available
+Claim:
+Reclaim Policy:  Retain
+Access Modes:    RWO
+Capacity:        200Mi
+Node Affinity:   <none>
+Message:
+Source:
+    Type:  LocalVolume (a persistent volume backed by local storage on a node)
+    Path:  /mnt/disks/vol/vol1
+Events:
+  Type    Reason           Age   From                                                                 Message
+  ----    ------           ----  ----                                                                 -------
+  Normal  MarkPVSucceeded  1m    local-volume-monitor-127.0.0.1-40a8fb4d-4206-11e8-8e52-080027765304  Mark PV successfully with annotation key: NotMountPoint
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,127 @@
...
+#### For local storage:
+
+We can put monitor in its provisioner package and each provisioner is responsible for checking volumes health condition in that specific node.
+
+At the first phase, we can support local storage monitoring first.
+
+Take local storage as an example, implementation may be like this:
+
+```
+ func (mon *minitor) CheckStatus(spec *v1.PersistentVolumeSpec) (string, error) {
```
NickrenREN: This needs to be discussed

redbaron:
It is not clear how apps can react to bad news and recover from failure. I'd be very hesitant to allow Statefulset to delete PVC should there be failure detected:
 - there is no guarantee that new PV created will be healthy
 - removing PVC removes data, so if PV has a chance to become healthy later on I'd much rather wait
 - usually PV failure is tied with storage failure, that means that all PVs in that failure domain are going to report bad health state and therefore deleted. This is catastrophic

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,127 @@
...
+* We should mark the PV if it is not healthy;
+* Controllers such as statefulset controller should be aware of the unhealthy volumes and react to them;
+
+## Implementation
+
+As mentioned above, we can split this into two parts and put them in the external repo at first.
+
+### Monitoring controller: 
+
+Like PV controller, monitoring controller should check PVs’ health condition periodically and taint them if PVs are unhealthy.
```
gnufied: What happens when PVs are tainted? Currently we do not have ability to taint PVs

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,127 @@
...
+For now, kubernetes has no way to monitor the PVs, which may cause serious problems. 
+For example: if volumes are unhealthy, and pods do not know that and still try to get and write data, 
+which will lead to data loss and unavailability of services. 
+So it is necessary to have a mechanism for monitoring PVs and react when PVs have problems.
+
+## Proposal
+
+We can separate the proposal into two parts:
+
+* monitoring PVs and marking them if they have problems
```
gnufied: What exactly will it monitor btw? Are you talking about bad blocks or fsck? Please add some examples.

msau42:
@redbaron agree.  That is why we should clearly document here exactly what use cases we are trying to solve.  And also this reaction part needs to be opt-in by the application, so that only those that can recover from these scenarios can benefit.

From a workloads point of view, we want to target distributed applications like Cassandra, that replica their data across multiple instances for redundancy.

From a platform point of view, the infrastructure failures I'm thinking of are:
* local disks in cloud providers, where if the VM is deleted, then your local disk is really gone forever.
* local disks in on-prem, and the disk hardware has failed (open issue on how to actually detect that).  If you don't have them in a RAID config, then the data is also gone.

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: yeah, sth like that,  e.g.
For local PV: its Path( directory) is deleted by mistake , node failure...

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: Yeah, at the first phase, do not need to change PV struct,  we will mark PV by adding annotations instead.

NickrenREN:
@redbaron yeah, if we want to delete the PVC and pods directly and reschedule them to a new node, the application needs to have data backup and can restore it or can tolerate data loss

msau42:
cc @vkamra

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,128 @@
...
+                return ...
+        }
+        
+        // check if PV belongs to this node
+        if checkNodeAffinity(...) {
+                glog.Infof("PV: %s does not belong to this node", pv.Name)
+                return ...
+        }
+
+        // check volume health condition depending on device type
```
msau42: Can you outline here what kind of checks you want to do?  What are some general checks, and what are some specific checks?

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,128 @@
...
+The reaction controller then can react to this PV depending on the annotations and timestamp.
+
+### Reaction controller:
+
+Reaction controller will react to the PV update event (tainting PV by monitoring controller). 
+Different kinds of apps should have different reactions.
+
+At the first phase, we just consider the statefulset reaction.
+
+statefulset reaction: check the annotation timestamp, if the PV can recover within the predefined time interval, we will do nothing,
```
msau42: How does this controller know which StatefulSets to monitor?

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,128 @@
...
+                glog.Infof("PV: %s does not belong to this node", pv.Name)
+                return ...
+        }
+
+        // check volume health condition depending on device type
+        ...
+
+}
+```
+If monitor finds that one PV is unhealthy, it will mark the PV by adding annotations including timestamp.
```
msau42: Can you define the annotation key/value?

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,128 @@
...
+* Taint PV if the volume status is abnormal
+* Remove the taints if the volume becomes normal again
+
+#### For out-tree volume plugins(except local storage):
+
+We can create a new controller called MonitorController like ProvisionController, 
+which is responsible for creating informers, watching Node and PV events and calling each plugin’s monitor functions. 
+And each volume plugin will create its own monitor to check its volumes’ status.
+
+#### For local storage:
```
msau42: I would be very interested in seeing an outline for monitoring node failures, which is a failure mode that local volumes are especially prone to compared to other volume types.

contributors/design-proposals/storage/pv-monitoring-proposal.md
msau42: It would be good to explicitly write down some specific error conditions somewhere.

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,128 @@
...
+Controller can call them periodically. The controller also needs to watch node events because local PVs will be unreachable if nodes break down.
+
+For reacting, different kinds of apps may have different methods,we can also create a controller for it. 
+
+At first phase, we may just react to statefulsets with local storage PVs.
+
+## User Experience
+### Use Cases
+
+* Users create many local PVs manually and want to monitor their health condition;
```
msau42: Since all of the use cases here are for local PVs, should we just have this document be specific to local PVs for now?  We can consider extending this to other PV types in the future if the need arises.

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,128 @@
...
+
+statefulset reaction: check the annotation timestamp, if the PV can recover within the predefined time interval, we will do nothing, 
+otherwise we need to delete the PVC bound to the unhealthy volume(PV) as well as pods referencing it.
+
+Reaction controller’s workflow is:
+
+* Fill PV cache from etcd;
+* Watch for PV update events;
+* Resync and populate periodically;
+* Delete related PVC and pods if needed (just for statefulsets and reclaim PV depending on reclaim policy);
```
msau42: BTW, what application are you planning to use with this?  It may be good to mention in the beginning use cases what kinds of applications this is suitable for.

NickrenREN:
@ddysher @msau42  update the use cases here, and will modify the rest accordingly.
simple example is here: https://github.com/kubernetes-incubator/external-storage/pull/528

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: general checks: volume accessing check, volume existing check ...
specific checks: mountpoint check, local path check,...
will update

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: by monitoring relatedPVs

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: yeah, sure, defined in demo PR: kubernetes-incubator/external-storage#528
will update this proposal

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: yeah, will do

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: if we want to delete the PV directly, the application should have data backup and can restore it or can tolerate data loss. btw: the PV protection feature may help

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: yeah, we want to focus on local storage at the first stage, and then move to other storage drivers if needed.

contributors/design-proposals/storage/pv-monitoring-proposal.md
msau42: How do we mark which PVs we should monitor?  For example, not all StatefulSets may want this behavior.  So we need a way to be able to say which StatefulSets/PVs should be handled by the reaction controller.

BTW, @jingxu97 just showed me some cool work related to a Snapshot policy controller using the [metacontroller framework](https://github.com/GoogleCloudPlatform/metacontroller).  A CRD is defined that specifies:
* How often a snapshot should be taken
* Label selector to choose which PVCs should follow this policy.

I think we could do something similar for this controller, and have a CRD to define a policy with:
* What conditions to react to
* Time thresholds before reacting
* What should the action be for each condition
* Label selector for which StatefulSets to monitor

contributors/design-proposals/storage/pv-monitoring-proposal.md
msau42: Can you define the annotation format and syntax in this design doc, and also put some of the examples here too?

contributors/design-proposals/storage/pv-monitoring-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+Controller can call them periodically. The controller also needs to watch node events because local PVs will be unreachable if nodes break down.
+
+For reacting, different kinds of apps may have different methods,we can also create a controller for it. 
+
+At first phase, we may just react to statefulsets with local storage PVs.
+
+## User Experience
+### Use Cases
+
+* If the local PV path is deleted, users should know that and the local PV should be marked and deleted;
```
msau42: It seems like there's two different reaction controllers being proposed here?

* One for StatefulSets that are using these PVs
* One for local PVs

Could you try to clarify this in the doc and separate out which behaviors will be handled by which controllers?  I think one of the confusing things here is there's multiple levels/layers of reaction and it's not clear how they will all interact with each other.

msau42: Or maybe this design should just focus on the local PV monitoring part.  And we can leave StatefulSet handling to a different design.

NickrenREN: yes, reaction needs more discussion, so i will move it to next stage.

contributors/design-proposals/storage/pv-monitoring-proposal.md
NickrenREN: done

ddysher:
@NickrenREN @msau42 what's the status of the proposal :)

NickrenREN:
@ddysher  I am prototyping monitor here in https://github.com/caicloud/kube-storage-monitor
And will update this proposal soon. I am planning to add local volume PV monitor support this Q,  WDYT, @msau42  ?

msau42:
I probably won't have time to thoroughly review this until Q3 timeframe.

But it would help if in this design proposal, you could summarize things into a table with the following columns:
* Failure type
* Reaction (and by who)

Also it would be good to think about some of these questions:
* Can reaction be configurable per PV/PVC?  Reaction should only be opt-in because it could have dangerous consequences for various workloads.
* What happens if the failure is only temporary? Can we have configurable timeouts before reacting?

I really like the idea of having CRDs where you can define your reaction policies, and then using something like the metacontroller to write a reaction controller.  I would sync up with @jingxu97 on the work she did for snapshot policies.

msau42:
I opened up https://github.com/kubernetes-incubator/external-storage/issues/817 to discuss ideas for handling the node deletion scenario.

NickrenREN:
Closing, in favor of https://github.com/kubernetes/enhancements/pull/1077

PR #1487, opened 2017-12-11
Update SIG-UI meeting time

floreks:


sigs.yaml
```diff
@@ -1082,8 +1082,8 @@ sigs:
       company: Fujitsu
     meetings:
     - day: Thursday
-      time: "16:00"
-      tz: "UTC"
+      time: "18:00"
+      tz: "CET"
```
cblecker: Can you change this to `tz: "CET (Central European Time)"` (and then re-run `make`)? This will ensure that the time zone convertor link works properly.

floreks:
Done. PTAL.

PR #1495, opened 2017-12-12
Update SIG-apps leads

erictune:
Adds @kow3ns.


cblecker:
/approve
Should get an LGTM from @michelleN, @prydonius, or @mattfarina.

PR #1501, opened 2017-12-13
Complete File an Issue section

spzala:
The File an Issue section is work in progress. Adding more detail with a broader
coverage on what to file, why and where to file a issue to complete this section.


spzala:
@castrojo thanks for the quick review, updated changes with your suggestion.

PR #1519, opened 2017-12-16
security-release-process: remove PGP requirement

jessfraz:


liggitt:
lgtm.  would like a second lgtm from PST

philips:
thanks @jessfraz

PR #1524, opened 2017-12-18
Propose a test frameworks git repo

totherme:
Following up on [this email thread](https://groups.google.com/a/kubernetes.io/forum/?utm_medium=email&utm_source=footer#!msg/steering/LA9WiFnl6PI/DjiPaN-2CgAJ), this is a proposal for a new git repository to house:

- the [integration testing framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test) which is currently being built in the kubectl repo
- possibly also the framework currently in `kubernetes/test/e2e/framework`

### DIFF
--- a/contributors/design-proposals/testing/test-frameworks-repo.md
+++ b/contributors/design-proposals/testing/test-frameworks-repo.md
@@ -36,6 +36,17 @@ desire](https://groups.google.com/a/kubernetes.io/d/msg/steering/LA9WiFnl6PI/os4
 to get `kubernetes/test/e2e/framework` out into a more reusable place. It could
 live in `k8s.io/testframeworks/e2e`
 
+### What repos can be imported (and hence vendored) into this repo?
+
+No kubernetes repos other than [utils](https://github.com/kubernetes/utils)
+should be imported (and hence vendored) by any code in this repo.
+
+This means that any testing framework in this repo which interacts with any
+part of kubernetes has to do so through some CLI. The [CLI integration testing
+framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test)
+which originally motivated this repo certainly does not have any kubernetes
+dependencies.
+
 ### Other naming concerns
 
 We would like the name of the repo to be a reasonable [golang package
--- a/contributors/design-proposals/testing/test-frameworks-repo.md
+++ b/contributors/design-proposals/testing/test-frameworks-repo.md
@@ -4,7 +4,7 @@ Status: Pending
 
 Version: N/A
 
-Implementation Owners: @totherme @hoegaarden and @apelisse
+Implementation Owners: @totherme, @hoegaarden, @apelisse, and @spiffxp
 
 ## Motivation: Why do we need a new repo?
 
@@ -22,19 +22,21 @@ frameworks. The intent is that this repo could be vendored anywhere.
 
 ## Proposal
 
-We propose a new git repository named `k8s.io/testframeworks`. This would be
-backed by a github repository `github.com/kubernetes/testframeworks`.
+We propose a new git repository named `testing.k8s.io/frameworks`. This would be
+backed by a github repository `github.com/kubernetes-sig-testing/frameworks`.
 
 ### What should live in it?
 
 The [integration testing
 framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test)
-should live in `k8s.io/testframeworks/integration`.
+should live in `testing.k8s.io/frameworks/integration`.
 
 There also [seems to be a
 desire](https://groups.google.com/a/kubernetes.io/d/msg/steering/LA9WiFnl6PI/os48-c3HCgAJ)
 to get `kubernetes/test/e2e/framework` out into a more reusable place. It could
-live in `k8s.io/testframeworks/e2e`
+live in `testing.k8s.io/frameworks/e2e`. However, to make this happen we'll
+need to do a lot of work to remove some of the less necessary dependencies of
+the existing e2e framework.
 
 ### What repos can be imported (and hence vendored) into this repo?
 
@@ -46,6 +48,11 @@ part of kubernetes has to do so through some CLI. The [CLI integration testing
 framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test)
 which originally motivated this repo certainly does not have any kubernetes
 dependencies.
+
+We might revisit this decision in future, for example to potentially allow the
+usage of a client library if that turns out to both:
+- Be necessary for some important framework (such as e2e)
+- Not harm the existing usage of other frameworks in the repo
 
 ### Other naming concerns
 
@@ -69,10 +76,16 @@ currently doing this in the kubectl repo, and are happy to keep doing so here.
 
 In [this email
 thread](https://groups.google.com/a/kubernetes.io/forum/?utm_medium=email&utm_source=footer#!msg/steering/LA9WiFnl6PI/DjiPaN-2CgAJ)
+and in [the sig-architecture meeting on
+2018-01-04](https://docs.google.com/document/d/1BlmHq5uPyBUDlppYqAAzslVbAO8hilgjqZUTaNXUhKM/edit#heading=h.dm9wr8ympgj8)
 the following alternatives have been considered and found to be unpopular:
 
-- Create a new github org for new repos (such as this one) to exist in.
-   + The consensus seems to be that this involves more organisational overhead than it's worth.
+- Create this repo in the `kubernetes` github org.
+  + There is already a proposal to create one org per sig
+  + @spiffxp from sig-testing expressed an interest in supporting this repo
+    from sig-testing
+  + it was decided to create an org for sig-testing to trial the whole
+    one-org-per-sig concept, with this repo
 - Create a new repo which only contains the integration testing framework, and
   cannot also be used to house the e2e framework.
    + The consensus seems to be that this leads to ugly package names
--- a/contributors/design-proposals/testing/test-frameworks-repo.md
+++ b/contributors/design-proposals/testing/test-frameworks-repo.md
@@ -4,7 +4,7 @@ Status: Pending
 
 Version: N/A
 
-Implementation Owner: TBD
+Implementation Owners: @totherme @hoegaarden and @apelisse
 
 ## Motivation: Why do we need a new repo?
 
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/testing/test-frameworks-repo.md
```diff
@@ -0,0 +1,65 @@
...
+
+It therefore seems like a good idea to create a new repo to contain testing
+frameworks. The intent is that this repo could be vendored anywhere.
+
+## Proposal
+
+We propose a new git repository named `k8s.io/testframeworks`. This would be
+backed by a github repository `github.com/kubernetes/testframeworks`.
+
+### What should live in it?
```
stevekuznetsov: We might want to make explicit which other `k8s.io/` repos this would be allowed to import

contributors/design-proposals/testing/test-frameworks-repo.md
```diff
@@ -0,0 +1,65 @@
...
+live in `k8s.io/testframeworks/e2e`
+
+### Other naming concerns
+
+We would like the name of the repo to be a reasonable [golang package
+name](https://blog.golang.org/package-names).  We should therefore probably
+avoid characters that are not lower case alphabetic characters. This precludes
+for example `k8s.io/test-frameworks`.
+
+### Who should own it?
```
stevekuznetsov: Also, what does ownership mean?

contributors/design-proposals/testing/test-frameworks-repo.md
totherme: Good call. Which repos would folks like? :)

contributors/design-proposals/testing/test-frameworks-repo.md
totherme: In [this email](https://groups.google.com/a/kubernetes.io/d/msg/steering/LA9WiFnl6PI/os48-c3HCgAJ) when @spiffxp said "would naturally fall under the ownership of sig-testing", I presumed that "ownership" effectively meant something like "whose names go in the top level OWNERS file". I may have been wrong...

contributors/design-proposals/testing/test-frameworks-repo.md
apelisse: Ideally, none.

stevekuznetsov: But the e2e might depend on the client? (Or does it shell out?)

apelisse: If it does, then I don't think it can live outside of k/k.

contributors/design-proposals/testing/test-frameworks-repo.md
spiffxp: I would expect the individuals proposing to work on this project to be the leads for it, and responsible for actually driving / implementing it.  This is not a thing I have the capacity to staff or drive.  So I'd like to see the "implementation owner TBD" part filled out.

But in the context of SIG's needing to own all things, in the same way that SIG Apps "owns" helm and charts, I think it makes sense for SIG Testing to own this.  I would like to hear from @timothysc if this aligns with his proposal to expand the charter of sig-testing but I'm personally +1 here

totherme: I think the "individuals proposing to work on this project" are initially me, @hoegaarden and @apelisse. So far all the PRs on the [integration framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test) have been authored by me and @hoegaarden, and reviewed by @apelisse.

I can't speak with as much certainty for @apelisse , but I know that @hoegaarden and I are happy to be responsible for driving/implementing this framework so long as that makes sense for everyone else.

totherme: I've added our (my, @apelisse, and @hoegaarden's) names to the proposal, and clarified my understanding of the difference between the "SIG's needing to own all things" context, and the "day to day looking after the code" context. Does that work for everyone?

spiffxp:
/approve
/hold
I am happy with this, but would like to see an /lgtm from @timothysc (or some other sig-testing lead)

stevekuznetsov:
I think we may want to frame specifically which other repos this framework could import, if any, and enumerate them in the proposal. As it's meant to be helping with import cycles, that seems important to me as part of the charter for the repo.

apelisse:
I can't edit the pull-request and its authors are on vacation. I'd still really like the repo to exists when they come back.

I'll mention it here again, and I think it definitely should be printed in **bold** in the README at the root of the repo, but this is *NOT* going to import any other kubernetes repository (maybe `k8s.io/utils` would be the exception).

totherme:
I've added a section saying explicitly that no k8s repos other than `utils` should be imported. I know for sure that the CLI integration testing framework which originally motivated this PR does not import any kubernetes repos.

I can see (using `ag k8s.io`) that there seem to be plenty of files in [kubernetes/test/e2e/framework](https://github.com/kubernetes/kubernetes/tree/master/test/e2e/framework) that import `apimachinery`, `client-go`, `api/extensions`, `pkg/kubelet`, `pkg/cloudprovider`, `apiserver`, and so on. I therefore suspect that when @spiffxp [said "Getting kubernetes/test/e2e/framework out into a reusable place is something I keep hearing a desire for"](https://groups.google.com/a/kubernetes.io/forum/#!msg/steering/LA9WiFnl6PI/os48-c3HCgAJ), they probably meant that it might be a good idea to have space available in case e2e is able to make the move at some point in future. I think they didn't mean that we should whitelist all those packages for import. Have I understood you ok @spiffxp ?

spiffxp:
@totherme you have understood correctly, thanks

contributors/design-proposals/testing/test-frameworks-repo.md
```diff
@@ -0,0 +1,81 @@
...
+live in `k8s.io/kubernetes`. The framework currently lives in `k8s.io/kubectl`.
+Since `k8s.io/kubectl` already vendors components _from_ `k8s.io/kubernetes`, it 
+cannot be vendored _into_ `k8s.io/kubernetes`. 
+
+It therefore seems like a good idea to create a new repo to contain testing
+frameworks. The intent is that this repo could be vendored anywhere.
+
+## Proposal
+
+We propose a new git repository named `k8s.io/testframeworks`. This would be
```
timothysc: I'd prefer just to call it the testing repo.  b/c there could also be separate libraries and possible binary artifacts.

contributors/design-proposals/testing/test-frameworks-repo.md
```diff
@@ -0,0 +1,81 @@
...
+### What should live in it?
+
+The [integration testing
+framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test)
+should live in `k8s.io/testframeworks/integration`.
+
+There also [seems to be a
+desire](https://groups.google.com/a/kubernetes.io/d/msg/steering/LA9WiFnl6PI/os48-c3HCgAJ)
+to get `kubernetes/test/e2e/framework` out into a more reusable place. It could
+live in `k8s.io/testframeworks/e2e`
```
timothysc: There are a number of test-specific utility libraries that should be housed in this repo.

contributors/design-proposals/testing/test-frameworks-repo.md
```diff
@@ -0,0 +1,81 @@
...
+
+### What repos can be imported (and hence vendored) into this repo?
+
+No kubernetes repos other than [utils](https://github.com/kubernetes/utils)
```
timothysc: you'll absolutely need either the internal or external client.

contributors/design-proposals/testing/test-frameworks-repo.md
```diff
@@ -0,0 +1,81 @@
...
+should be imported (and hence vendored) by any code in this repo.
+
+This means that any testing framework in this repo which interacts with any
+part of kubernetes has to do so through some CLI. The [CLI integration testing
+framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test)
+which originally motivated this repo certainly does not have any kubernetes
+dependencies.
```
timothysc: That's not possible, but it is possible to ensure it's a proper DAG.

timothysc:
This falls in line with the subgroup proposed for sig-testing that solely focuses on tests, and code organization as well.

spiffxp:
@timothysc @stevekuznetsov I'd like to leave it to you two to decide how much of this needs to be hashed out before /lgtm'ing

In the meantime, I'm working on creating https://github.com/kubernetes-sig-testing to host a repo called "frameworks" per https://groups.google.com/forum/#!topic/kubernetes-sig-testing/8vacWVjIXCg

contributors/design-proposals/testing/test-frameworks-repo.md
```diff
@@ -0,0 +1,81 @@
...
+
+Some folks in Sig-CLI have been writing an [integration testing
+framework](https://github.com/kubernetes/kubectl/tree/master/pkg/framework/test)
+which they would like to use to simplify some of their integration tests which
+live in `k8s.io/kubernetes`. The framework currently lives in `k8s.io/kubectl`.
+Since `k8s.io/kubectl` already vendors components _from_ `k8s.io/kubernetes`, it 
+cannot be vendored _into_ `k8s.io/kubernetes`. 
+
+It therefore seems like a good idea to create a new repo to contain testing
+frameworks. The intent is that this repo could be vendored anywhere.
```
sttts: "anywhere" includes k/k?

sttts: If yes, this implies that it must not import anything from the staging repos, i.e. client-go, apimachinery, api, etc.

sttts: discussed below.

timothysc:
Per discussions, I'm ok with this, but I don't think we should merge this proposal and we should use this example to serve as a test case for @brendandburns proposal.

spiffxp:
@timothysc I guess I would prefer to see this proposal amended to reference the repo I'm creating per said proposal, so there is written documentation somewhere explaining its motivation.

totherme:
I've amended this proposal to reference the repo that @spiffxp is creating, and to clarify my understanding of what happened in the recent sig-architecture meeting.

timothysc:
Sure, but lets not merge this unless we have to, in preference for vetting on the official doc.

spiffxp:
>in preference for vetting on the official doc

@timothysc I'm not clear on what this means, can you elaborate?

marun:
@timothysc: I received emailed comments (dated Jan 9) from @sttts regarding how the new repo won't be able to rely on any staging repos if it is to be vendored into k/k, but I don't see those comments reflected here.  I think it may be worth reconsidering the use of an independent repo, and pivot to a repo staged out of k/k instead.  I would like to ensure improving test tooling for both k/k (where the bulk of testing work is currently done) and external repos (where the bulk of testing is likely to be done in the future), and the current approach of an independent repo can only hope to accomplish the latter.

timothysc:
So it already exists and the spec is in flight for a general-ish solution, closing.

PR #1525, opened 2017-12-18
Tombstone content moved to kubernetes/sig-release

spiffxp:
xref https://github.com/kubernetes/sig-release/pull/48


contributors/devel/release/patch-release-manager.md
```diff
@@ -1,258 +1 @@
...
-
-The Product Security Team (PST) will contact you if a security release is needed
-on your branch.
-In contrast to a normal release, you should not make any public announcements
-or push tags or release artifacts to public repositories until the PST tells you to.
-
-See the [Security Release Process](../security-release-process.md) doc for more
-details.
-
+This file has been migrated to https://github.com/kubernetes/sig-release/blob/master/release-process-documentation/release-team-guides/patch-release-manager-playbook.md
```
jdumars: https://github.com/kubernetes/sig-release/blob/master/release-process-documentation/release-team-guides/patch-release-manager-playbook.md is 404

contributors/devel/release/README.md
```diff
@@ -1,118 +1 @@
...
-2. Create a PR with documentation for your feature in the [documents repo](https://github.com/kubernetes/kubernetes.github.io).
-    * **Your PR should target the release branch (e.g. [`release-1.8`](https://github.com/kubernetes/kubernetes.github.io/tree/release-1.8)), not the [`master`](https://github.com/kubernetes/kubernetes.github.io/tree/master) branch.**
-      * Any changes to the master branch become live on https://kubernetes.io/docs/ as soon as they are merged, and for releases we do not want documentation to go live until the release is cut.
-3. Add link to your docs PR in the release tracking board, and notify the docs lead for the release (e.g. [Steve Perry](https://github.com/steveperry-53) for 1.8).
-4. The docs lead will review your PR and give you feedback.
-5. Once approved, the docs lead will merge your PR into the release branch.
-6. When the release is cut, the docs lead will push the docs release branch to master, making your docs live on https://kubernetes.io/docs/.
+This file has been migrated to https://github.com/kubernetes/sig-release/blob/master/ephemera/README.md
```
cblecker: Can we use https://git.k8s.io/sig-release/ephemera/README.md for cross-repo links?

cblecker:
Oh, also, can we set a deadline to remove these tombstones (as a comment or such in the files)?

contributors/devel/release/patch-release-manager.md
spiffxp: now that kubernetes/sig-release#48 has merged it's not

contributors/devel/release/README.md
spiffxp: will do

spiffxp:
I'll set a deadline in the files, but I think it'd be better to set a policy for tombstones as a project that could be automated.  Not going to solve that problem here today though.

Arbitrary guess for deadline: remove after a release cycle.  So, when 1.10 goes out the door, these tombstones can be removed.  I'm totally fine with something more aggressive too.  I couldn't tell that these were actually being linked to amongst repos, but I have no insight into traffic from other sources.

cblecker:
+1 to all of that. I'm good with the 1 release cycle for deletion. Plenty of time for the googles to update with new address, and any other links to be updated.

spiffxp:
I'll let the files call themselves "placeholders" instead of tombstones ✨

PR #1529, opened 2017-12-19
Update mount-options.md

xuhuilong:
yaml syntax correct.


gnufied:
@saad-ali @childsb can we merge this?

PR #1532, opened 2017-12-20
Added contributors/devel/help-wanted.md

alpha-san:

### DIFF
--- a/contributors/devel/help-wanted.md
+++ b/contributors/devel/help-wanted.md
@@ -1,0 +1,13 @@
+# Help Wanted
+
+Items marked with the `help wanted` label need to ensure that they are:
+
+- Sufficiently actionable: clear description of what should be done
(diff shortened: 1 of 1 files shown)


contributors/devel/help-wanted.md
```diff
@@ -0,0 +1,8 @@
+# Help Wanted
+
+Items marked with the `help-wanted` label need to ensure that they are:
```
cblecker: The label is actually `help wanted` now

cblecker: Please also add a bit at the end about the `/help` and `/remove-help` commands (as detailed here: https://git.k8s.io/test-infra/commands.md)

alpha-san: Okay cool. Updated help-wanted.md based on your feedback.

alpha-san:
@cblecker squashed :)

PR #1547, opened 2017-12-28
Proposal: Support specifying NodePort IP range

m1093782566:
/sig network

@kubernetes/huawei

### DIFF
--- a/contributors/design-proposals/network/nodeport-ip-range.md
+++ b/contributors/design-proposals/network/nodeport-ip-range.md
@@ -1,0 +1,85 @@
+# Support specifying NodePort IP range
+
+Author: @m1093782566
+
+# Objective
+
+This document proposes creating a option for kube-proxy to specify NodePort IP range.
+
+# Background
+
+NodePort type service gives developers the freedom to set up their own load balancers, to expose one or more nodes’ IPs directly. The service will be visible as the nodes's IPs. For now, the NodePort addresses are the IPs from all available interfaces.
+
+With iptables magic, all the IPs whose `ADDRTYPE` matches `dst-type LOCAL` will be taken as the address of NodePort, which might look like,
+
+```shell
+Chain KUBE-SERVICES (2 references)
+target     prot opt source               destination         
+KUBE-NODEPORTS  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service nodeports; NOTE: this must be the last rule in this chain */ ADDRTYPE match dst-type LOCAL
+```
+By default, kube-proxy accepts everything from NodePort without any filter. It can be a problem for nodes which has both public and private NICs, and people only want to provide a service in private network and avoid exposing any internal service on the public IPs. 
+
+# Proposal
+
+This proposal builds off of earlier requests to [[proxy] Listening on a specific IP for nodePort ](https://github.com/kubernetes/kubernetes/issues/21070), but proposes that we should find a way to tell kube-proxy what the NodePort IP blocks are instead of a single IP.
+
+## Create new kube-proxy configuration option
+
+There should be an admin option to kube-proxy for specifying which IP to NodePort. The option is a list of IP blocks, say `--nodeport-addresses`. These IP blocks as a parameter to select the interfaces where nodeport works. In case someone would like to expose a service on localhost for local visit and some other interfaces for particular purpose, an array of IP blocks would do that. People can populate it from their private subnets the same on every node.
+
+The `--nodeport-addresses` is defaulted to `0.0.0.0/0`, which means select all available interfaces and is compliance with current NodePort behaviour.
+
+If people set the `--nodeport-addresses` option to "127.0.0.0/8", kube-proxy will only select the loopback interface for NodePort.
+
+If people set the `--nodeport-addresses` option to "default-route", kube-proxy will select the "who has the default route" interfaces. It's the same heuristic we use for `--advertise-address` in kube-apiserver and others. 
+
+If people provide a non-zero IP block for `--nodeport-addresses`, kube-proxy will filter that down to just the IPs that applied to the node. 
+
+So, the following values for `--nodeport-addresses` are all valid:
+
+```
+0.0.0.0/0
+127.0.0.0/8
+default-route
+127.0.0.1/32,default-route
+127.0.0.0/8,192.168.0.0/16
+```
+
+
+And an empty string for `--nodeport-addresses` is considered as invalid.
+
+> NOTE: There is already a option `--bind-address`, but it has nothing to do with nodeport and  we need IP blocks instead of single IP.
+
+kube-proxy will periodically refresh proxy rules based on the list of IP blocks specified by `--nodeport-addresses`, in case of something like DHCP. 
+
+For example, if IP address of `eth0` changes from `172.10.1.2` to `172.10.2.100` and user specifies `172.10.0.0/16` for `--advertise-address`. Kube-proxy will make sure proxy rules `-d 172.10.0.0/16` exist. 
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/network/specify-nodeport-ip-range.md
```diff
@@ -0,0 +1,64 @@
...
+
+This proposal builds off of earlier requests to [[proxy] Listening on a specific IP for nodePort ](https://github.com/kubernetes/kubernetes/issues/21070), but proposes that we should find a way to tell kube-proxy what the NodePort IP blocks are instead of a single IP.
+
+## Create new kube-proxy configuration flag
+
+There should be an admin flag to kube-proxy for specifying which IP to NodePort. The flag is a list of IP blocks, say `--nodeport-addresses`. These IP blocks as a parameter to select the interfaces where nodeport works. In case someone would like to expose a service on localhost for local visit and some other interfaces for particular purpose, an array of IP blocks would do that. People can populate it from their private subnets the same on every node.
+
+The `--nodeport-addresses` is defaulted to `0.0.0.0/0`, which means select all available interfaces and is compliance with current NodePort behaviour.
+
+If people set the `--nodeport-addresses` flag to empty, kube-proxy will select the "who has the default route" + loopback interfaces. It's the same heuristic we use for `--advertise-address` in kube-apiserver and others.
```
thockin: Do we need this?  If we default to 0.0.0.0/0, why not make "" an error?  If anyone complains about it, we can consider something more heuristic like this (and maybe "default-route" is better than "", so `127.0.0.0/8,default-route` becomes a valid value)

contributors/design-proposals/network/specify-nodeport-ip-range.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Create new kube-proxy configuration flag
+
+There should be an admin flag to kube-proxy for specifying which IP to NodePort. The flag is a list of IP blocks, say `--nodeport-addresses`. These IP blocks as a parameter to select the interfaces where nodeport works. In case someone would like to expose a service on localhost for local visit and some other interfaces for particular purpose, an array of IP blocks would do that. People can populate it from their private subnets the same on every node.
+
+The `--nodeport-addresses` is defaulted to `0.0.0.0/0`, which means select all available interfaces and is compliance with current NodePort behaviour.
+
+If people set the `--nodeport-addresses` flag to empty, kube-proxy will select the "who has the default route" + loopback interfaces. It's the same heuristic we use for `--advertise-address` in kube-apiserver and others. 
+
+If people provide a non-zero IP block for `--nodeport-addresses`, kube-proxy will filter that down to just the IPs that applied to the node.
```
thockin: Will kube-proxy periodically refresh the list of interfaces and recreate rules, in case of something like DHCP?

contributors/design-proposals/network/specify-nodeport-ip-range.md
```diff
@@ -0,0 +1,64 @@
...
+## Kube-proxy implementation suport
+
+The implementation is simple.
+
+### iptables
+
+iptables support specify multiple IPs in the destination parameter(`-d`). For example, 
+
+```shell
+iptables -A PREROUTING -d 192.168.1.5,192.168.1.6 -p tcp --dport 22 -j ACCEPT
```
thockin: Why not use the CIDRs as specified by the user?  Only the special default-route would need to be expanded, and that is IF we need it.

contributors/design-proposals/network/specify-nodeport-ip-range.md
m1093782566: I was seeking a way to express "default route". I agree an explicitly word, e.g. `default-route` you mentioned is better than "" to express "default route". THEN, we may need to modify the definition for `--nodeport-addresses` - it's a string list, valid values can either be IP block or "default-route".

So, the following values for `--nodeport-addresses` are all valid:

```
0.0.0.0/0
127.0.0.0/8
default-route
127.0.0.1/32,default-route
127.0.0.0/8,192.168.0.0/16
```

contributors/design-proposals/network/specify-nodeport-ip-range.md
m1093782566: > Will kube-proxy periodically refresh the list of interfaces and recreate rules

Sorry, `--advertise-address` for kube-proxy is a list of IP blocks instead of interface names.

kube-proxy will periodically refresh the rules created based on `--advertise-address`. However, for example, if IP address of `eth0` changes from `172.10.1.2` to `192.168.3.4` and users only specify `172.10.0.0/16` for `--advertise-address`. Kube-proxy will NOT create proxy rules for `192.168.3.4` unless `eth0` has the default route.

When refer to DHCP user case, network administrator usually reserves a RANGE of IP addresses for the DHCP server. So, IP address change will always fall in an IP range? That's to say an IP address of a interface will not change from `172.10.1.2` to `192.168.3.4` in DHCP scenario?

contributors/design-proposals/network/specify-nodeport-ip-range.md
m1093782566: > Why not use the CIDRs as specified by the user?

It's a good implementation way as well :+1: We can make use of IP blocks directly in iptables command. However, we need to use individual IP in IPVS proxier.

> Only the special default-route would need to be expanded, and that is IF we need it.

Yeah, that's true.

m1093782566:
@thockin 

All comments are addressed or clarified. Please to see if you are happy with the changes when you have a chance.

contributors/design-proposals/network/specify-nodeport-ip-range.md
```diff
@@ -0,0 +1,85 @@
+# Support specifying NodePort IP range
+
+Author: @m1093782566
+
+# Objective
+
+This document propose creating a flag for kube-proxy to specify NodePort IP range.
```
dankenigsberg: English nit: propose -> proposes

contributors/design-proposals/network/specify-nodeport-ip-range.md
```diff
@@ -0,0 +1,85 @@
...
+target     prot opt source               destination         
+KUBE-NODEPORTS  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service nodeports; NOTE: this must be the last rule in this chain */ ADDRTYPE match dst-type LOCAL
+```
+By default, kube-proxy accepts everything from NodePort without any filter. It can be a problem for nodes which has both public and private NICs, and people only want to provide a service in private network and avoid exposing any internal service on the public IPs. 
+
+# Proposal
+
+This proposal builds off of earlier requests to [[proxy] Listening on a specific IP for nodePort ](https://github.com/kubernetes/kubernetes/issues/21070), but proposes that we should find a way to tell kube-proxy what the NodePort IP blocks are instead of a single IP.
+
+## Create new kube-proxy configuration flag
```
dankenigsberg: another English nit: I don't think that "flag" fits here. A "flag" is either raised or not. Here you suggest a more complex command-line *option*.

contributors/design-proposals/network/specify-nodeport-ip-range.md
m1093782566: Fixed now.

contributors/design-proposals/network/specify-nodeport-ip-range.md
m1093782566: Sound fair. It's fixed now. Thanks!

thockin:
I'm approving, though the impl may change just a bit (discussing the need for the "all")

/lgtm
/approve

m1093782566:
Okay, we can fix it up in follow-ups.

PR #1564, opened 2018-01-05
Update scheduler code locations.

misterikkit:
The scheduler code was moved in [kubernetes #57852](https://github.com/kubernetes/kubernetes/pull/57852), and we should update docs do reflect the new location.
 
/sig scheduling
/sig docs


misterikkit:
/hold
until [kubernetes #57852](https://github.com/kubernetes/kubernetes/pull/57852) merges

liggitt:
Hold until it's released, so 1.10, right?

misterikkit:
@liggitt Hmm, the state of master will be out of sync with community docs for some time, unless I misunderstand the release process.

liggitt:
ah, these are referencing code paths in master... you are right, it doesn't need to wait for release

cblecker:
/ok-to-test
Yeah, this should be fine once https://github.com/kubernetes/kubernetes/pull/57852 is merged.

misterikkit:
I think someone with write access needs to explicitly merge this?

Issue #1407, opened 2017-11-15
Document List of Organizations


mattfarina:
I was attempting to look for a list of the GitHub organizations that are part of Kubernetes and couldn't find any form list. This caught my attention because of CLA checking and formalities. From the Prow configuration I captured:

* kubernetes
* kubernetes-incubator
* kubernetes-security

There are other organizations that exist including, but not limited to:

* kubernetes-ui
* kubernetes-client
* kubernetes-up-and-running
* kubernetes-incubator-retired
* kubernetes-digitalocean-terraform
* kubernetes-helm
* kubernetes-contrib

And... possibly others.

Are there any others aside from the top 3 that are part of the Kubernetes project?

cblecker:
cc: @kubernetes/steering-committee

dankohn:
For [DevStats](devstats.k8s.io), we do the following.

For data starting from 2015-01-01 it is:
https://github.com/cncf/devstats/blob/master/projects.yaml
Which lists organizations: "kubernetes,kubernetes-client,kubernetes-incubator,kubernetes-helm"
Before 2015-01-01 this is a special config:
https://github.com/cncf/devstats/blob/master/kubernetes/kubernetes.sh
It lists single repos (not orgs): "GoogleCloudPlatform/kubernetes,kubernetes,kubernetes-client"
The same repos "GoogleCloudPlatform/kubernetes,kubernetes,kubernetes-client" are also included for 2015-01-01 - 2015-08-14

If we're missing some, we'll correct.

bgrant0607:
kubernetes-ui, kubernetes-up-and-running, kubernetes-digitalocean-terraform, and kubernetes-contrib are not ours.

spiffxp:
I've been using https://github.com/kubernetes/community/issues/1527 to try and cover all repos that are part of this project.  Thus far I have opened PR's for all repos in:
- [kubernetes](https://github.com/kubernetes)
- [kubernetes-client](https://github.com/kubernetes-client)
- [kubernetes-helm](https://github.com/kubernetes-helm)
- [kubernetes-incubator](https://github.com/kubernetes-incubator)

Based on discussion above and some educated guesses from [this github user search](https://github.com/search?p=3&q=kubernetes-&type=Users&utf8=%E2%9C%93) it seems like I'm missing:
- [kubernetes-csi](https://github.com/kubernetes-csi)
- [kubernetes-graveyard](https://github.com/kubernetes-graveyard)
[...]

PR #1569, opened 2018-01-05
Adding initial directions for kubernetes organization owners

mattfarina:
cc @kubernetes/steering-committee @idvoretskyi 

I think I have this right but I'm not sure. After some conversations in December I wanted to capture what I'd learned in writing.

### DIFF
--- a/org-owners-guide.md
+++ b/org-owners-guide.md
@@ -1,0 +1,54 @@
+# Kubernetes Github Organization Guide
+
+The Kubernetes project leverages multiple GitHub organizations to store and
+organize code. This guide contains the details on how to run those organizations
+for CNCF compliance and for the guidelines of the community.
+
+## Organization Naming
+
+Kubernetes managed organizations should be in the form of `kubernetes-[thing]`.
+For example, [kubernetes-client](https://github.com/kubernetes-client) where the
+API clients are housed.
+
+Prior to creating an organization please contact the steering committee for
+direction and approval.
+
+Note: The CNCF, as part of the Linux Foundation, holds the trademark on the
+Kubernetes name. All GitHub organizations with Kubernetes in the name should be
+managed by the Kubernetes project or use a different name.
+
+## Transferring Outside Code Into A Kubernetes Organization
+
+Due to licensing and CLA issues, prior to transferring software into a Kubernetes
+managed organization there is some due diligence that needs to occur. Please
+contact the steering committee and CNCF prior to moving any code in.
+
+It is easier to start new code in a Kubernetes organization than it is to
+transfer in existing code.
+
+## Current Organizations In Use
+
+The following organizations are currently known to be part of the Kubernetes
+project:
+
+* [kubernetes](https://github.com/kubernetes)
+* [kubernetes-client](https://github.com/kubernetes-client)
+* [kubernetes-csi](https://github.com/kubernetes-csi)
+* [kubernetes-graveyard](https://github.com/kubernetes-graveyard) †
+* [kubernetes-helm](https://github.com/kubernetes-helm)
+* [kubernetes-incubator](https://github.com/kubernetes-incubator)
+* [kubernetes-retired](https://github.com/kubernetes-retired)
+* [kubernetes-security](https://github.com/kubernetes-security)
+
+† kubernetes-retired should be used instead of kubernetes-graveyard going forward.
+
+Note, this list is subject to change.
+
+There are more organization names that we are squatting on with possible future
+intentions. [For more details please see community issue #1407](https://github.com/kubernetes/community/issues/1407).
+
+## Repository Guidance
+
+Repositories have additional guidelines and requirements, such as the use of
+CLA checking on all contributions. For more details on those please see the
+[Kubernetes Template Project](https://github.com/kubernetes/kubernetes-template-project).


spiffxp:
/assign
/committee steering
I'll look at this today or next week, I'm now sitting on a number of kubernetes orgs and would like to make sure what I'm doing is correct and documented here

ref: #1407

idvoretskyi:
/assign

@spiffxp I'd like to work on this as well from the CNCF side.

org-owners-guide.md
```diff
@@ -0,0 +1,66 @@
+# GitHub Organization Owners Guide
+
+The Kubernetes project leverages multiple GitHub organizations to store and
+organize code. This guide contains the details on how to run those organizations
+to for CNCF compliance and for the guidelines of the community.
+
+## Organization Naming
+
```
bgrant0607: I agree with this section.

org-owners-guide.md
```diff
@@ -0,0 +1,66 @@
...
+Prior to creating an organization please contact the steering committee for
+direction and approval.
+
+Note: The CNCF, as part of the Linux Foundation, holds the trademark on the
+Kubernetes name. All GitHub organizations with Kubernetes in the name should be
+managed by the Kubernetes project or use a different name.
+
+## Contributor License Agreement (CLA)
+
+All code contribute to the Kubernetes project needs to be done by those who are
```
bgrant0607: This is transitioning from orgs to repos, for which we have other docs:

https://github.com/kubernetes/community/blob/master/governance.md#repository-guidelines
https://github.com/kubernetes/community/blob/master/governance.md#cla
https://github.com/kubernetes/community/blob/master/governance.md#code-of-conduct
https://github.com/kubernetes/community/blob/master/incubator.md#creation-of-the-incubator-project
https://github.com/kubernetes/kubernetes-template-project

Maybe just link to them from here? Or otherwise consolidate this content somehow?

org-owners-guide.md
```diff
@@ -0,0 +1,66 @@
...
+repository should be [the file found in the Kubernetes repo](https://github.com/kubernetes/kubernetes/blob/master/code-of-conduct.md).
+
+## Software License
+
+All software that is part of the Kubernetes project needs to be licensed under
+the Apache 2 License.
+
+## Transferring Outside Code Into A Kubernetes Organization
+
+Due to licensing and CLA issues, prior to transferring software into a Kubernetes
```
bgrant0607: This is fine for now.

org-owners-guide.md
```diff
@@ -0,0 +1,66 @@
...
+* kubernetes
+* kubernetes-incubator
+* kubernetes-security
+* kubernetes-client
+* kubernetes-incubator-retired
+* kubernetes-helm
+* kubernetes-graveyard
+* kubernetes-csi
+
+This list is likely incomplete. If you are aware of additions please contribute
```
bgrant0607: I believe the list is still being hashed out elsewhere.

org-owners-guide.md
mattfarina: @bgrant0607 Since the incubation process is not in use and repos, such as those under kubernetes-client, are not part of the incubator, I would suggest we move the requirements for orgs and repos out of there and reference back here. Going forward it appears we will have multiple types of things.

org-owners-guide.md
spiffxp: We've since hashed this list out https://github.com/kubernetes/community/issues/1407#issuecomment-357114370

org-owners-guide.md
spiffxp: CLA is activated/configured per-org AFAIK, that should be in this doc about kubernetes orgs.

Code of Conduct, Software License are are repo-level specs.  I don't think they belong in this doc. 
 I think it would be simpler to say all repos in a kubernetes org must adhere to the standards 
 set by https://github.com/kubernetes/kubernetes-template-project and make sure that's a living example (eg: I'm about go to PR out the wording about incubator)

bgrant0607: I like Aaron's proposal.

spiffxp: Opened https://github.com/kubernetes/kubernetes-template-project/pull/14

mattfarina: There are two things that catch my attention here.

1. We already have numerous existing projects not based on this template. Those managing those repos and the orgs they are on need clear guidance on what they MUST do apart from what's RECOMMENDED. Many of those org and repo owners don't have those details. This document was an attempt to serve existing as well as new. How do we concisely communicate to these folks?
1. The template repo has a lot of implementation opinions. For example, all maintainers must LGTM a release and for dependencies you use godep. Where is the line between what one MUST do and what's RECOMMENDED? The template project doesn't seem to share that.

Shouldn't a template implement a set of guidance to be helpful rather than be the source of truth? At least for all those cases of communication that a template isn't well suited for (e.g., after day-0 project creation)?

spiffxp: I guess if we're going to go to full blown RFC compliant wording, I'd still prefer to see that wording over in kubernetes/kubernetes-template-project.  I am open to PR's that clearly demarcate opinions from requirements.

As far as reconciling the state of kubernetes repos today with the desired state is concerned, I see no difference between updating a living example after the fact and updating docs after the fact.

mattfarina: @spiffxp so I can update the wording here as a pointer. Repo maintainers need to:

1. Watch for changes on the template repo
1. Figure out the ones that apply to them
1. Apply them to their own repos

What I don't know is what parts of the example apply to all repo and which don't. Some things are required, such as the code of conduct, which other things are optional, such as the use of godep (glide, dep, and godep are in use right now). How do repo maintainers distinguish the types of things? How do we communicate that as guidance?

spiffxp: @mattfarina I think clear wording in https://github.com/kubernetes/kubernetes-template-project/blob/master/README.md would help distinguish what is required vs. recommended.  For example, we don't say "use _this_ CONTRIBUTING.md file that references godep", we say "use _a_ CONTRIBUTING.md file" and provide an example.

Communicating this guidance would probably best take place on the kubernetes-dev mailing list as we come to a consensus on when and how best to enforce project-wide standards

org-owners-guide.md
```diff
@@ -0,0 +1,66 @@
+# GitHub Organization Owners Guide
```
spiffxp: Could we retitle this Kubernetes Github Organizations?

org-owners-guide.md
```diff
@@ -0,0 +1,66 @@
...
+Kubernetes managed organizations should be in the form of `kubernetes-[thing]`.
+For example, [kubernetes-client](https://github.com/kubernetes-client) where the
+API clients are housed.
+
+Prior to creating an organization please contact the steering committee for
+direction and approval.
+
+Note: The CNCF, as part of the Linux Foundation, holds the trademark on the
+Kubernetes name. All GitHub organizations with Kubernetes in the name should be
+managed by the Kubernetes project or use a different name.
```
michelleN: ^ is that something we can actually enforce? I'm not pushing back here, just curious.

org-owners-guide.md
michelleN: +1 on pointing to `kubernetes-template-project` and communicating guidance on what repo maintainers need to do via the kubernetes-dev mailing list for now.

What do we think about having a `Guide to Upgrading Your Kubernetes Repo` doc that you can link here with the directions Matt Farina mentioned and use that as the guidance we send to the `kubernetes-dev` mailing list?

org-owners-guide.md
mattfarina: @michelleN if the CNCF doesn't enforce the trademark on the name than it can loose the ability to do this. It's a general thing that you need to enforce usage on a name or you loose the right to. It's legal and as I'm not a lawyer I don't really know the details. I'm sure the Linux Foundation and CNCF folks know what to do, though.

Because the CNCF owns the name there are channels through GitHub to deal with this if it were to come to that (which it usually doesn't).

mattfarina:
@spiffxp and others... the document has had a bunch of updates including:

* New title
* Pointer for repo to the template project
* Updated list of orgs with links
* Removed repo guidance

spiffxp:
/lgtm
/hold
holding for comment from other steering committee members

org-owners-guide.md
```diff
@@ -0,0 +1,52 @@
...
+managed organization there is some due diligence that needs to occur. Please
+contact the steering committee and CNCF prior to moving any code in.
+
+It is easier to start new code in a Kubernetes organization than it is to
+transfer in existing code.
+
+## Current Organizations In Use
+
+The following organizations are currently known to be part of the Kubernetes
+project:
```
timothysc: The steering committee will be pushing out a doc soon regarding organizational structure which will likely cause the structure below to be reformed.

mattfarina: Does that need to block this PR? Can we update the doc after those changes come out?

timothysc: I'd put a clause and change the content saying "this is all subject to change", but here is the current state, then lgtm from me.

mattfarina: Added that clause

PR #1574, opened 2018-01-07
Fix inline link markup

techtonik:


stp-ip:
Thanks for your contribution. This fixes the link, but breaks the italic style. Either we remove it completely or don't break it.

techtonik:
Oh. I didn't notice that. Fixed.

stp-ip:
+1
/ok-to-test
/lgtm

PR #1575, opened 2018-01-08
Proposal: Add udp portforwarding support design docs

shiywang:
Releated to https://github.com/kubernetes/kubernetes/issues/47862
@smarterclayton @ncdc @deads2k @mfojtik  @kargakis @pavolloffay @jbeda @kubernetes/api-reviewers and @kubernetes/api-approvers any one would like to take a look or sponsor on this ? 
  
@ncdc part of the description content I copied from your original design doc of port-forwarding, hope you don't mind : )


contributors/design-proposals/api-machinery/udp-portforwarding.md
```diff
@@ -0,0 +1,109 @@
...
+- port forwarding to containers with UDP protocol
+
+## Background && Motivation
+
+Fixes user request of https://github.com/kubernetes/kubernetes/issues/47862
+
+
+## Constraints and Assumptions
+
+- We already have an implemention with SPDY for TCP, the UDP implement basically reuse the SPDY framework, so it would be UDP through SPDY, the change here is remove the stream copy since UDP doesn't not has the same stream attribute as SPDY/TCP, so UDP will use datagram and `write` and `send` to communicate, socat will use unix domain socket as a middle layer to cross the boundaries between pod and Kubelet. and since TCP/SPDY doesn't have a message boundaries but UDP does, we also need to design an simple protocol for how to divide message read from SPDY back and forth.
```
ncdc: Please break into multiple lines if you don't mind.

Here are some edits to split up long sentences and make it easier to understand. Feel free to accept/reject as you see fit.

We already have a port forwarding implementation for TCP using SPDY. TCP is connection-based and supports streaming, so it's easy to identify individual connections and we can use `io.Copy` to transfer the streams of data between all the components involved (local client, kubectl, apiserver, kubelet/CRI, pod). Unlike TCP, UDP is connectionless and it uses messages instead of a stream of data. To add support for UDP, we will continue to use SPDY, but we will not be able to use `io.Copy` for UDP datagram transmission. Instead, we will design a simple protocol to allow UDP message encapsulation over SPDY. Additionally, we will use a Unix domain socket as the means for transferring UPD messages between the kubelet/CRI and the pod.

contributors/design-proposals/api-machinery/udp-portforwarding.md
```diff
@@ -0,0 +1,109 @@
...
+## Use Cases
+
+- As a user of a Kubernetes cluster, I want to be able to connect to local ports on my computer and have them forwarded to ports in the container with UDP.
+
+- Basically like this: `kubectl port-forward pod-name --protocol=udp   8888:10001`
+
+## Process Flow
+
+### Port Forwarding Flow
+1. The Kubectl connects to the Kubernetes Master to initiate a remote command execution
```
ncdc: s/Master/apiserver/
s/remote command execution/port forwarding/

contributors/design-proposals/api-machinery/udp-portforwarding.md
```diff
@@ -0,0 +1,109 @@
...
+request.
+2. The Master proxies the request to the Kubelet where the container lives.
```
ncdc: s/Master/apiserver/

contributors/design-proposals/api-machinery/udp-portforwarding.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Backward Compatibility
+
+#### old kubectl - new api-server/kubelet
+ 
+Fully compatible with out compatibility design.
+
+#### new kubectl - old api-server/kubelet
+
+Add an `PortForwardProtocolV2Name` in client side, as well as deal it in api-server side, so the old server deosn't recognize will throw an error.
```
ncdc: Add a new port forwarding protocol, `v2.portforward.k8s.io`, that will indicate that UDP port forwarding is supported. If a newer kubectl requests `v2.portforward.k8s.io` and the apiserver does not support it, the apiserver returns an error.

contributors/design-proposals/api-machinery/udp-portforwarding.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Customized Protocol
+
+The customized protocol is pretty simple here, it contains two parts: the content length and actual data.
+
+![simple protocol](udp2.png)
```
ncdc: I would do this as a markdown table instead of an image.

ncdc: It might be worth noting that traditional UDP datagrams have a maximum length of ~ 65535 bytes (i.e. a 2-byte unsigned integer), whereas the length field of a jumbogram may use up to 4 bytes (2^32 - 1).

contributors/design-proposals/api-machinery/udp-portforwarding.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+### Streaming Protocol
+
+The current multiplexed streaming protocol used is SPDY.
```
ncdc: We would prefer to move everything away from SPDY and over to either websockets or HTTP/2. I don't think your proposal needs to be impacted by this desire, though. In other words, if we decide to proceed with UDP port forwarding, we can implement this design and then replace it with websockets or HTTP/2 along with everything else (TCP port forwarding, exec).

contributors/design-proposals/api-machinery/udp-portforwarding.md
```diff
@@ -0,0 +1,109 @@
...
+The customized protocol is pretty simple here, it contains two parts: the content length and actual data.
+
+![simple protocol](udp2.png)
+
+
+### Streaming Protocol
+
+The current multiplexed streaming protocol used is SPDY. 
+
+### Master as First Level Proxy
```
ncdc: I think you can remove everything from here down to the end, since it's largely a copy from https://github.com/kubernetes/community/blob/master/contributors/design-proposals/network/command_execution_port_forwarding.md and we already have port forwarding and remote command execution implemented.

contributors/design-proposals/api-machinery/udp-portforwarding.md
smarterclayton: Our goal with port-forward/exec is to remove SPDY support across the board and replace it with web sockets.  However, the websockets implementation for port-forward is not scalable (it cannot multiplex multiple streams like SPDY) and therefore cannot be used.  I don't want to add anything to exec/port-forward until we fix the websocket port forward issue by implementing a new multiple stream sub-protocol and deprecate SPDY.

I don't want to add anything new to SPDY.

contributors/design-proposals/api-machinery/udp-portforwarding.md
smarterclayton: I would go even further - I don't think we should even add anything to our streaming protocols until we get off SPDY and onto a supportable websocket base.  SPDY is dead.

contributors/design-proposals/api-machinery/udp-portforwarding.md
shiywang: thank you, this is much better : ) will update it.

contributors/design-proposals/api-machinery/udp-portforwarding.md
shiywang: done

contributors/design-proposals/api-machinery/udp-portforwarding.md
vfreex: Hi @smarterclayton I am also interested in extending the port-forwarding and other streaming functionalities. What's the status of websocket multiplexing sub-protocol? Any blockers or anything we can help with this effort?
I noticed that there is already minimum multiple stream sub-protocol support in current implementation of api-server and kubelet, but the number of streams, aka channels, is negotiated during handshaking and a stream is identified by the first byte of a frame. Am I correct?

vfreex:
Hi @smarterclayton I am also interested in extending the port-forwarding and other streaming functionalities. What's the status of websocket multiplexing sub-protocol? Any blockers or anything we can help with this effort?
I noticed that there is already minimum multiple stream sub-protocol support in current implementation of api-server and kubelet, but the number of streams, aka channels, is negotiated during handshaking and a stream is identified by the first byte of a frame. Am I correct?

contributors/design-proposals/api-machinery/udp-portforwarding.md
shiywang: Yes, from practical perspective, package length more than MTU length will be dropped, and even we take ipv6 into consideration (16-bits payload length), so I think `uint32` is enough.

vfreex:
Currently the interfaces for SPDY and WebSocket connections are not the same.
 Is it possible to define an interface for underlying multiplexed connection so that port-forward will not worry about the protocol?
This is an example (inspired by SPDY connection):
```go
// A Header represents the key-value pairs in a channel request.
type Header map[string][]string

type Channel io.ReadWriteCloser

type NewChannelRequest interface {
	GetHeader() Header
	Accept() Channel
}
type MultiplexedConnection interface {
        // initiate a new channel (stream)
	CreateChannel(header Header) (Channel, error)
        // returns a Go channel for new channels received from the other end
	NewChannelChan() <-chan NewChannelRequest
	Close() error
	CloseChan() <-chan interface{}
	SetIdleTimeout(timeout time.Duration) error
}
```

lavalamp:
I am not excited about apiserver doing this proxying. I would rather see a separate "bastion" component for bridging dev traffic into clusters. I would like to remove the proxy we have right now, not add more types.

ncdc:
@lavalamp I completely agree that we need to remove the proxying from the apiserver for things like exec and port forwarding. However, I would think something like this could be developed independently/simultaneously of that effort, without causing any additional interdependencies.

kevinsimard:
Any news?

dims:
/assign @deads2k
/assign @lavalamp 

There seems to be lot of 👍 on https://github.com/kubernetes/kubernetes/issues/47862 can you both please look?

PR #1580, opened 2018-01-08
adds additional approvers and reviewers to KEPs OWNERS

calebamiles:
Adds additional people who are active on iterating on the KEP process
as OWNERS to the extracted `kep` directory.

### DIFF
--- a/keps/OWNERS
+++ b/keps/OWNERS
@@ -1,8 +1,16 @@
 reviewers:
   - sig-architecture-leads
   - jbeda
+  - bgrant0607
+  - jdumars
+  - calebamiles
+  - idvoretskyi
 approvers:
   - sig-architecture-leads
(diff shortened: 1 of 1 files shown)


calebamiles:
@jbeda, @idvoretskyi, @jdumars PTAL. I'm also not sure where `sig-architecture-leads` is defined. It doesn't seem to be in the [alias file][], unless I just missed a change in its location or purpose, and I can't see any GitHub team with that name either.


[alias file]: https://github.com/kubernetes/kubernetes/blob/master/OWNERS_ALIASES

cblecker:
@calebamiles https://github.com/kubernetes/community/blob/4d715ddf08fc31e7d2b8bc247c55fe8ee9bf7f06/OWNERS_ALIASES#L9

PR #1581, opened 2018-01-09
Design document for gRPC-based KMS plugin service

wu-qiang:
This document describes the gRPC interface between API server and out-of-tree KMS provider.
* Define the protobuf file for the gRPC interface
* Communicate with unix domain socket, not support TCP, so no TLS support
* The configuration file


wu-qiang:
@destijl Should this document be a part of https://github.com/kubernetes/community/pull/607?

deads2k:
This was open as google doc for quite some time and discussed multiple times at sig-auth. There were no dissenters in either spot.  I plan to tag tomorrow unless some speaks up today.


lgtm

PR #128, opened 2016-12-01
Proposal for refactoring cloud provider API

wlan0:
@thockin @luxas @roberthbailey @chrislovecnm @liggitt @saad-ali @erictune 

This PR is a copy of the rectified contents of this PR - https://github.com/kubernetes/kubernetes/pull/37037

PTAL

PR #1583, opened 2018-01-09
move cloud provider kep

calebamiles:
this is work that was already approved under WG Cloud Provider. Mark the KEP as `approved` and give it a number.


cblecker:
@calebamiles This was marked "work-in-progress" due to the PR title. Would you mind editing it if it's not WIP?

jbeda:
In my mind we are still figuring out what "approved" means here.  This is part of the process that isn't nailed down and is part of the KEP process that I consider still as alpha.

Specifically, I'd expect that we'd have a list of approvers in the metadata (it is still marked TBD) and those are the folks that would make the call to mark it as approved.

I'm fine moving this and leaving it in draft while we work through that process.

calebamiles:
@jbeda, this KEP covers work already underway by WG Cloud Provider. There's nothing to "approve" in this case; their use of the process at all was an experiment for our behalf and I'm, personally not interested in holding them up for administrivia. As @cheftako finds additional people to shepherd the work, which is already in flight, then I imagine the KEP will be updated to reflect this information. As the working group rolls up to SIG API Machinery (@kubernetes/sig-architecture-feature-requests), I leave it up to them to comment.

**cc: @lavalamp, @deads2k**

jbeda:
Sure -- perhaps one of the things that we need to clarify is if a KEP is at a SIG scope (something like this) or if it is at a project wide (or SIG-arch) scope.

Regardless -- as part of the process I think it probably makes sense for (a) the set of reviewers to be separate from the authors and (b) the approvers to be drawn from the owning SIG.

@lavalamp @deads2k -- are you up for signing on as approver for this KEP and having your name listed?  Do you think that this is in a state ready to be "approved"?

jdumars:
I'm with Caleb on this.  I think we need to pilot the mechanics of the KEP template itself, and not gate this work. There will be other opportunities to pilot the approval process.

As I understand it, the KEP is primarily a _design doc_ with "just enough" review to ensure it is actionable, consistent with the overall architectural digest, and relevant to the Kubernetes ecosystem. It should be "This work is consistent with the mission/charter of the sponsor SIG, is detailed enough to be actionable, and it makes sense in the overall ecosystem."  If any of those are a no, the KEP authors should be able to adjust scope, or in the worst case withdraw it.  The cloud provider breakout has already undergone all of these sanity checks.

jbeda:
I'm not trying to gate here.  But if we mark something as "approved" then there has to be someone doing the approving. My question here is "who is doing the approving" in this case?

Marking a KEP as "approved" is a mark that there is agreement that this is the way forward.  It seems strange to me that this is being bundled into a PR that is about just moving the location of the KEP.

By all means move the KEP into this directory and give it a number.  But I'd love to see someone stand up and say "I approve" before we mark it as approved.

jbeda:
FWIW, the KEP process itself (KEP-1) is still marked `draft`.

idvoretskyi:
> As I understand it, the KEP is primarily a design doc with "just enough" review to ensure it is actionable, consistent with the overall architectural digest, and relevant to the Kubernetes ecosystem. 

Yes, I second this.

calebamiles:
@jbeda this particular work is _already_ approved. One of the reasons why there is a cloud provider working group is to shepherd the work described here under the auspices of SIG API Machinery. Marking the KEP as `approved` in this case is a formality. Thank you for [updating the KEP process explanation][] and I think there's some good discussion to be had there. Do you have any reason why _this_ PR shouldn't be merged as is?

[updating the KEP process explanation]: https://github.com/kubernetes/community/pull/1584

lavalamp:
Meta-meta point: @jbeda, people will probably have a better experience using KEPs--and you'll get more a more accurate view of how people will understand the template--if you just observe and fix the problems you see by modifying the template for the next users, rather than raising meta points in the PRs attempting to use the process, like this one.

Meta points, since the can of worms is already opened: maybe calling it "plan-of-record" instead of "approved" would give everyone what they want? Also it shouldn't have to be approved to have an official number. I don't want to put my name on an official approvers list right now only because that would require me to (re?) read this, which would drag this out unnecessarily. I am happy to trust the WG and if that causes abuse later, fix it then.

smarterclayton:
Would changing the state to "sig-approved" and putting comment `# needs to be updated when KEP is finalized` be sufficient?  Is the biggest concern that there be an approver who can be blamed, or that there isn't a non-sig approver?

jbeda:
I think that there is some question as to what "approved" means here.  Does it mean "The problem space is something we are working on" or does it mean "The solution outlined in the KEP is agreed upon to be the solution we are going to execute on"?

I've been assuming the second meaning but perhaps that is where the confusion is.

From my point of view, if @lavalamp is happy to put his name down as an approver and say that this is a done deal then I'm good.  I don't think that we have to separate out "sig approved" from some level of higher level "approved".  But my understanding is that the spirit of approved is that someone besides the authors have taken a look at the plan and given it a thumbs up.

calebamiles:
As this is a rewrite of an existing design proposal for the cloud provider extraction, I believe your criteria has been met @jbeda. I've only added myself as an author in order to provide minor editorial support to @cheftako.

jbeda:
I'd love to have @lavalamp (or someone else from the owning SIG) added to the approvers list and give a thumbs up here.

smarterclayton:
I'm an approver in that SIG - but I would like someone less busy than me be the approver - @kubernetes/sig-api-machinery-api-reviews

calebamiles:
so @jbeda can we add the `approvers` later, deferring to WG Cloud Provider to find those people as suggested by @lavalamp.

jbeda:
Sure -- let's leave the state as "draft" and have the approve mark it "approved" once the approver is found.

calebamiles:
I don't think that's what I'm suggesting, @jbeda. I'm suggesting that this work is _already approved_ per prior discussions between SIG {API Machinery, AWS, GKE, Azure, Cluster Lifecycle, Node, Storage} with one of the outcomes being the creation of a WG Cloud Provider that has as part of its [charter][] to undertake the work described in this KEP. This KEP _specifically_ was created to trial this format for describing changes and describes work which has already been agreed to as their is an [existing design proposal][] which covers this work.

[charter]: https://docs.google.com/document/d/1m4Kvnh_u_9cENEE9n1ifYowQEFSgiHnbw43urGJMB64/edit#heading=h.rkwrsiw228ev  
[existing design proposal]: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/cloud-provider/cloud-provider-refactoring.md

jbeda:
I'm super confused.  This PR changes it from `draft` to `approved` yet we don't have any names in the `approvers` field.  It has "been approved" but no one has named those that have approved beyond just naming SIGs.  A lead from the owning SIG (@smarterclayton) says that he wants someone else to sign up. On top of that you are happy to have me submit another change after this PR to change it back from `approved` to `draft`.

This just doesn't make any sense to me.  We should either name approvers and have them give it a thumbs up (should be easy if it is already approved) or we should mark it as draft.  I'm not sure why we have to transition through the `approved` state on our way back to `draft`.

I don't think it is at all unreasonable to expect that if something is marked as `approved` there are names down as `approvers`.

calebamiles:
@jbeda, I've added @thockin as the `approver` for this KEP as he [approved][] the original design proposal and has reviewed the work of @cheftako as he rewrote that proposal as a KEP. Given that he is unlikely to review the PRs which implement the KEP, I imagine that someone from WG Cloud Provider will update the `approvers` field as they are able

[approved]: https://github.com/kubernetes/community/pull/128#pullrequestreview-18038563

jbeda:
@thockin -- can you give a thumbs up here that you are cool with this being marked as `approved` as a KEP?

Also, I'd prefer the dates to stay in the metadata.  That is currently in the template and the discussions to change the template are still in flight.  I don't think we've come to a conclusion there yet.

lavalamp:
I think I figured out what happened in this thread-- Caleb thinks the
approvers are people who will be approving the implementation PRs, whereas
I (and probably most everyone else?) thinks approvers are the people who
approved the KEP. This should be clarified in the KEP template.

jbeda:
@lavalamp I think you are right -- there is some squishiness in the definitions here.  I'm trying to clarify in https://github.com/kubernetes/community/pull/1584#issuecomment-356748129 and we are going to try and chat about it in SIG-Architecture tomorrow morning.  Feel free to join us if you want to be involved.

bgrant0607:
@thockin PTAL and approve if you are happy with this.

PR #1584, opened 2018-01-10
Update KEP process

jbeda:
* Update KEP-1 metadata to match template
* Tighten up some of the language of KEP-1
* Clarify some sections based on early feedback of KEP process.
* Align stages and metadata description with template.

Signed-off-by: Joe Beda <joe.github@bedafamily.com>


idvoretskyi:
cc @jdumars

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -1,21 +1,11 @@
-# Kubernetes Enhancement Proposal Process
-
-## Metadata
```
calebamiles: nit: I don't think we should format the Metadata so it's a Markdown table because it can require horizontal scrolling in many cases when rendered and doesn't really make it all that much easier to extract with automation

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -5,7 +5,8 @@ authors:
   - "@calebamiles"
   - "@jbeda"
 owning-sig: sig-architecture
-participating-sigs: []
+participating-sigs:
+  - kubernetes-wide
```
calebamiles: I personally don't feel this distinction is useful

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -141,11 +130,16 @@ proposing governance changes. However, as changes start impacting other SIGs or
 the larger developer community outside of a SIG, the KEP process should be used
 to coordinate and communicate.
 
-### KEP Template
+Enhancements that have major impacts on multiple SIGs should use the KEP process.
+A single SIG will own the KEP but it is expected that the set of approvers will span the impacted SIGs.
+The KEP process is the way that SIGs can negotiate and communicate changes that cross boundaries.
+
+KEPs will also be used to drive large changes that will cut across all parts of the project.
+These KEPs will be owned by SIG-architecture and should be seen as a way to communicate the most fundamental aspects of what Kubernetes is.
```
calebamiles: I don't believe it's needed to call this out explicitly. A cross cutting KEP will probably be managed by SIG Architecture but, like the cloud provider extraction, another SIG like API Machinery might manage the KEP itself

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -141,11 +130,16 @@ proposing governance changes. However, as changes start impacting other SIGs or
...
 
-The template for a KEP is precisely defined in the [template proposal][]
+### KEP Template
 
-[template proposal]: https://github.com/kubernetes/community/pull/1124
+The template for a KEP is precisely defined [here](0000-kep-template.md)
```
calebamiles: probably better to not use a relative path here

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -173,10 +167,9 @@ Metadata items:
   * Must be one of `Draft`, `Deferred`, `Approved`, `Rejected`, `Withdrawn`,
     `Final`, `Replaced`.
 * **authors** Required
-  * A list of authors for the KEP.  We require a name (which can be a psuedonym)
-    along with a github ID.  Other ways to contact the author is strongly
-    encouraged.  This is a list of maps.  Subkeys of each item: `name`,
-    `github`, `email` (optional), `slack` (optional).
+  * A list of authors for the KEP.
```
calebamiles: seems fine for now, but in general we probably should work towards some standard contact information for KEPs and test owners

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -186,13 +179,17 @@ Metadata items:
 * **participating-sigs** Optional
   * A list of SIGs that are involved or impacted by this KEP.
   * A special value of `kubernetes-wide` will indicate that this KEP has impact
-    across the entire project.
+    across the entire project. These KEPs will often be owned by sig-architecture.
```
calebamiles: seems redundant and I personally have no affinity for special values

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
-their roadmap; fleshed out with a design; started; and finally implemented. As
-Kubernetes continues to mature, hopefully metrics on the utilization of features
-will drive decisions on what features to maintain and which to deprecate and so
-it is possible that a KEP would be retired if its functionality no longer provides
-sufficient value to the community.
+A KEP is has the following states
+
+- `Draft`: The KEP has been proposed and is actively being defined.
+- `Deferred`: The KEP is proposed but not actively being worked on.
+- `Approved`: Using lazy consensus, the approvers have approved this KEP.
```
calebamiles: I don't think anyone has decided to use a lazy consensus model here; and I think it's important to mention that the SIG is the body approving or rejecting a KEP while the _individuals_ listed in `approvers` and `reviewers` are their delegates

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
-- opened -> accepted -> scoped -> started -> implemented (j)
-- opened -> accepted -> scoped -> started -> implemented -> retired (k)
-
-the happy path is denoted by (j) where an KEP is opened; accepted by a SIG as in
-their roadmap; fleshed out with a design; started; and finally implemented. As
-Kubernetes continues to mature, hopefully metrics on the utilization of features
-will drive decisions on what features to maintain and which to deprecate and so
-it is possible that a KEP would be retired if its functionality no longer provides
-sufficient value to the community.
+A KEP is has the following states
```
calebamiles: thanks for reducing these states; I do still think an `Orphaned` state would be useful but we can add it later as needed I suppose. I think the work done to add statuses to issues/pull requests would suggest that adding a `Started` state would be helpful

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
+
+- `Draft`: The KEP has been proposed and is actively being defined.
```
calebamiles: I think this state is redundant given that one could simply use the fact that the PR hasn't been merged to determine if the KEP was a draft or not so that merging the PR would place the KEP into one of `{Deferred, Approved, Rejected}`

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
+- `Deferred`: The KEP is proposed but not actively being worked on.
+- `Approved`: Using lazy consensus, the approvers have approved this KEP.
+  It will move forward to implementation.
+- `Rejected`: The approvers and authors have decided that this KEP is not moving forward.
+  The KEP is kept around as a historical document.
+- `Withdrawn`: The KEP has been withdrawn by the authors.
+- `Final`: The KEP has been implemented and is no longer actively changed.
```
calebamiles: My personal intention and perhaps also that of @jdumars was that these KEPs are living documents so that I don't think a `Final` state is needed at this time. Also whether or not a KEP is not being actively changed can be derived from either source control activity or the `mtime` on the KEP (which is probably what any automation would do anyway)

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
-looking for a complete list of KEPs. There should be basic CI to ensure that an
-`index.md` remains up to date.
-
-Ideally no work would begin within the repositories of the Kubernetes organization
-before a KEP has been approved by the responsible SIG or working group. While the
-details of how SIGs organize their work is beyond the scope of this proposal one
-possibility would be for each charter SIG to create a top level repository within
-the Kubernetes org where implementation issues managed by that SIG would be filed.
+KEPs are checked into the community repo under the `/kep` directory.
+KEPs can either be put at the top level of that directory or in a SIG specific subdirectory.
```
calebamiles: I don't know if we want to do that yet, especially since we still need a way to generate these monotonically increasing numbers. I was personally envisioning the breakout in to separate directories to aid with automatically assigning SIGs to KEPs they need to review but since we're still bootstrapping this process I think we can have all KEPs land at the top level for now. The `Metadata` section should help us move them later

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
+KEPs in SIG specific subdirectories have limited impact outside of the SIG and can leverage SIG specific OWNERS files.
```
calebamiles: if this is so, then would the `kubernetes-wide` special value be needed at all? Also why then add an `owning-sig` field in the metadata?

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
+
+New KEPs can be checked in with a file name in the form of `draft-YYYYMMDD-my-title.md`.
```
calebamiles: if one purpose of quickly checking in KEPs was to assign them a number, why have the intermediate filename? Also creation information is derivable from source control

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
+As significant work is done on the KEP the authors can assign a KEP number.
```
calebamiles: do you mean editors can assign a KEP number? Really anyone should be able to submit a PR to assign the KEP number since it's pure administrivia (in my opinion) once the KEP has been approved by the SIG

keps/1-kubernetes-enhancement-proposal-process.md
```diff
@@ -221,106 +218,34 @@ Metadata items:
...
+This is done by taking the next number in the NEXT_KEP_NUMBER file, incrementing that number, and renaming the KEP.
```
calebamiles: seems like something to introduce when (and if) automation assigns the numbers but for O(100) KEPs I think this would be overkill. While some people are able to keep a large mapping of arbitrary numbers to unstructured text in their mind, I believe that in general it would be much more helpful to talk about "cloud provider extraction" than `KEP 0002` in discussions with other humans

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I'm in the process of working with contribx to create a website (kep.k8s.io or community.k8s.io) so that we can present this stuff in a more accessible way. Having the metadata in a form that is compatible with static site generators will make this much easier.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: yeah -- agreed.  As things stand it was just too verbose.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: This comes down to what "approved" means.  In my mind that means that the KEP is generally directionally right and the approvers have given it a thumbs up.  Draft is a great state for it to be in while that is being driven.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: This is something we should discuss.  I'd love to hear other options on this.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I'd love to find a way to have folks sort through KEPs and find the ones that are still being actively implemented/worked on vs. those that are mostly frozen.  I'd like something that is easier to filter than just mtime.  Most of the static site tools and index tools don't have easy access to git history or mtime.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I'm okay with just keeping things flat for now.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I'm thinking that there would be a small number of KEPs like that but they'll have a big impact.  The owning SIG would be the main driver of the change even though it can/would impact a large number of other SIGs.  This would be a shorthand for "this'll impact pretty much everyone".

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: While source control is great for history, I'm not sure I want to lean on that completely.  I want to provide a way for users to easily visualize this and the tooling for that (static site generators) generally don't make that data available.  Also, if we ever move repos that data might be lost.  I'm looking at things like the RFC archive or the way that the W3C manages documents.  They don't rely on outside data sources for metadata.

jbeda: Wrt assigning a number.  I'm okay with having it assigned as part of the first PR.  I do worry that we may decide that some KEPs are a bad idea soon after creation and the numbers will be sparse.  But we can solve that problem when we get to it.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: Does a KEP have to be "approved" by a SIG to be created?  I would assume that things get approved once the direction is agreed upon.  Until then the KEP is a draft that is being worked on.

I'm happy with assigning a KEP number sooner rather than later.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: One issue with flat directory here is that the OWNERS file here will be a bottleneck.  By having per-SIG directories we can shard this out a bit.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I disagree here.  Often times we have a lot of similarly named proposals floating around.  Having a completely unambiguous way to talk about a proposal is super useful in my mind.  This is one of the things that is *good* about using github issues as the thing that I'd like to try and keep.

Are you suggesting not having KEP numbers?

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I don't think the cloud provider KEP cuts across *all* parts of the project.  It only involves a limited number of SIGs.

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: Could you explain a bit more about how the format makes it easier to work with static site generators? I do think GitHub is _also_ an entirely reasonable way to interact with KEPs (especially given how few of them we have). I played around with a hacky sort of way to extract an arbitrary block of text between two non rendering Markdown lines using widely available tools (awk) but I need to throw that up as a gist

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: Relative paths are more useful for representing these KEPs someplace that isn't github.

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: do you feel that we have a need to distinguish that more strongly than an open PR at this point?

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: Personally I'm a big fan of lazy consensus, I just don't think we're ready to say that's how decisions are made yet

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: I suppose my point is that automated tooling when it exists can add that sort of information when the site is (re)deployed and doesn't need to be mentioned here, at least not yet

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: Only as SIGs sign up for the process. I'm happy to add more people to the top level `OWNERS` file until then

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: Is that not adequately conveyed by the `owning-sig` being SIG Architecture or SIG {ContribX, PM} for that matter?

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: filename information _should never_ be lost when moving repositories, nor should the history in my opinion although it may become a bit mangled. Both of the examples you mentioned for prior art on document management are from a time before distributed VCS systems were in wide use. I think we can cross the bridge of needing to store all the information within the file itself when and if we come to it. I think the sparseness is fine and that the only hard requirement was that the numbers be monotonically increasing

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: I think I envisioned that a SIG would signal resolution on {approved, rejected, deferred} by processing the PR
{approved, deferred} -> merge PR
{rejected} -> close PR

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: Looking at something like hugo or jekyll, they have an idea of "front matter" that can be used when generating sites based on a pile of markdown.  https://gohugo.io/content-management/front-matter/

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: My understanding is that the idea is to get stuff checked in quickly and iterate with multiple PRs instead of trying to handle everything in a single mondo PR.  I'd like to avoid PRs with 100s of comments.  A "draft" state allows iteration pre-approval.  Being checked in doesn't mean approved.  This is similar to code that is behind a feature gate or that is marked alpha.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I'd like to keep things compatible with off the shelf tooling (static site generators) vs. having to invent a lot of new tooling.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: Sure -- SGTM.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: If we have some UI to say "show me all of the KEPs that impact SIG-foo" we then need a special value that means `*`.  I'm happy using `*` instead of `kubernetes-wide`.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: I'm cool with assigning KEP numbers early.  Wrt relying on VCS metadata -- I'd rather not. It creates a more complicated tool chain and makes it harder to process the archive of KEPs without making assumptions about the VCS. It just feels like a layer violation to me.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: We had already decided that merging a PR does not equal approval.  We need to have a draft state as a working state before the KEP is "approved".

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: I'm trying to suggest that there might be other consumers besides static site generators so I wouldn't want to hamstring us up front. Also GitHub Markdown presentation is one of our _existing_ use cases so I think we should ensure whatever we do allows things to look nice here

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: I don't know if a wildcard is a hard requirement, as I've experienced it (like how it works for BASH) it's nice sugar but it isn't a "must have"

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: Not sure where the `Final` state impacts "looks nice" with github rendering. 

Github markdown presentation is very opinionated and not customizable.  Trying to optimize around that as this becomes a more comprehensive set of documents and want to make the presentation of those documents more and more useful.

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: I don't think I agree that it makes the took chain more complicated. At some point I imagine you'll want to have a CI build for deploying say a website. Git is already part of our tool chain so I'm not really seeing the additional overhead. I don't imagine we'll _ever_ move away from git as the VCS despite what hosting provider we might use (self, GitHub, GitLab, Bitbucket) as it would be an incredibly disruptive change.

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: I believe I'm suggesting that merging the PR indicates that the KEP has been considered by the SIG and now lives in one of `{approved, deferred, rejected}`. Making that an invariant might also help us build automated linting of KEPs when we get there. If we find, and here is one place we can pretty easily be data driven, that our PRs are dragging on and the comments are exploding, besides of course the case were a small group of people are generating a fair amount of traffic, we should try tweaking the process then. I'd be happy to work with you to collect that information

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: I'm not suggesting that we should drop KEP numbers. I'm suggesting that in conversations with humans it's probably easier to talk about "implementing the container storage interface for Kubernetes" rather than "implementing KEP 0003". [Rust][] has over 2000 of these type of documents (an order of magnitude more than our design proposals) and seems to be OK. Maybe we should try and get an update from their community on how their process is working for them

[Rust]: https://github.com/rust-lang/rfcs/tree/master/text

keps/1-kubernetes-enhancement-proposal-process.md
calebamiles: You're correct `Final` itself doesn't impact GitHub rendering and yes GitHub rendering is very opinionated; but it is documented and used in a variety of different rendering engines (or at least a compatible format). My point is that we _already_ have GitHub rendering as a reasonable design constraint since the KEPs are stored in a git repository hosted in GitHub. I don't want to speak too much for SIG Docs but they found it unworkable to deploy a static website directly from the branch where changes were being made (especially in order to support things like previews of a rendered site before publishing) so I think at a minimum a scalable automation story would have

1. contributor update KEPs
1. website automation pulls in changes
1. website automation lints KEPs and generates deployment artifacts and metadata (e.g. ToC, most commented KEPs, new KEPs, stale KEPs, recently changed KEPs)
1. website automation prepares preview for final human approval
1. website automation deploys new website

so I don't think it's required to bundle deployment specific (e.g. Hugo, GitHub Pages) metadata within the KEPs at least not until we have pretty wide adoption of any of the tooling we build on top of flat files checked into git

jbeda: One of the lessons from SIG Docs was the usage of netlify to help manage deployment.  This is something that we are doing as we build this site out.

keps/1-kubernetes-enhancement-proposal-process.md
jbeda: This is the nut of the disagreement here, IMO.  

My understanding is that we had already decided that "check in early" was the way we were going.  Making "check in early" square with "checked in equal approved" isn't possible.

The template already had "check in early" as part of the instructions at the very top.

I'm happy to continue to discuss this but I'm pretty set on "check in early" being the process here.

jbeda:
@calebamiles I've updated this and the template to reflect the parts we agree on. I think, at this point, the main points of disagreement are:

1. What does "approved" mean?  Does having an initial PR submitted mean "approved" or is it a separate state transition.  Earlier discussions (as reflected in the template) established a "check in early" process with a "draft" state.
2. Metadata tweaks.  We had decided on this earlier but you want to change this a bit now.  Specifically:
   * Remove the creation-date and last-updated and defer to git instead.
   * Remove the `kubernetes-wide` special value for `participating-sigs`
   * Remove the `Final` state and instead rely on git to determine that a KEP has "calmed".

I'm hoping we can discuss this stuff live at SIG-architecture tomorrow morning.  Are there any other disagreements that we should make sure we cover?

calebamiles:
I think that's probably a find list of points to discuss, @jbeda

keps/0000-kep-template.md
```diff
@@ -37,15 +37,16 @@ The title should be lowercased and spaces/punctuation should be replaced with `-
...
   Name it `draft-YYYYMMDD-my-title.md`.
 * Create a PR in the [`kubernetes/community`](https://github.com/kubernetes/community) repo.
 * Check in early.
   Do this once the document holds together and general direction is understood by many in the sponsoring SIG.
   View anything marked as a draft as a working document.
   Aim for single topic PRs to keep discussions focused.
   If you disagree with what is already in a document, open a new PR with suggested changes.
-* As a KEP is approved, rename the file yet again with the final KEP number.
+* A KEP number can be assigned at any time by (a) taking the next number in the NEXT_KEP_NUMBER file and (b) incrementing that number.
+  These PRs should be approved quickly to minimize merge conflicts.
```
timothysc: The strict enumeration kind of seems like a #fail , why do we need strict ordering?

keps/1-kubernetes-enhancement-proposal-process.md
timothysc: s/is has/has

keps/0000-kep-template.md
timothysc: If we do this we should just have a script that does this for us.

jbeda: The idea of assigning a number will help to (a) make these feel more official (b) show some ordering in terms of where these fit into the history of the project and (c) provide a way to unambiguously refer to a KEP.

RFCs and PEPs both have numbers and it helps bring the proposals together to feel more like a coherent whole, IMO.

I'd be open for figuring out how to have a script help out on the mechanics here.

jbeda:
This was discussed at length at the SIG-architecture meeting.  Notes at http://bit.ly/sig-architecture. Video at https://youtu.be/_PJcb5Xwtvc.

Decisions coming out:
* Rename states from `draft` -> `accepted`, `approved` -> `implementable`, `final` -> `implemented`
* Having a KEP checked in at all indicates that there is acceptance from the SIG that this is something that deserves serious work and discussion.  This means there is *some* level of approval by the owning SIG when a KEP is first commited.
* For now, we'll keep the extended metadata attributes.  As we find ways to use other systems we can always remove this data later.

I'll update the docs here in the next day or two and we can get this checked in.

jbeda:
Updated the proposal.

I've added @timothysc as reviewer and @bgrant0607 as approver.

Hopefully y'all are cool with that.

PR #1585, opened 2018-01-10
Update sig-auth meeting time to 11:00 PT

mattlandis:
Looking at the meeting times it looks like sig-auth has been meeting at 11:00 PT for the last 6 months or so.


liggitt:
thanks... the time we meet shifts with US daylight savings time, it isn't UTC

PR #1586, opened 2018-01-10
Update SIG leads for sig-openstack

xsgordon:
Update SIG leads for sig-openstack based on discussion at KubeCon NA
2017 deep dive session and confirmed in subsequent online
communications. New year, new responsibilities!


dims:
👍  LGTM

PR #1595, opened 2018-01-12
Spec update of Block volume to support for dynamic provisioning

mtanino:
This PR updates current specification of Block volume to support dynamic provisioning functionality.

Initial design doc proposal for block volume: https://github.com/kubernetes/community/pull/1265
Feature issue: https://github.com/kubernetes/features/issues/351

### DIFF
--- a/contributors/design-proposals/storage/raw-block-pv.md
+++ b/contributors/design-proposals/storage/raw-block-pv.md
@@ -41,7 +41,7 @@ This document presents a proposal for managing raw block storage in Kubernetes u
     as bootstrapping 
   * A user wishes to utilize block storage to fully realize the performance of an application tuned to using block devices
   * A user wishes to read from a block storage device and write to a filesystem (big data analytics processing)
-  Future use cases include dynamically provisioning and intelligent discovery of existing devices, which this proposal sets the 
+  Also use cases include dynamically provisioning and intelligent discovery of existing devices, which this proposal sets the
   foundation for more fully developing these methods. 
 
  
@@ -68,7 +68,7 @@ This document presents a proposal for managing raw block storage in Kubernetes u
   
   It is important to note that when a PV is bound, it is either bound as a raw block device or formatted with a filesystem. Therefore, 
   the PVC drives the request and intended usage of the device by specifying the volumeMode as part of the API. This design lends itself
-  to future support of dynamic provisioning by also letting the request initiate from the PVC defining the role for the PV. It also 
+  to support of dynamic provisioning by also letting the request initiate from the PVC defining the role for the PV. It also
   allows flexibility in the implementation and storage plugins to determine their support of this feature. Acceptable values for 
   volumeMode are 'Block' and 'Filesystem'. Where 'Filesystem' is the default value today and not required to be set in the PV/PVC.
   
@@ -355,7 +355,7 @@ spec:
  local:
     path: /dev/xvdc
 ```
-## [FUTURE] UC5: 
+## UC5:
 
 DESCRIPTION: StorageClass with dynamically created volumes 
 
@@ -374,9 +374,7 @@ provisioner: kubernetes.io/local-block-ssd
 parameters:
 ```
 
-***This has implementation details that have yet to be determined. It is included in this proposal for completeness of design ****
-
-## UC6: 
+## UC6:
 
 DESCRIPTION: The developer wishes to request a block device via a Storage Class.
 
@@ -398,6 +396,8 @@ spec:
    requests:
      storage: 10Gi
 ```  
+
+**Since the PVC object is passed to the provisioner, it will be responsible for validating and handling whether or not it supports the volumeMode being passed**
 
 ## UC7: 
 
@@ -423,7 +423,6 @@ Spec:
   gcePersistentDisk:
     pdName: "gce-disk-1"
 ```
-**Since the PVC object is passed to the provisioner, it will be responsible for validating and handling whether or not it supports the volumeMode being passed**
 
 ## UC8:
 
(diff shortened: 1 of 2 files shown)


mtanino:
/cc @msau42 @erinboyd @screeley44 @jsafrane 
@jingxu97 @rootfs 

Please give feedback for the update of dynamic provisioning support, thanks.

jsafrane:
@mtanino, can you please update also "Out of tree provisioning" section in contributors/design-proposals/storage/volume-provisioning.md so we have the whole protocol in one document?

Otherwise this PR is OK.

mtanino:
@jsafrane 
Good catch. I'll also update volume-provisioning.md. Thanks.

mtanino:
@jsafrane 

>> @mtanino, can you please update also "Out of tree provisioning" section in contributors/design-proposals/storage/volume-provisioning.md so we have the whole protocol in one document?

Updated.

jsafrane:
Thanks! LGTM, however I'd appreciate if someone else looks at it too. If not, let's merge it say early next week.

contributors/design-proposals/storage/raw-block-pv.md
```diff
@@ -729,7 +728,79 @@ The usage of pod device map path is;
...
+# Dynamically provisioning
+
+Using dynamic provisioning, user is able to create block volume via provisioners. Currently,
+we have two types of provisioners, internal provisioner and external provisioner.
+During volume creation via dynamic provisioner, user passes persistent volume claim which
+contains `volumeMode` parameter, then the persistent volume claim object is passed to
+provisioners. Therefore, in order to create block volume, provisioners need to support
+`volumeMode` and then create persistent volume with `volumeMode`.
+
+As for the external provisioner case, admin have to carefully provide Kubernetes environment
```
msau42: Can you outline both for internal and external, what happens if the plugin does not support block?

mtanino: done.

contributors/design-proposals/storage/raw-block-pv.md
```diff
@@ -729,7 +728,85 @@ The usage of pod device map path is;
...
+we have two types of provisioners, internal provisioner and external provisioner.
+During volume creation via dynamic provisioner, user passes persistent volume claim which
+contains `volumeMode` parameter, then the persistent volume claim object is passed to
+provisioners. Therefore, in order to create block volume, provisioners need to support
+`volumeMode` and then create persistent volume with `volumeMode`.
+
+If a storage and plugin don't have an ability to create raw block type of volume,
+then `both internal and external provisioner don't need any update` to support `volumeMode`
+because `volumeMode` in PV and PVC are automatically set to `Filesystem` as a default when
+these volume object are created.
```
msau42: So IIUC, a volume will be created, but because the volumeMode is filesystem, it won't bind to the PVC.  So PVC will stay pending forever.  And then it's up to the user to manually delete the provisioned PVs?  Is there anyway we can propagate some type of error to make it more clear to the user what's wrong?

mtanino: @msau42 @jsafrane 

Good catch. This is a problem, we should provide solution as you mentioned.

Can pv_controller check created volume object after Provision()?
Seems if plugin uses internal provisioner, pv_controller have a chance to inspect whether requested volumeMode in PVC and volumeMode in PV are same or not, so that we can add inspection and show error if they don't match.

But if plugin uses external-provisioner, do we have a chance to inspect PV and show error same as internal provisioner?

screeley44: so not sure I know of an easy way to catch and handle this other than somehow letting the user know via an event that they supplied volumeMode in their pvc but there are currently no supported plugins or matching volumes.  

So in the pv-controller --> index.go, we try to find matchingVolumes based on claimSpecs, etc... We perform this check on misMatchingVolumeModes and maybe we can `flag` that there is a volumeMode mismatch, if that `flag is flipped` on and there are no matches meaning we return:

``` nil, nil (no volume, no error)```

maybe we to return:
```nil, <some error type message>``` 
so the caller can see no volume was returned but an error was returned 

(or another return value indicating that we did find volumeMode being used but did not find a matching volume):
```nil,nil,<some message saying plugin may not support volumeMode or no volumes exist">```

 and maybe then let's put that into an event into the pv and pvc?  

Just thinking outloud?  @msau42 @jsafrane @mtanino  any thoughts?

Comment (msau42) [reply contributors/design-proposals/storage/raw-block-pv.md]: Yeah for external provisioners, probably the best you can do is generate an event if the PV is prebound to the PVC and if volume mode doesn't match.

Comment (mtanino) [reply contributors/design-proposals/storage/raw-block-pv.md]: @screeley44 @msau42 

IIUC, the conclusions are;
- For internal provisioner
Check of mismatching volumeMode between PV spec and PVC spec after provision
`pv_controller.go`

- For external provisioner
Check of mismatching volumeMode between PV spec and PVC spec when pv_controller finds matching volumes based on claimSpecs
`index.go`

Comment (mtanino) [reply contributors/design-proposals/storage/raw-block-pv.md]: update.

Comment (jsafrane) [reply contributors/design-proposals/storage/raw-block-pv.md]: + events on the PVC and the PVC is Pending forever (or until admin fixes either internal or external provisioner)

Comment (mtanino) [reply contributors/design-proposals/storage/raw-block-pv.md]: @jsafrane 
Which line are you suggesting to add above comment?

Comment (jsafrane) [reply contributors/design-proposals/storage/raw-block-pv.md]: It's just for the comment above, the proposal itself is OK.

**Comment by jsafrane** [comment]
@msau42, @screeley44, do you have any more comments or is this good enough to be merged?

**Comment by msau42** [comment]
lgtm

**Comment by screeley44** [comment]
lgtm

**Comment by erinboyd** [comment]
lgtm great job @mtanino

**Comment by mtanino** [comment]
@erinboyd 
Thanks.
To avoid confusion, I removed the line 357 `fsType: ext4` from the `Example 2) the created PV:` of volume-provisioning.md

On the other hands, in the `raw-block-pv.md` side, we mention below. Therefore I think example of `Internal provisioner` could have FSType. This makes implementation of provisioner simpler since provisioner doesn't need to change plugin's volumeSource depends on volumeMode.

```
+Regardless of the volumeMode, provisioner can set `FSType` into the plugin's volumeSource
+but the value will be ignored at the volume plugin side if `volumeMode` is `Block`.
```

mtanino:
https://github.com/kubernetes/kubernetes/issues/59942

PR #1599, opened 2018-01-12
Update employer

kris-nova:


cblecker:
@kris-nova do you also want to update your entry on the cluster-api working group?
https://github.com/kubernetes/community/blob/f1b08f86a77a137d4929b7410a8c2e9474da1f94/sigs.yaml#L1201-L1203

PR #1601, opened 2018-01-12
New design of volume reconstruction

jingxu97:
This PR describes the new design of volume reconstruction in kubelet
volume manager.

### DIFF
--- a/contributors/design-proposals/storage/volume-reconstruct.md
+++ b/contributors/design-proposals/storage/volume-reconstruct.md
@@ -1,0 +1,94 @@
+# New Design Volume Reconstruction
+
+**Author**: Jing Xu (@jingxu97)
+
+**Last Updated**: 1/11/2018
+
+**Status**: Proposal
+
+This document describes the new design of volume reconstruction in Kubelet
+volume manager running on cluster nodes.
+
+## Background and Motivation
+
+Today, kubelet volume manager uses desired/actual states to store some volume
+and pod information for reconciler to work upon. When kubelet restarts, all
+states become empty. Desired state populator will recover the desired states by
+getting the existing pods information from pod manager. Reconciler can then
+recover actual state through reconcile loop. 
+
+The problem is that if pod is deleted during kubelet restarts, desired state
+no longer has this pod and volume information so as the actual state. But the
+actual mount points and direcotries are still exist on the disk. In order to
+clean up those left over mounts and directories, we propose a volume
+reconstruction process. The basic idea is to scan pod directories and recover
+volume information (such as volume spec) from the mount path and device mount
+path which will be updated into the actual state. With this reconstructure
+process, actual state can match the real world and reconciler will tear down
+volumes accordingly. 
+
+However, this reconstruction design has issues for some plugins such as ()
+because there is not enough information to reconstruct volume spec correctly.
+For those volume plugins, volume tear down will fail.
+
+## Possible solution
+
+In searching of the solutions, we first need to understand why we need to reconstruct volume spec. 
+Volume spec is an internal representation of a volume.  All API volume types translate to Spec.
+type Spec struct {
+	Volume           *v1.Volume                // for inline volume use case
+	PersistentVolume *v1.PersistentVolume      // for pvc use case
+	ReadOnly         bool
+}
+Both Volume and PersistentVolume has a name and volume source. Different volume plugins have different representation of their sources. 
+Volume spec is constructed when adding pod volume information into desired state.
+The main use of this spec is to get the volume source name to construct unique volume name (which typicall includes plugin name and volume source name). 
+The unique volume name is used as a key in actual/desired state. The volume source name is also used on construct global mount path (device mount path)
+for volume mounter. When volume is teared down, unmounter uses the same function to construct the mount path from the spec again.
+Different from global mount path, the bind mount path uses plugin name and spec name (not the volume source name). 
+
+So look more closely, you can see for only the purpose of volume tear down, if we have a record of those mount paths (bind and global) in actual state, there is no need to 
+construct those paths from the volume spec again. In case of kubelet restarts, those mount paths could be scanned from the mount lists left on the disk. (bind mount path has
+a path structure which can be easily idenetified and global mount path could be obtained from the mount reference of the bind mount path). This observation leads to our first
+possible solution.
+
+1. Remove the dependency of volume spec. 
+Modify the data structure in actual state to keep a record of the mount paths so that volume tear down process (unmount) can use those records instead of reconstructing them
+directly from volume spec. In case the kubelet restarts, reconstruction process just need to find out the left mounting paths and mark them into the actual state. Even if the
+volume spec cannot be reconstructed correctly, volume tear down process can still work. 
+
+But this solution still has an issue. Because some plugins cannot reconstruct volume spec correctly, which means the generated unique volume name might not match with the real one.
+In case another new pod is added to use the same volume, two volume operations on the same device might happen concurrently and cause race condition because they have different
+unique volume names.
+
+2. Cleanup mounts directly without reconstruction.
+This approach will try to clean up mounts that are no longer needed before reconciler starts to work. It scanns the pod volume directories when kubelet restarts to find the mount
+paths (bind and global). If the pod volume is no longer exist in desired state, clean up the mounts. To avoid long delay or mount hanging problem, we can use operation executor to
+start a new routine. 
+
+The potentical issue of this approach is that cleanup process might fail and
+reconciler will not try to do this process again.
+
+## Proposal
+
+The current proposal is trying to take the advantages from the two possible
+solutions mentioned above. 
+1. If volume plugins can support reconstruction well (i.e., volume spec can
+   be fully recovered), reconciler will update this information into the states
+   and reconciler will take care of volume cleanup. Reconciler can retry if
+   cleanup fails.
+2. If volume plugins cannot support reconstruction, reconciler clean up the
+   mounts directly with a go routine (operation executor) to avoid delay. If it
+   fails, reconciler will not retry.
+
+There are a number of monitoring options supported by Kubernetes
+ - Resource Metrics API (Alpha feature in progress [Resource Metrics API](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-metrics-api.md)
+ - Collect metrics in Heapster. (metrics has to exposed by Kubelet in Summary API)
+ - Export metrics through StackDriver 
+ - Kubectl Command (kubectl top, kubectl describe)
+
+All the storage metrics mentioned in Background are collected by Kubelet Summary API so that it could be integrated with the above monitoring system. However, pod volume metrics is provided in PodStat object’s VolumeStat and VolumeStat only has volume name specified in pod specification. There is no direct way of checking volume usage information by PVC name, which is more preferable by users.
+
+## Implementation Timeline:
+The work is targeted for kubernetes v1.10
+
--- a/contributors/design-proposals/volume-reconstruct.md
+++ b/contributors/design-proposals/volume-reconstruct.md
@@ -1,0 +1,94 @@
+# New Design Volume Reconstruction
+
+**Author**: Jing Xu (@jingxu97)
+
+**Last Updated**: 1/11/2018
+
+**Status**: Proposal
+
+This document describes the new design of volume reconstruction in Kubelet
+volume manager running on cluster nodes.
+
+## Background and Motivation
+
+Today, kubelet volume manager uses desired/actual states to store some volume
+and pod information for reconciler to work upon. When kubelet restarts, all
+states become empty. Desired state populator will recover the desired states by
+getting the existing pods information from pod manager. Reconciler can then
+recover actual state through reconcile loop. 
+
+The problem is that if pod is deleted during kubelet restarts, desired state
+no longer has this pod and volume information so as the actual state. But the
+actual mount points and direcotries are still exist on the disk. In order to
+clean up those left over mounts and directories, we propose a volume
+reconstruction process. The basic idea is to scan pod directories and recover
+volume information (such as volume spec) from the mount path and device mount
+path which will be updated into the actual state. With this reconstructure
+process, actual state can match the real world and reconciler will tear down
+volumes accordingly. 
+
+However, this reconstruction design has issues for some plugins such as ()
+because there is not enough information to reconstruct volume spec correctly.
+For those volume plugins, volume tear down will fail.
+
+## Possible solution
+
+In searching of the solutions, we first need to understand why we need to reconstruct volume spec. 
+Volume spec is an internal representation of a volume.  All API volume types translate to Spec.
+type Spec struct {
+	Volume           *v1.Volume                // for inline volume use case
+	PersistentVolume *v1.PersistentVolume      // for pvc use case
+	ReadOnly         bool
+}
+Both Volume and PersistentVolume has a name and volume source. Different volume plugins have different representation of their sources. 
+Volume spec is constructed when adding pod volume information into desired state.
+The main use of this spec is to get the volume source name to construct unique volume name (which typicall includes plugin name and volume source name). 
+The unique volume name is used as a key in actual/desired state. The volume source name is also used on construct global mount path (device mount path)
+for volume mounter. When volume is teared down, unmounter uses the same function to construct the mount path from the spec again.
+Different from global mount path, the bind mount path uses plugin name and spec name (not the volume source name). 
+
+So look more closely, you can see for only the purpose of volume tear down, if we have a record of those mount paths (bind and global) in actual state, there is no need to 
+construct those paths from the volume spec again. In case of kubelet restarts, those mount paths could be scanned from the mount lists left on the disk. (bind mount path has
+a path structure which can be easily idenetified and global mount path could be obtained from the mount reference of the bind mount path). This observation leads to our first
+possible solution.
+
+1. Remove the dependency of volume spec. 
+Modify the data structure in actual state to keep a record of the mount paths so that volume tear down process (unmount) can use those records instead of reconstructing them
+directly from volume spec. In case the kubelet restarts, reconstruction process just need to find out the left mounting paths and mark them into the actual state. Even if the
+volume spec cannot be reconstructed correctly, volume tear down process can still work. 
+
+But this solution still has an issue. Because some plugins cannot reconstruct volume spec correctly, which means the generated unique volume name might not match with the real one.
+In case another new pod is added to use the same volume, two volume operations on the same device might happen concurrently and cause race condition because they have different
+unique volume names.
+
+2. Cleanup mounts directly without reconstruction.
+This approach will try to clean up mounts that are no longer needed before reconciler starts to work. It scanns the pod volume directories when kubelet restarts to find the mount
+paths (bind and global). If the pod volume is no longer exist in desired state, clean up the mounts. To avoid long delay or mount hanging problem, we can use operation executor to
+start a new routine. 
+
+The potentical issue of this approach is that cleanup process might fail and
+reconciler will not try to do this process again.
+
+## Proposal
+
+The current proposal is trying to take the advantages from the two possible
+solutions mentioned above. 
+1. If volume plugins can support reconstruction well (i.e., volume spec can
+   be fully recovered), reconciler will update this information into the states
+   and reconciler will take care of volume cleanup. Reconciler can retry if
+   cleanup fails.
+2. If volume plugins cannot support reconstruction, reconciler clean up the
+   mounts directly with a go routine (operation executor) to avoid delay. If it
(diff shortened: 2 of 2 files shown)


contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+But this solution still has an issue. Because some plugins cannot reconstruct volume spec correctly, which means the generated unique volume name might not match with the real one.
+In case another new pod is added to use the same volume, two volume operations on the same device might happen concurrently and cause race condition because they have different
+unique volume names.
+
+### Cleanup mounts directly without reconstruction.
+This approach will try to clean up mounts that are no longer needed before reconciler starts to work. It scanns the pod volume directories when kubelet restarts to find the mount
+paths (bind and global). If the pod volume is no longer exist in desired state, clean up the mounts. To avoid long delay or mount hanging problem, we can use operation executor to
+start a new routine. 
+
+The potentical issue of this approach is that cleanup process might fail and
```
mlmhl: Typo in word? Maybe `potential`?

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+no longer has this pod and volume information so as the actual state. But the
+actual mount points and direcotries are still exist on the disk. In order to
+clean up those left over mounts and directories, we propose a volume
+reconstruction process. The basic idea is to scan pod directories and recover
+volume information (such as volume spec) from the mount path and device mount
+path which will be updated into the actual state. With this reconstructure
+process, actual state can match the real world and reconciler will tear down
+volumes accordingly. 
+
+However, this reconstruction design has issues for some plugins such as ()
```
mlmhl: Is this empty `()` Intentionally or just a mistake?

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+
+Today, kubelet volume manager uses desired/actual states to store some volume
+and pod information for reconciler to work upon. When kubelet restarts, all
+states become empty. Desired state populator will recover the desired states by
+getting the existing pods information from pod manager. Reconciler can then
+recover actual state through reconcile loop. 
+
+The problem is that if pod is deleted during kubelet restarts, desired state
+no longer has this pod and volume information so as the actual state. But the
+actual mount points and direcotries are still exist on the disk. In order to
```
stmcginnis: s/are still exist/still exist/

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+```
+type Spec struct {
+	Volume           *v1.Volume                // for inline volume use case
+	PersistentVolume *v1.PersistentVolume      // for pvc use case
+	ReadOnly         bool
+}
+```
+Both Volume and PersistentVolume has a name and volume source. Different volume plugins have different representation of their sources. 
+Volume spec is constructed when adding pod volume information into desired state.
+The main use of this spec is to get the volume source name to construct unique volume name (which typicall includes plugin name and volume source name).
```
stmcginnis: s/which typicall/which typically/

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+	Volume           *v1.Volume                // for inline volume use case
+	PersistentVolume *v1.PersistentVolume      // for pvc use case
+	ReadOnly         bool
+}
+```
+Both Volume and PersistentVolume has a name and volume source. Different volume plugins have different representation of their sources. 
+Volume spec is constructed when adding pod volume information into desired state.
+The main use of this spec is to get the volume source name to construct unique volume name (which typicall includes plugin name and volume source name). 
+The unique volume name is used as a key in actual/desired state. The volume source name is also used on construct global mount path (device mount path)
+for volume mounter. When volume is teared down, unmounter uses the same function to construct the mount path from the spec again.
```
stmcginnis: s/teared down/torn down/

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+Modify the data structure in actual state to keep a record of the mount paths so that volume tear down process (unmount) can use those records instead of reconstructing them
+directly from volume spec. In case the kubelet restarts, reconstruction process just need to find out the left mounting paths and mark them into the actual state. Even if the
+volume spec cannot be reconstructed correctly, volume tear down process can still work. 
+
+But this solution still has an issue. Because some plugins cannot reconstruct volume spec correctly, which means the generated unique volume name might not match with the real one.
+In case another new pod is added to use the same volume, two volume operations on the same device might happen concurrently and cause race condition because they have different
+unique volume names.
+
+### Cleanup mounts directly without reconstruction.
+This approach will try to clean up mounts that are no longer needed before reconciler starts to work. It scanns the pod volume directories when kubelet restarts to find the mount
```
stmcginnis: s/scanns/scans/

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+paths (bind and global). If the pod volume is no longer exist in desired state, clean up the mounts. To avoid long delay or mount hanging problem, we can use operation executor to
```
stmcginnis: s/is no longer exist/no longer exists/

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+start a new routine. 
+
+The potentical issue of this approach is that cleanup process might fail and
+reconciler will not try to do this process again.
+
+## Proposal
+
+The current proposal is trying to take the advantages from the two possible
+solutions mentioned above. 
+1. If volume plugins can support reconstruction well (i.e., volume spec can
```
stmcginnis: Can you add some detail about how this will be determined? It would be nice to see a little more detail in the proposal about how this is planned to be implemented.

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+The main use of this spec is to get the volume source name to construct unique volume name (which typicall includes plugin name and volume source name). 
+The unique volume name is used as a key in actual/desired state. The volume source name is also used on construct global mount path (device mount path)
+for volume mounter. When volume is teared down, unmounter uses the same function to construct the mount path from the spec again.
+Different from global mount path, the bind mount path uses plugin name and spec name (not the volume source name). 
+
+So look more closely, you can see for only the purpose of volume tear down, if we have a record of those mount paths (bind and global) in actual state, there is no need to 
+construct those paths from the volume spec again. In case of kubelet restarts, those mount paths could be scanned from the mount lists left on the disk. (bind mount path has
+a path structure which can be easily idenetified and global mount path could be obtained from the mount reference of the bind mount path). This observation leads to our first
+possible solution.
+
```
mtanino: Can you also add update about `block volume` case into somewhere?
This spec only explains updates related to `filesystem volume` case.

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+Volume spec is constructed when adding pod volume information into desired state.
+The main use of this spec is to get the volume source name to construct unique volume name (which typicall includes plugin name and volume source name). 
+The unique volume name is used as a key in actual/desired state. The volume source name is also used on construct global mount path (device mount path)
+for volume mounter. When volume is teared down, unmounter uses the same function to construct the mount path from the spec again.
+Different from global mount path, the bind mount path uses plugin name and spec name (not the volume source name). 
+
+So look more closely, you can see for only the purpose of volume tear down, if we have a record of those mount paths (bind and global) in actual state, there is no need to 
+construct those paths from the volume spec again. In case of kubelet restarts, those mount paths could be scanned from the mount lists left on the disk. (bind mount path has
+a path structure which can be easily idenetified and global mount path could be obtained from the mount reference of the bind mount path). This observation leads to our first
```
mlmhl: Should we consider the scenario that bind mount path not exist but global mount path still remains? Or we just solve this problem in another PR?

jingxu97:
cc @gnufied

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+	ReadOnly         bool
+}
+```
+Both Volume and PersistentVolume has a name and volume source. Different volume plugins have different representation of their sources. 
+Volume spec is constructed when adding pod volume information into desired state.
+The main use of this spec is to get the volume source name to construct unique volume name (which typicall includes plugin name and volume source name). 
+The unique volume name is used as a key in actual/desired state. The volume source name is also used on construct global mount path (device mount path)
+for volume mounter. When volume is teared down, unmounter uses the same function to construct the mount path from the spec again.
+Different from global mount path, the bind mount path uses plugin name and spec name (not the volume source name). 
+
```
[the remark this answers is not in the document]
gnufied: +1 to this. Also I think - volume source name roughly maps to Outer spec name and it might be worth adding that bit to clarify.

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+So look more closely, you can see for only the purpose of volume tear down, if we have a record of those mount paths (bind and global) in actual state, there is no need to 
+construct those paths from the volume spec again. In case of kubelet restarts, those mount paths could be scanned from the mount lists left on the disk. (bind mount path has
+a path structure which can be easily idenetified and global mount path could be obtained from the mount reference of the bind mount path). This observation leads to our first
+possible solution.
+
+### Remove the dependency of volume spec. 
+Modify the data structure in actual state to keep a record of the mount paths so that volume tear down process (unmount) can use those records instead of reconstructing them
+directly from volume spec. In case the kubelet restarts, reconstruction process just need to find out the left mounting paths and mark them into the actual state. Even if the
```
gnufied: Can we add some example for what you mean by this?

gnufied:
There is another case which slightly aggravates some users - sometimes while a node is down and one or more pods get migrated pod from node, the reconstruction process kinda breaks down because reconstruction tries to find mount points and all that is left on the disk is pod/volume directories (and nothing is mounted). This leads to massive spam of "Cleaning up orphaned pods failed" in kubelet logs...

contributors/design-proposals/storage/volume-reconstruct.md
jingxu97: In this case, constructVolumeSpec should fail, then in this proposal reconcile will try to tear down volume based on the mount path. Tear down process will remove the directory, but the global mount still exist. We can get the global mount path through the mount references from the bind mount path. If bind mount path does not exist, we need to figure out another way.

gnufied:
@jingxu97 yep that should work this design. thank you.

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,88 @@
...
+for volume mounter. When volume is teared down, unmounter uses the same function to construct the mount path from the spec again.
+Different from global mount path, the bind mount path uses plugin name and spec name (not the volume source name). 
+
+So look more closely, you can see for only the purpose of volume tear down, if we have a record of those mount paths (bind and global) in actual state, there is no need to 
+construct those paths from the volume spec again. In case of kubelet restarts, those mount paths could be scanned from the mount lists left on the disk. (bind mount path has
+a path structure which can be easily idenetified and global mount path could be obtained from the mount reference of the bind mount path). This observation leads to our first
+possible solution.
+
+### Remove the dependency of volume spec. 
+Modify the data structure in actual state to keep a record of the mount paths so that volume tear down process (unmount) can use those records instead of reconstructing them
```
jsafrane: volume tear down process is often not simple unmount.

- Ceph RBD needs to unmap ("node-initiated detach") the volume
- iSCSI needs to detach the volume (again, node-initiated detach)
- CSI can do virtually anything

All three need the full volume spec. Right now, we use json files placed on the filesystem with the volume spec. This brings its own issues, mainly with compatibility - new kubelet must be able to read files created by old kubelet(s) and it is not ideal solution either.

gnufied: The unmount device depends on Spec while unmount volume can be done without the spec. I am not sure, how this proposal is going to workaround the fact that, `UnmountDevice` requires full device spec.

jingxu97: Maybe I should remove "(unmount)" here. I think I checked all plugins and from currently implementation, volume spec is only needed during volume set up (attach, waitforattach, mountdevice/mapdevice, setup). But for volume tear down (teardown, unmapdevice, TearDownDevice/unmountDevice, detach), only a couple of paths are needed (globalmountpath, mountpath/podDevicePath, devicePath)
For FC, Ceph RDB,  iSCSI: teardown includes unmapDevice (podDevicePath, globalMountPath), tearDownDevice(globalMountPath)
So I don't see they need full volume spec so far. Please let me know if I miss something.

gnufied: hmm, looking closer - I think I agree with @jingxu97 . Although volume spec is available in `GenerateUnmountDevice` func - we aren't using it. The `devicePath` is enough to unmount devices in most cases.

In case of iSCSI - we are saving disk configuration to a json file - https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/iscsi/iscsi_util.go#L201 similar to CSI.

mtanino: @gnufied @jingxu97 

In case of iSCSI, we need many parameters(portal ip, iface name, iqn, initiatorname) to logout and cleanup iscsi interface.
All of them are stored in volumeSpec once, but it isn't provided during unmountDevice().
Therefore, iSCSI currently saves connection configuration into json file. If we get same parameters via volumeSpec at unmountDevice(), we don't need this logic.

mtanino: But as for the volume reconstruction path, iSCSI plugin can't create full spec using global mount path, therefore the plugin still need to store configurations in the json file same as CSI.

contributors/design-proposals/storage/volume-reconstruct.md
jingxu97: actually there are not much difference between block volume and filesystem volume. The tear down process is different, the name of those mounts are different (e.g., globalMapPath), but still tear down needs a record of those paths.

contributors/design-proposals/storage/volume-reconstruct.md
jingxu97: fixed

contributors/design-proposals/storage/volume-reconstruct.md
jingxu97: @mtanino yes, they need those information during tear down. But we can get those information from plugin.Unmounter which only needs paratmers of InnerVolumeSpecName, and PodUID and we have those information. 
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/iscsi/iscsi.go#L163
This unmounter only has podUID, volName, etc. But here https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/iscsi/iscsi_util.go#L432, during DetachDisk, I do see it might use Portals, Iqn,  Iface etc. Then what does it mean, those values will be just empty?

contributors/design-proposals/storage/volume-reconstruct.md
jingxu97: Currently, for volumes such as AWS, GCE, the volume source name can be retrieved by checking the mount reference.

jingxu97:
cc @gnufied @jsafrane @mlmhl @stmcginnis 
The design doc is updated, PTAL. Thanks!

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,118 @@
...
+Currently, the following volume plugins can support reconstruction. These volume
+source names are just the device names which can be retrived from the mount paths.
+
+AWS EBS, Azure_dd, Cinder, CSI, GCE_PD, FC, iScsi, Photon, Rbd, Vsphere 
+
+2. If volume plugins cannot support reconstruction, reconciler clean up the
+   mounts directly with a go routine (operation executor) to avoid delay. If it
+   fails, reconciler will not retry.
+
+The following volume plugins cannot support reconstruction,
```
mtanino: Do we need to add `local volume plugin` to this group? @msau42 @dhirajh

msau42: I think local volume can support reconstruction even though the spec is not complete.

For filesystem volumes: it only needs the kubelet's pod volume path
For block volumes: it doesn't do anything on teardown and just relies on common kubelet teardown

Do you see any other issues with an incomplete spec?  The incomplete spec means that Setup will fail, but we only care about Teardown here right?

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,118 @@
...
+unique volume names.
+
+### Solution2: Cleanup mounts directly without reconstruction.
+This approach will try to clean up mounts that are no longer needed before reconciler starts to work. It scans the pod volume directories when kubelet restarts to find the mount
+paths (bind and global). If the pod volume no longer exist in desired state, clean up the mounts. To avoid long delay or mount hanging problem, we can use operation executor to
+start a new routine. 
+
+Issue: The potential issue of this approach is that cleanup process might fail and
+reconciler will not try to do this process again.
+
```
msau42: Have you also considered an alternative solution using kubelet checkpointing?

contributors/design-proposals/storage/volume-reconstruct.md
mtanino: >> Do you see any other issues with an incomplete spec?

I think if the incomplete spec satisfies following conditions,  there is no other issue for reconstruction.
- for filesystem: plugin.NewMounter() should be succeeded
- for mapperPlugin.NewBlockVolumeMapper() should be succeeded and also spec should have `volumeMode`.

>> but we only care about Teardown here right?

Right.

contributors/design-proposals/storage/volume-reconstruct.md
```diff
@@ -0,0 +1,118 @@
...
+solutions mentioned above. 
+1. If volume plugins can support reconstruction well (i.e., volume spec can
+   be fully recovered), reconciler will update this information into the states
+   and reconciler will take care of volume cleanup. Reconciler can retry if
+   cleanup fails.
+
+Currently, the following volume plugins can support reconstruction. These volume
+source names are just the device names which can be retrived from the mount paths.
+
+AWS EBS, Azure_dd, Cinder, CSI, GCE_PD, FC, iScsi, Photon, Rbd, Vsphere
```
mtanino: @jingxu97 
Please add `local volume plugin` to reconstruction supported plugin. cc @msau42

calvix:
Hello, any progress on this PR? I have hit issue https://github.com/kubernetes/kubernetes/issues/54108 and finishing this PR would be helpful.

andyzhangx:
Hi @jingxu97 I found a regression issue(https://github.com/kubernetes/kubernetes/issues/62540) of azure disk on v1.10.0 which could be related to this design change.  Other disk related volume may also has this issue if it supports both Linux and Windows.

This issue is caused by  `devicePath` updated after `WaitForAttach`, related code is here:
https://github.com/kubernetes/kubernetes/blob/ee4d90aaa61150139cdcd67a73e22da8cb226dc6/pkg/volume/util/operationexecutor/operation_generator.go#L517-L518

`devicePath` could not be changed in azure_dd since 
 - AttachDisk happens on master, `devicePath` is a LUN(Logical Unit Number) number after AttachDisk process in the beginning
 - [WaitForAttach](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/azure_dd/attacher.go#L151-L190) happens on agent node: on Linux node, `devicePath` could be `/dev/disk/azure/scsi1/lunx` while on Windows node, `devicePath` would be a disk numbe , so for windows node, when WaitForAttach is recalled, it could not tell whether it's a LUN or a disk number (**this is the key point**).

So my question is:
Is there any way to not update `devicePath` after `WaitForAttach` successfully?

andyzhangx:
@jingxu97 Here is the update:
I fixed [devicePath update issue for azure_dd](https://github.com/kubernetes/kubernetes/issues/62540) by not using `devicePath` in `WaitForAttach` func anymore, here is the PR: https://github.com/kubernetes/kubernetes/pull/62612, while in the fix, I also found aws disk is also using `devicePath`  in [WaitForAttach func](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/aws_ebs/attacher.go#L145), I think there is potentical issue for aws_ebs

cofyc:
I find this doc is still useful. How about moving it to KEP?

PR #1608, opened 2018-01-16
postpone pv deletion if it is bound to a pvc

NickrenREN:
add finalizer to postpone pv deletion if it is bound to a pvc
xref: https://github.com/kubernetes/features/issues/499

/sig storage

/assign @pospispa

### DIFF
--- a/contributors/design-proposals/storage/postpone-pv-deletion.md
+++ b/contributors/design-proposals/storage/postpone-pv-deletion.md
@@ -1,0 +1,71 @@
+# Postpone deletion of a Persistent Volume if it is bound by a PVC
+
+Status: Pending
+
+Version: Beta
+
+Implementation Owner: NickrenREN@ 
+
+## Motivation
+
+Admin can delete a Persistent Volume (PV) that is being used by a PVC.  It may result in data loss.
+
+## Proposal
+
+Postpone the PV deletion until the PV is not used by any PVC.
+
+
+## User Experience
+### Use Cases
+
+* Admin deletes a PV that is being used by a PVC and a pod referring that PVC is not aware of this. 
+This may result in data loss. As a user, I do not want to experience data loss.
+
+## Implementation
+
+### API Server, PV Admission Controller, PV Create: 
+
+We can rename and reuse the PVC admission controller, let it automatically add finalizer information into newly created PV's and PVC's metadata.
+
+PVC protection proposal: [PVC protection](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/postpone-pvc-deletion-if-used-in-a-pod.md)
+
+#### PV controller:
+
+When we look for best matched PV for the PVC, if the PV has the  deletionTimestamp set, we will not choose it even if the PV satisfies all the PVC’s requirement.
+
+For pre-bound PVC/PV, if the PV has  the  deletionTimestamp set, we will not perform the `bind` operation and keep the PVC `Pending`.
+
+#### PV Protection Controller:
+
+PV protection controller is a new internal controller.
+
+Since we already have PV controller which is responsible for synchronizing PVs and PVCs, here in PV protection controller, 
+we can just watch PV events.
+
+PV protection controller watches for PV events that are processed as described below:
+
+* PV add/update/delete events:
+    * If deletionTimestamp is nil and finalizer is missing, the PV is added to PV queue.
+    * If deletionTimestamp is non-nil and finalizer is present, the PV is added to PV queue.
+
+PV information is kept in a cache that is done inherently for an informer.
+
+The PV queue holds PVs that need to be processed according to the below rules:
+
+* If PV is not found in cache, the PV is skipped.
+* If PV is in cache with nil deletionTimestamp and missing finalizer, finalizer is added to the PV. In case the adding finalizer operation fails, the PV is re-queued into the PV queue.
+* If PV is in cache with non-nil deletionTimestamp and finalizer is present, we try to get the PV's status, if it is not `bound`(synchronized by PV controller), the finalizer removal is attempted. The PV will be re-queued if the finalizer removal operation fails.
+
+#### CLI:
+
+If a PV’s deletionTimestamp is set, the commands kubectl get pv and kubectl describe pv will display that the PV is in terminating state.
+
+
+### Client/Server Backwards/Forwards compatibility
+
+N/A
+
+## Alternatives considered
+
+### Add this logic to the existing PV controller instead of creating a new admission and protection controller
+When we bind PV to PVC, we add finalizer for PV and remove finalizer when PV is no longer bound to a PVC.


contributors/design-proposals/storage/postpone-pv-deletion.md
```diff
@@ -0,0 +1,73 @@
...
+
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: NickrenREN@ 
+
+## Motivation
+
+User can delete a Persistent Volume (PV) that is being used by a PVC.  It may result in data loss.
```
pospispa: Nit: actually, only an admin can delete a PV. So this feature is for admins and doesn't affect ordinary users.

contributors/design-proposals/storage/postpone-pv-deletion.md
```diff
@@ -0,0 +1,73 @@
...
+### Use Cases
+
+* User deletes a PV that is being used by a PVC and a pod referring that PVC is not aware of this. 
+This may result in data loss. As a user, I do not want to experience data loss.
+
+## Implementation
+
+### API Server, PV Admission Controller, PV Create: 
+
+A new plugin for PV admission controller will be created. The plugin will automatically add finalizer information into newly created PV's metadata, just like `PVC protection` does.
```
pospispa: Nit: IMO, a link to PVC Protection design or documentation may be useful.

contributors/design-proposals/storage/postpone-pv-deletion.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Client/Server Backwards/Forwards compatibility
+
+N/A
+
+## Alternatives considered
+
+### Add this logic to the existing PV controller instead of creating a new admission and protection controller
+When we bind PV to PVC, we add finalizer for PV and remove finalizer when PV is no longer bound to a PVC.
+
```
pospispa: Could we add a part or the whole logic to the existing PVC Protection admission plugin and PVC Protection controller?

contributors/design-proposals/storage/postpone-pv-deletion.md
NickrenREN: yes, thanks, will do

contributors/design-proposals/storage/postpone-pv-deletion.md
NickrenREN: sorry for the unclear expression, this is not to add the logic to the existing PVC protection admission plugin and PVC Protection controller.
I am considering if we can add PV protection logic to the existing PV controller under the package `pkg/controller/volume/persistentvolume`.

contributors/design-proposals/storage/postpone-pv-deletion.md
NickrenREN: yeah, will update this, thanks

contributors/design-proposals/storage/postpone-pv-deletion.md
pospispa: Your expression is clear. I understand that you're considering to extend PV Controller.
My question may have been better.
Anyway, in my opinion we should consider these options:
1. extend PV controller.
2. extend PVC Protection admission plugin and PVC Protection controller.

NickrenREN: Good point. I talked with @jsafrane . He thinks if there is big chunk of code that two controllers would share , we need to merge them. 
1. I do not think the existing PV controller shares much code with PV protection controller, so we do not need to bind them together;
2. Since you want to move PVC protection to beta and i plan to add PV protection as alpha in 1.10, and also, they two will not share much code, so maybe we can implement them as two separate controller at first, if necessary, merge them later .
WDYT.

pospispa: I was thinking about the joining PV and PVC protection admission plugins together and the PV and PVC protection controllers together from a perspective of monolithic approach versus small programs communicating with each other approach.

AFAIK, Kubernetes is moving towards small core plus many controllers and plugins around approach. That's why joining PV and PVC protection admission plugins together and joining PV and PVC protection controllers together will go against this approach. So I'm against joining PV and PVC protection plugins/controllers together.

In case you find functions in PVC protection controller that are useful also for PV protection controller you can reuse them in PV protection controller anyway.

pospispa:
I have very limited knowledge of PV controller so I'm not able to review the proposed changes for PV controller. The rest looks good to me.

NickrenREN:
Thanks @pospispa , and @jsafrane any comments ?

msau42:

jsafrane:
One thing to consider is the admission plugin. We don't want extra admission plugin for each class of objects we want protect - they need to be explicitly mentioned on API server command line and it's hard to maintain.

I propose there is only one admission plugin, say StorageProtection (or ObjectProtection?) that handles both PV and PVC (depending on alpha feature gates).

In addition, I'd like to have only one feature gate. I hope the PV protection won't be much different from PVC protection and it could go directly to beta. What do you think?

contributors/design-proposals/storage/postpone-pv-deletion.md
```diff
@@ -0,0 +1,73 @@
...
+### Use Cases
+
+* Admin deletes a PV that is being used by a PVC and a pod referring that PVC is not aware of this. 
+This may result in data loss. As a user, I do not want to experience data loss.
+
+## Implementation
+
+### API Server, PV Admission Controller, PV Create: 
+
+A new plugin for PV admission controller will be created. The plugin will automatically add finalizer information into newly created PV's metadata, just like [PVC protection](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/postpone-pvc-deletion-if-used-in-a-pod.md) does.
```
jsafrane: Please share the same admission plugin as PVC protection and rename it to something more general. It's tedious to maintain API server command line with list of all protection plugins that should run.

contributors/design-proposals/storage/postpone-pv-deletion.md
jsafrane: In the end, I'd like to have one "library" or common code that anyone could use to protect X from deleting when Y with some attributes exists. One would use (not copy!) the library to protect PVCs when pod with `pod.volumes[x].persistentVolumeClaim.name == PVC.name` filling only the right informers and callbacks with checks. It can be after alpha, when we can see how this library code would look like.

NickrenREN:
>It looks good for alpha, for GA we need to consider merging the common parts into one package.

yeah, we can implement PV protection at first, and then let's see if we need to merge them later

>I propose there is only one admission plugin, say StorageProtection (or ObjectProtection?) that handles 
both PV and PVC (depending on alpha feature gates).

Sounds good. What about StorageClass protection? should we add it too ?

>In addition, I'd like to have only one feature gate. I hope the PV protection won't be much different from PVC protection and it could go directly to beta. What do you think?

I do not have a strong opinion on what version `PV protection` should be. But if  PV protection goes directly to beta, feature gate is not needed since it is implemented for alpha features. 
BTW: I am coding for the feature implementation, but not very sure if i can have enough time to write e2e tests for this feature this quarter, is it ok to let it go to beta ?  or maybe alpha first ?

jsafrane:
> What about StorageClass protection? should we add it too ?

I'd skip it for now. We might bring a table what needs to be protected to sig-storage.

> But if PV protection goes directly to beta, feature gate is not needed since it is implemented for alpha features

Not really, admin can disable beta features if they're buggy or annoying.

NickrenREN:
>I'd skip it for now. We might bring a table what needs to be protected to sig-storage.

Ok, agree, cool

>Not really, admin can disable beta features if they're buggy or annoying.

oh, great to know it, thanks. so we can let it go to beta directly without e2e tests ?

jsafrane:
Regarding tests...  As feature owner you're responsible that there are enough e2e tests to cover the feature. You don't need to write them, find someone else who writes them. There is a large community waiting for their first Kubernetes commit :-).

NickrenREN:
Ok, i will try to add e2e tests myself. If i really do not have enough time, will find someone to help me.
Thanks @jsafrane.

liggitt:
> We don't want extra admission plugin for each class of objects we want protect - they need to be explicitly mentioned on API server command line and it's hard to maintain.

soon you will not need to maintain an explicit list of admission plugins on the command line. we're moving to a model where there is a default set of admission plugins, and you can enable things not in that default set if you want, or disable things in that set if you don't want them (https://github.com/kubernetes/kubernetes/pull/58123)

jsafrane:
/assign @saad-ali @childsb 
for approval

childsb:
This LGTM, but i'll leave it for @saad-ali to give any final comment/feedback.

PR #1619, opened 2018-01-18
Formalize volume topology specification

msau42:
Describing the PV NodeAffinity here instead of in the local storage design document, since this is more generic and can apply to any volume type.

### DIFF
--- a/contributors/design-proposals/storage/volume-topology-scheduling.md
+++ b/contributors/design-proposals/storage/volume-topology-scheduling.md
@@ -78,16 +78,158 @@ pod placement, volume binding and provisioning should be more tightly coupled wi
 pod scheduling.
 
 
-## Background
-In 1.7, we added alpha support for [local PVs](local-storage-pv.md) with node affinity.
-You can specify a PV object with node affinity, and if a pod is using such a PV,
-the scheduler will evaluate the PV node affinity in addition to the other
-scheduling predicates.  So far, the PV node affinity only influences pod
-scheduling once the PVC is already bound.  The initial PVC binding decision was
-unchanged.  This proposal addresses the initial PVC binding decision.
+## New Volume Topology Specification
+To specify a volume's topology constraints in Kubernetes, the PersistentVolume
+object will be extended with a new NodeAffinity field that specifies the
+constraints.  It will closely mirror the existing NodeAffinity type used by
+Pods, but we will use a new type so that we will not be bound by existing and
+future Pod NodeAffinity semantics.
+
+```
+type PersistentVolumeSpec struct {
+    ...
+
+    NodeAffinity *VolumeNodeAffinity
+}
+
+type VolumeNodeAffinity struct {
+    // The PersistentVolume can only be accessed by Nodes that meet
+    // these required constraints
+    Required *NodeSelector
+}
+```
+
+The `Required` field is a hard constraint and indicates that the PersistentVolume
+can only be accessed from Nodes that satisfy the NodeSelector.
+
+In the future, a `Preferred` field can be added to handle soft node contraints with
+weights, but will not be included in the initial implementation.
+
+The advantages of this NodeAffinity field vs the existing method of using zone labels
+on the PV are:
+* We don't need to expose first-class labels for every topology domain.
+* Implementation does not need to be updated every time a new topology domain
+  is added to the cluster.
+* NodeSelector is able to express more complex topology with ANDs and ORs.
+
+Some downsides include:
+* You can have a proliferation of Node labels if you are running many different
+  kinds of volume plugins, each with their own topology labeling scheme.
 
 
-## Design
+### Example PVs with NodeAffinity
+#### Local Volume
+```
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  Name: local-volume-1
+spec:
+  capacity:
+    storage: 100Gi
+  storageClassName: my-class
+  local:
+    path: /mnt/disks/ssd1
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname
+          operator: In
+          values:
+          - node-1
+```
+
+#### Zonal Volume
+```
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  Name: zonal-volume-1
+spec:
+  capacity:
+    storage: 100Gi
+  storageClassName: my-class
+  gcePersistentDisk:
+    diskName: my-disk
+    fsType: ext4
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: failure-domain.beta.kubernetes.io/zone
+          operator: In
+          values:
+          - us-central1-a
+```
+
+#### Multi-Zonal Volume
+```
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  Name: multi-zonal-volume-1
+spec:
+  capacity:
+    storage: 100Gi
+  storageClassName: my-class
+  gcePersistentDisk:
+    diskName: my-disk
+    fsType: ext4
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: failure-domain.beta.kubernetes.io/zone
+          operator: In
+          values:
+          - us-central1-a
+          - us-central1-b
+```
+
+### Default Specification
+Existing admission controllers and dynamic provisioners for zonal volumes
+will be updated to specify PV NodeAffinity in addition to the existing zone
+and region labels.  This will handle newly created PV objects.
+
+Existing PV objects will have to be upgraded to use the new NodeAffinity field.
+This does not have to occur instantaneously, and can be updated within the
+deprecation period.
+
+TODO: This can be done through one of the following methods:
+- Manual updates/scripts
+- cluster/update-storage-objects.sh?
+- A new PV update controller
+
+### Bound PVC Enforcement
+For PVCs that are already bound to a PV with NodeAffinity, enforcement is
+simple and will be done at two places:
+* Scheduler predicate: if a Pod references a PVC that is bound to a PV with
+NodeAffinity, the predicate will evaluate the `Required` NodeSelector against
+the Node's labels to filter the nodes that the Pod can be schedule to.  The
+existing VolumeZone scheduling predicate will coexist with this new predicate
+for several releases until PV NodeAffinity becomes GA and we can deprecate the
+old predicate.
+* Kubelet: PV NodeAffinity is verified against the Node when mounting PVs.
+
+### Unbound PVC Binding
+As mentioned in the problem statement, volume binding occurs without any input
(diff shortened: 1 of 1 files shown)


msau42:
/assign @thockin @jsafrane 

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -78,16 +78,139 @@ pod placement, volume binding and provisioning should be more tightly coupled wi
...
+          operator: In
+          values:
+          - us-central1-a
+          - us-central1-b
+```
+
+### Bound PVC Enforcement
+For PVCs that are already bound to a PV with NodeAffinity, enforcement is
+simple and will be done at two places:
+* Scheduler predicate: if a Pod references a PVC that is bound to a PV with
```
jsafrane: How is it / will it be related to existing predicate that evaluates hard-coded `failure-domain.beta.kubernetes.io/zone`? I assume that both old and new will be enabled then PVC.NodeAffinity feature gate is enabled. If the old predicate is going away, we need some admission plugin that automatically fills PVC.NodeAffinity on upgrade.

msau42: Ah good point.  They both will have to coexist until this becomes GA + deprecation period, but eventually we should remove the old zone predicate.

Can an admission plugin upgrade existing objects that don't get updated by the user?  Maybe we could do something similar like the pvc/pv protection controller and fill in the PV.NodeAffinity.

msau42: Added a section on defaults above

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -78,16 +78,158 @@ pod placement, volume binding and provisioning should be more tightly coupled wi
...
+and region labels.  This will handle newly created PV objects.
+
+Existing PV objects will have to be upgraded to use the new NodeAffinity field.
+This does not have to occur instantaneously, and can be updated within the
+deprecation period.
+
+TODO: This can be done through one of the following methods:
+- Manual updates/scripts
+- cluster/update-storage-objects.sh?
+- A new PV update controller
```
thockin: This should be automatic unless absolutely impossible?

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -78,16 +78,158 @@ pod placement, volume binding and provisioning should be more tightly coupled wi
...
+existing VolumeZone scheduling predicate will coexist with this new predicate
+for several releases until PV NodeAffinity becomes GA and we can deprecate the
+old predicate.
+* Kubelet: PV NodeAffinity is verified against the Node when mounting PVs.
+
+### Unbound PVC Binding
+As mentioned in the problem statement, volume binding occurs without any input
+about a Pod's scheduling constraints.  To fix this, we will delay volume binding
+and provisioning until a Pod is created.  This behavior change will be opt-in as a
+new StorageClass parameter.
```
thockin: link to some other discussion about the late-provisioning?

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: Are you saying that there is already a general object upgrade/backfill mechanism that exists in Kubernetes today?  As far as I know, there isn't, so if we want it to be automatic, then the best way I can think of is to add a PV update controller.

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: The detailed discussion is in this same document further down.  Are there any missing details that should be added?

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: Just had a discussion with @caesarxuchao.  api-machinery is working on an update controller, however it's for updating objects between two different versions.  Since this is just adding a new field in the same version, then their solution won't work for us.

So we discussed a possible solution for our use case:
- PV update controller to copy the zone/region labels from existing PVs and specify them in the new field
- Admission controller to reject conflicting updates (ie changing the zone in the label, but not the field)

saad-ali:
The `StroageClass` currently has no first class field to represent topology. This proposal adds `nodeAffinity` to storage class to do that. Let's make sure that we capture all the use cases that it will be replacing:

Today AWS and GCE provisioners allow setting topology via the opaque parameters list. Example:
```YAML
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: slow
provisioner: kubernetes.io/gce-pd
parameters:
  type: pd-standard
  zones: us-central1-a, us-central1-b
```

They accept the following parameters:
1. `zone: <a single GCE zone>`
    * Volume provisioned in specified zone.
    * Provisioning fails if zone is invalid.
2. `zones: <comma separated list of GCE zone(s)>`
    * Volume provisioned in zone randomly selected [1] from specified zones.
    * Provisioning fails if selected zone is invalid.
3. Neither zone or zones specified
    * Volume provisioned in zone randomly selected [1] from all active zones where Kubernetes cluster has a node.
    * Provisioning fails if selected zone is invalid.

[1] - The provisioner hashes the PVC name and use that to select the zone (so the same PV always "randomly" snaps to the same zone). If a PVC name ends with `-<integer>`, the provisioner hashes the prefix, and then add the integer to the hash. This allows a StatefulSet's volumes (`claimname-statefulsetname-id`) to be spread across availability zones (assuming the id values are consecutive).

To support 1, a standard `nodeAffinity` selecting on zone will work.
To support 3, the lack of a `nodeAffinity` will work.
To support 2, we would need the ability for `nodeAffinity` to represent OR statements. Does this ability exist?

Also it may be worth documenting these use cases explicitly in the proposal.

lichuqiang:
>To support 2, we would need the ability for nodeAffinity to represent OR statements. Does this ability exist?

I think `NodeSelectorOpIn`("In") may work on the case.

msau42:
@saad-ali actually this proposal adds NodeAffinity on the PV, not the StorageClass.  So currently, this proposal is only handling specification of topology for already provisioned PVs.

The use cases you outline are for restricting topology during dynamic provisioning.  I would argue that one of the main reasons we have the zone parameters in the first place is to workaround the issue that dynamic provisioning is not in sync with the pod's topology restrictions.  So once we have dynamic provisioning integrated with pod scheduling, then the provisioning can be directed by the pod's policies, and there is no need for an admin to manually restrict provisioning to align with those pod policies.

Maybe there are still some use cases where you would still like to restrict the storage class.  @lichuqiang and I are working on the dynamic provisioning design that should be able to accommodate this through StorageClass capacity reporting per topology dimension.

saad-ali:
Users should have some way to be able to control where storage is provisioned.

I can buy the argument that if dynamic provisioning will be controlled by the scheduler, then the scheduler would specify the topological constraints of a volume to the provisioner. And, in that case, the user could control where storage is provisioned by setting node affinity.

However, I have a few concerns:
1. How exactly does the scheduler determine the topological constraints that should apply to a volume to provision?
2. How does the scheduler communicate the topological constraints selected to the provisioner?
3. If late binding is not enabled (it is not enabled by default), what mechanism will end users have to control where storage is provisioned? Depending on the answer to 2 this may be solved by the same mechanism.

msau42:
1. Scheduler determines the constraints based on the pod's policies: resource requests, nodeaffinity, pod anti-affinity, etc.  + the constraints of the available static PVs + provisionable capacity (ie how much capacity is available on each node, each zone, etc)
2. We're thinking the scheduler will tell the provisioner its chosen node.  It's up to the provisioner to translate that node into their topology dimension (node, zone, rack, region, etc)
3. If late binding is not enabled, then users have to use the existing method today of specifying zones in the StorageClass.

saad-ali:
> 1. Scheduler determines the constraints based on the pod's policies: resource requests, nodeaffinity, pod anti-affinity, etc. + the constraints of the available static PVs + provisionable capacity (ie how much capacity is available on each node, each zone, etc)

I think what I'm missing here concretely how this will happen: Let's say we allow arbitrary labels on a node from various storage plugins. The scheduler will use the pod policies to narrow down the list of nodes that it wants to schedule to. At that point, how does the scheduler take that and translate it in to something it can tell the provisioner (i.e. realize that the selected pods all belong to the same zone, and pass that zone to the provisioner)?

> 2. We're thinking the scheduler will tell the provisioner its chosen node. It's up to the provisioner to translate that node into their topology dimension (node, zone, rack, region, etc)

Ah ok, that answers the question above: so the scheduler will select a node from the list of possible nodes, and then pass the selected node to the provisioner and let the provisioner do the translation from node to topological info.

So next question is how exactly will the node selection be communicated with the provisioner? And we'll still need to figure out how the translation from node to toplogy will happens in the provisioner (most likely the provisioner can read the labels from the node object for the corresponding plugin and use that).

> 3. If late binding is not enabled, then users have to use the existing method today of specifying zones in the StorageClass.

That's fine, we'll just need need to define what happens in the case where late binding is enabled but user also specifies zone in opaque StorageClass information. Feels like this should be a failure, but since topology won't be a first class field in StorageClass, we won't be able to use API validation.

msau42:
> So next question is how exactly will the node selection be communicated with the provisioner? And we'll still need to figure out how the translation from node to toplogy will happens in the provisioner (most likely the provisioner can read the labels from the node object for the corresponding plugin and use that).

Yup, that still needs to be ironed out in the dynamic provisioning design.  But I'm roughly thinking, in-tree plugins just add a new node argument to Provision() call, and we'll need some PVC annotation for external provisioners.  And the provisioners will have to lookup the node labels to translate to their topology key.

saad-ali:
Michelle and I spoke offline. We agreed, it would be useful to:
1. Add a way for cluster admin to override topology in `StorageClass` (new first class field, either string to string `topologies` map or `VolumeNodeAffinity`).
2. Have the scheduler communicate the selected node with the provisioner via a new `selectedNode` annotation on the `PersistentVolumeClaim` object.
3. Have the scheduler use the topology information in the `StorageClass` along with the `NodeAffinity` information from the pod when making scheduling decision.
4. Capture the `zone` vs `zones` vs `replica-zones` vs `replica-zones:auto` use cases and how they would work with the new fields.
5. Consider reporting capacity information based on topology key as part of a new `Status` field `StorageClass`.
6. Verify there is no overlap between this new capacity field and the existing `StroageClassQuota` object (one is cluster admin set on Kubernetes side, the other is actual available capacity as reported by the storage system).

liggitt:
> Have the scheduler communicate the selected node with the provisioner via a new selectedNode annotation on the PersistentVolumeClaim object.

PVC annotations are under the control of the PVC creator... is that what we want for this?

msau42:
Currently PVC annotations are the only way to communicate with external provisioners and the PV controller.  There already exists other external dynamic provisioning trigger annotations in the PVC today, so this would be further extending that model.

If the user wants to override the decision and fill in the annotation themselves, then they are free to do that, but we can't make guarantees that it can be successfully provisioned or scheduled.

PR #1623, opened 2018-01-18
Update SIG Architecture backlog

bgrant0607:


jdumars:
/lgtm
/approve
see: http://bit.ly/sig-architecture for context

PR #1624, opened 2018-01-18
Add Volume Snapshot Support Proposal

xing-yang:
This document proposes to add Kubernetes snapshot support
in-tree.


xing-yang:
Looks like I need to get the CLA issue resolved first.  There may be an email address mismatch.

xing-yang:
Will try to re-submit to resolve the cla issue.

PR #1638, opened 2018-01-20
Add Volume Snapshot Support

xing-yang:
This document proposes to add Kubernetes snapshot support in-tree.

### DIFF
--- a/contributors/design-proposals/storage/volume-snapshotting.md
+++ b/contributors/design-proposals/storage/volume-snapshotting.md
@@ -1,17 +1,21 @@
 Kubernetes Snapshotting Proposal
 ================================
 
-**Authors:** [Cindy Wang](https://github.com/ciwang)
+**Authors:** [Cindy Wang](https://github.com/ciwang), [Jing Xu](https://github.com/jinxu97),[Tomas Smetana](https://github.com/tsmetana), [Huamin Chen ](https://github.com/rootfs), [Xing Yang] (https://github.com/xing-yang)
 
 ## Background
 
-Many storage systems (GCE PD, Amazon EBS, etc.) provide the ability to create "snapshots" of a persistent volumes to protect against data loss. Snapshots can be used in place of a traditional backup system to back up and restore primary and critical data. Snapshots allow for quick data backup (for example, it takes a fraction of a second to create a GCE PD snapshot) and offer fast recovery time objectives (RTOs) and recovery point objectives (RPOs).
+Many storage systems (GCE PD, Amazon EBS, etc.) provide the ability to create "snapshots" of persistent volumes to protect against data loss. Snapshots can be used in place of a traditional backup system to back up and restore primary and critical data. Snapshots allow for quick data backup (for example, it takes a fraction of a second to create a GCE PD snapshot) and offer fast recovery time objectives (RTOs) and recovery point objectives (RPOs).
 
 Typical existing backup solutions offer on demand or scheduled snapshots.
 
 An application developer using a storage may want to create a snapshot before an update or other major event. Kubernetes does not currently offer a standardized snapshot API for creating, listing, deleting, and restoring snapshots on an arbitrary volume.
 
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/volume-snapshotting.md
```diff
@@ -21,498 +25,529 @@ For the first version of snapshotting support in Kubernetes, only on-demand snap
...
+	// This is useful for single-node development and testing only!
+	// On-host storage is not supported in any way and WILL NOT WORK in a multi-node cluster.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#hostpath
+	// +optional
+	HostPath *HostPathVolumeSnapshotSource `json:"hostPath,omitempty"`
+	// AWSElasticBlockStore represents an AWS Disk resource that is attached to a
+	// kubelet's host machine and then exposed to the pod.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#awselasticblockstore
+	// +optional
+	AWSElasticBlockStore *AWSElasticBlockStoreVolumeSnapshotSource `json:"awsElasticBlockStore,omitempty"`
```
liggitt: I was expecting the cloud provider-related API and controller to be out of tree. Isn't the ideal scenario that any storage provider would be able to support snapshotting their storage, interacting with the core kubernetes API only via VolumeSnapshot/VolumeSnapshotData objects?

liggitt: to that end, it would probably make sense for there to be a place for opaque, storage-provider-specific parameters in the VolumeSnapshotData, so the storage provider can persist the info required to locate the snapshot (similar to the parameters in StorageClass and FlexVolumeSource)

kmova:
Most of the storage vendors already support volume level snapshots. It will be good to include a section on how these will be made available via the CSI drivers.

contributors/design-proposals/storage/volume-snapshotting.md
xing-yang: Thanks @liggitt for the comments.  This design is modeled after the design of PersistentVolume and PersistentVolumeClaim. For example, PersistentVolumeSource contains storage provider data as follows:

type PersistentVolumeSource struct {
// GCEPersistentDisk represents a GCE Disk resource that is attached to a
// kubelet's host machine and then exposed to the pod.
// +optional
GCEPersistentDisk *GCEPersistentDiskVolumeSource

// AWSElasticBlockStore represents an AWS EBS disk that is attached to a
// kubelet's host machine and then exposed to the pod
// +optional
AWSElasticBlockStore *AWSElasticBlockStoreVolumeSource
...
}

Having opaque, storage-provider-specific parameters in the VolumeSnapshotData is an option we can discuss about.

contributors/design-proposals/storage/volume-snapshotting.md
```diff
@@ -1,18 +1,22 @@
 Kubernetes Snapshotting Proposal
 ================================
 
-**Authors:** [Cindy Wang](https://github.com/ciwang)
+**Authors:** [Cindy Wang](https://github.com/ciwang), [Jing Xu](https://github.com/jinxu97),[Tomas Smetana](https://github.com/tsmetana), [Huamin Chen ](https://github.com/rootfs), [Xing Yang] (https://github.com/xing-yang)
```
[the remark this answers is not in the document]
xing-yang: Will address it.  Thanks.

xing-yang:
@kmova Sure, will add a section for that later.

xing-yang:
Since I re-signed CLA with a new email address, I need to close this PR and open a new PR to get the CLA issue resolved.  Thanks.

cblecker:
@xing-yang You shouldn't need to. You should also be able to follow directions like https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit to reset the author on your commits to match your e-mail address

xing-yang:
@cblecker Thanks.  I tried that in my other PR's but couldn't resolve it until I re-submit a new one:(.  I have got help from Linux Foundation help desk as well.

Here's the new PR with CLA issue resolved: https://github.com/kubernetes/community/pull/1662

Issue #1413, opened 2017-11-16
Umbrella issue for contributor's guide


guineveresaenger:
As per https://github.com/kubernetes/website/issues/6102 we need a contributor guide that is easily found, complete, and organized logically. 

In https://github.com/kubernetes/community/pull/1409, a /guide folder is being created in k/community/contributors, and its README is set up to be the central starting point for contributing.

This issue is an umbrella tracking issue. Below is a list of check boxes for individual items that need fixing. Some are easy, others require some expertise/experience as a contributor. 

**To help out:**
1. Pick a checkbox item. You will find prompts in the README to where these improvements are needed; they should mostly parallel these check points.
2. Ping me or this issue so I can update this comment next to the checkpoint to say that you are working on this.
3. When finished with the item, submit a PR. In many cases, we will want to re-write, move, and/or remove a linked reference. This should be reflected in the PR. All contributing specific information (i.e. GitHub workflow) should ultimately live in k/community/contributors/guide. _The ultimate goal is to have a single source of truth for all these topics._
4. Comment on your PR with a note like "...in partial completion of..." and refer to this issue.
5. Follow up with me to make sure I crossed off the checkbox as completed.

**Pages to Update:**
https://github.com/kubernetes/community/tree/master/contributors/guide/README.md
https://github.com/kubernetes/community/tree/master/contributors/guide

**List of tasks**
- [x] kubernetes/kubernetes/contributing.md -> point to community/contributors/guide/README.md
_Thank you, @anubhakushwaha!_

- [x] kubernetes/community/CONTRIBUTING.md -> Needs a rewrite to reflect only k/community-specific contributing guidelines

- [x] kubernetes/community/README.md -> Needs a rewrite to talk about k/community specifically and what it does

- [x] Individual SIG contributing documents -> add a link to community/contributors/guide/README.md _List of SIGs with contributing docs in_ [comment below](https://github.com/kubernetes/community/issues/1413#issuecomment-358127594).

- [x] Generate TOC for community/contributors/guide/README.md

- [x] Document should point at developer guide _being worked on by RyanJ from RedHat (?)_ @castrojo _there's a link to resources pointing at top level README in /devel_

- [x] consolidate the Code Review principles from several sources, and find a good place in /guide/README.md where it flows best

[...]

PR #1646, opened 2018-01-22
Moving and revising community expectations

castrojo:
This is part of #1413 

This moves the community expectations from the old place to the new contributor guide. I've given it a once over, removed the duplicated content that is stated elsewhere (we don't need every single page to mention the CoC for example), and I've grabbed things that were in `collab.md` and integrated it into this document.

We will not need to bring collab.md over to the contributor guide.

### DIFF
--- a/contributors/guide/community-expectations.md
+++ b/contributors/guide/community-expectations.md
@@ -22,35 +22,13 @@ As a community we believe in the value of code review for all contributions.
 Code review increases both the quality and readability of our codebase, which
 in turn produces high quality software.
 
-However, the code review process can also introduce latency for contributors
-and additional work for reviewers that can frustrate both parties.
+See the [pull request documentation](/devel/pull-requests.md) for more information
+on code review.
 
 Consequently, as a community we expect that all active participants in the
 community will also be active reviewers. The 
-[community membership](../community-membership.md) outlines the responsibilities
+[community membership](/community-membership.md) outlines the responsibilities
 of the different contributor roles. 
-
-All changes must be code reviewed. For non-maintainers this is obvious, since
-you can't commit anyway. But even for maintainers, we want all changes to get at
-least one review, preferably (for non-trivial changes obligatorily) from someone
-who knows the areas the change touches. For non-trivial changes we may want two
-reviewers. The primary reviewer will make this decision and nominate a second
-reviewer, if needed. Except for trivial changes, PRs should not be committed
-until relevant parties (e.g. owners of the subsystem affected by the PR) have
-had a reasonable chance to look at PR in their local business hours.
-
-Most PRs will find reviewers organically. If a maintainer intends to be the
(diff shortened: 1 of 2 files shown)


contributors/guide/community-expectations.md
```diff
@@ -16,53 +16,56 @@ that shapes the rest of the project.  It currently covers the expectations
...
-be an active reviewer can also add their name to an *active reviewer* file
-(location tbd) which will make them an active reviewer for as long as they
-are included in the file.
-
-#### Expectations of reviewers: Review comments
+community will also be active reviewers. The 
+[community membership](../community-membership.md) outlines the responsibilities
+of the different contributor roles. 
+
+All changes must be code reviewed. For non-maintainers this is obvious, since
```
cblecker: This paragraph isn't entirely accurate. It doesn't mention approvers/OWNERS, and it feels very light on the requirement for reviews. We should at least mention that the OWNERS process should be respected.

contributors/guide/community-expectations.md
```diff
@@ -16,53 +16,56 @@ that shapes the rest of the project.  It currently covers the expectations
...
+you can't commit anyway. But even for maintainers, we want all changes to get at
+least one review, preferably (for non-trivial changes obligatorily) from someone
+who knows the areas the change touches. For non-trivial changes we may want two
+reviewers. The primary reviewer will make this decision and nominate a second
+reviewer, if needed. Except for trivial changes, PRs should not be committed
+until relevant parties (e.g. owners of the subsystem affected by the PR) have
+had a reasonable chance to look at PR in their local business hours.
+
+Most PRs will find reviewers organically. If a maintainer intends to be the
```
cblecker: This also isn't accurate, as we have bots in most repos to do this.

contributors/guide/community-expectations.md
castrojo: A bunch of this stuff is just duplicated from the PR doc, so I think in the interest of brevity I'm just going to cut it out of here and then link to that.

contributors/guide/community-expectations.md
castrojo: I'll remove all this and we'll just have all the bot docs in one place.

contributors/guide/community-expectations.md
```diff
@@ -16,53 +16,34 @@ that shapes the rest of the project.  It currently covers the expectations
...
-### Code review
-
-As a community we believe in the [value of code review for all contributions](collab.md).
+As a community we believe in the value of code review for all contributions.
 Code review increases both the quality and readability of our codebase, which
 in turn produces high quality software.
 
-However, the code review process can also introduce latency for contributors
-and additional work for reviewers that can frustrate both parties.
+See the [pull request documentation](/devel/pull-requests.md) for more information
```
cblecker: I think this should be `/contributors/devel/pull-requests.md`

PR #1680, opened 2018-01-25
Add Topology Manager proposal.

ConnorDoyle:
/sig node

As noted in the text, this proposal _**only**_ deals with aligning device and exclusive CPU assignment.

* @lmdaly
* @balajismaniam 
* @resouer 
* @sjenning 
* @derekwaynecarr 
* @jeremyeder 
* @vikaschoudhary16 
* @vishh 
* @RenaudWasTaken


contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+This proposal provides a mechanism to coordinate fine-grained hardware
+resource assignments for different components in Kubernetes.
+
+
+# Motivation
+
+Multiple components in the Kubelet make decisions about system
+topology-related assignments:
+
+- CPU manager.
```
vikaschoudhary16: extra '.'

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  - The CPU manager makes decisions about the set of CPUs a container is
+allowed to run on. The only implemented policy as of v1.8 is the static
+one, which does not change assignments for the lifetime of a container.
+- Device manager.
```
vikaschoudhary16: extra '.'

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  - The device manager makes concrete device assignments to satisfy
+container resource requirements. Generally devices are attached to one
+peripheral interconnect. If the device manager and the CPU manager are
+misaligned, all communication between the CPU and the device can incur
+an additional hop over the processor interconnect fabric.
+- Container network interface
```
vikaschoudhary16: s/network interface/Network Interface

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+- [Hardware topology awareness at node level (including NUMA)][k8s-issue-49964]
+- [Discover nodes with NUMA architecture][nfd-issue-84]
+- [Support VF interrupt binding to specified CPU][sriov-issue-10]
+- [Proposal: CPU Affinity and NUMA Topology Awareness][proposal-affinity]
+
+Note that all of these concerns pertain only to multi-socket systems.
+
+## Goals
+
+- Allow CPU manager and device plugin subsystem to agree on preferred
```
vikaschoudhary16: s/plugin subsystem/manager

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  NUMA node affinity for containers.
+- Provide an internal interface and pattern to integrate additional
+  topology-aware Kubelet components.
+
+## Non-Goals
+
+- _Inter-device connectivity:_ Decide device assignments based on direct
+  device interconnects. This issue can be separated from NUMA node
+  locality. Inter-device topology can be considered entirely within the
+  scope of the device plugin subsystem, after which it can emit possible
```
vikaschoudhary16: s/plugin subsystem/manager

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  NUMNA affinities. The policy to reach that decision can start simple
```
vikaschoudhary16: s/NUMA

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  and iterate toward support for arbitrary inter-device graphs.
```
vikaschoudhary16: s/toward/towards

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+
+### New Component: NUMA Manager
+
+This proposal is focused on a new component in the Kubelet called the
+NUMA Manager. The NUMA Manager implements the pod admit handler
+interface and participates in Kubelet pod admission. When the `Admit()`
+function is called, the NUMA manager collects NUMA hints from from other
+Kubelet components.
+
+If the NUMA hints do are not compatible, the NUMA
```
vikaschoudhary16: extra 'do'

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+1. Add two implementations of NUMA Manager interface and a feature gate.
+    1. As much NUMA Manager functionality as possible is stubbed when the
+       feature gate is disabled.
+    1. Add a functional NUMA manager that queries hint providers in order
+       to compute a preferred NUMA node mask for each container.
+1. Add `GetNUMAHints()` method to CPU Manager.
+    1. CPU Manager static policy calls `GetAffinity()` method of NUMA
+       manager when deciding CPU affinity.
+1. Add `GetNUMAHints()` method to Device Manager.
+    1. Add NUMA Node ID to Device structure in the device plugin
```
vikaschoudhary16: We might not want to add NUMA ID specifically in the device structure, instead use key-value attributes map:
https://github.com/kubernetes/kubernetes/pull/58388

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+from all providers, the NUMA Manager must choose some mask that is
+present in all lists. Here is a sketch:
+
+1. Apply a partial order on each list: number of bits set in the
+   mask, ascending. This biases the result to be more precise if
+   possible.
+1. Iterate over the permutations of preference lists and compute
+   bitwise-and over the masks in each permutation.
+1. Store the first non-empty result and break out early.
+1. If no non-empty result exists, return an error.
```
lmdaly: Should an error be returned in the case of non-strict numa assignment?

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  device interconnects. This issue can be separated from NUMA node
+  locality. Inter-device topology can be considered entirely within the
+  scope of the device plugin subsystem, after which it can emit possible
+  NUMNA affinities. The policy to reach that decision can start simple
+  and iterate toward support for arbitrary inter-device graphs.
+- _HugePages:_ This proposal assumes that pre-allocated HugePages are
+  spread among the available NUMA nodes in the system. We further assume
+  the operating system provides best-effort local page allocation for
+  containers (as long as sufficient HugePages are free on the local NUMA
+  node.
```
lmdaly: Future addition could be to extend cAdvisor to advertise huge pages available per NUMA node, enabling NUMA Manager to take huge pages available on each node into account when calculating NUMA node affinity

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Thanks for linking that PR. My preference would be to avoid using semi structured K-V for any values the Kubelet cares about. We can get some more opinions on that. Either way, fairly minor detail.

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Hm, good point! Want to add that as something we can do for beta? Just send a PR against my branch.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+An increasing number of systems leverage a combination of CPUs and
+hardware accelerators to support latency-critical execution and
+high-throughput parallel computation. These include workloads in fields
+such as telecommunications, scientific computing, machine learning,
+financial services and data analytics. Such hybrid systems comprise a
+high performance environment.
+
+In order to extract the best performance, optimizations related to CPU
+isolation and memory and device locality are required. However, in
+Kubernetes, these optimizations are handled by disjoint components.
```
dougbtv: Suggestion for legibility: instead of "disjoint components", either "disjointed components" or "a disjoint set of components".

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+1. Add `GetNUMAHints()` method to CPU Manager.
+    1. CPU Manager static policy calls `GetAffinity()` method of NUMA
+       manager when deciding CPU affinity.
+1. Add `GetNUMAHints()` method to Device Manager.
+    1. Add NUMA Node ID to Device structure in the device plugin
+       interface. Plugins should be able to determine the NUMA node
+       easily when enumerating supported devices. For example, Linux
+       exposes the node ID in sysfs for PCI devices:
+       `/sys/devices/pci*/*/numa_node`.
+    1. Device Manager calls `GetAffinity()` method of NUMA manager when
```
lichuqiang: Not sure if this is out of scope:

For Device Manager, if available NUMA hints returned here, it means we have enough devices for the container, and preparatory work has already been completed.

Thus, we may need to consider the relationship between things we do here and that of `pluginResourceUpdateFunc ` in `predicateAdmitHandler `(which could be duplicated)

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+function is called, the NUMA manager collects NUMA hints from from other
+Kubelet components.
+
+If the NUMA hints are not compatible, the NUMA manager could choose to
+reject the pod. The details of what to do in this situation needs more
+discussion. For example, the NUMA manager could enforce strict NUMA
+alignment for Guaranteed QoS pods. Alternatively, the NUMA manager could
+simply provide best-effort NUMA alignment for all pods.
+
+The NUMA Manager component will be disabled behind a feature gate until
```
derekwaynecarr: Is the NUMA manager only relevant in its first iteration if static cpu policy is enabled?

contributors/design-proposals/node/numa-manager.md
derekwaynecarr: I think I would prefer to start simple on hugepages behavior.  For roadmap, I don’t think anything special would be needed for alpha or even beta.

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Practically speaking, yes.
* static cpu policy
* G pod
* integer cpu request container
* consuming a device
* on a multi socket machine

We wanted to start with this narrow use case, but at the same time define a lightweight internal API for other NUMA concerns.

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Great point. Here are two options:
1. Make sure this admit handler always runs first
1. Eliminate the side effects of the device plugin’s admit handler and add some kind of post-admit hook to do the device allocation after all of the admit handlers decide to really admit the pod.

contributors/design-proposals/node/numa-manager.md
resouer: I am also not sure adding specific fields in DP, +1 for some general types (but not raw K-V, painful for programming ... )

contributors/design-proposals/node/numa-manager.md
iwankgb: In my opinion non-strict assignment is still non-erroneous operation unless user indicates that strict assignment is required.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+### New Component: NUMA Manager
+
+This proposal is focused on a new component in the Kubelet called the
+NUMA Manager. The NUMA Manager implements the pod admit handler
+interface and participates in Kubelet pod admission. When the `Admit()`
+function is called, the NUMA manager collects NUMA hints from from other
+Kubelet components.
+
+If the NUMA hints are not compatible, the NUMA manager could choose to
+reject the pod. The details of what to do in this situation needs more
```
iwankgb: According to [the docs](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L1117) if scheduler is unaware of the rejection rule then `softAdmitHandler` should be used. This is the case described here so any pod rejected at this phase will stay in `Pending` state indefinitely. 
Suggesting user to utilize a single-pod ReplicaSet (or any other controller capable of deleting and re-creating a pod) might be a good idea. Otherwise user might not be aware of failed scheduling.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  // GetAffinity returns the preferred NUMA affinity for the supplied
+  // pod and container.
+  GetAffinity(podName string, containerName string) NUMAMask
+}
+
+// HintProvider is implemented by Kubelet components that make
+// NUMA-related resource assignments. The NUMA manager consults each
+// hint provider at pod admission time.
+type HintProvider interface {
+  GetNUMAHints(pod v1.Pod, containerName string) []NUMAMask
```
iwankgb: When `GetNUMAHints()` returns empty slice of `NUMAMask` then calculating intersection of sets returned from all `HintProvider`s is impossible. The function should be able to indicate if the value returned indicates indifference (when device plugin does not manage requested resources) or inability to offer a hint matching user's requirements.

iwankgb: Perhaps:
```go
GetNUMAHints(pod v1.Pod, containerName string) []NUMAMask, bool
```

ConnorDoyle: So, you'd return `nil, false` to indicate "don't care"? That would work.

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Great points, will add.

contributors/design-proposals/node/numa-manager.md
iwankgb: We were just having a discussion about it with @ppalucki and we believe that there are following options:
* return `[]NUMAMask{}, false` when `HintProvider` is indifferent to pod spec and `[]NUMAMask{}, true` when `HintProvider` cannot find matching configuration
* return `[]NUMAMask{0b11}` when `HintProvider` is indifferent to pod spec and `[]NUMAMask{0b00}` when `HintProvider` cannot find matching configuration

Making a good choice here is about developer experience, I think - we should make the interface as straightforward to implement as possible.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+# Challenges
+
+* Testing the NUMA Manager in a continuous integration environment
+  depends on cloud infrastructure to expose multi-node NUMA topologies
+  to guest virtual machines.
+* Implementing the `GetNUMAHints()` interface may prove challenging.
+
+# Limitations
+
+* *TBD*
```
jeremyeder: Do you anticipate any limits to the number of NUMA nodes that would be supported?

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+
+# Alternatives
+
+* [AutoNUMA][numa-challenges]: This kernel feature affects memory
+  allocation and thread scheduling, but does not address device locality.
+
```
jeremyeder: It won't work for these use-cases if/until it's wired into PCI locality.  Don't think that's on anyone's radar (that I am aware of), even though it has been discussed at length at least inside RH.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+1. Add two implementations of NUMA Manager interface and a feature gate.
+    1. As much NUMA Manager functionality as possible is stubbed when the
+       feature gate is disabled.
+    1. Add a functional NUMA manager that queries hint providers in order
+       to compute a preferred NUMA node mask for each container.
+1. Add `GetNUMAHints()` method to CPU Manager.
+    1. CPU Manager static policy calls `GetAffinity()` method of NUMA
+       manager when deciding CPU affinity.
+1. Add `GetNUMAHints()` method to Device Manager.
+    1. Add NUMA Node ID to Device structure in the device plugin
```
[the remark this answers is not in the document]
lmdaly: Two options for this could be:
1.  NUMA Manager is only triggered on a multi-socket system. Perform check at NUMA Manager creation 
2. HintProviders will all return 1 (or maybe a -1) to NUMA Manager

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+- [Discover nodes with NUMA architecture][nfd-issue-84]
+- [Support VF interrupt binding to specified CPU][sriov-issue-10]
+- [Proposal: CPU Affinity and NUMA Topology Awareness][proposal-affinity]
+
+Note that all of these concerns pertain only to multi-socket systems.
+
+## Goals
+
+- Allow CPU manager and Device Manager to agree on preferred
+  NUMA node affinity for containers.
```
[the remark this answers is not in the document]
lmdaly: CPU Manager and Device Manager send back the user requested resources NUMA node availability, ie. Device Manager sends back [10] and CPU Manager sends back [01 10 11] - NUMA Manager will bitwise-and the masks and get a NUMA node preference of [10]. CPU Manager and Device Manager then consult the NUMA Manager when making the actual resource allocations. So neither influences the other directly.

contributors/design-proposals/node/numa-manager.md
lmdaly: I don't think it's called out specifically - but I think a Kubelet flag for strict/preferred NUMA alignment could be useful. Some users may want to fail guaranteed pods if they cannot get NUMA affinity across resources, whereas some may want the pod to run regardless - with a preference for NUMA affinity where possible.

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: It would be limited by the width of the mask we use to represent preferences, but that's up to us. Probably a `uint64` to start. What is the highest node count you've seen in the wild?

jeremyeder: 8 sockets would be enough.

RenaudWasTaken:
With the current implementation of the Device Manager, the NUMA manager would need to be called after the Device Manager in the PodAdmitHandler list (devices for a pod are selected during admission).

I'm wondering what's the right model:
1) Have the NUMA Admit Handler always be the last Admit Handler
2) Have the device manager logic (device selection, gRPC calls, ...) be executed in `GetNUMAHints`

Neither option look good from a code point of view. Though from an architecture POV I do prefer 2) any ideas or suggestions on this?

contributors/design-proposals/node/numa-manager.md
XiaoningDing: Maybe a new pod annotation like kubernetes.io/RequireStrictNUMAAlignment ? The NUMA manager rejects a pod only when the pod has the annotation and NUMA manager cannot find a strict NUMA assignment for it.

XiaoningDing:
@RenaudWasTaken I think this design requires NUMA  admit handler to run before device manager and cpu manager (or any other NUMA hint providers). 

About the second option, are you saying doing the Allocate() logic in GetNUMAHints() method? When GetNUMAHints() is called we don't know the final decision about NUMA placement yet, so I don't think we can do device selection logic there.

lmdaly:
@RenaudWasTaken what is your thinking behind NUMA Manager being called after Device Manager?
 as @XiaoningDing said - the GetNUMAHints() call should only return a list of devices and their sockets for a requesting container - no device allocation decisions are made at this time. 

@XiaoningDing agreed that we may have to call the NUMA admit handler before any hint providers.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  and iterate toward support for arbitrary inter-device graphs.
+- _HugePages:_ This proposal assumes that pre-allocated HugePages are
+  spread among the available NUMA nodes in the system. We further assume
+  the operating system provides best-effort local page allocation for
+  containers (as long as sufficient HugePages are free on the local NUMA
+  node.
+- _CNI:_ Changing the Container Networking Interface is out of scope for
+  this proposal. However, this design should be extensible enough to
+  accommodate network interface locality if the CNI adds support in the
+  future. This limitation is potentially mitigated by the possiblity to
```
fsoppelsa: possibility

contributors/design-proposals/node/numa-manager.md
jeremyeder: @kad this is interesting observation because I sort of came to this conclusion working through how I would partition a system for dejittering.

The way RH does this is through the use of a tuned profile called "cpu-partitioning".  It also uses cpu sets.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  to guest virtual machines.
+* Implementing the `GetNUMAHints()` interface may prove challenging.
+
+# Limitations
+
+* *TBD*
+
+# Alternatives
+
+* [AutoNUMA][numa-challenges]: This kernel feature affects memory
```
vishh: I'd like to see the performance of autonuma with CPU affinity before adding support for explicit NUMA node pinning. 
If a pod can be restricted to a single CPU core and a set of devices accessible from that CPU socket, I really wonder if autonuma would *just work*.

RenaudWasTaken: Hello @vishh !

I looked a bit into autonuma as I wasn't aware of it, [from what I understand of it](https://lwn.net/Articles/488709/):
> When the time comes for the scheduler to make a decision, it passes over the per-process statistics to determine whether the target process would be better off if it were moved to another node. If the process seems to be accessing most of its pages remotely, and it is better suited to the remote node than the processes already running there, it will be migrated over.

It seems to me that what it is trying to solve is maximizing locality of CPU node and Memory, which might be a good thing for certain use cases of this design but wouldn't solve the problem for devices as they are "sending" memory from the NUMA node to the device.

Please correct me if I misunderstood something :)

RenaudWasTaken:
> @RenaudWasTaken what is your thinking behind NUMA Manager being called after Device Manager?
as @XiaoningDing said - the GetNUMAHints() call should only return a list of devices and their sockets for a requesting container - no device allocation decisions are made at this time.

Yep that sounds like a better model than what I was imagining, somehow I got into thinking that we needed to do device assignment before.

contributors/design-proposals/node/numa-manager.md
jeremyeder: as soon as a process is launched with sched_setaffinity, autonuma no longer pays attention to those pids.  IOW cpumanager disables autonuma for the pids in that cgroup.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,281 @@
...
+  locality. Inter-device topology can be considered entirely within the
+  scope of the Device Manager, after which it can emit possible
+  NUMA affinities. The policy to reach that decision can start simple
+  and iterate to include support for arbitrary inter-device graphs.
+- _HugePages:_ This proposal assumes that pre-allocated HugePages are
+  spread among the available NUMA nodes in the system. We further assume
+  the operating system provides best-effort local page allocation for
+  containers (as long as sufficient HugePages are free on the local NUMA
+  node.
+- _CNI:_ Changing the Container Networking Interface is out of scope for
```
Levovar: if I may chime-in from Nokia, representing the often-quoted "NFVI use-cases" :)
could you elaborate a little bit more on the topic of interworking between CNI specification, CNI plugins and the newly introduced NUMA manager component? Possibly not relying on the Device Manager component. 

To put the question into context: in a performance sensitive VNF one of the main requirement is to 1: have isolated (what this means exactly is a topic of another discussion), exclusive CPU(s) allocated to a Pod 2: said CPU(s) shall belong to the same NUMA node as the network devices used by the Pod
However, network devices can be managed and consumed via a number of different ways, both implicit and explicit. The PFs can be directly pushed to the Pod's netns, VFs can be created on top of PFs and then pushed into the pod's netns, or the PF can be further divided via other means like a DPDK-capable OVS, or VPP; managing virtiousers and vhostusers. (VF and PF control planes can be also bound to DPDK kernel drivers in which case sharing is done via their PCI address; just to complicate the situation even more).

Most of these management tasks are handled by CNI plugins, namely SRIOV (either huscat's or Intel's DPDK capable fork), vhostuser plugin, or any other proprietary CNI implementation. The way how the Device Manager proposal shares devices (through their Unix socket) does not necessarily make it compatible with NFVI requirements, which in turn raises the question that how can be the current NUMA manager proposal enhanced to satisfy above requirements if the DeviceManager does not, or cannot play a role in these use-cases?

So, trying to distill this train of thought into two short questions, or even additional requirements:
- Dynamic, exclusive CPU allocation needs to be aligned with the configured CNI plugin; can the current NUMA manager proposal handle this UC in the future, and if yes, how?
- If not, have you considered the possibility of CPU allocation based on static NUMA request, e.g. "give me X exclusive CPUs from NUMA node A, and Y exclusive CPUs from NUMA node B?" (This mode would enable operators to at least manually align CPUs and NICs in a homogen cluster)

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: > If a pod can be restricted to a single CPU core and a set of devices accessible from that CPU socket  I really wonder if autonuma would just work.

We don't yet have a way to know what NUMA node devices live on, or align that with CPU affinity. I agree with the above, but the precondition is what this proposal tries to address. We're still not suggesting to configure `cpuset.mems`

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Added a note on opting in to strict mode.

kad:
To bring comments from another discussion related to this proposal: NUMA id vs. CPU set.

cpuset is native entity for current CPU manager, for kernel/cgroups (`sched_setaffinity`), for OCI runtime spec (https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cpu) , etc.
Just NUMA node number can be only in reality usable for setting memory allocation policy (`mbind`, `set_mempolicy`).

In most cases, passing just NUMA node between components will complicate implementation in various ways:
- plugin should be able to find mapping between CPUs and NUMA node numbers.
- CPU manager will require to map back from NUMA number to CPU set.
- on the systems with only one numa nodes it will be special tricks if kernel driver returns numa node = 0 or -1. For same hardware scenario CPU set will not have those complications.

Having in hint provider result only NUMA node id will also block implementation of use cases where device plugin can handle fine-tuned resource allocations per device. E.g. 
- 64 virtual functions for one physical device: plugin can give hint about partitioning related CPU resources.
- if device plugin want to hint that "by using that device, it must be used with exclusive non-HT cores", this will be achievable by plugin returning CPUset, but not achievable with just node id. 

As bottom line, I think if we start thinking about introducing topology awareness, then it will be good that device plugins can give hints about all topologically depended resources specified in OCI runtime spec:

- cpu.cpus hint bitmask that can be utilized together with existing CPU manager and then passed to runtime (https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cpu )
- cpu.mems  for memory allocation policies https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cpu 
- hugepages info: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#huge-page-limits

potentially as well in the long run: 
- cache allocation policies: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#intelrdt
- RDMA https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#rdma

lmdaly:
@kad apologies for the delayed reply! Having thought some more on this I have some follow up questions.

> plugin should be able to find mapping between CPUs and NUMA node numbers.

   The plugins can do this, but is this extra overhead for each plugin to have to find the CPUSet? Most devices have a numa node already associated so it is easy for the plugins to discover this.

> CPU manager will require to map back from NUMA number to CPU set.

   I would see this as a minor overhead for CPU Manager as it already has the NUMA topology stored.

> on the systems with only one numa nodes it will be special tricks if kernel driver returns numa node = 0 or -1. For same hardware scenario CPU set will not have those complications.

 Could you elaborate some more on the special tricks needed here? 

> if device plugin want to hint that "by using that device, it must be used with exclusive non-HT cores", this will be achievable by plugin returning CPUset, but not achievable with just node id.

I would be of the opinion that a device plugin shouldn't be responsible for this kind of information, there should be some abstraction that the user can request that their workload needs a device and exclusive CPUs and Kubernetes will facilitate this. 

Overall, I think adding all these capabilities will result in a too complex implementation. Though if the community strongly backs the need for this granularity of information then I am open to a design change.

contributors/design-proposals/node/numa-manager.md
```diff
@@ -0,0 +1,293 @@
+# NUMA Manager
```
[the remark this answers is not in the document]
ConnorDoyle: Added hugepages alignment to graduation criteria for beta.

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Added hugepages alignnment to the beta graduation criteria. If it's the case that no action is required (per performance test results) then it can be marked done via proper documentation of that fact.

contributors/design-proposals/node/numa-manager.md
ConnorDoyle: Would you all be OK with the following approach?

1. Generalize the hint structure from `NUMAHint` to something like `TopologyHints`.
1. `TopologyHints` structure has a field for `SocketID`

The material in this proposal, and known use cases, are really more about alignment with a physical socket and e.g. it's local PCI complex than "NUMA". That way, if we find use cases for specifically talking about memory nodes, or sub-socket CPU topologies, we can extend the hint structure.

We may even want to rename the component from "NUMA Manager" to something more generic like "Topology Manager" accordingly.

@kad @jeremyeder

contributors/design-proposals/node/numa-manager.md
Levovar: Considering that my colleagues already tested their DPDK based workload on vanilla(ish) kubernetes with CPUs and SRIOV VFs manually aligned, then observed that their unaligned hugepages caused additional packet processing delays in the 100ms range, I think this won't be an issue :)

contributors/design-proposals/node/numa-manager.md
jeremyeder: Agreed, since the eventual intent is to handle things like PCI locality on behalf of NVLINK or NICs, and that's only related to NUMA itself in the context of Intel's current chip design.

contributors/design-proposals/node/numa-manager.md
jeremyeder: Where is this test?

Levovar: in our lab, with our 4G "Cloudified" BTS application

but I'm fairly sure the same can be easily reproduced with testPMD too (I didn't put effort into it, so it is just an assumption from my side)

contributors/design-proposals/node/numa-manager.md
kad: "Topology Manager" sounds to me a lot better than "NUMA Manager".
Generalising topology hints as a struct makes sense. However, `SocketID` doesn't really make sense.
As a customer of that API in device plugin, for me it is hard to get which `SocketID` it will be. It will be a lot easier to read corresponding to device sysfs `local_cpulist` entry (e.g. `/sys/devices/pci0000:80/0000:80:01.0/0000:81:00.0/local_cpulist`) and pass it up to `Topology Manager`.
If `Topology Manager` policies will be interested in some kind of grouping of those, it can be done outside of plugins.

Making hints structure of containing two hints fields (cpu set and memory set, similar like it is defined in 
https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cpu for fields `cpu.cpus` and `cpu.mems`) will allow to cover universally any topology for both IO critical or Memory critical alignments.

ConnorDoyle: The points of mistakenly referring to a socket as a single NUMA node, and also the likely need for future kinds of locality hints are well-taken. Thanks for bringing those up.

In the interest of maintaining momentum on this proposal and leaving the door open to further hints, I've changed the naming and suggested API to a struct of hints as described above. We can explore more granular hints and related semantics in subsequent proposals. This is a deep topic. The semantics of more granularity are going to require some extended discussions, and currently the Kubelet does not set `cpuset.mems` at all, so it may be a stretch to get that (unenforceable) piece into the API at this point. At the same time, most participants in the discussion have been supportive / in agreement about the goal of at least socket-level alignment. We discussed using the CPU as a reference for internal node topology way back during the first resource management wg face-to-face in early 2017. Hopefully everyone is OK with this compromise for the first phase.

RenaudWasTaken:
I'm currently looking at this from a Device Plugin Author perspective.

To implement `GetTopologyHints()` I would expect to be aware of the current state to make a decision about which sockets should be used for this pod.
This is currently an issue because of the "lazy GPU reclaim" strategy, I will not be aware of the latest state until I actually get the `Allocate` call that this pod will trigger. 

What do you think?

ConnorDoyle:
@RenaudWasTaken perhaps the text needs clarification, but the expectation is that each Device descriptor from the plugin contains a socket ID. The device manager chooses the device to allocate based on this information, together with hints from other hint providers. Only after that does the allocate call occur. Does that work for you, or are you "faking" the device IDs in order to delay the choice of actual device to bind in your plugin?

RenaudWasTaken:
So I'll try to illustrate my thinking with a theoretical example, imagine a two NUMA node machine with one GPU on each node.

Story #1: The node receives two pod requesting one GPU each. 

For the first request, I expect the device plugin to return both sockets as a possible option. For the request I expect the device plugin to return only one socket.

Story #2: The node receives one pod requesting one GPU, that pod runs and terminates normally, the node then requests another pod requesting one GPU.

For the first request, I expect the device plugin to return both sockets as a possible option. For the request I also expect the device plugin to return both sockets as a possible option.


This is problematic today because of the following issues:
- The plugin is not aware of the socket chosen, this one is easy to fix, we can send the information
- The plugin is not aware of when a pod terminates, this is done "lazily" when a new pod comes in, the device manager looks at the active pods and reclaims GPUs used by pods which aren't active anymore.

Do these example make sense to you @ConnorDoyle ? Or am I misunderstanding the design doc?

ConnorDoyle:
Hi @RenaudWasTaken -- Based on the question, it looks like there might be a misunderstanding. The device plugin itself does not return hints. The Device Manager (internal to the Kubelet) implements the hint provider interface. It supplies info based on the set of unallocated devices and their declared affinity. So, the topology negotiation occurs before the plugin receives any allocate callback. I can update the text to make it clearer, and also outline more explicitly how the device plugin API would be affected. In short, the `Device` proto gets a field like `TopologyInfo` (mod naming) that initially contains just one member; the Socket ID.

From there, your scenario number 2 is supported because although the device plugin does not yet know that the pod is terminal and the device is available for a subsequent allocation, the Kubelet does know this and specifically the DeviceManager can accurately respond to the "get hints" callback from the TopologyManager during pod admission.

derekwaynecarr:
per sig-node discussion, i think this document is implementable and we can do any other adjustments as part of that process.

/approve
/lgtm

contributors/design-proposals/node/topology-manager.md
```diff
@@ -0,0 +1,322 @@
...
+# Graduation Criteria
+
+## Phase 1: Alpha (target v1.13)
+
+* Feature gate is disabled by default.
+* Alpha-level documentation.
+* Unit test coverage.
+* CPU Manager allocation policy takes topology hints into account.
+* Device plugin interface includes socket ID.
+* Device Manager allocation policy takes topology hints into account.
```
vishh: We need some way to introspect decisions may be via structured logs?

contributors/design-proposals/node/topology-manager.md
```diff
@@ -0,0 +1,322 @@
...
+This proposal is focused on a new component in the Kubelet called the
+Topology Manager. The Topology Manager implements the pod admit handler
+interface and participates in Kubelet pod admission. When the `Admit()`
+function is called, the Topology Manager collects topology hints from other
+Kubelet components.
+
+If the hints are not compatible, the Topology Manager may choose to
+reject the pod. Behavior in this case depends on a new Kubelet configuration
+value to choose the topology policy. The Topology Manager supports two
+modes: `strict` and `preferred` (default). In `strict` mode, the pod is
```
cbf123: Rather than a Kubelet configuration value, I think it would make more sense to have the pod determine whether it wants to either have "strict" NUMA-alignment or fail to schedule, or whether it prefers to prioritize being able to run at all over ensuring optimal performance.

After all, it's the workload that determines how performance-sensitive it is.

PR #1683, opened 2018-01-25
[WIP] local volume attach/dettach refactor

mlmhl:
This PR describes the refactor of local volumes' attach/detach mechanism.
ref: kubernetes/kubernetes#54108

### DIFF
--- a/contributors/design-proposals/storage/local-volume-management.md
+++ b/contributors/design-proposals/storage/local-volume-management.md
@@ -1,0 +1,219 @@
+# Local volume management
+
+**Authors:** [mlmhl](https://github.com/mlmhl)
+
+- [Background](#background)
+- [Object](#object)
+- [Roadmap](#roadmap)
+- [Implementation Design](#design)
+	- [Kubernetes Side Refactor](#k8s_refactor)
+		- [Volum Plugin Interfaces Refactor](#k8s_interfaces_refactor)
+		- [Kubelet Volume Manager Refactor](#k8s_vm_refactor)
+			- [Mount Device Operation](#k8s_vm_mount)
+			- [Umount Device Operation](#k8s_vm_umount)
+		- [A/D Controller Refactor](#k8s_ad_controller_refactor)
+	- [Volume Plugin Side Refactor](#volume_plugin_refactor)
+		- [Mount Device Operation](#volume_plugin_mount_refactor)
+		- [Umount Device Operation](#volume_plugin_umount_refactor)
+
+## <a name="background"></a>Background
+
+For all attachable volume plugins, there are three major steps when provisioning a volume for a `Pod`:
+
+1. Attach the volume to according node the pod is scheduled to
+2. Mount the attached device to a global mount path
+3. Bind the global mount path to a pod specified path
+
+The recycle of a volume is similar, umount/detach the path/device in a reverse order.
+
+Today, A/D controller is responsible for the attach/detach operation of volumes, kubelet is responsible for the other two steps. Before we start, let's make a quick review of the whole workflow:
+
+1. A/D controller watches all scheduled `Pods` from apiserver, and add all volumes in `Pod.Spec.Volumes` to its `DesiredStateOfWorld`. This cache maintains volumes need to be attached
+2. A/D controller periodically checks volumes in `DesiredStateOfWorld` to make sure if they are attached. If not, controller will invoke the attach operation of this volume. After attach operation finished, controller will add this volume to `Node.Status.VolumesAttached`
+3. Kubelet watches all pod scheduled to its own, and add all volumes in `Pod.Spec.Volumes` to its `DesiredStateOfWorld`
+4. Kubelet periodically checks volumes in `DesiredStateOfWorld` to make sure if they are mounted. If not, kubelet will:
+	- Wait for the attach operation finished
+	- Mount the attached device to the global mount path
+	- Bind the global mount path to a pod specified path
+
+This workflow if nicely for all cloud volumes(`AWS EBS`, `GCE PD`, etc.) which have a remote attach API, but not very suitable for local volumes(`fc`, `iscsi`, `rbd`, etc.) which need a local attach operation. So current implementation make the `Attach/Detach` method of these local volumes do nothing, and execute the real attach operations in `WaitForAttach` method. This led to the following problems:
+
+1. Device path missed in `Node.Status.VolumesAttached`.  A/D controller gets volumes' device path from the returned parameter of `Attach` method, and  updates according Node API object. Local volumes' `Attach` method does nothing, so the returned device path is empty. These absent fields may make user confused.
+
+2. Unexpected umount behavior in kubelet. Kubelet will umount the device from node if no pods use a volume, the `UnmountDeviceFunc` will use the device path stored in `Node.Status.VolumesAttached` to check if this device is still opened or not. As this device path is empty for local volumes, there may be unexpected behaviors.
+
+3. Complex and confusing code both in kubernetes side(`kubelet Volumemanager` and `A/D controller`) and volume plugin side, as local volumes have to implement `AttachableVolumePlugin` interface.
+
+## <a name="object"></a>Object
+
+To solve the problems as above, we need to split the `AttachableVolumePlugin` interface into two independent interfaces: One is still called `AttachableVolumePlugin` and another is `DeviceMountableVolumePlugin` . The new `AttachableVolumePlugin` has the same interface with the old one and implemented by the cloud volumes only, `DeviceMountableVolumePlugin` has the device mount/umount interface only and implemented by the local volumes. Accordingly, we also need to split `Attacher` intreface to `Attacher` and `DeviceMounter`, `Detacher` to `Detacher` and `DeviceUmounter`.
+
+## <a name="roadmap"></a>Roadmap
+
+* Volume relative interface refactor
+	* Refactor volume plugin interfaces(`AttachableVolumePlugin`) and volume interfaces(`Attacher/Detacher`)
+	* Refactor `A/D controller` and `Kubelet VolumeManager` to use new interfaces
+* Refactor local volumes to implement `DevicemountableVolumePlugin` only
+	* Refactor `rbd`
+	* Refactor `fc`, `iscsi`, etc.
+
+## <a name="design"></a>Implementation Design
+
+### <a name="k8s_refactor"></a>Kubernetes Side Refactor
+
+#### <a name="k8s_interfaces_refactor"></a>Volume Plugin Interfaces refactor
+
+We extract the device mount relative methods from `Attacher` to a new interface, and embed this new interface back to `Attacher` so that volumes implement `Attacher` need also implement `DeviceMounter`:
+
+```go
+// Attacher can attach a volume to a node.
+type Attacher interface {
+	DeviceMounter
+
+	// Attaches the volume specified by the given spec to the node with the given Name.
+	// On success, returns the device path where the device was attached on the
+	// node.
+	Attach(spec *Spec, nodeName types.NodeName) (string, error)
+
+	// VolumesAreAttached checks whether the list of volumes still attached to the specified
+	// node. It returns a map which maps from the volume spec to the checking result.
+	// If an error is occurred during checking, the error will be returned
+	VolumesAreAttached(specs []*Spec, nodeName types.NodeName) (map[*Spec]bool, error)
+
+	// WaitForAttach blocks until the device is attached to this
+	// node. If it successfully attaches, the path to the device
+	// is returned. Otherwise, if the device does not attach after
+	// the given timeout period, an error will be returned.
+	WaitForAttach(spec *Spec, devicePath string, pod *v1.Pod, timeout time.Duration) (string, error)
+}
+
+// DeviceMounter can mount a volume to node.
+type DeviceMounter interface {
+	// GetDeviceMountPath returns a path where the device should
+	// be mounted after it is attached. This is a global mount
+	// point which should be bind mounted for individual volumes.
+	GetDeviceMountPath(spec *Spec) (string, error)
+
+	// MountDevice mounts the disk to a global path which individual pods
+	// can then bind mount. Returns the global mount path if succeeded,
+	// otherwise returns an error.
+	MountDevice(spec *Spec, devicePath string, pod *v1.Pod) (string, error)
+}
+```
+
+A `*v1.Pod` pararmeter is added for `MountDevice` method because we need to use pod's namespace to fetch according secret if the volume is a pod inline volume, or the secret namespace is not set.
+
+Similarly we apply the refactor to `Detacher`:
+
+```go
+// Detacher can detach a volume from a node.
+type Detacher interface {
+	DeviceUmounter
+	// Detach the given volume from the node with the given Name.
+	// volumeName is name of the volume as returned from plugin's
+	// GetVolumeName().
+	Detach(volumeName string, nodeName types.NodeName) error
+}
+
+// Deviceumounter can umount a volume from a node.
+type DeviceUmounter interface {
+	// UnmountDevice unmounts the global mount of the disk. This
+	// should only be called once all bind mounts have been
+	// unmounted.
+	UnmountDevice(deviceMountPath string) error
+}
+```
+
+Accordingly, extract a `DeviceMountableVolumePlugin` interface from `AttachableVolumePlugin`:
+
+```go
+// AttachableVolumePlugin is an extended interface of VolumePlugin and is used for volumes that require attachment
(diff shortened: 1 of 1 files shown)


mtanino:

jingxu97:
@mlmhl thanks for the proposal. I also think we need refactor work around this area. But I am not a big fan of adding even more interfaces. The problem you mentioned devicePath is retrieved from node.Status can be solved by updating the the cached data in actual state. During unmount operation, the device path is retrieved from the cached data, not directly from node.Status.
My thought about refactor is just having two interface, one is for remote volume setup, one is for local volume setup. The operations for remote volume setup is attach/detach, the operations for local volume setup include waitforattach, mount etc. We are also discussing whether separate mountdevice and unmountdevice is necessary.

mlmhl:
@jingxu97 
IIUC, the difference between refactor you mentioned and this proposal is shall we maintain a separate interface for device mount/umount operation. This proposal keeps three interfaces: `VolumePlugin`(for mount/umount operation), `AttachableVolumePolugin`(for attach/detach/waitforattach operation) and `DeviceMountableVolumePlugin`(for device mount/umount), the refactor you mentioned merge device mount/umount operation to mount/umount operation, so we only need to keep two interface: `VolumePlugin`(for device mount/umount and mount/umount operations) and `AttachabelVolumePlugin`(for attach/detach/waitforattach operation).

I think keep only two interfaces is also work, and make the kubernetes side code(volume manager) more simple and clearer. So I'm not sure which one we should choose. Please let me know when we make the final decision and I will update this proposal for it.

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+2. A/D controller periodically checks volumes in `DesiredStateOfWorld` to make sure if they are attached. If not, controller will invoke the attach operation of this volume. After attach operation finished, controller will add this volume to `Node.Status.VolumesAttached`
+3. Kubelet watches all pod scheduled to its own, and add all volumes in `Pod.Spec.Volumes` to its `DesiredStateOfWorld`
+4. Kubelet periodically checks volumes in `DesiredStateOfWorld` to make sure if they are mounted. If not, kubelet will:
+	- Wait for the attach operation finished
+	- Mount the attached device to the global mount path
+	- Bind the global mount path to a pod specified path
+
+This workflow if nicely for all cloud volumes(`AWS EBS`, `GCE PD`, etc.) which have a remote attach API, but not very suitable for local volumes(`fc`, `iscsi`, `rbd`, etc.) which need a local attach operation. So current implementation make the `Attach/Detach` method of these local volumes do nothing, and execute the real attach operations in `WaitForAttach` method. This led to the following problems:
```
NickrenREN: If i understand your proposal correctly, you just extract the `MountDevice` out of `Attacher` and then put it into `DeviceMounter`.  Can we really solve this problem ?  We still need to execute `attach` for rbd in `WaitForAttach`.

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+
+A/D controller needn't to change any implementation details as it only handle `AttachableVolumePlugin` and we actually don't modify this interface and according volume plugins. The only thing we need to do is remove local volume plugins from the returned list of `ProbeAttachableVolumePlugins` method as they aren't `AttachableVolumePlugin` any more.
+
+### <a name="volume_plugin_refactor"></a>Volume Plugin Side Refactor
+
+Currently, some local volumes('rbd', `fc`, `iscsi`) also implement `AttachableVolumePlugin`. As these volumes don't have a remote `Attach` API, the local attach operation is executed inside `WaitForAttach` method, the umount and detach operation both performed inside `UnmountDevice` method, this implement leads to the problems mentioned in the beginning of this article. So we should make these volume plugins implementing `DeviceMountableVolumePlugin` only.
+
+#### <a name="volume_plugin_mount_refactor"></a>Mount Device Operation
+
+- Move the local attach operation from `Attacher.WaitForAttach` method to `DeviceMounter.UnmountDevice` method. This is a simple and relative small modification.
```
NickrenREN: s/UnmountDevice/MountDevice

contributors/design-proposals/storage/local-volume-management.md
NickrenREN: oh, you mentioned at the end of this proposal, move `attach` to `MountDevice` ?

saad-ali:
Thank you for opening this @mlmhl!! I like this proposal, it is what I was envisioning in https://github.com/kubernetes/kubernetes/issues/54108#issuecomment-358425835

However, @mtanino pointed out in that thread (https://github.com/kubernetes/kubernetes/issues/54108#issuecomment-358530942) that he has a requirement that local attach (`MountDevice(...)` in this proposal) should not be triggered by a kubelet, until `UnmountDevice(...)` has completed from previous node for a single-read-write volume plugin.

If we agree that this should be a requirement this design will get more complicated because it will require coordination between master/kubelet.

@mtanino thoughts?

msau42:
For plugins like iSCSI that do local attach, but may still want kubernetes to serialize attach across multiple nodes, then they may still have to implement the Attach interface which does nothing, and MountDevice which does the local attach.

For local volume plugin, MountDevice is needed, but Attach is not, so this split be useful.  And it will align with CSI capabilities better too.

Can we proceed with the current design?

msau42:
@mkimuram can you evaluate if you need local attach to be coordinated across nodes for iscsi and fc plugins?  We would like to proceed forward with the current design that will still benefit local and CSI volume plugins.

msau42:
cc @bswartz
who is doing some work on iscsi locking.  With iscsi locking, it may be possible for the iscsi plugin to not have to implement Attach() and could just use a local attacher interface.

mkimuram:
@msau42

> @mkimuram can you evaluate if you need local attach to be coordinated across nodes for iscsi and fc plugins? We would like to proceed forward with the current design that will still benefit local and CSI volume plugins. 

In short answer, I think that we can go ahead with this proposal for existing iSCSI and FC attachers and consider avoiding from multi-attaching in CSI version of iSCSI and FC attachers. The reason that I think so is as below. Correct me if I'm wrong.

IIUC, there are below two points that attachers handle backend storage to avoid multi-attaching volumes at the same time:

[Points to handle for avoiding multi-attaching issue]
1. Allowing volumes to be accessed from a node where a pod that uses the volumes is/will be running,
2. Making the volumes visible to the node

In existing FC driver and iSCSI plugin, there is no place to do 1, as a result, all volumes are visible from all the nodes, so an issue like "(a) a volume can be attached to multiple nodes at the same time because the volume is attached to one node before it is detached from another node" happens.
(Existing attachers are in tree, so there is no chance to do 1, so it has been done in external provisioner when no one knows where the volume will be consumed.)

However, if iSCSI and FC are moved to CSI, it will have a chance to do 1 on its own controller's or node's publish/unpublish volume. Therefore it will be much easier to handle this issue in CSI. 

> who is doing some work on iscsi locking. With iscsi locking, it may be possible for the iscsi plugin to not have to implement Attach() and could just use a local attacher interface.

As you mentioned, a similar issue like "(b) a volume that is detaching from a node is wrongly attached to the same node while attaching another volume on the same node" should be handled properly in 2. (I think that it is under development in https://github.com/kubernetes/kubernetes/issues/59946 )

However, I think we can separate it from the design of this proposal. (If this kind of locking also solves the issue (a) above, it is good for exisiting FC attacher and iSCSI attacher. However, I guess that this only focuses on issues on one node, not issues on multiple nodes.)

mkimuram:
@msau42

> Thanks @mkimuram, can you clarify why it will be easier to address this issue in CSI?

In short, iSCSI plugin and FC plugin in CSI will be more similar to cloud storage than local storage, in the point that controller's ```Attach``` call makes sense.

As I mentioned in my previous comment, existing iSCSI attacher doesn't restrict a certain volume to be accessible from a certain node, therefore it has a risk to be attached to multiple nodes and accessed from multiple nodes.([AttachDisk](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/iscsi/iscsi_util.go#L216) which is called from ```WaitForAttach``` just  scans volumes in the node.)

This is due to the restriction that there is no point to inject backend storage dependent codes in attacher.(Restricting access is backend storage dependent logic. And before CSI, external provisioner was the only place where backend storage dependent codes can be injected. However, it is not decided in which node the provisioned volume will be consumed on provision,  there's no other way for iSCSI plugin to make provisioned volumes accessible from all the possible [nodes](https://github.com/kubernetes-incubator/external-storage/blob/master/iscsi/targetd/provisioner/iscsi-provisioner.go#L127). So, volumes for existing iSCSI plugin has no guard for multiple node access.)

On the other hand, CSI attacher has a chance to restrict a certain volume to be accessible from a certain node, in [ControllerPublishVolume call]( https://github.com/kubernetes-csi/external-attacher/blob/master/pkg/connection/connection.go#L207) from external attacher, as cloud storages do. So, if the volume is restricted to be accessed from one node at the same time there, attempts to attach the volume to another inappropriate nodes from kubelet will fail properly. 
(Assuming that ```NodeStageVolume``` and ```NodePublishVolume``` will be called from kubelet, but they will fail if the requested volume is not for the node. )
As a result, it will be guaranteed that a certain volume will be accessed from only one node at the same time, at least for this issue's scenario.

Of course, to make it work well as described above, each CSI plugin for each iSCSI backend storage needs its own implementation of ControllerPublishVolume. For example in iscsi/targetd case, it needs to create proper acl mappings for LUNs and iSCSI initiators.

Also, I just mentioned about iSCSI above, FC is almost the same situation.

jsafrane:
> However, @mtanino pointed out in that thread (kubernetes/kubernetes#54108 (comment)) that he has a requirement that local attach (MountDevice(...) in this proposal) should not be triggered by a kubelet, until UnmountDevice(...) has completed from previous node for a single-read-write volume plugin.

This is IMO already solved by iSCSI implementing empty attach/detach interface. A/D controller won't allow attaching a volume to a node if it's still attached somewhere else. I personally don't like it (it's still prossible to run two pods with single RWO volume on the same node and these pods may overwrite data of each other), but for this particular purpose it serves well.

> Clearly there is some code that could be reused between iSCSI and other SAN-based volumes that just start from a block device and put a filesystem on it. That code doesn't need to be part of the iSCSI plugin -- it could be shared in a library.

There is `SafeFormatAndMount` used by all volume plugins that gets a device, formats it if it's necessary  and mounts it.

> The challenges that exist with iSCSI have more to do with the shared transport underneath where multiple volumes may rely on a single transport layer,

That's problem that only iSCSI plugin must solve and it is IMO unrelated to this PR.

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+3. Kubelet watches all pod scheduled to its own, and add all volumes in `Pod.Spec.Volumes` to its `DesiredStateOfWorld`
+4. Kubelet periodically checks volumes in `DesiredStateOfWorld` to make sure if they are mounted. If not, kubelet will:
+	- Wait for the attach operation finished
+	- Mount the attached device to the global mount path
+	- Bind the global mount path to a pod specified path
+
+This workflow if nicely for all cloud volumes(`AWS EBS`, `GCE PD`, etc.) which have a remote attach API, but not very suitable for local volumes(`fc`, `iscsi`, `rbd`, etc.) which need a local attach operation. So current implementation make the `Attach/Detach` method of these local volumes do nothing, and execute the real attach operations in `WaitForAttach` method. This led to the following problems:
+
+1. Device path missed in `Node.Status.VolumesAttached`.  A/D controller gets volumes' device path from the returned parameter of `Attach` method, and  updates according Node API object. Local volumes' `Attach` method does nothing, so the returned device path is empty. These absent fields may make user confused.
```
jsafrane: I don't see how this proposal changes this. We still need "empty" Attach()/Detach() for iSCSI and FC to make sure we don't attach the volume to multiple nodes.

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+
+2. Unexpected umount behavior in kubelet. Kubelet will umount the device from node if no pods use a volume, the `UnmountDeviceFunc` will use the device path stored in `Node.Status.VolumesAttached` to check if this device is still opened or not. As this device path is empty for local volumes, there may be unexpected behaviors.
+
+3. Complex and confusing code both in kubernetes side(`kubelet Volumemanager` and `A/D controller`) and volume plugin side, as local volumes have to implement `AttachableVolumePlugin` interface.
```
jsafrane: This PR does not propose any solution to this complicated code, it will remain roughly the same.

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+	* Refactor `rbd`
+	* Refactor `fc`, `iscsi`, etc.
+
+## <a name="design"></a>Implementation Design
+
+### <a name="k8s_refactor"></a>Kubernetes Side Refactor
+
+#### <a name="k8s_interfaces_refactor"></a>Volume Plugin Interfaces refactor
+
+We extract the device mount relative methods from `Attacher` to a new interface, and embed this new interface back to `Attacher` so that volumes implement `Attacher` need also implement `DeviceMounter`:
```
jsafrane: "volumes implement `Attacher` need also implement `DeviceMounter`" - that's a good thing to require, however I don't think that embedding `DeviceMounter` into `Attacher` is the right thing to do. I'd leave the interface independent and check it in runtime (i.e. in `GenerateMountVolumeFunc`).

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+	VolumesAreAttached(specs []*Spec, nodeName types.NodeName) (map[*Spec]bool, error)
+
+	// WaitForAttach blocks until the device is attached to this
+	// node. If it successfully attaches, the path to the device
+	// is returned. Otherwise, if the device does not attach after
+	// the given timeout period, an error will be returned.
+	WaitForAttach(spec *Spec, devicePath string, pod *v1.Pod, timeout time.Duration) (string, error)
+}
+
+// DeviceMounter can mount a volume to node.
```
jsafrane: I'd emphasize that it's for volumes backed by a block device. It's useless for Gluster or NFS.

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+	GetDeviceMountPath(spec *Spec) (string, error)
+
+	// MountDevice mounts the disk to a global path which individual pods
+	// can then bind mount. Returns the global mount path if succeeded,
+	// otherwise returns an error.
+	MountDevice(spec *Spec, devicePath string, pod *v1.Pod) (string, error)
+}
+```
+
+A `*v1.Pod` pararmeter is added for `MountDevice` method because we need to use pod's namespace to fetch according secret if the volume is a pod inline volume, or the secret namespace is not set.
```
jsafrane: Why it needs the Secrets at this point? If it's for Ceph RBD, the volume must be attached in `WaitForAttach()`, because after that it could be used as a raw block volume and MountDevice won't be called for it.

Perhaps we could rename `WaitForAttach()` to `CompleteNodeAttach()`, that would do local attach (if necessary) and find the block device (for both local and remote attach).

contributors/design-proposals/storage/local-volume-management.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### <a name="k8s_ad_controller_refactor"></a>A/D Controller refactor
+
+A/D controller needn't to change any implementation details as it only handle `AttachableVolumePlugin` and we actually don't modify this interface and according volume plugins. The only thing we need to do is remove local volume plugins from the returned list of `ProbeAttachableVolumePlugins` method as they aren't `AttachableVolumePlugin` any more.
+
+### <a name="volume_plugin_refactor"></a>Volume Plugin Side Refactor
+
+Currently, some local volumes('rbd', `fc`, `iscsi`) also implement `AttachableVolumePlugin`. As these volumes don't have a remote `Attach` API, the local attach operation is executed inside `WaitForAttach` method, the umount and detach operation both performed inside `UnmountDevice` method, this implement leads to the problems mentioned in the beginning of this article. So we should make these volume plugins implementing `DeviceMountableVolumePlugin` only.
```
jsafrane: No, empty `Attach()` and `Detach()` make sure that the volume is not attached to multiple nodes. iSCSI, FC and RBD should still use it. IMO, only LocalStorage can benefit from separate `DeviceMountableVolumePlugin`.

jsafrane:
Friendly ping?

msau42:
This has been replaced by https://github.com/kubernetes/community/pull/2438

PR #1717, opened 2018-02-01
Proposal: optionally disable recursive chown with fsGroup

tsmetana:
Proposal for a new `PodSecurityObject` attribute that would prevent recursive chown of the volumes when the `fsGroup` is set and prevent errors mounting volumes with large number of files.

Feature issue: https://github.com/kubernetes/features/issues/549

### DIFF
--- a/contributors/design-proposals/storage/fsgroup-optional-recursion.md
+++ b/contributors/design-proposals/storage/fsgroup-optional-recursion.md
@@ -1,0 +1,59 @@
+## PodSecurityContext and fsGroup
+
+To allow for pod-level security settings the API defines `PodSecurityObject` that lets the admin set-up
+several security attributes applied to all containers in the pod. One of the attributes is `fsGroup`
+integer allowing Kubelet to change ownership of the files on the containers' volumes for some volume
+types. When `fsGroup` is set, Kubelet would recursively traverse the volume mount and change GID ownership
+of all the files on it.
+
+### Known issues with current design
+
+The current desing may cause mount time-outs on volumes containing many files. Typically a restart of a
+pod that produces large number of files becomes impossible if `fsGroup` is set for the pod:
+
+```
+FirstSeen     LastSeen        Count   From  SubobjectPath     Type            Reason          Message
+---------     --------        -----   ----  -------------     --------        ------          -------
+2m            2m              1       {default-scheduler }    Normal          Scheduled       Successfully assigned <NODE_NAME>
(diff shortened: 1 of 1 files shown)


php-coder:
Could it be implemented as an annotation? Something like a hint for a Kubelet?

BTW, we can even solve this without introducing a new field, if we allow to specify negative `fsGroup` that will be treated in a special way. Yes, it looks strange but it even won't break backward compatibility as we didn't allow negative values before :)

gnufied:
Do we want to use this opportunity to allow users to disable recursive relabelling as well?

tsmetana:
@gnufied good idea. I should have thought of it too...
@php-coder annotation might be a fallback solution if the API change is rejected. I would rather have the flag in the PodSecurityContext since it feels more natural and hence more user-friendly. Negative number would work too but... I didn't expect having to imagine something as ugly. :)

tsmetana:
After examining how SELinux relabelling really works in the Kubelet, I'm not quite sure... It's different from fsGroup: the GID ownership is done by the Kubelet itself on mount, seLinuxOptions are only being parsed and passed to container runtime to cope with them. Meaning Kubelet has no control over the files re-labelling.

gnufied:
@tsmetana kubelet does have control over them. There are 2 container runtime options available ":Z" and ":z". The former does recursive relabelling while later does just the top level directory relabelling. So, if we were to expose this as a SELinux option in PSP or somewhere else, kubelet will honour it.

tsmetana:
Thanks @gnufied. I didn't know about the two options... I thought it was "all or nothing".

tsmetana:
So, no, we don't have control over the SELinux labels. The -z and -Z difference is in the label they apply not the recursion -- they both recursively relabel the volume, only -z always applies "system_u:object_r:container_file_t:s0" whereas -Z uses the context of the calling process (container runtime).

tsmetana:
I would like to revive the discussion about this. @saad-ali I know you consider the proposal to be insufficient... I was wondering: would the annotation workaround proposed by @php-coder be better? At least it would not require an API change... The current situation is very unfortunate, e.g. Jenkins which is a very popular CI automation tool tends to create lots of many small files, so its users might be affected by the problem quite frequently.

php-coder:
@tsmetana Why it has been closed?

mmclanerh:
@tsmetana was there a follow on to this?  this seems to be a bit of a problem today, noticed with some frequency in the Jenkins CI and CHE IDE application cases (as you had predicted).

tsmetana:
@mmclanerh I'm afraid this discussion died. I've seen other issues reporting basically the same problem (need to search github...) and in IIRC the resolution was "don't do it, find a workaround".

Also there is still the problem with SELinux labelling I have no idea how to deal with -- so my proposal would is only a partial fix to the issue.

PR #1723, opened 2018-02-02
Ihor's email updated

idvoretskyi:
@spiffxp


idvoretskyi:
It's a minor change :)

PR #1727, opened 2018-02-02
Update metadata-policy.md

bgrant0607:
Noting that this proposal was not implemented. Not marked obsolete because we might still want a flavor of it.


idvoretskyi:
PTAL @kubernetes/sig-api-machinery-pr-reviews

PR #1731, opened 2018-02-02
Flex deploy examples, and miscellaneous documentation updates.

verult:
* Updating documentation to reflect the newly updated Flexvolume example directory. Depends on https://github.com/kubernetes/kubernetes/pull/59149.
* Removing mentions of --enable-controller-attach-detach because setting it to false is not recommended.
* Adding documentation about pod information passed as mounter options.

/sig storage
/assign @chakri-nelluri @saad-ali

### DIFF
--- a/contributors/devel/flexvolume.md
+++ b/contributors/devel/flexvolume.md
@@ -1,6 +1,6 @@
 # Flexvolume
 
-Flexvolume enables users to write their own drivers and add support for their volumes in Kubernetes. Vendor drivers should be installed in the volume plugin path on every Kubelet node and on master node(s) if `--enable-controller-attach-detach` Kubelet option is enabled.
+Flexvolume enables users to write their own drivers and add support for their volumes in Kubernetes. Vendor drivers should be installed in the volume plugin path on every node, and on master if the driver requires attach capability (unless `--enable-controller-attach-detach` Kubelet option is set to false, but this is highly discouraged because it is a legacy mode of operation).
 
 Flexvolume is a GA feature from Kubernetes 1.8 release onwards.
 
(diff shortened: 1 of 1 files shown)


verult:
gentle ping

chakri-nelluri:
@verult we have two more options missing in the doc. 
```
"kubernetes.io/mountsDir"
"kubernetes.io/pvOrVolumeName"
```
Other than that everything looks good to me.

verult:
ACK, although `"kubernetes.io/mountsDir"` isn't used anywhere in code (besides being declared)?

chakri-nelluri:
My bad in copy & paste
```	"kubernetes.io/fsGroup" ```
is the other option missing.

**Comment by verult** [comment]
Addressed comments

File: contributors/devel/flexvolume.md [hunk 1]
```diff
@@ -1,6 +1,6 @@
 # Flexvolume
 
-Flexvolume enables users to write their own drivers and add support for their volumes in Kubernetes. Vendor drivers should be installed in the volume plugin path on every Kubelet node and on master node(s) if `--enable-controller-attach-detach` Kubelet option is enabled.
+Flexvolume enables users to write their own drivers and add support for their volumes in Kubernetes. Vendor drivers should be installed in the volume plugin path on every node, and on master if the driver requires attach capability.
```
chakri-nelluri: This statement is not completely true. If the --enable-controller-attach-detach is false, kubelet will do the attach.

verult: I think we should strongly discourage users from setting that flag to false; k8s is no longer designed to operate in that mode. Having this flag in the documentation gives users new to k8s the idea that disabling attach detach is a common and viable alternative, but in most cases it's not.

I'm open to wording it differently but I'd avoid including the flag in the documentation

verult:
PTAL

chakri-nelluri:
@verult Did you take a re-stab a the wording? It still sounds the same. I just want to make sure it won't confuse existing users.

verult:
@chakri-nelluri how about that?

verult:
Updated example links to the examples repo.

chakri-nelluri:
@verult Please rebase.

verult:
@chakri-nelluri ACK. @calebamiles PTAL, thanks!

PR #1733, opened 2018-02-04
Add processor cache management proposal

linericyang:
This is my first proposal to community repo, please advise if something need to change, such as process steps, format, etc. Thanks.

### DIFF
--- a/contributors/design-proposals/node/cache-management.md
+++ b/contributors/design-proposals/node/cache-management.md
@@ -1,0 +1,220 @@
+# Processor Cache Management 
+
+_Authors:_
+
+* Malini Bhandaru &lt;malini.k.bhandaru@intel.com&gt;
+* Connor Doyle &lt;connor.p.doyle@intel.com&gt;
+* Dakshina Ilangovan &lt;dakshina.ilangovan@intel.com&gt;
+* Lin Yang &lt;lin.a.yang@intel.com&gt;
+
+**Contents:**
+
+* [Overview](#overview)
+* [Proposed changes](#proposed-changes)
+* [Example Scenarios](#example-scenarios)
+* [Evaluation](#evaluation)
+* [Implementation roadmap](#implementation-roadmap)
+* [Background Details](#background-details)
+
+## Overview
+
+On shared platforms “noisy neighbors” affect application performance. It is a
+serious consideration for latency and jitter sensitive workloads such as
+Network Virtual Functions (NFV), Real Time and Interactive applications.
+Contention for processor cache affects performance in the same vein as
+resources such as CPU and RAM. **We introduce cache allocation to the arsenal
+of core pinning, RAM allocation, and job priority to reduce interference and
+ensure more predictable performance.** See [whitepaper](https://builders.intel.com/docs/networkbuilders/deterministic_network_functions_virtualization_with_Intel_Resource_Director_Technology.pdf) on the benefits that
+cache allocation provides NFV workloads.
+
+Cache is particularly important to hide memory access latency given the
+disparity in speed of the CPU core (typically measured in GigaHertz) and memory
+access that could be 100s of nanoseconds when there is a last level cache miss.
+Processor cache is typically allocated to an application on a first come, first
+served basis, with cache lines being evicted based on various [cache
+replacement policies](https://en.wikipedia.org/wiki/Cache_replacement_policies) such as Least Recently Used and Least Frequently Used
+among others. Different applications have different [working set](https://en.wikipedia.org/wiki/Working_set) sizes, and some known as 
+streaming applications, constantly pollute the cache with new lines. Consider
+for example a video player or a cryptography application. Without any dedicated
+allocation of cache, streaming applications or those with large working sets,
+for example App[0] running on Core 0 in the figure below, the majority of the
+cache lines may be unavailable to other workloads such as App[1] running on
+Core 1. Further, new workloads may come and go, further contributing to
+performance variability.
+
+![noisy neighbor](https://user-images.githubusercontent.com/3691428/35775410-e2907a88-093c-11e8-8c7f-6cec5f7f3b6f.png)
+
+A “noisy neighbor” on core zero over-utilizes the shared platform cache
+resources causing performance inversion (though the priority app on core one is
+higher priority, it runs slower than expected).
+
+In this initial enablement, given few users have a sense of cache significance
+(other than deployers of  Network Virtual Functions), we are not seeking to
+establish cache as a first class resource, that is a resource that is
+explicitly requested at container launch time. But seeking to control cache
+allocation to a container based on it QoS.
+
+## Proposed Changes
+
+### Extending CPU Manager
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/node/cache-management.md
```diff
@@ -0,0 +1,219 @@
...
+  launched containers except for a streaming application such as a video player
+  running first as  low priority and next as high priority. In the former
+  scenarios, cache disruption is limited to the bottom half while in the latter
+  it has full use of the cache and thus affects adversely more co-resident
+  container workloads. 
+
+## Implementation Roadmap
+
+### Phase 1
+V 1.10 or 1.11 to include the above experimental cache allocation.
```
ConnorDoyle: v1.11 at the earliest

ConnorDoyle:
/sig node
/kind feature

contributors/design-proposals/node/cache-management.md
LinEricYang: Thanks for pointing out this. Done.

jeremyeder:
The design does not seem to mention integration with the existing system-reserved scaffolding.  Had you considered a basic use-case, where we segment the cache into partitions (perhaps proportional in size) based on the use of system-reserved?

LinEricYang:
@jeremyeder Thanks for your comments. Actually segment the cache in size is the very first thought when we design the cache allocation strategy. However, cache allocation requires contiguous cache ways, it introduces fragmentation.  Current we don't have a perfect solution to resolve it without bringing in additional cache warm-up latency. So for "static" policy, we are seeking a simple solution which will meet most of requirements. And will try to improve it in "dynamic" policy by making it be able to track and understand workload requirements and do the proper action for it. Any thoughts?

derekwaynecarr:
i am going to defer reviewing this too much until 1.11 re-opens.

contributors/design-proposals/node/cache-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+The original [cpu manager design proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/cpu-manager.md) alluded to extending its scope
+to control processor cache allocation to obtain more predictable performance. 
+The CPU Manager intercepts all container related actions, in particular add,
+remove, start, and state update. Below we discuss changes for cache allocation. 
+
+### Pod QoS Class
+
+Kubernetes supports a notion of [pod QoS](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/), the higher the value, the higher
+the priority. We treat the Guaranteed QOS pod as deserving of more cache
```
thockin: I'm concerned that QoS is the wrong indicator.  It seems completely likely that batch work has limits specified and it seems equally likely (maybe more) that latency-sensitive work might not specify CPU limit (or might set it higher than request).  This latter is the default operating mode for Borg, and I explicitly advise people to not limit CPU unless they have very careful benchmarks and deep understanding of the implications.  It does not work how one would naively expect.

I think we may want some other indicator for this, e.g. an "archetype" selector or similar.  The nice thing about considering this as a new thing is that we can quota it on its own.  Having a limit on # of cores of latency-sensitive work seems useful.

contributors/design-proposals/node/cache-management.md
```diff
@@ -0,0 +1,220 @@
...
+### Alternative Cache Configuration
+
+Instead of just two (or more) overlapping COS, one might ask why not partition
+the cache and assign each exclusively to a CPU core, along the lines of CPU
+pinning. The underlying cache management library provides such support, but in
+this experimental  release, we opt for overlapping pools because of the greater
+sharing they provide (leading to better best-effort performance) yet providing
+less contention for workloads that have a higher QOS requirement. Further, when
+workloads are pinned to a CPU core, they already have exclusive access to the
+core’s L1 and L2 cache.
```
vishh: How does hyperthreading impact this assumption?

contributors/design-proposals/node/cache-management.md
```diff
@@ -0,0 +1,220 @@
...
+  workload is assigned a priority equal to high and all other workloads are
+  assigned priority low with either “Static” or “Dynamic” policy.
+- Perceptible difference in performance, all things being equal with respect to
+  launched containers except for a streaming application such as a video player
+  running first as  low priority and next as high priority. In the former
+  scenarios, cache disruption is limited to the bottom half while in the latter
+  it has full use of the cache and thus affects adversely more co-resident
+  container workloads. 
+
+## Implementation Roadmap
```
vishh: I'd like to see monitoring of cache allocations and memory bandwidth prior to adding a feature to restrict it. How can a k8s user know if they need to consume cache partitioning?

vishh:
@LinEricYang I find manual cache partitioning between `shared` and `reserved` a bit tedious since a user will then have to carefully plan capacity and scheduling. Would it be possible to dynamically adjust CLOS sizes based on usage metrics and policies (max limits for batch for example)? 
AFAIK, users are yet to run batch workloads at scale alongside their web services (or databases) on k8s and so as you mention this feature will be isolated to specialized workloads. Can you post some metrics on the impact of CAT on your NFV workloads? I'm interested in exploring the impact of this feature on ML workloads. Having an easy to use metrics pipeline would be very helpful to evaluate impact.

LinEricYang:
@vishh The static policy is our first step to enable CAT in k8s. It is trying to benefit the most of user case without any performance regression. And User doesn't have to do careful benchmark before having cache control functionality. Absolutely we will create a dynamical policy as next step to make it smarter, like a feedback algorithm to adjust the CLOS size based on the monitoring metrics.

You can find the whitepaper about how CAT benefit NFV case [here.](https://builders.intel.com/docs/networkbuilders/deterministic_network_functions_virtualization_with_Intel_Resource_Director_Technology.pdf)

contributors/design-proposals/node/cache-management.md
LinEricYang: True. We are working on another proposal to integrate cache monitoring technology with cadvisor, which will allow user to monitor the cache usage of the specific container.

contributors/design-proposals/node/cache-management.md
LinEricYang: Thanks for pointing out this, @vishh. Absolutely two logical thread will share the L1 and L2 cache in one cpu core. Will re-work this part.

contributors/design-proposals/node/cache-management.md
dakshinai: A small clarification here, though cache allocation happens alongside CPU pinning done by CPU manager, we do not exactly rely on how CPU manager assigns cores to a workload. We would map cache ways to container process id's. This is a feature through resctrl fs on kernel version >= 4.10

linericyang:
We did the benchmark to show the performance benefit of this proposal. The result shows it can reduce half of cache latency when the high-priority workload co-locates with the noisy neighbor. More result detail can be found [here](https://docs.google.com/presentation/d/1t2wB9i07CobU41MPT5-mvyl-uS3v8QykMSjGT8rDcIU/)

@ConnorDoyle @LuyaoZhong

PR #1738, opened 2018-02-05
Proposal: Add service port ranges support

m1093782566:
There are several applications like SIP apps or RTP which needs a lot of ports to run multiple calls or media streams. Currently there is no way to allow a range in ports in spec. So essentially users may have to do this:

```
- name: sip-udp5060
  containerPort: 5060
  protocol: UDP
- name: sip-udp5061
  containerPort: 5061
  protocol: UDP
  ...
- name: sip-udp5559
  containerPort: 5559
  protocol: UDP
```

Doing above for 500 ports is not pretty. Can we have a way to allow port ranges like 5060-5559?

## Objective

This proposal builds off of earlier requests to [Support port ranges in services](https://github.com/kubernetes/kubernetes/issues/23864) and aims to document all the requirements of supporting port ranges in services.

/sig network

### DIFF
--- a/contributors/design-proposals/network/service_port_ranges.md
+++ b/contributors/design-proposals/network/service_port_ranges.md
@@ -1,0 +1,130 @@
+# Support port ranges in services
+
+* Status: pending
+* Version: alpha
+* Implementation owner: @m1093782566
+
+## Background & Motivations
+
+There are several applications like SIP apps or RTP which needs a lot of ports to run multiple calls or media streams. Currently there is no way to allow a range in ports in spec. So essentially users may have to do this:
+
+```
+- name: sip-udp5060
+  containerPort: 5060
+  protocol: UDP
+- name: sip-udp5061
+  containerPort: 5061
+  protocol: UDP
+  ...
+- name: sip-udp5559
+  containerPort: 5559
+  protocol: UDP
+```
+
+Doing above for 500 ports is not pretty. Can we have a way to allow port ranges like 5060-5559?
+
+## Objective
+
+This proposal builds off of earlier requests to [Support port ranges in services](https://github.com/kubernetes/kubernetes/issues/23864) and aims to document all the requirements of supporting port ranges in services.
+
+## Goals
+
+A server side solution.
+
+Support multiple port ranges.
+
+## Non-goal
+
+Target port remapping in ranges.
+
+NodePort on ranges.
+
+Pod port ranges.
+
+## User Cases
+
+* Specifying port ranges in services declarations is mandatory for any VoIP-related application.
+* With the standard RTP port range which is 16384-32767. Clients connect on the standard RTP port range with a media server application.
+* Needs this change to make TeleStax (an RTC framework) work well in Kubernetes.
+
+## Proposal
+
+This proposal builds off earlier feature requrest to [Support port ranges in services](https://github.com/kubernetes/kubernetes/issues/23864). We aims to implement service port ranges at least for iptables and IPVS modes.
+
+### Service API changes
+
+Create a new filed `PortRange` in `ServicePort`.
+
+
+```go
+type ServiceSpec struct {
+  Ports []ServicePort
+}
+
+type ServicePort struct {
+  // portRange is the port range that are exposed by this service. Service can expose multiple port ranges and each port range should not be overlapped.
+  // Service port range does not support target port remapping, which means target port ranges are equal to service port ranges and targetPort should not be specified.
+  // Each port range is in the format of `X-Y` and non-negative nuimber X should be <= Y. A port range with single port X can be expressed by `X-X`
+  // port should not be specified when portRange is not empty and port should not be fall in any other port ranges.
+  // portRange should be not specified for NodePort and ExternalIPs Service.
+  // +optional
+  PortRange string `json: "portRange,omitempty"`
+  ...
+}
+```
+
+### Service port ranges example
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: service-port-ranges
+  namespace: default
+spec:
+  ports:
+  - name: http
+    port: 8080
+    protocol: TCP
+    targetPort: 80
+  - name: range1
+    portRange: 1000-2000
+    protocol: UDP
+  - name: range2
+    portRange: 3000-4000
+    protocol: TCP
+  selector:
+    app: foo
+```
+
+### Kube-proxy implementation limitations
+
+NodePort and ExternalIPs Services don't support port ranges since kube-proxy will hold local ports open for these services(nodeport and "external" IP happens to be an IP that is local to this machine). It's challenging to keep some huge port ranges open on host.
+
+Target port remapping is not supported since both iptables and ipvs does not do port translations in the mean time of doing port ranges.
+
+#### iptables mode
+
+iptables `multiport` modules support specifying multiple ports in a single line. Suppose service VIP is `1.2.3.4` and port ranges are `1000-2000` and `3000-4000`, then iptables rules will looks like
+
+```shell
+iptables -A KUBE-SERVICES -d 1.2.3.4/32 -p tcp -m multiport --dports 1000:2000,3000:4000 -j KUBE-SVC-FOO
+```
+
+#### ipvs mode
+
+IPVS + FWMARK can do port ranges. For example,
+
+```shell
+iptables -A PREROUTNG -t mangle -d 172.16.52.57 -p tcp --dport 1000:2000,3000:4000 -j MARK --set-mark 1
+
+ipvsadm -A -f 1 -s rr
+ipvsadm -a -f 1 -r 172.16.52.60 -m
+ipvsadm -a -f 1 -r 172.16.52.61 -m
+```
+
+Create a fwmark bit range [0-31], for stipulating the IPVS fwmark value range. It’s configurable and should not collison with masqBit and DropBit. By default, it's [0-13] (suppose masqBit is 14 and DropBit is 15), which means we can generate at most 8K fwmark values and cover 8K port ranges.
+
+Given VIP 1.2.3.4 and port ranges `1000:2000` and `3000:4000`. Kube-proxy will generate a unique fwmark value in the range of [1-8191], say F1. If the packet needs to be masqueraded, then `F1 = F1 OR 0x4000`. If the packets will be dropped later, then `F1 = F1 OR 0x8000`. Kube-proxy set F1 to the FWMARK field (will be created later) of IPVS virtual server and apply it to kernel.
+
+Netfilter will test if the packet should be masqueraded by bit testing `F1 & 0X4000 == 0X4000` in POSTROUTING chain. Similar thing will happen for checking if the packet should be dropped.
--- a/contributors/design-proposals/network/service_port_ranges.md
+++ b/contributors/design-proposals/network/service_port_ranges.md
@@ -1,8 +1,6 @@
 # Support port ranges in services
 
-* Status: pending
-* Version: alpha
-* Implementation owner: @m1093782566
+Authors: @m1093782566
 
 ## Background & Motivations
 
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,130 @@
...
+metadata:
+  name: service-port-ranges
+  namespace: default
+spec:
+  ports:
+  - name: http
+    port: 8080
+    protocol: TCP
+    targetPort: 80
+  - name: range1
```
bowei: What is the targetPort for a range?

m1093782566: Will be the same as service port range.

bgrant0607:
The proposed change would break API backward compatibility. The port field is a required field in Service and Endpoints.
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L3529
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L3732

This is one of the issues that has blocked progress in the past.

thockin:
You could do 

```
port: 8000
portRangeSize: 1000 # 1000 ports, starting at `port`
```

or

```
port: 8000
lastPort: 8999  # inclusive
```

The default portRangeSize has to be 1 for compat.

We'll need to validate that targetPort is not specified or is == port.

thockin:
Oh yeah, and `userspace` just can't work with ranges (need to EOL that?  some people still use it)

How about winuserspace?

thockin:
Also need to check how external LBs can handle it.

Also need to define the intersection of ranges and NodePorts (nodeport space is pretty limited by default)

There's a lot to do to make this work.

m1093782566:
Sorry for delay, @bgrant0607 @thockin, I was investigating how IPVS can implement port ranges. Proposal is updated now, please take a look when you have a chance.

Regarding API back compatibility, I think the following API pattern suggested by Tim would resolve it.

```
port: 8000
portRangeSize: 1000 # 1000 ports, starting at `port`
```

Since userspace mode in both Linux and Windows platform does not work with port ranges and we can't EOL them for now. Should them be the blocking issues of this feature? It seems like a endless loop if we don't make the decision. It's a fact that different proxy implementations have different functionalities.  Can Kube-proxy running in the modes which doesn't support port range simply log error/warning message when it happens to handle services with port range size >1?

> Also need to check how external LBs can handle it.

At least GCE and AWS LB can handle port ranges, please check:

GCE: https://cloud.google.com/compute/docs/load-balancing/network/forwarding-rules

AWS: https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_Listener.html

I think the most challenging part is how to implement it in IPVS mode.

AFAIK, IPVS + FWMARK can do port ranges. For example,

    iptables -A PREROUTNG -t mangle -d 172.16.52.57 -p tcp --dport 1000:2000,3000:4000 -j MARK --set-mark 1
    
    ipvsadm -A -f 1 -s rr
    ipvsadm -a -f 1 -r 172.16.52.60 -m
    ipvsadm -a -f 1 -r 172.16.52.61 -m

Create a fwmark bit range [0-31], for stipulating the IPVS fwmark value range. It’s configurable and should not collison with masqBit and DropBit. By default, it's [0-13] (suppose masqBit is 14 and DropBit is 15), which means we can generate at most 8K fwmark values and cover 8K port ranges.

Given VIP 1.2.3.4 and port ranges 1000:2000 and 3000:4000. Kube-proxy will generate a unique fwmark value in the range of [1-8191], say F1. If the packet needs to be masqueraded, then F1 = F1 OR 0x4000. If the packets will be dropped later, then F1 = F1 OR 0x8000. Kube-proxy set F1 to the FWMARK field (will be created later) of IPVS virtual server and apply it to kernel.

Netfilter will test if the packet should be masqueraded by bit testing F1 & 0X4000 == 0X4000 in POSTROUTING chain. Similar thing will happen for checking if the packet should be dropped.

In conclusion, this solution will 

* create an iptables rule for each port range

* occupy more than 1 fwmark bits depends on how many port ranges we want to support

I think it's not perfect, but can we do it in a heuristic way first and optimize it later?

contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,128 @@
...
+    protocol: UDP
+  - name: range2
+    port: 200
+    portRange: 300
+    protocol: TCP
+```
+
+### Kube-proxy implementation limitations
+
+NodePort and ExternalIPs Services don't support port ranges since kube-proxy will hold local ports open for these services(nodeport and "external" IP happens to be an IP that is local to this machine). It's challenging to keep some huge port ranges open on host.
```
danwinship: The SIP/RTP case that people talk about a lot requires having some sort of externally-reachable IP address. So, if we have this restriction, is the feature still useful?

(I think we need much more detailed and specific use cases.)

contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,128 @@
...
+
+Target port remapping is not supported since both iptables and ipvs does not do port translations in the mean time of doing port ranges.
+
+Userspace proxy mode in both linux and windows platform does not work with port ranges. Kube-proxy running in these modes will log error message if people create a service with `portRangeSize` > 1.
```
danwinship: As mentioned on https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-382029019, there are probably workable solutions for the linux userspace proxy. I don't know about Windows. (There is also a Windows kernel proxy, which isn't mentioned here, and which I also don't know about.)

contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,128 @@
...
+```shell
+iptables -A NODE-SERVICES -p tcp -m multiport --dports 1000:2000 --dst-type LOCAL -j KUBE-NODEPORTS
+```
+
+#### IPVS mode
+
+IPVS + FWMARK can do port ranges. For example,
+
+```shell
+iptables -A PREROUTNG -t mangle -d 172.16.52.57 -p tcp --dport 1000:2000,3000:4000 -j MARK --set-mark 1
```
danwinship: (typo `PREROUTNG`)

also, you probably mean `--set-mark 1/0x3fff`

contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,128 @@
...
+
+ipvsadm -A -f 1 -s rr
+ipvsadm -a -f 1 -r 172.16.52.60 -m
+ipvsadm -a -f 1 -r 172.16.52.61 -m
+```
+
+Create a fwmark bit range [0-31], for stipulating the IPVS fwmark value range. It’s configurable and should not collison with masqBit and DropBit. By default, it's [0-13] (suppose masqBit is 14 and DropBit is 15), which means we can generate at most 8K fwmark values and cover 8K port ranges.
```
danwinship: Using a large chunk of the fwmark may interfere with other subsystems... And it seems unlikely that you're going to have 8192 distinct service ranges in effect anyway... I'd recommend a much smaller default range.

contributors/design-proposals/network/service_port_ranges.md
cmluciano: Can you link to the mention of this external IP mention? If this is the case, does that mean this is more of a feature request to ingress?

thockin: I think this is a major limitation of the "concentric" model for NodePorts and LoadBalancer services.  I don't see a simple resolution except to allow those platforms that don't need the concentricness to opt out of them.

e.g. Google cloud doesn't need NodePort for LoadBalancer, but AWS does (or did).  I could add an annotation to say "cloud.google.com/bypass-node-port: true".  Or we could allow `nodePort: -1` to mean "skip it for this port".

I think that this needs to be built into the next generation of Service API.

contributors/design-proposals/network/service_port_ranges.md
thockin: @khenidak comment on the Windows side?

contributors/design-proposals/network/service_port_ranges.md
thockin: I'm really not fond of this approach still.  It will require us to allocate in a cluster-consistent way (etcd) or else to checkpoint local state so we can survive kube-proxy restarts (I think?).  And it requires carving off a chunk of the very limited fwmark.

I've asked before but not gotten a great answer - why not just forward the whole IP?  We don't have to deal with fwmark.

Or alternately, why not use ipvs to forward the whole IP and iptables to filter ports we don't want to pass thru (blech, did I just say that?)

contributors/design-proposals/network/service_port_ranges.md
danwinship: > Can you link to the mention of this external IP mention? If this is the case, does that mean this is more of a feature request to ingress?

Sorry, missed this question earlier. Eg, the initial comment on this PR refers to #23864, but the person who filed that bug [said](https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-206109433) "I do need to target the ports through a Service since that is the only way outside users would be able to place a call via SIP". So a solution that doesn't allow reserving the ports on some public IP isn't going to help them.

contributors/design-proposals/network/service_port_ranges.md
khenidak: @PatrickLang -- Can you help answering this?

khenidak:
@m1093782566 also Azure supports this on Standard SKU load balancers.

contributors/design-proposals/network/service_port_ranges.md
PatrickLang: @dineshgovindasamy or @madhanrm - can you help review this? Can you forward a group of ports, or does it makes sense to unroll the ranges into discrete calls from kube-proxy?


cc @jmesser81 @daschott

contributors/design-proposals/network/service_port_ranges.md
m1093782566: @thockin 

> why not use ipvs to forward the whole IP and iptables to filter ports we don't want to pass thru

It should work and can be an alternative if you really can't live with the fwmark approach.

kfox1111:
I've migrated some grid services into k8s and that use case would really benefit from port ranges too.

maoge:
port range proposal using kubernetes annotations
annotation: portrange.alpha.kubernetes.io/port-end-xxxx
https://github.com/kubernetes/kubernetes/commit/2cf5af1f8cc33e2d22ed1968aca05e5323ef9a0b

can this been merged to master branch?

maoge:
port range proposal using kubernetes annotations
annotation: portrange.alpha.kubernetes.io/port-end-xxxx
https://github.com/kubernetes/kubernetes/commit/2cf5af1f8cc33e2d22ed1968aca05e5323ef9a0b

can this been merged ?

thockin:
There are still open issues.  As interested as I am in this feature, I need a person to drive it.

contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,128 @@
+# Support port ranges in services
+
+Authors: @m1093782566
+
+## Background & Motivations
+
+There are several applications like SIP apps or RTP which needs a lot of ports to run multiple calls or media streams. Currently there is no way to allow a range in ports in spec. So essentially users may have to do this:
```
krmayankk: nit: should we be clear here that we are talking about pod spec ?

contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,128 @@
...
+* With the standard RTP port range which is 16384-32767. Clients connect on the standard RTP port range with a media server application.
+* Needs this change to make TeleStax (an RTC framework) work well in Kubernetes.
+
+## Proposal
+
+This proposal builds off earlier feature requrest to [Support port ranges in services](https://github.com/kubernetes/kubernetes/issues/23864). We aims to implement service port ranges at least for iptables and IPVS modes.
+
+### Service API changes
+
+Create a new filed `portRangeSize` in `ServicePort`.
```
krmayankk: nit: field

contributors/design-proposals/network/service_port_ranges.md
```diff
@@ -0,0 +1,128 @@
...
+kind: Service
+metadata:
+  name: service-port-ranges
+  namespace: default
+spec:
+  ports:
+  - name: range1
+    port: 90
+    protocol: TCP
+    portRangeSize: 100
```
krmayankk: nit: do we need port in the naming ?  Should we just say `rangeSize`

m1093782566:
If community folks are interested in this proposal, I will create a KEP. But I am not sure how far we can go forward if you don't accept the iptables + fwmark approach for IPVS proxier.

contributors/design-proposals/network/service_port_ranges.md
m1093782566: Thanks for comment, I will create a new KEP. You can continue to put comments there.

m1093782566:
@thockin

Just for clarification, when you say 'whole-IP forwarding', do you refer to 'use ipvs to forward the whole IP and iptables to filter ports we don't want to pass thru' OR change the feature from 'port range' to 'whole ip forward' as https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-435957275 said?

thockin:
I'm open to exploring both options.

kfox1111:
I'd prefer port ranges to whole ip's, but whole ip's are useful too.

Is this closed in favor of #23864?

thockin:
This was a pre-KEP "proposal".  If we want to pursue this, a KEP is appropriate

PR #1743, opened 2018-02-05
Add CRI testing policy

yujuhong:
@dchen1107 @derekwaynecarr

### DIFF
--- a/contributors/devel/cri-testing-policy.md
+++ b/contributors/devel/cri-testing-policy.md
@@ -1,0 +1,108 @@
+# Container Runtime Interface: Testing Policy
+
+**Owner: SIG-Node**
+
+This document describes testing policy and process for runtimes implementing the
+[Container Runtime Interface (CRI)](/contributors/devel/container-runtime-interface.md)
+to publish test results in a federated dashboard. The objective is to provide
+the Kubernetes community an easy way to track the conformance, stability, and
+supported features of a CRI runtime.
+
+This document focuses on Kubernetes node/cluster end-to-end (E2E) testing
+because many features require integration of runtime, OS, or even the cloud
+provider. A higher-level integration tests provider better signals on vertical
+stack compatibility to the Kubernetes community. On the other hand, runtime
+developers are strongly encouraged to run low-level
+[CRI validation test suite](https://github.com/kubernetes-incubator/cri-tools/blob/master/docs/validation.md)
+for validation as part of their development process.
+
+## Required and optional tests
+
+Runtime maintainers are **required** to submit the tests listed below.
+ 1. Node conformance test suite
+ 2. Node feature test suite
+
+Node E2E tests qualify an OS image with a pre-installed CRI runtime. The
+runtime maintainers are free to choose any OS distribution, packaging, and
+deployment mechanism. Please see the
+[tutorial](https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-node-tests.md)
+to know more about the Node E2E test framework and tests for validating a
+compatible OS image.
+
+The conformance suite is a set of platform-agnostic (e.g., OS, runtime, and
+cloud provider) tests that validate the conformance of the OS image. The feature
+suite allows the runtime to demonstrate what features are supported with the OS
+distribution.
+
+In addition to the required tests, the runtime maintainers are *strongly
+recommended to run and submit results from the Kubernetes conformance test
+suite*. This cluster-level E2E test suite provides extra test signal for areas
+such as Networking, which cannot be covered by CRI, or Node-level
+tests. Because networking requires deep integration between the runtime, the
+cloud provider, and/or other cluster components, runtime maintainers are
+recommended to reach out to other relevant SIGs (e.g., SIG-GCP or SIG-AWS) for
+guidance and/or sponsorship.
+
+## Process for publishing test results
+
+To publish tests results, please submit a proposal in the
+[Kubernetes community repository](https://github.com/kubernetes/community)
+briefly explaining your runtime, providing at least two maintainers, and
+assigning the proposal to the leads of SIG-Node.
+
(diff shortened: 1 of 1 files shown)


contributors/devel/cri-testing-policy.md
```diff
@@ -0,0 +1,108 @@
+# Container Runtime Interface: Testing Policy
+
+**Owner: SIG-Node**
+
+This document describes testing policy and process for runtimes implementing the
+[Container Runtime Interface (CRI)](/contributors/devel/container-runtime-interface.md)
+to publish test results in a federated dashboard. The objective is to provide
```
resouer: When is the target timeline or release of this work?

yujuhong: We can start doing this as soon as the proposal is approved. The majority of the change will be in the test-infra repository, which is not constrained by the k8s code freeze timeline. I think we can reasonably target this in 1.10.

At the same time, we are trying to re-categorize e2e tests (https://github.com/kubernetes/kubernetes/issues/59001) so that the node conformance suite can be more representative, and that we'll have a separate feature test suite required in the next section. The individual tests are tracked in the spreadsheet: https://docs.google.com/spreadsheets/d/1Af7zSLDEDM3i5g-8MZi0oZ7zPt85rWI1kHMQybhS9H0/edit?usp=sharing
This is also expected to be done by 1.10

PatrickLang:
@jchauncey @feiskyer any comments? Seems like we could just run the Windows tests as they're added to critest and upload to testgrid

contributors/devel/cri-testing-policy.md
```diff
@@ -0,0 +1,108 @@
...
+because many features require integration of runtime, OS, or even the cloud
+provider. A higher-level integration tests provider better signals on vertical
+stack compatibility to the Kubernetes community. On the other hand, runtime
+developers are strongly encouraged to run low-level
+[CRI validation test suite](https://github.com/kubernetes-incubator/cri-tools/blob/master/docs/validation.md)
+for validation as part of their development process.
+
+## Required and optional tests
+
+Runtime maintainers are **required** to submit the tests listed below.
```
mikebrow: /s/submit/submit results for/

contributors/devel/cri-testing-policy.md
```diff
@@ -0,0 +1,108 @@
...
+To publish tests results, please submit a proposal in the
+[Kubernetes community repository](https://github.com/kubernetes/community)
+briefly explaining your runtime, providing at least two maintainers, and
+assigning the proposal to the leads of SIG-Node.
+
+These test results should be published under the `sig-node` tab, grouped by the
+runtimes, organized as follows.
+
+```
+sig-node -> sig-node-{Kubernetes-version} -> sig-node-{runtime-name} -> [page containing all test jobs]
```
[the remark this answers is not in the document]
yujuhong: I think keeping up to two release versions + master should be sufficient.
```
sig-node-master
sig-node-1.9
sig-node-1.8
```
It should be okay if runtime maintainers want to publish results, for, say, the master branch.
WDYT?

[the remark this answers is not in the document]
yujuhong: /cc @dchen1107

feiskyer:
> any comments? Seems like we could just run the Windows tests as they're added to critest and upload to testgrid

This document is actually focus on Node E2E tests. But @kubernetes/sig-windows-misc may prefer to pushlish the results on a new sig-windows page. @PatrickLang  @michmike WDYH?

yujuhong:
> This document is actually focus on Node E2E tests. But @kubernetes/sig-windows-misc may prefer to pushlish the results on a new sig-windows page. @PatrickLang @michmike WDYH?

I think Windows is a special case and it still early in the development phase. The suggestion from previous sig-node meetings is for windows to whitelist select tests for Windows to run. Once it gets to a better phase, we can then consider how to incorporate windows in our conformance testing.

contributors/devel/cri-testing-policy.md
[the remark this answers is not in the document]
derekwaynecarr: @yujuhong i think as a starting point N-2 from the moment they integrate is fine.  thanks for clarification.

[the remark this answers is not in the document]
yujuhong: Updated the PR to reflect this.

derekwaynecarr:
@yujuhong this looks good to me.  i would like @mrunalp or @runcom to weigh in as well.

mrunalp:
:+1:

contributors/devel/cri-testing-policy.md
```diff
@@ -0,0 +1,118 @@
...
+stack compatibility to the Kubernetes community. On the other hand, runtime
+developers are strongly encouraged to run low-level
+[CRI validation test suite](https://github.com/kubernetes-incubator/cri-tools/blob/master/docs/validation.md)
+for validation as part of their development process.
+
+## Required and optional tests
+
+Runtime maintainers are **required** to submit the tests listed below.
+ 1. Node conformance test suite
+ 2. Node feature test suite
```
resouer: @yujuhong One thing I want to amend is, for hypervisor based runtimes, it actually need run in mixed runtime mode to pass all node conformance tests. We may need to mention this somewhere?

Mixed runtime mode can be like: Kata + Docker, Kata + containerd etc. They will expose one endpoint to kubelet or use a cri-proxy.

yujuhong: IMO, the runtime maintainers can decide whether they are going to use proxy or not, but that seems immaterial to the policy.
Pure hypervisor-based runtime is not going to pass all the tests, but the test results will still provide useful signals to the community. I think as a SIG, we can discuss case-by-case, when it comes up, to tolerate more failures for these non-traditional runtimes.

PR #1745, opened 2018-02-05
Update sig-api-machinery meeting and agenda links

jennybuckley:
API Machinery uses zoom for SIG meetings, not hangouts. Also, the link to the meeting agenda didn't actually lead to the meeting agenda.


idvoretskyi:
/lgtm

/hold

@kubernetes/sig-api-machinery-pr-reviews PTAL

PR #1758, opened 2018-02-07
fix Service Catalog SIG meeting time

kibbles-n-bytes:
The Service Catalog SIG meeting time did not account for the change in time due to daylight savings in the United States.


arschles:
LGTM

PR #1767, opened 2018-02-08
Update leads for wg-app-def

bryanl:
Adding garethr and bryanl as leads for wg-app-def.


bryanl:
CLA should be fixed up now.

PR #1768, opened 2018-02-09
Update company info in Policy WG

tsandall:


cblecker:
@tsandall This is an auto-generated document. You will need to edit sigs.yaml (the source of truth) and then run `make` to generate the documents.

tsandall:
Thanks @cblecker!

PR #1775, opened 2018-02-09
Change KEP state from accepted to draft

mattfarina:
This PR is a proposal to make the initial state of a KEP more clear.
There have been misunderstandings on the means because of the way
accepted has been used elsewhere on the project.

For example, some reviewing the Application object/CRD PR though expected meant implementable or that this was the accepted design direction that would be expected to implement. Since this is a new process I would argue the naming should follow the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and attempt to be intuitive even to those not familiar with the process (which is most people at the moment).

This proposed change moves the term to draft which is used buy other projects
such as the IETF standards track.

/cc @jdumars @bgrant0607 @jbeda 

/sig architecture

### DIFF
--- a/keps/0001-kubernetes-enhancement-proposal-process.md
+++ b/keps/0001-kubernetes-enhancement-proposal-process.md
@@ -164,7 +164,7 @@ Metadata items:
     KEP filename.  See the template for instructions and details.
 * **status** Required
   * The current state of the KEP.
-  * Must be one of `accepted`, `implementable`, `implemented`, `deferred`, `rejected`, `withdrawn`, or `replaced`.
+  * Must be one of `draft`, `implementable`, `implemented`, `deferred`, `rejected`, `withdrawn`, or `replaced`.
 * **authors** Required
   * A list of authors for the KEP.
     This is simply the github ID.
@@ -222,9 +222,9 @@ Metadata items:
 
 A KEP has the following states
 
-- `accepted`: The KEP has been proposed and is actively being defined.
+- `draft`: The KEP has been proposed and is actively being defined.
   This is the starting state while the KEP is being fleshed out and actively defined and discussed.
-  The owning SIG has accepted that this is work that needs to be done.
+  The owning SIG has agreed that this is work that needs to be done, but is not ready for implementation.
 - `implementable`: The approvers have approved this KEP for implementation.
 - `implemented`: The KEP has been implemented and is no longer actively changed.
 - `deferred`: The KEP is proposed but not actively being worked on.
--- a/keps/0000-kep-template.md
+++ b/keps/0000-kep-template.md
@@ -16,7 +16,7 @@ approvers:
 editor: TBD
 creation-date: yyyy-mm-dd
 last-updated: yyyy-mm-dd
-status: accepted
+status: draft
 see-also:
   - KEP-1
   - KEP-2
--- a/keps/0002-controller-manager.md
+++ b/keps/0002-controller-manager.md
@@ -15,7 +15,7 @@ reviewers:
 approvers:
   - "@thockin"
 editor: TBD
-status: approved
+status: draft
 replaces:
   - contributors/design-proposals/cloud-provider/cloud-provider-refactoring.md
 ```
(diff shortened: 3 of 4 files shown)


keps/0001-kubernetes-enhancement-proposal-process.md
```diff
@@ -222,9 +222,9 @@ Metadata items:
 
 A KEP has the following states
 
-- `accepted`: The KEP has been proposed and is actively being defined.
+- `draft`: The KEP has been proposed and is actively being defined.
   This is the starting state while the KEP is being fleshed out and actively defined and discussed.
-  The owning SIG has accepted that this is work that needs to be done.
+  The owning SIG has agreed that this is work that needs to be done.
```
liggitt: even this is likely to be misread... suggest being very explicit: "The owning SIG has agreed that this is work that needs to be done, but is not ready for implementation yet."

liggitt:
I'm think this is much clearer... without close reading, "accepted" sounds like "good to go"

keps/0001-kubernetes-enhancement-proposal-process.md
mattfarina: Sounds good. Updated.

jbeda:
hahahaha. It was `draft` to begin with but was changed to `accepted` after some long discussion in SIG-architecture.

That is all documented here: https://github.com/kubernetes/community/pull/1584#issuecomment-357017991.

/cc @calebamiles @bgrant0607 

Personally I like `draft` better (and had it that way to begin with) but I know that Caleb and Brian have stronger feelings here.

liggitt:
> It was draft to begin with but was changed to accepted after some long discussion in SIG-architecture.

that's funny. fwiw, I would never guess "accepted" meant "SIG agrees with goals, but not necessarily with described approach"

jbeda:
It's documented but it is unfortunately (a) people don't read docs and (b) it isn't obvious from context.  https://github.com/kubernetes/community/blob/master/keps/0001-kubernetes-enhancement-proposal-process.md#kep-workflow

Brian? Caleb?

cblecker:
Another issue related to KEP states: https://github.com/kubernetes/community/issues/1793

bgrant0607:
`Accepted` is used in some other processes, but I can understand the confusion.

However, I disagree that `Draft` is better. It gives no indication what the stage of the process it's in, or even that there is a process. We have many, many, many docs that stay in Draft forever, regardless whether they were implemented or abandoned.

How about Provisional? Or WIP?

jbeda:
I like `provisional`.  It implies some level of endorsement but without being final.  Others have thoughts?

spiffxp:
other ideas... reviewed, endorsed, sponsored, owned

calebamiles:
My thoughts were initially that the action of interacting with Source Control (git) and the Hosting Provider (GitHub) should be incorporated into the KEP process so that merging a portion of a KEP signaled agreement on the content of the KEP which had been merged. For example, when a SIG agreed to the `Motivation` that part of the KEP would be merged and the SIG could begin iterating on the other sections of the KEP as they saw fit; and again just an example but merging the `Proposal` section would signal agreement on the initial design and implementation could begin.

That was a long way of saying that I don't particularly care what string is used and I honestly don't understand the fuss, or the need to retroactively change documents that other SIGs are using, when we have <10 of these documents to manage. To @bgrant0607's point, I would really like to see us prune documentation that is clearly no longer useful such that we don't continue to accumulate cruft in a source control system which in theory is capable of allowing any interested party to walk through historical changes to a file.

Regarding the desire to avoid hundreds of commits on a pull request, I entirely agree with the goal and when it seems as if the comment thread isn't useful anymore _someone_ should probably start a mailing list discussion or a thread on slack to have a focused conversation around disagreements and link that discussion back into the metadata section so posterity can find it again. I am mostly of the opinion that in general people need to be more involved with the actual project management required to arrive at consensus on an approach and implementation rather than relying on a magical process to take care of this or to ensure that the right questions are asked. Perhaps this is one opportunity for a specific responsibility for SIG PM.

jbeda:
We discussed this in SIG-Architecture and decided on `provisional` instead of `accepted`.  Discussion here: https://youtu.be/Os0zQoYL4XA?t=53m54s. I'm going to put together a PR to do that and also encourage people to commit early.  Also going to make it clear that the processes is available but not required for individual SIGs.

Closing this in favor of that.  Will reference the parent issue (#1793).

PR #1783, opened 2018-02-10
fix a duplication problem in admission_control_resource_quota.md

lijianfeng1993:


cblecker:
This is intended. This is a quota object named quota.

/close

PR #1788, opened 2018-02-11
fix a duplication problem in rescheduling.md

lijianfeng1993:


liggitt:
It would be helpful to consolidate these into a single PR

lijianfeng1993:
ok, thank you! @liggitt

PR #1799, opened 2018-02-14
Add team guidance section to org-owners-guide

spiffxp:
@kubernetes/sig-contributor-experience-pr-reviews

/committee steering
ref: https://groups.google.com/a/kubernetes.io/d/msg/steering/qnZuXlqK44Y/ugGn8lXdAwAJ

### DIFF
--- a/org-owners-guide.md
+++ b/org-owners-guide.md
@@ -54,6 +54,8 @@ Each organization should have the following teams:
 - teams for each repo `foo`
   - `foo-admins`: granted admin access to the `foo` repo
   - `foo-maintainers`: granted write access to the `foo` repo
+  - `foo-reviewers`: granted read access to the `foo` repo; intended to be used as
+    a notification mechanism for interested/active contributors for the `foo` repo
 - a `bots` team
   - should contain bots such as @k8s-ci-robot and @linuxfoundation that are
     neccessary for org and repo automation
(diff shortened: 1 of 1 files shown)


spiffxp:
/hold
for comment

Notably, I'm not sure whether I want to drag in the idea of pollinating the kubernetes-sig-foo-suffix teams across all orgs.  The reason we might want to do that is that you can't notify another org's teams.  To do this level of mirroring would require we either really smooth automation of adding people to github orgs and creating teams, or that we at least invite the 'k8s-mirror' users and train people to consume github notifications more actively through the mirror-based google groups.  

All of that seems like a bigger issue beyond my attempt to document the status quo, but maybe this needs to be sorted out for the kubernetes-sigs org?

org-owners-guide.md
```diff
@@ -47,8 +47,36 @@ Note, this list is subject to change.
...
+- an `owners` team
+  - should be populated by everyone who has `owner` privileges to the org
+  - gives users the opportunity to ping owners as a group rather than having to
+    search for individuals
+
+**NB**: Not all organizations in use today currently follow this team guidance.
+We are looking to coalesce existing teams towards this model, and use this model
+for all orgs going forward.  Notable discrepancies at the moment:
+
+- `foo-reviewers` teams that grant read access to each `foo` repo haven't
```
spiffxp: discussed during steering committee meeting, and we decided this pattern would be useful for the kubernetes-sigs org, so... nevermind, I'll take this back

assuming in-repo or multi-repo subprojects work out, it'd probably make more sense to have `subproject-reviewers` teams that correspond to subprojects rather than repos

spiffxp:
>I'm not sure whether I want to drag in the idea of pollinating the kubernetes-sig-foo-suffix teams across all orgs

Discussed during steering committee meeting today and consensus was we don't need to sort this out prior to moving forward with the kubernetes-sigs org, so I think this is good to go

idvoretskyi:
/lgtm

@spiffxp feel free to unhold if you're ready to go.

PR #1801, opened 2018-02-14
Proposal: HPA using external metrics

MaciekPytel:
This proposal describes an extension to the HPA v2 API that would allow autoscaling based on metrics coming from outside kubernetes cluster. This is not possible using existing API.

Common use-cases are autoscaling pods using a hosted service (ex. hosted message queue) or autoscaling in systems that have (so far) only partly moved to Kubernetes.

https://github.com/kubernetes/community/pull/1802 describes External Metrics API used internally by HPA to retrieve metrics. 

/sig-autoscaling

### DIFF
--- a/contributors/design-proposals/autoscaling/hpa-external-metrics.md
+++ b/contributors/design-proposals/autoscaling/hpa-external-metrics.md
@@ -1,0 +1,205 @@
+# **HPA v2 API extension proposal**
+
+# Objective
+
+[Horizontal Pod Autoscaler v2 API](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/hpa-v2.md) allows users to autoscale based on custom metrics. However, there are some use-cases that are not well supported by the current API. The goal of this document is to propose the following changes to the API:
+
+*   Allow autoscaling based on metrics coming from outside of Kubernetes. Example use-case is autoscaling based on a hosted cloud service used by a pod.
+*   Allow specifying per-pod target for global metrics. This makes more sense for many metrics than a global target (ex. 200 QPS / pod makes sense while a global target for QPS doesn't).
+
+# Overview
+
+A new External metric source will be added. It will identify a specific metric to autoscale on based on metric name and a label selector. The assumed model is that specific time series in monitoring systems are identified with metric name and a set of key-value labels or tags. The details vary in different systems (ref: [Prometheus](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels), [Stackdriver](https://cloud.google.com/monitoring/api/v3/metrics#time_series), [Datadog](https://docs.datadoghq.com/agent/tagging/), [Sysdig](https://www.sysdig.org/wiki/sysdig-user-guide/#user-content-filtering)), however, in general the adapter should be able to use metric name and a set of labels to construct a query sufficient to identify specific time series in underlying system.
+
+External and Object metrics will specify the desired target by setting exactly one of two fields: TargetValue (global target) or TargetAverageValue (per-pod target).
+
+
+# Multiple metric values
+
+Label selector specified by user can match multiple time series, resulting in multiple values provided to HPA. In such case the sum of all those values will be used for autoscaling. This is meant to allow autoscaling using a metric that is drilled down by some criteria not relevant for autoscaling a particular workload (for example to allow autoscaling based on a total number of HTTP requests, regardless of which HTTP method is used).
+
+If the need arises we can easily add other simple aggregations and allow user to choose one of them.
+
+# Example
+
+This is an example HPA configuration autoscaling based on number of pending messages in a message queue (RabbtMQ) running outside of cluster.
+
+```yaml
+kind: HorizontalPodAutoscaler
+apiVersion: autoscaling/v2beta2
+spec:
+  scaleTargetRef:
+    kind: ReplicationController
+    name: Worker
+  minReplicas: 2
+  maxReplicas: 10
+  metrics:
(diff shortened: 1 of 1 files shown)


MaciekPytel:
@kubernetes/sig-autoscaling-proposals 
@kubernetes/api-reviewers 

@DirectXMan12 - we discussed this during sig-autoscaling, can you take a look?

contributors/design-proposals/autoscaling/hpa-external-metrics.md
```diff
@@ -0,0 +1,144 @@
...
+        PodsMetricSourceType MetricSourceType = "Pods"
+        // ResourceMetricSourceType is a resource metric known to Kubernetes, as
+        // specified in requests and limits, describing each pod in the current
+        // scale target (e.g. CPU or memory).  Such metrics are built in to
+        // Kubernetes, and have special scaling options on top of those available
+        // to normal per-pod metrics (the "pods" source).
+        ResourceMetricSourceType MetricSourceType = "Resource"
+        // ExternalMetricSourceType is a global metric coming from
+        // outside of Kubernetes. It allows autoscaling based on information
+        // coming from components running outside of cluster
```
DirectXMan12: this is a bit of a documentation issue, but I think "coming from outside the cluster" is the wrong term here.  I think it's better to stress that this is for the case when the metrics aren't immediately associatable with objects inside the cluster.

For instance, I could run a DBaaS that runs outside the cluster, but still provide metrics to track usage per namespace, and thus not have to use the `external` type.

contributors/design-proposals/autoscaling/hpa-external-metrics.md
```diff
@@ -0,0 +1,144 @@
...
+        // MetricName is the name of a metric used for autoscaling in
+        // metric system.
+        MetricName string `json:"metricName" protobuf:"bytes,1,name=metricName"`
+
+        // MetricSelector is used to identify a specific time series
+        // within a given metric.
+        MetricSelector metav1.LabelSelector `json:"metricSelector" protobuf:"bytes,2,name=metricSelector"`
+
+        // CurrentValue is the current value of the metric (as a quantity)
+        CurrentValue resource.Quantity `json:"currentValue" protobuf:"bytes,3,name=currentValue"`
```
DirectXMan12: This also needs `currentAverageValue *resource.Quantity` as an optional field.

DirectXMan12:
holding due to needing some further some further consideration of the alternatives

contributors/design-proposals/autoscaling/hpa-external-metrics.md
MaciekPytel: Good point. Updated all comments as per your suggestion.

contributors/design-proposals/autoscaling/hpa-external-metrics.md
MaciekPytel: Added.

MaciekPytel:
@DirectXMan12 I've added a section discussing considered alternatives. Can you take a look?

DirectXMan12:
So, after a bit discussion with @MaciekPytel offline, I'm inclined to accept this proposal, with a caveat.  I'm still hesitant about potential confusion on the part of end-users (when do I use `external` vs `object`), so I think that we'll need to make super certain to collect user feedback and re-evaluate next release how we want to move forward with these changes.

I believe we still need an @kubernetes/api-reviewers on the sibling proposal, so we should hold off merge until that gets the ok.

eghobo:
@MaciekPytel sorry for stupid question (I honestly read full proposal ;)). But how this different from existing API (e.g. in our use-case we wrote api-adapter for external system and use ObjectMetricSourceType in HPA)?

mwielgus:
@eghobo With ObjectMetricSource users have to bind their metrics to some kubernetes object (thus possibly modify their definition in the monitoring solution). With external metrics we allow to use unmodified metrics (like pubsub queue lenghth, external load balancer qps, etc) in HPA.

MaciekPytel:
@eghobo
Short answer: If you can identify a metric you want to use using Object, than use it. Otherwise you need External.

The goal is to allow autoscaling based on metrics that have no inherent relationship to any k8s object. For example if your application use a hosted service and you want to autoscale based on its metrics, which contain no reference to your kubernetes object. 

The design of Object metric (even the name :) ) is suited to use metrics inherently associated with some k8s object.  `Target` field in ObjectMetricSource is useless for identifying metrics that have no such association. As discussed in `Using Object metric source for external metrics` section there are some workarounds that allow using Object for autoscaling such metrics. However, regardless which workaround you choose, at the end of the day you will need to explicitly define the metric you want to use (ie. specify the same set of information you give to External metric). You may write adapter so that using a specific `Target` gives you some arbitrary metric of course, but that's stretching the Object metric API. Also that means everyone will also need to somehow do that in their adapter. Instead the idea is to just create a standard way of accessing such metrics.

DirectXMan12:
/hold

for api review

mwielgus:
#60096 is approved by API reviewer (@thockin). 

@DirectXMan12  Can we unhold?

DirectXMan12:
/unhold

yep

DirectXMan12:
(although I kind-of wanted approval on the external metrics API, too)

eghobo:
@MaciekPytel thx for details, we are using adapter to access external metrics now. Sorry, but I am very skeptical about `standard way of accessing such metrics`. E.g. we have three monitoring systems Graphite/JMX/Prometheus in place now and there is no way to find `standard way` for them without writing adapter ):

PR #1802, opened 2018-02-14
Proposal: External Metrics API

MaciekPytel:
This proposal introduces External Metrics API. It is based on existing Custom Metrics API and it provides metrics not related to any Kubernetes object.

https://github.com/kubernetes/community/pull/1801 describes how HPA would use this API.

/sig-instrumentation


MaciekPytel:
@kubernetes/sig-instrumentation-proposals 
cc: @kubernetes/sig-autoscaling-proposals

contributors/design-proposals/instrumentation/external-metrics-api.md
```diff
@@ -0,0 +1,87 @@
...
+
+A single value should always be returned if MetricSelector specifies a single value for every label defined for a given metric.
+
+## Metric names
+
+Custom Metrics API [doesn't allow using certain characters in metric names](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/custom-metrics-api.md#metric-names). The reason for that is a technical limitation in GO libraries. This list of forbidden characters includes slash (`/`). This is problematic as many systems use slashes in their metric naming convention.
+
+Rather than expect metric adapters to come up with their custom ways of handling that this document proposes introducing `\|` as a custom escape sequence for slash. HPA controller will automatically replace any slashes in MetricName field for External metric with this escape sequence.
+
+Otherwise the allowed metric names are the same as in Custom Metrics API.
```
piosz: Is there any plan to adjust this behaviour in the future if the limitation of Go library is addressed?

piosz:
LGTM from me, please wait for the approval from the other reviewers.

DirectXMan12:
So, after a bit discussion with @MaciekPytel offline, I'm inclined to accept this proposal, with a caveat.  I'm still hesitant about potential confusion on the part of end-users (when do I use `external` vs `object`), so I think that we'll need to make super certain to collect user feedback and re-evaluate next release.

I believe we still need an @kubernetes/api-reviewers on the sibling proposal for the API, but will merge once they give the sign-off.

contributors/design-proposals/instrumentation/external-metrics-api.md
```diff
@@ -0,0 +1,87 @@
...
+    // the value of the metric
+    Value resource.Quantity
+}
+```
+
+# Semantics
+
+## Namespaces
+
+Kubernetes namespaces don't have a natural 1-1 mapping to metrics coming from outside of Kubernetes. It is up to adapter implementing the API to decide which metric is available in which namespace. In particular a single metric may be available through many different namespaces.
```
brancz: I'm not sure I understand why we want this in the first place. One of the problems with the custom metrics API is as you stated that the metric may be completely decoupled from Kubernetes, so why introduce Kubernetes paradigms anyways? Seems like we should just leave this out and leave it up to the user to add any select by labels as they wish.

contributors/design-proposals/instrumentation/external-metrics-api.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Metric names
+
+Custom Metrics API [doesn't allow using certain characters in metric names](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/custom-metrics-api.md#metric-names). The reason for that is a technical limitation in GO libraries. This list of forbidden characters includes slash (`/`). This is problematic as many systems use slashes in their metric naming convention.
+
+Rather than expect metric adapters to come up with their custom ways of handling that this document proposes introducing `\|` as a custom escape sequence for slash. HPA controller will automatically replace any slashes in MetricName field for External metric with this escape sequence.
+
+Otherwise the allowed metric names are the same as in Custom Metrics API.
+
+## Access Control
```
brancz: A clear idea on access control (possibly with the mentioned ExternalMetricsPolicy) is definitely a GA requirement in my opinion. To get off the ground I think we can evaluate what makes sense once this lands.

contributors/design-proposals/instrumentation/external-metrics-api.md
MaciekPytel: The main point is to allow writing RBAC rules to allow or forbid access to a given metric in a given namespace. I think this is good enough reason to add it even if I expect some (most?) adapters may choose to just ignore this parameter. Of course adapters are free to use it for some custom access control if required.

contributors/design-proposals/instrumentation/external-metrics-api.md
MaciekPytel: Agreed. I left it out of initial version because adding a policy would be a backward compatible change and I'd like to get some feedback from beta before committing to it.

contributors/design-proposals/instrumentation/external-metrics-api.md
brancz: that makes sense :+1:

MaciekPytel:
/unhold
The API has been approved from api-reviewers side by @thockin on https://github.com/kubernetes/kubernetes/pull/60079.

contributors/design-proposals/instrumentation/external-metrics-api.md
```diff
@@ -0,0 +1,87 @@
...
+
+A request for a given metric may return multiple values if MetricSelector matches multiple time series. Each value should include a complete set of labels, which is sufficient to uniquely identify a timeseries.
+
+A single value should always be returned if MetricSelector specifies a single value for every label defined for a given metric.
+
+## Metric names
+
+Custom Metrics API [doesn't allow using certain characters in metric names](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/custom-metrics-api.md#metric-names). The reason for that is a technical limitation in GO libraries. This list of forbidden characters includes slash (`/`). This is problematic as many systems use slashes in their metric naming convention.
+
+Rather than expect metric adapters to come up with their custom ways of handling that this document proposes introducing `\|` as a custom escape sequence for slash. HPA controller will automatically replace any slashes in MetricName field for External metric with this escape sequence.
```
DirectXMan12: I really dislike this escape sequence.  Is it really necessary?

caipre:
There is presently a broken link from https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/hpa-external-metrics.md to this document. What's the status of this proposal?

DirectXMan12:
the feature is implemented, but I think the proposal somehow didn't get merged.   @brancz ok if I just merge this this and the other HPA v2 features into an HPA v2 kep, so we can track the evolution of HPA v2 in one place?

brancz:
@DirectXMan12 sounds good to me.

MaciekPytel:
As @DirectXMan12 wrote this has been implemented for a long time. Leads of both interested sigs (autoscaling and instrumentation) agreed to merging it in comments above, so I'm removing hold. 

/hold cancel

PR #1813, opened 2018-02-19
Update/clarify KEP process

### LINKED ISSUE #1793: Confusion over states in the KEP-1 proposal
This issue arises out of a discussion in SIG-Apps today, and applies to 0001-kubernetes-enhancement-proposal-process.md

For background, there has been movement in the community to use the KEP process for more general standards (e.g. CRDs, conventions, or ecosystem tooling), not just changes to Kubernetes core. This is causing confusion.

The states defined in the KEP-1 (implementable, implemented, deferred, etc.) are significantly different than the states define for RFCs, Rust RFCs, PEPs, and similar specification processes. This is causing confusion about how to actually use KEPs.

Specifically, it appears to be a process designed to handle only _one implementation_ of any given proposal. (`implementable`/`implemented`/`deferred`). Thus it seems only useful for one project that admits only one canonical implementation of a KEP. So it becomes generally unusable when trying to define interchange formats, protocols, schemata, etc.

The core confusion comes from conflating _specification_ and _implementation_. First, an implementation can be achieved against a non-existent or poor specification. Whether something is _implementable_ or _implemented_ does not speak to the quality of the specification. Neither should the existence of an implementation foreclose the possibility of writing a good specification. Second, a specification does not necessarily need "an implementation" to be useful. One might "specify" the purpose or meaning of labels or annotations, but without having "an implementation". (KEP-1, for example, does not have "an implementation".)

As a remedial action, I would propose that we do one of two things:

- Agree to not use KEP-1 for anything but proposals to extend the functionality of Kubernetes core, and leave everyone else (SIGs, etc) to figure out their own process
- Supersede KEP-1 with a new KEP that outlines states as they are explained in [the RFC process](https://en.wikipedia.org/wiki/Request_for_Comments) or another standards process that is not strongly tied to just one canonical implementation.

jbeda:
Based on discussions on SIG-Architecture we are changing the `accepted` state to `provisional`.

Also made the process in the template much more clear.  Expanded KEP README.md to add some FAQs.  Most critical is that SIGs are not required to use the KEP process.

Fixes #1793.
Supersedes #1775.

Signed-off-by: Joe Beda <joe.github@bedafamily.com>

### DIFF
--- a/keps/0000-kep-template.md
+++ b/keps/0000-kep-template.md
@@ -53,7 +53,7 @@ To get started with this template:
 1. **Merge early.**
   Avoid getting hung up on specific details and instead aim to get the goal of the KEP merged quickly.
   The best way to do this is to just start with the "Overview" sections and fill out details incrementally in follow on PRs.
-  View anything marked as a `provisional` as a working document.
+  View anything marked as a `provisional` as a working document and subject to change.
   Aim for single topic PRs to keep discussions focused.
   If you disagree with what is already in a document, open a new PR with suggested changes.
 
--- a/keps/0000-kep-template.md
+++ b/keps/0000-kep-template.md
@@ -1,5 +1,5 @@
 ---
-kep-number: draft-YYYYMMDD
+kep-number: 0
 title: My First KEP
 authors:
   - "@janedoe"
@@ -16,7 +16,7 @@ approvers:
 editor: TBD
 creation-date: yyyy-mm-dd
 last-updated: yyyy-mm-dd
-status: accepted
+status: provisional
 see-also:
   - KEP-1
   - KEP-2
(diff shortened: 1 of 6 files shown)


keps/README.md
```diff
@@ -1,7 +1,42 @@
 # Kubernetes Enhancement Proposals (KEPs)
 
-This directory contains both the template and process for KEPs.
+A Kubernetes Enhancement Proposal (KEP) is a way to propose, communicate and coordiante on new efforts for the Kubernetes project.
+You can read the full details of the project in [KEP-1](0001-kubernetes-enhancement-proposal-process.md).
 
-This process is still in an _alpha_ state and is opt-in for those that want to provide feedback for the process.
+This process is still in a _beta_ state and is opt-in for those that want to provide feedback for the process.
```
errordeveloper: Maybe we could define graduation requerements?

errordeveloper:
I don't want to slow down the process, but it'd be very nice to consider format enforcement and validation as part of graduation requirements.
Perhaps we could use [`markdownfmt`](https://github.com/shurcooL/markdownfmt) or something similar,  and discuss having a tool to validate YAML-frontmatter has the right fields, as it's hard for reviewers to track all the fields.

keps/0000-kep-template.md
```diff
@@ -37,16 +37,25 @@ The title should be lowercased and spaces/punctuation should be replaced with `-
...
+  Name it `NNNN-YYYYMMDD-my-title.md` where `NNNN` is the KEP number that was allocated.
+1. **Fill out the "overview" sections.**
+  This includes the Summary and Motivation sections.
+  These should be easy if you've preflighted the idea of the KEP with the appropriate SIG.
+1. **Create a PR.**
+  Assign it to folks in the SIG that are sponsoring this process.
+1. **Merge early.**
+  Avoid getting hung up on specific details and instead aim to get the goal of the KEP merged quickly.
+  The best way to do this is to just start with the "Overview" sections and fill out details incrementally in follow on PRs.
+  View anything marked as a `provisional` as a working document.
```
timothysc: working document, and subject to change.

keps/README.md
```diff
@@ -1,7 +1,42 @@
...
+
+1. Socialize an idea with a sponsoring SIG.
+   Make sure that others think the work is worth taking up and will help review the KEP and any code changes required.
+2. Follow the process outlined in the [KEP template](0000-kep-template.md)
+
+## FAQs
+
+### Do I have to use the KEP process?
+
+Generally, no!
```
timothysc: I'd like a "No, but... statement"   imo KEPs are pretty low overhead and provide some structure which has been lacking in the community for some time.  I think outlining the benefits would help.

jbeda:
@errordeveloper @timothysc -- PTAL!

keps/README.md
```diff
@@ -1,7 +1,59 @@
...
+
+If the KEP is mainly restricted to one SIG's purview then it should be in a KEP directory for that SIG.
+If the KEP is widely impacting much of Kubernetes, it should be put at the root of this directory.
+If in doubt ask [SIG-Architecture](../sig-architecture/README.md) and they can advise.
+
+### What will it take for KEPs to "graduate" out of "beta"?
+
+Things we'd like to see happen to consider KEPs well on their way:
+* A set of KEPs that show healthy process around describing an effort and recording decisions in a reasonable amount of time.
+* KEPs exposed on a searchable and indexable web site.
```
calebamiles: I don't believe we ever agreed that this was a requirement but a really nice to have at some point

keps/sig-cluster-lifecycle/0003-cluster-api.md
```diff
@@ -48,16 +45,16 @@ last-updated: 2018-01-22
 
 ## Summary
 
-We are building a set of Kubernetes cluster management APIs to enable common cluster lifecycle operations (install, upgrade, repair, delete) across disparate environments. 
-We represent nodes and other infrastructure in Kubernetes-style APIs to enable higher level controllers to update the desired state of the cluster (e.g. the autoscaling controller requesting additional machines) and reconcile the world with that state (e.g. communicating with cloud providers to create or delete virtual machines). 
+We are building a set of Kubernetes cluster management APIs to enable common cluster lifecycle operations (install, upgrade, repair, delete) across disparate environments.
```
calebamiles: can you move these whitespace changes to another PR or ask @cheftako to run a linting pass?

keps/README.md
jbeda: It is just eliminating trailing whitespace.  I'll make sure that I do that in a separate commit/PR in the future but I don't think it is worth the time to fix it up in this PR.

jbeda: FWIW -- my editor did this automatically.  As we make pre-submit checks here we can make no trailing whitespace part of that linting.

jdumars:
/lgtm
Discussed in SIG Arch office hour

PR #1814, opened 2018-02-19
Community site KEP

jbeda:
This is the start of a KEP aimed at creating a community facing web site.

I'd like to agree that this is worth exploring and get this PR in.  Details of the proposal will come in follow on PRs.

I also created a sub-directory for SIG-contributor-experience KEPs.

I'm tagging sig-architecture as I'm thinking this is how we publish KEPs themselves.

(Process is being updated in #1813 and I'm assuming that'll merge.)


pmorie:
I'm +1 on this idea. Sounds like this is the right place for an onramp to be linked from (or hosted?)

cblecker:
/hold
Holding for discussion. contribex has been discussing this already, and while there is consensus around having a better and more discoverable way to get to the information in this repo, there are certain implementation details (top level community.k8s domain, what information gets included, etc) that we haven't decided on yet.

cc: @grodrigues3 @parispittman @Phillels @tpepper @guineveresaenger @castrojo @idvoretskyi @spiffxp

jbeda:
@cblecker -- that can all be talked about in the KEP itself. This PR just says "we should scope out having a community oriented site".

The workflow of the KEP process is such that the first PR should be that it is an idea worth exploring.  Merging now doesn't mean that it is improved but gives us a place to hang details off of. 

"contribex has been discussing this already" should be enough to merge this.

cblecker:
@jbeda -- I think the specific thing that is included in this PR that I'm not sure if we have consensus on is "community.kubernetes.io", or if we are going to look at some other solution (wiki, static sub-site under kubernetes.io, etc).

The other bit that I think I'm confused about is the use of the KEP process for this. I know we do have an umbrella issue at the moment (https://github.com/kubernetes/community/issues/1413) covering work on the contributor guide content.

keps/sig-contributor-experience/0005-community-site.md
```diff
@@ -0,0 +1,104 @@
...
+
+* Provide a new site (`community.kubernetes.io`).
+* Publish information that is currently in the [community repo](https://github.com/kubernetes/community).
+* Build some simple tools to enhance discoverability within the site.
+  This could include features such as automatically linking KEP and SIG names.
+
+### Non-Goals
+
+* Discover and bring together information from multiple orgs/repos.
+* Create a super dynamic back end.  This is most likely served best with a static site.
```
errordeveloper: Some basic notes to recommend implementation maybe helpful. I hear that SIG-docs are looking to move `k8s.io` to Hugo, perhaps `community.k8s.io` could pioneer use of Hugo instead of copying anything from docs repo.

keps/sig-contributor-experience/0005-community-site.md
```diff
@@ -0,0 +1,104 @@
...
+How will we know that this has succeeded?
+Gathering user feedback is crucial for building high quality experiences and SIGs have the important responsibility of setting milestones for stability and completeness.
+Hopefully the content previously contained in [umbrella issues][] will be tracked in the `Graduation Criteria` section.
+
+[umbrella issues]: https://github.com/kubernetes/kubernetes/issues/42752
+-->
+
+## Implementation History
+
+## Drawbacks [optional]
```
errordeveloper: I think you can safely delete Drawbacks and Alternative sections.

keps/sig-contributor-experience/0005-community-site.md
```diff
@@ -0,0 +1,104 @@
...
+
+* Discover and bring together information from multiple orgs/repos.
+* Create a super dynamic back end.  This is most likely served best with a static site.
+
+## Proposal
+
+TBD.
+
+
+### Risks and Mitigations
```
errordeveloper: Not sure if there is anything meaningful to be said here...

keps/sig-contributor-experience/0005-community-site.md
```diff
@@ -0,0 +1,104 @@
...
+## Summary
+
+We need a way to organize and publish information targeted at contributors.
+In order to continue to scale the Kubernetes community we need a convenient, scalable and findable way to publish information.
+
+## Motivation
+
+While the current kubernetes.io site is great for end users, it isn't often used by or aimed at project contributors.
+Instead, most contributors look at documentation in markdown files that are spread throughout a wide set of repos and orgs.
+It is difficult for users to find this documentation.
```
errordeveloper: It would be worth mentioning that standalone website will provide more visibility into the community process. Right now, as a potential new contributor or curious user, you have to know what to look for, what KEPs are etc. We could have that have an index of e.g. in-flight KEPs by SIG, which I think would be very helpful.
Also an index of community events, meetup groups would be very helpful.

keps/sig-contributor-experience/0005-community-site.md
errordeveloper: Maybe we should also explicitly say about other things that we won't do, e.g. index of job vacancies and contact details for consulting service providers...

idvoretskyi:
cc @kubernetes/sig-docs-maintainers

keps/sig-contributor-experience/0005-community-site.md
jbeda: I'm avoiding going too deep into implementation.  The first step of a KEP is to have the goals outlined and merge that fast.  We can cover the *how* in a follow on PR.

keps/sig-contributor-experience/0005-community-site.md
jbeda: Once we do have an implementation proposed I would like to contrast.  The main set of drawbacks/alternatives is around a separate site vs. trying to do something combined with k8s.io.

keps/sig-contributor-experience/0005-community-site.md
jbeda: We can figure it out as we fill out the rest of the KEP.

jbeda:
Pushed some updates with some more clarity.

@cblecker -- I put in some question marks on the name as I don't want to bikeshed here.  As for the mechanism, I lean to a separate site and I think that is implied by the description but I'm open to exploring other ideas as long as we close on *something* soon.  I want to get this done to be able to host something like KEPs.

As for using the KEP process -- this was suggested by @castrojo.  I'm happy to use any other mechanism the SIG suggests but I'm also would like to have *some* timeline to getting to some decision here.  The KEP process lets us have structured conversations and lock down progress as we go.

As for #1413 -- this goes beyond the contributor guide.  This is a place where we would host the contributor guide but it doesn't need to be there.

@idvoretskyi -- We've discussed this in the past with SIG-docs folks but this proposal is that this site would have a very different target audience vs. our main site.

idvoretskyi:
> We've discussed this in the past with SIG-docs folks but this proposal is that this site would have a very different target audience vs. our main site.

@jbeda agree, but I'd like the SIG to have an insight here as well.

keps/sig-contributor-experience/0005-community-site.md
jberkus: Implementation will be determined by goals and tool availability; that belongs in the detail of the KEP, but shouldn't be in the goals section, as Joe says.

parispittman:
I'm all in on a newly designed community site and figuring out a better technical implementation of the discoverability of our docs.  

My concern for starting the work now is potential double efforts that may be going on with Linux Foundation to revamp; we've just heard about this effort. @castrojo is pulling LF folks in here for comment.  The Docs team will be having another Hugo discussion at their meeting next week and sounds positive. 

Me talking out loud vs needing comment - Also concerned with the 'code contributor' and 'community contributor' persona(s) that SIG Docs are working on - we'd have three sources for contributor items at that point: 1-github, 2-this new site, 3-personas on the k8s site. We can probably automate these but unsure.

castrojo:
Here are some notes from our discussion, including that the Docs team was already moving to hugo anyway:

Oustanding KEP: https://github.com/kubernetes/community/pull/1814

## ContribX Requirements
- Discoverable/Searchable
- Reflects the single source of truth -- community repo (user friendly facade)
- Community Photos - More welcoming
- Getting started guide (tips and tricks)
- List of communication platforms
- Blatantly obvious that you’re in a developer zone
- Anyone should be able to update community content transparently
- We want owners for community site stuff in docs repo - 
- Ephemeral content (e.g. Kubecon agenda)

## Docs Notes (Zach Cornelissen and Caityn)

- Docs considering migrating to Hugo - status: under discussion
- Hugo
- Plays well with netlify
- Works well for internationalization
- Andrew’s script going away! 
- May fix the problem entirely
- Blog moving off blogspot
- Needs a process to skip dual +1’s from SIG Docs members when pushing into the community section of the repo. (aka. don’t bottleneck the whole repo on SIG Docs)
- Simplest propsal: have approvers from Contribex in these directories in docs.

Other ephemeral contributor content (events etc.) can also he handled by Hugo.

errordeveloper:
@jbeda sure, LMGT - merge early, as we said. I was just commenting in case I don't get a chance to review the follow up.

castrojo:
Ihor explained the process to us, basically what we do is merge this, then open a new issue that is then the main discussion for this KEP. So I'll merge this and file a discussion issue. 

/unhold
/lgtm

parispittman:
We haven't heard from LF and I believe this to be a priority for surfacing docs and creating a centralized point for contributors. Let's go. 

+1 - happy to be an approver; @jbeda , @castrojo - pls be reviewers. This is my first KEP, too, so help with process would be great @jbeda 

/lgtm

PR #1835, opened 2018-02-21
Formal scalability processes design proposal

shyamjvs:
We've been having various scattered discussions on how scalability/performance regressions are affecting k8s and its eng productivity. To try to solve those issues, I've assimilated some of my thoughts for formal scalability processes into [this google doc](https://docs.google.com/document/d/1Jv_QbehMTf0t2YVS4pYM_iSqWpg777JDRRbIUKXErYI/edit?usp=sharing), which changed into the current form after some great inputs from sig-scale folks (@wojtek-t, @wasylkowski-a, @gmarek, @porridge - thanks!).

This PR is a markdown version of that doc. PTAL

/cc @thockin @smarterclayton (as steering committee members)
/cc @spiffxp (as sig-testing lead)
/cc @wojtek-t (as sig-scalability lead)

### DIFF
--- a/sig-scalability/processes/formal-scalability-processes.md
+++ b/sig-scalability/processes/formal-scalability-processes.md
@@ -1,0 +1,136 @@
+# Formal Scalability Processes
+
+_by Shyam JVS, Google Inc_
+
+**February 2018**
+
+## Introduction
+
+Scalability is a very crucial aspect of kubernetes and has allowed many customers to adopt it with confidence. K8s [started scaling to 5000](http://blog.kubernetes.io/2017/03/scalability-updates-in-kubernetes-1.6.html) nodes beginning from release 1.6. Building and maintaining a performant and scalable system needs conscious efforts from the whole developer community. Lack of solid measures have caused problems (both scalability and release-related) in the past - for e.g during [release-1.7](https://github.com/kubernetes/kubernetes/issues/47344), [release-1.8](https://github.com/kubernetes/kubernetes/issues/53255) and [in general](https://github.com/kubernetes/kubernetes/issues/56062). We need them to ensure that the effort is well-streamlined with proper checks and balances in place. Of course they may evolve over time to suit the community/project’s needs better.
+
+## Goal
+
+Our goal is to make scalability of kubernetes sustainable. That is, to let k8s scalability thrive despite the ever-increasing rate of developer contributions. More concretely, we want to achieve the following criteria (subject to future changes):
+
+`Engineering time spent on debugging of scalability regressions is < 2 days/month`
+
+## Overview
+
+Development process can broadly be divided into 3 phases - i.e design, implementation and testing. Each phase poses different challenges wrt scalability. And not every change needs to pass through each phase. For e.g, a refactoring, dependency update or a simple code change would likely not have a design phase, while a new feature (e.g apiserver webhooks) would. We propose few processes for each of these phases to help with scalability.
+
+At a high-level, these processes are (in the order in which we want to add them):
+
+- [Implementation (or) pre-submit phase](#implementation--pre-submit-phase):
+  - Add optional PR jobs for scalability (both medium and large scale)
+  - Add mandatory presubmit jobs for scalability (only medium scale - fast & healthy)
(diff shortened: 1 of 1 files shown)


shyamjvs:
cc @kubernetes/sig-scalability-misc @kubernetes/sig-testing-misc

sig-scalability/processes/formal-scalability-processes.md
```diff
@@ -0,0 +1,136 @@
...
+- Medium-scale jobs\
+  ([relevant feature request](https://github.com/kubernetes/test-infra/issues/5677))\
+  Jobs that do not consume too many resources (light on quota-usage), not that high on cost and time and low-risk wrt outages, etc. These can be triggered by ~anybody. The following would be available here (can change over time):
+
+  - Kubemark-500 (consumes ~80 vCPUs)
+  - GCE-100 (consumes ~100 vCPUs)
+
+- Large-scale jobs\
+  ([relevant feature request](https://github.com/kubernetes/test-infra/issues/5051))\
+  Jobs that are much heavier wrt resources, time, cost and are high-risk. Their trigger would be access-restricted to a group of few googlers for those reasons. To begin with, the group would consist of key googlers from sig-scalability. Also these jobs need quota-checking before trigger and are to be used sparingly. The following would be available here (can change over time):
```
thockin: We need to get past "key googlers" - are there no other sig-scale members tat could be authorized for this?

sig-scalability/processes/formal-scalability-processes.md
```diff
@@ -0,0 +1,136 @@
...
+
+  - Kubemark-2000 (consumes ~270 vCPUs)
+  - Kubemark-5000 (consumes ~700 vCPUs)
+  - GCE-500 (consumes ~520 vCPUs)
+  - GCE-2000 (consumes ~2030 vCPUs)
+
+#### Mandatory pre-merge jobs
+([relevant feature request](https://github.com/kubernetes/test-infra/issues/4445))
+
+These are scalability jobs that would run against the PR at the head of the submit-queue right before merging. Note that they wouldn’t be running by default against each opened PR due to resource, cost and other constraints. Further, these should be fast and healthy enough in order to not hurt our merge rate. If the failure rate of their corresponding post-submit jobs goes >2% (measured over last 1 week), it would lead to a P0 issue needing immediate attention from sig-scalability. If needed, this metric may be modified later to subtract false positives (i.e failures belonging to a regression that has been fixed later). Next, these jobs should also be available and open for the author of a PR to be able to kick them - as they might want to verify a fix in case the pre-merge job failed on their PR. The following would be available here (can change over time):
```
thockin: If they do not run on every PR, when do they run?

sig-scalability/processes/formal-scalability-processes.md
shyamjvs: Agree.. We should add more members from sig-scale. I'm just not super sure how granting access to our google-owned projects to outside folks works (as folks might need to check quota availability in the project before kicking them). Eventually we should try to get rid of such manual overhead somehow though.

wojtek-t: This should be solved at the test-infra level.

Before kicking the test, prow (or whatever else) should be checking how many other tests are running at a time, for how many we have quota in the project and depending on that accept or reject the request.

shyamjvs: Agree

sig-scalability/processes/formal-scalability-processes.md
shyamjvs: The idea is to try to run those only during pre-merge round (i.e not in the initial round of pre-submits which happens for ~every single PR opened and might be a waste). There is some discussion going around how this might cause inconsistencies in our current SQ frameowrk under https://github.com/kubernetes/test-infra/issues/4445. Hopefully we reach some consensus there.

shyamjvs:
Thanks for the review @thockin.
Wrt flakiness - I have some ideas to get rid of them by making our tests more robust. Hopefully I'll be able to progress there by this EOQ.

PR #1836, opened 2018-02-21
List desired cloud provider git repositories as subprojects

calebamiles:
This is a statement of intent to create new Git repositories for the
cloud providers to extract from `kubernetes/kubernetes`. Handling
repositories for newly donated cloud providers will be handled
separately

### DIFF
--- a/sigs.yaml
+++ b/sigs.yaml
@@ -1643,6 +1643,7 @@ workinggroups:
       mailing_list: https://groups.google.com/forum/#!forum/kubernetes-wg-app-def
   - name: Cloud Provider
     dir: wg-cloud-provider
+    label: cloud-provider
     mission_statement: >
       Charter can be found [here](https://docs.google.com/document/d/1m4Kvnh_u_9cENEE9n1ifYowQEFSgiHnbw43urGJMB64/edit#)
     leads:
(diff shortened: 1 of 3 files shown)


calebamiles:
**cc: @kubernetes/sig-architecture-feature-requests**

jdumars:
I don't think it affects this PR, but there was discussion about these repos being more than cloud provider code - meaning volume and other "implementation-specific" components.  There wasn't a good suggestion about what to call it that was generic enough, but perhaps @cheftako @jagosan or @wlan0 can weigh in on that.  Otherwise, LGTM

sigs.yaml
```diff
@@ -1663,6 +1663,31 @@ workinggroups:
...
+      - https://raw.githubusercontent.com/kubernetes/aws-cloud-provider/master/OWNERS
+    - name: azure-cloud-provider
+      owners:
+      - https://raw.githubusercontent.com/kubernetes/azure-cloud-provider/master/OWNERS
+    - name: cloudstack-cloud-provider
+      owners:
+      - https://raw.githubusercontent.com/kubernetes/cloudstack-cloud-provider/master/OWNERS
+    - name: gce-cloud-provider
+      owners:
+      - https://raw.githubusercontent.com/kubernetes/gce-cloud-provider/master/OWNERS
```
mikedanese: gcp cloud provider instead of gce?

idvoretskyi: +1. It should be GCP (based on the existing names conventions).

sigs.yaml
```diff
@@ -1663,6 +1663,31 @@ workinggroups:
...
+    - name: openstack-cloud-provider
+      owners:
+      - https://raw.githubusercontent.com/kubernetes/openstack-cloud-provider/master/OWNERS
+    - name: ovirt-cloud-provider
```
idvoretskyi: Do we have an existing actively supported piece of codebase for ovirt?

calebamiles: https://github.com/kubernetes/kubernetes/tree/master/pkg/cloudprovider/providers/ovirt

calebamiles:
changed, @mikedanese

calebamiles:
I'm not sure if the script properly handles changes to the template, @jdumars. Since I _definitely_ didn't edit any `README.md` files by hand I was going to ignore the failure

jdumars:
Ok, with non-submitter approval, and 2 legit LGTMs, removing hold and merging.

PR #1843, opened 2018-02-23
chore(*): fix typos

bbbmj:

### DIFF
--- a/CLA.md
+++ b/CLA.md
@@ -59,7 +59,7 @@ You must also set your [git e-mail](https://help.github.com/articles/setting-you
 to match this e-mail address as well.
 
 If you've already submitted a PR you can correct your user.name and user.email
-and then use use `git commit --ammend --reset-author` and then `git push` to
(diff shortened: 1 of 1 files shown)


grodrigues3:
/ok-to-test

Why the -f for the push command?  Is that necessary?

CLA.md
```diff
@@ -59,7 +59,7 @@ You must also set your [git e-mail](https://help.github.com/articles/setting-you
 to match this e-mail address as well.
 
 If you've already submitted a PR you can correct your user.name and user.email
-and then use use `git commit --ammend --reset-author` and then `git push` to
+and then use use `git commit --amend --reset-author` and then `git push -f` to
```
cblecker: To keep the style consistent, can we use the `--force` flag instead of `-f`?

bbbmj:
fixed @cblecker @grodrigues3 PTAL

CLA.md
bbbmj: done

PR #1849, opened 2018-02-24
Add ability to have custom content in OWNERS_ALIASES

cblecker:
Not super pretty, but it works.

### DIFF
--- a/generator/app.go
+++ b/generator/app.go
@@ -41,8 +41,10 @@ const (
 	aliasesOutput = "OWNERS_ALIASES"
 	indexFilename = "README.md"
 
-	beginMarker = "<!-- BEGIN CUSTOM CONTENT -->"
-	endMarker   = "<!-- END CUSTOM CONTENT -->"
+	beginCustomMarkdown = "<!-- BEGIN CUSTOM CONTENT -->"
+	endCustomMarkdown   = "<!-- END CUSTOM CONTENT -->"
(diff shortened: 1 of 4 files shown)


cblecker:
/cc @cjwagner 
Hey Cole.. Will the approve plugin still process the file properly if we have comments in it?

cjwagner:
> Hey Cole.. Will the approve plugin still process the file properly if we have comments in it?

Yes, we use a YAML parser to read the file. It should handle comments properly. Here is an OWNERS file that has a comment: https://github.com/kubernetes/steering/blob/4ba203d5dfa575aae1e230d411fc5ee2c0368923/OWNERS

cblecker:
awesome! thanks @cjwagner

generator/app.go
```diff
@@ -244,7 +269,7 @@ func createGroupReadme(groups []Group, prefix string) error {
 
 		outputPath := filepath.Join(outputDir, indexFilename)
 		readmePath := filepath.Join(baseGeneratorDir, templateDir, fmt.Sprintf("%s_%s", prefix, readmeTemplate))
-		if err := writeTemplate(readmePath, outputPath, true, group); err != nil {
+		if err := writeTemplate(readmePath, outputPath, "markdown", group); err != nil {
```
jamiehannaford: instead of adding a new arg, could we compute this by reading the path extension?

cblecker: `OWNERS_ALIASES` doesn’t have an extension :(

PR #1857, opened 2018-02-27
Topology aware dynamic provisioning design

msau42:
This design is for 1.11 work to support dynamic provisioning.

Update volume topology design doc with:
* Defining terms used to describe topology
* Dynamic provisioning new APIs
* In-tree and external provisioning interface changes
* Scheduler implementation details
* Examples for local, zonal, and multi-zonal volumes

/hold

### DIFF
--- a/contributors/design-proposals/storage/volume-topology-scheduling.md
+++ b/contributors/design-proposals/storage/volume-topology-scheduling.md
@@ -281,37 +281,44 @@ Today, volume binding occurs immediately once a PersistentVolumeClaim is
 created. In order for volume binding to take into account all of a pod's other scheduling
 constraints, volume binding must be delayed until a Pod is being scheduled.
 
-A new StorageClass field `VolumeBindingMode` will be added to control the volume
+A new StorageClass field `BindingMode` will be added to control the volume
 binding behavior.
 
 ```
 type StorageClass struct {
     ...
 
-    VolumeBindingMode *VolumeBindingMode
+    BindingMode *BindingMode
 }
 
-type VolumeBindingMode string
+type BindingMode string
 
 const (
-    VolumeBindingImmediate VolumeBindingMode = "Immediate"
-    VolumeBindingWaitForFirstConsumer VolumeBindingMode = "WaitForFirstConsumer"
+    BindingImmediate BindingMode = "Immediate"
+    BindingWaitForFirstConsumer BindingMode = "WaitForFirstConsumer"
 )
 ```
 
-`VolumeBindingImmediate`  is the default and current binding method.
+`BindingImmediate`  is the default and current binding method.
 
-This approach allows us to introduce the new binding behavior gradually and to
-be able to maintain backwards compatibility without deprecation of previous
-behavior. However, it has a few downsides:
+This approach allows us to:
+* Introduce the new binding behavior gradually.
+* Maintain backwards compatibility without deprecation of previous
+  behavior. Any automation that waits for PVCs to be bound before scheduling Pods
+  will not break.
+* Support scenarios where volume provisioning for globally-accessible volume
+  types could take a long time, where volume provisioning is a planned
+  event well in advance of workload deployment.
+
+However, it has a few downsides:
 * StorageClass will be required to get the new binding behavior, even if dynamic
-provisioning is not used (in the case of local storage).
+  provisioning is not used (in the case of local storage).
 * We have to maintain two different code paths for volume binding.
 * We will be depending on the storage admin to correctly configure the
-StorageClasses for the volume types that need the new binding behavior.
+  StorageClasses for the volume types that need the new binding behavior.
 * User experience can be confusing because PVCs could have different binding
-behavior depending on the StorageClass configuration.  We will mitigate this by
-adding a new PVC event to indicate if binding will follow the new behavior.
+  behavior depending on the StorageClass configuration.  We will mitigate this by
+  adding a new PVC event to indicate if binding will follow the new behavior.
 
 
 ## Dynamic Provisioning with Topology
@@ -335,7 +342,7 @@ kind: StorageClass
 metadata:
   name: standard
 provisioner: kubernetes.io/gce-pd
-volumeBindingMode: WaitForFirstConsumer
+bindingMode: WaitForFirstConsumer
 parameters:
   type: pd-standard
 ```
@@ -415,27 +422,30 @@ described below.
 type StorageClass struct {
     ...
 
-    AllowedTopologies *VolumeProvisioningTopologies
+    // Restrict the node topologies where volumes can be dynamically provisioned.
+    // Each volume plugin defines its own supported topology specifications.
+    // Each entry in AllowedTopologies is ORed.
+    AllowedTopologies *[]TopologySelector
 }
 
-type VolumeProvisioningTopologies struct {
-    // Volume must be provisioned from one of the required topologies.
-    // These VolumeTopology requirements are ORed.
-    RequiredTopologies []VolumeTopology
+type TopologySelector struct {
+    // Topology must meet all of the TopologySelectorLabelRequirements
+    // These requirements are ANDed.
+    MatchLabelExpressions []TopologySelectorLabelRequirement
 }
 
-type VolumeTopology struct{
-    // The provisioned volume must support all the specified node labels.
-    // Each entry in the map is ANDed.
-    MatchNodeLabels map[string]string
+// Topology requirement expressed as Node labels.
+type TopologySelectorLabelRequirement struct{
+    // Topology label key
+    Key string
+    // Topology must match at least one of the label Values for the given label Key.
+    // Each entry in Values is ORed.
+    Values []string
 }
 ```
 
 A nil value means there are no topology restrictions. A scheduler predicate
 will evaluate a non-nil value when considering dynamic provisioning for a node.
-When evaluating the allowed topology against a node, the node needs to have
-labels matching everything specified in a VolumeTopology for one of the
-RequiredTopologies.
 
 A StorageClass update controller can be provided to convert existing topology
 restrictions specified under plugin-specific parameters to the new AllowedTopologies
@@ -459,12 +469,13 @@ Kubernetes will leave validation and enforcement of the AllowedTopologies conten
 to the provisioner.
 
 ##### Alternatives
-VolumeNodeAffinity is not reused here because the provisioning operation requires
+A new restricted TopologySelector is used here instead of reusing
+VolumeNodeAffinity because the provisioning operation requires
 allowed topologies to be explicitly enumerated, while NodeAffinity and
 NodeSelectors allow for non-explicit expressions of topology values (i.e.,
-operators NotIn, Exists, DoesNotExist, Gt, Lt). The provisioning operation also
-does not operate against Nodes (unlike pod scheduling), so NodeAffinity cannot
-be easily evaluated by the provisioner.
+operators NotIn, Exists, DoesNotExist, Gt, Lt). It would be difficult for
+provisioners to evaluate all the expressions without having to enumerate all the
+Nodes in the cluster.
 
 Another alternative is to expand ResourceQuota to support topology constraints.
 However, ResourceQuota is currently only evaluated during admission, and not
@@ -477,37 +488,52 @@ us-central1-b.
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
-  Name: zonal-class
+  name: zonal-class
 provisioner: kubernetes.io/gce-pd
 parameters:
   type: pd-standard
 allowedTopologies:
-  requiredToplogies:
-  - matchNodeLabels:
-      failure-domain.beta.kubernetes.io/zone: us-central1-a
-  - matchNodeLabels:
-      failure-domain.beta.kubernetes.io/zone: us-central1-b
+- matchLabelExpressions:
+  - key: failure-domain.beta.kubernetes.io/zone
+    values: us-central1-a, us-central1-b
 ```
 
 #### Multi-Zonal Example
 This example restricts the volume's primary and failover zones
-to us-central1-a and us-central1-b. Topologies that are incompatible with the
-storage provider parameters will be enforced by the provisioner.
+to us-central1-a, us-central1-b and us-central1-c. The regional PD
+provisioner will pick two out of the three zones to provision in.
 ```
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
-  Name: multi-zonal-class
+  name: multi-zonal-class
 provisioner: kubernetes.io/gce-pd
 parameters:
   type: pd-standard
   replication-type: regional-pd
 allowedTopologies:
-  requiredToplogies:
-  - matchNodeLabels:
-      failure-domain.beta.kubernetes.io/zone: us-central1-a
-  - matchNodeLabels:
-      failure-domain.beta.kubernetes.io/zone: us-central1-b
+- matchLabelExpressions:
+  - key: failure-domain.beta.kubernetes.io/zone
+    values: us-central1-a, us-central1-b, us-central1-c
+```
+
+Topologies that are incompatible with the storage provider parameters
+will be enforced by the provisioner. For example, dynamic provisioning
+of regional PDs will fail if provisioning is restricted to fewer than
+two zones in all regions. This configuration will cause provisioning to fail:
+```
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: multi-zonal-class
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-standard
+  replication-type: regional-pd
+allowedTopologies:
+- matchLabelExpressions:
+  - key: failure-domain.beta.kubernetes.io/zone
+    values: us-central1-a
 ```
 
 #### Multi Label Example
@@ -522,20 +548,20 @@ have the following labels:
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
-  Name: something-fancy
+  name: something-fancy
 provisioner: rack-based-provisioner
 parameters:
 allowedTopologies:
-  requiredTopologies:
-  - matchNodeLabels:
-      zone: us-central1-a
-      rack: rack1
-  - matchNodeLabels:
-      zone: us-central1-b
-      rack: rack1
-  - matchNodeLabels:
-      zone: us-central1-b
-      rack: rack2
+- matchLabelExpressions:
+  - key: zone
+    values: us-central1-a
+  - key: rack
+    values: rack1
+- matchLabelExpressions:
+  - key: zone
+    values: us-central1-b
+  - key: rack
+    values: rack1, rack2
 ```
 
 ## API Upgrades
@@ -685,7 +711,7 @@ metadata:
 ```
 
 ## Feature Gates
-PersistentVolume.NodeAffinity and StorageClas.VolumeBindingMode fields will be
+PersistentVolume.NodeAffinity and StorageClas.BindingMode fields will be
 controlled by the VolumeScheduling feature gate, and must be configured in the
 kube-scheduler, kube-controller-manager, and all kubelets.
 
@@ -1032,7 +1058,7 @@ release time frame, and aligns better with the current Kubernetes architecture.
 
 #### Unsupported Use Cases
 The following use cases will not be supported for PVCs with a StorageClass with
-VolumeBindingWaitForFirstConsumer:
+BindingWaitForFirstConsumer:
 * Directly setting Pod.Spec.NodeName
 * DaemonSets
 
--- a/contributors/design-proposals/storage/volume-topology-scheduling.md
+++ b/contributors/design-proposals/storage/volume-topology-scheduling.md
@@ -1,24 +1,32 @@
 # Volume Topology-aware Scheduling
 
-Authors: @msau42
+Authors: @msau42, @lichuqiang
 
 This document presents a detailed design for making the default Kubernetes
 scheduler aware of volume topology constraints, and making the
 PersistentVolumeClaim (PVC) binding aware of scheduling decisions.
 
+## Definitions
+* Topology: Rules to describe accessibility of an object with respect to
+  location in a cluster.
+* Domain: A grouping of locations within a cluster. For example, 'node1',
+  'rack10', 'zone5'.
+* Topology Key: A description of a general class of domains. For example,
+  'node', 'rack', 'zone'.
+* Hierarchical domain: Domain that can be fully encompassed in a larger domain.
+  For example, the 'zone1' domain can be fully encompassed in the 'region1'
+  domain.
+* Failover domain: A domain that a workload intends to run in at a later time.
 
 ## Goals
-* Allow a Pod to request one or more topology-constrained Persistent
-Volumes (PV) that are compatible with the Pod's other scheduling
-constraints, such as resource requirements and affinity/anti-affinity
-policies.
-* Support arbitrary PV topology constraints (i.e. node,
-rack, zone, foo, bar).
-* Support topology constraints for statically created PVs and dynamically
-provisioned PVs.
+* Allow a Pod to request one or more Persistent Volumes (PV) with topology that
+  are compatible with the Pod's other scheduling constraints, such as resource
+  requirements and affinity/anti-affinity policies.
+* Support arbitrary PV topology domains (i.e. node, rack, zone, foo, bar).
+* Support topology for statically created PVs and dynamically provisioned PVs.
 * No scheduling latency performance regression for Pods that do not use
-topology-constrained PVs.
-
+  PVs with topology.
+* Allow administrators to restrict allowed topologies per StorageClass.
 
 ## Non Goals
 * Fitting a pod after the initial PVC binding has been completed.
@@ -36,13 +44,34 @@ operator to schedule them together.  Another alternative is to merge the two
 pods into one.
     * For two+ pods non-simultaneously sharing a PVC, this scenario could be
 handled by pod priorities and preemption.
+* Provisioning multi-domain volumes where all the domains will be able to run
+  the workload. For example, provisioning a multi-zonal volume and making sure
+  the pod can run in all zones.
+    * Scheduler cannot make decisions based off of future resource requirements,
+      especially if those resources can fluctuate over time. For applications that
+      use such multi-domain storage, the best practice is to either:
+        * Configure cluster autoscaling with enough resources to accommodate
+          failing over the workload to any of the other failover domains.
+        * Manually configure and overprovision the failover domains to
+          accommodate the resource requirements of the workload.
+* Scheduler supporting volume topologies that are independent of the node's
+  topologies.
+    * The Kubernetes scheduler only handles topologies with respect to the
+      workload and the nodes it runs on. If a storage system is deployed on an
+      independent topology, it will be up to provisioner to correctly spread the
+      volumes for a workload. This could be facilitated as a separate feature
+      by:
+        * Passing the Pod's OwnerRef to the provisioner, and the provisioner
+          spreading volumes for Pods with the same OwnerRef
+        * Adding Volume Anti-Affinity policies, and passing those to the
+          provisioner.
 
 
 ## Problem
 Volumes can have topology constraints that restrict the set of nodes that the
 volume can be accessed on.  For example, a GCE PD can only be accessed from a
 single zone, and a local disk can only be accessed from a single node.  In the
-future, there could be other topology constraints, such as rack or region.
+future, there could be other topology domains, such as rack or region.
 
 A pod that uses such a volume must be scheduled to a node that fits within the
 volume’s topology constraints.  In addition, a pod can have further constraints
@@ -70,16 +99,21 @@ binding happens without considering if multiple PVCs are related, it is very lik
 for the two PVCs to be bound to local disks on different nodes, making the pod
 unschedulable.
 * For multizone clusters and deployments requesting multiple dynamically provisioned
-zonal PVs, each PVC Is provisioned independently, and is likely to provision each PV
-In different zones, making the pod unschedulable.
+zonal PVs, each PVC is provisioned independently, and is likely to provision each PV
+in different zones, making the pod unschedulable.
 
 To solve the issue of initial volume binding and provisioning causing an impossible
 pod placement, volume binding and provisioning should be more tightly coupled with
 pod scheduling.
 
 
-## New Volume Topology Specification
-To specify a volume's topology constraints in Kubernetes, the PersistentVolume
+## Volume Topology Specification
+First, volumes need a way to express topology constraints against nodes. Today, it
+is done for zonal volumes by having explicit logic to process zone labels on the
+PersistentVolume. However, this is not easily extendable for volumes with other
+topology keys.
+
+Instead, to support a generic specification, the PersistentVolume
 object will be extended with a new NodeAffinity field that specifies the
 constraints.  It will closely mirror the existing NodeAffinity type used by
 Pods, but we will use a new type so that we will not be bound by existing and
@@ -107,18 +141,27 @@ weights, but will not be included in the initial implementation.
 
 The advantages of this NodeAffinity field vs the existing method of using zone labels
 on the PV are:
-* We don't need to expose first-class labels for every topology domain.
-* Implementation does not need to be updated every time a new topology domain
+* We don't need to expose first-class labels for every topology key.
+* Implementation does not need to be updated every time a new topology key
   is added to the cluster.
 * NodeSelector is able to express more complex topology with ANDs and ORs.
+* NodeAffinity aligns with how topology is represented with other Kubernetes
+  resources.
 
 Some downsides include:
 * You can have a proliferation of Node labels if you are running many different
   kinds of volume plugins, each with their own topology labeling scheme.
+* The NodeSelector is more expressive than what most storage providers will
+  need. Most storage providers only need a single topology key with
+  one or more domains.  Non-hierarchical domains may present implementation
+  challenges, and it will be difficult to express all the functionality
+  of a NodeSelector in a non-Kubernetes specification like CSI.
 
 
 ### Example PVs with NodeAffinity
 #### Local Volume
+In this example, the volume can only be accessed from nodes that have the
+label key `kubernetes.io/hostname` and label value `node-1`.
 ```
 apiVersion: v1
 kind: PersistentVolume
@@ -141,6 +184,9 @@ spec:
 ```
 
 #### Zonal Volume
+In this example, the volume can only be accessed from nodes that have the
+label key `failure-domain.beta.kubernetes.io/zone` and label value
+`us-central1-a`.
 ```
 apiVersion: v1
 kind: PersistentVolume
@@ -164,6 +210,9 @@ spec:
 ```
 
 #### Multi-Zonal Volume
+In this example, the volume can only be accessed from nodes that have the
+label key `failure-domain.beta.kubernetes.io/zone` and label value
+`us-central1-a` OR `us-central1-b`.
 ```
 apiVersion: v1
 kind: PersistentVolume
@@ -194,12 +243,11 @@ and region labels.  This will handle newly created PV objects.
 
 Existing PV objects will have to be upgraded to use the new NodeAffinity field.
 This does not have to occur instantaneously, and can be updated within the
-deprecation period.
+deprecation period.  This can be done through a new PV update controller that
+will take existing zone labels on PVs and translate that into PV.NodeAffinity.
 
-TODO: This can be done through one of the following methods:
-- Manual updates/scripts
-- cluster/update-storage-objects.sh?
-- A new PV update controller
+An admission controller can validate that the labels and PV.NodeAffinity for
+this existing zonal volume types are kept in sync.
 
 ### Bound PVC Enforcement
 For PVCs that are already bound to a PV with NodeAffinity, enforcement is
@@ -225,27 +273,16 @@ Both binding decisions of:
 will be considered by the scheduler, so that all of a Pod's scheduling
 constraints can be evaluated at once.
 
-The rest of this document describes the detailed design for implementing this
-new volume binding behavior.
+The detailed design for implementing this new volume binding behavior will be
+described later in the scheduler integration section.
 
+## Delayed Volume Binding
+Today, volume binding occurs immediately once a PersistentVolumeClaim is
+created. In order for volume binding to take into account all of a pod's other scheduling
+constraints, volume binding must be delayed until a Pod is being scheduled.
 
-## New Volume Binding Design
-The design can be broken up into a few areas:
-* User-facing API to invoke new behavior
-* Integrating PV binding with pod scheduling
-* Binding multiple PVCs as a single transaction
-* Recovery from kubelet rejection of pod
-* Making dynamic provisioning topology-aware
-
-For the alpha phase, only the user-facing API and PV binding and scheduler
-integration are necessary.  The remaining areas can be handled in beta and GA
-phases.
-
-### User-facing API
-In alpha, this feature is controlled by a feature gate, VolumeScheduling, and
-must be configured in the kube-scheduler and kube-controller-manager.
-
-A new StorageClass field will be added to control the volume binding behavior.
+A new StorageClass field `VolumeBindingMode` will be added to control the volume
+binding behavior.
 
 ```
 type StorageClass struct {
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,329 @@ const (
...
+
+    Status *StorageClassStatus
+}
+
+type StorageClassStatus struct {
+    Capacity *StorageClassCapacity
+}
+
+type StorageClassCapacity struct {
+    TopologyKey string
```
lichuqiang: Still I'm concerned about the Topology key in StorageClassCapacity, if we place it in the status field, and leave it to provisioners to specify, then for edge-case that:
 
 When scheduler find a storage with StorageClassCapacity empty, it is hard to tell whether the storage is indeed without capacity limit, or it's just of a state that capacity has not been initialized by provisioners yet.

 If we move the Topology key out of status field, and leave it to admin, it is thought to be fulfilled on StorageClass creation.
 And the edge-case above could be avoid.

msau42: Then it will open up a new mode for failure, where administrator specifies the wrong topologyKey and the provisioner values won't match with it.

If a pod gets scheduled before provisioners have reported their capacity, then:
* Provisioning could fail.  In that case we should retry scheduling
* Provisioning could succeed.  Then we will not have marked the PVC with provisioned capacity information, so our calculations for used capacity will be incorrect.  Is there a way we can handle this?  Ie, have a StorageClass modify handler, when StorageClassCapacity is initialized, then initialize the capacity cache?

lichuqiang: Oh, I would prefer the scheduler could recognize the StorageClass at the first beginning, but not manage things by trying Provisioning. That makes things complex and may have effect on performance.

If you are worried that admin may specify a wrong key, then a simple flag to imply whether the StorageClass is of  capacity limit may also work.
I think the admin should at least know the StorageClass he creates is of  capacity limit or not.

msau42: How will the scheduler know that all the provisioners have initialized and reported capacity? (in the case of local volume provisioners)

msau42: Nm, in that case, once the topologyKey is specified, then capacity is zero until otherwise updated.

lichuqiang: > How will the scheduler know that all the provisioners have initialized and reported capacity? (in the case of local volume provisioners)

I think it doesn't need to know, what the scheduler cares:
1. Whether the StorageClass is of capacity limit
2. If so, whether it's capacity could satisfy the requirement.

No matter it's capacity smaller than requirement, or it's in state of uninitialized, it's both in state of unavailable for scheduler.

msau42:
TODO: Add section about predicate ecache invalidation and active pod queues

redbaron:
Failure domain for storage in general case can be different from failure domain for compute. Lets say in bare-metal setup, there can be 2 SAN storage, each available to half of compute nodes, while compute nodes themselves are split into failure domains per-rack.

There is a hack for Local volumes in current kubernetes version, where the fact that storage failure domain is per node is special cased. Do  think  suggestion to add first class storage-specific failure domains is something which belongs to this proposal?

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,329 @@ const (
...
+      values:
+      - us-central1-a
+      - us-central1-b
+```
+
+### Storage Provider Capacity Reporting
+Some volume types have capacity limitations per topology domain. For example,
+local volumes have limited capacity per node. Provisioners for these volume
+types can report these capacity limitations through a new StorageClass status
+field.
```
verult: I've been warned before against having constantly changing metrics (like resource usage) be placed as an API status, and one argument is that it could easily overwhelm the API server. Maybe the capacity information should be plumbed to the scheduler through existing resource usage monitoring pipelines.

Might be good to get a review from sig-instrumentation?

msau42: The capacity reported by the provisioner should be total capacity, not available capacity.  I don't anticipate total capacity changing very frequently.

The challenge with existing resource usage tracking is that it is done on a per node level, whereas volume resource limits may not necessarily be at the node boundary.

verult: There are plans to eventually plumb PVC usage stats through Metrics Server, as part of the work to get these stats exposed to kubectl. I believe the scheduler also talks to the Metrics Server.

Currently there's a limitation that PVC usage stats can only be measured when the underlying volume is mounted in a pod, but maybe we can expand this feature and add the capability of having volume plugins report these stats as well.

@jingxu97

msau42: I believe for PVC usage stats, it's still tied to a Pod that is scheduled to a specific node.  That way you can hit all the kubelet metrics endpoints.  For this design to handle getting capacity from metrics, we would need to collect all the metrics from a central entity, and not per node to handle volumes with topologies spanning more than a single node.

I am unaware of the scheduler talking to the metrics server.  My understanding is it uses resource capacity reported in the Node object.  @bsalamat could you clarify?

But besides the initial StorageClass creation, I don't anticipate reported capacity to be updated very frequently.

lichuqiang: Can we find the details of the feature somewhere?

msau42: Here, we have also have the challenge that external provisioners need to report their capacity somewhere, whether on an API object (like proposed), or to some central server.

lichuqiang: Yeah, as I'm also focusing on design of local storage provisioning, I wonder what the way to talk to the metrics server looks like.

I guess users are more likely to look forward to things they are familiar

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,329 @@ const (
...
+StorageClassQuota to restrict capacity usage for each user.
+
+
+## Feature Gates
+PersistentVolume.NodeAffinity and StorageClas.VolumeBindingMode fields will be
+controlled by the VolumeScheduling feature gate, and must be configured in the
+kube-scheduler, kube-controller-manager, and all kubelets.
+
+StorageClass.AllowedTopology and StorageClassCapacity fields will be controlled
+by the DynamicProvisioningScheduling feature gate, and must be configured in the
```
lichuqiang: Seem the two are not completely independent, maybe we should control both binding and provisioning with a same Feature Gate?

If we enable `VolumeScheduling` and disable `DynamicProvisioningScheduling`, then what's the expected behavior in pv controller? for PVCs of mode `WaitForFirstConsumer`, skip binding but exec provisioning directly?

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: I discussed a bit with @verult, and we agreed that going the metrics server approach would require a lot of extra plumbing that doesn't exist today.  So we're still leaning towards the current approach of reporting total (not used) capacity in the StorageClass and calculating and caching used from provisioned PVCs.

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: Challenge is static local PVs is beta, and dynamic provisioning will be alpha, so we can't have one feature gate control both.  I see VolumeScheduling as a superset, so if that is disabled, then the delayed dynamic provisioning is also disabled.  Once delayed dynamic provisioning goes to beta, we can combine the two feature gates.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,329 @@ const (
...
+
+```
+type StorageClass struct {
+    ...
+
+    AllowedTopology *VolumeProvisioningTopology
+}
+
+type VolumeProvisioningTopology struct {
+    Required map[string]VolumeTopologyValues
```
msau42: @saad-ali found a use case that this cannot represent:
An administrator wants to restrict provisioning to either "zone A, rack 1" or "zone B, rack 1".

To fix this, we could go back to using a restricted node selector (only support the IN operator), or follow the CSI [proposed](https://github.com/container-storage-interface/spec/pull/188#issuecomment-366425374) representation which sort of does the reverse of the node selector: OR, then AND.

A restricted NodeSelector would be more familiar to Kubernetes users, so I'm leaning towards that approach.   As long as we restrict the operators to just IN, then it can be translated internally into the CSI representation.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -404,26 +753,30 @@ AssumePVCs(pod *v1.Pod, node *v1.Node) (pvcBindingRequired bool, err error)
...
 ```
-BindUnboundPVCs(pod *v1.Pod, node *v1.Node) (err error)
+BindVolumes(pod *v1.Pod, node *v1.Node) (err error)
 ```
 1. For static PV binding:
     1. Prebind the PV by updating the `PersistentVolume.ClaimRef` field.
     2. If the prebind fails, revert the cache updates.
 2. For in-tree and external dynamic provisioning:
-    1. Set `annStorageProvisioner` on the PVC.
+    1. Set `annScheduledNode` on the PVC.
```
lichuqiang: annSelectedNode

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: I'm flip flopping again.  A node selector specifies a subset off of a set of labels on the node object.  But in this case, we are not selecting from a set of labels on an existing object.  In this case, we want to limit the topology of a PV object that will be created in the future.  So I think following the CSI representation as a list of maps makes more sense, rather then specifying a selector on nothing.

lichuqiang: +1

verult: +1. Another semantic difference is: VolumeTopology is used to select **a set of nodes** at which a PV is available, whereas NodeSelector is used to select **a single node**.

liggitt: > NodeSelector is used to select a single node

that's a little surprising... selectors typically select a set of things

verult: @liggitt you are right, I was using a wrong frame of thinking, so I take that back.

Here's another point, similar to what @msau42  said, that helped me understand the difference: ultimately the scheduler needs a set of nodes to choose from; a NodeSelector **directly** specifies that set, whereas VolumeTopology **indirectly** specifies it by providing a super set which is then refined further by the volume plugin. Please correct me if this is the wrong understanding.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,329 @@ const (
...
+  name: my-pod
+spec:
+  containers:
+  ...
+  volumes:
+  - name: my-vol
+    persistentVolumeClaim:
+      claimName: my-pvc
+```
+5. Scheduler picks a node that can satisfy the Pod and passes it to the
```
kmova: @msau42 - Can the scheduler pass the "suggested" nodes where it can be scheduled? What if the node that is picked by the scheduler can't be used by the dynamic provisioner to create a PV. 

In the next step (6), how about the dynamic provisioner create a PV and set nodeAffinity (list of nodes, racks, zones) where it can be accessed and the Pod get scheduled on one of those nodes. This could tie into the way, static PVs provisioned and associated with nodeAffinity works.

lichuqiang: > What if the node that is picked by the scheduler can't be used by the dynamic provisioner to create a PV.

I think the topology and capacity limitation of the provisioner should have been specified in the Storageclass.

> In the next step (6), how about the dynamic provisioner create a PV and set nodeAffinity (list of nodes, racks, zones) where it can be accessed and the Pod get scheduled on one of those nodes.

This way, there is still a chance that the provisioned pv is conflict to the scheduled node.
I think the main purpose of the feature is to reconcile pod scheduling with volume binding/provisioning,

msau42: The general scheduler design requires that we have all the scheduler predicates and priorities we need so that the scheduler can make the right decision.  If there is some timing issue which causes the scheduler's view to be out of date, provisioning needs to fail, and the scheduler has to retry.  Eventually, the scheduler's view of the world has to be corrected and it will make the right choice.

If provisioning doesn't fail and it ends up provisioning in a topology domain that the scheduler didn't recommend, in the best case, the scheduler can still find a node that can satisfy the PV.  In the worst case, the pod will fail scheduling and stay pending.

Step 6 will work exactly like as you mentioned.  The PV provisioned must also specify the nodeAffinity, and on pod restarts, it will follow the same bound PVC flow.

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: I discussed with @saad-ali on this.  Our consensus is that it's better for topologyKey to be reported by the provisioner.  Depending on the user to specify the topologyKey, or even a flag, opens up the possibility that they could incorrectly set it.  And they don't have any choice in what to set as the value.  For local storage, it must always be set to "kubernetes.io/hostname", there is no other value that will work.

In the worst case, there will be a period of time during cluster initialization where the provisioners may not have reported capacity yet.  The consequences are not fatal.  Provisioning could fail until the provisioners are up and running (which is expected behavior in general).  Provisioning could succeed, and in that case we'll have to resync the capacity cache later.  We could try to mitigate this scenario, by recommending that provisioners do not process any provisioning requests until they have fully initialized (including reporting capacity).

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -383,19 +732,19 @@ PrioritizeUnboundPVCs(pod *v1.Pod, filteredNodes HostPriorityList) (rankedNodes
...
-For the alpha phase, the AssumePVCs function will be directly called by the
+For the alpha phase, the AssumeVolumes function will be directly called by the
 scheduler.  We’ll consider creating a generic scheduler interface in a
 subsequent phase.
 
 ```
-AssumePVCs(pod *v1.Pod, node *v1.Node) (pvcBindingRequired bool, err error)
+AssumeVolumes(pod *v1.pod, node *v1.node) (pvcbindingrequired bool, err error)
 ```
 1. If all the Pod’s PVCs are bound, return false.
```
orainxiong: I think the rough logic of AssumePodVolumes would be : 

- Could returns true if all volumes are fully bound
- Could returns false if any volume binding operation needs to be done afterward, and then the pod will be enqueued into _BindQueue_

orainxiong:
I have modified _volume-topology-scheduling.png_ a little.
[https://www.processon.com/view/link/59e6fc25e4b040dc8507bfd1](url)
Hope it helps.

contributors/design-proposals/storage/volume-topology-scheduling.md
kmova: Thanks @msau42 @laishiquan. Having the topology information to the StorageClass and using that as a scheduler predicate might just work.

msau42: Can you clarify the use case you are trying to support?  This proposal adds 2 ways to influence scheduling for dynamic provisioning:

1. Admin can restrict topology via AllowedTopologies
2. Provisioner can restrict topology via StorageClassCapacity

msau42:
Updated AllowedTopologies to align with how CSI specifies topology, and should have addressed all the review comments

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,368 @@ const (
...
+    // (free and used) for provisioning at each value of the
+    // TopologyKey
+    Capacities map[string]resource.Quantity
+}
+```
+
+If no StorageClassCapacity is specified, then there are no capacity
+limitations.
+
+Capacity reporting is limited to a single topology key. `TopologyKey`
```
msau42: I think it doesn't have to be restricted to a single key if the Capacities map key is a VolumeTopology type.  Not sure if that is expressible in the API though, I haven't seen any non-string map keys in the API.

If we can't do the above, then Capacities could be an array, but then lookups by the scheduler will be inefficient.  We need to figure out how important capacity reporting is for multi-dimensional storage systems.  The examples I can think of are only hypothetical at this point.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -328,30 +693,54 @@ avoid these error conditions are to:
...
+* allowedTopologies, when it is set in the StorageClass
+
+If selectedNode is set, the provisioner should get its appropriate topology
+labels from the Node object, and provision a volume based on those topology
+values. In the common use case for a volume supporting a single topology domain,
+if nodeName is set, then allowedTopologies can be ignored.
+
+In-tree provisioners:
+```
+Provision(selectedNode *string, allowedTopologies *storagev1.VolumeProvisioningTopology) (*v1.PersistentVolume, error)
```
verult: What if the entire Node spec is passed instead of just the node name? A topology-aware volume plugin has to fetch the node object anyway in order to get the node's topology info.

Alternatively - should there be a dedicated NodeTopology field inside Node spec? Then here we could just pass that field, and volume plugins encode their topology information in this field rather than labels.

A related issue: https://github.com/kubernetes/kubernetes/issues/61906 (TL;DR if we had this field, it could be passed to applications through downwardAPI, making it easy for applications to obtain topology info).

msau42: Having a dedicated topology field in the node object could be useful, but I think it has broader implications than just for storage, so would need to be considered more holistically.

Passing a Node object sounds ok to me. It means that the PV controller also needs permissions to get Node objects.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,368 @@ const (
...
+  provisioned in a domain that cannot run the Pod since it doesn't take any
+  scheduler input.
+* For volumes that span multiple domains, the AllowedTopologies can restrict those
+  additional domains.  However, special care must be taken to avoid specifying
+  conflicting topology constraints in the Pod. For example, the administrator could
+  restrict a multi-zonal volume to zones 'zone1' and 'zone2', but the Pod could have
+  constraints that restrict it to 'zone1' and 'zone3'.  If 'zone1'
+  fails, the Pod cannot be scheduled to the intended failover zone.
+
+Kubernetes will leave validation and enforcement of the AllowedTopologies content up
```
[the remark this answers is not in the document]
msau42: Labels are user defined and can be anything.  And each volume plugin may have a different idea of what topology labels it needs to operate on.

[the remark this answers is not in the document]
verult: But in the end these are directly compared against Node labels for scheduling purposes, so they should at least satisfy the Kubernetes label format. I'd imagine the reasoning is similar to why NodeSelector and NodeAffinity are validated against the label format as well.

contributors/design-proposals/storage/volume-topology-scheduling.md
lichuqiang: This would indeed bring inconvenience to provisioners and the scheduler,
I would suggest we do not introduce this before we can see practical use case of multi-dimensional  capacity reporting

contributors/design-proposals/storage/volume-topology-scheduling.md
lichuqiang: @msau42 As we discussed before, we need a way to calculate the capacity used by released PVs(i.e. provisioned PVs whose claims got deleted then)

so here we'll also need to pass `AnnProvisionedTopology`, and require the developers of provisioners to add the annotation onto PVs.

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: Do the nodes where the StoragePool exist map to the volume io storage pods?  Also, can a node serve multiple storage pools?

I'm trying to see if you could use the storage class capacity reporting instead of having the admin set allowedTopologies, because that will open up the possibility that the admin will configure it incorrectly.  But the capacity reporting scheme may not work if you can have a node serve multiple storage pools because you cannot have multiple label values for a single key.

msau42: One more question. Is it a requirement that the application pod is on the same node where the storagepool is in?

msau42: Hm IIUC, you will have one storageclass per storage pool.  In that case, then capacity reporting won't help, unless the provisioner does something like report some fake capacity for each node in the storagepool (and by default 0 capacity for the rest)

kmova: A node can server multiple StoragePools depending on the available storage and other resources on the node. 

Since the communication is like this: application pod -> volume io controller pod (stateless) -> storagepool (storage io pod), it is not a hard requirement to have the application pod on the same node as the storagepool, but it is a preferred way. 

@msau42 it is a good point though about the possibility that the admin may configure incorrectly, to circumvent this - we could use an operator to handle the updates to the allowedToplogies. 

Is the the allowedTopologies an editable parameter on the StorageClass object?

msau42: We could make it mutable. I don't see a reason not to.  Do you see a use case where the admin would want to configure some policy that is different from the operator though?  Maybe they would still want to allow more nodes outside of the storagepool because they are node resource constrained.  

It sounds like for your use case, topology preferences would be better than requirements.

msau42: Originally, the allowedTopologies was intended to be set by the user, hence, having it in the "spec".  And then capacity is intended to be reported by the system, therefore it's in the "status".  The use case you bring up doesn't fit quite nicely into this model. I need to think about it more

msau42: @kmova I pondered some more about your use case and I'm not actually sure you need to influence topology at all during provisioning.  Let me know if this makes any sense:

You have one storageclass per storagepool.  A user requests a specific storageclass ie a specific storagepool.  Since it's not a requirement that pods land on the same node where a storagepool spans, you should always have provisioning succeed even if it results in suboptimal placement.  After provisioning, you can specify your topology preferences in the PV.nodeaffinity (note, preferences are not supported today and would need to be added), and the scheduler can use the preferences to make the final node choice.

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: We can support multi-keys but still have efficient lookups by flattening the values into a single string separated by a delimiter.  Will update the spec shortly.

msau42: Also there is a limitation that a single storage pool should not be shared by multiple storage classes, otherwise you will have more capacity reported than is actually available.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,368 @@ const (
...
+metadata:
+  Name: zonal-class
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-standard
+allowedTopologies:
+  requiredToplogies:
+  - matchNodeLabels:
+      failure-domain.beta.kubernetes.io/zone: us-central1-a
+  - matchNodeLabels:
```
jsafrane: It took me a while to understand why there are two `matchNodeLabels` entries above. It could be useful to explicitly note somewhere that Kubernetes computes OR between matchNodeLabels and AND among individual labels inside.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,368 @@ const (
...
+will evaluate a non-nil value when considering dynamic provisioning for a node.
+When evaluating the allowed topology against a node, the node needs to have
+labels matching everything specified in a VolumeTopology for one of the
+RequiredTopologies.
+
+A StorageClass update controller can be provided to convert existing topology
+restrictions specified under plugin-specific parameters to the new AllowedTopologies
+field.
+
+The AllowedTopologies will also be provided to provisioners as a new field, detailed in
```
jsafrane: It would be nice to have a sentence or two why node selector was not used in StorageClass.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,368 @@ const (
...
+      node1: 100Gi
+      node2: 200Gi
+```
+
+TODO: should this be in the local volume spec instead?
+Some downsides specific to the local volume implementation:
+* StorageClass object size can be big for large clusters. Assuming that
+  each node entry in the Capacities map could take up around 100 bytes
+  (size of node name + resource.Quantity), a 1000 node cluster that uses
+  dynamic provisioning of local volumes could add 100Ki to the StorageClass
```
jsafrane: Just brainstorming: what if we move the StorageClassCapacity into its own object? For local storage there would be one StorageClass as it is today + every node would have its own StorageClassCapacity somehow linked to the StorageClass. Individual local storage provisioners would not conflict when updating their own StorageClassCapacity objects. On the other hand, multiple small objects would probably load API server more than one big one.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,368 @@ const (
...
+  seen such use cases so far.
+* There is no common mechanism to report capacity. Each provisioner has to
+  update their own fields in the StorageClass if necessary. CSI could potentially
+  address this.
+* There is a gap during cluster initialization where the StorageClass object
+  has been created, and the provisioner has not reported capacity yet. In this
+  scenario, the provisioner should fail provisioning until it has reported its
+  capacity and not process any PVCs that have not been updated with the chosen
+  topologyKey value.
+
```
jsafrane: Another downsides:

- When the storage is shared among several Kubernetes clusters *or* with something outside Kubernetes, Kubernetes cannot reliably compute amount of used storage. It can compute only the storage the particular cluster consumes (i.e. PVs), but it won't see PVs of other clusters and/or users outside of Kubernetes. It is recommended that Kubernetes gets dedicated storage or storage quota so it can reliably compute amount of used/remaining storage just by counting PVs.

- The storage may be shared by several StorageClasses. Typically, Gluster allows admins to choose nr. of replicas of PVs in StorageClass. There may be two StorageClasses that use the same Gluster servers, however each of them creates PVs with different nr. of replicas. Kubernetes treats StorageClasses as independent on each other, however in this case PVs in one StorageClass must be counted to used space in a different StorageClass.

- Capacity does not report granularity of the storage. E.g. there may be 1 GiB of available space on Gluster in a particular datacenter and scheduler can schedule one pod with two 0.5 GiB PVCs there. Gluster allocates storage in 1GiB chunks, so it can't accommodate both PVCs and the pod becomes unschedulable until someone frees up some space in the datacenter.
  - This is IMO rare corner case when a storage is almost full.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,368 @@ const (
...
+...
+status:
+  capacity:
+    topologyKey: kubernetes.io/hostname
+    capacities:
+      node1: 100Gi
+      node2: 200Gi
+```
+
+TODO: should this be in the local volume spec instead?
```
jsafrane: How this can be in local PV spec? We shouldn't put capacity of a node into all local PVs that are the node.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -328,30 +693,54 @@ avoid these error conditions are to:
...
 to let it come through the scheduler path.
 
 Dynamic provisioning will also be skipped if
-VolumBindingWaitForFirstConsumer is set.  The scheduler will signal to
+VolumBindingWaitForFirstConsumer is set. The scheduler will signal to
 the PV controller to start dynamic provisioning by setting the
-`annStorageProvisioner` annotation in the PVC.
+`annSelectedNode` annotation in the PVC. If provisioning fails, the PV
+controller can signal back to the scheduler to retry dynamic provisioning by
+removing the `annSelectedNode` annotation.
```
jsafrane: "removing the `annSelectedNode` annotation" is not described below. Who removes it and when? And can scheduler pick the same node again or will it ensure that it picks a different node this time?

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: Updated in API comments

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -194,12 +244,11 @@ and region labels.  This will handle newly created PV objects.
 
 Existing PV objects will have to be upgraded to use the new NodeAffinity field.
 This does not have to occur instantaneously, and can be updated within the
-deprecation period.
+deprecation period.  This can be done through a new PV update controller that
```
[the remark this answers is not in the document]
msau42: Added API update section

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: added

contributors/design-proposals/storage/volume-topology-scheduling.md
msau42: done

contributors/design-proposals/storage/volume-topology-scheduling.md
jsafrane: Ack, let's leave it in StorageClass.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,502 @@ const (
...
+reported by the storage system, while the StorageClassQuota represents policies
+set by the cluster admin and does not take into account topology domain.
+
+For example, a storage system is configured to support 200Gi in zone1 for a
+StorageClass that is shared by multiple users.  An administrator can use
+StorageClassQuota to restrict capacity usage for each user.
+
+## API Upgrades
+
+### Zonal PV Upgrade
```
msau42: Add an example for multi-zonal PV upgrade because the zone label value format is special.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -141,6 +185,9 @@ spec:
 ```
 
 #### Zonal Volume
+In this example, the volume can only be accessed from nodes that have the
+label key `failure-domain.beta.kubernetes.io/zone' and label value
```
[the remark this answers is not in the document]
msau42: done thanks

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,565 @@ const (
...
+for dynamic provisioning.
+```
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: standard
+...
+status:
+  capacity:
+    topologyKey: zone,rack
```
msau42: Keys

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,565 @@ const (
...
+types can report these capacity limitations through a new StorageClass status
+field. It should be noted that provisioners only need to report capacity in the
+StorageClass if the single StorageClass encompasses multiple storage pools with
+different capacities across different topology domains.
+
+```
+type StorageClass struct {
+    ...
+
+    Status *StorageClassStatus
```
lichuqiang: generally speaking, status field should not be a pointer?

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,565 @@ const (
...
+    // value.
+    Capacities map[string]resource.Quantity
+}
+```
+
+If no StorageClassCapacity is specified, then there are no capacity
+limitations.  A missing map entry for a topology domain will be treated
+as zero capacity.
+
+A scheduler predicate will use the reported capacity to determine if dynamic
```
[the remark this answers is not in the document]
thockin: Also it will scale badly for large clusters.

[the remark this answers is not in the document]
thockin: Also it is insufficient to express whether a given volume can actually fit (e.g. could be fragmented).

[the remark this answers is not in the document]
msau42: This is similar to how a node today reports total CPU/memory.  It doesn't actually report used vs available, instead the scheduler keeps track of used based on all the Pods that have been scheduled to that node.  This way you avoid the latency issue of waiting for reported values to updated after they are allocated.

So we're trying to follow a similar concept here, except the challenge with volumes is that their topology is not always scoped to a single node, so it doesn't make sense to report as part of node capacity.  The reported capacity would only change if you change the capacity of the storage pool, but not when you allocate/free volumes from it.  I expect the former will happen less frequently than the latter, so the scaling limits should be similar to how we process cpu/memory capacity.

But this does have some limitations to be accurate, like you mentioned:
* The storage pool has to be dedicated to this StorageClass and cannot be shared with other StorageClasses
* Provisioning from the storage pool cannot be prone to fragmentation

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,565 @@ const (
...
+
+    Status *StorageClassStatus
+}
+
+type StorageClassStatus struct {
+    Capacity *StorageClassCapacity
+}
+
+type StorageClassCapacity struct {
+    // Minimum size of a provisioned volume
```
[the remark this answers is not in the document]
lichuqiang: As @jsafrane said, "some clusters allocates storage in 1GiB chunks", thus, if a user requested 1.5GiB in PVC, we are expected to reject it, even if we have enough capacity left.
Agree we might need to consider whether the capacity field is a good place to specify it.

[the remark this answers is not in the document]
lichuqiang: Well, I misunderstood its meaning here, and thought we need to reject it if a user requested a volume under the minimum size.

contributors/design-proposals/storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
msau42: Here's an alternative idea.  Considering that only local volumes plan to use the capacity reporting api at the moment, and there are some restrictions on how you deploy the storage pools to ensure that this capacity handling is actually accurate, we could try going with a CRD + scheduler extender approach.

The downsides are:
- Deployment is more complicated
- We can't include this into Kubernetes core.  The provisioner is already out of tree, maybe this not an issue.
- The predicate will become best-effort, due to the delay of seeing the PVC provisioning API update in the extender.  The fact that the scheduler round robins on equal priority nodes could help mitigate this, so that the same node is not chosen again for the next Pod.

Advantages are:
- Since it's a CRD specific to local volumes, the API could be simpler and more flexible because we don't need to make it generic to all volume topologies
- You potentially could support scenarios like storage pools being shared across multiple StorageClasses, since it doesn't need to be strictly tied to a StorageClass anymore
- You can encode special behavior/requirements of your specific storage system into the extender logic.
- Keeps StorageClass lean and we don't have to regret not putting a Spec/Status in the beginning

@lichuqiang @jsafrane @thockin wdyt?

contributors/design-proposals/storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
jsafrane: IMO it should be admin decision what "rounding" is acceptable and what's too much. Someone finds using 50GB for 10GB of data just fine, others find it waste of space.

CSI introduced `required_bytes` and `limit_bytes` and I like it, however we don't have capability to express that in PVC and in storage class, where both user and admin might define what's acceptable.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,565 @@ const (
 
 This approach allows us to introduce the new binding behavior gradually and to
 be able to maintain backwards compatibility without deprecation of previous
-behavior.  However, it has a few downsides:
+behavior. However, it has a few downsides:
 * StorageClass will be required to get the new binding behavior, even if dynamic
 provisioning is not used (in the case of local storage).
-* We have to maintain two different paths for volume binding.
+* We have to maintain two different code paths for volume binding.
 * We will be depending on the storage admin to correctly configure the
```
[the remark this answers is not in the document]
jsafrane: I don't have any special use case for very slow-to-bind volumes, it would just break existing behavior. Users (and automation) are used to their PVCs being bound relatively quickly and they will be confused.

Someone on SIG-storage said that in some storage backends provisioning can take hours and it's well planned operation, so users want to create PVCs and let them provision before they start a pod. @kangarlou, was that you?

contributors/design-proposals/storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
msau42: PVC already has the fields to support request and limit, we just don't use limit currently.  I'm still not quite sure about the use case of putting it in the storageclass.  I believe for the other resources, admins can define a limit range to restrict at admission.

contributors/design-proposals/storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
kangarlou: @jsafrane No, it wasn't me, but here is my two cents: 
Today for dynamic provisioners, binding happens as soon as the provisioning is done. For the NFS and iSCSI volumes that we manage on NetApp storage, provisioning happens very quickly. To keep it fast, our provisioner doesn't format the iSCSI volumes and relies on kubelet to perform this functionality during attach. This is not a concern for us, but I can see cases where provisioning takes a long time for some plugins/provisioners (e.g., formatting large drives at the time of volume creation, provisioning from snapshots). Having said that, I don't think we should maintain two code paths. As far as external automation is concerned, because provisioning can take arbitrarily long time (in the worst case, the provisioning can fail), any external tooling should inspect the PVC status and not rely on the immediate binding behavior. I agree with @jsafrane that some users prefer to provision their PVCs before they start the pods, but for provisioners that take a long time, I don't think the proposed changes affect the user experience much as such users see provisioning and binding as one step, not as two separate steps.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -328,30 +890,62 @@ avoid these error conditions are to:
 * Separate out volumes that the user prebinds from the volumes that are
 available for the system to choose from by StorageClass.
 
-#### PV Controller Changes
+### PV Controller Changes
 When the feature gate is enabled, the PV controller needs to skip binding
 unbound PVCs with VolumBindingWaitForFirstConsumer and no prebound PVs
 to let it come through the scheduler path.
 
 Dynamic provisioning will also be skipped if
```
kangarlou: Dynamic provisioning will be skipped or delayed until `annSelectedNode` annotation is set? Does this mean dynamic provisioners should look for this annotation for the storage classes with `VolumeBindingMode` set to `VolumeBindingWaitForFirstConsumer` before they can act on the PVCs of that type?

contributors/design-proposals/storage/volume-topology-scheduling.md
jsafrane: > I propose to instead probe the storage provider if needed: "Could you create a volume of size X and type Y?" This could be a "dry run" variant of the volume create api.

Scheduler is under watch for its performance. Currently, it does not call any external API. Calling number of probe/dry-run methods would slow everything down. We need a quick and mostly reliable check where to start provisioning first, hoping it succeeds in vast majority of cases.

msau42-tmp: Another problem with doing a call out is that the "provisioning check" and "actual provisioning" steps are not synchronous.  So the following two scenarios would not be accurate:

* Pod requests multiple PVCs.  You would need to query the storage system in a way so that the 1st query would temporarily subtract from the total when doing the 2nd query.
* Scheduler does the actual binding operations asynchronously, so you could have a query for the next Pod before the provisioning for the 1st Pod is triggered.  This is similar to the 1st scenario, just for different Pods.

That's why the scheduler today doesn't do callouts to get the Node resources, and it tracks the used capacity internally.  So if we want to do something similar for storage, then I think we should follow this similar model of storing total capacity in an object.  I agree that there are many internal details that influence what is actually provisionable, so if we had such an API, it would be up to the provisioner to report its capacity wrt to the user's request.  However, the limitation of the current capacity API is that the storage pool is contiguous, which doesn't sound to be the case with gluster (discrete disks).

msau42:
I squashed the PR into 3 commits:
* Design up the most recent review iteration
* Addressed feedback from the most recent review, changing allowedTopologies to use a new TopologySelector field instead of an array of maps
* Separated out the capacity reporting design, so we can discuss it separately from the rest

msau42:
I removed the capacity api section, and will look at addressing it separately.

So the design proposal here includes:
* Basic changes to scheduler and PV controller to delay dynamic provisioning and pass the selected node to the provisioner
* API changes to allow an administrator to restrict topologies in the storageclass (replacement for plugin-specific zone/zones parameters today)

@saad-ali @thockin @jsafrane PTAL

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -266,41 +304,565 @@ const (
...
+    // These VolumeTopology requirements are ORed.
+    RequiredTopologies []VolumeTopology
+}
+
+type VolumeTopology struct{
+    // The provisioned volume must support all the specified node labels.
+    // Each entry in the map is ANDed.
+    MatchNodeLabels map[string]string
+}
+```
```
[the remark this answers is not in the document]
verult: > ```
> StorageClass:
>  topologyConstraints:
>    - zone in [a, b]
>      rack in ["1", "2"]
> ```
Wouldn't this allow "zone=b,rack=1"? To entirely capture the original topology, you'd need something like:
```
StorageClass:
  topologyConstraints:
    - zone in [a]
      rack in ["1"]
    - zone in [b]
      rack in ["2"]
```

One inconvenience with the selector approach is there are multiple ways to specify certain kinds of topology. For example, the following 3 are equivalent:
```
StorageClass:
  topologyConstraints:
    - zone in [a]
      rack in ["1"]
    - zone in [b]
      rack in ["1"]
    - zone in [b]
      rack in ["2"]
```
```
StorageClass:
  topologyConstraints:
    - zone in [a,b]
      rack in ["1"]
    - zone in [b]
      rack in ["2"]
```
```
StorageClass:
  topologyConstraints:
    - zone in [a]
      rack in ["1"]
    - zone in [b]
      rack in ["1", "2"]
```

I could potentially see users being confused which form Kubernetes expects, when in reality either of them work. To me it seems better to have one layer of ORs instead of two.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -328,30 +890,62 @@ avoid these error conditions are to:
...
 
-#### PV Controller Changes
+### PV Controller Changes
 When the feature gate is enabled, the PV controller needs to skip binding
 unbound PVCs with VolumBindingWaitForFirstConsumer and no prebound PVs
 to let it come through the scheduler path.
 
 Dynamic provisioning will also be skipped if
-VolumBindingWaitForFirstConsumer is set.  The scheduler will signal to
+VolumBindingWaitForFirstConsumer is set. The scheduler will signal to
```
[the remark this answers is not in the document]
msau42: 1. In-tree provisioners will be coordinated by the PV controller.
2. The external-provisioner library in external-storage will be updated to also coordinate this.
3. The CSI external-provisioner will coordinate this.
4. If you don't use the external-provisioner library, then yes, you need to handle this in your own provisioner.

contributors/design-proposals/storage/volume-topology-scheduling.md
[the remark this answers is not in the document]
msau42: My updated example [here](https://github.com/kubernetes/community/pull/1857/files#diff-9d29da618fae1244e58d194d3d70e2adR554) correctly represents (one way) the example topology.  I think having multiple representations is fine as long as they evaluate the same, and it is consistent with how selectors are specified for other Kubernetes objects.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -225,82 +273,477 @@ Both binding decisions of:
...
+described below.
+
+```
+type StorageClass struct {
+    ...
+
+    // Restrict the node topologies where volumes can be dynamically provisioned.
+    // Each volume plugin defines its own supported topology specifications.
+    // Each entry in AllowedTopologies is ORed.
+    AllowedTopologies *[]TopologySelector
```
lichuqiang: Does it have to be a pointer? I think it fairly enough to make it an array, and just leave it empty if we don't wanna specify the field.

and I tried to make it a pointer, but when running `make update`, seems the update scripts still treat the field as array, and tried to range over it:
```
# k8s.io/kubernetes/vendor/k8s.io/api/storage/v1
vendor/k8s.io/api/storage/v1/generated.pb.go:169:8: invalid argument m.AllowedTopologies (type *[]TopologySelector) for len
vendor/k8s.io/api/storage/v1/generated.pb.go:170:17: cannot range over m.AllowedTopologies (type *[]TopologySelector)
vendor/k8s.io/api/storage/v1/generated.pb.go:348:8: invalid argument m.AllowedTopologies (type *[]TopologySelector) for len
vendor/k8s.io/api/storage/v1/generated.pb.go:349:15: cannot range over m.AllowedTopologies (type *[]TopologySelector)
```

So if we have to make it this way, we may need a fix on the update scripts.

msau42: Oops yes this does not need a pointer. I mistakenly left it from the last iteration when it was a struct

lichuqiang: so `[]*TopologySelector` ?

msau42: `[] TopologySelector`

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -1,24 +1,32 @@
...
-Volumes (PV) that are compatible with the Pod's other scheduling
-constraints, such as resource requirements and affinity/anti-affinity
-policies.
-* Support arbitrary PV topology constraints (i.e. node,
-rack, zone, foo, bar).
-* Support topology constraints for statically created PVs and dynamically
-provisioned PVs.
+* Allow a Pod to request one or more Persistent Volumes (PV) with topology that
+  are compatible with the Pod's other scheduling constraints, such as resource
+  requirements and affinity/anti-affinity policies.
```
saad-ali: nit: This still doesn't sound very clear to me.

Pods don't request anything. Pods specify things that users request. So, to me, this reads "Allow a user to to specify in a pod one or more PV with topology that is compatible with the pod's other constraints."

That still doesn't make sense to me. Do we prevent that now? Is that not possible now?

The following sound a lot more clear to me:
* `Allow the Kubernetes scheduler to influence where a volume is provisioned based on scheduling constraints on the pod referencing the volume, such as pod resource requirements and affinity/anti-affinity policies.`
* `Ensure volume topology is used as a constraint when scheduling a pod referencing one or more Persistent Volumes (PV) with topology.`

These may be implied, but I'd like to be very explicit about them.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -1,24 +1,32 @@
...
+* Support arbitrary PV topology domains (i.e. node, rack, zone, foo, bar).
```
saad-ali: Maybe expand this to explicitly state that we want to this without having to encode as first class objects in the k8s API.

contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -1,24 +1,32 @@
...
+* Support topology for statically created PVs and dynamically provisioned PVs.
```
saad-ali: nit: personal preference "pre-provisioned" instead of "statically created". The symmetry with dynamically provisioned is nicer =)

msau42:
This is unicorning too frequently.  I addressed all the comments here, and opened up a new PR https://github.com/kubernetes/community/pull/1857

/close

jheiss:
@msau42 Did you mean to refer to #2168 when closing this? (You referred to this same issue in your last comment.)

msau42:
Thanks, updated!

PR #1873, opened 2018-03-01
E2e local

pohly:
The example for e2e testing with a local cluster does not work as-is for two different reasons:
- missing --provider parameter
- incorrect quoting of ginkgo.focus value


cblecker:
@kubernetes/sig-testing-misc Can someone have a look at this? Thanks!

PR #1876, opened 2018-03-02
fix typo

CaoShuFeng:


cblecker:
Thanks for this @CaoShuFeng .. Looks like this was already fixed before this PR merged.

/close

PR #1879, opened 2018-03-03
Update sig-aws leads: add countspongebob

justinsb:
As agreed in sig-aws.


justinsb:
cc @countspongebob @kris-nova @jbeda

PR #1885, opened 2018-03-04
Added privileges to Responsibilities and privileges section

ellenkorbes:
I had no idea about these and only found out about them when @nikhita told me on chat. I figured it'd be useful for other new members to have these specifically pointed out to them.

Feel free to suggest different wording, or add anything we missed.

### DIFF
--- a/community-membership.md
+++ b/community-membership.md
@@ -88,9 +88,10 @@ List of contributions:
   - Code is well tested
   - Tests consistently pass
   - Addresses bugs or issues discovered after code is accepted
-- Can do a `/lgtm` on open PRs.
-- Can be assigned to issues and PRs, and people can ask you for reviews with a `/cc @username`.
-- No `/ok-to-test` needed.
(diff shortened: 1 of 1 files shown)


community-membership.md
```diff
@@ -88,6 +88,9 @@ List of contributions:
   - Code is well tested
   - Tests consistently pass
   - Addresses bugs or issues discovered after code is accepted
+- Can do a `/lgtm` on open PRs.
+- Can be assigned to issues and PRs, and people can ask you for reviews with a `/cc @username`.
+- No `/ok-to-test` needed.
```
nikhita: Something like "Tests can be run against their PRs automatically".

nikhita: Members can do a `/ok-to-test` for PRs that have `needs-ok-to-test` and use commands like `/close` to close PRs as well.

ellenkorbes:
Thanks! I've added in those corrections.

PR #1893, opened 2018-03-05
Support traffic shaping for Kubelet CNI network plugin

m1093782566:
Currently the kubenet code supports applying basic traffic shaping during pod setup. This will happen if bandwidth-related annotations have been added to the pod's metadata, for example:

    {
        "kind": "Pod",
        "metadata": {
            "name": "iperf-slow",
            "annotations": {
                "kubernetes.io/ingress-bandwidth": "10M",
                "kubernetes.io/egress-bandwidth": "10M"
            }
        }
    }

Our current implementation uses the linux tc to add an download(ingress) and upload(egress) rate limiter using 1 root qdisc, 2 class(one for ingress and one for egress) and 2 filter(one for ingress and one for egress attached to the ingress and egress classes respectively).

Kubelet CNI code doesn't support it yet. However, there is enough functionality for us to add a simple CNI plugin to replicate the behavior we have today in kubenet if we feel this is an important feature. In fact, CNI has already supported traffic shaping plugin.

/sig network

### DIFF
--- a/contributors/design-proposals/network/support_traffic_shaping_for_kubelet_cni.md
+++ b/contributors/design-proposals/network/support_traffic_shaping_for_kubelet_cni.md
@@ -1,0 +1,89 @@
+# Support traffic shaping for CNI network plugin
+
+Version: Alpha
+
+Authors: @m1093782566
+
+## Motivation and background
+
+Currently the kubenet code supports applying basic traffic shaping during pod setup. This will happen if bandwidth-related annotations have been added to the pod's metadata, for example:
+
+```json
+{
+    "kind": "Pod",
+    "metadata": {
+        "name": "iperf-slow",
+        "annotations": {
+            "kubernetes.io/ingress-bandwidth": "10M",
+            "kubernetes.io/egress-bandwidth": "10M"
+        }
+    }
+}
+```
+
+Our current implementation uses the `linux tc` to add an download(ingress) and upload(egress) rate limiter using 1 root `qdisc`, 2 `class `(one for ingress and one for egress) and 2 `filter`(one for ingress and one for egress attached to the ingress and egress classes respectively).
+
+Kubelet CNI code doesn't support it yet, though CNI has already added a [traffic sharping plugin](https://github.com/containernetworking/plugins/tree/master/plugins/meta/bandwidth). We can replicate the behavior we have today in kubenet for kubelet CNI network plugin if we feel this is an important feature.
+
+## Goal
+
+Support traffic shaping for CNI network plugin in Kubernetes.
+
+## Non-goal
+
+CNI plugins to implement this sort of traffic shaping guarantee.
+
+## Proposal
+
+If kubelet starts up with `network-plugin = cni` and user enabled traffic shaping via the network plugin configuration, it would then populate the `runtimeConfig` section of the config when calling the `bandwidth` plugin.
+
+Traffic shaping in Kubelet CNI network plugin can work with ptp and bridge network plugins.
+
+### Pod Setup
+
+When we create a pod with bandwidth configuration in its metadata, for example, 
+
+```json
+{
+    "kind": "Pod",
+    "metadata": {
+        "name": "iperf-slow",
+        "annotations": {
+            "kubernetes.io/ingress-bandwidth": "10M",
+            "kubernetes.io/egress-bandwidth": "10M"
+        }
+    }
+}
+```
+
+Kubelet would firstly parse the ingress and egress bandwidth values and transform them to Kbps because both `ingressRate` and `egressRate` in cni bandwidth plugin are in Kbps. A user would add something like this to their CNI config list if they want to enable traffic shaping via the plugin:
+
+```json
+{
+  "type": "bandwidth",
+  "capabilities": {"trafficShaping": true}
+}
+```
+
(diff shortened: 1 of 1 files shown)


m1093782566:
/cc @cmluciano 

I guess you are interested in it? :)

contributors/design-proposals/network/support_traffic_shaping_for_kubelet_cni.md
```diff
@@ -0,0 +1,72 @@
...
+            "kubernetes.io/ingress-bandwidth": "10M",
+            "kubernetes.io/egress-bandwidth": "10M"
+        }
+    }
+}
+```
+
+Kubelet will parse the ingress and egress bandwidth values and transform them to Kb because both `ingressRate` and `egressRate` in cni bandwidth plugin are in Kbps. Eventually, kubelet will bulid the runtime config which like the following and call cni plugin `AddNetworkList` API.
+
+```json
```
caseydavenport: I've asked on the upstream plugin regarding support for runtimeConfig / capabilities similar to how the portmap plugin works today: https://github.com/containernetworking/plugins/pull/96#issuecomment-370952052

I think we want to agree a capability and use that to negotiate whether or not shaping will be done.

So a user would add something like this to their CNI config list if they wanted to enable traffic shaping via the plugin:

```json
{
  "type": "bandwidth",
  "capabilities": {"trafficShaping": true}
}
```

The kubelet would then populate the `runtimeConfig` section of the config when calling the `bandwidth` plugin:

```json
{
  "type": "bandwidth",
  "runtimeConfig": {
    "trafficShaping": {
      "ingressRate": "X",
      "egressRate": "Y"
    }
  }
}
```

m1093782566:
Proposal updated as the CNI traffic shaping plugin got merged.

Lion-Wei:
If we do this, then kubenet `BandwidthShaper` part can be replaced by traffic shaping plugin, big step of deprecating kubenet, I think.

bowei:
@jingax10 can you take a look?

m1093782566:
@jingax10 

> One quick question, can those ingress/egress limit be tuned later?

That's a good point. Technically, if `AddNetwork` and `AddNetworkList` of CNI APIs are idempotent, we can call these APIs as many times as required. So, ingress/egress limit be tuned via updating Pod annotations.

BTW, traffic shaping in kubenet only take effect in Pod setup.

jingax10:
> That's a good point. Technically, if AddNetwork and AddNetworkList of CNI APIs are idempotent, we can call these APIs as many times as required. So, ingress/egress limit can be tuned via updating Pod annotations.
> 
> BTW, traffic shaping in kubenet ONLY takes effect in Pod setup.

Thanks for your reply! If limits are immutable, I could imagine it could be only used for bound the network usages from this pod if pod itself could become the bottleneck to serve (e.g., kubenet). Even if limits are mutable, how to set the limit dynamically is an art. Moreover, the granularity of the flows (for now only in and out without knowing the destinations) may not be sufficient for the control system to calculate the limits if one understands where are the bottlenecks. I am just trying to understand your use cases :-)

thockin:
For the sake of my queue, I am unassigning myself.  Assign me (don't just CC me) when/if it is ready to be reviewed.

caseydavenport:
I'm in favor of this as a way to move towards deprecating kubenet. The alternative would be to drop support for these annotations altogether, or revisit more holistically how we want to do bandwidth shaping (not sure how much appetite there is for such a discussion though).

Related, are these documented as alpha or beta somewhere or are these considered "GA"? My understanding was that these are an alpha type feature, but the annotation name doesn't indicate that.

yujuhong:
> Related, are these documented as alpha or beta somewhere or are these considered "GA"?

Bandwidth shaping is an "experimental" feature, so I think the closest definition would be alpha.

> My understanding was that these are an alpha type feature, but the annotation name doesn't indicate that.

I believe the annotation was added before the alpha format was defined...

m1093782566:
@caseydavenport 

This feature target alpha, I updated this document to clarify it.

@sig-network-leads any more comments? Can we push it forward?

m1093782566:
/assign @thockin 

as it's ready for final review!

contributors/design-proposals/network/support_traffic_shaping_for_kubelet_cni.md
```diff
@@ -0,0 +1,89 @@
...
+Kubelet would firstly parse the ingress and egress bandwidth values and transform them to Kbps because both `ingressRate` and `egressRate` in cni bandwidth plugin are in Kbps. Kubelet would  then detect whether user has enabled the traffic shaping plugin by checking the following CNI config file:
+
+```json
+{
+  "type": "bandwidth",
+  "capabilities": {"trafficShaping": true}
+}
+```
+
+If traffic shaping plugin is enabled, kubelet would populate the `runtimeConfig` section of the config when call the `bandwidth` plugin:
```
thockin: Can we just always pass this, so no checking required?

thockin:
This is OK for now.  We need a long-term answer that respects the general resource model and plugin model.

/lgtm
/approve

m1093782566:
>  We need a long-term answer that respects the general resource model and plugin model.

Yeah, definitely. That's also what my team is exploring :)

PR #1910, opened 2018-03-08
Create new KEP directory for SIG Network with just a README for now.

rramkumar1:
/assign @calebamiles


jbeda:
Feel free to add an OWNERS file here too so that SIG-network can own the destiny of their own KEPs.

PR #1911, opened 2018-03-08
categorize and expand api machinery projects

lavalamp:


lavalamp:
@pwittrock

lavalamp:
will fix test after I repair my local go stack (had to reimage my machine not long ago and 'make generate' makes very poor assumptions about the environment).

cblecker:
@lavalamp `make generate-dockerized` runs the same within a golang docker container

lavalamp:
too late, already fixed my computer ;)

lavalamp:
Filed https://github.com/kubernetes/community/issues/1913 since I'm definitely never going to find time to maintain this list after this PR merges.

bgrant0607:
@lavalamp Yes, we plan to eventually generate these lists bottom-up from the OWNERS files, but we don't have that machinery yet.

bgrant0607:
Cool. Thanks for adding the additional directories and rationalizing the categories.

/lgtm

PR #1914, opened 2018-03-09
WIP: Design of local storage dynamic provisioning

lichuqiang:
Still in progress, more details need to be added
The change relies on https://github.com/kubernetes/community/pull/1857

@msau42 @NicolasT

### DIFF
--- a/contributors/design-proposals/storage/local-storage-pv.md
+++ b/contributors/design-proposals/storage/local-storage-pv.md
@@ -783,16 +783,16 @@ The local volume provisioner needs to be updated in several areas to support dyn
 
 Mechanisms to manage local storage sources are varied.
 In addition to current `populator`, `deletor` and `discoverer`, a new component named
-`storageManager` will be introduced in the plugin, to manage the backend storage.
+`dynamicProvisioningManager` will be introduced in the plugin, to manage the backend storage.
 
-According to the input of classes in the configuration, there will be a list of `backends`
-under the manager, to communicate with backend storage. `backend` is a general interface,
-each kind of storage should have its own implementation. storageManager store the backends
-according to storage classes, backends of different classes can be different.
+According to the input of classes in the configuration, there will be a list of `storageBackends`
+under the manager, to communicate with backend storages. `StorageBackend` is a general interface,
+each kind of storage should have its own implementation. `dynamicProvisioningManager` stores the
+backends according to storage classes, backends of different classes can be different.
 LVM backend will be introduced in alpha phase.
 
 ```
-Backends map[string]backend.DynamicProvisioningBackend
+storageBackends map[string]backend.StorageBackend
 ```
 
 During start-up, a new environment variable is needed for the provisioner DaemonSet:
@@ -826,10 +826,10 @@ data:
         - "2"
 ```
 
-A StorageManager instance is responsible for:
+A manager instance is responsible for:
 
-1. Monitor available capacity for provisioning, and update the capacity in StorageClass if
-there is a change.
+1. Implement interface `GetCapacity(storageClass string) int64` to report capacity of
+certain class.
 2. Implement interface `CreateLocalVolume(claim *v1.PersistentVolumeClaim) error`
 to create PVs and the local volumes behind them.
 3. Implement interface `DeleteLocalVolume(pv *v1.PersistentVolume) error` to
@@ -837,12 +837,16 @@ clean up the local volume behind a PV.
 
 ##### Reconcile Capacity
 
-The StorageManager is expected to reconcile capacity of StorageClasses in its local
-`storageClassMap` regularly. Particularly, for LVM, the proposed workflow is:
+A new component `capacityReconciler` is expected to reconcile capacity of StorageClasses
+in its local `storageClassMap` regularly. Interface `GetCapacity` of `dynamicProvisioningManager`
+will be passed into the reconciler as a hook (`capacityGetFunc`), to fetch capacity of certain
+class.
+
+The proposed workflow is:
 
 1. Walks through classes in its `storageClassMap` regularly
-2. Fetch the VG of each class if `volumeGroup` is specified
-3. Compare the value in `VG size` with the capacity in StorageClass API object
+2. Fetch actual capacity of a class via hook `capacityGetFunc`
+3. Compare the capacity with that in the StorageClass API object
 4. Update the capacity in StorageClass object if there is a difference
 
 **Note:** For the alpha phase, thinly-provisioned VG (overcommit) is not supported,
@@ -867,14 +871,14 @@ For these purposes, the following permissions are required:
 ##### Volume Creation
 
 During start-up of the provisioner, a new work queue `ProvisionQueue` will be created,
-and passed into both `populator` and `storageManager`:
+and passed into both `populator` and `dynamicProvisioningManager`:
 
 ```
 ProvisionQueue *workqueue.Type
 ```
-When it finds a PVC is in need of privsioning, as desicribed above,
+When it finds a PVC is in need of provisioning, as described above,
 the `populator` will pass it into the queue to trigger volume privisioning;
-`StorageManager` will be the consumer, it will watch the queue and
+`dynamicProvisioningManager` will be the consumer, it will watch the queue and
 call its `CreateLocalVolume` function to:
 
 1. Create volumes behind a PV accordingly
@@ -890,8 +894,8 @@ that the scheduler made a provisioning decision before the capacity initialized
 in StorageClasses. Return error.
 3. Check if the class is available for dynamic provisioning
 (e.g. has an input of `volumeGroup` for LVM). If not, return error.
-4. Create volumes behind a PV (e.g. for LVM, Create `logical volume (LV)`
-object in the VG).
+4. Call the volume creation funcion of related storageBackend to create volumes
+behind a PV (e.g. for LVM, Create `logical volume (LV)` object in the VG).
 5. Create a PV API object that is pre-bound to the PVC.
 
 **Note:** As it describes in [Volume Topology-aware Scheduling](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/volume-topology-scheduling.md),
@@ -919,10 +923,15 @@ Currently, Intended `VolumeMode` of a local volume is determined by the state of
 the input path. If a user need a local volume whose `VolumeMode` is filesystem,
 the path provided in `LocalVolumeSource` must be a directory. Thus, for dynamic
 provisioned volumes, the provisioner need to mount the LV to a global path first,
-to allow kubelet for another mount. It increases the complexity of the plugin.
+to allow kubelet for another mount. It increases the complexity of the provisioner,
+and makes handling reboot scenarios difficult.
 
 Therefore, in-tree local storage plugin will be updated to support both directory
-and block as volume source.
+and block as volume source. When the specified Path is detected to be a block device
+and the volumeMode is filesystem, the plugin will create the filesystem (on the first
+mount) and globally mount the device to the plugin path. This is similar to how other
+volume plugins handle filesystem on block.
+
 Specifically, we will implement the following interfaces, to manage the global mount
 when volume source is of block:
 
@@ -950,9 +959,9 @@ type LocalVolumeSource struct {
 	Path string
 	// Filesystem type to mount.
 	// Must be a filesystem type supported by the host operating system.
-	// Ex. "ext4", "xfs", "ntfs". Implicitly inferred to be "ext4" if unspecified.
+	// Ex. "ext4", "xfs", "ntfs". Default is "ext4".
 	// +optional
-	FSType string
+	FSType *string
 }
 ```
 And on the provisioner side, new field `volumeMode` will be added into the config,
@@ -987,7 +996,7 @@ in their claims.
 Instead of existing volume groups, we can take disk partitions as storage source.
 This way, the provisioner is in charge of underlying storage.
 
-**Advances:**
+**Advantages:**
 
 Cluster admins do not need to make extra preparations and maintenance work.
 The only thing needed is to prepare the disks for provisioning, and configure their paths
@@ -998,6 +1007,6 @@ in storage sources.
 1. The plugin will need to manage the life cycle of the storage resources behind the LVM.
 When some disks are added, removed or failed, the plugin need to process and recover accordingly.
 In addition, the underlying storage resources are varied, they may be  physical disks,
-RAIDs, and so on, so it's hard to cover all the scenes.
+RAIDs, and so on, so it's hard to cover all the scenarios.
 2. As provisioners running on all nodes share a same configuration ConfigMap,
 disk partitions of their nodes are required to be named the same.
--- a/contributors/design-proposals/storage/local-storage-pv.md
+++ b/contributors/design-proposals/storage/local-storage-pv.md
@@ -785,14 +785,14 @@ Mechanisms to manage local storage sources are varied.
 In addition to current `populator`, `deletor` and `discoverer`, a new component named
 `storageManager` will be introduced in the plugin, to manage the backend storage.
 
-According to the input of classes in the configuration, there will be a list of `wrappers`
-under the manager, to communicate with backend storage. `wrapper` is a general interface,
-each kind of storage should have its own implementation. storageManager store the wrappers
-according to storage classes, wrappers of different classes can be different.
-LVM wrapper will be introduced in alpha phase.
+According to the input of classes in the configuration, there will be a list of `backends`
+under the manager, to communicate with backend storage. `backend` is a general interface,
+each kind of storage should have its own implementation. storageManager store the backends
+according to storage classes, backends of different classes can be different.
+LVM backend will be introduced in alpha phase.
 
 ```
-Wrappers map[string]wrapper.LocalVolumeWrapper
+Backends map[string]backend.DynamicProvisioningBackend
 ```
 
 During start-up, a new environment variable is needed for the provisioner DaemonSet:
@@ -806,7 +806,7 @@ in PVs with `AnnProvisionedBy`, to indicate which PVs are up to a certain node.
 To distinguish it from the new env `PROVISIONER_NAME`, we will rename it to `provisionerTag`.
 
 New field will be added into the configuration ConfigMap, to specify available
-volume group for lvm:
+volume group for lvm, the group should be created by the admin in advance:
 
 ```
 kind: ConfigMap
@@ -821,6 +821,9 @@ data:
     local-dynamic:
       lvm:
         volumeGroup: "volumeGroup1"
+      blockCleanerCommand:
+        - "/scripts/dd_zero.sh"
+        - "2"
 ```
 
 A StorageManager instance is responsible for:
@@ -877,18 +880,19 @@ call its `CreateLocalVolume` function to:
 1. Create volumes behind a PV accordingly
 2. Create a PV object that is pre-bound to the PVC
 
-For LVM, the proposed workflow is:
+The proposed workflow is:
 
 1. Check if there is already a PV pre-bound to the PVC, if so, return directly
 (This can happen when a volume has already been provisioned for a PVC,
 but has not been bound to the PVC by the controller yet).
-2. Check if the its capacity has been initialized in the StorageClass.
-If not, return error.
-3. Check if the class is available for dynamic provisioning (i.e. has an input of `volumeGroup`).
-If not, return error.
-4. Create `logical volume (LV)` object in the VG.
-5. Create a PV API object that is pre-bound to the PVC, and the path
-of local storage in the PV should be the path of the LV created above.
+2. Check if `annProvisionerTopology` is set on the PVC. If not, then it means
+that the scheduler made a provisioning decision before the capacity initialized
+in StorageClasses. Return error.
+3. Check if the class is available for dynamic provisioning
+(e.g. has an input of `volumeGroup` for LVM). If not, return error.
+4. Create volumes behind a PV (e.g. for LVM, Create `logical volume (LV)`
+object in the VG).
+5. Create a PV API object that is pre-bound to the PVC.
 
 **Note:** As it describes in [Volume Topology-aware Scheduling](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/volume-topology-scheduling.md),
 in whichever step it fails, the provisioner will signal back to the scheduler
@@ -911,28 +915,38 @@ and for dynamic privisioned volumes, the deleting process should be:
 
 ##### Expansion of Local Volume Plugin
 
-Currently, the path in `LocalVolumeSource` must be a directory. Thus, for dynamic
-provisioned volumes, the provisioner need to mount the LV to a host path first,
+Currently, Intended `VolumeMode` of a local volume is determined by the state of
+the input path. If a user need a local volume whose `VolumeMode` is filesystem,
+the path provided in `LocalVolumeSource` must be a directory. Thus, for dynamic
+provisioned volumes, the provisioner need to mount the LV to a global path first,
 to allow kubelet for another mount. It increases the complexity of the plugin.
 
 Therefore, in-tree local storage plugin will be updated to support both directory
-and block as a source of mount. The mount process would be:
-1. Detect if given path in volumeSource is mounted or not, if mounted,
-continue to work in the existing way.
-2. If not, Check if global path `{pluginDir}/mounts/{volumeID}` exists.
-and create the directory if not.
-3. Check if the global path is of mount point, if not, mount given path to
-the global path.
-4. Manage `mkfs` for the global path if it is newly created or mounted.
-5. Mount the global path to the path in pods.
+and block as volume source.
+Specifically, we will implement the following interfaces, to manage the global mount
+when volume source is of block:
 
-New field `FSType` will be added in `LocalVolumeSource` to support `mkfs` in the mounter.
+* GetDeviceMountPath(spec *Spec) (string, error)
+* MountDevice(spec *Spec, devicePath string, deviceMountPath string) error
+* UnmountDevice(deviceMountPath string) error
+
+1. GetDeviceMountPath: Detect if given path in volumeSource is a directory,
+if so, then it means the path is already of filesystem, return it as global
+path directly; if not, return global path `{pluginDir}/mounts/{volumeID}`.
+2. MountDevice: if given path is of filesystem, return without doing anything;
+if it's of block, check if global path exist, if not, create the path; check
+if the global path is of mount point, if not, try to format and mount given path
+to the global path.
+3. UnmountDevice: if the mount path is in format of `{pluginDir}/mounts/...`,
+unmount it; otherwise return directly.
+
+New field `FSType` will be added in LocalVolumeSource, to indicate the filesystem
+type when formating a disk input path:
 
 ```
 type LocalVolumeSource struct {
 	// The full path to the volume on the node
-	// For alpha, this path must be a directory
-	// Once block as a source is supported, then this path can point to a block device
+	// The path can be a directory or point to a block device
 	Path string
 	// Filesystem type to mount.
 	// Must be a filesystem type supported by the host operating system.
@@ -941,9 +955,30 @@ type LocalVolumeSource struct {
 	FSType string
 }
 ```
+And on the provisioner side, new field `volumeMode` will be added into the config,
+to indicate intended volume mode of statically discovered volumes:
 
-**Note:** The change will only work on PersistentVolumes whose volumeMode are "Filesystem",
-PVs of "Block" mode are not affected.
+```
+kind: ConfigMap
+metadata:
+  name: local-volume-config
+  namespace: kube-system
+data:
+ storageClassMap: |
+    local-static:
+      hostDir: "/mnt/ssds"
+      mountDir: "/local-ssds"
+      volumeMode: "Filesystem"
+    local-dynamic:
+      lvm:
+        volumeGroup: "volumeGroup1"
+      blockCleanerCommand:
+        - "/scripts/dd_zero.sh"
+        - "2"
+```
+
+**Note:** `VolumeModes` of dynamically provisioned volumes are determined by the volume mode
+in their claims.
 
 ##### Alternatives
 
@@ -964,5 +999,5 @@ in storage sources.
 When some disks are added, removed or failed, the plugin need to process and recover accordingly.
 In addition, the underlying storage resources are varied, they may be  physical disks,
 RAIDs, and so on, so it's hard to cover all the scenes.
-2. As provisioners running on all nodes shall a same configuration ConfigMap,
+2. As provisioners running on all nodes share a same configuration ConfigMap,
 disk partitions of their nodes are required to be named the same.
--- a/contributors/design-proposals/storage/local-storage-pv.md
+++ b/contributors/design-proposals/storage/local-storage-pv.md
@@ -785,8 +785,16 @@ Mechanisms to manage local storage sources are varied.
 In addition to current `populator`, `deletor` and `discoverer`, a new component named
 `storageManager` will be introduced in the plugin, to manage the backend storage.
 
-`storageManager` is a general interface, each kind of storage should have its own implementation.
-LVM StorageManager will be introduced in alpha phase, acting as the default StorageManager.
+According to the input of classes in the configuration, there will be a list of `wrappers`
+under the manager, to communicate with backend storage. `wrapper` is a general interface,
+each kind of storage should have its own implementation. storageManager store the wrappers
+according to storage classes, wrappers of different classes can be different.
+LVM wrapper will be introduced in alpha phase.
+
+```
+Wrappers map[string]wrapper.LocalVolumeWrapper
+```
+
 During start-up, a new environment variable is needed for the provisioner DaemonSet:
 
 * `PROVISIONER_NAME` to specify the name of the provisioner, its value matches the Provisioner
@@ -819,14 +827,15 @@ A StorageManager instance is responsible for:
 
 1. Monitor available capacity for provisioning, and update the capacity in StorageClass if
 there is a change.
-2. Implement interface `CreateLocalVolume(*v1.PersistentVolumeClaim) *v1.PersistentVolume`
+2. Implement interface `CreateLocalVolume(claim *v1.PersistentVolumeClaim) error`
 to create PVs and the local volumes behind them.
-3. Implement interface `DeleteLocalVolume(string) error` to clean up the local volume behind a PV.
+3. Implement interface `DeleteLocalVolume(pv *v1.PersistentVolume) error` to
+clean up the local volume behind a PV.
 
 ##### Reconcile Capacity
 
 The StorageManager is expected to reconcile capacity of StorageClasses in its local
-`storageClassMap` regularly. Particularly, LVM StorageManager behaviors as follows:
+`storageClassMap` regularly. Particularly, for LVM, the proposed workflow is:
 
 1. Walks through classes in its `storageClassMap` regularly
 2. Fetch the VG of each class if `volumeGroup` is specified
@@ -838,19 +847,12 @@ We'll consider it in a subsequent phase.
 
 ##### Change in Populator
 
-In addition to PV, the `populator` will also be extended to watch StorageClasses and PVC objects of all namespaces.
-
-It stores latest StorageClass objects in cache, they are used:
-
-* To reconcile capacity, the StorageManager need to compare actual capacity with that recorded
-in StorageClass objects
-* During volume creation, some opitions (e.g. fsType for `mkfs`) are specified in StorageClasses as Parameters
-
-And for PVCs, when the `populator` finds:
+In addition to PV, the `populator` will also be extended to watch PVC objects of all namespaces.
+When the `populator` finds:
 
 1. An unbound PVC is annotated with `annSelectedNode`, and its node name match the value
-of the annotation
-2. Value of the `annStorageProvisioner` on the PVC matches the name of the provisioner
+of the annotation.
+2. Value of the `annStorageProvisioner` on the PVC matches the name of the provisioner.
 
 It then trigger volume provsioning, detailed in the volume creation section.
 
@@ -861,15 +863,15 @@ For these purposes, the following permissions are required:
 
 ##### Volume Creation
 
-During start-up of the provisioner, a new channel `provisionChan` will be created,
+During start-up of the provisioner, a new work queue `ProvisionQueue` will be created,
 and passed into both `populator` and `storageManager`:
 
 ```
-provisionChan chan *v1.PersistentVolumeClaim
+ProvisionQueue *workqueue.Type
 ```
 When it finds a PVC is in need of privsioning, as desicribed above,
-the `populator` will pass it into the channel to trigger volume privisioning;
-`StorageManager` will be the consumer, it will watch the channel and
+the `populator` will pass it into the queue to trigger volume privisioning;
+`StorageManager` will be the consumer, it will watch the queue and
 call its `CreateLocalVolume` function to:
 
 1. Create volumes behind a PV accordingly
@@ -880,15 +882,20 @@ For LVM, the proposed workflow is:
 1. Check if there is already a PV pre-bound to the PVC, if so, return directly
 (This can happen when a volume has already been provisioned for a PVC,
 but has not been bound to the PVC by the controller yet).
-2. Check if Storage Class of the PVC is in record of its storageClass map.
+2. Check if the its capacity has been initialized in the StorageClass.
 If not, return error.
 3. Check if the class is available for dynamic provisioning (i.e. has an input of `volumeGroup`).
 If not, return error.
-4. Check if `Free Size` of the VG under the class can satisfy the capacity requested
-in the PVC. If not, return error.
-5. Create `logical volume (LV)` object in the VG.
-6. Create a PV API object that is pre-bound to the PVC, and the path
+4. Create `logical volume (LV)` object in the VG.
+5. Create a PV API object that is pre-bound to the PVC, and the path
 of local storage in the PV should be the path of the LV created above.
+
+**Note:** As it describes in [Volume Topology-aware Scheduling](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/volume-topology-scheduling.md),
+in whichever step it fails, the provisioner will signal back to the scheduler
+to retry dynamic provisioning by removing the "annSelectedNode" annotation of
+the PVC.
+
+For the purpose, it requires the permission to update PersistentVolumeClaims.
 
 ##### Cleanup after Release
 
--- a/contributors/design-proposals/storage/local-storage-pv.md
+++ b/contributors/design-proposals/storage/local-storage-pv.md
@@ -787,25 +787,19 @@ In addition to current `populator`, `deletor` and `discoverer`, a new component 
 
 `storageManager` is a general interface, each kind of storage should have its own implementation.
 LVM StorageManager will be introduced in alpha phase, acting as the default StorageManager.
-During start-up, two new environment variables are needed for the provisioner DaemonSet:
+During start-up, a new environment variable is needed for the provisioner DaemonSet:
 
-* `STORAGE_BACKEND` to determine which StorageManager instance to start
 * `PROVISIONER_NAME` to specify the name of the provisioner, its value matches the Provisioner
-specified in the StorageClass.
+specified in the StorageClass. It will be default to `local-volume-provisioner` if not specified.
 
 **Note:** Currently we already have a `provisionerName` in `RuntimeConfig` of the plugin,
 which is in format of `local-volume-provisioner-${NODE_NAME}-${NODE_ID}`, it is annotated
 in PVs with `AnnProvisionedBy`, to indicate which PVs are up to a certain node.
 To distinguish it from the new env `PROVISIONER_NAME`, we will rename it to `provisionerTag`.
 
-A StorageManager instance is responsible for:
-
-1. Monit available capacity for provisioning, and update the capacity in StorageClass if
-there is a change.
-2. Provide the ability to create volumes, detailed in the volume creation section.
-
-A new field `storageSources` will be added into the configuration ConfigMap, to illustrate available
-storage source for provisioning.
+New field `storageBackend` will be added into the configuration ConfigMap, to specify the backend,
+and only lvm is supported for alpha phase; Field `volumeGroup` will also be added to illustrate
+available volume group for lvm.
 
 ```
 kind: ConfigMap
@@ -817,14 +811,17 @@ data:
     local-static:
       hostDir: "/mnt/ssds"
       mountDir: "/local-ssds"
-    local_dynamic:
-      storageSources:
-        - "volumeGroup1"
-      blockCleanerCommand: "/scripts/lv_release.sh"
+    local-dynamic:
+      storageBackend: "lvm"
+      volumeGroup: "volumeGroup1"
 ```
-For example for an LVM StorageManager, the list of storageSource could contain volume groups.
-For other StorageManagers, the list could contain other storage sources (e.g. unpartitioned disks).
-Each StorageManager should know the scope and what to do with the input.
+
+A StorageManager instance is responsible for:
+
+1. Monit available capacity for provisioning, and update the capacity in StorageClass if
+there is a change.
+2. Provide the ability to create volumes, detailed in the volume creation section.
+3. Implement interface `DeleteLocalVolume(string) error` to clean up the local volume behind a PV.
 
 ##### Reconcile Capacity
 
@@ -832,19 +829,12 @@ The StorageManager is expected to reconcile capacity of StorageClasses in its lo
 `storageClassMap` regularly. Particularly, LVM StorageManager behaviors as follows:
 
 1. Walks through classes in its `storageClassMap` regularly
-2. Fetch VGs (volume groups) of each class if `storageSources` is specified
-3. Sum up the `VG size`, and compare it with the capacity in StorageClass API object
+2. Fetch the VG of each class if `volumeGroup` is specified
+3. Compare the value in `VG size` with the capacity in StorageClass API object
 4. Update the capacity in StorageClass object if there is a difference
 
-**Note:** For the alpha phase, thinly-provisioned VGs (overcommit) is not supported,
+**Note:** For the alpha phase, thinly-provisioned VG (overcommit) is not supported,
 We'll consider it in a subsequent phase.
-
-Though a list of VGs can be configured as storage source, it's not recommend to
-pass in more than one VG. We may suffer from some edge-cases:
-
-Suppose we have two VGs on a node for provisioning, VG1 and VG2, and they each has 10GB available,
-then the reported capacity would be 20GB, but a pod with storage request of 15GB could not
-be satisfied indeed.
 
 ##### Change in Populator
 
@@ -893,36 +883,44 @@ For LVM, the proposed workflow is:
 but has not been bound to the PVC by the controller yet).
 2. Check if Storage Class of the PVC is in record of its storageClass map.
 If not, return error.
-3. Check if the class is available for dynamic provisioning (i.e. has an input of `storageSources`).
+3. Check if the class is available for dynamic provisioning (i.e. has an input of `volumeGroup`).
 If not, return error.
-4. Check if there is a VG under the class, whose `Free Size` can satisfy the
-capacity requested in the PVC. If not, return error; otherwise, choose a VG for provisioning.
-5. Create `logical volume (LV)` object in the chosen VG.
-6. Initialize its file system with `mkfs`. FsType and other potential options
-for FS creation can be specified in StorageClass as Parameters
-7. Create a PV API object that is pre-bound to the PVC, and the path
-of local storage in the PV is the path of the LV created above.
+4. Check if `Free Size` of the VG under the class can satisfy the capacity requested
+in the PVC. If not, return error.
+5. Create `logical volume (LV)` object in the VG.
+6. Create a PV API object that is pre-bound to the PVC, and the path
+of local storage in the PV should be the path of the LV created above.
 
 ##### Cleanup after Release
 
 Currently we already have a component `deleter` to cleanup volumes/PVs of static local volume,
 detail cleanup actions are managed via custom scripts/binaries that specified in `blockCleanerCommand`.
 
-The behavior is extensible enough to support cleanup task of provsioned volumes too.
-So nothing new will be introduced here in the provisioner, but the difference lies
-in the cleanup script/binary itself.
-
-Existing script `quick_reset.sh` calls `mkfs -F` and `wipefs` to cleanup a block,
-For dynamic privisioned volumes, the logical volume in LVM also need to be deleted after that.
+Existing script `quick_reset.sh` calls `mkfs -F` and `wipefs` to cleanup a block.
+In addition to that, a hook named `volumeDeleteFunc` will be added into `deleter`,
+and for dynamic privisioned volumes, the deleting process should be:
+1. Call scripts/binaries in `blockCleanerCommand` to clean up data in the volume.
+2. Call `volumeDeleteFunc` to recycle the volume behind the PV (e.g. delete the LV for lvm).
+3. Delete the PV object.
 
 ##### Expansion of LocalVolumeSource
 
 Currently, the path in `LocalVolumeSource` must be a directory. Thus, for dynamic
-provisioned volumes, the provisioner need to mount the LV to a host path first, to allow
-kubelet for another mount. It increases the probability of failure to mount twice.
+provisioned volumes, the provisioner need to mount the LV to a host path first,
+to allow kubelet for another mount. It increases the complexity of the plugin.
 
-Therefore, new field `Type` will be introduced into `LocalVolumeSource`, to support both
-directory and block as a source of mount:
+Therefore, in-tree local storage plugin will be updated to support both directory
+and block as a source of mount. The mount process would be:
+1. Detect if given path in volumeSource is mounted or not, if mounted,
+continue to work in the existing way.
+2. If not, Check if global path `{pluginDir}/mounts/{volumeID}` exists.
+and create the directory if not.
+3. Check if the global path is of mount point, if not, mount given path to
+the global path.
+4. Manage `mkfs` for the global path if it is newly created or mounted.
+5. Mount the global path to the path in pods.
+
+New field `FSType` will be added in `LocalVolumeSource` to support `mkfs` in the mounter.
 
 ```
 type LocalVolumeSource struct {
@@ -930,21 +928,16 @@ type LocalVolumeSource struct {
 	// For alpha, this path must be a directory
 	// Once block as a source is supported, then this path can point to a block device
 	Path string
-	// Type of the volume source, default to Directory
-	Type *LocalVolumeType
+	// Filesystem type to mount.
+	// Must be a filesystem type supported by the host operating system.
+	// Ex. "ext4", "xfs", "ntfs". Implicitly inferred to be "ext4" if unspecified.
+	// +optional
+	FSType string
 }
-
-type LocalVolumeType string
-
-const (
-	// A directory must exist at the given path
-	LocalVolumeDirectory LocalVolumeType = "Directory"
-	// A block device must exist at the given path
-	LocalVolumeBlockDev LocalVolumeType = "BlockDevice"
-)
 ```
 
-And the mounter of local volume will manage the mount task basing on the type in `LocalVolumeSource`.
+**Note:** The change will only work on PersistentVolumes whose volumeMode are "Filesystem",
+PVs of "Block" mode are not affected.
 
 ##### Alternatives
 
@@ -953,13 +946,13 @@ And the mounter of local volume will manage the mount task basing on the type in
 Instead of existing volume groups, we can take disk partitions as storage source.
 This way, the provisioner is in charge of underlying storage.
 
-**Advances**:
+**Advances:**
 
 Cluster admins do not need to make extra preparations and maintenance work.
 The only thing needed is to prepare the disks for provisioning, and configure their paths
 in storage sources.
 
-**Downsides**:
+**Downsides:**
 
 1. The plugin will need to manage the life cycle of the storage resources behind the LVM.
 When some disks are added, removed or failed, the plugin need to process and recover accordingly.
@@ -967,7 +960,3 @@ In addition, the underlying storage resources are varied, they may be  physical 
 RAIDs, and so on, so it's hard to cover all the scenes.
 2. As provisioners running on all nodes shall a same configuration ConfigMap,
 disk partitions of their nodes are required to be named the same.
-
-##### Testing
-
-TBD
--- a/contributors/design-proposals/storage/local-storage-pv.md
+++ b/contributors/design-proposals/storage/local-storage-pv.md
@@ -828,12 +828,61 @@ data:
 
 A manager instance is responsible for:
 
-1. Implement interface `GetCapacity(storageClass string) int64` to report capacity of
+1. Implement function `GetCapacity(storageClass string) int64` to report capacity of
 certain class.
-2. Implement interface `CreateLocalVolume(claim *v1.PersistentVolumeClaim) error`
+2. Implement function `CreateLocalVolume(claim *v1.PersistentVolumeClaim) error`
 to create PVs and the local volumes behind them.
-3. Implement interface `DeleteLocalVolume(pv *v1.PersistentVolume) error` to
+3. Implement function `DeleteLocalVolume(pv *v1.PersistentVolume) error` to
 clean up the local volume behind a PV.
+
+##### LVM Backend Details
+
+LVM backend will be introduced in alpha phase. Here are some details about it.
+
+###### Required Paths
+
+As the provisioner plugin runs as a container, we'll need to mount the root path of LVM into
+the container. The path is default to "/dev". Theoretically, the LVM root path can be modified
+to other paths, but we haven't see practical use cases so far. Besides, if a user decide to
+mount it to another path, he/she will have to modify the config of LVM and udev in the
+provisioner image manually. So for the first phase, we'll hard code the paths to "/dev",
+and make it configurable in later releases if the need ever arises.
+
+###### Parameters To Initialize a Backend
+
+Volume group name and the root path of LVM (default to "/dev") are needed to initialize
+a `LvmBackend`:
+
+```
+NewLvmBackend(volumeGroup, rootPath string) *LvmBackend
+```
+
+###### Responsibility
+
+The backend is responsible for creating and deleting volumes (LVs) when it's called by
+the `dynamicProvisioningManager`. To create a volume, the manager will need to pass
+volume name and volume size into the backend; and volume name is needed to delete a LV.
+
+The name of a volume will be same to its PV object. Thus, the path of a created LV will be:
+
+```
+/dev/{volumeGroupName}/{pvName}
+```
+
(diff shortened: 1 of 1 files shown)


ianchakeres:
FYI: @dhirajh

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+### Phase 3: 1.11 alpha
+
+#### Dynamic provisioning
+
+The local volume provisioner will be enhanced to support dynamic provisioning of local storage,
+which depends on the newly introduced changes in [Volume Topology-aware Scheduling](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/volume-topology-scheduling.md),
+that is:
+
+1. When feature `DynamicProvisioningScheduling` enabled, scheduler would verify a pod's volume requirement,
```
ianchakeres: Reword:

When the feature gate `DynamicProvisioningScheduling` is enabled, the scheduler will verify a pod's volume requirements,

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+and set `annScheduledNode` on PVCs that need provisioning.
```
ianchakeres: /and set/and set an annotation/

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+2. New fields would be added in `StorageClass` to allow storage providers to report
```
ianchakeres: What new fields need to be added?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+capacity limits per topology domain.
+
+The local volume provisioner need to be updated in several aspects to support the scenario.
```
ianchakeres: nit: /need/needs/
nit: /aspects/areas/
nit: /the scenario/dynamic provisioning/

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
 them.  When a PV is released, then it will send the cleanup request to the worker.
 
 Detailed design TBD
+
+
+### Phase 3: 1.11 alpha
+
+#### Dynamic provisioning
```
ianchakeres: Please capitalize all words in headings `####` and `#####`.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+    local-fast:
+      hostDir: "/mnt/ssds"
+    local-slow:
+      hostDir: "/mnt/hdds"
+    local_dynamic:
+      storageSource:
+        - "/dev/sdb"
+        - "/dev/sdc"
+```
+For LVM, its items are disk partitions, and for other mechanisms, they could be other things.
```
ianchakeres: For example for an LVM StorageManager, the list of `storageSource` could contain disk partitions. For other StorageManagers, the list could contain other storage sources (e.g. unpartitioned disks).

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+StorageManager should know the scope and what to do with the input.
```
ianchakeres: /StorageManager/Each StorageManager/

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+Major downsides of this approach:
+
+* As provisioners running on all nodes shall a same configuration ConfigMap,
+storage sources of their nodes are required to be configured in the same way.
+Take LVM as an example, for configmap above, disks used for provisioning are required
+to be partitioned as "/dev/sdb" and "/dev/sdc".
```
ianchakeres: I think this situation could be handled via multiple means.

Cluster operators could leverage multiple dameonset manifests and various selectors. For example, if you had two different types of hardware with a different number of partitions and different partition locations, then you could label the nodes accordingly and deploy two daemonsets where each daemonset config was tailored toward the appropriately labelled nodes.

Alternatively, another earlier process in the node's lifecycle (e.g. nodeprep - https://engineering.salesforce.com/provisioning-kubernetes-local-persistent-volumes-61a82d1d06b0) could perform operations before the lv dynamic provisioner is used, and the node preparation could make nodes fit the common configuration via symlinks or bindmounts.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+  namespace: kube-system
+data:
+ storageClassMap: |
+    local-fast:
+      hostDir: "/mnt/ssds"
+    local-slow:
+      hostDir: "/mnt/hdds"
+    local_dynamic:
+      storageSource:
+        - "/dev/sdb"
```
ianchakeres: I recommend that people use filesystems bind mounted by referencing their by-uuid or raw block devices symlinks by referencing their by-id, and that we do something similar in our examples. https://wiki.archlinux.org/index.php/persistent_block_device_naming

This scheme helps prevent some problems that could happen if you access volumes via names that remain static, when the underlying volumes can move.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+1. When feature `DynamicProvisioningScheduling` enabled, scheduler would verify a pod's volume requirement,
+and set `annScheduledNode` on PVCs that need provisioning.
+2. New fields would be added in `StorageClass` to allow storage providers to report
+capacity limits per topology domain.
+
+The local volume provisioner need to be updated in several aspects to support the scenario.
+
+##### Backend storage management
+
+Mechanisms to manage local storage sources are varied. They could be LVM, Ceph, or something like that.
```
msau42: For Ceph, why can't the Ceph plugin be used?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+* As provisioners running on all nodes shall a same configuration ConfigMap,
+storage sources of their nodes are required to be configured in the same way.
+Take LVM as an example, for configmap above, disks used for provisioning are required
+to be partitioned as "/dev/sdb" and "/dev/sdc".
+
+##### Change in populator
+
+In addition to PV, the `populator` will also be extended to watch PVC objects of all namespaces.
+When it finds that a PVC is newly annotated with `annScheduledNode`, and its node name match the value
```
msau42: It also needs to watch for the provisioner name to match itself.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+of the annotation, it then trigger volume provsioning, detailed in the volume creation section.
+
+For this purpose, the following permissions are required:
+
+* update/list/get PersistentVolumeClaims
```
msau42: Why does it need to update PVC?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+the StorageManager should:
+
+1. Check if related PV (Physical volume in LVM) exists, if not, create one with command:
+```
+pvcreate /dev/new
+```
+
+2. Check if the Physical volume exists in Volume group exStorageGroup, if not, add it:
+```
+vgextend exStorageGroup /dev/new
```
msau42: Who creates/configures the vg?

contributors/design-proposals/storage/local-storage-pv.md
msau42: Carefully consider the case where a disk might be added to the system and on reboot, the ordering in /dev changes.  Would be good to add a workflow detailing adding/removing capacity, and how to replace a failed disk.

contributors/design-proposals/storage/local-storage-pv.md
msau42: Agree, we should support the use case of different number and path to disks on every node, along with disks being dynamically added and removed. Maybe another discovery directory or local file is needed.

contributors/design-proposals/storage/local-storage-pv.md
msau42: Another use case to consider is configuring raid underneath.  We should decide where to draw the line for the provisioner.  Should it:
- manage raid? The lvm input could just be the raid device path instead of the physical disks
- manage lvm? The storage source could just take the vg name

contributors/design-proposals/storage/local-storage-pv.md
ianchakeres: Supporting wildcards or regex in the storageSources listed could provide some additional flexibility to support multiple node's and their individual configurations.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,124 @@ The central manager will poll each worker for their discovered volumes and creat
...
+  name: local-volume-config
+  namespace: kube-system
+data:
+ storageClassMap: |
+    local-fast:
+      hostDir: "/mnt/ssds"
+    local-slow:
+      hostDir: "/mnt/hdds"
+    local_dynamic:
+      storageSource:
```
ianchakeres: Should `storageSource` be `storageSources`, since there is a list below it?

lichuqiang:
@ianchakeres Comments on expression addressed, thanks for your patience

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Oh, you are right, at the first beginning, I thought the annotation of provisioner is set by the provisioner itself, and forgot to correct it here later when I realize the fact that it's set by the controller.
The `permissions` thing below is for the same reason

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Still the storageManager of LVM, when a new source disk comes, it should check if related vg exists, and create one first if not exists.
The description here is indeed not detailed enough, will have an update

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Really appreciate for you guys suggestions, I will look into that.

lichuqiang: Really appreciate for you guys suggestions, I will look into that.

lichuqiang:
> Another use case to consider is configuring raid underneath. We should decide where to draw the line for the provisioner. Should it:

> * manage raid? The lvm input could just be the raid device path instead of the physical disks
> * manage lvm? The storage source could just take the vg name

@msau42 I hope to manage raid. say `pvcreate /dev/md0` and add it to existing VG.
This way, the behavior could be the same, no matter the path passed in is of physical disks or of raids.
But as I don't have enough knowledge on RAID, could you help clarify whether it possible to mix physical disks and raids in a VG?

lichuqiang:
>Carefully consider the case where a disk might be added to the system and on reboot, the ordering in /dev changes. Would be good to add a workflow detailing adding/removing capacity, and how to replace a failed disk.

Yes, failure recovery is an issue that we haven't considered well in our internal release of HUAWEI.
For example, if we have three disks `/dev/sda`, `/dev/sdb` and `/dev/sdc` in the VG, if `/dev/sdc` failed, what do you think users would like to see?
Update its capcacity to `0` to inform admin for recovery, or just remove the `unknown device` from the VG, and leave the remaining two disks for provisioning?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+the scheduler will verify a pod's volume requirements,
+and set an annotation `annScheduledNode` on PVCs that need provisioning.
+2. New fields `Capacity` will be added in `Status` of `StorageClass` to
+allow storage providers to report capacity limits per topology domain.
+
+The local volume provisioner needs to be updated in several areas to support dynamic provisioning.
+
+##### Backend storage management
+
+Mechanisms to manage local storage sources are varied. They could be LVM, Ceph, or something like that.
```
NicolasT: Why would Ceph be 'local storage'?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+In addition to current `populator`, `deletor` and `discoverer`, a new component named
+`storageManager` will be introduced in the plugin, to manage the backend storage.
+
+`storageManager` is a general interface, each kind of storage should have its own implementation.
+During start-up, a new environment variable `STORAGE_BACKEND` is set in the provisioner DaemonSet,
+to determine which StorageManager instance to start.
+(TBD: Should we support the scenario of multi backends on a node?)
```
NicolasT: Wouldn't it be much simpler if a cluster operator just deploys multiple `DaemonSet`s, each for a different `STORAGE_BACKEND`?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+A StorageManager instance is responsible for:
+
+1. Monit changes in storage input, and handle corresponding initialization.
+2. Update capacity in StorageClass when storage source changed.
+3. Provide the ability to create volumes, detailed in the volume creation section.
+
+Take LVM as an example, when a new source disk "/dev/new" is added to its StorageClass `exStorage`,
+the StorageManager should:
+
+1. Check if related PV (physical volume in LVM) exists, if not, create one with command:
```
NicolasT: IMHO it's problematic if the provisioner is responsible for this. As mentioned below, disk naming etc could be problematic. Instead, I believe storage pool provisioning is a responsibility of the cluster operator (using e.g. Ansible or whatever), only volume provisioning should be performed by the Kubernetes infrastructure.

As such, in the LVM case: the Kubernetes provider creates and manages LVs in a given and pre-existing VG, an admin creates PVs and VGs using whichever mechanism seems suitable.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+```
+pvcreate /dev/new
+```
+
+2. Check if volume group exStorageGroup exists, if not, create it:
+```
+vgcreate exStorageGroup /dev/new
```
NicolasT: See above. If the VG doesn't exist, the provisioner should bail out IMHO.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+```
+
+
+3. If VG exists, check if the physical volume exists in volume group exStorageGroup,
+if not, add it:
+```
+vgextend exStorageGroup /dev/new
```
NicolasT: Nope ;-)

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+  name: local-volume-config
+  namespace: kube-system
+data:
+ storageClassMap: |
+    local-fast:
+      hostDir: "/mnt/ssds"
+    local-slow:
+      hostDir: "/mnt/hdds"
+    local_dynamic:
+      storageSources:
```
NicolasT: I'd say only the name of a VG should be specified for things to work. Nothing more required.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+```
+
+
+3. If VG exists, check if the physical volume exists in volume group exStorageGroup,
+if not, add it:
+```
+vgextend exStorageGroup /dev/new
+```
+
+4. Update capacity in StorageClass if there is a change in VG size.
```
NicolasT: How does this work for thinly-provisioned VGs?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+        - "/dev/sdc"
+```
+For example for an LVM StorageManager, the list of storageSource could contain disk partitions.
+For other StorageManagers, the list could contain other storage sources (e.g. unpartitioned disks).
+Each StorageManager should know the scope and what to do with the input.
+
+Major downsides of this approach:
+
+* As provisioners running on all nodes shall a same configuration ConfigMap,
+storage sources of their nodes are required to be configured in the same way.
```
NicolasT: When you have a couple of servers with 80 drives each, behind RAID controllers or whatnot, this can be *very* tricky! Any 'mistake' (e.g. due to disk relabeling after a reboot) could cause some... 'interesting' results.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+and `storageManager`:
+```
+provisionChan chan *common.LocalPVConfig
+```
+When it finds a PVC is in need of privsioning, as desicribed above,
+the `populator` will consturct an input with necessary config information, and pass it into the channel
+to trigger volume privisioning; `StorageManager` will be the consumer, it will watch the channel and
+manage provisioning when new input received, that is:
+
+1. Create volumes behind a PV accordingly(e.g. exec `lvcreate` for LVM)
```
NicolasT: Consider using bindings to `liblvm2` or `lvm2-dbusd` instead of shell-out?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,138 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+Currently we already have a component `deleter` to cleanup volumes/PVs of static local volume,
+detail cleanup actions are managed via custom scripts/binaries that specified in `blockCleanerCommand`.
+
+The behavior is extensible enough to support cleanup task of provsioned volumes too.
+So nothing new will be introduced here in the provisioner, but the difference lies
+in the cleanup script/binary itself.
+
+For example, dd-zero could be a way to cleanup static volumes,
+but for dynamically provisioned volumes, `lvremove` is also
```
NicolasT: Since the provisioner shall likely create an FS on a create LV, you'll want to `wipefs` before `lvremove`. Rationale: LVM doesn't scrub removed LVs, so if a new one is created immediately after, and `mkfs` is executed, it may fail with `An FS already exists`, and (for safety purposes), I don't think an automated system shall run `mkfs -f` as-is.

Also: how are FS creation options configured?

NicolasT:
Shouldn't there be some design explaining how filesystems created on provisioned LVs (which I assume will be necessary, though not yet described) are mounted after a system reboot, such that the CRI can find them and expose them in a Pod/container?

contributors/design-proposals/storage/local-storage-pv.md
msau42: This will also simplify the issue where all the disks on every node must be named the same.

lichuqiang: Seems you guys have a strong intention on this point.

We made it this way in our internal release, for it's mainly used in shared clusters. And we aimed to make things simpler for users/admins, and our use case may be much simpler: only physical disks are used for now, and the machines are created in batch batch creation with the same specifications.

Anyway, since we have seen several defects in it as a general solution, I’ll take it as the first choice to manage vgs directly in the doc, and make existing one alternative for more feedback

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Yes, I did lose a lot of details here.

Initially, I thought the doc is of overall design of local storage, but not limited to LVM, so I mainly described changes of the overall structure to coordinate the API side change, and take LVM things just as an illustration(and get we can detail things about LVMs in another doc, maybe in  repo external-storage) 

Anyway, if you guys agree it should be the place to address LVM details, I'll have an update  to detail the workflow :)

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Yes, I didn't describe it clearly, I mean support multi backends in a same plugin, which seems no need :)

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: @msau42 @NicolasT By the way, this way, do you think it necessary to support a list of VGs, or a unique VG should be enough?

If we have more than one VG, then we might meet some edge-cases, for example:
If we have two VGs, VG1 and VG2, they each has 10G left, then the reported capacity would be 20G, but a pod with storage request of 15G could not be satisfied indeed

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: It becomes complicated if we take overcommit into consideration. I'd like to leave it for future releases
@msau42 What's your opinion?

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: > Also: how are FS creation options configured?

fsType and other potential options for FS creation can be specified in StorageClass Parameters,

lichuqiang:
>Shouldn't there be some design explaining how filesystems created on provisioned LVs (which I assume will be necessary, though not yet described) are mounted after a system reboot, such that the CRI can find them and expose them in a Pod/container?

Well, in our internal release, we have an interface named `RecoverMountEntries` to manage the mount state recovery on reboot, I'll detail it if necessary.

But I'm wondering if we could make it easier. I noticed that both in [API](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L1611) and in the [local storage implementation](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/local/local.go#L305), only directory is supported for mount of local storage . Thus, for dynamic provisioned volumes, the provisioner need to mount the LV to a host path, to allow kubelet for another mount, which may not be a good idea. It increases the probability of failure to mount twice.

If we extend the mount source to support block too, the provisioner only need to create LV. `mkfs` and then specify the path of LV as volume path directly, thus, the provisioner doesn't need to care about the mount recovery on reboot too.

lichuqiang:
Updated to address the review comments :)

msau42:
I think it is feasible to update the local volume plugin to detect if the given path is mounted or not, and if not, mount it to some global path.  It could even do the mkfs, although then we have to add fstype to the API too.  This is more similar to the behavior of the other volume plugins.  The only confusing thing now is that the local volume path could refer to a block device path or a mounted fs path with PV.volumeMode=filesystem.  @jsafrane wdyt?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+1. Walks through classes in its `storageClassMap` regularly
+2. Fetch VGs (volume groups) of each class if `storageSources` is specified
+3. Sum up the `VG size`, and compare it with the capacity in StorageClass API object
+4. Update the capacity in StorageClass object if there is a difference
+
+**Note:** For the alpha phase, thinly-provisioned VGs (overcommit) is not supported,
+We'll consider it in a subsequent phase.
+
+Though a list of VGs can be configured as storage source, it's not recommend to
```
msau42: Is there a reason why you want to support more than 1 VG per storage class?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+metadata:
+  name: local-volume-config
+  namespace: kube-system
+data:
+ storageClassMap: |
+    local-static:
+      hostDir: "/mnt/ssds"
+      mountDir: "/local-ssds"
+    local_dynamic:
+      storageSources:
```
msau42: Would prefer lvm to be more explicit in the parameters, instead of as an environment variable.  That way it's very obvious what kind of provisioner it is, and we don't need to worry about defining generic arguments that can be used across multiple implementations.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+Mechanisms to manage local storage sources are varied.
+In addition to current `populator`, `deletor` and `discoverer`, a new component named
+`storageManager` will be introduced in the plugin, to manage the backend storage.
+
+`storageManager` is a general interface, each kind of storage should have its own implementation.
+LVM StorageManager will be introduced in alpha phase, acting as the default StorageManager.
+During start-up, two new environment variables are needed for the provisioner DaemonSet:
+
+* `STORAGE_BACKEND` to determine which StorageManager instance to start
+* `PROVISIONER_NAME` to specify the name of the provisioner, its value matches the Provisioner
```
msau42: Can we define a default to use, or not allow custom names at all?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+##### Change in Populator
+
+In addition to PV, the `populator` will also be extended to watch StorageClasses and PVC objects of all namespaces.
+
+It stores latest StorageClass objects in cache, they are used:
+
+* To reconcile capacity, the StorageManager need to compare actual capacity with that recorded
+in StorageClass objects
+* During volume creation, some opitions (e.g. fsType for `mkfs`) are specified in StorageClasses
```
msau42: Can you document in another section what StorageClass parameters the provisioner can support

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+2. Check if Storage Class of the PVC is in record of its storageClass map.
+If not, return error.
+3. Check if the class is available for dynamic provisioning (i.e. has an input of `storageSources`).
+If not, return error.
+4. Check if there is a VG under the class, whose `Free Size` can satisfy the
+capacity requested in the PVC. If not, return error; otherwise, choose a VG for provisioning.
+5. Create `logical volume (LV)` object in the chosen VG.
+6. Initialize its file system with `mkfs`. FsType and other potential options
+for FS creation can be specified in StorageClass as Parameters
+7. Create a PV API object that is pre-bound to the PVC, and the path
```
msau42: What is the path actually going to be?  Do we need the user to pass in a path where we can safely provision the lvs?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+Currently we already have a component `deleter` to cleanup volumes/PVs of static local volume,
+detail cleanup actions are managed via custom scripts/binaries that specified in `blockCleanerCommand`.
+
+The behavior is extensible enough to support cleanup task of provsioned volumes too.
+So nothing new will be introduced here in the provisioner, but the difference lies
+in the cleanup script/binary itself.
+
+Existing script `quick_reset.sh` calls `mkfs -F` and `wipefs` to cleanup a block,
+For dynamic privisioned volumes, the logical volume in LVM also need to be deleted after that.
```
msau42: Should lv deletion be done directly in the provisioner instead of a script?  Since lv creation is done directly in the provisioner?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+directory and block as a source of mount:
+
+```
+type LocalVolumeSource struct {
+	// The full path to the volume on the node
+	// For alpha, this path must be a directory
+	// Once block as a source is supported, then this path can point to a block device
+	Path string
+	// Type of the volume source, default to Directory
+	Type *LocalVolumeType
```
msau42: I think this will be too confusing because PV.VolumeMode can also be Filesystem or Block.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,213 @@ The central manager will poll each worker for their discovered volumes and creat
...
+(This can happen when a volume has already been provisioned for a PVC,
+but has not been bound to the PVC by the controller yet).
+2. Check if Storage Class of the PVC is in record of its storageClass map.
+If not, return error.
+3. Check if the class is available for dynamic provisioning (i.e. has an input of `storageSources`).
+If not, return error.
+4. Check if there is a VG under the class, whose `Free Size` can satisfy the
+capacity requested in the PVC. If not, return error; otherwise, choose a VG for provisioning.
+5. Create `logical volume (LV)` object in the chosen VG.
+6. Initialize its file system with `mkfs`. FsType and other potential options
```
msau42: Provisioner should also handle PVC.VolumeMode == Block and not format a fs.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Actually no need.
As the config is not limited to LVM, a list could be more general.
But as you suggested below, if we have more explicit parameters for LVM, we can certainly have a clearer limit on it.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Sure

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: I think LVM will create lvs under the path of vg automatically, so the path is determined once a vg is given.
And we just pass the path of the lv into the persistentvolume after that.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: In fact, I haven't thought of other options other than the fsType.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: That makes sense, will update tp add storage manager hook inside the `deleter`

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Ah ,sorry, I overlooked the scenario of BlockVolume here, will remove the API change and rely on local volume plugin to detect its type

lichuqiang:
>I think it is feasible to update the local volume plugin to detect if the given path is mounted or not, and if not, mount it to some global path. It could even do the mkfs, although then we have to add fstype to the API too. This is more similar to the behavior of the other volume plugins. The only confusing thing now is that the local volume path could refer to a block device path or a mounted fs path with PV.volumeMode=filesystem

@msau42 You mean we should also implement `Attacher` interface for local storage?
I'm not sure whether it a good idea to mount twice for a volume(once to the `global path`, once to container path), can we just pass the given path directly to the mounter, and manage the detect job in it?

msau42:
I'm not sure about implementing Attacher.  But the current plugin code assumes that the volume is already mounted globally somewhere.  If we want the plugin to handle the mounting too, then we also need to handle the scenario where multiple pods may want to mount the same volume.  Before, we accomplished this by having two mounts: 1 global mount, and 1 bind mount for each pod.

lichuqiang:
Another update to address the review comments

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,202 @@ The central manager will poll each worker for their discovered volumes and creat
...
+metadata:
+  name: local-volume-config
+  namespace: kube-system
+data:
+ storageClassMap: |
+    local-static:
+      hostDir: "/mnt/ssds"
+      mountDir: "/local-ssds"
+    local-dynamic:
+      storageBackend: "lvm"
```
msau42: you could do something more similar to how volume plugins today are specified in kubernetes, like:
```
local-dynamic:
  lvm:
    volumeGroup: "volumeGroup1"
```

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,202 @@ The central manager will poll each worker for their discovered volumes and creat
...
+      volumeGroup: "volumeGroup1"
+```
+
+A StorageManager instance is responsible for:
+
+1. Monit available capacity for provisioning, and update the capacity in StorageClass if
```
msau42: Monitor

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,202 @@ The central manager will poll each worker for their discovered volumes and creat
...
+there is a change.
+2. Provide the ability to create volumes, detailed in the volume creation section.
```
msau42: Maybe `CreateVolume` should also be an interface?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,202 @@ The central manager will poll each worker for their discovered volumes and creat
...
+detail cleanup actions are managed via custom scripts/binaries that specified in `blockCleanerCommand`.
+
+Existing script `quick_reset.sh` calls `mkfs -F` and `wipefs` to cleanup a block.
+In addition to that, a hook named `volumeDeleteFunc` will be added into `deleter`,
+and for dynamic privisioned volumes, the deleting process should be:
+1. Call scripts/binaries in `blockCleanerCommand` to clean up data in the volume.
+2. Call `volumeDeleteFunc` to recycle the volume behind the PV (e.g. delete the LV for lvm).
+3. Delete the PV object.
+
+##### Expansion of LocalVolumeSource
```
msau42: Call this LocalVolume plugin to be more clear

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,202 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+Therefore, in-tree local storage plugin will be updated to support both directory
+and block as a source of mount. The mount process would be:
+1. Detect if given path in volumeSource is mounted or not, if mounted,
+continue to work in the existing way.
+2. If not, Check if global path `{pluginDir}/mounts/{volumeID}` exists.
+and create the directory if not.
+3. Check if the global path is of mount point, if not, mount given path to
+the global path.
+4. Manage `mkfs` for the global path if it is newly created or mounted.
```
msau42: How will you handle multiple pods trying to mkfs and/or mount the same volume at the same time?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,202 @@ The central manager will poll each worker for their discovered volumes and creat
...
+	Path string
+	// Filesystem type to mount.
+	// Must be a filesystem type supported by the host operating system.
+	// Ex. "ext4", "xfs", "ntfs". Implicitly inferred to be "ext4" if unspecified.
+	// +optional
+	FSType string
+}
+```
+
+**Note:** The change will only work on PersistentVolumes whose volumeMode are "Filesystem",
```
msau42: Can you also add a workflow for how dynamic provisioning of Block volumes will work? (I think it will work with this design as long as PV is created with correct volumeMode)

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,202 @@ The central manager will poll each worker for their discovered volumes and creat
...
+Therefore, in-tree local storage plugin will be updated to support both directory
+and block as a source of mount. The mount process would be:
+1. Detect if given path in volumeSource is mounted or not, if mounted,
+continue to work in the existing way.
+2. If not, Check if global path `{pluginDir}/mounts/{volumeID}` exists.
+and create the directory if not.
+3. Check if the global path is of mount point, if not, mount given path to
+the global path.
+4. Manage `mkfs` for the global path if it is newly created or mounted.
+5. Mount the global path to the path in pods.
```
msau42: Can you also add "unmount" workflow?

contributors/design-proposals/storage/local-storage-pv.md
msau42: I think there were some updates made to the Attacher interface to better support local attaching volume types, that should only be called once per node, instead of once per pod.  I wonder if it could be used here.  @saad-ali

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Well, when creating PV objects, the provisioner will simply specify its volumeMode according to that in the PVC; and for PV of Block mode, the in-tree local storage plugin simply behavior in existing way.

So I think nothing new will be introduced for PV of Block mode

lichuqiang: Maybe I should stress the way to specify the volumeMode of PVs in the workflow of provisioning

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Oh, then the global path tasks should indeed be managed in Attacher/Detacher. Otherwise, it's hard to solve the race condition, and hard to decide when to unmount the global path.

contributors/design-proposals/storage/local-storage-pv.md
msau42: Yeah, I think nothing new needs to be done.  But it may be good to just explicitly show that here for completeness.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Walking through the code, I think the volumeManager process the mounts in order, seems it will not meet a race condition?

Anyway, refer to the behavior of other plugins, it seems reasonable to implement interface `MountDevice`/`UnmountDevice` of Attacher/Detacher to manage the mount of global path.
The downside is we haven't seen real requirement on attach for local storage, so interface like  `Attach`/`VolumesAreAttached`/`WaitForAttach` will be left empty.

By the way, the new `MountDevice` is still in the process of `MountVolume`, @msau42 , can you see race condition this way?

lichuqiang: Another thing to note is that other attachable storages will have a global path for sure, and unmount it in `UnmountDevice` without check.
But for local storage, if given path is a already directory, the global path is not needed, thus in `UnmountDevice`, we might need a basis to check whether to do unmount

msau42: I spoke with @saad-ali and @jingxu97.  If the plugin does not implement the Attach interface, then it is possible for Mount/Unmount to be called concurrently across pods sharing the same volume, so I think we'll have to implement an Attach interface to do the global mount.  However, there is a problem that the Attach call is run on the master, and not on the node.  So some plugins, like iscsi and fc, have implemented their "local attach" in the WaitForAttach call, which is not the cleanest solution.  @saad-ali prefers that we implement a new LocalAttach interface, and eventually we can also migrate other plugins to this interface too.  Some more discussion on local attach is [here](https://github.com/kubernetes/kubernetes/issues/54108)

cc @mlmhl

mlmhl: Yeah, I'm working on this, according doc and PR is #1683  and kubernetes/kubernetes#58810. We add a new `DeviceMounter` interface to do the local attach operation.

However, if we want the local attach operation called once per node, maybe we need to do more work: Currently the `WaitForAttach`, `MountDevice` and `SetUp` operations are all performed inside `operationGenerator.GenerateMountVolumeFunc`, which is called once per pod. One possible solution is we split the `WaitForAttach` and `MountDevice` operation into a single function(for example: `GenerateMountDeviceFunc`) which called once a node(i.e. only called if the volume isn't globally mounted).
@saad-ali  @msau42  Any thoughts?

lichuqiang:
Oh, I think I overlooked an important fact that the local volume provisioner plugin in the community is deployed as a container.
@NicolasT Do you have related experience on handle lvm of the host in a container?
@msau42 we'll also need a way to expose the path of LVs into the provisioner container, so that it can manage the cleanup task before deleting a LV. maybe mount the root path of VGs to the container on start-up?

msau42:
@lichuqiang exposing the lvm path into the container sounds fine.  It would be good to document all the paths required

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -785,8 +785,16 @@ Mechanisms to manage local storage sources are varied.
...
-`storageManager` is a general interface, each kind of storage should have its own implementation.
-LVM StorageManager will be introduced in alpha phase, acting as the default StorageManager.
+According to the input of classes in the configuration, there will be a list of `wrappers`
+under the manager, to communicate with backend storage. `wrapper` is a general interface,
+each kind of storage should have its own implementation. storageManager store the wrappers
+according to storage classes, wrappers of different classes can be different.
+LVM wrapper will be introduced in alpha phase.
+
+```
+Wrappers map[string]wrapper.LocalVolumeWrapper
```
msau42: How about calling it something like "DynamicProvisioningBackend"?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -838,19 +847,12 @@ We'll consider it in a subsequent phase.
...
-
-And for PVCs, when the `populator` finds:
+In addition to PV, the `populator` will also be extended to watch PVC objects of all namespaces.
+When the `populator` finds:
 
 1. An unbound PVC is annotated with `annSelectedNode`, and its node name match the value
-of the annotation
-2. Value of the `annStorageProvisioner` on the PVC matches the name of the provisioner
+of the annotation.
+2. Value of the `annStorageProvisioner` on the PVC matches the name of the provisioner.
```
msau42: Per my latest update of the topology design, we should also add 'annProvisionerTopology' to this list of required fields

contributors/design-proposals/storage/local-storage-pv.md
msau42: If volume manager can coordinate the serialization of MountDevice and MountVolume, then I think getting rid of WaitForAttach is fine.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,208 @@ The central manager will poll each worker for their discovered volumes and creat
...
+The only thing needed is to prepare the disks for provisioning, and configure their paths
+in storage sources.
+
+**Downsides:**
+
+1. The plugin will need to manage the life cycle of the storage resources behind the LVM.
+When some disks are added, removed or failed, the plugin need to process and recover accordingly.
+In addition, the underlying storage resources are varied, they may be  physical disks,
+RAIDs, and so on, so it's hard to cover all the scenes.
+2. As provisioners running on all nodes shall a same configuration ConfigMap,
```
msau42: s/shall/share

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: I think once `annSelectedNode` is set, the topology of the node should have been measured by the scheduler.
So here we need to make a double check to prevent the topology label of the node from being changed midway?

msau42: The check for `annProvisionerTopology` is to handle the coordination between provisioners that need to report capacity and initialization of the StorageClass.  If `annProvisionerTopology` was not set, then that means the scheduler made a decision not taking into account capacity.

msau42: Can the removal of the annotation be one level higher so that it's common across all provisioner implementations?

msau42: Yes, what I meant was the "general" storagemanager layer, not the lvm specific layer.

lichuqiang: sure thing :)

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,208 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+Therefore, in-tree local storage plugin will be updated to support both directory
+and block as a source of mount. The mount process would be:
+1. Detect if given path in volumeSource is mounted or not, if mounted,
+continue to work in the existing way.
+2. If not, Check if global path `{pluginDir}/mounts/{volumeID}` exists.
+and create the directory if not.
+3. Check if the global path is of mount point, if not, mount given path to
+the global path.
+4. Manage `mkfs` for the global path if it is newly created or mounted.
```
jsafrane: > 4. Manage `mkfs` for the global path

You must call `mkfs` before mounting it. Internal volume plugins do it even before step 1.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,208 @@ The central manager will poll each worker for their discovered volumes and creat
...
+  name: local-volume-config
+  namespace: kube-system
+data:
+ storageClassMap: |
+    local-static:
+      hostDir: "/mnt/ssds"
+      mountDir: "/local-ssds"
+    local-dynamic:
+      lvm:
+        volumeGroup: "volumeGroup1"
```
jsafrane: Who creates the volume group? Is it something that admin must do during installation?

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Yah, have not got a chance the update the section.
As discussed above, we need to consider this part along with the change in  #1683

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: yeah, sorry the discussion got flooded
> IMHO it's problematic if the provisioner is responsible for this. As mentioned below, disk naming etc could be problematic. Instead, I believe storage pool provisioning is a responsibility of the cluster operator (using e.g. Ansible or whatever), only volume provisioning should be performed by the Kubernetes infrastructure.

> As such, in the LVM case: the Kubernetes provider creates and manages LVs in a given and pre-existing VG, an admin creates PVs and VGs using whichever mechanism seems suitable.

> Agree, there are too many ways an admin can configure and manage vgs. It will be hard to support all of them.

> This will also simplify the issue where all the disks on every node must be named the same.

@msau42 and @NicolasT agree it reasonable to pass in existing groups as input.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Discussed with @mlmhl, [the refactor](https://github.com/kubernetes/kubernetes/pull/58810) does not cover the change in the process of `operationGenerator`, maybe we can raise another issue to address the race scenario?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,208 @@ The central manager will poll each worker for their discovered volumes and creat
...
+5. Mount the global path to the path in pods.
+
+New field `FSType` will be added in `LocalVolumeSource` to support `mkfs` in the mounter.
+
+```
+type LocalVolumeSource struct {
+	// The full path to the volume on the node
+	// For alpha, this path must be a directory
+	// Once block as a source is supported, then this path can point to a block device
+	Path string
```
msau42: I think the confusing part will be that there are 3 different behaviors for path depending on the underlying type:

1. Volume mode = FS, path is a FS.  We bind mount the path to pod directory
2. Volume mode = FS, path is a device. We mkfs if needed, and globally mount the device.  Then bind mount global path to pod directory.
3. Volume mode = block, path is a device. Passthrough.

I wonder if we could make this behavior more explicit if we supported true "fs-on-block" volume mode binding (ie allow PVC volumeMode fs to bind to PV volumeMode block if the plugin supports it).  I know it was initially rejected for simplicity, but it really does make the local volume behavior a bit confusing.

@jsafrane would appreciate your thoughts as you originally had strong objections to the overloading of the Path field in the initial design.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: It's indeed a common case, but a new interface seems duplicated to exsiting `MountDevice` interface. I think we can reuse the  interface to manage local attach. It's the fact that the interface included in `Attacher` that confusing us.
That is also what https://github.com/kubernetes/kubernetes/pull/58810 intend to address.

After that, we need a further step to make funcs in `DeviceMounter` called per node I think.
@msau42 @mlmhl WDYT?

msau42: I discussed this with @jingxu97.  We think the best approach would be to split this into 2 phases:

1. Separate Attach calls from MountDevice.  Handled by #1683 
2. Move Attach + MountDevice to outside the MountVolume operation.  This is a much bigger change and would impact all volume plugins, so needs to be very carefully considered.

So for now, we can implement MountDevice.  It's going to get called per pod, however, we must make sure operation executor will serialize the MountVolume operation per volume (right now it's only done for attachable plugins).

lichuqiang: sounds good to me.
>  It's going to get called per pod, however, we must make sure operation executor will serialize the MountVolume operation per volume (right now it's only done for attachable plugins).

Yah, maybe improve `pendingOperations` in operation executor to also keeps track of mount operations.
Anyway, I think we should open another  issue to address it. and what we should describe here is after #1683 in, how we should implement related interfaces in local volume plugin to support both block and file path as volume source.

contributors/design-proposals/storage/local-storage-pv.md
jsafrane: I am fine with that, however please note that explicitly in the proposal.

contributors/design-proposals/storage/local-storage-pv.md
jsafrane: Yeah, listen to me :-). It *is* confusing, however it's for separate PR on block storage.

With regard to local storage, I expect that "path is a FS" should be used in very special cases where underlying platform can't provide us block devices.

"path is a device" is what we should message to users, with a small note somewhere below that it might be a directory if it's really needed.

msau42: Thanks for your input. So IIUC, we can propose the volumeMode binding change separately, and go with this approach for now to be unblocked.  It also means for static discovery, we need a new parameter to indicate if block device should be volumeMode fs or block.

I really want to go with the explicit volumeModes approach before we make this behavior beta though.  Otherwise we'll still have to support both ways.

lichuqiang: So for now, besides `Path` and `FsType`, we'll need another field like `SourceType` in the LocalVolumeSource, to indicate its type, is it?

Another notable thing it that currently interface UnmountDevice looks like this:
```
UnmountDevice(deviceMountPath string) error
```
if only the interface is implemented, unmount will be managed for sure, but if the source path is of fs, we don't need global mount/unmount indeed.
So maybe we should add extra parameter to the interface for check?

lichuqiang: or in the implementation `UnmountDevice` for local storage, we can have a internal agreement on the format of global mount path, and basing on the format of given deviceMountPath, we can decide whether global unmount is needed or not.

msau42: I don't think we need a `SourceType` in the LocalVolumeSource.  I meant we need a new field in `MountConfig` of static provisioner. Right now, it assumes if the device is a block device, then PV.volumeMode is Block.

I wouldn't change the `UnmountDevice` interface, we can just check internally.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+`storageManager` will be introduced in the plugin, to manage the backend storage.
+
+According to the input of classes in the configuration, there will be a list of `backends`
+under the manager, to communicate with backend storage. `backend` is a general interface,
+each kind of storage should have its own implementation. storageManager store the backends
+according to storage classes, backends of different classes can be different.
+LVM backend will be introduced in alpha phase.
+
+```
+Backends map[string]backend.DynamicProvisioningBackend
```
msau42: How about renaming "storageManager" to "dynamicProvisioningManager", then "backends" could be "storageBackends"

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+**Note:** For the alpha phase, thinly-provisioned VG (overcommit) is not supported,
+We'll consider it in a subsequent phase.
+
+##### Change in Populator
+
+In addition to PV, the `populator` will also be extended to watch PVC objects of all namespaces.
+When the `populator` finds:
+
+1. An unbound PVC is annotated with `annSelectedNode`, and its node name match the value
```
msau42: Also needs to watch for `annSelectedTopology`

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+      mountDir: "/local-ssds"
+    local-dynamic:
+      lvm:
+        volumeGroup: "volumeGroup1"
+      blockCleanerCommand: "./quick_reset.sh"
+```
+
+A StorageManager instance is responsible for:
+
+1. Monitor available capacity for provisioning, and update the capacity in StorageClass if
```
msau42: Should capacity reporting be an interface as well, so we can implement a generic capacity reconciler?

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+1. Create volumes behind a PV accordingly
+2. Create a PV object that is pre-bound to the PVC
+
+The proposed workflow is:
+
+1. Check if there is already a PV pre-bound to the PVC, if so, return directly
+(This can happen when a volume has already been provisioned for a PVC,
+but has not been bound to the PVC by the controller yet).
+2. Check if `annProvisionerTopology` is set on the PVC. If not, then it means
```
msau42: I think this could be done one level up

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+2. Call `volumeDeleteFunc` to recycle the volume behind the PV (e.g. delete the LV for lvm).
+3. Delete the PV object.
+
+##### Expansion of Local Volume Plugin
+
+Currently, Intended `VolumeMode` of a local volume is determined by the state of
+the input path. If a user need a local volume whose `VolumeMode` is filesystem,
+the path provided in `LocalVolumeSource` must be a directory. Thus, for dynamic
+provisioned volumes, the provisioner need to mount the LV to a global path first,
+to allow kubelet for another mount. It increases the complexity of the plugin.
```
msau42: increases complexity of the provisioner, and makes handling reboot scenarios difficult.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+Therefore, in-tree local storage plugin will be updated to support both directory
+and block as volume source.
```
msau42: When the specified Path is detected to be a block device and the volumeMode is filesystem, the plugin will create the filesystem (on the first mount) and globally mount the device to the plugin path.  This is similar to how other volume plugins handle filesystem on block.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+type when formating a disk input path:
+
+```
+type LocalVolumeSource struct {
+	// The full path to the volume on the node
+	// The path can be a directory or point to a block device
+	Path string
+	// Filesystem type to mount.
+	// Must be a filesystem type supported by the host operating system.
+	// Ex. "ext4", "xfs", "ntfs". Implicitly inferred to be "ext4" if unspecified.
```
msau42: Default is "ext4"

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+	// +optional
+	FSType string
```
msau42: This needs to be a pointer because it's optional

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+in their claims.
+
+##### Alternatives
+
+###### Take Disk Partitions as Storage Source of LVM
+
+Instead of existing volume groups, we can take disk partitions as storage source.
+This way, the provisioner is in charge of underlying storage.
+
+**Advances:**
```
msau42: Advantages

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+Cluster admins do not need to make extra preparations and maintenance work.
+The only thing needed is to prepare the disks for provisioning, and configure their paths
+in storage sources.
+
+**Downsides:**
+
+1. The plugin will need to manage the life cycle of the storage resources behind the LVM.
+When some disks are added, removed or failed, the plugin need to process and recover accordingly.
+In addition, the underlying storage resources are varied, they may be  physical disks,
+RAIDs, and so on, so it's hard to cover all the scenes.
```
msau42: scenes -> scenarios

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,239 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+##### Volume Creation
+
+During start-up of the provisioner, a new work queue `ProvisionQueue` will be created,
+and passed into both `populator` and `storageManager`:
+
+```
+ProvisionQueue *workqueue.Type
+```
+When it finds a PVC is in need of privsioning, as desicribed above,
```
msau42: provisioning
described

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: you mean in the populator?
In the populator we only filter incoming pvcs, but for this case, we'll need to treat it as "failure" and signal back to the scheduler to retry dynamic provisioning by removing the "annSelectedNode" annotation of the PVC. So I think it should be better to place it here in the manager.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: Good idea.
It is better decoupling this way, though there is not a lot of content in the reconciler for now.

contributors/design-proposals/storage/local-storage-pv.md
msau42: I mean in the layer that is managing whether or not to call CreateVolume(). The main question I ask myself for every operation is "Is this something that is specific to the LVM implementation?"

lichuqiang: Have clarified to @msau42, `CreateVolume` here is a generic interface and not specific to the LVM.

contributors/design-proposals/storage/local-storage-pv.md
msau42: Can you add a section about the LVM details, and cover some things like:
* the provisioner arguments
* any hostpath volumes lvm requires
* what it will do for each of the backend interfaces
* format of the device name when lv is created
* how to handle lv naming with multiple vgs (multiple storage classes)
* how reboot/restart is handled

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,248 @@ The central manager will poll each worker for their discovered volumes and creat
...
+and makes handling reboot scenarios difficult.
+
+Therefore, in-tree local storage plugin will be updated to support both directory
+and block as volume source. When the specified Path is detected to be a block device
+and the volumeMode is filesystem, the plugin will create the filesystem (on the first
+mount) and globally mount the device to the plugin path. This is similar to how other
+volume plugins handle filesystem on block.
+
+Specifically, we will implement the following interfaces, to manage the global mount
+when volume source is of block:
```
NickrenREN: So you mean, implement the `Attacher` & `Detacher` for local volume plugin ?

lichuqiang: Nope, I think after https://github.com/kubernetes/kubernetes/pull/58810 merged, the interfaces will be separate from Attacher/Detacher

NickrenREN: ok, got it. thanks. Will take local plugin change task when the refactor is done.

contributors/design-proposals/storage/local-storage-pv.md
lichuqiang: done

KlavsKlavsen:
How is this faring.. does it work? I am looking for something to dynamicly deploy disks for pod's.. and I would be very happy with the "simple" approach og creating a config-map defining a VolumeGroup (LVM VG) for it to use to assign from - and then it would simply look at how many extents were available (compared to the storage requirement of the pod) - and figure out if it had room the local pod storage or not... 
ie. binding dynamic storage to LVM.. but that would IMHO be a good approach - and LVM do support stuff like dmcache (ssd 'hot block' write-back caching - greatly increases IOPS without impeeding disk size) and other things.

lichuqiang:
> How is this faring.. does it work?
Not really, see https://github.com/kubernetes-incubator/external-storage/pull/743#issuecomment-419917613

cpfeiffer:
One thing is not clear to me from reading the design document. Our use case is a build cluster for Jenkins jobs, where a workload (job) would be scheduled on *any node* in the cluster (depending on resource constraints). The workload should get a local persistent volume, which is either already available on the node, or dynamically created. It is not desired to schedule the same workload always on the same node (e.g. when it is already overloaded), just because its local PV lives there.

IOW: the workload should be scheduled to a node by resource constraints and then get a dynamically provisioned local PV on that node, even if another node already has a PV for it.

msau42:
@cpfeiffer do the pods have anti-affinity for nodes?  If so, then this design will handle that.  Even if there is a local PV already available on node A, if there is a Pod A already running there and your Pod B has anti-affinity to Pod A, then we will not choose the PV that's already on Node A

cpfeiffer:
@msau42 No, I mean just Pod A that should be scheduled to either Node A or Node B depending on resource availability. If Pod A stops on Node A (= Jenkins build job has finished), then it should be possible to get Pod A restarted on Node B when the next build is started.

It's not necessary to have the same volume contents (= build artifacts etc.) on Node A and Node B. The container will make sure it gets all that is needed. We just want to achieve that it won't have to retrieve *everything* again and again, while being able to use fast local storage.

lichuqiang:
> It is not desired to schedule the same workload always on the same node (e.g. when it is already overloaded), just because its local PV lives there.

It's an storage mechanism issue, but not limited to local pv.
during volume binding, we'll try to find available pv for a pvc first, and will not trigger dynamic provision if there exists available pv, even if its node is already of high load.
And for your case, if you expect pod to be scheduled to node B, make sure there are available pvs on the node as well

cpfeiffer:
@lichuqiang OK, if the PVs need to be available all nodes in advance, would the dynamic provisioner be able to create them on all nodes at once?

lichuqiang:
With existing static discovery pattern, PV will be created once you expose volumes to the plugin;
And for the upcoming dynamic provisioning pattern, as I described above, we'll not trigger provision if there exists available pv.

So for your case, you can
1. pre-create PVs across the nodes (static discovery pattern makes sense for this)
2. ensure no unbound PVs left in the cluster, and rely on the dynamic plugin to provision on demand.

msau42:
Hi @cpfeiffer it sounds a little bit like you don't need the persistence part of the feature, ie you want to be sticky to the data.

If all you need is ephemeral local storage, then you can use emptydir

kfox1111:
hmm.... actually, thinking about it a little bit, while emptyDir seems to be a good fit for @cpfeiffer's stated issue, I think there is one case it doesn't fit. The dynamic provisioning mentioned here can allocate local volumes as needed in order to not over allocate host storage. emptyDir currently doesn't guarantee an amount of storage. Maybe emptydir should be reimplemented on top of local dynamic storage, so that it can reserve an amount of storage for its lifecycle? or alternately, maybe a second, sizedEmptyDir (localEmptyDir?) driver is added that does this instead.

msau42:
You can request a limit on emptydir size now, and I believe we do soft eviction if you exceed.  I think there are plans to support project quota for hard enforcement too.

The other use case that @cpfeiffer mentioned was to avoid having to load data into the volume everytime, but still have the flexibility to schedule on different nodes.  In that case, then hostpath may be the simplest solution here.

kfox1111:
ah, I didn't realize it did that yet. thanks for the info. :)

cpfeiffer:
@msau42 emptyDir doesn't fit us because we want to keep the contents for longer than just a single run.

And indeed hostpath is the best we figured out so far. I just thought that dynamic provisioning might save us from manually provisioning the host volumes.

lichuqiang:
Update to list the thoughts inflight, to avoid readers being misled by the old information. It's far from a complete deisgn, but more of rough thoughts on which we hope to receive feedback from readers.

contributors/design-proposals/storage/local-storage-pv.md
```diff
@@ -761,3 +761,38 @@ The central manager will poll each worker for their discovered volumes and creat
...
+
+Note: The origin design that based on existing static discovery plugin is out-of-date. And currently the thing is blocked by the capcity report issue. Here I list the thoughts inflight, to avoid readers being misled by the old information. It's far from a complete deisgn, but more of rough thoughts on which we hope to receive feedback from readers.
+
+##### Overall Thoughts
+
+As we'll gradually switch all of the storages to CSI, we plan to construct a CSI driver for local storage, which will also be responsible for dynamically provisioning of local volumes. Users can still use the existing local storage plugin in external-storage repo to maintain local volumes in the "static way". The two plugins are completely independent.
+
+##### CSI Driver Details
+
+Different from other storages, we'll need to deploy a set of CSI provisioner on each of the nodes. We'll rely on feature volumeScheduling as a dependency. For local storage, the provisioner is only responsible for PVCs marked with "related plugin name + selected node annotation". That is, it only privision the pvcs that "scheduled" to its node.
```
mlmhl: Will all these provisioners list/watch all PV/PVCs? If so, maybe we need to take care of the pressure on apiserver/etcd.

lichuqiang: The existing static discovery plugin also behavior like this. So I suppose this acceptable.

KlavsKlavsen:
Hows this progressing.. sure would be nice with automatic LVM provisioning for local storage :)

PR #1915, opened 2018-03-09
KEP: kubeadm config HA support

mattkelly:
This KEP elaborates on the approach suggested by @fabriziopandini in https://github.com/kubernetes/kubeadm/issues/706#issuecomment-368068828 to augment the existing `kubeadm-config` with additional master-specific ConfigMaps in order to enable HA upgrades in the future.

An alternative approach (elaborated on in the KEP) would be to augment the existing `kubeadm-config` directly without the use of additional ConfigMaps. The multiple ConfigMap approach makes more sense to me so that's what I decided to propose. If there is a lot of opposition to this approach, then I'm happy to draft a new KEP.

I will also work on prototyping this as we agreed. I just wanted to get something out there to get a more serious discussion moving.

cc @timothysc @luxas @fabriziopandini @stealthybox @mbert

### DIFF
--- a/keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
+++ b/keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
@@ -1,0 +1,164 @@
+---
+kep-number: TBD
+title: Augment Kubeadm Config to Enable Upgrades of HA Clusters
+authors:
+  - "@mattkelly"
+owning-sig: sig-cluster-lifecycle
+reviewers:
+  - "@timothysc"
+  - "@fabriziopandini"
+approvers:
+  - TBD
+editor:
+  - "@mattkelly"
+creation-date: 2018-03-08
+last-updated: 2018-04-09
+status: provisional
+see-also:
+  - [Creating HA clusters with kubeadm](https://kubernetes.io/docs/setup/independent/high-availability/)
+  - [KEP kubeadm join --master workflow](https://github.com/kubernetes/community/pull/1707) (in progress)
+  - [Upgrading kubeadm HA clusters from 1.9.x to 1.9.y](https://kubernetes.io/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-ha/)
+---
+
+# Augment Kubeadm Config to Enable Upgrades of HA Clusters
+
+## Table of Contents
+
+* [Augment Kubeadm Config to Enable Upgrades of HA Clusters](#augment-kubeadm-config-to-enable-upgrades-of-ha-clusters)
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+   * [Goals](#goals)
+   * [Non-Goals](#non-goals)
+   * [Challenges and Open Questions](#challenges-and-open-questions)
+* [Proposal](#proposal)
+   * [Implementation Details](#implementation-details)
+      * [Background](#background)
+      * [Adding Additional Master-Specific ConfigMaps](#adding-additional-master-specific-configmaps)
+         * [Key Design Considerations and Benefits](#key-design-considerations-and-benefits)
+            * [Parallel Node Creation](#parallel-node-creation)
+            * [Guaranteed Consistent kubeadm-config](#guaranteed-consistent-kubeadm-config)
+   * [Risks and Mitigations](#risks-and-mitigations)
+      * [Migrating Existing Clusters](#migrating-existing-clusters)
+      * [More Complex User Experience for Overriding Configuration](#more-complex-user-experience-for-overriding-configuration)
+* [Implementation History](#implementation-history)
+* [Drawbacks](#drawbacks)
+* [Alternatives](#alternatives)
+
+## Summary
+
+One of the first steps of the upgrade process is to retrieve the `kubeadm-config` ConfigMap that is stored at `kubeadm init` time in order to get any relevant cluster configuration.
+The current `kubeadm-config` ConfigMap was designed to support upgrades of clusters with a single master.
+As we move towards supporting High Availability (HA) natively in kubeadm, the persistent configuration must be enhanced in order to store information unique to each master node.
+For example, master node-specific information is required in order for kubeadm to identify which control plane static pods belong to the current node during an upgrade (the static pods are identifiable by the `nodeName` field embedded in the pod name).
+If a kubeadm HA cluster is created today using the [workarounds recommended by the community](https://github.com/kubernetes/kubeadm/issues/546), then each subsequent `kubeadm init` on a master node will overwrite the master node-specific information from the previous `init`, which is not desired.
+
+This KEP outlines a possible solution for adding and retrieving master node-specific information through the use of additional kubeadm ConfigMaps.
+
+## Motivation
+
+Kubeadm is driving towards natively supporting highly available clusters.
+As part of HA support, a clean upgrade path is required.
+The purpose of this KEP is to introduce support for multiple masters in the kubeadm configuration that is stored in-cluster in order to enable that clean upgrade path.
+
+### Goals
+
+Enable `kubeadm upgrade` of highly available clusters by augmenting the existing persistent kubeadm configuration.
+
+### Non-Goals
+
+This proposal does not aim to solve the entire problem of upgrading HA clusters.
+This KEP specifically tackles the persistent configuration problem so that the information required at upgrade time is available.
+
+### Challenges and Open Questions
+
+The final implementation of this KEP will require deciding exactly what "master node-specific information" means.
+Currently, the `nodeName` and `advertiseAddress` of the master are two entries known to be node-specific.
+It may be possible that additional config entries could be split out into the node-specific area(s) of the config.
+This could result in asymmetric configuration across the masters, which may or may not be something that we wish to support.
+
+## Proposal
+
+### Implementation Details
+
+#### Background
+
+Currently, the `kubeadm-config` ConfigMap in the `kube-system` namespace serves as the single source of truth for how kubeadm has been used to create and modify a cluster.
+Because kubeadm is not a process that runs on the cluster (it is only run to perform operations, e.g. `init` and `upgrade`), this config is not modified during normal operation.
+In the non-HA case today, it is guaranteed to be an accurate representation of the kubeadm configuration.
+
+If kubeadm is used to create an HA cluster today, e.g. using the workarounds described in [kubeadm #546](https://github.com/kubernetes/kubeadm/issues/546) and/or @mbert's [document](https://docs.google.com/document/d/1rEMFuHo3rBJfFapKBInjCqm2d7xGkXzh0FpFO0cRuqg), then the `kubeadm-config` ConfigMap will be an accurate representation except for any master node-specific information.
+As explained in [Challenges and Open Questions](#challenges-and-open-questions), such node-specific information is not yet well-defined but minimally consists of the master's `nodeName` and `advertiseAddress`.
+The `nodeName` in `kubeadm-config` will correspond to the last master that happened to write to the ConfigMap.
+In the case of parallel node creation, this may not be well-defined.
+When `kubeadm upgrade` is run on a master and this `nodeName` is fetched, it may be incorrect and the upgrade process will fail.
+
+#### Adding Additional Master-Specific ConfigMaps
+
+The proposed solution is to add additional kubeadm ConfigMaps that are specific to each master (one ConfigMap for each master).
+Each master-specific ConfigMap will be created as part of the `kubeadm init` process for the initial master and as part of the to-be-implemented [`kubeadm join --master` process](https://github.com/kubernetes/community/pull/1707) for additional masters.
+Any master-specific information in the main `kubeadm-config` ConfigMap will be removed.
+Each master-specific ConfigMap can be automatically deleted via garbage collection (see [below](#guaranteed-consistent-kubeadm-config) for details).
+
+The names of these new ConfigMaps will be `kubeadm-config-<machine_UID>` where `machine_UID` is an identifier that is guaranteed to be unique for each node in the cluster.
+There is a precedent for using such a `machine_UID`, and in fact kubeadm already has a [prerequisite](https://kubernetes.io/docs/setup/independent/install-kubeadm/#verify-the-mac-address-and-product_uuid-are-unique-for-every-node) that such machine identifiers be unique for every node.
+For the purpose of this KEP, let us assume that `machine_UID` is the full `product_uuid` of the machine that the master node is running on.
+
+Kubeadm operations such as upgrade that require master-specific information should now also retrieve the corresponding ConfigMap for their node.
+This master-specific configuration will be explicitly provided to any functions that require it instead of e.g. merging everything into one configuration.
+
+##### Key Design Considerations and Benefits
+
+There are a few key benefits to the approach of adding additional ConfigMaps over an approach which would augment the existing `kubeadm-config` with master-specific information:
+
+###### Parallel Node Creation
+
+Node creation in parallel is a valid use-case that works today.
+By adding additional ConfigMaps instead of requiring each master to modify the existing `kubeadm-config`, we avoid the need to lock on that ConfigMap.
+
+###### Guaranteed Consistent kubeadm-config
+
+This approach allows us to continue to guarantee that the main `kubeadm-config` is consistent with the actual cluster configuration.
+If we put master-specific information into `kubeadm-config` itself, then we would require either a yet-to-be-defined `kubeadm leave` workflow or active reconciliation of `kubeadm-config` in order to ensure accurateness.
+This may not be critical, but it is a consideration.
+
+With this proposal, if a node unexpectedly leaves a cluster, then at worst a dangling ConfigMap will be left in the cluster.
+For the case where a node is explicitly deleted, we can leverage garbage collection to automatically delete the master-specific ConfigMap by listing the node as an `ownerReference` when the ConfigMap is created.
+
+### Risks and Mitigations
+
+#### Migrating Existing Clusters
+
+There will be situations in which a kubeadm operation (e.g. upgrade) that requires the new master-specific ConfigMap is run and finds that the expected ConfigMap does not exist.
+For example, this will happen for users who are upgrading HA clusters that were created using the aforementioned workarounds required before kubeadm HA support is available.
+We can automate the creation of missing node-specific ConfigMaps in the following manner when a `kubeadm upgrade` (or other operation requiring it) is performed:
+
+1. Determine which node kubeadm is currently running on by listing all master nodes in the cluster and looking at the existing `product_uuid` field
+2. Get the `nodeName` from this node's metadata
+3. Compare the current `nodeName` to the `nodeName` in `kubeadm-config`
+4. If they are equal, update `kubeadm-config` to remove node-specific information (to match the new `kubeadm-config` specification) 
+5. Create the `kubeadm-config-<machine_UID>` ConfigMap for this node
+6. Continue the upgrade process
+
+#### More Complex User Experience for Overriding Configuration
+
+Currently, users may override configuration items by providing a configuration file when running kubeadm.
+The existence of additional, disjoint ConfigMaps may make the user experience more complex for overriding configuration.
+One possibility for mitigating this would be to keep the `kubeadm-config` specification the same as it is today instead of removing fields.
+This would allow a user to specify any node-specific information in the same configuration file instead of having to provide multiple files.
+Placing this information in the appropriate node-specific ConfigMap would be an implementation detail not requiring any impact to user experience.
+
+## Implementation History
+
+- [Issue #546: Workarounds for the time before kubeadm HA becomes available](https://github.com/kubernetes/kubeadm/issues/546)
+- [Adding HA to kubeadm-deployed clusters](https://docs.google.com/document/d/1rEMFuHo3rBJfFapKBInjCqm2d7xGkXzh0FpFO0cRuqg)
+- [Issue #706: Make kubeadm upgrade HA ready](https://github.com/kubernetes/kubeadm/issues/706)
+
+## Drawbacks
+
+This KEP introduces additional ConfigMaps for kubeadm to use (one for each master).
+
+## Alternatives
+
+An alternative approach would be to augment the existing `kubeadm-config` ConfigMap with master-specific information.
+The advantages over this approach are detailed in the [Proposal](#proposal) section.


keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+With this proposal, if a node unexpectedly leaves a cluster, then at worst a stale ConfigMap will be left in the cluster.
+For the case where a node is explicitly deleted, we can leverage garbage collection to automatically delete the master-specific ConfigMap by listing the node as an `ownerReference` when the ConfigMap is created.
+
+### Risks and Mitigations
+
+There will be situations in which a kubeadm operation (e.g. upgrade) that requires the new master-specific ConfigMap is run and finds that the expected ConfigMap does not exist.
+For example, this will happen for users who are upgrading HA clusters that were created using the aforementioned workarounds required before kubeadm HA support is available.
+In this case, there are several options:
+
+1. Kubeadm can fall back to looking at the main `kubeadm-config`.
```
timothysc: Can't you just:
1. scan for existing non_UID configs 
2. read in and check local name = name in the config, if so delete old and drop new. 
3. if not, create new one.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+   The user would have to manually modify `kubeadm-config` for each master to set the `nodeName` to the current master.
+   This is the [recommended workaround](https://github.com/kubernetes/kubeadm/issues/546#issuecomment-365063404) today.
+
+2. Kubeadm can provide an additional command similar to (or an extension of) the existing [`kubeadm config upload`](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-config/) command.
```
timothysc: I think we can automate it.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+  - TBD
+approvers:
+  - TBD
+editor:
+  - "@mattkelly"
+creation-date: 2018-03-08
+last-updated: 2018-03-09
+status: provisional
+see-also:
+  - [KEP kubeadm join --master workflow](https://github.com/kubernetes/community/pull/1707) (in progress)
```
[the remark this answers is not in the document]
mattkelly: Added

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+
+Kubeadm operations such as upgrade that require master-specific information should now also grab the corresponding ConfigMap for their node.
+
+##### Key Design Considerations and Benefits
+
+There are a few key benefits to the approach of adding additional ConfigMaps over an approach which would augment the existing `kubeadm-config` with master-specific information:
+
+###### Parallel Node Creation
+
+Node creation in parallel is a valid use-case that works today.
```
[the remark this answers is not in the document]
mattkelly: Great point - fixed.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+If we required that each master modify `kubeadm-config`, then there may be concern around concurrent modification of this ConfigMap.
+
+###### Guaranteed Consistent kubeadm-config
+
+This approach allows us to continue to guarantee that the main `kubeadm-config` is consistent with the actual cluster configuration.
+If we put master-specific information into `kubeadm-config` itself, then we would require either a yet-to-be-defined `kubeadm leave` workflow or active reconciliation of `kubeadm-config` in order to ensure accurateness.
+This may not be critical, but it is a consideration.
+
+With this proposal, if a node unexpectedly leaves a cluster, then at worst a stale ConfigMap will be left in the cluster.
+For the case where a node is explicitly deleted, we can leverage garbage collection to automatically delete the master-specific ConfigMap by listing the node as an `ownerReference` when the ConfigMap is created.
```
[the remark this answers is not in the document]
mattkelly: I added a note above about deletion as well, referencing this section. I may further restructure this to make it flow better.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
+---
+kep-number: TBD
+title: Augment Kubeadm Config to Enable Upgrades of HA Clusters
+authors:
+  - "@mattkelly"
+owning-sig: sig-cluster-lifecycle
+reviewers:
+  - TBD
```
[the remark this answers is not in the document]
fabriziopandini: + me

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+              * [Parallel Node Creation](#parallel-node-creation)
+              * [Guaranteed Consistent kubeadm-config](#guaranteed-consistent-kubeadm-config)
+     * [Risks and Mitigations](#risks-and-mitigations)
+  * [Implementation History](#implementation-history)
+  * [Drawbacks](#drawbacks)
+  * [Alternatives](#alternatives)
+
+## Summary
+
+Currently, `kubeadm upgrade` of a master in a multi-master cluster will fail without workarounds because of a lack of node-specific master configuration in the `kubeadm-config` ConfigMap that is created at `init` time and later referenced during an upgrade.
```
fabriziopandini: IMO it is a little bit unfair saying that the current implementation will fail (it hints a wrong implementation)
What about rephrasing this a little bit and telling that the kubeadm-config-machineUID was originally designed for supporting upgrades in single-master clusters, and now, As we move towards supporting HA natively in kubeadm, this design should be improved?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+As explained in [Challenges and Open Questions](#challenges-and-open-questions), such node-specific information is not yet well-defined but minimally consists of the master's `nodeName`.
+The `nodeName` in `kubeadm-config` will correspond to the last master that happened to write to the ConfigMap.
+In the case of parallel node creation, this may not be well-defined.
+When `kubeadm upgrade` is run on a master and this `nodeName` is fetched, it may be incorrect and the upgrade process will fail.
+
+#### Adding Additional Master-Specific ConfigMaps
+
+The proposed solution is to add additional kubeadm ConfigMaps that are specific to each master (one ConfigMap for each master).
+Each master-specific ConfigMap will be created as part of the to-be-implemented [`kubeadm join --master` process](https://github.com/kubernetes/community/pull/1707).
```
fabriziopandini: I guess that also kubeadm init should create a configMap for the first master...

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,147 @@
...
+Any master-specific information in the main `kubeadm-config` ConfigMap will be removed.
+
+The names of these new ConfigMaps will be `kubeadm-config-<machine_UID>` where `machine_UID` is an identifier that is guaranteed to be unique for each node in the cluster.
+There is a precedent for using such a `machine_UID`, and in fact kubeadm already has a [prerequisite](https://kubernetes.io/docs/setup/independent/install-kubeadm/#verify-the-mac-address-and-product_uuid-are-unique-for-every-node) that such machine identifiers be unique for every node.
+For the purpose of this KEP, let us assume that `machine_UID` is the full `product_uuid` of the machine that the master node is running on.
+
+Kubeadm operations such as upgrade that require master-specific information should now also grab the corresponding ConfigMap for their node.
```
fabriziopandini: Just for checking if I got this right:
- kubeadm init will continue to use the current master Configuration struct (with both cluster info and master specific info)
- when creating the kubeadm configMap, master specific info will be stripped down and stored in a separated configMap named using the machine UID
- when joining a new master node, an additional master specific configMap must be created for the joining machine UID
- when upgrading, the master specific configMap with the current machine UID, and merge d together with the kubeadm-- config configMap into the current Master Configuration struct
Is that right?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
fabriziopandini: +1 
We should ensure an automatic upgrade flow for single--master nodes
For HA cluster, created with manual workaround, IMO It is acceptable to document  some manual step as well

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: I agree that this could be misinterpreted - I will reword it along the lines of your suggestion.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: I was picturing this slightly differently - I think my wording is unclear and I will have to edit it to clarify:

- The current `MasterConfiguration` struct type (which becomes the `kubeadm-config` ConfigMap) will be modified to remove master-specific entries (`NodeName` definitely, others TBD)
- A new struct type `MasterNodeConfiguration` (don't love the name, still thinking about it in POC implementation) and corresponding `kubeadm-config-<UID>` will be defined to contain anything stripped from `MasterConfiguration` (minimally `NodeName`, others TBD)
- `kubeadm init` will remain as-is today except it will additionally create the `kubeadm-config-<UID>` for the current node
- `kubeadm join --master` (when it exists) will reference `kubeadm-config` but it will not modify it - it will only create the `kubeadm-config-<UID>` for the current node
- `kubeadm upgrade` will reference both `kubeadm-config` and `kubeadm-config-<UID>` for this node, however no merging will be performed - it will simply look at both structs

Does this make sense? If so, I will update accordingly to clarify. The key points are that no explicit merging is performed and `kubeadm-config` and `kubeadm-config-<UID>` are disjoint.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: Yup, I will clarify this. Explained in more detail in a comment below.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: Yes, I believe we can automate this actually, which is great.

I was originally thinking it would be difficult / impossible due to not being able to decide whether the existing `NodeName` refers to the current node or not, given the lack of a UID in `kubeadm-config`. However, we can do this:

1. Figure out which node we are by listing all master nodes in the cluster and looking at the existing `product_uuid` field
2. Pull the `NodeName` from our node's metadata
3. Do as @timothysc describes in his comment

Or maybe there's an even simpler way to determine the current node's `NodeName` that I'm overlooking?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
detiber: What will be the expected workflow for init where a user would be overriding config that would end up in both the kubeadm-config and kubeadm-config-<UID> configmaps?

fabriziopandini: @mattkelly 
> The key points are that no explicit merging is performed and `kubeadm-config` and `kubeadm-config-<UID>` are disjoint.
> Does this make sense?

It makes sense, but I'm a little bit concerned about the impact on the user experience...
By designing two disjoint configMaps, do you expect also changes in the `kubeadm  init` user interface? e.g. from `keadm init --config` to `kubeadm init --config --node-config` (to pass two files instead of one)

I guess the prototype will answer to those questions as well; in the meantime, if you agree, we could just take note of the possibile alternative - that provide a less clean representation of the HA cluster - but preserves the current user interface (focused only on the node kubeadm is running) while managing the complexity of the cluster in behind the scene.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+              * [Guaranteed Consistent kubeadm-config](#guaranteed-consistent-kubeadm-config)
+     * [Risks and Mitigations](#risks-and-mitigations)
+  * [Implementation History](#implementation-history)
+  * [Drawbacks](#drawbacks)
+  * [Alternatives](#alternatives)
+
+## Summary
+
+Currently, `kubeadm upgrade` of a master in a multi-master cluster will fail without workarounds because of a lack of node-specific master configuration in the `kubeadm-config` ConfigMap that is created at `init` time and later referenced during an upgrade.
+In particular, node-specific information is required in order for kubeadm to identify which control plane static pods belong to the current node during an upgrade.
```
jamiehannaford: this sentence is a bit unclear to me. Could we better relate how "node-specific information" relate to a static pod?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+In the non-HA case, having a single `nodeName` property in `kubeadm-config` that corresponds to the single master is sufficient because there is no ambiguity.
+As we move towards supporting HA natively in kubeadm, a new approach is required to uniquely identify master nodes.
+
+## Motivation
```
jamiehannaford: do we need separate sections? they seem to talk about the same thing. maybe we could synthesize

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+
+Kubeadm is driving towards natively supporting highly available clusters.
+As part of HA support, a clean upgrade path is required.
+The purpose of this KEP is simply to introduce support for multiple masters in the kubeadm configuration that is stored in-cluster in order to enable that clean upgrade path.
```
jamiehannaford: s/simply//

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+
+Currently, `kubeadm upgrade` of a master in a multi-master cluster will fail without workarounds because of a lack of node-specific master configuration in the `kubeadm-config` ConfigMap that is created at `init` time and later referenced during an upgrade.
+In particular, node-specific information is required in order for kubeadm to identify which control plane static pods belong to the current node during an upgrade.
+In the non-HA case, having a single `nodeName` property in `kubeadm-config` that corresponds to the single master is sufficient because there is no ambiguity.
+As we move towards supporting HA natively in kubeadm, a new approach is required to uniquely identify master nodes.
+
+## Motivation
+
+Kubeadm is driving towards natively supporting highly available clusters.
+As part of HA support, a clean upgrade path is required.
```
jamiehannaford: nit: do we need these on separate lines?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Implementation Details
+
+#### Background
+
+Currently, the `kubeadm-config` ConfigMap in the `kube-system` namespace serves as the single source of truth for how kubeadm has been used to create and modify a cluster.
+Because kubeadm is not a process that runs on the cluster (it is only run to perform operations, e.g. `init` and `upgrade`), this config is not modified during normal operation.
+In the non-HA case today, it is guaranteed to be an accurate representation of the kubeadm configuration.
+
+If kubeadm is used to create an HA cluster today, e.g. using the workarounds described in [kubeadm #546](https://github.com/kubernetes/kubeadm/issues/546) and/or @mbert's [document](https://docs.google.com/document/d/1rEMFuHo3rBJfFapKBInjCqm2d7xGkXzh0FpFO0cRuqg), then the `kubeadm-config` ConfigMap will be an accurate representation except for any master node-specific information.
```
jamiehannaford: shouldn't we refer to the actual HA doc too?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+As explained in [Challenges and Open Questions](#challenges-and-open-questions), such node-specific information is not yet well-defined but minimally consists of the master's `nodeName`.
+The `nodeName` in `kubeadm-config` will correspond to the last master that happened to write to the ConfigMap.
```
jamiehannaford: I kinda feel this should be described earlier in the problem definition

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+In the case of parallel node creation, this may not be well-defined.
+When `kubeadm upgrade` is run on a master and this `nodeName` is fetched, it may be incorrect and the upgrade process will fail.
+
+#### Adding Additional Master-Specific ConfigMaps
+
+The proposed solution is to add additional kubeadm ConfigMaps that are specific to each master (one ConfigMap for each master).
+Each master-specific ConfigMap will be created as part of the to-be-implemented [`kubeadm join --master` process](https://github.com/kubernetes/community/pull/1707).
+Any master-specific information in the main `kubeadm-config` ConfigMap will be removed.
```
jamiehannaford: can we be specific on what master-specific info?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+Each master-specific ConfigMap can be automatically deleted via garbage collection (see [below](#guaranteed-consistent-kubeadm-config) for details).
+
+The names of these new ConfigMaps will be `kubeadm-config-<machine_UID>` where `machine_UID` is an identifier that is guaranteed to be unique for each node in the cluster.
```
jamiehannaford: why machine_UID and not nodeName? wouldn't the latter be more human-readable?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,149 @@
...
+Node creation in parallel is a valid use-case that works today.
+By adding additional ConfigMaps instead of requiring each master to modify the existing `kubeadm-config`, we avoid the need to lock on that ConfigMap.
+
+###### Guaranteed Consistent kubeadm-config
+
+This approach allows us to continue to guarantee that the main `kubeadm-config` is consistent with the actual cluster configuration.
+If we put master-specific information into `kubeadm-config` itself, then we would require either a yet-to-be-defined `kubeadm leave` workflow or active reconciliation of `kubeadm-config` in order to ensure accurateness.
+This may not be critical, but it is a consideration.
+
+With this proposal, if a node unexpectedly leaves a cluster, then at worst a stale ConfigMap will be left in the cluster.
```
jamiehannaford: can we replace "stale" with "dangling"?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: We're not yet sure exactly what fields are master-specific (see "Challenges and Open Questions" above). I could just add "(e.g. `NodeName`)" for now, as that's a definite.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: The core problem is that we don't actually know the `nodeName` of the current node without retrieving it from some persistent config. It can be anything - it's not guaranteed to be the hostname, for example. That's why we need to fetch it based on a machine UID, which is what's proposed here :)

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: I was questioning this myself - I ended up looking at the [KEP template](https://raw.githubusercontent.com/kubernetes/community/master/keps/0000-kep-template.md) to check the style of the raw markdown. This is the style used there. They're not on separate lines in the generated file FWIW (need two newlines for that).

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: This is following the [KEP template](https://github.com/kubernetes/community/blob/master/keps/0000-kep-template.md#motivation).

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: @fabriziopandini that's a good point, and I think also what @detiber was hinting at. I hadn't considered the user-facing impact of using disjoint ConfigMaps. I'll make a note in the meantime, and I agree that the POC will shed some light on how to best manage this.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
jamiehannaford: okay, thanks for clarifying

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: After thinking about this, I don't really see a good way to move it up in the document. I'd like to keep the Summary free of details like this and I think that the preceding background is important to understand this detail (readers need to know that specific workarounds are needed for kubeadm HA to work today anyway).

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: I decided that it would be better to leave this as-is since it would be difficult to shoehorn in a reference to the HA doc here and there are plenty of links out to it from the workarounds links. It also didn't seem to fit in well with the other link sections (`see-also` which is ideally just to other KEPs and `Implementation History`).

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: Updated

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: Updated

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: Updated

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: I added a note in the `Risks and Mitigations` section explaining this. Once I finally get the POC out, this can always be updated.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,161 @@
...
+* [Implementation History](#implementation-history)
+* [Drawbacks](#drawbacks)
+* [Alternatives](#alternatives)
+
+## Summary
+
+One of the first steps of the upgrade process is to retrieve the `kubeadm-config` ConfigMap that is stored at `kubeadm init` time in order to get any relevant cluster configuration.
+The current `kubeadm-config` ConfigMap was designed to support upgrades of clusters with a single master.
+As we move towards supporting High Availability (HA) natively in kubeadm, the persistent configuration must be enhanced in order to store information unique to each master node.
+In particular, master node-specific information is required in order for kubeadm to identify which control plane static pods belong to the current node during an upgrade (the static pods are identifiable by the `nodeName` field embedded in the pod name).
```
kargakis: Why is this specific to static pods?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
jamiehannaford: why is it difficult to add a link?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
jamiehannaford: when reading the summary I didn't understand the problem domain. Giving details there would have facilitated better understanding IMO.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,161 @@
...
+### Implementation Details
+
+#### Background
+
+Currently, the `kubeadm-config` ConfigMap in the `kube-system` namespace serves as the single source of truth for how kubeadm has been used to create and modify a cluster.
+Because kubeadm is not a process that runs on the cluster (it is only run to perform operations, e.g. `init` and `upgrade`), this config is not modified during normal operation.
+In the non-HA case today, it is guaranteed to be an accurate representation of the kubeadm configuration.
+
+If kubeadm is used to create an HA cluster today, e.g. using the workarounds described in [kubeadm #546](https://github.com/kubernetes/kubeadm/issues/546) and/or @mbert's [document](https://docs.google.com/document/d/1rEMFuHo3rBJfFapKBInjCqm2d7xGkXzh0FpFO0cRuqg), then the `kubeadm-config` ConfigMap will be an accurate representation except for any master node-specific information.
+As explained in [Challenges and Open Questions](#challenges-and-open-questions), such node-specific information is not yet well-defined but minimally consists of the master's `nodeName`.
```
liztio: this link doesn't go anywhere

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,161 @@
...
+The `nodeName` in `kubeadm-config` will correspond to the last master that happened to write to the ConfigMap.
+In the case of parallel node creation, this may not be well-defined.
+When `kubeadm upgrade` is run on a master and this `nodeName` is fetched, it may be incorrect and the upgrade process will fail.
+
+#### Adding Additional Master-Specific ConfigMaps
+
+The proposed solution is to add additional kubeadm ConfigMaps that are specific to each master (one ConfigMap for each master).
```
liztio: will the regular `kubeadm join` commands require information about these master nodes (for kubelet control)? How do they decide which master to connect to?

stealthybox: I understand that `kubeadm join` will still be done through an L4/L7 LB that is fronting all of the masters.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: It may not be - I'll make it more clear that this is just an example of how the node-specific information is used.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: It navigates to the `Challenges and Open Questions` section properly for me. Maybe you were scrolled in such a way that it didn't seem like it went anywhere (the section is not far about this line)?

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: I added a link to the official kubeadm HA doc in the `see-also` section.

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
mattkelly: I added [line 54](https://github.com/kubernetes/community/pull/1915/files#diff-adb4d7daed272f590d1032bafcfa7ad3R54) which I hope will clarify the problem at a deeper level earlier on.

mattkelly:
Made some quick updates:

- Address all new comments
- Update some stale links
- Call out `advertiseAddress` as another master node-specific config entry as @fabriziopandini pointed out on in a recent meeting

mattkelly:
Adding some thoughts here regarding the actual implementation details for a proof of concept of what is proposed in this KEP. Hopefully this is useful as a potential jumping off point for discussion during one of the weekly meetings. Keep in mind that I'm still fairly new to the codebase so my opinions on complexity, etc may not be calibrated yet :). 

After digging through the code and trying to introduce additional node-specific ConfigMaps (the new `kubeadm-config-<product_uuid>` CMs proposed here), I've now realized that this is potentially a pretty big change if we want to do it properly.

The main difficulty is that the `MasterConfiguration` struct (aka the "main" config) is used _everywhere_ in kubeadm. By introducing another source for configuration, we will have to do one of the following:

1. Define a new struct containing the "main" config and "node-specific" config and change the entire codebase to use this
    - Massive changeset
2. Intelligently pass the node-specific config only where it is needed
    - Will probably get pretty complex / confusing, especially if we continue to transfer things from the main config to node-specific configs
3. Merge the new node-specific config into the existing config before using it anywhere
    - Internal representation is inconsistent with external, feels pretty strange and too magical
    - May be hard to maintain depending on exactly what "merge" means

Some of the more obvious tradeoffs are listed. The complexity of all of them is compounded by the need to add (and/or modify) all of the defaulting + validation code, etc. This is then made a lot more difficult if we want to minimize negative impact on user experience, [as discussed here](https://github.com/kubernetes/community/pull/1915#discussion_r174856423). 

For these reasons (purely from an implementation complexity + UX viewpoint), my current opinion is that we should revisit the idea of extending the existing `kubeadm-config` without introducing additional node-specific ConfigMaps. I believe the main considerations there would be (1) reconciling the ConfigMap on a best-effort basis to resolve inconsistencies (masters can join/leave/die, etc) and (2) managing concurrent creation attempts / access to the ConfigMap for parallel master bringup.

cc @timothysc @detiber @fabriziopandini @liztio @stealthybox

timothysc:
I'm going to put a hold on this until we KEP out well defined versioning of the config file.  

/cc @liztio

keps/sig-cluster-lifecycle/kubeadm-config-ha-support.md
```diff
@@ -0,0 +1,164 @@
...
+            * [Parallel Node Creation](#parallel-node-creation)
+            * [Guaranteed Consistent kubeadm-config](#guaranteed-consistent-kubeadm-config)
+   * [Risks and Mitigations](#risks-and-mitigations)
+      * [Migrating Existing Clusters](#migrating-existing-clusters)
+      * [More Complex User Experience for Overriding Configuration](#more-complex-user-experience-for-overriding-configuration)
+* [Implementation History](#implementation-history)
+* [Drawbacks](#drawbacks)
+* [Alternatives](#alternatives)
+
+## Summary
```
calebamiles: it seems that your summary is a great motivation and vice versa but that's just how it read to me

timothysc:
We're going to close this one as @fabriziopandini 's PRs are currently up or in.

PR #1926, opened 2018-03-14
Fix broken links

liubin:
Fix two broken links.


castrojo:
Thanks for your fixes!

/lgtm

PR #1930, opened 2018-03-15
Update auditing.md

mayakacz:
"Header" level is not defined anywhere, presumably "Metadata" level is what is meant.

### DIFF
--- a/contributors/design-proposals/api-machinery/auditing.md
+++ b/contributors/design-proposals/api-machinery/auditing.md
@@ -141,7 +141,7 @@ const (
     AuditNone AuditLevel = "None"
     // AuditMetadata provides basic level of auditing, logging data at HTTP level
     AuditMetadata AuditLevel   = "Metadata"
-    // AuditRequestBody provides Header level of auditing, and additionally
+    // AuditRequestBody provides Metadata level of auditing, and additionally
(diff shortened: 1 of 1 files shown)


mayakacz:
My github account should now be associated with a visible work email (so have CLA)

contributors/design-proposals/api-machinery/auditing.md
```diff
@@ -141,7 +141,7 @@ const (
     AuditNone AuditLevel = "None"
     // AuditMetadata provides basic level of auditing, logging data at HTTP level
     AuditMetadata AuditLevel   = "Metadata"
-    // AuditRequestBody provides Header level of auditing, and additionally
+    // AuditRequestBody provides Metadata level of auditing, and additionally
```
deads2k: This edit doesn't look right. RequestBody is a lot of info

@tallclair

tallclair: I think this is right. Yes, audit events can be quite large. There's an issue to limit it: https://github.com/kubernetes/kubernetes/issues/60432

mayakacz: I was wrong here, was reading line 144 as a comment on 143, but it's a comment on 145. REverting

PR #1933, opened 2018-03-15
Update README.md

philoserf:
- update Bob’s organization (trivial change)


cblecker:
As mentioned at the top, this is an auto generated doc. The change would need to be made in sigs.yaml, and then use `make` to regenerate the docs

PR #1934, opened 2018-03-15
add ProcMount option

jessfraz:
/cc sig-node-proposals

see also:
https://github.com/moby/moby/issues/36597
https://github.com/opencontainers/runc/issues/1658#issuecomment-373122073

### DIFF
--- a/contributors/design-proposals/auth/proc-mount-type.md
+++ b/contributors/design-proposals/auth/proc-mount-type.md
@@ -1,0 +1,93 @@
+# ProcMount/ProcMountType Option
+
+## Background
+
+Currently the way docker and most other container runtimes work is by masking
+and setting as read-only certain paths in `/proc`. This is to prevent data
+from being exposed into a container that should not be. However, there are
+certain use-cases where it is necessary to turn this off.
+
+## Motivation
+
+For end-users who would like to run unprivileged containers using user namespaces
+_nested inside_ CRI containers, we need an option to have a `ProcMount`. That is,
+we need an option to designate explicitly turn off masking and setting 
+read-only of paths so that we can
+mount `/proc` in the nested container as an unprivileged user.
+
+Please see the following filed issues for more information:
+- [opencontainers/runc#1658](https://github.com/opencontainers/runc/issues/1658#issuecomment-373122073)
+- [moby/moby#36597](https://github.com/moby/moby/issues/36597)
+- [moby/moby#36644](https://github.com/moby/moby/pull/36644)
+
+Please also see the [use case for building images securely in kubernetes](https://github.com/jessfraz/blog/blob/master/content/post/building-container-images-securely-on-kubernetes.md).
+
+Unmasking the paths in `/proc` option really only makes sense for when a user 
+is nesting
+unprivileged containers with user namespaces as it will allow more information
+than is necessary to the program running in the container spawned by
+kubernetes.
+
+The main use case for this option is to run
+[genuinetools/img](https://github.com/genuinetools/img) inside a kubernetes
+container. That program then launches sub-containers that take advantage of
+user namespaces and re-mask /proc and set /proc as read-only. So therefore
+there is no concern with having an unmasked proc open in the top level container.
+
+It should be noted that this is different that the host /proc. It is still
+a newly mounted /proc just the container runtimes will not mask the paths.
+
+Since the only use case for this option is to run unprivileged nested
+containers,
+this option should only be allowed or used if the user in the container is not `root`.
+This can be easily enforced with `MustRunAs`.
+Since the user inside is still unprivileged,
+doing things to `/proc` would be off limits regardless, since linux user
+support already prevents this.
+
+## Existing SecurityContext objects
+
+Kubernetes defines `SecurityContext` for `Container` and `PodSecurityContext`
+for `PodSpec`. `SecurityContext` objects define the related security options
+for Kubernetes containers, e.g. selinux options.
+
+To support "ProcMount" options in Kubernetes, it is proposed to make
+the following changes:
+
+## Changes of SecurityContext objects
+
+Add a new `string` type field named `ProcMountType` will hold the viable
+options for `procMount` to the `SecurityContext`
+definition.
+
+By default,`procMount` is `default`, aka the same behavior as today and the
+paths are masked.
+
+This will look like the following in the spec:
+
+```go
+type ProcMountType string
+
+const (
+    // DefaultProcMount uses the container runtime default ProcType.  Most 
+    // container runtimes mask certain paths in /proc to avoid accidental security
+    // exposure of special devices or information.
+    DefaultProcMount ProcMountType = "Default"
+
+    // UnmaskedProcMount bypasses the default masking behavior of the container
+    // runtime and ensures the newly created /proc the container stays in tact with
+    // no modifications.  
+    UnmaskedProcMount ProcMountType = "Unmasked"
+)
+
+procMount *ProcMountType
+```
+
+This requires changes to the CRI runtime integrations so that
+kubelet will add the specific `unmasked` or `whatever_it_is_named` option.
+
+## Pod Security Policy changes
+
+A new `[]ProcMountType{}` field named `allowedProcMounts` will be added to the Pod
+Security Policy as well to gate the allowed ProcMountTypes a user is allowed to
+set. This field will default to `[]ProcMountType{ DefaultProcMount }`.


jessfraz:
cc @tallclair @dchen1107 happy to discuss at next sig-node meeting

jessfraz:
cc @rhatdan

contributors/design-proposals/node/cri-rawproc.md
```diff
@@ -0,0 +1,35 @@
+# CRI: RawProc Option
+
+## Background
+
+Currently the way docker and most other container runtimes work is by masking
+and setting as read-only certain paths in `/proc`. This is to prevent data
+from being exposed into a container that should not be. However, there are
+certain use-cases where it is necessary to turn this off.
```
tallclair: Are there any other constraints that are lifted by running privileged? Device hiding?

jessfraz: Privileged add all caps and all host devices. So this won’t do either of those :)

rhatdan: Yup --privileged turns off seccomp, adds all caps, turns off SELinux and AppArmor, adds all devices from most, turns off the device cgroup, this allows processes inside of the container to make and interact with any devices.  Exposes read/write kernel file systems.

rhatdan: You might want to add something equivalent to --device all where the Container Runtime would expose all devices from the host to the container and turn off device cgroup.  As a separate knob.   Not sure how many people other then people debugging would ever turn these knobs.

rhatdan: CRI-O Does whatever Kubernetes tells it to do, If K8s supports rawproc/CRI-O will support Rawproc.
podman currently does everything that docker does.
Know idea what kind of controls RKT has.
LXC can probably do anything, since it has knobs for everything.

jessfraz:
Added a bit more information on usecases hopefully that is useful.

contributors/design-proposals/node/cri-rawproc.md
tallclair: > You might want to add something equivalent to --device all where the Container Runtime would expose all devices from the host to the container and turn off device cgroup.

Yeah, that's what I was getting at. We already expose options for capabiltiise, seccomp, and AppArmor. I think we're missing a knob to bipass SELinux (or whatever the privileged behavior is there), disable the device cgroup, and rawproc. Is there anything else?

contributors/design-proposals/node/cri-rawproc.md
```diff
@@ -0,0 +1,50 @@
...
+
```
Random-Liu: Instead of adding a `rawproc` option at CRI level, why don't we just expose `maskedPaths` and `readonlyPaths` ([link](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#masked-paths)) in CRI?

I'm not saying that we should do that, but I want to make sure we've considered that option.

I think it is more generic. If needed, we can leverage that to mask/unmask or set read-only 
for other paths in the future, instead of adding another `rawXXX`.

contributors/design-proposals/node/cri-rawproc.md
```diff
@@ -0,0 +1,50 @@
...
+Which will inform the runtimes implementing CRI to not mask or set as read-only
+the paths in `/proc`.
+
+This option would also be exposed as a `bool` to the `securityContext` for
+containers in the form of `rawProc`.
+
+It will be false by default.
+
+This will also add a `AllowRawProc` to the `PodSecurityPolicy` to allow
```
yujuhong: My $.02: The proposal should be about adding a feature in kubernetes, and have sections on kubernetes API, and CRI changes, respectively. The title and the way that the content is arranged seem to suggest that this is only a CRI change.

Random-Liu: Similar questions with https://github.com/kubernetes/community/pull/1934/files#r175907931.

At Kubernetes level, should we expose highest level abstraction `privileged`? Or several lower level abstractions `rawproc`, `rawsys`, `rawXXX`? Or expose `maskedPaths` and `readonlyPaths` directly?

None of them seems idea to me.

derekwaynecarr: @yujuhong i had missed this section in sig-node today, agree it would be good to start from the user facing API down to the CRI.

contributors/design-proposals/node/cri-rawproc.md
mrunalp: I agree with @Random-Liu. We could just expose maskedPaths and readonlyPaths. 
A higher level abstraction makes more sense when it isn't portable between runtimes.

contributors/design-proposals/node/cri-rawproc.md
jessfraz: I think it is confusing for users at the top level to have the options of configuring everything but up to you all.

jessfraz: I really don't mind what the CRI api is, i really just need this in the top level, what do you suggest then? making this proposal just for the top level, you all can implement in cri like i did in docker...
https://github.com/moby/moby/pull/36644

I really don't even care about the name, lol, I just need a way to do it.

contributors/design-proposals/node/cri-rawproc.md
jessfraz: ok thats for cri, but what do you want for the top level, the same? I will do whatever everyone wants, I really have no fight in the naming, I just need a way to do this.

contributors/design-proposals/node/cri-rawproc.md
yujuhong: @jessfraz CRI is only a small part of this proposal. The more important questions are why Kubernetes needs this, what the use cases are, and how this would be represented in the Kubernetes API. We can discuss how to add this option to CRI after those are nailed down.

I am unclear about the use cases. Could you explain why nested containers are needed or point us to related issues and proposals? As for adding this to PodSecurityPolicy, I think there are more people with opinions on that (hence my suggestion on changing the proposal to get reviews from the right folks).

If I'm wrong and this is a pure CRI change to support finer-grained control (e.g., break down of `Privileged`), let me know and we can review as it is.

jessfraz: this is why I need it https://blog.jessfraz.com/post/building-container-images-securely-on-kubernetes/

contributors/design-proposals/node/cri-rawproc.md
```diff
@@ -0,0 +1,50 @@
...
+container. That program then launches sub-containers that take advantage of
+user namespaces and re-mask /proc and set /proc as read-only. So therefore
+there is no concern with having a raw proc open in the top level container.
+
+## Implementation
+
+This proposal suggests adding the following to `bool` `LinuxSandboxSecurityContext`:
+
+```
+bool raw_proc
```
AkihiroSuda: `repeated string raw_access` instead of bool?

jessfraz: thats sounds right :)

jessfraz:
I am going to redo this as a proposal for the high level k8s api, and you all can figure out naming for cri however you want, how does that sound?

contributors/design-proposals/node/cri-rawproc.md
smarterclayton: Which users would I definitely not give rawProc to?  Might be good to link/include a capsule "here's why /proc/sys being writable is / is-not dangerous" description here.

@rhatdan I know this was because we wanted to close off the option, if selinux is enough to protect those subsequently I'm less worried.

I *want* to run unprivileged nested containers if we have sufficient defense behind /proc now.

jessfraz: > Which users would I definitely not give rawProc to? 

Someone evil? Who would possibly try to exploit the info in proc with an 0day. Remember the user inside is still unprivileged so doing things to proc would be off limits regardless. So we could make it so rawProc is only allowed if the user !=0 :)

dchen1107:
/assign @liggitt

Looks good to me on node side. Assign to @liggitt for the top level API approval.

stmcginnis:
The title probably should be updated now to avoid any confusion.

contributors/design-proposals/auth/rawproc.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Changes of SecurityContext objects
+
+Add a new `bool` type field named `rawProc` to the `SecurityContext`
+definition.
+
+By default,`rawProc` is `false`.
+
+The API will reject as invalid `rawProc=true` and `user=0/root`, since `rawProc`
+only makes sense if you want to nest unprivileged user namespaces.
```
smarterclayton: Since PSP would be able to control this interaction anyway, this may not be necessary.  `nonRoot` + `allowRawProc`

contributors/design-proposals/auth/rawproc.md
```diff
@@ -0,0 +1,73 @@
+# RawProc Option
```
smarterclayton: Other than maybe spending a few minutes trying to come up with a better name, I think this is a reasonable small change.  Allowing containers to nest unprivileged containers is a good thing for other use cases, like isolating child processes.

rhatdan: SELinux would protect against most attacks against /proc and more specifically other kernel filesystems like /proc/sys /sys/fs/* ...  But I would be careful making this the norm.

smarterclayton: @rhatdan and does being non-root protect us at all?

rhatdan: Yes, most parts of the kernel file systems require root (CAP_SYS_ADMIN) to do any manipulation of the content.

frezbo: @jessfraz as @rhatdan  mentioned earlier this approach would not work on  a hardened machine with SELinux enabled for docker, and I came up with a strange idea which would still allow to build containers as non-root:
1. Run containerd as non-root on the host (still trying to fix the issues) (so all the issues with masked proc would go away)
2. Run a socat container that would forward the containerd socket to the buider image
3. Img/buildah or whatever tool uses this socket to build images

This is solve many issues, 
1. the containerd process is running as non-root on the host and uses runc to create/build images
2. Do not want to worry/put up hacks for the proc fix
3. SELinux would not be a limiting/blocking factor here

rhatdan: Having your build tool run as non root so that it can run inside of an orchestrator like Kubernetes launch builders in non privileged user namespaces, preferably a different UserNS for each builder.  In this case your container runtime like CRI-O or containerd would still be running as root and launching the build containers in different UserNS.

With the tool Buildah, we want to support this method, but we also want to have buildah running as its own Container Runtime, where each buildah container could create the build container in a different UserNS.

jessfraz: I agree with @rhatdan the true value comes from running as different unprivileged users not just one

contributors/design-proposals/auth/rawproc.md
jessfraz: should I take this part out and then do you think we are good to go? :)

jessfraz:
do we think we can get this through?

squillace:
Far be it from me to push, but I'd sure love to see this solution land.

contributors/design-proposals/auth/rawproc.md
jessfraz: I don't know about selinux specifically but apparmor yes

jessfraz: I will try on CoreOS w selinux

tallclair:
I'm a little concerned that this is too implementation-specific, but I don't have a better solution and don't want to block this. Do all the major runtimes mask proc?

jessfraz:
most seem to

contributors/design-proposals/auth/rawproc.md
dims: What would a better one-line title be? "Option to disable read only /proc for unprivileged non-root containers"

rhatdan:
Would RWProc be a better name?

smarterclayton:
We typically would have more expansive names, so WritableProcFilesystem
would probably be the longer form of that name.  One advantage is that
RawProc sounds like something you shouldn't ask for if you don't know what
you're doing, while WritableProcFilesystem does sound like something you
might want.  We could do ExposeRawProc but that's just adding a verb to the
noun (and we generally try to just have the noun, like User, Group,
HostPID).

rhatdan:
Fine.  I agree RawProc is scarier.  But to me the name means nothing.

jessfraz:
I think there will only ever but again this comes from the fact runtimes made the assumption no one would ever want to even do this in the first place so who really knows. I can't predict the future or else I would definitely not work in tech :)

liggitt:
other than the enum vs bool question (which I don't feel strongly about, just wanted to raise), I don't have further comments

dims:
+1 from me

squillace:
almost there... :-)

smarterclayton:
How about we change to 

```
type ProcMountType string

const (
  // ... uses the container runtime default ProcType.  Most container runtimes selective mount /proc to avoid accidental security exposure of special devices.
  DefaultProcMount ProcMountType = "Default"
  // ... bypasses the default mount behavior of the container runtime and ensures the host /proc is mounted into the container.  When the host /proc is mounted into the container, the container may have data that exposes host specific information unless other protections such as SELinux limit access.
  HostProcMount ProcMountType = "Host"
)

  ...
  procMount *ProcMountType
```

?

liggitt:
I like that... seems clearer and allows for the possibility of more-restrictive-than-default modes in the future

jessfraz:
cool will update :)

jessfraz:
so it isnt really the host proc, it is still a new proc, just the paths are not masked. so that would be a little weird wording :)

jessfraz:
@smarterclayton @liggitt any additional thoughts on the naming re: my comment above about this not being a host proc, it is a newly mounted proc, just the paths are not masked by the container runtime

smarterclayton:
UnmaskedProcMount = “Unmasked”

?

jessfraz:
that sgtm :)

squillace:
:-)

liggitt:
works for me, thanks

smarterclayton:
/lgtm
/approve

Thanks for the patience

jessfraz:
@Random-Liu  @mrunalp how do you want to do this with cri i am happy to open a PR :) or if you all want a certain way that can be as well

jessfraz:
am i the first in this chicken and egg problem lol

Random-Liu:
@jessfraz There are 2 options:
* Option 1: Expose MaskedPath and ReadonlyPaths in CRI, and let kubelet apply the default, and kubelet can then handle `ProcMountType` correspondingly.
  * Pros: Consistent behavior across different container runtimes.
* Option 2: Only have `ProcMountType` in CRI, and let container runtimes handle it separately.

I prefer option 1, as long as we are fine with letting kubelet control the default.

Another question is how do we support this in dockershim? I'm fine with not supporting it, but would like to know whether it is possible. :)

jessfraz:
I have no preference on the first since that is yours all to maintain :) so I am happy to make the PR any way you want

as far as docker, i made a PR but let's turn off support until that's in and then i can add support after :)

jessfraz:
opened https://github.com/kubernetes/kubernetes/pull/64283

dimm0:
> What is the security tradeoff a cluster operator is making by allowing unmasked ProcMounts in their pod security policies?

Could somebody answer this plz?

ThisGuyCodes:
@dimm0 does this help? https://github.com/kubernetes/community/pull/1934#discussion_r176509249

dimm0:
Yup, thanks

PR #1939, opened 2018-03-17
KEP kubeadm phases to beta

fabriziopandini:
This KEP is aimed at defining a roadmap for graduating `kubeadm alpha phase` commands to beta (and then GA) while trying to address at the same time concern/lesson learned so far about the additional effort for maintenance of this feature (e.g. flag proliferation, UX code duplication).

A prototype implementation is currently work in progress and will become a [WIP] PR asap.

/CC @kubernetes/sig-cluster-lifecycle-proposals


keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+* How to provide an intuitive user interface for both phases and sub-phases
+  (e.g  `kubeadm alpha phase certs` and `kubeadm alpha phase certs apiserver`)?
+* Should we preserve all the second/third level nested phases implemented today
+  (e.g. certs CA, certs api-server, certs sa etc.), or should we allow only second
+  level phases (e.g certs), or should we mantain & simplify second/third level phases?
+* Should we preserve or derecate the (few) phases implemented today that are not
+  part of the `kubeadm init` workflow? If preserve, in which form? as a new first
+  level commands?
+* While `kuneadm init` executes some phases conditially (e.g. self-hosting only
+  if the correspondig feature-gate enable), the current implementation of phases
```
cblecker: s/correspondig/corresponding

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+## Implementation History
+
+* Prototype implementation (in progress)
+
+## Drawbacks
+
+By merging phases into `kubeadm init` derives a reduced capability to customize
+the user interface for each phase. More specifically:
+
+* It would not be possible to provide any kind of advice to the user about which
```
liztio: is there no way to have `kubeadm init certs --help` spit out different help than just the same as `kubeadm init`?

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+
+```bash
+kubeadm init certs [flags]
+```
+
+* To invoke the beta (GA) version of  `kubeadm alpha phase certs apiserver` the user
+  will be required to execute:
+
+```bash
+kubeadm init certs/apiserver [flags]
```
liztio: is there prior art for the `phase/subphase` syntax anywhere else in kubeadm / k8s in general?

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+effort for maintenance of this feature.
+
+## Motivation
+
+The `kubeadm phase` command was introduced in v1.8 cycle under `kubeadm alpha phase`
+command with the goal to provide to users an interface for invoking individually any
+task/phase of the `kubeadm init` workflow.
+
+During this period,  `kubeadm phase`  proved to be a valuable and composable
+API/toolbox that can be used by any IT automation tool or by advanced user for
```
liztio: I'm happy to give this a copyediting once-over if that'd be helpful! this should be "by an advanced user" for example.

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+  that significantly reduces the effort for maintaining this feature up to date
+  with the continuous changes of the init workflow.
+* To provide a centralized point for controlling the proliferation of flags in
+  kubeadm.
+* To avoid inconsistencies between init and phases (both at UX and implementation
+  logic level)
+
+### Non-goals
+
+* This proposal doesn't include any changes/improvement to the init workflow
```
liztio: how is modifying `kubeadm init` to run through the phases not a change to the "init workflow internal logic?" is something else meant by this?

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+
+The `kubeadm phase` command was introduced in v1.8 cycle under `kubeadm alpha phase`
+command with the goal to provide to users an interface for invoking individually any
+task/phase of the `kubeadm init` workflow.
+
+During this period,  `kubeadm phase`  proved to be a valuable and composable
+API/toolbox that can be used by any IT automation tool or by advanced user for
+creating custom clusters.
+
+On the other side, it was necessary a continuous effort for maintaining this
```
liztio: this sentence is a little confusing. "It required a continuous effort to keep this component up to date with" maybe?

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
fabriziopandini: Thanks! this will be really appreciated!

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
fabriziopandini: Even if this KEP imply a refactoring of kubeadm init for allowing execution of the entire workflow or single phases, when you will execute the entire workflow you will get the same result as today e.g. the same certs will be created, the same manifest file will be created etc.

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
fabriziopandini: Strictly speaking of phase/subphase of a command, there isn't nothing I'm aware of.
If we consider the path generic pattern for expressing a hierarchy of things, it is already present in K8s e.g. in `kubectl scale deployment/nginx --replicas=3`

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
fabriziopandini: If [phases] are defined as argument of the `kubeadm init` command as discussed in this proposal, I don't think this is feasible. The only way I'm aware of for doing his is to maintain a separate three of command for phases (see alternatives)

fabriziopandini:
@liztio happy to have you on board! 🎉 
Minor changes are addressed; answers to your question are in line

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+```bash
+Usage:
+  kubeadm init [flags]
+```
+
+to
+
+```bash
+Usage:
+  kubeadm init [phases] [flags]
```
errordeveloper: Are you implying there would be like a comma-separated list of phases or just a name of a single phase?
Arguments like this make it too ambiguous, how about `--only=<listOfPhases>` or `--only=<singlePhase>`?

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
errordeveloper: How about more explicit naming, e.g. something like `allCerts`, `apiserverCerts` etc? I find that characters with special meanings in command-line arguments can be very misleading, someone might think phases map to files, or even consider that it's similar to `deployment/nginx`...

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
liztio: I'm only starting to get familiar with the codebase, but could we use [`SetHelpFunc`](https://github.com/spf13/cobra/blob/master/command.go#L223)? It looks like the second argument is the positional arguments, so we could switch based on that.

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
liztio: I think it's a tricky balance to strike. Having all those separate commands is going to make discovery / organising. Having the explicit hierarchy makes sense to me. I was hoping for something simpler like `kubeadm init certs apiserver`, but that obviates being able to run multiple commands. I like the implementation as-is, the prior art that exists is good enough for me.

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+
+NB. By executing `kubeadm init`  without [phases], all the task/phases which are part
+of the init workflow will be executed; this will guarantee consistency with the
+current behavior.
+
+#### Expected benefits
+
+The proposed implementation approach is expected to give following advantages:
+
+* `kubeadm phase` "by design" will be aligned to any change to `kubeadm init`
```
errordeveloper: Why two commands? This is very confusing.

fabriziopandini: Good point. `kubeadm phases` should be replaced with `kubeadm init [phases]`

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
errordeveloper: I wasn't clear enough, I was assuming implying my other suggestion to introduce an explicit `--only` flag, so the usage would be something like `--only= allCerts` or `--only=apiserverCerts`. To discover phases `--only=help` could be an option, or otherwise `--help` should be sufficient actually.

errordeveloper: > I like the implementation as-is, the prior art that exists is good enough for me.

It seems mostly okay to me too (except a few odd naming decisions).

However, if we want to move this functionality under `kubeadm init`, I'd strongly insist on having an explicit flag and an unambiguous naming scheme.

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
fabriziopandini: Open for discussion....
> Are you implying there would be like a comma-separated list of phases or just a name of a single phase?
With regards to the number of phases we can limit it to 1 or keep it open. [phases] can be changed in whatever we want accordingly (the current form was inspired by [flags], [args] etc.)
> Arguments like this make it too ambiguous, how about --only=<listOfPhases> or --only=<singlePhase>?
Personally I think that using arguments is the best choice for this use case

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+  of UX and API in  kubeadm (e.g. the code that implement phases writes to stdout)
+
+## Graduation Criteria
+
+* To create a periodic E2E test that bootstraps a cluster using phases
+* To document the new user interface for phases in kubernetes.io
+
+## Implementation History
+
+* Prototype implementation (in progress)
```
liztio: Could we link to this PR, now that it's in progress?

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
fabriziopandini: Unfortunately `setHelpFunc` seems a dead end; [here](https://github.com/spf13/cobra/blob/a1f051bc3eba734da4772d60e2d677f47cf93ef4/command.go#L322) arguments are not effectively passed, and [here](https://github.com/spf13/cobra/blob/a1f051bc3eba734da4772d60e2d677f47cf93ef4/command.go#L851) argument are passed but only in case the CMD is executed and fails...

However, apart from the technical solution, the real question is what kind of dedicated help we think is necessary for each function, and the effort for maintaining this for about thirty phases.

If I look at current phases implementation, It seems to me that the real added value of custom help is low, so I'm basically fine in getting rid of the maintenance of all those text and of the tedious task of keeping aligned flags between phases and init.

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,243 @@
...
+  with the continuous changes of the init workflow?
+* How to provide an intuitive user interface for both phases and sub-phases
+  (e.g  `kubeadm alpha phase certs` and `kubeadm alpha phase certs apiserver`)?
+* Should we preserve all the second/third level nested phases implemented today
+  (e.g. certs CA, certs api-server, certs sa etc.), or should we allow only second
+  level phases (e.g certs), or should we maintain & simplify second/third level phases?
+* Should we preserve or deprecate the (few) phases implemented today that are not
+  part of the `kubeadm init` workflow? If preserve, in which form? as a new first
+  level commands?
+* While `kubeadm init` executes some phases conditionally (e.g. self-hosting only
```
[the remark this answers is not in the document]
fabriziopandini: You are right, when developing the prototype I explored the option to preserve this behaviour. 
But this is just a proposed solution; let's see if there are different opinions about this

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
fabriziopandini: @errordeveloper message received strong and clear. I start tinkering around this

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
liztio: I'm looking over the existing phases and it seems like there's a pretty good number of phase- and subphase-specific arguments. I guess the question is whether individual users would get much help out of breaking them out. I would lean towards having _some_ machine-readable summary of which phases and subphases take which arguments. In the future we could use those to make sure our test coverage is complete over valid combinations of arguments.

keps/sig-cluster-lifecycle/draft-20180316-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,241 @@
...
+```yaml
+---
+kep-number: draft-20180316
+title: kubeadm phases to beta (GA)
+status: accepted
+authors:
+  - "@fabriziopandini"
+owning-sig: sig-cluster-lifecycle
+reviewers:
+  - TBD
```
timothysc: me and I'll poke @liztio to review as well.

fabriziopandini:
@timothysc @neolit123 @luxas 
Re-submitting an updated version to get feedbacks/consensus about executing this in the v1.13 cycle

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,184 @@
...
+
+to
+
+```bash
+Usage:
+  kubeadm init [workflow-flags] [flags]
+```
+
+[workflow-flags] will be a set of optional flags that allows user to select which phase
+of the `kubeadm init` logic to execute (`--only-phases`) or to skip (`--skip-phases`). e.g.
```
neolit123: my vote goes to something in the lines of:
`--apply-phases`

i.e. having a verb in the flag name.

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,184 @@
...
+- It would not be possible to provide any kind of advice to the user about which
+  flags are relevant for one specific phase (the help will always show all the flags).
+- It would not be possible to add long description and/or examples to each phase
+- It would not be possible to provide additional flags specific for one phase
+  (the flags are shared between init and all the phases).
+- It would not be possible to expose to the users phases which are not part of kubeadm workflows
+  ("extra" phases should be hosted on dedicated commands).
+
+This is considered an acceptable trade-off in light of the benefits of the suggested
+approach.
```
neolit123: i agree

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Alternatives
+
+It is possible to graduate phases by simply moving corresponding command to first level,
+but this approach provide less opportunities for reducing the effort
+for maintaining phases up to date with the changes of kubeadm.
```
neolit123: are we still planning to move all `kubeadm alpha phase` -> `kubeadm phase` early in 1.13?

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
fabriziopandini: @neolit123, thanks for this comments! They add more value to this proposal!

> we need to outline ordering to the user somehow

Yes, definitely. I'm going to add a sentence about this

> if --apply-phases=addons, kubelet is called,  would that imply that the user wants to first install addons and then run the kubelet

We can choose if to allow execution of a single phase only - as of today - or if to allow execution of one or many phases - as it is now in this proposal -. If we confirm many, IMO the executor should always respect the given order of phases for sake of consistency and for reducing complexity (only one behaviour to be implemented and tested).

Eventually we should consider if to give evidence of that to the user by printing the execution plan - the ordered list of phases that are going to be executed - when running in --dry-run  mode

> have you considered  `kubeadm init --phase-config=<some-config-structure>`

This is discussed in the drawbacks paragraph, and the answer to your question is yes, this was considered even if not in this exact form, but I discarded this option for several reasons, the most important are again consistency and the ambition to reduce complexity.

Phases are part of kubeadm init workflow, so flags of phases are expected to be a subset of the flags of the kubeadm workflow; if a flag is added to kubeadm init workflow, phases should get the same flag, and I think that the viceversa holds as well.

On the contrary, If we open to something different there will be again flag proliferation, and both the CLI help, the user documentation and the code for flag validation as well will be forced to manage an O(MxN) problem.

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
fabriziopandini: I hope so, but this is up to the Sig...

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,184 @@
...
+```bash
+Usage:
+  kubeadm init [workflow-flags] [flags]
+```
+
+[workflow-flags] will be a set of optional flags that allows user to select which phase
+of the `kubeadm init` logic to execute (`--only-phases`) or to skip (`--skip-phases`). e.g.
+
+```bash
+kubeadm init --only-phases certs [flags]
```
chuckha: I would really like to see a way to pass in a file containing a list of phases either in addition to replacing the --only-phases flag where you specify phases separated by commas. I think this would provide a nice way to modify and iterate on a list of phases to run without relying on bash history. It would also be nice to say `kubeadm init --phases > phases.txt` then modify the list of phases to be specific.

fabriziopandini: +1 on having this as part of init configuration

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
fabriziopandini: > We should have strict dependency ordering and go on that lines

Totally agree. See "IMO the executor should always respect the given order of phases for sake of consistency and for reducing complexity (only one behaviour to be implemented and tested)" above.

> This is no longer a full blown init, but rather kubeadm phases sub-command.

I got your point but I'm not sure I agree with you.
1) what will happen if we apply the phases concept to other kubeadm commands, like kubeadm upgrade or kubeadm join? Everything (all the phases for all the commands) will clash into kubeadm phases? 
2) having phases separated by the workflow they belong already created problems in the past, and today there is set of utility commands under phases which are not technically a phases of the kubeadm init workflow e.g certs renew. I want to prevent this happening again otherwise the user won`t understand anymore what a phase is.
3) executing a phase is nevertheless executing a piece of init , so why to move it in a different place?

fabriziopandini: As agree in kubeadm office hours we are giving the users only the capability to invoke one phase at time

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
fabriziopandini: As agree in kubeadm office hours we are giving the users only the capability to invoke one phase at time

keps/sig-cluster-lifecycle/0029-20180918-kubeadm-phases-beta.md
```diff
@@ -0,0 +1,184 @@
...
+    /kubelet        Generates a kubeconfig file for the kubelet to use.
+    ...
+  ...
+````
+
+Phases implementation will be designed in order to ensure that phases are reusable across different
+workflows thus allowing e.g. reuse of phase `certs` in both `kubeadm init` and
+`kubeadm join --master` workflows.
+
+Each kubeadm workflow will be defined and managed using an `Executor` that basically is a component
```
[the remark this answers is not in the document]
fabriziopandini: Done

PR #1942, opened 2018-03-19
Move provider Cloud Controller Manager repo requirements to KEP

hogepodge:
Moves the recently approved provider Cloud Controller Manager repository requirements to the Cloud Controller Manager kep. Also integrates corrections and clarifications from WG-Cloud-Provider.


hogepodge:
@wlan0 @calebamiles @andrewsykim

keps/0002-controller-manager.md
```diff
@@ -206,7 +210,7 @@ taints.
 ### API Server Changes
 
 Finally, in the kube-apiserver, the cloud provider is used for transferring SSH keys to all of the nodes, and within an a
-dmission controller for setting labels on persistent volumes.
+admission controller for setting labels on persistent volumes.
```
Fedosin: There is an "a" on the line above. After this change it looks like "...within an a admission..."

hogepodge:
@Fedosin thanks, fixed.

keps/0002-controller-manager.md
```diff
@@ -254,7 +258,100 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+Each cloud provider hosted within the `kubernetes` organization shall have a
+single repository named `kubernetes/cloud-provider-<provider_name>`. Those
+repositories shall have the following structure:
+
+* A `cloud-controller-manager` subdirectory that contains the implementation
+  of the provider-specific cloud controller.
+* A `docs` subdirectory.
+* A `docs/cloud-controller-manager.md` file that describes the options and
+  usage of the cloud controller manager code.
+* A `tests` subdirectory that contains testing code.
```
calebamiles: I don't think this should be a requirement. Although this is what we do for the e2e tests for Kubernetes. Some reasonable people think that having tests moved away from the code is an antipattern. Perhaps we should require a `TESTING.md` or similar which specifies how to run the tests for humans and a `Makefile` as a test entrypoint

chenopis:
Thoughts on what should be in the docs guidelines:
- What should be in the front matter
  - `title:`, should follow our imperative form style, e.g. "Setup Kubernetes on Foo Cloud" instead of "Setting up..."
  - `canonical-source:`, i.e. link to file in the upstream repo where changes should be made
  - `reviewers:` original owner or stakeholders
- Document structure; if it is a task oriented doc, it should have sections:
  - overview
  - prerequisites, usually titled "Before you begin" 
  - steps
  - (optional) discussion
  - (optional) what's next
- Should follow Kubernetes Style Guide [language](https://kubernetes.io/docs/home/contribute/style-guide/#language) and [formatting standards](https://kubernetes.io/docs/home/contribute/style-guide/#documentation-formatting-standards)
- Directory structure
  - Have a `/docs` directory which will be imported into kubernetes.io
- Import config file
  - Submit a PR for an import config file, e.g. [community](https://github.com/kubernetes/website/blob/master/update-imported-docs/community.yml), that specifies which files should be imported and to where in kubernetes.io they should go. **Note**: we are working on making the default behavior recursively import an entire directory so you don't have to list the files individually.

MVP next steps:
- Have one of the cloud providers setup a doc in their repo for us to import.
- Iterate on what needs to be in the guidelines.
- Automate the import process.

andrewsykim:
/lgtm 

@calebamiles PTAL

edqallen:
@hogepodge is there a sig/label for vSphere?  I was referenced to this issue by another regarding setting up a vsphere cluster, but I don't see it in the label list;  I'd like to make sure that it's being attended to as well.

hogepodge:
@edqallen We're in kind of a strange spot right now. We're looking at forming a new sig-cloud group, and the individual provider sigs will likely be closed down and folded into sig-cloud. At that point the new sig would start accepting new projects. At least, that's the plan as I understand it after kubecon.

keps/0002-controller-manager.md
```diff
@@ -257,6 +263,102 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+* `installer` for custom installer code.
+
+Each repository may have additional directories and files that are used for
+additional feature that include but are not limited to:
+
+* Other provider specific testing.
+* Additional documentation, including examples and developer documentation.
+* Dependencies on provider-hosted or other external code.
+
+
```
prydie: nit: ~~`\n`~~

keps/0002-controller-manager.md
```diff
@@ -257,6 +263,102 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+* A `docs/getting-started.md` file that describes the installation and basic
+  operation of the cloud controller manager code.
+
+Where the provider has additional capabilities, the repository should have
+the following subdirectories that contain the common features:
+
+* `dns` for DNS provider code.
+* `cni` for the Container Network Interface (CNI) driver.
+* `csi` for the Container Storage Interface (CSI) driver.
```
karataliu: What should be in put in `cni` or `csi` dir? Since their code are neutral to orchestrator, they should belong to a dedicate repo already ? `flex` makes sense here since it is for k8s only.

keps/0002-controller-manager.md
```diff
@@ -257,6 +263,102 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+single repository named `kubernetes/cloud-provider-<provider_name>`. Those
+repositories shall have the following structure:
+
+* A `cloud-controller-manager` subdirectory that contains the implementation
+  of the provider-specific cloud controller.
+* A `docs` subdirectory.
+* A `docs/cloud-controller-manager.md` file that describes the options and
+  usage of the cloud controller manager code.
+* A `docs/testing.md` file that describes how the provider code is tested.
+* A `Makefile` with a `test` entrypoint to run the provider tests.
```
karataliu: Shall this be enforced? This should be described in 'docs/testing.md' right?

keps/0002-controller-manager.md
andrewsykim: Good point about CSI/CNI being neutral to orchestrators, I wouldn't be surprised if we remove these directories in the near future, I think today though it's useful since we can put deployment specs for running CSI/CNI on top of Kubernetes (daemonests, configmaps, etc).

keps/0002-controller-manager.md
andrewsykim: Yeah, I think we may want to change this in the future, a lot of what we have now is subject to change and more of a starting point for us to iterate on :)

PR #1949, opened 2018-03-21
dir should not have blank

shenyuanyu:
should not have blank in dir


cblecker:
The space is intended.

/close

liggitt:
@cblecker then the `vendor/directory` on the same line probably should also have a space

shenyuanyu:
ok, thanks

PR #1937, opened 2018-03-16
Add initial design proposal for Scheduling Policy

arnaudmz:
Ref: kubernetes/kubernetes#61185

/assign @bsalamat @liggitt 
/cc @yastij

PR #1950, opened 2018-03-21
[RFC] PodRestriction proposal

tallclair:
This is a proposal for a new policy object (and associated admission controller) that adds basic restrictions to most fields on a Pod. This is motivated by a desire to unify the various policies we have now (which all behave in slightly different ways), and to unblock expanding the set of restricted fields. Furthermore, it aims to solve some of the key usability problems with PodSecurityPolicies, such as the inability to compose policies and the difficulty in binding policies.

**Help Wanted:** I don't actually have time to own the implementation of this proposal, so I'm looking for community members who are interested in realizing this. If we decide to move forward with this approach, my hope is that we can consolidate some of the efforts around other new restriction policies.

**Question**: Should this be rolled up into the overall [policy proposal](https://docs.google.com/document/d/1Ht8wpj4j9YfAA7aVv9Yn3Ci1T_MLMWt0DBr0QmxI2OM/edit#heading=h.6l8mozsiyzb1) or something like [OPA](http://www.openpolicyagent.org/docs/kubernetes-admission-control.html)? Alternatively, should this use a path-based OPA-style API?

Fix for https://github.com/kubernetes/kubernetes/issues/60001

/sig auth

### DIFF
--- a/contributors/design-proposals/auth/pod-restriction.md
+++ b/contributors/design-proposals/auth/pod-restriction.md
@@ -1,0 +1,477 @@
+# PodRestrictions
+
+## Overview
+
+This is a proposal for a new policy object (and associated admission controller) that adds basic
+restrictions to most fields on a Pod. This is motivated by a desire to unify the various policies we
+have now (which all behave in slightly different ways), and to unblock expanding the set of
+restricted fields. Furthermore, it aims to solve some of the key usability problems with
+PodSecurityPolicies, such as the inability to compose policies and the difficulty in binding
+policies.
+
+### Existing Policies
+
+There are multiple policies that constrain pod objects today. PodRestriction would supersede these
+policies, and they would eventually be deprecated.
+
+[AlwaysPullImages](https://kubernetes.io/docs/admin/admission-controllers/#alwayspullimages) -
+Admission controller that sets the PullAlways ImagePullPolicy to be set on every container.
+
+[PodSecurityPolicy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) - Restricts
+PodSecurityContext, SecurityContext, and some volume options. Bound to users and service accounts.
+
+[LimitRanger](https://kubernetes.io/docs/admin/admission-controllers/#limitranger) - Limits the
+range of resource limits that can be placed on a Pod. Intersection of LimitRanges in a namespace.
+
+[Toleration](https://kubernetes.io/docs/admin/admission-controllers/#podtolerationrestriction) &
+[NodeSelector](https://kubernetes.io/docs/admin/admission-controllers/#podnodeselector)
+restrictions - Restrict scheduling options through namespace annotations. Several proposals are out
+for new policies to supercede these ([#1937](https://github.com/kubernetes/community/pull/1937),
+[#48041](https://github.com/kubernetes/kubernetes/issues/48041),
+[#61185](https://github.com/kubernetes/kubernetes/issues/61185)).
+
+There have been other proposals for policies restricting fields like labels, and annotations.
+
+## Scope
+
+This proposal is about building a new policy type that can place restrictions on any or all pod
+fields (see [Monolith vs. Modular](#monolith-vs-modular)). However, the goal is not to combine all
+admission controllers into a single one. The following controls are explicitly out of scope:
+
+-   Defaulting (and mutations in general) - See [No Defaulting](#no-defaulting).
+
+-   External dependencies - The policy doesn't express any inter-object relationships, and doesn't
+    call out to external servers or have plugin points.
+
+-   Complex logic or interfield relationships - The goal is to keep the policy simple, both to
+    implement and reason about. The only exception is fields that explicitly reference others like
+    VolumeMounts and EnvVarSource, but most fields should be able to be validated without knowledge
+    of any other Pod fields.
+
+## Design
+
+### Policy Binding
+
+PodRestrictions are applied at the namespace level. Any Pod (or PodTemplate object) must conform to
+the policies matching the namespace it is created in.
+
+```go
+Binding:
+    Namespaces        []string
+    NamespaceSelector *metav1.LabelSelector
+```
+
+#### Justification
+
+There are a few different approaches taken by our current policies (see
+[#60001](https://github.com/kubernetes/kubernetes/issues/60001)).
+
+**Bind to Users**: Binding policy to users is the obvious first choice, but has a number of
+problems. First, users don't typically create pods, they create resources like Deployments that
+create pods on their behalf. That "on behalf" ends up being problematic to implement, especially
+with issues like confused deputy attacks, third party controllers, and changing policies.
+
+**Bind to ServiceAccounts**: To get around the user problem, PodSecurityPolicy adds the ability to
+bind to the target objects service account. ServiceAccount usage isn't actually authorized
+separately from namespace usage, so this just ends up being namespace-level authorization. However,
+because PodSecurityPolicy is authorized by either the source or target user, this ends up being less
+secure, since the Pod can use it's service account to create pods in other namespaces with the same
+access level. For instance, using PodSecurityPolicy you cannot have a controller that runs as
+privileged without opening up all pods it creates to running as privileged.
+
+**Namespaced Objects**: This approach has its advantages, but it doesn't scale well to large
+organizations with many namespaces (without a system on top managing all the policies across
+namespaces). See extensions below for how we might use namespaced objects with PodRestrictions.
+
+**Bind to Pod labels**: Pod labels don't have any authorization of usage without something like
+PodRestriction, so binding to labels isn't sufficient no it's own.
+
+**Bind to Namespaces**: The approach I'm recommending. Cluster level objects use a namespace
+selector to select the namespaces it applies to. It is consistent with the authorization of create
+requests, which are authorized at the namespace level. A small number of policies can be managed at
+the cluster level, and mapped to the namespaces they should apply to.
+
+#### Optional Extensions
+
+-   Namespaced PodRestrictions - PodRestriction objects can be created in a namespace, so that
+    namespace admins can place additional restrictions within their namespace.
+
+-   Additional Bindings - The NamespaceSelector is not a top-level field so that additional binding
+    approaches may be added in the future. Examples may include: ServiceAccounts, LabelSelector,
+    NodeSelector, etc.
+
+### Policy Matching - Union or Intersection?
+
+**Intersection** refers to the model where a Pod must be valid under all PodRestrictions that
+pertain to it. The advantage of intersecting is that restrictions can be composed. For example, I
+could have a policy that frontend & backend Pods cannot share a node, and a separate policy that all
+production Pods must drop privileges. Without the ability to intersect those policies, I need 3
+different versions (either, or, both). A problem with this approach is that running a privileged
+namespace (e.g. kube-system) means that every restriction needs an exception in the
+NamespaceSelector.
+
+**Union** refers to the model where a Pod must be valid under any PodRestriction that is bound to
+it. This approach makes it easier to reason about what the restrictions on a given namespace are,
+and easier to add exceptions for namespaces, but can lead to a combinatorial explosion of different
+policies that can become unmanageable.
+
+**Proposal**: Why not both? The default behavior is intersection, but you can punch a hole through
+for a specific namespace with a union approach. An optional `BindingMode` is added to the Binding
+which specifies the behavior. A pod must be valid under either:
+
+-   ANY `BindingModeAccept` policy that is bound to it, or
+
+-   ALL `BindingModeDrop` policies that are bound to it
+
+In other words, the Pod is first checked against Accept restrictions, and is accepted if any
+match. Then, it is checked against the Drop restrictions, and rejected if any don't match. The
+default behavior at the end is to accept the pod.
+
+```
+Binding:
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+privileged without opening up all pods it creates to running as privileged.
+
+**Namespaced Objects**: This approach has its advantages, but it doesn't scale well to large
+organizations with many namespaces (without a system on top managing all the policies across
+namespaces). See extensions below for how we might use namespaced objects with PodRestrictions.
+
+**Bind to Pod labels**: Pod labels don't have any authorization of usage without something like
+PodRestriction, so binding to labels isn't sufficient no it's own.
+
+**Bind to Namespaces**: The approach I'm recommending. Cluster level objects use a namespace
```
ravisantoshgudimetla: This is much more intuitive. I am not sure how well it works at scale. One case I could think of is having a lot of namespaces all having different labels based on which selection could happen resulting in delay of policy binding. This could be avoided by labeling all the namespaces with predefined value though.

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+selector to select the namespaces it applies to. It is consistent with the authorization of create
+requests, which are authorized at the namespace level. A small number of policies can be managed at
+the cluster level, and mapped to the namespaces they should apply to.
+
+#### Optional Extensions
+
+-   Namespaced PodRestrictions - PodRestriction objects can be created in a namespace, so that
```
ravisantoshgudimetla: Can you please elaborate on how to avoid conflicts between restriction objects created in namespace vs cluster-scoped ones? or are the restriction objects which are defined in cluster-scoped cannot be touched in namespace?

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+New CRDs that hold a PodTemplate and create Pods should be able to be registered against the
+PodRestriction controller. The registration should include a path (auto-discovery?) to extract the
+PodTemplate from the CRD.
+
+### API Design
+
+#### Monolith vs. Modular
+
+Our current approach to policies has been "modular", i.e. a different controller and API for each
+"area" of features. This has led to some inconsistencies and gaps in our policies, and makes adding
```
ravisantoshgudimetla: I think we have gone past that stage of "area" of features as well. While I am not completely against monolithic approach after your point, I think, having a delegation model to area would helpful and might be a good step. For example, podrestriction controller would delegate to scheduling admission controller which can add restrictions specific to scheduling. But a careful grouping of controllers to areas is very much needed.

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+-   Pods can only run as privileged if they run with a specific AntiAffinity
+
+-   Pods labeled as staging must not run on production nodes.
+
+#### No Defaulting
+
+There are 2 problems with mutating Pod fields: 1) it becomes hard to compose policies (what happens
+when the defaults conflict?) and 2) you need a policy ordering mechanism for when a Pod matches
+multiple policies. Instead, PodRestriction opts out of mutating defaults entirely, and leaves that
+for a different component to solve ([PodPreset?](../service-catalog/pod-preset.md)).
```
ravisantoshgudimetla: What if mutation is needed via webhooks, won't they be allowed as well? If yes, it may be too restrictive.

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+
+	RestartPolicy                 *StringRestriction
+	TerminationGracePeriodSeconds *NumericRestriction
+	ActiveDeadlineSeconds         *NumericRestriction
+	DNSPolicy                     *StringRestriction
+	NodeSelector                  *StringMapRestriction
+	AutomountServiceAccountToken  *BoolRestriction
+	SecurityContext               *PodSecurityContextRestriction
+
+    // TODO: Scheduling restrictions are deeply nested structures and require
```
ravisantoshgudimetla: I think grouping them might be helpful further. I could talk from scheduler perspective, how about having a schedulerrestriction which has affinity, schedulerName, tolerations etc. This is similar to composition of restrictions.

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+restrictions to most fields on a Pod. This is motivated by a desire to unify the various policies we
+have now (which all behave in slightly different ways), and to unblock expanding the set of
+restricted fields. Furthermore, it aims to solve some of the key usability problems with
+PodSecurityPolicies, such as the inability to compose policies and the difficulty in binding
+policies.
+
+### Existing Policies
+
+There are multiple policies that constrain pod objects today. PodRestriction would supersede these
+policies, and they would eventually be deprecated.
```
jhorwit2: The admission controllers mentioned here for the most part all perform mutation/defautling so how do you envision deprecating them in favor of this admission controller if it isn't supposed to mutate/default fields?

tallclair:
@erictune brought up a great point in another thread: Unifying all policies into a single resource means that you can't subdivide authorization on that resource. For instance, I can't have one admin who's responsible for setting the security constraints, and another responsible for setting the resource constraints.

A better approach might be to stick with the "modular" approach we have now, but define a consistent set of design principles that each policy should adhere to. In other words, all policies should bind, compose, and restrict in the same (or very similar) ways.

If we want to stick with the modular approach, I think we need to survey the scope of pod features and define a clearly delineated list of what policies we want, and what features fall into which policy. For instance, where should label restrictions go? SchedulingPolicy? LabelingPolicy?

contributors/design-proposals/auth/pod-restriction.md
jhorwit2: That sounds like it may get complex for PSP's since that admission controller has very specific logic around choosing a PSP to apply to a pod. 

cc @liggitt

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+for a specific namespace with a union approach. An optional `BindingMode` is added to the Binding
+which specifies the behavior. A pod must be valid under either:
+
+-   ANY `BindingModeAccept` policy that is bound to it, or
+
+-   ALL `BindingModeDrop` policies that are bound to it
+
+In other words, the Pod is first checked against Accept restrictions, and is accepted if any
+match. Then, it is checked against the Drop restrictions, and rejected if any don't match. The
+default behavior at the end is to accept the pod.
```
tallclair: Alternative approach - use a firewall analogy:

- Fields in the policy are just pattern-matching, there's no concept of per-field allow vs deny
- Policy action is `allow` or `deny`. If the field matchers in a policy match the pod, the action is taken.
- Policy has a priority specifying the order. Conflicts are resolved as deny (fail closed)

yastij: @tallclair - I'd prefer having a policy action which would be allow, deny or default

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+different versions (either, or, both). A problem with this approach is that running a privileged
+namespace (e.g. kube-system) means that every restriction needs an exception in the
+NamespaceSelector.
+
+**Union** refers to the model where a Pod must be valid under any PodRestriction that is bound to
+it. This approach makes it easier to reason about what the restrictions on a given namespace are,
+and easier to add exceptions for namespaces, but can lead to a combinatorial explosion of different
+policies that can become unmanageable.
+
+**Proposal**: Why not both? The default behavior is intersection, but you can punch a hole through
```
yastij: @tallclair - what if a Policy specifies intersection and another one union, and both are applicable, how do we resolve this ?

contributors/design-proposals/auth/pod-restriction.md
tallclair: What does a default action mean? Do you mean setting field defaults?

contributors/design-proposals/auth/pod-restriction.md
tallclair: I was proposing that the "union" (accept) policies are checked first. Alternatively, the "firewall" approach I mentioned uses priority and favors deny.

contributors/design-proposals/auth/pod-restriction.md
yastij: yes

contributors/design-proposals/auth/pod-restriction.md
ravisantoshgudimetla: ok, thanks for the clarification. After reading this first line the paragraph I got confused.

contributors/design-proposals/auth/pod-restriction.md
yastij: FYI i've added a [#1937(comment)](https://github.com/kubernetes/community/pull/1937#issuecomment-376107468) that sums up state of policy-related standards.

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+range of resource limits that can be placed on a Pod. Intersection of LimitRanges in a namespace.
+
+[Toleration](https://kubernetes.io/docs/admin/admission-controllers/#podtolerationrestriction) &
+[NodeSelector](https://kubernetes.io/docs/admin/admission-controllers/#podnodeselector)
+restrictions - Restrict scheduling options through namespace annotations. Several proposals are out
+for new policies to supercede these ([#1937](https://github.com/kubernetes/community/pull/1937),
+[#48041](https://github.com/kubernetes/kubernetes/issues/48041),
+[#61185](https://github.com/kubernetes/kubernetes/issues/61185)).
+
+There have been other proposals for policies restricting fields like labels, and annotations.
```
bgrant0607: https://github.com/kubernetes/kubernetes/issues/15390
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/metadata-policy.md

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+selector to select the namespaces it applies to. It is consistent with the authorization of create
+requests, which are authorized at the namespace level. A small number of policies can be managed at
+the cluster level, and mapped to the namespaces they should apply to.
+
+#### Optional Extensions
+
+-   Namespaced PodRestrictions - PodRestriction objects can be created in a namespace, so that
+    namespace admins can place additional restrictions within their namespace.
+
+-   Additional Bindings - The NamespaceSelector is not a top-level field so that additional binding
```
bgrant0607: It could be at the top level. It just needs to be optional.

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+
+2.  Restrictions are self-contained. There isn't complex interdependent logic here. If you want more
+    than basic whitelist / blacklist / pattern matching, then implement a new policy or controller
+    for that.
+
+3.  There are a lot of fields in the PodSpec, and thus this API gets large (see [Monolith
+    vs. Modular](#monolith-vs-modular)). We don't need to build it all at once, but I think over
+    time it will eventually grow into the full version.
+
+4.  Fields are unrestricted by default. This is important due to the size of the API. If I want to
```
bgrant0607: We need to think about API evolution: what should happen when new fields are added?

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+
+    Namespaces        []string
+    NamespaceSelector *meta.LabelSelector
+}
+
+type BindingMode string
+
+const (
+    BindingModeAccept BindingMode = "Accept"
+    BindingModeDrop   BindingMode = "Drop"
```
bgrant0607: I find the term "Drop" confusing. It's Deny, the opposite of Accept? (As opposed to, e.g., elide.)

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+	User  *StringRestriction
+	Role  *StringRestriction
+	Type  *StringRestriction
+	Level *StringRestriction
+}
+
+type CapabilitiesRestriction struct {
+	PointerRestriction
+	Add  *StringListRestriction
+	Drop *StringListRestriction
```
bgrant0607: Again, does this mean deny or elide?

contributors/design-proposals/auth/pod-restriction.md
```diff
@@ -0,0 +1,477 @@
...
+
+type CollectionRestriction struct {
+	ForbidEmpty  bool
+	RequireEmpty bool
+}
+
+type StringRestriction struct {
+	PointerRestriction
+
+	Whitelist []string
```
bgrant0607: Does this mean include/exclude, or accept/deny?

What's the rationale for per-field accept/deny rather than top-level accept/deny?

tallclair:
See updates on https://github.com/kubernetes/kubernetes/issues/60001#issuecomment-378776927

My recommendation is that we focus on the SchedulingPolicy proposal for now (possibly applying some of the ideas from this proposal), and close this specific proposal.

PR #1953, opened 2018-03-21
Update design doc based on code

orainxiong:
update design pattern doc with :
* Integrating binding with scheduling
* Scheduler Changes
* Update volume-topology-scheduling.png

@kubernetes/sig-storage
cc @msau42


orainxiong:
update volume-topology-scheduling.png

orainxiong:
cc @lichuqiang

msau42:
Thanks for taking a close look at this design doc.  I also made some similar updates in https://github.com/kubernetes/community/pull/1857.  I think we can just include it all in change to make it easier to merge.  Could you compare the two and see if anything is missing?

orainxiong:
OK.

PR #1956, opened 2018-03-21
KEP for CoreDNS to GA.

rajansandeep:
This KEP is defined to graduate CoreDNS to GA.

### DIFF
--- a/keps/sig-network/0010-20180314-coredns-GA-proposal.md
+++ b/keps/sig-network/0010-20180314-coredns-GA-proposal.md
@@ -1,0 +1,126 @@
+---
+kep-number: 10
+title: Graduate CoreDNS to GA
+authors:
+  - "@johnbelamaric"
+  - "@rajansandeep"
+owning-sig: sig-network
+participating-sigs:
+  - sig-cluster-lifecycle
+reviewers:
+  - "@bowei"
+  - "@thockin"
+approvers:
+  - "@thockin"
+editor: "@rajansandeep"
+creation-date: 2018-03-21
+last-updated: 2018-05-18
+status: provisional
+see-also: https://github.com/kubernetes/community/pull/2167
+---
+ 
+# Graduate CoreDNS to GA
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [User Cases](#use-cases)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+
+## Summary
+
+CoreDNS is sister CNCF project and is the successor to SkyDNS, on which kube-dns is based. It is a flexible, extensible
+authoritative DNS server and directly integrates with the Kubernetes API. It can serve as cluster DNS,
+complying with the [dns spec](https://git.k8s.io/dns/docs/specification.md). As an independent project,
+it is more actively developed than kube-dns and offers performance and functionality beyond what kube-dns has. For more details, see the [introductory presentation](https://docs.google.com/presentation/d/1v6Coq1JRlqZ8rQ6bv0Tg0usSictmnN9U80g8WKxiOjQ/edit#slide=id.g249092e088_0_181), or [coredns.io](https://coredns.io), or the [CNCF webinar](https://youtu.be/dz9S7R8r5gw).
+
+Currently, we are following the road-map defined [here](https://github.com/kubernetes/features/issues/427). CoreDNS is Beta in Kubernetes v1.10, which can be installed as an alternate to kube-dns.
+The purpose of this proposal is to graduate CoreDNS to GA.
+
+## Motivation
+
+* CoreDNS is more flexible and extensible than kube-dns. 
+* CoreDNS is easily extensible and maintainable using a plugin architecture.
+* CoreDNS has fewer moving parts than kube-dns, taking advantage of the plugin architecture, making it a single executable and single process.
+* It is written in Go, making it memory-safe (kube-dns includes dnsmasq which is not). 
+* CoreDNS has [better performance](https://github.com/kubernetes/community/pull/1100#issuecomment-337747482) than [kube-dns](https://github.com/kubernetes/community/pull/1100#issuecomment-338329100) in terms of greater QPS, lower latency, and lower memory consumption. 
+
+### Goals
+
+* Bump up CoreDNS to be GA.
+* Make CoreDNS available as an image in a Kubernetes repository (To Be Defined) and ensure a workflow/process to update the CoreDNS versions in the future.
+  May be deferred to [next KEP](https://github.com/kubernetes/community/pull/2167) if goal not achieved in time.
+* Provide a kube-dns to CoreDNS upgrade path with configuration translation in `kubeadm`.
+* Provide a CoreDNS to CoreDNS upgrade path in `kubeadm`.
+
+### Non-Goals
+
+* Translation of CoreDNS ConfigMap back to kube-dns (i.e., downgrade).
+* Translation configuration of kube-dns to equivalent CoreDNS that is defined outside of the kube-dns ConfigMap. For example, modifications to the manifest or `dnsmasq` configuration.
+* Fate of kube-dns in future releases, i.e. deprecation path.
+* Making [CoreDNS the default](https://github.com/kubernetes/community/pull/2167) in every installer.
+
+## Proposal
+
+The proposed solution is to enable the selection of CoreDNS as a GA cluster service discovery DNS for Kubernetes.
+Some of the most used deployment tools have been upgraded by the CoreDNS team, in cooperation of the owners of these tools, to be able to deploy CoreDNS:
+* kubeadm
+* kube-up
+* minikube
+* kops
+
+For other tools, each maintainer would have to add the upgrade to CoreDNS.
+
+### Use Cases
+
+* CoreDNS supports all functionality of kube-dns and also addresses [several use-cases kube-dns lacks](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/network/coredns.md#use-cases). Some of the Use Cases are as follows: 
+    * Supporting [Autopath](https://coredns.io/plugins/autopath/), which reduces the high query load caused by the long DNS search path in Kubernetes.
+    * Making an alias for an external name [#39792](https://github.com/kubernetes/kubernetes/issues/39792)
+    
+* By default, the user experience would be unchanged. For more advanced uses, existing users would need to modify the ConfigMap that contains the CoreDNS configuration file.
+* Since CoreDNS has more supporting features than kube-dns, there will be no path to retain the CoreDNS configuration in case a user wants to switch to kube-dns.
+
+#### Configuring CoreDNS
+
+The CoreDNS configuration file is called a `Corefile` and syntactically is the same as a [Caddyfile](https://caddyserver.com/docs/caddyfile). The file consists of multiple stanzas called _server blocks_.
+Each of these represents a set of zones for which that server block should respond, along with the list of plugins to apply to a given request. More details on this can be found in the 
+[Corefile Explained](https://coredns.io/2017/07/23/corefile-explained/) and [How Queries Are Processed](https://coredns.io/2017/06/08/how-queries-are-processed-in-coredns/) blog entries.
+
+The following can be expected when CoreDNS is graduated to GA.
+
+#### Kubeadm
+
+* The CoreDNS feature-gates flag will be marked as GA.
+* As Kubeadm maintainers chose to deploy CoreDNS as the default Cluster DNS for Kubernetes 1.11:
+    * CoreDNS will be installed by default in a fresh install of Kubernetes via kubeadm.
+    * For users upgrading Kubernetes via kubeadm, it will install CoreDNS by default whether the user had kube-dns or CoreDNS in a previous kubernetes version.
+    * In case a user wants to install kube-dns instead of CoreDNS, they have to set the feature-gate of CoreDNS to false. `--feature-gates=CoreDNS=false`
+* When choosing to install CoreDNS, the configmap of a previously installed kube-dns will be automatically translated to the equivalent CoreDNS configmap.
+
+#### Kube-up
+
+* CoreDNS will be installed when the environment variable `CLUSTER_DNS_CORE_DNS` is set to `true`. The default value is `false`.
+
+#### Minikube
+
+* CoreDNS to be an option in the add-on manager, with CoreDNS disabled by default.
+
+## Graduation Criteria
+
+* Verify that all e2e conformance and DNS related tests (xxx-kubernetes-e2e-gce, ci-kubernetes-e2e-gce-gci-ci-master and filtered by `--ginkgo.skip=\\[Slow\\]|\\[Serial\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]`) run successfully for CoreDNS.
+  None of the tests successful with Kube-DNS should be failing with CoreDNS.
+* Add CoreDNS as part of the e2e Kubernetes scale runs and ensure tests are not failing.
+* Extend [perf-tests](https://github.com/kubernetes/perf-tests/tree/master/dns) for CoreDNS.
+* Add a dedicated DNS related tests in e2e scalability test [Feature:performance].
+
+## Implementation History
+
+* 20170912 - [Feature proposal](https://github.com/kubernetes/features/issues/427) for CoreDNS to be implemented as the default DNS in Kubernetes.
+* 20171108 - Successfully released [CoreDNS as an Alpha feature-gate in Kubernetes v1.9](https://github.com/kubernetes/kubernetes/pull/52501).
+* 20180226 - CoreDNS graduation to Incubation in CNCF.
+* 20180305 - Support for Kube-dns configmap translation and move up [CoreDNS to Beta](https://github.com/kubernetes/kubernetes/pull/58828) for Kubernetes v1.10.


keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+authoritative DNS server and directly integrates with the Kubernetes API. It can serve as cluster DNS,
+complying with the [dns spec](https://git.k8s.io/dns/docs/specification.md). As an independent project,
+it is more actively developed than kube-dns and offers performance and functionality beyond what kube-dns has. For more details, see the [introductory presentation](https://docs.google.com/presentation/d/1v6Coq1JRlqZ8rQ6bv0Tg0usSictmnN9U80g8WKxiOjQ/edit#slide=id.g249092e088_0_181), or [coredns.io](https://coredns.io), or the [CNCF webinar](https://youtu.be/dz9S7R8r5gw).
+
+Currently, we are following the road-map defined [here](https://github.com/kubernetes/features/issues/427). CoreDNS is Beta in Kubernetes v1.10, which can be installed as an alternate to kube-dns.
+The purpose of this proposal is to graduate CoreDNS to GA and be the default DNS service, replacing the current implementation of kube-dns.
+
+## Motivation
+
+    * CoreDNS is more flexible and extensible than what kube-dns.
```
johnbelamaric: @rajansandeep let's not indent these, we don't want it showing up as pre-formatted.

johnbelamaric: s/than what/than/

fturib:
I think you need to add the file that hold the NEXT-KEP-NUMER.
increment the number, and add the file in this PR. (to be sure no one else will take the number)

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+
+* Have CoreDNS run as the default DNS server both after a new installation and an upgrade of the existing installation.
+* Make CoreDNS available as an [image in gcr.io](https://github.com/kubernetes/kubernetes/pull/59753) and define a workflow/process to update the CoreDNS versions in the future.
+* Provide a kube-dns to CoreDNS upgrade path with configuration translation.
+* Retain the configuration of a CoreDNS deployment from Alpha (1.9) or Beta (1.10).
+
+### Non-Goals
+
+* Translation of CoreDNS ConfigMap back to kube-dns (i.e., downgrade).
+* Retaining the ConfigMap of a custom/modified image of CoreDNS during upgrade. That is, if the user has modified the Corefile in the ConfigMap, those changes will not be retained (the user should back it up and redo the changes on the new ConfigMap).
```
stp-ip: Feels quite intrusive to overwrite user configuration on an upgrade without a path to prevent this. At least needs to be documented or warned for imo.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+* Translation configuration of kube-dns to equivalent CoreDNS that is defined outside of the kube-dns ConfigMap. For example, modifications to the manifest or `dnsmasq` configuration.
```
stp-ip: Could be considered a breaking change and needs to be documented imo.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
rajansandeep: Ok, I can document that and also make this statement a little clearer. 

Just to clarify, at upgrade time, the ConfigMap is not retained only if the coredns image we find is "unofficial" (not shipped with alpha (1.9) or beta (1.10)... ) and coredns will be upgraded to the latest image and default ConfigMap.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
bowei: This should be required as there a users of kube-dns with the configmap, and their setups will break if this is not provided.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+* CoreDNS to be enabled by default in the add-on manager, with kube-dns disabled by default.
+ 
+## Unresolved questions
+
+* Fate of kube-dns in future releases, i.e. deprecation path.
+
+## Graduation Criteria
+
+* Passing all e2e conformance and DNS related tests.
+
```
bowei: Scalability tests

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+authoritative DNS server and directly integrates with the Kubernetes API. It can serve as cluster DNS,
+complying with the [dns spec](https://git.k8s.io/dns/docs/specification.md). As an independent project,
+it is more actively developed than kube-dns and offers performance and functionality beyond what kube-dns has. For more details, see the [introductory presentation](https://docs.google.com/presentation/d/1v6Coq1JRlqZ8rQ6bv0Tg0usSictmnN9U80g8WKxiOjQ/edit#slide=id.g249092e088_0_181), or [coredns.io](https://coredns.io), or the [CNCF webinar](https://youtu.be/dz9S7R8r5gw).
+
+Currently, we are following the road-map defined [here](https://github.com/kubernetes/features/issues/427). CoreDNS is Beta in Kubernetes v1.10, which can be installed as an alternate to kube-dns.
+The purpose of this proposal is to graduate CoreDNS to GA and be the default DNS service, replacing the current implementation of kube-dns.
+
+## Motivation
+
+* CoreDNS is more flexible and extensible than what kube-dns.
```
miekg: drop "what" here.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+* CoreDNS is easily extensible and maintainable using a plugin architecture.
+* CoreDNS has fewer moving parts than kube-dns, taking advantage of the plugin architecture, making it a single executable and single process.
+* It is written in Go, making it memory-safe (kube-dns includes dnsmasq which is not). 
+* CoreDNS has [better performance](https://github.com/kubernetes/community/pull/1100#issuecomment-337747482) than [kube-dns](https://github.com/kubernetes/community/pull/1100#issuecomment-338329100) in terms of greater QPS, lower latency, and lower memory consumption. 
+* It supports a number of [use cases](https://coredns.io/2017/05/08/custom-dns-entries-for-kubernetes/) that kube-dns doesn't. As a general-purpose authoritative DNS server it has a lot of functionality that kube-dns couldn't reasonably be expected to add.
+
+### Goals
+
+* Have CoreDNS run as the default DNS server both after a new installation and an upgrade of the existing installation.
+* Make CoreDNS available as an [image in gcr.io](https://github.com/kubernetes/kubernetes/pull/59753) and define a workflow/process to update the CoreDNS versions in the future.
```
miekg: Should probably file a coredns issue for this as well.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
miekg: autoscaling?

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+* CoreDNS is more flexible and extensible than what kube-dns. 
+* CoreDNS is easily extensible and maintainable using a plugin architecture.
+* CoreDNS has fewer moving parts than kube-dns, taking advantage of the plugin architecture, making it a single executable and single process.
+* It is written in Go, making it memory-safe (kube-dns includes dnsmasq which is not). 
+* CoreDNS has [better performance](https://github.com/kubernetes/community/pull/1100#issuecomment-337747482) than [kube-dns](https://github.com/kubernetes/community/pull/1100#issuecomment-338329100) in terms of greater QPS, lower latency, and lower memory consumption. 
+* It supports a number of [use cases](https://coredns.io/2017/05/08/custom-dns-entries-for-kubernetes/) that kube-dns doesn't. As a general-purpose authoritative DNS server it has a lot of functionality that kube-dns couldn't reasonably be expected to add.
+
+### Goals
+
+* Have CoreDNS run as the default DNS server both after a new installation and an upgrade of the existing installation.
```
smarterclayton: +1 on this.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
smarterclayton: There isn’t anything that prevents them from continuing to use Kube-dns, which is backwards compatible.

I don’t think we should claim support for arbitrary customizations, and “customizer beware” is our default stance.

fturib: @bowei : we are converting the kube-dns ConfigMap to a CoreDns ConfigMap. this "non-goal" statement is to warn that a customized deployment of kube-dns with extra options on the cmd-lines of dnsmasq or kube-dns containers will not be translated.

in other words we support translation "StubDomains", "UpstreamNameServer" and "Federation" is supported.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
fturib: @miekg : autoscaling is generic and should work with coredns. 
However I noticed an issue in the e2e tests of autoscaling with DNS - because the test itself waits for kube-dns to run. This issue will have to be fixed. See https://github.com/kubernetes/kubernetes/issues/61176

keps/sig-network/0006-20180314-coredns-GA-proposal.md
fturib: @miekg : maybe, but I propose we wait to have a better understanding of who is doing what in the new process. Specifically who will have the responsibility to generating the image and push to gcr.io

keps/sig-network/0006-20180314-coredns-GA-proposal.md
fturib: @bowei : what have you in mind ?
- a manual scale test, as we did here : https://github.com/kubernetes/community/pull/1100#issuecomment-337747482 (using Kubemark)
- or an extension of these [perf-test for DNS ](https://github.com/kubernetes/perf-tests)(adapting those for CoreDNS) ? 
- or another kind of scalability included in kubernetes/tests/e2e ?

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+The purpose of this proposal is to graduate CoreDNS to GA and be the default DNS service, replacing the current implementation of kube-dns.
+
+## Motivation
+
+* CoreDNS is more flexible and extensible than what kube-dns. 
+* CoreDNS is easily extensible and maintainable using a plugin architecture.
+* CoreDNS has fewer moving parts than kube-dns, taking advantage of the plugin architecture, making it a single executable and single process.
+* It is written in Go, making it memory-safe (kube-dns includes dnsmasq which is not). 
+* CoreDNS has [better performance](https://github.com/kubernetes/community/pull/1100#issuecomment-337747482) than [kube-dns](https://github.com/kubernetes/community/pull/1100#issuecomment-338329100) in terms of greater QPS, lower latency, and lower memory consumption. 
+* It supports a number of [use cases](https://coredns.io/2017/05/08/custom-dns-entries-for-kubernetes/) that kube-dns doesn't. As a general-purpose authoritative DNS server it has a lot of functionality that kube-dns couldn't reasonably be expected to add.
```
thockin: This is outside the DNS spec for Kube, so it is pretty irrelevant to this plan

keps/sig-network/0006-20180314-coredns-GA-proposal.md
thockin: We should probably keep a fork of CoreDNS in github.com/kubernetes (or in a new, related org) and cut our own containers from there.  We should be doing that for everything we ship.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
thockin: Agree - this is user-hostile.  If you expose a ConfigMap you are publishing it as a mechanism to configure the app.  If you intend to blow it away, you should just not expose it. at all.

Now we have to talk about defaults vs user-overrides.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
thockin: To be clear - I read this as "we WILL support graceful upgrades from the kube-dns configmap, but we will NOT support upgrade from arbitrary other config" - is that right?

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+
+For other tools, each maintainer would have to add the upgrade to CoreDNS.
+
+### Use Cases
+
+* CoreDNS supports all functionality of kube-dns and also addresses several use-cases kube-dns lacks. Some of the Use Cases are as follows: 
+    * Supporting [Autopath](https://coredns.io/plugins/autopath/), which reduces the high query load caused by the long DNS search path in Kubernetes
+    * Making an alias for an external name [#39792](https://github.com/kubernetes/kubernetes/issues/39792)
+    * Dynamically adding services to another domain, without running another server [#55](https://github.com/kubernetes/dns/issues/55)
+    * Adding an arbitrary entry inside the cluster domain (for example TXT entries [#38](https://github.com/kubernetes/dns/issues/38))
```
thockin: I do not think we should document, endorse, or emphasize operations that do not go through Kubernetes API, except with the explicit warning that it is coredns specific and non-portable.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
bowei: Sorry about the terseness, I mean that we should get the CoreDNS running as a part of regular k8s scale tests.

Having it possible to be run as target of perf-test for DNS would be great.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
johnbelamaric: Ok, but this is just for the ConfigMap, right? That is, how should we handle changes to the manifest/deployment itself?

In this case then we should add the new, updated default ConfigMap under a different name, to make it easy for the user to switch to that if they want to, or at least compare any differences. This also means that in the event of non-backwards compatible changes to the Corefile (which we might do, but only when we bump the minor - not patch - version), we will need to do something special.

@rajansandeep can you make sure the 1.0.x -> 1.1.x remains backwards compatible for the Corefile? If so we don't need to worry about that for a while.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
johnbelamaric: @thockin, yes. The options that can be configured in the kube-dns ConfigMap will be retained in the conversion to the Corefile. But we aren't going to tweak things for arbitrary config changes to the dnsmasq CLI options, etc. If the user has that, they should probably retain the kube-dns installation until they can upgrade in a more considered manner.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
johnbelamaric: We have looked at the perf-test for DNS, we can do that but it will require a lot of changes to those tests. They are geared toward discovering/testing varying levels of resources devoted to dnsmasq and kube-dns.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
johnbelamaric: Ok. So are you thinking we should pull these from the KEP?

stp-ip: Maybe instead of listing the various extra features of CoreDNS, just link to a CoreDNS specific document. That way it's linked, but can be made clear it's out of scope for this KEP and k8s in general.

keps/sig-network/0006-20180314-coredns-GA-proposal.md
bowei: I think we should have an integration style test for CoreDNS in that style given the surprises we've experienced wrt to resource limits etc from the past.

fturib: @bowei : Let me verify I understand your 2 points:

### CoreDNS as part of the regular k8s scale runs (e.g. up to 5k nodes)

That means have CoreDNS be part of the regular k8s scale runs at 5k nodes
Therefore ensure we have a DNS e2e tests with the flag "[Feature:Performance]". 
It will automatically run with all perf-tests like "ci-kubernetes-kubemark-gce-scale" (which is the one running 5k nodes with Kubemark)


###  integration test style resource determination

From this sentence, and your comment before, I understand you advise to extend those **[perf-test for DNS]( https://github.com/kubernetes/perf-tests/tree/master/dns)** and create some dedicated tests for CoreDNS to measure the resources used by CoreDNS under stress of queries.

**is my understanding correct ?**

bowei: lgtm

keps/sig-network/0006-20180314-coredns-GA-proposal.md
fturib: @johnbelamaric : what about this proposition for  upgrade 1-ConfigMap, 2-Deployment
1- never update the ConfigMap (as you suggest). It is created first time of installation of CoreDNS (either by upgrade from kube-dns, either by installation from scratch), and on subsequent upgrade, we keep that ConfigMap - Until we reach the issue of non-backward compatible corefile. At that time we figure out with special upgrade.

NOTE: does it worth to create a "coredns-default" ConfigMap ? It would show no more than the one of the coredns/deployment repository since kube-dns config no longer exist.

2- For the deployment part, first installation creates the deployment of CoreDNS. Subsequent upgrades only change the image of CoreDNS - unless admin prevent it by removing the label "kubeadm-auto-upgrade" that the first installation is setting.

fturib:
@bowei @thockin : the KEP has been updated with the comments. Can you have a second look to check it integrates all your concerns. Thanks!

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,125 @@
...
+#### Minikube
+
+* CoreDNS to be enabled by default in the add-on manager, with kube-dns disabled by default.
+ 
+## Unresolved questions
+
+* Fate of kube-dns in future releases, i.e. deprecation path.
+
+## Graduation Criteria
+
```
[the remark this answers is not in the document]
rajansandeep: @rramkumar1 Just to clarify:
I submit a PR to merge now making CoreDNS default in kubeadm, kube-up... to see if the tests are "passing" and if we're satisfied with the result, we keep the PR, else we revert.

And what is the right number of runs to determine it's a success?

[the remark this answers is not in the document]
fturib: @rramkumar1 : why would we need to specify "amount of test runs" ? I would think that one is enough no ? as long as all test are OK with CoreDNS then CoreDNS is acceptable.

Are you worry about kind of flaking tests ? or other commits that would interfere and failed some of the  "conformance" tests (independently of DNS service) ?

To prevent to have a wrong view do to other errors )not linked to the DNS service), we could propose to have one of the "conformance" + "DNS" tests running with kube-dns and one with CoreDNS. so we can validate we have the same level of passing tests.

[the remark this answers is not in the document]
fturib: For e2e test, I would propose:

- all tests filtered by "--ginkgo.skip=\\[Slow\\]|\\[Serial\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]"
(the regular tests "xxx-kubernetes-e2e-gce").

As the success of each test depends of the code itself (not only the DNS Service), we would verify that CoreDNS is equivalent to kube-dns in regards of the passed e2e tests. That means validate
"ci-kubernetes-e2e-gce-gci-ci-master" against the exact same test "ci-kubernetes-e2e-gce-gci-ci-master-kube-dns" with kube-dns deployed. 

They are supposed to be always the same, or at least none of the success tests with kube-dns is failing with coredns.

fturib:
@rramkumar1 : the point on e2e tests to validate is modified to be more precise. We are not proposing a sequence of run to be green, but rather a comparison for the same commit between CoreDNS and the former kube-dns : we want to validate that CoreDNS is able to replace kube-dns.

For the e2e performance tests, the same way, we just want that those performances are not hit by CoreDNS, in other words we want the test to stay green.

Let me know if you agree with the new verbiage.
Thanks!

bowei:
It looks like this proposal does two things:

* Make CoreDNS GA.
* Make CoreDNS the default.

I think it would be good to have these be separate. I support taking CoreDNS to GA in this release, but would like to see one release of miles on the feature before flipping it to the default implementation.

fturib:
@bowei : CoreDNS is already running in v1.9 (as ALPHA), and v1.10 (as BETA).
The meaning of GA was to be "default". 

So what would you expect to see/monitor during that "one release" where CoreDNS is GA but not "default" ?

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,123 @@
...
+* [Proposal](#proposal)
+    * [User Cases](#use-cases)
+* [Unresolved Questions](#unresolved-questions)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+
+## Summary
+
+CoreDNS is sister CNCF project and is the successor to SkyDNS, on which kube-dns is based. It is a flexible, extensible
+authoritative DNS server and directly integrates with the Kubernetes API. It can serve as cluster DNS,
```
krmayankk: what does directly integrates with kubernetes API mean ? KubeDNS also does that right ?

keps/sig-network/0006-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,123 @@
...
+it is more actively developed than kube-dns and offers performance and functionality beyond what kube-dns has. For more details, see the [introductory presentation](https://docs.google.com/presentation/d/1v6Coq1JRlqZ8rQ6bv0Tg0usSictmnN9U80g8WKxiOjQ/edit#slide=id.g249092e088_0_181), or [coredns.io](https://coredns.io), or the [CNCF webinar](https://youtu.be/dz9S7R8r5gw).
+
+Currently, we are following the road-map defined [here](https://github.com/kubernetes/features/issues/427). CoreDNS is Beta in Kubernetes v1.10, which can be installed as an alternate to kube-dns.
+The purpose of this proposal is to graduate CoreDNS to GA.
+
+## Motivation
+
+* CoreDNS is more flexible and extensible than kube-dns. 
+* CoreDNS is easily extensible and maintainable using a plugin architecture.
+* CoreDNS has fewer moving parts than kube-dns, taking advantage of the plugin architecture, making it a single executable and single process.
```
krmayankk: i am guessing the `fewer moving parts` is explained later ?

krmayankk:
@rajansandeep i am working on enabling kubedns and then i saw these. After reading this proposal, i am still not clear what functionality does CoreDNS offer that it should be made the default dns solution for kubernetes. Few questions:-
 - is the user experience for normal(not advanced) uses cases exactly the same as kube-dns?
- is configuring core-dns any worse then kube-dns ?
- what exactly are the features in CoreDns that i should move to CoreDns rather than kube-dns.
Antyhing default always on in kubernetes should be very simple to use and offer majority of the common use cases without turning a lot of knobs. Does this achieve that ?

keps/sig-network/0010-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,124 @@
...
+* Make CoreDNS available as an image in a Kubernetes repository (To Be Defined) and ensure a workflow/process to update the CoreDNS versions in the future.
+  May be deferred to [next KEP](https://github.com/kubernetes/community/pull/2167) if goal not achieved in time.
+* Provide a kube-dns to CoreDNS upgrade path with configuration translation in `kubeadm`.
+* Provide a CoreDNS to CoreDNS upgrade path in `kubeadm`.
+
+### Non-Goals
+
+* Translation of CoreDNS ConfigMap back to kube-dns (i.e., downgrade).
+* Translation configuration of kube-dns to equivalent CoreDNS that is defined outside of the kube-dns ConfigMap. For example, modifications to the manifest or `dnsmasq` configuration.
+* Fate of kube-dns in future releases, i.e. deprecation path.
```
thockin: Making coredns the default in every installer

thockin:
one small nit - please xref the other KEP

rajansandeep:
@thockin done.
Let me know if there are any other comments to be addressed for approval! Thanks!

keps/sig-network/0010-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,126 @@
...
+reviewers:
+  - "@bowei"
+  - "@thockin"
+approvers:
+  - "@thockin"
+editor: "@rajansandeep"
+creation-date: 2018-03-21
+last-updated: 2018-05-18
+status: provisional
+succeeded by: https://github.com/kubernetes/community/pull/2167
```
johnbelamaric: I think you were thinking of "superseded-by", but we don't want to use that in this KEP (it is for ones that get replaced by a new KEP). Instead use "see-also" (which takes a list).

keps/sig-network/0010-20180314-coredns-GA-proposal.md
```diff
@@ -0,0 +1,126 @@
...
+  May be deferred to [next KEP](https://github.com/kubernetes/community/pull/2167) if goal not achieved in time.
+* Provide a kube-dns to CoreDNS upgrade path with configuration translation in `kubeadm`.
+* Provide a CoreDNS to CoreDNS upgrade path in `kubeadm`.
+
+### Non-Goals
+
+* Translation of CoreDNS ConfigMap back to kube-dns (i.e., downgrade).
+* Translation configuration of kube-dns to equivalent CoreDNS that is defined outside of the kube-dns ConfigMap. For example, modifications to the manifest or `dnsmasq` configuration.
+* Fate of kube-dns in future releases, i.e. deprecation path.
+* Making CoreDNS the default in every installer.
```
johnbelamaric: Reference/link the new KEP here too.

fturib:
Thanks @thockin !

PR #1957, opened 2018-03-21
Updated SIG-OpenStack meeting time for daylight saving

hogepodge:
Updated SIG-OpenStack entry and associated files with updated meeting information, changed to reflect daylight saving.


cblecker:
@hogepodge Would it be helpful to express the meeting time in a different timezone?
Example: https://github.com/kubernetes/community/blob/efc2221c7e1fe8f8ea13fc14a0b8cc2d6276c7af/sigs.yaml#L19-L23

Expressing it in a timezone that uses DST will ensure the conversion link stays up to date

hogepodge:
@cblecker good advice. Updated. Thanks!

sigs.yaml
```diff
@@ -1138,9 +1138,9 @@ sigs:
         company: Ticketmaster
     meetings:
     - description: Regular SIG Meeting
-      day: Thursday
-      time: "00:00"
-      tz: "UTC"
+      day: Wednesday
+      time: "16:00"
+      tz: "PT"
```
cblecker: Unfortunately, you have to actually use the format `"PT (Pacific Time)"` so that the time zone converter works

PR #1958, opened 2018-03-21
Move KEP-5 to implementable

jbeda:
This is a final call for comments and tweaks to KEP 5 before we move forward with implementation. Please let me know of any last concerns or tweaks that need to be made.

(Obviously, we can change things if this isn't working, but this marks that we are confident enough in this plan to start implementing it.  Experience with that can lead to further modifications here or abandoning this KEP altogether.)

### DIFF
--- a/keps/sig-contributor-experience/0005-contributor-site.md
+++ b/keps/sig-contributor-experience/0005-contributor-site.md
@@ -14,7 +14,7 @@ approvers:
 editor: TBD
 creation-date: "2018-02-19"
 last-updated: "2018-03-07"
-status: provisional
+status: implementable
 ---
 
 # Contributor Site


cblecker:
@jbeda Question about the process.. so moving to implementable, means that someone can work on it, and create a POC/designs/etc.. but depending on how that POC goes, we can either a) keep moving forward, b) go back to the drawing board, or c) scrap it, yes?

jbeda:
I'd view it as a commitment to invest in the effort.  The bar to undo that direction gets higher over time. Nothing is impossible but the assumption is that, unless there is a huge roadblock or a big change of situation, things move forward.

At the end of the day the sponsoring SIG is the owner of the process and can make changes as necessary. But this is about signaling where efforts are going. There is an assumption here that people are reasonable and open about how this evolves over time.  If not, the SIG is the tie breaker.

cblecker:
Yup, that makes total sense. This raises the bar from "we're talking about it" to "we have enough consensus to try and make this work", but not so far as "this is the point of no return, and we are committed to this no matter what".

cblecker:
Oh, and to be clear: I support moving this to implementable :)

jbeda:
There are no "point of no return".  It's just software.  We can change it. ;)

jbeda:
@parispittman ping?

parispittman:
Lets move to the next step @jbeda. No other comments have come in (to here, me, or through the group).

jbeda:
Thanks Paris! I'll be on vacation next week but hope to get started after that.  Was going to get it started this weekend but life happened.

PR #1963, opened 2018-03-22
Update seccomp design doc to enable default seccomp profile

wangzhen127:
Updates kubernetes/kubernetes#39845
Feature issue kubernetes/features#135

This PR updates the seccomp design doc to indicate that we are enabling a default seccomp profile: docker/default. It pulls a few pieces from #660, but does not include the seccomp spec. The reasoning here is that we can simply enable docker/default as the first step and see how it works. If strong needs of a different default k8s seccomp profile come up, we can add that later.

### DIFF
--- a/contributors/design-proposals/node/seccomp.md
+++ b/contributors/design-proposals/node/seccomp.md
@@ -197,7 +197,7 @@ profiles to be opaque to kubernetes for now.
 
 The following format is scoped as follows:
 
-1.  `docker/default` - the default profile for the container runtime, can be
+1.  `runtime/default` - the default profile for the container runtime, can be
     overwritten by the following two.
 2.  `unconfined` - unconfined profile, ie, no seccomp sandboxing
 3.  `localhost/<profile-name>` - the profile installed to the node's local seccomp profile root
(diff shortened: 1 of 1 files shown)


wangzhen127:
/cc @jessfraz @tallclair 

Can you guys take a look first?

contributors/design-proposals/node/seccomp.md
```diff
@@ -191,7 +197,8 @@ profiles to be opaque to kubernetes for now.
 
 The following format is scoped as follows:
 
-1.  `docker/default` - the default profile for the container runtime
+1.  `docker/default` - the default profile for the container runtime, can be
```
liggitt: this is the default for all runtimes? or just the docker runtime?

smarterclayton: Yeah, should be `runtime/default` or `container/default` or `default`.

wangzhen127: I agree that we should use a general term to cover all kinds of runtimes. The current immediate step is to enable docker/default for docker runtime, which covers a lot of our use cases. Having a general seccomp default will need spec, ect.

wangzhen127:
Updated docker/default to runtime/default now.

PR #1969, opened 2018-03-26
Add wg apply

apelisse:
Adding wg-apply in the sigs list. We probably don't have all the information (leads? calendar?) yet, but this is still better than nothing.


apelisse:
@lavalamp

lavalamp:
LGTM but I don't know what the requirements are for this :)

cblecker:
also, is this a separate working group, or would this be a subproject of an existing sig?

lavalamp:
TBH I just made a mailing list so I could have a group to email without spamming everyone.

The work we'll be doing has a green light from SIG Architecture. Members of SIG API Machinery and (at least) SIG CLI will be working on this.

cblecker:
probably best to classify it as a subproject then, but cc: @kubernetes/steering-committee for guidance.

bgrant0607:
WG kind of makes sense because it involves 2 SIGs for now: API machinery and CLI.

Eventually, it will be just API machinery, I assume. It's up to API machinery whether it makes sense to be permanently broken out as a subproject. If not, that reaffirms the WG direction.

lavalamp:
I imagine this group will dissolve after we are "done" if that makes a
difference.

bgrant0607:
Yes, that suggests not a subproject.

apelisse:
OK, Thanks everyone. Looks like we know where we're going. I'll make sure we get there and fill the gaps. I'll ping here when ready.

apelisse:
Proposal has been sent, and I've updated the document with the regular meeting description. Please proceed.

PR #1973, opened 2018-03-27
design: improved service account token volumes

mikedanese:
Refactoring of https://docs.google.com/document/d/1Ki7K9rAC5XIAgyYJXUis1UpgwmtRq2Xz68mF3_LggEY/edit?ts=5a5e7205

@liggitt @smarterclayton 
@kubernetes/sig-auth-feature-requests @kubernetes/sig-storage-feature-requests


contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+
+A new volume projection will be implemented with an API that closely matches the
+TokenRequest API.
+
+```go
+type ServiceAccountTokenProjection struct {
+  // Audiences are the intendend audiences of the token.
+  Audiences []string
+  // ExpirationSeconds is the requested duration of validity of the service
+  // account token
```
smarterclayton: Comment about what default value means

contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+To replace the current service account token secrets, we also need to inject the
+clusters CA certificate bundle. Initially we will deploy to data in a configmap
+per-namespace and reference it using a ConfigMapProjection.
+
+This fixes one scalability issue with the current service account token
+deployment model where secret GETs are a large portion of overall apiserver
+traffic. However it doesn't solve the per-namespace replicated CA data which
+needs to be stored in ETCD.
+
+To fix that issue we can implement a PublicConfigMapProjection that exactly
```
smarterclayton: Another option: extend config map source to perform an authorization check on the service accounts behalf to fetch a cross namespace config map.  That removes the need for a special config map source and makes users who have this use case happier.

Authorization is two checks - node authorizer for the reference, and a corresponding acl check as the service account or the node (arguably I should be able to grant my service account permission to fetch something and let the node fetch it)

smarterclayton: The node authorizer can do that check on the behalf of the service account:

1. If referenced
2. And if service account has access if outside of namespace

mikedanese: Authorization referring to secrets is tracked in https://github.com/kubernetes/kubernetes/issues/4957

jbeda:
It would be great if the reviewers for the doc included some folks that weren't also listed as authors.  The purpose of getting these things in PRs is to find independent review.

contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+  @liggitt
+  @mikedanese
+
+## Summary
+
+Kubernetes is able to provide pods with unique identity tokens that can prove
+the caller is a particular pod to a Kubernetes API server. These tokens are
+injected into pods as secrets. This proposal proposes a new mechanism of
+distribution with support for [improved service account
+tokens](https://github.com/kubernetes/community/pull/1460) and explores how to
```
jbeda: A PR is not sufficient here.  We should make sure that anything referenced by a design doc is checked in and agreed upon.  We *have* to get out of the habit of not closing these things out.  We need to be able to actually make decisions.

mikedanese: I can prioritize closing that out. The outstanding question is here:

https://github.com/kubernetes/community/pull/1460#discussion_r166636787

contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+1.  Implement a ServiceAccountToken volume projection that that maintains a
+    service account token requested by the node from the TokenRequest API.
+
+### Token Volume Projection
+
+A new volume projection will be implemented with an API that closely matches the
+TokenRequest API.
+
+```go
+type ServiceAccountTokenProjection struct {
```
jbeda: I don't see the name of the service account here.  Have you considered making it so that a Pod can use multiple SAs at the same time?  I can see scenarios where, say, a set of microservices get merged into one service and the things these things are talking to other services with unique SAs. Having the new merged service be able to use the existing SAs would be nice.

mikedanese:
I hope to get some sig-storage reviewers of this proposal after it's presented at the next sig meeting.

contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+A volume plugin implemented in the kubelet will project a service account token
+sourced from the TokenReview API into volumes created from
+ProjectedVolumeSources. As the token approaches expiration, the kubelet volume
+plugin will proactively rotate the service account token.
+
+To replace the current service account token secrets, we also need to inject the
+clusters CA certificate bundle. Initially we will deploy to data in a configmap
```
jbeda: Having a way to *just* take the cluster CA bundle would be nice. This would allow a service to receive jwts from other services without having a SA token itself.

jbeda: Ah -- ignore -- thought this was part of the SA token projection..

contributors/design-proposals/auth/svcacct-token-volume-source.md
jbeda: That issue is about to be closed due to inactivity.

contributors/design-proposals/auth/svcacct-token-volume-source.md
jbeda: If you want to use the KEP process (not required but might be nice -- hint hint) there is an idea of checking in stuff that is WIP and isolating the nut of disagreement to dedicated PRs.  That helps to find the areas agreement and get them locked down incrementally.

contributors/design-proposals/auth/svcacct-token-volume-source.md
smarterclayton: Jordan proved to me my idea wouldn't work because the node authorizer doesn't know the who.  I think his argument was to not put this in the API but make it be a kubelet detail.  I think it's pretty reasonable for the short term not to introduce a new API here but instead make it part of kube config.

contributors/design-proposals/auth/svcacct-token-volume-source.md
smarterclayton: Nothing prevents you mounting one of their secrets today, so i think it's reasonable to extend that.  I think it's a reasonable addition, but is also something that could be added later without impact (an optional field where the default is the pod's service account).  Do you think a ton of people have this right now Joe?  The only example I've seen has been prometheus which has a scraper token and a service account token - I would definitely use something like that there, but that's 1 out of several hundred apps I've seen people deploy, so i'd suspect its 5% case (a useful one).  I don't *think* any of the other core services have come up with this use case yet.

contributors/design-proposals/auth/svcacct-token-volume-source.md
smarterclayton: Although we still need a way to tie the ca info (the implicit service account data) to a pod that doesn't request the secret.  The API on ServiceAccountToken is token path specific, but without the CA the token is pretty useless, and in-cluster config is broken.

smarterclayton: Would the admission controller automatically inject a public config map projection?

contributors/design-proposals/auth/svcacct-token-volume-source.md
jbeda: I also think that this may be more interesting as people start mapping to Cloud SAs.  You may have one SA for talking within the cluster and another SA to talk to the cloud.  Ideally these would be the same but it depends on how flexible the mappings are.

I'm just allergic to any 1:1 bindings for stuff like this. In my experience it almost always comes back to bite you.  And if we don't have it in there earlier than folks will tend to work with what they have vs. come back and add something.  However, I wouldn't lay down on the tracks if this didn't make it in the first version.

I think it is also worth calling out (perhaps in an example) that a pod can project the same SA multiple times with different audiences. Without making this clear people will naively just add more audiences and pass that token around. This kind of defeats the whole purpose of the audience field. In my mind having multiple audiences on a token is a bad smell and we shouldn't make that the easy/suggested/obvious case.

mikedanese: This would add a lot of questions to the design (e.g. how do these extra service accounts interact with PodSecurityPolicy, ImagePullSecrets attached to the service account, etc...). As Clayton says, this functionality could be added backwards compatibly later and I would prefer to solve this with a holistic design for sub-pod isolation.

mikedanese: > think it is also worth calling out (perhaps in an example) that a pod can project the same SA multiple times with different audiences. Without making this clear people will naively just add more audiences and pass that token around.

Good point.

contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+per-namespace and reference it using a ConfigMapProjection.
+
+This fixes one scalability issue with the current service account token
+deployment model where secret GETs are a large portion of overall apiserver
+traffic. However it doesn't solve the per-namespace replicated CA data which
+needs to be stored in ETCD.
+
+To fix that issue we can implement a PublicConfigMapProjection that exactly
+copies the implementation of ConfigMapProjection but sources config from the
+`kube-public` namespace.
```
jbeda: I *think* that this is the first time that `kube-public` is being used as part of the core controllers.  We introduced it as part of the kubeadm bootstrapping process but it was completely optional. Not sure if other stuff has shown up in there.

It is a little strange to make a namespace special like this.  In fact, `kube-system` isn't all that special above and beyond convention (and RBAC defaults, I think?).  Introducing a PublicConfigMapProjection is probably something that deserves its own design doc.  I'd suggest something like listing the namespace and requiring that the target configmap be readable by anonymous.  That would allow folks to use this in other contexts and avoid naming `kube-public` explicitly.

bgrant0607:
@kubernetes/sig-auth-feature-proposals @kubernetes/sig-storage-feature-proposals

contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+  ExpirationSeconds *int64
+  // Path is the relative path of the file to project the token into.
+  Path string
+}
+```
+
+A volume plugin implemented in the kubelet will project a service account token
+sourced from the TokenReview API into volumes created from
+ProjectedVolumeSources. As the token approaches expiration, the kubelet volume
+plugin will proactively rotate the service account token.
```
saad-ali: Can you provide the struct for a `ProjectedVolumeSource`?

contributors/design-proposals/auth/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,138 @@
...
+  // account token
+  ExpirationSeconds *int64
+  // Path is the relative path of the file to project the token into.
+  Path string
+}
+```
+
+A volume plugin implemented in the kubelet will project a service account token
+sourced from the TokenReview API into volumes created from
+ProjectedVolumeSources. As the token approaches expiration, the kubelet volume
```
saad-ali: Can you provide the struct for `ProjectedVolumeSource` for clarity?

mikedanese: Done.

saad-ali:
/approve

@jsafrane could you PTAL, will leave it to you to LGTM

/assign @jsafrane

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,168 @@
...
+To replace the current service account token secrets, we also need to inject the
+clusters CA certificate bundle. Initially we will deploy to data in a configmap
+per-namespace and reference it using a ConfigMapProjection.
+
+This fixes one scalability issue with the current service account token
+deployment model where secret GETs are a large portion of overall apiserver
+traffic. However it doesn't solve the per-namespace replicated CA data which
+needs to be stored in ETCD.
+
+To fix that issue we can extend ConfigMapProjection to support cross namespace
```
mikedanese: @smarterclayton PTAL

jsafrane:
LGTM from storage point of view, the volume plugin spec is OK.

I personally don't like hardcoded `kube-public` namespace, but others already pointed that out.

Anyone from sig-auth to check the security part of this?

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,168 @@
...
+  // ExpirationSeconds is the requested duration of validity of the service
+  // account token
+  ExpirationSeconds *int64
+  // Path is the relative path of the file to project the token into.
+  Path string
+}
+```
+
+A volume plugin implemented in the kubelet will project a service account token
+sourced from the TokenReview API into volumes created from
```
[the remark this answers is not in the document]
mikedanese: This type is supposed to match TokenRequest. TokenReview here is a typo, thanks for the catch. TokenRequest supports tokens with multiple audiences (for better or worse):

https://github.com/kubernetes/kubernetes/blob/c36e66dd66d04eafeb5e6668b1939244c62720a8/pkg/apis/authentication/types.go#L108-L114

TokenReview doesn't have audiences yet:

https://github.com/kubernetes/kubernetes/blob/c36e66dd66d04eafeb5e6668b1939244c62720a8/pkg/apis/authentication/types.go#L59

 but when it does, it will be an array:

https://github.com/kubernetes/kubernetes/pull/62802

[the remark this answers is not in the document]
verult: cool, thanks for the clarification. I was looking at the design doc [here](https://github.com/mikedanese/community/blob/eb687869434b31ed6fa8c3553d7eea17034133de/contributors/design-proposals/auth/bound-service-account-tokens.md#modify-tokenreviewsauthenticationk8sio) which has a single Audience in the TokenReviewSpec.

Another follow-up question: `Audiences` and `ExpirationSeconds` is filled in by the TokenRequest, right? So the only pre-specified fields are the TokenRequest name (I assume you are using name to identify the TokenRequest) and `Path`?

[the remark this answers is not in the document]
mikedanese: Audiences and ExpirationSeconds are config provided by the pod author, but have defaults of ["k8s-api"] and 1 hour. However, a user might want a token with audience ["vault"] and expiration seconds 3 hours, and they should be able to configure that in the volume projection.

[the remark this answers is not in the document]
verult: Who creates the TokenRequest object? The TokenProjection volume plugin, not the pod author, right?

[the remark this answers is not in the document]
mikedanese: Yes, the token volume plugin (with kubelet's client) makes the TokenRequest.

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Proposal
+
+Kubernetes should:
+
+1.  Allow the automatic service account token injection to be disabled via
+    configuration.
+1.  Allow automatic service account token creation to occur only when a user
+    explicitly requests.
+1.  Implement a ServiceAccountToken volume projection that that maintains a
```
spikecurtis: "that that"

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,168 @@
...
+clusters CA certificate bundle. Initially we will deploy to data in a configmap
+per-namespace and reference it using a ConfigMapProjection.
+
+This fixes one scalability issue with the current service account token
+deployment model where secret GETs are a large portion of overall apiserver
+traffic. However it doesn't solve the per-namespace replicated CA data which
+needs to be stored in ETCD.
+
+To fix that issue we can extend ConfigMapProjection to support cross namespace
+configmap references. Initially, we can hardcode a rule int the NodeAuthorizer
```
spikecurtis: int -> into

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,168 @@
...
+that allows nodes and pods to reference a specific configmap in the kube-public
+namespace that contains the root ca.crt data with the goal of improving the
+permission model to support cross namespace configmap references generally. See
+[#4957](https://github.com/kubernetes/kubernetes/issues/4957).
+
+A projected volume source that is equivalent to the current service account
+secret:
```
liggitt: also the namespace, which you can replicate with a downwardAPI projection for metadata.namespace

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,168 @@
...
+  ExpirationSeconds *int64
+  // Path is the relative path of the file to project the token into.
+  Path string
+}
+```
+
+A volume plugin implemented in the kubelet will project a service account token
+sourced from the TokenRequest API into volumes created from
+ProjectedVolumeSources. As the token approaches expiration, the kubelet volume
+plugin will proactively rotate the service account token.
```
liggitt: should this rotation be optional and opt-in? (e.g. allow for a short-lived token that is only used at startup and then expires?)

liggitt: also, if the kubelet is going to maintain this (presumably rotating at some percentage of the token's lifetime), there should probably be a minimum ExpirationSeconds allowed (we don't want to make kubelets rotate every 3 seconds)

contributors/design-proposals/storage/svcacct-token-volume-source.md
liggitt: options I see:

1. per-namespace configmap (one per namespace, which is much better than one per serviceaccount token, but still requires controller maintenance and has etcd size and kubelet request issues)
2. single cross-namespace configmap (opens pandora's box to cross-namespace references, access model is unclear)
3. downward API for apiserver trust root (non-pod fields in downward API is a shift I'm unsure of, still requires the kubelet to get the data somehow, but avoids the source being part of the pod API)
4. additional projection source for apiserver trust root (still requires the kubelet to get the data somehow, but avoids the source being part of the pod API)

1 seems ok as a starting point, I think I prefer 3 or 4 to making cross-namespace references part of the API

contributors/design-proposals/storage/svcacct-token-volume-source.md
mikedanese: > should this rotation be optional and opt-in? (e.g. allow for a short-lived token that is only used at startup and then expires?)

interesting idea. I'd be happy to add this functionality when there is a non hypothetical usecase. we would probably need to refetch every container restart.

> also, if the kubelet is going to maintain this (presumably rotating at some percentage of the token's lifetime), there should probably be a minimum ExpirationSeconds allowed (we don't want to make kubelets rotate every 3 seconds)

I suggest validating that requests must have validity durations for longer than 10 minutes.

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,169 @@
...
+    audience: ca.istio.io
+```
+
+However this alone doesn't solve the per-namespace replicated CA data which
+needs to be stored in ETCD. To fix that issue we can extend ConfigMapProjection
+to support cross namespace configmap references. Initially, we can hardcode a
+rule into the NodeAuthorizer that allows nodes and pods to reference a specific
+configmap in the kube-public namespace that contains the root ca.crt data with
+the goal of improving the permission model to support cross namespace configmap
+references generally. See [kubernetes/kubernetes#4957][4597].
```
tallclair: s/4597/4957/

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,173 @@
...
+type VolumeProjection struct {
+  Secret *SecretProjection
+  DownwardAPI *DownwardAPIProjection
+  ConfigMap *ConfigMapProjection
+  ServiceAccountToken *ServiceAccountTokenProjection
+}
+
+type ServiceAccountTokenProjection struct {
+  // Audience is the intendend audience of the token.
+  Audience string
```
mikedanese: Note audience is singular here which is a divergence from the TokenRequest API. The hope here is that people will be less likely to use multiple audiences insecurely. If a use case for multiple audiences comes along, we can add support for it easily.

mikedanese:
I've pulled out the configmap stuff. I'm going to address it in a followup PR.

contributors/design-proposals/storage/svcacct-token-volume-source.md
liggitt: how does one indicate a request for an API token? omit audience, just like with TokenRequest?

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,136 @@
...
+  // ExpirationSeconds is the requested duration of validity of the service
+  // account token
+  ExpirationSeconds *int64
```
liggitt: is this required or optional? if optional, what is the default?

mikedanese: This is optional and the default matches the default for TokenRequest which is currently 1 hour. I'll add a note in the api doc.

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,136 @@
...
+  // Path is the relative path of the file to project the token into.
+  Path string
```
liggitt: verify with @kubernetes/sig-storage-proposals we have the ability to update content of individual files mounted into a container... I vaguely remember issues with that

contributors/design-proposals/storage/svcacct-token-volume-source.md
liggitt: minimum sounds reasonable to me, at least as long as the kubelet is responsible for rotation. briefly describe the rotation scheme ("start trying to rotate at x% or y seconds prior to expiration, whichever is earlier", etc)

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+A volume plugin implemented in the kubelet will project a service account token
+sourced from the TokenRequest API into volumes created from
+ProjectedVolumeSources. As the token approaches expiration, the kubelet volume
+plugin will proactively rotate the service account token.
+
+To replace the current service account token secrets, we also need to inject the
+clusters CA certificate bundle. Initially we will deploy to data in a configmap
+per-namespace and reference it using a ConfigMapProjection.
```
liggitt: include an example of the initial replacement for the service account token, including the token projection, downward API projection for the namespace, and configmap projection for the CA

liggitt:
pared down to a single new projection source, this LGTM as a building block we can use in a variety of approaches moving forward.

/assign @jsafrane @msau42 
would like an ack from someone in @kubernetes/sig-storage-api-reviews on the viability of refreshing content of single-file projected mounts

/assign @tallclair 
as Joe mentioned, would like an ack from a @kubernetes/sig-auth-api-reviews reviewer not listed as an author

/cc @wojtek-t 
in 2-3 releases, I expect this to greatly lessen the quantity of secret mounts the kubelet maintains, and replace it with specific token mounts (which the kubelet does not need to watch for, and can schedule refreshes of much more effectively)

contributors/design-proposals/storage/svcacct-token-volume-source.md
mikedanese: Done.

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Risks and Mitigations
+
+Reducing the scope of service account tokens by not creating them automatically
+on service account creation is technically an API break. This would have to be
+opt-in, but like RBAC is about reducing the scope of vulnerability. Many people
+may opt not to disable it. For those who do disable it, we can preserve the
+existing behavior of being able to create a secret of type service-account-token
+(with the annotation to the service account that links to it) and have the
+controller auto populate it.
```
tallclair: I don't understand this last sentence. Does "disable it" refer to disabling the automounting of tokens, or disabling the feature of not auto-mounting by default?

I also don't understand the workaround. Shouldn't `automountServiceAccountToken` always automount the credential, even if we change the default to false?

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,164 @@
...
+
+FlexVolume and CSI are the only way to deliver custom content to nodes today
+without persisting it in the API. In a virtual kubelet environment, these
+mechanisms may not work the same as on regular kubelets, so third party identity
+integrators may not be able to deliver their custom content. A container volume
+or init container might be a sufficient workaround.
```
tallclair: I also don't understand this paragraph. Is this an alternative implementation? I thought tokens from the tokenrequest API aren't persisted in the etcd, is that not true?

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,164 @@
...
+scalability of the platform.
+
+## Proposal
+
+Kubernetes should:
+
+1.  Allow the automatic service account token injection to be disabled via
+    configuration.
+1.  Allow automatic service account token creation to occur only when a user
+    explicitly requests.
```
tallclair: The risks section briefly touches on these first two points, but I wasn't clear on what you are proposing. Disabling credential auto-mounting seems like something that should be addressed independently of ServiceAccountTokenProjection volumes.

jsafrane:
> would like an ack from someone in @kubernetes/sig-storage-api-reviews on the viability of refreshing content of single-file projected mounts

From storage point of view, this is just another AtomicWriter volume and it will behave the same as a Secret with one key, right? I don't see any issue here.

liggitt:
> From storage point of view, this is just another AtomicWriter volume and it will behave the same as a Secret with one key, right? I don't see any issue here.

sounds good. so projected can refresh single-file, it was subPath mounts that could not?

msau42:
Correct because subpath is bind mounted by the runtime, the symlink gets resolved and fixed.

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,143 @@
...
+// the pods runtime filesystem for use against APIs (Kubernetes API Server or
+// otherwise).
+type ServiceAccountTokenProjection struct {
+  // Audience is the intended audience of the token. A recipient of a token
+  // must identify itself with an identifier specified in the audience of the
+  // token, and otherwise should reject the token.
+  Audience string
+  // ExpirationSeconds is the requested duration of validity of the service
+  // account token. Defaults to 1 hour.
+  ExpirationSeconds *int64
```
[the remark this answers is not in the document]
mikedanese: This was addressed in https://github.com/kubernetes/community/pull/1973#discussion_r183810989

The minimum will be 10 minutes checked in validation. I'll add that to the comment.

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,143 @@
...
+
+// ServiceAccountTokenProjection represents a projected service account token
+// volume. This projection can be used to insert a service account token into
+// the pods runtime filesystem for use against APIs (Kubernetes API Server or
+// otherwise).
+type ServiceAccountTokenProjection struct {
+  // Audience is the intended audience of the token. A recipient of a token
+  // must identify itself with an identifier specified in the audience of the
+  // token, and otherwise should reject the token.
+  Audience string
```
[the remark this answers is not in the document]
mikedanese: The default is the same as the token request API. The audience defaults to the identifier of the apiserver.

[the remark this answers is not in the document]
mikedanese: I had a conversation with @liggitt about this. We are opting for a single audience to encourage correct usage of the API. We discussed a few options and decided this was probably the best approach for now.

contributors/design-proposals/storage/svcacct-token-volume-source.md
```diff
@@ -0,0 +1,143 @@
...
+  Path string
+}
+```
+
+A volume plugin implemented in the kubelet will project a service account token
+sourced from the TokenRequest API into volumes created from
+ProjectedVolumeSources. As the token approaches expiration, the kubelet volume
+plugin will proactively rotate the service account token. The kubelet will start
+trying to rotate the token if the token is older than 80 percent of its time to
+live or if the token is older than 24 hours.
```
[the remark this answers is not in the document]
mikedanese: Done.

mikedanese:
@tallclair PTAL

PR #1975, opened 2018-03-27
TODO:  Equality-based & Set-based filtering by fields

andrewpsp:
lists that return objects with labels should support label filtering and most lists should support filtering by fields.

### DIFF
--- a/contributors/devel/api-conventions.md
+++ b/contributors/devel/api-conventions.md
@@ -123,9 +123,12 @@ filtering (see [the labels documentation](https://kubernetes.io/docs/user-guide/
 lists should support filtering by fields.
 
    Examples: PodLists, ServiceLists, NodeLists
-
-   TODO: Describe field filtering below or in a separate doc.
-
+   
+   *Equality-based filtering by label keys and values.*
+   ``` NodeLists != DaemonSet-foo ```  
+   
+    *Set-based filtering keys according to a set of values* 
+   ``` DaemonSet-bar in (PodLists, ServiceLists) ```
 3. **Simple** kinds are used for specific actions on objects and for
 non-persistent entities.
 


contributors/devel/api-conventions.md
```diff
@@ -123,9 +123,12 @@ filtering (see [the labels documentation](https://kubernetes.io/docs/user-guide/
 lists should support filtering by fields.
 
    Examples: PodLists, ServiceLists, NodeLists
-
-   TODO: Describe field filtering below or in a separate doc.
-
+   
+   *Equality-based filtering by label keys and values.*
+   ``` NodeLists != DaemonSet-foo ```
```
liggitt: this doesn't really make sense... I think the TODO was looking for an example API call that demonstrated field filtering

bgrant0607:
cc @kubernetes/sig-api-machinery-proposals @kubernetes/api-approvers

contributors/devel/api-conventions.md
lavalamp: +1 I don't understand the new text and it is not fulfilling the TODO

andrewpsp: My understanding for the TODO - provide use case of field filtering.  Under that assumption  Daemonset-foo would not be in  NodeLists. ie a Daemonset would be set to to more specific set of Nodes like Nodelists-bar.  It may make more sense to include an API call instead.

bgrant0607:
Some field filtering is supported today. We need to document what is actually supported, which isn't what this PR does.

bgrant0607:
@andrewpsp Thanks for the PR. In the future, I recommend to reaching out to people via slack or mailing lists to ask for help/clarification/direction. Github notifications aren't very effective due to the extreme volume.

Issue #1894, opened 2018-03-05
Other unanswered questions in cherry-picks.md


MikeSpreitzer:
I made a PR against the master branch of kubernetes, and it needs to be cherry-picked to some older releases.  I have read https://github.com/kubernetes/community/blob/master/contributors/devel/cherry-picks.md but it leaves me with some questions.

cherry-picks.md says, in the section on "Propose a Cherry Pick", that "Cherrypicks are managed with labels and milestones", and the last part is a hyperlink to https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md --- which is probably not what is intended.

I am very much in the dark regarding labels and milestones.  cherry-picks.md says "first ensure the following labels are on the original master branch PR" but I do not know how to make that happen.  I do not even know whether it is something I do for myself (if so, how?) or something that I get somebody else to do (if so, who?).

When there are multiple releases that need a cherry-pick, cherry-picks.md does not say whether they are done serially  or in parallel.  As noted in #994, cherry-picks.md does not say how to handle multiple cherry-picks.

MikeSpreitzer:
@kubernetes/sig-contributor-experience-bugs

cblecker:
cc: @spiffxp

nikhita:
/remove-lifecycle stale

There is a PR open to improve the cherrypicks doc.

nikhita:
The cherry-pick doc recently went through a round of improvements in https://github.com/kubernetes/community/pull/2408 and https://github.com/kubernetes/community/pull/1980. It was also updated from the test-infra side so that the bot reacts with useful comments for cherry-pick PRs https://github.com/kubernetes/test-infra/pull/9128.

Closing as this is now done.
/close

PR #1980, opened 2018-03-28
clarify some parts of the cherry pick process

tpepper:
The cherry-pick instruction guide had some readability issues in
its formatting, with numbered bullet lists not rendering correctly.
It also referenced an issue which was resolved, so I've added links
to the resulting tools and information.  Finally, based on watching
a new user attempt a cherry pick (a common occurance around release
time with the revolving staffing of the release team) a few
clarification and setup issues are described around the
hack/cherry_pick_pull.sh script.

### DIFF
--- a/contributors/devel/cherry-picks.md
+++ b/contributors/devel/cherry-picks.md
@@ -1,44 +1,34 @@
 # Overview
 
-This document explains how cherry picks are managed on release branches within the
-Kubernetes projects.  Patches are either applied in batches or individually
-depending on the point in the release cycle.
+This document explains how cherry picks are managed on release
+branches within the Kubernetes projects.
 
-## Propose a Cherry Pick
+## Prerequisites
+ * [Contributor License Agreement](http://git.k8s.io/community/CLA.md)
+   is considered implicit for all code within cherry-pick pull requests,
+   ***unless there is a large conflict***.
+ * A pull request merged against the master branch.
+ * [Release branch](https://git.k8s.io/release/docs/branching.md) exists.
+ * The normal git and GitHub configured shell environment for pushing
+   to your kubernetes `origin` fork on GitHub and making a pull request
+   against a configured remote `upstream` that tracks
+   "https://github.com/kubernetes/kubernetes.git", including
+   `GITHUB_USER`.
+ * Have `hub` installed, which is most easily installed via `go get
+   github.com/github/hub` assuming you have a standard golang development
+   environment.
 
-1. Cherrypicks are [managed with labels and milestones](/contributors/guide/release-notes.md)
-1. To get a PR merged to the release branch, first ensure the following labels
-   are on the original **master** branch PR:
-  * The milestone for the branch that you want to cherry-pick to (e.g. v1.7 if you want to backport your change in 1.7 branch)
-  * The `cherrypick-candidate` label
-1. If `release-note-none` is set on the master PR, the cherrypick PR will need
-   to set the same label to confirm that no release note is needed.
-1. `release-note` labeled PRs generate a release note using the PR title by
-   default OR the release-note block in the PR template if filled in.
-  * See the [PR template](https://git.k8s.io/kubernetes/.github/PULL_REQUEST_TEMPLATE.md) for more details.
-  * PR titles and body comments are mutable and can be modified at any time
-    prior to the release to reflect a release note friendly message.
-
-### How do cherrypick-candidates make it to the release branch?
-
-1. **BATCHING:** After a branch is first created and before the X.Y.0 release
-  * Branch owners review the list of `cherrypick-candidate` labeled PRs.
-  * PRs batched up and merged to the release branch get a `cherrypick-approved`
-label and lose the `cherrypick-candidate` label.
-  * PRs that won't be merged to the release branch, lose the
-`cherrypick-candidate` label.
-
-1. **INDIVIDUAL CHERRYPICKS:** After the first X.Y.0 on a branch
-  * Run the cherry pick script. This example applies a master branch PR #98765
-to the remote branch `upstream/release-3.14`:
-`hack/cherry_pick_pull.sh upstream/release-3.14 98765`
-  * Your cherrypick PR (targeted to the branch) will immediately get the
-`do-not-merge/cherry-pick-not-approved` label. The branch owner will triage PRs
-targeted to the branch and label the ones to be merged by applying the `lgtm`
-label.
-
-There is an [issue](https://github.com/kubernetes/kubernetes/issues/23347) open
-tracking the tool to automate the batching procedure.
+## Initiate a Cherry Pick
+ * Run the [cherry pick script](https://git.k8s.io/kubernetes/hack/cherry_pick_pull.sh).
+   This example applies a master branch PR #98765 to the remote branch
+   `upstream/release-3.14`: `hack/cherry_pick_pull.sh upstream/release-3.14
+   98765`
+ * Your cherrypick PR targeted to the release branch will immediately get the
+   `do-not-merge/cherry-pick-not-approved` label. The release branch owner
+   will triage PRs targeted to the branch.  Normal rules apply for code merge.
+   * Reviewers `/lgtm` and owners `/approve` as they deem appropriate.
+   * The approving release branch owner is responsible for applying the
+     `cherrypick-approved` label.
 
 ## Cherry Pick Review
 
@@ -51,8 +41,3 @@ requested - this should not be the norm, but it may happen.
 
 See the [cherrypick queue dashboard](http://cherrypick.k8s.io/#/queue) for
 status of PRs labeled as `cherrypick-candidate`.
-
-[Contributor License Agreements](http://releases.k8s.io/HEAD/CONTRIBUTING.md) is
-considered implicit for all code within cherry-pick pull requests, ***unless
-there is a large conflict***.
-


tpepper:
PTAL
/assign @dims

contributors/devel/cherry-picks.md
```diff
@@ -7,39 +7,42 @@ depending on the point in the release cycle.
...
+ * PRs that won't be merged to the release branch, lose the `cherrypick-candidate` label.
+
+There is a helper tool [`branchff`](https://git.k8s.io/release/branchff) and
+more information on this workflow in the
+[branching](https://git.k8s.io/release/docs/branching.md) documentation.
+
+#### Individual Cherrypicks
+ * Run the [cherry pick script](https://git.k8s.io/kubernetes/hack/cherry_pick_pull.sh). This example applies a master branch PR #98765 to the remote branch `upstream/release-3.14`: `hack/cherry_pick_pull.sh upstream/release-3.14 98765`
+ * NOTE:
+    * For this script to work you need the normal git and GitHub configured shell environment for pushing to your kubernetes `origin` fork on GitHub and making a pull request against a configured remote `upstream` that tracks "http://git.k8s.io/kubernetes".
```
dims: is `http://git.k8s.io/kubernetes` legit?

tpepper: @cblecker is this the preferred way to reference it versus https://github.com/kubernetes/kubernetes/ ?

spiffxp:
taking a look, untangling how cherry picks actually work is hurting my head

I'm also unsure whether we should have this instead document the prow-based cherrypicker that's being worked on

contributors/devel/cherry-picks.md
```diff
@@ -7,39 +7,42 @@ depending on the point in the release cycle.
...
-  * The milestone for the branch that you want to cherry-pick to (e.g. v1.7 if you want to backport your change in 1.7 branch)
-  * The `cherrypick-candidate` label
-1. If `release-note-none` is set on the master PR, the cherrypick PR will need
-   to set the same label to confirm that no release note is needed.
-1. `release-note` labeled PRs generate a release note using the PR title by
-   default OR the release-note block in the PR template if filled in.
-  * See the [PR template](https://git.k8s.io/kubernetes/.github/PULL_REQUEST_TEMPLATE.md) for more details.
-  * PR titles and body comments are mutable and can be modified at any time
-    prior to the release to reflect a release note friendly message.
+
```
spiffxp: This is more of a note to self as I work through this:
- `submit-queue` is actually the thing responsible for merging PR's to `release-1.x`
- `submit-queue` won't bother if there is a `do-not-merge/cherry-pick-unapproved` label
- the `label-pick-unapproved` munger run as part of the `misc-mungers` deployment applies this label
  - it removes the label if there is a `cherrypick-approved` label
  - it applies the label if a PR is not for `master`
- the `cherrypick-approved` label can either be applied by a human, or by the `cherrypick-auto-approve` munger
- the `cherrypick-auto-approve` munger
  - only looks at PR's not for `master`
  - skips PR's that have `cherrypick-approved` and a milestone
  - skips PR's that don't mention other PR's (hereafter called "parents") in their description
  - skips PR's that are for `release-1.x` but have milestone `v1.y`
  - skips PR's whose parents don't have the `cherrypick-approved` label
  - skips PR's that are for `release-1.x` whose parents have milestone `v1.y` or none
  - applies `cherrypick-approved` and milestone `1.x` if it's made it this far
- various mungers will remove the `cherrypick-candidate` label
  - from "parent" PR's that have successfully merged
  - from "parent" PR's that lack a milestone
  - etc.

spiffxp: Some thoughts:
- the current auto-approve process assumes one PR to master == one cherrypick, and discounts that some fixes may need to be backported to multiple release branches
- I had to double-check, but it sure looks like `cherrypick-candidate` is just used for display purposes, I'm not sure if there is automation that consumes the cherrypick queue
- Given that there is no bot-driven way to apply the `cherrypick-candidate` label, it's more likely a PR author without repo write access would just run the cherypick script themselves

I dislike this whole process and suggest that it's _way_ easier to just get a human to apply the `cherrypick-approved` label

cjwagner:
Please let me know of any pain points or feature requests related to the mungegithub cherrypick mungers. They are due for a revamp and migration to Prow.

contributors/devel/cherry-picks.md
```diff
@@ -7,39 +7,42 @@ depending on the point in the release cycle.
...
+ * Branch owners review the list of `cherrypick-candidate` labeled PRs.
+ * PRs batched up and merged to the release branch get a `cherrypick-approved` label and lose the `cherrypick-candidate` label.
+ * PRs that won't be merged to the release branch, lose the `cherrypick-candidate` label.
+
+There is a helper tool [`branchff`](https://git.k8s.io/release/branchff) and
+more information on this workflow in the
+[branching](https://git.k8s.io/release/docs/branching.md) documentation.
+
+#### Individual Cherrypicks
+ * Run the [cherry pick script](https://git.k8s.io/kubernetes/hack/cherry_pick_pull.sh). This example applies a master branch PR #98765 to the remote branch `upstream/release-3.14`: `hack/cherry_pick_pull.sh upstream/release-3.14 98765`
```
spiffxp: I would move the description of this script way, way up.  AFAICT it's the sole answer for how cherrypick-candidates turn into cherrypick PR's with the current setup.  A branch manager looks at the cherry pick queue, and calls this with N PR's for "batch" mode, or PR authors can do this on their own.

contributors/devel/cherry-picks.md
```diff
@@ -7,39 +7,42 @@ depending on the point in the release cycle.
...
+ * NOTE:
```
spiffxp: I think the script also needs the `GITHUB_USER` environment variable to be set to wherever your fork of kubernetes lives

contributors/devel/cherry-picks.md
liggitt: Re: multi-branch pick tracking/process, see suggestions in https://groups.google.com/forum/m/#!topic/kubernetes-wg-contribex/XxYn02YQcV0

contributors/devel/cherry-picks.md
```diff
@@ -7,39 +7,42 @@ depending on the point in the release cycle.
 ## Propose a Cherry Pick
 
 1. Cherrypicks are [managed with labels and milestones](/contributors/guide/release-notes.md)
```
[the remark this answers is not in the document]
tpepper: Oof indeed.  I should've caught that.

contributors/devel/cherry-picks.md
```diff
@@ -7,39 +7,42 @@ depending on the point in the release cycle.
...
+There are two ways to do cherry picks:
+1. Batched: This is used only during the time after a branch is first created but before the X.Y.0 release.
+1. Individually: This is done after the first X.Y.0 on a branch.
+
+#### Batching
+ * Branch owners review the list of `cherrypick-candidate` labeled PRs.
+ * PRs batched up and merged to the release branch get a `cherrypick-approved` label and lose the `cherrypick-candidate` label.
+ * PRs that won't be merged to the release branch, lose the `cherrypick-candidate` label.
+
+There is a helper tool [`branchff`](https://git.k8s.io/release/branchff) and
```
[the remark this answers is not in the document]
tpepper: Do folks think it would make sense to merge this cherry-pick.md and the https://github.com/kubernetes/release/blob/master/docs/branching.md  ?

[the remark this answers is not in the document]
spiffxp: Maybe if you want to rescope into "the lifecycle of a release branch and how commits get into it", which also sorta overlaps with https://github.com/kubernetes/sig-release/blob/master/release-process-documentation/release-team-guides/patch-release-manager-playbook.md and https://github.com/kubernetes/sig-release#during-a-patch-release and help I can't stop why do we have so many docs that also say different versions and facets of the same thing

for now I'd rather see this PR land to fixup this one doc to be the authoritative source for this one mechanical part of the release process

contributors/devel/cherry-picks.md
spiffxp: thanks for reminding me of that thread, replied over there

spiffxp:
This may address https://github.com/kubernetes/community/issues/1894

tpepper:
@spiffxp I've chopped most of the doc now.  I'm not marking this as fixing #1894.  I feel like the full fix is moving to prow and if there are any then also changes to the "cherrypick-*" label management and syncing docs to that.

contributors/devel/cherry-picks.md
```diff
@@ -1,44 +1,32 @@
...
+   github.com/github/hub` assuming you have a standard golang development
+   environment.
+
+## Initiate a Cherry Pick
+ * Run the [cherry pick script](https://git.k8s.io/kubernetes/hack/cherry_pick_pull.sh).
+   This example applies a master branch PR #98765 to the remote branch
+   `upstream/release-3.14`: `hack/cherry_pick_pull.sh upstream/release-3.14
+   98765`
+ * Your cherrypick PR (targeted to the branch) will immediately get
+   the `do-not-merge/cherry-pick-not-approved` label. The branch owner
```
cblecker: The branch owner will approve the cherry pick, but they won't LGTM it. Standard lgtm/approval rules still apply, even for cherry picks.

tpepper: I guess I'm happy the way I updated this doc made the few pieces of original text look like new text too.  I hadn't changed this line's contents but it is nevertheless unclear.  I've tweaked the text as suggested.  Thanks!

contributors/devel/cherry-picks.md
```diff
@@ -1,44 +1,33 @@
...
+## Initiate a Cherry Pick
+ * Run the [cherry pick script](https://git.k8s.io/kubernetes/hack/cherry_pick_pull.sh).
+   This example applies a master branch PR #98765 to the remote branch
+   `upstream/release-3.14`: `hack/cherry_pick_pull.sh upstream/release-3.14
+   98765`
+ * Your cherrypick PR targeted to the release branch will immediately get the
+   `do-not-merge/cherry-pick-not-approved` label. The release branch owner
+   will triage PRs targeted to the branch.  Normal rules apply for code
+   reviewers applying the `/lgtm` label as appropriate, with the branch
+   owner applying the `/approve` label.
```
cblecker: Sorry, I should have been clearer.. the branch manager will apply "cherry-pick-approved", but OWNERS will still need to "/approve" the PR.

tpepper: So that makes sense, but...is that what's actually implemented?  I'm not seeing that in prow.

cblecker: The `cherrypick-approved` label doesn't have an associated prow command. It's added manually by the branch manager.

Example PR that shows the process: https://github.com/kubernetes/kubernetes/pull/62847

tpepper: Got it!  Amended.

contributors/devel/cherry-picks.md
```diff
@@ -1,44 +1,34 @@
...
+ * Run the [cherry pick script](https://git.k8s.io/kubernetes/hack/cherry_pick_pull.sh).
+   This example applies a master branch PR #98765 to the remote branch
+   `upstream/release-3.14`: `hack/cherry_pick_pull.sh upstream/release-3.14
+   98765`
+ * Your cherrypick PR targeted to the release branch will immediately get the
+   `do-not-merge/cherry-pick-not-approved` label. The release branch owner
+   will triage PRs targeted to the branch.  Normal rules apply for code merge.
+   Reviewers apply the `/lgtm` label and owners `/approve` as they deem
+   appropriate. The approving release branch owner is responsible for applying
+   the `/cherrypick-approved` label.
```
cblecker: I'm gonna get real nit picky here.. when you're talking about commands, the command is `/lgtm` and the label it applies is `label`. So if you're talking about the context of the command someone will use, it's `/lgtm` and `/approve`, but if you're talking about the labels they apply, it's `lgtm`, `approve`, and `cherrypick-approved`. `cherrypick-approved` should never have a slash, as there is no slash command.

cblecker: It all depends on the context and how you write it. When you say "Reviewers apply the `/lgtm` label", this is incorrect because the label is "lgtm". You can either say "Reviewers apply the `lgtm` label" or "Reviewers use the `/lgtm` command". I don't really have a preference as long as we are consistent.

tpepper: Amended to simplified language on the commands and removed the errant slash on the non-comman `cherrypick-approved` label.

PR #1981, opened 2018-03-29
Apply KEP

lavalamp:
Quick draft of this.


jbeda:
/approve no-issue

I encourage you to get this merged early as long as the scope is acceptable.  Since this has been talked about quite a bit I'm going to give this a LGTM too.  I assume that @bgrant0607 is ok with being the approver here.

/lgtm

jbeda:
The ideal process after this is to make smaller targeted changes that can be discussed and closed on more quickly vs. creating a "decide it all at once" type PR.

Nothing is "final" until the KEP moves to `implementable`.  Even then we react to reality.  The PR to change the state is a bit of a "last call" and a way for people to raise any last objections.  It puts a pin in it and moves things to a phase so that hopefully we don't re-litigate as much as we currently seem to do.

apelisse:
@jbeda Thank you for approving this PR. I'm not sure we know exactly how to move this forward? You're saying that we should make smaller target changes. What do you think should change next?

PR #1983, opened 2018-03-29
Update sig.yaml with subprojects of SIG multicluster

pmorie:
Updates the sigs.yaml file with information about SIG multicluster's subprojects.

Note: the README file for this SIG already contains manually entered information about the subprojects which we would like to maintain - for now I've left it alone but am happy to shift around as folks want.

Note: I've opened https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress/issues/159 to add an OWNERS file to that repo so that it can be tracked correctly in sigs.yaml.

### DIFF
--- a/sig-multicluster/README.md
+++ b/sig-multicluster/README.md
@@ -31,6 +31,22 @@ The Chairs of the SIG run operations and processes governing the SIG.
 * [Mailing list](https://groups.google.com/forum/#!forum/kubernetes-sig-multicluster)
 * [Open Community Issues/PRs](https://github.com/kubernetes/community/labels/sig%2Fmulticluster)
 
+## Subprojects
+
+The following subprojects are owned by sig-multicluster:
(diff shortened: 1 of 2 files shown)


spiffxp:
@pmorie there are already some other OWNERS files listed in here that don't yet exist, it'd be easier from an auditing perspective to go ahead and add the entry for multicluster ingress

Also go ahead and add an entry for cluster federation v2, so we don't need to update this file again once we make a repo for that

The description and meeting info could certainly be migrated into sigs.yaml, but I'm willing to punt on this as a evidence that our existing formatting for subprojects info needs work.

nikhiljindal:
https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress/issues/159 is now fixed, so we can add kubemci here :)

pmorie:
@spiffxp comments should be addressed now

PR #1984, opened 2018-03-29
Catalog of known scalability issues with kubernetes services

shyamjvs:
This is a markdown version of my google doc. Sharing more widely to make it common knowledge.

cc @kubernetes/sig-scalability-misc @kubernetes/sig-network-misc 
/cc @wojtek-t @countspongebob (sig-scale leads)

### DIFF
--- a/sig-scalability/blogs/k8s-services-scalability-issues.md
+++ b/sig-scalability/blogs/k8s-services-scalability-issues.md
@@ -1,0 +1,133 @@
+# Known Scalability Issues with Kubernetes Services
+
+_by Shyam JVS, Google Inc (with inputs from Brian Grant, Wojciech Tyczynski & Dan Winship)_
+
+**June 2018**
+
+This document serves as a catalog of issues we've known/discovered with kubernetes services as of June 2018, focusing on their scalability/performance. The purpose of the document is to make the information common knowledge for the community, so we can work together towards improving it. Listing them below in no particular order.
+
+
+## Iptables packet processing performance when large number of service ports exist
+
+#### Issue
+
+Iptables can be slow in packet processing when a large number of services exist. As the number of service ports increases, the KUBE-SERVICES chain gets longer and since it’s also evaluated very frequently that can bring down the performance. There have been some recent improvements in this area ([#56164](https://github.com/kubernetes/kubernetes/pull/56164), [#57461](https://github.com/kubernetes/kubernetes/pull/57461), [#60306](https://github.com/kubernetes/kubernetes/pull/60306)) - but we still need to measure and come up with safe bounds for number of services that can be 'decently' supported (10k seems to be a reasonable estimate from [past discussions](https://github.com/kubernetes/kubernetes/issues/48938#issue-243000172)). Also, further improvements might be required based on those results.
+
+#### Relevant issues/links
+
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+
+#### Possible Solution(s)
+
+- Moving from iptables to IPVS may help here if that works stably. We have the IPVS alternative [implemented](https://github.com/kubernetes/kubernetes/pull/46580), but it still hasn’t gone to GA.
+- For a long time the official kernel upstream answer for this issue was that nftables was going to solve all iptables-related scalability problems (both with packet processing and with rule changes), and there's now an out-of-tree nftables kube-proxy backend too ([#62720](https://github.com/kubernetes/kubernetes/issues/62720)).
+- There's also an alternative plan to fix the packet processing speed problems by rewriting iptables to use eBPF inside the kernel (see - https://cilium.io/blog/2018/04/17/why-is-the-kernel-community-replacing-iptables).
+
+
+## Slow/failing iptables-restore operations when large number of rules exist
+
+#### Issue
+
+When we’re running services worth a large number of backends (> 100k) on our large-cluster scalability tests, we’re noticing that kube-proxy is timing out while trying to do iptables-restore due to failing to acquire lock over iptables. There are at least two parts to this problem:
+
+- If a process tried to do a very large iptables-restore, and other processes on the system were also doing iptables-related things, the iptables-restore might get restarted multiple times (inside the kernel) before completing, which could then cause another iptables operation running in another goroutine to time out before it ever got the xtables lock.
+- The `iptables-restore` implementation is such that it grabs the lock before it parses its input, and simply parsing tens of thousands of iptables rules takes a noticeable amount of time. So if two iptables commands start at the same time, the first might grab the lock, and then start parsing its input, and burn through half of the other iptables command's `--wait` time before it even gets to the point of passing the rules off to the kernel.
+
+TODO: Find if there's any issue arising here from k8s side (relevant bug linked below).
+
+#### Relevant issues/links
+
+- [#48107](https://github.com/kubernetes/kubernetes/issues/48107)
+
+#### Possible Solution(s)
+
+For the first problem:
+
+- There's [a claim](https://github.com/kubernetes/kubernetes/issues/48107#issuecomment-398081930) that upgrading to linux kernel version 4.15 may fix this. Though we haven't verified this (current COS versions using <= 4.14 kernel version).
+
+For the second problem:
+
+- It's not easy to fix it because of the way that `iptables-restore` was written to share code with the main `iptables` binary, and no one is working on fixing it because the official plan is to move to nftables instead. If the world ends up moving to iptables-over-eBPF rather than nftables then this will probably need to be fixed at some point.
+
+In general:
+
+- The issue requires more debugging, but it may be possible that we are able to (at least partially) mitigate the problem with some tweaks here and there on kubernetes side.
+- Decide to support a smaller number of backends for kubernetes officially if it can't be fixed.
+
+
+## Apiserver spending too much cpu/memory in duplicate Endpoints serializations
+
+#### Issue
+
+Currently while serving watches, the apiserver is deep-copying deserialized endpoints objects from etcd and serializing it once for each kube-proxy watch it serves (which is 5k watches in our largest clusters).
+
+#### Relevant issues/links
+
+- [#55779](https://github.com/kubernetes/kubernetes/issues/55779#issuecomment-354452477)
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+- [#58050](https://github.com/kubernetes/kubernetes/issues/58050)
+
+#### Possible Solution(s)
+
+- Avoid those duplicate serializations if/when possible ([#60067](https://github.com/kubernetes/kubernetes/pull/60067) was an attempt to do this, but it didn't seem to help too much).
+- Moving from 'endpoints' -> 'endpoint' API change (discussed in detail in below issues) would mitigate this problem as the serialization load would decrease with smaller objects.
+
+
+## Endpoints traffic is quadratic in the number of endpoints
+
+#### Issue
+
+Endpoints object for a service contains all the individual endpoints of that service. As a result, whenever even a single pod in a service is added/updated/deleted, the whole endpoints object (which includes even the other endpoints that didn't change) is re-computed, written to storage and sent to all readers. Not being able to efficiently read/update individual endpoint changes can lead to (for e.g during rolling upgrade of a service) endpoints operations that are quadratic in the number of its elements. If you consider watches in the picture (there's one from each kube-proxy), the situation becomes even worse as the quadratic traffic gets multiplied further with number of watches (usually equal to #nodes in the cluster).
+
+Overall, this is a serious performance drawback affecting multiple components in the control-plane (apiserver, etcd, endpoints-controller, kube-proxy). The current endpoints API (which was designed at a very early stage of kubernetes when people weren't really thinking too much about scalability/performance) makes it hard to solve this problem without introducing breaking changes.
+
+#### Relevant issues/links
+
+- [#47787](https://github.com/kubernetes/kubernetes/issues/47787)
+- [#24552](https://github.com/kubernetes/kubernetes/issues/24552)
+- [#24553](https://github.com/kubernetes/kubernetes/issues/24553)
+- [#8190](https://github.com/kubernetes/kubernetes/issues/8190)
+
+#### Possible Solution(s)
+
+- Redesign endpoints API with (at least) the following properties:
+  - Lives in an API group
+  - Not coupled to service resource lifecycle
+  - Can track both pod and non-pod endpoints
+  - Separate API object for each individual endpoint
+  - Has a 'Ready' property as part of the object
+  - Port ranges and other features we’ve put off for a long time
+
+
+## Define/measure performance SLIs for endpoints propagation
+
+#### Issue
+
+We need to measure the e2e endpoints propagation latency, i.e the time from when a 
+service is created until its endpoints are populated, given that all of the associated pods 
+are already running and ready. We also need to come up with reasonable SLOs for the 
+same and verify that they’re satisfied at scale.
+
+#### Relevant issues/links
+
+- [#10436](https://github.com/kubernetes/kubernetes/issues/10436)
+- This [doc on networking SLIs](https://docs.google.com/document/d/1kudI8uILO13ySkpE5HgNHvIGn-yXVQatpyM3YTVVUis) proposes one way of formulating the SLI
+
+#### Possible Solution(s)
+
+- Measure the SLI defined in the above doc and come up with a reasonable SLO
+
+
+## Master election shouldn’t consume unnecessary watch bandwidth
+
+#### Issue
+
+Currently we store the leader lock for master components like the scheduler in the corresponding endpoints object. This has the undesirable side effect of sending a notification down the kube-proxy <-> master (and kube-dns <-> master) watch every second, which is not actually required.
+
+#### Relevant issues/links
+
+- [#34627](https://github.com/kubernetes/kubernetes/issues/34627)
+
+#### Possible Solutions(s)
+
+- Avoid overloading endpoints as a lock API. We instead want to move to using the [new Lease API](https://github.com/kubernetes/kubernetes/pull/64246) (ETA 1.12). One potential challenge here is performing upgrades of HA clusters that are using the old endpoints-based locking approach to this new API approach without breaking them.


sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,131 @@
...
+#### Possible Solution(s)
+
+- Moving from iptables to IPVS may help here if that works stably. We have the IPVS alternative [implemented](https://github.com/kubernetes/kubernetes/pull/46580), but it still hasn’t gone to GA. Also we need to measure/confirm performance benefits from it.
+
+
+## Kube-proxy slow in updating iptables when large no. of backends exist
+
+#### Issue
+
+When we’re running services worth a total of 120k endpoints (backends) on our 5k-node scalability tests, we’re noticing that kube-proxy is timing out while trying to do iptables-restore due to failing to acquire lock over iptables.
```
wojtek-t: Did you verify that it's "timing-out" not crashing or anything like that?

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,131 @@
...
+Note: Need to confirm how bad the situation is currently after some improvements that were made in 1.8.
```
wojtek-t: From when the last measuremenets are? 1.8 happened half-an-year ago...

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,131 @@
...
+
+#### Relevant issues/links
+
+- [#55779](https://github.com/kubernetes/kubernetes/issues/55779#issuecomment-354452477)
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+- [#58050](https://github.com/kubernetes/kubernetes/issues/58050)
+
+#### Possible Solution(s)
+
+- Avoid those duplicate serializations if/when possible
```
wojtek-t: You can link this PR as a try to address it (which doesn't really help).
I should probably update some thread somewhere to describe why some things are impossible to do and why this may not be that easy...

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,131 @@
...
+- Do better batching of EP updates in endpoints-controller: This is more like a bandaid, as the duplicate serializations would still happen, but we at least reduce the total no. of watch events (at the cost of increased latency of ep updates potentially). Also, this would not help too much with cases where pods of services are toggling between Ready and Failed states
```
wojtek-t: Personally I don't think it's a generic solution. It pretty much works only when you're createing/deleting a significant amount of service backends at once.
Since we probably need to solve other cases too, doing that may be unnecessary complication.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,131 @@
...
+  - Sharded into multiple etcd key/value pairs? (potentially in its own etcd)
+  - Stable sharding & ordering, with a ready property instead of separate lists
+  - Port ranges and other features we’ve put off for a long time
+
+
+## Endpoints updates quadratic in the no. of endpoints
+
+#### Issue
+
+This is the equivalent issue for GET as that of the above issue for PUT. Every time a new endpoint is added to the endpoints object, the whole endpoints object is returned by the apiserver to watch. In essence, if the no. of endpoints are N, the total no. of endpoints delivered to a watch are O(N^2) in the worst case. We want to avoid this. Also note that this behavior can lead to O(N^2) space consumption in etcd due to storing revisions of the endpoints object, each time a new endpoint is appended.
```
wojtek-t: I don't fully agree with the description here. This is quadratic even if we won't do any GETs and watches.

The problem basically is:
- we want to do rolling upgrade of a service backends
- whenever one pod changes, you update the Endpoints object and for that you update the object consisting of N endpoints
- since you do that N times, in total you write N^2 pieces of data (because each object is of size O(N))

If you add watches into the picture it's even worse, because there is watch coming from every node, so that is multiplied by the size of cluster..

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,131 @@
...
+Currently we store the leader lock for master components like the scheduler in the corresponding endpoints object. This has the undesirable side effect of sending a notification down the kube-proxy <-> master (and kube-dns <-> master) watch every second, which is not actually required.
+
+#### Relevant issues/links
+
+- [#34627](https://github.com/kubernetes/kubernetes/issues/34627)
+
+#### Possible Solutions(s)
+
+- Avoid overloading endpoints as a lock API - we can use configmaps instead, which is a supported option already in our client-go leader election library. One tricky problem here is performing upgrades of HA clusters that are using the old endpoints-based locking approach to this configmaps-based approach without breaking them
+- Filter those watches to not get such updates
```
wojtek-t: That is hacking the problem, not solving it :)

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: I've sent https://github.com/kubernetes/test-infra/pull/7545 to verify. Let's see what happens.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: https://github.com/kubernetes/kubernetes/pull/60067 (just for my record - when I'll later address the comments)

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: > Did you verify that it's "timing-out" not crashing or anything like that?

Yes. We've now verified this is the case of "timing out" - https://github.com/kubernetes/kubernetes/issues/48107#issuecomment-398018717

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Removed the note. As we now have evidence of this problem at HEAD.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Linked the PR. Does it lgty?

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Incorporated those points, PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: I've rephrased the points to make the above clearer, PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Mentioned this.

shyamjvs:
@wojtek-t I finally got time to revisit this PR. Addressed your comments - PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
+# Known Scalability Issues with Kubernetes Services
+
+_by Shyam JVS, Google Inc (with inputs from Brian Grant & Wojciech Tyczysnki)_
```
wojtek-t: nit:
s/snki/nski/

:)

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+
+When we’re running services worth a total of 120k endpoints (backends) on our 5k-node scalability tests, we’re noticing that kube-proxy is timing out while trying to do iptables-restore due to failing to acquire lock over iptables.
+
+#### Relevant issues/links
+
+- [#48107](https://github.com/kubernetes/kubernetes/issues/48107)
+
+#### Possible Solution(s)
+
+- Give more time for kube-proxy to update iptables by increasing its batching interval for large clusters
```
wojtek-t: Batching interval is unrelated here - there will never be two executions of those in-flight (if there are, that's a bug, but looking into the code it shouldn't be the case).
So that doesn't seem to be an option.

What we should do is to verify where is this timeout coming from.
E.g. can you grab the contents and try running the command yourself? I don't know where this timeout is really coming from.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+- Decide to support a smaller no. of backends for kubernetes officially
+
```
wojtek-t: Can you mention that potentiall upgrading kernel version may help: https://github.com/kubernetes/kubernetes/issues/48107#issuecomment-398081930

[To be verified, but probably worth mentioning here anyway.]

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+#### Relevant issues/links
+
+- [#55779](https://github.com/kubernetes/kubernetes/issues/55779#issuecomment-354452477)
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+- [#58050](https://github.com/kubernetes/kubernetes/issues/58050)
+
+#### Possible Solution(s)
+
+- Avoid those duplicate serializations if/when possible ([#60067](https://github.com/kubernetes/kubernetes/pull/60067) was an attempt to do this, but it didn't seem to help too much)
+- Do better batching of EP updates in endpoints-controller: This is more like a bandaid, as the duplicate serializations would still happen, but we at least reduce the total no. of watch events (at the cost of increased latency of ep updates potentially). Also, this would not help too much with cases where pods of services are toggling between Ready and Unready states, or when pods are not created in a 'bursty' manner. Since we probably need to solve such cases too, solving the problem this way may be an unnecessary investment.
```
wojtek-t: If we decide to switch from Endpoints to Endpoint object at the api level at some point, this won't help at all.

If we add that to the second part of this line, I'm not convinced we should even put it here.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+## Endpoints write traffic is quadratic in the no. of endpoints
+
+#### Issue
+
+There were some [improvements](https://github.com/kubernetes/kubernetes/issues/47597#issuecomment-311698194) around endpoints-controller, but one issue that still remains that we couldn’t make progress on was to allow useful patching for EP objects (i.e patch contains only the endpoint entry to be changed). As a result, when one pod changes, you update the endpoints object with that and send the entire object (consisting of other already-present endpoints). So the average size of a write is O(N) and there are N writes in total, leading to O(N^2) write traffic. Also note how this can increase space consumption in etcd which stores multiple revisions of the object. The current endpoints API makes it hard to solve this problem without introducing breaking changes.
```
wojtek-t: This description is too verbose in my opinion - this after "So the average size" doesn't provide much value.

At the same time, important aspects are missing here.
Even if we would be able to send patches and if etcd space wouldn't be a problem, that is not the whole problem.

Because even if there is *one* reader of endpoints objects, that reader will be quadratic, because there is no way to read just deltas on the object - you read the whole object always.

So this whole section should be rephrased to reflect that (that basically proves that the only possible solution is redesigning api - nothing else can be done).
BTW - you should also mention that there are some caveats for the potential new API (like pretty much no matching is possible then, or more metadata is being send).

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+- [#24552](https://github.com/kubernetes/kubernetes/issues/24552)
+- [#24553](https://github.com/kubernetes/kubernetes/issues/24553)
+
+#### Possible Solution(s)
+
+- Redesign endpoints API. Some properties we may like to have for the new API:
+  - Lives in an API group
+  - Not coupled to service resource lifecycle
+  - Can track both pod and non-pod endpoints
+  - Sharded into multiple etcd key/value pairs? (potentially in its own etcd)
```
wojtek-t: I'm not convinced sharding is a solution here.

wojtek-t: In general, I think if we want to split, we need to split to "individual Endpoint" without any grouping at all.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+  - Stable sharding & ordering, with a ready property instead of separate lists
```
wojtek-t: Ready property yes - about sharding and ordering - I don't really agree with that - see comment above.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+  - Port ranges and other features we’ve put off for a long time
+
+
+## Endpoints read traffic is (super-)quadratic in the no. of endpoints
+
+#### Issue
+
+This is the equivalent issue for 'reads' as the above one which was for 'writes'. An endpoints read would return all the O(N) endpoints present in the object - not allowing us to efficiently retrieve only a single endpoint we may be interested in. This becomes an even bigger problem with watches on endpoints (e.g ones from kube-proxies) as the quadratic updates get multiplied further with a factor of #watches (which is at least O(#nodes)).
```
wojtek-t: I see that read aspect is described here.

Given that the issue is basically the same (and the solution is also the same), those two points should be merged into a single one.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+Currently we store the leader lock for master components like the scheduler in the corresponding endpoints object. This has the undesirable side effect of sending a notification down the kube-proxy <-> master (and kube-dns <-> master) watch every second, which is not actually required.
+
+#### Relevant issues/links
+
+- [#34627](https://github.com/kubernetes/kubernetes/issues/34627)
+
+#### Possible Solutions(s)
+
+- Avoid overloading endpoints as a lock API - we can use configmaps instead, which is a supported option already in our client-go leader election library. One tricky problem here is performing upgrades of HA clusters that are using the old endpoints-based locking approach to this configmaps-based approach without breaking them.
+- Filter those watches to not get such updates. This is more of a hack than a proper solution, so we choose not to do it.
```
wojtek-t: We shouldn't even mention that - we need to solve it properly.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,130 @@
...
+
+Currently we store the leader lock for master components like the scheduler in the corresponding endpoints object. This has the undesirable side effect of sending a notification down the kube-proxy <-> master (and kube-dns <-> master) watch every second, which is not actually required.
+
+#### Relevant issues/links
+
+- [#34627](https://github.com/kubernetes/kubernetes/issues/34627)
+
+#### Possible Solutions(s)
+
+- Avoid overloading endpoints as a lock API - we can use configmaps instead, which is a supported option already in our client-go leader election library. One tricky problem here is performing upgrades of HA clusters that are using the old endpoints-based locking approach to this configmaps-based approach without breaking them.
```
wojtek-t: confimap is not what we want.
We want to use the new Lease API that is coming in 1.12 for that purpose (the question about performing upgrades remains - though I have idea how to solve that):
https://github.com/kubernetes/kubernetes/pull/64246

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: I'm not sure if our understanding of 'sharding' is the same. My understanding was it meant exactly what you wrote above i.e broken into individual endpoints :)

That said, the original wordings for this suggestion was from @bgrant0607 - maybe he had something in mind.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: +1. I'm also unclear to me why we'd need it. + @bgrant0607 (who originally suggested this). My guess back then was it's probably some implementation-specific backward compatibility requirement (though I can't think of any).

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Done. It can sometimes get hard with Polish names ;)

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: You're right - I've tried to push in so many different concerns (well there indeed are so many problems with endpoints :p) without properly structuring it.
Reorganized now, PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Done.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Removed it.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: I actually wanted to exactly make this change in the doc couple days ago, but then forgot about it :)
Done - PTAL.

shyamjvs:
@wojtek-t - Addressed comments, PTAL.

cc @bgrant0607 (as some inputs were from you :)

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,123 @@
...
+When we’re running services worth a total of 120k endpoints (backends) on our 5k-node scalability tests, we’re noticing that kube-proxy is timing out while trying to do iptables-restore due to failing to acquire lock over iptables.
+TODO: Find root cause of this issue (linked below).
+
+#### Relevant issues/links
+
+- [#48107](https://github.com/kubernetes/kubernetes/issues/48107)
+
+#### Possible Solution(s)
+
+- There's [a claim](https://github.com/kubernetes/kubernetes/issues/48107#issuecomment-398081930) that upgrading to linux kernel version 4.15 may fix this. Though we haven't verified this (with current COS versions using 4.14 or below).
```
wojtek-t: It's possible that just bumping some timeouts on k8s side may help with that (maybe not fully solve, but at least help).
Probably worth mentioning that this requires debugging and it's possible that we may be able to (at least partially) mitigate the problem by just some tweaks here and there.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,123 @@
...
+#### Relevant issues/links
+
+- [#55779](https://github.com/kubernetes/kubernetes/issues/55779#issuecomment-354452477)
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+- [#58050](https://github.com/kubernetes/kubernetes/issues/58050)
+
+#### Possible Solution(s)
+
+- Avoid those duplicate serializations if/when possible ([#60067](https://github.com/kubernetes/kubernetes/pull/60067) was an attempt to do this, but it didn't seem to help too much).
+- The proper solution for this would be moving from 'endpoints' -> 'endpoint' API change (discussed in detail in below issues). As with smaller objects, the serialization load would significantly decrease.
```
wojtek-t: I'm not convinced that "significantly".

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,123 @@
...
+
+
+## Endpoints traffic is quadratic in the no. of endpoints
+
+#### Issue
+
+Endpoints object for a service contains all the individual endpoints of that service. As a result, whenever even a single pod in a service is added/updated/deleted, the whole endpoints object (which includes even the other endpoints that didn't change) is re-computed, written to storage and sent to all readers. The reason for this problem is that the current API is badly designed - not allowing for efficiently updating/reading individual endpoint changes. Explanation for quadratic behaviour:
+
+- let's say we want to do rolling upgrade of a service backends
```
wojtek-t: Please remove that - you already explained that above "the whole endpoints object (which includes even the other endpoints that didn't change) is re-computed, written to storage and sent to all readers"

This explanation doesn't bring any additional value.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,123 @@
...
+
+#### Possible Solution(s)
+
+- Redesign endpoints API. Some properties we may like to have for the new API:
+  - Lives in an API group
+  - Not coupled to service resource lifecycle
+  - Can track both pod and non-pod endpoints
+  - Sharded into multiple etcd key/value pairs? (potentially in its own etcd)
+  - Has a Ready property
+  - Stable sharding & ordering
```
wojtek-t: I still don't fully follow this one. If we decide for a singular "Endpoint" object, then we don't need any sharding at all (and ordering we get just for free, because things are ordered in API itself).

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,123 @@
...
+Currently we store the leader lock for master components like the scheduler in the corresponding endpoints object. This has the undesirable side effect of sending a notification down the kube-proxy <-> master (and kube-dns <-> master) watch every second, which is not actually required.
+
+#### Relevant issues/links
+
+- [#34627](https://github.com/kubernetes/kubernetes/issues/34627)
+
+#### Possible Solutions(s)
+
+- Avoid overloading endpoints as a lock API. We instead want to move to using the [new Lease API](https://github.com/kubernetes/kubernetes/pull/64246) (ETA 1.12). One potential challenge here is performing upgrades of HA clusters that are using the old endpoints-based locking approach to this new API approach without breaking them.
+- Another option is using configmaps, which is a supported option already in our client-go leader election library. But we don't choose this in favour of the above better option.
```
wojtek-t: Let's remove this - with Lease api that is designed for that particular purpose, this doesn't make sense.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Done.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Removed.

shyamjvs: I can later do some math to show the significance.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Removed it and changed the explanation a bit - PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Yeah - I feel the same. Removed it for now - we can re-add later if there's a proper reason to.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Done.

shyamjvs:
@wojtek-t Addressed, PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -54,20 +54,16 @@ Currently while serving watches, the apiserver is deep-copying deserialized endp
...
-- with N such changes (one per each pod in the service), we'd be writing O(N^2) amount of data (because each object is of size O(N))
-
-Now if you consider watches in the picture (there's one from each kube-proxy), the situation becomes even worse as the O(N^2) size traffic gets multiplied further with no. of watches (usually equal to #nodes in the cluster). Overall, this is a serious performance drawback affecting multiple components in the control-plane (apiserver, etcd, endpoints-controller, kube-proxy). The current endpoints API makes it hard to solve this problem without introducing breaking changes.
+If you consider watches in the picture (there's one from each kube-proxy), the situation becomes even worse as the quadratic traffic gets multiplied further with no. of watches (usually equal to #nodes in the cluster). Overall, this is a serious performance drawback affecting multiple components in the control-plane (apiserver, etcd, endpoints-controller, kube-proxy). The current endpoints API makes it hard to solve this problem without introducing breaking changes.
```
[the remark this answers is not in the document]
shyamjvs: Changed, here and elsewhere.

shyamjvs:
@wojtek-t Addressed comments and squashed - PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,117 @@
+# Known Scalability Issues with Kubernetes Services
+
+_by Shyam JVS, Google Inc (with inputs from Brian Grant & Wojciech Tyczynski)_
+
+**June 2018**
+
+This document serves as a catalog of issues we've known/discovered with kubernetes services as of today, focusing on their scalability/performance. The purpose of the document is to make the information common knowledge for the community, so we can work together towards improving it. Listing them below in no particular order.
```
wojtek-t: s/today/June 2018/

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,117 @@
...
+#### Possible Solution(s)
+
+- Moving from iptables to IPVS may help here if that works stably. We have the IPVS alternative [implemented](https://github.com/kubernetes/kubernetes/pull/46580), but it still hasn’t gone to GA. Also we need to measure/confirm performance benefits from it.
+
+
+## Kube-proxy slow in updating iptables when large number of backends exist
+
+#### Issue
+
+When we’re running services worth a total of 120k endpoints (backends) on our 5k-node scalability tests, we’re noticing that kube-proxy is timing out while trying to do iptables-restore due to failing to acquire lock over iptables.
```
wojtek-t: We're running 150k endpoints now.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,117 @@
...
+TODO: Find root cause of this issue (linked below).
+
+#### Relevant issues/links
+
+- [#48107](https://github.com/kubernetes/kubernetes/issues/48107)
+
+#### Possible Solution(s)
+
+- There's [a claim](https://github.com/kubernetes/kubernetes/issues/48107#issuecomment-398081930) that upgrading to linux kernel version 4.15 may fix this. Though we haven't verified this (current COS versions using <= 4.14 kernel version). The issue requires more debugging, but it's possible that we may be able to (at least partially) mitigate the problem with some tweaks here and there.
+- Decide to support a smaller number of backends for kubernetes officially if we isn't fixed.
```
wojtek-t: s/we/it/

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,117 @@
...
+#### Relevant issues/links
+
+- [#55779](https://github.com/kubernetes/kubernetes/issues/55779#issuecomment-354452477)
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+- [#58050](https://github.com/kubernetes/kubernetes/issues/58050)
+
+#### Possible Solution(s)
+
+- Avoid those duplicate serializations if/when possible ([#60067](https://github.com/kubernetes/kubernetes/pull/60067) was an attempt to do this, but it didn't seem to help too much).
+- The proper solution for this would be moving from 'endpoints' -> 'endpoint' API change (discussed in detail in below issues). As with smaller objects, the serialization load would decrease.
```
wojtek-t: But that isn't solution for the problem, because it still scale with the number of watchers.
This is just a mitigation - please make it clear.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+## Endpoints traffic is quadratic in the number of endpoints
+
+#### Issue
+
+Endpoints object for a service contains all the individual endpoints of that service. As a result, whenever even a single pod in a service is added/updated/deleted, the whole endpoints object (which includes even the other endpoints that didn't change) is re-computed, written to storage and sent to all readers. Not being able to efficiently read/update individual endpoint changes makes endpoints operations quadratic in the number of its elements. If you consider watches in the picture (there's one from each kube-proxy), the situation becomes even worse as the quadratic traffic gets multiplied further with number of watches (usually equal to #nodes in the cluster).
```
wojtek-t: > Not being able to efficiently read/update individual endpoint changes makes endpoints operations quadratic in the number of its elements.

That's not exactly true - a single operation is linear to number of endpoints.
It's thin like "rolling upgrade of a service" that is quadratic.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Done.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Yes, but if we're failing at a smaller no. we should mention that IMO.

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Done

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Done

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: Clarified, PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,117 @@
...
+**June 2018**
+
+This document serves as a catalog of issues we've known/discovered with kubernetes services as of June 2018, focusing on their scalability/performance. The purpose of the document is to make the information common knowledge for the community, so we can work together towards improving it. Listing them below in no particular order.
+
+
+## Performance of iptables when large number of service ports exist
+
+#### Issue
+
+Iptables can be slow in the presence of a large number of services. As the number of service ports increases, the KUBE-SERVICES chain gets longer and since it’s also evaluated very frequently that can bring down the performance. Also, it seems that manipulating a chain with lots of rules (e.g. KUBE-SERVICES) takes noticeably longer than others.
```
danwinship: The second sentence seems to be talking about packets moving through the kernel being slowed down by the presence of too many iptables rules, but the third then switches to talking about the speed of making changes to the iptables rules, which is a different problem and probably belongs in the next section?

If this is supposed to be about the speed of packet processing, rather than the speed of "iptables-restore", then it might be good to clarify that in the section heading, since just "Performance of iptables" is ambiguous as to which part of the system it's referring to.

(FWIW https://github.com/kubernetes/kubernetes/pull/56164 / https://github.com/kubernetes/kubernetes/pull/57461 / https://github.com/kubernetes/kubernetes/pull/60306 are some examples of fixes specifically for network performance as opposed to reload speed. I'm not sure if there are any currently-open issues about iptables-related network performance problems.)

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### Relevant issues/links
+
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+
+#### Possible Solution(s)
+
+- Moving from iptables to IPVS may help here if that works stably. We have the IPVS alternative [implemented](https://github.com/kubernetes/kubernetes/pull/46580), but it still hasn’t gone to GA. Also we need to measure/confirm performance benefits from it.
```
danwinship: For a long time the official kernel upstream answer was that nftables was going to solve all iptables-related scalability problems (both with packet processing and with rule changes), and there's now an out-of-tree nftables kube-proxy backend too. (https://github.com/kubernetes/kubernetes/issues/62720).

There's also an alternative plan to fix the packet processing speed problems by rewriting iptables to use eBPF inside the kernel (https://cilium.io/blog/2018/04/17/why-is-the-kernel-community-replacing-iptables/). (That doesn't necessarily address the reload-speed problem though, and there's a limit to how much that can be fixed without making incompatible changes to the userspace iptables "API".)

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: We don't currently test at that limit, but we have evidence from past that we see this issue at that number too (https://github.com/kubernetes/kubernetes/issues/48107). It may have changed now (but I'm not sure if some fixes were made for this bug in the meanwhile). So I feel we should have the lower number here to be rigorous.

danwinship: FWIW in terms of "timing out while trying to acquire lock", there's a second problem: `iptables-restore` grabs the lock *before* it parses its input, and simply parsing tens of thousands of iptables rules takes a noticeable amount of time. So if two iptables commands start at the same time, the first might grab the lock, and then start parsing its input, and burn through half of the other iptables command's `--wait` time before it even gets to the point of passing the rules off to the kernel.

This is not easy to fix because of the way that `iptables-restore` was written to share code with the main `iptables` binary, and no one is working on fixing it because the official plan is to move to nftables instead. If the world ends up moving to iptables-over-eBPF rather than nftables then this will probably need to be fixed at some point.

wojtek-t: I'm fine with having more rigorous number here. What I'm asking about is - why 120k?

If I would be reading it and would say 120k, I guess I would silently assume that 100k is fine (otherwise you would probably give more round number). Do we know that? (I think we don't).

So maybe it's safer to write 100k (if we are going this way?)

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: You're right. Let me make the distinction clear.

shyamjvs: I've incorporated the above - PTAL. Also thanks for those really useful leads!

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: I've incorporated those points - PTAL. Also thanks for those really useful leads!

sig-scalability/blogs/k8s-services-scalability-issues.md
shyamjvs: @danwinship I've incorporated those very useful leads. Thanks and PTAL.

> What I'm asking about is - why 120k?

@wojtek-t Because it's a known limit for which we have evidence that we seen the issue. Anyway, I've now changed it to `> 100k` - lgty?

shyamjvs:
@wojtek-t @danwinship Addressed above comments. I've also modified some issue descriptions to be more accurate in light of the extra info received - PTAL.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Possible Solution(s)
+
+For the first problem:
+
+- There's [a claim](https://github.com/kubernetes/kubernetes/issues/48107#issuecomment-398081930) that upgrading to linux kernel version 4.15 may fix this. Though we haven't verified this (current COS versions using <= 4.14 kernel version).
+
+For the second problem:
+
+- It's not easy to fix it because of the way that `iptables-restore` was written to share code with the main `iptables` binary, and no one is working on fixing it because the official plan is to move to nftables instead. If the world ends up moving to iptables-over-eBPF rather than nftables then this will probably need to be fixed at some point.
```
wojtek-t: In general, if our only problem are timeouts, maybe it would be enough to bump them?

I guess I'm saying that speeding this up may be difficult, but getting rid of errors sounds like something we may be able to do.

wojtek-t:
This looks good to me now. But I will @danwinship to also take another look

danwinship:
yeah, lgtm

shyamjvs:
Thanks for the reviews!

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,133 @@
...
+#### Relevant issues/links
+
+- [#55779](https://github.com/kubernetes/kubernetes/issues/55779#issuecomment-354452477)
+- [#48938](https://github.com/kubernetes/kubernetes/issues/48938)
+- [#58050](https://github.com/kubernetes/kubernetes/issues/58050)
+
+#### Possible Solution(s)
+
+- Avoid those duplicate serializations if/when possible ([#60067](https://github.com/kubernetes/kubernetes/pull/60067) was an attempt to do this, but it didn't seem to help too much).
+- Moving from 'endpoints' -> 'endpoint' API change (discussed in detail in below issues) would mitigate this problem as the serialization load would decrease with smaller objects.
```
smarterclayton: In general as we have more N scale watchers we probably need to consider optimizing the general watch stack.  I don’t think it’s enough to fix endpoints by itself.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,133 @@
...
+- [#24553](https://github.com/kubernetes/kubernetes/issues/24553)
+- [#8190](https://github.com/kubernetes/kubernetes/issues/8190)
+
+#### Possible Solution(s)
+
+- Redesign endpoints API with (at least) the following properties:
+  - Lives in an API group
+  - Not coupled to service resource lifecycle
+  - Can track both pod and non-pod endpoints
+  - Separate API object for each individual endpoint
```
smarterclayton: This is going to increase overhead a lot, so we may need spans or tranches.  Daemonset services are definitely the most likely to blow up first.

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,133 @@
...
+  - Port ranges and other features we’ve put off for a long time
+
+
+## Define/measure performance SLIs for endpoints propagation
+
+#### Issue
+
+We need to measure the e2e endpoints propagation latency, i.e the time from when a 
+service is created until its endpoints are populated, given that all of the associated pods 
+are already running and ready. We also need to come up with reasonable SLOs for the
```
smarterclayton: I’d like to do this for pod create -> endpoint distribution latency as that is a gating factor in unidle and serverless style use of Kube

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,133 @@
...
+#### Possible Solution(s)
+
+- Measure the SLI defined in the above doc and come up with a reasonable SLO
+
+
+## Master election shouldn’t consume unnecessary watch bandwidth
+
+#### Issue
+
+Currently we store the leader lock for master components like the scheduler in the corresponding endpoints object. This has the undesirable side effect of sending a notification down the kube-proxy <-> master (and kube-dns <-> master) watch every second, which is not actually required.
```
smarterclayton: Why haven’t we moved to configmaps?

sig-scalability/blogs/k8s-services-scalability-issues.md
wojtek-t: I agree with the optimizing watch stack - couple month ago I tried doing something - you can see: https://github.com/kubernetes/kubernetes/pull/60067, but that didn't help almost at all. And I've never had time to dig deeper why this doesn't help at all.

sig-scalability/blogs/k8s-services-scalability-issues.md
wojtek-t: > This is going to increase overhead a lot, so we may need spans or tranches. 

I think that's not clear to me. I'm not saying "this won't be needed for sure", but I think I'm not convinced it will for sure be needed.

> Daemonset services are definitely the most likely to blow up first.

I don't understand that. Can you clarify? What daemonset services? Kube-proxy e.g.? Why?

sig-scalability/blogs/k8s-services-scalability-issues.md
wojtek-t: I'm not a fan for pod created -> endpoint distribution. That contains e.g. "pod becoming ready" phase, which is completely application specific and not k8s dependent.
We already have this pod created -> pod running (not necessary ready) SLO (for ~3 years already: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos).
What is missing is the bit of pod ready -> endpoint distributed (which I would like to introduce).

Then what you're interested in is basically: (1) + (2) + some application specific time.

Anyway - I'm not opposed of measuring what you suggested too, but that isn't something you will be able to provide SLO for due to application-specific bits being part of it.

sig-scalability/blogs/k8s-services-scalability-issues.md
wojtek-t: I guess noone really picked this up (we would need to figure our migration in HA setups, which is doable but requires some work).

sig-scalability/blogs/k8s-services-scalability-issues.md
nevola: Note that keep using iptables interface it will inherit all the interface problems that iptables has. The evolution to nftables is not a whim, it was designed to fulfill all the lacks that iptables interface has but also much more features that permits to implement more use cases with better performance.

sig-scalability/blogs/k8s-services-scalability-issues.md
smarterclayton: Having one API object per endpoint is going to be more fixed overhead (serialization, key fetches, overhead of wrapping the object in memory) than having ranges of endpoints.

For reference, the metadata field itself is 200 bytes, even if every field is nil.  Maps can be up to 300 bytes in go, even empty, and metadata has 2 of them.

Having O(N) endpoints to fetch instead of O(1) endpoints with a small constant factor to fetch is going to be a tradeoff.

sig-scalability/blogs/k8s-services-scalability-issues.md
smarterclayton: The application specific time is not what I would measure (or keep it small to emulate a service that starts in 100ms), but since Kube's job is to run services that show up, I absolutely think 1-2 is the actual practical SLO for most users.

sig-scalability/blogs/k8s-services-scalability-issues.md
smarterclayton: Most of the code should have switched.  Scheduler and CM should both be on config map.  Some of the ecosystem listeners have moved.  OpenShift moved in 3.9 / 1.9 to force all users to use config map.

sig-scalability/blogs/k8s-services-scalability-issues.md
wojtek-t: > Having O(N) endpoints to fetch instead of O(1) endpoints with a small constant factor to fetch is going to be a tradeoff.

I agree with that. I was just trying to say that it's not obvious to me that what results of the experiment will be.

sig-scalability/blogs/k8s-services-scalability-issues.md
wojtek-t: So let me clarify a bit what I was trying to say.
For me, the official SLI is something that we can measure in running clusters (and is meaningful there). That pod create -> endpoint distribution latency is not really such thing. That's why I'm reluctant of making this an official SLI.

That said, I'm totally fine with measuring that in our tests and detecting regressions on that. And that's definitely a good point. I just don't think we can make an official SLI/SLO on top of that.

wojtek-t: So this will be addition to our tests and/or test framework, not really to the SLI.

sig-scalability/blogs/k8s-services-scalability-issues.md
wojtek-t: I agree that they should have switched. But given that they didn't, I think we shouldn't do it now and instead wait for Lease object (the API is already merged in fact) and switch things directly to that (and deprecate both endpoint and configmap for that purpose).

sig-scalability/blogs/k8s-services-scalability-issues.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+## Iptables packet processing performance when large number of service ports exist
+
+#### Issue
+
+Iptables can be slow in packet processing when a large number of services exist. As the number of service ports increases, the KUBE-SERVICES chain gets longer and since it’s also evaluated very frequently that can bring down the performance. There have been some recent improvements in this area ([#56164](https://github.com/kubernetes/kubernetes/pull/56164), [#57461](https://github.com/kubernetes/kubernetes/pull/57461), [#60306](https://github.com/kubernetes/kubernetes/pull/60306)) - but we still need to measure and come up with safe bounds for number of services that can be 'decently' supported (10k seems to be a reasonable estimate from [past discussions](https://github.com/kubernetes/kubernetes/issues/48938#issue-243000172)). Also, further improvements might be required based on those results.
```
thockin: We could turn the linear search into a binary search pretty easily, we have never had a driving demand for it.  IPVS and other modes seem to be better in almost every way, so we probably won't.

sig-scalability/blogs/k8s-services-scalability-issues.md
thockin: And there are other optimizations we could make - e.g. not using map, if that is really a problem.

shyamjvs: > Having O(N) endpoints to fetch instead of O(1) endpoints with a small constant factor to fetch is going to be a tradeoff.

While I agree the small constant factor will affect fetching 'N' O(1) endpoints, we also need to consider how many times do we actually perform the operation of "get all endpoints" - as we seem to be losing only in such cases. With watches in use (which is the case at all key places FMU), we actually should be much better off getting only O(1) update for a single endpoint change instead of O(N) update. There's still the initial list call which suffers (but that should be very infrequent). Wdyt?

thockin:
Should we run an experiment to see how reality maps to intuitions?  E.g.
100,000 1k objects, 10,000 10K objects, 1000 100K objects, 200 500K
objects.  Churn elements of those at some known rate and measure ...
something?

shyamjvs:
Sure - I definitely agree. We were just talking with @wojtek-t today about some experiments we'd want to do in this area.

wojtek-t:
> Should we run an experiment to see how reality maps to intuitions?  E.g.
100,000 1k objects, 10,000 10K objects, 1000 100K objects, 200 500K
objects.  Churn elements of those at some known rate and measure ...
something?

Yeah - that was exactly what we were talking about earlier today (well, not these exact numbers, but the high level idea).

PR #1985, opened 2018-03-29
Add more attributes into FlexVolume options field.

incfly:
Based on the source code, https://github.com/kubernetes/kubernetes/blob/4685df26ddc2c0778564ed77ec974da56bd52994/pkg/volume/flexvolume/driver-call.go#L48


incfly:
Do you want this to be merged or I should close this? @idealhack @cblecker

idealhack:
@incfly Thanks for your contribution and please keep this open for now. We just need some reviewers and approvers here.

/assign @spiffxp @idvoretskyi @calebamiles

contributors/devel/flexvolume.md
```diff
@@ -122,6 +122,12 @@ Note: Secrets are passed only to "mount/unmount" call-outs.
 ```
 "kubernetes.io/fsType":"<FS type>",
 "kubernetes.io/readwrite":"<rw>",
+
+"kubernetes.io/pod.name": <string>,
+"kubernetes.io/pod.namespace": <string>,
+"kubernetes.io/serviceAccount.name": <string>,
+"kubernetes.io/pod.uid": <string>,
+
```
idealhack: Seems a good PR, but why these attributes are not the same as the source you referenced?

spiffxp: This is missing a few options from
https://github.com/kubernetes/kubernetes/blob/4685df26ddc2c0778564ed77ec974da56bd52994/pkg/volume/flexvolume/driver-call.go#L48-L59

The arrangement/ordering is a bit weird.  I would either match what's in the source, or alphabetize

spiffxp:
ping @incfly

spiffxp:
/lifecycle stale
/unassign
ping @incfly

contributors/devel/flexvolume.md
incfly: Done.

However, because of unfamilarity of the options, I just leave added options as `<string>`, which might be fine, because they're pretty self-evident.

incfly:
ping @spiffxp

idealhack:
/sig storage

PR #1986, opened 2018-03-29
Include more instructions on WG creation.

kimsterv:

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -109,15 +109,15 @@ This is the purpose of Working Groups (WG). The intent is to make
 Working Groups relatively easy to create and to deprecate, once
 inactive.
 
-If you would like to propose a new working group, send a proposal to
-kubernetes-dev@googlegroups.com and also include any potentially 
-interested SIGs. Wait for public comment. If there's enough interest,
-a new Working Group should be formed. 
+To propose a new working group, first find a SIG to sponsor the group. 
+Next, send a proposal to kubernetes-dev@googlegroups.com and also include
+any potentially interested SIGs. Wait for public comment. If there's 
+enough interest, a new Working Group should be formed. 
 
-Create a new mailing list in the from of kubernetes-wg-group-name.
-Working groups typically have a Slack channel as well as regular meetings. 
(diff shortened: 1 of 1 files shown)


kimsterv:
CLA signed.

cblecker:
https://github.com/kubernetes/steering/issues/27

cc: @kubernetes/steering-committee

governance.md
```diff
@@ -109,6 +109,16 @@ This is the purpose of Working Groups (WG). The intent is to make
...
+If you would like to propose a new working group, send a proposal to
+kubernetes-dev@googlegroups.com and also include any potentially 
+interested SIGs. Wait for public comment. If there's enough interest,
+a new Working Group should be formed. 
+
+Create a new mailing list in the from of kubernetes-wg-group-name.
+Working groups typically have a Slack channel as well as regular meetings. 
+It's encouraged to keep a clear record of all accomplishments that's 
+publicly accessible.
+
```
kow3ns: Typically we have required SIGs to sponsor working groups. Do we require a SIG to sponsor a working group, or can a WG be formed independently of any SIG?

parispittman: I asked Kim to update with the SIG sponsor info (as well as some other things). Looking back on several previous WG creations, there was at least one sponsoring sig pre-proposal.

michelleN:
Looks like #1994 is also adding some detail on WG. 
I propose we accept this PR after Kim gets a chance to add the bit on sponsor SIGs and then review #1994 since it's a bit more involved. Is that okay with you @jbeda ?

jbeda:
Note that I'm also touching this stuff in https://github.com/kubernetes/community/pull/1994 but I don't think our PRs overlap.

jbeda:
It looks like they are touching the same files but don't significantly overlap.  My PR doesn't go into how WGs are formed but more about what they are compared to the other groups.  I think there there is some open questions around how WGs are formed and added a comment above.

idvoretskyi:
Anything that blocks us from merging it now?

PR #1989, opened 2018-03-29
Add a WIP charter for SIG multicluster

pmorie:
Initial stab at a WIP charter for SIG multicluster (mostly cribbed from README)

### DIFF
--- a/sig-multicluster/charter.md
+++ b/sig-multicluster/charter.md
@@ -1,16 +1,140 @@
 # SIG Multicluster Charter
 
-This charter is a WIP.
+This charter adheres to the conventions described in the [Kubernetes Charter README].
 
-SIG Multicluster focuses on solving common challenges related to:
+## Scope
 
-- Managing/administering multiple Kubernetes clusters
-- Deploying and managing applications into multiple Kubernetes clusters.
+This section defines the scope of things that would fall under ownership by this SIG.
+It must be used when determining whether subprojects should fall into this SIG.
 
-The SIG will be responsible for designing, discussing, implementing and
-maintaining API’s, tools and documentation related to multicluster
-administration and application management. This includes both active automated
-approaches as well as batch workflow-style continuous deployment systems.
-Standalone building blocks for these and other similar systems (for example a
-cluster registry), and proposed changes to kubernetes core where appropriate
-will also be in scope.
+### In scope
+
+- Solving common challenges related to management of multiple Kubernetes clusters and the applications that exist therein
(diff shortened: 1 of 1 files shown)


sig-multicluster/charter.md
```diff
@@ -0,0 +1,16 @@
+# SIG Multicluster Charter
+
+This charter is a WIP.
+
+SIG Multicluster focuses on solving common challenges related to:
+
+- Managing/administering multiple Kubernetes clusters
```
irfanurrehman: Probably rewording needed.
`Managing/administering centralised information about multiple kubernetes clusters` ?? or something else.
The prior (`cluster administration/management` as the term) seems to suggest stuff `cluster-lifecycle` folks might be involved with.
I understand that the same wording is currently used on the sig-multicluster page; IMHO needs correction.

pmorie: I think there's probably an overlap here. As I understand it, cluster-lifecycle is about creating new clusters. As I understand it, the wording currently in the README pertains to administering multiple clusters from the same source of truth. I'm fine striking it for now if people would like, but I expect this to be in scope for SIG multicluster.

irfanurrehman: Seems reasonable, lets go ahead with what this PR has.

timothysc:
@pmorie This needs a bit of work to follow the updated sig charter.  

https://github.com/kubernetes/community/tree/master/committee-steering/governance

Also requires scope boundaries.

michelleN:
@pmorie - Thanks for your time on this! We added a [charter template](https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-charter-template.md) in case that helps you a bit more. Feel free to reach out if you need clarification on anything. 

We're merging all charters with non-blocking items on Wednesday in case you want to close the loop on this but no pressure.

Cheers

philips:
Thank you for your patience as we work to your SIG charter merged.

Through the process of reviewing the [first 11 charters][charters] the Steering Committee (SC) found a lot of copy/paste language caused by our poorly designed initial charter template. To fix this there is a new [SIG charter template][template] which focuses on the main things we want to see in charters: scope and responsibilities.

Please consider updating to this charter template and putting focus into the scope and responsibilities. SC members will focus more on a deep review of scope more than anything else and using this template will help with that focus.

[charters]: https://docs.google.com/spreadsheets/d/1i_8oM1KlmXo90zIIgNTdOg-M6xesTggFLnTkgmscMlI/edit#gid=0
[template]: https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-charter-template.md

PR #1991, opened 2018-03-30
add KEP for pod ready++

freehan:
KEP for pod ready++

### DIFF
--- a/keps/sig-network/0007-pod-ready++.md
+++ b/keps/sig-network/0007-pod-ready++.md
@@ -1,0 +1,173 @@
+---
+kep-number: 1
+title: Pod Ready++
+authors:
+  - "freehan@"
+owning-sig: sig-network
+participating-sigs:
+  - sig-node
+  - sig-cli
+reviewers:
+  - thockin@
+  - dchen1107@
+approvers:
+  - thockin@
+  - dchen1107@
+editor: freehan@
+creation-date: 2018-04-01
+last-updated: 2018-04-01
+status: provisional
+
+---
+
+# Pod Ready++
+
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints-optional)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Alternatives](#alternatives-optional)
+
+
+## Summary
+
+This proposal aims to add extensibility to pod readiness. Besides container readiness, external feedback can be injected into PodStatus and influence pod readiness. Thus, achieving pod “ready++”. 
+
+## Motivation
+
+Pod readiness indicates whether the pod is ready to serve traffic. Pod readiness is dictated by kubelet with user specified readiness probe. On the other hand, pod readiness determines whether pod address shows up on the address list on related endpoints object. K8s primitives that manage pods, such as Deployment, only takes pod status into account for decision making, such as advancement during rolling update. 
+
+For example, during deployment rolling update, a new pod becomes ready. On the other hand, service, network policy and load-balancer are not yet ready for the new pod due to whatever reason (e.g. slowness in api machinery, endpoints controller, kube-proxy, iptables or infrastructure programming). This may cause service disruption or lost of backend capacity. In extreme cases, if rolling update completes before any new replacement pod actually start serving traffic, this will cause service outage. 
+
+
+### Goals
+
+- Allow extra signals for pod readiness.
+
+### Non-Goals
+
+- Provide generic framework to solve all transition problems in k8s (e.g. blue green deployment).
+
+## Proposal
+
+[K8s Proposal: Pod Ready++](https://docs.google.com/document/d/1VFZbc_IqPf_Msd-jul7LKTmGjvQ5qRldYOFV0lGqxf8/edit#)
+
+### PodSpec
+Introduce an extra field called ReadinessGates in PodSpec. The field stores a list of ReadinessGate structure as follows: 
+```yaml
+type ReadinessGate struct {
+	conditionType string	
+}
+```
+The ReadinessGate struct has only one string field called ConditionType. ConditionType refers to a condition in the PodCondition list in PodStatus. And the status of conditions specified in the ReadinessGates will be evaluated for pod readiness. If the condition does not exist in the PodCondition list, its status will be default to false. 
+
+#### Constraints:
+- ReadinessGates can only be specified at pod creation. 
+- No Update allowed on ReadinessGates.
+- ConditionType must conform to the naming convention of custom pod condition.
+
+### Pod Readiness
+Change the pod readiness definition to as follows:
+```
+Pod is ready == containers are ready AND conditions in ReadinessGates are True
+```
+Kubelet will evaluate conditions specified in ReadinessGates and update the pod “Ready” status. For example, in the following pod spec, two readinessGates are specified. The status of “www.example.com/feature-1” is false, hence the pod is not ready. 
+
+```yaml
+Kind: Pod 
+… 
+spec: 
+  readinessGates:
+  - conditionType: www.example.com/feature-1
+  - conditionType: www.example.com/feature-2
+… 
+status:
+  conditions:
+  - lastProbeTime: null
+    lastTransitionTime: 2018-01-01T00:00:00Z
+    status: "False"
+    type: Ready
+  - lastProbeTime: null
+    lastTransitionTime: 2018-01-01T00:00:00Z
+    status: "False"
+    type: www.example.com/feature-1
+  - lastProbeTime: null
+    lastTransitionTime: 2018-01-01T00:00:00Z
+    status: "True"
+    type: www.example.com/feature-2
+  containerStatuses:
+  - containerID: docker://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+    ready : true
+… 
+```
+
+Another pod condition `ContainerReady` will be introduced to capture the old pod `Ready` condition. 
+```
+ContainerReady is true == containers are ready
+```
+
+### Custom Pod Condition
+Custom pod condition can be injected thru PATCH action using KubeClient. Please be noted that “kubectl patch” does not support patching object status. Need to use client-go or other KubeClient implementations. 
+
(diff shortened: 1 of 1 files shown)


jbeda:
I encourage you to get this PR submitted once @thockin and/or @dchen1107 approve of the scope.  Err on the side of smaller targeted PRs that take on a single topic. Once things have settled create a PR to move it from "provisional" to "implementable".  That provides a "last call" to make sure that everyone is on the same page.  The goal is to avoid 100+ comment PRs that tackle all topics at once.

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Proposal
+
+[K8s Proposal: Pod Ready++](https://docs.google.com/document/d/1VFZbc_IqPf_Msd-jul7LKTmGjvQ5qRldYOFV0lGqxf8/edit#)
+
+
+
+### Implementation Details/Notes/Constraints
+
+This proposal mostly involves kubelet changes:
```
thockin: I would detail that the `ready` condition must continue to be the FINAL signal for controllers to proceed (for compat) which complicates the design overall.

MAYBE also add a short blurb about all-containers-ready (e.g. for endpoints) vs pod-ready?  Maybe that belongs in the doc (but sharing docs with community sucks)

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,96 @@
...
+- Allow extra signals for pod readiness.
+
+### Non-Goals
+
+- Provide generic framework to solve transition problems in k8s (e.g. blue green deployment).
+
+
+## Proposal
+
+[K8s Proposal: Pod Ready++](https://docs.google.com/document/d/1VFZbc_IqPf_Msd-jul7LKTmGjvQ5qRldYOFV0lGqxf8/edit#)
```
thockin: Is it worth converting that to markdown to be pasted inline here?  What is the precedent @calebamiles  ?

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,96 @@
...
+## Motivation
+
+Pod readiness indicates whether the pod is ready to serve traffic. Pod readiness is dictated by kubelet with user specified readiness probe. On the other hand, pod readiness determines whether pod address shows up on the address list on related endpoints object. K8s primitives that manage pods, such as Deployment, only takes pod status into account for decision making, such as advancement during rolling update. 
+
+For example, during deployment rolling update, a new pod becomes ready. On the other hand, service, network policy and load-balancer are not yet ready for the new pod due to whatever reason (e.g. slowness in api machinery, endpoints controller, kube-proxy, iptables or infrastructure programming). This may cause service disruption or lost of backend capacity. In extreme cases, if rolling update completes before any new replacement pod actually start serving traffic, this will cause service outage. 
+
+
+### Goals
+
+- Allow extra signals for pod readiness.
```
caseydavenport: Perhaps there are a set of signals that we should implement as part of completing this KEP and as part of proving the API?

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,96 @@
...
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+
+
+## Summary
+
+This proposal aims to add extensibility to pod readiness. Besides container readiness, external feedback can be injected into PodStatus and influence pod readiness. Thus, achieving pod “ready++”.
```
MikeSpreitzer: I think a better summary is to recognize that what will really happen is to split the concept of "pod ready" into two pieces, for two different sets of consumers.  The current text focuses one one set of consumers, the "workload controllers" that want "pod ready ++".  The other set of consumers is the load balancers, network policy implementors, and so on that contribute to the "pod ready ++" based on "plain old pod ready".  Remember, a load balancer (for example) has to react to a pod before all the load balancers and whatever have reacted to the pod.  Following is a suggested rewording.

This proposal splits the concept of pod readiness into two concepts, for two distinct sets of consumers.  One set of consumers is the workload controllers and other clients that simply want to know when a pod is fully ready for use.  They care about a concept of readiness that we may call "ready++" and is stricter than the concept that is supported today; ready++ also indicates that all the relevant load balancers, traffic filters, and whatever have been adjusted to account for the pod.  The other set of consumers is those very load balancers, traffic filters, and whatever.  They need to continue to react to today's looser concept of readiness, while contributing to the stricter concept.  For example, a load balancer needs to put a new pod into its backend set without waiting for ready++, while also contributing its own bit into ready++.

MikeSpreitzer:
This propsoal should explicitly explain how load balancers, NetworkPolicy implementations, and so on will continue to sense "plain old pod ready".

keps/sig-network/0007-pod-ready++.md
dcbw: +1 for markdown here in git

keps/sig-network/0007-pod-ready++.md
freehan: I added a user adoption in the graduation criteria. I would not tied this proposal with specific feature for now.

freehan:
Ported the proposal from google doc into the PR. PTAL

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Implementation Details/Notes/Constraints
+
+##### Workloads
+To conform with this proposals, workload controllers MUST take pod “Ready” condition as the final signal to proceed during transitions.  
+
+For the workloads that take pod readiness as a critical signal for its decision making, they will automatically comply with this proposal without any change. Majority, if not all, of the workloads satisfy this condition. 
+
+##### Kubelet
+- Use PATCH instead of PUT to update PodStatus fields that are dictated by kubelet.
```
krmayankk: What is the advantage of using Patch instead of Put here ?

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,173 @@
...
+status:
+  conditions:
+  - lastProbeTime: null
+    lastTransitionTime: 2018-01-01T00:00:00Z
+    status: "False"
+    type: Ready
+  - lastProbeTime: null
+    lastTransitionTime: 2018-01-01T00:00:00Z
+    status: "False"
+    type: www.example.com/feature-1
```
krmayankk: is this documented somewhere on how to create custom Conditions ?

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Summary
+
+This proposal aims to add extensibility to pod readiness. Besides container readiness, external feedback can be injected into PodStatus and influence pod readiness. Thus, achieving pod “ready++”. 
+
+## Motivation
+
+Pod readiness indicates whether the pod is ready to serve traffic. Pod readiness is dictated by kubelet with user specified readiness probe. On the other hand, pod readiness determines whether pod address shows up on the address list on related endpoints object. K8s primitives that manage pods, such as Deployment, only takes pod status into account for decision making, such as advancement during rolling update. 
+
+For example, during deployment rolling update, a new pod becomes ready. On the other hand, service, network policy and load-balancer are not yet ready for the new pod due to whatever reason (e.g. slowness in api machinery, endpoints controller, kube-proxy, iptables or infrastructure programming). This may cause service disruption or lost of backend capacity. In extreme cases, if rolling update completes before any new replacement pod actually start serving traffic, this will cause service outage.
```
krmayankk: how does this cause loss of backend capacity ?

keps/sig-network/0007-pod-ready++.md
freehan: Avoid conflicts. Similar to node status.

keps/sig-network/0007-pod-ready++.md
freehan: Patch is the way to go.

keps/sig-network/0007-pod-ready++.md
freehan: pod ready does not mean pod is serving traffic. Hence loss of backend capacity.

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,173 @@
...
+    lastTransitionTime: 2018-01-01T00:00:00Z
+    status: "True"
+    type: www.example.com/feature-2
+  containerStatuses:
+  - containerID: docker://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+    ready : true
+… 
+```
+
+Another pod condition `ContainerReady` will be introduced to capture the old pod `Ready` condition.
```
thockin: ContainersReady (plural) I think

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,173 @@
...
+Another pod condition `ContainerReady` will be introduced to capture the old pod `Ready` condition. 
+```
+ContainerReady is true == containers are ready
+```
+
+### Custom Pod Condition
+Custom pod condition can be injected thru PATCH action using KubeClient. Please be noted that “kubectl patch” does not support patching object status. Need to use client-go or other KubeClient implementations. 
+
+Naming Convention:
+The type of custom pod condition must comply with k8s label key format. For example, “www.example.com/feature-1”.
```
thockin: Will we allow "naked" named or do we require the prefix?

thockin:
May be worth documenting why we don't want to allow updates.

keps/sig-network/0007-pod-ready++.md
krmayankk: @freehan I mean how is this part better than without this feature . Overall I agree just want to understand the reasoning here

keps/sig-network/0007-pod-ready++.md
freehan: But there may be on one container in pod :P

keps/sig-network/0007-pod-ready++.md
```diff
@@ -0,0 +1,173 @@
...
+    type: www.example.com/feature-2
+  containerStatuses:
+  - containerID: docker://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+    ready : true
+… 
+```
+
+Another pod condition `ContainerReady` will be introduced to capture the old pod `Ready` condition. 
+```
+ContainerReady is true == containers are ready
```
krmayankk: Will this replace the old Ready condition or in addition. ? It has to be in addition otherwise this is a breaking change

PR #2009, opened 2018-04-06
Change 'Mac OS X' and 'OS X' to 'macOS'

patrikerdes:
Mac OS X was renamed to OS X in 2012, and then renamed again to [macOS](https://en.wikipedia.org/wiki/MacOS) in 2016.
I suggest that the current name (macOS) should be used whenever referencing the Apple OS.


castrojo:
Thanks for your attention to detail, and welcome to Kubernetes!

/lgtm

PR #2019, opened 2018-04-09
Fix for duplicate title on Contributors guide README

vijaymateti:
Removed duplicate title on contributors guide readme page

### DIFF
--- a/contributors/guide/README.md
+++ b/contributors/guide/README.md
@@ -1,7 +1,3 @@
----
-title: Kubernetes Contributor Guide
----
-
 # Kubernetes Contributor Guide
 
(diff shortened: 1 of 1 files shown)


cblecker:
/assign @castrojo 
I think this might be required

vijaymateti:
Not sure where else it is used but it looks like broken heading on this page https://github.com/kubernetes/community/blob/master/contributors/guide/README.md

castrojo:
This was used originally so we can build imported docs but I think we're going to go with building this page with the community portal instead, so I think we can remove it. 

Thanks for your contribution!

/lgtm

PR #2022, opened 2018-04-09
sig-auth: TPM-based TLS bootstrap proposal

awly:
Original docs shared with sig-auth and sig-node:
https://docs.google.com/document/d/12UgErB_46iHBOi0YEKbaFbEXv9E5O6XWQihwPkwB_7s/
https://docs.google.com/document/d/1Rrqyd9zXjbMsp05tywAxhUMTfaKypj7CHDG85DwXXbo/


contributors/design-proposals/auth/node-tls-bootstrap-via-tpm.md
```diff
@@ -0,0 +1,280 @@
...
+func MakeNodeCSR(privateKey, nodeName) (csrData, error)
+func RequestNodeCertificate(client, csrData, privateKey) (certData, error)
+```
+
+In `bootstap.LoadClientCert`:
+
+```
+csrData, err := csr.MakeNodeCSR(privateKey, nodeName)
+// handle err
+switch config.BootstrapCSRAttestation {
```
ericchiang: Lots of cloud based instance identity strategies will want to inject themselves here too. I'm skeptical of adding more hard coded hooks like this into the codebase. Maybe others can weigh in here?

contributors/design-proposals/auth/node-tls-bootstrap-via-tpm.md
```diff
@@ -0,0 +1,280 @@
...
+
+![Node bootstrap flow using TPM](node-tls-bootstrap-via-tpm.svg)
+
+After machine setup, node bootstrapping will go as follows:
+
+1. Kubelet generates an Elliptic Curve (EC) key pair and stores it on disk
+1. Kubelet triggers AIK creation in the TPM
+1. Kubelet requests that TPM certifies (signs) ECC public key with AIK
+1. Kubelet loads TPM’s AIK cert into memory
+1. Kubelet builds a Certificate Signing Request (CSR) using TLS key pair
```
ericchiang: Just to clarify. "TLS key pair" is the same as the "ECC key pair" discussed earlier and no the AIK key pair, right?

contributors/design-proposals/auth/node-tls-bootstrap-via-tpm.md
```diff
@@ -0,0 +1,280 @@
...
+        csrData = append(csrData, attestData...)
+}
+certData, err := csr.RequestNodeCertificate(client, csrData, privateKey)
+// handle err
+```
+
+Note: in this doc `privateKey` is always `crypto.PrivateKey`, not a PEM-encoded
+byte blob.
+
+### Option 2: exec authentication plugin
```
ericchiang: This section isn't very fleshed out, but based on our conversations this requires the exec auth plugin being able to interact with opaque private keys stored on the TPM. Would it be easier and more general to add pkcs11 support to our clients for using private keys?

contributors/design-proposals/auth/node-tls-bootstrap-via-tpm.md
awly: Yep, "TLS key pair" == "ECC key pair".

contributors/design-proposals/auth/node-tls-bootstrap-via-tpm.md
awly: In the case of exec plugin, TLS key would actually be created outside of the TPM and stored on disk. It would only be leaded in temporarily to do the signing (or "certification" in TPM lingo).
Keys created/stored in the TPM can't be exported so exec plugin wouldn't be able to pass it to kubelet.

Abstracting to PKCS11 is an interesting idea. I don't know if it would be easier from plugin author's perspective, but probably more universal. I'll think about this some more...

awly: @mikedanese had some thoughts on pkcs11.

Also, to clarify, this section isn't very fleshed out because exec plugin would do exactly what kubelet does currently plus changes described above. The code just happens to live in a different binary. Let me know if you'd like more detail here.

mikedanese: pkcs11 is a huge spec covering way more then we need, requires dynamic loading of symbols from object files to use, and is just a painful API to work with IMO:

http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/cs01/pkcs11-curr-v2.40-cs01.pdf

If I wanted to write a plugin that connected a key service to the kubelet TLS stack, I would much prefer a concise, designed to purpose API, vs implementing a pkcs11 backend.

awly:
ping

awly:
This should be possible to build via exec auth plugin and some kind of (unimplemented) remote TLS handshake mechanism to abstract TPM interactions.
Abandoning this proposal.

PR #2025, opened 2018-04-10
Remove SIG on-premise since its now a slack channel

dghubble:


cblecker:
Is it worth moving these docs as they are to `archive/` and removing the sigs.yaml entry completely?

dghubble:
@cblecker yes, agreed. I didn't realize there was a precedent we could follow with SIG Configuration.

dghubble:
Moved page as-is to `archive`, removed OWNERS, removed from `sigs.yaml`.

PR #2026, opened 2018-04-10
CRD versioning design proposal

mbohlool:
CRD versioning proposal is discussed in detail here:

https://docs.google.com/document/d/1Ucf7JwyHpy7QlgHIN2Rst_q6yT0eeN9euzUV6kte6aY/edit

This is the main part of the design. Look at #2054 first.

### DIFF
--- a/contributors/design-proposals/api-machinery/customresources-versioning.md
+++ b/contributors/design-proposals/api-machinery/customresources-versioning.md
@@ -1,0 +1,177 @@
+CRD Versioning
+=============
+
+The objective of this design document is to provide a machinery for Custom Resource Definition authors to define different resource version and a conversion mechanism between them.
+
+# **Background**
+
+Custom Resource Definitions ([CRDs](https://kubernetes.io/docs/concepts/api-extension/custom-resources/)) are a popular mechanism for extending Kubernetes, due to their ease of use compared with the main alternative of building an Aggregated API Servers. They are, however, lacking a very important feature that all other kubernetes objects support: Versioning. Today, each CR can only have one version and there is no clear way for authors to advance their resources to a newer version other than creating a completely new CRD and converting everything manually in sync with their client software.
+
+This document proposes a mechanism to support multiple CRD versions. A few alternatives are also explored at the end of the document. Notably the declarative approach which still has a lot of support among interested parties.
+
+**Goals:**
+
+* Support versioning on API level
+
+* Support conversion mechanism between versions
+
+* Support ability to change storage version
+
+* Support Validation/OpenAPI schema for all versions
+
+**Non-Goals:**
+
+* Support cohabitation (i.e. no group/kind move)
+
+# **Proposed Design**
+
+The basis of the design is a system that supports versioning and minimal conversions. The APIs here, is designed in a way that can be extended with more conversions later.
+
+The summary is to support a list of versions that will include current version. One of these versions can be flagged as the storage version and all versions ever marked as storage version will be listed in a stored_version field in the Status object to enable authors to plan a migration for their stored objects.
+
+The current `Version` field in planned to be deprecated in a later release and will be used to pre-populate the `Versions` field (The `Versions` field will be defaulted to a single version, constructed from top level `Version` field). The `Version` field will be also be mutable to give a way to the authors to remove it from the list.
+
+```golang
+// CustomResourceDefinitionSpec describes how a user wants their resource to appear
+type CustomResourceDefinitionSpec struct {
+  // Group is the group this resource belongs in
+  Group string
+  // Version is the version this resource belongs in
+  // must be always the first item in Versions field if provided.
+  Version string
+  // Names are the names used to describe this custom resource
+  Names CustomResourceDefinitionNames
+  // Scope indicates whether this resource is cluster or namespace scoped.  Default is namespaced
+  Scope ResourceScope
+  // Validation describes the validation methods for CustomResources
+  Validation *CustomResourceValidation
+
+  // ***************
+  // ** New Field **
+  // ***************
+  // Versions is the list of all supported versions for this resource.
+  // Validation: All versions must use the same validation schema for now. i.e., top 
+  // level Validation field is applied to all of these versions.
+  // Order: The order of these versions is used to determine the order in discovery API
+  // (preferred version first).
+  // The versions in this list may not be removed if they are in 
+  // CustomResourceDefinitionStatus.StoredVersions list.
+  Versions []CustomResourceDefinitionVersion
+}
+
+// ***************
+// ** New Type **
+// ***************
+type CustomResourceDefinitionVersion {
+  // Name is the version name, e.g. "v1", “v2beta1”, etc.
+  Name string
+  // Served is a flag enabling/disabling this version from being served via REST APIs
+  Served Boolean
+  // Storage flags the release as a storage version. There can be only one version
+  // flagged as Storage.
+  Storage Boolean
+}
+```
+
+The Status object will have a list of potential stored objects. This data is necessary to do a storage migration in future (the author can choose to do the migration themselves but there is a plan [TODO(lavalamp): link?] to solve the problem of migration, potentially for both standard and custom types).
+
+```golang
+// CustomResourceDefinitionStatus indicates the state of the CustomResourceDefinition
+type CustomResourceDefinitionStatus struct {
+  ...
+
+  // StoredVersions are all versions ever marked as storage in spec. Tracking these
+  // versions allow a migration path for stored version in etcd. The field is mutable
+  // so the migration controller can first make sure a version is certified (i.e. all
+  // stored objects is that version) then remove the rest of the versions from this list.
+  // None of the versions in this list can be removed from the spec.Versions field.
+  StoredVersions []string
+}
+```
+
+Basic validations needed for the `version` field are:
+
+* `Spec.Version` field exists in `Spec.Versions` field (serve should be True for that version if there is any).
+* If there is any `Served` `Version` in list of `Versions`, then the `Spec.Version`’s version in the `Spec.Versions` list should have a set `Served` flag. This is for backward compatibility. An old controller expect that version to be served but only the whole CRD is served. CRD Registration controller should unregister a CRD with no serving version.
+* None of the `Status.StoredVersion` can be removed from `Spec.Versions` list.
+* Only one of the versions in `spec.Versions` can flag as `Storage` version.
+
+# **Declarative conversion**
+
+Most of CRD conversions can be represented by simple operations like rename or move. The proposal is to support rename only and add more operations in future based on user feedback.
+
+```golang
+type CustomResourceDefinitionVersion {
+  // Name is the version name, e.g. "v1", “v2beta1”, etc.
+  Name string
+  // Served is a flag enabling/disabling this version from being served via REST APIs
+  Served Boolean
+  // Storage flags the release as a storage version. There can be only one version
+  // flagged as Storage.
+  Storage Boolean
+  //
+  Conversion CustomResourceConversion
+}
+
+type CustomResourceConversion struct {
+  Declarative map[string]CustomResourceDeclarativeConversion
+}
+
+type CustomResourceDeclarativeConversion {
+  // This type will have collection of conversion operations. Only one of them should be set.
+  Rename RenameConversion
+}
+
+type RenameConversion {
+  From JsonPath
+  To JsonPath
+}
+
+type JsonPath string
+```
+
+# **External Conversion/Webhooks**
+
+Webhooks gives flexibility to CRD author to convert from the specific version to any version. As the conversion needs to be reversible, here we only list versions this webhook converts to/from. 
+
+```golang
+type CustomResourceDefinitionVersion {
+  // Name is the version name, e.g. "v1", “v2beta1”, etc.
+  Name string
+  // Served is a flag enabling/disabling this version from being served via REST APIs
+  Served Boolean
+  // Storage flags the release as a storage version. There can be only one version
+  // flagged as Storage.
+  Storage Boolean
+  //
+  Conversion CustomResourceConversion
+}
+
+type CustomResourceConversion struct {
+  Declarative map[string]CustomResourceDeclarativeConversion
+  External CustomResourceConversionWebhook
+}
+
+type CustomResourceExternalConversion {
+  // List of versions this external conversion supports.
+  To []string
+  Webhook ConversionWebhook
+}
+
+type ConversionWebhook struct {
+  // ClientConfig defines how to communicate with the hook.
+  ClientConfig WebhookClientConfig `json:"clientConfig" protobuf:"bytes,2,opt,name=clientConfig"`
+}
+```
+
+This API design allows multiple webhooks but that should be prevented in validation. Also there is no need to redefine webhook for all versions. For example if on "v1" version we specified a webhook to convert it to “v2”, we do not need to add that webhook configuration to “v2” as the conversion is defined in both ways.
+
+**Note: ** Webhooks will, intentionally, be not included in the first version of the feature to make sure authors waiting for this feature is trying the declarative approach first. The Webhook approach is a fallback for all cases that declarative cannot support.
+
+# **Validation**
+
+The API structure above allows some invalid states in order to keep a clean easy API. There need to be validation to catch undesire states. The validations would be
+
+* Multiple webhooks are not allowed for a CRD. This will help us improve the performance by keeping the connection open to the webhook if the user defined a chain of version conversions.
+* There should be no loop in the conversion graph. This includes v1->v2->v1 or v1->v2->v3->v1. The former is implicit in the assumption that webhook conversion is round-trippable. The second one allows two way to convert v1 to v3 which is not desired.
+* Rename operation can be limited at first to not change the structure despite the fact that `To` field is also a json path.


contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
+CRD Versioning
+=============
+
+The objective of this design document is to provide a machinery for Custom Resource Definition authors to define different resource version and a conversion mechanism between them.
+
+# **Background**
+
+Custom Resource Definitions ([CRDs](https://kubernetes.io/docs/concepts/api-extension/custom-resources/)) are a popular mechanism for extending Kubernetes, due to their ease of use compared with the main alternative of building an Aggregated API Servers. They are, however, lacking a very important feature that all other kubernetes objects support: Versioning. Today, each CR can only have one version and there is no clear way for authors to advance their resources to a newer version other than creating a completely new CRD and converting everything manually in sync with their client software.
```
sttts: nit: server

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+
+**Goals:**
+
+* Support versioning on API level
+
+* Support conversion mechanism between versions
+
+* Support ability to change storage version
+
+* Support Validation/OpenAPI schema for all versions
```
sttts: for each or for all at once?

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+
+* Support cohabitation (i.e. no group/kind move)
+
+# **Proposed Design**
+
+The basis of the design is a system that supports versioning and minimal conversions. The APIs here, is designed in a way that can be extended with more conversions later.
+
+The summary is to support a list of versions that will include current version. One of these versions can be flagged as the storage version and all versions ever marked as storage version will be listed in a stored_version field in the Status object to enable authors to plan a migration for their stored objects.
+
+The current `Version` field in planned to be deprecated in a later release and will be used to pre-populate the `Versions` field (The `Versions` field will be defaulted to a single version, constructed from top level `Version` field). The `Version` field will be also be mutable to give a way to the authors to remove it from the list.
```
sttts: is

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+
+// ***************
+// ** New Type **
+// ***************
+type CustomResourceDefinitionVersion {
+  // Name is the version name, e.g. "v1", “v2beta1”, etc.
+  Name string
+  // Served is a flag enabling/disabling this version from being served via REST APIs
+  Served Boolean
+  // Storage flags the release as a storage version. There can be only one version
```
sttts: There **must be exactly one** version flagged as Storage.

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+  // flagged as Storage.
+  Storage Boolean
+}
+```
+
+The Status object will have a list of potential stored objects. This data is necessary to do a storage migration in future (the author can choose to do the migration themselves but there is a plan [TODO(lavalamp): link?] to solve the problem of migration, potentially for both standard and custom types).
```
sttts: stored versions.

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+  Scope ResourceScope
+  // Validation describes the validation methods for CustomResources
+  Validation *CustomResourceValidation
+
+  // ***************
+  // ** New Field **
+  // ***************
+  // Versions is the list of all supported versions for this resource.
+  // Validation: All versions must use the same validation schema for now. i.e., top 
+  // level Validation field is applied to all of these versions.
```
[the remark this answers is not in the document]
sttts: The same is true for subresources. Should we move them to `Versions` as well? I think we should. Today there is no graceful way to enable the `/status` subresource without breaking either old or new clients.

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+  // so the migration controller can first make sure a version is certified (i.e. all
+  // stored objects is that version) then remove the rest of the versions from this list.
+  // None of the versions in this list can be removed from the spec.Versions field.
+  StoredVersions []string
+}
+```
+
+Basic validations needed for the `version` field are:
+
+* `Spec.Version` field exists in `Spec.Versions` field (serve should be True for that version if there is any).
```
sttts: so we force at least one served version. Do we want that? Do we even have to (to allow e.g. the GC controller to do its work)?

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+* If there is any `Served` `Version` in list of `Versions`, then the `Spec.Version`’s version in the `Spec.Versions` list should have a set `Served` flag. This is for backward compatibility. An old controller expect that version to be served but only the whole CRD is served. CRD Registration controller should unregister a CRD with no serving version.
```
sttts: Isn't this "if" implicitly true by the "serve should be True" above?
Can `spec.version` be empty?

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+type CustomResourceDefinitionVersion {
+  // Name is the version name, e.g. "v1", “v2beta1”, etc.
+  Name string
+  // Served is a flag enabling/disabling this version from being served via REST APIs
+  Served Boolean
+  // Storage flags the release as a storage version. There can be only one version
+  // flagged as Storage.
+  Storage Boolean
+  //
+  Conversion CustomResourceConversion
```
sttts: pointer with omitempty.

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+}
+
+type CustomResourceConversion struct {
+  Declarative map[string]CustomResourceDeclarativeConversion
```
sttts: what's the meaning of the string? 
Are declarative conversions sorted?

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+  External CustomResourceConversionWebhook
```
sttts: pointer and omitempty

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+  // Served is a flag enabling/disabling this version from being served via REST APIs
+  Served Boolean
+  // Storage flags the release as a storage version. There can be only one version
+  // flagged as Storage.
+  Storage Boolean
+  //
+  Conversion CustomResourceConversion
+}
+
+type CustomResourceConversion struct {
```
sttts: are multiple conversions mutual exclusive for each version pair?

contributors/design-proposals/api-machinery/customresources-versioning.md
sttts: Having a webhook definition per version looks odd. I would expect it globally with a set of support conversions pairs. Then you don't have to maintain `n` webhook structs. Alternatively, we could think about a global named conversion webhook field and reference the names here (like volumes and mounts in pods).

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+
+**Note: ** Webhooks will, intentionally, be not included in the first version of the feature to make sure authors waiting for this feature is trying the declarative approach first. The Webhook approach is a fallback for all cases that declarative cannot support.
+
+# **Validation**
+
+The API structure above allows some invalid states in order to keep a clean easy API. There need to be validation to catch undesire states. The validations would be
+
+* Multiple webhooks are not allowed for a CRD. This will help us improve the performance by keeping the connection open to the webhook if the user defined a chain of version conversions.
+* There should be no loop in the conversion graph. This includes v1->v2->v1 or v1->v2->v3->v1. The former is implicit in the assumption that webhook conversion is round-trippable. The second one allows two way to convert v1 to v3 which is not desired.
```
sttts: Isn't a loop fine and necessary? Every path without a loop should be unique.

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+
+type CustomResourceDeclarativeConversion {
+  // This type will have collection of conversion operations. Only one of them should be set.
+  Rename RenameConversion
+}
+
+type RenameConversion {
+  From JsonPath
+  To JsonPath
+}
```
sttts: To avoid undesired renames, better use a base json path and a field rename:
```
type RenameConversion struct {
  Object JsonPath
  FieldName string
  NewFieldName string
}
```

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+
+**Note: ** Webhooks will, intentionally, be not included in the first version of the feature to make sure authors waiting for this feature is trying the declarative approach first. The Webhook approach is a fallback for all cases that declarative cannot support.
+
+# **Validation**
+
+The API structure above allows some invalid states in order to keep a clean easy API. There need to be validation to catch undesire states. The validations would be
+
+* Multiple webhooks are not allowed for a CRD. This will help us improve the performance by keeping the connection open to the webhook if the user defined a chain of version conversions.
+* There should be no loop in the conversion graph. This includes v1->v2->v1 or v1->v2->v3->v1. The former is implicit in the assumption that webhook conversion is round-trippable. The second one allows two way to convert v1 to v3 which is not desired.
+* Rename operation can be limited at first to not change the structure despite the fact that `To` field is also a json path.
```
sttts: See suggestion above: we only want to rename fields in objects, no moves, no structure changes.

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: Added a clarification.

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: I think we don't want that. We discussed this on the implementation PR. I will amend this.

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: I changed previous one. So the if is not true all the time. You can disable all versions and in that case the version defined in `Version` field does not need to have `Served` flag enabled. But if we serve ANY version, then the version defined in `Version` field should point to a `Served` version.

mbohlool: I updated the wording to make this more clear.

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: done

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: They are not sorted. The map key is the target version.

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: I explained the design further down. It is cleaner to define the webhook on the version but the validation will not let user define more than one webhook. It will just make sure if more than one webhook is defined, they are the same.

mbohlool:
@sttts I added the webhook to the original document and asked people to look at it and I thought there is no objection on the design. I see most of your comments is on webhooks. We can discuss that here, but as we do not want to implement webhooks in 1.11, Do you think it is better to remove it from here and continue discussion on the Google Doc one. It is easier to discuss things there.

contributors/design-proposals/api-machinery/customresources-versioning.md
sttts: Don't like this kind of redundancy. Why is it cleaner?

sttts:
> Do you think it is better to remove it from here and continue discussion on the Google Doc one

We can sketch webhook here, but clearly call it a sketch, i.e. to be clear that this must be worked out.

liggitt:
> Do you think it is better to remove it from here and continue discussion on the Google Doc one. It is easier to discuss things there.

It seems better to describe future webhook plans generally, not as API fields yet, if they are still being designed. The previous review concentrated on the parts under consideration for implementation in 1.10.

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: I was thinking we would let structural changes later (carefully) and we can prevent them with validation for now. This way we do not need to change API or add another conversion for structural changes but I am also OK if we are thinking about a `MoveConversion` later though I liked the end product to only have `MoveConversion` and not Move and Rename both.

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+  Conversion CustomResourceConversion
+}
+
+type CustomResourceConversion struct {
+  Declarative map[string]CustomResourceDeclarativeConversion
+}
+
+type CustomResourceDeclarativeConversion {
+  // This type will have collection of conversion operations. Only one of them should be set.
+  Rename RenameConversion
```
sttts: This have to be an array.

sttts: Why should only one of them be set? 

We can have renames, moves, deletions all at the same time. We need a logic to detect conflicts during validation of course.

contributors/design-proposals/api-machinery/customresources-versioning.md
sttts: If we call it `RenameConversion`, we should be strict. If we call it `MoveConversion` we could restrict it to renames for the moment. Both is fine with me.

contributors/design-proposals/api-machinery/customresources-versioning.md
deads2k: maps aren't allowed in our API

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+  // Served is a flag enabling/disabling this version from being served via REST APIs
+  Served Boolean
+  // Storage flags the release as a storage version. There can be only one version
+  // flagged as Storage.
+  Storage Boolean
+  //
+  Conversion CustomResourceConversion
+}
+
+type CustomResourceConversion struct {
```
deads2k: This is more involved than the "no changes" we had talked about before.  How about we start small?

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+Basic validations needed for the `version` field are:
+
+* `Spec.Version` field exists in `Spec.Versions` field (serve should be True for that version if there is any).
+* If there is any `Served` `Version` in list of `Versions`, then the `Spec.Version`’s version in the `Spec.Versions` list should have a set `Served` flag. This is for backward compatibility. An old controller expect that version to be served but only the whole CRD is served. CRD Registration controller should unregister a CRD with no serving version.
+* None of the `Status.StoredVersion` can be removed from `Spec.Versions` list.
+* Only one of the versions in `spec.Versions` can flag as `Storage` version.
+
+# **Declarative conversion**
```
deads2k: Ah, ok.  I think I see what happened.  You're saying "these are some thoughts I had", not "I want to build this an merge it in 1.11".  If that's the case, can you add a "Declarative conversion design not approved in this document" in big bold letters leading the section?

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: This can be an array or the `CustomResourceDeclarativeConversion` can be an array in the map. which one is better? I personally think it is more explicit to make the `CustomResourceDeclarativeConversion` an array then the conversion would be look like:

```
v1:
  - Rename: ...
  ...
  - Rename: ...
  ...
  - Move: ...
  ...
...
```

Also deletion cannot happened according to our API guidelines. User can rename the field to a in-name-marked deprecated to deleted field (e.g. deprecated_... or deleted_...).

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: Being more explicit is better I guess. I went with your `RenameConversion` suggestion.

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: The first PR will be no change, but for 1.11 we want at least rename. Do you have any objection to the rename design here?

contributors/design-proposals/api-machinery/customresources-versioning.md
mbohlool: Actually these are the things I want to have in 1.11 (only rename). Here we can discuss it if you have any thoughts. These are designed I discussed in one form or another with Stephan, Daniel, and Eric but none approved it yet. I guess that is the goal of this PR.

PR #2034, opened 2018-04-11
Topology updates to Kubernetes CSI implementation

verult:
This proposal depends on the following design changes:
Topology-aware dynamic provisioning design: https://github.com/kubernetes/community/pull/1857
CSI spec modifications related to topology: https://github.com/container-storage-interface/spec/pull/188/

/sig storage
/assign @saad-ali
/cc @vladimirvivien

### DIFF
--- a/contributors/design-proposals/storage/container-storage-interface.md
+++ b/contributors/design-proposals/storage/container-storage-interface.md
@@ -414,6 +414,97 @@ The alternative is invalid because the driver name could be up to 63 characters 
 
 Because CSI topology values only contain the topology name (`rack`) and topology value (`rack1`), the driver name and the constant prefix must be prepended to the label as part of the provisioner's topology translation step.
 
+#### Topology Representation in PersistentVolume Objects
+There exists multiple ways to represent a single topology as NodeAffinity. For example, suppose a `CreateVolumeResponse` contains the following accessible topology:
+
+```yaml
+- zone: "a"
+  rack: "1"
+- zone: "b"
+  rack: "1"
+- zone: "b"
+  rack: "2"
+```
+
+There are at least 3 ways to represent this in NodeAffinity (excluding `nodeAffinity`, `required`, and `nodeSelectorTerms` for simplicity):
+
+Form 1 - `values` contain exactly 1 element.
+```yaml
+- matchExpressions:
+  - key: zone
+    operator: In
+    values:
+    - "a"
+  - key: rack
+    operator: In
+    values:
+    - "1"
+- matchExpressions:
+  - key: zone
+    operator: In
+    values:
+    - "b"
+  - key: rack
+    operator: In
+    values:
+    - "1"
+- matchExpressions:
+  - key: zone
+    operator: In
+    values:
+    - "b"
+  - key: rack
+    operator: In
+    values:
+    - "2"
+```
+
+Form 2 - Reduced by `rack`.
+```yaml
+- matchExpressions:
+  - key: zone
+    operator: In
+    values:
+    - "a"
+    - "b"
+  - key: rack
+    operator: In
+    values:
+    - "1"
+- matchExpressions:
+  - key: zone
+    operator: In
+    values:
+    - "b"
+  - key: rack
+    operator: In
+    values:
+    - "2"
+```
+Form 3 - Reduced by `zone`.
+```yaml
+- matchExpressions:
+  - key: zone
+    operator: In
+    values:
+    - "a"
+  - key: rack
+    operator: In
+    values:
+    - "1"
+- matchExpressions:
+  - key: zone
+    operator: In
+    values:
+    - "b"
+  - key: rack
+    operator: In
+    values:
+    - "1"
+    - "2"
+```
+The provisioner will always choose Form 1, i.e. all `values` will have at most 1 element. Reduction logic could be added in future versions to arbitrarily choose a valid and simpler form like Forms 2 & 3.
+
 ### Example Walkthrough
 
 #### Provisioning Volumes
--- a/contributors/design-proposals/storage/container-storage-interface.md
+++ b/contributors/design-proposals/storage/container-storage-interface.md
@@ -89,14 +89,12 @@ CSI volume drivers should create a socket at the following path on the node mach
 
 `Sanitized CSIDriverName` is CSI driver name that does not contain dangerous character and can be used as annotation name. It can follow the same pattern that we use for [volume plugins](https://git.k8s.io/kubernetes/pkg/util/strings/escape.go#L27). Too long or too ugly driver names can be rejected, i.e. all components described in this document will report an error and won't talk to this CSI driver. Exact sanitization method is implementation detail (SHA in the worst case).
 
-Upon initialization of the external “CSI volume driver”, some external component must call the CSI method `GetNodeId` to get the mapping from Kubernetes Node names to CSI driver NodeID. It must then add the CSI driver NodeID to the `csi.volume.kubernetes.io/nodeid` annotation on the Kubernetes Node API object. The key of the annotation must be `csi.volume.kubernetes.io/nodeid`. The value of the annotation is a JSON blob, containing key/value pairs for each CSI driver.
-
-For example:
+Upon initialization of the external “CSI volume driver”, some external component must call the CSI method `GetNodeId` to get the mapping from Kubernetes Node names to CSI driver NodeID and the associated AccessibleTopology. It must then update the Kubernetes Node API object with the CSI driver NodeID as the `csi.volume.kubernetes.io/nodeid` annotation and AccessibleTopology as labels. The key of the NodeID annotation must be `csi.volume.kubernetes.io/nodeid`. The value of the annotation is a JSON blob, containing key/value pairs for each CSI driver. For example:
 ```
 csi.volume.kubernetes.io/nodeid: "{ \"driver1\": \"name1\", \"driver2\": \"name2\" }
 ```
 
-This will enable the component that will issue `ControllerPublishVolume` calls to use the annotation as a mapping from cluster node ID to storage node ID.
+This will enable the component that will issue `ControllerPublishVolume` calls to use the annotation as a mapping from cluster node ID to storage node ID. There are no hard restrictions on the label format, but for the format to be used by the recommended setup, please refer to [Topology Representation in Node Objects](#topology-representation-in-node-objects). 
 
 To enable easy deployment of an external containerized CSI volume driver, the Kubernetes team will provide a sidecar "Kubernetes CSI Helper" container that can manage the unix domain socket registration and NodeId initialization. This is detailed in the “Suggested Mechanism for Deploying CSI Drivers on Kubernetes” section below.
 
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -116,7 +114,7 @@ Provisioning and deletion operations are handled using the existing [external pr
...
...on of the PVC and issue the `CreateVolume` call against the CSI volume driver to provision the volume. The `CreateVolume` name will be auto-generated as it is for other dynamically provisioned volumes. The `CreateVolume` capacity will be taken from the `PersistentVolumeClaim` object. The `CreateVolume` parameters will be passed through from the `StorageClass` parameters (opaque to Kubernetes). If the `PersistentVolumeClaim` has the `selectedNode` annotation set (TODO verult update to actual annotation name) (only added if delayed volume binding is enabled in the `StorageClass`), the provisioner will get relevant topology labels from the corresponding `Node` and pass them to the `CreateVolume` call as preferred topology. `AllowedTopologies`from the `StorageClass` is passed through as permitted topology. Once the operation completes successfully, the external provisioner creates a `PersistentVolume` object to represent the volume using the information returned in the `CreateVolume` response. The topology of the returned volume is translated to the `PersistentVolume` `NodeAffinity` field. The `PersistentVolume` object is then bound to the `PersistentVolumeClaim` and available for use.
```
saad-ali: nit: Spacing around `AllowedTopologies`

saad-ali: Can you describe this in more detail: `The topology of the returned volume is translated to the `PersistentVolume` `NodeAffinity` field`

contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -387,6 +385,25 @@ To deploy a containerized third-party CSI volume driver, it is recommended that
...
+
+Topology information will be represented as labels.
+
+Requirements:
+* Must adhere to the [label format](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).
+* Must support different drivers on the same node.
+* The format of each key/value pair must match those in `PersistentVolume` and `StorageClass` objects. When a `StorageClass` has delayed volume binding enabled, the scheduler uses the topology information of a `Node` in the following ways:
+  1. During dynamic provisioning, the scheduler selects a candidate node for the provisioner by comparing each `Node`'s topology with the `AllowedTopology` in the `StorageClass`. (TODO verult Link to volume scheduling design doc)
+  1. During volume binding and pod scheduling, the scheduler selects a candidate node for the pod by comparing `Node` topology with `VolumeNodeAffinity` in `PersistentVolume`s. (TODO verult Link to volume scheduling design doc)
+* Must avoid collision with topology specified from sources other than CSI.
```
saad-ali: Must avoid collision with **labels** specified from...?

contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -387,6 +385,25 @@ To deploy a containerized third-party CSI volume driver, it is recommended that
...
 
+#### Topology Representation in Node Objects
+**Work In Progress**
+
+Topology information will be represented as labels.
+
+Requirements:
+* Must adhere to the [label format](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).
+* Must support different drivers on the same node.
+* The format of each key/value pair must match those in `PersistentVolume` and `StorageClass` objects. When a `StorageClass` has delayed volume binding enabled, the scheduler uses the topology information of a `Node` in the following ways:
```
saad-ali: > The format of each key/value pair must match those in `PersistentVolume` and `StorageClass` objects.

What does this mean?

contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -387,6 +385,25 @@ To deploy a containerized third-party CSI volume driver, it is recommended that
...
+  1. During dynamic provisioning, the scheduler selects a candidate node for the provisioner by comparing each `Node`'s topology with the `AllowedTopology` in the `StorageClass`. (TODO verult Link to volume scheduling design doc)
+  1. During volume binding and pod scheduling, the scheduler selects a candidate node for the pod by comparing `Node` topology with `VolumeNodeAffinity` in `PersistentVolume`s. (TODO verult Link to volume scheduling design doc)
+* Must avoid collision with topology specified from sources other than CSI.
+
+Proposal: `"csi.kubernetes.io.csi-driver.example.com/rack": "rack1"`
```
saad-ali: dash for delimiter is allowed?

verult: Yup, but not underscore

contributors/design-proposals/storage/container-storage-interface.md
verult: The topology format is ultimately controlled by the user, so the same applies to the translation. I'll clarify the steps for the recommended provisioner, though.

verult:
Added more discussion around PV NodeAffinity

contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -387,6 +450,126 @@ To deploy a containerized third-party CSI volume driver, it is recommended that
...
+* The same topology key could be used across different sources (different storage plugin, network plugin, etc.)
+* Once NodeRestriction is moved to the newer model (see [here](https://github.com/kubernetes/community/pull/911) for context), for each new label prefix introduced in a new driver, the cluster admin has to configure NodeRestrictions to allow the driver to update labels with the prefix. Cluster installations could include certain prefixes for pre-installed drivers by default. This is less convenient compared to the alternative, which can allow editing of all CSI drivers by default using the “csi.kubernetes.io” prefix, but often times cluster admins have to whitelist those prefixes anyway (for example ‘cloud.google.com’)
+
+Considerations:
+* Upon driver deletion/upgrade/downgrade, stale labels will be left untouched. It’s difficult for the driver to decide whether other sources rely on this label.
+* During driver installation/upgrade/downgrade, controller deployment must be brought down before node deployment, and node deployment must be deployed before the controller deployment, because provisioning relies on up-to-date node information.
```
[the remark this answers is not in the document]
verult: I can't think of an issue immediately. Will update this line to discuss why this isn't an issue and what's the behavior if both deployments are changed in parallel. I have a comment in the Upgrades & Downgrades sections about recommending users to still upgrade those deployments independently, and I'll leave that there just to be safe.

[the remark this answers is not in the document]
verult: One potential issue: if only topology values change while keys remain the same, and if AllowedTopologies is not specified, requisite topology will contain both old and new topology values, and CSI driver may fail the `CreateVolume()` call.

contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -387,6 +450,126 @@ To deploy a containerized third-party CSI volume driver, it is recommended that
...
+* Topology keys inside `CSINodeInfo` must reflect the topology keys from drivers currently installed on the node. If no driver is installed, the collection must be empty.
```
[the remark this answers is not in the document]
verult: A similar race problem arises as well with node rolling upgrades - we could end up with a requisite topology that doesn't contain the selected node.

contributors/design-proposals/storage/container-storage-interface.md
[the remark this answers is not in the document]
msau42: I think considering that topology labels can be shared with other components, we cannot override existing topology values if the key already exists on the node.

contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -116,7 +158,27 @@ Provisioning and deletion operations are handled using the existing [external pr
...
+
+If the `PersistentVolumeClaim` has the `volume.alpha.kubernetes.io/selected-node` annotation set (only added if delayed volume binding is enabled in the `StorageClass`), the provisioner will get relevant topology keys from the corresponding `CSINodeInfo` instance and the topology values from `Node` labels and use them to generate preferred topology in the `CreateVolume()` request. If the annotation is unset, preferred topology will not be specified. `AllowedTopologies` from the `StorageClass` is passed through as requisite topology. If `AllowedTopologies` is unspecified, the provisioner will pass in a set of aggregated topology values across the whole cluster as requisite topology.
```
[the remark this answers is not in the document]
verult: If the PVC has immediate binding and doesn't follow a StatefulSet naming scheme (or the external provisioner has StatefulSet spreading logic disabled), preferred topology will not be set

contributors/design-proposals/storage/container-storage-interface.md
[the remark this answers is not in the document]
verult: I think users should be free to do so if they can guarantee other components can handle new values too. Maybe we could call this out as a caveat when performing driver upgrades.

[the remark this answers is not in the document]
msau42: I would prefer in that case to have the user delete the label manually first instead of us just blindly overriding it and potentially messing up other components.

contributors/design-proposals/storage/container-storage-interface.md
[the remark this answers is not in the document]
verult: Synced with @msau42 offline. Like she said, there is always going to be a chance the provisioner has out-of-date topology information on the node, but if the provisioner acts on that case, we are OK - it's not any worse than how CSI works today, since volumes could totally be provisioned in a volume where it's not accessible.

There's a few things we could do to minimize this possibility as much as possible. For example, if the driver info could be deleted from `CSINodeInfo` as soon as possible after a node driver deletion or crash, the provisioner could do the following:
* Never use that node as the source of truth for topology keys.
* Don't include that node in the aggregated requisite topology.
In addition, the scheduler could check `CSINodeInfo` and invalidate nodes that don't have the appropriate driver information as potential scheduling candidates.

Will update the doc

verult:
Addressed comments

tallclair:
I don't want to block this, but I do want to point out that we've discussed moving CSI drivers into sandboxes. This would yield {driver x runtime} tuples, rather than {drixer x node}. Again, too early to do anything about it, but maybe worth keeping in mind.

See [storage attack surfaces](https://docs.google.com/document/d/1QQ5u1RBDLXWvC8K3pscTtTRThsOeBSts_imYEoRyw8A/edit#heading=h.d6xmreeqk1n6) for more details.

contributors/design-proposals/storage/container-storage-interface.md
```diff
@@ -387,6 +451,127 @@ To deploy a containerized third-party CSI volume driver, it is recommended that
...
+
+Considerations:
+* Upon driver deletion/upgrade/downgrade, stale labels will be left untouched. It’s difficult for the driver to decide whether other components outside CSI rely on this label.
+* During driver installation/upgrade/downgrade, controller deployment must be brought down before node deployment, and node deployment must be deployed before the controller deployment, because provisioning relies on up-to-date node information. One possible issue is if only topology values change while keys remain the same, and if AllowedTopologies is not specified, requisite topology will contain both old and new topology values, and CSI driver may fail the CreateVolume() call. Given that CSI driver should be backward compatible, this is more of an issue when a node rolling upgrade happens before the controller update. It's not an issue if keys are changed as well since requisite and preferred topology generation handles it appropriately.
```
msau42: Is this an issue if we don't override the label values?  I would say changing the value of an existing label is an incompatible change.  If a driver really needs to introduce a new value, then it should be through a new key.

verult: ACK, will update this paragraph and also indicate that a driver will be rejected by the node if there is a topology value conflict

saad-ali:
This was implemented in https://github.com/kubernetes-csi/external-provisioner/pull/141

/lgtm
/approve

PR #2036, opened 2018-04-11
renamed sig-contributor-experience google group

grantbow:


tpepper:
My only other request is to put something in the commit description indicating the what/why, eg:

```
As per
https://groups.google.com/d/topic/kubernetes-sig-contribex/sFWBwGUHszc/discussion
the "kubernetes-wg-contribex" group name is now "kubernetes-sig-contribex".
```

tpepper:
(and we'll want to squash on merge)

tpepper:
@cblecker @grodrigues3 @idvoretskyi @calebamiles we should get this merged ASAP so we don't have broken links lingering after yesterday's change.  For example the top level sig list file is very highly visible.

PR #2042, opened 2018-04-13
User namespaces KEP.

adelton:


adelton:
I'd appreciate any hints about the suitability of this PR at this stage. Specifically, I have some work-in-progress changes working towards the implementation, especially enhancing the API. Should I start creating the PRs right now and reference them from the KEP document already now, should I put diffs directly to the KEP document, or should we focus on describing the goals in the KEP for now and getting what we'd like to achieve, without necessarily looking at the code (the how) just yet?

derekwaynecarr:
@mrunalp @adelton @vikaschoudhary16 are going to iterate on this further in a separate google doc to be linked here.  plan is to present something to sig-node ~4/24 meeting.

derekwaynecarr:
@adelton - lets focus on the use cases and goals first and then we can move to the code after.

keps/0007-20180413-user-namespaces.md
```diff
@@ -0,0 +1,267 @@
...
+The goal is not to support use cases when the uids for processes get
+allocated in runtime and externally (IPA, Active Directory) for generic
+users, without specifying (hopefully narrow) range of uids to be mapped
+upfront, before the container gets started.
+
+## Proposal
+
+This is where we get down to the nitty gritty of what the proposal actually is.
+
+### User Stories / Options
```
derekwaynecarr: i think it may help describing this section in terms of distinct feature gates.

1.  GlobalUserNamespace
1.  NamedUserNamespace

For item 1, please expand upon the mechanism by which across CRI implementations the kubelet could discover the mapping.  This is similar to discussion from @filbranden on the idea of a "Runtime or Distro ships static configuration file" in `/etc/cri-runtime/xyz` so the kubelet can introspect in a common manner.

For item 2 and item 3, please expand on the API changes and CRI change required in this KEP.

Item 1 can deliver a lot of value with minimal end-user pain so I prefer to see that as the first phase.

krmayankk:
I love this effort please share the doc when you have it

keps/0007-20180413-user-namespaces.md
derekwaynecarr: also just capturing here for others, we are aware of a number of problems that need to be resolved at much lower levels to support "run containers with namespace specific userns remapping" especially around management of container filesystem.  as a result, the kep should describe the near term of what is possible today, and what the long term would be if those issues were resolved.

adelton:
The user namespace support effort was split into smaller steps, the first one using Node-level user namespace support in Container Runtime: https://github.com/kubernetes/community/pull/2067

PR #2045, opened 2018-04-13
Add KEP for topology-aware workload controllers

janetkuo:
Topology-aware workload controllers
Ref https://github.com/kubernetes/kubernetes/issues/41442

/sig apps
/sig scheduling
@kow3ns @bsalamat @enisoc @davidopp @erictune @smarterclayton


kargakis:
@tnozicka @mfojtik

mattfarina:
@janetkuo Could you talk about this in a SIG Apps meeting?

keps/sig-apps/0008-topology-aware-workload-controllers.md
```diff
@@ -0,0 +1,350 @@
...
+## Proposal
+
+We can think of updates as planned disruptions. When a workload controller
+performs a rolling update, it will evict (instead of deleting) old pods and
+creates new pods. Users will be able to use [`PodDisruptionBudget`][](PDB) to
+control all sorts of planned disruptions, including updates. 
+
+To achieve this, PDB should allow users to specify allowed pod disruptions for
+a collection of nodes (such as the nodes in one failure domain), using a new PDB
+field `.spec.nodeSelectorTerm`.
```
bsalamat: Please add that the existing label selector for pods is ANDed with the new node selector when the PDB is matched against pods.

keps/sig-apps/0008-topology-aware-workload-controllers.md
```diff
@@ -0,0 +1,350 @@
...
+kind: PodDisruptionBudget
+metadata:
+  name: app-pdb
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
+      app: foo
+  nodeSelectorTerm:
+      # Only allow 1 pod to be evicted/updated per zone
```
bsalamat: How does this allow 1 pod to be evicted per zone? The matchExpression wants the node to have the "zone" label and all the nodes of the cluster have such label. So, I think this rule causes only 1 pod to be evicted in the whole cluster, not 1 pod per zone.

keps/sig-apps/0008-topology-aware-workload-controllers.md
```diff
@@ -0,0 +1,350 @@
...
+
+To achieve this, PDB should allow users to specify allowed pod disruptions for
+a collection of nodes (such as the nodes in one failure domain), using a new PDB
+field `.spec.nodeSelectorTerm`. 
+
+This requires changes to workloads controllers code (change pod deletions to
+evictions), PDB API, and PDB controller code.
+
+[`PodDisruptionBudget`]: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
+
```
bsalamat: In a rolling update, I guess a common scenario is to allow only a certain number of pods to be evicted per failure domain. I am not sure if this requirement can be specified with the proposed solution. More on this in my comment on "Story 2".
Instead of a node selector, we may want to add a "topology" field to PDB which acts similar to "topology" of inter-pod affinity. That would allow us to specify a certain number of unavailable pods per failure domain.

keps/sig-apps/0008-topology-aware-workload-controllers.md
```diff
@@ -0,0 +1,350 @@
...
+specify things like: 
+
+* Update by failure domain: Update one failure domain at a time (staged domain)
+* Update across failure domains: Only allow X number of replicas to be down per
+  failure domain 
+
+## Proposal
+
+We can think of updates as planned disruptions. When a workload controller
+performs a rolling update, it will evict (instead of deleting) old pods and
```
tnozicka: With deletion you can delete a Pod on a Node that's down - it sets the deletionTimestamp and continue. Will eviction work even in case the node is down and in NotReady state?

keps/sig-apps/0008-topology-aware-workload-controllers.md
```diff
@@ -0,0 +1,350 @@
...
+
+### Risks and Mitigations
+
+Controllers will be changed to evict pods instead of delete pods.
+
+Risks:
+
+* If a pod eviction fails because of PDB, the controller may enter eviction
+  hotloop.
+* The controller may create more API calls than it should, because the
```
tnozicka: If I am getting it correctly, this makes me worried about performance, maybe the approach itself if the workload controller will be fighting with PDB with etcd in between. (Succeeding eventually, but producing unnecessary load.)

keps/sig-apps/0008-topology-aware-workload-controllers.md
kargakis: evict is delete plus the pdb checks so yes it should work

bgrant0607:
cc @thockin, who has thoughts on topology

bgrant0607:
FYI: volume topology discussion and presentation:
https://groups.google.com/forum/#!topic/kubernetes-sig-architecture/op7YJ-Fw5AM

justaugustus:

justaugustus:

/close

PR #2046, opened 2018-04-14
Fix mistakes about the example of resource account for burstable pods and typo mistake

fatsheep9146:
Fix mistake about the example about calculating the cpu.shares of a burstable pod (Pod3) with two containers. Since the container "bar" in Pod3 is only specified with resource.limit, so the resource.request of container "bar" is equal to its limits, so request.cpu is 100m, request.memory is 1Gi. 

Therefore, when you calculate the pod level cgroup's cpu.shares should be the sum of resource.request.cpu of container "foo" and "bar",  which is 120m. The cpu share of burstable cgroup should also be modified to 130m.


derekwaynecarr:
thank you for the fix.

/ok-to-test
/approve
/lgtm

Issue #1594, opened 2018-01-12
Add info on KUBE_TIMEOUT in testing doc for troubleshooting


johnarok:
Refering Doc: https://github.com/kubernetes/community/blob/master/contributors/devel/testing.md

While following the steps detailed on the above document, the unit tests for pkg/master would consistently error with a panic (see below) and the only way to get past it is to increase the KUBE_TIMEOUT on /hack/make-rules/test.sh to something greater than 120s in the kubernetes/kubernetes repo. Thinking 300s will be a safer.  

```sh
panic: test timed out after 2m0s

goroutine 4243 [running]:
testing.startAlarm.func1()
       ~/go/src/testing/testing.go:1145 +0xf9
created by time.goFunc
       ~/go/src/time/sleep.go:170 +0x44
````
I would like to add this as a pointer to the doc, under the run all unit tests section https://github.com/kubernetes/community/blob/master/contributors/devel/testing.md#run-all-unit-tests so it will help future readers. If there is a different place to document this or If it is better to increase the value on the script in kubernetes repo, please let me know. I can submit PR for either changes.

Below is the text I would like add.
--- START OF TEXT
if any of unit test fail with a timeout panic (see issue#here) on the testing package. you can try increasing the KUBE_TIMEOUT value as shown below.

```sh
# edit 
vim hack/make-rules/test.sh
# change the value from 120s to 300s
KUBE_TIMEOUT=${KUBE_TIMEOUT:--timeout 300s}
```
--- END OF TEXT

cblecker:
You shouldn't have to edit the file. You can override the variable with `make test KUBE_TIMEOUT=-timeout 300s`

cblecker:
Yes, putting a note with the syntax of how to boot the timeout would be good.

pisarenko-net:
I've been running into the same issue when trying to start with k8s development.

johnarok:
Submitted PR #2047 to add the note to testing.md

PR #2047, opened 2018-04-14
Add note on KUBE_TIMEOUT to resolve timeout panic

johnarok:
For #1594

### DIFF
--- a/contributors/devel/testing.md
+++ b/contributors/devel/testing.md
@@ -54,7 +54,7 @@ cd kubernetes
 make test  # Run all unit tests.
 ```
 
-if any unit test fails with a timeout panic (see #1594) on the testing package. you can increase the KUBE_TIMEOUT value as shown below.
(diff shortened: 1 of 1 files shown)


contributors/devel/testing.md
```diff
@@ -54,6 +54,12 @@ cd kubernetes
 make test  # Run all unit tests.
 ```
 
+if any unit test fails with a timeout panic (see #1594) on the testing package. you can increase the KUBE_TIMEOUT value as shown below.
```
cblecker: "If" and "You" should be capitalized at the beginning of sentences.

cblecker: Can you add a link to the issue number?
Something like this:
```
(see [#1594](https://github.com/kubernetes/community/issues/1594))
```

cblecker: Please wrap `KUBE_TIMEOUT` in single backticks.

johnarok: Thank you for the review, I have made the changes.

PR #2048, opened 2018-04-17
Add SIG IBMCloud

spzala:
A doc page for the newly created sig-ibmcloud.


sigs.yaml
```diff
@@ -826,7 +826,37 @@ sigs:
     - name: cloud-provider-gcp
       owners:
       - https://raw.githubusercontent.com/kubernetes/cloud-provider-gcp/master/OWNERS
-
+  - name: IBMCloud
+    dir: sig-ibmcloud
+    mission_statement: >
+      A Special Interest Group (SIG) for building, deploying, maintaining, supporting,
+      and using Kubernetes on IBM Public and Private Cloud.
```
duglin: s/Cloud/Clouds/

spzala: Sure. Thanks!

spzala:
@parispittman @idvoretskyi @castrojo do I need to modify the sig-list.md and OWNERS_ALIASES? as build seems complaining about it. (something like run `make` locally..) Thanks!

cblecker:
@spzala Yes, you need to run `make` after modifying sigs.yaml, and that will generate the autogenerated docs

spzala:
@cblecker thanks!!

PR #2051, opened 2018-04-18
Add a design proposal for dynamic volume limits

gnufied:
Add a proposal for dynamic volume limits.

cc @liggitt @saad-ali @derekwaynecarr 

xref - https://github.com/kubernetes/features/issues/554

/sig storage

### DIFF
--- a/contributors/design-proposals/storage/dynamic_volume_limit.md
+++ b/contributors/design-proposals/storage/dynamic_volume_limit.md
@@ -1,0 +1,105 @@
+# Dynamic attached volume limits
+
+## Goals
+
+Currently the number of volumes attachable to a node is either hard-coded or only configurable via an environment variable. Also
+existing limits only apply to well known volume types like EBS, GCE and is not available to all volume plugins.
+
+This proposal enables any volume plugin to specify those limits and also allows same volume type to have different volume
+limits depending on type of node.
+
+## Implementation Design
+
+### Prerequisite
+
+* This feature will be protected by an alpha feature gate, so as API and CLI changes needed for it. We are planning to call
+  the feature `AttachVolumeLimit`.
+
+### API Changes
+
+There is no API change needed for this feature. However existing `node.Status.Capacity` and `node.Status.Allocatable` will
+be extended to cover volume limits available on the node too.
+
+The key name that will store volume will be start with prefix `storage-attach-limits-`. The volume limit key will respect
+format restrictions applied to Kubernetes Resource names. Volume limit key for existing plugins might look like:
+
+
+* `storage-attach-limits-aws-ebs`
+* `storage-attach-limits-gce-pd`
+
+`IsScalarResourceName` check will be extended to cover storage limits:
+
+```go
+func IsStorageAttachLimit(name v1.ResourceName) bool {
+    return strings.HasPrefix(string(name), v1.ResourceStoragePrefix)
+}
+
+// Extended and Hugepages resources
+func IsScalarResourceName(name v1.ResourceName) bool {
+    return IsExtendedResourceName(name) || IsHugePageResourceName(name) ||
+        IsPrefixedNativeResource(name) || IsStorageAttachLimit(name)
+}
+```
+
+The prefix `storage-attach-limits-*` can not be used as a resource in pods, because it does not adhere to specs defined in following function:
+
+
+```go
+func IsStandardContainerResourceName(str string) bool {
+    return standardContainerResources.Has(str) || IsHugePageResourceName(core.ResourceName(str))
+}
+```
+
+Additional validation tests will be added to make sure we don't accidentally break this.
+
+#### Alternative to using "storage-" prefix
+We also considered using currently defined `GetPluginName` interface(of Volume Plugins) for using as key in the `node.Status.Capacity`. Ultimately
+we decided against using it, because most in-tree plugins start with `kubernetes.io/` and we needed a uniform way to identify storage
+related capacity limits in `node.Status`.
+
+### Changes to scheduler
+
+Scheduler will retrieve available attachable limit on a node from `node.Status.Allocatable` and store it in `nodeInfo` cache. Volume
+limits will be treated like any other scalar resource.
+
+For `AWS-EBS`, `AzureDisk` and `GCE-PD` volume types, existing `MaxPD*` predicates will be updated to use volume attach limits available
+from node's allocatable property. To be backward compatible - the scheduler will fallback to older logic, if no limit is set in `node.Status.Allocatable` for AWS, GCE and Azure volume types.
+
+### Setting of limit for existing in-tree volume plugins
+
+The volume limit for existing volume plugins will be set by querying the volume plugin. Following function
+will be added to volume plugin interface:
+
+```go
+type VolumePluginWithAttachLimits interface {
+    // Return key name that is used for storing volume limits inside node Capacity
+    // must start with storage- prefix
+    VolumeLimitKey(spec *Spec) string
+    // Return volume limits for plugin
+    GetVolumeLimits() (map[string]int64, error)
+}
+```
+
+When querying the plugin - plugin will use `ProviderName` function of CloudProvider to check
+if plugin is usable on the node. For example - querying for `GetVolumeLimits` from `aws-ebs` plugin with `gce` cloudprovider
+will result in error.
+
+Kubelet will query the volume plugins inside `kubelet.initialNode` function and populate `node.Status` with returned values.
+
+For GCE and AWS - `GetVolumeLimits` will return limits depending on node type. Plugin already has node name accessible
+via `VolumeHost` interface and hence it will check the node type and return the volume limits.
+
+We do not aim to cover all in-tree volume types. We will support dynamic volume limits proposed here for following volume types:
+
+* GCE-PD
+* AWS-EBS
+* AzureDisk
+
+We expect to add incremental support for other volume types.
+
+### Phase 2
+
+CSI implementation will be done in phase 2. The CSI spec change that reports
+volume limits from CSI plugin has been already merged, however we still need to decide
+what is the best way Kubernetes scheduler can identify individual CSI plugin and compare it against
+limit available from node's allocatable.
--- a/contributors/design-proposals/storage/dynamic_volume_limit.md
+++ b/contributors/design-proposals/storage/dynamic_volume_limit.md
@@ -104,7 +104,7 @@ For Kubernetes 1.12, we are adding support for CSI and moving the feature to bet
 
 #### CSI support
 
-A new function of will be added to `pkg/volume/util/attach_limit.go` which will return CSI attach limit
+A new function will be added to `pkg/volume/util/attach_limit.go` which will return CSI attach limit
 resource name.
 
 The interface of function will be:
@@ -133,6 +133,18 @@ func GetCSIAttachLimitKey(driverName string) string {
 
 This function will be used both on node and scheduler for determining CSI attach limit key.The value of the
 limit will be retrieved using `GetNodeInfo` CSI RPC call and set if non-zero.
+
+**Other options**
+
+Alternately we also considered storing attach limit resource name in `CSIDriver` introduced as part
+of https://github.com/kubernetes/community/pull/2514 proposal.
+
+This will work but depends on acceptance of proposal. We can always migrate attach limit resource names to
+values defined in `CSIDriver` object in later release. If `CSIDriver` object is available and has a attach limit key,
+then kubelet could use that key otherwise it will fallback to `GetCSIAttachLimitKey`.
+
+Scheduler can also check presence of `CSIDriver` object and corresponding key in node object, otherwise it will
+fallback to using `GetCSIAttachLimitKey` function.
 
 ##### Changes to scheduler
 
--- a/contributors/design-proposals/storage/dynamic_volume_limit.md
+++ b/contributors/design-proposals/storage/dynamic_volume_limit.md
@@ -12,20 +12,21 @@ limits depending on type of node.
 
 ### Prerequisite
 
-* This feature will be protected by an alpha feature gate, so as API and CLI changes needed for it. We are planning to call
+* 1.11 This feature will be protected by an alpha feature gate, so as API and CLI changes needed for it. We are planning to call
   the feature `AttachVolumeLimit`.
+* 1.12 This feature will be behind a beta feature gate and enabled by default.
 
 ### API Changes
 
 There is no API change needed for this feature. However existing `node.Status.Capacity` and `node.Status.Allocatable` will
 be extended to cover volume limits available on the node too.
 
-The key name that will store volume will be start with prefix `storage-attach-limits-`. The volume limit key will respect
+The key name that will store volume will be start with prefix `attachable-volumes-`. The volume limit key will respect
 format restrictions applied to Kubernetes Resource names. Volume limit key for existing plugins might look like:
 
 
-* `storage-attach-limits-aws-ebs`
-* `storage-attach-limits-gce-pd`
+* `attachable-volumes-aws-ebs`
+* `attachable-volumes-gce-pd`
 
 `IsScalarResourceName` check will be extended to cover storage limits:
 
@@ -97,9 +98,48 @@ We do not aim to cover all in-tree volume types. We will support dynamic volume 
 
 We expect to add incremental support for other volume types.
 
-### Phase 2
+### Changes for Kubernetes 1.12
 
-CSI implementation will be done in phase 2. The CSI spec change that reports
-volume limits from CSI plugin has been already merged, however we still need to decide
-what is the best way Kubernetes scheduler can identify individual CSI plugin and compare it against
-limit available from node's allocatable.
+For Kubernetes 1.12, we are adding support for CSI and moving the feature to beta.
+
+#### CSI support
+
+A new function of will be added to `pkg/volume/util/attach_limit.go` which will return CSI attach limit
+resource name.
+
+The interface of function will be:
+
+```go
+const (
+    // CSI attach prefix
+    CSIAttachLimitPrefix = "attachable-volumes-csi-"
+
+    // Resource Name length
+    ResourceNameLengthLimit = 63
+)
+
+func GetCSIAttachLimitKey(driverName string) string {
+    csiPrefixLength := len(CSIAttachLimitPrefix)
+    totalkeyLength := csiPrefixLength + len(driverName)
+    if totalkeyLength >= ResourceNameLengthLimit {
+        charsFromDriverName := driverName[:23]
+        // compute SHA1 of driverName and get first 16 chars
+        return CSIAttachLimitPrefix + charsFromDriverName + hashed
+
+    }
+    return CSIAttachLimitPrefix + driverName
+}
+```
+
+This function will be used both on node and scheduler for determining CSI attach limit key.The value of the
+limit will be retrieved using `GetNodeInfo` CSI RPC call and set if non-zero.
+
+##### Changes to scheduler
+
+To support attachable limit for CSI, a new predicate called `CSIMaxVolumeLimitChecker` will be added. It will use `GetCSIAttachLimitKey`
+function defined above for extracting attach limit resource name.
+
+The predicate will be NOOP if feature gate is not enabled or when attachable limits are not available from node object.
+
+Handling delayed binding is out of scope for this proposal and will be fixed in delayed binding and topology aware dynamic
+provisioning.


gnufied:
cc @childsb

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,67 @@
...
+
+Corresponding change is being made to CSI so as CSI drivers can output
+volume limits as part of Node capabilities - https://github.com/container-storage-interface/spec/pull/221
+
+csi node-registrar will be updated to set the `node.Status` property after fetching
+volume limit from CSI driver.
+
+### Changes to scheduler
+
+When a node is being considered for scheduling of a pod with volume, it will use currently running pods against value of node.Status.AttachedVolumeLimit.
```
msau42: For in-tree volumes, how do you map the volume type with the node.status.attachedvolumelimit key?  Is it just a hardcoded mapping?

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,67 @@
...
+The volume limit for existing volume plugins will be set by querying the cloudprovider. Following interface
+will be added to `pkg/cloudprovider/cloud.go`:
+
+```go
+type VolumeLimits interface {
+    GetVolumeLimits(ctx context.Context, nodeName types.NodeName) (map[string]int64, error)
+}
+```
+
+If given cloudprovider implements this interface - kubelet will query the interface and set(via merge) the returned values in
```
msau42: I believe there's a push towards moving cloud provider dependencies out of kubelet.

Could this be something the external cloud controller calls and updates for each node?

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: yeah - all volume types eventually resolve to something like - https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/aws_ebs/aws_ebs.go#L54 and hence for in-tree volume plugins this will be a very small change in scheduler. The key in `attachedVolumeLimit` will be same as what is being returned in `GetPluginName` for in-tree plugins.

msau42: Right now the scheduler does not plumb in the volumeplugin interface

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,67 @@
...
+## Implementation Design
+
+### Prerequisite
+
+* This feature will be protected by an alpha feature gate, so as API and CLI changes needed for it. We are planning to call
+  the feature `DynamicVolumeThreshold`.
+
+### API Changes
+
+We propose following changes to node's status field:
```
childsb: Shouldn't this go under node.capacity with the other resource constraints?

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,67 @@
...
+```
+node.Status.AttachedVolumeLimits map[string]int{}
+node.Status.AttachedVolumeLimit: {
+  "kubernetes.io/aws-ebs": 39
+}
+```
+
+For CSI the name used in key will be - `kubernetes.io/csi/<driver_name>`
+
+### Changes to CSI
```
childsb: Remove CSI call outs for now.

childsb: The design should move the 'default values' from the volume plugin to an admin modifiable location (such as CLI args).   The admin should set them per node as part of configuration.

We should not conflate CSI spec with this design.  Once we have the design correct for kubernetes, then we look at changes to CSI.

gnufied:
We had a sig-storage call today and following items were discussed:

1. We will need to perform dynamic look up of plugin type in scheduler. That means linking volume packages into scheduler. @gnufied is going to request some scheduler folks to review this proposal. cc @aveshagarwal 
2. We might end up pushing CSI out of 1.11 release for this proposal - if CSI spec proposal isn't accepted in time.  cc @saad-ali 
3. There is an effort underway to move cloudproviders out of tree and hence making a change to cloud interface(`GetVolumeLimits`) could be problematic. We need to document how external cloud providers can set this limit. For in-tree cloudproviders, @msau42 is going to talk to GCE folks about propose interface.  In case - we can't reach a consensus, we will drop proposed change to cloud provider interface and stick to existing defaults for GCE/AWS and just make necessary change so as it is possible to k8s admin to set newer limits via CLI.
4. @gnufied is going to update the proposal with flexvolume bits.

saad-ali:
I would support modifying the CloudProvider interface to add support for this.

> In case - we can't reach a consensus, we will drop proposed change to cloud provider interface and stick to existing defaults for GCE/AWS and just make necessary change so as it is possible for k8s admin to set newer limits via CLI.

Can you clarify what this means? Does it mean modifying the in-tree plugins without modifying the cloud provider to figure out what the limit should be (the plugin could maintain an internal map of limits, but the challenge I see with that is figuring what type of machine it is running on without relying on cloud provider code).

liggitt:
Adding to the interface we're trying to remove doesn't seem like a great approach. Making new features work from the outside in will drive actually figuring out how to configure the kubelet in various cloud envs, instead of just leaning on the existing in-tree interface

childsb:
To add to @liggitt and summarize a design discussion we had last week....   The cloud provider could just as easily set the attach limits dynamically by updating the Node object or some other mechanism.   

I feel the best first step design for this feature is creating knobs to set attach limits by the volume type, but not try and populate or determine the limit 'dynamically'.  The 'environment' (cloud provider or otherwise) can set the values as it determines, but the storage layer not create new API or tighter coupling with the cloud provider.  The onus would be on each cloud provider (or in the absence, some startup script) to set appropriate values based on the environment.

contributors/design-proposals/storage/dynamic_volume_limit.md
thockin: +1

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,67 @@
...
+We propose following changes to node's status field:
+
+```
+node.Status.AttachedVolumeLimits map[string]int{}
+node.Status.AttachedVolumeLimit: {
+  "kubernetes.io/aws-ebs": 39
+}
+```
+
+For CSI the name used in key will be - `kubernetes.io/csi/<driver_name>`
```
thockin: This is a format unlike any other in the system.  I don't think this is correct use of resource names.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Changes to scheduler
+
+When a node is being considered for scheduling of a pod with volume, it will use currently running pods against value of node.Status.AttachedVolumeLimit.
+
+To be backward compatible - the scheduler will fallback to older logic, if `AttachedVolumeLimit` is not set.  This is to make sure that - a updated scheduler can still work with older nodes.
+
+### Setting of limit for existing in-tree volume plugins
+
+The volume limit for existing volume plugins will be set by querying the cloudprovider. Following interface
```
thockin: I don't think this belongs in cloud provider - it's a facet of the storage driver.

saad-ali: +1

contributors/design-proposals/storage/dynamic_volume_limit.md
saad-ali: Drawback is that you have to request a resource type. But, according to @thockin, device plugin people are pushing a model where admission controller detects what type of volume you are requesting and it decorates the correct type of resource request. Agreed that we should try to repurpose this design.

msau42:
We had a discussion with @saad-ali @thockin and @cheftako and concluded that:
* Kubelet command line option to report plugin limits is not a great user experience.  The user could get it wrong, and it would be a pain to do on large clusters with varying node sizes.
* It should be up to the volume plugin, not cloud provider to report its limits
* CSI can support by adding an API for the volume plugin to report limits
* However, we cannot make in-tree volume plugins that rely on a cloud-provider actually call out to the cloud provider library.  As a temporary workaround until in-tree plugins are routed through CSI, the plugin can directly encode the limit table.
* The capacity reporting API could follow the device plugin design: reuse the same capacity fields, and have a plugin-specific admission webhook examine the Pod spec, count up the volumes of its type, and add it to the Pod capacity requests.  Then we can reuse the existing scheduler capacity code.  This also eliminates the need for the scheduler to have volume plugin callouts.

childsb:
I do not like 'discovering' this at the Volume/CSI layer...

It's impossible for the volume layer / CSI driver to have an accurate MAX that applies to most users / environments.... Kube can place conservative MAX default values but then admins will want to tweak it for performance.  Kube can use the published performant MAX as default but users may have other agreements for unpublished values or find the performant values overload the cluster.  I dont see a "set it and forget it" value for many storage system really.

As further counter-point to adding this to the volume layer, I dont know of any storage API which supports this discovery.... It's all discoverable via documentation matrix.

Is the conclusion of the discussion @msau42 that it would be better UX for the admin to 'tweak' the value through the volume layer / CSI config?

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: How do we know which plugins are available on a given node? The plugin initialization code gets initialized on all node types. For example - we initialize EBS plugin on a GCE cluster. So populating limits by querying the volume plugin will result in populating limits for all available in-tree volume types.  

Do we want to populate `node.Status.Capacity` with relatively large dict? I do not know if there is a way if we can determine if certain plugin can be initialized on certain node type.

gnufied:
@msau42 thank you for summing the discussion. 

@thockin @saad-ali I would have still preferred that CLI provides an option to override the limits. For example - even on EBS the performance of disk and the limit slightly depends on network bandwidth.  The default documented limit is 40 but if one attaches 40 volumes to a m1.medium instance type - likely the instance is not going to perform optimally.  In that case, an admin may want to override the limit to a  value that he/she knows best.  So my vote is - have some defaults but allow admin to override them. I do not buy the argument that it is clumsy or error prone. We have feature-gates flag as prior art which uses similar format and requires enabling on all nodes.

gnufied:
@msau42 Modifying `node.Status` is definitely worse. For one thing - a node restart could replace the value; for another - manually editing `node.Status.Capacity` is problematic because Status of most objects is populated by k8s. I do not think we encourage or support users directly editing `Status` field of api objects - so this will be an exception.

Lastly - how does a configuration management tool like puppet/ansible keep this value in sync? Does it needs to watch node object? does it need to poll node object and make sure value is always set? It arguably seems so much worse. :(

msau42:
I believe kubelet only does Patch on the Node object, so a restart would not cause the value to be overridden or lost.  I agree that in the case of managed instances, if the cloud provider deletes and creates a new instance, then those values would be lost.  So perhaps, for this admin override case, a place to persistently store the override is needed.  Could something like dynamic kubelet config help here?

gnufied:
> I believe kubelet only does Patch on the Node object, so a restart would not cause the value to be overridden or lost.

A node restart where node was down for sufficiently long time will cause node object to be overridden.  On AWS etc a node object is removed from api-server when it is shutdown.

msau42:
I had another quick discussion with @saad-ali.  In the long term, we can provide overrides through the CSI plugin.  In the short term, this is still an alpha feature, and the admin could disable it if they run into issues with the new limits.  I think it's better to not define a new flag/config parameter for something that we're going to replace soon.

contributors/design-proposals/storage/dynamic_volume_limit.md
saad-ali: > How do we know which plugins are available on a given node? 

Plugins register themselves with the node on startup.

> The plugin initialization code gets initialized on all node types. For example - we initialize EBS plugin on a GCE cluster. So populating limits by querying the volume plugin will result in populating limits for all available in-tree volume types.

In-tree plugin can have an additional check that verifies that it is running in cloud provider.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: We had a discussion today in sig-node call and there were valid concerns raised against using `node.Capacity` for scheduling. The main problem being - there is no corresponding pod level field. memory, cpu and GPU devices are all specified at container level but volume is counted at pod level.

At this point we either heard some support for using `node.Status.AttachedVolumeLimits` or no strong opinion against. cc @derekwaynecarr 

If we choose to use node and pod capacity then we will have to create a new field in pod object spec and there was concern that storage might be only sub system using that field. We expect more comments as follow up.

gnufied:
@msau42 @saad-ali @childsb PTAL. I pushed following changes to the proposal:

1. Query volume plugin rather than Cloudprovider for limits.
2. Drop the CLI options to kubelet.
3. Small change in naming convention for CSI to make in-line with flexvolume.

contributors/design-proposals/storage/dynamic_volume_limit.md
derekwaynecarr: additional clarification:

"pods" never appears in a container resource requirement

if it did, it would be a problem.

gnufied:
@derekwaynecarr @saad-ali @aveshagarwal @ravisantoshgudimetla @msau42 can you PTAL. I have updated the proposal with `node.Status.Capacity` and flexed out more details about how we are going to set those limits etc.

cc @jsafrane

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,102 @@
...
+## Implementation Design
+
+### Prerequisite
+
+* This feature will be protected by an alpha feature gate, so as API and CLI changes needed for it. We are planning to call
+  the feature `DynamicVolumeThreshold`.
+
+### API Changes
+
+There is no API change needed for this feature. However existing `node.Status.Capacity` will
```
aveshagarwal: node.Status.Allocatable too, right?

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,102 @@
...
+
+csi node-registrar will be updated to set the `node.Status` property after fetching
+volume limit from CSI driver.
+
+### Changes to scheduler
+
+Scheduler will retrieve available attachable limit on a node from `node.Status.Capacity` and store it in `nodeInfo` cache. Volume
+limits will be treated like any other scalar resource.
+
+Existing `NewMaxPDVolumeCountPredicate` will be extended to calculate volumes in-use on a node by counting all scheduled pods on the node
```
aveshagarwal: it is already done, see here: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithm/predicates/predicates.go#L437

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,102 @@
...
+and then comparing it with limit available from `nodeInfo`.
```
aveshagarwal: Yes so for this part, instead of limiting based on hardcoded value, the advertised allocatable values in nodeInfo should be used.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,102 @@
...
+
+To be backward compatible - the scheduler will fallback to older logic, if no limit is set in `node.Status.Capacity` for AWS, GCE and Azure volume
```
aveshagarwal: i think it would be better to use node.Status.Allocatable more than node.Status.Capacity.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,102 @@
...
+type VolumePlugin interface {
+    GetVolumeLimits() (map[string]int64, error)
+}
+```
+
+When querying the plugin - plugin will use `ProviderName` function of CloudProvider to check
+if plugin is usable on the node. For example - querying for `GetVolumeLimits` from `aws-ebs` plugin with `gce` cloudprovider
+will result in error.
+
+Kubelet will query the volume plugins inside `kubelet.initialNode` function and populate `node.Status` with returned values.
```
aveshagarwal: I am assuming that node.Status will be populated at the time of node registration before node is ready and before pods can be scheduled on it. I am assuming the limit wont change at run time, right?

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: That one only counts gce or aws or azure volumes. We will have to count volumes of all types as part of this proposal.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: Yes limit will not change at runtime and should be invariant through out lifetime of node.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: Yes both allocatable and capacity will be populated.  I will update the design to use allocatable rather than capacity.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,106 @@
...
+To be backward compatible - the scheduler will fallback to older logic, if no limit is set in `node.Status.Allocatable` for AWS, GCE and Azure volume
+types. This is to make sure that - a updated scheduler can still work with older nodes.
+
+### Setting of limit for existing in-tree volume plugins
+
+The volume limit for existing volume plugins will be set by querying the volume plugin. Following function
+will be added to volume plugin interface:
+
+```go
+type VolumePluginWithAttachLimits interface {
```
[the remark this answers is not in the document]
gnufied: When counting "in-use" volumes on a node by iterating through pods, scheduler will have to check if any of the volumes being used by pod under consideration implements `VolumePluginWithAttachLimits` and consider that volume.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,106 @@
...
+
+csi node-registrar will be updated to set the `node.Status` property after fetching
+volume limit from CSI driver.
+
+### Changes to scheduler
+
+Scheduler will retrieve available attachable limit on a node from `node.Status.Allocatable` and store it in `nodeInfo` cache. Volume
+limits will be treated like any other scalar resource.
+
+Existing `NewMaxPDVolumeCountPredicate` will be extended to calculate volumes in-use on a node by counting all scheduled pods on the node
```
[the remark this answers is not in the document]
gnufied: yes that will happen automatically because CSI plugin will implement `VolumePluginWithAttachLimits`

contributors/design-proposals/storage/dynamic_volume_limit.md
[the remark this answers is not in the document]
msau42: We should explicitly state that here because the scheduler today does not call volume plugin interfaces.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,106 @@
...
+  the feature `DynamicVolumeThreshold`.
+
+### API Changes
+
+There is no API change needed for this feature. However existing `node.Status.Capacity` and `node.Status.Allocatable` will
+be extended to cover volume limits available on the node too.
+
+The key name that will store volume will be start with prefix `storage-`. So for example:
+
+* `storage-aws-ebs`
```
msau42: Are there any restrictions on allowed characters?  Does this follow the existing resource naming convention?

contributors/design-proposals/storage/dynamic_volume_limit.md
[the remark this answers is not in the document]
gnufied: fixed

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: Yes it does. I will update the proposal:

```
// Resource names must be not more than 63 characters, consisting of upper- or lower-case alphanumeric characters,
// with the -, _, and . characters allowed anywhere, except the first or last character.
// The default convention, matching that for annotations, is to use lower-case names, with dashes, rather than
// camel case, separating compound words.
```

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Setting of limit for existing in-tree volume plugins
+
+The volume limit for existing volume plugins will be set by querying the volume plugin. Following function
+will be added to volume plugin interface:
+
+```go
+type VolumePluginWithAttachLimits interface {
+    // Return key name that is used for storing volume limits inside node Capacity
+    // must start with storage- prefix
```
msau42: Should the plugin prefix the "storage-" part, or should the system prefix it?

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,108 @@
...
+
+The key name that will store volume will be start with prefix `storage-limits-`. The volume limit key will respect
+format restrictions applied to Kubernetes Resource names. Volume limit key for existing plugins might look like:
+
+
+* `storage-limits-aws-ebs`
+* `storage-limits-gce-pd`
+* `storage-limits-csi-<driver_name>`
+
+`IsScalarResourceName` check will be extended to cover storage limits:
```
msau42: Does this function ensure that users cannot request this resource normally in their Pod?

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: It can go either way. For something like CSI the prefix will be added by the system. For in-tree currently the whole volumeLimitKey is supplied by the plugin.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: This check in itself will not prevent users from requesting this inside pods. But since name does not adhere to StandardResourceName conventions, so users can not request it. I will update validations to be double sure that users can not request this resource.

gnufied: Added a note about, how are are making sure that this resource name can not be used in pods.

jingxu97: How about NumberofAttachedVolumesLimit to make it more obvious?

gnufied: The `IsScalarResourceName` function is already used for checking whether a resource can be added to node's Capacity and Allocatable. We just have to extend the check to permit new volume limit values. We can't define new function without making bigger change in scheduler.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,119 @@
+# Dynamic attached volume limits
+
+## Goals
+
+Currently the number of volumes attachable to a node is either hard-coded or only configurable via an environment variable. Also
```
krmayankk: can you list down the those methods here for reference, i didnt know the limit existed somewhere

contributors/design-proposals/storage/dynamic_volume_limit.md
krmayankk: i agree `storage-limits` is very generic, if this mainly limiting the number of volumes attached per node, we should call it so

krmayankk: my point was what kind of storage limits is `storage-limits-aws-ebs` , is it size of volume that can be created, number of volumes attached to a node, or something else ?

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,119 @@
...
+
+
+* `storage-limits-aws-ebs`
+* `storage-limits-gce-pd`
+* `storage-limits-csi-<driver_name>`
+
+`IsScalarResourceName` check will be extended to cover storage limits:
+
+```go
+func IsStorageResourceName(name v1.ResourceName) bool {
```
bsalamat: `StorageResource` is very generic and could mean pretty much anything related to storage, like number of bytes, number of devices, etc. Why not calling it `AttachableVolumes` or `VolumeLimit` or something like that?

liggitt: I agree with being specific (both in the helper method name and in the actual resource prefix in the proposed API)

contributors/design-proposals/storage/dynamic_volume_limit.md
liggitt: > Does this function ensure that users cannot request this resource normally in their Pod?

I had the same question. Are we already using capacity/allocatable resources reported by nodes to represent resources not requested/consumed by pods?

> since name does not adhere to StandardResourceName conventions, so users can not request it. I will update validations to be double sure that users can not request this resource.

we have to be careful not to box ourselves in here... if I'm reading this correctly, this PR is adding a new use of allocatable resources reported by nodes:
* known resources pods can request (cpu, memory, hugepages, etc)
* arbitrary resources pods can request (device drivers, etc)
* known resources pods cannot request (added by this PR?)

if using allocatable resources for scheduling/tracking things other than pods makes sense, does it seem likely we'll end up wanting to use arbitrary resources in that area as well? (thinking about other finite resources CSI impls would want to report on the node side and have affect PV scheduling). If so, how would those be namespaced/partitioned?

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: Agreed. Will rename this function and also I think I will rename the actual key being used to `storage-attach-limits-` to make it more clear.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: Discussed this offline a bit and it is true that apart from this resource, there is no other resource that is available on the node but not directly requestable by the pod and it will be nice to segment resources that are available on the node but not requestable by the pods directly. 

However I think that, we can revisit that decision once we know more. cc @derekwaynecarr

contributors/design-proposals/storage/dynamic_volume_limit.md
bsalamat: Thanks, @gnufied. I think for the key `storage-volume-limits-` is better.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,119 @@
...
+There is no API change needed for this feature. However existing `node.Status.Capacity` and `node.Status.Allocatable` will
+be extended to cover volume limits available on the node too.
+
+The key name that will store volume will be start with prefix `storage-limits-`. The volume limit key will respect
+format restrictions applied to Kubernetes Resource names. Volume limit key for existing plugins might look like:
+
+
+* `storage-limits-aws-ebs`
+* `storage-limits-gce-pd`
+* `storage-limits-csi-<driver_name>`
```
liggitt: what is the max length of a CSI driver name? does that name, prefixed with `storage-limits-csi-` (or whatever prefix is chosen) fit within the resourcename length restrictions?

contributors/design-proposals/storage/dynamic_volume_limit.md
liggitt: > For something like CSI the prefix will be added by the system

that seems problematic for a couple reasons... first, we want to transition in-tree volume plugins to CSI plugins. that implies the resource name consumed by an attached volume should be able to be coherent between use of that volume type via in-tree volume sources and CSI volume sources (e.g. if I have 10 AWSElasticBlockStore volumes and 10 CSI volumes backed by AWS EBS volumes, they should both map to the same `storage-volume-limit-aws-ebs` resource)

second, you'll run into resource name length limits if you simply prefix `storage-csi-` to the CSI driver name (CSI drivers can be 63 chars long, just like resource names)

contributors/design-proposals/storage/dynamic_volume_limit.md
liggitt: hard-coding the attach resource name for in-tree volume types seems ok short-term, but for CSI volumes, I expected to see a representation in the PV CSIPersistentVolumeSource of what resource consumption was associated with the PV, among which could be the volume count resource.

that lets the scheduler avoid needing to call the volume plugin manager:
* in-tree volume types have hard-coded `f(pod) -> resources` and `f(pv) -> resources` transforms
* CSI volumes have `f(pv) -> resources` represented in the CSIPersistentVolumeSource API

then the scheduler can use normal resource fit calculations to see if a given pod and its PVs fit on a given node

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: The PV specs of csi ebs and in-tree EBS volumes are not interchangeable for various other reasons. In fact, we just went through a long discussion at sig f2f about how to handle this transition. I would not necessarily expect volume limit key to be same for in-tree EBS plugin and CSI plugin. 

I was thinking to workaround 63 char limit, we can take first 13 chars from sha1 key of plugin name.

contributors/design-proposals/storage/dynamic_volume_limit.md
msau42: Max pod per node limit is a node capacity resource that cannot be explicitly requested by pods

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: fixed.

gnufied:
@bsalamat updated the proposal to be explicit about volume plugin manager requirement inside predicate. Also renamed feature gate and resource name prefix that represents volume limits.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: I have tried addressing it here - https://github.com/kubernetes/kubernetes/pull/64154#discussion_r190955132

gnufied:
We discussed this offline. @bsalamat @liggitt @msau42 @saad-ali @jingxu97 were present and we agreed that, we are going to take out CSI for this release to reconsider how CSI plugin reports its resource requirements and how resource limits will be handled when two CSI drivers are running on a node that use same underlying volume type.

We are going ahead with implementation for AWS, GCE and Azure. We also agreed that, volume plugin manager interface will be dropped from scheduler and volume type checking will be done statically.

liggitt:
https://github.com/kubernetes/community/pull/2168 proposes delaying dynamic PV provisioning until the scheduler has chosen a node

How does that interact with the "attach count" resource computation being proposed here? This proposal requires the scheduler have PVs available so it can know the volume type, and account for attach limits on nodes when selecting a node.

If PV creation is delayed until after the scheduler selects a node as proposed in #2168, the scheduler won't have attach counts available to inform it if the node is at its limit for attached volumes of that type.

Do we need to enumerate consumed resources on the storage class object for the scheduler to use in its calculations for unprovisioned PVCs?

@msau42 ^

gnufied:
As such this proposal does not changes how we count in-use volumes. A pod that has any of its PVCs in unbound state can not be scheduled, so this predicate will be evaluated again until that condition is satisfied.

Currently we are over counting for unbound PVCs or deleted PVCs. My first implementation PR for this one actually dropped over counting (https://github.com/kubernetes/kubernetes/pull/64154#discussion_r190763038) because strictly speaking it is no longer necessary. After making pvc/pv protection features default - in usual course of action, it should no longer be possible to delete in-use pvc or bound PVs.

But I think fixing over counting mechanism can be done as separate PR.

liggitt:
> A pod that has any of its PVCs in unbound state can not be scheduled

how does that square with https://github.com/kubernetes/community/pull/2168?

> Currently we are over counting for unbound PVCs or deleted PVCs

How do you know what volume type to count that as without the PV available?

gnufied:
> How do you know what volume type to count that as without the PV available?

If for some reason we can not find PV from PVC, we assume that it is of type that the predicate has been hardcoded to evaluate with. I do not like this approach and I think this entire overcounting mechanism can be dropped, now that we have pv/pvc protection features.

liggitt:
> If for some reason we can not find PV from PVC, we assume that it is of type that the predicate has been hardcoded to evaluate with.

I'm not familiar with the predicate hard-coding today... does that mean the scheduler is configured different when running on AWS than on GCE? Is it a reasonable assumption that any arbitrary PVC resolves to a single PV type in a given cluster?

gnufied:
There are 3 different predicates for Aws, gce and azure. They all
statically match volume type based on source found in pv. If we can't find
pv for a pvc, then they will all over count the volume, assuming volume was
of their type.

gnufied:
@saad-ali @msau42 @liggitt I have pushed an update to the proposal that covers CSI and moves the feature to beta in 1.12. It is pretty minimal change - PTAL.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,135 @@
...
+
+    // Resource Name length
+    ResourceNameLengthLimit = 63
+)
+
+func GetCSIAttachLimitKey(driverName string) string {
+    csiPrefixLength := len(CSIAttachLimitPrefix)
+    totalkeyLength := csiPrefixLength + len(driverName)
+    if totalkeyLength >= ResourceNameLengthLimit {
+        // compute SHA1 of driverName and get first 36 chars
```
jsafrane: I think that keeping some prefix with the original name would help users deciphering what the key is, e.g. `reallyreallylongcsidrivername` -> `reallyreallylongcs-19e793491b1c6`.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -12,20 +12,21 @@ limits depending on type of node.
...
 be extended to cover volume limits available on the node too.
 
-The key name that will store volume will be start with prefix `storage-attach-limits-`. The volume limit key will respect
+The key name that will store volume will be start with prefix `attachable-volumes-`. The volume limit key will respect
 format restrictions applied to Kubernetes Resource names. Volume limit key for existing plugins might look like:
 
 
-* `storage-attach-limits-aws-ebs`
-* `storage-attach-limits-gce-pd`
+* `attachable-volumes-aws-ebs`
```
msau42: Given that we also want to add a max capacity limit, does "attachable-volumes" still make sense, or do we want to clarify that it is a count?

gnufied: I was thinking to use separate resource name for max attachable capacity. Something like - `attachable-capacity-xxx` and keep `attachable-volumes-xx` for representing count. I think capacity vs volumes word is "good enough" indication.

gnufied:
@msau42 @jsafrane fixed and updated the proposal. Also added a section about using https://github.com/kubernetes/community/pull/2514 as a possibility but we can always migrate to it, when it becomes available.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,157 @@
...
+
+This function will be used both on node and scheduler for determining CSI attach limit key.The value of the
+limit will be retrieved using `GetNodeInfo` CSI RPC call and set if non-zero.
+
+**Other options**
+
+Alternately we also considered storing attach limit resource name in `CSIDriver` introduced as part
+of https://github.com/kubernetes/community/pull/2514 proposal.
+
+This will work but depends on acceptance of proposal. We can always migrate attach limit resource names to
```
msau42: CSI doesn't have a way to report some alternative key though right?

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: fixed.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: i think there is a proposal to allow CSI drivers to register themselves and not use auto registration mechanism. In which case, it will be possible for a plugin to report alternative key as a part of registration process.

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,157 @@
...
+        // compute SHA1 of driverName and get first 16 chars
+        return CSIAttachLimitPrefix + charsFromDriverName + hashed
+
+    }
+    return CSIAttachLimitPrefix + driverName
+}
+```
+
+This function will be used both on node and scheduler for determining CSI attach limit key.The value of the
+limit will be retrieved using `GetNodeInfo` CSI RPC call and set if non-zero.
```
msau42: Are we going to handle the case of override and changing the limit on the node after the initial configuration?

contributors/design-proposals/storage/dynamic_volume_limit.md
```diff
@@ -0,0 +1,157 @@
...
+### API Changes
+
+There is no API change needed for this feature. However existing `node.Status.Capacity` and `node.Status.Allocatable` will
+be extended to cover volume limits available on the node too.
+
+The key name that will store volume will be start with prefix `attachable-volumes-`. The volume limit key will respect
+format restrictions applied to Kubernetes Resource names. Volume limit key for existing plugins might look like:
+
+
+* `attachable-volumes-aws-ebs`
```
thockin: I really dislike this sort of ad hoc syntax embedded in a string, and aside from that, I don't buy that this should be magical like this.  Why can't we do something less stringy?

Just thinking out loud...

e.g. Some volume have per-node attachment limits.  For those volumes, you install an admission controller that examines the volumes in play and decorates the pod as requiring certain resources, and then the scheduler simply schedules?

I feel like we have explored this, but if so, I don't recall why it was rejected.

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: No not in this release. We will tackle this problem in 1.13

contributors/design-proposals/storage/dynamic_volume_limit.md
gnufied: There are two part of it right?

1. The plugin must be able to specify a maximum attachment limit on per node basis. We chose to use node's allocatable/capacity field for that. And when we want to express this limit in node's allocatable regardless of how we do counting via admission controller or something else - the limit must have a key and a value. The key MUST be a string that adheres to resourceName naming conventions(allocatable is of type `map[v1.ResourceName]resource.Quantity`).  So isn't the `attachable-volumes-aws-ebs` prefix chosen for expressing attach limits in node's allocatable/capacity orthogonal to how we do counting?

2. We could still use a admission controller to do actual counting but I think it was agreed that, scheduler is better(or may be easier) place for counting actual volumes in use. Because:
a. It can consider all pods on the node. Not just the one being scheduled.
b. I think the way volume limits interacts with late binding and even distribution of pods with volume workloads(https://github.com/kubernetes/kubernetes/pull/60525) makes it better place to do in scheduler. 

And obviously a admission plugin may be disabled or not enabled(but same goes for a scheduler predicate I guess). I am not particularly opposed to using admission controller for counting but I think we did not see benefits we thought we will see.

thockin: I am going to reformat for comments:

`We chose to use node's allocatable/capacity field for that [to specify a maximum attachment limit]` ... `The key MUST be a string that adheres to resourceName naming conventions` ... `So isn't the attachable-volumes-aws-ebs prefix chosen for expressing attach limits in node's allocatable/capacity orthogonal to how we do counting`

Yes it is orthogonal.  Sorry for putting them in one message.

My first grump here is that we have invented non-obvious syntax without any specification for it.  I detest string-embedded syntaxes (stringly type APIs) and even if we DO go that way, we can't tolerate under-specified ones.  Where do these magical names come from?  How do they not conflict?  Jordan had some comments on that, too.  At the very very least this needs a strong spec, and perhaps we actually need something different.

Regarding scheduler, what I really don't want to do is build up a body of special -cases that we dump on the scheduler, when the problem is actually a pretty general problem.  How do we opaquely allocate and schedule resources WITHOUT changing the scheduler?   We may need to make other changes to the system to accomodate that, but I think it should be a preferred choice.  @bsalamat am I over-blowing this?

Late-binding is from a PVC to an PV via a class.  Presumably a single class can't use multiple drivers, so it's still feasible to reify the PVC -> class -> driver somewhere?  The fact that classes could, in theory change, is a problem that maybe we should think harder about and maybe we need a different kind of plugin (e.g. maybe we need explicit scheduler pre-processor plugins). "Let's just hack the scheduler' seems like a very slippery slope to me.

And, BTW, I *want* this capability (and have for years), so please don't think I am poo-pooing the idea :)

bsalamat: I second what @thockin said about dumping a lot of special case logic in the scheduler. That said, not everything can be checked at admission time. For example, dynamic volume binding happens after scheduler finds a node to run a pod. I guess there is no choice other than checking the number of attached volumes to the node at that point.

gnufied: > Where do these magical names come from? How do they not conflict? Jordan had some comments on that, too. At the very very least this needs a strong spec, and perhaps we actually need something different.

In current iteration any resource name(in node's allocatable) that uses `attachable-volumes-` prefix refers to attachable limits. We have added some tests to ensure that containers themselves aren't able to directly request this resource.  It is however possible that someone can set a new `attachable-volumes-xxx` value that does not apply to volume limits(via `patch` ).

I am not entirely sure about some external component inadvertently overriding any of existing limits though. Because - when kubelet starts, it sets those limits and it periodically syncs the value from same source. So if an external component does override those values it will be wiped out. 

I still see the need of speccing them better and perhaps there is a better way. I need to think and will bring this up on sig-storage.

msau42: Previous discussion [here](https://github.com/kubernetes/community/pull/2051#discussion_r183839230)

gnufied: Just had a discussion with @jsafrane as well about this and we discussed two problematic things about moving the count outside the scheduler:

1. For local volumes with delayed binding, storageClass could be dummy or unknown and hence counting unbound PVCs is going to be tricky outside the scheduler.
2. Currently the scheduler counts unique volumes. If we specify volumes as a pod level countable resource and scheduler counts volumes the way it counts memory or other resources, it can't accurately count unique volumes.

thockin:
Imagine we had a generic resource-finalizer plugin, at the beginning of the
scheduler.  It's job is to take a pod and return a resources block.  That
resources set is the sum of all of the pod's containers (or whatever the
policy is) plus anything the plugin wants to do to mutate it.  In this
case, add a volume-specific counted resource.  We don't need to change the
pod API per se.  We would need to change the scheduler, in a pretty big
way, but this would accomodate future situations, too.

As for unique volumes, this only impacts volumes that are ROX and used by
multiple pods.  I know some plugins allow shared mounting of RWO volumes,
but all of those drivers are *broken*.  Given that, this seems like a
pretty niche bug, and one I could probably live with.

contributors/design-proposals/storage/dynamic_volume_limit.md
thockin: > It is however possible that someone can set a new attachable-volumes-xxx value that does not apply to volume limits(via patch ).

What is the specification for xxx here?  Is it the CSI driver name?  Some other source?  Does it get some auto-mutation?  Are characters changed?

PR #2052, opened 2018-04-18
KEP: Security Profile Design Proposal

easeway:
This is the detailed design proposal for Security Profile.

Security Profile helps managing a secured Kubernetes cluster by solving problems of:

- Management complexity
  - Complicated configuration of control plane components' security-relevant command line flags, correctness of cert/key/config file permissions;
  - Requires deep understanding of Kubernetes policy objects like RBAC, PodSecurityPolicy, NetworkPolicy, ResourceQuota etc;
  - No support for automatic provisioning of namespaced policy objects upon namespace creation;
  - Requires of domain knowledge about security;
  - Requires keeping up with fast evolving Kubernetes security/isolation features, and update configurations accordingly;

- Extra efforts building in-house solutions
  - No standard/common ways to solve the problems;
  - It’s very difficult to make in-house solutions sharable/reusable, as a result, efforts are duplicated solving same/similar problems.

by defining curated settings of:

- Cluster bootstrapping configuration, like command line flags, file permissions, etc
- Cluster-scoped policy objects, like PodSecurityPolicy
- Namespace-scoped policy objects, like NetworkPolicy

And simplify the cluster management by specifying the name of security profile.

_This proposal is looking for sponsorship from sig-auth,
to create a public repo as [https://github.com/kubernetes-sigs/security-profiles](),
that community can help curate commonly used security profiles.
Details are described in Release section in the proposal._

/sig auth

/cc @davidopp 
/cc @tallclair 
/cc @ericchiang 
/cc @liggitt

### DIFF
--- a/keps/draft-20180418-security-profile.md
+++ b/keps/draft-20180418-security-profile.md
@@ -1,0 +1,1440 @@
+---
+kep-number: draft-20180418
+title: Security Profile
+authors:
+  - "@easeway"
+owning-sig: sig-auth
+status: provisional
+reviewers:
+  - "@tallclair"
+  - "@ericchiang"
+  - "@liggitt"
+  - "@davidopp"
+  - TBD
+approvers:
+  - "@tallclair"
+  - "@ericchiang"
+  - "@liggitt"
+editor:
+  - "@easeway"
+creation-date: 2018-04-18
+---
+# Security Profile
+
+Author: @easeway (yisuihu@google.com)
+
+_This proposal is looking for sponsorship from sig-auth,
+to create a public repo as [https://github.com/kubernetes-sigs/security-profiles](),
+that community can help curate commonly used security profiles.
+Details are described in [Release](#release) section below._
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+  * [Goals](#goals)
+  * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+  * [Proposed Security Profiles](#proposed-security-profiles)
+    * [default](#default)
+    * [saas-multitenancy](#saas-multitenancy)
+    * [more](#more)
+* [User Experience](#user-experience)
+  * [Use Cases](#use-cases)
+* [Implementation](#implementation)
+  * [The Security Profile](#the-security-profile)
+    * [Bootstrapping Rules](#bootstrapping-rules)
+    * [Runtime Rules](#runtime-rules)
+    * [Effectiveness](#effectiveness)
+    * [Switch Security Profiles](#switch-security-profiles)
+    * [Enforcement Status](#enforcement-status)
+    * [Cluster-scope Policy Objects](#cluster-scope-policy-objects)
+    * [Namespace Population](#namespace-population)
+  * [Customize and Extend](#customize-and-extend)
+  * [Versioning Upgrade and Deprecation](#versioning-upgrade-and-deprecation)
+    * [Release](#release)
+    * [Version Schema](#version-schema)
+    * [Upgrade](#upgrade)
+    * [Deprecation](#deprecation)
+  * [Work with Config-as-Code Systems](#work-with-config-as-code-systems)
+* [References](#references)
+* [Appendix](#Appendix)
+  * [Predefined Rules](#predefined-rules)
+    * [Bootstrapping Rule Definitions](#bootstrapping-rule-definitions)
+      * [Addons](#addons)
+      * [Command line flags](#command-line-flags)
+      * [Admission Controls](#admission-controls)
+      * [Authorization Mode](#authorization-mode)
+      * [Feature Gates](#feature-gates)
+      * [Runtime Configs](#runtime-configs)
+      * [File Permissions](#file-permissions)
+    * [Runtime Rule Definitions](#runtime-rule-definitions)
+      * [Namespace-scope Policy Objects](#namespace-scope-policy-objects)
+      * [Node Taints](#node-taints)
+  * [Proposed Security Profile Examples](#proposed-security-profile-examples)
+    * [The default profile](#the-default-profile)
+    * [The saas-multitenancy profile](#the-saas-multitenancy-profile)
+
+## Summary
+
+SecurityProfile is a way of grouping a set of k8s control plane component flag settings,
+templates for cluster- and namespace-scope policy objects, into a named, versioned bundle.
+When a cluster admin specifies a security profile at cluster creation time,
+they can be assured that the cluster will provide the security and isolation properties they need,
+without them having to understand the low-level configuration used to achieve it.
+Security Profile provides optional, additive functionality on top of core Kubernetes.
+
+## Motivation
+
+Configuring Kubernetes securely is difficult, because of:
+
+- Management complexity
+  - Complicated configuration of control plane components' security-relevant command line flags, correctness of cert/key/config file permissions;
+  - Requires deep understanding of Kubernetes policy objects like RBAC, PodSecurityPolicy, NetworkPolicy, ResourceQuota etc;
+  - No support for automatic provisioning of namespaced policy objects upon namespace creation;
+  - Requires of domain knowledge about security;
+  - Requires keeping up with fast evolving Kubernetes security/isolation features, and update configurations accordingly;
+- Extra efforts building in-house solutions
+  - No standard/common ways to solve the problems;
+  - It’s very difficult to make in-house solutions sharable/reusable, as a result, efforts are duplicated solving same/similar problems.
+
+### Goals
+
+This design presents the proposal of Security Profile for Kubernetes and demonstrates
+
+- The particular problems the proposal is going to solve
+- The impact on user experience using Kubernetes
+- The details how the proposal solves the problem
+
+### Non-Goals
+
+The following are not covered in this proposal
+
+- Policy objects defined for a particular Kubernetes cluster
+- How bootstrapping rules are enforced
+- Verification mechanisms
+- Continuously detect rule violations
+- Continuously fix rule violations
+- Conformance tests
+- The details integration with Config-as-Code systems
+
+## Proposal
+
+_Security Profile_ defines a common and portable way to describe how to
+secure a Kubernetes cluster, with content covering
+
+- Cluster bootstrapping configuration, like command line flags, file permissions, etc
+- Cluster-scoped policy objects, like PodSecurityPolicy
+- Namespace-scoped policy objects, like NetworkPolicy
+
+For example a _kick-the-tires_ security profile:
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: kickthetires-1.0-0
+spec:
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+---
+apiVersion: profile.security.k8s.io/v1alpha1
+kind: NamespaceTemplate
+metadata:
+  name: kickthetires-1.0-0
+spec:
+  namespaces:
+    selector:
+      matchExpressions:
+        - key: profile.security.k8s.io/opt-out
+          operator: DoesNotExist
+    excludes:
+    - kube-system
+  templates:
+  - apiVersion: rbac.authorizatoin.k8s.io/v1
+    kind: Role
+    metadata:
+      name: use-psp
+    rules:
+    - apiGroups: ['policy']
+      resources: ['podsecuritypolicy']
+      verbs: ['use']
+      resourceNames: ['kickthetires-1.0-0']
+    - apiVersion: rbac.authorization.k8s.io/v1
+      kind: RoleBinding
+      metadata:
+        name: use-psp-binding
+      roleRef:
+        apiGroup: rbac.authorization.k8s.io/v1
+        kind: Role
+        name: use-psp
+      subjects:
+      - kind: Group
+        apiGroup: rbac.authorization.k8s.io/v1
+        name: system:serviceaccounts
+---
+apiVersion: profile.security.k8s.io/v1alpha1
+kind: SecurityProfile
+metadata:
+  name: kickthetires-1.0-0
+spec:
+  bootstrapping-rules:
+  - name: flags
+    spec:
+      apiserver:
+        anonymous-auth:
+          expect: unset
+        basic-auth-file:
+          expect: unset
+  - name: admission-controls
+    spec:
+      excludes:
+      - AlwaysAdmit
+  - name: authorization-mode
+    spec:
+      excludes:
+      - AlwaysAllow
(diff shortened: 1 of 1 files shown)


ericchiang:
I'm concerned with the lack of scoping in this proposal. Autopopulating namespaces, config as code, bootstrapping config, conformance testing, meta-policies… there are at least three or four unique sub-projects in this document.

Before we get into the nitty gritty of the proposal, does it make sense to keep all of these concepts as one framework? Naively I could see conformance profiles and namespace templating being separate. Conformance profiles are a way to express the capabilities a cluster provides. Namespace templates are a tool for admins to quickly provision namespaces for users with sane security defaults.

easeway:
@ericchiang you are right. The unique pieces covered in the proposal are relatively independent/separate. They can be used individually. The reason putting all things together is for a really simple user experience and also the possibility to share the experience across Kubernetes clusters in different environments.

I agree it will be better to break up and focus on individual pieces. I will send out individual proposals, covering:

- Conformance profiles
- Namespace population
- Cluster bootstrapping
- Config as code

And later update this KEP as a complete solution to demonstrate the ultimate user experience we want to achieve after things are integrated.

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+Minimum security practice to be compatible with most workloads.
+Best fit for sharing the cluster with trusted workloads.
+
+Enforcement:
+
+- etcd
+  - TLS required for communication
+  - Explicitly specify `--wal-dir`, and `--max-wals=0`
+- kube-apiserver
+  - TLS required for communication
```
mayakacz: how is this enforced/ verified?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+  - Audit logging must be explicitly configured
+  - No anonymous auth, basic auth (unset `--anonymous-auth`, `--basic-auth-file`)
+  - Required authorization-mode: Node, RBAC
+  - Excluded authorization-mode: ABAC, AlwaysAllow
```
mayakacz: please add: automountServiceAccountToken: false

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+  - Explicitly set `--terminated-pod-gc-threshold`
+- kubelet
+  - Disable hostname override (unset `--hostname-override`)
+  - No privileged
+- addons
+  - No dashboard
+- file-permissions
+  - ... # correct setup of cert/key/config file permissions
+- PodSecurityPolicy
+  - No priviledged
```
mayakacz: can you specify? I'm thinking this is: privileged: false, allowPrivilegeEscalation: false, MustRunAsNonRoot

Or is this stronger than what you were thinking for 'default'?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+#### saas-multitenancy
+
+Full Example: [saas-multitenancy-1.0-0](#the-saas-multitenancy-profile)
+
+For SaaS scenarios which allows only plain workloads without the need to gain privileged access.
+
+Enforcement: in addition to `default`
+
+- PodSecurityPolicy
+  - No privileged pods
```
mayakacz: Isn't this already in 'default'?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+  - No host namespaces (network, IPC, PID)
+  - Limit volume types to
+    - configMap
+    - downwardAPI
+    - emptyDir
+    - nfs
+    - persistentVolumeClaim
+    - projected
+    - secret
+- NetworkPolicy
```
mayakacz: Why wouldn't we include this in 'default'?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+  - Block all ingress by default
+  - Block all egress by default
+
```
mayakacz: Please add secrets encryption

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+These rules can also be interpreted as a set of check conditions for tools to validate if the cluster has been setup properly.
+
+Please see [Predefined Rules](#predefined-rules) as a reference of the rules.
+
+As currently, there's no way (or limited) to inspect the host operating system from inside the cluster,
+these rules are only enforced by the deployers (or cluster lifecycle management system).
+
+#### Runtime Rules
+Runtime rules are enforced after the cluster is bootstrapped,
+by _enforcers_ (the controllers deployed with security profiles in the cluster).
```
mayakacz: Would a runtime enforcer just alert on a rule violation? Or force move to a new state? or....?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+
+_Alternative design_
+
+Runtime rules can also be defined similarly to bootstrapping rules, in a schemaless way.
+However it loses the support from Kubernetes API to perform schema validation and other functionality,
+even though it's currently very limited for CRDs, in future it may become better.
+
+#### Effectiveness
+Multiple security profiles can exist in the same cluster at the same time,
+but only one can be active (being enforced with `runtime-rules`, not `bootstrapping-rules`).
```
mayakacz: I find this confusing - which one would be enforced? (based on the SecurityProfileSelector below?) Is there a hierarchy, if they are in conflict?

Why allow multiple profiles, instead of just forcing users to adopt a different, merged profile?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+- Create/patch policy objects with updated labels;
+- Scanning and removing labeled objects of the same type but with mismatch values.
+
+#### Enforcement Status
+The controller/admission control enforcing a particular rule will be responsible to report the status of the enforcement.
+The status is defined as
+
+- `none`: no enforcers recognize this rule
+- `needed`: enforcement is needed, but not performed yet
+- `inprogress`: the enforcers are working to enforce the rule, not completed yet
```
mayakacz: does 'inprogress' this only apply for runtime enforcers?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+- Give it a new name
+- Add custom rules
+- Apply it using `kubectl apply`
+- Select it in `SecurityProfileSelector`
+
+### Versioning Upgrade and Deprecation
+
+#### Release
+Security profiles are maintained as an individual project similarly to other Kubernetes open source projects, like heapster, dashboard etc.
+It has its own release lifecycle.
```
mayakacz: What is the goal? To roughly release in line with K8s major releases, for example?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+The general flow of upgrading to a new Security Profile is performed by:
+
+- Install a new version of security profile release bundle
+- Select a new version of security profile in `SecurityProfileSelector`
+- Uninstall the old security profile bundle
+
+Uninstall is easy as resources from a bundle are labeled with bundle name.
+With multiple bundles installed, it's possible multiple admission controls, controllers are running side-by-side.
+With help of labeled bundle name, the admission controls, controllers will stop enforcement if the currently selected security profile is not from the same bundle.
+With this mechanism, it's very simple and lightweight to switch back to the previous security profile if something unexpected happens with a new security profile.
```
mayakacz: to the previous security profile version?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+
+When upgrading a cluster, it's possible the current security profile may not work on the new cluster version.
```
mayakacz: may not work, or may not be supported? (i.e., 'deprecated')

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+In this case, upgrade the security profile to a new version first, and then upgrade the cluster.
+
+#### Deprecation
+
+A security profile is deprecated when:
+
+- Any bug found in enforcement logic (admission control, controllers etc)
+- Improper/Incomplete rules in the profile
+- Other security concerns
```
mayakacz: would we ever deprecate a security profile when it no longer ad some critical security functionality, even though it was the best practice for that K8s version?

For example, would we deprecate a security profile for K8s 1.5, because it does not have the ability to enforce RBAC? Is there some minimum K8s version (or recency) we require for security profiles?

keps/draft-20180418-security-profile.md
```diff
@@ -0,0 +1,1516 @@
...
+
+Conformance tests can be developed according to curated security profiles.
+Each security profile will have a corresponding conformance test suite to cover all checking criteria.
+The test suite can sit in the same repository as the security profile and the conformance test tools
+fetch the test suite from this repo.
+
+The alternative is put test suite together with Kubernetes E2E tests in the official Kubernetes repository,
+It will be a little difficult to synchronize the development work in two repos.
+
+The details is out of scope of this doc. A separate doc will be focus on conformance tests.
```
mayakacz: please share a link to the KEP

keps/draft-20180418-security-profile.md
easeway: They are defined in bootstrapping rules, the enforcement of bootstrapping rules are implementation specific, thus out of scope in this proposal.
Verification is out of scope.
I will call out these explicitly.

keps/draft-20180418-security-profile.md
easeway: `automountServiceAccountToken` is a field in ServiceAccount or Pod spec. Currently, it's difficult to define that in security profile as it's can't be enforced by any component (though security profile can introduce a admission webhook to mutate that field, however, the intention of security profile is to leveraging existing mechanism, not introduce something new). So I think this will be a good feature in PodSecurityPolicy, then it can be defined in security profile.

keps/draft-20180418-security-profile.md
easeway: I think it's stronger than what the `default` is proposed. I'm even thinking of removing `No privileged` in this profile. The `default` profile is designed to give the maximum compatibility with existing workloads. Add restrictions to block `privileged` will definitely break some workloads. So I think I need rephrase the goal for each security profiles, and add one more between `default` and `saas-multitenancy`.

keps/draft-20180418-security-profile.md
easeway: I should remove that from `default`.

keps/draft-20180418-security-profile.md
easeway: Adding NetworkPolicy (blocking all traffic by default) will break a lot of existing workloads which doesn't assume the existence of NetworkPolicy. So I think it doesn't belong to `default`.

keps/draft-20180418-security-profile.md
easeway: I will add more details.

keps/draft-20180418-security-profile.md
easeway: to be accurate: may not be supported. (it's not 'deprecated' which means the profile may still work, but may not guarantee the security level it's promised to provide because of imperfection of rules, bugs in implementations, etc.)

keps/draft-20180418-security-profile.md
easeway: In this case, a security profile won't be deprecated for the supported K8S version. In the example mentioned, the profile will not be deprecated for K8s 1.5, however it's not supported in K8s 1.6 or above with RBAC. So for users upgrading K8s from 1.5 to 1.6 they will have to upgrade the security profile to a new version which supports both 1.5 and 1.6.

keps/draft-20180418-security-profile.md
easeway: I will add that.

keps/draft-20180418-security-profile.md
easeway: not exactly the version. The user may upgrade from `default` to `saas`, and find out breaking workloads and decided to switch back to `default`.

keps/draft-20180418-security-profile.md
easeway: I will add that in bootstrapping rules via requirement of --experimental-encryption-provider-config

keps/draft-20180418-security-profile.md
easeway: yes

keps/draft-20180418-security-profile.md
easeway: It's possible in the cluster, there are multiple security profiles, like default, saas. User can use SecurityProfileSelector to select one of them. There's no hierarchy, and no conflict. Because only one security profile is selected. It's not allowed to select multiple security profiles.

keps/draft-20180418-security-profile.md
easeway: The runtime enforcers are currently enforcing by creating policy objects in newly created namespaces. There are multiple levels of enforcement:

1. Onetime enforcement when a namespace is created
2. Continuously monitoring all namespaces and detect violations
3. Continuously fixing detected violations

The proposal focuses on 1, and keep 2 and 3 out of scope. I will call this out.

easeway:
Removed Conformance tests which is on a different layer and should be separate from security profile.

keps/draft-20180418-security-profile.md
mayakacz: ack - that makes sense

keps/draft-20180418-security-profile.md
mayakacz: so 'default' would allow privileged pods?
Obviously I'm coming at this from a security angle, but if I turn on security profiles, I'm probably OK with it warning me that I shouldn't be running pods as privileged and then fix that - or are you saying that there isn't a lower level so they're stuck with that issue?

keps/draft-20180418-security-profile.md
mayakacz: got it, agreed.

keps/draft-20180418-security-profile.md
mayakacz: Makes sense

keps/draft-20180418-security-profile.md
easeway: This is my personal opinion (I'd like to hear other thoughts, as I'm probably wrong): the `default` security profile is provided to bring in the lowest level (there's no lower levels) of security (eliminate some really bad insecure configurations) without breaking compatibility of how people are using Kubernetes today. So we can use `default` to help Kubernetes users smoothly transit from insecure to secure. I believe privileged containers are allowed in most Kubernetes clusters with trusted users (like inside the same corp or team) and sometimes useful. That's the reason I removed "no-privileged containers" from `default` profile.

mayakacz: got it, I see you defined default as "to be compatible with most workloads". That makes sense then.

justaugustus:

justaugustus:

/close

PR #2054, opened 2018-04-19
CRD versioning design proposal [No Conversion]

mbohlool:
CRD versioning proposal is discussed in detail here:

https://docs.google.com/document/d/1Ucf7JwyHpy7QlgHIN2Rst_q6yT0eeN9euzUV6kte6aY/edit

This is the main no conversion part of the design agreed on by the community/API machinery SIG.

@sttts @deads2k @lavalamp

### DIFF
--- a/contributors/design-proposals/api-machinery/customresources-versioning.md
+++ b/contributors/design-proposals/api-machinery/customresources-versioning.md
@@ -1,0 +1,100 @@
+CRD Versioning
+=============
+
+The objective of this design document is to provide a machinery for Custom Resource Definition authors to define different resource version and a conversion mechanism between them.
+
+# **Background**
+
+Custom Resource Definitions ([CRDs](https://kubernetes.io/docs/concepts/api-extension/custom-resources/)) are a popular mechanism for extending Kubernetes, due to their ease of use compared with the main alternative of building an Aggregated API Server. They are, however, lacking a very important feature that all other kubernetes objects support: Versioning. Today, each CR can only have one version and there is no clear way for authors to advance their resources to a newer version other than creating a completely new CRD and converting everything manually in sync with their client software.
+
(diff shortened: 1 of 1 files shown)


deads2k:
I think this matches what we talked about before.

@liggitt chance for final comments.  I'll probably tag on Monday.

@mbohlool there may be a little bikeshedding on Name versus Version in the substruct, but in looks well scoped.

deads2k:
> @liggitt chance for final comments. I'll probably tag on Monday.

@sttts you too.

mbohlool:
> there may be a little bikeshedding on Name versus Version in the substruct, but in looks well scoped.

'Version.Version' vs 'Version.Name'? I vote for `Version.Name` :)

liggitt:
> @liggitt chance for final comments. I'll probably tag on Monday.

no other comments from me

contributors/design-proposals/api-machinery/customresources-versioning.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+# **Validation**
+
+Basic validations needed for the `version` field are:
+
+* `Spec.Version` field exists in `Spec.Versions` field.
+* The version defined in `Spec.Version` field should point to a `Served` Version in `Spec.Versions` list except when we do not serve any version (i.e. all versions in `Spec.Versions` field are disabled by `Served` set to `False`). This is for backward compatibility. An old controller expect that version to be served but only the whole CRD is served. CRD Registration controller should unregister a CRD with no serving version.
+* None of the `Status.StoredVersion` can be removed from `Spec.Versions` list.
+* Only one of the versions in `spec.Versions` can flag as `Storage` version.
```
sttts: Exactly one...

sttts:
Lgtm.

PR #2055, opened 2018-04-19
KEP: Enhance HPA Metrics Specificity

DirectXMan12:
This KEP introduces additional tools into the autoscaling/v2 API for
working with metrics, including support for a metrics label selector,
and expanded support for the `targetAverageValue` field.

This PR also introduces a KEP folder for SIG Autoscaling.


MaciekPytel:
If we're already changing the layout of the API maybe it would be worth it to move target to a separate unified struct (similar to what you do with moving fields describing metric to MetricSpec)? With that change your example would look like that:

```yaml
- type: Object
  object:
    target:
      kind: StatefulSet 
      apiVersion: apps/v1
      name: foo
    target:
      type: Value
      value: 2
    metric:
      name: queue_length
      selector: "queue=some-jobs"
```

@thockin suggested that approach during his review of external metric API: https://github.com/kubernetes/kubernetes/pull/60096#pullrequestreview-98399617 (you need to unfold the comment). WDYT?

keps/sig-autoscaling/draft-20180419-enhance-hpa-metrics-specificity.md
```diff
@@ -0,0 +1,247 @@
...
+- The testing adapter is updated to support these changes (for e2e
+  purposes).
+
+### Non-Goals
+
+It is outside of the purview of the KEP to ensure that current custom
+metrics adapters support the new changes -- this is up to those adapters
+maintainers.
+
+## Proposal
```
MaciekPytel: I think this misses a single part - what do we do if multiple time series are matched by selector? In external metrics API we sum all the values (we assume the user may not want to drill down on some dimension captured by metric, for example use all HTTP requests, regardless of response code). My assumption is we want to do the same? We should say it explicitly though.

keps/sig-autoscaling/draft-20180419-enhance-hpa-metrics-specificity.md
```diff
@@ -0,0 +1,247 @@
...
+and `Object` sources.
+
+These changes necessitate a second beta of `autoscaling/v2`:
+`autoscaling/v2beta2`.
+
+Similarly, corresponding changes need to be made to the custom metrics
+API:
+
+```go
+type MetricValue struct {
```
MaciekPytel: In external metrics API we return a list of values, one for each time series. I think we need to do that in CM API as well or we need to say that it's adapters job to aggregate all time series into a single value.

DirectXMan12: Yeah, the intention was that it's the adapter's job to aggregate, if necessary.  That's the approach we took before the specificity, and I missed that the external metrics API doesn't do that (I'd prefer if it did).  I'll note that explicitly

keps/sig-autoscaling/draft-20180419-enhance-hpa-metrics-specificity.md
DirectXMan12: If multiple series are matched, the adapter must deal with it, similarly to the current CM API (see my other comment).

DirectXMan12:
Factored in `targetXXX` unification, added note about what to do in case of multiple series.

@kubernetes/api-reviewers I'd also like to make the utilization serialize more sanely, but I want to make sure it's acceptible.  Utilization would still be an `int` (e.g. `40`) internally, but it would serialize as a string with a percent sign (e.g. `40%`) to make it more obvious what's going on.  Utilization continues to be one of the more confusing subjects for users, and I figure this will at least help a bit.

DirectXMan12:
NB: for round-tripping into v2beta1, serialize any `MetricSpec` and `MetricStatus` objects with label selectors as annotations, while keeping those without selectors as normal items in the `metrics` list.  This is analogous to what we do for `v1` with `TargetCPUUtilizationPercentage`.

justaugustus:

DirectXMan12:
@jdumars @justaugustus this is ready to be merged (the implementation was done a bit ago), but it needs higher-level LGTM to be merged because it introduces the sig-autoscaling directory for the first time.

keps/sig-autoscaling/0032-enhance-hpa-metrics-specificity.md
```diff
@@ -0,0 +1,277 @@
...
+	Selector *metav1.LabelSelector
+}
+
+type MetricTarget struct {
+	Type MetricTargetType // Utilization, Value, AverageValue
+    // value is the raw value of the single metric (valid for object metrics)
+	Value *resource.Quantity
+
+    // averageValue is the raw value or values averaged across the number
+    // of pods targetted by the HPA (valid for all metric types).
```
justaugustus:
```suggestion
    // of pods targeted by the HPA (valid for all metric types).
```

justaugustus: @DirectXMan12 -- this is failing a spell check test right now:

```
I1127 23:07:48.573] Verifying verify-spelling.sh
W1127 23:07:55.797] keps/sig-autoscaling/0032-enhance-hpa-metrics-specificity.md:150:15: "targetted" is a misspelling of "targeted"
I1127 23:07:55.898] FAILED   verify-spelling.sh	7s
```

justaugustus:
KEP structure:
/lgtm

PR #2056, opened 2018-04-19
Adding chair for SIG-Windows

PatrickLang:
Last SIG-Windows meeting @michmike nominated me to co-chair SIG-Windows with approval from the SIG members, which I'm happy to do :) We need to update sigs.yaml and generated files to make it official.

/assign michmike


cblecker:
I didn't see anything in the mailing list/minutes about this, but it looks like the last meeting recording has it discussed right at the start of the call (https://youtu.be/_oCgKOqYByw?t=27s).

@PatrickLang @michmike It might be worth making sure the mailing list and/or the minutes are updated with this information.

/ok-to-test
/approve
LGTM to @michmike.

PR #2064, opened 2018-04-22
Update device manager doc to match 1.10 device plugin API changes.

RenaudWasTaken:
Hello!

/cc @vikaschoudhary16 @jiayingz @vishh @derekwaynecarr @3xx0 @flx42

I made the following changes
- Updated with the new 1.10 API
- Added a naming section
  - This is a big one, I want to make sure we all use the same names to designate the same components
  - Feel free to comment on this
- Some small rewording

The names introduced here are:
- **Device Plugin**: The pod running on each node which allows kubelet and kubernetes to be "device aware"
- **Device Plugin API**: The gRPC API which allows kubelet and the device plugins to communicate. 
- **Device Plugin API Container**: The container which communicates with kubelet through the Device Plugin API
- **device manager**: The part of Kubelet in charge of managing the device plugins


flx42:
> 
>     Device Plugin: The pod running on each node which allows kubelet and kubernetes to be "device aware"
>     Device Plugin API Container: The container which communicates with kubelet through the Device Plugin API

This is the most important part of the PR, it might shape future directions.

To give some more context, @RenaudWasTaken and I were debating the meaning of "Device Plugin".
My stance was that Device Plugin == Device Plugin API Container. I don't see a distinction today. We might agree that this distinction should exist in the future, but then we would need to agree on what else should be part of this larger pod.

Is it monitoring? Is it driver installation?
In those cases, should we really call it "device plugin pod"?

Thoughts?

RenaudWasTaken:
> We might agree that this distinction should exist in the future, but then we would need to agree on what else should be part of this larger pod.
> Is it monitoring? Is it driver installation?

My opinion on this this is what we agreed in the F2F when we said that monitoring is part of the device plugin user story. And before when we agreed that driver installation is also part of that story.

We didn't agree that monitoring or driver installation should be part of the API or not.
We did however agree on the following story: "When a cluster admin deploys a device plugin daemonset in his cluster he will get monitoring of that device".

contributors/design-proposals/resource-management/device-plugin.md
```diff
@@ -61,51 +61,72 @@ the following simple steps:
 
 # Use Cases
 
- * I want to use a particular device type (GPU, InfiniBand, FPGA, etc.)
-   in my pod.
- * I should be able to use that device without writing custom Kubernetes code.
+ * As a cluster admin I want my cluster to be aware of the devices on my nodes:
+    * My nodes should report what devices are present.
+    * My users should be able to request "device containers" (GPU, InfiniBand,
```
vikaschoudhary16: will it be cleaner to simply say "devices for containers"

vikaschoudhary16: For Infiniband/FPGA, do we have a roadmap on how and when these will be supported? API changes in this doc are not sufficient for these resources, in that case, should we add these in some "Futuristic" or similar section?

contributors/design-proposals/resource-management/device-plugin.md
```diff
@@ -61,51 +61,72 @@ the following simple steps:
...
-
-These servers implement the gRPC interface defined later in this design
-document and once the device plugin makes itself known to kubelet, kubelet
-will interact with the device through two simple functions:
-  1. A `ListAndWatch` function for the kubelet to Discover the devices and
-     their properties as well as notify of any status change (device
-     became unhealthy).
-  2. An `Allocate` function which is called before creating a user container
-     consuming any exported devices
+At their core, device plugins are a plugin which runs on each device node.
```
vikaschoudhary16: s/ their core, device plugins are/its core, device plugin is

vikaschoudhary16:
Since pod is the elemental object in k8s API, IMO,  "device plugin" == "device plugin pod". Monitoring should be part of the device plugin. It could be in same container as DPAPI container or a separate container. Though, not  absolutely sure about driver installation but mostly it could be outside of device plugin.

contributors/design-proposals/resource-management/device-plugin.md
RenaudWasTaken: > will it be cleaner to simply say "devices for containers"

"devices aware containers" should be a better term

> For Infiniband/FPGA, do we have a roadmap on how and when these will be supported? API changes in this doc are not sufficient for these resources, in that case, should we add these in some "Futuristic" or similar section?

This seems a bit out of scope of this PR, I believe an issue might be a better format than editing this PR.

contributors/design-proposals/resource-management/device-plugin.md
```diff
@@ -61,51 +61,72 @@ the following simple steps:
...
-   in my pod.
- * I should be able to use that device without writing custom Kubernetes code.
+ * As a cluster admin I want my cluster to be aware of the devices on my nodes:
+    * My nodes should report what devices are present.
+    * My users should be able to request "device aware containers" (GPU,
+      InfiniBand, FPGA, etc...).
+    * My monitoring tool should be able to report device metrics.
+
+ * I should be able to enable my cluster to be aware of a device without
+   writing custom Kubernetes code.
```
vishh: Another way to phrase this `I should be able to enable access to devices in kubernetes by extending it rather than forking it`

contributors/design-proposals/resource-management/device-plugin.md
```diff
@@ -61,51 +61,72 @@ the following simple steps:
...
+* Add support for vendor specific Devices in kubelet:
     * Through an extension mechanism.
     * Which allows discovery and health check of devices.
-    * Which allows hooking the runtime to make devices available in containers
-      and cleaning them up.
-2. Define a deployment mechanism for this new API.
-3. Define a versioning mechanism for this new API.
+    * Which allows hooking the lifecycle of pods and containers to make devices
+      available in containers requesting these devices.
+* Define a deployment mechanism for this new API.
```
vishh: I'd add lifecycle management to the list of Objectives

PR #2067, opened 2018-04-24
Add proposal for node-level User Namespaces remapping support

vikaschoudhary16:
/cc @mrunalp @adelton @derekwaynecarr @sjenning

### DIFF
--- a/contributors/design-proposals/node/node-usernamespace-remapping.md
+++ b/contributors/design-proposals/node/node-usernamespace-remapping.md
@@ -1,0 +1,209 @@
+# Support Node-Level User Namespaces Remapping
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non-Goals](#non-goals)
+- [Use Stories](#user-stories)
+- [Proposal](#proposal)
+- [Future Work](#future-work)
+- [Risks and Mitigations](risks-and-mitigations)
+- [Graduation Criteria](graduation-criteria)
+- [Alternatives](alternatives)
+
+
+_Authors:_
+
+* Mrunal Patel &lt;mpatel@redhat.com&gt;
+* Jan Pazdziora &lt;jpazdziora@redhat.com&gt;
+* Vikas Choudhary &lt;vichoudh@redhat.com&gt;
+
+## Summary
+Container security consists of many different kernel features that work together to make containers secure. User namespaces is one such feature that enables interesting possibilities for containers by allowing them to be root inside the container while not being root on the host. This gives more capabilities to the containers while protecting the host from the container being root and adds one more layer to container security. 
+In this proposal we discuss:
+- use-cases/user-stories that benefit from this enhancement
+- implementation design and scope for alpha release
+- long-term roadmap to fully support this feature beyond alpha
+
+## Motivation
+From user_namespaces(7):
+> User namespaces isolate security-related identifiers and attributes, in particular, user IDs and group IDs, the root directory, keys, and capabilities. A process's user and group IDs can be different inside and outside a user namespace. In particular, a process can have a normal unprivileged user ID outside a user namespace while at the same time having a user ID of 0 inside the namespace; in other words, the process has full privileges for operations inside the user namespace, but is unprivileged for operations outside the namespace.
+
+In order to run Pods with software which expects to run as root or with elevated privileges while still containing the processes and protecting both the Nodes and other Pods, Linux kernel mechanism of user namespaces can be used make the processes in the Pods view their environment as having the privileges, while on the host (Node) level these processes appear as without privileges or with privileges only affecting processes in the same Pods
+
+The purpose of using user namespaces in Kubernetes is to let the processes in Pods think they run as one uid set when in fact they run as different “real” uids on the Nodes.
+
+In this text, most everything said about uids can also be applied to gids.
+
+## Goals
+Enable user namespace support in a kubernetes cluster so that workloads that work today also work with user namespaces enabled at runtime. Furthermore, make workloads that require root/privileged user inside the container, safer for the node using the additional security of user namespaces. Containers will run in a user namespace different from user-namespace of the underlying host.
+
+## Non-Goals
+- Non-goal is to support pod/container level user namespace isolation. There can be images using different users but on the node, pods/containers running with these images will share common user namespace remapping configuration. In other words, all containers on a node share a common user-namespace range.
+- Remote volumes support eg. NFS
+
+## User Stories
+- As a cluster admin, I want to protect the node from the rogue container process(es) running inside pod containers with root privileges. If such a process is able to break out into the node, it could be a security issue.
+- As a cluster admin, I want to support all the images irrespective of what user/group that image is using.
+- As a cluster admin, I want to allow some pods to disable user namespaces if they require elevated privileges.
+
+## Proposal
+Proposal is to support user-namespaces for the pod containers. This can be done at two levels:
+- Node-level : This proposal explains this part in detail.
+- Namespace-Level/Pod-level: Plan is to target this in future due to missing support in the low level system components such as runtimes and kernel. More on this in the `Future Work` section.
+
+Node-level user-namespace support means that, if feature is enabled, all pods on a node will share a common user-namespace, common UID(and GID) range (which is a subset of node’s total UIDs(and GIDs)). This common user-namespace is runtime’s default user-namespace range which is remapped to containers’ UIDs(and GID), starting with the first UID as container’s ‘root’.
+In general Linux convention, UID(or GID) mapping consists of three parts:
+1. Host (U/G)ID: First (U/G)ID of the range on the host that is being remapped to the (U/G)IDs in the container user-namespace
+2. Container (U/G)ID: First (U/G)ID of the range in the container namespace and this is mapped to the first (U/G)ID on the host(mentioned in previous point).
+3. Count/Size: Total number of consecutive mapping between host and container user-namespaces, starting from the first one (including) mentioned above.
+
+As an example, `host_id 1000, container_id 0, size 10`
+In this case, 1000 to 1009 on host will be mapped to 0 to 9 inside the container.
+
+User-namespace support should be enabled only when container runtime on the node supports user-namespace remapping and is enabled in its configuration. To enable user-namespaces, feature-gate flag will need to be passed to Kubelet like this `--feature-gates=”NodeUserNamespace=true”`
+
+A new CRI API, `GetRuntimeConfigInfo` will be added. Kubelet will use this API:
+- To verify if user-namespace remapping is enabled at runtime. If found disabled, kubelet will fail to start
+- To determine the default user-namespace range at the runtime, starting UID of which is mapped to the UID '0' of the container.
+
+### Volume Permissions
+Kubelet will change the file permissions, i.e chown, at `/var/lib/kubelet/pods` prior to any container start to get file permissions updated according to remapped UID and GID.
+This proposal will work only for local volumes and not with remote volumes such as NFS.
+
+### How to disable `NodeUserNamespace` for a specific pod
+This can be done in two ways:
+- **Alpha:** Implicitly using host namespace for the pod containers
+This support is already present (currently it seems broken, will be fixed) in Kubernetes as an experimental functionality, which can be enabled using `feature-gates=”ExperimentalHostUserNamespaceDefaulting=true”`.
+If Pod-Security-Policy is configured to allow the following to be requested by a  pod, host user-namespace will be enabled for the container:
+  - host namespaces (pid, ipc, net)
+  - non-namespaced capabilities (mknod, sys_time, sys_module)
+  - the pod contains a privileged container or using host path volumes.
+  - https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c#diff-a53fa76e941e0bdaee26dcbc435ad2ffR437 introduced via https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c.
+
+- **Beta:** Explicit API to request host user-namespace in pod spec
+          This is being targeted under Beta graduation plans.
+
+### CRI API Changes
+Proposed CRI API changes:
+
+```golang
+// Runtime service defines the public APIs for remote container runtimes
+service RuntimeService {
+    // Version returns the runtime name, runtime version, and runtime API version.
+    rpc Version(VersionRequest) returns (VersionResponse) {}
+    …….
+    …….
+    //  GetRuntimeConfigInfo returns the configuration details of the runtime.
+    rpc GetRuntimeConfigInfo(GetRuntimeConfigInfoRequest) returns (GetRuntimeConfigInfoResponse) {}
+}
+// LinuxIDMapping represents a single user namespace mapping in Linux.
+message LinuxIDMapping {
+   // container_id is the starting id for the mapping inside the container.
+   uint32 container_id = 1;
+   // host_id is the starting id for the mapping on the host.
+   uint32 host_id = 2;
+   // size is the length of the mapping.
+   uint32 size = 3;
+}
+
+message LinuxUserNamespaceConfig {
+   // is_enabled, if true indicates that user-namespaces are supported and enabled in the container runtime
+   bool is_enabled = 1;
+   // uid_mappings is an array of user id mappings.
+   repeated LinuxIDMapping uid_mappings = 1;
+   // gid_mappings is an array of group id mappings.
+   repeated LinuxIDMapping gid_mappings = 2;
+}
+message GetRuntimeConfig {
+    LinuxUserNamespaceConfig user_namespace_config = 1;
+}
+
+message GetRuntimeConfigInfoRequest {}
+
+message GetRuntimeConfigInfoResponse {
+    GetRuntimeConfig runtime_config = 1
+}
+
+...
+
+// NamespaceOption provides options for Linux namespaces.
+message NamespaceOption {
+	// Network namespace for this container/sandbox.
+	// Note: There is currently no way to set CONTAINER scoped network in the Kubernetes API.
+	// Namespaces currently set by the kubelet: POD, NODE
+	NamespaceMode network = 1;
+	// PID namespace for this container/sandbox.
+	// Note: The CRI default is POD, but the v1.PodSpec default is CONTAINER.
+	// The kubelet's runtime manager will set this to CONTAINER explicitly for v1 pods.
+	// Namespaces currently set by the kubelet: POD, CONTAINER, NODE
+	NamespaceMode pid = 2;
+	// IPC namespace for this container/sandbox.
+	// Note: There is currently no way to set CONTAINER scoped IPC in the Kubernetes API.
+	// Namespaces currently set by the kubelet: POD, NODE
+	NamespaceMode ipc = 3;
+	// User namespace for this container/sandbox.
+	// Note: There is currently no way to set CONTAINER scoped user namespace in the Kubernetes API.
+	// The container runtime should ignore this if user namespace is NOT enabled.
+	// POD is the default value. Kubelet will set it to NODE when trying to use host user-namespace
+	// Namespaces currently set by the kubelet: POD, NODE
+	NamespaceMode user = 4;
+}
+
+```
+
+### Runtime Support
+- Docker: Here is the [user-namespace documentation](https://docs.docker.com/engine/security/userns-remap/) and this is the [implementation PR](https://github.com/moby/moby/pull/12648)
+    - Concerns:
+Docker API does not provide user-namespace mapping. Therefore to handle `GetRuntimeConfigInfo` API, changes will be done in `dockershim` to read system files, `/etc/subuid` and `/etc/subgid`, for figuring out default user-namespace mapping. `/info` api will be used to figure out if user-namespace is enabled and `Docker Root Dir` will be used to figure out host uid mapped to the uid `0` in container. eg. `Docker Root Dir: /var/lib/docker/2131616.2131616` this shows host uid `2131616` will be mapped to uid `0`
+- CRI-O: https://github.com/kubernetes-incubator/cri-o/pull/1519
+- Containerd: https://github.com/containerd/containerd/blob/129167132c5e0dbd1b031badae201a432d1bd681/container_opts_unix.go#L149
+
+### Implementation Roadmap
+#### Phase 1: Support in Kubelet, Alpha, [Target: Kubernetes v1.11]
+- Add feature gate `NodeUserNamespace`, disabled by default
+- Add new CRI API, `GetRuntimeConfigInfo()`
+- Add logic in Kubelet to handle pod creation which includes parsing GetRuntimeConfigInfo response and changing file-permissions in /var/lib/kubelet with learned userns mapping.
+- Add changes in dockershim to implement GetRuntimeConfigInfo() for docker runtime
+- Add changes in CRI-O to implement userns support and GetRuntimeConfigInfo() support
+- Unit test cases
+- e2e tests
+
+#### Phase 2: Beta Support [Target: Kubernetes v1.12]
+- PSP integration
+- To grow ExperimentalHostUserNamespaceDefaulting from experimental feature gate to a Kubelet flag
+- API changes to allow pod able to request HostUserNamespace in pod spec
+- e2e tests
+
+### References
+- Default host user namespace via experimental flag
+  - https://github.com/kubernetes/kubernetes/pull/31169
+- Enable userns support for containers launched by kubelet
+  - https://github.com/kubernetes/features/issues/127
+- Track Linux User Namespaces in the Pod Security Policy
+  - https://github.com/kubernetes/kubernetes/issues/59152
+- Add support for experimental-userns-remap-root-uid and experimental-userns-remap-root-gid options to match the remapping used by the container runtime.
+  - https://github.com/kubernetes/kubernetes/pull/55707
+- rkt User Namespaces Background
+  - https://coreos.com/rkt/docs/latest/devel/user-namespaces.html
+
+## Future Work
+### Namespace-Level/Pod-Level user-namespace support
+There is no runtime today which supports creating containers with a specified user namespace configuration. For example here is the discussion related to this support in Docker https://github.com/moby/moby/issues/28593
+Once user-namespace feature in the runtimes has evolved to support container’s request for a specific user-namespace mapping(UID and GID range), we can extend current Node-Level user-namespace support in Kubernetes to support Namespace-level isolation(or if desired even pod-level isolation) by dividing and allocating learned mapping from runtime among Kubernetes namespaces (or pods, if desired). From end-user UI perspective, we dont expect any change in the UI related to user namespaces support.
+### Remote Volumes
+Remote Volumes support should be investigated and should be targeted in future once support is there at lower infra layers.
+
+
+## Risks and Mitigations
+The main risk with this change stems from the fact that processes in Pods will run with different “real” uids than they used to, while expecting the original uids to make operations on the Nodes or consistently access shared persistent storage.
+- This can be mitigated by turning the feature on gradually, per-Pod or per Kubernetes namespace.
+- For the Kubernetes' cluster Pods (that provide the Kubernetes functionality), testing of their behaviour and ability to run in user namespaced setups is crucial.
+
+## Graduation Criteria
+- PSP integration
+- API changes to allow pod able to request host user namespace using for example, `HostUserNamespace: True`, in pod spec
+- e2e tests
+
+## Alternatives
+User Namespace mappings can be passed explicitly through kubelet flags similar to https://github.com/kubernetes/kubernetes/pull/55707 but we do not prefer this option because this is very much prone to mis-configuration.


contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Runtime Support
+- Docker: https://docs.docker.com/engine/security/userns-remap/
+    - Concerns:
+Docker API does not provide user-namespace mapping. Therefore to handle `RuntimeConfigInfo` API, changes will be done in `dockershim` to read system files, `/etc/subuid` and `/etc/subgid`, for figuring out default user-namespace mapping. It won’t be possible to use custom user/group name to `--userns-remap parameter`, the implementation will expect `dockremap`, which is the default username.
+- CRI-O: WIP https://github.com/nalind/cri-o/tree/idmapping
+
+### Implementation Roadmap
+#### Phase 1: Support in Kubelet, Alpha, [Target: Kubernetes v1.11]
+- Add feature gate `NodeUserNamespace`, disabled by default
```
adelton: So are we adding new gate `NodeUserNamespace`, or reviving and fixing the existing `ExperimentalHostUserNamespaceDefaulting`?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,216 @@
...
+   repeated LinuxIDMapping uid_mappings = 1;
+   // gid_mappings is an array of group id mappings.
+   repeated LinuxIDMapping gid_mappings = 2;
+}
+
+message UserNamespaceConfig {
+   // is_enabled, if true indicates that user-namespaces are supported and enabled in the container runtime
+   string is_enabled = 1;
+   // node_usernamespace_mapping is the user namespace mapping set for the container runtime
+   UserNSMapping node_usernamespace_mapping = 2;
```
adelton: This is a name mismatch to the above `message UserNamespaceMapping`?

But, could we drop this completely and just use `repeated LinuxIDMapping uid_mappings` and `repeated LinuxIDMapping gid_mappings` here directly?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,216 @@
...
+   // container_id is the starting id for the mapping inside the container.
+   uint32 container_id = 1;
+   // host_id is the starting id for the mapping on the host.
+   uint32 host_id = 2;
+   // size is the length of the mapping.
+   uint32 size = 3;
+}
+
+// UserNamespaceMapping is a user namespace mapping for uids and gids.
+message UserNamespaceMapping {
```
adelton: Can we drop this and just put the `uid_mappings` / `gid_mappings` below directly?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,216 @@
...
+   // uid_mappings is an array of user id mappings.
+   repeated LinuxIDMapping uid_mappings = 1;
+   // gid_mappings is an array of group id mappings.
+   repeated LinuxIDMapping gid_mappings = 2;
+}
+
+message UserNamespaceConfig {
```
adelton: Could we name this something like `RuntimeConfigInfoUserNamespace`, to make it clearer that this is about getting information from the runtime, not about specifying user namespace for a pod.

adelton:
With the Beta phase of this proposal, users will be able to "escape" the user namespace using `"hostUserNamespace": true` in Pod Spec, controlled by PSP.

However, users will also be able to escape the user namespace by simply having their Pod scheduled on a Node where docker is not configured with `--userns-remap`. If we do not specify the `User` `NamespaceOption`, we will get the Container Runtime default, which might either be no mapping (== `host`), or some non-host mapping.

Are we thinking of a Node classification (does not support user namespaces (current docker without `--userns-remap`), supports one Node-level user namespace (current docker with `--userns-remap`, future CRI-O with `--userns-uid-map` and `--userns-uid-map`), supports per-Pod user namespace (future CRI-O)) which could close this loophole?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,216 @@
...
+	// User namespace for this container/sandbox.
+	// Note: There is currently no way to set CONTAINER scoped user namespace in the
+	// Kubernetes API. The container runtime has to run with user namespace enabled,
+	// and NODE then forces the use of host user namespace.
+	// Namespaces currently set by the kubelet: POD, NODE
+	NamespaceModeValue user = 4;
+}
+
+// NamespaceModeValue is the wrapper of NamespaceMode, to allow non-POD default.
+message NamespaceModeValue {
```
adelton: The use of `NamespaceModeValue` wrapper for `user` seems strange. Ideally we might want to go with just `NamespaceMode user = 4;` like we do for `network`, `pid`, and `ipc`. The problem is in the default. Ideally we'd like to go with `POD` ... but with existing docker Container Runtimes without `--userns-remap` specified, we always get no remapping (== effectively `NODE`) anyway.

Do we have some options to resolve this issue?

adelton:
@vikaschoudhary16, I thought the plan was to create this as KEP, using the KEP process and naming.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,216 @@
...
+message UserNamespaceMapping {
+   // uid_mappings is an array of user id mappings.
+   repeated LinuxIDMapping uid_mappings = 1;
+   // gid_mappings is an array of group id mappings.
+   repeated LinuxIDMapping gid_mappings = 2;
+}
+
+message UserNamespaceConfig {
+   // is_enabled, if true indicates that user-namespaces are supported and enabled in the container runtime
+   string is_enabled = 1;
```
AkihiroSuda: bool?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,216 @@
...
+	NamespaceMode value = 1;
+}
+```
+
+### Runtime Support
+- Docker: https://docs.docker.com/engine/security/userns-remap/
+    - Concerns:
+Docker API does not provide user-namespace mapping. Therefore to handle `RuntimeConfigInfo` API, changes will be done in `dockershim` to read system files, `/etc/subuid` and `/etc/subgid`, for figuring out default user-namespace mapping. It won’t be possible to use custom user/group name to `--userns-remap parameter`, the implementation will expect `dockremap`, which is the default username.
+- CRI-O: WIP https://github.com/nalind/cri-o/tree/idmapping
+
```
AkihiroSuda: containerd has support for remapped snapshot.

https://github.com/containerd/containerd/blob/129167132c5e0dbd1b031badae201a432d1bd681/container_opts_unix.go#L149

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: @adelton we are adding `NodeUserNamespace` for making it possible to run containers with remapped uid 0 and  will be fixing `ExperimentalHostUserNamespaceDefaulting` to allow privileged pod containers escape from user namespace feature enablement and use host namespace. Both flags have different roles.

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: does not that seem to be a bit redundant because at L#120, we are using it `RuntimeConfig`?

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: thanks, will update.

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: +1 on name mismatch. On keeping uid and gids directly in `RuntimeConfig` --  I am trying to follow existing pattern where all the network configurations are abstracted in `NetworkConfig`. If we put uids and gids directly in RuntimeConfig, would not that demand to put all the NetworkConfig members also directly for the sake of consistency?

vikaschoudhary16: +1

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: +1

vikaschoudhary16:
> With the Beta phase of this proposal, users will be able to "escape" the user namespace using "hostUserNamespace": true in Pod Spec, controlled by PSP.
>
>However, users will also be able to escape the user namespace by simply having their Pod scheduled on a Node where docker is not configured with --userns-remap. If we do not specify the User NamespaceOption, we will get the Container Runtime default, which might either be no mapping (== host), or some non-host mapping.
>
> Are we thinking of a Node classification (does not support user namespaces (current docker without --userns-remap), supports one Node-level user namespace (current docker with --userns-remap, future CRI-O with --userns-uid-map and --userns-uid-map), supports per-Pod user namespace (future CRI-O)) which could close this loophole?

Node classification based on runtime capabilities is an important point to discuss and is already being discussed [here to implement using a `RuntimeController`](https://docs.google.com/document/d/1EJKT4gyl58-kzt2bnwkv08MIUZ6lkDpXcxkHqCvvAp4/edit#heading=h.grgs3u22u7ze) . I would prefer to scope Node classification discussion out of this proposal. Thoughts?

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: Thats what we are trying to solve here by using this wrapper `NamespaceModeValue`. in proto3, sub-messages can be set to `nil`.

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: Well, that's the problem -- the existing `RuntimeConfig` only has `NetworkConfig` which seems to be used to *set* `pod_cidr`.

And now we seem to try to use the same message in `RuntimeConfigInfoResponse`, to *get* information from the Container Runtime. Maybe `RuntimeConfigInfoResponse` shouldn't have `RuntimeConfig` but `UserNamespaceConfig` directly.

vikaschoudhary16:
> @vikaschoudhary16, I thought the plan was to create this as KEP, using the KEP process and naming.

right, will update.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- Add feature gate `NodeUserNamespace`, disabled by default
+- Add new CRI API, `RuntimeConfigInfo()`
+- Add logic in Kubelet to handle pod creation which includes parsing RuntimeConfigInfo response and changing file-permissions in /var/lib/kubelet with learned userns mapping.
+- Add changes in dockershim to implement RuntimeConfigInfo() for docker runtime
+- Add changes in CRI-O to implement userns support and RuntimeConfigInfo() support
+- Unit test cases
+
+#### Phase 2: Beta Support [Target: Kubernetes v1.12]
+- PSP integration
+- To grow ExperimentalHostUserNamespaceDefaulting from experimental feature gate to a Kubelet flag
```
pweil-: should we wrap `ExperimentalHostUserNamespaceDefaulting` into `NodeUserNamespace` behavior?  It seems like you'd want them both.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- Once user-namespace feature in the runtimes has evolved to support container’s request for a specific user-namespace mapping(UID and GID range) by name(range identifier), Kubernetes might leverage this to support Namespace level isolation by allocating a different runtime user-namespace mapping (UID and GID range) to each Kubernetes namespace.
+
+## Risks and Mitigations
+The main risk with this change stems from the fact that processes in Pods will run with different “real” uids than they used to, while expecting the original uids to make operations on the Nodes or consistently access shared persistent storage.
+- This can be mitigated by turning the feature on gradually, per-Pod or per Kubernetes namespace.
+- For the Kubernetes' cluster Pods (that provide the Kubernetes functionality), testing of their behaviour and ability to run in user namespaced setups is crucial.
+
+## Graduation Criteria
+- PSP integration
+- API changes to allow pod able to request host user namespace using for example, `HostUserNamespace: True`, in pod spec
```
pweil-: Is there a use case for this outside of the items mentioned in defaulting (priv, other host ns, host volumes, etc)?

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: I'd much prefer for users to be explicit with `"hostUserNamespace": true` when needed, than rely on the implicit magic behaviour.

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: It is pretty much to be implicit about using the host / `NODE` user namespace, without having rely on the defaulting behaviour. For a start, the reasons might be those that you've listed.

Long term, we will probably want the default to be for every Pod to get unique per-Pod user namespace, unless the Pod Spec requests some specific user namespace mapping, or unless the Pod Spec requests to run with host / `NODE` user namespace, for whatever reason -- just for root to be a host root.

contributors/design-proposals/node/node-usernamespace-remapping.md
pweil-: I wonder about the experience there when the system knows that they need `hostUserNamespace: true`.  Seems like tying node level configuration knowledge to a user trying to run a pod.  

I consider it node level configuration because on a non-userns remap node not setting `hostUserNamespace` would work fine, on the userns remap node it would fail runtime validation during container creation.  The user expectation was just that they are allowed to user `hostNetwork` (or whatever) and it should work.

contributors/design-proposals/node/node-usernamespace-remapping.md
AkihiroSuda: I thought the list in the document is intended to show the current implementation status of several CRI runtimes including containerd, but no strong opinion.

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: > should we wrap ExperimentalHostUserNamespaceDefaulting into NodeUserNamespace behavior? It seems like you'd want them both.

only motivation for keeping them seperate is the flexibility that it provides to admin to choose whether he wants to allow privileged pods or not, in case NodeUserNamespace is enabled

adelton: @pweil-, would you have some link to discussion which would explain why having one host namespace implies the user namespace to be host as well? Is that a general thing in the kernel which will always be true, or merely behaviour of the current Container Runtimes like docker?

vikaschoudhary16: @adelton `"hostUserNamespace": true` seems complementary to the `ExperimentalHostUserNamespaceDefaulting`, rather an alternative to it. IMO, `ExperimentalHostUserNamespaceDefaulting` functionality is required even with `"hostUserNamespace": true` support.

mrunalp: @pweil- I think it would make sense to fold them into one. I can't think of cases where users will be able to get away without enabling HostUserNamespaceDefaulting.

contributors/design-proposals/node/node-usernamespace-remapping.md
mrunalp: I think handling what users mean by asking for privileged is hard to figure out automatically. It could imply give me all CAPs in a non host user namespace or give me all CAPs in the host user namespace.

contributors/design-proposals/node/node-usernamespace-remapping.md
derekwaynecarr: @AkihiroSuda thanks for the pointer, the intention is to show the capability is supportable across a variety of runtime choices.  adding a section on containerd is helpful for everyone.

mrunalp: :+1: We should add a line for containerd support as well.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+// NamespaceModeValue is the wrapper of NamespaceMode, to allow non-POD default.
+message NamespaceModeValue {
+	NamespaceMode value = 1;
+}
+```
+
+### Runtime Support
+- Docker: https://docs.docker.com/engine/security/userns-remap/
+    - Concerns:
+Docker API does not provide user-namespace mapping. Therefore to handle `RuntimeConfigInfo` API, changes will be done in `dockershim` to read system files, `/etc/subuid` and `/etc/subgid`, for figuring out default user-namespace mapping. It won’t be possible to use custom user/group name to `--userns-remap parameter`, the implementation will expect `dockremap`, which is the default username.
```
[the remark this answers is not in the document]
pweil-: The docker info api tells you if userns remap is enabled.  Should we use that to tell if we should read /etc files?  https://github.com/moby/moby/blob/6cd806aa53de41e6f04346150d08e19ec63311e0/daemon/info.go#L75-L78

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: > IMO, ExperimentalHostUserNamespaceDefaulting functionality is required even with "hostUserNamespace": true support.

In my testing, having the explicit `"hostUserNamespace": true` seems to work even with the unfixed `ExperimentalHostUserNamespaceDefaulting`. Granted, I've only checked if the process in the container is indeed running as uid 0 on the host, didn't do any functional testing.

adelton: > Capabilities for a non-user namespace are checked against the owning user namespace so any time you're sharing a non-user namespace you need to share the owning user namespace in order to have permissions to use the non-user namespace

Thanks.

contributors/design-proposals/node/node-usernamespace-remapping.md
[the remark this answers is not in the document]
vikaschoudhary16: based on docker info response, dockershim can decide when it should read /etc/subuid.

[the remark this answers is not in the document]
mrunalp: Yeah, that should work. I thought it wasn't supported on docker 1.13 but it is.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+   repeated LinuxIDMapping uid_mappings = 1;
+   // gid_mappings is an array of group id mappings.
+   repeated LinuxIDMapping gid_mappings = 2;
+}
+message RuntimeConfig {
+    NetworkConfig network_config = 1;
+    UserNamespaceConfig user_namespace_config = 2;
+}
+
+message UpdateRuntimeConfigRequest {
```
AkihiroSuda: Could you explain when/how kublet is expected to call this API?

mrunalp:
@jessfraz @cyphar @rhatdan PTAL

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- As a cluster admin, I want to protect the node from the rogue container process(es) running inside pod containers with root privileges. If such a process is able to break out into the node, it could be a security issue.
+- As a cluster admin, I want to support all the images irrespective of what user/group that image is using.
+- As a cluster admin, I want to allow some pods to disable user namespaces if they require elevated privileges.
+
+## Proposal
+Proposal is to support user-namespaces for the pod containers. This can be done at two levels:
+- Node-level : This proposal explains this part in detail.
+- Pod-level: Plan is to target this in future due to missing support in the low level system components such as runtimes and kernel. More on this in the later sections.
+
+Node-level  user-namespace support means that, if feature is enabled, all pods on a node will share a common user-namespace, common UID(and GID) range (which is a subset of node’s total UIDs(and GIDs)). This common user-namespace is runtime’s default user-namespace range which is remapped to containers’ UIDs(and GID), starting with the first UID as container’s ‘root’.
```
tallclair: > starting with the first UID as container’s ‘root’

Does this mean that all IDs will be offset by 1?

tallclair: Would the UID mapping range include the entire UID space (except 0)?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+
+User-namespace support should be enabled only when container runtime on the node supports user-namespace remapping and is enabled in its configuration. To enable user-namespaces, feature-gate flag will need to be passed to Kubelet like this `--feature-gates=”NodeUserNamespace=true”`
+
+A new CRI API, `RuntimeConfigInfo` will be added. Kubelet will use this API:
```
tallclair: This should be Linux-specific (or have a linux-specific sub object)

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- To verify if user-namespace remapping is enabled at runtime. If found disabled, kubelet will fail to start
+- To determine the default user-namespace range at the runtime, starting UID of which is mapped to the UID '0' of the container.
+
+### Volume Permissions
+Kubelet will change the file permissions at `/var/lib/kubelet/pods` prior to any container start to get file permissions updated according to remapped UID and GID.
```
tallclair: `change the file permissions` == chown?

What happens if a remote volume type is attached, and then unattached, and then reattached (i.e. pod is evicted, and rescheduled). How do you prevent the IDs by being shifted each time?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+### How to disable `NodeUserNamespace` for a specific pod
+This can be done in two ways:
+- **Alpha:** Implicitly using host namespace for the pod containers
+This support is already present (currently it seems broken, will be fixed) in Kubernetes as an experimental functionality, which can be enabled using `feature-gates=”ExperimentalHostUserNamespaceDefaulting=true”`.
+If Pod-Security-Policy is configured to allow the following to be requested by a  pod, host user-namespace will be enabled for the container:
+  - host namespaces (pid, ipc, net)
+  - non-namespaced capabilities (mknod, sys_time, sys_module)
+  - the pod contains a privileged container or using host path volumes.
+  - https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/container/runtime.go#L451 introduced via https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c.
+- **Beta:** Explicit API to request host user-namespace in pod spec
```
tallclair: :+1:

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+
+### How to disable `NodeUserNamespace` for a specific pod
+This can be done in two ways:
+- **Alpha:** Implicitly using host namespace for the pod containers
+This support is already present (currently it seems broken, will be fixed) in Kubernetes as an experimental functionality, which can be enabled using `feature-gates=”ExperimentalHostUserNamespaceDefaulting=true”`.
+If Pod-Security-Policy is configured to allow the following to be requested by a  pod, host user-namespace will be enabled for the container:
+  - host namespaces (pid, ipc, net)
+  - non-namespaced capabilities (mknod, sys_time, sys_module)
+  - the pod contains a privileged container or using host path volumes.
+  - https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/container/runtime.go#L451 introduced via https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c.
```
tallclair: nit: reference a specific commit for runtime line (press `y` in github) so this reference doesn't go out of date.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+   // is_enabled, if true indicates that user-namespaces are supported and enabled in the container runtime
+   bool is_enabled = 1;
+   // uid_mappings is an array of user id mappings.
+   repeated LinuxIDMapping uid_mappings = 1;
+   // gid_mappings is an array of group id mappings.
+   repeated LinuxIDMapping gid_mappings = 2;
+}
+message RuntimeConfig {
+    NetworkConfig network_config = 1;
+    UserNamespaceConfig user_namespace_config = 2;
```
tallclair: `LinuxRuntimeConfig`?

contributors/design-proposals/node/node-usernamespace-remapping.md
tallclair: Shouldn't the default be set by the kubelet?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+	NamespaceModeValue user = 4;
+}
+
+// NamespaceModeValue is the wrapper of NamespaceMode, to allow non-POD default.
+message NamespaceModeValue {
+	NamespaceMode value = 1;
+}
+```
+
+### Runtime Support
```
tallclair: How are the mapped UIDs applied to the container image layers? Does docker `chown` the files when the image is unpacked? What happens if a container running in the host user namespace shares a base layer with a container that is running in a different user namespace?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+
+## User Stories
+- As a cluster admin, I want to protect the node from the rogue container process(es) running inside pod containers with root privileges. If such a process is able to break out into the node, it could be a security issue.
+- As a cluster admin, I want to support all the images irrespective of what user/group that image is using.
+- As a cluster admin, I want to allow some pods to disable user namespaces if they require elevated privileges.
+
+## Proposal
+Proposal is to support user-namespaces for the pod containers. This can be done at two levels:
+- Node-level : This proposal explains this part in detail.
+- Pod-level: Plan is to target this in future due to missing support in the low level system components such as runtimes and kernel. More on this in the later sections.
```
AkihiroSuda: > More on this in the later sections.

Could you add a section about this?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Volume Permissions
+Kubelet will change the file permissions at `/var/lib/kubelet/pods` prior to any container start to get file permissions updated according to remapped UID and GID.
+
+### How to disable `NodeUserNamespace` for a specific pod
+This can be done in two ways:
+- **Alpha:** Implicitly using host namespace for the pod containers
+This support is already present (currently it seems broken, will be fixed) in Kubernetes as an experimental functionality, which can be enabled using `feature-gates=”ExperimentalHostUserNamespaceDefaulting=true”`.
+If Pod-Security-Policy is configured to allow the following to be requested by a  pod, host user-namespace will be enabled for the container:
+  - host namespaces (pid, ipc, net)
```
jessfraz: you should be able to combine these with user namespaces...

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+
+User-namespace support should be enabled only when container runtime on the node supports user-namespace remapping and is enabled in its configuration. To enable user-namespaces, feature-gate flag will need to be passed to Kubelet like this `--feature-gates=”NodeUserNamespace=true”`
+
+A new CRI API, `RuntimeConfigInfo` will be added. Kubelet will use this API:
+- To verify if user-namespace remapping is enabled at runtime. If found disabled, kubelet will fail to start
+- To determine the default user-namespace range at the runtime, starting UID of which is mapped to the UID '0' of the container.
+
+### Volume Permissions
+Kubelet will change the file permissions at `/var/lib/kubelet/pods` prior to any container start to get file permissions updated according to remapped UID and GID.
+
```
AkihiroSuda: How does it work with shared volumes such as NFS?
IIUC all nodes in a cluster must have consistent `subuid` and `subgid` file, right?

jessfraz:
If you are then planning on using the userns containers as builders for something like kaniko then you are going to want to use `newuidmap` tools so you can map multiple users and `setgroups`
(but that might be a part two)
cc @AkihiroSuda

contributors/design-proposals/node/node-usernamespace-remapping.md
mrunalp: @tallclair e.g. host_id 1000, container_id 0, size 10
In this case, 1000 to 1009 on host will be mapped to 0 to 9 inside the container.

mrunalp: In this proposal kubelet learns the range from the runtime through the API. It will be upto the admin to setup the runtime according to their needs but yes 64K is a nice value :)

contributors/design-proposals/node/node-usernamespace-remapping.md
mrunalp: :+1: We can update it.

mrunalp:
@jessfraz Yes, per pod or k8s namespace mappings will be addressed later on. This is to get the feature going as a first step.

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: It's be nice to be able to start setting the default from now on. But what should the default be? `NODE`, and `POD` only when we detect that the Container Runtime can support it and will do it for us (like docker with `--userns-remap` enabled), but not when `NODE` is requested by `ExperimentalHostUserNamespaceDefaulting` gate or `hostUserNamespace`?

contributors/design-proposals/node/node-usernamespace-remapping.md
mrunalp: @jessfraz Yes, technically possible. We can look at the existing use cases for host ipc, host pid and host net and see what works as expected when mixed with non-host user namespaces.

contributors/design-proposals/node/node-usernamespace-remapping.md
yujuhong: I don't think this will work with remote volumes across different types. Some persistent disks can also be attached and mounted to multiple nodes. Need more clarification on what "volumes" will be handled.

contributors/design-proposals/node/node-usernamespace-remapping.md
yujuhong: +1

/cc @Random-Liu

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+### Runtime Support
+- Docker: https://docs.docker.com/engine/security/userns-remap/
+    - Concerns:
+Docker API does not provide user-namespace mapping. Therefore to handle `RuntimeConfigInfo` API, changes will be done in `dockershim` to read system files, `/etc/subuid` and `/etc/subgid`, for figuring out default user-namespace mapping. It won’t be possible to use custom user/group name to `--userns-remap parameter`, the implementation will expect `dockremap`, which is the default username.
+- CRI-O: WIP https://github.com/nalind/cri-o/tree/idmapping
+
+### Implementation Roadmap
+#### Phase 1: Support in Kubelet, Alpha, [Target: Kubernetes v1.11]
+- Add feature gate `NodeUserNamespace`, disabled by default
+- Add new CRI API, `RuntimeConfigInfo()`
```
yujuhong: This needs a CRI validation test.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- Add logic in Kubelet to handle pod creation which includes parsing RuntimeConfigInfo response and changing file-permissions in /var/lib/kubelet with learned userns mapping.
+- Add changes in dockershim to implement RuntimeConfigInfo() for docker runtime
+- Add changes in CRI-O to implement userns support and RuntimeConfigInfo() support
+- Unit test cases
```
yujuhong: Also need an end-to-end test.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+  - https://github.com/kubernetes/features/issues/127
+- Track Linux User Namespaces in the Pod Security Policy
+  - https://github.com/kubernetes/kubernetes/issues/59152
+- Add support for experimental-userns-remap-root-uid and experimental-userns-remap-root-gid options to match the remapping used by the container runtime.
+  - https://github.com/kubernetes/kubernetes/pull/55707
+- rkt User Namespaces Background
+  - https://coreos.com/rkt/docs/latest/devel/user-namespaces.html
+
+## Future Work
+- Named User namespaces: There is no runtime today which supports creating containers with a named user namespace. For example here is the discussion related to this support in Docker https://github.com/moby/moby/issues/28593 and support is being created for CRI-O https://github.com/nalind/cri-o/tree/idmapping.
```
kargakis: Link to the cri-o PR instead of the fork's branch: https://github.com/kubernetes-incubator/cri-o/pull/1519

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- The goal is to be able to take container images that would all want to run as the same uid (0 for root, 48 for apache, some 997 for nginx) or set of uids (systemd-based multi-daemon container which starts as uid 0 and forks a number of service running as different uids) and run them without clashing with the same uids on the host.
+- The effort will be successful when it will be possible to use any container image, built using any set of uids and run it in Kubernetes cluster safely with host/node  uids not matching those in the image. Specifically, Pods with containers that need to be run as root (uid 0) should be runnable with non-root uids.
+- To allow images using privileged users securely where such containers will run in a user namespace different from user-namespace of the underlying host.  
+
+## Non-Goals
+- Non-goal is to support use cases when the uids for processes get allocated in runtime and externally (IPA, Active Directory) for generic users, without specifying (hopefully narrow) range of uids to be mapped upfront, before the container gets started.
```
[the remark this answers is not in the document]
vikaschoudhary16: agree, i think this can be skipped all-together.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- Non-goal is to support pod/container level user namespace isolation. There can be images using different users but on the node, pods/containers running with these images will share common user namespace remapping configuration. In other words, all containers on a node share a common user-namespace range.
```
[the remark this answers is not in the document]
vikaschoudhary16: We are trying to use user-namespace feature to the level it is being supported in more than one runtimes and not just docker.  No runtime today supports passing a per-container UID range. Even `rkt` chooses a random UID range from host and provides no guarantee that those ranges will not collide. Therefore, we thought of going incrementally in adding layers of security that user-namespaces can provide.
 
With what we have today in runtimes in terms of user-namespace support, through this proposal we are trying to secure nodes from rough pod containers. Once runtimes have evolved enough to be able to allow per-container UID range, we will then design to add another layer of security to protect pods of one namespace from the pods of another namespace.

> the only real change is that root in containers is no longer "real root" but we already have PSPs that can be used to restrict running as root

Thats the point. We dont want to limit and restrict those images which are supposed to run as root user.

contributors/design-proposals/node/node-usernamespace-remapping.md
pweil-: @adelton ^ looks like it is not as clear cut as I thought.  Also, should probably record that if this is done there needs to be work to remove some of the [daemon validations](https://github.com/moby/moby/blob/65c575f592a247b3dd0d10dd707f1bbf3d459ae1/daemon/daemon_unix.go#L603-L614) that would prevent this configuration.

contributors/design-proposals/node/node-usernamespace-remapping.md
pweil-: for correctness, some others have noted that some usecases of non-host user ns and host pid/net/ipc DO work.  TIL https://github.com/kubernetes/community/pull/2067#discussion_r183835918

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+- To support Pod-Security-Policy integration is a non-goal for this KEP.
```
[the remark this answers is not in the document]
vikaschoudhary16: Because this will be required for the knob, `HostUserNamespace: true`, and that we are planning to add  beyond alpha.

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: sure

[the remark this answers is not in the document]
vikaschoudhary16: sure

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: sure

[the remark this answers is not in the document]
vikaschoudhary16: sure

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+Proposal is to support user-namespaces for the pod containers. This can be done at two levels:
+- Node-level : This proposal explains this part in detail.
+- Pod-level: Plan is to target this in future due to missing support in the low level system components such as runtimes and kernel. More on this in the later sections.
+
+Node-level  user-namespace support means that, if feature is enabled, all pods on a node will share a common user-namespace, common UID(and GID) range (which is a subset of node’s total UIDs(and GIDs)). This common user-namespace is runtime’s default user-namespace range which is remapped to containers’ UIDs(and GID), starting with the first UID as container’s ‘root’.
+
+User-namespace support should be enabled only when container runtime on the node supports user-namespace remapping and is enabled in its configuration. To enable user-namespaces, feature-gate flag will need to be passed to Kubelet like this `--feature-gates=”NodeUserNamespace=true”`
+
+A new CRI API, `RuntimeConfigInfo` will be added. Kubelet will use this API:
+- To verify if user-namespace remapping is enabled at runtime. If found disabled, kubelet will fail to start
```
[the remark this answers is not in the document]
vikaschoudhary16: This is similar behavior to how today kubelet fails if cgroup driver configuration at kubelet does not match with cgroup driver at runtime. Wondering if you have something in mind to suggest alternatively.

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: @adelton @tallclair By default kubelet will not set this variable and thus keeping behavior same as it is today when this member does not exist. Runtime will see a `nil` value. `POD` will be set by kubelet if `NodeUserNamespace` is enabled and `NODE` will be set if `NodeUserNamespace` is enabled but a pod wants to escape user-namespace.
Runtime will have clear indication about each condition without ambiguity.

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: +1

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: Well, and the question is, could we say that from now one, not having the member in the message will be considered the default for `NamespaceMode` which is `POD`, and we immediately change the code to always set it to `NODE` because after all, that's what it semantically means today.

And the for alpha / 1.11, we will add the code to set the `POD` when we detect `NodeUserNamespace` and escape to `NODE` is not requested nor allowed.

What I'm aiming for with these thoughts is to make the `user` consistent with the existing `network`, `pid`, and `ipc`, actually defining the behaviour, without that `nil` state when we say "it behaves like it always did, except we actually don't know how the Pod will be run".

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: @vikaschoudhary16, could we add a bullet point about the containerd work, to make the document more completely describe the landscape?

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: Or at least not add similar once to CRI-O without much testing.

contributors/design-proposals/node/node-usernamespace-remapping.md
estesp: As an aside, Netflix just recently released their [Titus project](https://github.com/netflix/titus) as open source; in combination with their own use of userns via the Docker engine, they have some not-yet-upstreamed patches that make NFS userns aware; one example (in a series of patches from Sargun): https://github.com/sargun/linux/commit/b6f64c0809a30731258e3eac89d691fc99f44b68 

I realize this can't be relied on yet, but just as a note that there are people working on making NFS userns aware and handling some of the mapping challenges.

contributors/design-proposals/node/node-usernamespace-remapping.md
estesp: Correct

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: @yujuhong, are you concerned with the direction of the information flow and purpose of `RuntimeConfig`, similar to https://github.com/kubernetes/community/pull/2067#discussion_r183638511?

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: @tallclair, would you recommend renaming `UserNamespaceConfig` to `LinuxUserNamespaceConfig` and potentially putting it directly to `RuntimeConfigInfoResponse`?

adelton: Ah, now I see you propose `LinuxRuntimeConfig` below.

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: If you run with host user namespace, won't docker daemon pull and create completely different set of layers?

estesp: I assumed @yujuhong was referring to the ability to override (for the `--privileged`/admin use case) with `docker run --userns=host` which starts a container using the same layers already downloaded and chown'ed as the shifted ID range. It is a bit kludgy but strongly requested to be possible for certain use cases. If you add a `--user` declaration to that run, you now have a non-shifted non-privileged user running the container process, and probably not having access to many of the files in the layers.

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: To be able to access remote volumes without permission issues, yes.

On the other hand, there can be use-cases where you decide to explicitly partition the cluster into sets of nodes with different mappings, and drive different payloads to different nodes, protecting each others remote data.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,210 @@
...
+	// Namespaces currently set by the kubelet: POD, CONTAINER, NODE
+	NamespaceMode pid = 2;
+	// IPC namespace for this container/sandbox.
+	// Note: There is currently no way to set CONTAINER scoped IPC in the Kubernetes API.
+	// Namespaces currently set by the kubelet: POD, NODE
+	NamespaceMode ipc = 3;
+	// User namespace for this container/sandbox.
+	// Note: There is currently no way to set CONTAINER scoped user namespace in the
+	// Kubernetes API. The container runtime has to run with user namespace enabled,
+	// and kubelet then forces the use of host user namespace.
```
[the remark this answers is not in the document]
adelton: It's "use" of host user namespace, not "user" of host user namespace.

contributors/design-proposals/node/node-usernamespace-remapping.md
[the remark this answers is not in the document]
adelton: @cyphar, we see the security benefits in being able to isolate the Node from the Pods in most cases, while being able to for example run systemd-based and multi-daemon containers without allowing root on the host.

While going all-in with per-Pod user namespaces would be good and it was the goal of the previous PR https://github.com/kubernetes/community/pull/2042, we've been advised that the proposal should focus on something that Container Runtimes can actually currently provide. There is an ongoing effort brewing in https://github.com/kubernetes-incubator/cri-o/pull/1519 and I've been following that and was able to specify uid/gid mapping per-Pod when using that branch of CRI-O.

Do you feel that the current proposal hinders future adoption of per-Pod user namespace support? Our hope was that it will actually validate the user namespace functionality and prepare way for future more flexible setups.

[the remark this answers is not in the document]
cyphar: > No runtime today supports passing a per-container UID range. Even rkt chooses a random UID range from host and provides no guarantee that those ranges will not collide.

LXC **definitely** does support this, and `rkt` might not provide such a guarantee but it doesn't change that they do support it too (and in principle could implement such a guarantee if they wanted to). Not to mention that `runc` and the OCI spec explicitly support this. So I disagree with this conclusion.

To be fair, `cri-o` doesn't support this but from a high-level perspective the implementation would not be *significantly* harder than the current plan of implementing user namespaces (Docker is more difficult for some more nuanced reasons -- not to mention that it is already implemented and changing it after-the-fact is much harder). It however would be significantly harder to retro-fit a pod-specific UID mapping to implementations like `cri-o` (as is evident from how difficult it would be to add that design to Docker).

> Therefore, we thought of going incrementally in adding layers of security that user-namespaces can provide.

I don't have a problem with doing this incrementally, I'm just worried that you cannot incrementally add the design I described above with this design -- you would need to create a separate design to this (and then you have two different ways of setting up user namespaces).

> Thats the point. We dont want to limit and restrict those images which are supposed to run as root user.

But there are many benefits of user namespaces other than just this one feature -- which is the point I'm trying to make. A large portion of container images don't *need* root privileges to run (and user namespaces still are very useful for images which don't run as root -- because it protects against cross-container attacks).

contributors/design-proposals/node/node-usernamespace-remapping.md
cyphar: Can someone please explain what a "named user namespace" means?

contributors/design-proposals/node/node-usernamespace-remapping.md
cyphar: @pweil- @adelton 

> Capabilities for a non-user namespace are checked against the owning user namespace so any time you're sharing a non-user namespace you need to share the owning user namespace in order to have permissions to use the non-user namespace

This is mostly true, though there are some exceptions (and there is no framework in the kernel to guarantee this, it's implemented for each namespace independently using `ns_capable`). But as a rule of thumb, yes this is pretty much true (except in the cases where you need a host capability or a capability in a parent namespace or something similar).

contributors/design-proposals/node/node-usernamespace-remapping.md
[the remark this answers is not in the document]
cyphar: @adelton 

> we see the security benefits in being able to isolate the Node from the Pods in most cases, while being able to for example run systemd-based and multi-daemon containers without allowing root on the host.

I agree that this is *a* benefit of user namespaces, but you can (in principle) get the same benefit by not running as root inside a non-user namespaced container. To be fair, there are some additional protections you get from running containers in a user namespace versus just running as an unprivileged user in the host user namespace, but they are *very* specific and nuanced (for instance the `ptrace` permission checks are more rigid and protect against some join-race attacks). I get that the main benefit of this is that it allows people who currently run containers as root to get more security, but I feel that making that *one of the few benefits* of using user namespaces in Kubernetes is restricting ourselves needlessly.

> Do you feel that the current proposal hinders future adoption of per-Pod user namespace support? Our hope was that it will actually validate the user namespace functionality and prepare way for future more flexible setups.

Yes, I believe it might (but I might be mistaken -- I've only skimmed the full design doc, I will re-read it tomorrow).

The design I outlined above -- where you by default have an unspecified user namespace mapping (so it's unique for each container with no overlaps), and for special cases like `hostPath` you specify what UIDs have to be the same in all containers that access that `hostPath` (which allows you to then have only overlaps for those UIDs and only for containers that require such access). Maybe this links into the "named user namespaces" thing that gets mentioned later (though I don't know what the term means), but this design would permit **far more** out-of-the-box security and also provides far more semantic control to users to specify what UIDs they actually want to share.

But the design I outlined requires more changes to the Kubernetes API, so it's *possible* that iteration on this design could be done to implement the design that I outlined (my main concern is whether or not the mappings are entirely controlled by Kubernetes -- because there are a lot of things that are runtime-specific that Kubernetes doesn't really know about and shouldn't care about when it comes to user namespaces).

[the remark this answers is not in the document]
mrunalp: @cyphar The end goal is definitely going towards per pod (or per namespace) user namespaces. However, we feel that we should go there step by step. Even getting this first step working seamlessly in kubernetes would take quite some effort. I think one of the principles we can follow is minimize user burden while enabling user namespaces. They simply shouldn't know or care or should have to modify their workloads or pod/resource settings. They should benefit from the additional security of user namespaces when they are enabled in the cluster. 

For the use case of per-pod user namespace, I think it would make sense to allocate a user namespace to a k8s namespace and then have all pods within that k8s namespaces share the same mappings. I can't think of anything that is being proposed as a first step here that would prevent us to move towards that in the future. But again, there is still a lot of work to do to even get this first step working nicely :)

contributors/design-proposals/node/node-usernamespace-remapping.md
mrunalp: Can't the kubelet just default to POD and set to NODE when trying to use host user namespace. The runtime can ignore the value when it has not been configured with user namespaces. We don't need a wrapper if we are okay with that.

vikaschoudhary16: This too is fine with me. @adelton @tallclair thoughts?

contributors/design-proposals/node/node-usernamespace-remapping.md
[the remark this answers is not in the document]
vikaschoudhary16: @cyphar thanks a lot and i really appreciate your inputs. Just want to clarify on this:

> . But I'm not entirely sure whether we would end up with two different ways of specifying user namespaces

In our understanding this is not going to be case and we will not end-up specifying user namespaces in two different ways. Today we are reading a single default mapping from runtime and using it for all the pods on a node. In future also, to enable pod (k8s namespace level) user-namespaces, this mapping will be read from runtime and then a sub-part will be used for each pod (or k8s namespace). Externally, there will be no change in the way to enable user-namespaces or configuring user-namespaces. Ofcourse, this future approach still need more rigorous thinking to become a complete approach, but does seem to us that we are going in a direction where we will end up into specifying user-namespaces in two different ways.

May be you have noticed something in the current design which made you think so. We would really appreciate if you point that out and help us reaching a better design.

contributors/design-proposals/node/node-usernamespace-remapping.md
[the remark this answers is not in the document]
vikaschoudhary16: Update: re-thinking on this, i think it makes sense to remove it from non-goal because it is actually part of plan.

vikaschoudhary16:
Hey folks, thanks a lot for the valuable inputs. Tried to address all, request to please point out if i missed something important.
Thanks!!

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+- Track Linux User Namespaces in the Pod Security Policy
+  - https://github.com/kubernetes/kubernetes/issues/59152
+- Add support for experimental-userns-remap-root-uid and experimental-userns-remap-root-gid options to match the remapping used by the container runtime.
+  - https://github.com/kubernetes/kubernetes/pull/55707
+- rkt User Namespaces Background
+  - https://coreos.com/rkt/docs/latest/devel/user-namespaces.html
+
+## Future Work
+### Namespace-Level/Pod-Level user-namespace support
+There is no runtime today which supports creating containers with a specified user namespace configuration. For example here is the discussion related to this support in Docker https://github.com/moby/moby/issues/28593 and support is being created for CRI-O https://github.com/nalind/cri-o/tree/idmapping.
```
kargakis: You've missed linking to the crio PR I added in my previous comment.

vikaschoudhary16: I have updated that in the `Runtime Support` section and missed here. WIll update

contributors/design-proposals/node/node-usernamespace-remapping.md
adelton: Right, that will work fine as well.

I'm not sure we want to declare that the Container Runtime is free to ignore the value (that `POD` default) though. After all, running with `--userns=host` when the kubelet thinks user namespaces are applied means that the payload runs with different security attributes and intended.

contributors/design-proposals/node/node-usernamespace-remapping.md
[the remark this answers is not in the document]
adelton: > In future also, to enable pod (k8s namespace level) user-namespaces, this mapping will be read from runtime and then a sub-part will be used for each pod (or k8s namespace).

We might be getting ahead of ourselves with this here but for the per-Pod user namespace scenarios, why would the Container Runtime userns mapping defaults play any role? I view it as a default the setups falls back to when no uid/gid mapping will be provided with the Pod (via Pod Spec or from Kubernetes namespace default). But if explicit mapping is provided, it should be able to map to any host ids it wants, unrestricted by the Container Runtime default settings.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+
+The purpose of using user namespaces in Kubernetes is to let the processes in Pods think they run as one uid set when in fact they run as different “real” uids on the Nodes.
+
+In this text, most everything said about uids can also be applied to gids.
+
+## Goals
+Enable user namespace support in a kubernetes cluster so that workloads that work today also work with user namespaces enabled at runtime. Furthermore, make workloads that require root/privileged user inside the container, safer for the node using the additional security of user namespaces. Containers will run in a user namespace different from user-namespace of the underlying host.
+
+## Non-Goals
+- Non-goal is to support pod/container level user namespace isolation. There can be images using different users but on the node, pods/containers running with these images will share common user namespace remapping configuration. In other words, all containers on a node share a common user-namespace range.
```
adelton: Another non-goal worth listing is ensuring the mapping is the same on all Nodes in the cluster. That is admin's responsiblity, outside of this proposal.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+- Remote volumes support eg. NFS
+
```
adelton: If the remote volume ownership matches the host_ids used by the Container Runtime mappins, remote volumes should work. The non-goal is any sort of changing ownership for remote volumes.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+## User Stories
+- As a cluster admin, I want to protect the node from the rogue container process(es) running inside pod containers with root privileges. If such a process is able to break out into the node, it could be a security issue.
+- As a cluster admin, I want to support all the images irrespective of what user/group that image is using.
+- As a cluster admin, I want to allow some pods to disable user namespaces if they require elevated privileges.
+
+## Proposal
+Proposal is to support user-namespaces for the pod containers. This can be done at two levels:
+- Node-level : This proposal explains this part in detail.
+- Namespace-Level/Pod-level: Plan is to target this in future due to missing support in the low level system components such as runtimes and kernel. More on this in the `Future Work` section.
+
```
adelton: Make it a link: `[Future Work](#future-work)`

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+3. Count/Size: Total number of consecutive mapping between host and container user-namespaces, starting from the first one (including) mentioned above.
+
+As an example, `host_id 1000, container_id 0, size 10`
+In this case, 1000 to 1009 on host will be mapped to 0 to 9 inside the container.
+
+User-namespace support should be enabled only when container runtime on the node supports user-namespace remapping and is enabled in its configuration. To enable user-namespaces, feature-gate flag will need to be passed to Kubelet like this `--feature-gates=”NodeUserNamespace=true”`
+
+A new CRI API, `GetRuntimeConfigInfo` will be added. Kubelet will use this API:
+- To verify if user-namespace remapping is enabled at runtime. If found disabled, kubelet will fail to start
+- To determine the default user-namespace range at the runtime, starting UID of which is mapped to the UID '0' of the container.
```
adelton: We might add here that kubelet needs this to properly prepare the local volumes and directories for the Pod.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Volume Permissions
+Kubelet will change the file permissions, i.e chown, at `/var/lib/kubelet/pods` prior to any container start to get file permissions updated according to remapped UID and GID.
+This proposal will work only for local volumes and not with remote volumes such as NFS.
+
+### How to disable `NodeUserNamespace` for a specific pod
+This can be done in two ways:
+- **Alpha:** Implicitly using host namespace for the pod containers
```
adelton: Add `.` to the end of the sentence.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+This support is already present (currently it seems broken, will be fixed) in Kubernetes as an experimental functionality, which can be enabled using `feature-gates=”ExperimentalHostUserNamespaceDefaulting=true”`.
+If Pod-Security-Policy is configured to allow the following to be requested by a  pod, host user-namespace will be enabled for the container:
+  - host namespaces (pid, ipc, net)
+  - non-namespaced capabilities (mknod, sys_time, sys_module)
+  - the pod contains a privileged container or using host path volumes.
+  - https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c#diff-a53fa76e941e0bdaee26dcbc435ad2ffR437 introduced via https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c.
+
+- **Beta:** Explicit API to request host user-namespace in pod spec
```
adelton: Add `.` to the end of the sentence.

derekwaynecarr:
For the past few days, I have been unable to render the PR in the web ui
without getting a github unicorn.

From my mail history, it appears that all comments have been addressed.

A follow-on PR is needed to align this document with the KEP process:
https://github.com/kubernetes/community/blob/master/keps/0001-kubernetes-enhancement-proposal-process.md

In particular, it needs a number, and the state (provisional,
implementable).  In this case, I think its fair to say its "implementable"
and any further discussion can happen on the follow-on PR or the
implementation.

/approve
/lgtm

contributors/design-proposals/node/node-usernamespace-remapping.md
yujuhong: Yes, @estesp that's what I meant. When per-pod configuration is supported in the future, this can cause great confusion for the users. I don't see a clear way to mitigate this...

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -0,0 +1,209 @@
...
+
+Node-level user-namespace support means that, if feature is enabled, all pods on a node will share a common user-namespace, common UID(and GID) range (which is a subset of node’s total UIDs(and GIDs)). This common user-namespace is runtime’s default user-namespace range which is remapped to containers’ UIDs(and GID), starting with the first UID as container’s ‘root’.
+In general Linux convention, UID(or GID) mapping consists of three parts:
+1. Host (U/G)ID: First (U/G)ID of the range on the host that is being remapped to the (U/G)IDs in the container user-namespace
+2. Container (U/G)ID: First (U/G)ID of the range in the container namespace and this is mapped to the first (U/G)ID on the host(mentioned in previous point).
+3. Count/Size: Total number of consecutive mapping between host and container user-namespaces, starting from the first one (including) mentioned above.
+
+As an example, `host_id 1000, container_id 0, size 10`
+In this case, 1000 to 1009 on host will be mapped to 0 to 9 inside the container.
```
krmayankk: does this mean that all containers belonging to all pods in this node, will have the same mapping ? meaning pid 0 inside any container on a given host will always map pid 1000 on the host ?

PR #2071, opened 2018-04-24
Enhance governance template

pmorie:
Clarifies some minor points that confused me while trying to sort out what the rules for governance/charters are.

/sig steering


timothysc:
per our discussion yesterday 

/assign @pwittrock 
/cc @quinton-hoole

PR #2081, opened 2018-04-26
KEP: security conformance proposal

easeway:
## Summary

Security Conformance defines citeria to certify a Kubernetes cluster to be minimally secure.

## Motivation

Configuring a Kubernetes cluster is complicated,
and it's very easy to mis-configure a cluster to be insecure.
The current Kubernetes conformance test suite barely covers security aspects.

/sig auth
/sig testing
/assign @tallclair 

@davidopp @tallclair @ericchiang @liggitt @bgrant0607

### DIFF
--- a/keps/draft-20180426-security-conformance.md
+++ b/keps/draft-20180426-security-conformance.md
@@ -1,0 +1,136 @@
+---
+kep-number: draft-20180426
+title: Security Conformance
+authors:
+  - "@easeway"
+owning-sig: sig-auth
+participating-sigs:
+  - sig-testing
+  - sig-api-machinery
+  - sig-network
+  - sig-node
+  - sig-cluster-lifecycle
+reviewers:
+  - "@tallclair"
+  - "@ericchiang"
+  - "@liggitt"
+  - "@davidopp"
+approvers:
+  - "@tallclair"
+  - "@ericchiang"
+  - "@liggitt"
+editor: "@easeway"
+creation-date: 2018-04-26
+status: provisional
+---
+
+# Security Conformance
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [List of Criteria](#list-of-criteria)
+* [Graduation Criteria](#graduation-criteria)
+
+## Summary
+
+Security Conformance defines citeria to certify a Kubernetes cluster to be minimally secure.
+
+## Motivation
+
+Configuring a Kubernetes cluster is complicated,
+and it's very easy to mis-configure a cluster to be insecure.
+The current Kubernetes conformance test suite barely covers security aspects.
+
+### Goals
+
+- Define the initial list (to evolve later) of criteria for minimal security requirements
+  - Define the role of the actor (unprivileged user, namespace admin, cluster admin, etc.)
+  - Define pre-requisites to run test for verifying the criteria
+  - Define test behaviors
+  - Define test expectations
+
+### Non-Goals
+
+- A full list of criteria at the beginning
+  - The list of criteria is to be curated by the community
+  - This proposal starts with minimal criteria to be evolved quickly
+- Mechanisms to implement the test
+- Mechanisms to setup environment for the tests
+- Host setup and control plane configurations (covered by CIS benchmark)
+- Non-security related
+
+## Proposal
+
+### List of Criteria
+
+#### API privileges of default service accounts
+
+##### Rationale
+
+Every namespace has a service account named `default`.
+This is used by Pods created without specifying a particular service account.
+When the process running in the Pod invokes Kubernetes API, 
+this `default` service account is used as the identity for authn/authz.
+
+##### Pre-requisites
+
+- A Kubernetes cluster should be up and running
+- A `kubeconfig` to allow the test to
+  - Create a Pod in a namespace
+  - It doesn't matter what the additional permissions the user have
+- A namespace created for test
+  - This can be `default` namespace, but not `kube-system`
+  - The `default` service account in the namespace should not be granted extra permissions
+- The DNS name of API server inside the cluster, if it's not default
+
+##### Test Behavior
+
+- Create a Pod in the test namespace without specifying a service account (using `default`)
+- The Pod contains test logic which invokes Kubernetes API to
+  - `get/list/watch` namespaces
+  - `create` another Pod in the namespace
+  - `create` a new namespace
+  - `update/patch` labels/annotations of current Pod
+  - `delete` current Pod
+
+##### Test Expectation
+
+- All attempts using Kubernetes API should fail (Forbidden).
+
+#### Kubelet r/w API Access
+
+##### Rationale
+
+Kubelet read/write API (listening on port 10250) is designed to be consumed by API server.
+It's a security concern if other components, 
+including the workload running on the cluster is able to access the API.
+
+##### Pre-requisites
+
+- A Kubernetes cluster should be up and running
+- A `kubeconfig` to allow the test to
+  - Create a Pod in a namespace
+- A namespace for test Pod
+- The DNS name of API server inside the cluster, if it's not default
+- The DNS name/IP of Kubelet to be tested
+
+##### Test Behavior
+
+- Create a Pod in the test namespace running the test code
+- The test code in the Pod attempts to access R/W API on Kubelet directly
+  - `GET /exec/{namespace}/{currentPodID}/{currentContainerName}`
+
+##### Test Expectation
+
+- The API invocation should fail.
+
+## Graduation Criteria
+
+- The list of criteria is agreed by community
+- The tests are integrated into Kubernetes conformance test suite


keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
+---
+kep-number: draft-20180426
+title: Security Conformance
+authors:
+  - "@easeway"
+owning-sig: sig-auth
+participating-sigs:
+  - sig-testing
```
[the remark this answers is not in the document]
easeway: Yes, I will add related sigs.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+not among user workloads (e.g. single-tenancy model);
+- `Isolated`: in addition to `Restricted`, it covers the threaten among user workloads (e.g. multi-tenancy model).
+
+### Behavioral Criteria
+
+#### Config File Permissions
+
+  *Conformance Profile: Essential*
+
+  The configuration files for API server, controller manager, scheduler and kubelet should have the correct setup of ownership and file permissions.
```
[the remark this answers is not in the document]
ericchiang: ...or self-hosted setups where the control plane configuration is in the Kubernetes API itself

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+- A list of non-behavioral security criteria that can be validated by tests
+
+### Non-Goals
+
+- Mechanisms to satisfy security criteria
+- Details about Kubernetes conformance test
+
+## Proposal
+
+### Conformance Profiles
```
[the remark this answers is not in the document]
easeway: This proposal is security focused. Other profiles are out-of-scope. So "conformance profiles" is not a good name. I feel using "security conformance profiles" will be better. Regarding testing against restricted/isolated profiles, I have some details in the Implementation Details section. I will refine the content according to your question.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+The behavioral/non-behavioral criteria are categorized into a few conformance profiles for different levels of security requirements:
+
+- `Essential`: the baseline security requirements for the cluster,
+focusing on the threaten from outside the cluster (via APIs),
+non-privileged (non-root) access on the host running master and kubelet,
+not covering the threaten from user workloads (non-critical, non-system Pods) running inside the cluster.
+The criteria in this category must always be satisfied;
```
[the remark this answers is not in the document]
easeway: I will further clarify in the doc. This is an optional profile if security is optional. Otherwise it's mandatory.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+  - non-root user should not be able to access etcd data directory
+  - non-root user should not be able to access any kubeconfig files
+  - non-root user should not be able to access any private key files
+
+#### Anonymous Access
+
+  *Conformance Profile: Essential*
+
+  Anonymous access to Kubernetes API should be denied as unauthorized.
```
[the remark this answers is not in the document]
easeway: I think denying anonymous access is unreasonable. I will remove this. Because at least for healthz checks, it should allow anonymous access, from external monitoring systems. For discovery, swagger/openapi, it's OK to allow anonymous access. For the rest of APIs, as long as authorizers will block "system:unauthorized", it won't be a security concern.

ericchiang:
cc @kubernetes/sig-auth-proposals

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### API privileges of default service accounts
+
+  *Conformance Profile: Essential*
+
+  Accessing Kubernetes API using default service accounts from any namespace should be rejected as forbidden.
```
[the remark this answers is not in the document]
easeway: I will rephrase this. The idea behind that is the service account named `default` should have minimum privilege. I will spec the _minimum privilege_ for discussion.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Kubelet r/w API Access
+
+  *Conformance Profile: Essential*
+
+  Anonymous access to Kubelet r/w APIs should be denied as unauthorized.
+  Default service accounts should also be denied as forbidden accessing Kubelet r/w APIs.
```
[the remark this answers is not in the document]
easeway: I will further clarify this.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Privileged Containers
+
+  *Conformance Profile: Restricted*
+
+  Creation of Pods with privileged containers should not be allowed.
```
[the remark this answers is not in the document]
easeway: Agree. In general, the context applies to pod creation in a namespace without explicit RBAC rules granting specific permissions (or use a privileged PodSecurityPolicy). I will further clarify this.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+  *Conformance Profile: Restricted*
+
+  Creation of Pods with containers requesting any host namespaces (user, ipc, pid, net, uts) should not be allowed.
+
+#### Container Capabilities
+
+  *Conformance Profile: Restricted*
+
+  Pods are created with all capabilities dropped.
```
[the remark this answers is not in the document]
easeway: I agree. I researched a little about existing Linux capabilities, and opinionated dropping ALL will be reasonable. I'd love to hear feedback.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Non-behavioral Criteria
+
+#### API server configuration
+
+  *Conformance Profile: Essential*
+
+  The following are required:
+
+  - All communication over TLS
```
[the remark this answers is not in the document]
easeway: Yes. I will update this to be more accurate.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+- Host configuration validation
+
+  The tests will need to check the configurations on the host running master or kubelet.
+  The accessibility to the host is required.
+  There are a few options to achieve that:
+
+  - Create a Pod with special privilege to run with host namespaces, like pid
+  and bind mount host file system into testing container for inspection.
+  With host pid namespace and host file system,
+  the test is able to verify the command line flags and file permissions.
```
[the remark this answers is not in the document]
easeway: I will put more details in Implementation Details section to cover these issues.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+  This is the most consistent way with other tests without requiring user to perform additional operations before running the tests.
+
+  - Remote access to master, like SSH.
+  This is the alternative way if it's impossible to schedule a Pod on master (the kubelet on master doesn't register).
+  The verification part of the testing code can be reused,
+  however the orchestration part (test setup) will be separate.
+
+- API validation
+  Similar to existing test cases, directly calling Kubernetes APIs. Certain API validations are performed by running a Pod inside a namespace, with default service account.
```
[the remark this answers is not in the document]
easeway: The service account named `default`. I will clarify this.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Config File Permissions
+
+  *Conformance Profile: Essential*
+
+  The configuration files for API server, controller manager, scheduler and kubelet should have the correct setup of ownership and file permissions.
+
+  - non-root user should not be able to access etcd data directory
+  - non-root user should not be able to access any kubeconfig files
+  - non-root user should not be able to access any private key files
```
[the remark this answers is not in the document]
easeway: Good point. I will update.

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
easeway: Yes. I will put more details.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+  Creation of Pods with containers run as root (UID=0) should not be allowed.
+
+  _Note: this may change if a better isolation runtime (e.g. VM) is used for the container._
+
+#### Network Isolation
+
+  *Conformance Profile: Isolated*
+
+  Ingress/Egress are blocked for all traffic by default in namespaces.
+
```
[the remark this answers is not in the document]
easeway: Good point. I will add these in.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,213 @@
...
+  - use service account credentials
+
+#### Addons
+
+  *Conformance Profile: Essential*
+
+  The following addons should be disabled:
+
+  - dashboard
```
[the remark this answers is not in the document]
easeway: I'm still uncertain if addons should be covered in security conformance, as user may install any addons. Maybe the best way for now is to exclude addons from security conformance.

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
easeway: @tallclair could you comments on Kubelet r/w APIs? I don't have enough context here.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,465 @@
...
+approvers:
+  - "@tallclair"
+  - "@ericchiang"
+  - "@liggitt"
+editor: "@easeway"
+creation-date: 2018-04-26
+status: provisional
+---
+
+# Security Conformance
```
davidopp: I like this proposal, but it seems seems like there are some things missing, for example which admission controllers should be enabled. Basically more of the stuff discuss in
https://docs.google.com/document/d/1jAcsC4sLgEV9__TdgJrMvPa3G73G62tFtMcKQgeIlHM/edit
(though I think you did hit most of the things mentioned in that doc).

BTW I would suggest adding a section discussing the tradeoff between having the test just directly check whether a particular policy is installed, vs. check that the behavior dictated by the policy is actulally enforced. It seems you chose the latter; I think that's fine but it would be good to justify that choice.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,465 @@
...
+- Non-security related
+- Details about Kubernetes conformance test
+
+## Proposal
+
+### Security Conformance Profiles
+
+The behavioral/non-behavioral criteria are categorized into a few security conformance profiles for different levels of security requirements:
+
+- `Essential`: the baseline security requirements for the cluster,
```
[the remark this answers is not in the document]
easeway: The three definitions of _Security Conformance Profiles_ are based on where the threat comes from, and who will be impacted. 

Generally, the threat comes from:

1. Outside of the cluster: the attack surface will be K8S APIs, the hosts including master and nodes.
2. Inside the cluster: the workload running on K8S cluster

The impact will be on:
1. The cluster: the cluster may not function or leaking information
2. The workload: the cluster is still functioning, and secure, while some workload will be compromised, and may compromise other workload

_Essential_ is actually defined for _trusted_ workloads: it focuses on the threat from outside of the cluster and impact on the cluster. It will assume the workload is trusted, and allow privileges on workload. The common use case is a K8S cluster used internally in a team, or an org with all trusted users.

_Restricted_ is based on _Essential_, and additionally targets the threat from inside. The targeted impact is still on the cluster, not workloads. The common use case will be for untrusted workloads, regardless of tenants. For example, the operators of K8S cluster is responsible for keeping the cluster secure and stable, not considering the influence between workloads.

_Isolated_ based on _Restricted_, targets threat from both inside and outside, impact on both cluster and workload. It's for multi-tenancy use cases. It will prevent the attack from some tenants' workload to other tenants.

I will give more details on the definitions - mentioning use case categories.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,465 @@
...
+focusing on the threaten from outside the cluster (via APIs),
+non-privileged (non-root) access on the host running master and kubelet,
+not covering the threaten from user workloads (non-critical, non-system Pods) running inside the cluster.
+The criteria in this category must be satisfied for a Kubernetes cluster to be secure minimally;
+- `Restricted`: in addition to `Essential`,
+it covers the threaten from user workloads to the cluster,
+not among user workloads (e.g. single-tenancy model);
+- `Isolated`: in addition to `Restricted`, it covers the threaten among user workloads (e.g. multi-tenancy model).
```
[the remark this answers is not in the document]
easeway: Could you give a little more details about "soft isolation" and "hard isolation"?

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
easeway: I will move the section up here. The command line flags should be defined in security profiles. Different profiles define the command line flags differently. I will list them explicitly here according to _Essential_, _Restricted_ and _Isolated_.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,465 @@
...
+  _TODO: need more context_
+
+  Anonymous access to Kubelet r/w APIs should be denied as unauthorized.
+  Default service accounts should also be denied as forbidden accessing Kubelet r/w APIs.
+
+#### Privileged Containers
+
+  *Conformance Profile: Restricted*
+
+  Pod creation with `privileged` set to `true` in `SecurityContext` (Pod or container level)
```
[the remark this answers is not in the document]
easeway: It's true today, but not because of `privileged` in `SecurityContext`. Once PodSecurityPolicy admission control is enabled, no Pod will be created if there's no RBAC rolebinding/clusterrolebinding is associated with the Pod service account.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,465 @@
...
+  Test teardown
+
+  - Delete the namespace
+
+#### Network Isolation
+
+  *Conformance Profile: Isolated*
+
+  Ingress/Egress are blocked for all traffic for all Pods by default in namespaces,
+  except `kube-system` and those with special configurations.
```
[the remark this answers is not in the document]
easeway: `kube-system` is not touched, and left as-is. non-kube-system (matching namespace selector) pods, by default, are not allowed to talk to any services in other namespaces, including those in `kube-system` - not sure if it's too restricted, like shall we open dns to all namespaces, but I think K8S API can be blocked by default - open for discussion.

keps/draft-20180426-security-conformance.md
easeway: Yes. I will add a section in alternatives.

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
easeway: according to discussion, the `isolation` needs to be split into fine grained with different container runtime. E.g. plain container vs HyperVisor based vs gVisor etc.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+  - Delete the namespace
+
+#### Container Privilege Escalation
+
+  *Conformance Profile: Restricted*
+
+  Pod creation with `allowPrivilegeEscalation` set to `true` in `SecurityContext` (Pod or container level)
+  should fail if there's no explicit RBAC associated with the Pod service account
+  (`default` or explicitly specified in `ServiceAccountName`).
```
tallclair: I don't think we should go too far down the rabbit whole of testing pod security policy. If the goal is to test that there is a restricted PSP by default, I think we should just test that directly.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  Test teardown
+
+  - Delete the namespace
+
+#### Run as root
+
+  *Conformance Profile: Isolated*
+
+  Containers in Pod created without explicit RBAC associated with the Pod service account,
+  should run as non-root user and group.
```
tallclair: > and group.

Run as group is an alpha feature.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  Test teardown
+
+  - Delete the namespace
+
+#### Network Isolation
+
+  *Conformance Profile: Isolated*
+
+  Ingress/Egress are blocked for all traffic for all Pods by default in namespaces,
+  except the Egress to access `kube-dns`.
```
tallclair: I'm not sure this makes sense as the default option. But maybe someone from networking has more opinions.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+  Test setup
+
+  - Create two namespaces, ns1 and ns2
+
+
+  Test steps
+
+  - Create two Pods in ns1
+  - Check the output to make sure they can communicate with each other
```
tallclair: The description said that all traffic is blocked, but it sounds like you meant cross-namespace traffic?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  - Create two Pods, one in ns1, and the other in ns2
+  - Check the output to make sure they can't talk to each other.
```
tallclair: Verify that pods can't reach Kubelet RO ports?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+The criteria described below is based on a namespace with default settings.
+Like a newly created namespace, with only service account `default`.
+Namespaces with special/privileged settings, like `kube-system`,
+or special namespaces required by some system components
+(e.g. nginx ingress controller) are not covered.
+
+### Behavioral Criteria
+
+#### Config File Permissions
```
tallclair: Seems like this belongs under non-behavioral?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+  The following should be disabled:
+
+  - insecure listening port
+  - basic-auth
+  - token-auth-file
+  - AlwaysAllow authorizer
+  - Admission control:
+    - AlwaysAllow
+    - NamespaceAutoProvision
```
tallclair: why?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  - profiling
+
+  The following are required:
+
+  - RBAC authorizer
+  - Admission controls:
+    - PodSecurityPolicy
```
tallclair: I don't think PSP is an essential controller.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+    - NamespaceLifecycle
```
tallclair: Why? How is this security related?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+### Goals
+
+- A list of behavioral security criteria that can be validated by tests
+- A list of non-behavioral security criteria that can be validated by tests
+
+### Non-Goals
+
+- Mechanisms to satisfy security criteria
+- Non-security related
+- Details about Kubernetes conformance test
```
tallclair: Specifically, this seems to be totally separate from the Kubernetes conformance tests. Maybe it needs another name then?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  Same as [API server configuration](#api-server-configuration)
+
+#### API server configuration - Restricted
+
+  *Conformance Profile: Restricted, Isolated*
+
+  The following admission controls are required:
+  - PodSecurityPolicy
+  - NamespaceLifecycle
+  - NamespaceExists
```
tallclair: Deprecated? https://github.com/kubernetes/kubernetes/issues/31195

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  - DenyEscalatingExec
+  - DenyExecOnPrivileged
```
tallclair: This is redundant w/ DenyEscalatingExec

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  - EventRateLimit
+  - LimitRanger
+  - NodeRestriction
+  - PodTolerationRestriction
+  - AlwaysPullImages
```
tallclair: why?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+#### API server configuration - Restricted
+
+  *Conformance Profile: Restricted, Isolated*
+
+  The following admission controls are required:
+  - PodSecurityPolicy
+  - NamespaceLifecycle
+  - NamespaceExists
+  - DenyEscalatingExec
```
tallclair: Maybe for isolated, but I don't think this is required for restricted.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  - LimitRanger
+  - NodeRestriction
+  - PodTolerationRestriction
+  - AlwaysPullImages
+
+#### Use Sandboxed Isolation
+
+  *Conformance Profile: Sandboxed*
+
+  The container runtime must be something using HyperVisor or gVisor.
```
tallclair: I would just say TBD here, since we haven't defined sandboxes in k8s yet.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+  Same as [API server configuration](#api-server-configuration), the command line
+  of `kubelet` should be verified to use the correct container runtime.
+  The list of supported container runtimes is TO BE DETERMINED.
+
+## Graduation Criteria
+
+- Conformance Profiles are well defined and finalized
+- Definition of security criteria are curated and documented
+- Tests are added to conformance test suite to cover security criteria and categorized according to conformance profiles
```
tallclair: Which milestone are these for?
How will we measure success?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+## Alternatives
```
tallclair: Can you discuss how this compares with the CIS benchmark, and linters like [kube-bench](https://github.com/aquasecurity/kube-bench)

keps/draft-20180426-security-conformance.md
easeway: Security Conformance can be validated using tests which will be contributed as part of Kubernetes conformance test.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+Like a newly created namespace, with only service account `default`.
+Namespaces with special/privileged settings, like `kube-system`,
+or special namespaces required by some system components
+(e.g. nginx ingress controller) are not covered.
+
+### Behavioral Criteria
+
+#### Config File Permissions
+
+  *Conformance Profile: Essential*
```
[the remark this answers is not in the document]
easeway: "Essential" is defined "focusing on the threaten from outside the cluster (via APIs) and non-privileged (non-root) access on the host running master and kubelet". So local attacks are included. The way the profiles are categorized is based on:

- The source of threats: is workload involved (_Restricted_ or higher), or not (_Essential_)
  not involving workload means the threat can compromise the cluster without a particular (normally malicious) workload to be scheduled
- The components to be protected: cluster only (_Essential_, _Restricted_), or both cluster and workloads (_Isolated_, _Sandboxed_)
  cluster only means cluster is protected from being compromised, while it's possible the workloads will be compromised
  both means cluster is protected, and workloads are also protected from being compromised from other tenant or Pod (_Sandboxed_ profile).

I will add this after the definition of profiles for better explanation.

keps/draft-20180426-security-conformance.md
easeway: It can be behavioral, like we can clearly define the behavior of threat if file permissions are not properly set. I will update the description to focus on behaviors.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+  The configuration files for API server, controller manager, scheduler and kubelet should have the correct setup of ownership and file permissions.
+
+  - etcd data directory is owned and accessible only by the user running etcd;
```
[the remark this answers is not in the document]
easeway: Yes. Basically all the tests requiring host access (command line flags, file permissions) can be opt-out, as users may be limited to access the hosts.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  - kubeconfig files for control plane is owned and accessible only by the user runs the control plane;
+  - private key files is owned and accessible only by the user who runs the process which consumes the key;
+  - kubeconfig for administration purpose is owned and accessible only by the admin user/group.
```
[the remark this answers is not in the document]
easeway: I will explain in the better way. This is for the file `/etc/kubernetes/admin.conf` which contains credentials to access the cluster as cluster-admin.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  overriding mechanism (e.g. via environment variable) must be supported by test code.
+
+  With the processes identified by the names,
+  test code will be able to validate command line flags by looking into `/proc` file system (must run in host pid namespace), and also find out the user running these processes.
+  From the command line flags, the test code will identify the config/cert/key files being used,
+  and validate the file permission from there.
+
+#### API privileges of default service accounts
+
+  *Conformance Profile: Essential*
```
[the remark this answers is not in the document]
easeway: This should be essential. The attack surface is on API. My intentional definition for the threat coming from workload is non-API based. I will refine the criteria defining profiles.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+  Every namespace has a service account named `default`.
+  Without explicit permission grant from RBAC RoleBinding or ClusterRoleBinding,
```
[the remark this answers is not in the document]
easeway: Yes. No RBAC rules should be defined for the `default` service account, as the _default_ scenario.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+  This service account has the minimum privilege to access Kubernetes API,
+  it's forbidden for the service account to `get/list/watch/create/update/patch/delete` any of API objects
+
+###### Test Recommendation
+
+  The test code consumes Kubernetes API as a serviceaccount/user with privileges:
+
+  - Create/Delete a namespace
+  - Impersonate the `default` service account in the new namespace
```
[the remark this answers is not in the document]
easeway: Yes. That's closer to reflect the use case.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,532 @@
...
+
+  - Create a new namespace
+
+
+  Test steps  
+
+  - Create a ReplicaSet, in PodTemplate, without specifying `SecurityContext`, `ServiceAccountName`, without any volumes
+  - Poll the ReplicaSet and the namespace, and a new Pod should be created
+  - Create a ReplicaSet, in PodTemplate, specifying `privileged` as `true` in `SecurityContext`
+  - Poll the ReplicaSet, should got `ReplicaFailure` condition from `status`, and no Pod is created
```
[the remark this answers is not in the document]
easeway: Yes.

keps/draft-20180426-security-conformance.md
easeway: I'm thinking a simple checking is more from implementation perspective, not reflecting the behavior or scenario. The test should make sure, in a behavioral way, all the pieces are setup properly, including PSP, RBAC etc.

keps/draft-20180426-security-conformance.md
easeway: So, shall we keep it out for now? We can add back later when the feature is in beta or GA.

keps/draft-20180426-security-conformance.md
easeway: I'm not sure either. But I feel reasonable, as all workloads are running on the same overlay network - they are logically connected together. There are bunch of attacking techniques over network.

keps/draft-20180426-security-conformance.md
easeway: Oh, yes. I will clarify.

keps/draft-20180426-security-conformance.md
easeway: Good point!

keps/draft-20180426-security-conformance.md
easeway: Agreed, I will split this.

keps/draft-20180426-security-conformance.md
easeway: Added in Alternatives.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+
+- `Essential` (aka `Trusted`): the baseline security requirements for the cluster.
+
+  It covers the threat sources of:
+  - Use of Kubernetes APIs
+  - Host access on Kubernetes masters and nodes
+  - Access prohibited information
+
+  It protects the components of:
+  - Kubernetes cluster
```
tallclair: Cluster isn't well defined. I think it's worth enumerating the components here.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+  to be considered secure at minimal.
+
+- `Restricted` (aka `Untrusted`):
+
+  In addition to `Essential`, it covers the threat sources of:
+  - Malicious workloads (not through Kubernetes API)
+
+  This profile limits the compatibility/capabilities to workloads to prevent
+  them from compromising the cluster.
+  It's still possible a malicious workload will compromise other workloads.
```
tallclair: How are you defining workload here? If you can compromise a privileged workload, that's a pretty straightforward path to a full cluster compromise.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+  It's best for the use case to operate a secure Kubernetes cluster and share
+  with untrusted users without taking multi-tenancy into consideration.
```
tallclair: I'm confused by this. Isn't sharing the cluster with untrusted users what we call "hard multitenancy"?

keps/draft-20180426-security-conformance.md
tallclair: I mean we have e2e tests for PSP, RBAC, etc. If those tests don't provide sufficient coverage of the functionality, we should fix those tests. I think here you should just test 1) expected PSP is configured, and 2) PSP is enforced (a single check is sufficient).

Although actually, now that I say that I realize that you should be able to run a conformant cluster without using PSP - e.g. if you're enforcing that policy with a third party controller like OPA.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+  In addition to `Restricted`, it protects workloads from being compromised by
+  other workloads across the boundary of another tenant.
+
+  It's for the use case supporting multi-tenancy by sharing a Kubernetes cluster.
+
+- `Sandboxed`:
+
+  Similar to `Isolated`, but using a different container runtime for better isolation,
+  and possibly offering better compatibility and capabilities.
+  For example, using hypervisors, gVisor etc. which provides an additional layer of protection.
```
tallclair: How are you going to test this? We've been talking about how to define sandboxes concretely, and that's proving to be difficult on its own. I'm not sure it's going to be possible to test for those properties.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+  - etcd data directory is owned and accessible only by the user running etcd;
+  - kubeconfig files for control plane is owned and accessible only by the user runs the control plane;
+  - private key files is owned and accessible only by the user who runs the process which consumes the key;
+  - kubeconfig for administration purpose (if present, e.g. `/etc/kubernetes/admin.conf`)
+    is owned and accessible only by the admin user (perm 0400)/group (perm 0440)
+    (the test doesn't verify particular user/group, but only the permissions).
+
+
+  Notes:
+  - the above statements exclude _root_ user who always have full access to the system.
```
tallclair: This statement isn't accurate. A root user running unconfined with full privileges in the "root" namespace has full access.

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
tallclair: I'm still not convinced that this section is going to be useful as a conformance test. I think this should be discussed with @kubernetes/sig-cluster-lifecycle-pr-reviews - as @liggitt mentioned we haven't tested detailed configuration like this in the past.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+    E.g. creating a Pod using DaemonSet.
+    This is the most consistent way with other tests without requiring user to perform additional operations before running the tests.
+
+    - Remote access to master, like SSH.
+    This is the alternative way if it's impossible to schedule a Pod on master (the kubelet on master doesn't register).
+
+  The test code will inspect the master/node system by first looking up the processes by names.
+  The names are default to `etcd`, `kube-apiserver`, `kube-controller-manager`, `kube-scheduler`, `kubelet`.
+  In the case when the cluster admin changes these names,
+  overriding mechanism (e.g. via environment variable) must be supported by test code.
```
tallclair: I'm worried you're going to end up needing to define the full cluster API to fully configure these tests...

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
tallclair: I mean this should just say "Without an explicit authorization grant, this service account doesn't has minimal privileges on the Kubernetes API"

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+  This test or part of the test can be opt-out in the case host access is impossible.
+
+#### API privileges of default service accounts
+
+  *Conformance Profile: Essential*
+
+  Every namespace has a service account named `default`.
+  Without explicit permission grant from RBAC RoleBinding or ClusterRoleBinding,
+  This service account has the minimum privilege to access Kubernetes API.
+  The cluster is vulnerable if a malicious Pod can be scheduled with default service account
```
tallclair: nit: "The cluster is vulnerable..." depends on the threat model. I'd just leave this out.

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
tallclair: (still needs update)

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+
+  - Create a Pod with empty `SecurityContext` in the namespace
+  - The Pod will run a container trying to connect kubelet port 10250 and call
+    API GET `/exec/{namespace}/{podID}/{containerName}`
+  - The request should be rejected
+
+
+  Test teardown
+
+  - Delete the namesp
```
tallclair: nit: I think listing the test teardown here is too detailed for this proposal.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+  - Delete the namespaces
+
+#### API server configuration
+
+  *Conformance Profile: Essential*
+
+  The following are required:
+
+  - All communication over mTLS/TLS
+    - apiserver talks to etcd via HTTPS
```
tallclair: Why? If it's talking over a local socket this shouldn't matter.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+
+  - Delete the namespaces
+
+#### API server configuration
+
+  *Conformance Profile: Essential*
+
+  The following are required:
+
+  - All communication over mTLS/TLS
```
tallclair: nit: just TLS

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+  - token-auth-file
+    threat: weak authentication
+  - AlwaysAllow authorizer
+    threat: it by-pass authorizers
+  - Admission control:
+    - AlwaysAllow
+      threat: it by-pass admission controls
+    - NamespaceAutoProvision
+      threat: it allows accidental namespace creation
+  - profiling
```
tallclair: Why?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+
+  The following are required:
+
+  - RBAC authorizer
```
tallclair: I don't think we should require RBAC. We're already testing for authorization behavior, as long as that's enforced, do we care about the mechanism?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,682 @@
...
+    threat: without this, a pod with privileged security context,
+    host namespaces,
+    volumes with insecure volume types (e.g. host), etc may be scheduled
+  - DenyEscalatingExec
+    threat: a potential attack can use exec/attach command to connect to a privileged pod
+  - EventRateLimit
+    threat: API server may be flooded by event requests
+  - LimitRanger
+    threat: unconstrainted access to LimitRange objects
+  - NodeRestriction
```
tallclair: I think this should be moved to essential.

keps/draft-20180426-security-conformance.md
easeway: Thanks!

keps/draft-20180426-security-conformance.md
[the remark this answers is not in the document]
easeway: Here describes the desired permissions. Below I put an alternative: schedule a pod in the namespace to perform above operations.

keps/draft-20180426-security-conformance.md
easeway: I mean mTLS or TLS. It's OK to setup with stronger security both sides authenticate each other's certificates.

keps/draft-20180426-security-conformance.md
ericavonb: I think this might be trying to say it's the best case given the current capabilities of Kubernetes and containers, and does not address additional hardening that multi-tenancy might require to be fully secured.

easeway:
After discussion with @bgrant0607 and @tallclair, this proposal will focus on Kubernetes conformance. I'm updating the proposal to be decoupled from the Security Profile proposal.

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Summary
+
+Security Conformance defines citeria to certify a Kubernetes cluster to be minimally secure.
+
+## Motivation
+
+Configuring a Kubernetes cluster is complicated,
+and it's very easy to mis-configure a cluster to be insecure.
+The current Kubernetes conformance test suite barely covers security aspects.
```
mayakacz: Is this a problem we should be addressing?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Goals
+
+- Define the initial list (to evolve later) of criteria for minimal security requirements
```
mayakacz: is this tied 1:1 to the security profiles? i.e. is there a minimal security profile that would pass these tests? which one?

keps/draft-20180426-security-conformance.md
```diff
@@ -0,0 +1,136 @@
...
+  - The list of criteria is to be curated by the community
+  - This proposal starts with minimal criteria to be evolved quickly
+- Mechanisms to implement the test
+- Mechanisms to setup environment for the tests
+- Host setup and control plane configurations (covered by CIS benchmark)
+- Non-security related
+
+## Proposal
+
+### List of Criteria
```
mayakacz: Is this a complete list, or a starter list? (It seems you wanted to design the list of criteria elsewhere - where is that happenign?)

keps/draft-20180426-security-conformance.md
easeway: No. Security profile is more implementation specific, as it defines non-portable implementation specific rules like RBAC objects.

keps/draft-20180426-security-conformance.md
easeway: It's a starter list.

easeway:
The scope of Security Profile was changed. This is no longer in scope. I will close this, and probably reopen if we decide to put this back.

PR #2085, opened 2018-04-27
Move Kubelet CNI built-in module to a DaemonSet process

Lion-Wei:
This proposal is to move CNI built-in modules to a Daemonset process, and use gRPC interface to implement the invoke from kubelet to cni driver. 

We have been talk about this issue in this doc for a while:  https://docs.google.com/document/d/16WBndDxQ3eNTWmGxmZ6_dtzfLIDL-aFq8JkBilgMRHg/edit?usp=sharing 

Send this pr for more opinion or suggestion.

### DIFF
--- a/contributors/design-proposals/network/grpc-cni-plugin.md
+++ b/contributors/design-proposals/network/grpc-cni-plugin.md
@@ -1,0 +1,268 @@
+# Move Kubelet CNI built-in module to a DaemonSet process
+
+Author: @Lion-Wei @m1093782566
+
+## Background & Motivations
+
+CNI drivers are independent binaries and CNI API is understood by a kubelet built-in module, which is linked, compiled, built, and shipped with kubelet. Third party network plugins have to provide an executable CNI plugin binary and a network configuration file. This is undesirable for many reasons:
+
+- CNI module is coupled and dependent on Kubernetes releases.
+- Added bunch of CNI and network related flags to kubelet: cni-bin-dir/cni-conf-dir/network-plugin,etc.
+- Network plugins require write access to the root filesystem of the nodes in order to provide CNI plugin binary and network configuration file.
+- Move some more net config out of kubelet (implied APIs via iptables are bad)
+- There are places where kube-proxy could do better configuration if if had access to information that only the CNI drivers know. For example, the name that veth interfaces are given or the correct CIDRs for pods on that node.
+- Kube-proxy(service driver) is keeping doing so many low-level things, it would be better if the concepts of network driver and service "driver" should be closer to each other.
+- Use kube mechanisms to manage plugins (consistent with most other plugins)
+
+Beyond just the CNI plugins, Kubernetes implements Services, with kube-proxy being the default and most prevalent implementation. Both CNI plugins and kube-proxy are manipulating aspects of the network configuration. There are assumptions that kube-proxy makes about how networking works, which may not be true if the CNI plugins are changed.  There are assumptions that kube-proxy COULD make if it had more intimate knowledge of the CNI drivers in use.
+
+As we talked in kubernetes-sig-network, topic “Combining CNI + service”, it was proposed to build a DaemonSet responsible for container network and “service network”. Either combine CNI driver with kube-proxy or make CNI driver work with third-party provided network plugin. This is a long term proposal, need further discussion.
+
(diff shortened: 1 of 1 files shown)


thockin:
I don't think we should characterize this as CNI.  Or rather, that's a BIG decision that CNI people need to help make.  I think the only reference this doc should make to CNI is that we can produce an implementation of this gRPC API that calls existing CNI drivers.

I'd like to consider Services and maybe NetworkPolicy at the same time, though I *think* we can continue to think about those orthogonally. It might be possible to lump a Services implementation into the same DS or to run them as 2 DS's.

Need to think about local state for things like IPAM across upgrades of the DaemonSet.  HostPath is ugly for this.  I have long thought about how to have a daemonset ask for a local directory but not exactly which directory.  More like a local volume.  Maybe we can use local volumes for this or maybe we need a different primitive.  Like a PersistentEmptyDir.  BUt then we have lifecycle issues.  Maybe a PersistentTmpDir - it is guaranteed to stick around until the node reboots?  We should think about this.

This necessarily intersects with multi-network, which is a challenge.  The good news is that a CNI implementation (not the API itself) can consume a configmap (or N of them) to populate the networks.

Overall this captures the ideas we have discussed so far, so thanks.  I'd like other eyes to read it over and nit-pick for intent, and then we could move to a KEP?

PTAL @bowei @dcbw @caseydavenport 

@saad-ali if you have a minute, PTAL for comparison to CSI experience.

contributors/design-proposals/network/grpc-cni-plugin.md
```diff
@@ -0,0 +1,268 @@
...
+We will build a new repo for out-tree CNI driver, maybe named `kubernetes-cni`, which will run as a DeamonSet program,  and offer a gRPC server that in-tree network plugin can invoke.  This CNI driver will acts as a bridge between third-party CNI plugin binaries and kubelet.
+
+And the DaemonSet CNI driver will use a configmap as the network config file, so don’t have to access the root filesystem of the node or master.
+
+```go
+type cniDriver struct {
+  // cni.NetworkConfig is the network config specified by
+  // containernetworking/cni
+  // CNI driver will build this config according to network
+  // information from configmap
```
caseydavenport: If this is coming from a config map, does it mean we'll require this CNI config to be the same across all nodes? What about node-specific configuration? I think this would break a number of plugins (e.g. host-local) where per-node config is passed to the plugin, or it would require one configmap per-node (super ick).

Another example is Calico, where it supports various configuration options in the CNI network config file that some users want to be able to set on a per-node basis. I think requiring that a configmap be used to distribute the CNI network configuration might be too strict. Certainly that is one way it can be done, but in Calico for example we often use a network config _template_ in a ConfigMap, which each node can populate with its own node-specific data.

contributors/design-proposals/network/grpc-cni-plugin.md
```diff
@@ -0,0 +1,268 @@
...
+          volumeMounts:
+            - mountPath: /etc/cni/net.d
+              name: cni-conf
+              readOnly: false
+            - mountPath: /host/opt/cni/plugin
+              name: cni-plugin
+              readOnly: false
+      volumes:
+        # cni plugin binaries path
+        - hostPath:
```
caseydavenport: Perhaps I misunderstood some of the goals, but I thought we were trying to remove the need to have a host path volume mount in order to install CNI plugins.

Couldn't we run two containers in the pod - one with the "kubernetes-cni" image and one with the third-party CNI binaries and config, with sharing done using an emptyDir volume?

contributors/design-proposals/network/grpc-cni-plugin.md
```diff
@@ -0,0 +1,268 @@
...
+  // cniDriver will use this interface to Add/Del network
+  cniConfig     cni.CNI
+
+  // cniPluginDir is the path of third-party CNI plugin binaries
+  // which from environment variables of DaemonSet
+  cniPluginDir     []string
+
+  // networkConfigDir is the network config path
+  // which from environment variables of DaemonSet
+  networkConfigDir string
```
caseydavenport: I'm a bit confused - what is this parameter used for? I thought the network config was specified in a config map.

Is this just to configure the driver to know where the configmap has been mounted into the pod?

PR #2090, opened 2018-04-30
KEP: efficient node heartbeats

wojtek-t:
/assign @dchen1107 @derekwaynecarr 
/reviewers @lavalamp @deads2k 

/cc @bgrant0607 @yujuhong @liggitt @smarterclayton 
@kubernetes/sig-node-api-reviews @kubernetes/sig-api-machinery-api-reviews

### DIFF
--- a/keps/0008-node-heartbeat.md
+++ b/keps/0008-node-heartbeat.md
@@ -1,0 +1,390 @@
+---
+kep-number: 8
+title: Efficient Node Heartbeat
+authors:
+  - "@wojtek-t"
+  - "with input from @bgrant0607, @dchen1107, @yujuhong, @lavalamp"
+owning-sig: sig-node
+participating-sigs:
+  - sig-scalability
+  - sig-apimachinery
+  - sig-scheduling
+reviewers:
+  - "@deads2k"
+  - "@lavalamp"
+approvers:
+  - "@dchen1107"
+  - "@derekwaynecarr"
+editor: TBD
+creation-date: 2018-04-27
+last-updated: 2018-04-27
+status: provisional
+see-also:
+  - https://github.com/kubernetes/kubernetes/issues/14733
+  - https://github.com/kubernetes/kubernetes/pull/14735
+replaces:
+  - n/a
+superseded-by:
+  - n/a
+---
+
+# Efficient Node Heartbeats
+
+## Table of Contents
+
+Table of Contents
+=================
+
+* [Efficient Node Heartbeats](#efficient-node-heartbeats)
+   * [Table of Contents](#table-of-contents)
+   * [Summary](#summary)
+   * [Motivation](#motivation)
+      * [Goals](#goals)
+      * [Non-Goals](#non-goals)
+   * [Proposal](#proposal)
+      * [Risks and Mitigations](#risks-and-mitigations)
+   * [Graduation Criteria](#graduation-criteria)
+   * [Implementation History](#implementation-history)
+   * [Alternatives](#alternatives)
+      * [Dedicated “heartbeat” object instead of “leader election” one](#dedicated-heartbeat-object-instead-of-leader-election-one)
+      * [Events instead of dedicated heartbeat object](#events-instead-of-dedicated-heartbeat-object)
+      * [Reuse the Component Registration mechanisms](#reuse-the-component-registration-mechanisms)
+      * [Split Node object into two parts at etcd level](#split-node-object-into-two-parts-at-etcd-level)
+      * [Delta compression in etcd](#delta-compression-in-etcd)
+      * [Replace etcd with other database](#replace-etcd-with-other-database)
+
+## Summary
+
+Node heartbeats are necessary for correct functioning of Kubernetes cluster.
+This proposal makes them significantly cheaper from both scalability and
+performance perspective.
+
+## Motivation
+
+While running different scalability tests we observed that in big enough clusters
+(more than 2000 nodes) with non-trivial number of images used by pods on all
+nodes (10-15), we were hitting etcd limits for its database size. That effectively
+means that etcd enters "alert mode" and stops accepting all write requests.
+
+The underlying root cause is combination of:
+
+- etcd keeping both current state and transaction log with copy-on-write
+- node heartbeats being pontetially very large objects (note that images
+  are only one potential problem, the second are volumes and customers
+  want to mount 100+ volumes to a single node) - they may easily exceed 15kB;
+  even though the patch send over network is small, in etcd we store the
+	whole Node object
+- Kubelet sending heartbeats every 10s
+
+This proposal presents a proper solution for that problem.
+
+
+Note that currently (by default):
+
+- Lack of NodeStatus update for `<node-monitor-grace-period>` (default: 40s)
+  results in NodeController marking node as NotReady (pods are no longer
+  scheduled on that node)
+- Lack of NodeStatus updates for `<pod-eviction-timeout>` (default: 5m)
+  results in NodeController starting pod evictions from that node
+
+We would like to preserve that behavior.
+
+
+### Goals
+
+- Reduce size of etcd by making node heartbeats cheaper
+
+### Non-Goals
+
+The following are nice-to-haves, but not primary goals:
+
+- Reduce resource usage (cpu/memory) of control plane (e.g. due to processing
+  less and/or smaller objects)
+- Reduce watch-related load on Node objects
+
+## Proposal
+
+We propose introducing a new `Lease` built-in API in the newly create API group
+`coordination.k8s.io`). To make it easily reusable for other purposes it will
+be namespaced. Its schema will be as following:
+
+```
+type Lease struct {
+  metav1.TypeMeta `json:",inline"`
+  // Standard object's metadata.
+  // More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+  // +optional
+  ObjectMeta metav1.ObjectMeta `json:"metadata,omitempty"`
+
+  // Specification of the Lease.
+  // More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status
+  // +optional
+  Spec LeaseSpec `json:"spec,omitempty"`
+}
+
+type LeaseSpec struct {
+  HolderIdentity       string           `json:"holderIdentity"`
+  LeaseDurationSeconds int32            `json:"leaseDurationSeconds"`
+  AcquireTime          metav1.MicroTime `json:"acquireTime"`
+  RenewTime            metav1.MicroTime `json:"renewTime"`
+  LeaseTransitions     int32            `json:"leaseTransitions"`
+}
+```
+
+The Spec is effectively of already existing (and thus proved) [LeaderElectionRecord][].
+The only difference is using `MicroTime` instead of `Time` for better precision.
+That would hopefully allow us go get directly to Beta.
+
+We will use that object to represent node heartbeat - for each Node there will
+be a corresponding `Lease` object with Name equal to Node name in a newly
+created dedicated namespace (we considered using `kube-system` namespace but
+decided that it's already too overloaded).
+That namespace should be created automatically (similarly to "default" and
+"kube-system", probably by NodeController) and never be deleted (so that nodes
+don't require permission for it).
+
+We considered using CRD instead of built-in API. However, even though CRDs are
+`the new way` for creating new APIs, they don't yet have versioning support
+and are significantly less performant (due to lack of protobuf support yet).
+We also don't know whether we could seamlessly transition storage from a CRD
+to a built-in API if we ran into a performance or any other problems.
+As a result, we decided to proceed with built-in API.
+
+
+We this new API in place, we will change Kubelet so that:
+
+1. Kubelet is periodically computing NodeStatus every 10s (at it is now), but that will
+   be independent from reporting status
+1. Kubelet is reporting NodeStatus if:
+   - there was a meaningful change in it (initially we can probably assume that every
+     change is meaningful, including e.g. images on the node)
+   - or it didn’t report it over last `node-status-update-period` seconds
+1. Kubelet creates and periodically updates its own Lease object and frequency
+   of those updates is independent from NodeStatus update frequency.
+
+In the meantime, we will change `NodeController` to treat both updates of NodeStatus
+object as well as updates of the new `Lease` object corresponding to a given
+node as healthiness signal from a given Kubelet. This will make it work for both old
+and new Kubelets.
+
+We should also:
+
+1. audit all other existing core controllers to verify if they also don’t require
+   similar changes in their logic ([ttl controller][] being one of the examples)
+1. change controller manager to auto-register that `Lease` CRD
+1. ensure that `Lease` resource is deleted when corresponding node is
+   deleted (probably via owner references)
+1. [out-of-scope] migrate all LeaderElection code to use that CRD
+
+Once all the code changes are done, we will:
+
+1. start updating `Lease` object every 10s by default, at the same time
+   reducing frequency of NodeStatus updates initially to 40s by default.
+   We will reduce it further later.
+   <br> TODO: That still results in higher average QPS. It should be acceptable but
+   needs to be verified.
+1. announce that we are going to reduce frequency of NodeStatus updates further
+   and give people 1-2 releases to switch their code to use `Lease`
+   object (if they relied on frequent NodeStatus changes)
+1. further reduce NodeStatus updates frequency to not less often than once per
+   1 minute.
+   We can’t stop periodically updating NodeStatus as it would be API breaking change,
+   but it’s fine to reduce its frequency (though we should continue writing it at
+   least once per eviction period).
+
+
+To be considered:
+
+1. We may consider reducing frequency of NodeStatus updates to once every 5 minutes
+   (instead of 1 minute). That would help with performance/scalability even more.
+   Caveats:
+   - NodeProblemDetector is currently updating (some) node conditions every 1 minute
+     (unconditionally, because lastHeartbeatTime always changes). To make reduction
+     of NodeStatus updates frequency really useful, we should also change NPD to
+     work in a similar mode (check periodically if condition changes, but report only
+     when something changed or no status was reported for a given time) and decrease
+     its reporting frequency too.
+   - In general, we recommend to keep frequencies of NodeStatus reporting in both
+     Kubelet and NodeProblemDetector in sync (once all changes will be done) and
+     that should be reflected in [NPD documentation][].
+   - Note that reducing frequency to 1 minute already gives us almost 6x improvment.
+     It seems more than enough for any foreseeable future assuming we won’t
+     significantly increase the size of object Node.
+     Note that if we keep adding node conditions owned by other components, the
+     number of writes of Node object will go up. But that issue is separate from
+     that proposal.
+
+Other notes:
+
+1. Additional advantage of using Lease for that purpose would be the
+   ability to exclude it from audit profile and thus reduce the audit logs footprint.
+
+[LeaderElectionRecord]: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/leaderelection/resourcelock/interface.go#L37
+[ttl controller]: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/ttl/ttl_controller.go#L155
+[NPD documentation]: https://kubernetes.io/docs/tasks/debug-application-cluster/monitor-node-health/
+[kubernetes/kubernetes#63667]: https://github.com/kubernetes/kubernetes/issues/63677
+
+### Risks and Mitigations
+
+Increasing default frequency of NodeStatus updates may potentially break clients
+relying on frequent Node object updates. However, in non-managed solutions, customers
+will still be able to restore previous behavior by setting appropriate flag values.
+Thus, changing defaults to what we recommend is the path to go with.
+
+## Graduation Criteria
+
+The API can be immediately promoted to Beta, as the API is effectively a copy of
+already existing LeaderElectionRecord. It will be promoted to GA once it's gone
+a sufficient amount of time as Beta with no changes.
+
+The changes in components logic (Kubelet, NodeController) should be done behind
+a feature gate. We suggest making that enabled by default once the feature is
+implemented.
+
+## Implementation History
+
+- RRRR-MM-DD: KEP Summary, Motivation and Proposal merged
+
+## Alternatives
+
+We considered a number of alternatives, most important mentioned below.
+
+### Dedicated “heartbeat” object instead of “leader election” one
+
+Instead of introducing and using “lease” object, we considered
+introducing a dedicated “heartbeat” object for that purpose. Apart from that,
+all the details about the solution remain pretty much the same.
+
+Pros:
+
+- Conceptually easier to understand what the object is for
+
+Cons:
+
+- Introduces a new, narrow-purpose API. Lease is already used by other
+  components, implemented using annotations on Endpoints and ConfigMaps.
+
+### Events instead of dedicated heartbeat object
+
+Instead of introducing a dedicated object, we considered using “Event” object
+for that purpose. At the high-level the solution looks very similar. 
+The differences from the initial proposal are:
+
+- we use existing “Event” api instead of introducing a new API
+- we create a dedicated namespace; events that should be treated as healthiness
+  signal by NodeController will be written by Kubelets (unconditionally) to that
+  namespace
+- NodeController will be watching only Events from that namespace to avoid
+  processing all events in the system (the volume of all events will be huge)
+- dedicated namespace also helps with security - we can give access to write to
+  that namespace only to Kubelets
+
+Pros:
+
+- No need to introduce new API
+   - We can use that approach much earlier due to that.
+- We already need to optimize event throughput - separate etcd instance we have
+  for them may help with tuning
+- Low-risk roll-forward/roll-back: no new objects is involved (node controller
+  starts watching events, kubelet just reduces the frequency of heartbeats)
+
+Cons:
+
+- Events are conceptually “best-effort” in the system:
+   - they may be silently dropped in case of problems in the system (the event recorder
+     library doesn’t retry on errors, e.g. to not make things worse when control-plane
+     is starved)
+   - currently, components reporting events don’t even know if it succeeded or not (the
+     library is built in a way that you throw the event into it and are not notified if
+     that was successfully submitted or not).
+     Kubelet sending any other update has full control on how/if retry errors.
+   - lack of fairness mechanisms means that even when some events are being successfully
+     send, there is no guarantee that any event from  a given Kubelet will be submitted
+     over a given time period
+	So this would require a different mechanism of reporting those “heartbeat” events.
+- Once we have “request priority” concept, I think events should have the lowest one.
+  Even though no particular heartbeat is important, guarantee that some heartbeats will
+  be successfully send it crucial (not delivering any of them will result in unnecessary
+  evictions or not-scheduling to a given node). So heartbeats should be of the highest
+  priority. OTOH, node heartbeats are one of the most important things in the system
+  (not delivering them may result in unnecessary evictions), so they should have the
+  highest priority.
+- No core component in the system is currently watching events
+   - it would make system’s operation harder to explain
+- Users watch Node objects for heartbeats (even though we didn’t recommend it).
+  Introducing a new object for the purpose of heartbeat will allow those users to
+  migrate, while using events for that purpose breaks that ability. (Watching events
+  may put us in tough situation also from performance reasons.)
+- Deleting all events (e.g. event etcd failure + playbook response) should continue to
+  not cause a catastrophic failure and the design will need to account for this.
+
+### Reuse the Component Registration mechanisms
+
+Kubelet is one of control-place components (shared controller). Some time ago, Component
+Registration proposal converged into three parts:
+
+- Introducing an API for registering non-pod endpoints, including readiness information: #18610
+- Changing endpoints controller to also watch those endpoints
+- Identifying some of those endpoints as “components”
+
+We could reuse that mechanism to represent Kubelets as non-pod endpoint API.
+
+Pros:
+
+- Utilizes desired API
+
+Cons:
+
+- Requires introducing that new API
+- Stabilizing the API would take some time
+- Implementing that API requires multiple changes in different components
+
+### Split Node object into two parts at etcd level
+
+We may stick to existing Node API and solve the problem at storage layer. At the
+high level, this means splitting the Node object into two parts in etcd (frequently
+modified one and the rest).
+
+Pros:
+
+- No need to introduce new API
+- No need to change any components other than kube-apiserver
+
+Cons:
+
+- Very complicated to support watch
+- Not very generic (e.g. splitting Spec and Status doesn’t help, it needs to be just
+  heartbeat part)
+- [minor] Doesn’t reduce amount of data that should be processed in the system (writes,
+  reads, watches, …)
+
+### Delta compression in etcd
+
+An alternative for the above can be solving this completely at the etcd layer. To
+achieve that, instead of storing full updates in etcd transaction log, we will just
+store “deltas” and snapshot the whole object only every X seconds/minutes.
+
+Pros:
+
+- Doesn’t require any changes to any Kubernetes components
+
+Cons:
+
+- Computing delta is tricky (etcd doesn’t understand Kubernetes data model, and
+  delta between two protobuf-encoded objects is not necessary small)
+- May require a major rewrite of etcd code and not even be accepted by its maintainers
+- More expensive computationally to get an object in a given resource version (which
+  is what e.g. watch is doing)
+
+### Replace etcd with other database
+
+Instead of using etcd, we may also consider using some other open-source solution.
+
+Pros:
+
+- Doesn’t require new API
+
+Cons:
+
+- We don’t even know if there exists solution that solves our problems and can be used.
+- Migration will take us years.


keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+  LeaderTransitions    int32       `json:"leaderTransitions"`
+}
+```
+
+The Spec is effectively of already existing (and thus proved) [LeaderElectionRecord][].
+That would hopefully allow us go get directly to Beta.
+
+We will use CRD (as opposed to built-in API), because:
+
+- CRDs are `the` way to created new APIs
```
liggitt: I don't think this is quite accurate. It doesn't seem proper for a core kubernetes component to depend on the existence of a "custom" resource, which would be required to exist in all clusters to be functional

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+
+While running different scalability tests we observed that in big enough clusters
+(more than 2000 noodes) with non-trivial number of images used by pods on all
+nodes (10-15), we were hitting etcd limits for its database size. That effectively
+means that etcd enters "alert mode" and stops accepting all write requests.
+
+The underlying root cause is combination of:
+
+- etcd keeping both current state and transaction log with copy-on-write
+- node heartbeats being pontetially very large objects (even though
```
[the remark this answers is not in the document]
wojtek-t: Images are only one potential problem (yes - we may consider moving them to a separate object, but that doesn't solve the generic problem of "heartbeats are expensive").

The second big problems are Volumes - we will soon support 100+ volumes per VM in GCE and that has exactly the same issue.

So we really should solve this problem in more generic way.

keps/0008-node-heartbeat.md
wojtek-t: I'm not the API-expert. I've been discussing it with @bgrant0607 and he explicitly said that it should be done with CRD.
Personally, I don't really have preference here, but I trust that Brian knows what he is saying.

liggitt: > I'm not the API-expert. I've been discussing it with @bgrant0607 and he explicitly said that it should be done with CRD. Personally, I don't really have preference here, but I trust that Brian knows what he is saying.

that doesn't seem proper... node heartbeating isn't an extension, add-on, or optional component... it's fundamental to a kubernetes cluster.

wojtek-t: @liggitt - is your concerns only about CRD vs built-in API or more?

To cite Brian on CRDs:
```
1. CRD is a mechanism for building APIs
2. Most new APIs will at least start as CRDs
3. Most new APIs will stay CRDs forever
```
But I will let @bgrant0607 to comment more on that.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+  the patch send over network is small, in etcd we store the whole Node
+  object) - they may easily exceed 15kB
+- Kubelet sending heartbeats every 10s
+
+This proposal presents a proper solution for that problem.
+
+
+Note that currently:
```
zhouhaibing089: Note that currently by default?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+     number of writes of Node object will go up. But that issue is separate from
+     that proposal.
+1. Increasing default frequency of NodeStatus updates may potentially break customers
+   relying on frequent Node object updates. However, in non-managed solutions,
+   customers will still be able to restore previous behavior by setting appropriate
+   flag values. Thus, changing defaults to what we recommend is the path to go with.
+
+Other notes:
+
+1. Additional advantage of using LeaderElection for that purpose would be the ability
```
zhouhaibing089: exclude?

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
bgrant0607: I agree with Wojtek. 

Also, the image info is unnecessarily large, regardless where it is stored.

And I want to mitigate the proliferation of APIs.

keps/0008-node-heartbeat.md
liggitt: > is your concerns only about CRD vs built-in API or more?

That's my main concern. An API the kubelet depends on is an API the core repo has to define, document, build clients for, and support. Most of those are easier and more reliable for the kube-apiserver to do as a compiled-in type, rather than a CRD. Given that the schema stability and iteration time is tied to kube releases, I don't see many benefits to building these as a CRD, and see plenty of downsides.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+
+We will change Kubelet so that:
+
+1. Kubelet is periodically computing NodeStatus every 10s (at it is now), but that will
+   be independent from reporting status
+1. Kubelet is reporting NodeStatus if:
+   - there was a meaningful change in it (initially we can probably assume that every
+     change is meaningful, including e.g. images on the node)
+   - or it didn’t report it over last `node-status-update-period` seconds
+1. Kubelet creates and periodically updates its own LeaderElection object and frequency
```
liggitt: > its own LeaderElection object

what name is reserved for kubelets to use?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+
+The following are nice-to-haves, but not primary goals:
+
+- Reduce resource usage (cpu/memory) of control plane (e.g. due to processing
+  less and/or smaller objects)
+- Reduce watch-related load on Node objects
+
+## Proposal
+
+We propose introducing a new `LeaderElection` CRD with the following schema:
```
liggitt: is this a namespaced or cluster-scoped object?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+all the details about the solution remain pretty much the same.
+
+Pros:
+
+- Conceptually easier to understand what the object is for
+
+Cons:
+
+- Introduces a new, narrow-purpose API. Leader election is already used by other
+  components, implemented using annotations on Endpoints and ConfigMaps.
```
liggitt: I don't see this as a con. Using a leader election API object for a single actor which doesn't make use of any of the "leader" aspects of it is pretty confusing.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
liggitt: ok

keps/0008-node-heartbeat.md
bgrant0607: I actually see the arguments against CRDs as pretty weak.

Extension mechanisms need to be in the Nucleus so that we can use them to build the rest of the system. We're trying to make the apiserver dumb, generic functionality as much as possible. In particular, we're trying to make CRDs good enough to build APIs as good as built-in ones for most resources. And we'd like most new APIs to at least start as CRDs. 

This resource is just CRUD. It's implemented in annotations on ConfigMaps currently. We could just use that, with the ConfigMaps in a new namespace to reduce the watch cost.

If this were a built-in API, it would probably start as alpha, in which case Kubelet also couldn't depend on it. The node controller also can't depend on it, since we need to support version-skewed kubelets.

keps/0008-node-heartbeat.md
bgrant0607: I think only the name "leader election" is confusing. This is a Lease API, which is what I'd call it.

The other actor is the node controller, which takes action when the Kubelet doesn't renew its lease in time. It just doesn't acquire the lease.

keps/0008-node-heartbeat.md
smarterclayton: If write path efficiency matters, CRDs are at least 2-5x more expensive than a proper api with generated types.

Prototyping as a CRD makes some sense.  Splitting other pieces out into their own CRD types makes sense.  But if heartbeating is significant enough to make special its unlikely CRD is enough

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+   be independent from reporting status
+1. Kubelet is reporting NodeStatus if:
+   - there was a meaningful change in it (initially we can probably assume that every
+     change is meaningful, including e.g. images on the node)
+   - or it didn’t report it over last `node-status-update-period` seconds
+1. Kubelet creates and periodically updates its own LeaderElection object and frequency
+   of those updates is independent from NodeStatus update frequency.
+
+In the meantime, we will change `NodeController` to treat both updates of NodeStatus
+object as well as updates of the new `LeaderElection` object corresponding to a given
```
smarterclayton: This breaks backwards compatibility for any integration anyone has done to watching node objects for changes.  That’s any custom scheduler, and alerting based function, and probably a non trivial amount of custom node integrations built by users.

keps/0008-node-heartbeat.md
bgrant0607: Also note that my opinion on a "node heartbeat" API hasn't changed since the original proposal (https://github.com/kubernetes/kubernetes/pull/14735) in 2015. In fact, with ~60 APIs now it's even more critical that we introduce new ones with great care.

keps/0008-node-heartbeat.md
smarterclayton: Given that etcd has a special api specifically for leases, I’m having some trouble believing we should implement that same abstraction. it as a generic abstraction on top of the generic key value store.  I’d much prefer to leverage the underlying semantics of the store to make lease as efficient as possible.  Abstractions are great except when we push them too far.

smarterclayton: For instance, this is a great place to actually uses separate key ordered inline with nodes which is transactionally updated and synthsized into node or another resource as needed, preserving watch.

keps/0008-node-heartbeat.md
wojtek-t: I think it should be namespaced, so that paople would also be able to take advantage of it in their applications.

In that case, I would say that the the object corresponding to a given Kubelet should be:
namespace: kube-system
name: kubelet-`<node name>`

Does that sound ok?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+## Proposal
+
+We propose introducing a new `LeaderElection` CRD with the following schema:
+```
+type LeaderElection struct {
+  metav1.TypeMeta `json:",inline"`
+  // Standard object's metadata.
+  // More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+  // +optional
+	ObjectMeta metav1.ObjectMeta `json:"metadata,omitempty"`
```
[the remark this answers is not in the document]
wojtek-t: I think the name. As I mentioned above, the object corresponding to  given Kubelet should be:
namespace: kube-system
name: kubelet-<node name>

[holderIdentity will probably be the same as name though]

keps/0008-node-heartbeat.md
wojtek-t: Treating both as a healthiness signal? Why? Can you clarify that?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,374 @@
...
+1. Kubelet is reporting NodeStatus if:
+   - there was a meaningful change in it (initially we can probably assume that every
+     change is meaningful, including e.g. images on the node)
+   - or it didn’t report it over last `node-status-update-period` seconds
+1. Kubelet creates and periodically updates its own LeaderElection object and frequency
+   of those updates is independent from NodeStatus update frequency.
+
+In the meantime, we will change `NodeController` to treat both updates of NodeStatus
+object as well as updates of the new `LeaderElection` object corresponding to a given
+node as healthiness signal from a given Kubelet. This will make it work for both old
```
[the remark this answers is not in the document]
wojtek-t: I would say we should be watching kube-system namespace.

keps/0008-node-heartbeat.md
wojtek-t: @smarterclayton 
I did some first benchmarks of it and indeed CRDs vs built-in (which would mostly be json vs protobuf) would be ~5-6x slower. But, the object is fairly small, and even if that is 5x slower, it's still better the serialization of the whole Node object. Given that serialization/deserialization of Node objects is not our main painpoint in apiserver, I'm not that worried about that.

Re leases - I thought about it and I think that's actually not the best idea due to one reason:

Leases are not watchable (i.e. only creation and deletion of lease are watchable IIUC).
If we would use leases for that purpose, we would break all people who would like to watch for all heartbeats.
If we use regular API, they would just need to migrate to use this object instead.


> For instance, this is a great place to actually uses separate key ordered inline with nodes which is transactionally updated and synthsized into node or another resource as needed, preserving watch.

This is some variation of `Split Node object into two parts at etcd level` alternative mentioned below. Splitting object into two an synthetizing them to preserve a watch is pretty difficult. I fleshed some more thoughts about it in https://github.com/kubernetes/kubernetes/issues/14733#issuecomment-362217568 couple weeks ago. This would be a lot of very complicated code that wouldn't be very generic, I thinkg (@lavalamp for thoughts too).
You know that we already had bugs in cacher, and that is place that would have to be significantly changed for that and I guess we would have more.
So after more thinking about that, I'm quite against about presenting that in watch.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
wojtek-t: Incorporated into text

keps/0008-node-heartbeat.md
wojtek-t: Done

keps/0008-node-heartbeat.md
wojtek-t: Done

keps/0008-node-heartbeat.md
bgrant0607: I agree that should be explicitly specified in the proposal. It should be the node name, or possibly trivially derived from the node name, but isolating the resources by namespace should be good enough to not require prefixing IMO.

keps/0008-node-heartbeat.md
bgrant0607: Yes, it needs to be namespaced. 

I'd put them in a new namespace and skip the prefix.

keps/0008-node-heartbeat.md
bgrant0607: We'll still write the status, just less frequently. It's effectively a configuration change.

bgrant0607: Also, schedulers shouldn't be watching elapsed time since last heartbeat directly. The node controller encodes cordoning and eviction in taints and conditions.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,376 @@
...
+   reducing frequency of NodeStatus updates initially to 20s by default.
+   We will reduce it further later.
+   <br> TODO: That still results in higher average QPS. It should be acceptable but
+   needs to be verified.
+1. announce that we are going to reduce frequency of NodeStatus updates further
+   and give people 1-2 releases to switch their code to use `LeaderElection`
+   object (if they relied on frequent NodeStatus changes)
+1. further reduce NodeStatus updates frequency to not less often than once per
+   1 minute.
+   We can’t stop periodically updating NodeStatus as it would be API breaking change,
```
resouer: Just to confirm, today's `NodeStatus` update will be kept, as other components/clients still need a way to consume status of nodes, right?

While if that's the case, isn't it true that consumers may get stale status of Node? Is this expected?

keps/0008-node-heartbeat.md
resouer: One of the concerns I have for using CRD is: where & how these CRD would be maintained? e.g. Separate repo?

bgrant0607: The existing leader election code is in client-go. It could go there, or elsewhere in k/k.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
bgrant0607: The name

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
bgrant0607: Or a dedicated namespace

keps/0008-node-heartbeat.md
bgrant0607: If something material changes in the node status, it will be updated by kubelet, as it is now.

If the node is deemed MIA, the node controller will update the node status and taints, as it does now.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,376 @@
...
+  even though the patch send over network is small, in etcd we store the
+	whole Node object
+- Kubelet sending heartbeats every 10s
+
+This proposal presents a proper solution for that problem.
+
+
+Note that currently (by default):
+
+- Lack of NodeStatus update for 40s results in NodeController marking node
```
bgrant0607: Actually, if some other component changed the Ready condition to Unknown, would node controller then set the taint correctly?

If not, it might be useful to decouple that code in node controller and create a way to disable the node controller's status checking so that someone could implement their own probing system if they wanted to.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,376 @@
...
+
+- Reduce resource usage (cpu/memory) of control plane (e.g. due to processing
+  less and/or smaller objects)
+- Reduce watch-related load on Node objects
+
+## Proposal
+
+We propose introducing a new `LeaderElection` CRD with the following schema:
+```
+type LeaderElection struct {
```
bgrant0607: I would call this Lease.

It's fine that the leader-election library would use it, since that's what that library does

keps/0008-node-heartbeat.md
wojtek-t: Added to doc.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
wojtek-t: Added that explanation below.

keps/0008-node-heartbeat.md
wojtek-t: Added to the doc above

keps/0008-node-heartbeat.md
resouer: Then what would be the API breaking change if we stop the periodically NodeStatus update?

keps/0008-node-heartbeat.md
wojtek-t: I initially wanted to have this somewhere in k/k. But it's not a strong preference - client-go sounds also fine for me.

wojtek-t: Clarified where to put that code in the doc.

keps/0008-node-heartbeat.md
wojtek-t: Switched to Lease - that sounds fine to me too.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,385 @@
+---
+kep-number: 8
+title: Efficient Node Heartbeat
+authors:
+  - "@wojtek-t"
+  - "with input from @bgrant0607, @dchen1107, @yujuhong, @lavalamp"
+owning-sig: sig-node
+participating-sigs:
+  - sig-scalability
+  - sig-apimachinery
```
bgrant0607: I would add sig-scheduling, due to the image issue

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,385 @@
...
+1. Additional advantage of using Lease for that purpose would be the
+   ability to exclude it from audit profile and thus reduce the audit logs footprint.
+
+[LeaderElectionRecord]: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/leaderelection/resourcelock/interface.go#L37
+[ttl controller]: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/ttl/ttl_controller.go#L155
+[NPD documentation]: https://kubernetes.io/docs/tasks/debug-application-cluster/monitor-node-health/
+
+### Risks and Mitigations
+
+Increasing default frequency of of NodeStatus updates may potential break customers
```
bgrant0607: s/customers/clients/g

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,385 @@
...
+	So this would require a different mechanism of reporting those “heartbeat” events.
+- Once we have “request priority” concept, I think events should have the lowest one.
+  Even though no particular heartbeat is important, guarantee that some heartbeats will
+  be successfully send it crucial (not delivering any of them will result in unnecessary
+  evictions or not-scheduling to a given node). So heartbeats should be of the highest
+  priority. OTOH, node heartbeats are one of the most important things in the system
+  (not delivering them may result in unnecessary evictions), so they should have the
+  highest priority.
+- No core component in the system is currently watching events
+   - it would make system`s operation harder to explain
```
bgrant0607: The backtick should be a forward tick

keps/0008-node-heartbeat.md
wojtek-t: done

keps/0008-node-heartbeat.md
wojtek-t: done

keps/0008-node-heartbeat.md
wojtek-t: done

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,386 @@
...
+## Summary
+
+Node heartbeats are necessary for correct functioning of Kubernetes cluster.
+This proposal makes them significantly cheaper from both scalability and
+performance perspective.
+
+## Motivation
+
+While running different scalability tests we observed that in big enough clusters
+(more than 2000 noodes) with non-trivial number of images used by pods on all
```
dchen1107: s/noodes/nodes

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,386 @@
...
+
+The Spec is effectively of already existing (and thus proved) [LeaderElectionRecord][].
+That would hopefully allow us go get directly to Beta.
+
+Definition of `Lease` object will be added to main Kubernetes repo or
+alternatively to client-go repository next to leader election library.
+
+We will use that object to represent node heartbeat - for each Node
+there will be corresponding `Lease` object with Name equal
+to Node name in the `kube-system` namespace.
```
dchen1107: @derekwaynecarr in another thread mentioned that we are overloading kube-system namespace nowadays. and suggested to use a separate one. I like the idea of separating node-related objects and operations into a different ns. cc/ @derekwaynecarr

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,386 @@
...
+   similar changes in their logic ([ttl controller][] being one of the examples)
+1. change controller manager to auto-register that `Lease` CRD
+1. ensure that `Lease` resource is deleted when corresponding node is
+   deleted (probably via owner references)
+1. [out-of-scope] migrate all LeaderElection code to use that CRD
+
+Once all the code changes are done, we will:
+
+1. start updating `Lease` object every 10s by default, at the same time
+   reducing frequency of NodeStatus updates initially to 20s by default.
```
dchen1107: How bad to change node-status-update-period to 40s, even 1min to reduce QPS?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,386 @@
...
+   but it’s fine to reduce its frequency (though we should continue writing it at
+   least once per eviction period).
+
+
+To be considered:
+
+1. We may consider reducing frequency of NodeStatus updates to once every 5 minutes
+   (instead of 1 minute). That would help with performance/scalability even more.
+   Caveats:
+   - NodeProblemDetector is currently updating (some) node conditions every 1 minute
```
dchen1107: The only reason of updating it periodically is NodeCondition API definition now, which can be changed. NPD can read NodeStatus, and only update upon the changes.

keps/0008-node-heartbeat.md
dchen1107: @smarterclayton We thought about such compatibility issues and this is why we still updated lastHeartbeatTime periodically, but with a increasing period.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
dchen1107: Basically the node object is unbound today. Besides images and volumes, there could be labels and annotations to that API object. And periodically (now 10s) updating makes it very expensive.

keps/0008-node-heartbeat.md
dchen1107: @wojtek-t Can you include your experimental benchmark on CRD vs. built-in into this KEP?  I am fine with CRD, at least we can start with CRD first. If the performance concern remains, we can graduate that to built-in later.

keps/0008-node-heartbeat.md
wojtek-t: done

keps/0008-node-heartbeat.md
wojtek-t: switched to dedicated namespace (name of that is TBD - it doesn't really matter for this proposal, but if you have suggestion, I'm happy to add it here too)

keps/0008-node-heartbeat.md
wojtek-t: From my personal perspective, the lower frequency is, the better. I'm just not sure how far we can go initially.
@bgrant0607 - do you think we can go straight to 1m (or 40s)?

keps/0008-node-heartbeat.md
wojtek-t: I tried to capture that in this point that we generally may change that. If you have any suggestions how to make it more clear, let me know.

keps/0008-node-heartbeat.md
wojtek-t: Technically, so far I only did json vs protobuf benchmark on this particular object with some ~real data. I can create a better benchmark when I'm back from leave in couple days.

@dchen1107 that said - how exactly you want this to be incorporated into this KEP? Just the results? Or maybe link to PR to main repo? Or sth else?

dchen1107:
The proposal looks good to me, but I hope @derekwaynecarr can take a look before I apply /lgtm label. He is out for OpenShift Summit this week.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,388 @@
...
+  // More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status
+  // +optional
+  Spec LeaseSpec `json:"spec,omitempty"`
+}
+
+type LeaseSpec struct {
+  HolderIdentity       string      `json:"holderIdentity"`
+  LeaseDurationSeconds int32       `json:"leaseDurationSeconds"`
+  AcquireTime          metav1.Time `json:"acquireTime"`
+  RenewTime            metav1.Time `json:"renewTime"`
```
smarterclayton: Probably should be microtime (higher precision) except we haven’t gotten that in.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,388 @@
...
+to Node name in a newly created dedicated namespace (we considered
+using `kube-system` namespace but decided that it's already too overloaded).
+
+
+We will use CRD (as opposed to built-in API), because:
+
+- CRDs are `the` way to created new APIs
+- even though CRDs aren't very efficient due to lack of protobuf support now:
+    - performance should be acceptable (because Lease object will be small)
+      <br> TODO(wojtek-t): Working on microbenchmark to prove with data.
```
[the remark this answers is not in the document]
wojtek-t: We can simulate the solution (once implemented) with kubemark, but we are not really measuring there e.g. serialization time, which is quite crucial from perspective of it.
But yes - from Kubemark we will know if it regressed our SLOs or not.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,388 @@
...
+    - protobuf support for CRDs is on the roadmap (though not near term)
```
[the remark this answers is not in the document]
wojtek-t: Added.

keps/0008-node-heartbeat.md
wojtek-t: Changed to 40s then.

keps/0008-node-heartbeat.md
wojtek-t: Actually, we have that already in:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/micro_time.go#L33
And it's already used in new events and in audit logs:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/apis/audit/v1alpha1/types.go#L131

And I agree we should use it here (probably for both acquire and renew times). Changed.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,390 @@
...
+
+The following are nice-to-haves, but not primary goals:
+
+- Reduce resource usage (cpu/memory) of control plane (e.g. due to processing
+  less and/or smaller objects)
+- Reduce watch-related load on Node objects
+
+## Proposal
+
+We propose introducing a new `Lease` CRD. To make it easily
```
liggitt: given where the discussion in https://github.com/kubernetes/community/pull/2090#discussion_r184988877 ended up, I was surprised to see this still proposed as a CRD. Isn't it odd to build core function on top of extension APIs? The things CRDs let you do (easily disable, change schema, validation, etc) are all things I would not want any chance of happening to the resource my nodes used for heartbeating.

gyuho:
Just note:

Delta compression in etcd would be interesting, but just not sure how it will be implemented when etcd also stores Protocol Buffer encoded values. Same for other compression algorithms. We once evaluated transport-layer compression https://github.com/coreos/etcd/issues/5186, but not much benefit. Would love to experiment again!

For the near future, I would look into periodic compaction on etcd storage and run defragmentation to reclaim the disk space. There are plans to make this better as well (https://github.com/coreos/etcd/issues/9222).

/cc @jpbetz @joelegasse

keps/0008-node-heartbeat.md
wojtek-t: @bgrant0607 for more comments

Personally my slight preference was also towards built-in API, but Brian has very strong opinion on using CRD for it and ultimately he is the API reviewer.
That said, as long as this is not yet enabled by default, we will be free to change that. And purely from code-perspective, using CRD vs built-in API doesn't change much.

keps/0008-node-heartbeat.md
derekwaynecarr: @PatrickLang its configurable via node-monitor-grace-period and pod-eviction-timeout arguments to node lifecycle controller.  its probably best to refer to these timings generically and just reference their defaults.

wojtek-t: Changed to mention appropriate settings mentioned by Derek.

Answering Brian`s comment - yes, we should probably decouple that logic. But that's independent from this proposal (we should probably do it even without this proposal), so I will not mention it here to avoid making it to wide/complex.

keps/0008-node-heartbeat.md
bgrant0607: I put this on the agenda for SIG Architecture for tomorrow (the 17th).

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,391 @@
...
+  LeaseDurationSeconds int32            `json:"leaseDurationSeconds"`
+  AcquireTime          metav1.MicroTime `json:"acquireTime"`
+  RenewTime            metav1.MicroTime `json:"renewTime"`
+  LeaseTransitions     int32            `json:"leaseTransitions"`
+}
+```
+
+The Spec is effectively of already existing (and thus proved) [LeaderElectionRecord][].
+The only difference is using `MicroTime` instead of `Time` for better precision.
+That would hopefully allow us go get directly to Beta.
```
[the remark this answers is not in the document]
bgrant0607: I think there's a question of how we validate the new API other than in Kubemark and other synthetic tests if it starts as alpha and disabled by default.

keps/0008-node-heartbeat.md
bgrant0607: I think of this as significance filtering. If there is no material change to report, how often should that be confirmed? If there is a material change, how quickly should it be reported?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,391 @@
...
+
+### Risks and Mitigations
+
+Increasing default frequency of of NodeStatus updates may potential break clients
+relying on frequent Node object updates. However, in non-managed solution, customers
+will still be able to restore previous behavior by setting appropriate flag values.
+
+## Graduation Criteria
+
+This can be immediately promoted to Beta, as the API is effectively a copy of
```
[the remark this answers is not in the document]
bgrant0607: Regarding the feature gate: I agree that the kubelet and node controller behavior should be configurable. That should be decoupled from the API. 

The API itself is effectively just data. It should be largely equivalent to the long-standing annotation-based implementation on ConfigMap and other resources. There isn't a controller side to the API implementation.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
wojtek-t: > I think whether the API goes directly from inception to beta is orthogonal to the design, 

It's orthogonal to the design, it influences timeline when we can start using it.

> but seems to possibly influence the usage of CRDs versus built-in API.

Why? I consider CRD vs built-in API a separate thing from versioning.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,391 @@
...
+The only difference is using `MicroTime` instead of `Time` for better precision.
+That would hopefully allow us go get directly to Beta.
+
+Definition of `Lease` object will be added to main Kubernetes repo or
+alternatively to client-go repository next to leader election library.
+
+We will use that object to represent node heartbeat - for each Node
+there will be corresponding `Lease` object with Name equal
+to Node name in a newly created dedicated namespace (we considered
+using `kube-system` namespace but decided that it's already too overloaded).
```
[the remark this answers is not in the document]
wojtek-t: Yes - I completely agree - added to the doc.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,391 @@
...
+1. Kubelet is periodically computing NodeStatus every 10s (at it is now), but that will
+   be independent from reporting status
+1. Kubelet is reporting NodeStatus if:
+   - there was a meaningful change in it (initially we can probably assume that every
+     change is meaningful, including e.g. images on the node)
+   - or it didn’t report it over last `node-status-update-period` seconds
+1. Kubelet creates and periodically updates its own Lease object and frequency
+   of those updates is independent from NodeStatus update frequency.
+
+In the meantime, we will change `NodeController` to treat both updates of NodeStatus
```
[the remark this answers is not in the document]
wojtek-t: Not sure if I understand - can you please clarify?

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,391 @@
...
+
+1. start updating `Lease` object every 10s by default, at the same time
+   reducing frequency of NodeStatus updates initially to 40s by default.
+   We will reduce it further later.
+   <br> TODO: That still results in higher average QPS. It should be acceptable but
+   needs to be verified.
+1. announce that we are going to reduce frequency of NodeStatus updates further
+   and give people 1-2 releases to switch their code to use `Lease`
+   object (if they relied on frequent NodeStatus changes)
+1. further reduce NodeStatus updates frequency to not less often than once per
```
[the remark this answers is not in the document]
wojtek-t: That sentence from above is supposed to explain that:
```
Kubelet is reporting NodeStatus if:
  - there was a meaningful change in it (initially we can probably assume that every
    change is meaningful, including e.g. images on the node)
  - or it didn’t report it over last `node-status-update-period` seconds
```

So if any condition changed (no matter if the node is ready or not), Kubelet should report its status every 10s.
Only when all conditions (and other stuff) didn't change over last 10s (or sth), we are allowed to not report it up to 1m (or whatever period we will decide for).

@derekwaynecarr Is that more clear now? Do you have any suggestion to make that initial senstence I pasted above more cleaner?

keps/0008-node-heartbeat.md
wojtek-t: I already tried to answer in the comment above. But the TL;DR is Kubelet:
- if any condition (or any other field of NodeStatus) changed, it will be reported every 10s (as it is now)
- if nothing changed, we guarantee to update the NodeStatus every 1m (or whatever period we decide for)

The logic in NPD should be consistent with that (i.e. compute the NodeStatus every X seconds, report it only if something changed, but not less often than every Y seconds).

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,391 @@
...
+   - In general, we recommend to keep frequencies of NodeStatus reporting in both
+     Kubelet and NodeProblemDetector in sync (once all changes will be done) and
+     that should be reflected in [NPD documentation][].
+   - Note that reducing frequency to 1 minute already gives us almost 6x improvment.
+     It seems more than enough for any foreseeable future assuming we won’t
+     significantly increase the size of object Node.
+     Note that if we keep adding node conditions owned by other components, the
+     number of writes of Node object will go up. But that issue is separate from
+     that proposal.
+1. Increasing default frequency of NodeStatus updates may potentially break clients
```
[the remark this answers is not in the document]
wojtek-t: done

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
wojtek-t: Yeah - that's a good point.
I made that explicit that I was thinking about API (to promote it directly to Beta).

I added a point to put the changes to kubelet/NodeController logic behind feature gate. I think it should be enabled by default once done, but that's probably a separate discussion (though put that into doc).

wojtek-t:
> I have a number of questions that would help to clarify.

@derekwaynecarr I clarified those in the comments. TL;DR

> It is not clear to me if the kubelet updates node conditions at a frequency less than the node status
> update frequency in the case of meaningful change. If so, at what frequency?

Yes - if something changes it will be updated (with the same frequency as it is today - 10s).

> Its not clear why this is not behind a feature gate of some kind as its evaluated rather than immediately 
> moving to beta.

Yes - it should be behind feature gate - I added to the doc. I think it should be enabled by default once implemented (I mean the feature gate for it), but that's a separate discussion.

keps/0008-node-heartbeat.md
ncdc: Sorry I missed the arch meeting, but 💯 to being in the built-in API

wojtek-t: We have just discussed this during SIG Architecture.
Long story short, it seems that we want to pursue with using CRDs for even important APIs, but given that CRDs are not feature-complete it and given how core this object will be (necessary for Kubelets to work correctly), there seem to be a decision to use a built-in API.

The question, that I have now, is - what API groups it should belong then. IIUC, we don't want to add anything new to "core" api group. And I don't really see any other apigroup to be a good match. So should we create a new one?

@bgrant0607 @smarterclayton @liggitt @lavalamp - I would like to hear your suggestions here.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
ncdc: CRDs only currently support a single version per api group.

keps/0008-node-heartbeat.md
bgrant0607: The convincing argument was about risk. 

We could (and should) benchmark CRDs, but since they don't have versioning support yet and since we don't know whether we could seamlessly transition storage from a CRD to a built-in API, if we ran into a performance or other problem, it may not be easy to resolve if we started as a CRD.

Yes, don't add any new APIs to the core API group. Besides, we don't want to add a beta-ish API to a v1 group.

We'd need to choose a group name even if it were a CRD.

Conceptually, I'd categorize it as "coordination".

And I believe we've agreed that new groups will be suffixed with `.k8s.io`.

So that would suggest `coordination.k8s.io`.

wojtek-t: Thanks Brian!

I changed the doc to talk about built-in API (and the coordination.k8s.io api group). And added some explanation for future reference that we lso considered CRDs and why we decided for built-in api.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
wojtek-t: Got it.
That said, with a decision to proceed with built-in api, this part is no longer valid.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,390 @@
...
+
+We considered using CRD instead of built-in API. However, even though CRDs are
+`the new way` for creating new APIs, they don't yet have versioning support
+and are significantly less performant (due to lack of protobuf support yet).
+We also don't know whether we could seamlessly transition storage from a CRD
+to a built-in API if we ran into a performance or any other problems.
+As a result, we decided to proceed with built-in API.
+
+
+We this new API in place, we will change Kubelet so that:
```
derekwaynecarr: nit: With

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
derekwaynecarr: i think the updated text lets me answer my own question.  non renewal of a lease will cause the node controller to mark the node ready condition to condition unknown.  any other conditions source from the node will still be set when there is meaningful change, and the node controller will convert those conditions to taints as usual.

keps/0008-node-heartbeat.md
```diff
@@ -0,0 +1,390 @@
...
+1. audit all other existing core controllers to verify if they also don’t require
+   similar changes in their logic ([ttl controller][] being one of the examples)
+1. change controller manager to auto-register that `Lease` CRD
+1. ensure that `Lease` resource is deleted when corresponding node is
+   deleted (probably via owner references)
+1. [out-of-scope] migrate all LeaderElection code to use that CRD
+
+Once all the code changes are done, we will:
+
+1. start updating `Lease` object every 10s by default, at the same time
```
derekwaynecarr: nit: i know this is clarified earlier, but this is just changing the default frequency, but does not impact the frequency by which the node sends "meaningful" changes (i.e. changes in node condition) to the apiserver.  it only impacts the frequency of non-meaningful changes being updated.

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
derekwaynecarr: further iterations of this proposal will need to debate what is meaningful or not, but yes, this is clear enough for now.

keps/0008-node-heartbeat.md
derekwaynecarr: sgtm

keps/0008-node-heartbeat.md
wojtek-t: Fixed in https://github.com/kubernetes/community/pull/2172

keps/0008-node-heartbeat.md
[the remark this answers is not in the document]
wojtek-t: Yes - that is correct.

keps/0008-node-heartbeat.md
wojtek-t: Done in follow up PR.

keps/0008-node-heartbeat.md
mtaufen: What would the new namespace be called?
I assume the entire namespace is just scoped to node leases? 
I guess we should add it to the set of [immortal namespaces](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/namespace/lifecycle/admission.go#L65).

wojtek-t: > What would the new namespace be called?

TBD

> I assume the entire namespace is just scoped to node leases?

Yes.

> I guess we should add it to the set of immortal namespaces.

Yes

PR #2093, opened 2018-04-30
WIP: Promote sysctl annotations to fields

ingvagabund:
TBD


ingvagabund:
@sjenning @derekwaynecarr @sttts

sjenning:
Thanks Jan :+1:

/sig node

keps/0008-20180430-promote-sysctl-annotations-to-fields.md
```diff
@@ -0,0 +1,187 @@
+---
+kep-number: 8
+title: Protomote sysctl annotations to fields
+authors:
+  - "@ingvagabund"
+owning-sig: sig-node
+participating-sigs:
```
php-coder: As it touches PSP, I think that we may mention sig-auth here.

keps/0008-20180430-promote-sysctl-annotations-to-fields.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Implementation Details/Notes/Constraints
+
+Extending `SecurityContext` struct with `Sysctls` field:
+
+```go
+// SecurityContext holds security configuration that will be applied to a container.
+// Some fields are present in both SecurityContext and PodSecurityContext.  When both
+// are set, the values in SecurityContext take precedence.
+type SecurityContext struct {
```
php-coder: `SecurityContext` is for container-scoped restrictions. If we want to make it pod-scoped only, we need to modify `PodSecurityContext` instead.

php-coder:
@ingvagabund Thank you for working on this!

derekwaynecarr:
Please add section for following:

How to enable unsafe sysctls on node should move away from an experimental flag and become kubelet config api option

derekwaynecarr:
The original support predates feature gates, but it’s move to first class fields should come with feature gate.  Can you add details on feature gate?

keps/0008-20180430-promote-sysctl-annotations-to-fields.md
```diff
@@ -0,0 +1,187 @@
...
+spec:
+  sysctls:
+  - kernel.shmmax
+  - kernel.shmall
+  - net.*
+```
+
+More examples at [design-proposals/node/sysctl.md#allowing-only-certain-sysctls](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/sysctl.md#allowing-only-certain-sysctls)
+
+## Proposal
```
sttts: The kubelet has `--experimental-allowed-unsafe-sysctls`. Should we drop the "experimental" ?

php-coder: There is request for removing this options or moving it to the config file: https://github.com/kubernetes/kubernetes/issues/61669

sttts: The later is fine. But we need it. Otherwise, unsafe sysctls cannot be used.

ingvagabund:
@derekwaynecarr updated, PTAL

keps/0008-20180430-promote-sysctl-annotations-to-fields.md
```diff
@@ -0,0 +1,220 @@
...
+  // Whitelist of unsafe sysctls or unsafe sysctl patterns (ending in *).
+  // Default: nil
+  // +optional
+  AllowedUnsafeSysctls []string `json:"allowedUnsafeSysctls,omitempty"`
+}
+```
+
+Upstream issue: https://github.com/kubernetes/kubernetes/issues/61669
+
+### Gate the feature
```
derekwaynecarr: Please just state what the feature gate flag is and the default behavior.

I have no problem with what @sjenning has started, but this document will be read in the future and linking out like that is painful for future readers.

derekwaynecarr:
/cc @kubernetes/sig-node-proposals 

this is the document that describes the promotion of sysctl to fields for beta support.

if folks have any comments, please provide them this week

keps/0008-20180430-promote-sysctl-annotations-to-fields.md
ingvagabund: Updated

derekwaynecarr:
Thank you for the updates @ingvagabund 

/approve
/lgtm

sttts:
🎉

php-coder:
After 2 weeks after merge, it became outdated. Do we care about making it up-to-date?

keps/0008-20180430-promote-sysctl-annotations-to-fields.md
```diff
@@ -0,0 +1,225 @@
...
+  apiVersion: v1
+  kind: Pod
+  metadata:
+    name: sysctl-example
+  spec:
+    securityContext:
+      sysctls:
+      - name: kernel.shm_rmid_forced
+        value: 1
+      - name: net.ipv4.route.min_pmtu
```
thockin: Why are networking-related knobs in the security context?  This seems like the wrong structure to me?

PR #2110, opened 2018-05-01
This is a practice commit

cantbewong:
This describes why this is being done


contributors/new-contributor-playground/my-new-contribution.md.txt
```diff
@@ -0,0 +1,2 @@
+hello everyone
+this is line 2
```
msohn: isn't that obvious ? 
I am not convinced this line adds any value

PR #2111, opened 2018-05-01
Add first practice commit

mitsutaka:
This is description part.


mohamed-mehany:
Need to resolve conflict

PR #2121, opened 2018-05-02
Remove broken link for helm OWNERS

stigsb:


prydonius:
@stigsb can you please make the change in `sigs.yaml` and run `make`?

cblecker:
it may be broken, but the project it refers to is still owned by sig-apps. links were put into where owners files should be, even if they don't yet exist.

/lgtm cancel
/close

PR #2125, opened 2018-05-03
Add the minimum go version for release 1.11+

warmchang:
ref: https://github.com/kubernetes/kubernetes/pull/60597

### DIFF
--- a/contributors/devel/development.md
+++ b/contributors/devel/development.md
@@ -134,7 +134,9 @@ development environment, please [set one up](http://golang.org/doc/code.html).
 | 1.5, 1.6       | 1.7 - 1.7.5 |
 | 1.7            | 1.8.1       |
 | 1.8            | 1.8.3       |
-| 1.9+           | 1.9.1       |
+| 1.9            | 1.9.1       |
+| 1.10           | 1.9.1       |
+| 1.11+          | 1.10.1      |
 
 Ensure your GOPATH and PATH have been configured in accordance with the Go
 environment instructions.


contributors/devel/development.md
```diff
@@ -134,7 +134,9 @@ development environment, please [set one up](http://golang.org/doc/code.html).
 | 1.5, 1.6       | 1.7 - 1.7.5 |
 | 1.7            | 1.8.1       |
 | 1.8            | 1.8.3       |
-| 1.9+           | 1.9.1       |
+| 1.9            | 1.9.2       |
```
cblecker: this is go1.9.1: https://github.com/kubernetes/kubernetes/blob/release-1.9/hack/lib/golang.sh#L325

contributors/devel/development.md
```diff
@@ -134,7 +134,9 @@ development environment, please [set one up](http://golang.org/doc/code.html).
...
+| 1.10           | 1.9.3       |
```
cblecker: this is go1.9.1: https://github.com/kubernetes/kubernetes/blob/release-1.10/hack/lib/golang.sh#L329

warmchang: What's your opinion? For the 1.9 and 1.10 branches, only show the minimum go-lang version instead of the actual used of each branch?

cblecker: yes, it should be the minimum version for that branch.

warmchang: ok, I've modified and please check it.

PR #2135, opened 2018-05-07
Typo fix: change 'focussed' to 'focused'

angao:


cblecker:
This needs to be changed here: https://github.com/kubernetes/community/blob/master/sigs.yaml#L933

and then run `make` to regenerate the files.

angao:
thank you!

PR #2139, opened 2018-05-09
Updated link to CFSSL

pcgeek86:
The old link to CFSSL simply redirected to a couple of different pages. The GitHub project seems like the most logical place to link to.


cblecker:
The existing link is correct. There are additional links in the top bar to binary packages, if you want to use them instead of compiling from source. This is also listed as the homepage from their github project.

/close

PR #2146, opened 2018-05-11
Fix incorrect path in api_changes doc

austbot:
Corrects location of core defaults.go based on current kubernetes/kubernetes repo.

### DIFF
--- a/contributors/devel/api_changes.md
+++ b/contributors/devel/api_changes.md
@@ -365,10 +365,14 @@ being required otherwise.
 ### Edit defaults.go
 
 If your change includes new fields for which you will need default values, you
-need to add cases to `pkg/apis/<group>/<version>/defaults.go` (the core v1 API
-is special, its defaults.go is at `pkg/api/v1/defaults.go`. For simplicity, we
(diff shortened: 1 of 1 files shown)


contributors/devel/api_changes.md
```diff
@@ -366,7 +366,7 @@ being required otherwise.
 
 If your change includes new fields for which you will need default values, you
 need to add cases to `pkg/apis/<group>/<version>/defaults.go` (the core v1 API
-is special, its defaults.go is at `pkg/api/v1/defaults.go`. For simplicity, we
+is special, its defaults.go is at `pkg/apis/v1/defaults.go`. For simplicity, we
```
cblecker: The change as you have it is incorrect. There is no file at that default location. I believe this "special case" no longer exists, and the core v1 api is at `pkg/apis/core/v1/defaults.go` like everything else.

/cc @deads2k @liggitt

austbot: @cblecker you are right! core would now be a <group> I'll update pr to remove special case. Sorry for mis reading the path.

PR #2153, opened 2018-05-15
Remove doc that is in perf-tests repo now

wojtek-t:
/assign shyamjvs


wojtek-t:
@cblecker - shouldn't I be able to approve it as a sig-scalability lead?

cblecker:
@wojtek-t yes, you should be able to `/approve`

PR #2177, opened 2018-05-22
KEP: Namespace Population

easeway:
## Summary

Namespace Population is an automated mechanism to make sure the predefined policy objects
(e.g. NetworkPolicy, Role, RoleBinding) are present in selected namespaces.

## Motivation

Kubernetes users create namespaces (self-service namespace creation) and want to
ensure security/isolation policies are enforced between them and other users.
Just like addon manager creates cluster-scope policy objects like PodSecurityPolicy,
ClusterRole, ClusterRoleBinding at cluster creation time.
It creates namespace-scope policy objects during namespace creation time.

This is part of [Security Profile proposal](https://github.com/kubernetes/community/pull/2052)

/sig auth
/sig api-machinery
/sig cluster-lifecycle

/cc @davidopp 
/cc @tallclair 
/cc @ericchiang 
/cc @liggitt 
/cc @bgrant0607 
/cc @roberthbailey 
/cc @lavalamp

### DIFF
--- a/keps/draft-20180426-namespace-template.md
+++ b/keps/draft-20180426-namespace-template.md
@@ -1,0 +1,684 @@
+---
+kep-number: draft-20180426
+title: Namespace Population
+authors:
+  - "@easeway"
+owning-sig: sig-auth
+participating-sigs:
+  - sig-api-machinery
+  - sig-cluster-lifecycle
+reviewers:
+  - "@davidopp"
+  - "@tallclair"
+  - "@ericchiang"
+  - "@liggitt"
+  - "@roberthbailey"  
+approvers:
+  - "@tallclair"
+  - "@ericchiang"
+  - "@liggitt"
+  - "@roberthbailey"
+editor: "@easeway"
+creation-date: 2018-04-26
+status: provisional
+---
+
+# Namespace Population
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [User Stories](#user-stories)
+      * [Security Defaults](#security-defaults)
+    * [Implementation Details](#implementation-details)
+      * [Namespace Creation Flow](#namespace-creation-flow)
+      * [The Controller](#the-controller)
+      * [Namespace Match](#namespace-match)
+      * [Apply the template](#apply-the-template)
+      * [Schema validation](#schema-validation)
+      * [Opt-out](#opt-out)
+      * [Namespace initialization](#namespace-initialization)
+      * [Extending Initializers](#extending-initializers)
+      * [Manual Update of Objects](#manual-update-of-objects)
+* [Graduation Criteria](#graduation-criteria)
+
+## Summary
+
+Namespace Population is an automated mechanism to make sure the predefined policy objects
+(e.g. NetworkPolicy, Role, RoleBinding) are present in selected namespaces.
+
+## Motivation
+
+It's a common problem that Kubernetes cluster administrators want to apply pre-defined
+policy objects into namespaces, for example, setting up default ResourceQuota,
+binding `default` service account with a PodSecurityPolicy,
+restricting network ingress/egress using a NetworkPolicy etc.
+
+Today, people can achieve that in a few ways:
+
+1. Use scripts or automation tools (e.g. Terraform) to create a namespace and then populate
+  these policy objects;
+2. Adopt a namespace request process: users request namespaces and cluster admins create for
+  them with all policy objects populated;
+3. Build some in-house services/APIs that users can use to create a namespace by themselves.
+
+The approach `#1` is very basic,
+and assumes all users will collaboratively use these scripts in a good way;
+`#2` introduces more controls over `#1` at the cost of more lengthy process;
+`#3` works pretty automated, but requires significant development effort,
+and result in a non-standard way that users must learn before they can use.
+
+This proposal demonstrates a solution that achieves the same result with standard approach
+that the users already know:
+
+```
+kubectl create namespace my-namespace
+```
+
+The above command will create the namespace and also populated all policy objects pre-defined
+by cluster administrators.
+
+Please read forward for more details in [User Stories](#user-stories) section.
+
+### Goals
+
+- The namespace population mechanism described in this proposal is effective to
+populate identical configurations into large number of namespaces;
+- It's also effective to populate different sets of predefined configurations into
+different sets of namespaces;
+- The mechanism is agnostic to object types as long as they are namespace scoped;
+- The mechanism works in a single cluster;
+- This proposal doesn't require any changes in core Kubernetes.
+
+#### MVP Scope
+
+- `NamespaceTemplate` is applied into newly created namespaces; the existing namespaces are left as is;
+- All policy objects are defined inline in `NamespaceTemplate` custom resource, no external sources of policies supported;
+- Not support namespace matching - apply to all newly created namespaces;
+- NO opt-out mechanism;
+- Not watching changes to policy objects already created.
+
+### Non-Goals
+
+- The namespace population mechanism is NOT effective for customizing namespaces individually;
+- Limit the number of namespaces to be created by users/groups (defend against namespace exhaustion or malicious namespace creators);
+- Conflicts of namespace names;
+- No dynamic objects (generating the object definition during population,
+e.g. string substitution) is supported - all objects are defined statically, 
+except `$(CREATOR)` and `$(NAMESPACE)` (explained in [Apply the template](#apply-the-template));
+- No validation is performed over static object definitions, until they are applied;
+- The mechanism does NOT work across clusters.
+- This proposal relies on initializers mechanism in core Kubernetes, 
+  however, the deprecation/change of initializers in core Kubernetes is out-of-scope and should be discussed in a separate proposal,
+  as this proposal doesn't require any core Kubernetes changes.
+
+## Proposal
+
+The namespace population is performed by defining one or more cluster-scope
+Custom Resource with kind `NamespaceTemplate`,
+and deploy a controller in the cluster.
+
+The Custom Resource Definition of `NamespaceTemplate`:
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: namespacetemplates.policy
+spec:
+  group: policy
+  version: v1alpha1
+  scope: Cluster
+  names:
+    plural: namespacetemplates
+    singular: namespacetemplate
+    kind: NamespaceTemplate
+    shortNames:
+    - nstpl
+```
+
+The example of `NamespaceTemplate`:
+
+```yaml
+apiVersion: policy/v1alpha1
+kind: NamespaceTemplate
+metadata:
+  name: default
+spec:
+  namespaces:
+    # labelSelector selects namespaces by labels to populate defined objects
+    labelSelector: {}
+  templates:
+    # list of objects to be populated into namespaces defined here
+    - apiVersion: networking.k8s.io/v1
+      kind: NetworkPolicy
+      metadata:
+        name: default
+      spec:
+        podSelector: {}
+        policyTypes: ["Ingress", "Egress"]
+    - apiVersion: rbac.authorization.k8s.io/v1
+      kind: RoleBinding
+      metadata:
+        name: use-podsecuritypolicy
+      roleRef:
+        apiGroup: rbac.authorization.k8s.io
+        kind: ClusterRole
+        name: use-psp-default
+      subjects:
+        - apiGroup: rbac.authorization.k8s.io
+          kind: Group
+          name: system:serviceaccounts
+    - apiVersion: rbac.authorization.k8s.io/v1
+      kind: RoleBinding
+      metadata:
+        name: creator
+      roleRef:
+        apiGroup: rbac.authorization.k8s.io
+        kind: ClusterRole
+        name: admin
+      subjects:
+        - apiGroup: rbac.authorization.k8s.io
+          kind: User
+          name: '$(CREATOR)'
+```
+
+There's a size limit of any object created inside Kubernetes (due to etcd).
+If there are too many policy objects inlined in the template, the creation of `NamespaceTemplate` will fail.
+In practice, it's uncommon to hit the limit.
+If that's the case, it's recommended the cluster admin redesign the `NamespaceTemplate` into multiple objects
+and separate policy objects into different tiers of groups.
+
+Alternatively, the policy objects can be defined in some other places, for example:
+
+- As real policy objects in a real namespace which serves as a template only.
+  This approach has advantages that all policy objects are handled through Kubernetes APIs with all supported features, like schema validation.
+  And is more intuitive for cluster admin to understand.
+  Easier to control the access via RBAC.
+  However it's difficult to dynamically create a policy object (e.g. RoleBinding for the creator) according to
+  the creator and namespace name.
+- Remote file server, git etc. and use some tools like kinflator to compose on the fly
+  Out-of-scope in this proposal.
+
+The `NamespaceTemplate` is potentially extensive to support arbitrary source of policy objects,
+by introduce a property `source` to define how to locate the policy objects.
+This is out-of-scope in this proposal.
+
+### User Stories
+
+#### Security Defaults
+
+The cluster administrator manages a shared Kubernetes clusters and allows
+developers self-service to create namespaces by themselves, with predefined
+NetworkPolicy, RoleBinding using PodSecurityPolicy, ResourceQuota,
+ServiceAccounts and Roles present in the namespace, as security defaults.
+
+The cluster administrator creates `NamespaceTemplate` with policy objects in `templates`.
+With the help of `NamespaceTemplate` enforcement controller, these objects are
+created automatically when a user creates a new namespace.
+
+### Implementation Details
+
+#### Namespace Creation Flow
+
+The overall namespace creation flow with `NamespaceTemplate` is:
+
+1. A namespace creator creates a namespace, e.g. with CLI `kubectl create namespace ns1`
+2. The _MutatingAdmissionWebhook_ intercepts the creation request, and
+   - add label `authorization.k8s.io/creator=username`
+   - add `namespace-template` to `metadata/initializers.pending` list
+3. The `NamespaceTemplate` controller is notified of creation of namespace, it
+   - find all matching `NamespaceTemplate` objects by labels and sort them by name
+   - Concate policy objects defined in these objects in the order
+   - Create these policy objects into the namespace
+   - Remove `namespace-template` from `metadata/initializers.pending` list
+4. The client gets the response that the namespace is created
+
+#### The Controller
+
+A controller is deployed in the cluster (recommended in the `kube-system` namespace).
+It runs with a service account with privilege to
+
+- get/list/watch Namespace objects
+- get/list/watch NamespaceTemplate objects
+- get/list/create/update/patch/delete namespace-scope objects in all namespaces, except `kube-system`
+
+The controller watches Namespace and NamespaceTemplate objects,
+when a namespace is added/updated, it applies the objects in namespace templates matching the namespace;
+when a namespace template is added/updated, it applies objects in the template to all matching namespaces.
+
+#### Namespace Match
+
+The `NamespaceTemplate` uses `labelSelector` to efficiently select namespaces.
+This design is not proposing extra selection mechanism, 
+even though it's almost impossible to select/deselect a namespace without labels.
+This problem is out-of-scope here.
+
+When multiple `NamespaceTemplate` objects are matched, the items in the templates are concated
+according to the order of names of `NamespaceTemplate` objects (ascend).
+It's cluster admin's responsibility to design the policy objects carefully into multiple `NamespaceTemplate` objects,
+the population mechanism is not aware of any conflicts introduced by the policy objects.
+
+Using `labelSelector` imposes a potential security concern that 
+a malicious namespace creator creates a namespace with carefully selected labels to pick the  `NamespaceTemplate` which is different from cluster admin's intention.
+And it's also possible the owner of namespace later updates the labels to matching a different set of
+`NamespaceTemplate` from those at creation time.
+
+To mitigate the problem, the RBAC system can be utilized to define the permission that a namespace creator must have to use certain `NamespaceTemplate`.
+The mechanism introduces the `use` verb in RBAC ClusterRole to express the permission to access a `NamespaceTemplate`. (Similar to PodSecurityPolicy)
+
+Here's an example:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: use-namespace-template-default
+rules:
+  - apiGroups: ["poliy"]
+    resources: ["namespacetemplates"]
+    resourceNames: ["default"]
+    verbs: ["use"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: namespace-template-usage
+roleRef:
+  kind: ClusterRole
+  name: use-namespace-template-default
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: User
+  name: "user1"
+  apiGroup: rbac.authorization.k8s.io
+```
+
+The namespace creation flow becomes:
+
+1. A namespace creator creates a namespace, e.g. with CLI `kubectl create namespace ns1` with or without certain labels
+2. The _MutatingAdmissionWebhook_ intercepts the creation request, and
+   - add label `authorization.k8s.io/creator=username`
+   - add `namespace-template` to `metadata/initializers.pending` list
+3. The _ValidatingAdmissionWebhook_ 
+   - find out all matched `NamespaceTemplate` objects
+   - verify if the creator is allowed to apply **ALL** of the matched `NamespaceTemplate`. It fails the request if the creator doesn't have `use` permission on **ANY** of the mached `NamespaceTemplate`.
+4. Same as the original flow
+
+Using this mechanism, the cluster admin is able to design `NamespaceTemplate` intended to be used by a limited number of users.
+
+For example
+
+```yaml
+apiVersion: policy/v1alpha1
+kind: NamespaceTemplate
+metadata:
+  name: nonprivileged
+spec:
+  namespaces:
+    # matches namespaces without being labeled as privileged
+    labelSelector:
+      matchExpressions:
+      - key: namespace-class
+        operator: NotIn
+        values: [privileged]
+  templates: [] # ... details omitted
+---
+apiVersion: policy/v1alpha1
+kind: NamespaceTemplate
+metadata:
+  name: privileged
+  annotations:
+    policy/namespace-template/require-permission: true
+spec:
+  namespaces:
+    # matches namespaces being labeled as privileged
+    labelSelector:
+      matchLabels:
+        namespace-class: privileged
+  templates: [] # ... details omitted
+```
+
+To create a namespace allowing privileged workloads, 
+the creator must put `namespace-class=privileged` in labels of namespace.
+To limit the users who are allowed to create _privileged_ namespaces, 
+the cluster admin sets additional RBAC rules:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: privileged-namespace-creator
+rules:
+  - apiGroups: ["policy"]
+    resources: ["namespacetemplates"]
+    resourceNames: ["privileged"]
+    verbs: ["use"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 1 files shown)


keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: "system:serviceaccount:kube-system:name"
+  apiGroup: rbac.authorization.k8s.io
+```
+
+Using external admission webhook increases the complexity, and also latency.
+As the webhook must be reliable all the time, it must be deployed in HA.
+It's recommended to create a built-in admission controller for namespace initialization.
```
smarterclayton: I don't think this statement is true.  I certainly don't think we should put this admission controller in code if it can be built with an external admission controller.

I'm on the fence about whether this feels like a part of Kube or not, but if this is a controller and a CRD and is plugging in to extension, I don't see anything that requires built in web hook admission yet, certainly not until it graduates from beta.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+The mechanism is agnostic to object types as long as they are namespace scoped.
+The mechanism works in a single cluster.
+
+### Non-Goals
+
+The namespace population mechanism is NOT effective for customizing namespaces
+individually.
+No dynamic objects (generating the object definition during population,
+e.g. string substitution) is supported - all objects are defined statically, 
+except `$(CREATOR)` and `$(NAMESPACE)`.
```
smarterclayton: Please reference prior proposals here on templating.  This is still somewhat problematic - you'll need to define what the rules are here, what fields it can be applied to, what happens when it is applied incorrectly (like to an array or to a string key).

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+    labelSelector: {}
+    # excludes lists names of namespaces which must be excluded from being applied 
+    # with current namespace template.
+    excludes: []
+  common:
+    # labels to be injected into all objects defined in templates
+    labels: {}
+    # annotations to be injected into all objects defined in templates
+    annotations: {}
+  templates:
```
smarterclayton: This object has some scaling problems - include at minimum a discussion on what would happen at the max limit for templates.  Also you'll need to define ordering (is ordering in this significant) in case you didn't already and I just missed it.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+metadata:
+  name: default
+spec:
+  namespaces:
+    # labelSelector selects namespaces by labels to populate defined objects
+    labelSelector: {}
+    # excludes lists names of namespaces which must be excluded from being applied 
+    # with current namespace template.
+    excludes: []
+  common:
```
smarterclayton: I'd rather not do this - it's too clever.  Users can easily duplicate labels and be precise.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+creation request. 
+It adds an annotation `policy/namespace-creator=name` to the namespace being created.
+This annotation is immutable from further update/patch requests.
+
+#### Readiness of a namespace
+
+A newly created namespace is not ready before all objects in `NamespaceTemplate`
+are fully populated.
+To determine the readiness of a namespace, simply inspect `phase` in `status`
+subresource of the namespace. It's ready when the `phase` is `Active`.
```
derekwaynecarr: There are only two phases: active and terminating.  Are you saying there will be a third phase?

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+apiVersion: policy/v1alpha1
+kind: NamespaceTemplate
+metadata:
+  name: default
+spec:
+  namespaces:
+    # labelSelector selects namespaces by labels to populate defined objects
+    labelSelector: {}
+    # excludes lists names of namespaces which must be excluded from being applied 
+    # with current namespace template.
```
smarterclayton: This doesn't scale at all.  I don't know that this is worth it.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+This annotation is immutable from further update/patch requests.
+
+#### Readiness of a namespace
+
+A newly created namespace is not ready before all objects in `NamespaceTemplate`
+are fully populated.
+To determine the readiness of a namespace, simply inspect `phase` in `status`
+subresource of the namespace. It's ready when the `phase` is `Active`.
+Alternatively, the `NamespaceTemplate` controller is registered as an _Initialization Controller_
+and uses `Initializer` admission control to hide the namespace before it's fully populated.
```
derekwaynecarr: Please don’t do this (hide ns)

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+- kind: User
+  name: "$(CREATOR)"
+  apiGroup: rbac.authorization.k8s.io
+```
+
+There is one issue that when the controller gets informed to create objects inside a new
+namespace from `NamespaceTemplate`, it has no idea who created the namespace.
+To be able to substitute `$(CREATOR)`, a `MutatingAdmissionWebhook` is involved in namespace
+creation request. 
+It adds an annotation `policy/namespace-creator=name` to the namespace being created.
```
smarterclayton: This should be a proper kubernetes namespace, probably "kubernetes.io".  I don't see why namespace creation is special here.  And I don't know that this is the correct naming "creator" - it probably needs to match the syntax and semantics of the user object.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+The webhook will fail the request if the opt-out annotation is present without the RBAC binding for the requestor.
+
+#### Race Condition
+When a namespace is created, there's a gap before all objects in the namespace
+templates are populated.
+During the gap, the users may have the access to create objects into namespaces.
+
+To avoid the situation, a few mechanisms can be used:
+
+##### Creating namespace through service
```
derekwaynecarr: We have a similar concept in openshift via a feature called projectrequest

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+  name: "$(CREATOR)"
+  apiGroup: rbac.authorization.k8s.io
+```
+
+There is one issue that when the controller gets informed to create objects inside a new
+namespace from `NamespaceTemplate`, it has no idea who created the namespace.
+To be able to substitute `$(CREATOR)`, a `MutatingAdmissionWebhook` is involved in namespace
+creation request. 
+It adds an annotation `policy/namespace-creator=name` to the namespace being created.
+This annotation is immutable from further update/patch requests.
```
smarterclayton: This is a pretty big deal, it means that any naive code that changes the annotation on namespaces will break when this is introduced.  I'm not sure we should be doing a one off "this annotation is special" approach without a bit more discussion on next steps.

keps/draft-20180426-namespace-template.md
smarterclayton: It's not clear to me whether you are proposing adding a new phase.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+##### Setup RBAC properly
+
+See above [Self-serviced namespace creation](#self-serviced-namespace-creation)
+for details of RBAC setup.
+
+This approach requires pre-setup work from cluster admin for ClusterRoleBinding to grant
+users/groups the privilege to create namespaces.
+Beyond that, it is the simplest and recommended approach.
+
+##### Use of Initializers
```
derekwaynecarr: I feel like just creating namespace through delegated service avoids the need for this complexity

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+
+There are other alternatives that the objects can be defined somewhere else from `NamespaceTemplate`.
+The implementation can be pluggable (e.g. adding a `source` property) to support different sources of templates.
+While the initial proposal will focus on inlined objects.
+
+The MVP scope covers the inlining templates only.
+
+#### Opt-out
+
+Cluster administrator is able to opt-out some namespaces from being populated by the controller.
```
smarterclayton: I don't think the mechanism described here is going to fit well in existing clusters. 

Since this mechanism does nothing to rate limit or control namespace creation under quota, your proposed solution doesn't really solve "defend against malicious creators", so I don't think namespace template should default to covering all namespaces or attempt to solve that problem.  Instead, this proposal is about "automated defaulting", and so it needs to be opt in by the user specifying a label.  If the administrator sets a namespace template label that intersects with a system namespace, that's the administrator's fault.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+### Goals
+
+The namespace population mechanism described in this proposal is effective to
+populate identical configurations into large number of namespaces.
+It's also effective to populate different sets of predefined configurations into
+different sets of namespaces.
+The mechanism is agnostic to object types as long as they are namespace scoped.
+The mechanism works in a single cluster.
+
+### Non-Goals
```
smarterclayton: I think you need to add as a non-goal (because this solution can't provide it)

* Defend against namespace exhaustion or malicious namespace creators.

and any elements of the proposed solution that overlap should be dropped.  Instead it is a defaulting solution that could be composed with a future change that defends namespace creation itself via some pattern or hook.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+or it's almost impossible to customize the way how they create namespaces,
+an automated opt-out mechanism is required.
+The mechanism is based on the service account running the controllers.
+With a `MutatingAdmissionWebhook`, 
+a namespace created by these controllers will be automatically annotated for opt-out.
+This mechanism is out of MVP scope.
+
+#### Self-serviced namespace creation
+
+Self-serviced namespace creation allows certain users/groups to create namespaces
```
smarterclayton: I don't think this proposal as written can correctly protect against this, so it either needs to be dropped, split, or a lot more work put into it.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,513 @@
...
+
+The objects created by namespace population may be altered manually or through
+other ways.
+The recommended way is setting up RBAC properly to mitigate the risk.
+The proposal doesn't watch update of objects already created.
+The update may be reverted in the case triggers template re-apply (namespace change or template change).
+
+## Graduation Criteria
+
+Namespaces can be securely populated.
```
smarterclayton: This is something this proposal doesn't provide, so this needs to be rethought.

keps/draft-20180426-namespace-template.md
smarterclayton: Self-service namespace creation is a much harder problem than this proposal.

keps/draft-20180426-namespace-template.md
derekwaynecarr: Agree with opt-in

keps/draft-20180426-namespace-template.md
easeway: I'm OK with building an external webhook. I will remove this statement.

keps/draft-20180426-namespace-template.md
easeway: The intention was for other components to specify labels they are interested in when integrating namespace population mechanism, not for users. I will remove it, as it's not necessary for now.

keps/draft-20180426-namespace-template.md
easeway: I agree. Mentioning this as an alternative. We will not go this path.

keps/draft-20180426-namespace-template.md
easeway: With help of a delegated service, it makes the problem simpler. However that requires the delegated service, with concerns in some use cases:

1. The user experience is inconsistent using delegated service, as in different environment/teams/corps, the delegated services are different
2. The cluster admin will either spend effort developing their own delegated service or use an existing solution, and they have to educate the users of the cluster about how to create a namespace.

In a self-service namespace creation use case, the users of the cluster want to simply use "kubectl" to create namespaces.

keps/draft-20180426-namespace-template.md
davidopp: @bgrant0607 had suggested that we could put the namespace templates in `kube-system` namespace. In that case we could use ResourceQuota to protect against creating too many. (His suggestion wasn't related to quota, but I think it would allow quota.) AIUI intention is that only cluster admin can create namespace templates, since they are part of the security profile.

keps/draft-20180426-namespace-template.md
davidopp: I may be misunderstanding, but I believe that making this opt-in runs counter to the goal here, which is to allow the cluster admin to force policies on new namespaces.

Of course obviously it's opt-in in the sense that if the cluster admin decides not to use the feature, then nothing happens, i.e. no change from today. 

It seems that your main objection to this proposal is that it does not address quota for namespace creation. Am I understanding correctly?

keps/draft-20180426-namespace-template.md
davidopp: @smarterclayton What is the "this" you're referring to where you say "protect against this"? I think you are referring to something that is not in this sentence, but I wasn't sure exactly what.

Certainly people are doing self-service namespace creation today on stock upstream Kubernetes clusters, so I assume you are referring to doing that with some additional guarantees. Can you specify which guarantees you think this proposal does not cover? The one you have mentioned elsewhere is quota on namespace creation. Are there others?

keps/draft-20180426-namespace-template.md
davidopp: @smarterclayton Would calling this something like `authentication.alpha.kubernetes.io/creator` and putting the username, uid, groups, and extra (assuming all of those are available an admission controller) as the value, satisfy what you're asking for?

keps/draft-20180426-namespace-template.md
davidopp: @smarterclayton Is there an aspect of security that is missing other than quota on namespace creation?

keps/draft-20180426-namespace-template.md
tallclair: Crazy idea: what if the template was a namespace? I.e. have some way of indicating that a namespace is a template and shouldn't actually be instantiated, and then clone all its objects into new namespaces when the template applies.

smarterclayton:
Jordan and I had a discussion about the security aspects and I think we felt we had a good way out of the problems we have today that ends up better for everyone.

For context, in the “untrusted user” case there are really three things that need to be defended:

1. The user in question can’t bypass or abuse namespace creation itself
2. The user can’t act on the content in the namespace before it is fully initialized - Ie they can’t create pods before quota is created
3. They can’t muck with the metadata of the namespace such that they can abuse 1 or 2

There’s a follow on problem - any solution that enforces 1-3 can’t create a usability gap so that the Kubernetes api behaves sufficiently differently for admins vs untrusted users.  The initializers proposal was created to solve this problem - even though we want to roll that back because it’s overly general, we can use its pattern to solve this problem.

So how do we bring this together?

1 can be solved with an external admission webhook (trivial).

2 Can be solved by not initializing the users role on the namespace until all content is protected - rbac saves the day.

3 can be handled via a mutating admission webhook - we can either be generic or not, but it’s tractable.

The problem that is left is that even with the above, an untrusted user who sees the namespace created might try to act on that namespace and get forbidden errors until the content is initialized (simple example is kubectl create with a namespace at the top and a list of resources in the namespace - all the namespaces resources would fail with forbidden errors because they are racing defaulting). We can change all clients, but that’s somewhat impractical at this stage in our lifecycle.  We can’t add a new phase because that’s not backwards compatible for old clients.

However, we can take the basics of initializers and special case them for namespaces.  Specifically, we want to record remaining initializers as a spec field on namespace, and then delay returning the original creation response until all initializers are removed or a timeout has been hit.  Then a naive client just waits a bit longer for namespace create to return, and all other operations complete.

The initializer proposal already covers this, and practical experience showed we can at least make the delay work.  By focusing initializers on namespaces as a spec field, we are consistent with namespace finalization and ensure we can provide a solution for all four problems without existing clients having to change.

keps/draft-20180426-namespace-template.md
smarterclayton: Unfortunately we can't change phase (it's an enum, so clients won't expect a new phase).  See my comments at the bottom - I think we want to adapt the spirit of initializers (which were designed around this use case) without requiring the existing initializer concept (of which this is the primary use case we have, and we can do it as first class on namespace rather than implicit).

keps/draft-20180426-namespace-template.md
smarterclayton: After talking with jordan I'm almost convinced we can kill project request off in favor of this, with the security caveats.

easeway:
Thanks @smarterclayton, your comment reveals a lot of design thoughts behind the scene which I didn't understand before.

> The problem that is left is that even with the above, an untrusted user who sees the namespace created might try to act on that namespace and get forbidden errors until the content is initialized (simple example is kubectl create with a namespace at the top and a list of resources in the namespace - all the namespaces resources would fail with forbidden errors because they are racing defaulting).

This is a really good point that clears out my confusion why the Initializer delays the response of the creation request. Now I understand your concerns about adding a new phase.

> We can’t add a new phase because that’s not backwards compatible for old clients.

This is a good point, though I don't fully agree on this. IMO it's not backward compatible problem, but an issue that the behavior/implementation of the old clients is wrong. I think the API is designed in a good way, as the status of namespace has `phase`. It's a good design that reserves the space for future extension. A properly implemented client should respect this, even it doesn't care about other phases (or future phases), it should respect the value it cares about (e.g. Active). And it should wait until the phase turn out the value it expects. I don't think it should be a reason preventing us from introducing reasonable new phases, and I don't think Initializer should delay the response for client with incorrect implementation. We should fix incorrect implementations, starting from kubectl. And the owners of other clients should be responsible for fixing the issue.

This also reminds me one special resource type CustomResourceDefinition. It's tricky, because when a CRD is created using kubectl, the client has to wait because it takes a while before the new type is populated across the cluster. So applying in a manifest with CRD sits on the top, followed by other resources using the CRD will always fail because the new type defined by CRD is not available immediately. I mention that as it already demonstrates the existence of delay scenario in the Kubernetes resource system, and it's reasonable for namespace fall into the similar situation.

easeway:
BTW, the mechanism introduced by Initializer is so useful for namespace population. As it's being deprecated, I think we should propose something specific, like NamespaceInitializer, and take more security into consideration. E.g. not hiding the namespace, but only allow whitelisted initialization controllers (by checking the service accounts) to populate objects into the namespace, not other users.

smarterclayton:
> This is a good point, though I don't fully agree on this. IMO it's not backward compatible problem, but an issue that the behavior/implementation of the old clients is wrong. I think the API is designed in a good way, as the status of namespace has phase. It's a good design that reserves the space for future extension

See for more description why (about phase).  https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#typical-status-properties  Phases can't be changed safely because we still have to support broken clients.  It's just one of those Kubernetes things (we don't use phase for new Apis because it can't be safely extended).

> And the owners of other clients should be responsible for fixing the issue.

Unfortunately, our strict API compatibility story prevents us from ever doing things like this - we take the hard line on API because it reduces pain for us and for users.  A future v2 namespace might address this, but for now we'll have to live with what we have.

> I mention that as it already demonstrates the existence of delay scenario in the Kubernetes resource system, and it's reasonable for namespace fall into the similar situation.

Yes, CRDs are ugly in this respect.  It's something we might address in the future.

> BTW, the mechanism introduced by Initializer is so useful for namespace population. As it's being deprecated, I think we should propose something specific, like NamespaceInitializer, and take more security into consideration

Yes, I'm effectively proposing we update this KEP to include the rough idea that:

1. we add a spec field to namespace `initializers` that has a similar structure to `metadata.initializers`
2. we describe that we will delay the return of a `POST /namespaces` call up to X seconds or when initializers field is empty, using the same underlying implementation as generic initializers
3. we describe how someone would implement a webhook admission plugin that controls namespace addition

and then that removes the last use case that generic initializers solves better than web hook, which allows us to declare generic initializers as dead and remove them.

keps/draft-20180426-namespace-template.md
easeway: Details in _Apply the template_ section. Will update.

keps/draft-20180426-namespace-template.md
easeway: Remove the `excludes`. This problem should not belong to here. I will call out in Non-Goals.

smarterclayton:
With those changes to the proposal I think this would meet the minimum bar for both security and extensibility.  We also need a section in the proposal on properly securing admission controllers - ie how do we prevent a namespace creator from bypassing label selection in detail and a minimum proposal for limiting label changes.

easeway:
Thanks @smarterclayton ! What you asked (securing admission controllers, limiting label changes) is a little tricky. I updated the proposal with a new section _Label Restrictions_ with a solution which extensively uses RBAC for labels - that doesn't require any change of existing code, but I'm not sure because no body is using RBAC in that way (maybe PodSecurityPolicy is an example which introduces a new verb _use_, while mine extends the subresource semantics). Please advise.

easeway:
Some good feedback received from policy working group meeting:

> Need details how namespace initialization mechanism is pluggable

Described the details in the doc, though not in MVP scope

> The label protection is more complicated, move details to a separate proposal

Use a better solution. No longer need to protect labels (except the creator one). The `Label Restrictions` section is removed, and all details merged into `Namespace Match` section

> Limit the MVP scope to cover new namespaces, not existing namespaces

Agreed and called out.

> Mention how to handle the upper limit of all inlined templates in one object

Added details in the proposal.

> Alternative mechanisms to compose the templated objects instead of inlining (kinflator?)

Mentioned some thoughts in the proposal, though not MVP scope.

keps/draft-20180426-namespace-template.md
```diff
@@ -0,0 +1,657 @@
...
+  name: "system:serviceaccount:kube-system:name"
+  apiGroup: rbac.authorization.k8s.io
+```
+
+#### Extending Initializers
+
+Kubernetes provides resource initialization mechanism
+(injecting items into `metadata/initializers.pending` list) in a _MutatingAdmissionWebhook_ to
+allow controllers to finish initialization work before the creation request is completed and response is 
+delivered to the client.
```
jbeda: Last I heard initializers were alpha not going to move forward.  Am I reading right that this is depending on those?

jbeda:
Some procedural questions:

* If I'm reading this right there are no changes to core k8s APIs and build code into our core control plane components.  That is good as it means that the level of impact here is a lot less.
* Has this been agreed by folks in SIG-auth to sponsor this KEP?  Apparently this was discussed as part of a larger discussion but taking on this as a separable sub-project doesn't look to be in the notes from SIG-auth that I see.
* I assume that this would show up as a sub-project to SIG-auth?

I want to note that, as designed, there is no need to use the KEP process or have this be part of official upstream Kubernetes at all.  This could easily also be an Associated Repository or exist outside of the Kubernetes governance model altogether.  See [Kubernetes Repositories](https://github.com/kubernetes/community/blob/master/kubernetes-repositories.md) for details on the different types of repos and how this relates to subprojects.

If you do want to continue this as a KEP, it would be great to get *something* checked in right away to agree on the scope of the KEP before diving into the details.  That also is a way to get an explicit ACK on the set of approvers/reviewers for the KEP and the involved SIGs.

/cc @calebamiles on the KEP process.

keps/draft-20180426-namespace-template.md
easeway: This doesn't depends on Initializers admission controller (alpha feature). It's proposed to use the metadata.initializers mechanism: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go

jbeda: It isn't clear to me (and maybe @smarterclayton can shine a light) that the admission controller and the `metadata.initializers` mechanism are separate.  My assumption/understanding is that these were both alpha.  Is the mechanism you are relying on really GA?

jbeda:
Reading the comments around initializers it sounds like that *may* require some changes to the core API machinery.

Above @smarterclayton says:
> However, we can take the basics of initializers and special case them for namespaces. Specifically, we want to record remaining initializers as a spec field on namespace, and then delay returning the original creation response (in the storage layer, very similar to how it is done for generic initializers today) until all initializers are removed or a timeout has been hit. Then a naive client just waits a bit longer for namespace create to return, and all other operations complete. Initialization needs to be fast anyway

It would be great to call out this very explicitly and make sure that this type of thing is understood and reviewed.  It also might be useful to split that into its own KEP to ensure that generic built in features like that are generic enough to be useful beyond just a single scenario like this and that it is justified to extend our core resource/API work flow.

(But I could be reading this wrong -- if so I'm sorry for the distraction!)

easeway:
@smarterclayton cloud you please clarify:

1. Is metadata.initializers mechanism (https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go) alpha, same as Initializers admission controller?
2. Are you proposing adding Initializers into NamespaceSpec, because metadata.initializers is alpha and going to be deprecated?

If yes, then the core API change is required because of adding Initializers field in NamespaceSpec.

smarterclayton:
For clarity, we have generally agreed to remove alpha generic initializers
because they don’t offer enough value over webhooks.  This is a more
targeted namespace-specific lifecycle with is consistent to the special
nature of namespaces as a “ACL bucket” that doesn’t apply to other objects.

I suppose we could preserve the generic field, but I might argue that each
object has its own lifecycle (pods vs ns vs nodes) and so init containers
vs namespace initialization are both domain specific.  Also, the things
that use objects are inextricably linked to how they are initialized - a
Kube proxy behaves one way with services as opposed to how an ingress
behaves with secrets.


On Jun 21, 2018, at 2:03 PM, Joe Beda <notifications@github.com> wrote:

Reading the comments around initializers it sounds like that *may* require
some changes to the core API machinery.

Above @smarterclayton <https://github.com/smarterclayton> says:

However, we can take the basics of initializers and special case them for
namespaces. Specifically, we want to record remaining initializers as a
spec field on namespace, and then delay returning the original creation
response (in the storage layer, very similar to how it is done for generic
initializers today) until all initializers are removed or a timeout has
been hit. Then a naive client just waits a bit longer for namespace create
to return, and all other operations complete. Initialization needs to be
fast anyway

It would be great to call out this very explicitly and make sure that this
type of thing is understood and reviewed. It also might be useful to split
that into its own KEP to ensure that generic built in features like that
are generic enough to be useful beyond just a single scenario like this and
that it is justified to extend our core resource/API work flow.

(But I could be reading this wrong -- if so I'm sorry for the distraction!)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/kubernetes/community/pull/2177#issuecomment-399192733>,
or mute the thread

smarterclayton:
I'll open a PR to the initializers proposal to describe the current state and that we desire to deprecate and remove the feature as not having enough value relative to web hooks to justify the additional complexity.  I think this proposal covers the core use case we do care about (namespace initialization).  In the short term, we could use the alpha initializers field in metadata to test this, and then if we reach beta on it we can move the field out of metadata and into namespace (`metadata.initializers` -> `spec.initializers`) as a beta field and then drop metadata.initializers.

Or we could leave the initializers field present, and remove the infrastructure that supports hiding the objects from end users.  We would still have a namespace finalizers field, but others could use `metadata.initializers` as they like with no semantics imposed by the apiserver.

jbeda:
In any case, I think it makes a ton of sense to make this KEP explicit in what the changes to "core" kubernetes are and what is being built outside.  I'd suggest perhaps even splitting it into 2 KEPs to ensure that the "core" features can stand on their own.

easeway:
Updated to explicitly call out this proposal doesn't include anything related to changes in core Kubernetes which should be discussed in a separate proposal.

ericavonb:
Can we be explicit about:
1. What are the exact user problems this proposal aims to solve? 
    * Can you give some user stories, specific examples of users encountering these issues?
2. Why can't they be solved with the current set of features in Kubernetes?
    * What user requirements are left unmet or are compromised to other needs with past attempts to address this issue? 
3. What external solutions have been tried?
    * Why are these attempts insufficient?
    * What blocks users from adopting the successful strategies out there?
    * Why should we have a standardized solution upstream for this problem in particular?
4. What are some other ways these problems could be solved within Kubernetes?
    * What are the pros and cons to each?
    * Why do you recommend this particular solution?

I am especially unclear about no. 2 and 4. It's difficult to evaluate this from a policy management perspective without a better understanding of the exact problem.

re @easeway

easeway:
@ericavonb Updated motivation section

easeway:
@smarterclayton @jbeda Separate KEP filed containing core Kubernetes changes: #2543 and this KEP doesn't contain anything requiring changes in core Kubernetes.

easeway:
@mattfarina #2543 is a dependency of this KEP. #2543 is for core Kubernetes changes required by this feature and this KEP focuses on non-core Kubernetes changes. This can't land without #2543.

justaugustus:

justaugustus:

/close

PR #2178, opened 2018-05-23
KEP test cluster framework

hoegaarden:
**Proposal**
> We propose an abstraction so that users and tools (e.g. CI) have a consistent
> way to spin up and configure clusters they can test against. This abstraction
> is targeted mainly to the needs of the following personas and should work
> equally well used interactively and non-interactively / automated.

See also: https://docs.google.com/document/d/13bMjmWpsdkgbY-JayrcU-e_QNwRJCP-rHjtqdeeoQHo/edit#

/sig testing

/cc @totherme 
/cc @mariantalla
/cc @timothysc 
/cc @liztio
/cc @marun


timothysc:
This is definitely on my todo list, but right now I'm booked trying to get 1.11 out the door.

keps/sig-testing/0014-20180523-test-cluster-framework.md
```diff
@@ -0,0 +1,169 @@
...
+<!--
+TODO: Add note that we did interviews and talked to different people, noted in
+      the "Validation" section of the personas?
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+### Goals
+
+1. To design an abstraction of “standing up a kubernetes cluster” which is:
```
fejta: How does this differ from what kubetest is trying to accomplish?

keps/sig-testing/0014-20180523-test-cluster-framework.md
```diff
@@ -0,0 +1,169 @@
...
+[alice]: https://docs.google.com/document/d/13bMjmWpsdkgbY-JayrcU-e_QNwRJCP-rHjtqdeeoQHo/edit#heading=h.17sl3t5a8nu
+[bob]: https://docs.google.com/document/d/13bMjmWpsdkgbY-JayrcU-e_QNwRJCP-rHjtqdeeoQHo/edit#heading=h.mecveaw12b4d
+[carol]: https://docs.google.com/document/d/13bMjmWpsdkgbY-JayrcU-e_QNwRJCP-rHjtqdeeoQHo/edit#heading=h.4i5imqvwxv3a
+[testing_commons_notes]: https://docs.google.com/document/d/1TOC8vnmlkWw6HRNHoe5xSv5-qv7LelX6XK3UVCHuwb0/edit#
+[the_doc]: https://docs.google.com/document/d/13bMjmWpsdkgbY-JayrcU-e_QNwRJCP-rHjtqdeeoQHo/edit#
+
+## Motivation
+
+With this proposal we would like to address these three points of friction:
+- For Carol, we introduce an abstraction which is intended to unify existing
```
derekwaynecarr: is there something special about a test cluster versus any other cluster?

keps/sig-testing/0014-20180523-test-cluster-framework.md
```diff
@@ -0,0 +1,169 @@
...
+### Goals
+
+1. To design an abstraction of “standing up a kubernetes cluster” which is:
+    - expressive enough that most or all e2e tests ought to be able to use it,
+      rather than standing things up directly
+    - abstract enough that it could in principle be wrapped around many existing
+      methods of standing up clusters
+    - equally easy to use interactively and in CI
+2. To build a lightweight testing framework for CLI use which
+    - is not dependant on `k8s.io/kubernetes`
```
derekwaynecarr: this implies that k8.io/kubernetes vendors this test framework?

keps/sig-testing/0014-20180523-test-cluster-framework.md
```diff
@@ -0,0 +1,169 @@
...
+    - is fully functional
+    - makes use of the abstraction above
+
+
+### Non-Goals
+
+- To wrap any existing “ways of standing up a cluster” in our abstraction
+   - Of course, wrapping existing technology may well be the most efficient way
+     to meet goals 2 and 3 (building CLI and API/Controller testing frameworks). 
+- To port large numbers of existing tests to use our abstraction.
```
derekwaynecarr: absent this, are we just creating a new standard?

BenTheElder:
How does this relate to [the cluster api](https://github.com/kubernetes-sigs/cluster-api)? AFAIK having a consistent way to spin up clusters is a goal already, more broadly than just test clusters.

krzyzacy:
cc @krousey ^^

keps/sig-testing/0014-20180523-test-cluster-framework.md
hoegaarden: We didn't know about `kubetest` before. The intent seems indeed very similar! From an (admittedly very quick) first try there might be room for some UX improvements, specifically for the use case of (new-ish) contributors that want to run it as part of their unit/integration test suite (and would be after fast feedback and low setup overhead).

We'll keep looking at `kubetest` and come back with more specific suggestions.

Is there additional documentation besides the README in the [`test-infra` repo](https://github.com/kubernetes/test-infra/tree/master/kubetest) and the godocs?

cc @mariantalla

keps/sig-testing/0014-20180523-test-cluster-framework.md
hoegaarden: Yes. It should be possible for `k/k` to vendor this framework.

keps/sig-testing/0014-20180523-test-cluster-framework.md
hoegaarden: To understand if this framework is useful we were thinking of porting a couple of tests for each (or most) of the test suites. An example of how such a port could look can be found [here](https://github.com/pivotal-k8s/kubernetes/compare/master...pivotal-k8s:translating-test-cmd-to-cluster-api#diff-efd84624607a7f9efbb2aaeeb004557aR10), where we experimented with porting different tests from `test-cmd.sh`.  And I believe we should port not only tests from `test-cmd.sh` but also other.

But porting *all* the tests is not in the scope of this KEP.

keps/sig-testing/0014-20180523-test-cluster-framework.md
hoegaarden: One example: A partial cluster (running only the apiserver & etcd) might be perfectly fine for a certain type of tests.

hoegaarden:
@BenTheElder 
That's a good point. I think we need to have a close look on the work that sig-cluster-api doing.

Ultimately, users of the test framework would need to be able to describe the shape of clusters they need in their tests, i.e. number of worker nodes, number of master nodes. In this context, we're not clear yet what the relationship would be between Cluster API and the testing framework. Would using the cluster api become an implementation detail of a specific test cluster implementation - put differently, would some cluster types leverage the cluster API and others not? Or, would the testing framework implement the cluster API itself?

cc @mariantalla

BenTheElder:
I'd be concerned that test clusters are being provisioned differently from
actual end user clusters if we have a special test cluster framework, we
try to avoid this today by just having EG kubetest call upstream
provisioners (like cluster/* scripts).

My understanding is that the cluster API will provide portable and
configurable provisioning of this type and the KEP mentions the aim is to
cover all major existing providers / provisioners. I'm not sure why future
tests and test tools (and any other tools!) shouldn't just integrate with
this directly..?

keps/sig-testing/0014-20180523-test-cluster-framework.md
fejta: Yeah `kubetest` is hard to use on its own right now, but I'd be interested in trying to fix that. 

The other documentation is: https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md (`hack/e2e.go` is just a wrapper to call `kubetest`)

keps/sig-testing/0014-20180523-test-cluster-framework.md
timothysc: @derekwaynecarr - Also (N) nodes on a single host is another example.  Similar to other DIND clusters.

keps/sig-testing/0014-20180523-test-cluster-framework.md
timothysc: We're trying to standardize the client tooling abstraction for re-use.  I think there is some overlap of concerns, but there are other constraints we are trying to solve here.

keps/sig-testing/0014-20180523-test-cluster-framework.md
rdodev: Ability to specify cluster and node type for a suite of tests with specific requirements 
 is a huge advantage. If the abstraction is provider agnostic even better.

keps/sig-testing/0014-20180523-test-cluster-framework.md
```diff
@@ -0,0 +1,169 @@
...
+TODO: Add note that we did interviews and talked to different people, noted in
+      the "Validation" section of the personas?
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+### Goals
+
+1. To design an abstraction of “standing up a kubernetes cluster” which is:
+    - expressive enough that most or all e2e tests ought to be able to use it,
```
BenTheElder: our e2e tests do not stand up clusters directly today? they expect a cluster to be available in $KUBECONFIG, tooling above that does standup.

keps/sig-testing/0014-20180523-test-cluster-framework.md
```diff
@@ -0,0 +1,169 @@
...
+the development of a new lightweight testing infrastructure outside
+`k8s.io/kubernetes`.
+
+Meanwhile, Carol, an API/Controller dev (see [Key Persona Carol][carol]), is
+frustrated that there are already too many ways to spin up test clusters.
+
+There is a clear tension between Alice and Carol’s frustrations. Left to their
+own devices, Alice is likely to contribute to the proliferation of test
+infrastructure that frustrates Carol, and Carol may find herself advocating
+that Alice be prevented from acting to resolve her own frustration with the
```
BenTheElder: Perhaps I've misread this persona section, but it reads to me as somewhat hostile to Alice, which I don't approve of.  

I don't think anything we check in as a KEP should suggest preventing others from developing anything, let alone test infrastructure...? Why can't "Carol" let "Alice" build whatever tooling they want?

BenTheElder: This whole section beyond "We propose an abstraction so ..." (anything after the first paragraph) seems fairly unnecessary. I'd suggest a motivation more along the lines of "we'd like a standard framework for spinning up clusters for testing with varying requirements and backing implementations" or something along those lines.

keps/sig-testing/0014-20180523-test-cluster-framework.md
BenTheElder: Additionally, this is necessary for say, conformance test e2es, which must not be aware of the underlying provider at all, just talking to "a cluster" of some sorts, provisioned elsewhere.

stevekuznetsov:
I would second Ben's comments that I think it would be good for this to sync with cluster API as that approach (declarative k8s-style objects backed by controllers) seems to be a very powerful approach and unless there is some architectural reason why clusters that tests run against should be stood up using an unrelated approach that is not also used for production clusters. I concur that the concept of a "test cluster' is a little problematic as it necessarily divides that from "cluster". If the current cluster API controllers don't allow specific cluster configurations, why not create custom controllers on those objects to achieve what would be necessary for testing?

BenTheElder:
This came up today at sig-testing, I think I might understand the goals a bit better now... See https://github.com/kubernetes/test-infra/tree/master/kind for some early tooling that may relate.

mariantalla:
@BenTheElder Will take a look at `kind`, thanks. Is there anything that can be improved in the KEP phrasing to help with expressing the goals more clearly? I'll have a look at the notes from the last meeting too.

cc @hoegaarden

justaugustus:

justaugustus:

/close

PR #2181, opened 2018-05-23
Add SECURITY_CONTACTS for k/community

cblecker:


wking:
Looks good to me :).  Can this PR or follow up work adjust [our current docs][1] to link to the new local file?  When I asked about linking a local example [here][2], @jessfraz [expected this repo to not need its own `SECURITY_CONTACTS`][3].  With the local example to link to, I think we can drop [the current template reference][4], but I don't have a problem keeping it either if folks want to link to both the local example and the template.

[1]: https://github.com/kubernetes/community/blame/e1f3ea4543d40c93aaf6581a84a9ab83533b3ef3/committee-steering/governance/sig-governance-template-short.md#L72
[2]: https://github.com/kubernetes/community/pull/2062#discussion_r189463883
[3]: https://github.com/kubernetes/community/pull/2062#issuecomment-390525219
[4]: https://github.com/kubernetes/community/blame/e1f3ea4543d40c93aaf6581a84a9ab83533b3ef3/committee-steering/governance/sig-governance-template-short.md#L74

cblecker:
No, I want to continue referencing the template project from the governance template.

jessfraz:
Thanks for being awesome :)

wking:
> No, I want to continue referencing the template project from the governance template.

It doesn't have to be either/or, it can be both/and.  For example:

```markdown
  - Defined in [`SECURITY_CONTACTS`](/SECURITY_CONTACTS) files, this is only relevant to the root file in
    the repository, there is a template
    [here](https://github.com/kubernetes/kubernetes-template-project/blob/master/SECURITY_CONTACTS)
```

or:

```markdown
  - Defined in `SECURITY_CONTACTS` files (like [this one](/SECURITY_CONTACTS)), this is only relevant to the root file in
    the repository, there is a template
    [here](https://github.com/kubernetes/kubernetes-template-project/blob/master/SECURITY_CONTACTS)
```

cblecker:
There's no need to reference the SECURITY_CONTACTS file in this repo from the template description. The authoritative location if someone is referencing the template is the template project. This file is irrelevant to the template.

wking:
> There's no need to reference the SECURITY_CONTACTS file in this repo from the template description. The authoritative location if someone is referencing the template is the template project. This file is irrelevant to the template.

Ah, maybe I'm just getting confused by having project template docs in the sig-charter template.  It looks like [it's designed to cover `sigs.yaml`][1], but now some of the information is actually stored in external repositories (like `OWNERS`).  So now I agree that we don't need a local `OWNERS` link, and am 100% on board with this PR as it stands.  I'll file a separate PR to try and refocus the sig-charter template section on `sigs.yaml`.

[1]: https://github.com/kubernetes/community/blame/e1f3ea4543d40c93aaf6581a84a9ab83533b3ef3/committee-steering/governance/sig-governance-template-short.md#L20

PR #2184, opened 2018-05-24
Fix name of linux foundation bot: thelinuxfoundation

justinsb:
linuxfoundation appears to be the org, thelinuxfoundation is the bot.


idvoretskyi:
@justinsb thanks!

PR #2185, opened 2018-05-24
Better organize performance SLOs - initial cleanup

wojtek-t:
/assign @shyamjvs 
/assign @countspongebob


sig-scalability/slos/slos.md
```diff
@@ -0,0 +1,108 @@
...
+in these areas.
+
+## What do we require from SLIs/SLOs?
+
+We are going to define more SLIs and SLOs based on the most important indicators
+in the system.
+
+Our SLOs need to have the following properties:
+- <b> They need to be testable </b> <br/>
+  That means that iwe need to have a benchmark to measure it it's met.
```
shyamjvs: iwe -> we
it -> if

sig-scalability/slos/slos.md
```diff
@@ -0,0 +1,108 @@
...
+
+As a consequence we introduce the two types of SLOs.
+
+### Steady state SLOs
+
+With steady state SLOs, we provide guarantees about system's behavior during
+normal operations. We are able to provide much more guarantees in that situation.
+
+```Definition
+We define system to be in steady state when the cluster churn per second is <= 20, where
```
shyamjvs: So we want to use the same number (20) for all sizes of clusters? As large clusters IIUC would expectedly have more churn, no?

wojtek-t: Longer term - maybe. For now, we don't really support bigger churn for larger clusters.

sig-scalability/slos/system_throughput.md
```diff
@@ -0,0 +1,26 @@
+### User stories
+- As a user, I want a guarantee that my workload of X pods can be started
+  within a given time
+- As a user, I want to understand how quickly I can react to a dramatic
+  change in workload profile when my workload exhibits very bursty behavior
+  (e.g. shop during Back Friday Sale)
+- As a user, I want a guarantee how auickly I can recreate the whole setup
```
shyamjvs: quickly

sig-scalability/slos/system_throughput.md
```diff
@@ -0,0 +1,26 @@
...
+- Single namespace should run 5000 `Pods` in the following configuration:
+  - one big `Deployment` running ~1/3 of all `Pods` from this `namespace`
+  - medium `Deployments`, each with 120 `Pods`, in total running ~1/3 of all
+    `Pods` from this `namespace`
+  - small `Deployment`, each with 10 `Pods`, in total running ~1/3 of all `Pods`
+    from this `Namespace`
+- Each `Deployment` should be covered by a single `Service`.
+- Each `Pod` in any `Deployment` contains two pause containers, one `Secret`
+  other than default `ServiceAccount` and one `ConfigMap`. Additionally it has
+  resource requests set and doesn't use any advanced scheduling features or
```
shyamjvs: Any reason why you left out mentioning the e.g of antiaffinity?

sig-scalability/slos/slos.md
wojtek-t: done

sig-scalability/slos/system_throughput.md
wojtek-t: done

sig-scalability/slos/system_throughput.md
wojtek-t: If we mention some features, people will think about it as an extensive list. And that probably wasn't an extensive list.

If you feel very strongly about it though, I can revert it.

PR #2188, opened 2018-05-24
KEP: Dynamic Audit Configuration

pbarker:
Proposal to allow for the dynamic configuration of advanced auditing features.


hh:
After sending out https://lists.cncf.io/g/cncf-k8s-conformance/message/417
It was nice to wake up this morning to your KEP.
Thank you!

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+- Provide a generic interface to configure all kube-apiserver flags
+
+## Proposal
+
+### Dynamic Configuration of Policy
+Dynamic configuration of the existing policy object will be provided through use of a shared informer.
+```yaml
+apiVersion: audit.k8s.io/v1beta1
+kind: Policy
+omitStages:
```
loburm: I suppose omitStages should be under the rules: https://github.com/kubernetes/apiserver/blob/master/pkg/apis/audit/types.go#L228

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+```
+In kube-apiserver today this is only accepted as a file, this KEP would allow it to either be configured as a file on the master node or applied to the `kube-system` namespace as an object and be dynamically configured. This will require the audit API to be registered with the apiserver. If both file and object exist the dynamic object will take precedence.
+
+### Dynamic Configuration of Flags
+There are a number of configuration options that are currently set through runtime flags. This KEP would provide a new API for the dynamic configuration of the webhook parameters.
+```yaml
+apiVersion: audit.k8s.io/v1alpha1
+kind: AuditConfiguration
+metadata:
+  name: <name>
```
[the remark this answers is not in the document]
hh: Use Cases for multiple webhooks / policies:

* There is an ongoing permanent audit policy, but a temporary audit with a different policy (and likely webhook endpoint) needs to be added. ie. Enabling APISnoop for a specific application
* [Joint auditing](https://en.wikipedia.org/wiki/Joint_audit) with same policy but dual webhooks to independent auditors.

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+```
+The webhook would share the model of validating and mutating admission control plugins, leveraging a shared informer to provide online configuration. If runtime flags are also provided, the dynamic configuration will override any parameters provided there.
+
+### User Stories
+
+#### Story 1
+As a cluster admin I will easily be able to enable the interal auditing features of an existing cluster, and tweak the configurations as necessary.
+
+#### Story 2
+As a Kubernetes extension developer, I will be able to provide drop in extensions that utilize audit data.
```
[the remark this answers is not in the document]
hh: A possible a third story to cherry pick, which will help with [APISnoop](https://github.com/cncf/apisnoop/) style audits: https://github.com/kubernetes/community/commit/5f46645799627040ae95d513f8602643c689858

#### Story 3
As a cluster admin, I will be able configure multiple audit-policies and webhook endpoints to provide independent auditing facilities.

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
pbarker: I was shooting for a single webhook initially just to match parity with the current system, but if we think there are sufficient use cases for multiples, I'm happy to change direction.

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Implementation Details/Notes/Constraints
+
+A misunderstanding may occur of both flags are defined and an object is provided. This should be explicitly called out in the 
+documentation and monitored within the community. If this becomes a point of contention we will need to reevaluate the implementation.
+
+### Risks and Mitigations
+
+This does open up the attack surface of the audit mechanisms. Having them strictly configured through the api server has 
+the advantage of limiting the access of those configurations to those that have access to the master node. However, Dynamic Admission has shown that this can be done in a safe manner.
```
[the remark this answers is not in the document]
liggitt: > As a potential mitigation strategy, what if we said that the flag-defined configuration always takes precedence, but can be combined with the dynamic config? This would let the cluster operator define a policy subset that must be abided by (e.g. ALWAYS log secrets access, at metadata level; NEVER log events, ALWAYS log dynamic audit log changes). This could easily be done by just appending the dynamic config to the static config.

that's tricky... I can see valid use cases both ways:

* flag-defined config that limits what dynamic configs can do
* flag-defined config for operating in normal mode, dynamic config that allows temporary verbose logging to debug a particular issue

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+  initialBackoff: <10s>
+  batch:
+    bufferSize: <400>
+    maxSize: <10000>
+    maxWait: <30s>
+    throttleBurst: <15>
+    throttleEnabled: <true>
+    throttleQPS: <10>
+```
+The webhook would share the model of validating and mutating admission control plugins, leveraging a shared informer to provide online configuration. If runtime flags are also provided, the dynamic configuration will override any parameters provided there.
```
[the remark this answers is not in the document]
pbarker: If we go that route I agree, we could add a `policy` field to the `AuditConfiguration` object, that would take the name of the policy object

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+Sucess will be determined by stability of the provided mechanisms and ease of understanding for the end user.
+
+* alpha: Policy and api server flags can be dynamically configured, known issues are tested and resolved.
+* beta: Mechanisms have been hardened against any known bugs and the process is validated by the community
+
+## Implementation History
+
+- 05/18/2018: initial design
+
+## Alternatives
```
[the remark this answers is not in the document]
pbarker: I like this idea, I think inherently you'll want stricter policy on certain namespaces while others cause bloat. We'll track this as a later feature

keps/0007-dynamic-audit-control.md
pbarker: `omitStages` can also be global https://github.com/kubernetes/apiserver/blob/master/pkg/apis/audit/types.go#L168

keps/NEXT_KEP_NUMBER
```diff
@@ -1 +1 @@
-7
+8
```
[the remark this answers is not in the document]
pbarker: ah ok, misunderstood that part, updating

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
pbarker: good idea 👍 everyone wins with this, updated

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+kind: Policy
+omitStages:
+  - "RequestReceived"
+rules:
+  - level: <request level>
+    resources:
+    - group: <group>
+      resources: <resources>
+```
+In kube-apiserver today this is only accepted as a file, this KEP would allow it to either be configured as a file on the master node or applied to the `kube-system` namespace as an object and be dynamically configured. This will require the audit API to be registered with the apiserver. If both file and object exist the dynamic object will take precedence.
```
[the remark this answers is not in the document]
pbarker: agreed, updating

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
pbarker: Temporary verbose logging is an interesting use case. Could we settle on assuming that if you want that feature you would need to leave your static config open to it, and configure the rest dynamically?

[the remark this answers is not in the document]
ericavonb: Having dynamic audit config overriding flag-defined config as defined is very problematic, IMHO. Dynamic policies, e.g. defined as an API object, would be letting the API monitoring itself. It would remove the ability to provide any guarantee on the integrity of an audit.

The only tenable way to allow some dynamism to audit policy would be to have flag-defined config be the ultimate authority and let it specify the bounds and policy for applying any dynamic configuration.

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Dynamic Configuration of Flags
+There are a number of configuration options that are currently set through runtime flags. This KEP would provide a new API for the dynamic configuration of the webhook parameters.
+```yaml
```
CaoShuFeng: +1 to combine configuration flags into an object.
We have 30 audit related flast in kube-apiserver now.
```
$ kube-apiserver --help |grep audit | wc -l
30
```

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
tallclair: https://github.com/kubernetes/kubernetes/issues/53455 is for a pull-model (at least for setting up the initial connection), where's multiple webhooks would be for a push model. Definitely overlap, but I think there's a use case for both. One thing to note is that if there is a policy-per-webhook, that gets confusing with https://github.com/kubernetes/kubernetes/issues/53455, which policy does it use?

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
CaoShuFeng: > As an extension, do you think it would ever make sense to allow per-namespace policies

This is what my boss want. I'am happy to implement it.
I think a quota for each namespace's webhooks is necessary.

CaoShuFeng:
I can't wait to see this happens.
@pbarker What's the developing plan for this proposal?

keps/0007-dynamic-audit-control.md
CaoShuFeng: >   applied to the `kube-system` namespace as an object and be dynamically configured

Why policy is not a cluster-scoped object?

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
CaoShuFeng: > As a potential mitigation strategy, what if we said that the flag-defined configuration always takes precedence, but can be combined with the dynamic config? This would let the cluster operator define a policy subset that must be abided by (e.g. ALWAYS log secrets access, at metadata level; NEVER log events, ALWAYS log dynamic audit log changes). This could easily be done by just appending the dynamic config to the static config.

Agree to make the flag-defined configuration read only.
And the dynamic audit config will send audits to it's own backends, they should not effect each other.

keps/0007-dynamic-audit-control.md
pbarker: good catch, updating

pbarker:
@CaoShuFeng I plan on beginning the implementation this week, and will tweak things as feedback comes in

bgrant0607:
We need to figure out the API review process for these types of changes. I just put a hold on this proposal for now.

fabioy:
@bgrant0607 What does this mean? Is there a proposal for how these API review processes should proceed? What needs to be done to unblock this proposal?

fabioy:
Nvm, found thread on sig-arch. Will follow up there.

bgrant0607:
Please add API machinery as a participating SIG.  The webhooks should not be inconsistent without good reasons. 

There are also no approver, reviewer(s), or editor assigned yet. I think we need to assign those before merging.

jbeda:
Feel free to check this stuff in early to iterate.  Before this moves past the `provisional` stage we want to make sure that we get all the right parties involved and an API review.  We need to get the metadata fields nailed down.  

Brian suggested bringing in API machinery to ensure consistency. But I'm not sure that is necessary if we go through an API review process for consistency.

bgrant0607:
At least one API approver should be an approver on this.

@liggitt Are you planning/willing to do that?

bgrant0607:
@tallclair Does SIG Auth agree that this proposal should move forward?

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+### Non-Goals
+- Provide a generic interface to configure all kube-apiserver flags
+
+## Proposal
+
+### Dynamic Configuration of Policy
+Dynamic configuration of the existing policy object will be provided through use of a shared informer.
+```yaml
+apiVersion: audit.k8s.io/v1beta1
+kind: Policy
```
bgrant0607: Is this an API resource or a configuration specification?

bgrant0607:
@jbeda 

According to the workflow, the owning SIG needs to accept that this should move ahead:
https://github.com/kubernetes/community/blob/master/keps/0001-kubernetes-enhancement-proposal-process.md#kep-workflow

I haven't seen that stated in any of the comments.

tallclair:
SIG wise, I think audit mostly falls under the umbrella of SIG-auth, with SIG-apimachinery providing substantial input on the mechanisms. I agree with the conceptual direction of this proposal, and we should move forward with discussions. I'm not ready to sign off on the proposed mechanisms though.

tallclair:
I was reflecting more on the dynamic policy configuration, and the suggestions to have per-backend policies or per namespace policies, and I'm wondering if we should actually move in the opposite direction - away from client-side policy.

Client-side in this case is the k8s apiserver, which pushes the policy to the webhook server. Moving the policy decisions server-side means the apiserver basically pushes audit events for everything, and it's up to the server to ignore the events they don't care about. The client side policy should then be more of a description _how_ to log things (e.g. which stages to include, don't include GET responses, drop sensitive secrets request bodies, etc.), as opposed to _what_ to log. This shrinks the scope of the audit policy, and makes it much easier to define a general policy for the whole cluster. The one piece that I think should remain per-backend (per-webhook) is the ability to only register a subset of namespaces, to support multitenant setups.

The tradeoff is that this means the server is pushing more data, which consumes memory & CPU to serialize the audit events on the apiserver & receiving webhook.

keps/0007-dynamic-audit-control.md
pbarker: We're considering having it as both, but the details are being hashed out on that. We want a static configuration file on the server to define a root policy and dynamic api resources to further refine and direct the events.

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
bgrant0607: Please read the API conventions. In particular:
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#naming-conventions

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
pbarker: This would require write access to the master node which is highly privileged. The main goal of the proposal is to provide drop in components that can utilize the audit data, which in turn enhances the security and conformance of the cluster. 

Based on the conversations were looking to provide a root policy that cannot be tampered with but enable the ability to dynamically configure secondary policies and webhooks.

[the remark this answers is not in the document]
CaoShuFeng: > This would require write access to the master node which is highly privileged. 

Agree. This is a reason.

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+    resources:
+    - group: <group>
+      resources: <resources>
+```
+In kube-apiserver today this is only accepted as a file, this KEP would allow it to either be configured as a file on the master node or applied to the `kube-system` namespace as an object and be dynamically configured. This will require the audit API to be registered with the apiserver. If both file and object exist the dynamic object will take precedence.
+
+### Dynamic Configuration of Flags
+There are a number of configuration options that are currently set through runtime flags. This KEP would provide a new API for the dynamic configuration of the webhook parameters.
+```yaml
+apiVersion: audit.k8s.io/v1alpha1
```
sttts: How does this play with aggregated apiservers? All of those get the same AuditConfig?

pbarker: yeah I believe the idea with aggregated apiservers is that all of their traffic still passes through the main apiserver so it will be audited there.

sttts: No, it won't. It is passed through, but only proxied. We do not look into the connection. The aggregated apiservers will have to watch these objects and pick the config and rules.

pbarker: I see, is the best path forward to simply document what's needed? It would seem we would also need to enforce that the aggregated apiserver is implementing this properly or it opens a potential security gap. Perhaps a conformance test could be created to provide a guarantee.

sttts: We will implement this in k8s.io/apiserver. So at least every newer apiserver will pick it up automatically. There might be old binaries around though which miss this support.

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+  - "RequestReceived"
+rules:
+  - level: <request level>
+    resources:
+    - group: <group>
+      resources: <resources>
+```
+In kube-apiserver today this is only accepted as a file, this KEP would allow it to either be configured as a file on the master node or applied to the `kube-system` namespace as an object and be dynamically configured. This will require the audit API to be registered with the apiserver. If both file and object exist the dynamic object will take precedence.
+
+### Dynamic Configuration of Flags
```
yliaog: https://goo.gl/opSc2o laid out the general recommendation for kubernetes component configuration
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/dynamic-kubelet-configuration.md is proposal for kubelet

I think kube api server would move towards this in the future. I'd like the change proposed here to be consistent with this direction.

keps/0007-dynamic-audit-control.md
sttts: The aggregated API servers will read the configs, not write or serve them. Similary to admission webhooks today. Every aggregated API server will pick them up.

keps/0007-dynamic-audit-control.md
sttts: +1

I can easily see the policies to be API objects as these are files in the traditional flag based configuration. But the audit plugin flags ought to be in-line with the componentconfig direction.

pbarker: @yliaog thank you for sharing the doc. The main takeaway I'm seeing that applies here is the movement away from static disk configuration. The only motivation for keeping the root policy as an optionally static configuration was to scope down access to admins with rights to the master node. I strongly agree with the sentiments and direction of the doc, but would want the sign off from sig-auth before removing static disk configuration entirely.

Please inform if there are other pieces of this proposal that aren't in line with the doc.

pbarker: along these lines, I'm curious if there is interest in providing a common dynamic webhook interface. Dynamic admission currently has https://github.com/kubernetes/kubernetes/blob/v1.10.0-beta.1/staging/src/k8s.io/api/admissionregistration/v1beta1/types.go#L218 but that lacks the robust auth options that a kubeconfig file offers. Would it make sense for there to be a common interface for outgoing webhooks that can load secrets for its auth infromation?

liggitt: > Would it make sense for there to be a common interface for outgoing webhooks that can load secrets for its auth infromation?

Auth was intentionally omitted from that structure, since it is useful and sometimes necessary to be able to distinguish the apiserver calling the webhook. For example, I would not expect all extension servers to have the ability to identify themselves as the main kube-apiserver when calling webhooks.

pbarker: That makes sense, I wonder if theres still value in providing a common mechanism for how outgoing dynamic webhooks are handled or whether we should duplicate that pattern within this implementation for the time being

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+  mode: <batch or blocking>
+  initialBackoff: <10s>
+  batch:
+    bufferSize: <400>
+    maxSize: <10000>
+    maxWait: <30s>
+    throttleBurst: <15>
+    throttleEnabled: <true>
+    throttleQPS: <10>
+```
```
yliaog: is the above the complete list of params you're going to specify dynamically, or is it just a sample list?

also how is it going to work with HA apiserver setup? Is there a need for an API server to have a different param?

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+### Goals
+- Provide an api and set of objects to configure the advanced auditing kube-apiserver webhook flags dynamically
+- Allow the current audit policy object to be configured dynamically
+
+### Non-Goals
+- Provide a generic interface to configure all kube-apiserver flags
+
+## Proposal
+
+### Dynamic Configuration of Policy
```
caesarxuchao: It's useful to log the changes to the `Policy` in the audit log, then it's easy to tell if changes in the audit log is a result of policy changes. To enforce this, we can hard code the rule for `create/update/delete/patch` of `audit.k8s.io/v1beta1/policy` to always be `level=RequestResponse`, and is not subject to the global `omitStages`.

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,126 @@
...
+In kube-apiserver today this is only accepted as a file, this KEP would allow it to either be configured as a file on the master node or applied to the `kube-system` namespace as an object and be dynamically configured. This will require the audit API to be registered with the apiserver. If both file and object exist the dynamic object will take precedence.
+
+### Dynamic Configuration of Flags
+There are a number of configuration options that are currently set through runtime flags. This KEP would provide a new API for the dynamic configuration of the webhook parameters.
+```yaml
+apiVersion: audit.k8s.io/v1alpha1
+kind: AuditConfiguration
+metadata:
+  name: <name>
+webhook:
```
caesarxuchao: Do we need to dynamically config the url of the webhook? Like https://github.com/kubernetes/kubernetes/blob/6decdaa775f5feb480d06e05c049fa564187f30e/staging/src/k8s.io/api/admissionregistration/v1beta1/types.go#L222

keps/0007-dynamic-audit-control.md
caesarxuchao: The proposal needs to describe how an aggregated API server knows an `AuditConfiguration` is for it.

keps/0007-dynamic-audit-control.md
pbarker: I'm also seeing the lack of auth here problematic. With the dynamic admission control its not really needed as your simply checking a token, with audit this opens an attack vector of an on cluster agent pushing phony events to the backend.

pbarker:
Based on the feedback I'm considering providing 3 levels of configuration.
`Root` - A singleton policy and configuration object, all other objects inherit its restrictions. This enables a cluster admin to enforce certain logging behavior. I understand there is no precedence for such an object, but it appears to make sense here.   
`Cluster` - Cluster scoped policy and configuration objects. These enable policies to be tied to a number of backends which apply to all cluster actions. An auditing backend can be added for all events but restricted in a manner that is acceptable. The policies will inherit the root policy's restrictions.   
`Namespace` - Namespace scoped policy and configuration objects. These enable multitenant scenarios where a tenant may want events sent to a specific backend. These policies will inherit the root policy's restrictions.   

This may appear fairly complex but I think it satiates the valid asks that have been presented. Let me know your thoughts

keps/0007-dynamic-audit-control.md
yliaog: A few followup questions to discuss:
1. what would be ideal config structure? I think this should be considered holistically for API server. we don't want to have a config struct for audit, another config struct for, say, storage, and yet another config for something else. Probably we want to have ApiServerConfiguration as top level struct, and AuditConfiguration as a sub struct
2. There is a difference between MutatingWebhookConfiguration and KubeletConfiguration in that one is a top level k8s API, the other is using ConfigMap. Hence it bears the question of whether AuditConfiguration should be a k8s API object. There is pros/cons of each, it is worthy of further discussion.
3. It seems that even with this dynamic config, there will still be audit related flags configured through command line (is there 30-ish of them?). would it make sense to group them together as a single config struct? and as a single disk file provided to API server.
4. Generally, what is the planned process to rollout config, rollback config? how to report / monitor the status of the rollout / rollback?

pbarker: Here's what I have so far:
1. I think a holistic config makes sense, but with audit attempting to be a dynamic configuration, I'm not sure how those two would play together. We could have a single `APIServer` configuration object, but I fear that may get bloated. We could potentially have a set of configuration objects that are namespaced in a certain way
2. A well defined API object seems to be more clear, what are the advantages of nesting it as a json blob in a configmap?
3. I would hope to get rid of the majority of the flags, or as you said roll them up into a single config object. Ideally that would be a singleton object that could be read from disk or provided via the api.
4. The current solution is to use a shared informer to listen for config changes for the dynamic config, and on rollout log the changes. If the config if invalid, that should be logged, and we could potentially provide a 'last known good config' to roll back to

keps/0007-dynamic-audit-control.md
pbarker: good idea 👍 will add

keps/0007-dynamic-audit-control.md
pbarker: @caesarxuchao I would assume the aggregates would all adhere to the same dynamic config. If a specific config is wanted for that aggregate it could be configured as a static file on the node. Is there a reason an aggregate would need a separate dynamic object?

keps/0007-dynamic-audit-control.md
pbarker: Adding the full golang definitions to make that more clear. I don't think API servers would need separate params, but if that is desired a static config could be placed on the master node for each server.

keps/0007-dynamic-audit-control.md
sttts: They could be based on a different k8s.io/apiserver version and hence have slightly different requirements, e.g. using another backend.

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+
+### Dynamic Configuration
+Dynamic configuration of the advanced auditing features will be provided through use of a shared informer. Two levels of configuration will be provided, a cluster scoped object and a namespace scoped object.
+
+#### Cluster Scoped Configuration
+A cluster scoped configuration object will be provided that applies to all events in the cluster.
+
+```golang
+// ClusterAuditConfiguration represents a cluster level audit configuration
+type ClusterAuditConfiguration struct {
```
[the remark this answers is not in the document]
pbarker: multiple, it's discussed below

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+    Policy *Policy `json:"policy"`
+
+    // Backends to send events
+    Backends *Backends `json:"backends"`
+}
+
+// Backends holds the configuration for the various event backends
+type Backends struct {
+    // Logs holds the file backends
+    Logs []*LogBackend  `json:"logs,omitempty"`
```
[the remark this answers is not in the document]
pbarker: I have separate external auditing facilities that can work off the same policy

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+    MaxWait  *int64  `json:"maxWait,omitempty"`
+
+    // ThrottleBurst is the maximum number of events sent at the same moment
+    ThrottleBurst *int `json:"throttleBurst,omitempty"`
+
+    // ThrottleEnabled determines whether throttling is enabled
+    ThrottleEnabled *bool `json:"throttleEnabled,omitempty"`
+
+    // ThrottleQPS maximum number of batches per second
+    ThrottleQPS *float32 `json:"thottleQPS,omitempty"`
```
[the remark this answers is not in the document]
pbarker: this is the batch webhook configuration, I believe this has parity with the flags, are you seeing something I'm missing?

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+
+```golang
+// AuditConfiguration represents a namespace level audit configuration
+type AuditConfiguration struct {
+    metav1.TypeMeta
+
+    v1.ObjectMeta
+
+    // Policy is the current audit v1beta1 Policy object
+    Policy Policy `json:"policy"`
```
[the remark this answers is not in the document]
pbarker: yeah we would have to decide whether we want a policy subset, or just ignore items that don't apply to the namespace. I'm leaning toward the latter

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+
+    // Backends to send events
+    Backends NamespacedBackends `json:"backends"`
+}
+
+// NamespacedBackends holds the configuration for the various namespaced scoped event backends
+type NamespacedBackends struct {
+    // Webhooks holds the webhook backends
+    Webhooks []WebhookBackend `json:"webhooks"`
```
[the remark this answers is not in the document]
pbarker: I have separate external auditing facilities that can work off the same policy

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+As a Kubernetes extension developer, I will be able to provide drop in extensions that utilize audit data.
+
+#### Story 3
+As a cluster admin, I will be able configure multiple audit-policies and webhook endpoints to provide independent auditing facilities.
+
+### Implementation Details/Notes/Constraints
+
+Configurations will be cached in memory, as a means of storing the last good configuration. If an update to an existing configuration is unsuccessfull, the system will roll back to the previous configuration and log that action. If no previous configuration exists the application of the config will fail, and log the result. Any actions to the audit configuration objects will be hard coded to log at the `level=RequestResponse` to the previous backend and the new backend.
+
+Inherently, apiserver aggregates and HA apiserver setups will work off the same dynamic configuration object. If separate objects are needed they should be configured as static objects on the node and set through the runtime flags. Aggregated servers will implement the same audit handling mechanisms, and a conformance test should be provided as assurance.
```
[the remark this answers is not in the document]
pbarker: thats a possibility we should track

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+policy:
+  omitStages:
+    - <stage>
+  rules:
+    - level: <request level>
+      resources:
+      - group: <group>
+        resources: <resources>
+backends:
+  webhooks:
```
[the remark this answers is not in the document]
pbarker: ResourceQuota? or are you referring to something else?

keps/0007-dynamic-audit-control.md
```diff
@@ -0,0 +1,334 @@
...
+
+#### Story 2
+As a Kubernetes extension developer, I will be able to provide drop in extensions that utilize audit data.
+
+#### Story 3
+As a cluster admin, I will be able configure multiple audit-policies and webhook endpoints to provide independent auditing facilities.
+
+### Implementation Details/Notes/Constraints
+
+Configurations will be cached in memory, as a means of storing the last good configuration. If an update to an existing configuration is unsuccessfull, the system will roll back to the previous configuration and log that action. If no previous configuration exists the application of the config will fail, and log the result. Any actions to the audit configuration objects will be hard coded to log at the `level=RequestResponse` to the previous backend and the new backend.
```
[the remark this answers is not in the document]
pbarker: in the event of a local file configuration there would be no status code, I'll add a piece on the api returning 400

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
CaoShuFeng: Yes. ResourceQuota.

If one tenant create too many webhook backends, it may require  too much network bandwidth.

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
CaoShuFeng: Sorry, the batch configuration is OK.

I mean this:
```
--audit-log-truncate-max-event-size int                   Maximum size of the audit event sent to the underlying backend. If the size of an event is greater than this number, first request and response are removed, andif this doesn't reduce the size enough, event is discarded. (default 102400)
```

keps/0007-dynamic-audit-control.md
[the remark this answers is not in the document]
pbarker: gotchya, I think that may be something to tackle in a future release, we will track

jbeda:
The idea behind the [KEP process](https://github.com/kubernetes/community/blob/master/keps/0000-kep-template.md) is to start with an agreement around scope and that the problem is worth solving and check in early.  Then we can expand and take issues one at a time so that we can keep discussions focused.  Nothing is official until the `status` field moves to "implementable".

In addition, we probably want the list of `approvers` to be smaller than the set of `reviewers`.  

Any objections to removing the stuff that is somewhat controversial from this PR (@pbarker will keep it for future PRs!) and getting a rough agreement on the fact that *something* here makes sense?

Also -- seems like either @liggitt or @tallclair would be appropriate people to be approvers.  Any objections there or do others feel super strongly?

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,342 @@
...
+
+#### Story 2
+As a Kubernetes extension developer, I will be able to provide drop in extensions that utilize audit data.
+
+#### Story 3
+As a cluster admin, I will be able configure multiple audit-policies and webhook endpoints to provide independent auditing facilities.
+
+### Implementation Details/Notes/Constraints
+
+Configurations will be cached in memory, as a means of storing the last good configuration. If an update to an existing configuration is unsuccessful, the system will roll back to the previous configuration and log that action, if configured through the api a 400 will be returned. If no previous configuration exists the application of the config will fail, and log the result. Any actions to the audit configuration objects will be hard coded to log at the `level=RequestResponse` to the previous backend and the new backend.
```
tallclair: > If an update to an existing configuration is unsuccessful

How is unsuccessful defined in this context?

tallclair: > Configurations will be cached in memory

This section has me worried. Is there any precedent for this type of change / rollback functionality anywhere else in the API machinery?

Perhaps a better approach is to do a phased-rollout approach (I.e. configure new backend, confirm it works, delete old backend).

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,342 @@
...
+
+Inherently, apiserver aggregates and HA apiserver setups will work off the same dynamic configuration object. If separate objects are needed they should be configured as static objects on the node and set through the runtime flags. Aggregated servers will implement the same audit handling mechanisms, and a conformance test should be provided as assurance.
+
```
tallclair: I'd like to see a bit more discussion of how this will be implemented, but I'd accept a TODO with a follow up PR. Specifically, the audit policy is currently checked in place in the audit pipeline, but the whole pipeline will need to be refactored to support multiple policies. Or we can just drop multi-policy from this proposal :wink:

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,342 @@
...
+    Policy *Policy `json:"policy"`
+
+    // Backends to send events
+    Backends *Backends `json:"backends"`
+}
+
+// Backends holds the configuration for the various event backends
+type Backends struct {
+    // Logs holds the file backends
+    Logs []*LogBackend  `json:"logs,omitempty"`
```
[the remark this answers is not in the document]
jbeda: SGTM -- I think that there was the idea here to rationalize this across the command line flags and the dynamic config.  But we can take the dynamic config first and leave the command line flags out of scope for now.

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,342 @@
...
+
+```golang
+// ClusterAuditConfiguration represents a cluster level audit configuration
+type ClusterAuditConfiguration struct {
+    metav1.TypeMeta
+
+    v1.ObjectMeta
+
+    // Policy is the current audit v1beta1 Policy object
+    Policy *Policy `json:"policy"`
```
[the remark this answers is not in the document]
jbeda: I think that this deserves some more discussion but we can elide it from the KEP to get it checked in and move that discussion to a different PR.  We'll have to close on this topic before this becomes `implementable` however.

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,342 @@
...
+    // TruncateOptions follows the truncate options 
+    // https://github.com/kubernetes/apiserver/blob/master/pkg/server/options/audit.go#L157
+    TruncateOptions *AuditTruncateOptions `json:truncateOptions,omitEmpty"`
+}
+
+type WebhookMode string
+
+const (
+  WebhookModeBatch WebhookMode = "batch"
+  WebhookModeBlocking WebhookMode = "blocking"
```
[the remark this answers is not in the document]
jbeda: Agree that blocking mode is dangerous.  How is this different from mutating admission controllers?  Is there a way to turn them off so that you can reconfigure them? 

Regardless, this can be removed from now and discussed in a different PR if we think we really need it.

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,342 @@
...
+
+#### Cluster Scoped Configuration
+A cluster scoped configuration object will be provided that applies to all events in the cluster.
+
+```golang
+// ClusterAuditConfiguration represents a cluster level audit configuration
+type ClusterAuditConfiguration struct {
+    metav1.TypeMeta
+
+    v1.ObjectMeta
```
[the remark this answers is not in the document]
pbarker: The current idea is that they work as independent solutions, where a policy could define what events you allow to an external auditing facility

keps/0007-dynamic-audit-configuration.md
pbarker: yeah this just solves the issue of 'I have multiple external audit facilities and want to scope what kind of data is allowed to each'. This is something we can move to a later PR

keps/0007-dynamic-audit-configuration.md
pbarker: in the situation of a single api server a rollout wouldn't be possible, which lead me to the cached configuration. I've updated to include a phased rollout where possible

pbarker:
Ok thank you for the feedback @jbeda and @tallclair I've updated to reflect the changes to scope. 
Removed Items:
* Dynamic policy
* Namespaced configuration
* Log backend
* Blocking webhook
* New apiserver flags

These will be discussed in later PRs as needed. The current scope simply takes the existing runtime flags that configure the webhook today and allows them to be configured by the API. Currently, multiple configurations will exist independent of one another. I understand that may cause changes to the existing pipeline, we can reduce this to a singleton if that is yet controversial.

keps/0007-dynamic-audit-configuration.md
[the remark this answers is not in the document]
tallclair: > We'll have to close on this topic before this becomes implementable however.

Do we? This seems like something that could be added later in a backwards-compatible way. I.e. audit configurations without a policy specification default to the static cluster policy.

keps/0007-dynamic-audit-configuration.md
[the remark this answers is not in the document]
tallclair: Dynamic admission controllers do have this danger, but there are a bunch of escape hatches (e.g. excluding the admission controllers namespace from the namespaces it admits). I suppose we could do something similar by excluding ClusterAuditConfiguration changes from the blocked requests, but that feels strange since we want EXTRA auditing guarantees around changes to ClusterAuditConfiguration

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+    // ClientConfig holds the connection parameters for the webhook
+    ClientConfig *ClientConfig `json:clientConfig"`
+
+    // TruncateOptions follows the truncate options 
+    // https://github.com/kubernetes/apiserver/blob/master/pkg/server/options/audit.go#L157
+    TruncateOptions *AuditTruncateOptions `json:truncateOptions,omitempty"`
+}
+
+// BatchConfig holds the configuration for batch webhooks
+type BatchConfig struct {
```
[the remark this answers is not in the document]
pbarker: I'm thinking we should aim to keep parity with the existing options if we choose to expose them, otherwise it seems like an odd mismatch between the flags and configuration object. Is there a reason why we want to limit them?

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+## Implementation History
+
+- 05/18/2018: initial design
+- 06/13/2018: updated design
+
+## Alternatives
+
+### Generalized Dynamic Configuration
+
+We could strive for all kube-apiserver flags to be able to be dynamically provisioned in a common way. This is likely a large task and out of the scope of the intentions of this feature.
```
deads2k: Audit seems more like admission where you expect uniformity across a cluster.  Most of the kube-apiserver config would not be uniform in the cluster.

keps/0007-dynamic-audit-configuration.md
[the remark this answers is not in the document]
tallclair: Because a lot of those have to do with performance tuning of the apiserver process, which isn't something that users setting up external webhooks should be concerned with, IMO. A few of these do affect the receiving server though, so I could see keeping those (ThrottleBurst, ThrottleQPS, TruncateOptions)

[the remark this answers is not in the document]
pbarker: ok I could see `BufferSize` and `MaxWait` as only api-server specific. `MaxSize` could potentially be backend specific, unless we find that to be too dangerous. The difficulty with this is that now those flags need to be denoted as global options, are we comfortable changing that here?

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+type Backends struct {
+    // Webhooks holds the webhook backends
+    Webhooks []*WebhookBackend `json:"webhooks,omitempty"`
+}
+
+type WebhookMode string
+
+const (
+    // WebhookModeBatch sends requests in batches
+    // bloking mode is currently disable but will be discussed at a later date
```
x13n: What does it mean? When dynamic config is enabled in the cluster, will it disable the ability to use blocking mode?

pbarker: There was a concern raised about dynamically configuring a webhook to be blocking with a bad configuration which kills the apiserver. You could still configure a blocking webhook through flags.

keps/0007-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+
+#### Story 2
+As a Kubernetes extension developer, I will be able to provide drop in extensions that utilize audit data.
+
+#### Story 3
+As a cluster admin, I will be able configure multiple audit-policies and webhook endpoints to provide independent auditing facilities.
+
+### Implementation Details/Notes/Constraints
+
+When an update to a configuration is issued, the server will cache the current working configuration. If the update fails to apply, the system will roll back to the previous configuration and log that action, if configured through the api a 400 will be returned. If no previous configuration exists the application of the config will fail, and log the result. Any actions to the audit configuration objects will be hard coded to log at the `level=RequestResponse` to the previous backend and the new backend. If multiple API servers exist, the configuration will be rolled out in increments.
```
tallclair: Sorry, I still don't understand this. What specifically does this mean?

> If the update fails to apply, ...

Maybe it would help if you describe the implementation of that condition a bit more?

keps/0007-dynamic-audit-configuration.md
x13n: Ack. So, is the idea for now is to keep one "static" configuration via flags, as it exists today, and in addition to that, provide any number of additional dynamic configurations that effectively enable multiple backends? If yes, can you state that explicitly in the proposal?

keps/0007-dynamic-audit-configuration.md
pbarker: I think this only applies to policy, since were not handling that here I'll remove

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+
+#### Story 2
+As a Kubernetes extension developer, I will be able to provide drop in extensions that utilize audit data.
+
+#### Story 3
+As a cluster admin, I will be able configure multiple audit-policies and webhook endpoints to provide independent auditing facilities.
+
+### Implementation Details/Notes/Constraints
+
+Any actions to the audit configuration objects will be hard coded to log at the `level=RequestResponse` to the previous backend and the new backend. If multiple API servers exist, the configuration will be rolled out in increments.
```
tallclair: > If multiple API servers exist, the configuration will be rolled out in increments.

What does this mean? Is this talking about HA masters, or aggregated api servers?

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+
+It may also be reasonable to provide a dynamic auth configuration from secrets, with the understanding that it is shared by the api servers.
+
+This needs further discussion.
+
+## Graduation Criteria
+
+Sucess will be determined by stability of the provided mechanisms and ease of understanding for the end user.
+
+* alpha: Policy and api server flags can be dynamically configured, known issues are tested and resolved.
```
tallclair: drop policy

pbarker:
@tallclair @liggitt @yliaog I think we are in a good place with this here, is there anything else you would like to see? or can this move to next steps? Thanks

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+- Provide a generic interface to configure all kube-apiserver flags
+- composable audit policies per-endpoint
+- configuring non-webhook backends
+- configuring audit output (format or per-field filtering)
+- authorization of audit output
+
+## Proposal
+
+### Dynamic Configuration
+Dynamic configuration of the advanced auditing features will be provided through use of a shared informer.
```
tallclair: I'm not sure what this means... what do informers have to do with configuration? Do you mean the audit pipeline will get the configuration from an informer?

If so, how will it be used? I assume event handlers to add, remove & update backends to sync with the configuration?

**Suggestion:**
Introduce a new "dynamic" audit backend type. The dynamic audit backend looks like the [union](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/audit/union.go) backend, but keeps a map of config name -> BE, and keeps the list in sync using the informer event hooks.

The dynamic backend is enabled through a commandline flag. It is always configured with a wrapping truncate & batch backend. The implications of this are that options to the truncate & batch backends should not be included in the dynamic configuration (I think that's already the case). IMO the performance benefits of sharing those are worth it.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+
+```golang
+// ClusterAuditConfiguration represents a cluster level audit configuration
+type ClusterAuditConfiguration struct {
+    metav1.TypeMeta
+
+    v1.ObjectMeta
+
+    // Backends to send events
+    Backends *Backends `json:"backends"`
```
tallclair: nit: I think this should be singular. Update the comments too.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// Backends holds the configuration for the various event backends
+type Backends struct {
+    // Webhook holds the webhook backends
+    Webhook *WebhookBackend `json:"webhooks,omitempty"`
```
tallclair: nit: json doesn't match. I suggest just removing the tags from the KEP - they can be dealt with in the API PR review.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+    ThrottleBurst *int `json:"throttleBurst,omitempty"`
+
+    // ThrottleEnabled determines whether throttling is enabled
+    ThrottleEnabled *bool `json:"throttleEnabled,omitempty"`
+
+    // ThrottleQPS maximum number of batches per second
+    ThrottleQPS *float32 `json:"thottleQPS,omitempty"`
+
+    // ClientConfig holds the connection parameters for the webhook
+    ClientConfig *ClientConfig `json:clientConfig"`
```
tallclair: nit: `ClientConfig *WebhookClientConfig`

tallclair: If this is required it shouldn't be a pointer (I think)

keps/sig-auth/0014-dynamic-audit-configuration.md
pbarker: 👍 I think were on the same page here, I updated to make it more clear. Let me know if there's anything else we're missing

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -114,27 +114,27 @@ type WebhookBackend struct {
...
 
     // `caBundle` is a PEM encoded CA bundle which will be used to validate
     // the webhook's server certificate.
-    CABundle []byte `json:"caBundle"`
+    CABundle []byte
 }
 ```
 
-This configuration object will be accepted both as the existing runtime flags, and a dynamic api object. Multiple definitions can exist as independent solutions. These updates will require the audit API to be registered with the apiserver. Dynamic configuration will be enabled by a feature gate. If existing flags are provided to configure the audit backend they will be taken as a separate configuration object.
+Multiple definitions can exist as independent solutions. These updates will require the audit API to be registered with the apiserver. The dynamic configurations will be wrapped by truncate and batch options, which are set statically through existing flags. Dynamic configuration will be enabled by a feature gate for pre-stable releases, as well as a runtime flag `--audit-dynamic-enabled`. If existing flags are provided to configure the audit backend they will be taken as a separate backend configuration.
```
tallclair: I think the feature gate is sufficient... I don't think we also need the `--audit-dynamic-enabled` flag.

tallclair:
LGTM. Please squash your commits.

Any last thoughts from other reviewers? @liggitt @caesarxuchao @yliaog

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+
+The reason for static configuration today is that a single configured credential would have no way of differentiating apiserver replicas or their aggregates. There is a possible mitigation by providing a bound service account token and using the calling server's dns name as the audience.
+
+It may also be reasonable to provide a dynamic auth configuration from secrets, with the understanding that it is shared by the api servers.
+
+This needs further discussion.
+
+## Graduation Criteria
+
+Sucess will be determined by stability of the provided mechanisms and ease of understanding for the end user.
```
tallclair: s/sucess/success/

pbarker:
ok @tallclair I fixed up the spelling error and updated the next kep number due to an issue with https://github.com/kubernetes/community/pull/2331, should be cleared up 👍

PR #2198, opened 2018-05-28
Better organize performance SLIs/SLOs - part 2

wojtek-t:
/assign @shyamjvs @countspongebob


sig-scalability/slos/api_call_latency.md
```diff
@@ -0,0 +1,42 @@
...
+impact API call latencies.
+- As a result, we define the SLIs to be very generic (no matter how your
+cluster is set up), but we provide SLO only for default installations (where we
+have control over what apiserver is doing). This doesn’t provide a false
+impression, that we provide guarantee no matter how the cluster is setup and
+what is installed on top of it.
+- At the same time, API calls are part of pretty much every non-trivial workflow
+in Kubernetes, so this metric is a building block for less trivial SLIs and
+SLOs.
+- The SLO for latency for read-only API calls may have significant buffer in
```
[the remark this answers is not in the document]
wojtek-t: done

sig-scalability/slos/api_call_latency.md
```diff
@@ -0,0 +1,42 @@
...
+cluster is set up), but we provide SLO only for default installations (where we
+have control over what apiserver is doing). This doesn’t provide a false
+impression, that we provide guarantee no matter how the cluster is setup and
+what is installed on top of it.
+- At the same time, API calls are part of pretty much every non-trivial workflow
+in Kubernetes, so this metric is a building block for less trivial SLIs and
+SLOs.
+- The SLO for latency for read-only API calls of a given type may have significant
+buffer in threshold. In fact, the latency of the request should be proportional to
+the amount of work to do (which is number of objects og a given type in a given
```
shyamjvs: s/og/of

sig-scalability/slos/slos.md
```diff
@@ -92,17 +92,53 @@ of the cases.
...
+sliding window. However, for the purpose of reporting the SLO, it means one
+point per day (whether SLO was satisfied on a given day or not).
+
+<a name="footnote4">\[4\]</a>By non-streaming read-only API calls we mean GET
+requests without `watch=true` option set. (Note that in Kubernetes internally
+it translates to both GET and LIST calls).
+
+<a name="footnote5">\[5\]</a>A scope of a request can be either (a) `resource`
+if the request is about a single object, (b) `namespace` if it is about objects
+from a single namespace or (c) `cluster` if it spawns objects from multiple
```
shyamjvs: This might have already been discussed earlier - but it seems like non-namespaced resources (like nodes) will be put under 'cluster' category? If so, it means we give as much as 30s for 'LIST nodes' calls for e.g. Are we ok with that?

sig-scalability/slos/watch_latency.md
```diff
@@ -0,0 +1,17 @@
+## Watch latency SLI details
+
+### User stories
+- As an aministrator, if Kubernetes is slow, I would like to know if the root
```
shyamjvs: administrator

sig-scalability/slos/watch_latency.md
```diff
@@ -0,0 +1,17 @@
...
+cause of it is slow api-machinery (slow watch) or something farther the path
+(lack of network bandwidth, slow or cpu-starved controllers, ...)
+
+### Other notes
+- Pretty much all control loops in Kubernetes are watch-based. As a result
+slow watch means slow system in general.
+- Note that how we measure it silently assumes no clock-skew in case of
+cluster with multiple masters.
+
+### TODOs
```
shyamjvs: Btw - are we not having a test scenario here (like for other ones)?

sig-scalability/slos/api_extensions_latency.md
```diff
@@ -0,0 +1,6 @@
+## API call extension points latency SLIs details
+
+### User stories
+- As an administrator, if API calls are slow, I would like to know if this is
+because slow extension points (admission plugins, webhooks, initializers) and
+if so which ones are responsible for it.
```
shyamjvs: Are we planning to define some SLIs here?

shyamjvs:
I've left few comments. Lgtm besides those.

sig-scalability/slos/api_call_latency.md
wojtek-t: done

sig-scalability/slos/api_extensions_latency.md
wojtek-t: There are already SLIs - but definition of SLIs/SLOs are in the slos.md file.

sig-scalability/slos/slos.md
wojtek-t: I think it's what we should do. For things like PVs it will be possible to have a looot of them.

sig-scalability/slos/watch_latency.md
wojtek-t: done

sig-scalability/slos/watch_latency.md
wojtek-t: We don't have test cases for SLIs - test cases make sense only for things for which we have SLO.

If we will have an SLO for that in the future, there should be a test case.

sig-scalability/slos/slos.md
shyamjvs: I partly agree with you - but I don't have any data to prove how they compare with the order of namespaced resources :)

One idea is to have a different bound for non-namespaced category (10s) which will for e.g align well with our current latency for "list nodes" that's ~5s on 5k. It'll also help us to not link this category with other existing ones. But I'll leave the decision here to you.

sig-scalability/slos/watch_latency.md
shyamjvs: Yeah, that makes sense. Thanks.

sig-scalability/slos/slos.md
shyamjvs: Ok, actually that value seems to be <2s without those slow calls coming from cpu-starved heapster nanny.

sig-scalability/slos/api_extensions_latency.md
wojtek-t: I'm already having more the SLI per file, e.g.:
1. one file for both read-only and muting API calls
2. one file for all intiailized, admission plugins and webhooks

I think merging it further is a bad idea, because they pretty much don't share anything.

sig-scalability/slos/slos.md
wojtek-t: I would say that we may consider tuning this in the past, but as long as we don't have cluster loader and won't measure exactly what is here (5-min window, etc.), let's not tune it before then.

sig-scalability/slos/api_extensions_latency.md
shyamjvs: Sorry, I think I'm misunderstanding. I was suggesting above that we move the SLI definitions into the relevant files instead of putting them in the root file - there's no merging involved there.

sig-scalability/slos/slos.md
shyamjvs: Assuming you meant 'future' by 'past', that sounds reasonable to me :)
Maybe add a TODO for revisiting the SLO for non-namespaced resources?

wojtek-t: Yes - sorry, I meant future :)

Sure - adding TODO sounds good to me - adding now.

wojtek-t: Added a TODO.

shyamjvs: Lgtm, thx.

sig-scalability/slos/api_extensions_latency.md
shyamjvs: Ping.

wojtek-t: Sorry - apparently I forgot to send this comment.

I definitely want to have a single table with just all SLIs/SLOs definition (wihtout any comments) - that is what 90% of people would be interested in (and not interested in anything else).

We may consider duplicating the SLI/SLO definition into those files, but personally, I don't think that's a good idea, because that would make people read those dosc instead of the main one.

People should just read the "slos.md", and only if they are intersted in more details, go into those smaller files.

shyamjvs: > 90% of people would be interested in

Seems like you also want some SLO for our readers :)

And yeah, that sgtm.

shyamjvs:
/lgtm

Thanks!

shyamjvs:
Also @wojtek-t, why do PRs by you not have approve label?

wojtek-t:
> Also @wojtek-t, why do PRs by you not have approve label?

I think it's different in this repo when approve need to be explicit.

/approve

shyamjvs:
I see.. But I still don't understand why it can't be done automatically when the author can do it manually (which most likely he/she of course would)?
And if you need to be blocked on sth, I'd expect that sth shouldn't be you :)

cc @cjwagner

cjwagner:
@shyamjvs This is the behavior @cblecker requested when we enabled the `approve` plugin for this repo: https://github.com/kubernetes/test-infra/pull/5742#pullrequestreview-80336018

shyamjvs:
Thanks for the lead Cole. However wrt:

> no implicit self-approval from PR authors

the question I had above (in rephrased words) is - what is the usecase of disabling "implicit" self-approvals but allowing "explicit" ones?

cjwagner:
If a repo is *only* using the `approve` plugin process to gate automated merges, implicit self approval should probably be disabled so that a PR is not immediately merged if the author is an approver. In this case the author could approve their own PR, but they may prefer to have someone else approve it.

IMO that is the only use case where it really makes sense to disable implicit self approval. I guess some people prefer explicit self approval because it is especially clear and requires the PR author to acknowledge that they are an approver for the PR?

shyamjvs:
Won't argue on that :)
But thanks for the lead Cole.

PR #2203, opened 2018-06-01
Proposal for indexing labels to support efficient queries by label selectors

ccding:
kubernetes/kubernetes#4817


smarterclayton:
@kubernetes/sig-api-machinery-api-reviews

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
...
+This proposal presents solutions to make the queries by label selectors in the
+list API efficient.
+
+## Motivation
+
+Currently, only namespace index is supported in the cache. Controllers such as
+ReplicaSet, when selecting pods, need to select all the pods by namespace and
+then select the ones wanted filtering by index. While running a large number
+of pods, we observed that listing all pods and then filtering takes several
+seconds. This proposal presents solutions to improve this.
```
smarterclayton: Can you describe the scenario you are hitting?  With the watch cache enabled I have never seen a filtered label query take longer than 1s even on large sets, so I’d want to understand the data you have and how it could be applied.

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
...
+creation-date: "2018-05-31"
+last-updated: "2018-05-31"
+status: provisional
+---
+
+## Summary
+
+This proposal presents solutions to make the queries by label selectors in the
+list API efficient.
+
```
adohe: To make it clear, this proposal focus on client cache or apiserver watch cache?

ccding: See Proposal section 1 and 4. It focuses on client cash, while apiserver watch cache is optional.

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
...
+Since objects are created/deleted dynamically, we also need to update the index
+efficiently.
+
+2. Calculates the count for each index and evaluate the most selective first.
+
+3. Provides the ability in API for users to mark which labels to index or build
+all the index for all labels as the first step.
+
+4. `Optional` Adds index in API-server cache to improve queries directly going
+to API-server.
```
adohe: actually there are two things: first we need to evaluate controller side cache label index, then we could go to API-server cache.

keps/0012-label-index.md
lavalamp: Eventually we will have to do this. However, I don't want to do it too early because it's a lot of complexity, and I'm not sure apiserver is the right place to do it. I think it makes more sense to have the storage layer do this. This requires either adding a layer between apiserver and etcd, or implementing a bunch of features in etcd. Neither is going to happen soon. In the meantime, I don't want to do this in apiserver unless it's absolutely necessary.

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
...
+seconds. This proposal presents solutions to improve this.
+
+
+### Goals
+
+Make the queries by label selectors in the list API efficient.
+
+### Proposal
+
+1. Adds label based index in controllers to facilitate lookup by label.
```
lavalamp: This is not really a design :)

How will this be accomplished? What components will change? How will you support all of our selector options, in particular IN?

(I'm not saying these questions are not answerable and we can't do it, I'm saying they are and this document should provide the answers)

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
...
+
+Since objects are created/deleted dynamically, we also need to update the index
+efficiently.
+
+2. Calculates the count for each index and evaluate the most selective first.
```
lavalamp: So, ah, I don't want to own a query planning engine without a lot of justification. We don't add value by implementing DB functions, at least not so far. Even though this sounds like a lot of fun to write.

keps/0012-label-index.md
lavalamp: You can already build a client-side index for isolated queries pretty easily.

keps/0012-label-index.md
lavalamp: Oh I missed the word "controllers", you can already add indexes. Which controllers do you want to change?

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
...
+
+3. Provides the ability in API for users to mark which labels to index or build
```
lavalamp: I'm against passing this complexity on to users. Or are you talking about an API for configuring controller manager (as a substitute for a flag)? Even there I'd expect this to be just on or off, not require an administrator to figure out what their expensive label selectors are.

keps/0012-label-index.md
lavalamp: These two things shouldn't be in the same KEP, they don't have much in common.

keps/0012-label-index.md
lavalamp: I am not sure if you need to have a super detailed design in the kep itself but it needs to be somewhere.

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
+---
+kep-number: 12
+title: Index labels to support efficient queries by label selectors
+authors:
+  - "@jian-he, @ccding"
+owning-sig: sig-api-machinery
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
lavalamp: are these people approving the KEP or approving the PRs that result? If the former, add me & deads2k.

keps/0012-label-index.md
```diff
@@ -0,0 +1,51 @@
...
+creation-date: "2018-05-31"
+last-updated: "2018-05-31"
+status: provisional
```
lavalamp: This should not be merged it its current state. It needs: 1) better justification 2) focus on apiserver or clients and 3) a more detailed mechanism design.

keps/0012-label-index.md
ccding: Added evaluation results. PTAL

keps/0012-label-index.md
ccding: Removed the apiserver side cache. We will do it in the client side first.

keps/0012-label-index.md
ccding: Removed the apiserver side cache. We will do it in the client side first.

keps/0012-label-index.md
```diff
@@ -0,0 +1,89 @@
...
+| 2000 |2.768|1.200 |
+| 5000 |7.760|2.863 |
+| 10000 |14.756| 5.590|
+| 15000 |20.727|8.328 |
+
+#### Evaluation conclusion
+
+* With increasing number of pods, the response time of the list operations
+  increases linearly.
+* With more than 10K pods, it takes 14 seconds to list all the pods, and 5
```
lavalamp: And with client side filtering, it will take 14 seconds in both cases? I don't understand how this scenario can measure client performance?

keps/0012-label-index.md
adohe: @lavalamp Adding a layer between apiserver and etcd sounds like a reasonable approach, build DB like index query into etcd seems like not that possible. So this KEP just focus on client side.

lavalamp:
/hold

This KEP is still ambiguous in multiple places about what it wants to do clientside vs in the server. I think that should be very clear throughout before the discussion continues too much further.

PR #2204, opened 2018-06-01
Document pod startup SLO

wojtek-t:
/assign @shyamjvs @countspongebob

### DIFF
--- a/sig-scalability/slos/pod_startup_latency.md
+++ b/sig-scalability/slos/pod_startup_latency.md
@@ -1,0 +1,54 @@
+## Pod startup latency SLI/SLO details
+
+### User stories
+- As a user of vanilla Kubernetes, I want some guarantee how quickly my pods
+will be started.
+
+### Other notes
+- Only schedulable and stateless pods contribute to the SLI:
+  - If there is no space in the cluster to place the pod, there is not much
+    we can do about it (it is task for Cluster Autoscaler which should have
+    separate SLIs/SLOs).
+  - If placing a pod requires preempting other pods, that may heavily depend
+    on the application (e.g. on their graceful termination period). We don't
+    want that to contribute to this SLI.
+  - Mounting disks required by non-stateless pods may potentially also require
+    non-negligible time, not fully dependent on Kubernetes.
+- We are explicitly excluding image pulling from time the SLI. This is
+because it highly depends on locality of the image, image registry performance
+characteristic (e.g. throughput), image size itself, etc. Since we have
+no control over any of those (and all of those would significantly affect SLI)
+we decided to simply exclude it.
+- We are also explicitly excluding time to run init containers, as, again, this
+is heavily application-dependent (and does't depend on Kubernetes itself).
+- The answer to question "when pod should be considered as started" is also
+not obvious. We decided for the semantic of "when all its containers are
+reported as started and observed via watch", because:
+  - we require all containers to be started (not e.g. the first one) to ensure
+    that the pod is started. We need to ensure that pontential regressions like
+    linearization of container startups within a pod will be catch by this SLI.
+  - note that we don't require all container to be running - if some of them
+    finished before the last one was started that is also fine. It is just
+    required that all of them has been started (at least once).
+  - we don't want to rely on "readiness checks", because they heavily
+    depend on the application. If the application takes couple minutes to
+    initialize before it starts responding to readiness checks, that shouldn't
+		count towards Kubernetes performance.
+  - even if your application started, many control loops in Kubernetes will
+    not fire before they will observe that. If Kubelet is not able to report
+    the status due to some reason, other parts of the system will not have
+    a way to learn about it - this is why reporting part is so important
+    here.
+  - since watch is so centric to Kubernetes (and many control loops are
+    triggered by specific watch events), observing the status of pod is
+    also part of the SLI (as this is the moment when next control loops
+    can potentially be fired).
+
+### TODOs
+- We should try to provide guarantees for non-stateless pods (the threshold
+may be higher for them though).
+- Revisit whether we want "watch pod status" part to be included in the SLI.
+
+### Test scenario
+
+__TODO: Descibe test scenario.__
--- a/sig-scalability/slos/slos.md
+++ b/sig-scalability/slos/slos.md
@@ -102,6 +102,7 @@ Prerequisite: Kubernetes cluster is available and serving.
 | --- | --- | --- | --- |
 | __Official__ | Latency<sup>[1](#footnote1)</sup> of mutating<sup>[2](#footnote2)</sup> API calls for single objects for every (resource, verb) pair, measured as 99th percentile over last 5 minutes | In default Kubernetes installation, for every (resource, verb) pair, excluding virtual and aggregated resources and Custom Resource Definitions, 99th percentile per cluster-day<sup>[3](#footnote3)</sup> <= 1s | [Details](./api_call_latency.md) |
 | __Official__ | Latency<sup>[1](#footnote1)</sup> of non-streaming read-only<sup>[4](#footnote3)</sup> API calls for every (resource, scope<sup>[5](#footnote4)</sup>) pair, measured as 99th percentile over last 5 minutes | In default Kubernetes installation, for every (resource, scope) pair, excluding virtual and aggregated resources and Custom Resource Definitions, 99th percentile per cluster-day (a) <= 1s if `scope=resource` (b) <= 5s if `scope=namespace` (c) <= 30s if `scope=cluster` | [Details](./api_call_latency.md) |
+| __Official__ | Startup latency of stateless<sup>[6](#footnode6)</sup> and schedulable<sup>[7](#footnote7)</sup> pods, excluding time to pull images and run init containers, measured from pod creation timestamp to when all its containers are reported as started and observed via watch, measured as 99th percentile over last 5 minutes | In default Kubernetes installation, 99th percentile per cluster-day <= 5s | [Details](./pod_startup_latency.md) |
 
 <a name="footnote1">\[1\]</a>By latency of API call in this doc we mean time
 from the moment when apiserver gets the request to last byte of response sent
@@ -123,9 +124,12 @@ if the request is about a single object, (b) `namespace` if it is about objects
 from a single namespace or (c) `cluster` if it spawns objects from multiple
 namespaces.
 
+<a name="footnode6">[6\]</a>A `stateless pod` is defined as a pod that doesn't
+mount volumes with sources other than secrets, config maps, downward API and
+empty dir.
 
-__TODO: Migrate existing SLIs/SLOs here:__
-- __Pod startup time__
+<a name="footnode7">[7\]</a>By schedulable pod we mean a pod that can be
+scheduled in the cluster without causing any preemption.
 
 ### Burst SLIs/SLOs
 


sig-scalability/slos/pod_startup_latency.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Other notes
+- Only schedulable and stateless pods contribute to the SLI:
+  - If there is no space in the cluster to place the pod, there is not much
+    we can do about it (it is task for Cluster Autoscaler which should have
+    separate SLIs/SLOs).
+  - If placing a pod requires preempting other pods, that may heavily depend
+    on the application (e.g. on their graceful termination period). We don't
+    want that to contribute to this SLI.
+  - Mounting disks required by non-stateless pods may potentially also require
```
shyamjvs: I need to think if there are more scenarios under which the SLI may undesirably be affected. But maybe we don't need to block this PR on that.

A thought - it might be easier to define what scenario we "include" instead of explicitly providing all scenarios we "exclude"?

sig-scalability/slos/pod_startup_latency.md
```diff
@@ -0,0 +1,47 @@
...
+    non-negligible time, not fully dependent on Kubernetes.
+- We are explicitly excluding image pulling from time the SLI. This is
+because it highly depends on locality of the image, image registry performance
+characteristic (e.g. throughput), image size itself, etc. Since we have
+no control over any of those (and all of those would significantly affect SLI)
+we decided to simply exclude it.
+- We are also explicitly excluding time to run init containers, as, again, this
+is heavily application-dependent (and does't depend on Kubernetes itself).
+- The answer to question "when pod should be considered as started" is also
+not obvious. We decided for the semantic of "when all its containers are
```
shyamjvs: Can you please confirm this? I'm not sure if it's "all containers are reported as running". IIRC it is podrunning phase (which is at least one container running or sth like that).

sig-scalability/slos/pod_startup_latency.md
```diff
@@ -0,0 +1,47 @@
...
+  - we don't want to rely on "readiness checks", because they heavily
+    depend on the application. If the application takes couple minutes to
+    initialize before it starts responding to readiness checks, that shouldn't
+		count towards Kubernetes performance.
+  - even if your application started, many control loops in Kubernetes will
+    not fire before they will observe that. If Kubelet is not able to report
+    the status due to some reason, other parts of the system will not have
+    a way to learn about it - this is why reporting part is so important
+    here.
+  - since watch is so centric to Kubernetes (and many control loops are
```
shyamjvs: I'm sure this was a conscious decision in the past and not asking to change this, but one might also reason why watch latency shouldn't be included. For e.g a client who's listing directly from etcd or even from watch cache (where the pod is updated but the corresponding event wasn't delivered yet) may not suffer from this latency :)

sig-scalability/slos/pod_startup_latency.md
wojtek-t: I'm not necessary saying that it's what we measure - I'm saying this is what we SHOULD measure. ;)

sig-scalability/slos/pod_startup_latency.md
wojtek-t: Agree - but if we're talking about core components (e.g. endpoints controller), it watches.

sig-scalability/slos/slos.md
```diff
@@ -123,9 +124,12 @@ if the request is about a single object, (b) `namespace` if it is about objects
 from a single namespace or (c) `cluster` if it spawns objects from multiple
 namespaces.
 
+<a name="footnode6">[6\]</a>A `stateless pod` is defined as a pod that doesn't
```
shyamjvs: Actually, IIUC even with just secrets and configmaps it might be still possible to affect the SLI. For e.g if there are too many or too big of them.

wojtek-t: For both - we may tune it if/when we have numbers. I don't think we can say more now.

sig-scalability/slos/pod_startup_latency.md
shyamjvs: There are at least following other "exclude" scenarios I could think of in the meanwhile which might affect the latency:

- too many secrets/configmaps used by the pod
- too big secrets/configmaps?
- pod with complicated scheduling (anti-affinity, etc)

My feeling is we might be able to express our include/exclude scenarios in a more succinct way (sth like "stateless and schedulable pods that don't exercise custom scheduling/setup features that can cause increased delays in running the pod (e.g antiaffinity. volume mounts)"). I know that isn't perfect, but just giving the idea. Feel free to disagree.

wojtek-t: 1. too many/too big secrets/configmaps - if we have numbers we may consider changing that; I don't want to provide something that isn't very clear.
2. scheduling features - that isn't something that should be excluded from SLI; it may be TODO to improve it, but they should be covered by SLI

sig-scalability/slos/pod_startup_latency.md
shyamjvs: In that case do we have a rationale why we "should" measure it that way? (because our ideas of "pod started" can be different:)

sig-scalability/slos/pod_startup_latency.md
shyamjvs: Sure.. I'm just saying that there are users whose view of 'pod startup' doesn't have this latency. And controllers are typically not user-facing - wdyt?

sig-scalability/slos/slos.md
shyamjvs: Sgtm - let's not complicate this already too-complicated SLI definition for now :)

sig-scalability/slos/pod_startup_latency.md
```diff
@@ -0,0 +1,47 @@
...
+  - If there is no space in the cluster to place the pod, there is not much
+    we can do about it (it is task for Cluster Autoscaler which should have
+    separate SLIs/SLOs).
+  - If placing a pod requires preempting other pods, that may heavily depend
+    on the application (e.g. on their graceful termination period). We don't
+    want that to contribute to this SLI.
+  - Mounting disks required by non-stateless pods may potentially also require
+    non-negligible time, not fully dependent on Kubernetes.
+- We are explicitly excluding image pulling from time the SLI. This is
+because it highly depends on locality of the image, image registry performance
```
hectorj2f: Sorry to jump in this PR! Should you specify that this timing could vary based on container runtime ? Also, when you mean Kubernetes default installation means you use Docker as container runtime. Thanks

sig-scalability/slos/pod_startup_latency.md
wojtek-t: the points below describe that

sig-scalability/slos/pod_startup_latency.md
wojtek-t: > I'm just saying that there are users whose view of 'pod startup' doesn't have this latency. 

Yes

> And controllers are typically not user-facing - wdyt?

But if your pod is part of service, what you care in the end is how fast it will be added to your service.
So this generally matters (if your pods is not part of service, in many cases you don't even care about its status being reported).

That said, yes - it's debatable whether we want watch or no. Added a TODO to revisit that.

sig-scalability/slos/pod_startup_latency.md
wojtek-t: > Should you specify that this timing could vary based on container runtime ?

That may matter for SLO (the fact that some runtimes are slower than other doesn't affect SLI - i.e. what we measure).

For SLO - we assume default k8s installation, which is Docker.

So yes, while the container runtime matters, it's not important here.

hectorj2f: @wojtek-t Thanks for the clarification

sig-scalability/slos/pod_startup_latency.md
shyamjvs: Yes, but I can't find the reasoning for e.g why we should consider "all containers running", but not "at least one container running". Not that the latter is what we [seem to define](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) as "PodRunning" status.

sig-scalability/slos/pod_startup_latency.md
shyamjvs: That lgtm - thnx.

shyamjvs: Btw - I think the part you mention about pod being available in a service sounds like a different SLI to me (which of course would be influenced by watch latency - and we anyway seem to have another SLI for watch latency) :)

wojtek-t: Yes - being available in a service is a different SLI.
But it cannot even become part of service until kubelet reports its status (and that status will be observed) - only then something may start processing.
I can imagine an artificial scenario that "pod status update latency" have 10x higher watch latency. Should this then contribute to service SLI - I think no. This will ideally be a separate SLI, but since we're not there yet, I think it's better to have it here...
But yeah - TODO is already there.

shyamjvs: Yes, and that is why I was mentioning "we anyway seem to have another SLI for watch latency" for those kind of situations (which is an SLI we're already having (at least in our docs)?).
Anyway - I raised that just for discussion.

sig-scalability/slos/pod_startup_latency.md
wojtek-t: The definition isn't fully precise. What it's going to address are cases when there are more container within a pod but some of them already finished (e.g. succeeded).

Looking into kubelet code where this is computed seems to confirm my theory:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L1293

So what happens:
- if there are container in waiting state => pod pending
- if there are containers in unkwonwn state => pod pending


Now: why I think we should wait for *all*, not just for some?
The reason is that a Pod is defined as "all its containers". So if 2/3 of containers are running, the pod itself is not running (at least I can imagine situations like that very easily - some functionality may simply be not available).
Now, imagine that one container is simply misconfigured (e.g. has non-existing image), so it will never be started. but all other container are already running. Can we report pod as running? No.

I added a comment with explanation.

wojtek-t:
PTAL

sig-scalability/slos/pod_startup_latency.md
shyamjvs: There are few points on which I disagree:

1. If a pod has a crashlooping container - we shouldn't wait for it for the purpose of our SLI. IMO that's out of the scope of "pod startup" from k8s perspective:
  - we don't know why it's failing and it might be some user-space problem (e.g image doesn't exist, seg-fault in the code, etc)
  - it's super easy to break the SLO with that definition

2. If all containers except one start and let's say they finish instantaneously while the remaining is still starting, your SLI will be waiting indefinitely for "all containers running". While "at least one container running (and all containers created)" will be a bound condition.

3. Your argument around "pod is not functional if only 2/3 of it's containers are running" may or may not be true - and it depends on the user pod in question (e.g some pods may be functional with 2/3 running). And I think that is for readiness gates to check, which again I believe is a user-space problem we shouldn't be including in our SLI.

In general, I believe that from k8s perspective our measurement should actually be limited to container creations (i.e setting up their cgroups, network, namespaces, etc) and we're done - what happens next is probably user-space. I wasn't around back then, but the [current semantic](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L1301-L1303) of "PodRunning" was probably a conscious choice for some reason?

cc @yujuhong @freehan (as they may have some useful inputs here)

smarterclayton: The semantic was chosen as a boundary that most people perceive as running.  In restart never pods, for instance, two related containers may start at different times, and one may complete before the other has started, but the end user perceives start as the moment the first container lights up.

If you are omitting pull latency, then I think this SLI should require all containers to have started because an accidental start linearization regression would have an impact on both services and jobs, and an end user that had 10 containers that started linearly would rightly conclude Kubernetes is “broken” when using multiple containers.

sig-scalability/slos/pod_startup_latency.md
```diff
@@ -0,0 +1,51 @@
...
+- The answer to question "when pod should be considered as started" is also
+not obvious. We decided for the semantic of "when all its containers are
+reported as running and observed via watch", because:
+  - we require all containers to be running (not e.g. the first one) to ensure
+    that the pod is started. We need to exclude situations when e.g. one of
+    containers has bad image and will never be able to start.
+  - we don't want to rely on "readiness checks", because they heavily
+    depend on the application. If the application takes couple minutes to
+    initialize before it starts responding to readiness checks, that shouldn't
+		count towards Kubernetes performance.
```
smarterclayton: Except in a practical sense an application can’t be used until it’s ready.  But that feels more like a service SLI than a pod SLI (time from first container start -> another pod can contact the container via a service).

sig-scalability/slos/pod_startup_latency.md
wojtek-t: > If you are omitting pull latency, then I think this SLI should require all containers to have started because an accidental start linearization regression would have an impact on both services and jobs, and an end user that had 10 containers that started linearly would rightly conclude Kubernetes is “broken” when using multiple containers.

That is exactly my main argument (which apparently I wasn't able to describe as I wanted - thanks @smarterclayton )

Also, to address some of @shyamjvs comments.

> we don't know why it's failing and it might be some user-space problem (e.g image doesn't exist, seg-fault in the code, etc)

That's not true - the definition says that all container started (not that they are still running). If your app crashes, that's fine - it was started and failed, but will already be reported as started (potentially even more than once).

> In general, I believe that from k8s perspective our measurement should actually be limited to container creations

+100 for that and that is exactly what we're doing.
But "container creations" means *all* containers, not some containers.

wojtek-t: I just realized that i wrote not exactly what i wanted to write :)
So this should be now clarified to capture all the above.

sig-scalability/slos/pod_startup_latency.md
shyamjvs: Ok - that sounds reasonable to me.

sig-scalability/slos/pod_startup_latency.md
shyamjvs: > But "container creations" means all containers, not some containers.

So to clarify - I do agree with that we should consider `all container creations` as part of the SLI. What I was against was to consider `all containers running` (which is what seemed to be indicated in your comment above mine) - my reason being that 'container running' is more of a user-space thing than k8s-specific. With that SLI, we would catch the linearization argument Clayton pointed above.

That said, the current formulation lgtm now.

shyamjvs:
@wojtek-t Thanks for addressing those comments!

/lgtm

PR #2217, opened 2018-06-05
contribx: link charter doc in readme

tpepper:
The main readme for the SIG Contributor Experience covers at a high
level some of the information that is detailed in the new charter
document and is a bit out of date relative to it.  This freshens things a little.

### DIFF
--- a/sigs.yaml
+++ b/sigs.yaml
@@ -746,11 +746,13 @@ sigs:
   - name: Contributor Experience
     dir: sig-contributor-experience
     mission_statement: >
-      Developing and sustaining a healthy community of contributors is critical
-      to scaling the project and growing the ecosystem. We need to ensure our
-      contributors are happy and productive, we need to break the commit-rate
-      ceiling we hit in July 2015, and we need to get the monotonically growing
-      PR merge latency and numbers of open PRs and issues under control.
+      Developing and sustaining a healthy community of contributors
+      is critical to scaling the project and growing the ecosystem.
+      We need to ensure our contributors are happy and productive,
+      and that there are not bottlenecks hindering the project in,
+      for example: feature velocity, community scaling, pull request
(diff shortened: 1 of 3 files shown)


cblecker:
This is part of the autogenerated section of the doc.

https://github.com/kubernetes/community/blob/9151896ac90a1a3bacb4f33768e7bde96abbde22/sigs.yaml#L759-L764

tpepper:
Fixed up after the auto test failed and showed I was editing a generated file...in parallel to @cblecker pointing out the same.

sigs.yaml
```diff
@@ -757,11 +757,16 @@ sigs:
...
-      PR merge latency and numbers of open PRs and issues under control.
+      Developing and sustaining a healthy community of contributors
+      is critical to scaling the project and growing the ecosystem.
+      We need to ensure our contributors are happy and productive,
+      and that there are not bottlenecks hindering the project in,
+      for example: feature velocity, community scaling, pull request
+      latency, and absolute numbers of open pull requests and open
+      issues.
+
+      See the SIG Contributor Experience [charter document](charter.md)
```
cblecker: Can you add an extra new line here?

tpepper: Done.

tpepper:
Updated to match new "charter_link:" format in the sigs.yaml instead of manually linking in the markdown mission statement.

cblecker:
I'm okay with this, but a change to the mission statement should be reviewed by contribex as a whole.

/assign @parispittman @Phillels

parispittman:
I'm ok with all changes. I think we need a better mission statement in general but the change is better than the current. I'll discuss with Elsie, Christoph, and Garrett.

parispittman:
/lgtm
/approve

note: spoke to elsie, garrett, and christoph

PR #2218, opened 2018-06-06
Remove not need word "or"

chenyb4:
Signed-off-by: Yuanbin.Chen <cybing4@gmail.com>


cblecker:
Unneeded change. The current form is also correct.
/close

PR #2224, opened 2018-06-07
SIG Cloud Provider KEP: Reporting Conformance Test Results to Testgrid

andrewsykim:
KEP for reporting conformance test results from all cloud providers.

@AishSundar @BenTheElder (sig-testing)
@calebamiles @jdumars (sig-release)

### DIFF
--- a/keps/sig-cloud-provider/0003-testgrid-conformance-e2e.md
+++ b/keps/sig-cloud-provider/0003-testgrid-conformance-e2e.md
@@ -1,0 +1,270 @@
+---
+kep-number: 003
+title: Reporting Conformance Test Results to Testgrid
+authors:
+  - "@andrewsykim"
+owning-sig: sig-cloud-provider
+participating-sigs:
+  - sig-testing
+  - sig-release
+  - sig-aws
+  - sig-azure
+  - sig-gcp
+  - sig-ibmcloud
+  - sig-openstack
+  - sig-vmware
+reviewers:
+  - TBD
+approvers:
+  - TBD
+editor: TBD
+creation-date: 2018-06-06
+last-updated: 2018-06-06
+status: provisional
+
+---
+
+# Reporting Conformance Test Results to Testgrid
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+
+## Summary
+
+This is a KEP outlining the motivation behind why cloud providers should periodically upload E2E conformance test results to [Testgrid](https://github.com/kubernetes/test-infra/tree/master/testgrid) and how a cloud provider can go about doing this.
+
+## Motivation
+
+The primary motivation behind collecting conformance test results from various cloud providers on a regular basis is to inform sig-release of any critical bugs. It's important to collect results from various cloud providers to increase coverage and inform other cloud providers of bugs they may be impacted by.
+
+### Goals
+
+* All SIGs / Subprojects owners representing a cloud provider is aware of the importance of frequently uploading conformance test results to testgrid
+* There is a clear and detailed process in place for any cloud provider to upload conformance test results to Testgrid.
+
+### Non-Goals
+
+* Test coverage - increasing test coverage is outside the scope of this KEP
+* CI/CD - what CI/CD tool is used to run E2E tests and upload results is outside the scope of this KEP. It is up to the cloud provider to decide where/how tests are run.
+* Cluster Provisioning - how a cluster is provisioned is outside the scope of this KEP.
+
+## Proposal
+
+We would like to propose that every Kubernetes cloud provider reports conformance test results for every patch version of Kubernetes at the minimum. Running conformance tests against master and on pre-release versions are highly encouraged but will not be a requirement.
+
+### Implementation Details/Notes/Constraints
+
+Before continuing, it is highly recommended you read the following documentation provided by sig-testing:
+* [Testgrid Configuration](https://github.com/kubernetes/test-infra/tree/master/testgrid#testgrid)
+* [Display Conformance Tests with Testgrid](https://github.com/kubernetes/test-infra/blob/master/testgrid/conformance/README.md)
+
+#### How to Run E2E Conformance Tests
+
+This KEP outlines two ways of running conformance tests, the first using [Sonobuoy](https://github.com/heptio/sonobuoy) and the second using [kubetest](https://github.com/kubernetes/test-infra/tree/master/kubetest). Though Sonobuoy is easier to setup, it does not guarantee that you will run the latest set of conformance tests. Kubetest though requiring a bit more work to setup, can ensure that you are running the latest set of conformance tests.
+
+At this point we will assume that you have a running cluster and that `kubectl` is configured to point to that cluster with admin access.
+
+#### Sonobuoy
+
+You should use Sonobuoy if you would like to run the standard set of CNCF conformance tests. This may exclude any new tests added by the latest versions of Kubernetes.
+
+##### Installing Sonobuoy
+
+The following is mostly a [copy of the sonobuoy documentation](https://github.com/heptio/sonobuoy#download-and-run).
+
+First install sonobuoy. The following command adds it to the `$GOBIN` environment variable which is expected to be part of your `$PATH` environment variable.
+```
+$ go get -u github.com/heptio/sonobuoy
+```
+
+##### Running Conformance Tests with Sonobuoy
+
+You can then start e2e tests by simply running:
+```
+$ sonobuoy run
+Running plugins: e2e, systemd-logs
+INFO[0001] created object                                name=heptio-sonobuoy namespace= resource=namespaces
+INFO[0001] created object                                name=sonobuoy-serviceaccount namespace=heptio-sonobuoy resource=serviceaccounts
+INFO[0001] created object                                name=sonobuoy-serviceaccount-heptio-sonobuoy namespace= resource=clusterrolebindings
+INFO[0001] created object                                name=sonobuoy-serviceaccount namespace= resource=clusterroles
+INFO[0001] created object                                name=sonobuoy-config-cm namespace=heptio-sonobuoy resource=configmaps
+INFO[0002] created object                                name=sonobuoy-plugins-cm namespace=heptio-sonobuoy resource=configmaps
+INFO[0002] created object                                name=sonobuoy namespace=heptio-sonobuoy resource=pods
+INFO[0002] created object                                name=sonobuoy-master namespace=heptio-sonobuoy resource=services
+```
+
+You can then check the status of your e2e tests by running
+```
+$ sonobuoy status
+PLUGIN		STATUS	COUNT
+e2e		running	1
+systemd_logs	running	3
+
+Sonobuoy is still running. Runs can take up to 60 minutes.
+```
+
+E2E tests can take up to an hour. Once your tests are done, you can download a snapshot of your results like so:
+```
+$ sonobuoy retrieve .
+```
+
+Once you have the following snapshot, extract it's contents like so
+```
+$ mkdir ./results; tar xzf *.tar.gz -C ./results
+```
+
+At this point you should have the log file and JUnit results from your tests:
+```
+results/plugins/e2e/results/e2e.log
+results/plugins/e2e/results/junit_01.xml
+```
+
+#### Kubetest
+
+You should use kubetest if you want to run the latest set of tests in upstream Kubernetes. This is highly recommended by sig-testing so that new tests can be accounted for in new releases.
+
+##### Installing Kubetest
+
+Install kubetest using the following command which adds it to your `$GOBIN` environment variable which is expected to be part of your `$PATH` environment variable.
+```
+go get -u k8s.io/test-infra/kubetest
+```
+
+##### Running Conformance Test with kubetest
+
+Now you can run conformance test with the following:
+```
+cd /path/to/k8s.io/kubernetes
+export KUBERNETES_CONFORMANCE_TEST=y
+kubetest \
+    # conformance tests aren't supposed to be aware of providers
+    --provider=skeleton \
+    # tell ginkgo to only run conformance tests
+    --test --test_args="--ginkgo.focus=\[Conformance\]" \
+    # grab the most recent CI tarball of kubernetes 1.10, including the tests
+    --extract=ci/latest-1.10 \
+    # directory to store junit results
+    --dump=$(pwd)/_artifacts | tee ./e2e.log
+```
+
+Note that `--extract=ci/latest-1.10` indicates that we want to use the binaries/tests on the latest version of 1.10. You can use `--extract=ci/latest` to run the latest set of conformance tests from master.
+
+Once the tests have finished (takes about an hour) you should have the log file and JUnit results from your tests:
+```
+e2e.log
+_artifacts/junit_01.xml
+```
+
+#### How to Upload Conformance Test Results to Testgrid
+
+##### Requesting a GCS Bucket
+
+Testgrid requires that you store results in a publicly readable GCS bucket. If for whatever reason you cannot set up a GCS bucket, please contact @BenTheElder or more generally the [gke-kubernetes-engprod](mailto:gke-kubernetes-engprod@google.com) team to arrange for a Google [GKE](https://cloud.google.com/kubernetes-engine/) EngProd provided / maintained bucket for hosting your results.
+
+##### Authenticating to your Testgrid Bucket
+
+Assuming that you have a publicly readable bucket provided by the GKE team, you should have been provided a service account JSON file which you can use with [gcloud](https://cloud.google.com/sdk/downloads) to authenticate with your GCS bucket.
+
+```
+$ gcloud auth activate-service-account --key-file /path/to/k8s-conformance-serivce-accout.json
+Activated service account credentials for: [demo-bucket-upload@k8s-federated-conformance.iam.gserviceaccount.com]
+```
+
+##### Uploading results to Testgrid
+
+At this point you should be able to upload your testgrid results to your GCS bucket. You can do so by running a python script availabile [here](https://github.com/kubernetes/test-infra/tree/master/testgrid/conformance). For this example, we upload results for v1.10 into it's own GCS prefix.
+```
+git clone https://github.com/kubernetes/test-infra
+cd test-infra/testgrid/conformance
+./upload_e2e.py --junit /path/to/junit_01.xml \
+       --log /path/to/e2e.log \
+       --bucket=gs://k8s-conformance-demo/cloud-provider-demo/e2e-conformance-release-v1.10
+Uploading entry to: gs://k8s-conformance-demo/cloud-provider-demo/e2e-conformance-release-v1.10/1528333637
+Run: ['gsutil', '-q', '-h', 'Content-Type:text/plain', 'cp', '-', 'gs://k8s-conformance-demo/cloud-provider-demo/e2e-conformance-release-v1.10/1528333637/started.json'] stdin={"timestamp": 1528333637}
+Run: ['gsutil', '-q', '-h', 'Content-Type:text/plain', 'cp', '-', 'gs://k8s-conformance-demo/cloud-provider-demo/e2e-conformance-release-v1.10/1528333637/finished.json'] stdin={"timestamp": 1528337316, "result": "SUCCESS"}
+Run: ['gsutil', '-q', '-h', 'Content-Type:text/plain', 'cp', '~/go/src/k8s.io/kubernetes/results/plugins/e2e/results/e2e.log', 'gs://k8s-conformance-demo/cloud-provider-demo/e2e-conformance-release-v1.10/1528333637/build-log.txt']
+Run: ['gsutil', '-q', '-h', 'Content-Type:text/plain', 'cp', '~/go/src/k8s.io/kubernetes/results/plugins/e2e/results/e2e.log', 'gs://k8s-conformance-demo/cloud-provider-demo/e2e-conformance-release-v1.10/1528333637/artifacts/e2e.log']
+Done.
+```
+
+##### Testgrid Configuration
+
+Next thing you want to do is configure testgrid to read results from your GCS bucket. There are two [configuration](https://github.com/kubernetes/test-infra/tree/master/testgrid#configuration) steps required. One for your [test group](https://github.com/kubernetes/test-infra/tree/master/testgrid#test-groups) and one for your [dashboard](https://github.com/kubernetes/test-infra/tree/master/testgrid#dashboards).
+
+To add a test group update [config.yaml](https://github.com/kubernetes/test-infra/blob/master/testgrid/config.yaml) with something like the following:
+```
+test_groups:
+...
+...
+- name: cloud-provider-demo-e2e-conformance-release-v1.10
+  gcs_prefix: k8s-conformance-demo/cloud-provider-demo/e2e-conformance-release-v1.10
+```
+
+To add a link to your results in the testgrid dashboard, update [config.yaml](https://github.com/kubernetes/test-infra/blob/master/testgrid/config.yaml) with something like the following:
+```
+dashboards:
+...
+...
+- name: conformance-demo-cloud-provider
+  dashboard_tab:
+  - name: "Demo Cloud Provider, v1.10"
+    description: "Runs conformance tests for cloud provier demo on release v1.10"
+    test_group_name: cloud-provider-demo-e2e-conformance-release-v1.10
+```
+
+Once you've made the following changes, open a PR against the test-infra repo adding the sig testing label (`/sig testing`) and cc'ing @kubernetes/sig-testing-pr-reviews. Once your PR merges you should be able to view your results on https://k8s-testgrid.appspot.com/ which should be ready to be consumed by the necessary stakeholders (sig-release, sig-testing, etc).
+
+#### Lifecycle of Test Results
+
+You can configure the lifecycle of testgrid results by specifying fields like `days_of_results` on your test group configuration. More details about this in the [Testgrid Advanced Configuration](https://github.com/kubernetes/test-infra/tree/master/testgrid#advanced-configuration) docs. If for whatever reason you urgently need to delete testgrid results, you can contact someone from sig-testing.
+
+#### Examples
+
+Here are some more concrete examples of how other cloud providers are running conformance tests and uploading results to testgrid:
+* Open Stack
+    * [OpenLab zuul job for running/uploading testgrid results](https://github.com/theopenlab/openlab-zuul-jobs/tree/master/playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance)
+    * [OpenStack testgrid config](https://github.com/kubernetes/test-infra/pull/7670)
+    * [OpenStack conformance tests dashboard](https://github.com/kubernetes/test-infra/pull/8154)
+
+
+### Risks and Mitigations
+
+#### Operational Overhead
+
+Operating CI/CD system to run conformance tests on a regular basis may incur extra work from every cloud provider. Though we anticipate the benefits of running conformance tests to outweight the operational overhead, in some cases it may not.
+
+Mitigation: TODO
+
+#### Misconfigured Tests
+
+There are various scenarios where cloud providers may mistakenly upload incorrect conformance tests results. One example being uploading results for the wrong Kubernetes version.
+
+Mitigation: TODO
+
+#### Flaky Tests
+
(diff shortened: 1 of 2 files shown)


keps/sig-cloud-provider/003-testgrid-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+
+Operating CI/CD system to run conformance tests on a regular basis may incur extra work from every cloud provider. Though we anticipate the benefits of running conformance tests to outweight the operational overhead, in some cases it may not.
+
+Mitigation: TODO
+
+#### Misconfigured Tests
+
+There are various scenarios where cloud providers may mistakenly upload incorrect conformance tests results. One example being uploading results for the wrong Kubernetes version.
+
+Mitigation: TODO
```
andrewsykim: @AishSundar @BenTheElder would love some feedback on how to mitigate this going forward

andrewsykim: same with flaky tests below

keps/sig-cloud-provider/003-testgrid-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### Flaky Tests
+
+Tests can fail for various reasons in any cloud environment and may raise false negatives for the release team.
```
BenTheElder: we expect some level of flaking to occur, but the release team will probably not move/keep jobs into the release blocking dashboards if they are very flaky. (they will still be in other dashboards of course). for the moment which jobs are in the release dashboard is ~up to the release team, and they may choose to ignore other dashboards.

keps/sig-cloud-provider/003-testgrid-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+Once you've made the following changes, open a PR against the test-infra repo. Once your PR merges you should be able to view your results on https://k8s-testgrid.appspot.com/ which should be ready to be consumed by the necessary stakeholders (sig-release, sig-testing, etc).
+
+
+#### Examples
+
+Here are some more concrete examples of how other cloud providers are running conformance tests and uploading results to testgrid:
+* Open Stack
+    * [OpenLan zuul job for running/uploading testgrid results](https://github.com/theopenlab/openlab-zuul-jobs/tree/master/playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance)
```
BenTheElder: s/OpenLan/OpenLab/

keps/sig-cloud-provider/003-testgrid-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Implementation Details/Notes/Constraints
+
+Before continuing, it is highly recommended you read the following documentation provided by sig-testing:
+* [Testgrid Configuration](https://github.com/kubernetes/test-infra/tree/master/testgrid#testgrid)
+* [Display Conformance Tests with Testgrid](https://github.com/kubernetes/test-infra/blob/master/testgrid/conformance/README.md)
+
+#### How to Run E2E Conformance Tests
+
+[Sonobuoy](https://github.com/heptio/sonobuoy) is the recommended tool to run conformance tests on your cluster. At this point we will assume that you have a running cluster and that `kubectl` is configured to point to that cluster with admin access.
```
BenTheElder: There is actually a major downside to using Sonobuoy vs building the e2e tests with kubernetes in continuous integration in particular, AFAIK. If you use the pre-released Sonobuoy images then you will not be testing the latest conformance tests as they are developed. Running these directly or with [kubetest](https://github.com/kubernetes/test-infra/tree/master/kubetest) avoids this.

BenTheElder:
I expect providers will want to perform two kinds of testing:

1) My current hosted product / platform is conformant with the released tests, IE passes sonobuoy

2) Continuous integration where I've built kubernetes from a release branch / master, gotten it running on my platform, and tested it with the tests built with it. 

The first is obviously important to providers, but the release team and myself are going to be far more interested in 2) so we can validate:
- the core components in any ongoing development to kubernetes
- the conformance tests themselves as they are developed.

I would like to push towards 2), which does have a bit more overhead and needs a bit more clarity regarding versions.

For OpenStack the tests run now are 2) by checking eg `release-1.11` branch, building k8s + tests, then running.
For GCE they are 2) the same as OpenStack, and also another set of tests checking out the latest tagged release, IE `1.10.2`. The OpenStack type are the most valuable signal for the release team, the tagged release signal is also useful but less important, imho.

Also for 2) [making sure we're actually using a current build of kubernetes and the tests is a concern](https://github.com/kubernetes/community/pull/2224#discussion_r193852529).

keps/sig-cloud-provider/003-testgrid-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+##### Requesting a GCS Bucket
+
+Testgrid requires that you store results in a publicly readable GCS bucket. If for whatever reason you cannot set up a GCS bucket, please contact @BenTheElder or more generally the [gke-kubernetes-engprod](mailto:gke-kubernetes-engprod@google.com) team to arrange for a Google [GKE](https://cloud.google.com/kubernetes-engine/) EngProd provided / maintained bucket for hosting your results.
+
+##### Authenticating to your Testgrid Bucket
+
+Assuming that you have a publicly readable bucket provided by the GKE team, you should have been provided a service account JSON file which you can use with [gcloud](https://cloud.google.com/sdk/downloads) to authenticate with your GCS bucket.
+
+```
+$ gcloud auth activate-service-account --key-file /path/to/k8s-conformance-serivce-accout.json
```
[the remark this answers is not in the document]
andrewsykim: oh.. that's quite handy. Is that the preferred method going forward?

andrewsykim:
I definitely agree that pushing for 2) would be ideal so we can catch bugs before they're released. Because of the overhead I had felt that settling for 1) as a requirement and 2) as strongly recommended seemed like a good middle ground (at least for now). My gut feeling is that most providers will end up doing 2) anyways for the reasons you brought up.

keps/sig-cloud-provider/003-testgrid-e2e.md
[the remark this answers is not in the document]
BenTheElder: either works fine, but I think for some setups this will be easier? It also means the cloud providers just need to have the gcloud tools available, they don't necessarily need to be aware of how to use them

BenTheElder:
Totally agree!
Though I really hope we can push to get plenty up to speed on 2) to catch
Kubernetes-wide conformance problems quickly like
https://github.com/kubernetes/kubernetes/issues/64359,
https://github.com/kubernetes/kubernetes/issues/64362.

We should also try to make 1) vs 2) clear, since 2) is more likely to fail
due to new tests etc. and does not necessarily mean that the provider is
non-conformant. One way could just be standardizing how the TestGrid
entries are named to clarify (?) I'm not sure what the best solution is
just yet though.

keps/sig-cloud-provider/003-testgrid-e2e.md
AishSundar: +1 on this comment. In the past whenever I wished to run conformance on the master to get latest coverage or test latest changes I used kubetest. If Sonobuoy supports this it will be good to call that out here. If not, plz include instructions on how to achieve this using Kubetest or the hack scripts. Thanks.

keps/sig-cloud-provider/003-testgrid-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+...
+...
+- name: conformance-demo-cloud-provider
+  dashboard_tab:
+  - name: "Demo Cloud Provider, v1.10"
+    description: "Runs conformance tests for cloud provier demo on release v1.10"
+    test_group_name: cloud-provider-demo-e2e-conformance-release-v1.10
+```
+
+Once you've made the following changes, open a PR against the test-infra repo. Once your PR merges you should be able to view your results on https://k8s-testgrid.appspot.com/ which should be ready to be consumed by the necessary stakeholders (sig-release, sig-testing, etc).
```
AishSundar: May be suggest sending the PR to '@kubernetes/sig-testing-pr-reviews' and CCing /sig testing ?

keps/sig-cloud-provider/003-testgrid-e2e.md
AishSundar: +1 on this. One of the major criteria in promoting a e2e test to conformance is it should flake <1% of the time. If for some reason this isnt met we will remove the test from conformance after investigation of rootcause and fix.

keps/sig-cloud-provider/003-testgrid-e2e.md
AishSundar: If the providers own their own GCS bucket, then they can go into the appropriate folder and delete data corresponsing to the run / move it to the right directory in the bucket etc. If the GCS bucket is owned by GKE EngProd, then they can mail gke-kubernetes-engprod@google.com, with details of the run and bucket, for fixing the data. 

@BenTheElder wdyt? accurate?

michelle192837: So TestGrid processes X results back until it stops finding changes, but that means that it varies how many columns back it looks. If you remove a GCS entry and want that run to not show up in TestGrid, the best way to enforce that is to force TestGrid to re-collect results for that test group.

To force TestGrid to re-collect, you can:
- Increase the test group's `days_of_results` field, [Recommended]
- Change the test group's `query` or `column_header` fields, or
- Change the test group name (which TestGrid will treat as a new test group).

If you need to wipe something urgently, you can also request that I or someone else on the team delete the state file for the test group, which will remove the results until TestGrid updates that group again.

BenTheElder: Small note: the changes mentioned above would be to the testgrid [config file](https://github.com/kubernetes/test-infra/blob/master/testgrid/config.yaml) in test-infra, by a PR.
thanks @michelle192837 :-)

andrewsykim: This is really good to know, I'll add a section about this.

keps/sig-cloud-provider/003-testgrid-e2e.md
andrewsykim: Sounds good, I will update to use `kubetest`, I was in favour of Sonobuoy originally because it was easier to run since it doesn't require setting up a development environment but I don't think the tradeoff is worth it.

keps/sig-cloud-provider/003-testgrid-e2e.md
andrewsykim: Sounds good, I will update to use kubetest, I was in favour of Sonobuoy originally because it was easier to run since it doesn't require setting up a development environment but I don't think the tradeoff is worth it.

AishSundar: I am not suggesting to remove Sonobuoy completely from here. But plz do mention Kubetest also as a means of triggering tests, esp if you want to test on latest changes.

BenTheElder: Yeah, I don't want to say that we necessarily need to mention any particular tool, I've mentioned Sonobuoy and kubetest in [the testgrid conformance usage guide](https://github.com/kubernetes/test-infra/tree/master/testgrid/conformance#usage-guide).
It should also be possible to run gingko / the conformance tests directly or via the hack scripts in kubernetes, but we should probably mention at least one way that involves using freshly built e2e tests, and how to use it correctly.

BenTheElder: Also: I definitely hear those concerns 😬 

I'd hope providers can eventually set up a matrix covering many versions, but even in small matrices I'd hope they can set aside one job to run (latest, latest), these tests are generally quite stable, which is expected of conformance tests.

The providers ought to be able to cross reference each other if we can track upstream Kubernetes together w/o pinning, at least for one periodic "job" each. If all of the providers start failing around the same time, it's probably Kubernetes, not the provider.

andrewsykim: This will be one of the top concerns for sure. I don't think we need a full proof solution yet and I think KEPs are supposed to be updated frequently so we can iterate on the KEP once we have a better story, likely once all the providers have some experience with the workflow. I like that with `kubetest` we do give every provider the ability to choose that (cluster version is version tested against, kubernetes repo version is version for conformance tests) which I think is enough for us to move forward.

timothysc: So sonobuoy will lag for the release just due the way artifacts are published.  If we were to continuously publish an upstream conformance test similar to what we do downstream the original statement is a non-starter.  This has been a todo to publish the artifact, but it's never been top of mind.  I'd be happy to take it up on 1.12.

I'll bring this up during the next sig-testing meeting.

BenTheElder: Sure, will these artifacts work for each branch? 
Conformance tests in particular are pretty easy to run from ginkgo / e2e.test directly without many args but for continuous integration really ought to be in sync with the k8s binaries directly... 
My concern is that the Sonobuoy e2e binaries won't be able to match the k8s updates when testing `master` in particular, unless the test can also build sonobuoy images ..?

Building and running the e2e tests from source for master continuous integration in particular seems unavoidable, but not necessarily for actual releases.

BenTheElder: The [testgrid conformance docs in test-infra](https://github.com/kubernetes/test-infra/tree/master/testgrid/conformance) call out using the official Sonobuoy config at https://github.com/cncf/k8s-conformance as the first option, which should *Definitely* be the default for "am I conformant", as that's what you should submit to the CNCF.

For CI though, to test with the latest build something like:
```sh
kubetest \
# grab the most recent CI tarball of kubernetes, including the tests
    --extract=ci/latest \ 
# conformance tests aren't supposed to be aware of providers
    --provider=skeleton \ 
# tell ginkgo to only run conformance tests
    --test --test_args="--ginkgo.focus=\[Conformance\]" \  
# make sure to grab the log and junit
     --dump=./_artifacts | tee ./e2e.log 
```
should work in the meantime.

timothysc: All we need todo is push some of the details from https://github.com/heptio/kube-conformance into an upstream artifact that is cleanly factored (does just 1 thing).  Some of the test containers from upstream, last I looked, rolled in a bunch of other stuff which from a community perspective can be confusing and unnecessary.  @mml and I wanted todo this several releases ago, but just never got to it.

timothysc: Sonobuoy itself is a more elaborate runner to gather multiple artifacts in a more reproduciable manner.  You don't need to rebuild those containers and they are widely used, vetted, versioned and tested.

BenTheElder: there aren't any other artifacts to gather in this case...?
But you should use the latest e2es in some cases (continuous integration...) and I don't see how that will work without rebuilding https://github.com/heptio/kube-conformance/blob/c43a9d2d1f096ccff8d5b069691a523b1b517b8e/Dockerfile#L24 to match the newest binaries.

timothysc: Depends on what they want todo.  A lot of folks create their own custom tests that they collect together.  Also, it's a reproducible artifact that you can handoff.  Can't do that with the e2es today.

BenTheElder: xref: https://github.com/kubernetes/test-infra/issues/8299 issue for sonobuoy automation instructions with the CNCF config + the testgrid uploader.

andrewsykim:
Comments addressed, PTAL @AishSundar @BenTheElder

BenTheElder:
er habit there on the hold, not necessary.
This looks pretty good to me, we should probably fill out the TODO though :-)
/hold cancel

keps/sig-cloud-provider/0003-testgrid-e2e.md
```diff
@@ -0,0 +1,270 @@
...
+    --provider=skeleton \
+    # tell ginkgo to only run conformance tests
+    --test --test_args="--ginkgo.focus=\[Conformance\]" \
+    # grab the most recent CI tarball of kubernetes 1.10, including the tests
+    --extract=ci/latest-1.10 \
+    # directory to store junit results
+    --dump=$(pwd)/_artifacts | tee ./e2e.log
+```
+
+Note that `--extrace=ci/latest-1.10` indicates that we want to use the binaries/tests on the latest version of 1.10. You can use `--extract=ci/latest` to run the latest set of conformance tests from master.
```
mml: s/extrace/extract

frapposelli:
/lgtm

:+1: from VMware

fedebongio:
/LGTM

thanks Andrew

keps/sig-cloud-provider/0003-testgrid-e2e.md
```diff
@@ -0,0 +1,270 @@
+---
+kep-number: 003
```
cheftako: I believe https://github.com/kubernetes/community/tree/master/keps/sig-cluster-lifecycle is already using 0003.

keps/sig-cloud-provider/0003-testgrid-e2e.md
```diff
@@ -0,0 +1,270 @@
...
+The primary motivation behind collecting conformance test results from various cloud providers on a regular basis is to inform sig-release of any critical bugs. It's important to collect results from various cloud providers to increase coverage and inform other cloud providers of bugs they may be impacted by.
+
+### Goals
+
+* All SIGs / Subprojects owners representing a cloud provider is aware of the importance of frequently uploading conformance test results to testgrid
+* There is a clear and detailed process in place for any cloud provider to upload conformance test results to Testgrid.
+
+### Non-Goals
+
+* Test coverage - increasing test coverage is outside the scope of this KEP
```
cheftako: While I don't think we want to address general test coverage issues, I wonder if we should either address the number of if not "gce" skip test or at minimum mention that conformance tests should be covered. It seems like if a sig is going to own getting rid of cloud-provider skip tests it should be sig-cloud-provider. Whats more any test which is skipped according to cloud provider is probably an important test to run. The functionality under test would seem to be related to something about cloud-provider.

cheftako: @mml suggested that we might at least want to track how many tests were skipped based on cloud provider in testgrid.

keps/sig-cloud-provider/0003-testgrid-e2e.md
justaugustus: There's open discussion on KEP numbering here: https://github.com/kubernetes/community/issues/2245

justaugustus:
/cc @shubheksha @khenidak @colemickens @feiskyer @andyzhangx 
(for Azure review)

karataliu:
@justaugustus I'm following this

keps/sig-cloud-provider/0003-testgrid-e2e.md
andrewsykim: Will revisit file name and KEP number once that's resolved :)

keps/sig-cloud-provider/0003-testgrid-e2e.md
andrewsykim: This is not the final version of the proposal so we can refine some of these details in the future when we have more context.

keps/sig-cloud-provider/0003-testgrid-e2e.md
```diff
@@ -0,0 +1,270 @@
...
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+
+## Summary
+
+This is a KEP outlining the motivation behind why cloud providers should frequently upload E2E conformance test results to [Testgrid](https://github.com/kubernetes/test-infra/tree/master/testgrid) and how a cloud provider can go about doing this.
```
karataliu: 'frequently' implies high infra cost, cloud provider test vendors shall have control over the frequency.

'periodically' could be an alternate word.

keps/sig-cloud-provider/0003-testgrid-e2e.md
```diff
@@ -0,0 +1,270 @@
...
+* [Testgrid Configuration](https://github.com/kubernetes/test-infra/tree/master/testgrid#testgrid)
+* [Display Conformance Tests with Testgrid](https://github.com/kubernetes/test-infra/blob/master/testgrid/conformance/README.md)
+
+#### How to Run E2E Conformance Tests
+
+This KEP outlines two ways of running conformance tests, the first using [Sonobuoy](https://github.com/heptio/sonobuoy) and the second using [kubetest](https://github.com/kubernetes/test-infra/tree/master/kubetest). Though Sonobuoy is easier to setup, it does not guarantee that you will run the latest set of conformance tests. Kubetest though requiring a bit more work to setup, can ensure that you are running the latest set of conformance tests.
+
+At this point we will assume that you have a running cluster and that `kubectl` is configured to point to that cluster with admin access.
+
+#### Sonobuoy
```
karataliu: We can just put a link for those testing tools, since the step details are subject to change.

If we're to go with latest master change (of conformance test) , shall it mention 'Sonobuoy' for now? It is not used for running against latest master change, thus it'll be confusing here.

andrewsykim: Maybe I'm missing something, "latest master change" to me implies we should use kubetest since Sonobuoy does not always let us run the latest set of tests in master as you mentioned.

andrewsykim: Regardless, I think it's worth mentioning both here as long as we highlight the differences (see paragraph above). Running conformance tests on the latest master is going to be preferred going forward but not strictly a requirement _yet_.

keps/sig-cloud-provider/0003-testgrid-e2e.md
BenTheElder: yes, definitely. frequently is great, but... any results at all can be useful. `periodically` SGTM :-)

keps/sig-cloud-provider/0003-testgrid-e2e.md
karataliu: What I'm thinking about is sometimes it might not makes sense to run against fixed e2e version and upload periodically. 
For example:
Since conformance test is considered as important, some provider would run conformance test (say e2e version-1) as pre-submit check.
Later, the periodical run also validates 'cloud-provider master' against 'e2e version-1', then for sure the uploaded report would be all pass. The report will be less useful if it is intrinsically always pass. (suppose it is not used for identifying flaky tests)

If the periodical run is against (e2e version-master), then some conformance test change in master would reveal and may show up as failure in report. Was that the original intention?

BenTheElder: > Later, the periodical run also validates 'cloud-provider master' against 'e2e version-1', then for sure the uploaded report would be all pass. The report will be less useful if it is intrinsically always pass. (suppose it is not used for identifying flaky tests)

There's no reason to assume the cloud provider AND tests are not flaky. The tests are not supposed to be, but they have occasion been so.

> If the periodical run is against (e2e version-master), then some conformance test change in master would reveal and may show up as failure in report. Was that the original intention?

I'm not sure if this was @andrewsykim's intention here, but we are leveraging something like this now and it has caught regressions. Validating conformance tests / core components from kubernetes master across providers helps catch regresssions.

For GCE we're running a matrix of these periodically with kubernetes + e2es built from the release branches at HEAD, last tagged release version within each supported minor release, and against master at HEAD. OpenStack also has a smaller matrix, and Gardener.
Providers will obviously have to choose what kind of matrix coverage they want, but getting the ongoing development particularly on master tested widely is definitely helpful for the project.

andrewsykim: > If the periodical run is against (e2e version-master), then some conformance test change in master would reveal and may show up as failure in report. Was that the original intention?

Yes, this was the original intention. In the future as @BenTheElder had mentioned we may want something more elaborate but this seems like a good starting point.

keps/sig-cloud-provider/0003-testgrid-conformance-e2e.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Implementation Details/Notes/Constraints
+
+Before continuing, it is highly recommended you read the following documentation provided by sig-testing:
+* [Testgrid Configuration](https://github.com/kubernetes/test-infra/tree/master/testgrid#testgrid)
+* [Display Conformance Tests with Testgrid](https://github.com/kubernetes/test-infra/blob/master/testgrid/conformance/README.md)
+
+#### How to Run E2E Conformance Tests
+
+This KEP outlines two ways of running conformance tests, the first using [Sonobuoy](https://github.com/heptio/sonobuoy) and the second using [kubetest](https://github.com/kubernetes/test-infra/tree/master/kubetest). Though Sonobuoy is easier to setup, it does not guarantee that you will run the latest set of conformance tests. Kubetest though requiring a bit more work to setup, can ensure that you are running the latest set of conformance tests.
```
nckturner: I believe the set of tests included in the default e2e run skips quite a few tests that might be relevant for cloud provider to consider their cloud controller manager healthy.  You might want to mention that the cloud provider should spend some effort determining which existing tests are relevant and whether or not more tests need to be added.

keps/sig-cloud-provider/0003-testgrid-conformance-e2e.md
```diff
@@ -0,0 +1,270 @@
...
+
+We would like to propose that every Kubernetes cloud provider reports conformance test results for every patch version of Kubernetes at the minimum. Running conformance tests against master and on pre-release versions are highly encouraged but will not be a requirement.
+
+### Implementation Details/Notes/Constraints
+
+Before continuing, it is highly recommended you read the following documentation provided by sig-testing:
+* [Testgrid Configuration](https://github.com/kubernetes/test-infra/tree/master/testgrid#testgrid)
+* [Display Conformance Tests with Testgrid](https://github.com/kubernetes/test-infra/blob/master/testgrid/conformance/README.md)
+
+#### How to Run E2E Conformance Tests
```
nckturner: I'm assuming the goal of this is to find issues with cloud provider implementation specific stuff.  Do you need a bit more information about the set of binaries that the cloud provider should ensure are under test here (at least CCM, but are CSI impls included?  Anything else?)

keps/sig-cloud-provider/0003-testgrid-conformance-e2e.md
```diff
@@ -0,0 +1,270 @@
...
+
+##### Running Conformance Test with kubetest
+
+Now you can run conformance test with the following:
+```
+cd /path/to/k8s.io/kubernetes
+export KUBERNETES_CONFORMANCE_TEST=y
+kubetest \
+    # conformance tests aren't supposed to be aware of providers
+    --provider=skeleton \
```
nckturner: Is `skeleton` what we want?  Or is my previous comment incorrect re: the goal is testing cloud provider specific implementations?  I should take a closer look at what tests are actually run, but my hunch is we might want a custom set of tests vs conformance for each cloud provider to actually catch issues.

BenTheElder:
If the conformance tests aren't good enough we should probably make them better? That seems orthogonal.

Seperately, providers may also want to run and optionally report other tests suites, kubernetes/kubernetes CI runs many, many, many of these in a matrix.

andrewsykim:
@nckturner The initial goal here is actually to get cloud providers reporting _something_ and the standard set of conformance tests seem like a good start. Once we have the infrastructure in place, incrementally increasing coverage like @BenTheElder had mentioned will be significantly easier.

BenTheElder:
> If we're going to ask CPs to submit conformance test results, I think that we should also authoritatively prescribe which tool we will expect conformance results to come from.

The _results_ come from the output of the `e2e.test` binary / ginkgo. Sonobuoy and kubetest orchestrate this and help copy the files. 

For some use cases specifically https://github.com/cncf/k8s-conformance really ought to be used as it's the official config for submitting results to the CNCF, but for testing more recent builds something else may be necessary.

Similarly there's no requirement to use the `upload_e2e.py` script, just some tool the effectively produces the same results.

justaugustus:
Thanks for clarifying, @BenTheElder!
/lgtm
/hold cancel

keps/sig-cloud-provider/0003-testgrid-conformance-e2e.md
nckturner: Yeah of course, my general question was whether or not conformance test is the correct thing for this KEP to require, and it sounds like for now it should be scoped to conformance tests exclusively.

keps/sig-cloud-provider/0003-testgrid-conformance-e2e.md
nckturner: 👍

nckturner:
@andrewsykim @BenTheElder That all makes sense.  This KEP is scoped to just getting conformance tests run on a cloud provider (in any way possible) and the results reported back.  Out of scope is testing the cloud provider specific stuff, like the CCM binary, or any tests outside of conformance.

And tangentially, as a community, we should improve conformance tests where they fall short when possible.  

Sounds like a great start.

hogepodge:
Since part of the intent is to get -some- form of testing, the how feels less important. Conformance is the first natural and lowest bar to cross on that.

/approve

PR #2230, opened 2018-06-08
system:node-bootstrapper is not a group name

youyongsong:
`system:node-bootstrapper` is a ClusterRole not a group.

### DIFF
--- a/keps/sig-cluster-lifecycle/draft-20180130-kubeadm-join-master.md
+++ b/keps/sig-cluster-lifecycle/draft-20180130-kubeadm-join-master.md
@@ -280,7 +280,7 @@ existing `kubeadm join` flow:
       in `kube-system` namespace.
 
       > This requires to grant access to the above configMap for
-      `system:node-bootstrapper` group (or to provide the same information
+      `system:bootstrappers` group (or to provide the same information
       provided in a file like in 1.).
 
    2. Check if the cluster is ready for joining a new master node:


keps/sig-cluster-lifecycle/draft-20180130-kubeadm-join-master.md
```diff
@@ -280,7 +280,7 @@ existing `kubeadm join` flow:
       in `kube-system` namespace.
 
       > This requires to grant access to the above configMap for
-      `system:node-bootstrapper` group (or to provide the same information
+      `system:node-bootstrapper` ClusterRole (or to provide the same information
```
mikedanese: You can't grant a role permission and you shouldn't modify system roles. I think this sentence has to change a little bit more.

youyongsong: What about change `system:node-bootstraper` to `system:bootstrappers`? `system:bootstrappers` is the group of `system:bootstrap:<token-id>` user.

mikedanese: That works.

youyongsong: ok, I changed that in a new commit.

youyongsong:
@roberthbailey thanks for reminding, I have squashed the two commits.

PR #2235, opened 2018-06-08
Proto is our default storage serialization; we can't break it

lavalamp:


lavalamp:
/assign @smarterclayton 
/assign @liggitt 
cc @bgrant0607

lavalamp:
/assign @grodrigues3 

Can you rubber stamp this please?

Should we move this file into an api machinery specific location?

cblecker:
/approve

Up to you. Yes we can totally move this to a more specific folder with an OWNERS file.

PR #2241, opened 2018-06-11
KEP: pull based audit event stream

CaoShuFeng:
This tries to fix: https://github.com/kubernetes/kubernetes/issues/53455

### DIFF
--- a/keps/0013-pull-based-audit-events-stream.md
+++ b/keps/0013-pull-based-audit-events-stream.md
@@ -12,7 +12,7 @@ approvers:
   - TBD
 editor: TBD
 creation-date: 2018-06-11
-last-updated: 2018-06-11
+last-updated: 2018-07-18
 status: provisional
 see-also:
   - n/a
@@ -26,112 +26,79 @@ superseded-by:
 # Pull based audit events stream
 
 ## Table of Contents
-
 * [Pull based audit events stream](#pull-based-audit-events-stream)
-* [Summary](#summary)
-* [Motivation](#motivation)
-    * [Goals](#goals)
-    * [Non-Goals](#non-goals)
-* [Proposal](#proposal)
-    * [Make audit backends dynamicly registerable](#make-audit-backends-dynamicly-registerable)
-    * [endpoints and query parameters](#endpoints-and-query-parameters)
-    * [Bandwidth limit Configuration](#bandwidth-limit-configuration)
-    * [User Stories](#user-stories)
+   * [Table of Contents](#table-of-contents)
+   * [Summary](#summary)
+   * [Motivation](#motivation)
+      * [Goals](#goals)
+      * [Non-Goals](#non-goals)
+   * [User Stories](#user-stories)
       * [Story 1](#story-1)
       * [Story 2](#story-2)
-    * [Risks and Mitigations](#risks-and-mitigations)
-* [Graduation Criteria](#graduation-criteria)
-* [Implementation History](#implementation-history)
-* [Drawbacks [optional]](#drawbacks-optional)
-* [Alternatives [optional]](#alternatives-optional)
-
-[Tools for generating]: https://github.com/ekalinin/github-markdown-toc
+   * [Proposal](#proposal)
+      * [new resource endpoints](#new-resource-endpoints)
+      * [Field Selector for subset of audit events](#field-selector-for-subset-of-audit-events)
+      * [Bandwidth limit Configuration](#bandwidth-limit-configuration)
+      * [Risks and Mitigations](#risks-and-mitigations)
+   * [Graduation Criteria](#graduation-criteria)
+   * [Implementation History](#implementation-history)
+   * [Alternatives](#alternatives)
 
 ## Summary
 
-This proposal aims at providing a pull-based stream for audit logging. This would look similar to a watch request to the API server, and provide stream of audit events. The stream does not return any previous events (i.e. no associated storage), it would only stream events that occurred after the stream was opened. Two non resource-url endpoints are installed to provide the stream. '/audits/' for all audit events generated in the API server, and '/audits/{namespace}' for audits events in the specified namespace. And query parameters could be used to get subset of audit events.
+This proposal aims at providing a pull-based stream for audit logging. This would look similar to a watch request to the API server, and provide stream of audit events. The stream does not return any previous events (i.e. no associated storage), it would only stream events that occurred after the stream was opened. Two new resource endpoints is installed to provide the stream: '/apis/audit.k8s.io/v1beta1/events' and '/apis/audit.k8s.io/v1alpha1/namespaces/{namespace}/namespacedevents'. These endpoints would only support the WATCH verb in the first.
 
 ## Motivation
-The current audit backends (log & webhook) require very high privileges (file access to the master) or prior configuration (webhook). There are a number of use cases for dynamic and unprivileged access to the audit logs. Examples include:
-- allow tenants to get audit events in their own namespace
+This endpoint provides a good way for short-term debugging. Both tenants and cluster administrator could use this endpoint to get event stream without any changes to the cluster. Examples include:
+- allow tenants to get audit events in their own namespaces
 - provide a subset of audit events, so it will be easier to query and analyze (e.g. https://github.com/kubernetes/kubernetes/issues/56683)
 
 
 ### Goals
 - Provide a new pull based audit event stream, which returns events that occurred after the stream was opened.
 - Allow users go get a subset of audit events.
-- Access control. Allow unprivileged users to get audits in their own namespace.
+- Access control. Allow cluster administrator to get all events in cluster, allow namespace administrators to get all events in their own namespace.
 - Bandwidth limit in server scope or per-namespace.
 
 ### Non-Goals
-- Access control at user/group level(e.g. allow user A to only get audits events from users A,B,C)
+- Access control beyond namespace granularity(e.g. allow user A to only get audits events from users A,B,C or allow user A to only get audit events of resource foo)
+- Provide a custom per-stream policies. The audit stream should be based on the cluster-wide audit policy.
 
+## User Stories
+
+### Story 1
+As a tenant, I will be able to known all things happened to my own namespace. When problem happens, I can use the audit events for trouble shooting.
+
+### Story 2
+As a developer, I will easily be able to get a subset of audit events. This will save a lot of time for debugging. I don't need to grep audit events from a couple of audit files.
 
 ## Proposal
 
-### Make audit backends dynamicly registerable
+### new resource endpoints
+Two new resource endpoint '/apis/audit.k8s.io/v1beta1/events' and '/apis/audit.k8s.io/v1alpha1/namespaces/{namespace}/namespacedevents' are installed.
+Unlike other resource endpoints which usually serve resource objects from key-value storage, these endpoints read events from a audit backend and only support the WATCH verb. '/apis/audit.k8s.io/v1beta1/events' provides a stream of all audit events in the cluster. '/apis/audit.k8s.io/v1alpha1/namespaces/{namespace}/namespacedevents' provides a stream of all audit events in the namespace. A new API object NamespacedEvent is introduced for the namespaced endpoint. NamespacedEvent has the same definition with the audit Event object which we already have, but `metadata.namespace` element is required for it.
 
-```golang
-// Registry is a decorator for the audit backend, it allows caller to dynamicly register/unregister new backends to it.
-type Registry interface {
-       Backend
+The audit events returned to users are the events generated according to the audit policy file. If an event is omitted by the audit policy file, the event would not be sent to user by these endpoints. For aggregated API servers, those endpoints only supports metadata-level, because kube-apiserver only record these events at metadata level.
 
-       // Register register a new Backend to the registry.
-       Register(Backend)
+To implement these watch endpoints, a new audit backend is introduced. This new backend will provide [watch.Interface](https://github.com/kubernetes/kubernetes/blob/release-1.11/staging/src/k8s.io/apimachinery/pkg/watch/watch.go#L28) to apimachinery. And apimachinery use this watch.Interface to implement the watch endpoint.
 
-       // UnRegister remove the Backend from registry.
-       UnRegister(Backend)
-}
-```
-When a request comes, a new backend is registered to the registry. Audit Events sent to this backend will be sent to users in the end.
-
-### endpoints and query parameters
-
-|endpoint            |definition                                  |supported paramsters                                 |
-|--------------------|--------------------------------------------|-----------------------------------------------------|
-|/audits             |all audit events generated in the API server|username, group, namespace, apiGroup, resource, verbs|
-|/audits/{namespaces}|audits events in the specified namespace    |username, group, apiGroup, resource, verbs           |
-
-definition of query parameters
-```golang
-type queryParm struct {
-       // username in the audit event.
-       // optional
-       username string
-       // groups in the audit event, if all specified groups exist in the audit event, then it will be sent to user.
-       // optional
-       groups []string
-       // namespace of the audit event, <none> for cluster scoped request, empty string for all namespaces.
-       // only supported for /audits endpoint.
-       // optional
-       namespace string
-       // apiGroup in the audit event, <core> for the core api group.
-       // optional
-       apiGroup string
-       // resource in the audit event, for example: pods.
-       // optional
-       resource string
-       // verb in the audit event, for example: create, put. If the verbs contains the verb of audit event, then
-       // the audit event will be sent to user.
-       // optional
-       verbs []string
-}
-```
+### Field Selector for subset of audit events
+Some field selector would be supported for users to get a subset of audit events. (e.g. objectRef.namespace, user.username)
 
 ### Bandwidth limit Configuration
+
 ```golang
 // LimitType is the type of the limit.
 type LimitType string
 
 const (
 	// ServerLimitType is a type of limit where there is one bucket shared by
-	// all of the audit events sent by the API Server, including '/audits' and
-	// 'audits/{namespace}' endpoints.
+	// all of the audit events sent by the API Server, including cluster-scoped and
+	// namespaced endpoints.
 	ServerLimitType LimitType = "server"
 
 	// NamespaceLimitType is a type of limit where there is one bucket used by
-	// each namespace. When the namespace is set to empty string, it limit the
-	// bandwidth of all requests to '/audits' endpoint.
+	// each namespace.
 	NamespaceLimitType LimitType = "namespace"
 )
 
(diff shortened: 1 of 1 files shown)


keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Drawbacks [optional]](#drawbacks-optional)
+* [Alternatives [optional]](#alternatives-optional)
+
+[Tools for generating]: https://github.com/ekalinin/github-markdown-toc
+
+## Summary
+
+This proposal aims at providing a pull-based stream for audit logging. This would look similar to a watch request to the API server, and provide stream of audit events. The stream does not return any previous events (i.e. no associated storage), it would only stream events that occurred after the stream was opened. Two non resource-url endpoints are installed to provide the stream. '/audits/' for all audit events generated in the API server, and '/audits/{namespace}' for audits events in the specified namespace. And query parameters could be used to get subset of audit events.
```
tallclair: > Two non resource-url endpoints

I think these should be resource endpoints, even though they're "pseudo resources"

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Motivation
+The current audit backends (log & webhook) require very high privileges (file access to the master) or prior configuration (webhook). There are a number of use cases for dynamic and unprivileged access to the audit logs. Examples include:
```
tallclair: Most of these are addressed by [dynamic audit configuration](https://github.com/kubernetes/community/pull/2188). I think it's worth highlighting the different use cases for each here. In particular, I see the streaming endpoint as useful for short-term debugging usecases, e.g. where I run a client-side tool to pull events for a short period, and the push model is better for long-running usecases, but requires more setup.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+
+
+### Goals
+- Provide a new pull based audit event stream, which returns events that occurred after the stream was opened.
+- Allow users go get a subset of audit events.
+- Access control. Allow unprivileged users to get audits in their own namespace.
+- Bandwidth limit in server scope or per-namespace.
+
+### Non-Goals
+- Access control at user/group level(e.g. allow user A to only get audits events from users A,B,C)
```
tallclair: I would generalize this to include any granularity beyond namespaces (including users, resources, etc.)

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+- provide a subset of audit events, so it will be easier to query and analyze (e.g. https://github.com/kubernetes/kubernetes/issues/56683)
+
+
+### Goals
+- Provide a new pull based audit event stream, which returns events that occurred after the stream was opened.
+- Allow users go get a subset of audit events.
+- Access control. Allow unprivileged users to get audits in their own namespace.
+- Bandwidth limit in server scope or per-namespace.
+
+### Non-Goals
```
tallclair: Also custom per-stream policies should be out of scope (at least for the initial proposal). The audit stream should be based on the cluster-wide audit policy.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+- Access control at user/group level(e.g. allow user A to only get audits events from users A,B,C)
+
+
+## Proposal
+
+### Make audit backends dynamicly registerable
```
tallclair: This seems out-of-scope for this proposal. https://github.com/kubernetes/community/pull/2188 covers this case, and isn't required for a streaming backend.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+}
+```
+When a request comes, a new backend is registered to the registry. Audit Events sent to this backend will be sent to users in the end.
+
+### endpoints and query parameters
+
+|endpoint            |definition                                  |supported paramsters                                 |
+|--------------------|--------------------------------------------|-----------------------------------------------------|
+|/audits             |all audit events generated in the API server|username, group, namespace, apiGroup, resource, verbs|
+|/audits/{namespaces}|audits events in the specified namespace    |username, group, apiGroup, resource, verbs           |
```
tallclair: These should be pseudo-resources. Something like:
```
/apis/audit.k8s.io/v1beta1/events
/apis/audit.k8s.io/v1beta1/namespaces/{namespace}/events
```
(I'm not sure how it works having both a cluster-scoped & namespace scoped version, but this is the basic idea)

tallclair: Initially, these should only support `watch`. But I could imagine expanding the scope of verbs someday.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+
+definition of query parameters
+```golang
+type queryParm struct {
+       // username in the audit event.
+       // optional
+       username string
+       // groups in the audit event, if all specified groups exist in the audit event, then it will be sent to user.
+       // optional
+       groups []string
```
tallclair: I thought you said username & groups were out of scope under "non goals"?

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+When a request comes, a new backend is registered to the registry. Audit Events sent to this backend will be sent to users in the end.
+
+### endpoints and query parameters
+
+|endpoint            |definition                                  |supported paramsters                                 |
+|--------------------|--------------------------------------------|-----------------------------------------------------|
+|/audits             |all audit events generated in the API server|username, group, namespace, apiGroup, resource, verbs|
+|/audits/{namespaces}|audits events in the specified namespace    |username, group, apiGroup, resource, verbs           |
+
+definition of query parameters
```
tallclair: I don't think we should use any of these, at least not initially.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+	// before emitting to client.
+	// optional
+	MaxEventSize int64 `json:"maxEventSize"`
+}
+```
+
+When the sent bytes exceeds the bandwidth, API server will try to truncate the audit event, and check the limit again. If bandwidth limit is still exceeded, the event would be droped.
+
+
+### User Stories
```
tallclair: nit: move user-stories to before the proposal.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+
+#### Story 1
+As a tenant, I will be able to known all things happened to my own namespace. When problem happens, I can use the audit events for trouble shooting.
+
+#### Story 2
+As a developer, I will easily be able to get a subset of audit events. This will save a lot of time for debugging. I don't need to grep audit events from a couple of audit files.
+
+### Risks and Mitigations
```
tallclair: Would be good to talk about SLO. In particular, I think these endpoints should be treated as debugging & development endpoints, not for actual secure audit logging.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+- Provide a new pull based audit event stream, which returns events that occurred after the stream was opened.
+- Allow users go get a subset of audit events.
+- Access control. Allow unprivileged users to get audits in their own namespace.
+- Bandwidth limit in server scope or per-namespace.
+
+### Non-Goals
+- Access control at user/group level(e.g. allow user A to only get audits events from users A,B,C)
+
+
+## Proposal
```
tallclair: Can you add more implementation details? In particular, I'd like to see:

- Is this a new [audit backend](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apiserver/plugin/pkg/audit)?
- How are events routed to the watchers?
- How are the watch connections maintained? (Does apimachinery solve this?)

keps/0013-pull-based-audit-events-stream.md
CaoShuFeng: I will investigate this way.
Thanks.

keps/0013-pull-based-audit-events-stream.md
sttts: Can RBAC rules express this?

keps/0013-pull-based-audit-events-stream.md
sttts: What about the aggregated API server case?
- will the kube-aggregator aggregate audit events?

keps/0013-pull-based-audit-events-stream.md
sttts: At least discovery will not work with cluster and namespace scoped resources, either or. So we need at least two resources. 

In theory we could serve `/apis/audit.k8s.io/v1beta1/events`, even though events are namespaced. This would mean though that each event has a namespace set, even those for cluster scoped resources.

For the other search criteria like "user" we could use labels, i.e. a label selector.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,199 @@
...
+## Motivation
+The current audit backends (log & webhook) require very high privileges (file access to the master) or prior configuration (webhook). There are a number of use cases for dynamic and unprivileged access to the audit logs. Examples include:
+- allow tenants to get audit events in their own namespace
+- provide a subset of audit events, so it will be easier to query and analyze (e.g. https://github.com/kubernetes/kubernetes/issues/56683)
+
+
+### Goals
+- Provide a new pull based audit event stream, which returns events that occurred after the stream was opened.
+- Allow users go get a subset of audit events.
+- Access control. Allow unprivileged users to get audits in their own namespace.
```
sttts: How do we make sure that the audit events do not leak information the user cannot access via "normal" resource queries otherwise (i.e. which are restricted via RBAC) ?

CaoShuFeng: We can have such level restriction:
Allow A to read audit events in namespace-1.

We can't have this:
Allow A to read audit events `about pods,servers` in namespace-1.

CaoShuFeng: Since we treat audit events as `pseudo-resource`. I think we can restrict the access control of audit events as other normal resources.

But we don't support `resourceNames`, since audit events don't have a name.

keps/0013-pull-based-audit-events-stream.md
sttts: If we turn this into a resource endpoints (as @tallclair suggests above), it will be part of discovery. And in theory the aggregator could pick it up. But those api groups will conflict. So the same resource can only be used for one API server at a time. So this topic needs some more thought, to be at least extensible in the future.

keps/0013-pull-based-audit-events-stream.md
sttts: They could have a pseudo-name.

I was wondering whether we really want to invent another authz system for audit events or whether we can piggy pack on RBAC rules. I.e. when querying (via /audit or a real resource URL) I see all events of objects I could also read via normal APIs. Or we add another verb to RBAC called "AUDIT". Just thinking out loud.

/cc @kubernetes/sig-auth-misc @liggitt

keps/0013-pull-based-audit-events-stream.md
sttts: When implementing pseudo resources, there are dragons I heard.

keps/0013-pull-based-audit-events-stream.md
CaoShuFeng: Will remove it.

keps/0013-pull-based-audit-events-stream.md
CaoShuFeng: Will add it.

keps/0013-pull-based-audit-events-stream.md
CaoShuFeng: Will remove it.

keps/0013-pull-based-audit-events-stream.md
sttts: Doubling the load for each request.

keps/0013-pull-based-audit-events-stream.md
sttts: As written above, storing to disk sounds like doubling the etcd load, our bottleneck resource in big clusters.

keps/0013-pull-based-audit-events-stream.md
CaoShuFeng: > Doubling the load for each request.

:cry:  :cry:  :cry: 
Yes, but I haven't figure out another way to implement this in HA environment.

keps/0013-pull-based-audit-events-stream.md
tallclair: I think for an initial implementation it's acceptable to just authorize it at the audit resource level (i.e. I need read permissions on the `audit/namespaces/foo` resource), and document this as a something to watch out for. Then it's up to the admin / namespace admin to restrict audit permissions to those who should actually be able to read everything in the cluster/namespace.

Also, sensitive information (secrets) should be filtered from audit logs anyway.

tallclair:
@kubernetes/sig-api-machinery-feature-requests

keps/0013-pull-based-audit-events-stream.md
tallclair: I mean this should just be handled in the policy. e.g. https://github.com/kubernetes/kubernetes/blob/56d9d73f38489f09d55ccee442ed77169484b146/cluster/gce/gci/configure-helper.sh#L837-L839. The whole object body is excluded in this case.

lavalamp:
On Tue, Jun 19, 2018, 5:24 PM Jordan Liggitt <notifications@github.com>
wrote:

> *@liggitt* commented on this pull request.
> ------------------------------
>
> In keps/0013-pull-based-audit-events-stream.md
> <https://github.com/kubernetes/community/pull/2241#discussion_r196617894>:
>
> > +[Tools for generating]: https://github.com/ekalinin/github-markdown-toc
> +
> +## Summary
> +
> +This proposal aims at providing a pull-based stream for audit logging. This would look similar to a watch request to the API server, and provide stream of audit events. The stream does not return any previous events (i.e. no associated storage), it would only stream events that occurred after the stream was opened. Two non resource-url endpoints are installed to provide the stream. '/audits/' for all audit events generated in the API server, and '/audits/{namespace}' for audits events in the specified namespace. And query parameters could be used to get subset of audit events.
> +
> +## Motivation
> +The current audit backends (log & webhook) require very high privileges (file access to the master) or prior configuration (webhook). There are a number of use cases for dynamic and unprivileged access to the audit logs. Examples include:
> +- allow tenants to get audit events in their own namespace
> +- provide a subset of audit events, so it will be easier to query and analyze (e.g. https://github.com/kubernetes/kubernetes/issues/56683)
> +
> +
> +### Goals
> +- Provide a new pull based audit event stream, which returns events that occurred after the stream was opened.
> +- Allow users go get a subset of audit events.
> +- Access control. Allow unprivileged users to get audits in their own namespace.
>
> Also, sensitive information (secrets) should be filtered from audit logs
> anyway.
>
> if we add a mechanism for filtering object content for specific resources
> (e.g. not allowing or ignoring an audit policy that tries to include that
> info), let's ensure it doesn't hard-code the Secret resource
>
I think we should mark fields as sensitive in the openAPI spec. Many places
could use that info.

in the internals of the audit handler... each API server should be able to

keps/0013-pull-based-audit-events-stream.md
CaoShuFeng: I did some research about aggregation server. kube-aggregator could also aggregate audit events if the aggregated api-server has the same endpoint exposed.

But if the kube-apiserver is behind a load balancer, audit events in another kube-apiserver could not be returned, because no storage is involved.
Is that OK?

sttts: It is a major constraint, but might be acceptable depending on the environment. It just means that your auditing system has to know and query all HA instances of kube-apiserver.

keps/0013-pull-based-audit-events-stream.md
tallclair: In order to get a first-pass of this feature in, we could simply say that the streaming endpoint only supports metadata-level, and then it can be handled entirely by kube-aggregator. This simplification also addresses the concerns about sensitive fields.

CaoShuFeng: > we could simply say that the streaming endpoint only supports metadata-level,

Test in my environment that could work. Thanks for the idea.

> In that case, we have no guarantees that we collected all events with a pull mechanism because we don't see the instance anymore in such a case.

We only trace audit events in aggregator, and request distributed to aggregated apiservers will be logged at metadata-level.

sttts: > We only trace audit events in aggregator

Misread @tallclair suggestion. Do you have a WIP PR? Sounds reasonably easy to do, if there are no surprises.

tallclair: Hmm, I forgot about HA masters. I think this still doesn't address that. I guess this would need some pubsub/message solution - does anything else in kubernetes solve a similar problem, or does everything just rely on etcd?

liggitt: > just etcd

tallclair: Actually, this seems pretty similar to (non-audit) `Events`.

CaoShuFeng:
@tallclair @sttts 
I updated it, please take a look again.

keps/0013-pull-based-audit-events-stream.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Proposal
+
+### new resource endpoints
+Two new resource endpoint '/apis/audit.k8s.io/v1beta1/events' and '/apis/audit.k8s.io/v1alpha1/namespaces/{namespace}/namespacedevents' are installed.
+Unlike other resource endpoints which usually serve resource objects from key-value storage, these endpoints read events from a audit backend and only support the WATCH verb. '/apis/audit.k8s.io/v1beta1/events' provides a stream of all audit events in the cluster. '/apis/audit.k8s.io/v1alpha1/namespaces/{namespace}/namespacedevents' provides a stream of all audit events in the namespace. A new API object NamespacedEvent is introduced for the namespaced endpoint. NamespacedEvent has the same definition with the audit Event object which we already have, but `metadata.namespace` element is required for it.
+
+The audit events returned to users are the events generated according to the audit policy file. If an event is omitted by the audit policy file, the event would not be sent to user by these endpoints. For aggregated API servers, those endpoints only supports metadata-level, because kube-apiserver only record these events at metadata level.
```
liggitt: can you outline the data flow of audit information for a set of HA api servers, and how audit information from the various servers makes it to this handler, and then to the user?

CaoShuFeng: Sorry, for a HA environment, users have to know and query all HA instances of kube-apiserver.
I think this would work like a [headless services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services).

tallclair: That doesn't work if the HA instances are behind a loadbalancer. I think it should be OK to document this as a known issue, and move forward with the current proposal to alpha. I do think HA support should be a blocker before going to beta. @liggitt @sttts WDYT?

tallclair: As for solutions, I think either the audit events need to be put into a etcd circular buffer, or the apiserver that ends up serving the request makes proxy requests to the other apiservers. That assumes apiserver <-> apiserver connectivity though, which might not be viable.

liggitt: I think we should at least have a sketch of how this would work both with multiple instances of the same apiserver and with multiple distinct servers for different API groups to decide if it seems feasible before proceeding.

liggitt: For example, would satisfying this API require all apiservers to be forwarding their audit events to a collating sink, would the filters/verbosity of that forwarding be determined by active watchers on the audit pull API, etc

sttts: Having a pull mechanism with a collating sink sounds odd. Actually, if one really wants that, why not use the existing webhook push mechanism to send events to the sink and then pull from there?

In other words, I think having to pull all apiservers "manually" (and knowing about the server topology) is a fair compromise with a pull mechamism in HA setups.

CaoShuFeng: > why not use the existing webhook push mechanism to send events to the sink and then pull from there?


This could be an [kubernetes aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/)

tallclair: Maybe we should try prototyping this out-of-tree using the push-based dynamic webhook?

CaoShuFeng: >   prototyping this out-of-tree

This is what Aggregated API Server do, I think.

justaugustus:

justaugustus:

/close

PR #2244, opened 2018-06-12
update SIG scheduling meeting times and info

bsalamat:


cblecker:
Per the notice at the top of these files, they are auto generated, and the source of truth is the sigs.yaml file. Instructions for using the generator are here: https://github.com/kubernetes/community/blob/master/generator/README.md

bsalamat:
@cblecker I followed the same instructions. Is there anything that I did incorrectly?

cblecker:
This works!
/approve

PR #2249, opened 2018-06-12
Reserving KEP # 13 for 'IPv4/IPv6 Dual Stack' specification.

leblancd:


idvoretskyi:
/uncc

PTAL @calebamiles @jbeda @bgrant0607

PR #2259, opened 2018-06-13
Remove an unexpected colon

whs-dot-hk:
For Hugo

### DIFF
--- a/sig-autoscaling/README.md
+++ b/sig-autoscaling/README.md
@@ -8,7 +8,7 @@ To understand how this file is generated, see https://git.k8s.io/community/gener
 -->
 # Autoscaling Special Interest Group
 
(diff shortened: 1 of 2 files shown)


cblecker:
@whs-dot-hk Why is this change necessary? Need a bit more background than "For Hugo"

castrojo:
This is to allow hugo to build the community repo markdown pages, [KEP-005](https://github.com/kubernetes/community/blob/master/keps/sig-contributor-experience/0005-contributor-site.md).

cblecker:
But what is the issue with it? Are colons aren't allowed as a part of the mission statement? This is valid YAML, and Go Templates consume it properly.

whs-dot-hk:
Colons are allowed the html generated fine, but when looking at the result, you will spot the colon immediately, it looks unnatural to me in html

cblecker:
@whs-dot-hk Okay, so this is purely cosmetic?

If so, then I'm okay with this change. You'll need to rebase, make the change, and then regenerate the docs (directions here: https://git.k8s.io/community/generator/README.md)

PR #2261, opened 2018-06-13
api_changes: remove proto tags from examples

### LINKED ISSUE #1159: api_changes.md: add protobuf definitions to the Frobber struct
Most of the [examples with the `Frobber` struct](https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#on-compatibility) has only JSON definitions while some of them also have a protobuf definitions. In order to make the doc consistent, all the examples should be written in am uniform way.

nikhita:
Fixes #1159 
See https://github.com/kubernetes/community/issues/1159#issuecomment-397008148

Some examples have proto tags and some don't. Since proto tags are generated and not manually written, this PR removes the proto tags from the examples to remain consistent.

### DIFF
--- a/contributors/devel/api_changes.md
+++ b/contributors/devel/api_changes.md
@@ -791,9 +791,9 @@ For example, consider the following object:
 // API v6.
 type Frobber struct {
   // height ...
-  Height *int32 `json:"height" protobuf:"varint,1,opt,name=height"`
+  Height *int32 `json:"height"
   // param ...
-  Param  string `json:"param" protobuf:"bytes,2,opt,name=param"`
+  Param  string `json:"param"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sttts:
> Since proto tags are generated and not manually written, this PR removes the proto tags from the examples to remain consistent.

They are also manually written.

nikhita:
> They are also manually written.

Interesting! @sttts do you have an example for where this is done?

sttts:
> nteresting! @sttts do you have an example for where they are manually written?

The tags are all manually written (or I missed a tool to do that the last 2 years :-D). Or maybe we talk about a different thing? The tags are manual, the codecs and proto files are genarated.

nikhita:
> Or maybe we talk about a different thing? The tags are manual, the codecs and proto files are genarated.

I meant the proto tags here:

https://github.com/kubernetes/kubernetes/blob/ed95abe72fb914e20052b4480be1f5e42c680ca1/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1/types_jsonschema.go#L21


> The tags are all manually written (or I missed a tool to do that the last 2 years :-D).

`hack/update-generated-protobuf.sh` (https://github.com/kubernetes/kubernetes/blob/master/hack/update-generated-protobuf.sh) does the tag generation as well! I definitely did not write all the proto tags for the `JSONSchemaProps` struct above. :)

Also double checked by removing some tags randomly from `types.go` above and running the script. It generated the proto tags (albiet updating the tags wrongly because removing tags randomly is not a good idea :))

sttts:
> Also double checked by removing some tags randomly from types.go above and running the script. It generated the proto tags (albiet updating the tags wrongly because removing tags randomly is not a good idea :))

wow. I never noticed 🤔

nikhita:
@sttts lgty?

nikhita:
/assign @cblecker 
for approval

PR #2262, opened 2018-06-13
Add link to charter from sig readme

### LINKED ISSUE #2243: There should be a field in sigs.yaml for location of charter
Currently the charter link is sprinkled into mission statement in many cases, while there is no first-class field for the location of a charter. We should add a first-class field for this information.

cblecker:
fixes #2243.

Adds first class field to sig readme for link to charter

### DIFF
--- a/sigs.yaml
+++ b/sigs.yaml
@@ -6,6 +6,7 @@ sigs:
       API CRUD semantics, admission control, encoding/decoding, conversion,
       defaulting, persistence layer (etcd), OpenAPI, CustomResourceDefinition,
       garbage collection, and client libraries.
+    charter_link:
     label: api-machinery
     leadership:
       chairs:
@@ -116,6 +117,7 @@ sigs:
       We discuss how to define and run apps in Kubernetes, demo relevant tools
       and projects, and discuss areas of friction that can lead to suggesting
       improvements or feature requests.
+    charter_link:
     label: apps
     leadership:
       chairs:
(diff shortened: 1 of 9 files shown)


sigs.yaml
```diff
@@ -229,6 +231,7 @@ sigs:
     mission_statement: >
       The Architecture SIG maintains and evolves the design principles of Kubernetes, and provides a consistent
       body of expertise necessary to ensure architectural consistency over time.
+    charter_link: chater.md
```
nikhita: s/chater/charter

sigs.yaml
```diff
@@ -1906,6 +1937,7 @@ workinggroups:
       mailing_list: https://groups.google.com/forum/#!forum/kubernetes-wg-apply
   - name: Cloud Provider
     dir: wg-cloud-provider
+    charter_link:
```
nikhita: are working groups supposed to have a charter? https://github.com/kubernetes/community/blob/master/committee-steering/governance/README.md#sig-charter-guide says that only SIGs need to have one.

nikhita: if WGs are also supposed to have a charter, can we add the `charter_link` field for all WGs for consistency? this only adds for wg-cloud-provider.

- wg-apply has a comment which says "we might want to add a charter here".
- wg-app-def has a gdoc as their charter in the mission statement.

cblecker:
@carolynvs @nikhita Updated. PTAL.

nikhita:
We should also update this doc https://github.com/kubernetes/community/blob/master/committee-steering/governance/README.md to say that even WGs have to have a charter (but that can be a separate PR).

/lgtm

nikhita:
/assign @idvoretskyi 
for approval

cblecker:
I don't strictly know that they *have* to have one yet, but I don't mind providing a linking structure for those that do.

/approve

Issue #15, opened 2016-06-29
Community office hours should be added to calendar


rothgar:
I randomly found The [community office hours](https://www.youtube.com/watch?v=nKHYnvTr7LE) video from [@sarahnovotny's retweet](https://twitter.com/surajd_/status/748091857071026176)

They should be added to the [shared calendar](https://calendar.google.com/calendar/embed?src=cgnt364vd8s86hr2phapfjc6uk%40group.calendar.google.com&ctz=America/Los_Angeles) so people can get involved, ask questions, etc.

sarahnovotny:
aaah!  i'd added them to the community events, not community meetings calendar.  I'll be sure to add them to both next time. thanks for the reminder.

PR #2270, opened 2018-06-15
Reservation of KEP ID #15 for "SCTP support in Kubernetes"

janosi:
I would like to reserve KEP ID #15 for "SCTP support in Kubernetes"
I jumped from 14 to 16 in the NEXT_KEP_NUMBER file, as there is another pending PR to reserve KEP_ID 14 (and that PR increased the next ID value to 15 already).


janosi:
@calebamiles Maybe the KEP process needs an update? At least, when I started this one was in force. Here the 2nd step is the reservation of a KEP number
https://github.com/kubernetes/community/blob/master/keps/0000-kep-template.md

PR #2273, opened 2018-06-15
CSI: support for in-line volumes in pods.

jsafrane:
Current CSI implementation supports only PersistentVolumes. This proposal adds support for in-line CSI volumes in pods.

/sig storage

/assign @davidz627 @saad-ali @vladimirvivien @liggitt


contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,119 @@
...
+// Exactly one member can be set.
+type VolumeAttachmentSource struct {
+	// Name of the persistent volume to attach.
+	// +optional
+	PersistentVolumeName *string
+
+	// VolumeSource represents the source location of a volume to attach.
+	// Only CSIVolumeSource can be specified.
+	// +optional
+    VolumeSource *v1.VolumeSource
```
liggitt: why is this data denormalized instead of referencing the pod?

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,119 @@
...
+}
+```
+
+* A/D controller **copies whole `VolumeSource`**  from `Pod` into `VolumeAttachment`. This allows external CSI attacher to detach volumes for deleted pods without keeping any internal database of attached VolumeSources.
```
liggitt: * what will create VolumeAttachment objects corresponding to pods?
* what will clean up VolumeAttachment objects after pods are deleted?
* how will the VolumeAttachment objects be named? is it derived from the pod namespace and name?
* what is the behavior if two pods both contain identical VolumeSources?

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,119 @@
...
+* Using whole `VolumeSource` allows us to re-use `VolumeAttachment` for any other in-line volume in the future. We provide validation that this `VolumeSource` contains only `CSIVolumeSource` to clearly state that only CSI is supported now.
+	* TBD: `CSIVolumeSource` would be enough...
+* External CSI attacher must be extended to  process either `PersistentVolumeName` or `VolumeSource`.
+* Since in-line volume in a pod can refer to a secret in the same namespace as the pod, **external attacher must get permissions to read any Secrets in any namespace**.
```
liggitt: that determination should be made by the deployer of the attacher. it may not be expected to service all namespaces, and may not require user-specified attach secrets

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: Because the pod can be deleted when external A/D controller needs to detach a volume. Current CSI implementation expects that PV is not deleted (and uses finalizes for that). I think it would be too much to add finalizers to  pods.

liggitt: what indicates whether an inline volume requires an attach/detach phase? many potential uses of inline CSI volumes only require mount/unmount

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: > what will create VolumeAttachment objects corresponding to pods?
> what will clean up VolumeAttachment objects after pods are deleted?

A/D controller. There is no difference there from attaching PVs.

> how will the VolumeAttachment objects be named? is it derived from the pod namespace and name?

It will be the same as PVs, i.e. CSI driver name + CSI volume handle + node name, all hashed to conform to naming rules.

> what is the behavior if two pods both contain identical VolumeSources?
￼
They both get the volume, assuming the volume is ReadWriteMany. And it will work even if one of the pods uses PV and the other has the same volume inlined.

liggitt: > It will be the same as PVs, i.e. CSI driver name + CSI volume handle + node name, all hashed to conform to naming rules.

CSI volume handle in a PV is considered trustworthy, because creating a PV is a highly privileged operation. The entity creating the PV is assumed to create unique volume handles, right?

CSI volume handle in a pod is not particularly trustworthy. What happens if two pods on the same node both specify the same CSI driver and volume handle, but different VolumeAttributes, fstype, etc? which one wins and gets its content in the single VolumeAttachment object?

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
liggitt: the object references in v1.VolumeSource (secret, endpoints, etc) are ambiguous when this is embedded in an object located outside the source pod's namespace.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: The first one wins. Its the same as when two users create non-CSI in-line volume, such as Ceph RBD or AWS EBS.

Which brings another topic: we should extend PSP / SCC to allow/disallow individual CSI drivers inlined in pods, similar to Flex:

https://kubernetes.io/docs/concepts/policy/pod-security-policy/#flexvolume-drivers

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: > what indicates whether an inline volume requires an attach/detach phase? many potential uses of inline CSI volumes only require mount/unmount

We currently have no answer here, all CSI volumes (both inline and PVs) must go through external attacher and have VolumeAttachment. It is probably for another proposal to solve this (and we need a smart idea how to do it - A/D controller currently does not know what drivers are deployed to the cluster and which of them need attach/detach).

> the object references in v1.VolumeSource (secret, endpoints, etc) are ambiguous when this is embedded in an object located outside the source pod's namespace.

Touche!

1. We could add pointer to pod to VolumeAttachment, but that complicates things when two pods share the same volume and the one that's referenced dies and user deletes its secrets. Having two pods sharing the same volume and having *different* Secrets looks quite bizarre though..

2. We could copy the secrets to VolumeAttachment.

3. We could change CSI not to require per-volume secrets on Detach time. IMO, all in-tree volume plugins don't need it. All of them need it either at attach or mount time.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: Added a note about this.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: Added PSP change.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: > 3. We could change CSI not to require per-volume secrets on Detach time. IMO, all in-tree volume plugins don't need it. All of them need it either at attach or mount time.

This may not be true when Cinder is updated to process secrets introduced in https://github.com/kubernetes/kubernetes/pull/63826

@dims, do you expect that Secrets introduced there will be used to detach volumes? If so, how do you make sure that the Secrets are available at detach time? Especially for the in-line volume in a pod, user can delete the Secrets at any time.

jsafrane:
@liggitt, I got one idea about the secrets for in-line volumes.

1. CSI plugin adds *reference* to pod to VolumeAttachment (as a field, exact API is pending for now)

2. VolumeAttachment reads the VolumeSource to attach, reads referenced Secrets (using given namespace) and calls CSI to attach the volume.

3. If attach succeeded, external attacher **creates a new secret in the namespace where the attacher runs**. It uses `VolumeAttachment.Name` as `Secret.Name` so it's unique and  can be found easily.
    
    The secret must be copied by external attacher and not by A/D controller, because A/D controller does not know in which namespace the attacher runs.

4. When the pod is deleted, the external attacher deletes the secret after successful detach.

This way, users may delete any Secrets in their namespaces, attacher keeps a copy. This needs to be carefully documented and admins must make sure that namespace with the attacher is safe enough. Would it be a way?

jsafrane:
hmm, and the external attacher should copy secrets also from PVs, in case the same volume is used inlined in another pod.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,148 @@
...
+Author: @jsafrane
+
+## Goal
+* Define API and high level design for in-line CSI volumes in Pod
+
+## Motivation
+Currently, CSI can be used only though PersistentVolume object. All other persistent volume sources support in-line volumes in Pods, CSI should be no exception. There are two main drivers:
+* We want to move away from in-tree volume plugins to CSI, as designed in a separate proposal https://github.com/kubernetes/community/pull/2199/. In-line volumes should use CSI too.
+* CSI drivers can be used to provide Secrets-like volumes to pods, e.g. providing secrets from a remote vault. We don't want to force users to create PVs for each secret, we should allow to use them in-line in pods as regular Secrets or Secrets-like Flex volumes.
+
```
kfox1111: You hint at it a bit, but might be good to explicitly add it as a motivation:
* Reach feature parity with FlexVolume api and allow for its deprecation.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,148 @@
...
+// Represents storage that is managed by an external CSI volume driver (Beta feature)
+type CSIVolumeSource struct {
+	// Driver is the name of the driver to use for this volume.
+	// Required.
+	Driver string
+
+	// VolumeHandle is the unique volume name returned by the CSI volume
+	// plugin’s CreateVolume to refer to the volume on all subsequent calls.
+	// Required.
```
kfox1111: There needs to be a mechanism to make this optional. CSI allows a driver to not implement CreateVolume, in which case VolumeHandle could be generated by kubelet itself.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,148 @@
+# In-line CSI volumes in Pods
+
+Author: @jsafrane
+
+## Goal
+* Define API and high level design for in-line CSI volumes in Pod
```
kfox1111: After reading through the spec once, there may be two different use cases wrapped up into one here.

1. secrety like volumes. per pod volumes.
2. more traditional volume drivers. volume create/delete/attach/etc.

While both need solving, the vast majority of difficult design/coding I think is around use case number 2.

Since use case number one may be much easier to implement, and would garner a bunch of drivers quickly should we consider breaking the spec into 2, to allow the first use case to be implemented faster then the second?

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: Added a new motivation item.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
kfox1111: ah. I was confusing the two. even still, the use case I see for one of the drivers looks like the following. There isn't really any benefit in requiring VolumeHandle I think:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: nginx:1.13-alpine
    ports:
    - containerPort: 80
    volumeMount:
    - name: data
      mountPath: /usr/share/nginx/html
  volumes:
  - name: data
    csi:
      driver: drivers.k8s.io/image-driver
      image: kfox1111/misc:test
```

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
kfox1111: +1. but just hoping to see at least part 1 implemented in 1.12. If the design of part 2 would push implementation out to 1.13, then maybe we should separate. If not, then lets keep it all together.

msau42:
Topology support in kubernetes is only provided at the PV layer.  So inline CSI volumes won't be able to support topology.

jsafrane:
> So inline CSI volumes won't be able to support topology.

That's IMO ok, no in-line volume supports any volume topology. Anyone who creates a pod with topology-constrained volume must set the right node selectors and affinities in the pod itself.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: There is no `CSIVolumeSource.Image`. If `kfox1111/misc:test` is unique volume ID (and it looks like it uniquely identifies source of the data to mount), it should be used as `VolumeHandle`.

kfox1111: Hmm... I haven't figured out the actual mapping to k8s. Reading through the pv->csi api spec, I think it actually should be:
volumes:
  - name: data
    csi:
      driver: drivers.k8s.io/image-driver
      volumeAttributes:
        image: kfox1111/misc:test

I have a prototype csi driver working here: https://github.com/kfox1111/drivers/blob/volume-driver/pkg/image/nodeserver.go#L143

It uniquely identifies the container image to use for the volume. the volume identifier though should match the pod's lifecycle. so the easiest thing I can think of is to use the podid as the volumehandle. or maybe just a random id (but consistent for the pods lifecycle).

For things that follow pod lifecycle, VolumeHandle doesn't really make sense for the user to specify. It is supposed to be unique, but you could have multiple instances of the same pod on the same host. If the user specified something, then it could conflict between the pods on the same host.

kfox1111: actually, thinking about it a little more, lets say you make a volume driver that was as trivial as possible. say, emptyDir for sake of discussion. Now say you had two instances of it within a pod:

```yaml
volumes:
  - name: foo
    csi:
      driver: drivers.k8s.io/emptyDir-driver
  - name: bar
    csi:
      driver: drivers.k8s.io/emptyDir-driver
 ```

Each instance would need to be unique per pod. so using pod_id as volume_id wouldn't be the right thing to do. Each should get its own volume_id. so maybe just randomly created ones might be the best option.

Drivers still might want to have access to the pod_id though. So maybe that could be passed through via an automatic volumeAttribute?

File: contributors/design-proposals/storage/container-storage-interface-inline-volumes.md [hunk 7]
```diff
@@ -3,13 +3,14 @@
 Author: @jsafrane
 
 ## Goal
-* Define API and high level design for in-line CSI volumes in Pod
+* Define API and high level design for in-line CSI volumes in Pod.
+* Make in-line CSI volumes secure for using ephemeral volumes (such as Secrets or ConfigMap).
 
 ## Motivation
-Currently, CSI can be used only though PersistentVolume object. All other persistent volume sources support in-line volumes in Pods, CSI should be no exception. There are two main drivers:
+Currently, CSI can be used only though PersistentVolume object. All other persistent volume sources support in-line volumes in Pods, CSI should be no exception. There are three main drivers:
```
sbezverk: nit: s/though/through/

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -102,29 +128,42 @@ type VolumeAttachmentSource struct {
...
+* Secrets for PVs are "global", out of user's namespace, so this assumption is probably OK.
+* Secrets for in-line volumes must be in the same namespace as the pod that contains the volume. Users can delete them before the volume is detached. We deliberately choose to let the external attacher to fail when such Secret cannot be found on detach time and keep the volume attached, reporting errors about missing Secrets to user.
+	* Since access to in-line volumes can be configured by `PodSecurityPolicy` (see below), we expect that cluster admin gives access to CSI drivers that require secrets at detach time only to educated users that know they should not delete Secrets used in volumes.
+	* Number of CSI drivers that require Secrets on detach is probably very limited. No in-tree Kubernetes volume plugin requires them on detach.
+	* We will provide clear documentation that using in-line volumes with drivers that require credentials on detach may leave orphaned attached volumes that Kubernetes is not able to detach. It's up to the cluster admin to decide if using such CSI driver is worth it.
```
sbezverk: Curious, if it happens, what would be a recovery steps for a cluster admin to clear this condition, it cannot stay orphaned forever.

kfox1111: I think this is a good compromise.

Maybe we should specify finer grained behaviour with respect to pod security policy though. Maybe that belongs in a different spec. Being able to white list particular csi drivers such as image-driver while making a driver such as nfs-driver, pv only could be handy. Looking at the pod security policy, 'volumes'  is currently a string array so would only allow/disallow all of csi. Perhaps the psp volume list could ignore just csi though and use the ".volume.csi.driver" info instead.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
kfox1111: Where does the VolumeHandlePrefix come from? is it specified in the pod spec? Should it come from the driver registrar (https://github.com/kubernetes-csi/driver-registrar) sidecar instead? Then the packager of the csi driver could tell k8s which type it expects so the user won't have to specify it most of the time.

jsafrane: VolumeHandlePrefix comes in pod spec, `pod.spec.volumes[x].csi.volumeHandlePrefix`.

jsafrane:
Added PSP update to allow/deny attachable in-line CSI volumes, as they could abuse nodes by deleting Secrets needed for detach.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: @sbezverk, there already is metric of failed detaches and admin can look at `VolumeAttachments` with failed detach.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
jsafrane: fixed

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
kfox1111: So under my current understanding of the spec, the theoretical csi emtpyDir example would need to be:
```yaml
volumes:
  - name: foo
    csi:
      driver: drivers.k8s.io/emptyDir-driver
      volumeHandle: foo
      volumeHandlePrefix: pod
  - name: bar
    csi:
      driver: drivers.k8s.io/emptyDir-driver
      volumeHandle: bar
      volumeHandlePrefix: pod
```
Thats quite verbose. And a miss copied entry could have bad results. It really feels like volumeHandlePrefix or at least its default should come from the driver some how.

Alternately, could the VolumeHandlePrefix be made optional and default to Pod (should be a safe default) and VolumeHandle default to the volume's name and be optional (guaranteed to be unique within a pod). That would restore the ease of use for using any csi driver where you want to map volume lifecycle to the pod. and still allow for specifying the exact volumes you want to mount when you need to.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,211 @@
...
+	VolumeHandlePrefix CSIVolumeHandlePrefix
+
+	// Optional: The value to pass to ControllerPublishVolumeRequest.
+	// Defaults to false (read/write).
+	// +optional
+	ReadOnly bool
+
+	// Filesystem type to mount.
+	// Must be a filesystem type supported by the host operating system.
+	// Ex. "ext4", "xfs", "ntfs". Implicitly inferred to be "ext4" if unspecified.
```
codenrhoden: Given https://github.com/kubernetes/kubernetes/issues/65122, I'm not sure that `ext4` should be inferred.

kfox1111:
@thockin not sure who suggested secrets and configmaps be moved to CSI. I don't see it in the proposal.

One of the use cases covered though is not about secrets as a k8s object, but a volume type that may contain per pod volumes. one subtype of that could be a volume plugin that generates "secrets" per pod. Not the kind=Secret definition of secret.

For example, say you wanted a volume type that gave a volume with an Active Directory Kerberos keytab for every pod that requested one. The pod would launch, the driver would get the request for a new volume for pod d18e74d3-f77a-4a4a-a916-b68a9aacba98, it would do some handshaking with the domain controller to create a service account and copy the credentials into a keytab file into the volume and return it. When the pod is deleted, the volume would get deleted and the driver could unregister the account before letting the volume be completely destroyed. CSI has all the plumbing to do this, provided it had access to the pod id and the pod spec allowed calling the CSI driver directly. This could not be done with an emptyDir + initContainer today because there is no way to do the unregister reliably, and you don't want end user pods being able to access the AD credentials needed to register/delete an account.

My interests are in writing a few other drivers that follow the per pod volume allocation behaviour (emptyDir like) but can live out of tree (so CSI). Not used for secrets at all, but for bringing new types of volumes that will fill some of the missing gaps when it comes to launching advanced workloads on k8s when you don't have full admin permissions to the cluster so hostPath is not an option.

So, lets back up a little bit and let me ask the question again but in a slightly different way. We kind of have 2 different use cases in this spec I think.
1. per pod csi drivers that behave like an enhanced emptyDir.
2. regular csi drivers.

Tthis proposal tries to tie both together. Which is one legitimate way of solving it. But, should we consider breaking it up into two proposals so we can focus on what is the best api for either in isolation? 
I agree with @thockin the prefix stuff makes it a bit hard on the users. prefix should be optional and the driver (or its packaging) should be able to provide enough info to set which mode it behaves in. Otherwise the user is forced to tell k8s how the driver should behave, and that is an implementation detail they should not have to know.

I think there is a pressing need for use case 1 to be solved soon. I've needed and worked around the lack of some use case 1 drivers in my code for a year and a half now. so its important.  I have not personally needed use case 2 like things yet. PV's cover those pretty well today. That proposal can simmer for a while while we figure out the best way to implement it. I think there are a lot more potential gotchas in use case 2 then use case 1.

thockin:
Can we force-fit this onto PersistentVolumes?  E.g. create a StorageClass
for "kerberos-keytab", claiming read-write-many capable.  User creates a
PVC against this class, controller wakes up and takes not that user X wants
one or more keytab.  When a pod references the PVC, the node-local portion
of the driver does what you described above.  Each pod gets a different
value, but they are all referencing the same conceptual volume.

This provides a hook wherein the "provisioner" can verify that this is
allowed, maybe unpack per-user credentials (so it only has creds for active
users), log stuff, etc.  If any of that fails, it fails long before the
user tries to run a pod.

kfox1111:
Hmm.. That provides a mechanism for doing the keytab like thing, yeah... Interesting.

What about the data image driver use case? https://github.com/kubernetes-csi/drivers/issues/85

If multiple different helm charts were trying to use the same image as a data volume, you might get one or more identical looking pvc's. Either in the same namespace or different ones. Is that ok? There would also need to be some mechanism in the storageClass to allow some kind of pvc attributes (imageName) to be copied into the resulting pv request so the driver would have access to them. Is there a new can of worms there? Or another solution to that?

jsafrane:
We don't want to move Secrets and ConfigMaps to CSI. We want to make it possible to make CSI drivers for ephemeral secrets-like volumes in Pods.

I want them in-line to keep consistency with existing API, e.g. injected them into Pods using pod presets or admission hook. Sure, PVC can be used for it, still it seems as ugly hack to me. We should not force users to create *PersistentVolume*Claim for something as ephemeral as kerberos ticket or temporary key.

thockin:
I understand that, but I want to explore every alternative.  To be viable,
this API can't come with as many sharp-edges as it has now.

kfox1111:
Hmm... lets keep exploring the pvc idea for a bit...

What about something kind of like what the StatefulSet does?

The volumeClaimTemplates section causes pvc's to be created magically per pod.
What if the in pod spec csi driver section ended up creating a pvc per pod, there being a storage class per driver, and with all the settings in the csi driver section being passed to the storageClass dynamic provisioner so they could be optionally passed to the pv?

This would allow a dummy dynamic provisioner to be written that just passes the arguments from the pvc to the allocated pv for drivers that don't need to do anything special, and custom dynamic provisioners to do things too.

Benefits:
 * more flexible
Drawbacks:
 * every pod gets a corresponding pv when perhaps may not be needed. this adds more api server load.
Unknowns:
 * may cause issues with multiattaching a volume to multiple pods? multiple pvc's per pv or multiple pv's pointing at the same volume? is this ok?

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,211 @@
...
+// PersistentVolume or a volume in-lined in a Pod.
+// Exactly one member can be set.
+type VolumeAttachmentSource struct {
+	// Name of the persistent volume to attach.
+	// +optional
+	PersistentVolumeName *string
+
+	// InlineVolumeSource represents the source location of a in-line volume in a pod to attach.
+	// +optional
+    InlineVolumeSource *InlineVolumeSource
```
mikedanese: What does inline mean here? Is this actually inline if the VolumeSource spec isn't embedded in the PodSpec?

jsafrane: Pod yaml has an in-line volume:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: testpod
spec:
  containers:
    ...
  volumes:
      - name: vol
        csi:
          driver: io.kubernetes.storage.mock
          volumeAttributes:
              name: "Mock Volume 1"
          volumeHandle: "1"
```

This `csi:` gets copied to `VolumeAttachment` because Pod does not need to exist at detach time and we need the volume source somewhere to detach volumes.

kfox1111:
Still a chance of the pod following driver being available in 1.12 or is that unlikely?

kfox1111:
Thinking about this some more, I still think we should tackle this in two pieces. per pod drivers and regular drivers.

In the first implementation, we only support per pod drivers. The user api would be
csi:
  driver: xxxx
  volumeAttributes:
     whatever: whatever

The packager would have to specify to the driver-registrar the --per-pod flag. If not, for now, when the pod was launched and the volume attempted to be bound, kubelet would fail it.

kubelet could automatically inject some volumeAttributes that are pod related like k8s.pod_id so the driver could know some things about the pod. It would make up a uuid for the volumeattachment that it would keep consistent.

This would allow per pod implementation to move forward while the rest of the implementation details on how stateful volume things like ebs/rbd/etc could be done safely.

This would be the shortest path to something working I think.

Going forward to use case 2, running everything through storageclass via @thockin's idea sounds like a good plan to me. I think the api proposed above does not preclude the implementation being switched to that model once the needed plumbing for that is in place.

kfox1111:
What can I do to help move this along?

vladimirvivien:
At the risk of revisiting what has already been said... if this proposal gets rejected, is the workaround here:

- emptyDir/hostpath/config/secrets remain in-tree vol drivers
- All CSI volume must use PV/PVC
- For inline, declare a PV/PVC and reference PVC in pod

kfox1111:
@vladimirvivien the work around for my use case is to write a flex driver instead. :/ can't be done with pvc.

thockin:
I think there's no reason to move EVERYTHING to CSI.  emptyDir, hostPath, secrets, and so on make sense as built-ins.  Should we make it possible to implement similar functions as CSI?  Sure.  Yeah.  Probably.  I mean, as long as it is not too complex.  That is the option 1 in @kfox1111 breakdown.

I'm not sure I understand the rest of your taxonomy.

Do we really need 2.1?  Like REALLY REALLY *need*?  Sorry, I may have lost the use case in this feeble brain.  Maybe it shakes out from enabling option1 ?

2.2 seems to be more generically served by the old idea of inline claims, which goes back to the notion that we support CSI through PVs.

kfox1111:
Yeah, I think that as csi support in k8s is defined today, (hosting csi drivers in containers), you actually need kubelet to handle at least hostpath and emptyDir at minimum. Some drivers probably require Secrets (rbd? for service account tokens?) too. So I think all of those probably should remain in kubelet. ConfigMaps pretty simple too if you are doing Secrets. So, +1 to keeping all those in tree and using them as a base to build custom drivers on (CSI)

I was just mentioning 2.1 as it is currently a case that kubernetes supports today in non csi drivers. For example, you can use the nfs driver directly in the pod spec and specify all the contact info (ip address, nfs version, etc). If we are migrating those type of drivers to CSI, then should we support that going forward? I'm not sure we should either. But we do need to discuss the use case since we need to decide to either deprecate it or embrace it moving forward.

jsafrane:
I think nobody is proposing to move *everything* to CSI. The plan is to move real persistent storage to CSI, e.g. NFS, iSCSI, Ceph, AWS EBS, GCE PD, OpenStack Cinder, vmWare, StorageOS, Portworx, ScaleIO and things like that.

For ephemeral storage (Downward API, Secrets, ConfigMap, Projected, EmptyDir) and HostPath (which is persistent, but very special) we want to stick to in-tree volume plugins in Kubernetes.

That said, it *should* be eventually possible to write CSI drivers for Secrets-like volumes that would provide similar functionality like Secrets, but get the actual secret content from a remote vault instead of Kubernetes API object.

I created new feature https://github.com/kubernetes/features/issues/603 to pass pod information from kubelet to CSI driver in NodePublish. It is completely orthogonal to this proposal, it does not matter if a volume is in-line in a pod or has a PV in this new feature.

thockin:
So if @kfox1111 point 1 is "eventually", and 2.1 derives from that, and 2.2 is better served with inline claims (or maybe even out-of-line claims), what is this about, again?

kfox1111:
The confusion to me is I don't think use case 2.1 is derived from use case 1.

In use case 1, I see the csi driver getting a unique id for each instantiation of the driver. so,
```yaml
spec:
  volume:
  - name: foo
    driver: image
    volumeAttributes:
      image: kfox1111/foo
  - name: bar
    driver: image
    volumeAttributes:
      image: kfox1111/bar
```
Would end up calling the image drivers NodePublishVolume with two different unique volume id's, one for foo and one for bar. Along with passing all the k8s info from https://github.com/kubernetes/community/pull/2439/files and the volumeAttributes directly. Drivers of this type would never have a dynamic provisioner, or a ControllerPublishVolume function?

Folks want a migration path from precsi pods to csi ones so some drivers can leave kubelet. So, for example, look at a pod that could exist today:
```yaml
spec:
  volume:
    - name: data
      nfs:
        server: foo.example.com
        path: /mythingy
```
That can be switched to use a csi based nfs driver and the nfs code can be removed from kubelet.
Something like:
```yaml
spec:
  volume:
    - name: data
      csi:
        driver: nfs
        volumeAttributes:
          server: foo.example.com
      path: /mythingy
```
This is use case 2.1 from above. I am not expecting the csi drivers in this class to know k8s pod specific things though as in use case 1. Drivers of this type may have a dynamic provisioner, or a ControllerPublishVolume function? Is 2.1 something we should consider supporting at all, or should we deprecate it? I'm not sure it is derived from use case 1 though. In some cases, doesn't the driver require you to use the volumeID the create function returned? How does this map into an nfs driver, or an rbd driver? I called it 2.1 cause I think in some ways, use case 2.2 could be used to handle it though. Maybe it ends up triggering a PVC request with the volumeAttributes called out in the PVC so that the dynamic provisioner can call any required cluster level attaching things? maybe nfs is a bad example. maybe ebs/cinder is a better example? Not really sure here. Still feels like there is a gotcha in this use case somewhere. But maybe you've thought it through more then I have, and it really is just solvable via use case 1 from above.

I'm perfectly happy just skipping 2.1 though and just covering use case 1 and maybe use case 2.2 in this PR and we can talk about 2.1 more in https://github.com/kubernetes/community/pull/2199 ?

kfox1111:
So, where are we at?

thockin:
Sorry I was out for a few days.

My point was that both usecase 1 and 2.1 require to add a new CSI block to `pod.spec.volumes[]`.  If you enable that for case 1, haven't you implicitly enabled it for case 2.1?

You said "I am not expecting the csi drivers in this class to know k8s pod specific things though as in use case 1".  What things?  And why make those cases different?

Just like built-in PD or EBS support, specifying it in the pod does NOT trigger provisioning.  It's a direct linkage, whereas PVC is indirect.  If you specify an NFS server and path, you get exactly _that_ volume.  I find this case to be an unfortunate side-effect of history and I would RATHER we didn't support it for CSI drivers.  That said, kerberos or whatever as local-ish drivers makes sense.  If we can't pretend those are PVCs (which I am still not convinced we can't?) then they have to be present in pod spec.  Right?

If you want to trigger provisioning for pod-related volumes, this is where the idea of inline claims comes in, but I'm still not convinced that is justified above normal claims. 

We might do better face-to-face.  Is there a SIG meeting or something where we can talk?  I want to explain my thinking a bit more but have a meeting now. will follow up soon.

thockin:
Here's my thought process.  Sorry it got long.

The fact that we support EBS and GCE PD (and maybe even NFS?) directly-linked from a pod is unfortunate.  That API predates indirect linkage (PV/PVC), which seems better in pretty much every way.  If we could kill  those off, we would probably try.  Ergo, adding more of those - via CSI - is undesireable.

That said, the local drivers like emptydir, tmpfs, secret, etc are not obviously better as PV/PVC.  They really are pod-bound, so it makes sense that their API linkage is direct.

It's about life-cycle.  PVs are designed to represent things that exist regardless of whether a pod is using them at a given point in time, and have their own lifetimes.  Pod-linked volumes are coupled to the pod's lifetime.

Now there's a desire to offer local-ish or pod-bound volumes through CSI.  That seems reasonable.

If we add a CSI block to `pod.spec.volumes[]`, does that mean that any CSI driver can work?  Even PD/EBS/NFS?  That is the "undesireable" part above.  But if it is a side-effect of a necessary change, I guess I can live with it.

But before we concede that, I want to revisit an assumption.  Earlier I said that "local drivers ... are not obviously better as PV/PVC".  Are they obviously worse?  I am not sure (literally thinking it thru as I type).  Imagine for a minute that we made an `ephemeralEmptyDir` driver.  To use it, you create a PVC for a class which is bound to a PV.  That PV is fake.  It doesn't represent a concrete volume, but a placeholder.  When kubelet calls the node-half of the CSI driver, it makes an empty directory somewhere and mounts it.  When I delete my pod that is unmounted and the empty dir is deleted.

That seems to fit the CSI model well enough, and it satisfies the emptyDir UX well enough.  It's a little weird in that I specified a PVC for something that is not really persistent.  But it works, right?

Repeat for Secret or ConfigMap (yes there are RBAC issues there, so you probably would not ACTUALLY do this), but stick with me :).  Can we think of a local-ish volume type that truly CANNOT work this way?

(assuming my thought experiment holds this far...) So now we have demonstrated that we really only need PV/PVC.  Now we can argue whether the UX of creating a PVC *and* a Pod is bad enough to justify adding inline PVCs to pod.

So the main thrust is - how little can we get away with?  We already have PV/PVC - can we bend that to fit our needs?  Is that to horrible to actually live with, or is it just the right amount of horrible?

saad-ali:
>We already have PV/PVC - can we bend that to fit our needs? 

We could probably get it to work, but I don't think it would be that straight forward and would get pretty ugly.

Ignoring the whole life cycle mismatch, it's also worth thinking about how users would use it.

You mentioned:

> To use it, you create a PVC for a class which is bound to a PV. That PV is fake. It doesn't represent a concrete volume, but a placeholder. When kubelet calls the node-half of the CSI driver, it makes an empty directory somewhere and mounts it. When I delete my pod that is unmounted and the empty dir is deleted

PVCs are designed to be agnostic to the specific type of storage. That's why we have `StorageClasses`, but `StorageClasses` are designed to be generic templates for any newly provisioned volume, not necessarily a description exactly one volume (that would be a PV). With ephemeral local volumes we generally want to specify a different set of parameters for each volume (e.g. the name of the secrets, config maps, etc.). So either we would have to put those parameters on the PVC object some how, or have a different StorageClass for every volume. At that point things would be getting very ugly.

vladimirvivien:
@jsafrane for clarification, why is type `VolumeHandlePrefix` is introduced here to mitigate volume handle collision.  Why other inline volumes not using similar approach?  What's different here?

kfox1111:
If pvc supported custom parameters from the user so that the dynamic provisioner could access properties from the pvc and copy them into the generated pv's, that would work I think. If that was the case, use case 1 could be done without a direct csi inline driver at all but instead a pvc that looked like:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: myclaim
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      volumeAttributes:
        image: kfox1111/testimg
  storageClassName: csiImage
```

thockin:
Being able to provide args through PVC all the way to drivers was always
part of the PV/PVC plan.

sbezverk:
Interesting, every time when we tried to extend PVC with any attributes, we were shot down by API folks, saying that it would impact PVC portability.

thockin:
It's a tradeoff - if portability matters, then you don't want args.
But sometimes you need to pass a param.  Perhaps we haven't needed it
thus far...

I don't want to be belligerent about this.  I just want to explore every option.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
vladimirvivien: @jsafrane couple of things I would like to suggest here:
1) I don't like the name `InlineVolumeSource` as it seems to prematurely leak the usage/impl of component in its name.  Maybe 
2) I am thinking of removing one level of object abstraction by putting `PersistentVolumeName` and `Source` directly in the `VolumeAttachmentSpec`

```go
type VolumeAttachmentSpec struct {
    // <snip>
    // Source represents the volume that should be attached.
    PersistentVolumeName *string

    Source *VolumeAttachmentSource
}

type VolumeAttachmentSource struct {
	// VolumeSource is copied from the pod. It ensures that attacher has enough
	// information to detach a volume when the pod is deleted before detaching.
	// Only CSIVolumeSource can be set.
	// Required.
	VolumeSource v1.VolumeSource
 	// Namespace of the pod with in-line volume. It is used to resolve
	// references to Secrets in VolumeSource.
	// Required.
	Namespace string
}
```

Let me know if those changes work for you... If so, I can continue with implementation.

kfox1111:
Another interesting ramification of always using pvc, would be that since it goes through the dynamic provisioner, it can add a flag to the generated pv's that state the driver has pod semantics. Rather then using a new object to describe that state, or a new api on the registrar. It also only shows up through the admin api's too which means users cant inappropriately muck with it.

kfox1111:
I think @sbezverk's comment is a good one. If we are to explore that route, we need buyin from the api team that patches to enable the use case will not be blocked. If we can't get that agreement, then we're just spinning our wheels exploring non options. What is the right path to get buyin?

sbezverk:
@kfox1111 you need to have ok from @liggitt for PVC API changes

liggitt:
> @kfox1111 you need to have ok from @liggitt for PVC API changes

(or another of @kubernetes/api-approvers ... @thockin is much more of a domain expert on storage bits than me)

The desire to keep parameters limited to the storage class and PV objects has traditionally been two-fold:
* preserve portability of PVCs in typical uses
* keep control of parameters with security-related impact limited to cluster storage admins (the ones setting up the storage classes, provisioners, etc)

Storage-related parameters aren't always neatly divided into safe/unsafe buckets (see the discussion around [mount options](https://github.com/kubernetes/community/pull/321) for an example). If we decided to add a way to parameterize PVCs, we should think through:
* discoverability - how do consumers know what parameters and allowed values are available?
* guidelines for what types of control should be exposed to namespaced PVC consumers via parameters (and the implications of exposing "unsafe" parameters to namespaced PVCs)
* expected behavior of provisioners if unknown/disallowed parameters are specified

saad-ali:
> Being able to provide args through PVC all the way to drivers was always part of the PV/PVC plan.

We've thought about it a lot, and how we can do this without sacrificing portability (e.g. provisioner ignores PVC parameters that it doesn't understand, for example). But the drawbacks are ugly, and we have yet to agree on an approach there.

Consider we don't have a way to specify PVC specific parameters today, I don't think having ephemeral volumes use PVCs is a viable option.

What's wrong with this approach:

* Validation rejects inline volumes for CSI Volume Plugin by default.
* If a CSI plugin registers using CSI Cluster Registry (https://github.com/kubernetes/features/issues/594) and indicates that it permits inline, validation permits inline.

vladimirvivien:
The VolumeHandlePrefix does feel like it can a bit burdensome to user. I am thinking that  the implementation plumbing should use one way of applying prefix (pod id always) to avoid volume handle id collision.  Which would imply that a pod within the same namespace would never have the same id either.

As far as volumeHandle becoming completely optional, not sure that would work.  The VolumeHandle value is used as a way to identify specific volume to the external driver.  Remember the volume must be pre-existing for inline volumes.

As far as rebasing this proposal on top of https://github.com/kubernetes/features/issues/594, I think the two should evolve separately as #594 has much broader implication than just inline volume.

jsafrane:
> What's wrong with this approach:
>
> * Validation rejects inline volumes for CSI Volume Plugin by default.
> * If a CSI plugin registers using CSI Cluster Registry (kubernetes/features#594) and indicates that it permits inline, validation permits inline.

We don't do cross-object validation in API server. Validation webhook would be a viable option here.

kfox1111:
VolumeHandle is a way to give an identifier to csi for tracking the volume lifecycle. I'm arguing, in the case of per pod volume drivers, the lifecycle is defined by the pod itself. It actually is very counterintuitive to have the user ever set it. That would be like adding a volumeHandle property to emptyDir. Sure, every emptyDir does have a unique identifier enough that it lands on disk in a unique place, but the user never has to worry about it. In fact, I can only think of bad ways a user could abuse it if they were setting it as the driver would expect to delete the storage on main pod removal, but if some other pod was referencing it, some undefined bad behavior could happen.

kfox1111:
Or, following up on that idea a bit, what about in the CSIRegistry having the option for setting the VolumeHandlePrefix as a default, or forcing it to be a particular value, and having an option to set VolumeHande by default to be the volume.name from the spec with an option of saying if its a default or if it is forced?

This would allow the csi driver packager to completely set the behavior the inline spec could have appropriately for the driver while still allowing the user to specify things if that is a reasonable thing to allow enabling use case 2.1 still.

vladimirvivien:
It seems the consensus is coming down to doing inline CSI the natural way (without StorageClass/PVC). Maybe it's time to kinda reset the expectations as to what it means to be inline pod.

**Inline Per Pod Mode 1**
- volume and volume attribs specified in pod spec
- the CSI driver (or internal CSI plumbing) generates the `VolumeHandle` and handles VolumeHandle naming collision strategy (default pod-prefix), etc
- CSI pod runs on its own lifecycle

Example spec:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: some-pod
spec:
  containers:
    ...
  volumes:
      - name: some-vol
        csi:
          driver: some.driver
          volumeAttributes:
              foo: bar
```

**Inline Per Pod Mode 2**
- volume specified in pod spec
- Human specifies VolumeHandle in pod spec, 
- VolumeHandle naming collision can be specified somewhere (if not, defaults to pod-prefix)
- CSI pod runs on its own lifecycle

Example spec:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: some-pod
spec:
  containers:
    ...
  volumes:
      - name: some-vol
        csi:
          driver: some.driver
          volumeAttributes:
              foo: bar
          volumeHandle: some-handle
```

bgrant0607:
@saad-ali @jsafrane 

Is there a roadmap of where we plan to go with storage volumes? I'm concerned about continued organic growth in this area. I'd like to see a sketch of the desired end state before adding more features.

cc @thockin

kfox1111:
My understanding is that the end goal is to deprecate FlexVolumes in favor of CSI, and to migrate most of the native volume drivers from in tree to csi except for "emtypDir, hostPath, configmap and secret"
So ultimate goal is to minimize code in k8s and use csi.

FlexVolumes currently provide per pod like behaviors. We're trying to figure out how to do them with CSI.

bgrant0607:
@kfox1111 Thanks. Could we please add a statement like that to:
https://github.com/kubernetes/community/blob/master/sig-storage/volume-plugin-faq.md

or add a separate roadmap doc to that subdir?

bgrant0607:
And would we deprecate the other volume sources in the API in addition to moving their implementations out of tree?

kfox1111:
@vladimirvivien  with flexvolume you can only deploy it on the host, not in pods currently. That is one of the selling points for getting support into CSI drivers.

but, you can create a flex driver, that behaves as though it is an enhanced emtpyDir+content in that it gets the extended pod info and can create/delete a backing volume following the pod's create/deletelifecycle and customizing it before handing it over to the pod.

So, for trivial example, you could write a flex driver that put the pod's id into a file in the root of the volume. Similar to the downwardapi volume type. Each pod gets its own dynamic temp volume.

We'd like to do the same kind of thing with CSI drivers.

There are two parts of the puzzle to allow this use case. This issue allowing the definition of csi volumes inline, the same as you can for flex volumes,
and this issue:
https://github.com/kubernetes/community/pull/2439
for passing pod info to the driver.

saad-ali:
>>> My understanding is that the end goal is to deprecate FlexVolumes in favor of CSI, and to migrate most of the native volume drivers from in tree to csi except for "emtypDir, hostPath, configmap and secret"
>>> So ultimate goal is to minimize code in k8s and use csi.
>>> 
>>> FlexVolumes currently provide per pod like behaviors. We're trying to figure out how to do them with CSI.
>> 
>> @kfox1111 Thanks. Could we please add a statement like that to:
https://github.com/kubernetes/community/blob/master/sig-storage/volume-plugin-faq.md
>> 
>> or add a separate roadmap doc to that subdir?
>>
>> And would we deprecate the other volume sources in the API in addition to moving their implementations out of tree?
>
>  I think that may be covered in #2199

@bgrant0607 let me know if #2199 answers your questions, if not happy to expand that or the FAQ to clarify.

jsafrane:
Updated with `CSIDriver.Spec.VolumeHandleMode` to configure generation of `VolumeHandles` for inline volumes.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,190 @@
...
+
+### VolumeHandle generation
+The VolumeHandle, for certain CSI drivers, may be omitted by its users for in-line volumes (i.e. secrets, configMaps, etc). When this is the case, the Kubelet will employ a naming strategy to generate the value for the volumeHandle.  The Kubelet will use the [`CSIDriver` configuration object](https://github.com/kubernetes/community/pull/2514) to figure out how to auto-generate a volumeHandle.
+
+The `CSIDriverSpec` type will expose field `VolumeHandleMode` which can have be set to:
+
+* `AutomaticVolumeHandleGeneration`
+* `NoVolumeHandleGeneration`
+
+When the driver is configured with `CSIDriverSpec.VolumeHandleMode = AutomaticVolumeHandleGeneration` and the volumeHandle is not specified, the Kubelet will automatically generate the volume handle to be sent to the CSI driver.  The generated value will be a combination of podUID and pod namespace.
```
kfox1111: Hmm... I don't think this is unique enough.

It should be possible to launch two different image driver volume instances in a pod with different images. But with podUID and pod namespace, it would end up being the same value for both volumes with this scheme.

What about podUID and Volume Name? I think podUID is unique even over different namespaces, and volume name is guaranteed unique in the pod spec.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+// Represents storage that is managed by an external CSI volume driver (Alpha feature)
+type CSIVolumeSource struct {
+	// Driver is the name of the driver to use for this volume.
+	// Required.
+	Driver string
+
+	// VolumeHandle is the unique ID of the volume. It is the volume ID used in
+	// all CSI calls, optionally with a prefix based on VolumeHandlePrefix
```
kfox1111: VolumeHandlePrefix isn't in the spec anymore.

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
vladimirvivien: Ok thanks for pointing this out.

vladimirvivien:
@thockin I addressed the remaining minor comments.

AishSundar:
/milestone v1.13

adding this to 1.13 milestone since it seems to be related to https://github.com/kubernetes/features/issues/603

contributors/design-proposals/storage/container-storage-interface-inline-volumes.md
kfox1111: looks like we very well may miss the 1.13 window even for part 1 because we're still debating how to solve part 2. Can we please set aside part 2 for now?

thockin:
@jsafrane @vladimirvivien @saad-ali @msau42 

What is the status of this proposal and the associated PR?  The last I looked at the API PR it had a significant compat problem.  This spec has outstanding comments.

Is it going to progress?  When will it be ready for re-review?  If it needs my attention, I need to know. :)

vladimirvivien:
Hi @thockin

There is progress. But we've been focusing on implementation which is revealing some issues we're working through here - https://github.com/kubernetes/kubernetes/pull/68232 (which will change some of the design). If we cant get these issues resolved by Monday, this may not go in 1.13.

kfox1111:
I think we have come up with a plan to get a sold enough base implemented this weekend for the 1.13 alpha functionality.

corentone:
Will there be any path to integration with local-ephemeral-storage metrics? Or thoughts if that'd make sense?
https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#local-ephemeral-storage

msau42:
In the initial alpha implementation, it will not be integrated, but before beta we should have a design.  Thanks for bringing that up!

msau42:
(btw, this PR has been replaced by a kep. Latest update [here](https://github.com/kubernetes/enhancements/pull/790#discussion_r259543973))

vladimirvivien:
This PR has been replaced by KEP - https://github.com/kubernetes/enhancements/pull/790#discussion_r259543973 . Will be going as alpha in 1.15.

davidz627:
@vladimirvivien @jsafrane if the KEP is a full replacement please close this PR

vladimirvivien:
@jsafrane I can't close it.

PR #2274, opened 2018-06-15
contributors/devel: request better godep vendoring commit messages

tpepper:
On the release team I've seen a number of issues go by this year
where vendor code was updated and later implicated in issue triage
and bisection, but the original commits didn't sufficiently document
the change.  Reverting the update then reintroduced a previously
fixed issue, which was insufficiently documented in the changelog.  The
subsequent churn led to a lot of work by multiple people to resolve.

Terse godep vendor code update commit messages seem to be the norm and
we should aim higher.

### DIFF
--- a/contributors/devel/godep.md
+++ b/contributors/devel/godep.md
@@ -156,6 +156,22 @@ transitively), you'll also need to update deps there:
 ./hack/update-staging-godeps.sh
 ```
 
+## Commit messages
+
+Terse messages like "Update foo.org/bar to 0.42" are problematic
+for maintainability.  Please include in your commit message the
+detailed reason why the dependencies were modified.
+
+Too commonly dependency changes have a ripple effect where something
+else breaks unexpectedly.  The first instinct during issue triage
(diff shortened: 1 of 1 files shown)


tpepper:
I feel this can be a slightly fuzzy cultural process thing that varies with project and developer, if you're coming at problem triage from a GitHub artifact perspective (PR and issue text) versus from a git perspective ('git bisect', 'git log' or 'git whatchanged'), and also whether or not merge commits are in use.  Since merge commits are in use, I'd like it in both places, but would be happy to more consistently have it in at least one and if only one then the commit itself.  But that's me biasing to to git.  In k8s any more git feels less authoritative on rationales than GitHub, so if the reasoning is being discussed in the GH PR then the PR message is perhaps the better place if reviewers ask for clarifying text and that's easier perhaps to quickly edit in a browser versus git commit amend?  ¯\_(ツ)_/¯

thockin:
FWIW, I think "both" is correct :(

PR #2275, opened 2018-06-15
Fixing cloud provider KEP.

cheftako:
Fixed reference to NodeController to reflect IPAM/Lifecycle breakout.
Added proposal for building CCM and generating CP specific deployments.


keps/sig-cloud-provider/0002-cloud-controller-manager.md
```diff
@@ -266,9 +298,317 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+like ranges. However Google then wires in the cloud provider to provide custom functionality on things like CIDRs. The 
+thought then is that in the short term we allow it to be run in either KCM or CCM. Most cloud providers would run it in 
+the KCM, Google would run it in the CCM. When we are ready to move cloud provider code out of K8s/K8s, we remove the 
+cloud provider code from the version which is in K8s/K8s and continue to have flags to control if it runs. 
+K8s/Cloud-Provider-Google could then have an enhanced version which is run in the CCM. Other controllers such as Route 
+and Service needs to run in either the KCM or CCM. For things like K8s/K8s e2e tests we will always want these 
+controllers in the K8s/K8s repo. Having it in the K8s/K8s repo is also useful for keeping the behavior of these sort of 
+core systems consistent.
+
+#### Staging
```
andrewsykim: I personally think we should pitch moving code to k8s/k8s/staging in a separate KEP. Given we'll want every in-tree cloud provider to approve it might be easier to consume if it's presented in a more consise / smaller scoped KEP.

keps/sig-cloud-provider/0002-cloud-controller-manager.md
```diff
@@ -434,9 +774,58 @@ Community norms.
...
+
+### Staging Alternatives
+
+#### Git Filter Branch
+
+One possible alternative is to make use of a Git Filter Branch. This would be an alternative to the Staging mechanism. 
+This has issues such as not giving K8s/K8s developers any indications of what the dependencies various 
+K8s/Cloud-providers have. Without that information it becomes very easy to accidentally break various cloud providers 
+and time you change dependencies in the K8s/K8s repo. With staging the dependency line is simple and automatically 
+enforced. Things in Staging are not allowed to depend on things outside of Staging. If you want to add such a
```
nckturner: Two things that in my opinion could be slightly clearer: 1. Is the git filter branch proposal to just move the code straight to the cloud provider repository?  2. The mechanism which enforces "Things in Staging are not allowed to depend on things outside of Staging" (maybe a link would suffice)

keps/sig-cloud-provider/0002-cloud-controller-manager.md
```diff
@@ -266,9 +298,317 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+
+### Design Options
+
+For the short term problem of sharing code between K8s/K8s repo, the K8s/cloud-provider repos and K8s SIGs/library repos; 
+there are 2 fundamental solutions. We can push more code into Staging to make the code available to the 
+K8s/cloud-provider repos. Currently for code which needs to be shared between K8s/K8s itself and other projects we put 
+the code in Staging. This allows the build machinery to detect things like attempts to make the shared code depend on 
+code which is not shared (i.e. disallowing code in Staging from depending Iton code in K8s/K8s but not in Staging) 
+In addition using Staging means that we can benefit from work to properly break out the libraries/code which exists in 
+Staging. Most of the Staging SIGs/library repos. The K8s/cloud-provider repos should be the top of any build dependency
```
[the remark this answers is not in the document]
andrewsykim: We have to hash this out a bit more, but I _think_ what we want as an intermediate step is to move all the code in `k8s/k8s/pkg/cloudprovider/` to `k8s/k8s/staging/pkg/cloudprovider` which you then import from `k8s/cloud-provider-*`. We should avoid importing `k8s/cloud-provider-*` in `k8s/k8s`, but we may end up making exceptions while we transition to account for dependencies in `k8s/k8s` (though in theory, we should have `k8s/k8s/staging/cloudprovider` for as long as we have those dependencies).  

For other sig owned repos / external projects you would import `k8s/k8s/staging/pkg/cloudprovider` until we're ready to move everything into `k8s/cloud-provider-*`.

keps/sig-cloud-provider/0002-cloud-controller-manager.md
```diff
@@ -266,9 +298,317 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+cloud provider code from the K8s/K8s repo. We need to keep this code in K8s/K8s while we still support cloud provider 
+deployments from K8s/K8s. In the long term this code should be part of each cloud providers repo and that code should 
+be removed from K8s/K8s. This suggests that in the short term that code should have one source of truth. However it 
+should probably not end up in the vendor directory as that is not its intended final home. Other code such as the 
+controllers should end up in the vendor directory. Additionally each provider will need their own copy of 
+pkg/cloudprovider/providers/providers.go and related build file to properly control which Cloud Provider Implementation 
+get linked in.
+
+We also need to be able to package a combination of binaries from K8s/K8s and K8s/cloud-provider-<cp> into a deployable 
+package. The code for this will need to accommodate things like differing side cars for each cloud provider’s CSI
```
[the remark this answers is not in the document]
cheftako: I think there will need to be some of guidance on this. That having been said I think things like https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/, https://github.com/kubernetes/community/blob/master/keps/sig-cloud-provider/0002-cloud-controller-manager.md#repository-requirements and the cluster lifecycle team may have opinions.

keps/sig-cloud-provider/0002-cloud-controller-manager.md
```diff
@@ -266,9 +298,317 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+Controller-Manager the --cloud-provider flag should be set appropriately for that cloud provider. In addition we need 
+to set the set of controllers running in the Kubernetes Controller-Manager. More on that later.
+
+For further background material please look at [running cloud controller](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/).
+
+### Design Options
+
+For the short term problem of sharing code between K8s/K8s repo, the K8s/cloud-provider repos and K8s SIGs/library repos; 
+there are 2 fundamental solutions. We can push more code into Staging to make the code available to the 
+K8s/cloud-provider repos. Currently for code which needs to be shared between K8s/K8s itself and other projects we put
```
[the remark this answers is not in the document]
cheftako: Good catch, thanks

keps/sig-cloud-provider/0002-cloud-controller-manager.md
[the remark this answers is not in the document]
cheftako: +1 to Andrews explanation. 

Additonally I'm trying to set up a dependency ordering to avoid cyclic graphs and similar problems. So I'm suggesting we have a final state something like K8s/cloud-provider-aws (This is the repo which builds the deployable) which depends on K8s/K8s (the repo containing the K8s "kernel") and K8s Sigs/storage-csi-aws (the aws csi side car) which depends on K8s Sigs/cp-aws (the aws cloud provider interface). Right now I just want to be clear about ensuring that we do create any K8s/K8s  dependencies on K8s/cloud-provider-*.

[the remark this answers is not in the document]
cheftako: Good catch. How about "Most of the repositories we are adding to staging should end up as K8s SIGs/library repos."

keps/sig-cloud-provider/0002-cloud-controller-manager.md
```diff
@@ -266,9 +298,317 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+manager framework its own K8s/K8s Staging repo. 
+
+Lastly we have any controllers which might run in either KCM or CCM. This includes things like NodeIpamController, 
+which will be shared both in the short and long term. Currently it needs to take a cloud provider to allow it to do GCE 
+CIDR management. We could handle this by leaving the cloud provider interface with the controller manager framework 
+code. The GCE controller manager could then inject the cloud provider for that controller. For everyone else 
+(especially the KCM) NodeIpamController is interesting because currently everyone needs its generic behavior for things 
+like ranges. However Google then wires in the cloud provider to provide custom functionality on things like CIDRs. The 
+thought then is that in the short term we allow it to be run in either KCM or CCM. Most cloud providers would run it in 
+the KCM, Google would run it in the CCM. When we are ready to move cloud provider code out of K8s/K8s, we remove the
```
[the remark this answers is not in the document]
cheftako: Yes :smile: Beyond that a CP may want to run even a stock controller in something other than the KCM to do things like limit permission domains or blast radius.

keps/sig-cloud-provider/0002-cloud-controller-manager.md
cheftako: Adding to KEP 13

keps/sig-cloud-provider/0002-cloud-controller-manager.md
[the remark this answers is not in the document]
cheftako: Changed the wording to try to convey this.

keps/sig-cloud-provider/0002-cloud-controller-manager.md
```diff
@@ -266,9 +298,317 @@ In case of the cloud-controller-manager, the deployment should be deleted using
...
+- **Kube-utilities**
+- Metrics
+- Sample-apiserver
+- Sample-Controller
+
+#### Cloud Provider Instances
+
+Currently in K8s/K8s the cloud providers are actually included by including [providers.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/providers.go)
+file which then includes each of the in-tree cloud providers. In the short term we would leave that file where it is 
+and adjust it to point at the new homes under Staging. For the K8s/cloud-provider repo, we would have a this CCM
```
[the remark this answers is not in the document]
cheftako: Thanks

andrewsykim:
This might be useful information form ComponentConfig KEP:
```
Support ComponentConfiguration for the cloud-controller-manager, as it’s a stop-gap for the cloud providers to move out of tree. This effort is in progress.
When the currently in-tree cloud providers have move out of tree, e.g. to `k8s.io/cloud-provider-gcp`, they should create their own external and internal types and make the command support loading configuration files.
The new repo can reuse the generic types from the to-be-created, `k8s.io/controller-manager` repo eventually.
Meanwhile, the cloud-controller-manager will reference the parts it needs from the main repo, and live privately in `cmd/cloud-controller-manager`

```

keps/sig-cloud-provider/0002-cloud-controller-manager.md
cheftako: Added

keps/sig-cloud-provider/0013-build-deploy-ccm.md
```diff
@@ -0,0 +1,652 @@
+---
+kep-number: 13
+title: Cloud Provider Controller Manager
```
andrewsykim: Does this title need to be changed?

keps/sig-cloud-provider/0013-build-deploy-ccm.md
```diff
@@ -0,0 +1,652 @@
...
+K8s/CP releases are largely decoupled from K8s/K8s releases. The cadence of K8s/K8s releases should not change. We are 
+assuming that K8s/CP releases will be on a similar or faster cadence as needed. It is desirable for the community for 
+all cloud providers to support recent releases. As such it would be good for them to minimize the lag between K8s/K8s 
+releases and K8s/CP releases. A K8s/CP cannot however release a Kubernetes version prior to that version having been 
+released by K8s/K8s. (So for example the cloud provider cannot make a release with a 1.13 Kubernetes core, prior to 
+K8s/K8s having released 1.13) The ownership and responsibility of publishing K8s/K8s will not change with this project. 
+The K8s/CP releases must necessarily move to being the responsibility of that cloud provider or a set of owners 
+delegated by that cloud provider.
+
+#### Migrating to a CCM Build [WIP]
```
andrewsykim: IMO I think we should try to reduce the scope of this by removing all the WIPs.

keps/sig-cloud-provider/0013-build-deploy-ccm.md
cheftako: Yes. How about "Switching To Cloud Provider Repo And Builds"

keps/sig-cloud-provider/0013-build-deploy-ccm.md
```diff
@@ -0,0 +1,652 @@
...
+  - "@thockin"
+editor: TBD
+status: provisional
+---
+
+# Remove Cloud Provider Code From Kubernetes Core
+
+## Table of Contents
+
+- [Remove Cloud Provider Code From Kubernetes Core](#remove-cloud-provider-code-from-kubernetes-core)
```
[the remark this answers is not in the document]
cheftako: Sorry fixed.

keps/sig-cloud-provider/0013-build-deploy-ccm.md
cheftako: Lets chat. Willing to remove the WIP flag from things like [Migrating to a CCM Build] and make it part of the proposal. I think for things like KubeAdm and TestGrid we need a proper answer in the KEP. I don't want to hold up the initial work while we get those answers but we need to track that we need those answers.

keps/sig-cloud-provider/0013-build-deploy-ccm.md
```diff
@@ -0,0 +1,652 @@
...
+is broken it could be the result of a code change in either, and modifications by an engineer testing against one cloud 
+provider could have unintended consequences in another. 
+
+To address this issue, SIG-Testing has invested in a process by which participating cloud providers can report results 
+from their own CI testing processes back to TestGrid, the Kubernetes community tool for tracking project health. 
+Results become part of a central dashboard to display test results across multiple cloud providers. 
+
+More thought should be put into managing multiple, independently evolving components. 
+
+## Implementation Details/Notes/Constraints
```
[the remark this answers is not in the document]
cheftako: My bad. Removed the C&P mistake.

keps/sig-cloud-provider/0013-build-deploy-ccm.md
```diff
@@ -0,0 +1,494 @@
...
+- Apimachinery
+- Apiserver
+- Client-go
+- **Controller**
+  - **Cloud**
+  - **Service**
+  - **NodeIpam**
+  - **Route**
+  - **?Volume?**
+- **Controller-manager**
```
andrewsykim: Been thinking about this more lately. 

I think we should be consistent with what is happening with `kube-apiserver` & `kube-controller-manager`. From what I can tell it seems like it's unclear whether all the Kubernetes components will be broken up into its own repos (unless this was stated somewhere and I missed it). Until we know for sure, I think we should hold off on trying to move controller-manager and controllers code. 

I also think it'll be easier to get buy-in if we scope this KEP for only moving cloud provider code and worry about controllers when the story around that is clearer.

andrewsykim: I forgot about https://github.com/kubernetes/community/pull/2354 but it's unclear if `k8s.io/{component}` repos are for component types AND the actual code for the component. Will check and report back.

andrewsykim: nvm, I asked the sig cluster lifecycle folks, all the code will be moved to k8s.io/components

andrewsykim: Leaving controllers where they're at now means they can be used in-tree and still be vendored in K8s/CP. Not saying that I don't want to see controllers be pulled out of K8s/K8s but it may be easier to move provider code out first and move controllers when we have a better idea of what that would look like (also to reduce # of high impact changes in the k8s/k8s codebase). I'm a bit on the fence about this since I don't know what the implications of moving controllers (cloud node controller, service, routes, etc) into staging would be 🤔.

cheftako: So if leave them where they are they can definitely be used in in-tree. However we have been trying for awhile to use staging as a delimiter for what can be vendored in externally. If we move the relevant controllers to staging they can be used in-tree and they are then part of the code we have officially marked as being vendorable for K8s/CP. I would like to see this kept in the KEP as its the right thing to do. If we want to mark it as a longer term goal which is the right thing to do, I would be fine with that.

cheftako: Worth reading @smarterclayton thoughts on staging found at https://github.com/kubernetes/kubernetes/issues/28559. It references client-go but I think it generally applies to packages which need to be shared beyond K8s/K8s.

keps/sig-cloud-provider/0013-build-deploy-ccm.md
```diff
@@ -0,0 +1,496 @@
...
+      - [Background](#background)
+      - [Design Options](#design-options)
+          - [Staging](#staging)
+          - [Importing the Cloud Provider Implementation](#importing-the-cloud-provider-implementation)
+          - [Build Targets](#build-targets)
+          - [Releases](#releases)
+          - [Migrating to a CCM Build](#migrating-to-a-ccm-build)
+          - [Flags, Service Accounts, etc](#flags,-service-accounts,-etc)
+          - [Deployment Scripts](#deployment-scripts)
+          - [KubeAdm](#kubeadm)
```
neolit123: `KubeAdm` should be lowercase `kubeadm` everywhere in the doc.

keps/sig-cloud-provider/0013-build-deploy-ccm.md
```diff
@@ -0,0 +1,496 @@
...
+own cloud provider owned repositories.
+
+Currently there are eight cloud provider implementations. To prevent this number from growing we have locked Kubernetes
+Core to the addition of any new dependencies. This means all new cloud providers have to implement all their pieces
+outside of the Core. However, the current cloud providers are still used by core Kubernetes components. In order to
+remove the dependencies, changes are required in the following areas:
+
+- [Kube Controller Manager](https://kubernetes.io/docs/reference/generated/kube-controller-manager/) - Track usages of [CMServer.CloudProvider](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-controller-manager/app/options/options.go)
+- [API Server](https://kubernetes.io/docs/reference/generated/kube-apiserver/) - Track usages of [ServerRunOptions.CloudProvider](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go)
+- [Kubelet](https://kubernetes.io/docs/reference/generated/kubelet/) - Track usages of [KubeletFlags.CloudProvider](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/options.go)
```
neolit123: `Kubelet` -> `kubelet` in the doc.
except if it's about API types such as `KubeletFlags`

keps/sig-cloud-provider/0013-build-deploy-ccm.md
cheftako: Fixed

keps/sig-cloud-provider/0013-build-deploy-ccm.md
cheftako: Fixed

frapposelli:
thumbs up 👍 from sig-vmware

PR #2281, opened 2018-06-18
Add scheduling framework design proposal

bsalamat:
Add scheduling framework design proposal. This is converted from [this Google Doc](https://docs.google.com/document/d/1vITsfsbNBJTUklxrxU6D_gp8E4FHVDKLGYVvex8lrWg). You must join [kubernetes-dev](https://groups.google.com/forum/#!forum/kubernetes-dev) group to gain access to the doc.


contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+### Reserve
+
+This is not a plugin point. At this point scheduler updates its cache by
+"reserving" a Node (partially or fully) for the Pod. In scheduler v1 this stage
+is called "assume". At this point, only the scheduler cache is updated to
+reflect that the Node is (partially) reserved for the Pod. The actual assignment
+of the Node to the Pod happens during the "Bind" phase. That is when the API
+server updates the Pod object with the Node information.
+
+### Admit
```
smarterclayton: Is the use of the term admit going to confuse users with admission control?

bsalamat: That's a good point. It could. I will find a better name or if you have a better name, please share.

Thanks for reviewing the doc, by the way.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+# DETAILED DESIGN
+
+## Bare bones of scheduling
+
+Pods that are not assigned to any node go to a scheduling queue and sorted by
+order specified by plugins (described [here](#heading=h.dggq0ff44y2y)). The
+scheduling framework picks the head of the queue and starts a **scheduling
+cycle** to schedule the pod. At the end of the cycle scheduler determines
+whether the pod is schedulable or not. If the pod is not schedulable, its status
+is updated and goes back to the scheduling queue. If the pod is schedulable (one
```
ravisantoshgudimetla: `goes back to the scheduling queue` or based on priority could it preempt one of the existing pods? Or did I miss something?

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+The main scheduling thread processes pods one by one and reserves nodes for
+them. The gang scheduling plugin at the admit stage is invoked for each pod.
+When it finds that the pod belongs to a gang, it checks the properties of the
+gang. If there are not enough members of the gang which are scheduled or in
+"wait" state, the plugin returns "wait". When the number reaches the desired
+value, all the Pods in wait state are approved and sent for binding.
+
+# OUT OF PROCESS PLUGINS
+
+Out of process plugins (OOPP) are called via JSON over an HTTP interface. In
```
ravisantoshgudimetla: Why JSON?

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+### Bind
+
+Once all pre-bind plugins return true, the Bind plugins are executed. Multiple
+plugins may be registered at this extension point. Each plugin may return true
+or false (or an error). If a plugin returns false, the next plugin will be
+called until a plugin returns true. Once a true is returned **the remaining
+plugins are skipped**. If any of the plugins returns an error or all of them
+return false, the Pod is rejected and sent back to the scheduling queue.
+
+### Post Bind
```
ravisantoshgudimetla: Curious if there is any use-case for this extension point

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+timeout expires, the pod is rejected and goes back to the scheduling queue.
+
+#### Approving a Pod binding
+
+While any plugin can receive the list of reserved Pod from the cache and approve
+them,  we expect only the "Admit" plugins to approve binding of reserved Pods
+that are in "waiting" state. Once a Pod is approved, it is sent to the Bind
+stage.
+
+### Reject
```
ravisantoshgudimetla: Why can't we combine this into admit extension point, when there is  a failure, we can undo all the activities done till now.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: The scheduling framework does not have preemption logic of its own. Preemption must be built as a plugin installed at "post filter".

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: The reason is that "admit" may have multiple plugins. They are called in order. Let's say the last one fails, while all the other plugins have already performed their tasks. There must be another extension point that allows those plugins to undo their operations when needed.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: If bind fails, certain things may need to be undone. For example, if we dynamically bind volumes to a node at "pre-bind" and then bind to that node fails, we probably want to detach the volumes.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: I am open to explore ideas here and don't have a strong opinion about one or the other. That said, there are a several reasons I think JSON might be a better choice here:

1. One cannot expect very high performance from OOPPs. If high performance is required, we highly recommend in-process plugins. If performance is not an issue, JSON should work fine.
1. Protobuf requires the addition of all the protobuf (and probably the rest of GRPC) libraries.
1. Webhooks are more aligned with other K8s extensions.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+at one of these extension points. They are called by the framework in the same
+order they are registered for each extension point.  
+In the following sections we describe each extension point in the same order
+they are called in a schedule cycle.
+
+![image](images/scheduling-framework-extensions.png)
+
+### Scheduling queue sort
+
+These plugins indicate how Pods should be sorted in the scheduling queue. A
```
losipiuk: Is there a mechanism which ensures that Pods further in the queue will not be starved by Pods with higher priority.
E.g. can we defend ourselves from situation where we pick a first not-scheduled Pod from sorted queue. Pass it through scheduling steps but at some point we decide to Reject it. IIUC it now gets back to the queue. And if it ends up (after resorting of queue) to be the first one in the queue again, the situation may repeat (possibly indefinitely).

Or maybe the resorting only happens after all the pods from the queue have been considered? I did not get that from the doc.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+## Plugin registration
+
+Plugin registration is done by providing an extension point and a function that
+should be called at that extension point. This step will be something like:
+
+<table>
+<thead>
+<tr>
+<th><p><pre>
+register("pre-filter", plugin.foo)
```
losipiuk: Will there be possibility to pass plugin specific parameters when plugin is registered?
Or plugin parameters will be read by plugin itself from scheduler config map?

If parameters are passed at registration time let's make sure that multiple instances of same plugin (but parametrized differently) can be registered under different names.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+The following picture shows the scheduling cycle of a Pod and the extension
+points that the scheduling framework exposes. In this picture "Filter" is
+equivalent to "Predicate" in scheduler v1 and "Scoring" is equivalent to
+"Priority function". Plugins are go functions. They are registered to be called
+at one of these extension points. They are called by the framework in the same
+order they are registered for each extension point.  
+In the following sections we describe each extension point in the same order
+they are called in a schedule cycle.
+
+![image](images/scheduling-framework-extensions.png)
```
k82cn: Following the discussion at google doc about preemption hooks, I think we need to clarify which part are provided by scheduling framework and which part are pluggable; for example, pending queue, cache, assuming, binding, preempting and so on.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+The above process is the same as what Kubernetes scheduler v1 does. Some of the
+essential features of scheduler v1, such as leader election, will also be
+transferred to the scheduling framework.  
+In the rest of this section we describe how various plugins are used to enrich
+this basic workflow. In this section, we focus on in-process plugins.
+Out-of-process plugins are discussed later in the doc.
+
+## Communication and statefulness of plugins
+
+The scheduling framework provides a library that plugins can use to pass
```
msau42: Is this where something like informers can be shared by multiple plugins?

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+of the Pod to this Node and reject the node if needed.
+
+![image](images/scheduling-framework-threads.png)
+
+### Reserve
+
+This is not a plugin point. At this point scheduler updates its cache by
+"reserving" a Node (partially or fully) for the Pod. In scheduler v1 this stage
+is called "assume". At this point, only the scheduler cache is updated to
+reflect that the Node is (partially) reserved for the Pod. The actual assignment
```
msau42: Other plugins may need to do their own reserve as well.  For example, for volumes, we want to mark PV objects as taken in the cache so that we don't choose those PVs again for other pods.

I suppose this could be done at the Admit phase in the plugin, but because Admit is done in a separate thread than the Filter/Scoring/Reserve, then you could have more rejections.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+routine. The same go routine runs "Bind" after these plugins when all of them
+return true.
+
+### Bind
+
+Once all pre-bind plugins return true, the Bind plugins are executed. Multiple
+plugins may be registered at this extension point. Each plugin may return true
+or false (or an error). If a plugin returns false, the next plugin will be
+called until a plugin returns true. Once a true is returned **the remaining
+plugins are skipped**. If any of the plugins returns an error or all of them
```
msau42: Is the idea that Pre-Bind is used by a plugin to bind non pod->node resources, and Bind is used to bind the Pod to the Node?

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+</th>
+</tr>
+</thead>
+<tbody>
+</tbody>
+</table>
+
+The details of the function signature will be provided later.
+
+## Extension points
```
msau42: Are there any requirements on the plugins regarding time limits?  For example, dynamically provisioning resources could take on the order of minutes.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+### Pre-filter
+
+These plugins are generally useful to check certain conditions that the cluster
+or the Pod must meet. These are also useful to perform pre-processing on the pod
+and store some information about the pod that can be used by other plugins.  
+The pod pointer is passed as an argument to these plugins. If any of these
+plugins return an error, the scheduling cycle is aborted.  
+These plugins are called serially in the same order registered.
+
+### Filter
```
msau42: Are there any plans on how to integrate this new framework with the predicate equivalence cache?

contributors/design-proposals/scheduling/scheduling-framework.md
msau42: Similarily, current assume cache has a TTL of 30 seconds, which will be too short for dynamic resource provisioning.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Bare bones of scheduling
+
+Pods that are not assigned to any node go to a scheduling queue and sorted by
+order specified by plugins (described [here](#heading=h.dggq0ff44y2y)). The
+scheduling framework picks the head of the queue and starts a **scheduling
+cycle** to schedule the pod. At the end of the cycle scheduler determines
+whether the pod is schedulable or not. If the pod is not schedulable, its status
+is updated and goes back to the scheduling queue. If the pod is schedulable (one
+or more nodes are found that can run the Pod), scoring process is started. The
```
adohe: IIUC this means `scheduling cycle` not include scoring process?

contributors/design-proposals/scheduling/scheduling-framework.md
adohe: I would think it as central cache, we could put information like `metadata` to the cache, thus multi plugins can access the meta.

contributors/design-proposals/scheduling/scheduling-framework.md
adohe: same doubt here.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+be registered at this point. Plugins registered at this point are called in
+order and the invocation continues as long as plugins return "equal". Once a
+plugin returns "greater" or "smaller" the invocation of these plugins are
+stopped.
+
+### Pre-filter
+
+These plugins are generally useful to check certain conditions that the cluster
+or the Pod must meet. These are also useful to perform pre-processing on the pod
+and store some information about the pod that can be used by other plugins.
```
adohe: metadata calculation fits well with this plugin type.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+behavior of scheduler v1 in placing Pods on nodes is preserved. This includes
+building plugins that replicate default predicate and priority functions of
+scheduler v1 and its binding mechanism, but scheduler extenders built for
+scheduler v1 won't be compatible with scheduler v2. Also, predicate and priority
+functions which are not enabled by default (such as service affinity) are not
+guaranteed to exist in scheduler v2.
+
+# DEVELOPMENT PLAN
+
+We will develop the scheduling framework as an incubator project in SIG
```
adohe: we have this project now?

contributors/design-proposals/scheduling/scheduling-framework.md
lichuqiang: +1

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -0,0 +1,406 @@
...
+resources to Nodes at the time of scheduling Pods as dynamic resource binding.  
+Dynamic resource binding has proven to be a challenge in Scheduler v1, because
+Scheduler v1 is not flexible enough to support various types of plugins at
+different phases of scheduling. As a result, binding of storage volumes is
+integrated in the scheduler code and some non-trivial changes are done to the
+scheduler extender to support dynamic binding of network GPUs.  
+The scheduling framework allows such dynamic bindings in a cleaner way. The main
+thread of scheduling framework process a pending Pod that requests a network
+resource and finds a node for the Pod and reserves the Pod. A dynamic resource
+binder plugin installed at "Pre-Bind" stage is invoked (in a separate thread).
```
lichuqiang: AFAIK, volume scheduling requires us to take volume details into consideration when making node chosen decision,
but seems we've already have a chosen node before "Pre-Bind" here. So, won't it be too late?
Or it turned out to be best-effort? i.e. try to bind to a chosen node, but not ensure it available before selecting a node.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: Our current scheduler (with Pod Priority and Preemption enabled) has a smarter scheduling queue that does not retry scheduling a Pod unconditionally after it has determined unschedulable. The scheduler stores a Pod in a separate structure and brings it back to the scheduling queue only if there is an event in the cluster that could make the Pod schedulable, for example, another Pod is terminated, .... This will effectively prevent a high priority Pod from hugging the front of the scheduling queue in most practical scenarios. The starvation scenario could still happen if the cluster has very high churn and many pods are terminated every second. In that case a high priority unschedulable pod may be retried many times per second as well. We are building some backoff mechanism to address this unlikely scenario as well. A similar approach should be used here too.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: Why does the plugin need to modify the scheduler cache? It can keep a separate data structure in its own package to track resources, etc. for its use.  In your case, the PV plugin can keep a set of taken PVs for its own use.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: Plugings may have many use-cases, but an obvious use-case for pre-bing is to bind cluster resources dynamically to nodes before a Pod is bound to the node.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: There are two scenarios:
1. a PV is already bound to a node and the Pod needs that PV: in this case a "filter" function must choose that node and filter out the rest of the nodes.
2. Pod requires a PV, but the PV is not bound to any node. Scheduler picks a node that satisfies other scheduling requirements of the Pod and then a pre-bind plugin is invoked to bind the PV to the chosen node. If the binding fails, the pre-bind plugin returns an error and the Pod goes back to the scheduling queue for another try. If binding succeeds, the Pod goes to the next stage and is bound to the node.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: no

contributors/design-proposals/scheduling/scheduling-framework.md
msau42: Yes, the plugin should keep its own cache, however, what I meant was that I think there needs to be a plugin-level Reserve call so that the plugin can update its cache before more pods are processed.  

Just like today where the scheduler pod cache is updated to account for that pod's resources on the node before it actually gets bound to a node, the volume plugin needs to update its PV cache to mark the PV as taken so that subsequent pods will not try to take the same one.

contributors/design-proposals/scheduling/scheduling-framework.md
lichuqiang: My main concern is for the scenario 2: for now, we take volume binding/provisioning things into consideration during predicates, and scheduler will trigger binding/provisioning for unbound pvcs on a pod, to ensure the volume can work on the chosen node.

By "satisfies other scheduling requirements of the Pod", you mean we'll skip this in the V2 framework? Then there will be a chance that the Pre-bind plugin cannot find available volume to bind/provision for the chosen node. Or you'd suggest the scheduler keep retrying scheduling, until it finds available volume for the chosen node that time?

Also, currently scheduling will fail if there are unbound pvcs on a pod, I guess it will not step into the "Pre-Bind" stage.

lichuqiang: Or the plugin of predicate still exists, and the main difference lies in the "Reserve"?
Currently we may experience multi scheduling processes if the volume binding/provision is slow. And with the new mechanism, we can manage the thing in a same scheduling cycle?

bsalamat: > By "satisfies other scheduling requirements of the Pod", you mean we'll skip this in the V2 framework?

No! I didn't mean that.

> there will be a chance that the Pre-bind plugin cannot find available volume to bind/provision for the chosen node. Or you'd suggest the scheduler keep retrying scheduling, until it finds available volume for the chosen node that time?

Yes, the scheduler will retry unschedulable pods and when there are volumes available, the pod may be scheduled.

> Also, currently scheduling will fail if there are unbound pvcs on a pod, I guess it will not step into the "Pre-Bind" stage.

Such pod could proceed to Pre-Bind and then binding can happen there, but as I mentioned above in my comment to Michelle, that logic has a potential race condition. So, we may want to consider a different design where reservation of a volume happens in earlier stages of a scheduling cycle.

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: The use-case make sense. I updated the proposal and added an extension point at "Reserve".

contributors/design-proposals/scheduling/scheduling-framework.md
bsalamat: What I think we should do is to provide an equivalence cache library in the framework and leave it up to the plugins to use it or not. In more details:
- The framework creates a structure similar to our equivalence cache.
- The framework provides an interface to interact with the eCache.
- Filter plugins can call a function to obtain a pointer to the cache and use its interface to store or retrieve its data. It will be up to the plugin to use the stored data or not.

contributors/design-proposals/scheduling/scheduling-framework.md
```diff
@@ -244,12 +244,16 @@ of the Pod to this Node and reject the node if needed.
...
-reflect that the Node is (partially) reserved for the Pod. The actual assignment
-of the Node to the Pod happens during the "Bind" phase. That is when the API
-server updates the Pod object with the Node information.
+At this point scheduler updates its cache by "reserving" a Node (partially or
+fully) for the Pod. In scheduler v1 this stage is called "assume".
+At this point, only the scheduler cache is updated to
+reflect that the Node is (partially) reserved for the Pod. The scheduling
+framework calls plugins registered at this extension points so that they get a
+chance to perform cache updates or other accounting activities. These plugins
+do not return any value.
```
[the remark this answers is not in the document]
bsalamat: I was thinking about returning error a few minutes ago. :) I think we should support returning error from all plugins. I should clarify this in the doc. Other than errors, Reserve plugins do not return values that change the scheduling decision.

contributors/design-proposals/scheduling/scheduling-framework.md
msau42: The main interaction between plugin and ecache I think would be the informer handlers to invalidate the predicates.

bsalamat: I agree that it is also an important part of the interaction of plugins and eCache. Generally, the framework should allow plugins to register their own handler at various object events, such as pod update, node delete, etc. The framework calls their handles at the events.
Invalidating eCache is one of many operations that a plugin may want to perform at any of the events. Similarly, plugins may want to invalidate their other internal caches or update them at object events.

msau42:
lgtm from volume scheduler

bsalamat:
If there is no more comments about this design, we should merge it. We can, of course, change it in the future if needed.

k82cn:
+1 to get this finalized :)

bsalamat:
I approve it then and as I said we can change it if needed.

/approve

contributors/design-proposals/scheduling/scheduling-framework.md
msau42: The latest design adds a "Reserve" phase that is similar to the current scheduler's "Assume" phase. I think this is sufficient for volume binding (along with predicates, priorities and bind phases)

msau42: That's correct, the current predicate we have today maps to the filter phase of the scheduler v2

PR #2287, opened 2018-06-19
KEP for garbage collection

janetkuo:
Expand Kubernetes garbage collection to handle common uses which are awkward today. We will add an object-lifecycle-respecting TTL mechanism to the garbage collector. We will also allow users to mark resource objects to be cleaned up by the garbage collector after those objects are no longer in use: either terminated, completed, or not referenced by other resource objects. 

This KEP covers existing problems and an overview of a proposed solution. Implementation details are discussed separately (also linked in KEP):
1. Object-lifecycle-respecting TTL mechanism: https://goo.gl/9AUQ3n
2. Garbage collect resources that finish execution: https://goo.gl/ws9t3b
3. Garbage collect resources that are not used by other resources: https://goo.gl/KeHPiQ

/assign @kow3ns @lavalamp @erictune 
/cc @kow3ns @lavalamp @erictune 
@liggitt @deads2k @smarterclayton @bgrant0607 

ref https://github.com/kubernetes/kubernetes/issues/22368, https://github.com/kubernetes/kubernetes/issues/28486, https://github.com/kubernetes/kubernetes/issues/64470


keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+  (see [#28486]).
+
+[resource quotas]: https://kubernetes.io/docs/concepts/policy/resource-quotas/
+[#28486]: https://github.com/kubernetes/kubernetes/issues/28486
+
+#### Non-generic TTL-based Garbage Collection
+
+Current TTL-based garbage collection has the following issues:
+
+* Not generic. Only supports Events.
```
liggitt: that's true for kubernetes resources... other apiservers using the apimachinery storage layer can make use of this today to set variable-length TTLs on objects based on fields in the object

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+and becomes useless. 
+
+[UID]: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids
+
+## Proposal
+
+The design has three main parts: 
+
+1. **Object-lifecycle-respecting TTL mechanism.** Provide observability before
+   deletion, and address the resource creation ordering problems. See [detailed
```
liggitt: > address the resource creation ordering problems

it wasn't clear to me how this addressed that... was it because the TTL gives you a window within which the owning objects can be created? what would we recommend someone wanting to create a set of interdependent objects without worrying about creation order set this to? a few seconds? a few minutes?

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+   design of TTL][].
+1. **Garbage collect resources that finish execution.** Delete a resource object
+   when it finishes execution (either terminates or completes). See [detailed
+   design of GC after use (part 1)][].
+1. **Garbage collect resources that are not used by other resources.** Extract
+   usage references of a resource object and delete the object when it is no
+   longer referenced. Garbage collector will count usage references and address
```
liggitt: inverting the graph direction garbage collection is responsible for has some significant implications:
* latency or disruption in the GC controller's informers means that new usages of a `GCIfUnused: true` resource are unobserved, and can incorrectly treat that resource as unused and delete it
* any inability to discover or reach aggregated APIs has to halt GC of `GCIfUnused: true` resources until resolved, since the undiscovered resources could contain usages that would prevent deletion

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: Yes. I should clarify that this point is specific to built-in apiserver.

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: The main problem I want to solve is the observability before deletion. We can solve the creation ordering problem with something more reliable, such as having GC controller to remember if it has ever observed the owner or not. 

If we don't want to encourage using TTLs to solve problems like creation order, we can make TTLs to be "GC after use" only. We can even combine TTL and GCIfUnused fields. For example, have a "durability" field, which can be `durable` (default, don't GC after use) or `ephemeral` (GC after not used for 2d).

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: GC controller can do a sanity check before deleting a resource. This check could be expensive, though. 

It'd be useful for GCIfUnused resources to specify the api group versions of its potential users. This would make queries less expensive and make unreachable aggregated API only affect GCIfUnused resources that cares about the API.

janetkuo:
Thanks @liggitt for the review. Great feedback.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+resource around for a while before garbage collection kicks in.
+
+Note that we will not use etcd TTL to support this feature, because etcd TTLs
+won’t respect other lifecycle guarantees (e.g. finalizers). This also allows the
+feature to be independent of etcd.
+
+### Clean Up Resources After Use
+
+A resource is no longer *in use* either when it finishes execution (such as a
+terminated Pod or a completed Job), or when it’s not referenced/used by other
```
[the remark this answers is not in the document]
janetkuo: CronJob only needs to see the jobs that haven't completed so that it can react based on its concurrency policy. It doesn't need to see completed jobs. 

The proposed GC will respect owner reference; therefore, Jobs owned by a CronJob won't be deleted by new GC after it completes. It's taken care of by CronJob controller using job history limit.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+## Proposal
+
+The design has three main parts: 
+
+1. **Object-lifecycle-respecting TTL mechanism.** Provide observability before
+   deletion, and address the resource creation ordering problems. See [detailed
+   design of TTL][].
+1. **Garbage collect resources that finish execution.** Delete a resource object
+   when it finishes execution (either terminates or completes). See [detailed
+   design of GC after use (part 1)][].
```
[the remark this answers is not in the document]
janetkuo: Totally agree with the advice.

In most cases "finishable" resources should be hooks. The Job used in the Minio chart is actually a hook -- it has hook related annotations. Finishable resources shouldn't be GC'd after use and their lifecycle should be managed by Helm. Helm provides cleanup after hook success/failure. 

I like the idea of adding hooks to `kubectl apply`. Users applying configs in bulk should follow the same advice for Charts.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+This is similar to [mark-and-sweep][] garbage collection. 
+
+This timestamp represents the last time this resource is needed and shouldn’t be
+garbage collected. It will be empty when the resource is still in use, and will
+only be set when this resource is no longer needed. This is important, so that
+we can avoid the need for frequent updates (like a heartbeat) which could cause
+performance and scalability issues. The timestamp only needs to be updated when
+a resource is either in use or not in use.
+
+Combining with TTL, users can specify the earliest safe time to clean up
```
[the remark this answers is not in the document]
janetkuo: Adding default ttl should be fine, as it'll only make GC take longer before deleting things. 

I think what you meant is making resources being cleaned up by default (`metadata.gcifunused = true`) with some ttl value? I'm worried that it would break people unexpectedly and there's no way to recover from delete. 

In the "Future Work - Namespace-wide Garbage Collection Policy" section in this KEP, I proposed allowing users to create a "clean up policy" object that will set a default cleanup policy for resources. We can also create a global policy resource or allowing cluster admins to use a flag to set a global default value.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+
+## Alternatives
+
+We have considered using owner references, instead of counting usage reference,
+to clean up resources after use. However, it has the following drawbacks:
+
+* Owner references of a resource may be excessively long. Resources like
+  ConfigMap can be referenced by Pods, ReplicaSets, and Deployments. If a
+  Deployment has 1k replicas, it will create 1k Pods and the ConfigMap needs to
+  add those 1k Pods to the ConfigMap’s owner references list. It will be easy
```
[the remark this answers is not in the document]
janetkuo: Configmaps can be "owned" by lots of pods (any pod that uses the configmap owns it), but pods will usually only be "owned" 1 replicaset.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+
+1. In the ConfigMap and Secret’s metadata, the user sets TTL to 1 hour and that
+   the ConfigMap and Secret should be cleaned up after they are no longer used
+   by other resources (Pods, ReplicaSets, Deployments, etc.)
+1. The user then deploys applications, including ConfigMaps and Secrets of those
+   applications, as usual. Users can create new ConfigMaps and Secrets when they
+   need to change the configuration of their applications. They don’t need to
+   manually check which ConfigMaps and Secrets are not used and can be cleaned
+   up. If the ConfigMaps and Secrets are shared among applications, other
+   applications can still use them before TTL expires.
```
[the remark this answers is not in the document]
janetkuo: Thanks that's a good example!

Actually I start thinking about combining `.metadata.gcttl` and `.metadata.gcifunused` into one field named `durability` (https://github.com/kubernetes/community/pull/2287#discussion_r197017961), as TTL seems less useful for owner references. An object's `durability` can be `durable`, `ephemeral`, and something else. Each `durability` comes with a sane TTL value.

keps/sig-api-machinery/0014-garbage-collection.md
lavalamp: To solve this problem, I suggested making the GC keep a per-group-version used/not used marker on objects opting in to this mechanism.

If the apiserver serving the group-version is down, it's true that GC can't observe new uses, but also true that people can't create new uses. So this leaves only a small window where someone adds a new use and then the apiserver goes down before GC observes it. This is likely preferable to blocking the whole GC (which could render the control plane inoperable) if an apiserver is down.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+
+Resources that are not terminated Pods, Events, or controllers history
+resources, can only be garbage collected with ownerReference-based garbage
+collection. 
+
+However, OwnerReference-based garbage collection can only handle the case when a
+resource has at least an owner. However, not all resources have owners (i.e.
+non-empty `.metadata.ownerReferences`). You can think of the resources that
+don't have owners as root nodes in a DAG (directed acyclic graph). Those
+resources without owners may never be deleted after they're no longer needed.
```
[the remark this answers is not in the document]
janetkuo: `kubectl apply --prune` requires users to manually manage those objects, i.e. think about whether a manifest can be removed or not.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+[detailed design of GC after use (part 2)]: https://goo.gl/KeHPiQ
+
+### Generic TTL-based Garbage Collection
+
+First, we need to clarify what TTL means in this context. TTL represents the
+earliest safe time when a resource can be garbage collected. 
+
+Generic TTL-based garbage collection will allow the user to set a TTL on each of
+Kubernetes resources. The TTL starts when a resource is eligible for garbage
+collection and stops when the resource is no longer eligible; the system will
```
crimsonfaith91: Does the TTL reset to zero when the resource becomes no longer eligible for garbage collection?

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+only garbage collect the resource after the TTL expires. This is useful when the
+user expects a resource to be needed intermittently and/or want to keep the
+resource around for a while before garbage collection kicks in.
+
+Note that we will not use etcd TTL to support this feature, because etcd TTLs
```
crimsonfaith91: It sounds to me that the proposed TTL GC is a generic version of the etcd TTL that works only on events. Is the proposed TTL GC able to replace etcd TTL in future? To make it a step further, I think the proposed TTL GC should have the long-run goal of unifying all GC flavors (OwnerReference, Capacity, and TTL) into one.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+#### Story 2: Secrets and ConfigMaps
+
+Secrets and ConfigMaps are used and referenced by other resources, such as Pods,
+ReplicaSets, and Deployments. An advised pattern is to treat application
+configurations as immutable so that application revisions are intact and
+rollbacks are reliable. When the users deploy their applications, they treat
+Secrets and ConfigMaps as immutable resources, and therefore Secrets and
+ConfigMaps are frequently created and deployed to a Kubernetes cluster, but
+never updated or cleaned up. Secrets and ConfigMaps accumulate over time, which
+could cause master component downtime.
```
crimsonfaith91: +1 for having immutable Secrets, but I am neutral on ConfigMaps. I think ConfigMaps should provide enough versatility for minor config or script injection changes.

keps/sig-api-machinery/0014-garbage-collection.md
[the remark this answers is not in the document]
crimsonfaith91: I think the notion of `durability` is hard to be achieved because different applications will have different sane TTL value. +1 for usage reference to handle the case of scaling down to zero.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+
+This feature will also be feature gated.
+
+## Graduation Criteria
+
+This can be promoted to beta when it satisfies users' need for cleaning up
+resources after use, without regressions.
+
+This will be promoted to GA once it's gone a sufficient amount of time as beta
+with no changes.
```
crimsonfaith91: +1 on this. For a GC to become GA, the best yardstick is to have it working for a predefined period of time (e.g., 6 months) without encountering any substantial issue. I am not sure whether we should start thinking about unifying all GC flavors into one when the proposed TTL GC is beta or GA.

keps/sig-api-machinery/0014-garbage-collection.md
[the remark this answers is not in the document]
crimsonfaith91: This brings upon the question that should we enforce the requirement that pods should be owned by at most (0 or 1) parent at one time? This one-to-one mapping (if the pod is owned) may contribute to a GC optimization in future.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+* Be generic and configurable.
+* Remove resources based on their last used time, instead of creation time.
+* Be coherent with other existing garbage collection mechanism.
+
+### Namespace-wide Garbage Collection Policy
+
+We can implement a garbage collection policy that applies to all resources in
+the same namespace by default for easier garbage collection policy management.
+
+### Move Event TTL out of etcd TTL
```
crimsonfaith91: +1 on this.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+We may consider moving Event TTL out of etcd TTL, and use the TTL mechanism in
+garbage collector for Events. 
+
+### Make Pod GC and Event GC Work Better with Garbage Collector
+
+Pod GC and Event GC don't respect OwnerReference-based garbage collector today.
+The biggest problem caused by Pod GC is that it breaks Job controller
+([#28486][]).
+
+We may move Pod GC and Event GC into garbage collector, or deprecate them and
```
crimsonfaith91: +1 on deprecating and deleting them entirely in the long run (i.e., not agree on moving Pod GC and Event GC into the garbage collector). Ideally, we want to maintain as small number of GC flavors as possible.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+
+### Move deletionGracePeriodSeconds from metadata to Pod spec
+
+`deletionGracePeriodSeconds` is currently in `.metadata` of all resources,
+although it only applies to Pods. This is confusing to the users.
+
+Because users cannot set or modify this field on a non-Pod resource (the field
+will be silently dropped), it will be compatible to move it outside of
+`.metadata` of resources other than Pod. This field can be part of a
+Pod-specific `.metadata`, or can even move this field to Pod spec in the future
```
crimsonfaith91: `.metadata` attributes across various types of resources should be as similar as possible. Hence, I think it makes more sense to move `deletionGracePeriodSeconds` field to pod spec in future.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+no longer used by another other resources, and a Pod without an owner gets
+deleted 1 hour after it terminates.
+
+[mark-and-sweep]: https://www.geeksforgeeks.org/mark-and-sweep-garbage-collection-algorithm/
+
+### Summary
+
+In summary, resource garbage collection with TTL is illustrated below:
+
+![Garbage Collection Diagram](garbage-collection-diagram.png)
```
crimsonfaith91: In the bottom right corner of the diagram, do you mean: "Finalizers are gone `or` Pod's deletion grace period has passed"? IIUC, finalizers are different from deletion grace period (e.g., I added a finalizer intended to be removed manually, which is irrelevant to deletion grace period).

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+  add those 1k Pods to the ConfigMap’s owner references list. It will be easy
+  for the ConfigMap to hit Kubernetes resource object size limit. After the
+    limit is hit, the resource object cannot be updated anymore. 
+* Using owner references to implement this feature will make the feature harder
+  to be safely disabled after it’s enabled, because that requires all the added
+  owner references to be removed immediately before the garbage collector starts
+  reacting to it. It will cause resource objects to be cleaned up unexpectedly. 
+* Using owner references is conceptually more complex.
+
+We have also considered implementing a cluster wide cleanup policy. However we
```
crimsonfaith91: I believe cluster wide cleanup policy has some value as cluster admins can technically be users too.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+
+## Implementation History
+
+TBD
+
+## Future Work
+
+### Generic Capacity-based Garbage Collection
+
+We may potentially want to support generic capacity-based garbage collection.
```
crimsonfaith91: I think the main reason behind supporting capacity-based GC is the users specify too-big TTL values for TTL GC. If so, perhaps we can look at embedding an automated version of capacity-based GC into the TTL garbage collector. In other words, the TTL garbage collector deletes resources when the quota is reached _regardless_ whether TTL value has expired of the resources.

For this to be implemented with correct semantics, the TTL garbage collector should auto-delete the resources with highest TTL value.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+and ConfigMaps, are often treated as _immutable_ resources -- they are
+frequently created and not updated by users. We also have discussed making such
+resources immutable ([#10179][]).
+
+However, it's difficult for the users to clean up those resources automatically,
+because those resources don't have owners. Without a way to clean up those
+frequently created, short-lived resources, those resources can accumulate and
+overload a Kubernetes cluster very easily. Even if we can avoid the overload
+issue by implementing a cluster-wide (global) resource quota, users won't be
+able to create new resources without cleaning up old ones first.
```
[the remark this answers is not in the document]
janetkuo: TTL is complementary to quota. Quota only stops you from creating more than allowed capacity, but you cannot move forward without deleting objects. Enhanced GC allows you to clean up objects easily and even avoid hitting resource quota. 

A global parent only allows those resource objects to be gone with the parent. However, we'd like to clean up objects when they're not needed. One of the child resources isn't needed doesn't imply that parent object isn't needed.

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: How we should handle Event TTL is still under discussion. TTL GC won't replace existing Event TTL mechanism in the first (alpha) release.

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: The downside of updating ConfigMaps is that it doesn't work well with rolling update features for workloads APIs. You don't have control over when a ConfigMap change will be updated to a Pod, and you cannot roll back.

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: Yes, but it has to be v2 Pod for compatibility.

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: Yes they're different, and that's why I separated them. Pods can also have finalizers on them. I'm not sure if I understand your comment here. Would you clarify?

keps/sig-api-machinery/0014-garbage-collection.md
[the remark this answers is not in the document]
tnozicka: If you set default GC delay by gcttl will it affect cascading deletion?

keps/sig-api-machinery/0014-garbage-collection.md
tnozicka: We were/are actually thinking about creating a top level object to allow for ConfigMap/Secret snapshoting and also triggering rollouts. (It would also have a revisionHistoryLimit and be bound to particular RS or ControllerRevision.)    /cc @mfojtik 

I don't think the Secret/ConfigMap should be immutable per type, but rather per instance - like having metadata.immutable=true as there are always different use cases across the ecosystem.

keps/sig-api-machinery/0014-garbage-collection.md
crimsonfaith91: sgtm, thanks!

keps/sig-api-machinery/0014-garbage-collection.md
crimsonfaith91: +1 on @tnozicka 's suggestion on creating a top level object for managing ConfigMap/Secret snapshots and rollouts.

If we are going to introduce a field indicating immutability preference, it may be better to add the field to a standalone `.immutability` field outside of `.metadata`.

keps/sig-api-machinery/0014-garbage-collection.md
crimsonfaith91: Oh, I meant it should be `or`, not `and`.

janetkuo:
@tnozicka thanks for the review! I replied to your comments above; not sure why GitHub UI doesn't show my replies at the bottom. :confused:

deads2k:
Inverting the direction of GC references means that inspection of a resource is insufficient to determine whether that resource can be deleted.  Instead, the current state of every resource must be known in order to make that decision.

There is no reasonable timeout that will be sufficient.  For instance, an aggregated API server can go down for an extended periods of time, during which it would not be safe to GC any object.  The direction of our current links ensure that such a situation does not freeze unaffected objects.  The new design proposed here can't make a safe decision in those conditions.

I don't see why we would trade a design that is not vulnerable to such a problem (what we have today) for one where we do have that problem and have to build various and numerous compensation methods to avoid those repercussions.

Is there a separate proposal just for the TTL aspect?  A TTL that respects finalizers is a thing I could see being generally useful.

janetkuo:
@deads2k thank you for the review! 

Current owner references GC is not enough for cleaning up unused objects. We can implement "clean up when unused" with owner references by treating each "user object" as "owner", but that would require "GC when there's no owners/users", which is not currently supported. I also added a diagram ([link](https://docs.google.com/document/d/1FNsGTxae0agzzqdQcQu7BwxcYjo92GCoCuzHiwT9Bks/edit#heading=h.g7eaxr7xpe7f)) to compare owner reference and usage reference: 
public garbage collect resources after use 2 2f2_ usage references

> For instance, an aggregated API server can go down for an extended periods of time, during which it would not be safe to GC any object.

Only the object that enables "clean up when unused" will be affected. If GC controller keeps track of the user objects of a used object, GC controller won't need to pause when there's no user objects behind the aggregated API server. 

> Is there a separate proposal just for the TTL aspect? A TTL that respects finalizers is a thing I could see being generally useful.

Yes, I split the KEP into 3 parts so that they can be discussed and implemented separately. 

1. The TTL design doc is here: https://goo.gl/9AUQ3n
2. The design for GC after finished: https://goo.gl/ws9t3b - this is also useful and doesn't introduce new GC references. 
3. The design for GC after unused: https://goo.gl/KeHPiQ - this seems to be much more controversial

Would it help if I make "3" its own KEP and don't mention it in this KEP? Alternatively, we can also make this KEP a high level overview of the problems we'd like to solve, and leave the design details to the design docs I linked above.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+## Motivation
+
+Make it easy for users to clean up frequently-created, short-lived Kubernetes
+resources objects.
+
+One common use case is [ConfigMaps][] and [Secrets][]. The users roll out
+ConfigMaps and Secrets by creating new ConfigMaps or Secrets and updating the
+references to those ConfigMaps or Secrets in containers ([#22368][]). The old
+ConfigMaps and Secrets, if not referenced by any container, are no longer
+needed.
```
mfojtik: ConfigMaps and secrets can be consumed by other means than containers right? You can have an operator that is reading the content of an configMap or secret (can be a CA or something)...

mfojtik: (also you can reference configMap via downward api)

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+
+One common use case is [ConfigMaps][] and [Secrets][]. The users roll out
+ConfigMaps and Secrets by creating new ConfigMaps or Secrets and updating the
+references to those ConfigMaps or Secrets in containers ([#22368][]). The old
+ConfigMaps and Secrets, if not referenced by any container, are no longer
+needed. 
+
+Another common use case is completed [Jobs][]. Users keep creating Jobs to run
+some batch tasks, and those Jobs, once complete, are no longer needed in the
+system.
```
mfojtik: @soltysh might prove me wrong, but I thought there is something like revisionHistoryLimit for Jobs which will cause the old jobs to be deleted by the jobs controller.

janetkuo: Here I'm talking about the Jobs that aren't managed by CronJobs. Some users create Jobs directly with their own tools.

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: Resource objects, including ConfigMaps/Secrets, can decide whether they want to use this feature (GC after unused) or not. 

Referencing via downward api is also covered by this proposal. The schema information can be encoded in OpenAPI spec. If the operator uses an API object to read/reference a ConfigMap/Secret, the usage can be discovered in the OpenAPI spec. If the operator simply reads/gets the ConfigMap/Secret, I would imagine that such ConfigMap/Secret won't be massively created or short-lived but be updated directly. In this case, they don't need GC after unused.

deads2k:
> Would it help if I make "3" its own KEP and don't mention it in this KEP? Alternatively, we can also make this KEP a high level overview of the problems we'd like to solve, and leave the design details to the design docs I linked above

I think that 3 is the most controversial.  I would suggest separate KEPs for separate features.

A generic TTL (1) that mirrors etcd TTL and respects finalizers is a thing that I can see being broadly useful.  Without inverted GC links (3), I don't think I see why I'd tie it to GC, but an etcd-like TTL seems pretty useful.

Deleting finished resources (2) doesn't seem related to GC to me.  Given an object, a separate controller could watch for finished things (I think I'd suggest using a condition) without any ownerRefs.  Without inverted GC links (3), I don't think I see why I'd tie it to GC, but a "clean up completed things" may be useful.  I suggest a separate KEP this too.  Removing resources without express user permission is a pretty aggressive thing to do so I don't expect controllers to set it.

janetkuo:
(1) and (2) ties to GC because an object is only safe to be cleaned up after all GC mechanisms agree it's safe to be cleaned up. If an object's owners still exist, it shouldn't be cleaned up even after it finishes and TTL expires. 

> Removing resources without express user permission is a pretty aggressive thing to do so I don't expect controllers to set it.

Users can decide if they want to clean up a completed thing or not (opt-in/out in a field of `.metadata`). GC controller then provides a generic way to clean up finished resources with TTL and respects owner references and finalizers.

deads2k:
> (1) and (2) ties to GC because an object is only safe to be cleaned up after all GC mechanisms agree it's safe to be cleaned up. If an object's owners still exist, it shouldn't be cleaned up even after it finishes and TTL expires.

A TTL is a general machinery item, but if a TTL is set, then GC doesn't enter into it until the item is deleted.  Same as an etcd TTL today like we use for events.  The only reason it seemed related if links are inverted is because you need a mechanism to avoid "early" reaping.  If you don't invert links, you don't have the problem.

Similarly, for deleted resources.  GC doesn't get involved until the item is actually deleted.  Deciding to delete the item belongs to something else.  The only reason it seemed related is that if links are inverted you need a mechanism to build the entire graph and chase the links.  if you don't invert the links, you don't have the problem because the single object is all the context you need.

janetkuo:
The use cases I want to address is cleaning up frequent-created, short-lived resources in a generic way. That includes the resources that already finish (2) or are unused (3). Both use cases can benefit from TTL (1) to allow some delays before cleanup actually happens. We also want to avoid early reaping for finished resource objects, for example, after a Job completes users need some time to collect results before it can be deleted. If it doesn't respect owner references, it'd be confusing to users. Similarly, if user declares a TTL, GC controller shouldn't delete an object earlier, even when the object's owners are deleted. How might we support the aforementioned use cases?

deads2k:
> The use cases I want to address is cleaning up frequently-created, short-lived resource objects in a generic way. That includes the resources that already finish (2) or are unused (3). Both use cases can benefit from TTL (1) to allow some delays before cleanup actually happens. We also want to avoid early reaping for finished resource objects, for example, after a Job completes users need some time to collect results before it can be deleted. If it doesn't respect owner references, it'd be confusing to users. Similarly, if user declares a TTL, GC controller shouldn't delete an object earlier, even when the object's owners are deleted. How might we support the aforementioned use cases?

You've described a new TTL controller which can delete a resource after a certain time and what amounts to a graceful deletion controller that gives a separate entity a deadline to do work after a deletion timestamp set and removes a finalizer after that time.  If you need a guarantee, not grace you've described a new finalizer for your type.

Neither of those things involve GC which serves the narrow purpose of traversing links to reap dead children.

keps/sig-api-machinery/0014-garbage-collection.md
```diff
@@ -0,0 +1,555 @@
...
+collector. The generic garbage collector deletes Kubernetes objects that once
+had an owner, but no longer have an owner. A Kubernetes object defines its owner
+by adding its owner to its own  `.metadata.ownerReferences` field. You can think
+of `.metadata.ownerReferences` as a list of pointers to a resource's parents
+(owners). 
+
+For example, when a Deployment owns a ReplicaSet, that ReplicaSet's
+`.metadata.ownerReferences` will include the Deployment. If the generic garbage
+collector finds that the Deployment doesn't exist in the system anymore, and if
+the Deployment is the ReplicaSet's only owner, it will delete the ReplicaSet.
```
krmayankk: what about the case when the selector is changed. Currently we have a case where because of selector changes, all of our olds Pods became orphaned. I have the following questions:-
- The DaemonSet's selector changed,which orphaned the old pods and DaemonSet created new pods. I saw that the ownerReferences of the old Pods doesnt exist any more.  My question is who removed those ownerReferences ?
- Does your proposal cover deleting such orphans whose parents are gone because of selector change ?
If we didnt remove the ownerReferences of these old Pods , but changed the uid of the owner since the selector of the owner change, we could achieve this. Thoughts @janetkuo ?

kargakis:
Selector changes are invalid in apps/v1

keps/sig-api-machinery/0014-garbage-collection.md
janetkuo: > Currently we have a case where because of selector changes, all of our olds Pods became orphaned. 

That is exactly why we made selector immutable in apps/v1. Changing selector of a workload controller is dangerous, and this specific use case is not something we want to support.

krmayankk: thanks @janetkuo  i am fine with this. But why when the selector changed, did the orphan'd pods loose their ownerReferences ?

janetkuo: ControllerReferences (OwnerReferences with controller==true) of Pods are updated automatically by workload controllers, to make sure that no two controllers are fighting against each others controlling the same group of Pods (think of it as a lock). When a controller no longer controls a group of Pods (selector no longer matches), it removes itself from the Pods' ControllerReferences to release the lock. 

Label selectors are used for identifying relationships between controllers and controllee. When a controller's selector is changed to something that doesn't match its controllees' labels, it has the same effect as orphaning the controllees by changing their labels, as the controllers finds controllees through "label/selector matches". 

If you really want to change the selector of a controller, the recommended way is to just create a new one with the new selector / template labels, and delete the old one.

krmayankk: Thanks @Janetkuo for helping understand. the intention is not to change the selectors . We changed it by accident but just be able to properly reason about its behavior . How does the controller know that a pod’s label no longer matches it’s selector . The controller can know which pods matches its label and rest everything else does not match its label . So does the controlller actually goes and looks at all pods that didn’t match its selector and see if it has an owner reference pointing to it and then Remove it self from those pods? Isn’t that too costly? Is the selector immutable in apps/v1 for all workload controllers ? Do you see any issues with GC always deleting a pod which doesn’t have any ownerreferneces or if it’s owner doesn’t exist . Why does t have to be triggered by owners deletion ? Sorry lot of questions :-)

janetkuo: > So does the controlller actually goes and looks at all pods that didn’t match its selector and see if it has an owner reference pointing to it and then Remove it self from those pods? 

This is correct. 

> Isn’t that too costly? 

Most controllers use informer framework, so instead of listing all controllees (e.g. pods), the controller watches them and acts upon their changes. So it's not too costly. 

> Do you see any issues with GC always deleting a pod which doesn’t have any ownerreferneces or if it’s owner doesn’t exist. Why does t have to be triggered by owners deletion ?

GC doesn't delete things that don't have ownerRefs. It deletes things whose owners are all gone. Think of ownerRefs as a signal for cascading deletion: the owner wants its dependents to be gone with it; if there are more than one owners, the GC should only happen after all owners are deleted. Note that if an item is removed from a dependent's ownerRefs list, that item is no longer considered the owner of the dependent.

mattfarina:
/sig apps

Since it's listed as a participating sig

keps/sig-api-machinery/0014-garbage-collection.md
mattfarina: @janetkuo Just to verify, it's opt-in to non-referenced ConfigMaps/Secrets being GC'd if not attached to a container rather than opt-out. Right?

justaugustus:

justaugustus:

/close

PR #2290, opened 2018-06-20
RuntimeClass KEP

tallclair:
As [previously discussed](https://docs.google.com/document/d/1mpjPTZkoeFV3tG59149KaIuI8LhIE-cekjFq8rRmNoQ/preview) in SIG-Node, here is the initial KEP for `RuntimeClass`.

This initial proposal excludes 2 important pieces: the supported features API, and the RuntimeController admission controller. My intent is to followup with another PR to add those, but let me know if those should be included here first.

/kind design
/sig node
/sig architecture

### DIFF
--- a/keps/sig-node/0014-runtime-class.md
+++ b/keps/sig-node/0014-runtime-class.md
@@ -106,7 +106,7 @@ iteration:
 - As a cluster operator, I want to provide multiple runtime options to support a wide variety of
   workloads. Examples include native linux containers, "sandboxed" containers, and windows
   containers.
-- As a cluster operator, I want to provide stable rolling upgrades or runtimes. For
+- As a cluster operator, I want to provide stable rolling upgrades of runtimes. For
   example, rolling out an update with backwards incompatible changes or previously unsupported
   features.
 - As an application developer, I want to select the runtime that best fits my workload.
@@ -222,10 +222,6 @@ Examples of runtime parameters include:
 // way of surfacing behaviors that were previously implicit in docker but not
 // well defined in Kubernetes.
 "default-capabilities": "[]"
-
-// Set the infrastructure container image. More generally, configuring options
-// not otherwise exposed in the Kubrenetes API.
-"infra-image": "example.com/self-monitoring-pod:v2"
 ```
 
 _Note: RuntimeClass is intended to be consumed by a single runtime implementation, so parameter keys
--- a/keps/sig-node/0014-runtime-class.md
+++ b/keps/sig-node/0014-runtime-class.md
@@ -255,7 +255,9 @@ CRI implementation, but it should be cached if needed for subsequent calls.
 If the RuntimeClass cannot be resolved (e.g. doesn't exist) at Pod creation, then the request will
 be rejected in admission (controller to be detailed in a following update). If the RuntimeClass
 cannot be resolved by the Kubelet when `RunPodSandbox` should be called, then the Kubelet will fail
-the Pod. The admission check on a replica recreation will prevent the scheduler from thrashing.
+the Pod. The admission check on a replica recreation will prevent the scheduler from thrashing. If
+the `RuntimeHandler` is not recognized by the CRI implementation, then `RunPodSandbox` will return
+an error.
 
 [RunPodSandboxRequest]: https://github.com/kubernetes/kubernetes/blob/b05a61e299777c2030fbcf27a396aff21b35f01b/pkg/kubelet/apis/cri/runtime/v1alpha2/api.proto#L344
 
@@ -294,6 +296,9 @@ Alpha:
 
 - Everything described in the current proposal
 - Supported features and associated admission controller
+- [CRI validation test][cri-validation]
+
+[cri-validation]: https://github.com/kubernetes-incubator/cri-tools/blob/master/docs/validation.md
 
 Beta:
 
--- a/keps/sig-node/0014-runtime-class.md
+++ b/keps/sig-node/0014-runtime-class.md
@@ -141,7 +141,7 @@ controller for performing validation & defaulting.
 
 There are 3 categories of fields that may be included in the `RuntimeClassSpec`:
 
-1. **Runtime parameters** - These fields control the behavior of the runtime in some way,
+1. **Runtime setup** - These fields control the behavior of the runtime in some way,
    including configuring arbitrary options in the runtime (`Parameters`), and eventually specifying
    resource overhead. This is the focus of this initial proposal.
 2. **Compatibility validation** - These fields declare support for various features, and pods are
@@ -200,7 +200,55 @@ const (
 An unspecified RuntimeClassName `""` is equivalent to the `legacy` RuntimeClass, though the field is
 not defaulted to `legacy` (to leave room for configurable defaults in a future update).
 
-#### CRI API Specification
+#### Runtime Parameters
+
+Runtime parameters are provided as a mechanism for passing non-portable runtime-specific arguments
+that are generally static for the RuntimeClass, but dynamically interpreted by the runtime. They
+provide a way of specifying options that are tightly coupled with the class of runtime, but should
+not be used to enumerate general pod options as this would lead to a combinatorial explosion of
+RuntimeClass definitions.
+
+Examples of runtime parameters include:
+
+```
+// Decide which underlying OCI runtime should be used. This replaces the
+// annotations used by CRI-o, containerd & frakti to select between sandboxed
+// runtimes.
+"oci-runtime": "runsc"
+
+// Change the default set of capabilities used by the runtime. This is a good
+// way of surfacing behaviors that were previously implicit in docker but not
+// well defined in Kubernetes.
+"default-capabilities": "[]"
+
+// Set the infrastructure container image. More generally, configuring options
+// not otherwise exposed in the Kubrenetes API.
+"infra-image": "example.com/self-monitoring-pod:v2"
+```
+
+_Note: RuntimeClass is intended to be consumed by a single runtime implementation, so parameter keys
+do not need to be "namespaced"._
+
+In contrast, the following are examples of parameters that are better left to annotations or other
+extension mechanisms:
+
+```
+// Configuring the smack LSM for a runtime that adds support. This is an example
+// of a feature that will typically have multiple valid configurations, and is
+// better left to a pod-scoped mechanism like annotations.
+"smack-profile": "foo-bar"
+
+// Propagate Kubelet version information to the runtime. Tight coupling of a
+// RuntimeClass to nodes should generally be avoided. This will create churn
+// during upgrades.
+"kubelet-version": "1.2.3"
+```
+
+Minimal validation is done on parameters (copied from StorageClass):
+
+1. Keys must not be empty.
+2. A maximum of 512 parameters may be specified.
+3. Total size of keys+values must be less than 256K
 
 The runtime parameters are passed to the CRI as part of the `RunPodSandboxRequest`:
 
--- a/keps/sig-node/0014-runtime-class.md
+++ b/keps/sig-node/0014-runtime-class.md
@@ -40,8 +40,7 @@ status: provisional
 
 `RuntimeClass` is a new cluster-scoped resource that surfaces container runtime properties to the
 control plane. RuntimeClasses are assigned to pods through a `runtimeClass` field on the
-`PodSpec`. This provides a new mechanism for supporting multiple runtimes in the cluster and on the
-nodes.
+`PodSpec`. This provides a new mechanism for supporting multiple runtimes in a cluster and/or node.
 
 ## Motivation
 
@@ -107,7 +106,7 @@ iteration:
 - As a cluster operator, I want to provide multiple runtime options to support a wide variety of
   workloads. Examples include native linux containers, "sandboxed" containers, and windows
   containers.
-- As a cluster operator, I want to provide stable rolling upgrades of nodes or runtimes. For
+- As a cluster operator, I want to provide stable rolling upgrades or runtimes. For
   example, rolling out an update with backwards incompatible changes or previously unsupported
   features.
 - As an application developer, I want to select the runtime that best fits my workload.
@@ -355,7 +354,7 @@ Beta:
   limited to linux. As we build out Windows container support, we'll need to add windows-specific
   features as well.
 - Host namespaces (Network,PID,IPC) may not be supported by virtualization-based runtimes
-  (e.g. Kata-containers & visor).
+  (e.g. Kata-containers & gVisor).
 - Per-pod and Per-container resource overhead varies by runtime.
 - Device support (e.g. GPUs) varies wildly by runtime & nodes.
 - Supported volume types varies by node - it remains TBD whether this information belongs in
--- a/keps/sig-node/0014-runtime-class.md
+++ b/keps/sig-node/0014-runtime-class.md
@@ -1,0 +1,311 @@
+---
+kep-number: 14
+title: Runtime Class
+authors:
+  - "@tallclair"
+owning-sig: sig-node
+participating-sigs:
+  - sig-architecture
+reviewers:
+  - TBD
+approvers:
+  - TBD
+editor: TBD
+creation-date: 2018-06-19
+status: provisional
+---
+
+# Runtime Class
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Motivation](#motivation)
+  * [Goals](#goals)
+  * [Non\-Goals](#non-goals)
+  * [User Stories](#user-stories)
+* [Proposal](#proposal)
+  * [API](#api)
+    * [API Specification](#api-specification)
+    * [CRI API Specification](#cri-api-specification)
+  * [Versioning, Updates, and Rollouts](#versioning-updates-and-rollouts)
+  * [Implementation Details](#implementation-details)
+  * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+* [Appendix](#appendix)
+  * [Examples of runtime variation](#examples-of-runtime-variation)
+
+## Summary
+
+`RuntimeClass` is a new cluster-scoped resource that surfaces container runtime properties to the
+control plane. RuntimeClasses are assigned to pods through a `runtimeClass` field on the
+`PodSpec`. This provides a new mechanism for supporting multiple runtimes in the cluster and on the
+nodes.
+
+## Motivation
+
+RuntimeClasses address two distinct problems: surfacing support and configuration of optional
+features, and supporting multiple runtimes.
+
+Today's runtimes have varying support for features that are expressible through the PodSpec (see
+[examples of runtime variation](#examples-of-runtime-variation) below). In some cases such as
+SELinux, features are silently ignored when they aren't supported. This is problematic from a
+security perspective since it makes it difficult to audit actual runtime options. Other features
+such as AppArmor are checked for support, but on failure the pod is just held in a "Blocked" state
+to avoid the churn of recreating the rejected pod. Still other features maybe be applied in slightly
+different configurations, as is the case with capabilities today (the default set is not
+well-defined), and `privileged` containers. By raising these properties to the control plane errors
+can be surfaced much earlier, and configuration can be made explicit.
+
+The second problem that RuntimeClass addresses is supporting multiple runtimes in a cluster, or even
+on the same node. [Sandboxes][] are the primary motivator for this right now, with both Kata
+containers and gVisor looking to integrate with Kubernetes. Other runtime models such as Windows
+containers or even remote runtimes will also require support in the future. RuntimeClass provides a
+way to configure different runtimes in the cluster, surface their properties (both to the cluster &
+the user), and provides a mechanism for pods to select a runtime.
+
+[Sandboxes]: https://docs.google.com/document/d/1QQ5u1RBDLXWvC8K3pscTtTRThsOeBSts_imYEoRyw8A/edit
+
+### Goals
+
+- Provide a mechanism for surfacing container runtime properties to the control plane
+- Support multiple runtimes per-cluster, and provide a mechanism for users to select the desired
+  runtime
+
+### Non-Goals
+
+- RuntimeClass is NOT a general policy mechanism
+- RuntimeClass is NOT "NodeClass". Although different nodes may run different runtimes, in general
+  RuntimeClass should not be a cross product of runtime properties and node properties.
+
+The following goals are out-of-scope for the initial implementation, but may be explored in a future
+iteration:
+
+- Automatic runtime or feature discovery - initially RuntimeClasses are manually defined (by the
+  cluster admin or provider), and are asserted to be an accurate representation of the runtime.
+- Scheduling in heterogeneous clusters - it is possible to operate a heterogeneous cluster
+  (different runtime configurations on different nodes) through scheduling primitives like
+  `NodeAffinity` and `Taints+Tolerations`, but the user is responsible for setting these up and
+  automatic runtime-aware scheduling is out-of-scope.
+- Define standardized or conformant runtime classes - although I would like to declare some
+  predefined RuntimeClasses with specific properties, doing so is out-of-scope for this initial KEP.
+- [Pod Overhead][] - Although RuntimeClass is likely to be the configuration mechanism of choice,
+  the details of how pod resource overhead will be implemented is out of scope for this KEP.
+- Provide a mechanism to dynamically register or provision additional runtimes.
+- Requiring specific RuntimeClasses according to policy. This should be addressed by other
+  cluster-level policy mechanisms, such as PodSecurityPolicy.
+- "Fitting" a RuntimeClass to pod requirements - In other words, specifying runtime properties and
+  letting the system match an appropriate RuntimeClass, rather than explicitly assigning a
+  RuntimeClass by name. This approach can increase portability, but can be added seamlessly in a
+  future iteration.
+
+[Pod Overhead]: https://docs.google.com/document/d/1EJKT4gyl58-kzt2bnwkv08MIUZ6lkDpXcxkHqCvvAp4/edit
+
+### User Stories
+
+- As a cluster operator, I want to provide multiple runtime options to support a wide variety of
+  workloads. Examples include native linux containers, "sandboxed" containers, and windows
+  containers.
+- As a cluster operator, I want to provide stable rolling upgrades of nodes.
+- As an application developer, I want to select the runtime that best fits my workload.
+- As an application developer, I don't want to study the nitty-gritty details of different runtime
+  implementations.
+- As an application developer, I want my application to be portable across clusters that use similar
+  but different variants of a "class" of runtimes.
+- As an application operator, I want to know precisely how the application is running, and which
+  features are being used.
+- As an application operator, I want pods to fail fast when trying to use an optional feature that
+  isn't supported by the chosen runtime.
+- As a security operator, I want to ensure that pods are running with the security features they
+  claim.
+
+## Proposal
+
+The initial design includes:
+
+- `RuntimeClass` API resource definition
+- `RuntimeClass` pod field for specifying the RuntimeClass the pod should be run with
+- Kubelet implementation for fetching & interpreting the RuntimeClass
+- CRI API & implementation for passing along the RuntimeClass definition
+
+A subsequent update will include the API for describing supported features and an admission
+controller for performing validation & defaulting.
+
+### API
+
+`RuntimeClass` is a new cluster-scoped resource in the `node.k8s.io` API group.
+
+> _The `node.k8s.io` API group would eventually hold the Node resource when `core` is retired.
+> Alternatives considered: `runtime.k8s.io`, `cluster.k8s.io`_
+
+There are 3 categories of fields that may be included in the `RuntimeClassSpec`:
+
+1. **Runtime parameters** - These fields control the behavior of the runtime in some way,
+   including configuring arbitrary options in the runtime (`Parameters`), and eventually specifying
+   resource overhead. This is the focus of this initial proposal.
+2. **Compatibility validation** - These fields declare support for various features, and pods are
+   validated against them in admission. This category, along with the admission controller, will be
+   addressed in a subsequent update to this proposal.
+3. **Informational** - These fields surface semantic properties of the runtime to guide runtime
+   selection. Examples could include hardware virtualization, stability, and I/O performance
+   class. These fields are out-of-scope for the initial proposal.
+
+#### API Specification
+
+_(This is a simplified declaration, syntactic details will be covered in the API PR review)_
+
+```go
+type RuntimeClass struct {
+    metav1.TypeMeta
+    // ObjectMeta minimally includes the RuntimeClass name, which is used to reference the class.
+    // Namespace should be left blank.
+    metav1.ObjectMeta
+
+    Spec RuntimeClassSpec
+}
+
+type RuntimeClassSpec struct {
+    // Parameters holds opaque parameters that are interpreted by the runtime (CRI implementation).
+    Parameters map[string]string
+}
+```
+
+The runtime is selected by the pod by specifying the RuntimeClass in the PodSpec:
+
+```go
+type PodSpec struct {
+    ...
+    // RuntimeClassName refers to a RuntimeClass object with the same name,
+    // which should be used to run this pod.
+    // +optional
+    RuntimeClassName string
+    ...
+}
+```
+
+The `legacy` RuntimeClass name is reserved. The legacy RuntimeClass is defined to be fully backwards
+compatible with current Kubernetes. This means that the legacy runtime does not specify any runtime
+parameters or perform any feature validation (all features are "supported").
+
+```go
+const (
+    // RuntimeClassNameLegacy is a reserved RuntimeClass name. The legacy
+    // RuntimeClass does not specify any runtime parameters or perform any
+    // feature validation.
+    RuntimeClassNameLegacy = "legacy"
+)
+```
+
+An unspecified RuntimeClassName `""` is equivalent to the `legacy` RuntimeClass, though the field is
+not defaulted to `legacy` (to leave room for configurable defaults in a future update).
+
+#### CRI API Specification
+
+The runtime parameters are passed to the CRI as part of the `RunPodSandboxRequest`:
+
+```proto
+message RunPodSandboxRequest {
+    // Configuration for creating a PodSandbox.
+    PodSandboxConfig config = 1;
+    // Configuration for the runtime.
+    map<string, string> RuntimeParameters = 2;
+}
+```
+
+### Versioning, Updates, and Rollouts
+
+_Axiom: Rerunning a pod on a different node in the same cluster with the same RuntimeClassName should
+not change its behavior._
+
+This means that if a runtime is updated in a fully backwards-compatible way, then the RuntimeClass
+does not _need_ to change. It also means that if the updated runtime breaks backwards compatibility,
+then it **MUST** be renamed. RuntimeClass does not provide a predefined versioning or naming scheme,
+so versioned upgrades are left to the cluster operator to handle.
+
+Since breaking runtime changes require a new RuntimeClass definition, it means that pods must be
+updated to reference the new RuntimeClass. This comes with the advantage of native support for
+rolling updates through the same mechanisms as any other application update.
+
+### Implementation Details
+
+The Kubelet uses an Informer to keep a local cache of all RuntimeClass objects. When a new pod is
+added, the Kubelet resolves the Pod's RuntimeClass against the local RuntimeClass cache.  Once
+resolved, a subset of its fields are passed to the CRI as part of the [`RunPodSandboxRequest`][]. At
+that point, the interpretation of the parameters is left to the CRI implementation, but they should
+be cached if needed for subsequent calls.
+
+[RunPodSandboxRequest]: https://github.com/kubernetes/kubernetes/blob/b05a61e299777c2030fbcf27a396aff21b35f01b/pkg/kubelet/apis/cri/runtime/v1alpha2/api.proto#L344
+
+### Risks and Mitigations
+
+**Scope creep.** RuntimeClass has a fairly broad charter, but it should not become a default
+dumping ground for every new feature exposed by the node. For each feature, careful consideration
+should be made about whether it belongs on the Pod, Node, RuntimeClass, or some other resource. The
+[non-goals](#non-goals) should be kept in mind when considering RuntimeClass features. Features that
+are too specific to a single runtime implementation should use `Parameters` rather than expanding
+the `RuntimeClassSpec`.
+
+**Becoming a general policy mechanism.** RuntimeClass should not be used a replacement for
+PodSecurityPolicy. The use cases for defining multiple RuntimeClasses for the same underlying
+runtime implementation should be extremely limited (generally only around updates & rollouts). To
+enforce this, no authorization or restrictions are placed directly on RuntimeClass use; in order to
+restrict a user to a specific RuntimeClass, you must use another policy mechanism such as
+PodSecurityPolicy.
+
+**Pushing complexity to the user.** RuntimeClass is a new resource in order to hide the complexity
+of runtime configuration from most users (aside from the cluster admin or provisioner). However, we
+are still side-stepping the issue of precisely defining specific types of runtimes like
+"Sandboxed". However, it is still up for debate whether precisely defining such runtime categories
+is even possible. RuntimeClass allows us to decouple this specification from the implementation, but
+it is still something I hope we can address in a future iteration through the concept of pre-defined
+or "conformant" RuntimeClasses.
+
+**Non-portability.** We are already in a world of non-portability for many features (see [examples
+of runtime variation](#examples-of-runtime-variation). RuntimeClasses actually address this issue by
+formally declaring supported features. Another issue is that pods need to refer to a RuntimeClass by
+name, which may not be defined in every cluster. This is something that can be addressed through
+pre-defined runtime classes (see previous risk), and/or by "fitting" pod requirements to compatible
+RuntimeClasses.
+
+## Graduation Criteria
+
+Alpha:
+
+- Everything described in the current proposal
+- Supported features and associated admission controller
+
+Beta:
+
+- Major runtimes support RuntimeClass
+- RuntimeClasses are configured in the E2E environment with test coverage of a non-legacy RuntimeClass
+- TBD
+
+## Implementation History
+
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
@tallclair i had expected to see an association or mapping of some kind that enumerated what set of `RuntimeClass` definitions are supported by a particular `Node` to aid in scheduling, but it's possible that you intended to express that elsewhere.   In general, for `Parameters` is the assumption that the set of opaque parameters provided are not statically configured by a particular runtime and therefore can be updated on a per container/pod basis?  Can we provide some representative example syntaxes for some of the items in the appendix to know if the structure is sufficient?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,311 @@
...
+
+- Linux Security Module (LSM) choice - Kubernetes supports both AppArmor & SELinux options on pods,
+  but those are mutually exclusive, and support of either is not required by the runtime. The
+  default configuration is also not well defined.
+- Seccomp-bpf - Kubernetes has alpha support for specifying a seccomp profile, but the default is
+  defined by the runtime, and support is not guaranteed.
+- Windows containers - isolation features are very OS-specific, and most of the current features are
+  limited to linux. As we build out Windows container support, we'll need to add windows-specific
+  features as well.
+- Host namespaces (Network,PID,IPC) may not be supported by virtualization-based runtimes
```
derekwaynecarr: i was hoping that userns defaulting logic could be applied via runtimeclass as well for similar reasons.

/cc @vikaschoudhary16

tallclair:
> i had expected to see an association or mapping of some kind that enumerated what set of RuntimeClass definitions are supported by a particular Node to aid in scheduling, but it's possible that you intended to express that elsewhere.

Yeah, I see this as somewhere in the intersection of built-in scheduling support & runtime discovery, listed as areas for future exploration. I'd like to keep this out of the alpha for a couple reasons: limit the initial scope, it can be cleanly layered on to the existing proposal, and it's a hard problem that needs careful thought (best considered without other distractions).

> In general, for Parameters is the assumption that the set of opaque parameters provided are not statically configured by a particular runtime and therefore can be updated on a per container/pod basis? Can we provide some representative example syntaxes for some of the items in the appendix to know if the structure is sufficient?

Right, these are parameters that are expected to have dynamic effect. Examples include:
- selecting OCI runtime, as is currently done through annotations with cri-o, containerd & frakti
- configuring new runtime defaults, as a way of rolling out changes to things that would otherwise break backward compatibility (e.g. default seccomp profile, default capability set, usernamespace default). This runs the risk of overlapping too much with "general policy" though.

However, if you have a bunch of completely independent features, then the RuntimeClass parameters are probably not a good fit (don't want a combinatorial explosion of different runtime classes), and those features that really vary pod-to-pod should stay as annotations.

I'll add this to the KEP.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+
+A subsequent update will include the API for describing supported features and an admission
+controller for performing validation & defaulting.
+
+### API
+
+`RuntimeClass` is a new cluster-scoped resource in the `node.k8s.io` API group.
+
+> _The `node.k8s.io` API group would eventually hold the Node resource when `core` is retired.
+> Alternatives considered: `runtime.k8s.io`, `cluster.k8s.io`_
```
tallclair: I'm open to other suggestions for a better group.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+```
+// Decide which underlying OCI runtime should be used. This replaces the
+// annotations used by CRI-o, containerd & frakti to select between sandboxed
+// runtimes.
+"oci-runtime": "runsc"
+
+// Change the default set of capabilities used by the runtime. This is a good
+// way of surfacing behaviors that were previously implicit in docker but not
+// well defined in Kubernetes.
+"default-capabilities": "[]"
```
tallclair: Configuring defaults is eeking over towards general policy territory, but I see 2 reasons to include defaults in RuntimeClass parameters:

1. Our defaulting / policy story is a mess right now, and we don't have a good composable way of associating defaults with another field. In this case, setting defaults according to runtime class, irrespective of user.
2. Defaulting of fields otherwise not exposed in the API. If it's not configured on the pod spec, it can't be defaulted in the control plane.

keps/sig-node/0014-runtime-class.md
vikaschoudhary16: @derekwaynecarr userns feature will require a new runtimeclass because it is not backward compatible. If scheduling awareness is there, runtimeclasses will be more helpful in upgrades. New pods with userns runtimeclass will get steered to upgraded nodes. And it will be possible to have `hostUserNamespace=false` by default if `NodeLevelUserNamespace` is enabled. But for now, i think keeping `hostUserNamespace=true` as default makes more sense.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,358 @@
...
+
+Today's runtimes have varying support for features that are expressible through the PodSpec (see
+[examples of runtime variation](#examples-of-runtime-variation) below). In some cases such as
+SELinux, features are silently ignored when they aren't supported. This is problematic from a
+security perspective since it makes it difficult to audit actual runtime options. Other features
+such as AppArmor are checked for support, but on failure the pod is just held in a "Blocked" state
+to avoid the churn of recreating the rejected pod. Still other features maybe be applied in slightly
+different configurations, as is the case with capabilities today (the default set is not
+well-defined), and `privileged` containers. By raising these properties to the control plane errors
+can be surfaced much earlier, and configuration can be made explicit.
```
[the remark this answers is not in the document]
tallclair: Use of AppArmor will be rejected in validation if the feature gate is disabled, but that doesn't tell you if the nodes actually support it, also the feature gate is not intended to be a cluster-wide "is AppArmor supported" bit, just gating the maturity of the feature (and will go away once the feature is GA)

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,358 @@
...
+
+The second problem that RuntimeClass addresses is supporting multiple runtimes in a cluster, or even
+on the same node. [Sandboxes][] are the primary motivator for this right now, with both Kata
```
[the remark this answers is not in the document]
tallclair: Yeah, we're sticking with a single CRI implementation from the Kubelet's perspective for now. The existing approach is to fork the behavior at the OCI layer, but you could also imagine a CRI-multiplexer that could handle multiple CRIs while still exposing a single one to the kubelet.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,358 @@
...
+3. Total size of keys+values must be less than 256K
+
+The runtime parameters are passed to the CRI as part of the `RunPodSandboxRequest`:
+
+```proto
+message RunPodSandboxRequest {
+    // Configuration for creating a PodSandbox.
+    PodSandboxConfig config = 1;
+    // Configuration for the runtime.
+    map<string, string> RuntimeParameters = 2;
```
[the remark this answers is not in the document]
tallclair: Ah, I changed that and forgot to update line 130. I'll fix that.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,358 @@
...
+to avoid the churn of recreating the rejected pod. Still other features maybe be applied in slightly
+different configurations, as is the case with capabilities today (the default set is not
+well-defined), and `privileged` containers. By raising these properties to the control plane errors
+can be surfaced much earlier, and configuration can be made explicit.
+
+The second problem that RuntimeClass addresses is supporting multiple runtimes in a cluster, or even
+on the same node. [Sandboxes][] are the primary motivator for this right now, with both Kata
+containers and gVisor looking to integrate with Kubernetes. Other runtime models such as Windows
+containers or even remote runtimes will also require support in the future. RuntimeClass provides a
+way to configure different runtimes in the cluster, surface their properties (both to the cluster &
```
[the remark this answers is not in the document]
tallclair: Sort of both. Configuring different runtimes & selecting between those configurations.

> "selecting a runtime out of many running with different configurations based on matching details specified in the runtimeclass"?

I think this is what I called "fitting" below (non-goals), and is out of scope initially, but should be explored in the future. As an aside, I think labels & label selectors might be a good solution.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,358 @@
...
+  containers.
+- As a cluster operator, I want to provide stable rolling upgrades of nodes.
+- As an application developer, I want to select the runtime that best fits my workload.
+- As an application developer, I don't want to study the nitty-gritty details of different runtime
+  implementations.
+- As an application developer, I want my application to be portable across clusters that use similar
+  but different variants of a "class" of runtimes.
+- As an application operator, I want to know precisely how the application is running, and which
+  features are being used.
+- As an application operator, I want pods to fail fast when trying to use an optional feature that
```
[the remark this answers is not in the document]
tallclair: It will fail at the node level, and then we hold it in the blocked state to avoid scheduler thrashing. By failing the request in the admission stage, the error is surfaced in a more obvious way.

keps/sig-node/0014-runtime-class.md
tallclair: I'd like to be careful about expanding the scope to setting backwards-incompatible defaults. I think options that are expressed in the PodSpec might be better off defaulted in cluster policy, even though our current cluster policy story (e.g. PodSecurityPolicy) is weak right now. I'm not saying RuntimeClass can't be used for that, I just don't think it's a usecase we should focus on right now. WDYT?

tallclair:
Oops, just realized this was supposed to be KEP 13. I'll change it before submitting, but leave it for now to preserve comments.

keps/sig-node/0014-runtime-class.md
lizrice: I like that this gives a mechanism for sane defaulting of capabilities. Similarly I wonder if this could be a mechanism for defaults for the other fields in the OCI runtime config - for example [hooks](https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks), which afaict don't have any way to be configured at all right now)?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+// well defined in Kubernetes.
+"default-capabilities": "[]"
+
+// Set the infrastructure container image. More generally, configuring options
+// not otherwise exposed in the Kubrenetes API.
+"infra-image": "example.com/self-monitoring-pod:v2"
+```
+
+_Note: RuntimeClass is intended to be consumed by a single runtime implementation, so parameter keys
+do not need to be "namespaced"._
```
lizrice: For my understanding: you could have multiple runtimeClasses on the same cluster, referring to the same runtime but with different parameters, right?  I'm thinking of, say, runc with and without SELinux, and a future when scheduling could take RuntimeClass into consideration.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+  future iteration.
+
+[Pod Overhead]: https://docs.google.com/document/d/1EJKT4gyl58-kzt2bnwkv08MIUZ6lkDpXcxkHqCvvAp4/edit
+
+### User Stories
+
+- As a cluster operator, I want to provide multiple runtime options to support a wide variety of
+  workloads. Examples include native linux containers, "sandboxed" containers, and windows
+  containers.
+- As a cluster operator, I want to provide stable rolling upgrades of nodes.
```
sjenning: How is this a user story for RuntimeClasses?

keps/sig-node/0014-runtime-class.md
tallclair: Yeah, definitely. 2 usecases I have in mind for this:

1. Same CRI runtime, different OCI runtime (not sure if you were including this)
2. "Native" RuntimeClass - This would be the default runc runtime, but applying the constraints & specs that "should" have been there from the start. The `legacy` RuntimeClass would apply to the same runc runtime, but without any constraints / supported features. (will make more sense when I follow up with the supported features addition)

To your example of runc with & without SELinux, if you're running mixed nodes where some do & don't have the SELinux LSM installed, then this makes sense. However, if you're talking about configuring a runtimeclass on a single node that supports SELinux, that feels like something that would be better configured in cluster-level policy configuration. Does that make sense?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+- As an application developer, I want to select the runtime that best fits my workload.
+- As an application developer, I don't want to study the nitty-gritty details of different runtime
```
sjenning: Yet, the application developer will need to bear the cognitive load of understanding the trade-offs between runtimes since they will have to write the pod spec to select the runtime explicitly rather than expressing the attributes they require of the runtime ("Fitting" from above)

keps/sig-node/0014-runtime-class.md
tallclair: Good question! It's more about updating the runtime, which is typically tied to the node version in most deployments. If I want to roll out a new runtime version that has some new features or breaking changes, then RuntimeClass is a mechanism that could enable this. To really solve this use case, we need to add better scheduling support (which I want to do eventually).

keps/sig-node/0014-runtime-class.md
tallclair: Yes, but this shifts some of the responsibility. For example, the cluster admin can set up the `native` RuntimeClass that accurately reflects the features supported & used in the specific cluster, and tell developers to use that instead of `legacy` (or even require it with policy), that way the developer doesn't need to read the details of the specific node OS / runtime versions - they can just see the supported features directly in the RuntimeClass. If the admin sets up a `sandboxed` class, the developer can use it with a reasonable assumption of security (if they trust the admin), without needing to learn all the knobs to tune - it's a simplification. We also might add "fitting" in the future, it's not off the table.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+The `legacy` RuntimeClass name is reserved. The legacy RuntimeClass is defined to be fully backwards
+compatible with current Kubernetes. This means that the legacy runtime does not specify any runtime
+parameters or perform any feature validation (all features are "supported").
+
+```go
+const (
+    // RuntimeClassNameLegacy is a reserved RuntimeClass name. The legacy
+    // RuntimeClass does not specify any runtime parameters or perform any
+    // feature validation.
+    RuntimeClassNameLegacy = "legacy"
```
sjenning: I'm not a huge fan of the name "legacy".  In my mind, it conveys the idea of "old", "possible a target of deprecation", and "insecure", when really what we are meaning to convey is "the native kubernetes RuntimeClass".

How about "native" instead?

tallclair: > I'm not a huge fan of the name "legacy". In my mind, it conveys the idea of "old", "possible a target of deprecation", and "insecure",

Great, that's exactly what I was going for :D

> How about "native" instead?

I would like to add `native` as a standard RuntimeClass name that users can expect to find on clusters. The difference between `native` and `legacy` is that `native` would include feature validation. This will make more sense once I add the "supported features" update (I went back-and-forth about whether to include it in the initial proposal). But basically `native` will be the runtime that users _should_ use, but `legacy` needs to be included for backwards compatibility. Make sense?

mrunalp:
Looks good :+1:

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+formally declaring supported features. Another issue is that pods need to refer to a RuntimeClass by
+name, which may not be defined in every cluster. This is something that can be addressed through
+pre-defined runtime classes (see previous risk), and/or by "fitting" pod requirements to compatible
+RuntimeClasses.
+
+## Graduation Criteria
+
+Alpha:
+
+- Everything described in the current proposal
```
derekwaynecarr: can we enumerate a set of proposed RuntimeParameters we expect to support as part of Alpha success criteria?  i feel like we need to at least demonstrate support for 1...X parameter options otherwise its just an envelope with no paper.  I assume runtime selection is the first parameter choice, but it would be good to know a minimum set.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+not be used to enumerate general pod options as this would lead to a combinatorial explosion of
+RuntimeClass definitions.
+
+Examples of runtime parameters include:
+
+```
+// Decide which underlying OCI runtime should be used. This replaces the
+// annotations used by CRI-o, containerd & frakti to select between sandboxed
+// runtimes.
+"oci-runtime": "runsc"
```
resouer: There should be some runtimes are not oci specified but still integrated with CRI, maybe the name of `oci-runtime` a bit arbitrary？

keps/sig-node/0014-runtime-class.md
resouer: I also think defaulting capabilities in `RuntimeClass` is practical approach: it does seem more like a runtime configuration than user specific.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+- Linux Security Module (LSM) choice - Kubernetes supports both AppArmor & SELinux options on pods,
+  but those are mutually exclusive, and support of either is not required by the runtime. The
+  default configuration is also not well defined.
+- Seccomp-bpf - Kubernetes has alpha support for specifying a seccomp profile, but the default is
+  defined by the runtime, and support is not guaranteed.
+- Windows containers - isolation features are very OS-specific, and most of the current features are
+  limited to linux. As we build out Windows container support, we'll need to add windows-specific
+  features as well.
+- Host namespaces (Network,PID,IPC) may not be supported by virtualization-based runtimes
+  (e.g. Kata-containers & visor).
```
resouer: typo: gvisor? but gvisor should support host network.

keps/sig-node/0014-runtime-class.md
resouer: By saying that ` is that native would include feature validation.` Do you mean validation on runtime implementation side? Or the following admission design?

tallclair: Following admission design.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+* [Implementation History](#implementation-history)
+* [Appendix](#appendix)
+  * [Examples of runtime variation](#examples-of-runtime-variation)
+
+## Summary
+
+`RuntimeClass` is a new cluster-scoped resource that surfaces container runtime properties to the
+control plane. RuntimeClasses are assigned to pods through a `runtimeClass` field on the
+`PodSpec`. This provides a new mechanism for supporting multiple runtimes in the cluster and on the
+nodes.
```
krmayankk: nit: i think you mean we can support multiple runtimes on the same node at the same time. what would it mean to support multiple runtimes in the cluster ?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+- Automatic runtime or feature discovery - initially RuntimeClasses are manually defined (by the
+  cluster admin or provider), and are asserted to be an accurate representation of the runtime.
+- Scheduling in heterogeneous clusters - it is possible to operate a heterogeneous cluster
+  (different runtime configurations on different nodes) through scheduling primitives like
+  `NodeAffinity` and `Taints+Tolerations`, but the user is responsible for setting these up and
+  automatic runtime-aware scheduling is out-of-scope.
+- Define standardized or conformant runtime classes - although I would like to declare some
+  predefined RuntimeClasses with specific properties, doing so is out-of-scope for this initial KEP.
+- [Pod Overhead][] - Although RuntimeClass is likely to be the configuration mechanism of choice,
+  the details of how pod resource overhead will be implemented is out of scope for this KEP.
```
krmayankk: not clear what does pod resource overhead mean here ?

keps/sig-node/0014-runtime-class.md
krmayankk: i would reword this as providing rolling upgrades of runtimes. Thats would this enables, rolling upgrade of nodes could be many things like upgrading os on the nodes, etc.

keps/sig-node/0014-runtime-class.md
krmayankk: i can imagine runtime classes like containerd-default, containerd-secure, etc or default, secure, super-secure :-)

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+  workloads. Examples include native linux containers, "sandboxed" containers, and windows
+  containers.
+- As a cluster operator, I want to provide stable rolling upgrades of nodes.
+- As an application developer, I want to select the runtime that best fits my workload.
+- As an application developer, I don't want to study the nitty-gritty details of different runtime
+  implementations.
+- As an application developer, I want my application to be portable across clusters that use similar
+  but different variants of a "class" of runtimes.
+- As an application operator, I want to know precisely how the application is running, and which
+  features are being used.
```
krmayankk: you mean auditing applications by finding the runtimeclass and looking at what features of the runtime those class enable ?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+#### API Specification
+
+_(This is a simplified declaration, syntactic details will be covered in the API PR review)_
+
+```go
+type RuntimeClass struct {
+    metav1.TypeMeta
+    // ObjectMeta minimally includes the RuntimeClass name, which is used to reference the class.
+    // Namespace should be left blank.
+    metav1.ObjectMeta
```
krmayankk: do you envision this to also be namespaced as well in future ? for e.g. tenant A supports runtime A and tenant B supports runtime B ?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+}
+
+type RuntimeClassSpec struct {
+    // Parameters holds opaque parameters that are interpreted by the runtime (CRI implementation).
+    Parameters map[string]string
+}
+```
+
+The runtime is selected by the pod by specifying the RuntimeClass in the PodSpec. Once the pod is
+scheduled, the RuntimeClass cannot be changed.
```
krmayankk: should be enforced in the admission.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+
+```go
+type PodSpec struct {
+    ...
+    // RuntimeClassName refers to a RuntimeClass object with the same name,
+    // which should be used to run this pod.
+    // +optional
+    RuntimeClassName string
```
krmayankk: when not specified this would select the dockershim as the default runtime ?

keps/sig-node/0014-runtime-class.md
krmayankk: why do we need the name legacy ? Not specifying the class should invoke this behavior

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+"oci-runtime": "runsc"
+
+// Change the default set of capabilities used by the runtime. This is a good
+// way of surfacing behaviors that were previously implicit in docker but not
+// well defined in Kubernetes.
+"default-capabilities": "[]"
+
+// Set the infrastructure container image. More generally, configuring options
+// not otherwise exposed in the Kubrenetes API.
+"infra-image": "example.com/self-monitoring-pod:v2"
```
krmayankk: are you referring to pause ? Does this really belong to runtime configuration ?

keps/sig-node/0014-runtime-class.md
krmayankk: dumb question what is the difference between CRI runtime and OCI runtime ?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+
+Getting upgrades and rollouts right is a very nuanced and complicated problem. For the initial alpha
+implementation, we will kick the can down the road by making the `RuntimeClassSpec` **immutable**,
+thereby requiring changes to be pushed as a newly named RuntimeClass instance. This means that pods
+must be updated to reference the new RuntimeClass, and comes with the advantage of native support
+for rolling updates through the same mechanisms as any other application update.
+
+This conservative approach is preferred since it's much easier to relax constraints in a backwards
+compatible way than tighten them. We should revisit this decision prior to graduating RuntimeClass
+to beta.
```
krmayankk: Are you thinking in beta to be able to update a given RuntimeClass params, and that kicks the pod rolling update automatically without updating the class name ?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,359 @@
...
+  defined by the runtime, and support is not guaranteed.
+- Windows containers - isolation features are very OS-specific, and most of the current features are
+  limited to linux. As we build out Windows container support, we'll need to add windows-specific
+  features as well.
+- Host namespaces (Network,PID,IPC) may not be supported by virtualization-based runtimes
+  (e.g. Kata-containers & visor).
+- Per-pod and Per-container resource overhead varies by runtime.
+- Device support (e.g. GPUs) varies wildly by runtime & nodes.
+- Supported volume types varies by node - it remains TBD whether this information belongs in
+  RuntimeClass.
```
krmayankk: i think runtime is limited to environment needed to run pods and hence volume types doesnt belong here

keps/sig-node/0014-runtime-class.md
lizrice: Could this simply be called "runtime"?

keps/sig-node/0014-runtime-class.md
lizrice: Let me see if I understand about the cluster-level policy config: I suppose there are two ways for users to come at determining what runtime to use: 

- a "developer intent" that a given workload should really have a certain level of isolation or requires certain features from its runtime, and this is defined in the PodSpec

- a "security policy" defined cluster-wide that says that workloads, or a certain class of workloads, require a certain level of isolation or runtime features. This is defined in some cluster-wide policy - should that also refer to this RuntimeClass definition? Is the idea that this will be policed at admission time?

keps/sig-node/0014-runtime-class.md
resouer: @lizrice I once commented similar before but later I found Tim was just listing an example here. :D

So I believe any key:value pairs can be passed in this part.

keps/sig-node/0014-runtime-class.md
tallclair: clarified. It could be the case that you have some nodes that support one runtime and other nodes that support a different runtime (e.g. linux & windows), or a bunch of nodes that support 2 runtimes (e.g. native & sandboxed).

keps/sig-node/0014-runtime-class.md
tallclair: See the link. This is referring to another proposal.

keps/sig-node/0014-runtime-class.md
tallclair: See the second paragraph of the motivation section. Just because a feature is specified in the podspec doesn't mean the pod is running with that feature.

keps/sig-node/0014-runtime-class.md
tallclair: No. RuntimeClasses are a property of the underlying infrastructure. I don't think it makes any more sense to namespace them than it does to namespace nodes. You could restrict some tenants to a certain runtime class through namespaced policy though.

keps/sig-node/0014-runtime-class.md
tallclair: I think this would be enforced in validation (https://github.com/kubernetes/kubernetes/blob/44073e6f43af29c0fb8cad66593d850435c6e2be/pkg/apis/core/validation/validation.go#L3434)

keps/sig-node/0014-runtime-class.md
tallclair: Not sure I understand this comment. RuntimeClass isn't selecting between different CRI servers - it's up to the single CRI server the kubelet talks to interpret how to run the RuntimeClass.

keps/sig-node/0014-runtime-class.md
tallclair: Having the legacy class lets you explicitly use it. For example, if a cluster policy defaults RuntimeClassName to `native`, but a given app is incompatible with it, then that app can explicitly state `legacy`. It also makes it easier to reference from policy (e.g. `forbid: ["legacy"]` is more readable than `forbid: [""]`)

keps/sig-node/0014-runtime-class.md
krmayankk: ok so we are saying that by specifying the runtimeclass, the control plane knows what features that class supports, so i can match what is specified in the podspec with what is supported in the class to surface errors. May be this needs to be written somewhere or its obvious  to everyone

keps/sig-node/0014-runtime-class.md
resouer: @krmayankk "Automatic runtime or feature discovery" is not covered right now as mentioned above, so we need to do manual configuration right now.

tallclair:
After our previous discussion of RuntimeClass in sig-node, I was asked by @derekwaynecarr to come up with a set of concrete use cases for runtime `Parameters`, to make sure we aren't building an "envelope with nothing to put in it". This was an excellent suggestion, and when I dug deeper I couldn't come up with any compelling use cases other than selecting between different runtime setups. There were 2 other use cases that were tempting, but I ultimately decided against:

1. **Configure Defaults** - For instance, setting the default user namespace value. This is definitely tempting, I think mostly because we don't currently have a good story around configuring defaults. However, this is a more general problem, and I don't think it's something we should include in RuntimeClass simply because it would be easy to do so. This requires more thought and a holistic look at the space of cluster policies defaulting in general. That isn't to say that it will never be included in RuntimeClass, but I'm not ready to include it now.
2. **Arbitrary runtime configuration** - Examples I previously gave included the default set of capabilities, the infrastructure container image, and the path to the oci-runtime binary. Including these parameters makes the RuntimeClass less portable, and makes upgrades more complicated. For instance, moving the location of the binary should not require me to redeploy my workload just because the RuntimeClass configuration changed. These problems are better dealt with independently.

Since the only use case I came up with was selecting the underlying runtime setup (and it turns out the "untrusted" runtime feature has been adopted by most CRI implementations), I decided to remove `Parameters` and replace it with a dedicated field for selecting the name.

This brings us to naming. I think almost every possible term we could use is already overloaded many times over, so the only hope is to come up with a unique combination. The best I came up with is `RuntimeHandler`, but I am definitely open to suggestions. Other contenders included:
- ExecutionHandler
- ExecutionDriver (Docker terminology)
- RuntimeConfigurationName
- RuntimeTag

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,332 @@
...
+such as AppArmor are checked for support, but on failure the pod is just held in a "Blocked" state
+to avoid the churn of recreating the rejected pod. Still other features maybe be applied in slightly
+different configurations, as is the case with capabilities today (the default set is not
+well-defined), and `privileged` containers. By raising these properties to the control plane errors
+can be surfaced much earlier (admission), and configuration can be made explicit.
+
+The second problem that RuntimeClass addresses is supporting multiple runtimes in a cluster, or even
+on the same node. [Sandboxes][] are the primary motivator for this right now, with both Kata
+containers and gVisor looking to integrate with Kubernetes. Other runtime models such as Windows
+containers or even remote runtimes will also require support in the future. RuntimeClass provides a
```
yujuhong: Windows may not be the best example unless we plan not to surface the information on the pod or node.

keps/sig-node/0014-runtime-class.md
yujuhong: Yes, restrictions on pod update will need to be enforced in validation (as it won't go through admission)

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,332 @@
...
+The Kubelet uses an Informer to keep a local cache of all RuntimeClass objects. When a new pod is
+added, the Kubelet resolves the Pod's RuntimeClass against the local RuntimeClass cache.  Once
+resolved, the RuntimeHandler field is passed to the CRI as part of the
+[`RunPodSandboxRequest`][]. At that point, the interpretation of the RuntimeHandler is left to the
+CRI implementation, but it should be cached if needed for subsequent calls.
+
+If the RuntimeClass cannot be resolved (e.g. doesn't exist) at Pod creation, then the request will
+be rejected in admission (controller to be detailed in a following update). If the RuntimeClass
+cannot be resolved by the Kubelet when `RunPodSandbox` should be called, then the Kubelet will fail
+the Pod. The admission check on a replica recreation will prevent the scheduler from thrashing.
```
yujuhong: If the CRI implementation does not recognize the `RuntimeHandler`, it should also return an error.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,332 @@
...
+
+[RunPodSandboxRequest]: https://github.com/kubernetes/kubernetes/blob/b05a61e299777c2030fbcf27a396aff21b35f01b/pkg/kubelet/apis/cri/runtime/v1alpha2/api.proto#L344
+
+### Risks and Mitigations
+
+**Scope creep.** RuntimeClass has a fairly broad charter, but it should not become a default
+dumping ground for every new feature exposed by the node. For each feature, careful consideration
```
yujuhong: +1. We should be careful about overloading RuntimeClass for all kinds of node resources.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,332 @@
...
+name, which may not be defined in every cluster. This is something that can be addressed through
+pre-defined runtime classes (see previous risk), and/or by "fitting" pod requirements to compatible
+RuntimeClasses.
+
+## Graduation Criteria
+
+Alpha:
+
+- Everything described in the current proposal
+- Supported features and associated admission controller
```
yujuhong: This needs a CRI validation test in the cri-tools repo, at the least.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,362 @@
...
+`PodSpec`. This provides a new mechanism for supporting multiple runtimes in a cluster and/or node.
+
+## Motivation
+
+RuntimeClasses address two distinct problems: surfacing support and configuration of optional
+features, and supporting multiple runtimes.
+
+Today's runtimes have varying support for features that are expressible through the PodSpec (see
+[examples of runtime variation](#examples-of-runtime-variation) below). In some cases such as
+SELinux, features are silently ignored when they aren't supported. This is problematic from a
```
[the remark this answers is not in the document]
tallclair: That's why it's motivating ;-)

One thing I have been wondering about though is whether this is problematic for portability. For instance I can't have one yaml that runs in a apparmor OR selinux cluster. I think that's OK though, and we should just recommend using a tool like kustomize to deal with that.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,366 @@
...
+security perspective since it makes it difficult to audit actual runtime options. Other features
+such as AppArmor are checked for support, but on failure the pod is just held in a "Blocked" state
+to avoid the churn of recreating the rejected pod. Still other features maybe be applied in slightly
+different configurations, as is the case with capabilities today (the default set is not
+well-defined), and `privileged` containers. By raising these properties to the control plane errors
+can be surfaced much earlier (admission), and configuration can be made explicit.
```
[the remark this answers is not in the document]
tallclair: Admission is the stage where we decide whether a resource is valid, and then potentially reject the create or update request. We might want to update the scheduler to handle heterogeneous clusters, but I think feature checking should be done in admission.

keps/sig-node/0014-runtime-class.md
tallclair: I'm not sure I understand this. Do you mean we might want a top-level `OS` field on the podspec?

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,362 @@
...
+  RuntimeClass by name. This approach can increase portability, but can be added seamlessly in a
+  future iteration.
+
+[Pod Overhead]: https://docs.google.com/document/d/1EJKT4gyl58-kzt2bnwkv08MIUZ6lkDpXcxkHqCvvAp4/edit
+
+### User Stories
+
+- As a cluster operator, I want to provide multiple runtime options to support a wide variety of
+  workloads. Examples include native linux containers, "sandboxed" containers, and windows
+  containers.
```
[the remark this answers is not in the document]
tallclair: Yeah, but you could also have a VM-based runtime that runs windows containers on a linux node. This is the case for most node-level properties (except hardware devices), so I'm thinking about just removing that non-goal.

keps/sig-node/0014-runtime-class.md
```diff
@@ -0,0 +1,362 @@
...
+- As a cluster operator, I want to provide stable rolling upgrades of runtimes. For
+  example, rolling out an update with backwards incompatible changes or previously unsupported
+  features.
+- As an application developer, I want to select the runtime that best fits my workload.
+- As an application developer, I don't want to study the nitty-gritty details of different runtime
+  implementations.
+- As an application developer, I want my application to be portable across clusters that use similar
```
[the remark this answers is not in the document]
tallclair: Examples:
- "sandboxed" runtime class, gvisor & kata-containers implementations
- "vm" runtime class, KVM and xen solutions
- "windows" runtime class, native on windows host vs. windows in VM solutions

Really what I'm getting at is that runtime class separates the "intent" from the implementation.

keps/sig-node/0014-runtime-class.md
tallclair: @yujuhong yes, that's the idea. Right now that's not the case (e.g. selinux), but with "supported features" enforcement, you could assume that if you specified something in the PodSpec it's actually being used.

keps/sig-node/0014-runtime-class.md
tallclair: Agreed. Added.

keps/sig-node/0014-runtime-class.md
tallclair: Added.

keps/sig-node/0014-runtime-class.md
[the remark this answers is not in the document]
derekwaynecarr: i feel similarly about support on non-linux platforms.  the pod should not run if its spec cannot be realized.

keps/sig-node/0014-runtime-class.md
derekwaynecarr: its inevitable to me that we will need to declare the target os of a pod spec to simplify management of associated features.  defaulting must differ based on target platform and right now there is nothing but an opaque label in many cases to know if this is windows/linux.  to me, its inevitable that we will need to know the target platform on the pod spec itself similar to https://github.com/opencontainers/runtime-spec/blob/master/config.md#platform-specific-configuration

derekwaynecarr:
I am torn on the proposal.

I would flip the ordering of motivations as the proposal clearly addresses variability within runtime selection, but less adequately describes in my opinion how varying support for features expressible through the pod API spec are addressed.  I agree the latter motivation is important, but I think its only properly solved by doing a better job distinguishing on the pod API spec itself the features or concepts that vary by platform similar to oci runtime-spec rather than via a secondary object.  For varying support among a common platform (i.e. linux using apparmor vs selinux), its not clear how this API facilitates that distinction just yet and at some level this motivation also contradicts with the scope creep.

As for the actual goals targeted in the proposal, I think it does a satisfactory job of describing an opaque bundle of potential properties that could be ingested by the container runtime itself to support running more than one type of sandbox approach on a single node.  If we accept that some pods will always need to be run on a host with a looser sandbox configuration to provide host level monitoring, this seems like a sensible step to supporting varying types of sandboxes on a common node.

It's just not clear to me what (if anything else) would be also added to this resource type.  Another option I speculated on was storage driver variance across runtimes (overlay, devicemapper, etc.) but typically they never vary on a single host but could vary across hosts.

keps/sig-node/0014-runtime-class.md
yujuhong: What I was trying to say is that "Windows" is not just container runtime feature. It can require Windows node (i.e., host OS), and as @derekwaynecarr pointed out, specific Windows configurations in the pod spec. Given that, there are more obvious signals that we can/should use for pod scheduling. Using RuntimeClass to capture node-level property may be out of scope?

raravena80: This sounds like you would have to some sort of validation in where the runtime specified in the RuntimeClass would have to comply with the NodeSpec and/or PodSpec. But this may contradict with the scope creep like @derekwaynecarr was mentioning.

tallclair:
Thanks for the feedback. I've scrubbed the "supported features" references from the initial proposal, and moved it to the "explore in the future" section. I feel good about this change, because supported features is a higher risk scope creep, and adds a lot of complexity to the API and design. For the initial implementation, my top priority is enabling multiple runtime selection.

PTAL.

keps/sig-node/0014-runtime-class.md
dchen1107: +1 on node.k8s.io

keps/sig-node/0014-runtime-class.md
dchen1107: Yes, the initial scope requires a manual configuration, but I think providing an example of how admin to use RuntimeClass might help some of the readers to understand the entire workflow better.

keps/sig-node/0014-runtime-class.md
dchen1107: +1 on explicitly listing what we plan to support in alpha. For example, I am not very clear on if we plan to address Windows Containers usage here. It was mentioned above, but later only focusing on trusted vs. unstrusted and gvisor vs. kata.

keps/sig-node/0014-runtime-class.md
dchen1107: +1 on both @vikaschoudhary16 and @tallclair said. More specifically, we might extend runtimeclass in the future for backward-incompatible features, but let's keep the alpha version's scope smaller. We can iterate it later.

keps/sig-node/0014-runtime-class.md
derekwaynecarr: i think windows support really needs a clarified pod spec on what has meaning for windows.  i don't think its the job of runtime class to do much with windows imho.

derekwaynecarr:
I think the design provides enough detail to proceed on an alpha implementation for the identified use case.  I think further use cases will need more study.

/approve
/lgtm

PR #2297, opened 2018-06-21
godep: add instruction to commit after godep save

### LINKED ISSUE #1213: Wrong order of workflow steps
The [workflow instructions](https://github.com/kubernetes/community/blob/master/contributors/devel/development.md#workflow) suggest to first run `make verify` and later commit the change.

However `make verify` will stop asking for the commit:

```
Verifying hack/make-rules/../../hack/verify-staging-godeps.sh

Unexpected dirty working directory:

   M pkg/kubelet/pleg/generic.go

Commit your changes in another terminal and then continue here by pressing enter.
```

so it would be more useful to advise to commit first, and later `--amend` as necessary.

nikhita:
Fixes  #1213

Would be nice to be explicit about committing before generating staging godeps. The staging godeps script will also stop to ask you to commit the changes first.

/assign @cblecker


contributors/devel/godep.md
```diff
@@ -156,6 +163,13 @@ transitively), you'll also need to update deps there:
 ./hack/update-staging-godeps.sh
 ```
 
+Commit the changes generated by the above script:
+
+```sh
+git add .
```
nikhita: is it needed to explicitly mention the commands? Maybe just asking to commit the changes is enough? :thinking:

nikhita: or `git commit --amend`? is there a best practice here - both changes should go into one commit? :)

cblecker: I would suggest that we don't call out specific commands. Sometimes you want to have them be in the same commit, other times not.

nikhita: > I would suggest that we don't call out specific commands. Sometimes you want to have them be in the same commit, other times not.

:+1: done.

PR #2301, opened 2018-06-21
New KEP for Dry-run

apelisse:


apelisse:
@kubernetes/sig-api-machinery-feature-requests @lavalamp @deads2k

keps/0041-dry-run.md
```diff
@@ -0,0 +1,59 @@
...
+
+Dry-run is a new feature that we intend to implement in the api-server. The goal
+is to be able to send requests to modifying endpoints, and see if the request
+would have succeeded (admission chain, validation, merge conflicts, ...) and/or
+what would have happened without having it actually happen. The response body
+for the request should be as close as possible to a non dry-run response.
+
+## Specifying dry-run
+
+Dry-run is triggered by setting the “dryRun” query parameter to true on
```
lavalamp: List the allowed values of the parameter and their meanings?

keps/0041-dry-run.md
```diff
@@ -0,0 +1,59 @@
...
+
+Admission controllers need to be modified to understand that the request
+is a “dry-run” request. While admission controllers are not supposed to
+have side-effects when triggered, some of them still do. Quotas for
+example uses the current request values to change the available
+quotas. Providing a parameter, either has an argument for static
+admission controllers or through a dryRun query-parameter for dynamic
+webhooks, will give them a chance not to have any side-effect.
+
+All admission controllers will have to be verified and changed. A new flag will
```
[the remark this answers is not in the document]
apelisse: I haven't addressed this comment yet, the refactor make it look like it was resolved. I'll think about it.

apelisse:
> Can we instead rev the admission review api object, and just test whether they support v2 of it?

This needs to be addressed.

keps/0041-dry-run.md
```diff
@@ -0,0 +1,66 @@
...
+side-effects when triggered, some of them still do. Quotas for example uses the
+current request values to change the available quotas. Providing a parameter,
+either as an argument for built-in admission plugins or through a dryRun
+query-parameter for dynamic webhooks, will give them a chance not to have any
+side-effect.
+
+All admission controllers will have to be verified and changed. A new flag will
+be added so that webhooks can explicitly register with dry-run support. If
+dry-run is requested on a non-supported webhook, the request will be completely
+rejected.
```
adohe: means the webhook will reject the request?

soltysh:
My only question is how this will correlate with the existing `--dry-run` flag, at least for those commands that do implement that?

apelisse:
@soltysh I'm planning on coming to the next CLI meeting to talk about that. I'll probably also write a short proposal before that.

Eventually we'll probably want the `--dry-run` flag to use server-side dry-run and completely remove the existing logic. In the mean time, I think we'll probably have two flags (one for client side and one for server-side), and we'll take some time to change the behavior of the `--dry-run` flag: client-side initially, same as today, and then server-side later.

liggitt:
> Eventually we'll probably want the --dry-run flag to use server-side dry-run and completely remove the existing logic.

we don't want to lose the ability to generate/output things from those client commands without ever talking to a server. it's possible --dry-run isn't the best name for that, but that's what drives that behavior today

apelisse:
> we don't want to lose the ability to generate/output things from those client commands without ever talking to a server.

You're right, I'm not disagreeing, but what use-cases do you have in mind?

juanvallejo:
> we don't want to lose the ability to generate/output things from those client commands without ever talking to a server. it's possible --dry-run isn't the best name for that, but that's what drives that behavior today

Could we require --local to be passed along with dry-run for commands that currently support those cases?

soltysh:
Next sig-cli meeting sounds like a perfect place to have all this discussion.

apelisse:
I wrote this short [proposal](https://docs.google.com/document/d/1R8Q-thJCn2qCtvNkH5zRxcfF15Bl0yG_sQIjK4YEruI/).

keps/0041-dry-run.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Storage
+
+The storage layer will be modified, so that it can know if request is dry-run,
+most likely by looking for the field in the “Options” structure (missing for
+some handlers, to be added). If it is, it will NOT store the object, but return
+success. That success can be forwarded back to the user.
```
liggitt: * Will update/delete preconditions (resourceVersion, uid) be checked?
* What object will be returned for create dryRun operations? Will it have a populated uid? A populated resourceVersion?

keps/0041-dry-run.md
```diff
@@ -0,0 +1,66 @@
...
+Admission controllers need to be modified to understand that the request is a
+“dry-run” request. While admission controllers are not supposed to have
+side-effects when triggered, some of them still do. Quotas for example uses the
+current request values to change the available quotas. Providing a parameter,
+either as an argument for built-in admission plugins or through a dryRun
+query-parameter for dynamic webhooks, will give them a chance not to have any
+side-effect.
+
+All admission controllers will have to be verified and changed. A new flag will
+be added so that webhooks can explicitly register with dry-run support. If
```
liggitt: Detail the change to the webhook config API (and maybe the plan for default values… false in v1beta1, true+deprecated in v1, and removed in v2… all must support dry run by then)

liggitt: Also detail what happens if dry run is requested and one of the registered webhooks doesn't support dryRun mode (status code, type of error)

keps/0041-dry-run.md
```diff
@@ -0,0 +1,66 @@
...
+finer-grained mechanism later, if necessary.
+
+## Admission controllers
+
+Admission controllers need to be modified to understand that the request is a
+“dry-run” request. While admission controllers are not supposed to have
+side-effects when triggered, some of them still do. Quotas for example uses the
+current request values to change the available quotas. Providing a parameter,
+either as an argument for built-in admission plugins or through a dryRun
+query-parameter for dynamic webhooks, will give them a chance not to have any
```
liggitt: Detail the change to the AdmissionReview API object

keps/0041-dry-run.md
apelisse: I've started working on that:

> Will update/delete preconditions (resourceVersion, uid) be checked?

Yes,

> What object will be returned for create dryRun operations?

I'm expecting the object to be very close to the object that would have been created if dry-run had not been specified. 

> Will it have a populated uid?

I'll double check that tomorrow

> A populated resourceVersion?

As far as I understand, resourceVersion is meant to be storage specific, we can decide that the "dry-run storage" populate this.

keps/0041-dry-run.md
```diff
@@ -0,0 +1,66 @@
...
+what would have happened without having it actually happen. The response body
+for the request should be as close as possible to a non dry-run response.
+
+## Specifying dry-run
+
+Dry-run is triggered by setting the “dryRun” query parameter on modifying
+verbs: POST, PUT, PATCH and DELETE.
+
+This parameter is a string, working as an enum:
+- All: Everything will run as normal, except for the storage that won’t be
```
liggitt: @kubernetes/api-reviewers question here on what to call the dry run option to include all phases of the apiserver handling chain. Suggestions I've seen so far are `All`, `True`, and `*`

keps/0041-dry-run.md
liggitt: not sure what we want for fields that are generated, and will not match the results of an actual submit (uid, creationTimestamp, generateName->name, etc)

calebamiles:
@liggitt, @juanvallejo would you be alright merging the KEP in it's current form. It's been open for almost a month now and I would like to see more changes made to KEP once this initial draft is checked in. @apelisse, I do wish you had used the template for consistency but whatever. You can also drop the change to `NEXT_KEP_NUMBER` as I'm about to formally declare bankruptcy on anyone besides designated KEP editors working with the file.

liggitt:
> would you be alright merging the KEP in it's current form.

yes, spoke with @apelisse yesterday and with the status in provisional, this seems fine for iteration.

follow-ups needed:
* move into a `sig-api-machinery` subfolder (along with `0006-apply.md`)
* nail down the actual API bits (from https://github.com/kubernetes/community/pull/2301#pullrequestreview-137666788) and get an ack from @kubernetes/api-approvers

/lgtm

keps/0041-dry-run.md
apelisse: @jennybuckley

keps/0041-dry-run.md
apelisse: @jennybuckley

PR #2309, opened 2018-06-25
remove initial-resources proposal

wackxu:
initial-resource plugin has been removed in https://github.com/kubernetes/kubernetes/pull/58784.


DirectXMan12:
Hmm... I not sure what the appropriate process is here?  Do we keep the proposal around and mark as removed, or do we remove the proposal entirely?

nikhita:
> Hmm... I not sure what the appropriate process is here? Do we keep the proposal around and mark as removed, or do we remove the proposal entirely?

@cblecker probably knows?

cblecker:
I'd suggest marking as historical/deprecated as opposed to removing. Makes it easier for those looking it up to know what happened to it (rather than look through git history)

PR #2310, opened 2018-06-25
Add proposal for new PR test jobs triggers

matthyx:
as discussed in https://github.com/kubernetes/test-infra/issues/3827


contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+1. modify `trigger` plugin automatic comment and `tide` documentation to explain the new `ok-to-test` label requirements
+1. batch add `ok-to-test` label to non-members trusted PRs
+1. modify `trigger` plugin to **only** look for `ok-to-test` label when checking PR trust (code simplification)
+1. stop adding `needs-ok-to-test` label to untrusted PRs
+1. batch remove `needs-ok-to-test` label to untrusted PRs
+1. remove `needs-ok-to-test` from `label_sync`
+
+## Future evolutions
+
+In the future, we might decide to require the new label for all PRs, which means that organisation members will also need the `ok-to-test` label applied to their PRs before automatic testing can be triggered.
```
fejta: Automatically add the `ok-to-test` label to PRs created by org members?

contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+
+> Using a positive label instead of a negative one avoids all the race conditions that force us to list comments and check for `/ok-to-test` instead of just checking for a label.
+
+## Proposal
+
+We suggest introducing a new label named `ok-to-test` that would be required on any non-member PR before automatic test jobs can be started by the `trigger` plugin.
+
+This label could only be added by members of the *trusted organization* for the repository.
+
+We will also deprecate the `/ok-to-test` command.
```
fejta: How will people add/remove the `ok-to-test` label?

contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+## Objective
+
+This document describes a major change to the way the `trigger` plugin determines if test jobs should be started on a pull request (PR).
+
+## Current situation
+
+PR test jobs are started by the trigger plugin on **trusted PR** events, or when a untrusted PR becomes trusted.
+> A PR is considered trusted if the author is a member of the *trusted organization* for the repository or if such a member has left an `/ok-to-test` command on the PR.
+
+You can easily spot an untrusted PR by its `needs-ok-to-test` label.
```
fejta: We should continue to provide people with an expedient means of searching for PRs that need the `ok-to-test` label.

There is not a way to search for PRs that are "authored by non-members without an `ok-to-test`" label.

contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+This label could only be added by members of the *trusted organization* for the repository.
+
+We will also deprecate the `/ok-to-test` command.
+
+## Caveats
+
+Switching from a negative to a positive label could create confusion and add clutter to the PR.
+
+@cblecker
+> .. label overload. Having a label just to say testing is okay.. bleh.
```
fejta: Could we make this a github context? This would:
* allow us to only flag PRs that are unsafe for testing
* accept ok-to-test from either an issue or review comment
* simplify searching (just look for the contexts, rather than all comments

BenTheElder: other pros/cons:

\+ A github context would also not be human editable which avoids accidentally removing it when editing labels

\- A context is not searchable by humans AFAIK, right now we have [go.k8s.io/needs-ok-to-test](https://go.k8s.io/needs-ok-to-test) for contribex to help onboard new people, if it's still a label we can update this github search link to still work..

\+/- github contexts are cleared when a PR is closed/reopened

\- our current workflow mostly involves labels and commands that manipulate them, should approval / lgtm also be a context..?

contributors/design-proposals/testing/trigger-pr-tests.md
cjwagner: We will still need this command. This is how org members will add the `ok-to-test` label to authorize non-member PRs.

contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+This is why documentation and education are key here, and like @cjwagner reminds us:
+> we are already leaving a comment on the PR if the author needs `/ok-to-test` from a member, so authors are already receiving specific instructions on how to proceed
+
+We would just have to modify this comment.
+
+## Migration path
+
+1. declare `ok-to-test` in `label_sync`
+1. modify `trigger` plugin to check `ok-to-test` label as another condition for a trusted PR
+1. modify `trigger` plugin automatic comment and `tide` documentation to explain the new `ok-to-test` label requirements
```
cjwagner: This shouldn't affect Tide?

contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+1. batch add `ok-to-test` label to non-members trusted PRs
```
cjwagner: We may need to write a tool to do this or at least to identify non-member PRs that are missing `needs-ok-to-test`.

contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+
+This is why documentation and education are key here, and like @cjwagner reminds us:
+> we are already leaving a comment on the PR if the author needs `/ok-to-test` from a member, so authors are already receiving specific instructions on how to proceed
+
+We would just have to modify this comment.
+
+## Migration path
+
+1. declare `ok-to-test` in `label_sync`
+1. modify `trigger` plugin to check `ok-to-test` label as another condition for a trusted PR
```
cjwagner: I don't know how we would support both types of checks simultaneously. I think we may need to make an instantaneous switch because the positive and negative processes are conflicting (how would we handle PRs missing both an `ok-to-test` and `needs-ok-to-test` label?).

contributors/design-proposals/testing/trigger-pr-tests.md
```diff
@@ -0,0 +1,92 @@
...
+1. modify `trigger` plugin to **only** look for `ok-to-test` label when checking PR trust (code simplification)
+1. stop adding `needs-ok-to-test` label to untrusted PRs
+1. batch remove `needs-ok-to-test` label to untrusted PRs
+1. remove `needs-ok-to-test` from `label_sync`
+
+## Future evolutions
+
+In the future, we might decide to require the new label for all PRs, which means that organisation members will also need the `ok-to-test` label applied to their PRs before automatic testing can be triggered.
+
+This would allow us to choose at which point automatic test jobs should be triggered, saving even more API tokens.
```
cjwagner: I'm not sure what the first clause of the this sentence is referring to. We would save API tokens by avoiding org membership checks.

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Good point, I will add it.

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Ok, my bad, I thought we wanted to stop parsing the comments (and still not parse reviews) to save the API tokens.

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Ok then this is a no show. I will add this option in the doc and explain why it's not a good idea.

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Apparently not, but I think saw the label documented there...

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Hmm, if we do things in the right order, normally this double labeling should not happen.

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: +1

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Ok, makes more sense. I had a wrong understanding on how API tokens are consumed... maybe I should search and document here my findings. And also explain why it's a problem to use too much of them?

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Indeed, so we might propose `ok-to-test` required for everybody.

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Indeed the [way](https://help.github.com/articles/applying-labels-to-issues-and-pull-requests/) to add labels by hand isn't really practical... I thought it was available directly in the issue/PR page.

BenTheElder: yeah, there's a way in the page but it performs a set on the labels from the browser, not an "add to whatever currently exists" so if the labels change while you're editing them (IE prow labels something) and then you finish it will set the labels you see, and remove any you didn't have locally 🙄

cjwagner: > Ok, my bad, I thought we wanted to stop parsing the comments (and still not parse reviews) to save the API tokens.

We will save a lot of API tokens by avoiding listing the comments, reviews, and review comments every time we need to check if a PR is trusted.
Instead we can handle the corresponding events (issue_comment, pull_request_review, and pull_request_review_comment) with a single GenericCommentEvent handler that adds the `ok-to-test` label in response to `/ok-to-test` from an org-member. From then on whenever we need to check if the PR is trusted, we just see if it is authored by an org member or has the `ok-to-test` label (no need to list everything).

contributors/design-proposals/testing/trigger-pr-tests.md
cjwagner: I agree that double labeling could be avoided. I'm asking how we would handle the lack of either label.
If a PR is missing both `ok-to-test` and `needs-ok-to-test` there is no way to tell if it should be trusted or not.

contributors/design-proposals/testing/trigger-pr-tests.md
cjwagner: Sure. It is a problem to use too many tokens because we have an hourly limit of 5000 and that limit is shared by all components that are using the @k8s-ci-robot account. If we use too many tokens here other Prow components are negatively impacted. Either failing to complete important tasks involving GitHub or at a minimum being rate limited.

contributors/design-proposals/testing/trigger-pr-tests.md
matthyx: Well, these will be caught by the batch, if there is no `needs-ok-to-test` label, add `ok-to-test`.

BenTheElder: another thought, this may affect other prow consumers, we should write down exaclty how we do each migration step and try to broadcast this.

@stevekuznetsov @sebastienvas

cjwagner: > Use the commenter to /ok-to-test to any PR more than 4 hours old that doesn't have needs-ok-to-test.

We only want to do this to non-member PRs. I don't think there is any way to specify that in a GH search query so the commenter is probably not suitable?

> Well, these will be caught by the batch, if there is no needs-ok-to-test label, add ok-to-test.

Ok, that makes sense. The batch step needs to come before this step then.

matthyx:
Thanks for all your reviews! I have detailed the migration steps...
Now the last blocker (for me) is how to help members spot untrusted PRs needing review and `/ok-to-test`...

stevekuznetsov:
Discussed this in our SIG meeting this week, the proposal here LGTM now -- @fejta @cblecker do we have any other outstanding comments for this KEP?

matthyx:
@mattlandis proposed during our SIG to keep `need-ok-to-test` label for user experience, and we could use it to spot PRs needing `/ok-to-test` as it's the case today.
So finally the new label `ok-to-test` would mostly be used to simplify our code, and save API tokens!
I will be on vacation in two weeks, so roll-out won't start until mid-August.

cblecker:
The biggest negative here is label sprawl. I think the advantages though, are worth it in this case.

We need to ensure that a PR has only one status (needs-ok-to-test OR ok-to-test). Should these labels be classed (`something/something`)? I'd like to but I can't think of a good name.

We should also clarify that it's org members or repo collabs by default (recent code change).

fejta:
> keep need-ok-to-test label for user experience

What is the user experience benefit?

> spot PRs needing `/ok-to-test`

Search for `-label:ok-to-test` and/or notice the absence of this label. Functionally equivalent to the current strategy of search `label:needs-ok-to-test` and note the absence of this label means testing is ok.

> classed (something/something)

`trust/ok-to-test` and `trust/needs-ok-to-test`? or maybe /yes /no?

cblecker:
The label wouldn’t be applied to member PRs, so if you’re a member/repo collab, you don’t get either label (so searching for absence doesn’t work)

BenTheElder:
why not add `ok-to-test` to member / collaborator PRs as well?

fejta:
Just add the `ok-to-test` label to all PRs that are... ok to test?

matthyx:
Alright, I didn't want to light up another comment war...

> We need to ensure that a PR has only one status (needs-ok-to-test OR ok-to-test).

Absolutely, they will need to be mutually exclusive, even if the only label used as a reference should be `ok-to-test`.

> Should these labels be classed (something/something)?

Why change something that seems logical?

> What is the user experience benefit?

Non members can see this dark red label and immediately understand why tests are not run. Once we issue a `/ok-to-test` command, the dark red label transforms into a green label reminding the update of the status. To me it's even more logical for the user experience to see the progression.

> Why not add ok-to-test to member / collaborator PRs as well?

It was agreed before to only add this to non-members PR, at least for the beginning. It will require to add automatically the label to member authored PRs, as stated in the KEP.

Let's see what @cjwagner has to comment on that, and I will probably update the KEP to keep the `needs-ok-to-test` label for consistency and ease of comprehension.

cjwagner:
@matthyx All of that SGTM.
I like the idea of prefixing most of our labels with a class (`something/something`), but adding a class would necessitate a label migration which is easy for us, but potentially confusing for users. I'd be happy either way on this part. Since there are only 2 states/labels the class probably isn't important.

matthyx:
I have updated the KEP to keep `needs-ok-to-test` label, which will make our lives easier, if `trigger` (or maybe another command would be better?) could check for both labels applied to a PR.
It could happen if we're not careful when adding/removing labels after a `/ok-to-test`.

matthyx:
@stevekuznetsov @cjwagner now that I am officially back from vacation, I will move forward with this...
What's the process to continue with this KEP?

cjwagner:
> What's the process to continue with this KEP?

Hmm, I'm not sure. The [README](https://github.com/kubernetes/community/tree/master/contributors/design-proposals#kubernetes-design-documents-and-proposals) has a TODO to document the process.

BenTheElder:
@spiffxp

matthyx:
/assign @bgrant0607
Let's restart that process, I have applied the template and reformatted the KEP.

spiffxp:
if we're going to move forward with this, early in the release cycle is as good a time as any, I'm going to raise this during contribex meeting tomorrow one last time?

spiffxp:
/approve
/hold
I have added to https://github.com/orgs/kubernetes/projects/1 to discuss during tomorrow's contribex meeting

cblecker:
The one thought that I had on this just recently -- how would we handle things in the case of a prow outage or dropped hooks? I think this would actually handle those situations better, as instead of a negative state, we're looking for a positive state. If we don't have the label, we would err on the side of *not* testing, which is safer.

Is this accurate?

spiffxp:
I heard no objections to this on the meeting, and am looking for /lgtm's on this

keps/sig-testing/0028-20180625-new-label-for-trusted-pr-identification.md
```diff
@@ -0,0 +1,127 @@
...
+1. PR: switch to `ok-to-test`
+   * remove `needs-ok-to-test` from `missingLabels` in `prow/config.yaml`
+   * edit `prow/config/jobs_test.go`
+   * edit `prow/cmd/deck/static/style.css`
+   * edit `prow/cmd/tide/README.md`
+   * code changes in `trigger`:
+      * `/ok-to-test` adds `ok-to-test`
+      * PR trust relies on `ok-to-test`
+      * if PR has both labels, drop `needs-ok-to-test`
+      * edit all references to `needs-ok-to-test`
```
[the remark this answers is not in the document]
matthyx: Thanks @tpepper I will add these steps in the next revision of the KEP... for the moment I am just going to merge it and start working.

PR #2314, opened 2018-06-26
svcacct token issuer should support oidc discovery

mikedanese:
This was omitted from the original document for the purpose of expediting it's merge. Reintroducing now that we've made progress. 

This change allows clients with knowledge of oidc discovery to validate service account tokens. Simpler clients can choose to ignore the discovery document and validate tokens against the keys directly by fetching the JWKs URL.

@kubernetes/sig-auth-feature-requests @kubernetes/sig-auth-api-reviews

### DIFF
--- a/contributors/design-proposals/auth/bound-service-account-tokens.md
+++ b/contributors/design-proposals/auth/bound-service-account-tokens.md
@@ -213,6 +213,67 @@ The token payload will be:
 }
 ```
 
+### Add Service Account signing key retrieval API
+
+A non-resource API will be added to the apiserver that will expose an [OIDC
+discovery](https://openid.net/specs/openid-connect-discovery-1_0.html)
+compatible configuration. This will allow clients to retrieve a list of
+currently valid signing keys as an array of JWKs. For example, if the service
+account issuer is configured with the name `https://dev.cluster.internal`, the
+apiserver could expose the following configuration:
+
+```
+> GET /.well-known/openid-configuration
+{
+  "issuer": "https://dev.cluster.internal",
+  "jwks_uri": "https://dev.cluster.internal/serviceaccountkeys/v1",
+  "authorization_endpoint": "urn:kubernetes:programmatic_authorization",
+  "response_types_supported": [
+    "id_token"
+  ],
+  "subject_types_supported": [
+    "public"
+  ],
+  "id_token_signing_alg_values_supported": [
+    "RS256",
+    "ES256"
+  ],
+  "claims_supported": [
+    "sub",
+    "iss"
+  ]
+}
+> GET /serviceaccountkeys/v1
+{
+  "keys": [
+    {
+      "kty": "RSA",
+      "alg": "RS256",
+      "use": "sig",
+      "kid": "ccab4acb107920dc284c96c6205b313270672039",
+      "n": "wWGfvdCEjJJy7CQpGcTq6GghmqWLi9H4SNHNTtFMfIDPsv-aWj1e_iSO22505BlC9UcL9LvlSyVH8HmQUy5916YNqxCbhPFPabBAv0a-CpVuzbbyhpDNP3RkRIJgxlzPDh_dB11cbPTQ3yz0A0JARX3QNZfIQ8LFiZ1vh0iZAIm-I3eZeI4QZigImNDviZstSoHB2Ny1tsRmpZn-neYZCxYq717buFctnCVvot4iCwcQpeaGdniqYNDxzN4KlQwwDeCVJm-K0rG9nkiqZ_rq8SgCxi_l7NyF2ZURNTTzZyDwYfBR7jZUhbmjxIDoDZalsa1Tzzy1vzqBfxkFD5Z03w",
+      "e": "AQAB"
+    },
+    {
+      "kty": "RSA",
+      "alg": "RS256",
+      "use": "sig",
+      "kid": "8770f6158b125040b98e50a1e0e6790ff2f9ea09",
+      "n": "vpgsIIPqDO3A3dEuRCIZvQinyfME0BjH_RbeyLAAvrQx-Sv08ryFPjplqxm5t9mC0yULrhOmaIZCVfIuYn5n_dOblZNhpIpoy89bP0qNwV7gxsNv-0Tdu9nj4ymxeoaby6SFiv_c8P2JZ0CSqif_qXgj-o0TqU20FEv1hkizzQWDzFsKZ__IABAkdKfpGqQTOBTylFG9HFLV1tdh9AAdhVRf40982rksaOSDWvN_sfxiz6midGPgG0OOnMnwKAW-3BBNNd_uUrD9baSXZPFA8zo9dlkhQhfrFgg_U6ke4M5DPyFiPKOVitBzpL1Kth_patVZvnBGXtq2frbReF-6pw",
+      "e": "AQAB"
+    },
+    {
+      "kty": "RSA",
+      "alg": "RS256",
+      "use": "sig",
+      "kid": "68241231bbf0df8f9123d018cf9e601e2aa3673a",
+      "n": "rHozcxeim9flTWQxqC1ObpGP0EjpkUHVHpHNX8WGHHnMcVi63_9PaHn2cJeFuPF9qkI1dMPXeoX0m33N0tgM9-KOSmTg1oGbyJGoUYMFI-A7tdxoobb91LGjeNWJC0la0gLOGPcQ6zQLEU5RGftCZT0wElxMuwEH7FZoVBn5i8Ddvc2ADd4bFW0f_FckwFYN1rIU1uLf6coku_1xBfae3b_JiBq38QOGXPdPgxfPzmJEvIz_LB2WOIcwhl97DY32BQU7l_lNLYz6wMg9HeCKolypPIFEGNxLj1TcuOhwP5-BnSja-PvrB-1FN1JyzlL3nh__uJv8SoKPn0CoBPueWw",
+      "e": "AQAB"
+    }
+  ]
+}
+```
+
 ## Service Account Authenticator Modification
 
 The service account token authenticator will be extended to support validation


contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -213,6 +213,67 @@ The token payload will be:
...
+  ]
+}
+> GET /serviceaccountkeys/v1
+{
+  "keys": [
+    {
+      "kty": "RSA",
+      "alg": "RS256",
+      "use": "sig",
+      "kid": "ccab4acb107920dc284c96c6205b313270672039",
```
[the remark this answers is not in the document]
liggitt: I assume these would be populated from the `--service-account-key-file` flag, which is used for verifying presented tokens, and does take multiple files (or files with multiple key PEM blocks):

> File containing PEM-encoded x509 RSA or ECDSA private or public keys, used to verify ServiceAccount tokens. The specified file can contain multiple keys, and the flag can be specified multiple times with different files.

a server restart is required to pick up new public keys today

> How would the apiserver determine which key is used for a TokenRequest response?

the signing key specified by `--service-account-signing-key-file` is used

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -213,6 +213,67 @@ The token payload will be:
...
+currently valid signing keys as an array of JWKs. For example, if the service
+account issuer is configured with the name `https://dev.cluster.internal`, the
+apiserver could expose the following configuration:
+
+```
+> GET /.well-known/openid-configuration
+{
+  "issuer": "https://dev.cluster.internal",
+  "jwks_uri": "https://dev.cluster.internal/serviceaccountkeys/v1",
+  "authorization_endpoint": "urn:kubernetes:programmatic_authorization",
```
liggitt: is this compatible with https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata:
```
authorization_endpoint
    REQUIRED. URL of the OP's OAuth 2.0 Authorization Endpoint
```
that specifies a URL, not a URI (which would allow a URN)

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -213,6 +213,67 @@ The token payload will be:
...
+    "public"
+  ],
+  "id_token_signing_alg_values_supported": [
+    "RS256",
+    "ES256"
+  ],
+  "claims_supported": [
+    "sub",
+    "iss"
+  ]
```
liggitt: if `grant_types` is omitted, the default value is `["authorization_code", "implicit"]`, which means that `token_endpoint` is also required

liggitt: using an openid discovery doc as a vehicle for publishing public keys gets weird when the tokens we're wanting to allow someone to validate can't actually be obtained via an openid flow... we'd have to do things like advertise `grant_types_supported:[]` and make our `authorization_endpoint` point to a URL that rejected all requests. that seems pretty weird.

mikedanese:
@mdietz

contributors/design-proposals/auth/bound-service-account-tokens.md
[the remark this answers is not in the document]
micahhausler: > the signing key specified by --service-account-signing-key-file is used

Sorry I wasn't clear on the last question. When given a list of keys, which one is used to sign a `TokenRequest`? Ex: I want to rotate out a certain key and stop signing new tokens with it, but still keep it around for verification of existing tokens.

Also, is `kid` derived from the key? I'm just fuzzy on where that comes from, and how it is coordinated across API servers.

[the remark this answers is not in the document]
liggitt: > When given a list of keys, which one is used to sign a TokenRequest

`--service-account-signing-key-file` takes a single key, and is used to sign TokenRequests.

`--service-account-key-file` takes multiple keys for verification purposes.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -213,6 +213,67 @@ The token payload will be:
...
+compatible configuration. This will allow clients to retrieve a list of
+currently valid signing keys as an array of JWKs. For example, if the service
+account issuer is configured with the name `https://dev.cluster.internal`, the
+apiserver could expose the following configuration:
+
+```
+> GET /.well-known/openid-configuration
+{
+  "issuer": "https://dev.cluster.internal",
+  "jwks_uri": "https://dev.cluster.internal/serviceaccountkeys/v1",
```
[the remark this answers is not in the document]
mikedanese: That should work.

contributors/design-proposals/auth/bound-service-account-tokens.md
mikedanese: However, grant types are extensible:

https://tools.ietf.org/html/rfc6749#section-8.3

An alternative is to make `"grant_types":["urn:kubernetes:grant_type:programmatic_authorization"]` and make `authorization_endpoint` point somewhere bogus.

mikedanese: @BrenoDeMedeiros do you have any thoughts on how we should express the Kubernetes id_token flow in a openid-configuration to support oidc discovery?

contributors/design-proposals/auth/bound-service-account-tokens.md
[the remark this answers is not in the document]
micahhausler: ah that makes sense. I wasn't paying close enough attention and was merging `--service-account-key-file` and `--service-account-signing-key-file`.

contributors/design-proposals/auth/bound-service-account-tokens.md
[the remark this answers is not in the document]
enj: You want both `system:unauthenticated` and `system:authenticated`.

contributors/design-proposals/auth/bound-service-account-tokens.md
enj: Does `authorization_endpoint` even make sense?  There is no way to actually do an OAuth flow with the API master.

enj:
While discovery in general sounds like a good idea, especially to allow distributed auth models, the `OpenID Provider Metadata` is clearly focused on Open ID flows.

If we take OpenShift as an example, today you can do:

```
$ oc get --raw '/.well-known/oauth-authorization-server'
```

Which returns:

```json
{
  "issuer": "https://<master>",
  "authorization_endpoint": "https://<master>/oauth/authorize",
  "token_endpoint": "https://<master>/oauth/token",
  "scopes_supported": [
    "user:check-access",
    "user:full",
    "user:info",
    "user:list-projects",
    "user:list-scoped-projects"
  ],
  "response_types_supported": [
    "code",
    "token"
  ],
  "grant_types_supported": [
    "authorization_code",
    "implicit"
  ],
  "code_challenge_methods_supported": [
    "plain",
    "S256"
  ]
}
```

But that makes sense because a user can actually go through an OAuth flow with the server and have an OAuth token returned to them.  No such flows exists for SA tokens, and I do not think anyone actually wants such a flow to exist.

Basically, Kube SA tokens are JWTs, but that does not have any specific relationship to OIDC.  Perhaps there is a different way to discover a `jwks_uri`?

mikedanese:
OIDC discovery would be exposed because it is widely adopted and supported by target integrations (e.g. AWS, vault, dex etc...). I think we need to meet integrations where they are rather then force them to implement a new spec to validate k8s service account identities. I'm open to considering other discovery mechanisms but level of support is my top consideration.

enj:
> OIDC discovery would be exposed because it is widely adopted and supported by target integrations (e.g. AWS, vault, dex etc...)

I do not think this actually succeeds in making any of those integrations easier.  This still have to know that this is a Kube OIDC discovery doc, and that only meaningful field is `jwks_uri`.  @liggitt's concerns in https://github.com/kubernetes/community/pull/2314#discussion_r199673997 basically point out that this proposed endpoint simply fails to meet the OIDC contract.  Having custom grant types just further couples any integration with Kube specifics.  So if integrations already need to know that "this is a Kube SA token and all you can do is verify it - you cannot do any other OIDC things with it," than at best you have allowed them to reuse some existing structs for decoding.

I guess with this doc a generic verify method could be written for OIDC like endpoints:

```go
func Verify(token, issuer string) bool {
    // get discovery doc
    // use it to get all info needed to verify token
}
```

I am not sure how critical reuse of any such logic would be to any integrator.

liggitt:
> I do not think this actually succeeds in making any of those integrations easier.

It would support discovery/loading of the info needed to verify a token obtained out of band (this is the same level of info the kube OIDC token verifier uses from the issuer discovery docs)

enj:
> It would support discovery/loading of the info needed to verify a token obtained out of band (this is the same level of info the kube OIDC token verifier uses from the issuer discovery docs)

Right, no disagreement there.  I am mostly referring to the fact that integrators still have Kube specific knowledge in their codebase, so I not seeing how this is different than a custom Kube specific endpoint (non-resource or resource) that does not pretend to be OIDC.

mikedanese:
> I am mostly referring to the fact that integrators still have Kube specific knowledge in their codebase

I think this is not true. This doc would be compatible with AWS and vault without any changes on their side.

enj:
> I think this is not true. This doc would be compatible with AWS and vault without any changes on their side.

Can you describe their use case?  Or point me to the docs or code?  I can see how they could use this endpoint (because it is close enough to everything else and they only care about the `jwks_uri`).  As a user I would find it very confusing to be presented an OIDC discovery doc from a server that cannot perform any OIDC flow.

micahhausler:
@enj A use case we're exploring based on this proposal is outlined here https://docs.google.com/document/d/1V_B24zNLInLPRIPRXQeKj4JAJDUvKe-hZPJvrgx80yQ/edit

We would like for Kubernetes would implement enough of OIDC to act as a federated identity provider in AWS.

liggitt:
> How do people feel about only exposing a JWKs URL and proceeding from there?

I think that's a much easier step to reason about, and makes sense to me

cceckman:
If a relying party wants to validate the token (including the `iss` claim), they'd still need a provider-specific discovery mechanism to find that issuer. 

Could we offer a subset of the OIDC metadata (`issuer`, `jwks_uri`, anything else necessary) at a non-standard endpoint? e.g. `/serviceaccountissuer/{keys,metadata}` rather than `/.well-known/openid-configuration`?

cceckman:
Some notes from this week's sig-auth meeting- I'm probably misrepresenting these a bit, apoligies:

- @liggitt: "We're using this for validation, not for request flows. Not going to be helpful to offer at a nonstandard path" (as I suggested above; fair.) "Work out the minimal spec-compliant thing that won't cause validators to crash."
- @mikedanese: "Authz endpoint is an actual URL, but always returns 403 / 404. Nothing that says grant types have to be well-known."

liggitt:
can we move this into a KEP in enhancements? would be good to reconcile this with the external signing proposal to make sure whatever we end up with there supports this goal as well

mikedanese:
Ya, was planning on doing that.

contributors/design-proposals/auth/bound-service-account-tokens.md
```diff
@@ -213,6 +213,67 @@ The token payload will be:
 }
 ```
 
+### Add Service Account signing key retrieval API
+
+A non-resource API will be added to the apiserver that will expose an [OIDC
+discovery](https://openid.net/specs/openid-connect-discovery-1_0.html)
```
Engpablo, line 219: #8343

Issue #2321, opened 2018-06-27
New Contributor Playground


cblecker:
https://kubernetes.slack.com/archives/C1TU9EB9S/p1529008553000790

There was a Slack discussion on setting up a new contributor playground, not in k/community. Opening up an issue to track it so we don't forget.

cblecker:
@guineveresaenger @carolynvs @parispittman Do you know what the "due date" for this is? When will a playground be needed by?

carolynvs:
I am doing an OSS workshop at GopherCon the last week of August that if it were done by then, I would point people towards it. But otherwise I would guess KubeCon Seattle, right @guineveresaenger?

FYI, since this involves setting up a new repo, I'm not sure how I can help move this forward until that's created. Let me know if I'm missing something and I can work on it this week on my vacation!

parispittman:
they will have a new contributor workshop in Shanghai that @jberkus is leading

jberkus:
We definitely need one by Shanghai.  Having one by GopherCon would be even better.

cblecker:
This is done: https://github.com/kubernetes/org/issues/26

/close

PR #2322, opened 2018-06-27
Remove new-contributor-playground

cblecker:
Removes the old contributor playground that was extremely spammy.

Opened #2321 to set up a new one.


idvoretskyi:
cc @guineveresaenger

PR #2330, opened 2018-06-29
GitHub handle is required at the request body

idvoretskyi:
Sometimes it's much easier to copy the GH handle from the request body than from a subject. Adding an update the requires the GitHub handle at the community membership request body.

cc @cblecker @calebamiles 

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>


cblecker:
Post merge +1 :)

PR #2331, opened 2018-06-29
Update KEP kubeadm join --master

fabriziopandini:
Update to the kubeadm join --master KEP

/area kubeadm
/sig sig-cluster-lifecycle

@kubernetes/sig-cluster-lifecycle-pr-reviews

/CC @luxas @timothysc


chuckha:
How would folks feel about `kubeadm join --control-plane` ? 

Thumbs up/Thumbs down would work as a response

neolit123:
> How would folks feel about kubeadm join --control-plane ?

i think `--master` is fine, unless `--control-plane` is considered more descriptive.

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+```yaml
+---
+kep-number: 15
+title: kubeadm join --master workflow
+status: accepted
+authors:
+  - "@fabriziopandini"
+owning-sig: sig-cluster-lifecycle
+reviewers:
+  - "@cha”
```
luxas: +luxas

luxas: on github @chuckha

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+  - "@jdtibier"
```
luxas: @detiber

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+## Motivation
+
+Support for high availability is one of the most requested features for kubeadm.
+
+Even if, as of today, there is already the possibility to create an HA cluster
+using kubeadm in combination with some scripts and/or automation tools (e.g.
+[this](https://kubernetes.io/docs/setup/independent/high-availability/)), this KEP was
+designed with the objective to introduce an upstream simple and reliable solution for
+achieving the same goal.
```
luxas: ... with important non-goals. Say clearly (also in the summary) that this doesn't solve every case or even the full end-to-end flow automatically

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+### Goals
+
+- "Divide and conquer”
+
+  This proposal - at least in its initial release - does not address all the possible
+  user stories for creating an highly available Kubernetes cluster, but instead
+  focuses on:
+
+  - Defining a generic and extensible flow for bootstrapping an HA cluster, the
```
luxas: `a cluster consisting of multiple masters` instead of HA?

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+  We expect higher-level and tooling will leverage on kubeadm for creating HA clusters;
+  accordingly, the `kubeadm join --master` workflow should provide support for
+  the following operational practices used by higher level tools:
+
+  - Parallel node creation
+
+    Higher-level tools could create nodes in parallel (both masters and workers)
+    for reducing the overall cluster startup time.
+    `kubeadm join --master` should support natively this practice without requiring
```
luxas: and how does this work?

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+    the implementation of any synchronization mechanics by higher-level tools.
+
+  - Replace reconciliation strategies
+
+    Especially in case of cloud deployments, higher-level automation tools could
+    decide for any reason to replace existing nodes with new ones (instead of apply
+    changes in-place to existing nodes).
+    `kubeadm join --master` will support this practice by making easier to replace
+    existing master nodes with new ones.
```
luxas: do you mean that graduating a node to master is a non-goal?
I don't get what you're saying here really

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+- Provide support both for dynamic and static bootstrap flow
+
+  At the time a user is running `kubeadm init`, s/he/an operator might not know what
```
luxas: `the user` or `the operator`

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+### Non-goals
+
+- This proposal doesn't include a solution for etcd cluster management (but nothing in this proposal should
+  prevent to address this in future).
+
+  > At the time of writing, the CoreOS recommended approach for etcd is to run
+  > the etcd cluster outside kubernetes (see discussion in [kubeadm office hours](https://goo.gl/fjyeqo)).
+
+- Nothing in this proposal should prevent users to run master nodes components
```
luxas: why is this in non-goals? I'd say this is a goal -- to be able to co-locate etcd and the API server _as well_

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+  and etcd on the same machines.
+
+- This proposal doesn't include a solution for API server load balancing (Nothing in this proposal
+  should prevent users from choosing their preferred solution for API server load balancing.
```
luxas: missing )

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+- This proposal doesn't address the ongoing discussion about kubeadm self-hosting; in light of
```
luxas: I think we should say at this point we won't support self hosting at all at this point and the foreseeable future

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+  divide and conquer goal stated before, it is not planned to provide initial support
+  for self-hosted clusters (but nothing in this proposal should prevent to address this in future).
+
+- This proposal doesn't address the ongoing discussion about storage of kubeadm TLS assets
```
luxas: We won't do secrets for a very long time. Say this doesn't address transferring the CA key and other required certs from one master to the other

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+  - Kubeadm binary can execute actions _only_ on the machine where it is running
+    e.g. it is not possible to execute actions on other nodes, to copy files across
+    nodes etc.
+  - During the join workflow, kubeadm can access the cluster _only_ using identities
+    with limited grants, `system:unauthenticated` or `system:node-bootstrapper`.
+
+- Upgradability
+
+  - How to setup an high available cluster in order to simplify the execution
```
luxas: If we implement  `kubeadm join --master` we MUST support upgrades pretty seamlessly

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+  provide initial support for clusters with TLS stored in secrets (but nothing in this
+  proposal should prevent to address this in future).
+
+- Nothing in this proposal should prevent practices that exist today.
+
+### Challenges and Open Questions
+
+- Keep the UX simple.
+
+  - _What are the acceptable trade-offs between the need to have a clean and simple
```
luxas: should you/we answer this?

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+  - _How to combine into a single UX support for both static and dynamic bootstrap
+    workflows?_
+
+- Kubeadm limited scope of action
+
+  - Kubeadm binary can execute actions _only_ on the machine where it is running
+    e.g. it is not possible to execute actions on other nodes, to copy files across
+    nodes etc.
+  - During the join workflow, kubeadm can access the cluster _only_ using identities
```
luxas: ??

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+<!-- TOC -->
+
+- [kubeadm join --master workflow](#kubeadm-join---master-workflow)
+    - [Metadata](#metadata)
+    - [Table of Contents](#table-of-contents)
+    - [Summary](#summary)
+    - [Motivation](#motivation)
+        - [Goals](#goals)
+        - [Non-goals](#non-goals)
```
luxas: can you add prerequisites?

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+#### Preparing for execution of kubeadm join --master
+
+Before invoking `kubeadm join --master`, the user/higher level tools
+should copy control plane certificates from an existing master node, e.g. bootstrap master
+
+> NB. kubeadm is limited to execute actions *only*
+> in the machine where it is running, so it is not possible to copy automatically
+> certificates from remote locations.
+
+Please note that strictly speaking only ca certificate and and service account key pair
```
luxas: hmm, also the front-proxy CA key/cert...

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+certificate are required to be equal among all masters. Accordingly:
+
+- `kubeadm join --master` will check for the mandatory certificates and fail fast if
+  they are missing
+- If some/all of the other certificates are provided by the user as well, `kubeadm join --master`
```
luxas: provided and valid

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+  will use them without further checks.
+- If any other certificates are missing, `kubeadm join --master` will create them.
```
luxas: given the required CA key exists

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+1. Discovery cluster info [No changes to this step]
+
+   > NB This step waits for a first instance of the kube-apiserver to become ready
+   > (the bootstrap master); And thus it acts as embedded mechanism for handling the sequence
+   > `kubeadm init` and `kubeadm join` actions in case of parallel node creation.
+
+2. In case of `join --master` [New step]
+
+   1. Using the bootstrap token as identity, read the `kubeadm-config` configMap
```
luxas: I want us to dive into whether we should use different bootstrap token roles for joining a master and a node

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+      a. In case of control plane deployed as static pods, create related kubeconfig files
+      and static pod manifests.
+
+      > see "Strategies for deploying control plane components" paragraph
+      > for additional info about this step.
+
+   4. Create the admin.conf kubeconfig file
+
+3. Executes the kubelet TLS bootstrap process [No changes to this step]:
```
luxas: Why doesn't this happen as step 2 instead of 3?
I think having this as step 2 and making everything `--master` post-node join makes a lot of sense (e.g. in case the normal node join fails)

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+         > to a dynamic workflow or to a static workflow (and to apiserver advertise address
+         > of the joining node). see following paragraphs for more details for additional info.
+
+      a. In case of control plane deployed as static pods, create related kubeconfig files
+      and static pod manifests.
+
+      > see "Strategies for deploying control plane components" paragraph
+      > for additional info about this step.
+
+   4. Create the admin.conf kubeconfig file
```
luxas: Say here that security-wise this is maybe not the greatest thing to have `n` root certificates lying around but for good UX we should do this

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+prevent to address this in subsequent phases.
+
+#### Strategies for distributing cluster certificates
+
+As of today kubeadm supports two solutions for storing cluster certificates:
+
+1. Cluster certificates stored on file system in case of:
+   - Control plane deployed as static pods (current kubeadm default)
+   - Self-hosted control plane in case of `--feature-gates=SelfHosting=true,StoreCertsInSecrets=false`
+2. Cluster certificates stored in secrets in case of:
```
luxas: We actually don't just alpha code lying around. Please remove this consideration

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+  operator is required to provide during `kubeadm init` the list of masters/the list of IP
+  addresses for all the masters as alternative names for the API servers certificate, thus
+  allowing the proper functioning of all the API server instances that will join
+
+#### Strategies for deploying control plane components
+
+As of today kubeadm supports two solutions for deploying control plane components:
+
+1. Control plane deployed as static pods (current kubeadm default)
+2. Self-hosted control plane in case of `--feature-gates=SelfHosting=true`
```
luxas: You don't need to say this here as we actually don't support this on a beta or higher level.

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+There are many ways to configure an highly available cluster.
+
+Among them, the approach best suited for a dynamic bootstrap workflow requires the
+user to set the `--apiserver-advertise-address` of each master, including the bootstrap master
+itself, equal to the `controlplaneAddress` endpoint provided during kubeadm init
+(the IP/DNS of the external load balancer).
+
+By using the same advertise address for all the IP masters, `kubeadm init` can create
+a unique API server serving certificate that could be shared across many masters nodes;
```
luxas: ??
I think the certificate should be **always** be unique per-master but always have `controlPlaneAddress` in the SANs if specified.

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+In case of a static bootstrap workflow the final layout of the controlplane - the number, the
+name and the IP of master nodes - is know in advance.
+
+Given such information, the user can choose a different approach where each master has a
+specific apiserver advertise address different from the `controlplaneAddress`.
+
+Please note that:
+
+- if the user is not planning to distribute the apiserver certificate among masters, kubeadm
```
luxas: I don't think the user ever should distribute/share the API server serving cert

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+  will generate a new apiserver serving certificate with the required SANS
+- if the user is planning to distribute the apiserver certificate among masters, he/she/the
```
luxas: can you just remove this?

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+Please note that:
+
+- if the user is not planning to distribute the apiserver serving certificate among masters,
+  kubeadm will generate a new apiserver serving certificate “almost equal” to the certificate
+  created on the bootstrap master (it differs only for the domain name of the joining master)
+
+#### Static workflow (advertise-address != `controlplaneAddress`)
+
+In case of a static bootstrap workflow the final layout of the controlplane - the number, the
```
luxas: what is technically the difference between these?

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+
+The proposed solution for case 1. "Control plane deployed as static pods", assumes
+that the `kubeadm join --master` flow will take care of creating required kubeconfig
+files and required static pod manifests.
+
+Self-hosted control plane are not supported in this release of the proposal
+(`kubeadm join --master` will fail immediately), but nothing in this proposal should
+prevent to address this in subsequent phases.
+
+#### Strategies for distributing cluster certificates
```
luxas: Please dive into how the CSR API could be utilized in an external CA-world.
It'd be cool to just use the CSR API, but we're stuck needing the front-proxy clientcert for the API server and the SA key that is not bound to the CA.

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+"Cluster certificates stored in secrets", are not supported in this release of the
+proposal (`kubeadm join --master` will fail immediately), but nothing in this proposal
+should prevent to address this in subsequent phases.
+
+#### `kubeadm upgrade` for HA clusters
+
+Nothing in this proposal prevents implementation of `kubeadm upgrade` for HA cluster.
+
+Further detail will be provided in a subsequent release of this KEP when all the detail
+of the `v1Beta1` release of kubeadm api will be available (including a proper modeling
```
luxas: nit: `v1beta1`

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+## Implementation History
+
+- original HA proposals [#1](https://goo.gl/QNtj5T) and [#2](https://goo.gl/C8V8PV)
+- merged [Kubeadm HA design doc](https://goo.gl/QpD5h8)
+- HA prototype [demo](https://goo.gl/2WLUUc) and [notes](https://goo.gl/NmTahy)
+- [PR #58261](https://github.com/kubernetes/kubernetes/pull/58261) with the showcase implementation of the first release of this KEP
+
+## Drawbacks
+
+The kubeadm join --master workflow requires that some condition are satisfied at `kubeadm-init` time,
```
luxas: kubeadm init

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,431 @@
...
+certificates from remote locations.
+
+Then, the `kubeadm join --master` flow will take care of checking certificates
+existence and conformance.
+
+"Cluster certificates stored in secrets", are not supported in this release of the
+proposal (`kubeadm join --master` will fail immediately), but nothing in this proposal
+should prevent to address this in subsequent phases.
+
+#### `kubeadm upgrade` for HA clusters
```
luxas: can you dive into exactly _how_ and in which order the upgrade should be performed.
i.e. upgrade etcd fully first (in _some_ way out of scope for this proposal probably), then remove one master from the loadbalancer, upgrade it, (add it back to the LB, or not). Upgrades without downtime supported officially or not? etc. etc.

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: You can create in parallel bootstrap master, secondary masters and workers because:
1. all the joining nodes (secondary masters and workers) use the same discovery mechanism, that "automatically" waits for the bootstrap master to complete its setup
2. the joining nodes (secondary masters and workers) can join in any order

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: I removed this paragraph that is not relevant anymore and added a specific non-goal; I confirm that graduating an existing node to master is a non-goal.

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: 100% agreed! never the less it is a challenge worthy of note here

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: yes! in the implementation is described how we can support different types of HA clusters / join workflow by making only few changes to the `kubeadm join` UX

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: In the discover phase, the joining node has limited access to the cluster (with respect to the "root" access that you have when doing `kubeadm init`)

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: Prerequisites is not part of the KEP template.. What do you foresee in this chapter? all the prerequisites for this specific proposal are already described in the Implementation Details paragraph ...

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: Nothing prevents us to this, but we should consider also that having different bootstrap token makes the UX more complex. 
IMO, considering the short TTL now assigned by default to tokens and all the existing facilities for managing token available under `kubeadm token`, we should keep this simple

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: This is not necessary in case of advertise-address == `controlplaneAddress`, but nothing prevents you to do so.

As stated in the note, kubeadm will act as you suggest, creating unique certificates per-master unless the user intentionally decide to distribute the same certificate everywhere (that is something we will not recommend in the user guide)

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: I prefer to leave it, because it documents "what happen if".
Instead let's deprecate this in the user guide

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: This was discussed in some kubeadm office hours meetings, and I wrote also a blog post with all the details https://blog.heptio.com/kubernetes-ha-under-x-ray-5d05f552c9f

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: I would prefer to keep this out of scope of this scope, because it does not apply specifically to the multi-master scenario (e.g. it is relevant also for the existing `kubeadm join` workflow).

Never the less I totally agree the we should document how the CSR API could be used in an external CA-world, but I don't have real expertise in this. Eventually, let's raise the topic at the sig-meeting

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: The only prior art I'm aware of is https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha/ , but I'm not sure there is already a consensus about how to do this, and so I prefer to get a first release of the KEP approved and the iterate on the upgrade part.

fabriziopandini:
@luxas I provided answer to your question and addressed all comments. Let me know if there is something that requires further deep dives.
PS. I'm planning to discuss briefly this KEP during kubeadm office hours on wendesday, as well

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: done

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: done

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
fabriziopandini: see to following comment

fabriziopandini:
@chuckha @detiber kindly ping

keps/sig-cluster-lifecycle/0015-kubeadm-join-master.md
```diff
@@ -0,0 +1,427 @@
...
+
+- This proposal doesn't include a solution for API server load balancing (Nothing in this proposal
+  should prevent users from choosing their preferred solution for API server load balancing).
+
+- This proposal doesn't address the ongoing discussion about kubeadm self-hosting; in light of
+  divide and conquer goal stated before, it is not planned to provide support for self-hosted clusters
+  neither in the initial proposal nor in the foreseeable future (but nothing in this proposal should
+  explicitly prevent to reconsider this in future as well).
+
+- This proposal doesn't provide an automated solution for transferring the CA key and other required
```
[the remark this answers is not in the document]
fabriziopandini: 👍

fabriziopandini:
@chuckha thanks for your comments! 
Feedbacks are addressed;, with regards to you general question, yes, you are right, the proposed solution re-uses a lot of what already exists in `kubeadm init` and  `kubeadm join`, but this is intentional.

What we are getting on top is the fact that we are going to simplify what the user has to do for getting a multi-master cluster up and running, and at the same time, we are reducing the probability to make errors along the way.  This is basically achieved by:
- delegating to kubeadm the task to ensure that _exactly_ the same kubeadm-config_configMap is used on all masters
- checking that the cluster/the kubeadm-config is properly configured for HA (kind of pre-flight checks on the "carefully creafted `config.yaml`" )
- blocking users trying to create multi masters with configurations we don't want to support as a sig (e.g. HA with self-hosted control plane)

On top of that :
- There are also things that `kubeadm join --master` doesn't do with respect to `kubeadm init` (it doesn't create a new token, it doesn't re-write / re-create all the RBAC rules, config maps, addons)  
- We can provide a better/cleaner solution for steps that should be done in a slightly different way on a secondary master with respect to the bootstrap master (e.g. updating the kubeadm-config map adding info about the new master instead of creating a new configMap from scratch *this is related to API redesing).  

Does this sound good to you?

PS. the above points are documented in the alternatives paragraphs; let me know if this should be further clarified

neolit123:
BTW, during one of the meeting last week we've agreed that `--master` should to be used, and `--control-plane` is fine instead.
should that be reflected in the KEP?

chuckha:
/lgtm

Thanks for the detailed response @fabianofranz. I think the alternatives section is good. I think you have to intimately know the code to get `kubeadm init --config config.yaml` to function correctly and `kubeadm join --master` will be a great addition.

pbarker:
@fabianofranz I believe this KEP should be  number `14` and the `NEXT_KEP_NUMBER` should be `15` looking at https://github.com/kubernetes/community/blob/master/keps/0000-kep-template.md#title

PR #2335, opened 2018-07-02
CSI snapshot design

xing-yang:
This document proposes to add Kubernetes snapshot support for CSI Volume Drivers. It proposes standardized snapshot APIs in the form of CRD, external snapshot controller to handle snapshotting operations, and a way to restore volumes from snapshots.


contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,316 @@
+Kubernetes CSI Snapshot Proposal
+================================
+
+**Authors:** [Jing Xu](https://github.com/jingxu97), [Xing Yang](https://github.com/xing-yang), [Tomas Smetana](https://github.com/tsmetana), [Huamin Chen ](https://github.com/rootfs)
+
+## Background
```
jsafrane: There should be some relation to the previous snapshot implementation. Will there be a supported path how to move from the old implementation to the new one?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+	VolumeSnapshotRef *core_v1.ObjectReference `json:"volumeSnapshotRef" protobuf:"bytes,2,opt,name=volumeSnapshotRef"`
+
+	// PersistentVolumeRef represents the PersistentVolume that the snapshot has been
+	// taken from
+	// +optional
+	PersistentVolumeRef *core_v1.ObjectReference `json:"persistentVolumeRef" protobuf:"bytes,3,opt,name=persistentVolumeRef"`
+}
+
+// Represents the actual location and type of the snapshot. Only one of its members may be specified.
+type VolumeSnapshotDataSource struct {
```
jsafrane: Since only CSI is going to be supported, this struct should probably contain only `CSISnapshotSource`:

```go
type VolumeSnapshotDataSource struct {
    CSI *CSISnapshotSource
}

type CSISnapshotSource struct {
    ID string
}
```

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+Note: SnapshotSource will be added in this phase. PVCSource will be added next.
+
+
+#### The `SnapshotClass` Object
+
+A new SnapshotClass API object will be added to avoid mixing parameters between snapshots and volumes. Each CSI Volume Driver can have its own default SnapshotClass. If SnapshotClass is not provided, a default will be used. It allows to add new parameters for snapshots.
+
+```
+
+type SnapshotClass struct {
```
jsafrane: Missing `TypeMeta` and `ObjectMeta`?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+	// This value may not be empty.
+	Provisioner string
```
jsafrane: Since only CSI is going to be supported, should it just contain `CSIDriver` instead of `Provisioner`?

In addition, `Provisioner` is already used to define provisioner that provisions PVs. Should it be `Snapshotter` or something similar?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+
+```
+
+
+### CSI External Snapshot Controller
+* External snapshotter is responsible for creating/deleting snapshots and binding snapshot and snapshotData objects.
+
+* External snapshotter is running in the sidecar along with external-attacher and external-provisioner for each CSI Volume Driver.
+
+* For dynamically created snapshot, it should have a SnapshotClass associated with it. User can explicitly specify a SnapshotClass in the VolumeSnapshot API object. If not, a default SnapshotClass will be used.
```
jsafrane: "a default SnapshotClass will be used." - how is the default SnapshotClass marked?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+
+* For statically binding snapshot, user/admin must specify the pointers in both directions (VolumeSnapshot → VolumeSnapshotData, and VolumeSnapshotData → VolumeSnapshot) so that controller will bind them.
```
jsafrane: "so that controller will bind them" - if both pointers are set, what exactly will the controller do to "bind them"? In PV-PVC case, these pointers *are* the binding and PV controller won't do anything when both pointers are set (except updating PV/PVC status to "Bound").

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: The previous snapshot implementation is not for CSI.  Is there a way to move PV/PVC from non-CSI drivers to CSI drivers?  If there's a way, then maybe we can do similar move for snapshots.  Maybe some kind of export/import?

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Sure.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: You are right.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Ok, we can call it Snapshotter.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Ok.  This will be re-worded.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: A default SnapshotClass should still be created ahead of time by the admin, similar to a default StorageClass.  We'll add some clarifications.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: I'll take a look of the migration of in-tree plugin to CSI spec.  Maybe we can do something similar here.

jsafrane: I am not advocating for full backward compatibility, I think a paragraph or two about relation of these two implementations and perhaps high-level list of (manual) steps necessary to move from the old to the new one would be enough.

xing-yang: Sure.

xing-yang: @jsafrane I added a section "Transition to the New Snapshot Support" to address your comments.  Please take a look.  Thanks.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,380 @@
...
+
+Refer to [Container Storage Interface (CSI)](https://github.com/container-storage-interface/spec) for detailed instructions on how CSI Volume Driver shall implement snapshot functions.
+
+
+## Transition to the New Snapshot Support
+
+To use the snapshot implementation in [external storage repo](https://github.com/kubernetes-incubator/external-storage/tree/master/snapshot), an external snapshot controller and external provisioner need to bedeployed.
+
+To use the new snapshot implementation for CSI, a sidecar container for the external snapshot controller needs to be deployed in addition to the sidecar container for the external provisioner.
+
```
jingxu97: Maybe put the deployment differences you mentioned at the beginning into the following section "New design" "Old design" too? 

Just say something like "The following lists the main differences between this snapshot proposal and the the old snapshot support in external storage repo"?

Or we have two part, the first part talks about the user facing differences, the second part talks about deployment differences.

xing-yang: Sure.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,380 @@
...
+}
+
+// VolumeSnapshotDataList is a list of VolumeSnapshotData objects
+type VolumeSnapshotDataList struct {
+        metav1.TypeMeta `json:",inline"`
+        Metadata        metav1.ListMeta      `json:"metadata"`
+        Items           []VolumeSnapshotData `json:"items"`
+}
+
+// The desired state of the volume snapshot
```
wackxu: s/volume snapshot/volume snapshot data/

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,380 @@
...
+Old Design
+* Does not support CSI volume drivers.
+* SnapshotClass is not needed to create a snapshot and this concept does not exist in the old design.
+* To restore a volume from the snapshot, however, user needs to create a new StorageClass that is different from the original one for the PVC.
+
+Here is an example yaml file to create a snapshot in the old design:
+
+```GO
+
+apiVersion: volumesnapshot.external-storage.k8s.io/v1
```
wackxu: maybe `volumesnapshot.csi.k8s.io/v1alpha1`

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,380 @@
...
+
+// VolumeSnapshotData represents the actual "on-disk" snapshot object
+type VolumeSnapshotData struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	Metadata metav1.ObjectMeta `json:"metadata"`
+
+	// Spec represents the desired state of the snapshot
+	// +optional
+	Spec VolumeSnapshotDataSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
```
wackxu: Do we need status for `VolumeSnapshotData `?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,380 @@
...
+#### The `SnapshotClass` Object
+
+A new SnapshotClass API object will be added to avoid mixing parameters between snapshots and volumes. Each CSI Volume Driver can have its own default SnapshotClass. If SnapshotClass is not provided, a default will be used. It allows to add new parameters for snapshots.
+
+```
+
+type SnapshotClass struct {
+        metav1.TypeMeta `json:",inline"`
+        // +optional
+        Metadata metav1.ObjectMeta `json:"metadata"`
```
wackxu:
```
metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,380 @@
...
+
+Here is an example to create a SnapshotClass and create a snapshot in the new design:
+
+```GO
+
+apiVersion: volumesnapshot.csi.k8s.io/v1alpha1
+kind: SnapshotClass
+metadata:
+  name: csi-hostpath-snapclass
+  snapshotter: csi-hostpath
```
wackxu: the format maybe wrong, remove the white space

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: This is an example of the old design from external-storage.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Ok, I'll check.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,380 @@
...
+	Conditions []VolumeSnapshotCondition `json:"conditions" protobuf:"bytes,2,rep,name=conditions"`
+}
+
+type VolumeSnapshotConditionType string
+
+// These are valid conditions of a volume snapshot.
+const (
+        // VolumeSnapshotConditionCreating means the snapshot is being created but
+        // it is not cut yet.
+        VolumeSnapshotConditionCreating VolumeSnapshotConditionType = "Creating"
```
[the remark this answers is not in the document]
xing-yang: API needs to return immediately while CSI driver call is blocking until the snapshot is cut.  That's why we need a "Creating" status for API.  It might map to UNKNOWN in CSI.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Since we are not binding in the in-tree controller any more, we don't need a status for VolumeSnapshotData any more because external-snapshotter will be handling the status creation and binding.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,381 @@
...
+
+#### The `VolumeSnapshotData` Object
+
+```GO
+
+// +genclient=true
+// +nonNamespaced=true
+
+// VolumeSnapshotData represents the actual "on-disk" snapshot object
+type VolumeSnapshotData struct {
```
humblec: As discussed in today's design meeting,` VolumeSnapshotContent` looks better than `VolumeSnapshotData`.

xing-yang: Done.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,395 @@
...
+        // +optional
+        FailedAt metav1.Time `json:"failedAt" protobuf:"bytes,1,opt,name=failedAt"`
+}
+
+type VolumeSnapshotStatus struct {
+	// VolumeSnapshotCreated indicates whether the snapshot was successfully created.
+	// If the timestamp CreateAt is set, it means the snapshot was created;
+	// Otherwise the snapshot was not created.
+        // +optional
+        Created VolumeSnapshotCreated
```
jsafrane: Since it's optional, should it be a pointer?

```go
if snap.Status.Created == nil  {
}
```

vs.

```go
if snap.Status.Created.CreatedAt.IsZero() {
}
```

tsmetana: Why do we need this to be a struct at all? The "IsZero()" test would work with the plain timestamp too and should be sufficient to figure out if the snapshot exists and since when.

xing-yang: Since we only have timestamp in the struct now, it is possible to get rid of the struct.  A struct would allow us to add more parameters to it later if needed.

jingxu97: I think that's a good suggestion. We can make it like this

```
Type VolumeSnapshotStatus struct {
    CreatedAt *metav1.Time
    AvailableAt *metav1.Time
    FailedAt *metav1.Time
    Reason string
    Message string
}
```

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+    * PersistentVolumeClaim: clone volume which is represented by PVC
+
+```
+
+type PersistentVolumeClaimSpec struct {
+        // If specified, volume will be prepopulated with data from the PVCDataSourceRef.
+        // +optional
+        PVCDataSourceRef *core_v1.LocalObjectReference `json:"dataSourceRef" protobuf:"bytes,2,opt,name=dataSourceRef"`
+
+}
```
garthy: Would a v1.ObjectReference be better? This also allows sources to be in other namespaces.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+        Name string
+	// Kind indicates the type of the object reference.
+	// +optional
+	Kind string
+}
+
+```
+
+In the first version, only VolumeSnapshot will be a supported `Type` for data source object reference. PersistentVolumeClaim will be added in a future version. If unsupported `Type` is used, the PV Controller SHALL bail out of the operation.
+
```
garthy: bail out -> fail

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+        // +optional
+	Parameters map[string]string
+}
+
+```
+
+#### Add `DataSource` to `StorageClass`
+
+DataSource will be added as an optional parameter in `StorageClass` to allow volumes belonging to the same `StorageClass` to be created with prepopulated data.
+
```
garthy: This leans towards making the LocalObjectReference an ObjectReference as it supports shared sources in a public namespace usable my other in multiple other namespaces.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+* CreateSnapshot, DeleteSnapshot, and create volume from snapshot if it supports CREATE_DELETE_SNAPSHOT.
+* ListSnapshots if it supports LIST_SNAPSHOTS.
+
+ListSnapshots can be an expensive operation because it will try to list all snapshots on the storage system. For a storage system that takes nightly periodic snapshots, the total number of snapshots on the system can be huge. Kubernetes should try to avoid this call if possible. Instead, calling ListSnapshots with a specific snapshot_id as filtering to query the status of the snapshot will be more desirable and efficient.
+
+CreateSnapshot is a synchronous function and it must be blocking until the snapshot is cut. For cloud providers that support the uploading of a snapshot as part of the creating snapshot operation, CreateSnapshot function must also be blocking until the snapshot is cut and after that it shall return an operation pending gRPC error code until the uploading process is complete.
+
```
garthy: You mention that is is a "synchronous function" but then state that operation pending will be returned until the upload is complete. 
Does this mean that while a snapshot is in progress on a volume you can't take another snapshot until the current one has finished? (In some cases uploadin to cloud?)
This seems overly restrictive.

xing-yang: "Synchronous" refers to the CSI driver interface part. Actually all the volume operations such as create snapshot, create volume are synchronous in CSI.  You can still take another snapshot while the current one is in progress.  If you call create snapshot with the same input parameters while the current one is in progress (pending), the call will be pending as well because it is idempotent.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: The intention is to only allow sources in the same namespace to be used.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: You are right.  We'll need to discuss about this one.  May be we need to add some limitations here.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Sure.

contributors/design-proposals/storage/csi-snapshot.md
garthy: Why is this restriction in place? Is it a technical reason or something else?

garthy:
On thing that is not covered in this is how snapshots interact with the PV/PVC life cycle. 
Specifically https://kubernetes.io/docs/tasks/administer-cluster/storage-object-in-use-protection/.

Is it valid to delete a PV/PVC that a snapshot references?  should we extend the `kubernetes.io/pvc-protection` finaliser or add our own finaliser `kubernetes.io/pvc-snapshot-protection`

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: It is for technical reason.  VolumeSnapshot itself is name-spaced.

garthy:
OK So having read this one thing that makes me a bit uneasy is the reference to PVC as the storage from the snapshot. In my mind it should match the PVC/PV split.

* PV (PersistentVolume) - a piece of storage in the cluster
* PVC (PersistentVolumeClaim) - a request for storage by a user

So the snapshot API should Mirror this.

* SV (SnapshotVolume) -  a snapshot of a volume in the cluster
* SVC (SnapshotVolumeClaim) - a request for a snapshot by a user

The SnapVolume would then have a SnapshotVolumeSource like the PersistentVolumeSource that is a list of pointers to supported snapshot sources (In this case CSI)

The SVC lifecycle would then match closly to the PVC lifecycle.

A user would create a SVC and this claim for a snapshot of a volume would be satisfied (or not).

The SV would reference a PV not a PVC.

This matches the PVC/PV API much more closely.

xing-yang:
@garthy We discussed about whether to follow PV/PVC naming convention as well.  Please see comments here: https://github.com/kubernetes-csi/external-snapshotter/pull/6#pullrequestreview-138422466

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+The snapshotter talks to out-of-tree CSI Volume Driver over socket (`/run/csi/socket` by default, configurable by `-csi-address`). The snapshotter tries to connect for `-connection-timeout` (1 minute by default), allowing CSI Volume Driver to start and create its server socket a bit later.
+
+The snapshotter creates CreateSnapshotRequest and calls CreateSnapshot through CSI ControllerClient. It gets CreateSnapshotResponse from the out-of-tree CSI Volume Driver and creates a VolumeSnapshotContent API object with VolumeSnapshotSource.
+
+The snapshotter checks status in VolumeSnapshot to decide whether to bind. It binds VolumeSnapshot and VolumeSnapshotContent when ready.
+
+When the storage system fails to create snapshot, retry will not be performed in the first version. This is because users may not want to retry when taking consistent snapshots or scheduled snapshots when the timing of the snapshot creation is important. In a future version, a maxRetries flag will be added to allow users to control whether retries are needed.
```
saad-ali: @garthy had a good recommendation on the call: should have retry termination timestamp instead of number of retries. Which will work better with workload "quiecing".

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+        // both GCE and AWS support uploading of the snapshot after it is cut as part of
+        // the Create Snapshot process.
+        // If the timestamp AvailableAt is set, it means the snapshot was available;
+        // Otherwise the snapshot was not available.
+        // +optional
+        AvailableAt *metav1.Time `json:"availableAt" protobuf:"bytes,2,opt,name=availableAt"`
+
+        // The time the error occurred during the snapshot creation (or uploading) process
+        // +optional
+        FailedAt *metav1.Time `json:"failedAt" protobuf:"bytes,3,opt,name=failedAt"`
```
saad-ali: Let's resuse `VolumeError struct` from https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/storage/types.go
instead of three fields (FailedAt, Reason, Message)

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+        // it is possible that the snapshot is being uploaded to the cloud. For example,
+        // both GCE and AWS support uploading of the snapshot after it is cut as part of
+        // the Create Snapshot process.
+        // If the timestamp AvailableAt is set, it means the snapshot was available;
+        // Otherwise the snapshot was not available.
+        // +optional
+        AvailableAt *metav1.Time `json:"availableAt" protobuf:"bytes,2,opt,name=availableAt"`
+
+        // The time the error occurred during the snapshot creation (or uploading) process
+        // +optional
```
saad-ali: What happens if creation fails and `FailedAt` is set, but creation succeeds on retry, and then fails on `AvailableAt`

Let's document what gets set when, and what it means.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,377 @@
...
+```
+
+#### Add `DataSource` to `StorageClass`
+
+DataSource will be added as an optional parameter in `StorageClass` to allow volumes belonging to the same `StorageClass` to be created with prepopulated data.
+
+```
+
+type StorageClass struct {
+        // If specified, volume will be prepouplated with data from the DataSource
```
saad-ali: Adam on the call had a good suggestion of adding DataSource to PV as well as PVC because if an external-provisioner doesn't understand the new DataSource field and doesn't pre-populate the new volume, PV/PVC provisioner should be able to detect that (e.g. PVC has data source but PV does not)

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Done.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Done.

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Done

contributors/design-proposals/storage/csi-snapshot.md
xing-yang: Done

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+
+Possible `DataSource` types may include the following:
+
+    * VolumeSnapshot: restore snapshot to a new volume
+    * PersistentVolumeClaim: clone volume which is represented by PVC
+
+```
+
+type PersistentVolumeClaimSpec struct {
+        // If specified, volume will be prepopulated with data from the PVCDataSourceRef.
```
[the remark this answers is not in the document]
thockin: comment says "PVCDataSourceRef" but field is "DataSourceRef"

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+        // +optional
+        DataSourceRef *TypedLocalObjectReference `json:"dataSourceRef" protobuf:"bytes,2,opt,name=dataSourceRef"`
+}
+
+type PersistentVolumeSpec struct {
```
thockin: Do we really need this?  Is it not sufficient for it to be in the PVC?

if so, same comments as above

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+
+Add a TypedLocalObjectReference in core API.
+
+```
+
+// TypedLocalObjectReference contains enough information to let you locate the referenced object inside the same namespace.
+type TypedLocalObjectReference struct {
+        // Name of the object reference.
+        Name string
+	// Kind indicates the type of the object reference.
```
thockin: indents

contributors/design-proposals/storage/csi-snapshot.md
thockin: What if the PVC and PV have different sources?

thockin: Is this going to be used as part of the PV<->PVC match logic?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+#### Add `DataSource` to `StorageClass`
+
+DataSource will be added as an optional parameter in `StorageClass` to allow volumes belonging to the same `StorageClass` to be created with prepopulated data.
+
+```
+
+type StorageClass struct {
+        // If specified, volume will be prepouplated with data from the DataSource
+        // +optional
+        DataSource TypedLocalObjectReference
```
thockin: Do we really need this? How common is it?

What if this and the PVC both specify source, and they disagree?

thockin: For all of these source fields - what happens if I point them at a nonsensical kind, like "Service" or at a kind that doesn't exist?  What if I point them at a valid Kind but the named instance doesn't exist.  they ALL need to document this.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+type VolumeSnapshotList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	Items           []VolumeSnapshot `json:"items"`
+}
+
+// VolumeSnapshotSpec is the desired state of the volume snapshot
+type VolumeSnapshotSpec struct {
+        // PersistentVolumeClaimName is the name of the PVC being snapshotted
```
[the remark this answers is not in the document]
jingxu97: volumesnapshot creations supports two types: dynamically creation of volumesnapshot and statically binding of snapshot and snapshotContent. In case of dynamically creation, PVCName must be provided. For statically binding, PVC can be nil, and user has to create VolumeSnapshotContent and bind it with VolumeSnapshot

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+	// it is created, so the snapshot will become available some time after snapshot is created.
+	// For other plugins, the snapshot might be available at the same time when the snapshot 
+	// is created.
+        // +optional
+        AvailableAt *metav1.Time `json:"availableAt" protobuf:"bytes,2,opt,name=availableAt"`
+	
+	// Setting Bound to true means that the snapshot is created (and uploaded for cloud providers) sucuessfully, and 
+	// VolumeSnapshotContent is created and bound with the VolumeSnapshot. After Bound becomes true, users
+	// can start to use VolumeSnapshot to provision new volumes.
+	Bound bool `json:"bound" protobuf:"varint,4,opt,name=bound"`
```
garthy: Can we rename this to ready? Arguably the snapshot is bound when it is created it is just not read to be use.

bound to me implies that something is associated which is true once the snapshot is created. ready seem more appropriate as the volume is ready to use.

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+
+### Snapshot API Design
+
+The API design of VolumeSnapshot and VolumeSnapshotContent is modeled after PersistentVolumeClaim and PersistentVolume. In the first version, the VolumeSnapshot lifecycle is completely independent of its volumes source (PVC). When PVC/PV is deleted, the corresponding VolumeSnapshot and VolumeSnapshotContents objects will continue to exist.
+
+#### The `VolumeSnapshot` Object
+
+```GO
+
+// The volume snapshot object accessible to the user. Upon successful creation of the actual
```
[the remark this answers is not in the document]
jingxu97: Now changed it to

// VolumeSnapshot is a user's request for taking a snapshot. Upon successful creation of the actual
// snapshot by the volume provider it is bound to the corresponding VolumeSnapshotContent. 
// Only the VolumeSnapshot object is accessible to the users in their namespaces. 

After we add snapshot document files, we can also have something like this

//More info: https://kubernetes.io/docs/concepts/storage/volumesnapshot

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+type VolumeSnapshotContentList struct {
+        metav1.TypeMeta `json:",inline"`
+        Metadata        metav1.ListMeta      `json:"metadata"`
+        Items           []VolumeSnapshotContent `json:"items"`
+}
+
+// The desired state of the volume snapshot data
+type VolumeSnapshotContentSpec struct {
+	// Source represents the location and type of the volume snapshot
+	VolumeSnapshotSource `json:",inline" protobuf:"bytes,1,opt,name=volumeSnapshotSource"`
```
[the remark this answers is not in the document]
jingxu97: what is a Source sub-struct?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+
+	// PersistentVolumeRef represents the PersistentVolume that the snapshot has been
+	// taken from
+	// +optional
+	PersistentVolumeRef *core_v1.ObjectReference `json:"persistentVolumeRef" protobuf:"bytes,3,opt,name=persistentVolumeRef"`
+}
+
+// Represents the actual location and type of the snapshot.
+// Only CSI volume snapshot source is supported now.
+type VolumeSnapshotSource struct {
```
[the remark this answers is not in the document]
jingxu97: could you give some more details about this? Thanks!

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+        // plugin’s CreateSnapshot to refer to the snapshot on all subsequent calls.
+        // Required.
+        SnapshotHandle string `json:"snapshotHandle"`
+
+        // Timestamp when the point-in-time snapshot is taken on the storage
+        // system. The format of this field should be a Unix nanoseconds time
+        // encoded as an int64. On Unix, the command `date +%s%N` returns
+        // the  current time in nanoseconds since 1970-01-01 00:00:00 UTC.
+        // This field is REQUIRED.
+	CreatedAt int64 `json:"createdAt,omitempty" protobuf:"varint,3,opt,name=createdAt"`
```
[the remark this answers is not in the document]
jingxu97: fixed

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+#### The `VolumeSnapshotClass` Object
+
+A new VolumeSnapshotClass API object will be added instead of resuing the existing StorageClass, in order to avoid mixing parameters between snapshots and volumes. Each CSI Volume Driver can have its own default VolumeSnapshotClass. If VolumeSnapshotClass is not provided, a default will be used. It allows to add new parameters for snapshots.
+
+```
+
+type VolumeSnapshotClass struct {
+        metav1.TypeMeta `json:",inline"`
+        // +optional
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```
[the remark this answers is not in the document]
jingxu97: fixed

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+
+	// Snapshotter is the driver expected to handle this VolumeSnapshotClass.
+	// This value may not be empty.
```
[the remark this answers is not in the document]
jingxu97: it should be the same as StorageClass provisioner name. We can make it as optional. 
If user does not specify it, snapshot controller can find the StorageClass through the PVC specified in snapshotSpec and then use the provisioner from the StorageClass. 
If user specify it, the snapshotter name should match the provisioner name from StorageClass.
Even though this name is not very necessary considering the snapshot creation case, but could be userful for copying snapshot operations which might be added in the future

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+	Snapshotter string
+
+	// Parameters holds parameters for the snapshotter.
```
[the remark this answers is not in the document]
jingxu97: fixed

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+        // These values are opaque to the system and are passed directly
+        // to the snapshotter. The only validation done on keys is that they are
+        // not empty. The maximum number of parameters is
+        // 512, with a cumulative max size of 256K
+        // +optional
+	Parameters map[string]string
```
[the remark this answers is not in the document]
jingxu97: fixed

contributors/design-proposals/storage/csi-snapshot.md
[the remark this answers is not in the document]
jingxu97: fixed

contributors/design-proposals/storage/csi-snapshot.md
jingxu97: fixed

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,364 @@
...
+        SnapshotContentName string `json:"snapshotContentName" protobuf:"bytes,2,opt,name=snapshotContentName"`
+
+        // Name of the VolumeSnapshotClass required by the volume snapshot.
+        // +optional
+        VolumeSnapshotClassName string `json:"storageClassName" protobuf:"bytes,3,opt,name=storageClassName"`
+}
+
+type VolumeSnapshotStatus struct {
+        // CreatedAt is the time when the snapshot was successfully created. When it is set,
+	// users can start to resume/unfreeze the application if it was previously frozen
```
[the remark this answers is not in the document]
jingxu97: fixed

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,388 @@
...
+
+```
+
+type PersistentVolumeClaimSpec struct {
+        // If specified when creating, volume will be prepopulated with data from the DataSource.
+        // +optional
+        DataSource *TypedLocalObjectReference `json:"dataSource" protobuf:"bytes,2,opt,name=dataSource"`
+}
+
+type PersistentVolumeSpec struct {
```
thockin: I thought we agreed we didn't need this any more?

jingxu97: yes, just removed it. Thank you!

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,376 @@
...
+
+// VolumeSnapshotSpec describes the common attributes of a volume snapshot
+type VolumeSnapshotSpec struct {
+	// Source has the information about where the snapshot is created from.
+	// In Alpha version, only PersistentVolumeClaim is supported as the source.
+	// If not specified, user can create VolumeSnapshotContent and bind it with VolumeSnapshot manually.
+	// +optional
+	Source *TypedLocalObjectReference `json:"source" protobuf:"bytes,1,opt,name=source"`
+
+	// SnapshotContentName binds the VolumeSnapshot object with the VolumeSnapshotContent
```
thockin: We will need detailed docs for what happens in all the various corner-cases.  e.g. what if a user specifies this field?

contributors/design-proposals/storage/csi-snapshot.md
```diff
@@ -0,0 +1,376 @@
...
+Add a new `DataSource` field into PVC to represent the source of the data which is populated to the provisioned volume. External-provisioner will check `DataSource` field and try to provision volume from the sources. In the first version, only VolumeSnapshot is the supported `Type` for data source object reference. Other types will be added in a future version. If unsupported `Type` is used, the PV Controller SHALL fail the operation. Please see more details in [here](https://github.com/kubernetes/community/pull/2495)
+
+Possible `DataSource` types may include the following:
+
+    * VolumeSnapshot: restore snapshot to a new volume
+    * PersistentVolumeClaim: clone volume which is represented by PVC
+
+```
+type PersistentVolumeClaimSpec struct {
+        // If specified when creating, volume will be prepopulated with data from the DataSource.
```
thockin: In code this has to be more detailed.  What types are supported, what happens if the type specified is not supported?  Who decides what is supported...

Issue #484, opened 2017-03-28
New release notes process and guidance


bgrant0607:
Forked from https://github.com/kubernetes/features/pull/204 and other discussions.

We put our release notes here:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md

The original release-note mechanism design and release changelog template were described here:
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release-notes.md

A decision about whether a PR should generate a release note was made merge-blocking in the PR workflow, as described here:
https://github.com/kubernetes/community/blob/master/contributors/devel/pull-requests.md#release-notes

We subsequently made it possible to both set the labels automatically from the PR template, if properly filled out, and created /release-note commands to enable non-maintainers to set the labels, as described here:

https://github.com/kubernetes/community/blob/master/contributors/devel/pull-request-commands.md#set-release-notes

We also subsequently developed the features repo and process, from which some of the release notes are created by hand, based on which features target the release.

It's clear that the release-note mechanism in the kubernetes repo is causing too much contributor friction, that most people don't know what deserves a release note (e.g., examples, tests, and build scripts shouldn't have release notes) or what would make a good release note, and that the features repo provides another potential mechanism for surfacing release notes.

** What is the purpose of release notes **

Release notes give our users a heads-up about what is coming when they install or upgrade to a particular release of Kubernetes.

They are not primarily for contributors to Kubernetes, who could look at the PRs associated with a particular git tag.

** What deserves a release note **

Any significant user- or admin-visible change. It could be an API change, a CLI change, a UI change, a configuration schema change, a behavioral change, a change in non-functional attributes such as efficiency or availability, availability of a new platform, advance warning about deprecation of a feature, etc.

Fixes of issues previously cited as Known Issues due to the breadth and/or severity of their impact also qualify, as do CVE fixes.

** What doesn't deserve a release note **

Not exhaustive:
* tests (e.g., `Fix TestServiceAlloc flakes`)
* examples
* build infrastructure
* fixes of unreleased bugs
* fixes of minor bugs
* ideally, incomplete/inaccessible features

** What should be in a release note **

* Clear, concise description of the change
  * Explicitly name affected API or configuration fields, CLI commands or flags, relevant feature, measured non-functional changes, etc., in user-visible terms (e.g., using json field names rather than Go field names)
  * State whether it was added, removed, changed, deprecated, advanced in maturity (alpha, beta, stable), etc.
[...]

PR #2339, opened 2018-07-03
Update the release notes guide to reference recently added labels

marpaia:
This PR updates the copy of the Release Notes Guide to include information about using the `release-notes-action-required` and `release-note-none` label.

For context, we were talking with @jberkus in the 1.11 release team retro about how there is insufficient documented guidance on policies around release notes.  This all also seems relevant to #484. Perhaps this guide is where we can expand and document the expectations and policies for release notes?


contributors/guide/release-notes.md
```diff
@@ -16,20 +16,14 @@ For pull requests with a release note:
     Your release note here
     ```
 
-For pull requests that require additional action from users switching to the new release, include the string "action required" (case insensitive) in the release note:
+For pull requests that require additional action from users switching to the new release, write your Release Note as usual and add the `release-note-action-required` label to the PR.
```
nikhita: If the words "action required" are not mentioned in the release note, the bot will automatically add the `release-note` label.

If the words "action required" are present in the release note, the bot will automatically add the `release-note-action-required` label.

I think writing the words "action required" are still relevant. If we add the release note as usual and then add then use the command `/release-note-action-required`, it will have two release note labels: `release-note-none` and `release-note-action-required`.

nikhita: Oh, I just checked: https://prow.k8s.io/command-help. There is no `/release-note-action-required` command. Looks the words "action required" are really needed.

contributors/guide/release-notes.md
```diff
@@ -16,20 +16,14 @@ For pull requests with a release note:
...
+To see how to format your release notes, view the kubernetes/kubernetes [pull request template](https://git.k8s.io/kubernetes/.github/PULL_REQUEST_TEMPLATE.md) for a brief example. Pull Request titles and body comments can be modified at any time prior to the release to make them friendly for release notes.
 
-Release notes apply to pull requests on the master branch. For cherry-pick pull requests, see the [cherry-pick instructions](contributors/devel/cherry-picks.md). The only exception to these rules is when a pull request is not a cherry-pick and is targeted directly to the non-master branch.  In this case, a `release-note-*` label is required for that non-master pull request.
\ No newline at end of file
+Release notes apply to pull requests on the master branch. For cherry-pick pull requests, see the [cherry-pick instructions](contributors/devel/cherry-picks.md). The only exception to these rules is when a pull request is not a cherry-pick and is targeted directly to the non-master branch.  In this case, a `release-note-*` label is required for that non-master pull request.
```
nikhita: nit: modify the relative link to `/contributors/devel/cherry-picks.md`?

nikhita: > In this case, a `release-note-*` label is required for that non-master pull request.

This would mean the non-cherry-pick PR to a non-master branch can have `release-note-none`? I think it makes sense to enforce that it should have a release note. Wdyt?

tpepper: I wonder...are all post-release patches on a release branch implicitly not 'release-note-none'?  If so, it would be good to not burden the release patch manager with manually enforcing and instead have automation that makes it clear to the cherry pick proposer that a note is required.

tpepper: /cc @feiskyer @foxish @MaciekPytel @mbohlool 

I'm curious your thoughts on this.

tpepper:
I'm still curious if @feiskyer @foxish @MaciekPytel @mbohlool have thoughts, but I suppose we should leave off on trying improve the cherry pick process as a part of this.  That's it own large task.

So that leaves the prow and labels side.  Rather than just tweaking the docs, we really should make this an improvement to /release-note.  @marpaia do you have automation in place today for the 1.12 release notes that depends on this correct labelling?  Or is this general cleanup longer term?  If the latter I'd like to see this include prow and a simple release-notes labels state machine flow that's obvious to users, implemented in prow commands, and then yields structure that enables some automated scanning for release notes.

nikhita:
>  Perhaps prow adds it automatically if the action required text is found in the note. This would make the existing instructions correct on that point.

Yup, this is what happens.

> but I suppose we should leave off on trying improve the cherry pick process as a part of this. That's it own large task.

Agreed.

@marpaia Thanks for cleaning this up! Lgtm once commits are squashed. :+1:

marpaia:
@nikhita awesome! I've squashed all the commits, let me know if there's anything else I can do.

contributors/guide/release-notes.md
```diff
@@ -22,14 +22,12 @@ For pull requests that require additional action from users switching to the new
     action required: your release note here
     ```
 
-For pull requests that don't need to be mentioned at release time, just write "NONE" (case insensitive):
+For pull requests that don't need to be mentioned at release time, use the `/release-note-none` Prow command to add the `release-note-none` label to the PR. You can also write the string "NONE" as a release note in your PR description:
```
nikhita: oops, looks like this shows only the first commit?

nikhita: My bad, I don't know why I read that as /release-note-action-required. :woman_facepalming:

nikhita:
/lgtm

/assign @castrojo  
for approval

/hold
To make sure that @tpepper gets a chance to say if he has any further comments in mind

tpepper:
@castrojo was there something for a hiccup on your approval?  This looks like it got stuck a month ago?

PR #2350, opened 2018-07-06
Document conditions vs fields in API conventions

thockin:


contributors/devel/api-conventions.md
```diff
@@ -306,34 +306,49 @@ response reduces the complexity of these clients.
...
 
 **Conditions** represent the latest available observations of an object's
-current state. Objects may report multiple conditions, and new types of
-conditions may be added in the future. Therefore, conditions are represented
-using a list/slice, where all have similar structure.
+state.  They are an extension mechanism intended to be used when the details of
+an observation are not a priori known or would not apply to all instances of a
+given Kind.  For observations that are well known and apply to all instances, a
+regular field is preferred.  An example of a Condition that probably should
+have been regular field is Pod's "Ready" condition - it is managed by core
```
liggitt: have been *a* regular field

contributors/devel/api-conventions.md
```diff
@@ -306,34 +306,49 @@ response reduces the complexity of these clients.
...
   // +optional
-  LastHeartbeatTime  unversioned.Time  `json:"lastHeartbeatTime,omitempty" description:"last time we got an update on a given condition"`
+  Reason             *string            `json:"reason,omitempty" description:"one-word CamelCase reason for the condition's last transition"`
   // +optional
-  LastTransitionTime unversioned.Time  `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
+  Message            *string            `json:"message,omitempty" description:"human-readable message indicating details about last transition"`
+
   // +optional
-  Reason             string            `json:"reason,omitempty" description:"one-word CamelCase reason for the condition's last transition"`
+  LastHeartbeatTime  *unversioned.Time  `json:"lastHeartbeatTime,omitempty" description:"last time we got an update on a given condition"`
```
liggitt: do we want to note issues with write load in steady state with heartbeat timestamps?

contributors/devel/api-conventions.md
```diff
@@ -306,34 +306,49 @@ response reduces the complexity of these clients.
...
 **Conditions** represent the latest available observations of an object's
-current state. Objects may report multiple conditions, and new types of
-conditions may be added in the future. Therefore, conditions are represented
-using a list/slice, where all have similar structure.
+state.  They are an extension mechanism intended to be used when the details of
+an observation are not a priori known or would not apply to all instances of a
+given Kind.  For observations that are well known and apply to all instances, a
+regular field is preferred.  An example of a Condition that probably should
+have been regular field is Pod's "Ready" condition - it is managed by core
+controllers, it is well understood and applies to all Pods.
```
liggitt: probably worth noting that it is difficult/impossible to change the factors that combine to determine a single field value. For example, pod readiness could not decide in the future to also include other factors like service endpoint propagation, etc, without potentially breaking API clients that depended on a particular meaning of pod readiness.

contributors/devel/api-conventions.md
thockin: will fix

contributors/devel/api-conventions.md
thockin: will add a note.

contributors/devel/api-conventions.md
thockin: Interesting choice, since we are, in fact, changing the factors that combine to define pod readiness :)

I'll take a stab at words.

thockin:
re-pushed

liggitt:
lgtm

contributors/devel/api-conventions.md
```diff
@@ -306,34 +306,57 @@ response reduces the complexity of these clients.
 ##### Typical status properties
 
 **Conditions** represent the latest available observations of an object's
-current state. Objects may report multiple conditions, and new types of
-conditions may be added in the future. Therefore, conditions are represented
-using a list/slice, where all have similar structure.
+state.  They are an extension mechanism intended to be used when the details of
+an observation are not a priori known or would not apply to all instances of a
```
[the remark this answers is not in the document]
thockin: I list "ready" as an example a couple lines later.  Not sufficient?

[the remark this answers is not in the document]
krmayankk: thanks @thockin we were basically writing a operator and struggling to find some guidelines on when to use conditions in it when i stumbled upon this.  The `ready` example below is mainly an example of what should not be a condition since it applies to all instances of the Pod. I am mainly trying to decipher this line `when the details of an observation are not a priori known` for an example of a condition/observation to which that applies.   Thanks for updating the guidelines.

thockin:
Examples could be anything that we don't know when designing a type.  An
example of the opposite of "Ready" might be a security scanner - it could
add a "Vulnerable" condition, but we're not defining a field `vulnerable`
for it because we obviously don't have cause for that.

krmayankk:
thanks @thockin would that mean, that we should add fields in status when we know that fields represents a condition that applies to all instances of that object AND has a well defined/know cause ? I am just clarifying my own thinking. Feel free to include that in the doc if you think that will help

thockin:
Yes, exactly.  A condition like "ready" should have just been a field.

thockin:
This is still not merged, no reason why not.

/approve

contributors/devel/api-conventions.md
```diff
@@ -306,34 +306,57 @@ response reduces the complexity of these clients.
...
+and forwards-compatibility concerns of any other part of the API.
 
 Condition status values may be `True`, `False`, or `Unknown`. The absence of a
-condition should be interpreted the same as `Unknown`.
+condition should be interpreted the same as `Unknown`.  How controllers handle
+`Unknown` depends on the Condition in question.
+
+Condition types should indicate state in the "abnormal-true" polarity.  For
+example, if the condition indicates when a policy is invalid, the "is valid"
+case is probably the norm, so the condition should be called "Invalid".
```
dmage: The next paragraphs say about the "Ready" and "Succeeded" conditions. Should these examples be updated? It's a bit confusing to have counterexamples in the same document.

thockin:
@dmage thanks for this.  I saved it until I had a minute..

contributors/devel/api-conventions.md
thockin: https://github.com/kubernetes/community/pull/3049

contributors/devel/api-conventions.md
```diff
@@ -306,34 +306,57 @@ response reduces the complexity of these clients.
...
+  LastTransitionTime *unversioned.Time  `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
 ```
 
 Additional fields may be added in the future.
 
+Do not use fields that you don't need - simpler is better.
+
+Use of the `Reason` field is encouraged.
+
+Use the `LastHeartbeatTime` with great caution - frequent changes to this field
```
bgrant0607: `LastProbeTime` is also in use.

The general version of this would be `LastUpdateTime`. 

Given that it's discouraged, isn't urgent to clarify.

contributors/devel/api-conventions.md
smarterclayton: I don't know how I read this and agreed with it.  Is there an earlier discussion?  For abnormal conditions, sure.  Available = true is not abnormal, and it is the behavior we want (wait for this condition to be true) when new conditions are added (a new client waiting for an old controller to report a condition it shouldn't know about has to wait for both presence and value).

Forcing humans to negate positives is bad UX.

Issue #2351, opened 2018-07-06
No pod.yaml in kubernetes/test/fixtures/doc-yaml/user-guide/ folder


jiatongw:
In the document ```community/contributors/devel/running-locally.md```  line 118 "**Running a user defined pod**" title, (please check [https://github.com/kubernetes/community/blob/master/contributors/devel/running-locally.md](https://github.com/kubernetes/community/blob/master/contributors/devel/running-locally.md)), it guides users to try ```cluster/kubectl.sh create -f test/fixtures/doc-yaml/user-guide/pod.yaml``` as an example. But there is no ```pod.yaml``` in this folder. Please check : [https://github.com/kubernetes/kubernetes/tree/master/test/fixtures/doc-yaml/user-guide](https://github.com/kubernetes/kubernetes/tree/master/test/fixtures/doc-yaml/user-guide) 
Instead, we have ```deployment.yaml```,```job.yaml```, ```multi-pod.yaml``` and ```replication.yaml```. 

I think it's better to have a ```pod.yaml``` here in order not to mislead users. Or simply we can just change the `.md` file - replace ```pod.yaml``` to ```deployment.yaml``` or others.

I have created a PR for that. Thanks.

/kind bug

PR #2352, opened 2018-07-06
Fix a mismatch in contributors/devel/running-locally.md

jiatongw:
What this **PR dose/why we need it**

This PR is related with issue #2351 
I just simply changed the` running-locally.md` file. If needed, I will also add a `pod.yaml` file

### DIFF
--- a/contributors/devel/running-locally.md
+++ b/contributors/devel/running-locally.md
@@ -121,10 +121,10 @@ Note the difference between a [container](https://kubernetes.io/docs/user-guide/
 and a [pod](https://kubernetes.io/docs/user-guide/pods/). Since you only asked for the former, Kubernetes will create a wrapper pod for you.
 However you cannot view the nginx start page on localhost. To verify that nginx is running you need to run `curl` within the docker container (try `docker exec`).
 
-You can control the specifications of a pod via a user defined manifest, and reach nginx through your browser on the port specified therein:
+You can control the specifications of a pod via a user defined manifest, and reach nginx through your browser on the port specified there in:
 
 ```sh
-cluster/kubectl.sh create -f test/fixtures/doc-yaml/user-guide/pod.yaml
+cluster/kubectl.sh create -f test/fixtures/doc-yaml/user-guide/multi-pod.yaml
 ```
 
 Congratulations!


contributors/devel/running-locally.md
```diff
@@ -121,10 +121,10 @@ Note the difference between a [container](https://kubernetes.io/docs/user-guide/
 and a [pod](https://kubernetes.io/docs/user-guide/pods/). Since you only asked for the former, Kubernetes will create a wrapper pod for you.
 However you cannot view the nginx start page on localhost. To verify that nginx is running you need to run `curl` within the docker container (try `docker exec`).
 
-You can control the specifications of a pod via a user defined manifest, and reach nginx through your browser on the port specified therein:
+You can control the specifications of a pod via a user defined manifest, and reach nginx through your browser on the port specified there in:
```
nikhita: I think therein is one word: https://www.merriam-webster.com/dictionary/therein :)

contributors/devel/running-locally.md
```diff
@@ -121,10 +121,10 @@ Note the difference between a [container](https://kubernetes.io/docs/user-guide/
...
 
 ```sh
-cluster/kubectl.sh create -f test/fixtures/doc-yaml/user-guide/pod.yaml
+cluster/kubectl.sh create -f test/fixtures/doc-yaml/user-guide/multi-pod.yaml
```
nikhita: `git log -- test/fixtures/doc-yaml/user-guide/pod.yaml` shows that this was removed in https://github.com/kubernetes/kubernetes/commit/2ae916a0ba0363c0381699c2c32b686180fbda08 (https://github.com/kubernetes/kubernetes/pull/54342) as it was thought as not used anywhere.

IMO we should it add it back again instead of changing this to `multi-pod`.yaml. Created https://github.com/kubernetes/kubernetes/pull/65962 to add it back. Ptal.

contributors/devel/running-locally.md
nikhita: Also, this line won't hold true if we change it to `multi-pod.yaml`.

contributors/devel/running-locally.md
jiatongw: Thanks for your response, Nikhita!  I tried ```deployment.yaml``` and ```multi-pod.yaml```, they both worked well.  And the guide is to use a pod as an example so I thought maybe the easiest way was to just change the guideline. But yes, I agree that it's better to add the ```pod.yaml``` back :)

jiatongw: Closed this PR.

PR #2353, opened 2018-07-07
fix a typo in contributors/devel/controllers.md

Huang-Wei:
There is a missing `:` in sample code.


Huang-Wei:
CLA signed.

/assign @idvoretskyi

PR #2354, opened 2018-07-07
Add componentconfig API types to staging KEP

luxas:
Sorry in beforehand for all the team mentions, but this touches many SIGs:

@kubernetes/sig-cluster-lifecycle-pr-reviews @kubernetes/sig-api-machinery-pr-reviews @kubernetes/sig-architecture-api-reviews @kubernetes/sig-node-pr-reviews @kubernetes/sig-cloud-provider-pr-reviews @kubernetes/sig-scheduling-pr-reviews @kubernetes/sig-network-pr-reviews 

This idea was prototyped in [this Google Doc](https://docs.google.com/document/d/1-u2y03ufX7FzBDWv9dVI_HyiIQZL8iz3o3vabeOpP5Y/edit#) and has been reviewed over the course of the last week. Myself and @sttts think that now we have enough agreement and encouragement to move on with the proposal process and turn this into a KEP for more "formal" discussions.

Unfortunately, I have to do my military service starting from next week, which means I'll be unavailable for a fast response to review comments. I might have internet connection every now and then, but I don't know when I'll be able to answer this. Luckily @sttts will chime in here and help out by pushing agreed changes to this branch to make it mergeable. (This was semi-automatically converted from gdocs, there are probably a couple of formatting errors here still)

/assign @thockin @jbeda @liggitt @bgrant0607 @mtaufen @wojtek-t 
/cc @stewart-yu @dixudx @timothysc


keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+
+* Graduate the API versions
+    * For v1.12, we’re working incrementally and will keep the versions of the existing ComponentConfigs.
+* Change the components to support reading a config file, do flag precedence correctly or add e2e testing
+    * Further, the "load-versioned-config-from-flag" feature in this proposal *should not* be confused with the
+      [Dynamic Kubelet Configuration](https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/) feature. There is nothing in this proposal advocating for that
+      a component should support a similar feature. This is all about the making the one-off “read bytes from a source and unmarshal into internal config state” possible for
+      both the internal components and external consumers of these APIs
+    * This is work to be done after this proposal is implemented (for every component but the kubelet which has this implemented already), and might or might not require
+      futher, more component-specific proposals/KEPs
```
neolit123: `verify` complains about a typo, but i'd say `s/futher,//`

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+    * GenericComponentConfiguration (should be renamed to GenericControllerManagerConfiguration at the same time)
+* Assignee: @stewart-yu
+
+#### kube-controller-manager changes
+
+* **GroupVersionKind**: `kubecontrollermanager.config.k8s.io/v1alpha1.KubeControllerManagerConfiguration`
+* **External Package**: `k8s.io/kube-controller-manager/config/v1alpha1/types.go`
+* **Internal Package**: `k8s.io/kubernetes/pkg/controller/apis/config/types.go`
+* **Internal Scheme**: `k8s.io/kubernetes/pkg/controller/apis/config/scheme/scheme.go`
+* **Conversions & defaulting Package:** `k8s.io/kubernetes/pkg/controller/apis/config/v1alpha1/register.go`
```
stewart-yu: `s/register.go/default.go` ?

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+* **Internal Scheme**: `k8s.io/kubernetes/pkg/scheduler/apis/config/scheme/scheme.go`
+* **Conversions & defaulting Package:** `k8s.io/kubernetes/pkg/scheduler/apis/config/v1alpha1/register.go`
+* **Future Internal Package:** `k8s.io/kube-scheduler/pkg/apis/config/types.go`
+* Start referencing `ClientConnectionConfiguration` & `LeaderElectionConfiguration` from the generic ComponentConfig packages
+* Assignee: @dixudx
+
+#### k8s.io/controller-manager changes
+
+* **Not a "real" API group, instead shared packages only with both external and internal types.**
+* **External Package with defaulting & conversions**: `k8s.io/controller-manager/pkg/apis/config/v1alpha1/types.go`
```
stewart-yu: If we want to change the `default value` easy, can we put the `default.go` in core k8s directory? such `pkg/`. May be a immature suggestions

rosti:
@luxas @sttts Given the fact, that kubeadm is part of Kubernetes, do you think that we should do the same for its config? Some folks (Cluster API for example) may be interested in that in order to avoid pulling the whole Kubernetes repo as an external dependency just because of kubeadm config.

P.S: I am available if you need a hand in implementing this KEP.

sttts:
> Given the fact, that kubeadm is part of Kubernetes, do you think that we should do the same for its config?

Sounds reasonable. How do you consume kubeadm? Via binaries or containers? Is there anything else why you need to vendor k/k?

rosti:
At least in my case it is to generate the kubeadm config file and then invoke the executable with it. I assume, that most of the other people use kubeadm in the same way.
Basically kubeadm is mostly used by other external tools to deploy K8s on the current node, thus allowing for easier kubeadm config generation by external tools will be great.

luxas:
We plan to eventually move kubeadm to its own repo, yes. But that requires these other core ComponentConfigs to move out first. Moving kubeadm out is scope for an other proposal, I don't feel a need to include that in this proposal. Right now we do the same thing as proposed for the cloud-controller-manager, store our API under `cmd/`.

@rosti can we assign you the kube-proxy work items? k8s.io/apimachinery and k8s.io/apiserver needs to be done first, but if you feel like that is okay @sttts can add you as an assignee to the doc :)

rosti:
Yes, I can work on these. :)

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+          * [kube-controller-manager changes](#kube-controller-manager-changes)
+          * [cloud-controller-manager changes](#cloud-controller-manager-changes)
+          * [kube-apiserver changes](#kube-apiserver-changes)
+       * [Timeframe and Implementation Order](#timeframe-and-implementation-order)
+       * [OWNERS files for new packages and repos](#owners-files-for-new-packages-and-repos)
+
+## Summary
+
+Currently all ComponentConfiguration API types are in the core Kubernetes repo, which make them practically inaccessible to for any third-party tool,
+and blocks e.g. kubeadm to move out of the core repo. This KEP is about creating new staging repos, `k8s.io/{component}`, which will host the external
```
mattfarina: Two things.

1) A nit. The first sentence is hard to read. It might be wroth breaking into 2 or 3 sentences.
2) Given that k8s 1.11 makes it harder to vendor (missing generated files) this may be more important now than before.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+* **Package**: [k8s.io/kubernetes/pkg/kubelet/apis/kubeletconfig](https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/kubelet/apis/kubeletconfig/types.go)
+* **GroupVersionKind:** `kubelet.config.k8s.io/v1beta.KubeletConfiguration`
+* **Supports** reading **config from file** with **flag precedence**, **well-tested**.
+
+#### Current kube-proxy
+
+* **Package**: [k8s.io/kubernetes/pkg/proxy/apis/kubeproxyconfig](https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/proxy/apis/kubeproxyconfig/types.go)
+* **GroupVersionKind**: `kubeproxy.config.k8s.io/v1alpha1.KubeProxyConfiguration`
+* **Supports** reading **config from file**, **without flag precedence**, **not tested**.
+* This API group embeds `ClientConnectionConfiguration`, which is **wrong**
```
mtaufen: For things that are "wrong," it would be helpful to cross-reference a related issue.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### kubelet changes
+
+* **GroupVersionKind:** `kubelet.config.k8s.io/v1beta.KubeletConfiguration`
+* **External Package:** `k8s.io/kubelet/config/v1beta1/types.go`
+* **Internal Package:** `k8s.io/kubernetes/pkg/kubelet/apis/config/types.go`
+* **Internal Scheme:** `k8s.io/kubernetes/pkg/kubelet/apis/config/scheme/scheme.go`
+* **Conversions & defaulting Package:** `k8s.io/kubernetes/pkg/kubelet/apis/config/v1beta1/register.go`
+* **Future Internal Package:** `k8s.io/kubelet/pkg/apis/config/types.go`
+* Assignee: @mtaufen?
```
mtaufen: sgtm

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+### Goals
+
+* Find a home for the ComponentConfig API types, hosted as a staging repo in the "core" repo that is kubernetes/kubernetes
+* Make ComponentConfig API types consumable from *projects outside of kube* and from different parts of kube itself
+    * Resolve dependencies from the external ComponentConfig API types so that everything can depend on them
+    * The only dependency of the ComponentConfig API types should be `k8s.io/apimachinery`
+* Split internal types from versioned types
+* Remove the monolithic `componentconfig/v1alpha1` API group
+* Enable the staging bot so that a `[https://github.com/kubernetes/](https://github.com/kubernetes/){component}`
+  (imported as `k8s.io/{component}`) repos are created (this also requires changes to the k8s.io Golang import meta data).
```
thockin: The `go get` redirects are automatic now.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+#### kube-proxy changes
+
+* **GroupVersionKind**: `kubeproxy.config.k8s.io/v1alpha1.KubeProxyConfiguration`
+* **External Package**: `k8s.io/kube-proxy/config/v1alpha1/types.go`
+* **Internal Package**: `k8s.io/kubernetes/pkg/proxy/apis/config/types.go`
+* **Internal Scheme**: `k8s.io/kubernetes/pkg/proxy/apis/config/scheme/scheme.go`
+* **Conversions & defaulting Package:** `k8s.io/kubernetes/pkg/proxy/apis/config/v1alpha1/register.go`
+* **Future Internal Package:** `k8s.io/kube-proxy/pkg/apis/config/types.go`
+* Start referencing `ClientConnectionConfiguration` from the generic ComponentConfig packages
+* Assignee: ?
```
thockin: @rramkumar1 or @dcbw or @danwinship  - anyone willing to be assigned this?  I don't know how much actual urgency is here, but it's about time to start getting the config cleaned up?

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### k8s.io/apimachinery changes
+
+* **Not a "real" API group, instead shared packages only with both external and internal types.**
+* **External Package with defaulting & conversions**: `k8s.io/apimachinery/pkg/apis/config/v1alpha1/types.go`
+* **Internal Package**: `k8s.io/apimachinery/pkg/apis/config/types.go`
+* Structs to be hosted:
+    * ClientConnectionConfiguration
+* Assignee: ?
+
```
hanxiaoshuai: I want to do this, can I join it? @sttts

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+* **Not a "real" API group, instead shared packages only with both external and internal types.**
+* **External Package with defaulting & conversions**: `k8s.io/apiserver/pkg/apis/config/v1alpha1/types.go`
+* **Internal Package**: `k8s.io/apiserver/pkg/apis/config/types.go`
+* Structs to be hosted:
+    * LeaderElectionConfiguration
+    * DebuggingConfiguration
+    * later to be created: SecureServingConfiguration, AuthenticationConfiguration, AuthorizationConfiguration, etc.
+* Assignee: ?
+
```
hanxiaoshuai: I want to do this, can I join it? @sttts

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+### Migration strategy per component or k8s.io repo
+
+#### k8s.io/apimachinery changes
+
+* **Not a "real" API group, instead shared packages only with both external and internal types.**
+* **External Package with defaulting & conversions**: `k8s.io/apimachinery/pkg/apis/config/v1alpha1/types.go`
+* **Internal Package**: `k8s.io/apimachinery/pkg/apis/config/types.go`
+* Structs to be hosted:
+    * ClientConnectionConfiguration
+* Assignee: ?
```
stewart-yu: Does we need the work `k8s.io/kubernetes/pkg/util/pointer `package to` k8s.io/apimachinery`? i think can combine with this
If no others for help, i can help it too

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
hanxiaoshuai: I want to do this, can I join it? @sttts

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
m1093782566: @thockin You can assign it to here if all of other guys are busy :)

hanxiaoshuai: Maybe I can join it ？

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
hanxiaoshuai: I want to do this, can I join it? @sttts

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
stewart-yu: Does we need the work: move `k8s.io/kubernetes/pkg/util/pointer `package to` k8s.io/apimachinery' ? seems should combine with it @sttts

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,296 @@
...
+#### k8s.io/apiserver changes
+
+* **Not a "real" API group, instead shared packages only with both external and internal types.**
+* **External Package with defaulting & conversions**: `k8s.io/apiserver/pkg/apis/config/v1alpha1/types.go`
+* **Internal Package**: `k8s.io/apiserver/pkg/apis/config/types.go`
+* Structs to be hosted:
+    * LeaderElectionConfiguration
+    * DebuggingConfiguration
+    * later to be created: SecureServingConfiguration, AuthenticationConfiguration, AuthorizationConfiguration, etc.
+* Assignee: ?
```
hanxiaoshuai: I want to do this, can I join it? @sttts

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
sttts: 1. done 2. done

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
sttts: It's less about "wrong" here. Those types are duplicated in each API group instead of being shared. It's something we should unify. I updated the text.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
sttts: fixed

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
sttts: done

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
sttts: Added @m1093782566. @hanxiaoshuai feel free to collaborate on this one of course.

sttts:
Addressed all comments. ptal.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
m1093782566: Thanks @sttts , I would collaborate with @hanxiaoshuai :)

jbeda:
I'm supportive of this KEP. It is critical to find ways to (a) continue to move in the direction of richer config for k8s itself (beyond command line flags) and (b) decouple that config so that it isn't a big hairball.

Unfortunately I'm not as close to this stuff on a day to day basis to provide helpful actionable feedback on the details without significant time to dig in.  Let me know if you need that but it looks like others have it handled.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,297 @@
...
+    * ClientConnectionConfiguration
+* Assignee: @hanxiaoshuai
+
+#### k8s.io/apiserver changes
+
+* **Not a "real" API group, instead shared packages only with both external and internal types.**
+* **External Package with defaulting & conversions**: `k8s.io/apiserver/pkg/apis/config/v1alpha1/types.go`
+* **Internal Package**: `k8s.io/apiserver/pkg/apis/config/types.go`
+* Structs to be hosted:
+    * LeaderElectionConfiguration
```
liggitt: this would seem to fit under apimachinery, not apiserver (looking at the existing consumers)

thockin:
Feedback from sig-arch review today:

There was a discussion about whether to use `k8s.io/$component/config` or `k8s.io/$component-config`.  The important point is that it MUST be possible to vendor *just* the config definitions without vendoring the rest of the component's repo.  So the request for this KEP is to demonstrate and document that this is possible in the `k8s.io/$component/config` model with both dep and vgo as vendoring tools.  If it is not possible, we should consider the alternate model.

liggitt:
> the request for this KEP is to demonstrate and document that this is possible in the k8s.io/$component/config model with both dep and vgo as vendoring tools. If it is not possible, we should consider the alternate model.

for vgo specifically, this might require definition of a subdirectory module (e.g. "k8s.io/$component/config/go.mod", if the example at https://research.swtch.com/vgo-module is still accurate) and tags in the form "config/vX.Y". we should verify that such a subdirectory module is able to be vendored and version-tagged

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,297 @@
...
+    * DebuggingConfiguration
```
roycaihw: `DebuggingConfiguration` (which contains two boolean `EnableProfiling` and `EnableContentionProfiling`) is used by controller-manager, kube-CM and cloud-CM (it was originally split from `KubeControllerManagerConfiguration`). I wonder if it should belong to controller-manager, or is there plan to make it generic for other apiserver components.

stewart-yu: Have saw it, `apiserver` have `FeatureOptions` struct 
```
type FeatureOptions struct {
	EnableProfiling           bool
	EnableContentionProfiling bool
	EnableSwaggerUI           bool
}
```
May be we can split `EnableSwaggerUI ` from it, and make it more `generic`, Or @sttts  @luxas have another thoughts

sttts: Sounds reasonable to merge them into one generic part and additions for API servers separately. I am hestitant though to do these kind of refactoring here in the PR and also in the first move PRs. There will be a longer phase of cleanup of the existing config types. Luckily the apiserver and the controller managers do not use their types yet, and kube-scheduler only uses them as an alpha feature. So we are pretty free to refactor and clean up a lot there. I feel like we will possibly have another doc / KEP as a follow-up about the actual structured component config structure.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,297 @@
...
+  - sig-cloud-provider
+reviewers:
+  - "@thockin"
+  - "@liggitt"
+  - "@wojtek-t"
+  - "@stewart-yu"
+  - "@dixudx"
+approvers:
+  - "@thockin"
+  - "@jbeda"
```
timothysc: This should really be someone from api-machinery b/c it would be a updated policy.

luxas: @lavalamp as well?

luxas:
>  I wholeheartedly support this effort but how do we get and enforce accountability across the sigs on this?

The first step here @timothysc is to align the code structure across all the places (this proposal), and then after that we need individual (KEP?) proposals for the API structures to graduate to beta and beyond IMO, which is on the SIGs' own plates. We might wanna add some additional verify jobs for this after we see what makes sense and where this is going

luxas:
Also moving @sttts great comment top-level: https://github.com/kubernetes/community/pull/2354#discussion_r202260418

> Sounds reasonable to merge them into one generic part and additions for API servers separately. I am hestitant though to do these kind of refactoring here in the PR and also in the first move PRs. There will be a longer phase of cleanup of the existing config types. Luckily the apiserver and the controller managers do not use their types yet, and kube-scheduler only uses them as an alpha feature. So we are pretty free to refactor and clean up a lot there. I feel like we will possibly have another doc / KEP as a follow-up about the actual structured component config structure.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
sttts: @lavalamp is out of office for some time. Would add @deads2k instead.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
sttts: For the lack of a better place right now, I leave this in apiserver. We can move that elsewhere at another point in time if we find a better place. Apimachinery and Apiserver are equally wrong.

sttts:
Added @deads2k as approver, added a section about vendoring, added a non-goal of major refactoring of the actual ComponentConfig types.

ptal.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,324 @@
...
+* Set up staging publishing bot (async, non-critical)
+* The kubelet, kube-proxy and controller-manager follow, each one independently.
+
+### OWNERS files for new packages and repos
+
+* Approvers:
+    * @kubernetes/api-approvers
+    * @sttts
+    * @luxas
+    * @mtaufen
```
luxas: @mtaufen ok for you?

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,324 @@
...
+* Reviewers:
+    * @sttts
```
luxas: also the kubernetes/api-reviewers OWNERS alias

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
luxas: And the other place, which would be `k8s.io/controller-manager` is also kinda wrong as any component can do leader election, not only custom controller-managers.

keps/sig-cluster-lifecycle/0014-20180707-componentconfig-api-types-to-staging.md
```diff
@@ -0,0 +1,325 @@
...
+
+### OWNERS files for new packages and repos
+
+* Approvers:
+    * @kubernetes/api-approvers
+    * @sttts
+    * @luxas
+    * @mtaufen
+* Reviewers:
+	* @kubernetes/api-reviewers
```
m1093782566: nit: extra whitespaces?

sttts:
Are there any further comments?

According to the [KEP process](https://kubernetes.io/docs/imported/community/keps/#kep-workflow) next step is "implementable: The approvers have approved this KEP for implementation.".

@thockin @jbeda @deads2k is this ready for approval?

jbeda:
/approve

I'm giving this an approve.  KEPs are still mutable until they move into the "implementable" state (even then, reality means they can change, but the bar should be higher).  The goal is to merge early and iterate on smaller topics in separate PRs.

Can someone else give it the LGTM?

deads2k:
I think that API level defaulting for re-useable configuration types is an anti-goal since that significantly constrains the usage by composers who need or want different defaults.   How about removing defaulting as a thing we do for those types.  Defaults could be set per-process with a recommendation perhaps, but per-process, post deserialization and conversion.

lgtm otherwise.

andrewsykim:
LGTM from sig-cloud-provider, thanks for putting this together @luxas!

deads2k:
Once the bit about defaulting is gone, we can merge the rest and talk about defaulting in its own pull.  Changing the defaulting behavior when there is only one component actively using pieces is comparatively easy now.  You'd leave the defaulting function in place and simply call it post-serialization before usage.  But doing so avoids a pattern that will prevent composition of your config types later.

sttts:
@deads2k clarified defaulting that we post-pone that for a later discussion, and:
- we keep defaulting funcs only where necessary to keep existing behaviour
- defaulting should be moved to the leaf group if possible.

deads2k:
> @deads2k clarified defaulting that we post-pone that for a later discussion, and:
> 
> we keep defaulting funcs only where necessary to keep existing behaviour
> defaulting should be moved to the leaf group if possible.

Thanks

/lgtm

neolit123:
as mentioned by @luxas here:
https://github.com/kubernetes-sigs/kind/issues/107

some of these comments should be in this thread.

strict unmarshaling is something that will warn/error for problems in your YAML - duplicate fields and unknown fields. k8s is currently lacking this functionally which is not ideal...

we recently enabled this library fork to support it:
sigs.k8s.io/yaml.

the serializers in API machinery on the other hand do not use that library directly for JSON unmarshal, but only use the YAML->JSON functionality that it has. for JSON unmarshal the library json-iter is used.
https://github.com/kubernetes/kubernetes/pull/70839

^ errors are out of scope for now, but for warnings this is the place you can dig into (i.e. implementing a knob).

in the meantime the proposed solution is for consumers to implement a pre-step to validate input bytes using `sigs.k8s.io/yaml`. validation happens in `#1` in the example bellow.

```
sigs.k8s.io/yaml.UnmarshalStrict(bytes, externalConfigStructInstance) // #1
...
k8s.io/apimachinery/pkg/runtime.DecodeInto(myScheme.Codecs.UniversalDecoder(), bytes, internalConfigStructInstance);
```

liggitt:
we should work on an option to enable a strict codec, rather than reimplement kind/apiVersion detection and direct calls to yaml.UnmarshalStrict with external types in lots of places

Places that wanted to be strict could then enable that option

neolit123:
definitely, a strict codec would be ideal.

for reference, since i don't know how to write one (yet) we ended up with this in kubeadm for 1.13:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/config/strict/strict.go#L34

it finds a known GVK in known Schemas and creates an New() external config target to validate with `sigs.k8s.io/yaml.UnmarshalStrict()`.

luxas:
The follow-up KEP of this one is being created/discussed here: https://docs.google.com/document/d/1nZnzJD9dC0xrtEla2Xa-J6zobbC9oltdHIJ3KKSSIhk/edit. Please redirect comments/questions/concerns to that doc. And yes, totally we should be (to begin with) giving warnings when deserializing ComponentConfig files with unknown fields, using a stricter codec.

PR #2358, opened 2018-07-09
Change SIG Charter to reference base governance instead of copy-pasti…

pwittrock:
…ng it


philips:
I agree with these changes and think it will help us from creating a copy/paste mess.

/lgtm

PR #2359, opened 2018-07-09
Clean up SIG governance docs

pwittrock:


cblecker:
@pwittrock Did `/approve` not work for you?

PR #2363, opened 2018-07-11
WIP: KEP: Who are we and why are we here?

hh:
* Ask some introspective questions of our software interactions
* allow cluster wide aggregation of the answers
* analyse community wide results for non-obvious patterns


keps/sig-architecture/0016-20180601-identifying-api-usage-patterns.md
```diff
@@ -0,0 +1,243 @@
...
+Possibly something like:
+
+```
+kubelet/v1.12.0 (linux/amd64) k8s.io/kubelet b143093 built by test-infra@buildbot-10 02/03/85 23:44
+```
+
+### API interaction Purpose (Why are you here?)
+We must define a simple to implement, but contextually significant, answer to the question: *Why are you here?* It’s difficult to glean the purpose of an application interaction by external inspection without asking this obvious question.
+
+At the moment of making the API call, the application has access it’s own stack and history of source code location/lines and functions that brought it to make a request of an external API. Disabled by default, it could be enabled by setting a variable such as `KUBE_CLIENT_SUBMIT_PURPOSE`.
```
seh: s/it's/its/

hh: Thanks.

keps/sig-architecture/0016-20180601-identifying-api-usage-patterns.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Tying it all together: (How do I turn this on?)
+
+If all applications are compiled against a client-go (or other supported library)
+and support the env var `KUBE_CLIENT_SUBMIT_PURPOSE`,
+then deploying kubernetes itself with it set should
+enable all kubernetes components to begin transmitting identity and purpose.
+
+Setting this variable on all pods could be accomplished with
+an admission or initialization controller allowng all other
```
neolit123: `allowing`

hh: Thanks, fixed.

keps/sig-architecture/0016-20180601-identifying-api-usage-patterns.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Implementation History
+
+- 06/27/2018: initial design via google doc
+- 07/11/2018: submission of KEP
+- 07/??/2018: sponsorship by sig-????
+
+## Drawbacks [optional]
+
+Some feel this level of client ‘debugging’ doesn't belong in server-side logs.
```
hh: https://github.com/cncf/apisnoop/issues/23#issuecomment-399280296

DirectXMan12:
At a glance, some of this looks similar to tracing.  can you compare/contrast this a bit with opentracing?  It seems like you could inject the user agent and traceback stuff optionally as persistent tracing information, but I'm not super familiar with the details.

hh:
My goal is to make it simple to submit and gather API usage data both cluster and community wide. If we want to identify the _yet unseen patterns of use_, we need the ability to easily instrument all K8s API consumers.

The core of this KEP is to explore easy methods to enable any application to verbosely convey for each API call the _identity of the code behind this particular interaction_. I'd prefer to enable this behavior via a variable vs adding compile time instrumentation libraries specific to the our data collection runs.

Jaeger and Opentracing are great for debugging and development purposes. They do however require instrumentation logic to be added to the applications themselves. They do provide some beautiful insight, _if you already know what your looking for_. This would also limit the data we would be able to collect.

The initial APISnoop approach used mitm injection via a sidecar, similar to Envoy, and provided a non-kubernetes specific approach.
We explored using Istio / Envoy, but they were not designed to intercept communications destined for the APIServer.
At this point that things were looking a bit Rube Goldberg, hence the switch in focus to audit-logging.

While I'm enjoying the simplicity of audit-logging, it might be worth re-exploring the mitm / Envoy approach. 

If we route APIServer traffic to route through a proxy like Envoy we could support APISnooping more than just the K8s API.

dankohn:
@bhs could you please add any comments you have re OpenTracing.

bhs:
There are three pieces here:
1. Literally adding the necessary instrumentation to pass context / causality information from the clients into k8s, then through k8s while servicing the client API request(s)
1. Adding information (e.g., "purpose," though also what actually *happens* (beyond mere context propagation) to decorate the transactions for later analysis
1. Collection and analysis of all of the above

The third piece almost certainly does require "something new"... general-purpose tracing tools like Jaeger/Zipkin/etc are designed for performance analysis and root cause analysis, and it sounds like @hh intends for more bespoke sorts of insight as an output to the entire effort. It *is* conceivable to use a general-purpose distributed tracing system to collect the data, then write an analytical script to generate the human-ready output via API calls, but in any case "something new" will still need t obe written.

The second piece is kind of an "it depends"... OpenTracing has the notion of "baggage" which is designed for just this sort of user-level context information that has some user-level interpretation. At first glance it's very similar in spirit, though perhaps I'm missing something. Note that some people are nervous about baggage, though.

The first piece is where I see strong overlap with OpenTracing specifically. The actual context propagation and context management is what OT was designed for, and the API is thin, with no dependencies of its own (i.e., not a Rube Goldberg nightmare per the KEP's language).

I think the determinant would be (a) how much multi-process context propagation we expect to see here, and (b) what k8s wants to do with more debugging-centric context propagation. The main insight with (b) is that "context is context" regardless of the application, and the true "Rube Goldberg" outcome would be to record and propagate it multiple times in parallel.

If there was a decision to proceed with OT, I'd imagine that this effort could be / would be supported via a `Tracer` implementation that does some sort of passthrough/multiplexing with a "normal" `Tracer`... i.e., the `WhyAreWeHereTracer` / `WAWHTracer` would record the precise info it needed to, then pass things along to a (potentially noop) general-purpose `Tracer`.

Hope this helps.

dankohn:
@bhs thanks for the quick review. Let's plan to have @hh present this work to the OpenTracing community in the next couple months, once he has some progress to show.

quinton-hoole-2:
@hh Very entertaining :-).  Can this be closed?  If not can it be abbreviated to to the bare essentials to facilitate effective review?

hh:
I'm working on abbreviating it to the bare essentials. I'll set it back to WIP.

justaugustus:

justaugustus:

/close

PR #2364, opened 2018-07-11
KEP for Azure Availability Zones

feiskyer:
The KEP adds Azure Availability Zones support to Kubernetes, which includes:

- Detect availability zones automatically when registering new nodes and node's label `failure-domain.beta.kubernetes.io/zone` will be replaced with AZ instead of fault domain
- LoadBalancer and PublicIP will be provisioned with zone redundant
- `GetLabelsForVolume` interface will be implemented for Azure managed disks and it will also be added to `PersistentVolumeLabel` admission controller so as to support DynamicProvisioningScheduling

/sig azure
/sig storage


sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,238 @@
...
+
+```sh
+$ kubectl get pv --show-labels
+NAME           CAPACITY   ACCESSMODES   STATUS    CLAIM            REASON    AGE       LABELS
+pv-managed-abc 5Gi        RWO           Bound     default/claim1             46s       failure-domain.beta.kubernetes.io/region=centralus,failure-domain.beta.kubernetes.io/zone=centralus-1
+```
+
+### PersistentVolumeLabel
+
+Besides PVLabeler interface, [PersistentVolumeLabel](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/storage/persistentvolume/label/admission.go) admission controller should also updated with AzureDisk support, so that new PVs could be applied with above labels automatically.
```
msau42: This admission controller is deprecated now and I believe its replacement is in the cloud controller manager

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,238 @@
...
+kind: StorageClass
+metadata:
+  annotations:
+  labels:
+    kubernetes.io/cluster-service: "true"
+  name: managed-premium
+parameters:
+  kind: Managed
+  storageaccounttype: Premium_LRS
+  zone: "centralus-1"
```
msau42: The new topology aware dynamic provisioning feature adds a new allowedTopologies field to Storageclass which is intended to replace the zone/zones parameters.

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,238 @@
...
+
+LoadBalancer with standard SKU will be created and all available nodes (including VirtualMachines and VirtualMachineScaleSetVms, together with optional masters configured via excludeMasterFromStandardLB) are added to LoadBalancer backend pools.
+
+PublicIPs will also be created with standard SKU, and they are zone redundant by default.
+
+Note that zonal PublicIPs are not supported. We may add this easily if there’re clear use-cases in the future.
+
+## AzureDisk
+
+When Azure managed disks are created, the `PersistentVolumeLabel` admission controller automatically adds zone labels to them. The scheduler (via `VolumeZonePredicate`) will then ensure that pods that claim a given volume are only placed into the same zone as that volume, as volumes cannot be attached across zones.
```
msau42: Eventually the VolumeZone predicate will be deprecated and replaced by PV.NodeAffinity

sig-azure/0018-20180711-azure-availability-zones.md
feiskyer: Sorry, didn't notice this has been deprecated. Cloud controller manager will GA in v1.13 or v1.4. What's the replacement before its GA?

sig-azure/0018-20180711-azure-availability-zones.md
feiskyer: Topology-aware dynamic provisioning feature is still alpha in v1.12 if I'm not wrong, so these two parameters are still required before its GA, right?

And thanks of reminding this. `allowedTopologies` feature should also be added here.

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,238 @@
...
+
+This proposal aims to add [Azure Availability Zones (AZ)](https://azure.microsoft.com/en-us/global-infrastructure/availability-zones/) support to Kubernetes.
+
+## Scopes and Non-scopes
+
+### Scopes
+
+The proposal includes required changes to support availability zones for various functions in Azure cloud provider and AzureDisk volumes:
+
+- Detect availability zones automatically when registering new nodes and node's label `failure-domain.beta.kubernetes.io/zone` will be replaced with AZ instead of fault domain
```
liggitt: rather than having the kubelet self-register this, can the node controller pull this information and add it into the node object?

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,238 @@
...
+    "loadBalancerSku": "standard",
+    "excludeMasterFromStandardLB": true
+}
+```
+
+Note that with standard SKU LoadBalancer, `primaryAvailabitySetName` and `primaryScaleSetName` is not required because all available nodes (with configurable masters via `excludeMasterFromStandardLB`) are added to LoadBalancer backend pools.
+
+## Node registration
+
+When nodes are started, kubelet automatically adds labels to them with region and zone information:
```
liggitt: allowing nodes to self-label has been problematic (enumerated in https://github.com/kubernetes/community/pull/911). is there a reason we wouldn't have the node controller pull this info from the cloud provider and manage the labels on the node objects?

liggitt: cc @mikedanese

feiskyer: there're actually two ways supported today:

- Without cloud-controller-manager (kubelet --cloud-privider=azure), [kubelet registers itself](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L320) with zone labels
- With cloud-controller-manager (kubelet --cloud-provider=external), [node controller](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cloud/node_controller.go#L391) adds zone lables

Until cloud-controller-manager GA (probably v1.13 or v1.14), both way should be supported.

sig-azure/0018-20180711-azure-availability-zones.md
msau42: Correct just wanted to make sure you know that this area is changing and as part of beta, we want to get all the 3 in tree cloud provider plugins to start moving to the new method as well.

sig-azure/0018-20180711-azure-availability-zones.md
msau42: I'm not sure if there is a replacement in the meantime, but both methods should be implemented

sig-azure/0018-20180711-azure-availability-zones.md
liggitt: > Until cloud-controller-manager GA (probably v1.13 or v1.14), both way should be supported

I agree we need to support the old way until we can transition away from it, but I don't think we should expand it to new labels if possible

sig-azure/0018-20180711-azure-availability-zones.md
thockin: I don't understand the statement. You want to replace a standard label with ... what?

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Non-scopes
+
+Provisioning Kubernetes masters and nodes with availability zone support is not included in this proposal. It should be done in the provisioning tools (e.g. acs-engine). Azure cloud provider will auto-detect the node's availability zone if `availabilityZones` option is configured for the Azure cloud provider.
+
+## AZ label format
+
+Currently, Azure nodes are registered with label `failure-domain.beta.kubernetes.io/zone=faultDomain`.
+
+The format of fault domain is numbers (e.g. `1` or `2`), which is in same format with AZ (e.g. `1` or `3`). If AZ is using same format with faultDomain, then there'll be scheduler issues for clusters with both AZ and non-AZ nodes. So AZ will use a different format in kubernetes: `<region>-<AZ>`, e.g. `centralus-1`.
```
thockin: Maybe you should lead in with some details about what AZ means and what faultDomain means?  Why are they not interchangeable terms (they are both pretty commonly used to mean the same thing)

sig-azure/0018-20180711-azure-availability-zones.md
feiskyer: @msau42 you're right, 

- if cloud-controller-manager is used, then that is actually PVLabeler interface which is used persistent volume label controller
- or else, this admission controller is used.

Will add this info to both this and PVLabler part.

sig-azure/0018-20180711-azure-availability-zones.md
feiskyer: > I agree we need to support the old way until we can transition away from it, but I don't think we should expand it to new labels if possible

Agreed. this is only for existing zone labels.

sig-azure/0018-20180711-azure-availability-zones.md
feiskyer: Good catch. This should be definitely addressed.

sig-azure/0018-20180711-azure-availability-zones.md
liggitt: > Agreed. this is only for existing zone labels.

great, thanks for confirming

feiskyer:
Added a new commit with topology-aware provisioning and difference between fault domain and availability zones.

justaugustus:
@brendandburns / @thockin --

So from a Features Maintainer perspective, Feature Owners are required to submit design proposals for units of work that qualify as Features i.e., those tracked within `k/features`. The feature submission for Azure AZs can be found here: https://github.com/kubernetes/features/issues/586
I'd expect to see an uptick in KEPs across release if contributors are following the Features submission process.

From SIG Azure Chair perspective, anything that moves us along the line of transparency w.r.t. SIG roadmap is a step in the right direction. 
One thing that I appreciate about what's currently defined in the KEP metadata is that we consider scope; whether a KEP is `kubernetes-wide`, multi-SIG, single SIG, which lends itself to considering KEPs for what might be considered a smaller feature.

That said, I'm happy to firm up guidelines on expectations for KEP submission in SIG Azure! :)

feiskyer:
@thockin @brendandburns @justaugustus Thanks for clarification for KEP process. As Brendan's suggestion, could we continue the KEP and develop guidelines in sig-azure?

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -139,13 +146,13 @@ Note that zonal PublicIPs are not supported. We may add this easily if there’r
 
 ## AzureDisk
 
-When Azure managed disks are created, the `PersistentVolumeLabel` admission controller automatically adds zone labels to them. The scheduler (via `VolumeZonePredicate`) will then ensure that pods that claim a given volume are only placed into the same zone as that volume, as volumes cannot be attached across zones.
+When Azure managed disks are created, the `PersistentVolumeLabel` admission controller or PV controller automatically adds zone labels to them. The scheduler (via `VolumeZonePredicate` or `PV.NodeAffinity` in the future) will then ensure that pods that claim a given volume are only placed into the same zone as that volume, as volumes cannot be attached across zones.
```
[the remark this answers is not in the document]
feiskyer: Yep, exactly. Updated with that information.

justaugustus:
Absolutely @feiskyer. Let's keep this going!

feiskyer:
Rebased and squashed. @thockin @brendandburns @justaugustus @msau42 PTAL

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,279 @@
...
+- Zoned AzureDisk will only be provisioned when there are zoned nodes
+- If there are multiple zones, then those zones are chosen by round-robin
+
+Note that there are risks if the cluster is running with both zoned and non-zoned nodes. In such case, zoned AzureDisk can't be attached to non-zoned nodes. This means
+
+- new pods with zoned AzureDisks are always scheduled to zoned nodes
+- old pods using non-zoned AzureDisks can't be scheduled to zoned nodes
+
+So if users are planning to migrate workloads to zoned nodes, old AzureDisks should be recreated (probably backup first and restore to the new one).
+
```
ddebroy: Maybe you can add a section to call out the work needed in AzureDisks's `Provision()` to support Topology Aware Dynamic provisioning feature. This can cover the considerations in this [EBS PR](https://github.com/kubernetes/kubernetes/pull/65730) which will also be added for GCP.

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,279 @@
...
+## Scopes and Non-scopes
+
+### Scopes
+
+The proposal includes required changes to support availability zones for various functions in Azure cloud provider and AzureDisk volumes:
+
+- Detect availability zones automatically when registering new nodes (by kubelet or node controller) and node's label `failure-domain.beta.kubernetes.io/zone` will be replaced with AZ instead of fault domain
+- LoadBalancer and PublicIP will be provisioned with zone redundant
+- `GetLabelsForVolume` interface will be implemented for Azure managed disks and it will also be added to `PersistentVolumeLabel` admission controller so as to support DynamicProvisioningScheduling
```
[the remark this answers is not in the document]
feiskyer: Yep, exactly. Added the missing info in the new commit.

sig-azure/0018-20180711-azure-availability-zones.md
feiskyer: Thanks. I thought active zones means what you have said above. But for clarity, I also added this in the new commit.

For AzureDisks's Provision(), it should be the same logic as EBS. The EBS PR is indeed a good refer for azuredisk.

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,281 @@
...
+provisioner: kubernetes.io/azure-disk
+allowedTopologies:
+- matchLabelExpressions:
+  - key: failure-domain.beta.kubernetes.io/zone
+    values:
+    - centralus-1
+    - centralus-2
+```
+
+AzureDisk can only be created with one specific zone, so if multiple zones are specified in the storage class, then new disks will be provisioned with zone chosen arbitrarily among them.
```
ddebroy: Enabling support for `DynamicProvisioningScheduling` introduces several scenarios where the zone to provision a disk should not be chosen arbitrarily and specifying multiple zones in storage class should lead to an error:

If `DynamicProvisioningScheduling` is enabled and `VolumeBindingMode: WaitForFirstConsumer` is specified in the storage class, zone of the disk should be set to the zone of the node passed to `Provision()`. Specifying `zone`/`zones` in storage class should be considered an error in this scenario.

If `DynamicProvisioningScheduling` is enabled and `VolumeBindingMode: WaitForFirstConsumer` is not specified in StorageClass, zone of disk should be chosen from `allowedTopologies` or `zones` depending on which is specified. Specifying both `allowedTopologies` and `zones` should lead to error.

If `DynamicProvisioningScheduling` is disabled then the zone maybe arbitrarily chosen as specified above by arbitrarily choosing from the zones specified in the storage class.

sig-azure/0018-20180711-azure-availability-zones.md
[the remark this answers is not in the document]
ddebroy: Thanks. Suggestion/nit: something along the lines of this may capture the above better:

+ `GetLabelsForVolume` interface will be implemented for Azure managed disks so that PV label controller in cloud-controller-manager can appropriately add `Labels` and `NodeAffinity` to the Azure managed disk PVs. Additionally, `PersistentVolumeLabel` admission controller will be enhanced to achieve the same for Azure managed disks.

+  Azure Disk's `Provision()` function will be enhanced to take into account the zone of the node as well as `allowedTopologies` when determining the zone to create a disk in.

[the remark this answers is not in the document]
feiskyer: Thanks for the excellent suggestion. Fixed and squashed the commits.

feiskyer:
@thockin @brendandburns @justaugustus  @ddebroy Addressed comments and squashed commits. PTAL

justaugustus:
Overall LGTM, but I'd like to hold pending a review from Kal or Cole as well.
/lgtm
/hold
/cc @khenidak @colemickens

justaugustus:
Awesome, thanks for the review, Kal!
@feiskyer -- thanks for putting this together! :)

/hold cancel

justaugustus:
Ooooooo, I'm going to need to add an `OWNERS` for the sig-azure folder. 
Will handle next week.

khenidak:
:-) thanks

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

justaugustus:
> Ooooooo, I'm going to need to add an OWNERS for the sig-azure folder.
Will handle next week.

Fixed in https://github.com/kubernetes/community/pull/2399

/approve

justaugustus:
@jdumars @calebamiles @idvoretskyi -- could you `/approve` this?
It's approved from the SIG Azure side.

justaugustus:
Thanks @idvoretskyi :tada:

feiskyer:
Thanks @idvoretskyi and @justaugustus

sig-azure/0018-20180711-azure-availability-zones.md
```diff
@@ -0,0 +1,284 @@
...
+- Azure Disk's `Provision()` function will be enhanced to take into account the zone of the node as well as `allowedTopologies` when determining the zone to create a disk in.
+
+> Note that unlike most cases, fault domain and availability zones mean different on Azure:
+>
+> - A Fault Domain (FD) is essentially a rack of servers. It consumes subsystems like network, power, cooling etc.
+> - Availability Zones are unique physical locations within an Azure region. Each zone is made up of one or more data centers equipped with independent power, cooling, and networking.
+>
+> An Availability Zone in an Azure region is a combination of a fault domain and an update domain (Same like FD, but for updates. When upgrading a deployment, it is carried out one update domain at a time). For example, if you create three or more VMs across three zones in an Azure region, your VMs are effectively distributed across three fault domains and three update domains.
+
```
krmayankk: is there further classification like does an update domain have multiple fault domains ?

feiskyer: No, each VM is assigned with a single update domain.

PR #2383, opened 2018-07-18
fix a mis-placed section in schedule-DS-pod-by-scheduler.md

Huang-Wei:
This PR is to fix an issue that `NodeAffinity` section is mis-placed, and also add 4 spaces to make it indented properly.


nikhita:
I think k/community needs an explicit `/approve` for approval. `/lgtm` does not count as approval.

@bsalamat @k82cn can either of you approve? thanks!

PR #2384, opened 2018-07-18
add code of conduct committee bootstrapping doc

michelleN:


philips:
/approve 

we discussed this in the steering group and mailing list so I think this is just finalizing it on github. If you feel like it has had enough discussion feel free to hold cancel @michelleN 

/hold

PR #2385, opened 2018-07-18
move governance/ to sig-governance/ for clarity

michelleN:


michelleN:
Gonna close this for now bc I don't want to mess up links and confuse people.

PR #2387, opened 2018-07-18
Add more detailed plan for admission controller and webhook support to dry-run kep

jennybuckley:
This is to address https://github.com/kubernetes/community/pull/2301#discussion_r202886027, https://github.com/kubernetes/community/pull/2301#discussion_r202886261, and https://github.com/kubernetes/community/pull/2301#discussion_r202886329

### DIFF
--- a/keps/sig-api-machinery/0015-dry-run.md
+++ b/keps/sig-api-machinery/0015-dry-run.md
@@ -46,17 +46,69 @@ finer-grained mechanism later, if necessary.
 ## Admission controllers
 
 Admission controllers need to be modified to understand that the request is a
-“dry-run” request. While admission controllers are not supposed to have
-side-effects when triggered, some of them still do. Quotas for example uses the
-current request values to change the available quotas. Providing a parameter,
-either as an argument for built-in admission plugins or through a dryRun
-query-parameter for dynamic webhooks, will give them a chance not to have any
-side-effect.
+“dry-run” request. Admission controllers are allowed to have side-effects
+when triggered, as long as there is a reconciliation system, because it is not
+guaranteed that subsequent validating will permit the request to finish.
+Quotas for example uses the current request values to change the available quotas.
+The ```admission.Attributes``` interface will be edited like this, to inform the
+built-in admission controllers if a request is a dry-run:
+```golang
+type Attributes interface {
+	...
+	// IsDryRun indicates that modifications will definitely not be persisted for this request. This is to prevent
+	// admission controllers with side effects and a method of reconciliation from being overwhelmed.
+	// However, a value of false for this does not mean that the modification will be persisted, because it
+	// could still be rejected by a subsequent validation step.
+	IsDryRun() bool
+	...
+}
+```
 
-All admission controllers will have to be verified and changed. A new flag will
-be added so that webhooks can explicitly register with dry-run support. If
-dry-run is requested on a non-supported webhook, the request will be completely
-rejected.
+All built-in admission controllers will then have to be checked, and the ones with side
+effects will have to be changed to handle the dry-run case correctly. Some examples of
+built-in admission controllers with the possibility for side-effects are:
+- ResourceQuota
+- EventRateLimit
+- NamespaceAutoProvision
+- (Valid|Mut)atingAdmissionWebhook
+
+To address the possibility of webhook authors [relying on side effects](https://github.com/kubernetes/website/blame/836629cb118e0f74545cc7d6d97aa6b9edfa1a16/content/en/docs/reference/access-authn-authz/admission-controllers.md#L582-L584), a new field
+will be added to ```admissionregistration.k8s.io/v1beta1.ValidatingWebhookConfiguration``` and
+```admissionregistration.k8s.io/v1beta1.MutatingWebhookConfiguration``` so that webhooks
+can explicitly register as having dry-run support.
+If dry-run is requested on a non-supported webhook, the request will be completely rejected,
+as a 400: Bad Request. This field will be defaulted to true and deprecated in v1, and completely removed in v2.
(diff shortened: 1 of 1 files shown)


jennybuckley:
cc @liggitt since you noticed these details being missing in the first place
cc @apelisse since you wrote the kep in the first place
/sig api-machinery

keps/0015-dry-run.md
```diff
@@ -53,10 +53,52 @@ either as an argument for built-in admission plugins or through a dryRun
 query-parameter for dynamic webhooks, will give them a chance not to have any
```
liggitt: clarify the following:

> While admission controllers are not supposed to have side-effects when triggered, some of them still do. Quotas for example uses the side-effects when triggered

Side-effects are not prohibited as long as there is a reconciliation loop that accounts for the admission plugin getting called and the mutation not being persisted. The reason we want to provide a dry run indicator to admission plugins is so they can improve user experience in cases where waiting for that reconciliation would be disruptive (like quota).

keps/0015-dry-run.md
```diff
@@ -53,10 +53,52 @@ either as an argument for built-in admission plugins or through a dryRun
...
+// admissionregistration.k8s.io/v1beta1
+...
+type Webhook struct {
+	...
+	// SupportsDryRun defines whether this webhook will correctly handle dryRun requests.
+	// If false, any dryRun requests to resources/subresources this webhook applies to
+	// will be completely rejected and the webhook will not be called.
+	// Defaults to false.
+	// +optional
+	SupportsDryRun *bool `json:"supportsDryRun,omitempty" protobuf:"varint,6,number,opt,name=supportsDryRun"`
```
liggitt: `DryRunnable *bool`? Per https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#naming-conventions:

> The name of a field expressing a boolean property called 'fooable' should be called Fooable, not IsFooable.

That doc also says to think twice about bool fields. Having thought twice about it here, I think subdividing aspects of an admission plugin's dry runnability is too fine-grained, and a binary "supported" / "not supported" is better (especially since we want to default this on in future versions and require full support.

keps/0015-dry-run.md
```diff
@@ -53,10 +53,52 @@ either as an argument for built-in admission plugins or through a dryRun
...
+type AdmissionRequest struct {
+	...
+	// DryRun indicates that modifications will definitely not be persisted for this request.
+	// Defaults to false.
+	// +optional
+	DryRun *bool `json:"dryRun,omitempty" protobuf:"varint,11,number,opt,name=dryRun"`
+}
+```
+
+
```
liggitt: also need an update to the `admission.Attributes` interface (e.g. `DryRun() bool`)

liggitt:
thanks for the update, this file moved under the api-machinery folder, will need to rebase there

@apelisse were you planning to add details about the proposed behavior of generated fields separately (uid, creationTimestamp, resourceVersion, generateName, etc)?

keps/0015-dry-run.md
jennybuckley: I wasn't sure if that goes in the KEP because it wasn't an API change, but since it's an interface people could import I see how it could still go here.

keps/0015-dry-run.md
jennybuckley: I'll add some justification for this to the doc

jennybuckley: Do you think it should be DryRunnable?

jennybuckley:
Addressed comments  and rebased

keps/sig-api-machinery/0015-dry-run.md
```diff
@@ -46,17 +46,70 @@ finer-grained mechanism later, if necessary.
...
+// admissionregistration.k8s.io/v1beta1
+...
+type Webhook struct {
+	...
+	// DryRunnable defines whether this webhook will correctly handle dryRun requests.
+	// If false, any dryRun requests to resources/subresources this webhook applies to
+	// will be completely rejected and the webhook will not be called.
+	// Defaults to false.
+	// +optional
+	DryRunnable *bool `json:"supportsDryRun,omitempty" protobuf:"varint,6,number,opt,name=supportsDryRun"`
```
liggitt: nit: switch json and proto tags to match field name (`dryRunnable`)

keps/sig-api-machinery/0015-dry-run.md
```diff
@@ -46,17 +46,70 @@ finer-grained mechanism later, if necessary.
...
+- NamespaceAutoProvision
+- (Valid|Mut)atingAdmissionWebhook
+
+To address the possibility of webhook authors [relying on side effects](https://github.com/kubernetes/website/blame/836629cb118e0f74545cc7d6d97aa6b9edfa1a16/content/en/docs/reference/access-authn-authz/admission-controllers.md#L582-L584), a new field
+will be added to ```admissionregistration.k8s.io/v1beta1.ValidatingWebhookConfiguration``` and
+```admissionregistration.k8s.io/v1beta1.MutatingWebhookConfiguration``` so that webhooks
+can explicitly register as having dry-run support.
+If dry-run is requested on a non-supported webhook, the request will be completely rejected,
+as a 400: Bad Request, and the name of the webhook/webhooks which don't support dry run will be
+listed. This field will be defaulted to true and deprecated in v1, and completely removed in v2.
```
liggitt: > the name of the webhook/webhooks which don't support dry run will be listed

I wouldn't promise that this will surface via the API. We can log it, and consider whether we want this information leaking to the end user.

liggitt:
a couple nits, then this iteration looks good to go in. thanks

liggitt:
/assign @deads2k 
for approval

PR #2402, opened 2018-07-22
Reformat SIGs List

whs-dot-hk:
Reformat SIGs list so it now support unordered list
screen shot 2018-07-22 at 23 41 10


castrojo:
I assume moving the table to html is to avoid rendering differences between the markdown generator in hugo vs. the one github uses? 

/cc @cblecker to see if this affects any tooling.

whs-dot-hk:
@castrojo I turned it to html so unordered list could be display in table cells

PR #2404, opened 2018-07-23
Mention possibility to use -v flag when running godep-restore.sh

quasoft:
The documentation at [godep.md](https://github.com/kubernetes/community/blob/master/contributors/devel/godep.md) explains the use of the `hack/run-in-gopath.sh hack/godep-restore.sh` command for restoring dependencies.

This is fine, but the `hack/run-in-gopath.sh hack/godep-restore.sh` command shows no progress information and often takes a lot of time.

So it's good to mention that if you want to see what is happening, you can use the `-v` flag to turn on the verbose mode of `godep restore`.

This pull request is about adding a sentence at the end of section [Restoring deps](https://github.com/kubernetes/community/blob/master/contributors/devel/godep.md) explaining the possible use of the `-v` flag.

### DIFF
--- a/contributors/devel/godep.md
+++ b/contributors/devel/godep.md
@@ -63,6 +63,13 @@ simply `make clean` or `rm -rf _output`, and run it again.
 
 Now you should have a clean copy of all of the Kubernetes dependencies.
 
+Downloading dependencies might take a while, so if you want to see progress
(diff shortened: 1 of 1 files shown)


quasoft:
Signed CLA.

PR #2407, opened 2018-07-23
adds policy to dynamic audit kep

pbarker:
Adds ability to configure policy dynamically in the `AuditConfiguration` resource.

### DIFF
--- a/keps/sig-auth/0014-dynamic-audit-configuration.md
+++ b/keps/sig-auth/0014-dynamic-audit-configuration.md
@@ -17,8 +17,8 @@ approvers:
   - "@yliaog"
 editor: TBD
 creation-date: 2018-05-18
-last-updated: 2018-07-13
-status: provisional
+last-updated: 2018-07-31
+status: implementable
 ---
 
 # Dynamic Audit Control
@@ -38,29 +38,41 @@ status: provisional
         * [Story 1](#story-1)
         * [Story 2](#story-2)
         * [Story 3](#story-3)
+        * [Story 4](#story-4)
       * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+        * [Feature Gating](#feature-gating)
+        * [Policy Enforcement](#policy-enforcement)
+        * [Aggregated Servers](#aggregated-servers)
       * [Risks and Mitigations](#risks-and-mitigations)
         * [Privilege Escalation](#privilege-escalation)
+        * [Leaked Resources](#leaked-resources)
         * [Webhook Authentication](#webhook-authentication)
+        * [Performance](#performance)
   * [Graduation Criteria](#graduation-criteria)
   * [Implementation History](#implementation-history)
   * [Alternatives](#alternatives)
       * [Generalized Dynamic Configuration](#generalized-dynamic-configuration)
+      * [Policy Override](#policy-override)
 
 ## Summary
 
-We want to allow the advanced auditing features to be dynamically configured. Following in the same vein as [Dynamic Admission Control](https://kubernetes.io/docs/admin/extensible-admission-controllers/) we would like to provide a means of configuring the auditing features post cluster provisioning.
+We want to allow the advanced auditing features to be dynamically configured. Following in the same vein as 
+[Dynamic Admission Control](https://kubernetes.io/docs/admin/extensible-admission-controllers/) we would like to provide 
+a means of configuring the auditing features post cluster provisioning.
 
 ## Motivation
 
-The advanced auditing features are a powerful tool, yet difficult to configure. The configuration requires deep insight into the deployment mechanism of choice and often takes many iterations to configure properly requiring a restart of the apiserver each time. Moreover, the ability to install addon tools that configure and enhance audting is hindered by the overhead in configuration. Such tools frequently run on the cluster requiring future knowledge of how to reach them when the cluster is live. These tools could enhance the security and conformance of the cluster and its applications.
+The advanced auditing features are a powerful tool, yet difficult to configure. The configuration requires deep insight 
+into the deployment mechanism of choice and often takes many iterations to configure properly requiring a restart of 
+the apiserver each time. Moreover, the ability to install addon tools that configure and enhance auditing is hindered 
+by the overhead in configuration. Such tools frequently run on the cluster requiring future knowledge of how to reach 
+them when the cluster is live. These tools could enhance the security and conformance of the cluster and its applications.
 
 ### Goals
 - Provide an api and set of objects to configure the advanced auditing kube-apiserver configuration dynamically
 
 ### Non-Goals
 - Provide a generic interface to configure all kube-apiserver flags
-- composable audit policies per-endpoint
 - configuring non-webhook backends
 - configuring audit output (format or per-field filtering)
 - authorization of audit output
@@ -74,13 +86,18 @@ A new dynamic audit backend will be introduced that follows suit with the existi
 A cluster scoped configuration object will be provided that applies to all events in the cluster.
 
 ```golang
-// ClusterAuditConfiguration represents a cluster level audit configuration
-type ClusterAuditConfiguration struct {
+// AuditConfiguration represents a dynamic audit configuration
+type AuditConfiguration struct {
     metav1.TypeMeta
 
     v1.ObjectMeta
 
-    // Backends to send events
+    // Policy is the current audit v1beta1 Policy object
+    // if undefined it will default to the statically configured cluster policy if available
+    // if neither exist the backend will fail
+    Policy *Policy
+
+    // Backend to send events
     Backend *Backend
 }
 
@@ -127,12 +144,17 @@ type WebhookClientConfig struct {
 
 Multiple definitions can exist as independent solutions. These updates will require the audit API to be registered with the apiserver. The dynamic configurations will be wrapped by truncate and batch options, which are set statically through existing flags. Dynamic configuration will be enabled by a feature gate for pre-stable releases. If existing flags are provided to configure the audit backend they will be taken as a separate backend configuration.
 
-Example cluster yaml config:   
+Example configuration yaml config:   
 ```yaml
 apiVersion: audit.k8s.io/v1beta1
-kind: ClusterAuditConfiguration
+kind: AuditConfiguration
 metadata:
   name: <name>
+policy:
+  rules:
+  - level: <level>
+  omitStages:
+  - stage: <stage>
 backend:
   webhook:
   - initialBackoff: <10s>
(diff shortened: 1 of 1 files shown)


pbarker:
@tallclair updated with details on implementation and performance. PTAL

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -148,19 +155,32 @@ backend:
...
+
+Implementing the [attribute interface](/Users/patrick/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/authorization/authorizer/interfaces.go) based on the Event struct was also explored. This would allow us to keep the existing `Sink` interfaces, however it would require parsing the request URI twice in the pipeline due to how that field is represented in the Event. This was determined to not be worth the cost.
```
tallclair: I'm not sure I understand this, but parsing the request URI twice is likely to be much cheaper than copying the audit event N times.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -148,19 +155,32 @@ backend:
...
 
-Inherently apiserver aggregates and HA apiserver setups will work off the same dynamic configuration object. If separate objects are needed they should be configured as static objects on the node and set through the runtime flags. Aggregated servers will implement the same audit handling mechanisms. A conformance test should be provided as assurance. This needs further discussion with the participating sigs.
+#### Aggregated Servers
+Inherently apiserver aggregates and HA apiserver setups will work off the same dynamic configuration object. If separate objects are needed they should be configured as static objects on the node and set through the runtime flags. Aggregated servers will implement the same audit handling mechanisms. A conformance test should be provided as assurance. Metadata level logging will happen by default at the main api server as it proxies the traffic. The aggregated server will then watch the same configuration objects and only log on resource types that it handles.
```
tallclair: note: this means that there will be duplicate events at each stage. This isn't necessarily wrong, but it should be documented. The receiving servers should not expect to key off { Audit-ID x Stage }

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -173,6 +193,11 @@ This does open up the attack surface of the audit mechanisms. Having them strict
 
 As a mitigation strategy policy configured through a static file on the api server will not be accessible through the api. This file ensures that an escalation attack cannot tamper with a root configuration, but works independently of any dynamically configured objects.
 
+#### Leaked Resources
+A user is granted access to create audit policies, and inadvertently exposes secret resources.
```
tallclair: To make the risk clearer, prefer:
```
A user with permissions to create audit policies effectively has read access to the entire cluster (including all secrets data).
```

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -182,6 +207,10 @@ It may also be reasonable to provide a dynamic auth configuration from secrets,
 
 This needs further discussion.
 
+#### Performance
+
+These changes will likely have an O(n) performance impact on the api server per policy.  A `DeepCopy` of the event will be required for each backend. Also, the request/response object would now be serialized on every [request](https://github.com/kubernetes/kubernetes/blob/cef2d325ee1be894e883d63013f75cfac5cb1246/staging/src/k8s.io/apiserver/pkg/audit/request.go#L150-L152). Benchmark testing will be required to understand the scope of the impact and what optimizations may be required. This impact is gated by opt-in feature flags, which largely mitigates the concern.
```
tallclair: > This impact is gated by opt-in feature flags, which largely mitigates the concern.

I disagree with this statement. I think it unblocks the alpha implementation & lets us start benchmarking, but it is still a concern that needs to be addressed (or dismissed with data) before going to beta.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -193,9 +222,20 @@ Success will be determined by stability of the provided mechanisms and ease of u
...
+
+### Policy Override
+
+There has been discussion over whether the policy configured by api server flags should limit the policies configured dynamically. This would allow a cluster admin to narrowly define what is allowed to be logged by the dynamic configurations. While this has upsides it was ruled out for the following reasons: 
+
+* It would limit user story #4 in the ability to quickly turn up logging when needed 
+* It could prove difficult to understand as the policies themselves are fairly complex 
+* The use of CRDs would be difficult to bound
+
+The dynamic policy feature is gated by runtime flags. This still provides the cluster provisioner a means to limit audit logging to the single runtime object if needed.
```
tallclair: feature gates aren't intended as long-term configuration options. If this is a desirable knob, we should add a non-feature gate flag for it.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -17,8 +17,8 @@ approvers:
   - "@yliaog"
 editor: TBD
 creation-date: 2018-05-18
-last-updated: 2018-07-13
-status: provisional
+last-updated: 2018-07-30
+status: implementable
```
tallclair: Please either revert this or close https://github.com/kubernetes/community/pull/2417

pbarker: closed

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -80,7 +79,10 @@ type ClusterAuditConfiguration struct {
 
     v1.ObjectMeta
 
-    // Backends to send events
+    // Policy is the current audit v1beta1 Policy object
+    Policy Policy
```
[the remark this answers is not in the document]
tallclair: I think it makes sense. The default policy is a bit complicated - not just "log everything @request-response", so not requiring users to come up with their own version or copy that into the config here would simplify the setup quite a bit.

If you agree, make this a pointer, and update the comment to explain that.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -80,7 +85,11 @@ type ClusterAuditConfiguration struct {
 
     v1.ObjectMeta
 
-    // Backends to send events
+    // Policy is the current audit v1beta1 Policy object
+    // if undefined it will default to the runtime flag provided policy if available
```
tallclair: prefer: s/runtime flag provided/statically configured cluster/

tallclair: "if available" - what if it isn't?

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -144,23 +158,34 @@ backend:
...
+Implementing the [attribute interface](/Users/patrick/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/authorization/authorizer/interfaces.go) based on the Event struct was also explored. This would allow us to keep the existing `Sink` interfaces, however it would require parsing the request URI twice in the pipeline due to how that field is represented in the Event. This was determined to not be worth the cost.
+
+#### Aggregated Servers
+Inherently apiserver aggregates and HA apiserver setups will work off the same dynamic configuration object. If separate objects are needed they should be configured as static objects on the node and set through the runtime flags. Aggregated servers will implement the same audit handling mechanisms. A conformance test should be provided as assurance. Metadata level logging will happen by default at the main api server as it proxies the traffic. The aggregated server will then watch the same configuration objects and only log on resource types that it handles. This will duplicate the events sent to the receiving servers should they not expect to key off `{ Audit-ID x Stage }`.
```
tallclair: nit: grammar in the last sentence. I think you're missing "so"

tallclair: > This will duplicate ...

Do we want to provide an option to _not_ duplicate it? E.g. if you're logging at metadata, there's not much reason to duplicate those events with the extension server (unless you're debugging an issue with that server). I'm not sure how we'd implement that though - do api servers know whether they're extension or aggregator servers?

tallclair:
Should we change the resource name from `ClusterAuditConfiguration` to `AuditConfiguration`? I think you went with `Cluster` with the intention of adding a namespaced config later.

keps/sig-auth/0014-dynamic-audit-configuration.md
pbarker: > do api servers know whether they're extension or aggregator servers?
not sure on that one, I could see this as being a useful add but maybe something we track as a later feature.

keps/sig-auth/0014-dynamic-audit-configuration.md
pbarker: updated

pbarker:
@tallclair went ahead and renamed to `AuditConfiguration` to simplify things

tallclair:
/lgtm
/hold

I'm looking into the approval process before letting this merge.

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -144,23 +166,53 @@ backend:
...
+based on a shared informer. For the existing backends to comply, a method will be added that implements the `EnforcedSink` interface.
+
+Implementing the [attribute interface](/Users/patrick/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/authorization/authorizer/interfaces.go) 
+based on the Event struct was also explored. This would allow us to keep the existing `Sink` interfaces, however it would 
+require parsing the request URI twice in the pipeline due to how that field is represented in the Event. This was determined 
+to not be worth the cost.
+
+#### Aggregated Servers
+Inherently apiserver aggregates and HA apiserver setups will work off the same dynamic configuration object. If separate 
+objects are needed they should be configured as static objects on the node and set through the runtime flags. Aggregated
```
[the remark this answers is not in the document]
pbarker: Yeah separate audit configuration objects, I'll update to reflect

keps/sig-auth/0014-dynamic-audit-configuration.md
```diff
@@ -144,23 +166,53 @@ backend:
...
+require parsing the request URI twice in the pipeline due to how that field is represented in the Event. This was determined 
+to not be worth the cost.
+
+#### Aggregated Servers
+Inherently apiserver aggregates and HA apiserver setups will work off the same dynamic configuration object. If separate 
+objects are needed they should be configured as static objects on the node and set through the runtime flags. Aggregated 
+servers will implement the same audit handling mechanisms. A conformance test should be provided as assurance. Metadata level 
+logging will happen by default at the main api server as it proxies the traffic. The aggregated server will then watch the same 
+configuration objects and only log on resource types that it handles. This will duplicate the events sent to the receiving servers 
+so they should not expect to key off `{ Audit-ID x Stage }`.
```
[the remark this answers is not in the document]
pbarker: This just implies that in some scenarios the event may occur in both servers, so to be cognizant of that when developing features.

pbarker:
thank you @caesarxuchao! @tallclair and I just wanted input from api machinery before merging, I updated the doc based on your input

keps/sig-auth/0014-dynamic-audit-configuration.md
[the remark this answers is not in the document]
caesarxuchao: Are all the event sent to all the servers, and is up to the servers to filter the events?

PR #2411, opened 2018-07-24
Document rules and best practices of default values

janetkuo:
For context: https://github.com/kubernetes/kubernetes/issues/66135

/assign @bgrant0607 @liggitt @enisoc


contributors/devel/api_changes.md
```diff
@@ -365,9 +371,18 @@ being required otherwise.
...
-
-*Note:* In the past the core v1 API
+need to add cases to `pkg/apis/<group>/<version>/defaults.go`.
+
+**Note:** When adding default values to new fields, you *must* also add default
+values in all API versions, instead of leaving new fields unset (e.g. `nil`) in
+old API versions. This is required because defaulting happens whenever a
+serialized version is read (see [#66135]). When possible, pick meaningful values
+as sentinels for unset values. For example,
+`DeploymentSpec.ProgressDeadlineSeconds *int32` can have "max int32" (i.e.
```
thockin: would -1 be a cleaner example?

thockin:
Hmm, I know I can memorize -1.  I might even memorize INT32_MAX.  BUt I
sure have not memorized JSON's int-as-float max.  That means I have to hit
the docs any time I want to use this - whether to specify "infinite" or
just to compare against it?  I guess as the default I specify infinite by
leaving it blank, and comparison can be a constant.  Meh...

This set precendent..

janetkuo:
> Hmm, I know I can memorize -1.  I might even memorize INT32_MAX.  BUt I
sure have not memorized JSON's int-as-float max.  That means I have to hit
the docs any time I want to use this - whether to specify "infinite" or
just to compare against it?  I guess as the default I specify infinite by
leaving it blank, and comparison can be a constant.  Meh...

For some fields, `-1` isn't allowed in validation. Relaxing validation isn't allowed, and even if we treat this as a special case to relax validation, objects might be in a weird state after the users downgrade k8s to a previous version. 

We documented "infinite" value in API doc when setting default values: https://github.com/kubernetes/kubernetes/pull/66581/files#diff-718f149c23dd5e15fc407bd7a2f9d840L357.

thockin:
We probably can't readjudicate old APIs.  Let's just be sure this is what
we want for new ones.  It will be copied.

liggitt:
> We probably can't readjudicate old APIs. Let's just be sure this is what we want for new ones. It will be copied.

Agree. A marker value like -1 or 0 is way easier to deal with than 2^53-1.

> For some fields, -1 isn't allowed in validation.

When introducing a new field and determining a default value and validation, we could allow 0 or -1 at that point.

thockin:
@liggitt Do you think it's worth continuing the debate?  I'd much rather see new fields do sane sentinels and opportunistically enable sane sentinels in new api versions.

I don't mean to block this PR indefinitely, though.

janetkuo:
I further clarified the rules and removed the example. We can merge the rule first and add the example/convention later.

liggitt:
I'm happy with this now.

PR #2412, opened 2018-07-24
Improve dry-run KEP

apelisse:

### DIFF
--- a/keps/sig-api-machinery/0015-dry-run.md
+++ b/keps/sig-api-machinery/0015-dry-run.md
@@ -110,9 +110,32 @@ type AdmissionRequest struct {
 }
 ```
 
+## Generated values
+
+Some values of the object are typically generated before the object is persisted:
+- generateName can be used to assign a unique random name to the object,
+- creationTimeStamp records the time of creation,
+- UID uniquely identifies the object and is randomly generated (non-deterministic),
+- resourceVersion tracks the persisted version of the object.
+
+Most of these values are not useful in the context of dry-run, and could create
+some confusion. The UID and the generated name would have a different value in a
+dry-run and non-dry-run creation. These values will be left empty when
+performing a dry-run.
+
+CreationTimeStamp is also generated on creation, but there are less ways to
+abuse it so it will be generated as it would for a regular request.
+
+ResourceVersion will also be left empty on creation.
+
 ## Storage
 
 The storage layer will be modified, so that it can know if request is dry-run,
 most likely by looking for the field in the “Options” structure (missing for
 some handlers, to be added). If it is, it will NOT store the object, but return
 success. That success can be forwarded back to the user.
+
+A dry-run request should behave as close as possible to regular
+request. Attempting to dry-run create en existing object will result in an
+`AlreadyExists` error to be returned. Similarly, if an dr-run update is
+performed on a non-existing object, a `NotFound` error will be returned.
--- a/keps/sig-api-machinery/0015-dry-run.md
+++ b/keps/sig-api-machinery/0015-dry-run.md
@@ -15,7 +15,7 @@ approvers:
 editor: apelisse
 creation-date: 2018-06-21
 last-updated: 2018-06-21
-status: provisional
+status: implementable
 ---
 # Kubernetes Dry-run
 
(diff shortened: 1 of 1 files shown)


apelisse:
@liggitt

keps/sig-api-machinery/0015-dry-run.md
```diff
@@ -110,9 +110,32 @@ type AdmissionRequest struct {
 }
 ```
 
+## Generated values
+
+Some values of the object are typically generated before the object is persisted:
+- generateName can be used to assign a unique random name to the object,
+- creationTimeStamp records the time of creation,
+- UID uniquely identifies the object and is randomly generated (non-deterministic),
+- revisionVersion tracks the persisted version of the object.
```
liggitt: `resourceVersion`

apelisse:
Addressed both comments, thanks!

keps/sig-api-machinery/0015-dry-run.md
```diff
@@ -110,9 +110,32 @@ type AdmissionRequest struct {
...
+
+Most of these values are not useful in the context of dry-run, and could create
+some confusion. The UID and the generated name would have a different value in a
+dry-run and non-dry-run creation. These values will be left empty when
+performing a dry-run.
+
+CreationTimeStamp is also generated on creation, but there are less ways to
+abuse it so it will be generated as it would for a regular request.
+
+ResourceVersion will also be left empty on creation.
```
jennybuckley: I agree that we should leave it empty on creation. And on updates, I think the value should be unchanged for dry-run requests

jennybuckley: @liggitt do you think this is the way we want to go? I think it would be better than leaking some information (probably incorrect information) about what the next rv is going to be

keps/sig-api-machinery/0015-dry-run.md
```diff
@@ -110,9 +110,33 @@ type AdmissionRequest struct {
...
+
 ## Storage
 
 The storage layer will be modified, so that it can know if request is dry-run,
 most likely by looking for the field in the “Options” structure (missing for
 some handlers, to be added). If it is, it will NOT store the object, but return
 success. That success can be forwarded back to the user.
+
+A dry-run request should behave as close as possible to regular
+request. Attempting to dry-run create en existing object will result in an
```
liggitt: a couple typos: `an existing object`, and `if a dry-run update`

liggitt:
a couple nits, lgtm otherwise

apelisse:
@liggitt Fixed, thanks

keps/sig-api-machinery/0015-dry-run.md
liggitt: yeah, leaving empty/unchanged makes sense to me

liggitt:
/lgtm

@deads2k for approval

currently tracked in https://github.com/kubernetes-sigs/architecture-tracking/issues/13 and is in the "sig-review" list in https://github.com/kubernetes-sigs/architecture-tracking/projects/2. @jdumars - what's the process for getting it into the API review pipeline in https://github.com/kubernetes-sigs/architecture-tracking/projects/3?

deads2k:
As a user of this API, I suspect that at some point I'll want to see **all** the errors I'd get, not just the first one, but that can be a later extension.

I'd like the clarification on deletiontimestamp.  

/approve
/hold

holding to get the deletiontimestamp clarified, but I suspect it parallels creationtimestamp

keps/sig-api-machinery/0015-dry-run.md
```diff
@@ -110,9 +110,33 @@ type AdmissionRequest struct {
...
+- creationTimeStamp records the time of creation,
+- UID uniquely identifies the object and is randomly generated (non-deterministic),
+- resourceVersion tracks the persisted version of the object.
+
+Most of these values are not useful in the context of dry-run, and could create
+some confusion. The UID and the generated name would have a different value in a
+dry-run and non-dry-run creation. These values will be left empty when
+performing a dry-run.
+
+CreationTimeStamp is also generated on creation, but there are less ways to
```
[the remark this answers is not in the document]
liggitt: you think this will be used as an expensive "get the current server time" API?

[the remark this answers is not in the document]
deads2k: > you think this will be used as an expensive "get the current server time" API?

Delete is weird in general.  I think an exploration of deletion semantics will be required as we try to make it useable.

About creationtimestamp, as a consumer, wouldn't you do it to set the skew?  We've had requests for similar things.  It wouldn't stop me allowing it to be set if that were easier though.

apelisse:
@deads2k @liggitt Added a note for deletion (I made it similar to creation)

PR #2413, opened 2018-07-24
KEP: apiserver validates etcd

ccding:


keps/sig-api-machinery/0017-apiserver-validates-etcd-instance.md
```diff
@@ -0,0 +1,61 @@
...
+instance, it checks the cluster id stored in etcd. There are three possible
+outcomes.
+
+- if the etcd doesn't have a cluster id, it means the etcd is a fresh etcd.
+  apiserver will write the cluster id to etcd, which will be used for
+  validation in future. We have to make sure there is no race condition here.
+- if the etcd stores a cluster id, and it matches the one configured to
+  apiserver, apiserver uses the etcd normally.
+- if the etcd stores a cluster id, but it doesn't match the one configured to
+  apiserver, apiserver reports error and quits.
```
smarterclayton: How would you restore from a backup?  That will give you a new cluster ID

keps/sig-api-machinery/0017-apiserver-validates-etcd-instance.md
```diff
@@ -0,0 +1,61 @@
...
+# apiserver validates etcd instance
+
+## Summary
+
+We would like to have apiserver to validate the etcd instance to make sure it
+is connected to the proper one.
+
+## Motivation
+
+When there are multiple deployments of kubernetes and etcd, it may cause a
```
smarterclayton: Can you give more evidence to support this?  I can't say that I've heard this as a common complaint.  When would you have a Kubernetes apiserver that is confused about which etcd to talk to?

smarterclayton: Some examples of where this happened, what caused it, and any other issues or discussion in the community that indicates this is a common problem would be helpful for making a case for this change.

keps/sig-api-machinery/0017-apiserver-validates-etcd-instance.md
ccding: when an apiserver crashes, and a new one is initiated, the new one has to use the same cluster id as the crashed one.

keps/sig-api-machinery/0017-apiserver-validates-etcd-instance.md
ccding: Human makes mistakes. It happens when the apiserver is misconfigured, which messes up all the nodes--a disaster to the whole system. It happened a few times in our production environment.

liggitt: > It happened a few times in our production environment.

Securing etcd instances with per-cluster credentials would prevent apiservers messing with etcd instances belonging to other clusters.

That is a strong first recommendation (for many reasons) that is possible today, and would ensure the apiserver only talks to the intended etcd cluster.

liggitt: I'd support deprecating unsecured apiserver<->communication for this reason among others

ccding: it makes sense. thanks

PR #2417, opened 2018-07-25
marks kep 14 as implementable

pbarker:
marks kep 14 for dynamic auditing as implementable


pbarker:
closing in favor of https://github.com/kubernetes/community/pull/2407

PR #2420, opened 2018-07-25
KEP for CRD conversion webhook

mbohlool:
CRD supports multiple version but no conversion between them (something called nopConverter which only change the apiVersion of the CR). With this proposal we introduced a conversion mechanism for CRDs based on an external webhook. Detail API changes, use cases and upgrade/downgrade scenarios are discussed.

/assign @deads2k @lavalamp @erictune @liggitt 
/cc @sttts @enisoc

You can find the original design doc and discussions [here](http://bit.ly/k8s-crd-conversion)

/sig api-machinery

### DIFF
--- a/contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
+++ b/contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
@@ -1,0 +1,859 @@
+# CRD Conversion Webhook
+
+Status: Approved
+
+Version: Alpha
+
+Implementation Owner: @mbohlool
+
+Authors: @mbohlool, @erictune
+
+Thanks: @dbsmith, @deads2k, @sttts, @liggit, @enisoc
+
+### Summary
+
+This document proposes a detailed plan for adding support for version-conversion of  Kubernetes resources defined via Custom Resource Definitions (CRD).  The API Server is extended to call out to a webhook at appropriate parts of the handler stack for CRDs.  
+
+No new resources are added; the [CRD resource](https://github.com/kubernetes/kubernetes/blob/34383aa0a49ab916d74ea897cebc79ce0acfc9dd/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/types.go#L187) is extended to include conversion information as well as multiple schema definitions, one for each apiVersion that is to be served.
+
+
+## Definitions
+
+**Webhook Resource**: a Kubernetes resource (or portion of a resource) that informs the API Server that it should call out to a Webhook Host for certain operations. 
+
+**Webhook Host**: a process / binary which accepts HTTP connections, intended to be called by the Kubernetes API Server as part of a Webhook.
+
+**Webhook**: In Kubernetes, refers to the idea of having the API server make an HTTP request to another service at a point in its request processing stack.  Examples are [Authentication webhooks](https://kubernetes.io/docs/reference/access-authn-authz/webhook/) and [Admission Webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/).  Usually refers to the system of Webhook Host and Webhook Resource together, but occasionally used to mean just Host or just Resource.
+
+**Conversion Webhook**: Webhook that can convert an object from one version to another.
+
+**Custom Resource**: In the context of this document, it refers to resources defined as Custom Resource Definition (in contrast with extension API server’s resources).
+
+**CRD Package**: CRD definition, plus associated associated controller deployment, RBAC roles, etc, which is released by a developer who uses CRDs to create new APIs.
+
+  
+## Motivation
+
+Version conversion is, in our experience, the most requested improvement to CRDs.  Prospective CRD users want to be certain they can evolve their API before they start down the path of developing a CRD + controller. 
+
+
+## Requirements
+
+* As an existing author of a CRD, I can update my API's schema, without breaking existing clients.  To that end, I can write a CRD(s) that supports one kind with two (or more) versions.  Users of this API can access an object via either version (v1 or v2), and are accessing the same underlying storage (assuming that I have properly defined how to convert between v1 and v2.)
+
+* As a prospective user of CRDs, I don't know what schema changes I may need in the future, but I want to know that they will be possible before I chose CRDs (over EAS, or over a non-Kubernetes API).
+
+* As an author of a CRD Package, my users can upgrade to a new version of my package, and can downgrade to a prior version of my package (assuming that they follow proper upgrade and downgrade procedures; these should not require direct etcd access.)
+
+* As a user, I should be able to request CR in any supported version defined by CRD and get an object has been properly converted to the requested version (assuming the CRD Package Author has properly defined how to convert).
+
+* As an author of a CRD that does not use validation, I can still have different versions which undergo conversion.
+
+* As a user, when I request an object, and webhook-conversion fails, I get an error message that helps me understand the problem.
+
+* As an API machinery code maintainer, this change should not make the API machinery code harder to maintain
+
+* As a cluster owner, when I upgrade to the version of Kubernetes that supports CRD multiple versions, but I don't use the new feature, my existing CRDs work fine.  I can roll back to the previous version without any special action.
+
+
+## Summary of Changes
+
+1. A CRD object now represents a group/kind with one or more versions.
+
+2. The CRD API (CustomResourceDefinitionSpec) is extended as follows:
+
+    1. It has a place to register 1 webhook.
+
+    2. it holds multiple "versions".
+
+    3. Some fields which were part of the .spec are now per-version; namely Schema, Subresources, and AdditionalPrinterColumns.
+
+3. A Webhook Host is used to do conversion for a CRD.
+
+    4. CRD authors will need to write a Webhook Host that accepts any version and returns any version.
+
+    5. Toolkits like kube-builder and operator-sdk are expected to provide flows to assist users to generate Webhook Hosts.
+
+
+## Detailed Design
+
+
+### CRD API Changes
+
+The CustomResourceDefinitionSpec is extended to have a new section where webhooks are defined: 
+
+```golang
+// CustomResourceDefinitionSpec describes how a user wants their resource to appear
+type CustomResourceDefinitionSpec struct {
+  Group string
+  Version string
+  Names CustomResourceDefinitionNames
+  Scope ResourceScope
+  // This optional and correspond to the first version in the versions list
+  Validation *CustomResourceValidation
+  // Optional, correspond to the first version in the versions list 
+  Subresources *CustomResourceSubresources
+  Versions []CustomResourceDefinitionVersion
+  // Optional, and correspond to the first version in the versions list  
+  AdditionalPrinterColumns []CustomResourceColumnDefinition
+
+  Conversion *CustomResourceConversion
+}
+
+type CustomResourceDefinitionVersion struct {
+  Name string
+  Served Boolean
+  Storage Boolean
+  // These three fields should not be set for first item in Versions list
+  Schema *JSONSchemaProp
+  Subresources *CustomResourceSubresources
+  AdditionalPrinterColumns []CustomResourceColumnDefinition
+}
+
+Type CustomResourceConversion struct {
+  // Conversion strategy, either "nop” or "webhook”. If webhook is set, Webhook field is required.
+  Strategy string
+
+  // Additional information for external conversion if strategy is set to external
+  // +optional
+  Webhook *CustomResourceConversionWebhook
+}
+
+type CustomResourceConversionWebhook {
+  // ClientConfig defines how to communicate with the webhook. This is the same config used for validating/mutating webhooks.
+  ClientConfig WebhookClientConfig
+}
+```
+
+### Defaulting
+
+In case that there is no versions list, a single version with values defaulted to top level version will be created. That means a single version with a name set to spec.version.
+All newly added per version fields (schema, additionalPrinterColumns or subresources) will be defaulted to the coresponding top level field except for the first version in the list that will remain empty.
+
+
+### Validation
+
+To keep backward compatibility, the top level fields (schema, additionalPrinterColumns or subresources) stay the same and source of truth for first (top) version. The first item in the versions list must not set any of those fields. The plan is to use unified version list for v1.
+
+
+### Support Level
+
+The feature will be alpha in the first implementation and will have a feature gate that is defaulted to false. The roll-back story with a feature gate is much more clear. if we have the features as alpha in kubernetes release Y (>X where the feature is missing) and we make it beta in kubernetes release Z, it is not safe to use the feature and downgrade from Y to X but the feature is alpha in Y which is fine. It is safe to downgrade from Z to Y (given that we enable the feature gate in Y) and that is desirable as the feature is beta in Z.
+
+
+### Rollback
+
+Users that need to rollback to version X (but may currently be running version Y > X) of apiserver should not use CRD Webhook Conversion if X is not a version that supports these features.  If a user were to create a CRD that uses CRD Webhook Conversion and then rolls back to version X that does not support conversion then the following would happen:
+
+1. The stored custom resources in etcd will not be deleted.
+
+2. Any clients that try to get the custom resources will get a 500 (internal server error). this is distinguishable from a deleted object for get and the list operation will also fail. That means the CRD is not served at all and Clients that try to garbage collect related resources to missing CRs should be aware of this. 
+
+3. Any client (e.g. controller) that tries to list the resource (in preparation for watching it) will get a 500 (this is distinguishable from an empty list or a 404).
+
+4. If the user rolls forward again, then custom resources will be served again.
+
+If a user does not use the webhook feature but uses the versioned schema, additionalPrinterColumns, and/or subresources and rollback to a version that does not support them per version, any value set per version will be ignored and only values in top level spec.* will be honor.
+
+Please note that any of the fields added in this design that is not supported in previous kubernetes releases can be removed on an update operation (e.g. status update). The kubernetes release where defined the types but gate them with an alpha feature gate, however, can keep these fields but ignore there value.
+
+### Webhook Request/Response
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+For operations that need more than one conversion (e.g. LIST), no partial result will be returned. Instead the whole operation will fail the same way with detailed error messages. To help debugging these kind of operations, the UID of the first failing conversion will also be included in the error message. 
+
+
+## Caching
+
+No new caching is planned as part of this work, but the API Server may in the future cache webhook POST responses.
+
+Most API operations are reads.  The most common kind of read is a watch.  All watched objects are cached in memory.  Each GVK is cached, regardless of whether some GVKs share storage.  Thus, watches do not not cause conversion.  So, conversion webhooks will not add overhead to the watch path.  Watch cache is per api server and eventually consistent.
+
+Non-watch reads are also cached (if requested resourceVersion is 0 which is true for generated informers by default, but not for calls like `kubectl get ...`, namespace cleanup, etc). The cached objects are converted and per version (TODO: fact check). So, conversion webhooks will not add overhead here too.
```
mbohlool: original comment by @liggitt  need to be addressed?

> it does still add overhead, but I think the watch cache makes it scale with the number of versions * number of apiservers, not by the number of overall watchers

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+When a conversion webhook fails, e.g. for the GET operation, then the error message from the apiserver to its client should reflect that conversion failed and include additional information to help debug the problem. The error message and HTTP error code returned by the webhook should be included in the error message API server returns to the user.  For example:
+
+```bash
+$ kubectl get mykind somename
+error on server: conversion from stored version v1 to requested version v2 for somename: "408 request timeout" while calling service "mywebhookhost.somens.cluster.local:443"
+```
+
+
+For operations that need more than one conversion (e.g. LIST), no partial result will be returned. Instead the whole operation will fail the same way with detailed error messages. To help debugging these kind of operations, the UID of the first failing conversion will also be included in the error message.
```
mbohlool: original comments need to be addressed:

by @enisoc:
```
Is this what currently happens if one of the compiled-in conversion functions returns an error?

Also, this seems to imply that conversion hooks must not have stricter expectations than validation. Otherwise, one "valid" but unconvertible object could cause denial of service to many other objects.

Do we ever check that an object in a CR is valid, other than during the admission chain? That is, do we validate objects that we read from storage?
```
by @erictune:
```
What does List do if one object fails conversion? If it return a partial list? If so, I argue that is unexpected and probably going to cause bugs. If not, then we have the same DoS problem.
```

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+  Strategy string
+
+  // Additional information for external conversion if strategy is set to external
+  // +optional
+  Webhook *CustomResourceConversionWebhook
+}
+
+type CustomResourceConversionWebhook {
+  // ClientConfig defines how to communicate with the webhook. This is the same config used for validating/mutating webhooks.
+  ClientConfig WebhookClientConfig
```
mbohlool: original comment by @deads2k asking advice from @liggitt:
> I can't think of a reason for a conversion webhook to require authentication to keep the information a secret, but it may be useful for managing flow control concerns. @liggitt  thoughts? If we need to do it, we should build it in now.

mbohlool:
@liggitt @deads2k @erictune @sttts I moved all outstanding comments from the doc to this PR. Please let me know if I missed anything. I assume the one I (or somebody else) answered are resolved.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
+# CRD Conversion Webhook Alpha
+
+
+## Public
```
sttts: There is a template here: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/Design_Proposal_TEMPLATE.md

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+type CustomResourceConversionWebhook {
+  // ClientConfig defines how to communicate with the webhook. This is the same config used for validating/mutating webhooks.
+  ClientConfig WebhookClientConfig
+}
+```
+
+## Defaulting
+
+In case that there is no versions list, a single version with values defaulted to top level version will be created. That means a single version with a name set to spec.version.
```
sttts: The global values (schema, subresource, columns) only apply to the first version, right? Don't we need some logic to transition CRDs with NoConversion by defaulting all but the first versions values to the old global values?

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+
+## Validation
+
+To keep backward compatibility, the top level fields (schema, additionalPrinterColumns or subresources) are stay the same and source of truth for first (top) version. The first item in the versions list should not set any of those fields. The plan is to use unified version list for v1.
```
sttts: -are

sttts: s/should/must/

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+
+## Support Level
+
+The feature will be alpha in the first supported version and will have a feature gate that is defaulted to false. The roll-back story with a feature gate is much more clear. if we have the features as alpha in version Y (>X where the feature is missing) and we make it beta in version Z, it is not safe to use the feature and downgrade from Y to X but the feature is alpha in Y which is fine. It is safe to downgrade from Z to Y (given that we enable the feature gate in Y) and that is desirable as the feature is beta in Z. I will add this detail to the doc.
```
sttts: s/in the first supported version/in the first implementation/

sttts: better write "kubernetes release" instead of version. Here are so many versions in the doc with different meaning.

sttts: s/I will add this detail to the doc./?/

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+
+## Rollback
+
+Users that need to rollback to version X (but may currently be running version Y > X) of apiserver should not use CRD Webhook Conversion until X is a version that supports these features.  If a user were to create a CRD that uses CRD Webhook Conversion and then rolls back then the following would happen:
+
+1. The stored custom resources in etcd will not be deleted.
+
+2. Any clients that try to get the custom resources will get a 500 (internal server error). this is distinguishable from a deleted object for get and the list operation will also fail. That means the CRD is not served at all and Clients that try to garbage collect related resources to missing CRs should be aware of this.
```
sttts: Why does this happen? X must at least support webhook versioning as alpha. Then we can throw 500 errors if the gate is disabled.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
sttts: Here is the code: https://github.com/kubernetes/kubernetes/blob/dbb448bbdcb9e440eee57024ffa5f1698956a054/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L355

The logic:
- if resourceVersion is empty, we don't cache
- if requested resourceVesion is old than the cached version (especially for `resourceVersion=0`), serve the cached object.

The cache is in the storage layer using internal versions. I.e. it is not per version. So we webhooks add overhead.

Looking at the watcher code in the apiserver http handler, I am not sure at all that we have a watch cache per version. I see versioning encoders there for the object in the watch events. @liggitt have you looked into that in more detail? My impression is that we cache internal objects and convert on each outgoing stream. I hope I am wrong.

/cc @wojtek-t can you comment our current caching logic?

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+
+And create it:
+
+```bash
+Kubectl create -f cr2.yaml
+```
+
+
+**Step 6**: storage now has two custom resources in two different versions. To downgrade to previous CRD, one can apply crd1.yaml but that will fail as the status.storedVersions has both v1 and v2 and those cannot be removed from the spec.versions list. To downgrade, first create a crd2-b.yaml file that sets v1 as storage version and apply it, then follow "*Upgrade existing objects to a new stored version*“ in [this document](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/). After all CRs in the storage has v1 version, you can apply crd1.yaml.
```
sttts: Why does the storage have two versions? I see v1 is still the storage version. We only have v1 in etcd.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: I tried to follow the template with some changes to this doc

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: added

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: done

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: done (last part was the copy/paste monkey's mistake :) )

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+
+## Support Level
+
+The feature will be alpha in the first supported version and will have a feature gate that is defaulted to false. The roll-back story with a feature gate is much more clear. if we have the features as alpha in version Y (>X where the feature is missing) and we make it beta in version Z, it is not safe to use the feature and downgrade from Y to X but the feature is alpha in Y which is fine. It is safe to downgrade from Z to Y (given that we enable the feature gate in Y) and that is desirable as the feature is beta in Z. I will add this detail to the doc.
+
+
+## Rollback
+
+Users that need to rollback to version X (but may currently be running version Y > X) of apiserver should not use CRD Webhook Conversion until X is a version that supports these features.  If a user were to create a CRD that uses CRD Webhook Conversion and then rolls back then the following would happen:
```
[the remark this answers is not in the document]
mbohlool: edited

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: you mean CRD versioning with alpha webhook conversion? The assumption of this section is version X does not support webhook conversion and we do have a webhook conversion. No support for webhook conversion means the feature gate is off.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,861 @@
...
+
+1. The stored custom resources in etcd will not be deleted.
+
+2. Any clients that try to get the custom resources will get a 500 (internal server error). this is distinguishable from a deleted object for get and the list operation will also fail. That means the CRD is not served at all and Clients that try to garbage collect related resources to missing CRs should be aware of this. 
+
+3. Any client (e.g. controller) that tries to list the resource (in preparation for watching it) will get a 500 (this is distinguishable from an empty list or a 404).
+
+4. If the user rolls forward again, then custom resources will be served again.
+
+If a user does not use the webhook feature but uses the versioned schema, additionalPrinterColumns, and/or subresources and rollback to a version that does not support them per version, any value set per version will be ignored and only values in top level spec.* will be honor.
```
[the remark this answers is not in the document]
mbohlool: If X is a version that supports conversion in alpha, we can still keep these values but ignoring them (we do have them in types.go) but versions before that, we will lose them on status or any update.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: > The cache is in the storage layer using internal versions.

We do not store internal versions, I assume it is one of the external versions in the storage layer?

sttts: Ok, the "one storage per version" saves us for now. But this important and special: we do not have a hub version, we always convert directly to the target version. Until now this was an arbitrary design choice, i.e. we could have done a NoConversion on decode and the real conversion on encode. With the caching arguments here, we turn this arbitrary choice into a required design. Can you add a remake about this?

deads2k:
With the understand that the API will still go through an API review, this design looks good to me.

@kubernetes/sig-api-machinery-feature-requests @kubernetes/sig-api-machinery-api-reviews @kubernetes/sig-api-machinery-misc I plan to announce last call at the sig meeting and I anticipate this merging towards the end of this week.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: done

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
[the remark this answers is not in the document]
mbohlool: I assume that is an implementation detail for now and we can discuss on the implementation PR. Will update this doc when we had a decision on that.

mbohlool:
I stashed commits and address final comment from @sttts where was only adding reference in the text to the fact that we use one RESTStore object per version thus the watch cache is per version. I will reapply @sttts LGTM label but this still needs approval from @deads2k.

mbohlool:
@deads2k look like I can't apply labels here. Can you please approve and also lgtm (based on sttts's lgtm) this.

deads2k:
> With the understand that the API will still go through an API review, this design looks good to me.
> 
> @kubernetes/sig-api-machinery-feature-requests @kubernetes/sig-api-machinery-api-reviews @kubernetes/sig-api-machinery-misc I plan to announce last call at the sig meeting and I anticipate this merging towards the end of this week.

I haven't seen anything blocking come in.

/lgtm

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,859 @@
...
+
+### Defaulting
+
+In case that there is no versions list, a single version with values defaulted to top level version will be created. That means a single version with a name set to spec.version.
+All newly added per version fields (schema, additionalPrinterColumns or subresources) will be defaulted to the coresponding top level field except for the first version in the list that will remain empty.
+
+
+### Validation
+
+To keep backward compatibility, the top level fields (schema, additionalPrinterColumns or subresources) stay the same and source of truth for first (top) version. The first item in the versions list must not set any of those fields. The plan is to use unified version list for v1.
```
lavalamp: I don't think this is 100% necessary but it is probably expedient.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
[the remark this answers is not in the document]
lavalamp: Remove them if the feature is off.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,859 @@
...
+type ConversionResponse struct {
+  // UID is an identifier for the individual request/response.
+  // This should be copied over from the corresponding ConversionRequest.
+  UID types.UID
+  // ConvertedObject is the converted version of request.Object.
+  ConvertedObject runtime.RawExtension
+}
+```
+
+If the conversion is failed, the webhook should fail the HTTP request with a proper error code and message that will be used to create a status error for the original API caller.
```
lavalamp: Please explicitly define the error return type, e.g. by adding an `Error string` to the ConversionResponse type, or by requiring it to send back a metav1.Status object.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
lavalamp: Please clarify whether the webhook is supposed to accept an entire List or if it will always be called for each individual item-- I expect the latter, but it needs to be explicitly stated.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -0,0 +1,859 @@
...
+type CustomResourceConversionWebhook {
+  // ClientConfig defines how to communicate with the webhook. This is the same config used for validating/mutating webhooks.
+  ClientConfig WebhookClientConfig
+}
+```
+
+### Defaulting
+
+In case that there is no versions list, a single version with values defaulted to top level version will be created. That means a single version with a name set to spec.version.
+All newly added per version fields (schema, additionalPrinterColumns or subresources) will be defaulted to the coresponding top level field except for the first version in the list that will remain empty.
```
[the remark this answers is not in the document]
mbohlool: Because of backward compatibility explained down there. We cannot duplicate data and we cannot remove top level fields so we do not use first version's fields. This will be cleaned-up in v1.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: If we change the meaning of top level fields it is backward compatible, if we duplicate data of top level fields into first version, then an update with an old client will make inconsistent object (first version fields may have different values than top level). What is the case you thinking about that this is not necessary and it is still backward compatible?

PR #2266, opened 2018-06-14
sig-instrumentation: Add charter document

brancz:
This PR is currently work in progress and meant for sig-instrumentation to iterate on its charter document.

@kubernetes/sig-instrumentation-misc

PR #2423, opened 2018-07-26
commitee-steering: link SIG Charter to SIG Governance

philips:
This was a relative link before but as people copy the charter that
won't work. Make it a full URL instead.


philips:
Identified as an issue by the SIG Instrumentation use of the template: https://github.com/kubernetes/community/pull/2266#discussion_r205524019

PR #2425, opened 2018-07-26
SIG Autoscaling Charter and sigs.yaml updates

DirectXMan12:
This adds a SIG Autoscaling charter, and updates sigs.yaml to make sure it has the right subprojects assigned to sig autoscaling vs sig instrumentation


sig-autoscaling/charter.md
```diff
@@ -0,0 +1,74 @@
...
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+Covers development and maintenance of componets for automated scaling in
+Kubernetes.  This includes automated vertical and horizontal pod
+autoscaling, initial resource estimation, cluster-proportional system
+component autoscaling, and autoscaling of Kubernetes clusters themselves.
```
MaciekPytel: Currently cluster proportional autoscaling is not owned by sig-autoscaling:
* addon-resizer is maintained by sig-instrumentation (AFAIK it's primarily used for heapster and metrics-server)
* kube-dns-autoscaler is maintained by sig-networking

Ideally sig-autoscaling should take ownership of this use-case, but we're not there yet.

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Autoscaling-related API objects, such as the HorizontalPodAutoscaler and
+  VerticalPodAutoscaler
+
+- Autoscaling-related tools, such as the cluster autoscaler and
+  cluster-proportional scaling tools (e.g. pod-nanny)
+
+- Ensuring API interfaces (the scale subresource) are availble and usable
+  to enable other SIG to write autoscalable objects, and enable people to
+  interact with those interfaces.
```
MaciekPytel: I'd argue this goes requires a lot of apimachinery know-how, which is not too closely related to anything else we do in sig. I suspect you're the only one who actually knows how the polymorphic scale client actually works.

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,74 @@
...
+  API objects (e.g. HPA), as well as scaling interfaces (e.g. the
+  polymorphic scale client).
+
+#### Cross-cutting and Externally Facing Processes
+
+- Reviewing implementations of the scale subresource to ensure that
+  autoscaling behaves properly.
+
+- Coordinating with SIG Instrumentation to ensure that metrics APIs are
+  suitable for autoscaling on.
```
MaciekPytel: Can we add coordinating with sig-scheduling to ensure new scheduling features are compatible with cluster autoscaling?

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Out of scope
+
+- Testing performance at scale (this falls under the purview of [SIG
+  Scalability].
```
MaciekPytel: We certainly do it for cluster autoscaler. And I don't think sig-scalability is doing any tests for HPA. Maybe we should reconsider? I would argue some level of scalability should be pre-requisite for HPAv2 GA, which would make it very much our problem.

sig-autoscaling/charter.md
DirectXMan12: oh, do we not own the addon-resizer?  I thought it was under our purview from the pre-sig-instrumentation Heapster days, and I didn't think it got transferred over, since it's an autoscaler.  

We probably *should* own those, at any rate. @kubernetes/sig-instrumentation-misc

sig-autoscaling/charter.md
DirectXMan12: SIG scalability and SIG autoscaling are often confused in issue requests, and I kind-of wanted to make a clear deliniation.  It doesn't mean we can't test our *own* components, but we're not responsible for general cluster scale testing (e.g. if you can't launch 100000 pods at once, it's not our job to look into why).  I'll try to clarify.

sig-autoscaling/charter.md
brancz: > addon-resizer is maintained by sig-instrumentation (AFAIK it's primarily used for heapster and metrics-server)

I was not aware of this, and it's also not in our sig components. I was under the impression that sig-autoscaling own it.

MaciekPytel: I mean no-one is really doing much work on addon-resizer, but I believe @kawych and @x13n are the ones maintaining it.

sig-autoscaling/charter.md
MaciekPytel: Ah, gotcha, I completely misunderstood what you meant.

sig-autoscaling/charter.md
DirectXMan12: Every kubernetes piece of code needs to be owned by a SIG.  I think it makes sense for us to own it, especially since it doesn't fall under the responsibilities of SIG instrumentation normally.  @kawych and @x13n can continue to be the ones to work on it, and we can list them as the owners of the subproject, but it'll fall under the purview of sig autoscaling.

brancz: @DirectXMan12’s suggestion sounds good to me.

DirectXMan12: @kubernetes/sig-network-misc the cluster-proportional-autoscaler doesn't seem to be listed under `sigs.yaml` (unless my searching is failing me).  I think it makes some amount of sense for it to be a subproject of sig-autoscaling (no change in ownership necessary, but it should fall under a sig umbrella), so I'd appreciate your opinions.

sig-autoscaling/charter.md
DirectXMan12: ack

sig-autoscaling/charter.md
DirectXMan12: Sure, but we're responsible for maintaining the HPA, and the HPA can't work w/o the scale subresource, and the scale subresource's object is part of the autoscaling API group.

@kubernetes/sig-api-machinery-misc any opinions?

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,78 @@
+# SIG Autoscaling Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+Covers development and maintenance of componets for automated scaling in
```
MrHohn: nit: componets -> components

sig-autoscaling/charter.md
MrHohn: @DirectXMan12 what you suggested sounds great. Also wanted to point out yet another component that might fall under this same umbrella: [cluster-proportional-vertical-autoscaler](https://github.com/kubernetes-incubator/cluster-proportional-vertical-autoscaler).

DirectXMan12: cool, thanks for the heads up.  I'll update sigs.yaml with all of those

sig-autoscaling/charter.md
liggitt: In general, api-machinery owns mechanisms, specific sigs own specific APIs.

In this case, I'd expect api-machinery to own the apiserver code that enables subresources, and autoscaling to own the in-tree scale subresource implementations, the autoscaling/v1 Scale API types, and custom code in the scale client.

DirectXMan12: ack, that's more-or-less what I expected

DirectXMan12:
See #2534

sigs.yaml
```diff
@@ -345,12 +346,28 @@ sigs:
       - name: sig-autoscaling-test-failures
         description: Test Failures and Triage
     subprojects:
-    - name: autoscaler
+    - name: scale-client
+      owners:
+      - https://raw.githubusercontent.com/kubernetes/client-go/master/scale/OWNERS # doesn't exist yet
```
lavalamp: api machinery owns the level above this. Is it OK for OWNERS files in this list to overlap? I assume it must be. I don't have a strong opinion about who should own this code at the moment. If we had a client-focussed SIG, they would own it, I think, but we don't.

sigs.yaml
```diff
@@ -345,12 +346,28 @@ sigs:
...
+    - name: cluster-autoscaler
       owners:
       - https://raw.githubusercontent.com/kubernetes/autoscaler/master/OWNERS
-    - name: metrics
+    - name: vertical-pod-autoscaler
       owners:
-      - https://raw.githubusercontent.com/kubernetes/metrics/master/OWNERS
+      - https://raw.githubusercontent.com/kubernetes/autoscaler/master/OWNERS
+    - name: horizntal-pod-autoscaler
```
lavalamp: spelling

spiffxp:
/committee steering
/assign @brendanburns @michelleN 
per https://github.com/kubernetes/steering/issues/31

spiffxp:
ping @DirectXMan12 @mwielgus
needs rebase

DirectXMan12:
@spiffxp @brendanburns rebased.  What else has to be done?

DirectXMan12:
This should fix the verify issues.

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Coordinating with SIG Scheduling to make sure scheduling decisions can
+  interact well with the cluster autoscaler
+
+### Out of scope
+
+- Testing general cluster performance at scale (this falls under the
+  purview of [SIG Scalability]).
+
+- Owning metrics APIs (this falls under the purview of [SIG
```
spiffxp: missing close paren

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Autoscaling-related API objects, such as the HorizontalPodAutoscaler and
+  VerticalPodAutoscaler
+
+- Autoscaling-related tools, such as the cluster autoscaler,
+  single-component scaling tools (e.g. pod-nanny), and
+  cluster-proportional scaling tools
+
+- Ensuring API interfaces (the scale subresource) are available and usable
+  to enable other SIG to write autoscalable objects, and enable people to
```
spiffxp: nit: SIGs

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,78 @@
+# SIG Autoscaling Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+Covers development and maintenance of components for automated scaling in
```
brendandburns: I think that there should be something about "integrated into kubernetes by default" or something like that.

Basically, I want it to be clear that if someone develops a random autoscaler in their own repo, it's not covered by this SIG.  I think that's pretty clearly understood by everyone involved, but this text seems to lay claim to all Kubernetes autoscaling

(and the thread below that points to an incubator project as part of the scope makes this distinction even more important to make)

sig-autoscaling/charter.md
brendandburns: I don't think this SIG owns an incubator project, unless it was created by SIG members who want it to be within the scope of the SIG...

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Ensuring API interfaces (the scale subresource) are available and usable
+  to enable other SIG to write autoscalable objects, and enable people to
+  interact with those interfaces.
+
+[Link to SIG section in sigs.yaml][sigs.yaml]
+
+#### Code, Binaries and Services
+
+- Components and utilities that take automated action to scale a component
```
brendandburns: These are all kind of vague, pointers to actual repos/folders?

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,78 @@
...
+  on the cluster
+
+- Components and utilities that take automated action to scale the cluster
+  itself
+
+- Special parts of client-go for interacting with with the scaling
```
brendandburns: Definitely need a pointer into the project here.

sig-autoscaling/charter.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Reviewing implementations of the scale subresource to ensure that
+  autoscaling behaves properly
+
+- Coordinating with SIG Instrumentation to ensure that metrics APIs are
+  suitable for autoscaling on.
+
+- Coordinating with SIG Scheduling to make sure scheduling decisions can
+  interact well with the cluster autoscaler
+
```
brendandburns: Possibly something about cluster lifecycle wrt to cluster autoscale?

brendandburns: also SIG node for vertical scaling needs in the kubelet?

sig-autoscaling/charter.md
DirectXMan12: We also own officially "sponsored" projects.  The term "integrated in to Kubernetes by default" is pretty ambiguous, especially in the era of "do it out-of-tree using a CRD" (i.e. vertical pod autoscaler and cluster autoscaler).  If someone wants to get their project sponsored by a SIG, SIG Autoscaling is the right place, yet it may not be "integrated into Kubernetes by default".

sig-autoscaling/charter.md
DirectXMan12: I was under the impression that concrete folders is what sigs.yaml was supposed to be for, and that this was more of "conceptually, what do you cover".  That being said, I can certainly provide some examples here, or actually links if my impression was incorrection.

sig-autoscaling/charter.md
DirectXMan12: yep, those are both correct.  I'll add them.

sig-autoscaling/charter.md
DirectXMan12: done -- mentioned the specific subproject.  Actual folder definitions are in sigs.yaml

sig-autoscaling/charter.md
DirectXMan12: fixed to mention the actual subproject names

DirectXMan12:
@spiffxp typos fixed
@brendanburns fixed the comments re: specific links by mentioning specific subprojects that can be crossreferenced with sigs.yaml to find folders.  Adding specific folders in the charter seems like a recipe for getting out of date when we already have sigs.yaml.

@spiffxp I don't see any comments from @mwielgus ... what do you mean?

spiffxp:
@DirectXMan12 wow yeah where did I get that name from, I meant @MaciekPytel, I _think_ their comments are satisfied? just looking for a confirm

DirectXMan12:
looks like @MaciekPytel has responded to most of the threads that they raised confirming that, but just in case ping @MaciekPytel to confirm.

MaciekPytel:
I'm happy with how my comments were addressed, lgtm.

DirectXMan12:
@spiffxp @brendanburns good to go then?

spiffxp:
as do we, I poked @brendandburns and @michelleN both during steering meeting today, if you're still hanging tomorrow I can serve as another SC reviewer lgtm

spiffxp:
/lgtm
Here's your proxy LGTM from @brendandburns 
/approve
Here's a second SC reviewer saying LGTM

Thanks for working through this!

PR #2426, opened 2018-07-26
Advocate for running conformance tests in multiple architectures

dims:
We have been talking about and progressing towards multi-arch support for a long while now: https://github.com/kubernetes/kubernetes/issues/38067

We have active work being done to ensure that container images used in e2e test suite can run across multiple architectures. So we should make sure we don't add to this debt.

- https://github.com/kubernetes/kubernetes/issues/66626
- https://github.com/kubernetes/kubernetes/issues/66618

PR #2428, opened 2018-07-27
Advocate for running conformance tests in multiple architectures

dims:
We have been talking about and progressing towards multi-arch support for a long while now: https://github.com/kubernetes/kubernetes/issues/38067

We have active work being done to ensure that container images used in e2e test suite can run across multiple architectures. So we should make sure we don't add to this debt.

- https://github.com/kubernetes/kubernetes/issues/66626
- https://github.com/kubernetes/kubernetes/issues/66618

### DIFF
--- a/contributors/devel/conformance-tests.md
+++ b/contributors/devel/conformance-tests.md
@@ -34,6 +34,7 @@ especially in the Nucleus or Application layers as described
 (example: the default list of admission plugins should not have to be tweaked for passing conformance).
 - cannot rely on any binaries that are not required for the
 linux kernel or for a kubelet to run (i.e. git)
+- any container images used in the test must support all architectures for which kubernetes releases are built
 
 ### Conformance Test Version Skew Policy
 


dims:
apologies for screwing up the other PR https://github.com/kubernetes/community/pull/2426

dims:
cc @jdumars @cblecker @liggitt

bgrant0607:
@dims 

Which other architectures? ARM?

Any other architecture that's officially supported needs a multi-person team dedicated towards running and monitoring continuous testing and keeping it working, and someone needs to provide the infrastructure. It would be another provider + OS + architecture. Until there is a track record of stability (well, comparable to existing providers), I wouldn't block releases on it.

dims:
@bgrant0607 The section is about `A test is eligible for promotion to conformance`. All i am saying here is that such tests should use multi-arch images (which we already have support for in test/images makefiles etc). 

No one is talking about blocking releases. Just the criteria when things get added so we have a easier time when we will be able to do run e2e tests fully on other architectures.

dims:
Also, the manifest.go has already been updated already with multi-arch images:
https://github.com/kubernetes/kubernetes/blob/master/test/utils/image/manifest.go#L51-L93

One remaining PR is for nginx which @mkumatag and i are working on:
https://github.com/kubernetes/kubernetes/pull/66055

Once that PR is in, we can run e2e tests (on other arches) and fix whatever is broken.

cc @vielmetti for arm*
cc @sudeeshjohn and @mkumatag for ppc64el

bgrant0607:
@dims Ok, but don't we need to specify which architectures, even if just referring to the list in https://github.com/kubernetes/community/blob/master/contributors/design-proposals/multi-platform.md? Not all of these images are built by our build system and/or as part of our release.

dims:
@bgrant0607 yes, i deliberately left that out so we can see which arch(es) end up having full CI jobs and environments to run them in and people to look after exactly like you said.

At this point, we just want to be sure that we are using the `fat manifest` and when folks add images to say test/images they follow the pattern already established there.

In this phase, we are just getting the scaffolding ready and not trying to pick which arch to support in an official release. We have gotten to the point where all the kubernetes component images are multi-arch, all the images in the e2e tests are multi-arch (except the nginx one in progress), So we have gotten very far thanks to some solid work by folks like @mkumatag 

Other related discussion:
- https://github.com/kubernetes/kubeadm/issues/1030
- https://github.com/kubernetes/kubernetes/issues/38067
- https://github.com/kubernetes/kubernetes/issues/25730

neolit123:
+1
everything is going for fat images and the conf. tests should follow.

contributors/devel/conformance-tests.md
```diff
@@ -34,6 +34,7 @@ especially in the Nucleus or Application layers as described
 (example: the default list of admission plugins should not have to be tweaked for passing conformance).
 - cannot rely on any binaries that are not required for the
 linux kernel or for a kubelet to run (i.e. git)
+- container images used in the test should support multiple architectures (should not be limited to just amd64)
```
spiffxp: I am 100% fine with the spirit of this.  If Brian wants to tighten up the phrasing, maybe 
"any container images used in the test must support all architectures for which kubernetes releases are built"

bgrant0607: I like Aaron's phrasing. If it's changed to that, then LGTM

dims: On it!

spiffxp:
/approve
/lgtm
given that @timothysc and @bgrant0607 have given their LGTM's above

PR #2000, opened 2018-04-04
[RFC] SIG-Auth initial charter

tallclair:
Co-authored by: @liggitt @ericchiang @tallclair

Adapted from [sig-governance-template-short](https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance-template-short.md) and [sig-governance-template-long](https://github.com/kubernetes/community/pull/1650).

/sig auth

PR #2431, opened 2018-07-28
sig-auth charter (v2)

tallclair:
Add a sig-auth charter, based on the [sig-charter-template](https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-charter-template.md).

The first attempt at our charter is in https://github.com/kubernetes/community/pull/2000.

Co-authored by: @liggitt @mikedanese @tallclair

/sig auth

### DIFF
--- a/sig-auth/charter.md
+++ b/sig-auth/charter.md
@@ -7,8 +7,8 @@ the Roles and Organization Management outlined in [sig-governance].
 
 SIG Auth is responsible for the design, implementation, and maintenance of features in
 Kubernetes that control and protect access to the API and other core components. This includes
-authentication and authorization, but also encompasses features like auditing and some policy
-(see below).
+authentication and authorization, but also encompasses features like auditing and some security
+policy (see below).
 
 ### In scope
 
@@ -16,7 +16,8 @@ Link to SIG section in [sigs.yaml]
 
 #### Code, Binaries and Services
 
-- Kubernetes authentication, authorization, audit and policy features. Examples include:
+- Kubernetes authentication, authorization, audit and security policy features. Examples
+  include:
     - Authentication, authorization and audit interfaces and extension points
     - Authentication implementations (service accounts, OIDC, authenticating proxy, webhook,
       ...)
--- a/sig-auth/charter.md
+++ b/sig-auth/charter.md
@@ -1,0 +1,66 @@
+# SIG YOURSIG Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+Include a 2-3 sentence summary of what work SIG TODO does. Imagine trying to
+explain your work to a colleague who is familiar with Kubernetes but not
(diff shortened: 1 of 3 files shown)


sig-auth/charter.md
```diff
@@ -0,0 +1,68 @@
...
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+SIG Auth is responsible for the design, implementation, and maintenance of features in
+Kubernetes that control and protect access to the API and other core components. This includes
+authentication and authorization, but also encompasses features like auditing and some policy
+(see below).
```
tallclair: @liggitt @mikedanese - This is the only piece that isn't copied from the scope we discussed. PTAL

sig-auth/charter.md
```diff
@@ -0,0 +1,68 @@
...
+
+### In scope
+
+Link to SIG section in [sigs.yaml]
+
+#### Code, Binaries and Services
+
+- Kubernetes authentication, authorization, audit and policy features. Examples include:
```
jbeda: s/policy/security policy/? How does this overlap with other policy aspects?  Or do you think that it doesn't?

sig-auth/charter.md
```diff
@@ -0,0 +1,68 @@
...
+  - Bug bounty (prefer: product security team)
+  - Resource quota (prefer: sig-scheduling)
+  - Resource availability / DOS protection (prefer: sig-apimachinery, sig-network, sig-node)
+
+## Roles and Organization Management
+
+This sig follows adheres to the Roles and Organization Management outlined in [sig-governance]
+and opts-in to updates and modifications to [sig-governance].
+
+### Subproject Creation
```
jbeda: Can be done in a different PR, but there are no subprojects defined for SIG-auth defined in sigs.yaml.  How are y'all thinking about breaking this down into smaller chunks?

sig-auth/charter.md
tallclair: Done. We actually were going to change that to security policy, but I guess that change got lost.

I agree this is a fuzzy area, and the policy scope of sig-auth isn't super well defined. I think that is partly because most policy features are cross-cutting. For example, PodSecruityPolicy should really be co-owned by sig-auth & sig-node. [SchedulingPolicy](https://github.com/kubernetes/community/pull/1937) should be sig-auth & sig-scheduling. Network policy is probably just sig-networking, since it's only a runtime policy (comparable to PodSecurityContext).

I see our (sig-auth) role around policy as:
- helping to drive consistency across policies (really co-owned with wg-policy)
- owning the policy primitives & mechanisms (co-owned with api-machinery)
- ensuring that important use cases are covered by the existing authorization & policy mechanisms

Does this make sense? I can try to distill this into our scope if we all agree.

sig-auth/charter.md
spiffxp: opened https://github.com/kubernetes/community/issues/2505

PR #2439, opened 2018-07-31
CSI: send pod information in NodePublishVolumeRequest

jsafrane:
This proposal adds possibility to send pod information in CSI `NodePublishVolumeRequest` call.

Fixes https://github.com/kubernetes/kubernetes/issues/64984

It depends on https://github.com/kubernetes/community/pull/2514

/sig storage

### DIFF
--- a/contributors/design-proposals/storage/container-storage-interface-pod-information.md
+++ b/contributors/design-proposals/storage/container-storage-interface-pod-information.md
@@ -1,0 +1,48 @@
+# Pod in CSI NodePublish request
+Author: @jsafrane
+
+## Goal
+* Pass Pod information (pod name/namespace/UID + service account) to CSI drivers in `NodePublish` request as CSI volume attributes.
+
+## Motivation
+We'd like to move away from exec based Flex to gRPC based CSI volumes. In Flex, kubelet always passes `pod.namespace`, `pod.name`, `pod.uid` and `pod.spec.serviceAccountName` ("pod information") in every `mount` call. In Kubernetes community we've seen some Flex drivers that use pod or service account information to authorize or audit usage of a volume or generate content of the volume tailored to the pod (e.g. https://github.com/Azure/kubernetes-keyvault-flexvol).
+
+CSI is agnostic to container orchestrators (such as Kubernetes, Mesos or CloudFoundry) and as such does not understand  concept of pods and service accounts. [Enhancement of CSI protocol](https://github.com/container-storage-interface/spec/pull/252) to pass "workload" (~pod) information from Kubernetes to CSI driver has met some resistance. 
+
+## High-level design
+We decided to pass the pod information as `NodePublishVolumeRequest.volume_attributes`.
+
+* Kubernetes passes pod information only to CSI drivers that explicitly require that information in their [`CSIDriver` instance](https://github.com/kubernetes/community/pull/2523). These drivers are tightly coupled to Kubernetes and may not work or may require reconfiguration on other cloud orchestrators. It is expected (but not limited to) that these drivers will provide ephemeral volumes similar to Secrets or ConfigMap, extending Kubernetes secret or configuration sources.
+* Kubernetes will not pass pod information to CSI drivers that don't know or don't care about pods and service accounts. It is expected (but not limited to) that these drivers will provide real persistent storage. Such CSI driver would reject a CSI call with pod information as invalid. This is current behavior of Kubernetes and it will be the default behavior.
+
+## Detailed design
+
+### API changes
+No API changes.
+
+### CSI enhancement
+We don't need to change CSI protocol in any way. It allows kubelet to pass `pod.name`, `pod.uid` and `pod.spec.serviceAccountName` in [`NodePublish` call as `volume_attributes`]((https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume)). `NodePublish` is roughly equivalent to Flex `mount` call.
+
+The only thing we need to do is to **define** names of the `volume_attributes` keys that CSI drivers can expect:
+	*	`csi.storage.k8s.io/pod.name`: name of the pod that wants the volume.
+	*	`csi.storage.k8s.io/pod.namespace`: namespace of the pod that wants the volume.
+	*	`csi.storage.k8s.io/pod.uid`: uid of the pod that wants the volume.
+	*	`csi.storage.k8s.io/serviceAccount.name`: name of the service account under which the pod operates. Namespace of the service account is the same as `pod.namespace`.
+
+Note that these attribute names are very similar to [parameters we pass to flex volume plugin](https://github.com/kubernetes/kubernetes/blob/10688257e63e4d778c499ba30cddbc8c6219abe9/pkg/volume/flexvolume/driver-call.go#L55).
+
+### Kubelet
+Kubelet needs to create informer to cache `CSIDriver` instances. It passes the informer to CSI volume plugin as a new argument of [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/43f805b7bdda7a5b491d34611f85c249a63d7f97/pkg/volume/csi/csi_plugin.go#L58).
+
+### CSI volume plugin
+In `SetUpAt()`, the CSI volume plugin checks the `CSIDriver` informer if `CSIDriver` instance exists for a particular CSI driver that handles the volume. If the instance exists and has `PodInfoRequiredOnMount` set, the volume plugin adds `csi.storage.k8s.io/*` attributes to `volume_attributes` of the CSI volume. It blindly overwrites any existing values there.
+
+Kubelet and the volume plugin must tolerate when CRD for `CSIDriver` is not created (yet). Kubelet and CSI volume plugin falls back to original behavior, i.e. does not pass any pod information to CSI. We expect that CSI drivers will return reasonable error code instead of mounting a wrong volume. 
+
+TODO(jsafrane): check what (shared?) informer does when it's created for non-existing CRD. Will it start working automatically when the CRD is created? Or shall we retry creation of the informer every X seconds until the CRD is created? Alternatively, we may GEt fresh `CSIDriver` from API server in `SetUpAt()`, without any informer.
+
+## Implementation
+
+* Alpha in 1.12 (behind `CSIPodInfo` feature gate)
+* Beta in 1.13 (behind `CSIPodInfo` feature gate)
+* GA 1.14?


contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,123 @@
...
+## Detailed design
+
+### CSI enhancement
+We don't need to change CSI protocol in any way. It allows kubelet to pass `pod.name`, `pod.uid` and `pod.spec.serviceAccountName` in [`NodePublish` call as `volume_attributes`]((https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume)). `NodePublish` is roughly equivalent to Flex `mount` call.
+
+The only thing we need to do is to **define** names of the `volume_attributes` keys that CSI drivers can expect:
+	*	`csi.kubernetes.io/pod.name`: name of the pod that wants the volume.
+	*	`csi.kubernetes.io/pod.namespace`: namespace of the pod that wants the volume.
+	*	`csi.kubernetes.io/pod.uid`: uid of the pod that wants the volume.
+	*	`csi.kubernetes.io/serviceAccount.name`: name of the service account under which the pod operates. Namespace is the same as `pod.namespace`.
```
tamalsaha: Should be the annotation names be like `x.y.z/w1-w2-w3` ?

Example: https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L98

contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,123 @@
...
+## High-level design
+We decided to pass the pod information as `NodePublishVolumeRequest.volume_attributes`.
+
+* Kubernetes passes pod information only to CSI drivers that explicitly require that information. These drivers are tightly coupled to Kubernetes and may not work or may require reconfiguration on other cloud orchestrators. It is expected (but not limited to) that these drivers will provider ephemeral volumes similar to Secrets or ConfigMap, extending Kubernetes secret or configuration sources.
+* Kubernetes will not pass pod information to CSI drivers that don't know or don't care about pods and service accounts. It is expected (but not limited to) that these drivers will provide real persistent storage. Such CSI driver would reject a CSI call with pod information as invalid.
+
+## Detailed design
+
+### CSI enhancement
+We don't need to change CSI protocol in any way. It allows kubelet to pass `pod.name`, `pod.uid` and `pod.spec.serviceAccountName` in [`NodePublish` call as `volume_attributes`]((https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume)). `NodePublish` is roughly equivalent to Flex `mount` call.
```
kfox1111: The NodeServer has two sets of calls that I think are applicable. NodePublishVolume/Unpublish, but also NodeStageVolume/Unstage https://github.com/container-storage-interface/spec/blob/master/spec.md#nodestagevolume. Those calls also need the additional info to function properly.

contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,123 @@
...
+3. On a node, CSI Driver Registrar places UNIX domain socket to `/var/lib/kubelet/plugins/foo-reg.sock`.
+4. Kubelet finds the socket there and calls `GetInfo` on it (no change from current behavior).
+5. CSI Driver Registrar receives `GetInfo` call and passes CSI driver socket to kubelet (no change from current behavior).
+6. Kubelet notifies CSI volume plugin that there is a new CSI driver installed and passes **both** CSI DriverRegistrar socket and CSI driver socket (no change from current behavior).
+7. CSI volume plugin calls `GetCSIDriverInfo` on the CSI Driver Registrar socket.
+8. CSI Driver Registrar, based on `--requires-publish-pod` presence on command line, returns `CSIDriverInfo` with `requires_publish_pod = true`.
+9. CSI volume plugin stores this flag for the CSI driver.
+
+When a volume is being mounted for a pod:
+
```
kfox1111: optional csi stage call here

contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,123 @@
...
+It runs as a sidecar container in every pod with a CSI driver on every node. Each CSI driver runs its own CSI Driver Registrar.
+* It has a new command line option `--requires-publish-pod` that admin (or CSI driver author) sets if the CSI driver needs pod information in `NodePublish`.
+* It implements server part of `CSIRegistration` service, i.e. listens for  `GetCSIDriverInfo` call and returns `CSIDriverInfo` with the flag.
+
+
+## End to end examples
+
+### CSI driver wants pod information in `NodePublish`
+1. CSI driver author provides documentation and templates that the driver need  pod information in `NodePublish`.
+2. Admin installs CSI driver with CSI Driver Registrar `--requires-publish-pod`, e.g. by applying helm template from CSI diver author.
```
sarjeet2013: s/diver/driver

contributors/design-proposals/storage/container-storage-interface-pod-information.md
jsafrane: It's not annotation.

As I wrote, I follow names introduced in Flex, but I don't have strong opinion about the format. CSI itself does not describe how volume_attributes should look like.

contributors/design-proposals/storage/container-storage-interface-pod-information.md
jsafrane: In case several pods use the same volume on the same node, NodeStageVolume is called only once. It would include just the first Pod. NodePublishVolume is called is called separately for each pod that uses the volume and CSI driver can do the right thing there.

In addition, Flex driver did not get any pod in `mountdevice` call (which is equivalent of NodeStageVolume in CSI` and nobody complained.

Flex did not provide pod information in `unmoun` (=NodeUnpublishVolume) and nobody complained. Do you have concrete use case when pod can be useful in NodeUnpublishVolume?

contributors/design-proposals/storage/container-storage-interface-pod-information.md
jsafrane: Fixed, thanks

contributors/design-proposals/storage/container-storage-interface-pod-information.md
jsafrane: I added NodeStageVolume to the example, with a note that it won't get any pod information.

contributors/design-proposals/storage/container-storage-interface-pod-information.md
kfox1111: Oh, really? I thought stage was called once per pod, but mount was called per container mounting it? I guess I was mistaken. So I'll have to rewrite the image driver to have those semantics then. Thanks.

jsafrane:
We're considering using cluster-global API object to hold "configuration" of CSI driver in Kubernetes, i.e. how Kubernetes talks to the driver. It should contain also a flag if kubelet should pass pod name/namespace/uid in NodePublish. We won't need new gRPC API in kubelet this way.

The design is it flux right now, stay tuned.

kfox1111:
hmm. cluster global info might allow for pod inline validation to happen earlier too. +1.

jsafrane:
I reworked the proposal to use `CSIDriver` introduced in https://github.com/kubernetes/community/pull/2514. It's much shorter and easier now.

jsafrane:
And I kept the original commit there, just in case we need to revert to it.

contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,42 @@
...
+
+CSI is agnostic to container orchestrators (such as Kubernetes, Mesos or CloudFoundry) and as such does not understand  concept of pods and service accounts. [Enhancement of CSI protocol](https://github.com/container-storage-interface/spec/pull/252) to pass "workload" (~pod) information from Kubernetes to CSI driver has met some resistance. 
+
+## High-level design
+We decided to pass the pod information as `NodePublishVolumeRequest.volume_attributes`.
+
+* Kubernetes passes pod information only to CSI drivers that explicitly require that information in their [`CSIDriver` instance](https://github.com/kubernetes/community/pull/2523). These drivers are tightly coupled to Kubernetes and may not work or may require reconfiguration on other cloud orchestrators. It is expected (but not limited to) that these drivers will provider ephemeral volumes similar to Secrets or ConfigMap, extending Kubernetes secret or configuration sources.
```
kfox1111: typo at 'provider ephemeral'

contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,42 @@
...
+* Kubernetes will not pass pod information to CSI drivers that don't know or don't care about pods and service accounts. It is expected (but not limited to) that these drivers will provide real persistent storage. Such CSI driver would reject a CSI call with pod information as invalid. This is current behavior of Kubernete and it will be the default behavior.
```
kfox1111: Kubernete typo

jsafrane: fixed

contributors/design-proposals/storage/container-storage-interface-pod-information.md
jsafrane: fixed, thanks!

jsafrane:
@kfox1111, do you have concrete use case for `csi.kubernetes.io/volume.name`? I am not against per se, but I'd like to know why can be this information useful.

contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,42 @@
...
+No API changes.
+
+### CSI enhancement
+We don't need to change CSI protocol in any way. It allows kubelet to pass `pod.name`, `pod.uid` and `pod.spec.serviceAccountName` in [`NodePublish` call as `volume_attributes`]((https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume)). `NodePublish` is roughly equivalent to Flex `mount` call.
+
+The only thing we need to do is to **define** names of the `volume_attributes` keys that CSI drivers can expect:
+	*	`csi.kubernetes.io/pod.name`: name of the pod that wants the volume.
+	*	`csi.kubernetes.io/pod.namespace`: namespace of the pod that wants the volume.
+	*	`csi.kubernetes.io/pod.uid`: uid of the pod that wants the volume.
+	*	`csi.kubernetes.io/serviceAccount.name`: name of the service account under which the pod operates. Namespace of the service account is the same as `pod.namespace`.
```
jsafrane: The strings above define sort of API between Kubernetes and kubernetes-aware CSI drivers. @liggitt, does it need any special approval? It will be set to stone when we implement this proposal.

Background: `volume_attributes` is generic `map[string]string`, without any prescribed format of keys in [CSI standard](https://github.com/container-storage-interface/spec/blob/master/spec.md). I used the above keys because similar ones are already used in flex volumes. Now is the time to change the keys if we don't like them.

kfox1111:
At very least, volume.name + pod.id is a guaranteed unique identifier for the volume instance. So could be useful to drivers that are pod aware and want to track lifecycle? There may be other uses too. Haven't thought too much about it.

contributors/design-proposals/storage/container-storage-interface-pod-information.md
```diff
@@ -0,0 +1,42 @@
...
+## Detailed design
+
+### API changes
+No API changes.
+
+### CSI enhancement
+We don't need to change CSI protocol in any way. It allows kubelet to pass `pod.name`, `pod.uid` and `pod.spec.serviceAccountName` in [`NodePublish` call as `volume_attributes`]((https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume)). `NodePublish` is roughly equivalent to Flex `mount` call.
+
+The only thing we need to do is to **define** names of the `volume_attributes` keys that CSI drivers can expect:
+	*	`csi.kubernetes.io/pod.name`: name of the pod that wants the volume.
```
saad-ali: `csi.storage.kubernetes.io` to align with `CSIDriver` group name

jsafrane: Renamed to `csi.storage.k8s.io/*`

jsafrane:
Renamed parameters to `csi.storage.k8s.io/*` and squashed.

jsafrane:
And added feature gate `CSIPodInfo`.

jsafrane:
Implementation: https://github.com/kubernetes/kubernetes/pull/67945

contributors/design-proposals/storage/container-storage-interface-pod-information.md
ritazh: @jsafrane For inline CSI volume, what would be the equivalent of flexvolume's `options`? Currently we can only pass in `volumeAttributes` to `csi` as part of a PersistentVolume deployment. Is there a way to support `volumeAttributes` as part of inline CSI volume as part of the pod spec?

saad-ali:
This was implemented as alpha in k8s 1.12 (https://github.com/kubernetes/kubernetes/pull/67945). Merging the doc so that it can be updated.

/approve
/lgtm

PR #2442, opened 2018-07-31
Add KEP for cloud provider doc maintenance

d-nishi:


andrewsykim:
A lot more to work to be done here but this is a good starting point. 

@timothysc noted! We had agreed to use KEPs to standardize doc format across providers since we need to be coordinated for easier consumption from sig-docs. Maybe it is overkill but this is a good way for us to provide guidance to other subprojects (providers) on how to document things. We're happy to consider other approaches here :)! 

/approve

andrewsykim:
ref: https://github.com/kubernetes/kubernetes/issues/66825

PR #2443, opened 2018-07-31
KEP: Pruning for CustomResources

sttts:


deads2k:
@mbohlool @apelisse fyi

@sttts can you replicate the outstanding comments here like mehdy did in https://github.com/kubernetes/community/pull/2420 so we can re-find our comments?

contributors/design-proposals/api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+```
+
+An empty object `{}` would be defaulted to `{"a":{"x":42}}`
+
+
+Obviously, the default values should validate. But OpenAPI validation schemata do not enforce that. Also one default in a branch of `anyOf`, `allOf`, `oneOf` might force another validation run onto another branch of these quantors.
+
+Even worse: forcing onto another branch might make more defaults applicable. In other words, we would need to run validation with defaulting until it converges. This is a hint that the semantics of defaulting in full OpenAPI schemata are not well defined either.
+
+**We propose to allow defaults only outside of `anyOf`, `allOf`, `oneOf`, `not`.**
```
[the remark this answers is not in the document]
sttts: Adding a comment.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+>   "anyOf": [
+>	    { "properties": {"c":{}} }, 
+>       { "properties": {"d":{"not":{}}}}
+>   ]
+>}
+>```
+>
+> Then `a`, `b`, `b.x`, `c`, `d` are specified, `b.y` and `e` are not specified.
+
+**Question:** is it natural semantics to assume `d` to be specified and not to prune it? Note that there is no object with `d` that validates against `"d":{not:{}}`. But due to the `anyOf` there are objects which will validate against the complete schema.
```
sttts: Mehdy Bohlool: There should be a difference between not specifying "d" and specifying it like this. Not specifying it means pruning and define it like this means validation fail

Stefan Schimanski: OpenAPI is too powerful: you can specifiy it this way, still the validatio succeeds. That's basically our semantics problem: we have the three cases:
1. not specified
2. specified and validating
3. specified and not validating (but the whole schema validates).

sttts: **David Eads:** I would expect `d` to remain even if it is not used to satisfy validation. I expect `d` to remain because it's potential presence is acknowledged and the fact that the subrule can never be satisfied isn't a level of introspection we should perform.

**Antoine Pelisse:** +1

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+#### Pruning Options
+
+Motivated by the examples, we have different options to define the pruning semantics:
+1. Use the described semantics of `specified`.
+2. Only consider `properties` fields in the schema which actually successfully validate a given object (then `d` would be pruned in example 3).
```
sttts: **David Eads:** I'd prefer to not do this.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+3. Only consider `properties` fields outside of `anyOf`, `allOf`, `oneOf`, `not` during pruning.
```
sttts: **Antoine Pelisse:** I don't understand what this means. Does this mean pruning fields in the AllOf (for example), forcing it to fail validation?

**Stefan Schimanski:** No, it means that we only keep fields specified outside of any of anyOf, allOf, oneOf, not. So if you forget to specify a field outside, but only inside an anyOf, it will be pruned. Probably not what we want.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+4. Only consider `properties` fields outside of `anyOf`, `allOf`, `oneOf`, `not` during pruning, but enforce that every `properties` key inside `anyOf`, `allOf`, `oneOf`, `not` also appears outside all of those.
```
sttts: **Antoine Pelisse:** Isn't that the same as the one before? Since if you drop the fields in oneOf, allOf, anyOf, they are always going to be invalid, so they have to also appear in `properties`.

**Stefan Schimanski:** We would reject the CRD if the outer properties are not a super-set of the ones under all/any/oneOf/not.

So option 4 protects at least the user from mistakes.

Note: all these are not about validating a CR or not, but how to specifiy what you don't want to prune via CRD OpenAPI validation schemata.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+The OpenAPI validation schema allows to define defaults via the `default` field. For example:
+
+```json
+{"properties": {"a": {"default": {"x": 42}}}}
+```
+
+An empty object `{}` would be defaulted to `{"a":{"x":42}}`
+
+
+Obviously, the default values should validate. But OpenAPI validation schemata do not enforce that. Also one default in a branch of `anyOf`, `allOf`, `oneOf` might force another validation run onto another branch of these quantors.
```
sttts: **David Eads:** this doesn't bother me

**Stefan Schimanski:** We do a second round of validation in the registry validation step. So we catch them.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+Even worse: forcing onto another branch might make more defaults applicable. In other words, we would need to run validation with defaulting until it converges. This is a hint that the semantics of defaulting in full OpenAPI schemata are not well defined either.
+
+**We propose to allow defaults only outside of `anyOf`, `allOf`, `oneOf`, `not`.**
```
sttts: **David Eads:** I'm ok with this restriction to start, but I'm also ok with simply running multiple times. Eventually defaulting will stop defaulting.

**Stefan Schimanski:** also note that this convergence happens at some point. But there is no guarantee that it always converges against the same value (depending on which branches we take first) We should better avoid it altogether. IMO it's a strong hint for bad semantics.

**Antoine Pelisse:** I agree with Stefan. I'm not sold on the fact that defaults on properties (outside xxxOf/not) is completely solving that problem though?

**Antoine Pelisse:** To clarify my point, some structures might validate before defaults are applied, some might not validate after, so the order still matters.

**Stefan Schimanski:** The skeleton has no "require". So if it validates after defaulting, it was validating before as well. So this is fine.

**Antoine Pelisse:** I don't think that is true because of 
"not". Validation can be OK because a value wasn't there (before being defaulted). `{"properties": {"a": {"default": 42}}, "anyOf": [{"not": {"properties": {"a": {}}, "additionalProperties": false}}]}`

**Stefan Schimanski:** You claim to have a counterexample of the inverse of my statement :-) Of course, defaulting can make an object invalid. But dropping constrains (including all negative constains (i.e. under not or oneOf)) makes a schema weaker. So the skeleton is weaker than the full schema.

sttts: **Mehdy Bohlool:** why we don't allow it but fail on confusion/conflicts. That means validating the schema before actually using it for defaulting (and even pruning) and while the schema can be a valid OpenAPI schema, we reject it because, for example, it has two defaults for a single value using anyOf.

**Stefan Schimanski:** the conflicts depend on inputs. Computing the generic overlap for all input is computationally hard and needs complete validation schemata. Structural restriction is far less complex and allow to have unspecified validatoin schemata.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+>```json
+>{
+>  "anyOf":[
+>      {"properties": {"a": {"pattern": "1"}, "b": {"type": "string"}, "d": {"default": 1}}},
+>      {"properties": {"a": {"pattern": "2"}, "c": {"type": "string"}, "d": {"default": 2}}}
+>  ]
+>}
+>```
+>
+> Take an object `{"a": "1", "b":"foo", "c":"bar"}`. Assuming we follow pruning option 1, it is pruned to `{"a": "1", "b":"foo", "c":"bar"}` and to default to `{"a":"1", "b": "foo", "c":"bar", "d": 1}`.
```
sttts: **David Eads:** I would not have expected to be pruned. I'd expect `c` to be kept and `d=1` (first rule in what I described).

I'm also ok with restricting this for a first impl.

**Antoine Pelisse:** how would this work with `additionalProperties: false`?

**Stefan Schimanski:** We forbid that additionProperties and properties are used in parallel for the same object/map. So `additionalProperties: false` means to prune everything.

sttts: @deads2k added this as a comment further down that this is the starting point with the options for relaxing this later if needed.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+>
+> But what about the object `{"a": "12", "b":"foo", "c":"bar"}`? Both patterns apply. With pruning option 1 we keep `"b"` and `"c"`. But what about the default? Should we default to `1` or `2`?
```
sttts: **Mehdy Bohlool:** this is another example think can be easily detected and we can reject the schema because of confusion/conflicts in it. It should be clear from the schema what we are pruning and what we are defaulting. We can even try to make a list of allowed field with their default from schema (I believe that's proposal number 1) and if that process failed, we reject the schema.

**Stefan Schimanski:** I don't think we can easily detect overlaps. E.g. we would have to intersect regular languages (which is possible, but has high complexity).

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+>```json
+>{
+>   "properties": {
+>       "a": {"type": "string", "pattern": "<some-long-regex-for-ips>", "format": "ip"},
+>       "b": {"properties": {"x": {}, "y": {}}}
+>   }
+>}
+>```
+>
+> The type and properties are usually called schema validation, while regex patterns and formats are about values. The latter would be tested in the validation phase, the former would be checked during decoding for native types in the JSON decoder.
```
sttts: **Mehdy Bohlool:** I am not sure about format. it usually uses as an extra identifier with type and may not actually validate the value. Though the border here is blurry. When we say the type is string and the format is date, is it considered a validation or it is the "date" type? My understanding was "type" is primary type and "format" is secondary type.

**Stefan Schimanski:** Added a paragraph below about our discussion around keeping type and format in the skeleton.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+* which does not contain value validations
+* which is complete enough for pruning and defaulting.
+
+**Remark:** we have two options to define and implement pruning and defaulting:
+1. directly on the full OpenAPI validation schema with two custom algorithms doing parallel recursion over the schema and the input object.
+2. via an intermediate representation (the skeleton schema) and using go-openapi pruning and defaulting. Both algorithms are ten-liners based on the go-openapi/validate output. With the skeleton schema the go-openapi pruning algorithm coincides with our pruning option 1.
+
+Both routes lead to the same algorithm: the intermediate representation of the skeleton schema makes merging of OpenAPI validation schema constraints explicit, while the custom algorithms would hide that in its recursion code.
+
+Moreover, the main reason for the intermediate schema: the custom algorithm would have to replicate a lot of the validation logic of go-openapi, e.g. the semantics of `properties`, `additionalPropoerties`, `patternProperties` and the same for items of an array. With route 2 we get all of this for free.
```
sttts: Looks like we reject `patternProperties` today completely, although it exists in JSONSchemaProps. So for completeness it does not hurt to keep it in the doc here, even though in practice it does not matter. Maybe some day we add patternProperties.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+> * `additionalProperties`,
+> * `default`,
+> giving `drop(s)`,
+> 2. then merging `s_i` into it’s containing object.
+>
+>    I.e.: `skel(s) := merge(drop(s), s_1, …, s_n)` with
+>
+>```
+>merge(x_1, …, x_n) := { 
+>   "type": t if all x_i agree on t as type, undefined otherwise
```
sttts: **Mehdy Bohlool:** does type matter for pruning/defaulting?

**Stefan Schimanski:** no, see discussion about type and format below. We can leave out type+format completely.

**Stefan Schimanski:** added comment about that.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+>	    "c": {}
+>    }
+>}
+>```
+>
+> The `type` of `a` does not match on all paths. Hence, it is omitted in the skeleton. In contrast, `b` has a unique `type` of `"string"`, so it stays in the skeleton.
+
+### Types and Formats
+
+Note, that having less `type` constraints in the skeleton than in the whole OpenAPI validation schema means that admission and conversion will see possibly wrong types. The final validation in the registry validation phase will check for the complete OpenAPI validation schema and catch those type errors.
```
sttts: **Mehdy Bohlool:** so the plan is to pass `skel` schema to these webhooks instead of the full schema?

**Stefan Schimanski:** no, the webhooks won't see the schema at all. It is embedded in the decoder. But the webhooks will see the result of decoding, and without type+format the result might be unexpected in the sense that the Golang JSON decoder used by the webhook implementation – if that uses a Golang representation of CustomResources – will fail.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+{"properties": {"x": {"format": "json"}}}
+```
+
+In this example "x" is excluded from pruning. Note that you can still use any kind of OpenAPI validation schema constructs to restrict `"x"` further.
+
+Note that we lose expressiveness of the existing `format` strings: we either apply the format to `"json"` or any other pre-defined format. I.e. we cannot express that pruning should be disabled, but if it is an integer, it should be an int32.
+
+**Question:** this feels like a reasonable loss of expressivity. Do we accept that?
+
+**Raw JSON Option 2:** add an extension property, e.g.
```
sttts: **Nikhita Raghunath:** I like this more. I don't see any downsides to this. Do you?

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+```json
+{"properties": {"x": {"x-no-pruning": true}}}
```
sttts: **Mehdy Bohlool:** While I think we don't need first approach and map[string]object is expressing that, I like to have extension. There could be cases that I want to express some validations on some of the fields of a json object but I don't want pruning of other fields in that sub-object. e.g. is when the sub-object has ObjectMeta or TypeMeta but the rest of the object is a free json.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+```
+
+This would not lead to a loss in expressivity. It is formulated negatively intentionally to have `false` as the default with `omitempty`.
+
+As OpenAPI extensions are prefixed with `x-kubernetes` elsewhere, we probably should name it `x-kubernetes-no-pruning: true`.
+
+**We propose to follow option with the `x-kubernetes` prefix** because it is more explicit, does not reduce expressivity and does not mangle with the already very vague `format` field definition.
+
+### Nested x-no-pruning
```
sttts: @mbohlool @erictune @deads2k Do you have an opinion?

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+In https://github.com/kubernetes/kubernetes/pull/64907 a ratcheting mechanism during validation is introduced: if a specific proposition during validation is not fulfilled in an object stored in etcd, it is not required either during updates of that object.
+
+Ratcheting does not work well with arbitrary propositional OpenAPI validation schemata because propositions in negative positions (under `not` or `oneOf`) that are weakened via ratcheting will make the whole validation stricter instead of weaker (compare https://github.com/kubernetes/kubernetes/pull/64907#issuecomment-396526377).
+
+We have multiple options to restrict the use of ratcheting (which could be the property of certain OpenAPI validation schemata constructs):
+1. Allow ratcheting only in the skeleton schema. Everything in the skeleton schema is positive.
+2. Disallow ratcheting under `not` and under `oneOf`, but allow it everywhere else.
+3. Disallow ratcheting in negative positions. This implies that it is disallowed under `oneOf`, but is allowed e.g. under `not: {not: ….}`.
+The first is the most strict one, the last is the most relaxed one. On the other hand, 1 is the simplest to reason about.
```
sttts: **Nikhita Raghunath:** If I were using this, I would prefer 1 because it is easy to reason about and IMO if the author is changing the schema, they are aware that they might break their users...I think 1 should be sufficient.

Would be nice to factor in @liggitt 's views on this too.

sttts:
@erictune welcome back! Could you take a look, especially at the pruning-opt-out mechanism?

erictune:
Yes! I am excited about this and will take a look monday!

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+### Types and Formats
+
+Note, that having less `type` constraints in the skeleton than in the whole OpenAPI validation schema means that admission and conversion will see possibly wrong types. The final validation in the registry validation phase will check for the complete OpenAPI validation schema and catch those type errors.
+
+**Question:** we could extent the `skel` function to keep a list of `type` values. As long as all branches define the type, we can add `anyOf: [{type: "type1", type: "type2, ….}]` to the skeleton. If one branch does not define the type though, this is still not possible.
+
+In native types, we have custom unmarshallers for date / timestamps. In OpenAPI these would be rendered as `{type: "string", format: "date"}`. With the proposed skeleton algorithm, we would not verify these fields before full OpenAPI validation in the registry. We could move non-contradicting `format` constraints into the skeleton as well, like we do for `type` already. Then admission would be protected from invalid format (if admission uses Golang decoding, this might be relevant to get proper error messages).
```
erictune: Is this important, beyond trying to mimic what we did for core types, which grew organically?  I talk to users of admission webhooks and I haven't heard anyone complain that they are seeing insufficiently validated data.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+        "b": 2,
+        "y": {
+            "c": 3,
+            "z": 42
+        }
+    }
+}
+```
+
+is pruned to  `{"x":{"b": 2, "y":{"z": 42}}}`.
```
erictune: The only use case I know for no pruning is an object that has a unschematized sub-tree.

So, a go type like this:

```
type FooSpec struct {
  // How many things to make
  Replicas int32
  // Object definition for things to make, can be any Kubernetes type.
  // No validation is applied, in order to support future types.
  // Just needs to be valid json/yaml.
  AnythingTemplate runtime.RawExtension
}
```
I think would be transformed into OpenAPI like this:
```
"properties":   {
    "replicas": {
        "type": "integer"
    },
    "anythingTemplate": {
        "additionalProperties": true
    }
}
```

Therefore, I think the way you express "no pruning" is by specifying that a properties has `additionalProperties: true`.  This would apply to the subtree.  No non-bool additionalProperties would be supported.


I think we disallow `additionalProperties: `

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+## Non-Goals
+* Add a strict mode to the REST API which rejects objects with unknown fields.
+* Propose or decide anything about ratcheting. [We discuss the topic](#ratcheting-(slightly-related)) because it is related, but this needs a follow-up KEP.
+
+## Motivation
+### Pruning
+Pruning of a JSON value means to remove "unknown fields". A field (given as a JSON path) is unknown if it is not specified in the OpenAPI validation schema.
+
+A JSON path `<JSON path>.x` is specified in an OpenAPI validation schema if `"x" is a key in a corresponding `properties` field in the schema.
```
zhouhaibing089: looks like here is a format issue. :)

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
liggitt: there was more discussion on the ratcheting issue/PR at https://github.com/kubernetes/kubernetes/issues/64841#issuecomment-405431399, specifically around:
* what cases we allow tightening validating for (current plan is only for problems that result in fatally flawed objects)
* when ratcheting validation can be promoted to standard validation (only once we can verify there is no persisted data that would fail the tighter check)

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+    * [References](#references)
+    * [Alternatives Considered](#alternatives-considered)
+
+## Overview
+
+Native Golang based resources do not persist JSON fields which are not part of the Golang structs which are backing them in the API server memory. This artifact of the use of typed Golang structs inside of the REST implementation has turned into an API convention. Major parts of the Kubernetes depend on this to ensure consistency of data persisted to etcd and returned from the REST API. 
+
+Without consistency of data in etcd, objects can suddenly render unaccessible on version upgrade because unexpected data may break decoding. Even if persisted data has correct format and decodes correctly, having it not gone through validation and admission when it was stored can break cluster-wide invariants. For example assume the `privileged: true/false` field is added to a type in Kubernetes version X+1. In version X, there is no security check around this. So every user could set that flag if we didn’t drop unknown fields. When the field is added in X+1, that user suddenly has escalated access (note: on read we do not run admission). This is a serious security risk.
```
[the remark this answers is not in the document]
liggitt: > I can see that bad data might break decoding for types with generated decoders that make assumptions, such as that a property value is numeric not string. But, would this happen for CRD decoding? Since they decode to Unstructured, there presumably aren't the same assumptions.

CRDs are often paired with clients generated from strongly typed structs. Those clients would fail to get/list/watch if the API server returned schema-invalid data. xref https://github.com/kubernetes/kubernetes/issues/57705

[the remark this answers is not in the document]
sttts: @erictune I had those clients in mind as @liggitt writes. I will clarify the text.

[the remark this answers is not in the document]
sttts: done

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+CustomResources are persisted as JSON blobs today (with the exception of `ObjectMeta` which is pruned since Kubernetes 1.11), i.e. we do not drop unknown fields, with the described consequences. This proposal is about adding a decoding step named "pruning" into the decoder from JSON to `unstructured.Unstructured` inside the apiextensions-apiserver. This pruning step will drop fields not specified in the OpenAPI validation spec, leading to the same persistence semantics as for native types.
+
+Algorithmically deeply related to validation and pruning is "defaulting", i.e. the application of the OpenAPI "default" values. Defaulting is done after decoding for native types, before conversion. This proposal suggests to add defaulting just after pruning to the CustomResource JSON decoder.
+
+## Goals
+
+* Prune unknown fields from CustomResources silently. Unknown means not to be specified in the OpenAPI validation spec.
```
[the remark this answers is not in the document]
sttts: fixed

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: fixed

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: We propose below to only allow defaults outside of `anyOf`, `allOf`, ... So we don't have to decide about values and which branches they fulfill for them, making application of defaults very simple as well.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: Yes, I agree about `not`. IDLs (like e.g. Golang types) are probably always strictly-positive.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+Obviously, the default values should validate. But OpenAPI validation schemata do not enforce that. Also one default in a branch of `anyOf`, `allOf`, `oneOf` might force another validation run onto another branch of these quantors.
+
+Even worse: forcing onto another branch might make more defaults applicable. In other words, we would need to run validation with defaulting until it converges. This is a hint that the semantics of defaulting in full OpenAPI schemata are not well defined either.
+
+**We propose to allow defaults only outside of `anyOf`, `allOf`, `oneOf`, `not`.**
+
+Without that, we would get ambiguous defaulting semantics:
+
+#### Example 4:
```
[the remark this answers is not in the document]
sttts: Even for schemata coming from Golang types, I can imagine people want defaults depending on other fields. But we probably don't add support for those cases in defaulting Godoc tags in such an IDL. So webhooks are the natural way to express these if needed.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+>
+> But what about the object `{"a": "12", "b":"foo", "c":"bar"}`? Both patterns apply. With pruning option 1 we keep `"b"` and `"c"`. But what about the default? Should we default to `1` or `2`?
+
+Clearly, we can make the algorithm deterministic by choosing one, e.g. the first matching branch. This choice is arbitrary, hard to understand, and gets even more confusing with even more nested `allOf`, `anyOf` or even `not`.
+
+We can later become less restrictive about the allowed positions of default values in the scheme. I.e. the restriction of only allowing default outside of `allOf`, `anyOf`, `oneOf` and `not` is a starting point with a way forward if necessary.
+
+### Mixing of Schema and Value Validation
+
+The OpenAPI validation schema mixes the actual structural schema validation (which is usually done by the Golang struct JSON decoding for native types)
```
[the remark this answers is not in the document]
sttts: With "usually" I meant that most of the structural validation is done in the decoder. This does not exclude that more structural validation is done in the validation step of the generic registry. I am pretty sure we have those cases somewhere in our core types.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: Right. It's mostly about getting similar error messages.

I would claim we have a general, high-level goal for CRDs to behave like native resources. This is not special for pruning and defaulting. Merely, this goal is one driver to implement pruning and defaulting (and many other features we are adding to CRDs).

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+
+For pruning and defaulting we have to apply the OpenAPI validation schema inside of the decoder step (the left box inside apiextensions-apiserver in the figure above). This is considerably earlier than for native types. This has a number of implications:
+* We have to apply full OpenAPI value validation (e.g. regular expressions, propositional evaluation) during decoding.
+* Our generic registry applies validation after defaulting. We need the other way around for CRDs.
+* Our generic registry expects defaulting to always succeed (no error result type). CRD validation needed by defaulting can fail.
+
+To avoid these and to get sane semantics, **we propose to split the CRD validation into two steps:**
+1. During decoding we validate using a skeleton schema which lacks value validations and propositional logic (`anyOf`, `allOf`, `oneOf`, `not`).
+2. During defaulting we validate using the skeleton schema and apply the resulting defaulter.
```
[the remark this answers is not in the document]
sttts: Yes.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,486 @@
...
+During standard generic registry validation phase we validate using the full validation schema.
```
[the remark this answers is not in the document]
sttts: Yes, we will. We have this hardcoded in the mutating admission webhook code (https://github.com/kubernetes/kubernetes/blob/e4c219df42c77ecb8f0588197072bef81bca7429/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go#L152) and this will be called the same way for CRDs.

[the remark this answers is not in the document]
sttts: added a note

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: If we don't do type validation during decoding, the admission webhooks have to cope with wrong types **and report good error messages**. Technically this is not a problem, but it puts more responsibility on the admission webhook implementation. If we do that already in decoding, the behaviour might be more uniform.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: @erictune any opinion here?

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: additionalProperties only applies to one level, not to subtrees. I don't like to divert from its normal semantics. Not having `additionalProperties` is equivalent to `"additionalProperties": true`. Differentiating between those two is odd.

sttts:
> Can you confirm: after this is implemented, will users need to put additionalProperties: false in their OpenAPI in order to benefit from pruning? Do they need to put it "at each level of the tree"?

No. `additionalProperties: false` would mean to reject unknown fields. We don't do that. Unknown fields are completely valid. Pruning is a step after validation, i.e. a different layer. And we let the user to opt-in in v1beta1 CRDs by defaulting of `CRD.spec.pruning` to `false`. But we will switch to `CRD.spec.pruning: true` in v1 CRDs and even enforce it during creation and rely on `x-kubernetes-not-prune: true` for opt-out.

I have added a section about this.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
sttts: Ok, looks like ratcheting for native type is up for debate. I will leave this section here. In the non-goal we write that implementing ratcheting for CRDs is out of scope here. So this is merely a comment to be in-line with ratcheting if it is ever agreed on.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
erictune: Actually, I am saying something different.
I think a goal of pruning is to increase security for all types, reduce unforseen errors and provide a consistent experience for Kubernetes users.  We think that pruning should be in most cases.   But, in order to get this benefit, developers have to actively ask for pruning by laboriously adding `additionalProperties: false` to every single level.   Based on that, a likely result is that most developers don't bother, or they sometimes forget to add additionalProperties at some levels.  

Therefore, a plan is needed, I think, to encourage consistency.  Some options I see are:

1. use the OpenAPI 2.0 semantics , which are defaulting to `additionalProperties: false`.  Pro: devs have to actively opt out of pruning. Con: This would make us not be OpenAPI 3.0 standard compliant. 
2. expect that OpenAPI will be generated by tools we influence, like KubeBuilder and Operator Framework.  We will direct those frameworks to always add `additionalProperties: false` except in the case of `map` and `RawExtension` types. Pro: achieves result; con: may miss some developer workflows.
3. add a nested  `x-kubernetes-no-prune` tag. pro: devs only have to put the tag in a few places, at the root of spec and status.  con: devs still have to remember to put it in those places.

erictune:
Ah, so "pruning" means "silently remove", and "additionalProperties: false" means "reject"?  Obvious now but I did miss that distinction at first.

sttts:
> Ah, so "pruning" means "silently remove", and "additionalProperties: false" means "reject"? Obvious now but I did miss that distinction at first.

There was discussion about adding a strict mode to our rest model, especially for CLI clients. Then pruning and additionalProperties would conincide in that special mode. Currently we depend on OpenAPI client side validatoin in the CLI for the same use-case.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,510 @@
...
+> Which branch should be chosen while applying the defaults.
+
+#### Example 5:
+
+> Assume
+>
+>```json
+>{
+>  "anyOf":[
+>      {"properties": {"a": {"pattern": "1"}, "b": {"type": "string"}, "d": {"default": 1}}},
```
lavalamp: We should **not** be allowing users to mix value based and schema based requirements in the same check. Our entire decoding stack depends on a schema check before values can even be observed. What you have written here is valid OpenAPI but it is not valid for a kube resource. We should not support this. If we currently support this, we need to figure out how to stop doing so.

Crisply defining what is and is not allowed is maybe not going to be easy.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
lavalamp: We need to engineer our schema-acceptance criteria so we never face this question, I think.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
lavalamp: >  it will converge as the number of branches is finite and defaulting is idem-potent. But it is a strange behaviour nevertheless.

I'm pretty sure that this is turing complete or close to it if you can engineer branches that drop fields. Which is more or less what this KEP is proposing.

I really think we need to look hard at the the schema and come up with some rules that prevent people from saying things that they can't say with go types. I have some thoughts on how to do this, but this KEP isn't the place to hash it out. We should NOT be able to get turing-complete behavior by iterated default application.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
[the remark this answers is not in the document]
lavalamp: Not only should we make that assumption, we should actively work to reject all of these "interesting" examples at validation time.

We should NOT be allowing defaults to depend on other field values. People must use webhooks if they want that behavior.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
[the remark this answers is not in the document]
lavalamp: Sorry if I'm beating a dead horse by now, but I think this is a major bug report for the schemas we allow people to put in `validation`. As the text says, we do value validation and (limited) schema checks in completely different spots in our stack, and we depend on important stuff (e.g., webhooks) happening in between. Inseparably combining the two checks makes it impossible to reconcile with our stack. As @sttts pointed out in an earlier thread on this, it is NOT always possible to separate an openapi rule into schema-check and value-check. Therefore, I think we must start detecting and forbidding such schemas.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,510 @@
...
+2. During defaulting we validate using the skeleton schema and apply the resulting defaulter.
+During standard generic registry validation phase we validate using the full validation schema.
+
+Because the first step succeeds, also the second step succeeds. So we don’t need error handling during defaulting. The third step will catch wrong types of the used defaults.
+
+Note that defaulting – like for native types – will be done
+1. before conversion of the incoming API objects and of objects read from storage and
+2. after mutating admission webhooks (compare [k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go](https://github.com/kubernetes/kubernetes/blob/e4c219df42c77ecb8f0588197072bef81bca7429/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go#L152)).
+
+## Formal Proposal – following pruning option 1
```
lavalamp: I think I would rather approach it this way: reject bad schemas early.

1. Define a schema that expresses exactly and only the things we want to allow folks to express (i.e., things that can be done with go structs).
2. Translate both ways between that schema and OpenAPI spec.
3. Round-trip a CRD's `validation` tag through that. If it comes through unchanged, it is acceptable. Otherwise, it is unacceptable.

How exactly to roll this out into the validation of CRDs given that it's tightening our restrictions is another discussion.

I appreciate the cleverness that went into this section of the KEP. I think, however, we should keep in mind the overall goal of this chunk of the code base: we're allowing more ways of getting a specific chunk of functionality. OpenAPI spec allows a bunch of features that are about different functionality. Since we're not trying to implement arbitrary APIs, we shouldn't be clever in supporting those features, we should be preventing people from using them.

Nothing made by kube-openapi or kube-builder should be using these features. We are not taking features away from people by rejecting specs mixing schema and value validations.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,510 @@
...
+	
+  // Prune enables pruning of unspecified fields. Defaults to false.
+  // Note: this will default to true in version v1.
+  Prune *bool
+}
+```
+
+I.e. for `apiextensions.k8s.io/v1beta1` this will default to `false`.
+
+For `apiextensions.k8s.io/v1` we will change the default to `true` and forbid `false` during creation and updates. In `v1` the only way to opt-out from pruning is via setting `x-kubernetes-no-prune: true` in the schema.
```
lavalamp: Are you going to automatically propagate this with a version conversion function? Otherwise I don't see how you avoid a sudden behavior shift.

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,510 @@
...
+
+## Opt-in and Opt-out of Pruning on CRD Level
+
+We will add a pruning flag to `CustomResourceDefinitionSpec` of `apiextensions.k8s.io/v1beta1`:
+
+```golang
+type CustomResourceDefinitionSpec struct {
+  ...
+	
+  // Prune enables pruning of unspecified fields. Defaults to false.
```
lavalamp: I'm not sure about this. Yes, it's a behavior change. But the people who need it are the people who won't turn this flag on. How do we avoid surprising them?

A tool that you can run that scans your CRs and flips this bit if none of them would get pruned, and sends you a email otherwise? It seems hard.

The only reason not to turn it on is that people might have a spec that's missing an important field. Maybe in 1.12 we can somehow make the CRD's status field indicate whether CRs are stored that need pruning, and then tell people they can't upgrade to 1.13 until they fix it?

keps/sig-api-machinery/customresources-pruning-and-defaulting.md
```diff
@@ -0,0 +1,510 @@
...
+reviewers:
+  - "@deads2k"
+  - "@lavalamp"
+  - "@liggitt"
+  - "@erictune"
+  - "@mbohlool"
+  - "@apelisse"
+approvers:
+  - "@deads2k"
+  - "@lavalamp - substituted by @fedebongio while @lavalamp is out"
```
lavalamp: I'm back, sorry!

lavalamp:
Given the difficulty of fixing specs, and the fact that we have tools that should give correct specs, can we prune assuming the spec is correct and generated by kube-openapi / kube-builder? I think we should do this or just not prune at all until we can tell if the spec is acceptable or not.

I'm not in favor of pruning based on specs that use features we don't support.

deads2k:
> Given the difficulty of fixing specs, and the fact that we have tools that should give correct specs, can we prune assuming the spec is correct and generated by kube-openapi / kube-builder? I think we should do this or just not prune at all until we can tell if the spec is acceptable or not.
> 
> I'm not in favor of pruning based on specs that use features we don't support.

Pruning or defaulting?  Pruning based on the union of all mentioned fields won't strip any expected fields and will leave the reconciliation of different potential branches to validation where it is more appropriate.  Pruning is to prevent field squatting and to have structural consistency in your stored data.

I think the question about "correct" specs is more about defaulting where results become slightly harder to reason about.  However, crisp rules about only defaulting on non-branching paths make the behavior predictable and easy to decide about.

Separately, a decision to tighten the rules for specs we allow may be useful and we could have a path forward to to doing that based on status/failure to serve unless the "I know I'm going to be unservable in six months" flag is set.  I don't think that we need to block on such a tightening to implement the pruning or defaulting though.

lavalamp:
> Pruning or defaulting?

Both, if they are writing special code to handle the weird specs--as much as I want defaulting. We shouldn't go fishing while our house is on fire. Honestly I would have blocked adding the validation field in the first place until this was fixed if I had realized how flexible the openapi schema was. I'm also extra interested in schemas being correct due to working on apply.

> Pruning based on the union of all mentioned fields won't strip any expected fields and will leave the reconciliation of different potential branches to validation where it is more appropriate. 

It might fail to prune items that should be pruned.

Now, I think it is acceptable to not prune at all, either just simply rejecting unknown fields or letting the user deal with the fallout (ok for beta, likely not ok for GA). I also don't object to pruning. I think I object to pruning only most of the time, though.

> I think the question about "correct" specs is more about defaulting where results become slightly harder to reason about. However, crisp rules about only defaulting on non-branching paths make the behavior predictable and easy to decide about.

It is certainly laudable that sttts noticed these issues and is working to address them. However I think it is fundamentally wrong that we accept specs that express behaviors we never intend to support or allow. I think that a stopgap solution of writing code that notices and works around the issue is actually worse than doing nothing, because of the precedent it sets (that we have to respond appropriately to every OpenAPI feature). In my view, all of our effort to rectifying the situation needs to go into explicitly defining the subset of specs we will support.

> Separately, a decision to tighten the rules for specs we allow may be useful and we could have a path forward to to doing that based on status/failure to serve unless the "I know I'm going to be unservable in six months" flag is set.

#64841 might be homing in on a general solution to that thorny problem. CRD is still beta so we have some wiggle room.

> I don't think that we need to block on such a tightening to implement the pruning or defaulting though.

From my perspective, code that notices non-compliant specs and doesn't act on them (ideally getting an error somewhere visible) is fine and possibly our only choice given that tightening the validation will be a multi-release effort.

But I view code that attempts to repair/skeletonize/reinterpret non-compliant OpenAPI specs and then use them anyway as detrimental. Code like that is incredibly difficult to get correct in my experience, and incredibly frustrating when it does the wrong thing. I especially object if the code is silent and doesn't give users any indication that they're working off of our non-standard interpretation of their OpenAPI spec.

sttts:
> However I think it is fundamentally wrong that we accept specs that express behaviors we never intend to support or allow.
>> Pruning based on the union of all mentioned fields won't strip any expected fields and will leave the reconciliation of different potential branches to validation where it is more appropriate.
> 
> It might fail to prune items that should be pruned.
> 
> Now, I think it is acceptable to not prune at all, either just simply rejecting unknown fields or letting the user deal with the fallout (ok for beta, likely not ok for GA). I also don't object to pruning. I think I object to pruning only most of the time, though.
> 
>> I think the question about "correct" specs is more about defaulting where results become slightly harder to reason about. However, crisp rules about only defaulting on non-branching paths make the behavior predictable and easy to decide about.
> 
> It is certainly laudable that sttts noticed these issues and is working to address them. However I think it is fundamentally wrong that we accept specs that express behaviors we never intend to support or allow. I think that a stopgap solution of writing code that notices and works around the issue is actually worse than doing nothing, because of the precedent it sets (that we have to respond appropriately to every OpenAPI feature). In my view, all of our effort to rectifying the situation needs to go into explicitly defining the subset of specs we will support.

I have tried hard to follow these harsh judgements about the approach, and I would have hoped to have an open and also early-on discussion about these instead of this form. And not surprisingly I disagree with them:

- OpenAPI validation is and always has been about the validation step in our stack, not about decoding. Today's validation code for native resources does a lot of crazy things and the language we use there (Golang) is turing complete by nature. Hence, people can formulate any crazy restrictions on objects and they do. Our OpenAPI validation schema language gives us some limited way to define validations without writing code. With admission webhooks everybody is able to define even more crazy APIs without any limitation.

  In short: it does not matter that OpenAPI validation allows to restrict APIs in ways which do not comply with our API conventions.

- Without pruning people define tons of APIs today which do not follow our conventions. With versioning we allow them to turn them into beta and GA APIs. Not having pruning is dangerous if we care about API conventions.
- I claim that pruning using the skeleton schema enforces sane Golang-like schemata, without any embedded propositional logic.

  > It might fail to prune items that should be pruned.

  Which are they? I read from the comments that there is either fundamental disagreement about what is intuitively expected from pruning or misunderstanding of the approach. Would be happy to understand.

- If we start inventing our own small schema language, we will end up enforcing a schema like the skeleton, taking away expressivity from the user for more advanced validations. This is fine. But it will force ppl towards writing admission webhooks for anything more advanced. But how does it help to enforce API convention (i.e. sane Kubernetes-like APIs)? IMO it won't either, see point 1. Users need expressivity and expressivity allows them to shoot themselves into their foot. Whether expressibity is called OpenAPI or admission webhook does not matter.

liggitt:
> There is an element of end-user experience in this one. What do we want them to experience?
> 
> If we value consistency, the default experience for custom resources should be the same as built-in resources. If we apply the principle of least astonishment what should the experience be and how do we build an implementation from there?

We value both consistency and experience.

CRDs should behave consistently with built-in resources.

If there are experience issues with some of the built-in resource behaviors, we should determine if we can improve those (and ensure consistent treatment for CRDs as well).

erictune:
Recording this for posterity....
CRD Conversion it set to go into v1.13.  This KEP is not going to.  This is okay for the following reasons:
* CRD Conversion is important to a broad set of CRD users.  Pruning, less so.
  *  I've spoken to a number of sophisticated CRD users.  They included:
      * contributors to what was then known as CoreOS Operators
      * Istio contributors
      * Knative contributors
  * All said that CRD conversion was an important and urgent feature.  None said that Pruning is an urgent feature.  (However, some core type developers said pruning was essential for security sensitive use cases.)
* Most CRDs cannot benefit from this KEP.  But all can benefit from CRD Conversion.
  * 90% of CRDs on Github didn't use `.spec.validation`, in a survey I did of over 400 CRDs.    This doesn't mean they don't do some form of validation - they may validate in an admission webhook, or the controller may set error conditions in the status for invalid CRs.    CRD Conversion does not depend on having a `.spec.validation`, so it benefits all CRD users.
* We can add this KEP later.
  * Pruning can be enabled as a flag per-apiversion, later on. (It is a per-CRD flag in this KEP, but it could easily be per apiversion.)
  * Suppose The User has a CRD with version v1, which uses validation, but of course no pruning.  Later, pruning support is added to CRDs in Kubernetes version v1.x.0.   Now, The User adds version v2, and enables pruning for version v2.   Pruning is per apiVersion, so this should work.  And v1 clients with incorrect fields will still keep working.
*  There isn't consensus about pruning.  However, support for apiVersion conversion appears to be universal. 
  * In hindsight, some Kubernetes API experts feel it is still the right behavior, while others think we should instead have done to rejection for unknown fields.  Not listing the arguments here.

I summary, we move forward where there is greatest demand and certainty,  without blocking future options.

bgrant0607:
I agree with @erictune. Pruning can be added later.

CRDs, especially schema-less ones, are like annotations in many ways.

Related past discussions:
https://github.com/kubernetes/kubernetes/issues/30819
https://github.com/kubernetes/kubernetes/issues/5889

Either strict validation OR pruning can prevent the problems of circumventing validation, creating timebombs, and confusing controllers.

Each has different gotchas for clients.

bgrant0607:
And, for the record, while I haven't had time to read this KEP, I am in favor of pruning, API consistency, and declarative schema specification. But I don't think pruning can be required for all CRDs.

justaugustus:

justaugustus:

/close

Issue #2245, opened 2018-06-12
KEP numbering is broken


tallclair:
Here are the current keps, sorted by filename number and displayed alongside the metadata number:

```
0000:
  ./0000-kep-template.md  -  0

0001:
  ./0001-kubernetes-enhancement-proposal-process.md  -  1

0002:
  ./sig-cloud-provider/0002-cloud-controller-manager.md  -  2

0003:
  ./sig-cluster-lifecycle/0003-cluster-api.md  -  3

0004:
  ./sig-cluster-lifecycle/0004-bootstrap-checkpointing.md  -  4

0005:
  ./sig-contributor-experience/0005-contributor-site.md  -  5

0006:
  ./0006-apply.md  -  6

0007:
  ./sig-contributor-experience/0007-20180403-community-forum.md  -  0007
  ./sig-network/0007-pod-ready++.md  -  1

0008:
  ./sig-cluster-lifecycle/0008-20180504-kubeadm-config-beta.md  -  draft-20180412
  ./0008-20180430-promote-sysctl-annotations-to-fields.md  -  8
  ./sig-cli/0008-kustomize.md  -  8

0009:
  ./0009-node-heartbeat.md  -  8

0010:
  ./sig-network/0010-20180314-coredns-GA-proposal.md  -  10

0011:
  ./sig-network/0011-ipvs-proxier.md  -  TBD

0012:
  ./sig-network/0012-20180518-coredns-default-proposal.md  -  11
```

The problems include number collisions & metadata / filename mismatch. Due to these problems, the current numbering only provides friction & confusion. I think the numbers either need to be abandoned, or automation needs to be added to ensure the correct numbering process is followed.

justaugustus:
+1 
Do we think it's enough to just have KEPs organized in folders by SIG and abandon the numbering scheme?

...or perhaps have the numbering be based on the date of the proposal e.g., `sig-pm/180625-really-cool-kep.md`?

jdumars:
I'm putting this on the SIG-Architecture agenda for this week. Hopefully we'll get to it.

bgrant0607:
This came up in the original KEP discussions. I think there was some assumption that an "editor" would assign numbers post-merge. 

I am a fan of the date approach.

I think the intent of the numbering is, at a glance, to get some sense of which KEPs are more recent.

tallclair:
[...]

PR #2448, opened 2018-07-31
[RuntimeClass] Add examples & alpha criteria, mark as implementable

tallclair:
Address a few of @dchen1107 's comments that were left open when https://github.com/kubernetes/community/pull/2290 merged: added examples & alpha criteria.

Also marked as implementable in its current state, as this is the feature set we want to pursue for the initial alpha in 1.12.

/sig node
/milestone v1.12

### DIFF
--- a/keps/sig-node/0014-runtime-class.md
+++ b/keps/sig-node/0014-runtime-class.md
@@ -1,5 +1,5 @@
 ---
-kep-number: 14 FIXME(13)
+kep-number: 14
 title: Runtime Class
 authors:
   - "@tallclair"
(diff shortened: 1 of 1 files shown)


keps/sig-node/0014-runtime-class.md
```diff
@@ -1,5 +1,5 @@
 ---
-kep-number: 14 FIXME(13)
+kep-number: 14
```
tallclair: This was supposed to be 13, but since sig-arch is [redoing the KEP numbering](https://github.com/kubernetes/community/issues/2245), I just left it as is.

tallclair:
/hold
Needs @kubernetes/sig-architecture-pr-reviews

keps/sig-node/0014-runtime-class.md
```diff
@@ -176,6 +176,69 @@ const (
...
+    runtimeHandler: gvisor
+----
+kind: RuntimeClass
+apiVersion: node.k8s.io/v1alpha1
+metadata:
+    name: kata-containers
+spec:
+    runtimeHandler: kata-containers
+----
+# provides the default sandbox runtime when users don't care about which they're getting.
```
yujuhong: How is the default supplied? Sorry it's been a while since I last read it so this may have already been covered before.

tallclair: In this example, I meant the user just specifies "RuntimeClass = sandboxed", so not actually API defaulting - and this RuntimeClass defines what "sandboxed" means.

tallclair:
I dropped the change to mark this as implementable, pending sig-arch review.
/hold cancel

PR #2449, opened 2018-08-01
Expand requirements for new cloud providers

andrewsykim:
This ensures that new (out-of-tree) providers are up to the same technical standards & meets the same requirements as existing in-tree providers. 

cc @calebamiles


aoxn:
cool

PR #2452, opened 2018-08-01
Include out-of-tree provider examples to cloud provider docs KEP

andrewsykim:
Add manifest examples for out-of-tree providers as part of cloud provider docs. 

Follow up to https://github.com/kubernetes/community/pull/2442


hogepodge:
The cleanup looks good to me, adding a description would be helpful for the history.

andrewsykim:
cc @kubernetes/sig-cloud-provider-pr-reviews

PR #2137, opened 2018-05-08
First draft of SIG CLI charter

seans3:
* First draft of SIG CLI charter
* Updates SIG CLI chairs and leads.

PR #2453, opened 2018-08-02
First version of SIG CLI Charter

seans3:
Implements the initial governance rules for SIG CLI.


cblecker:
@kubernetes/steering-committee 
/committee steering

spiffxp:
Hi @seans3 is this a redo of https://github.com/kubernetes/community/pull/2137? If so please close that one and point to here

philips:
cc @bgrant0607 SIG CLI review

sig-cli/charter.md
```diff
@@ -0,0 +1,50 @@
...
+
+#### Code, Binaries and Services
+
+SIG CLI code include command line tools and binaries for working with Kubernetes
+API's. Examples of these binaries include: kubectl and kustomize.
+
+### Out of scope
+
+SIG CLI is NOT responsible for defining the Kubernetes API that it
+interfaces with. The Kubernetes API is the responsibility of SIG API Machinery.
```
philips: It is probably worthwhile to call out all of the other CLIs that SIG CLI doesn't oversee and why. For example kubeadm, kube-apiserver, etc, etc.

philips: The why being that these CLIs don't interact with the API I suppose.

sigs.yaml
```diff
@@ -492,23 +492,30 @@ sigs:
       establishment of conventions for writing CLI commands, POSIX compliance,
       and improving the command line tools from a developer and devops user
       experience and usability perspective.
-    charter_link:
+    charter_link: charter.md
     label: cli
     leadership:
       chairs:
+      - name: Maciej Szulik
```
liggitt: duplicated entry

sigs.yaml
```diff
@@ -492,23 +492,30 @@ sigs:
...
+        github: soltysh
+        company: Red Hat
+      - name: Sean Sullivan
```
liggitt: seems like it might be better to make chair changes separately from the charter addition

sig-cli/charter.md
```diff
@@ -0,0 +1,50 @@
...
+### In scope
+
+SIG CLI section in [sigs.yaml]
+
+#### Code, Binaries and Services
+
+SIG CLI code include command line tools and binaries for working with Kubernetes
+API's. Examples of these binaries include: kubectl and kustomize.
+
+### Out of scope
```
liggitt: would it make sense to include some "Cross-cutting and Externally Facing Processes" aspects of how sig-cli collaborates with the sigs that own other binaries to improve command consistency?

sig-cli/charter.md
soltysh: There's nothing stopping us from looking after them, but nobody reached out to us for so, that's why since none of us (sig participants) authored these we don't feel like owning them nor don't want to 'steal' other's work.

sigs.yaml
soltysh: I'd just move this to separate commit, which should be sufficient.

sigs.yaml
soltysh: Nope, there are chairs which are responsible for organizational issue with sig and leads which are responsible for technicalities. Sean and I are both chairs, on top of that Phil and I are tech-leads, which makes me deserve double entry :1st_place_medal:

sig-cli/charter.md
soltysh: On a second note, I think that calling out what we don't oversee might be an overkill. IMO it's sufficient we clearly define ownership in the previous paragraph. The fact that API is being mentioned here is specifically to call out that we don't and never will change the API, that's sig-apimachinery responsibility.

sig-cli/charter.md
soltysh: Not quite sure what you mean by that?

sig-cli/charter.md
philips: I am not going to hold strong on this point. But, I do still believe it would be helpful to provide illustrative counter examples.

seans3: To recap, following Maciej's advice:

1) We won't mention kubeadm or kube-apiserver.
2) We will keep the "Out of scope" statement as is.

If misunderstood, please let me know.

sig-cli/charter.md
```diff
@@ -0,0 +1,50 @@
...
+  shown in the test on-call schedule of the [Test Playbook] are included in this
+  group. Test Health Maintainers are SIG CLI Members.
+
+### Subproject Creation
+
+Option 1: by [SIG Technical Leads](https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md#L100)
+
+
+[sig-governance]: https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md
+[sigs.yaml]: https://github.com/kubernetes/community/blob/master/sigs.yaml#L487
```
spzala: Use of sigs.yaml with line number is not recommended by the latest changes in the charter template for a good reason that the number for sig section may not remain same - https://github.com/kubernetes/community/commit/e0d60fdd7cfdafb71335c9501c458278bf7fdf97

sig-cli/charter.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Scope
+
+The Command Line Interface SIG (SIG CLI) is responsible for kubectl and
+related tools. This group focuses on command line tools and
+libraries to interface with Kubernetes API's.
+
+### In scope
+
+SIG CLI section in [sigs.yaml]
```
bgrant0607: Please link to the corresponding information in the sig-cli README instead.

sig-cli/charter.md
bgrant0607: I don't think it's helpful to lump all Kubernetes components, such as kube-apiserver, under "CLI".

I do think it's helpful to indicate that SIG CLI doesn't own all conceivable command-line tools within the project: kubeadm, kops, kompose, kubefed, etc.

I would add something like:

SIG CLI is not responsible for command-line tools built and maintained by other SIGs, such as kubeadm, which is owned by SIG Cluster Lifecycle.

(Just to include one example and not make it appear to be an exhaustive list.)

sig-cli/charter.md
```diff
@@ -0,0 +1,50 @@
...
+
+#### Code, Binaries and Services
+
+SIG CLI code include command line tools and binaries for working with Kubernetes
```
bgrant0607: I suggest "general-purpose command-line tools and binaries" as a way to distinguish these from the special-purpose tools built by other SIGs.

philips:
Since @pwittrock, @bgrant0607, and myself from the steering committee have all taken a look at this and #2137 I am going to approve this.

However, if @jbeda would like to take one final look as one of the designated [Steering reviewers](https://docs.google.com/spreadsheets/d/1i_8oM1KlmXo90zIIgNTdOg-M6xesTggFLnTkgmscMlI/edit#gid=0) I will hold until EOD.

/lgtm
/approve
/hold

spiffxp:
/lgtm
from me too, FWIW

PR #2459, opened 2018-08-02
[wip] Tightening up conformance requirements

spiffxp:
Pushing for comment during sig-arch, still in progress


contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
 # Conformance Testing in Kubernetes
 
-The Kubernetes conformance test suite is a set of testcases, currently a
-subset of the integration/e2e tests, that the Architecture SIG has approved
-to define the core set of interoperable features that all Kubernetes
-deployments must support.
+The Kubernetes Conformance test suite is a subset of e2e tests that 
+SIG Architecture has approved to define the core set of interoperable 
+features that all Kubernetes clusters must support. The tests verify
```
timothysc: I wouldn't use the word "must".  Folks could fail tests and not get the badging.

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
+that the expected functionality works as a user might encounter it in
```
timothysc: expected "behavior" works as an end user

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
-Contributors must write and submit e2e tests first (approved by owning Sigs). 
-Once the new tests prove to be stable in CI runs, later create a follow up PR 
-to add the test to conformance. This approach also decouples the development 
-of useful tests from their promotion to conformance.
+The process to add new conformance tests is intended to decouple the
+development of useful tests from their promotion to conformance:
+
+- Contributors write and submit e2e tests, to be approved by owning SIGs
+- Tests are proven to meet the conformance test requirements
+- A follow up PR is submitted to promote the PR to conformance
```
timothysc: promote the test to conformance.

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
-- should be stable and run consistently
-- cannot skip providers (there should be no Skip like directives added to the test), 
-especially in the Nucleus or Application layers as described 
-[here](https://github.com/kubernetes/community/blob/master/contributors/devel/architectural-roadmap.md).
-- cannot test cloud provider specific features (i.e. GCE monitoring, S3 Bucketing, ...)
-- should work with default settings for all configuration parameters 
-(example: the default list of admission plugins should not have to be tweaked for passing conformance).
-- cannot rely on any binaries that are not required for the
-linux kernel or for a kubelet to run (i.e. git)
+Conformance tests should be boring. They should rely on nothing special, test only
```
timothysc: I'd revise this statement.

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
+features that are available everywhere, and generate the same results every time.
+
+More specifically, a test is eligble for promotion to conformance if:
+
+- it tests only GA, non-optional features or APIs (ie: no alpha or beta endpoints, 
+  no feature flags required)
+- it works for all providers (ie: no `SkipIfProviderIs`/`SkipUnlessProviderIs` calls)
+- it is non-privileged (ie: no root on nodes, network, nor cluster)
```
timothysc: That's not true^

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
+- it works without access to the public internet
+- it works without non-standard filesystem permissions granted to containers or users
+  (eg: can't share a writable host /tmp with a container)
+- it does not rely on any binaries that would be required for the linux kernel or
+  kubelet to run (ie: can't rely on git)
```
timothysc: Some of these constraints seem weirdly arbitrary and not really a requirement for some tests.

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
+- it is stable and runs consistently (ie: no flakes)
+
+Examples of features which are not eligible for conformance tests:
+
+- node/platform-reliant features, eg: multiple disk mounts, GPUs, high density, etc.
```
timothysc: This violates the long term ark around profiles.

contributors/devel/conformance-tests.md
duglin: Do we have examples of tests that are allowed to fail but people can still claim conformance?

contributors/devel/conformance-tests.md
spiffxp: Profiles don't exist and aren't a thing yet.  When they are, we can iterate on this doc.

contributors/devel/conformance-tests.md
duglin: Can we just abstract this a bit and talk about how it can't require platform, or vendor, specific artifacts? Being too specific here might lead to people wanting/needing to add more and more over time, and I don't think this should be the complete list, rather the guidelines.

contributors/devel/conformance-tests.md
spiffxp: I don't understand, are you saying folks should be allowed to fail conformance tests but get certified as conformant anyway?

contributors/devel/conformance-tests.md
spiffxp: How?

contributors/devel/conformance-tests.md
spiffxp: How?  There are conformance tests today that require privileges?  Or are you saying that there's functionality that is core that requires privileges to verify that we haven't written tests for yet?

contributors/devel/conformance-tests.md
duglin: +1 to dealing with it later but we should talk about how to deal with optional features at some point. For example, I'd like to see svc-cat have a conformance test suite and I'd like to discuss the best process to make that happen.

contributors/devel/conformance-tests.md
spiffxp: How so?  Trying to spell out explicit violations we've seen in the past

duglin: Yep, I get the intent. I'm just not keen on having to list each and every way people may violate our goals here, and keep updating it as they find new ways :-)   So, if possible, let's talk about the idea in the abstract.  E.g., perhaps:
it works on all providers without any proprietary add-ons and without requiring a configuration that is not deemed acceptable (by SIG-Arch) for running the test suite.

The point of the last part is that I'm not sure we can enumerate all of the possible config options in this list, e.g. doing things like requiring sharable/write /tmp with the host might, or might not, be ok in the future. SIG-Arch should define what kind of setup in which the tests should run and what the requirements are on the environment in-which it runs. That doc can, and will, change over time and maybe per release so it should be versioned, but in this doc (which is more like our charter to me) should be written to be pretty static across all versions - if we can.

spiffxp: This list needs to be specific enough that it can be used as a checklist for those reviewing whether or not a given test is eligible for promotion to conformance

contributors/devel/conformance-tests.md
dankohn: I'd suggest changing "all Kubernetes clusters must support" to "all conformant Kubernetes clusters must support".

contributors/devel/conformance-tests.md
timothysc: There are tests that require elevated privs already b/c they are in the API.

contributors/devel/conformance-tests.md
timothysc: The conformance tests  tests only GA, non-optional features or APIs 
 - no alpha or beta apis
 - no feature flags required

bgrant0607: Yeah, this statement is weird. The detailed list of requirements is below.

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
+More specifically, a test is eligble for promotion to conformance if:
+
+- it tests only GA, non-optional features or APIs (ie: no alpha or beta endpoints, 
+  no feature flags required)
+- it works for all providers (ie: no `SkipIfProviderIs`/`SkipUnlessProviderIs` calls)
+- it is non-privileged (ie: no root on nodes, network, nor cluster)
+- it works without access to the public internet
+- it works without non-standard filesystem permissions granted to containers or users
+  (eg: can't share a writable host /tmp with a container)
+- it does not rely on any binaries that would be required for the linux kernel or
```
bgrant0607: Should be would NOT be required.

SIG Node may have a list of system dependencies.

bgrant0607:
It would be useful to add a conventions/recommendations section to record decisions like using the pod log API (or termination message) to look at container output to verify expected container behavior.

bgrant0607:
Another class of "don't do this in conformance tests": Some information in the API isn't guaranteed to be stable over time. For instance, we don't guarantee that any specific Events will be generated.

Status and condition reason and message fields also aren't guaranteed to not change in the future:
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#typical-status-properties

So specific values shouldn't be matched in tests. In some cases, reason/message fields could be tested to be non-empty, however.

I just encountered this in #67144

spiffxp:
I have pushed updates to this PR to rebase and address existing comments.  However [as discussed at cncf-conformance-wg](https://docs.google.com/document/d/1D51Qzwff1nn_acK_cbRIdocLJcrK61UAcnZgKyTS2Y4/edit#heading=h.2h8mrtbyvenw) I will be moving this to a google doc for further iteration

spiffxp:
The aforementioned google doc https://docs.google.com/document/d/1kwsS0d0cVTg4idIOGrWzNHBoZ3UyEGi0fELbBlADgmI/edit

spiffxp:
/assign @smarterclayton @bgrant0607 @timothysc 
Assigning for approval

I've updated this PR description and contents according to the google doc linked in the description.  Last call for comments, I will merge this by noon Monday PST once I get approves/lgtms from the above.

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
+- Once we resolve the above issues, we plan on identifying the appropriate areas
+  to relax requirements to allow for the concept of conformance Profiles that
+  cover optional or additional behaviors
+
+## Conformance Test Requirements
+
+Conformance tests currently test only GA, non-optional features or APIs. More
+specifically, a test is eligible for promotion to conformance if:
+
+- it tests only GA, non-optional features or APIs (ie: no alpha or beta
```
dims: Do we want to say anything about
- addons?
- admission plugins?

dims: never mind, you said admission plugins below (line 57)

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
+| v1.1.x          | no     | no          | no          | yes         |
 
-### Running Conformance Tests
+## Running Conformance Tests
 
 Conformance tests are designed to be run with no cloud provider configured.
-Conformance tests can be run against clusters that have not been created with
-`hack/e2e.go`, just provide a kubeconfig with the appropriate endpoint and
-credentials.
+Conformance tests must be able to be run against clusters that have not been
```
dims: "to be run even when there are no cloud provider is configured"? (seems to be saying that we don't avoid them to run in an env when there is a cloud provider)

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
-  `framework.ConformanceIt()` call that includes all of the required
-  metadata about the test (see the [Test Metadata](#test-metadata) section)
-- use "Promote xxx e2e test to Conformance" as template of your PR title
-- tag your PR with "/area conformance" label
-- send your PR to Sig-Architecture for review by adding "@kubernetes/sig-architecture-pr-reviews" 
-also CC the relevant Sig and Sig-Architecture
+## Promoting Tests to Conformance
 
+To promote a test to the conformance test suite, open a PR as follows:
+- is titled "Promte xxx e2e test to Conformance"
```
dims: typo - Promote

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
+## Conformance Test Requirements
+
+Conformance tests currently test only GA, non-optional features or APIs. More
+specifically, a test is eligible for promotion to conformance if:
+
+- it tests only GA, non-optional features or APIs (ie: no alpha or beta
+  endpoints, no feature flags required, no deprecated features)
+- it works for all providers (ie: no `SkipIfProviderIs`/`SkipUnlessProviderIs`
+  calls)
+- it is non-privileged (ie: no root on nodes, network, nor cluster)
```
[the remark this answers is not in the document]
smarterclayton: Maybe rephrase as “does not require root on nodes, access to raw network interfaces, or cluster admin permissions)

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
+- is titled "Promte xxx e2e test to Conformance"
+- includes information and metadata in the description as follows:
+  - "/area conformance" on a newline
+  - "@kubernetes/sig-architecture-pr-reviews @kubernetes/sig-foo-pr-reviews
+    @kubernetes/cncf-conformance-wg" on a new line, where sig-foo is whichever
+    sig owns this test
+  - any necessary information in the description to verify that the test meets
+    [conformance test requirements], such as links to reports or dashboards that
+    prove lack of flakiness
+- modifies code as follows:
```
smarterclayton: Let’s add a bullet “- contains no other modifications to test source code” - that’s been very useful.

contributors/devel/conformance-tests.md
smarterclayton: We need to qualify that the test may create or set objects or fields that are alpha or beta that bypass policies or protections that are not yet GA, but which may reasonably be enabled on a conformant cluster.  Examples are attempting to set pod security policy to allow certain actions in test setup (and warning or allowing failure), or setting not-yet-GA scheduler annotations that may be used by admission control to protect against certain scheduling features.

Obviously we don’t need to do all of these up front, but I want conformant implementations to have flexibility to request certain security limiting choices to be relaxed by the test.

 if the test may be reasonably expected

duglin: @smarterclayton I want to understand what you're suggesting. I my view the tests will define, for the most part, what an end-user will see if they try to perform certain actions and a conformant impl must adhere to that definition. How it achieves that goal is really up to it since our testing isn't (and probably can't) get under the covers to know what's going on with each possible config knob. Having said that, if the tests explicitly talk about certain features or config settings then those must match what the impl has and does. An impl can not be consider conformant if the tests says "x must be set to y" if the impl either rejects that request or changes x to something other than y.  So, if (outside the scope of what the tests define) an impl uses alpha features that's its own choice that we have no say over.  Is this consistent with that you're saying?

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
+- it works for all providers (ie: no `SkipIfProviderIs`/`SkipUnlessProviderIs`
+  calls)
+- it is non-privileged (ie: no root on nodes, network, nor cluster)
+- it works without access to the public internet
+- it works without non-standard filesystem permissions granted to pods
+- it does not rely on any binaries that would not be required for the linux
+  kernel or kubelet to run (ie: can't rely on git)
+- any container images used within the test support all architectures for which
+  kubernetes releases are built
+- it passes against the previous release of kubernetes as well as the current
```
duglin: Is this consistent with the table below where it talks about the previous 2 releases, not just one? Or is there just missing an 's' on "release" in this sentence?

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
 Conformance tests are designed to be run with no cloud provider configured.
-Conformance tests can be run against clusters that have not been created with
-`hack/e2e.go`, just provide a kubeconfig with the appropriate endpoint and
-credentials.
+Conformance tests must be able to be run against clusters that have not been
+created with `hack/e2e.go`, just provide a kubeconfig with the appropriate
+endpoint and credentials.
 
 ```sh
+# ensure kubetest is installed
```
duglin: For completeness, do you want to say that they have to have the k8s src tree cloned first since not everyone running the tests could be a k8s dev?

contributors/devel/e2e-tests.md
```diff
@@ -580,6 +569,21 @@ breaking changes, it does *not* block the merge-queue, and thus should run in
 some separate test suites owned by the feature owner(s)
 (see [Continuous Integration](#continuous-integration) below).
 
+  - `[Conformance]`: Designate that is test is included in the Conformance
```
duglin: s/that is/that this/  I think

contributors/devel/e2e-tests.md
```diff
@@ -580,6 +569,21 @@ breaking changes, it does *not* block the merge-queue, and thus should run in
...
+test suite for [Conformance Testing](conformance-tests.md). This test must
+meet a number of [requirements](conformance-tests.md#conformance-test-requirements)
+to be eligible for this tag. This tag does not supersed any other labels.
+
+  - The following tags are not considered to be exhaustively applied, but are
```
duglin: I'm not sure what the point of these tags are since they seem to be orthogonal to conformance testing. From this doc's POV we should focus just on what it takes to be a conformance test. If people want to add tags for other purposes (e.g. to know which tests are 'alpha') then we should talk about that in some other testing doc. Talking about it here is confusing and could be misleading , for example, it could imply conformance tests might be related to "alpha" test/features when we specifically say above that they're not. I'd suggest removing this.

dims: @duglin we need this info to later organize the stuff into `buckets`. we need these to live in the test case themselves.

duglin: @dims during the last sig-arch call we did agree to determine what the categorized of "other" tests are so that we can have a discussion about what to do with them. However, I think its premature to then put text in here that might encourage people to open PRs to add these tags to our tests. I say this because, 1) we haven't agreed on what the tags should be, 2) after the analysis of these categories we may decide that each of them can be merged into the single conformance test suite that we have, thus then requiring follow-on PRs to remove these conformance-wise unnecessary tags.  As of now, we only have one tag, Conformance, and that's the only tag we should mention until we have further discussions in the conformance SIG.

If there's a desire to include something about this "other" stuff, I would prefer something that describes our current plans, something perhaps like:
- - -
There are certain features that are relatively easy to classify as a natural fit for our conformance test suite, such as CRUD operations on Pods. However, there are some features in Kubernetes that are less clear, often due to the variability, or optionality, of their support in certain implementations. SIG-Conformance is under-going an analysis of these features to determine the best path forward to determine if, or how, to include them within the conformance test suite.
- - -

In the meantime, we should gather the list of potential categories (the list in this PR is a good start) in a google doc, encourage people to add more (if needed) along with a justification for why they should, or should not, be part of conformance. Then we can discuss it in the next conformance meeting.

duglin: Only reason I suggested a google doc is because its easier for quick adds/removes of ideas. PR'ing something as part of a brainstorming/analysis process (which is the phase I think we're in) seems kind of heavy-weight.  I would also like us to go to weekly conformance SIG meetings for this very reason - I've added it to the agenda for our next call. Monthly meetings, IMO, discourage the community aspect of our SIG - which relates to the next point.

re: SIG-Conformance vs SIG-Arch... I think we probably need to get some clarity on process here. I definitely agree that whatever we do for conformance needs to be approved by SIG-Arch, but if all of our day-to-day brainstorming is done within the context of SIG-Arch then we might as well shutdown SIG-Conformance.  I would prefer a process, like all other SIGs, where we work on stuff and report our decisions up to SIG-Arch - and in some cases report them up for approval and not just when we need a big decision made. Granted we may get more oversight than some other SIGs due to the nature of our work, but I do want us to feel like we can have some degree of independence.

dims: The way i see it. this document is providing guidelines from sig-arch for the day to day work in the conformance sig.

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
+
+Conformance tests currently test only GA, non-optional features or APIs. More
+specifically, a test is eligible for promotion to conformance if:
+
+- it tests only GA, non-optional features or APIs (ie: no alpha or beta
+  endpoints, no feature flags required, no deprecated features)
+- it works for all providers (ie: no `SkipIfProviderIs`/`SkipUnlessProviderIs`
+  calls)
+- it is non-privileged (ie: no root on nodes, network, nor cluster)
+- it works without access to the public internet
```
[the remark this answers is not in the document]
spiffxp: Adding "short of whatever is required to pre-pull images for conformance tests" for now.  Down the line we may consider pre-pulled images part of the artifacts used to run conformance tests.

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,150 @@
...
+The Kubernetes Conformance test suite is a subset of e2e tests that SIG
+Architecture has approved to define the core set of interoperable features that
+all conformant Kubernetes clusters must support. The tests verify that the
+expected behavior works as a user might encounter it in the wild.
+
+The process to add new conformance tests is intended to decouple the development
+of useful tests from their promotion to conformance:
+- Contributors write and submit e2e tests, to be approved by owning SIGs
+  - Tests should begin by covering the most important and visible aspects of the
+    function
```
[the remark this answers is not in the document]
spiffxp: I'm going to drop the clarifying statement for now

contributors/devel/conformance-tests.md
spiffxp: Rewording to be more explicit, but this is consistent. v1.3.x is expected to pass against v1.3.x and v1.2.x, but not v1.1.x.

contributors/devel/conformance-tests.md
spiffxp: It is possible to do this without the k8s src tree, just using the released `.tar.gz` and `-test.tar.gz` artifacts

duglin: can we mention that then?  ie. show the URL where they can download the tar-ball from?  I prefer to assume we need to be pretty explicit for people - just to be extra helpful.

contributors/devel/e2e-tests.md
duglin: I can't find this section in https://docs.google.com/document/d/1kwsS0d0cVTg4idIOGrWzNHBoZ3UyEGi0fELbBlADgmI/edit  so I'm not sure its fair to imply that its been iterated on and agreed to.

spiffxp: Sourced from https://docs.google.com/document/d/1kwsS0d0cVTg4idIOGrWzNHBoZ3UyEGi0fELbBlADgmI/edit?disco=AAAACLaeHK4

duglin: yup- same doc I pointed to.  No mention of this section in there.

bgrant0607: @duglin This was raised in comments on that document and in 2 SIG Architecture meetings.

It is not just about "other" tests. Some tests that are currently part of the conformance suite are believed to not adhere to the aspirational criteria set forth by this doc. We want to assess how many tests fall into those categories and tag the tests so that they don't have to be re-identified later.

duglin: From a process perspective I would prefer that text being proposed, reviewed and implicitly agreed to, be put into the body of the document for each review and not left in a comment where it is easily missed - or interpreted as a WIP thought exercise.

contributors/devel/conformance-tests.md
bgrant0607: @smarterclayton 

I think we need to move towards the tests being run in a blackbox fashion, where cluster configuration and maybe even namespace creation and configuration are performed outside the context of the tests. Otherwise, depending on how authorization, security, networking, and other policies have been configured, various tests may fail.

If at all possible, I'd also like to define a set of functionality that really should work out of the box in all reasonable production-worthy environments (so, maybe not minikube or kubernetes-in-docker), including ones that are fairly locked down, which is the reason for the "non-privileged" point below.

As these guidelines are written to be aspirational and intended to trigger discussion about exceptions, I'd prefer to not add these types of caveats to this doc at this time. Let's assess the whole body of tests for divergence from these principles and then re-evaluate.

cc @spiffxp @dims

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,155 @@
...
+
+NB: This should be viewed as a living document in a few key areas:
+- The desired set of conformant behaviors is not adequately expressed by the
+  current set of e2e tests, as such this document is currently intended to
+  guide us in the addition of new e2e tests than can fill this gap
+- This document currently focuses solely on the requirements for GA,
+  non-optional features or APIs. The list of requirements will be refined over
+  time to the point where it as concrete and complete as possible.
+- There are currently conformance tests that violate some of the requirements
+  (ie: require privileged access), we will be categorizing these tests and
```
bgrant0607: nit: s/ie:/e.g.,/

contributors/devel/conformance-tests.md
dims: @bgrant0607 we need a list of principles in a living document. which is exactly what this doc is. So i am quite ok with listing them here since we have the caveat that this is not the full list and may be modified in the future. we have to start somewhere!

contributors/devel/conformance-tests.md
```diff
@@ -1,110 +1,155 @@
...
+- it is non-privileged (ie: does not require root on nodes, access to raw
+  network interfaces, or cluster admin permissions)
+- it works without access to the public internet (short of whatever is required
+  to pre-pull images for conformance tests)
+- it works without non-standard filesystem permissions granted to pods
+- it does not rely on any binaries that would not be required for the linux
+  kernel or kubelet to run (ie: can't rely on git)
+- any container images used within the test support all architectures for which
+  kubernetes releases are built
+- it passes against both the current development version of kubernetes under
```
bgrant0607: I don't think this bullet is consistent with the version-skew section below. How about we just forward-reference that section here?

contributors/devel/conformance-tests.md
```diff
@@ -1,57 +1,69 @@
...
- - A v1.1 cluster should pass v1.0, v1.1 conformance tests, and fail v1.2
-conformance tests
+For example, suppose we're in the midst of developing kubernetes v1.3. The
+following clusters must pass conformance tests built from the following branches:
 
+| cluster version | master | release-1.3 | release-1.2 | release-1.1 |
+| --------------- | -----  | ----------- | ----------- | ----------- |
+| v1.3.0-alpha    | yes    | yes         | yes         | no          |
+| v1.2.x          | no     | no          | yes         | yes         |
+| v1.1.x          | no     | no          | no          | yes         |
```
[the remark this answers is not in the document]
bgrant0607: @tpepper Yes, it would be reasonable to impose an earlier deadline. We continue to have the problem that new features are pushed through late in the process, are found to be buggy, and then are fixed in subsequent patch releases.

contributors/devel/conformance-tests.md
spiffxp: I'm going to put this under "example of areas we _may_ want to relax" so it's not just living in a PR

contributors/devel/e2e-tests.md
spiffxp: Let's start with this list, and if we find it's too granular, or not granular enough, we can amend the list

smarterclayton:
/lgtm

We can continue to evolve but this accomplishes the bulk of the short term goals for moving this process forward.

PR #2481, opened 2018-08-06
github management: tweak repo removal language

spiffxp:
Added a new case I just ran into, and removed SIG Architecture from having
to be involved in the decision to archive (I view consulting SIG Arch as
part of the process for finding new active maintainers of the project, we
could bake this into the language if it needs to be explicit)

Formatting changes:
- Trimmed whitespace in the FAQ
- Unindented headers
- FAQ questions are now bold rather than italic


spiffxp:
/hold
changing my mind, want this to be a checklist

spiffxp:
/hold cancel
checklist it now is

PR #2484, opened 2018-08-07
Improvements to CRD webhook conversion KEP

mbohlool:
A follow up to #2420 to address @lavalamp comments. @sttts @deads2k @erictune please let me know if you have any comment on the KEP and I can address them here or in a new PR.

/assign @lavalamp 
/sig api-machinery

### DIFF
--- a/contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
+++ b/contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
@@ -89,21 +89,21 @@ type CustomResourceDefinitionSpec struct {
   Version string
   Names CustomResourceDefinitionNames
   Scope ResourceScope
-  // This optional and correspond to the first version in the versions list
-  Validation *CustomResourceValidation
-  // Optional, correspond to the first version in the versions list 
-  Subresources *CustomResourceSubresources
   Versions []CustomResourceDefinitionVersion
-  // Optional, and correspond to the first version in the versions list  
+
+
+  // These fields are optional and correspond to the first version in the versions list
+  Validation *CustomResourceValidation
   AdditionalPrinterColumns []CustomResourceColumnDefinition
+  Subresources *CustomResourceSubresources
 
   Conversion *CustomResourceConversion
 }
 
 type CustomResourceDefinitionVersion struct {
   Name string
-  Served Boolean
-  Storage Boolean
+  Served bool
+  Storage bool
   // These three fields should not be set for first item in Versions list
   Schema *JSONSchemaProp
   Subresources *CustomResourceSubresources
@@ -111,7 +111,7 @@ type CustomResourceDefinitionVersion struct {
 }
 
 Type CustomResourceConversion struct {
-  // Conversion strategy, either "nop” or "webhook”. If webhook is set, Webhook field is required.
+  // Conversion strategy, either "None” or "Webhook”. If Webhook is set, Webhook field is required.
   Strategy string
 
   // Additional information for external conversion if strategy is set to external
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -225,7 +230,7 @@ error on server: conversion from stored version v1 to requested version v2 for s
 ```
 
 
-For operations that need more than one conversion (e.g. LIST), no partial result will be returned. Instead the whole operation will fail the same way with detailed error messages. To help debugging these kind of operations, the UID of the first failing conversion will also be included in the error message. 
+For operations that need more than one conversion (e.g. LIST), no partial result will be returned. Instead the whole operation will fail the same way with detailed error messages. Also the webhook will recieve the whole list to convert instead of items one by one.
```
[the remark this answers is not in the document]
mbohlool: fixed. Where do we have LIST as an example of multiple calls? It needs multiple conversion but it will be done with a single http call.

mbohlool:
@sttts PTAL
@lavalamp ping

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -178,20 +178,27 @@ type ConversionRequest struct {
   UID types.UID
   // The version to convert given object to. E.g. "stable.example.com/v1"
   APIVersion string
-  // Object is the CRD object to be converted.
+  // Object is the CRD object or list of objects to be converted.
```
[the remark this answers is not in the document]
lavalamp: Object will not be a CRD object, right? It will be of the type described in the CRD?

Also +1 to being explicit about the list type

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -178,20 +178,27 @@ type ConversionRequest struct {
...
   Object runtime.RawExtension
+  // IsList indicates that this is a List operation and Object contains a list of items.
+  IsList bool
```
lavalamp: Why is this necessary?

Another way to represent would be to have `List []runtime.RawExtension` and require that exactly one of `object` or `list` be set.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -178,20 +178,27 @@ type ConversionRequest struct {
...
 }
 
 type ConversionResponse struct {
   // UID is an identifier for the individual request/response.
   // This should be copied over from the corresponding ConversionRequest.
   UID types.UID
-  // ConvertedObject is the converted version of request.Object.
-  ConvertedObject runtime.RawExtension
+  // ConvertedObject is the converted version of request.Object if the Result field is nil.
+  // +optional
```
lavalamp: Probably worth being explicit about what to do for lists.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -225,7 +232,7 @@ error on server: conversion from stored version v1 to requested version v2 for s
 ```
 
 
-For operations that need more than one conversion (e.g. LIST), no partial result will be returned. Instead the whole operation will fail the same way with detailed error messages. To help debugging these kind of operations, the UID of the first failing conversion will also be included in the error message. 
+For operations that need more than one conversion (e.g. LIST), no partial result will be returned. Instead the whole operation will fail the same way with detailed error message. Also the webhook will receive the whole list to convert instead of items one by one.
```
lavalamp: Is it actually easy to always group items in a list like this, given our stack?

I think we shouldn't promise the *entire* list in one call. We will probably wish to call it in multiple batches, to avoid causing the webhook to need super high memory usage if for no other reason.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
```diff
@@ -178,20 +178,27 @@ type ConversionRequest struct {
...
+  ConvertedObject *runtime.RawExtension
+  // Result contains extra details into why a conversion request was failed. If it is not nil, it means
+  // the conversion has been failed.
+  // +optional
+  Result *metav1.Status
 }
 ```
 
-If the conversion is failed, the webhook should fail the HTTP request with a proper error code and message that will be used to create a status error for the original API caller.
+If the conversion is failed, the webhook is prefered to fill up the Result field of the response with a proper error code and message instead of returning failure HTTP codes.
```
lavalamp: Perhaps: "Instead of returning failure HTTP status codes, the preferred way for a webhook to indicate a conversion failure is to fill the `result` field of the response with an error code and message. Non-200 status return codes will be taken to mean there was a problem with the webhook rather than a problem with the object to be converted."

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: List has its own meta data. Do we care let conversion webhook to modify that? right now the webhook is responsible for any change to that meta data (specifically changing the APIVersion). If we pass a list of objects, then the caller(APIServer) not the webhook is responsible for any list metadata changes.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: refer to the other comment about list metadata. We pass a list item and implementation is strightforward right now. Even with current design and implementation, we can divide List object (UnstructuredList) into smaller valid List items and send them to webhook. We promise here that we send an UnstructuredList to the webhook. The webhook is not aware of the actual user request and we do not promise we send that whole user requested objects in one call.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
mbohlool: Fixed that in the implementation and will fix it here too.

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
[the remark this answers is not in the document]
mbohlool: nit: CR not CRD. Will fix that.

The list is a `UnstructuredList` (not sure what `CustomResourceList` is but I assume that is what you meant).

contributors/design-proposals/api-machinery/customresource-conversion-webhook.md
sttts: I don't see why the actual list object should be exposed to the conversion hook. Conversion is by list item by definition. @lavalamp's slice suggestion would make the interface obvious. We could even drop the `Object` field and just have an `Objects` instead.

liggitt: +1 for unifying the interface to something like `objects []runtime.RawExtension`, treating single and multi-item conversion requests consistently.

mbohlool:
@lavalamp @sttts PTAL.

mattfarina:
@lavalamp @deads2k what's the next step on this one?

liggitt:
obsoleted by https://github.com/kubernetes/community/pull/2737, I thought

justaugustus:

justaugustus:

/close

PR #2487, opened 2018-08-07
Add list of CSI drivers to Node.Status

jsafrane:
In https://github.com/kubernetes/community/pull/2034 we identified we need to store more information in a Node object. This proposal defines the API (`Node.Status.CSIDrivers`) and producer and consumers of the new `Node.Status` field.

It is on the edge between sig-node and sig-storage, tagging both.

/sig storage
/sig node

### DIFF
--- a/contributors/design-proposals/storage/container-storage-interface-node-status.md
+++ b/contributors/design-proposals/storage/container-storage-interface-node-status.md
@@ -1,0 +1,105 @@
+# Node.Status usage for CSI
+
+***Version:*** Alpha
+
+***Author:*** @jsafrane
+
+## Terminology
+
+Term | Definition
+---|---
+Container Storage Interface (CSI) | A specification attempting to establish an industry standard interface that Container Orchestration Systems (COs) can use to expose arbitrary storage systems to their containerized workloads.
+CSI Volume Plugin | A in-tree volume plugin that acts as an adapter and enables out-of-tree, third-party CSI volume drivers to be used in Kubernetes.
+CSI Volume Driver | An out-of-tree CSI compatible implementation of a volume plugin that can be used in Kubernetes through the Kubernetes CSI Volume Plugin.
+
+## Background & Motivations
+
+Storage backends may use different node IDs than Kubernetes - they may run on a distinct network with their own IP addresses or use separate DNS domains. Therefore when Kubernetes talks to a CSI Volume Driver, it must use node ID that is understandable to the CSI Volume Driver. A CSI Volume Driver reports ID of a node where it runs during registration of the driver to kubelet. Kubelet then collects node IDs of all CSI drivers installed on the node and currently publishes it as JSON dictionary in `csi.volume.kubernetes.io/nodeid`. In the end, each node has node IDs for all CSI drivers that run on the node.
+
+Example of a 1.11 node running hypothetical CSI drivers for AWS and Gluster:
+
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  name: node1.example.com
+  annotations:
+    csi.volume.kubernetes.io/nodeid: { \"csi.amazon.com/ebs\": \"ip-172-31-60-59.ec2.internal\", \"csi.gluster.org\": \"172.31.60.59\" }
+```
+
+This annotation cannot be extended with additional information as CSI specification evolves. In this proposal, we want to:
+
+  * Move this annotation into proper Node field.
+  * Add topology information introduced in CSI v0.3.0.
+
+This is part of
+
+* https://github.com/kubernetes/features/issues/557
+* https://github.com/kubernetes/community/pull/2034
+
+This fixes:
+
+* https://github.com/kubernetes/kubernetes/issues/66497
+
+
+## API
+
+```go
+// NodeStatus is information about the current status of a node.
+type NodeStatus struct {
+    // ...
+
+    // List of CSI drivers running on the node and their properties.
+    CSIDrivers []CSIDriverStatus
+}
+
+type CSIDriverStatus struct {
+    // CSI driver name.
+    Name string
+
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/container-storage-interface-node-status.md
```diff
@@ -0,0 +1,105 @@
...
+```go
+// NodeStatus is information about the current status of a node.
+type NodeStatus struct {
+    // ...
+
+    // List of CSI drivers running on the node and their properties.
+    CSIDrivers []CSIDriverStatus
+}
+
+type CSIDriverStatus struct {
```
PatrickLang: Since this was named Status I was expecting some field representing the health of the CSI driver. Should there be a status or health property?

thockin:
I don't think we should add this to node, at least not in this way.  Node is already a dumping ground for too many things, and is ballooning in size.

My first preference would be to add a new resource (via CRD) that is explicitly this information.  It can be 1:1 associated with Nodes by having the same name.  In some proposals this has been described as "structured annotations" or "out-of-line annotations" or "decorator objects".  The new resource can still be "official" in that Kubernetes supports CSI, and this is part of that support.  It's just not part of the Node resource.  Honestly, we should probably do that for a bunch of things.

An alternative (which I like less) would be to add a generic "drivers" concept and add a field that indicates which type of driver (e.g. driver.kubernetes.io/csi or something).  I like this less because it doesn't solve the growth of Node.

Thoughts?

vishh:
+1 on what Tim said. I have been exploring the idea of having distinct state objects for individual node extensions which can optionally influence the overall health of a node. 
We can re-use the recently proposed Lease APIs for tracking health and extend that object in ways specific to CSI.
We'd still need a pointer from the node objects to all the extension state objects that influence it's health.

jsafrane:
@thockin, I can live with a new API object loosely "bound" to Node by having the same name. Kubelet will get more complicated - it has nice infrastructure to report status, I need to make a separate one to report this new object. Not a big deal though.

I don't like CRD. I heard that there won't be any new API objects and every new object needs to be CRD. I don't think Kubernetes is prepared for this move. I can easily generate clients / informers for CRD, that's great, but that's about it.

* There is no distinct place in Kubernetes that would make sure that necessary CRD is installed during deployment. The CRD is part of Kubernetes API and *must* be installed when appropriate alpha feature is enabled. IMO it's not an addon. I'll try to find some place, starting with the thing that inserts default RBAC rules when the cluster (API Server?) starts...

* AFAIK, CRD means JSON as payload. There is no protobuff generated. Do we really want to use JSON again for **every** new API object?

* There is no version conversion. It's being added in 1.12 (?), but it requires a webhook. Who will serve the webhook? Probably not controller-manager. So <something> needs to run a pod with the webhook when cluster starts.

* There is no validation in API server, only regexp in CRD (which could be fine in this CSI case, so it's not that important).

We would need to do all this exploratory work to implement a stupid list of CSI drivers. Adding a new API object is less work right now. Why should be CSI the first? Others (e.g. https://github.com/kubernetes/community/pull/2354) still use traditional API objects in 1.12.

thockin:
>
> @thockin <https://github.com/thockin>, I can live with a new API object
> loosely "bound" to Node by having the same name. Kubelet will get more
> complicated - it has nice infrastructure to report status, I need to make a
> separate one to report this new object. Not a big deal though.
>

This is something to talk about with sig-arch (Saad said this is
scheduled?).  If this is a pattern we want to encourage, we may way to
formalize it more to handle this sort of use case.  E.g. maybe what we need
is a link from Node -> CSINode in a well-known field (e.g.
metadata.decorators), which tools can follow and ingest.


> I don't like CRD. I heard that there won't be any new API objects and
> every new object needs to be CRD. I don't think Kubernetes is prepared for
> this move. I can easily generate clients / informers for CRD, that's great,
> but that's about it.
>

It's not a hard rule that EVERY new object be CRD, but CRD is evolving to
be THE definition mechanism for resources.  It's not capable of
representing a lot of things, still, but growing.


>
>    -
>
>    There is no distinct place in Kubernetes that would make sure that
>    necessary CRD is installed during deployment. The CRD is part of Kubernetes
>    API and *must* be installed when appropriate alpha feature is enabled.
>    IMO it's not an addon. I'll try to find some place, starting with the thing
>    that inserts default RBAC rules when the cluster (API Server?) starts...
>
>
You are not alone in this journey, so we should find the right answer for
this


>    - AFAIK, CRD means JSON as payload. There is no protobuff generated.
>    Do we really want to use JSON again for *every* new API object?\
>
>
How frequently do you expect this object to be accessed?  CRD will have to
support better encodings, over time, but is it really an issue NOW?


>    - There is no version conversion. It's being added in 1.12 (?), but it
>    requires a webhook. Who will serve the webhook? Probably not
>    controller-manager. So needs to run a pod with the webhook when cluster
>    starts.
>
>
As above, you are not alone in this journey.


>    - There is no validation in API server, only regexp in CRD (which
>    could be fine in this CSI case, so it's not that important).
>
>
And getting more robust over time.

We would need to do all this exploratory work to implement a stupid list of
> CSI drivers. Adding a new API object is less work right now. Why should be
> CSI the first? Others (e.g. #2354
> <https://github.com/kubernetes/community/pull/2354>) still use
> traditional API objects in 1.12.
>

Those objects are not served by the API server but are centralized in
components.  Not a fair comparison.

I am not trying to penalize anyone (though I know storage has taken this
hit twice now - CF snapshots).  We need to start the process eventually.  I
am not 100% confident in my assertions on API machinery state, so bringing
the discussion to sig-arch is best.  That said, I believe my thinking here
aligns with the general direction we want to go, and someone has to go
first...

jsafrane:
We're exploring using CRD (`CSINodeInfo`) in https://github.com/kubernetes/community/pull/2034/

/close

PR #2491, opened 2018-08-07
Centralize list of administrators and moderators

### LINKED ISSUE #2490: Explicitly declare all moderators/admins of k8s communication channels
From contribex-list

@cblecker : "I would also like to suggest we have an official list of moderation/admin roles (youtube, discuss, etc) and have that PR'd into the community repo. We should have an official list somewhere so we know who holds these roles so we can audit, and we know who to reach out to if there is a problem."

How about a MODERATORS file similar to OWNERS-style in k/kubernetes/community/communication? Or I can just make it a normal markdown file with a list of moderators per property. 

/sig contributor-experience
/assign @castrojo

castrojo:
This gives us one list of all the moderators/admins on community properties in one place. 

Fixes #2490 

/sig contributor-experience
/hold

### DIFF
--- a/communication/K8sYoutubeCollaboration.md
+++ b/communication/K8sYoutubeCollaboration.md
@@ -40,4 +40,4 @@ Collaboration should simplify things for everyone, but with privilege comes resp
 
 Your community managers are happy to help with any questions that you may have and will do their best to help if anything goes wrong. Please get in touch via [SIG Contributor Experience](https://git.kubernetes.io/community/sig-contributor-experience).
 
- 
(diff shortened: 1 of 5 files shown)


mrbobbytables:
Should those that have access to the twitter account be included on the list as well?

parispittman:
@mrbobbytables twitter is managed by CNCF and we don't have access (contribex or those not employed by CNCF)

cblecker:
It might be worth noting that the GitHub moderators are https://github.com/kubernetes/community/tree/master/github-management#github-administration-team. If there is moderation/CoC action needed, then it would be that owners list who would take the action.

Issue #2476, opened 2018-08-04
kubernetes-sigs/architecture-tracking repo needs to be added to sigs.yaml and have an OWNERS file


spiffxp:
ref: #2464
ref: #1721

Ownership of this code must fall on a single sig.  In this case it's sig-architecture.  The repo is also lacking OWNERS files.

I recognize this repo is pretty much just for the project boards, but it's got two files in it and some open PR's.

/sig architecture
@kubernetes/sig-architecture-misc-use-only-as-a-last-resort 

/assign @jdumars

/area github-management

jdumars:
https://github.com/kubernetes-sigs/architecture-tracking/pull/46 is in and I'll add the info to sigs.yaml when this gets merged.

jdumars:
This work is complete.

PR #2499, opened 2018-08-08
added architecture tracking to sigs.yaml

jdumars:
Reference: https://github.com/kubernetes/community/issues/2476


bgrant0607:
Thanks. All of these subprojects are totally bogus, but this at least puts the link there.

/lgtm
/approve

PR #2502, opened 2018-08-08
committee-steering: remove sigs.yml link

philips:
A sigs.yml link isn't very helpful because the line numbers will keep
changing. Instead suggest people link to their subprojects in the SIG
README which is generated from the sigs.yml.

### DIFF
--- a/committee-steering/governance/sig-charter-template.md
+++ b/committee-steering/governance/sig-charter-template.md
@@ -11,8 +11,6 @@ necessarily all of the internals.
 
 ### In scope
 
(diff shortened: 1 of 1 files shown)


bgrant0607:
I noticed the same. Thanks.
/lgtm
/approve

spiffxp:
Not sure why tide is showing this as not mergeable...

philips:
@spiffxp reconciling...

cblecker:
@philips this isn't merging because https://github.com/kubernetes/test-infra/issues/9001 :(

PR #2514, opened 2018-08-10
[WIP] Proposal for CSI Cluster Registry

saad-ali:
This proposal introduces a mechanism for CSI volume plugins to register themselves with Kubernetes and  configure how Kubernetes interacts with them.

Still a work in progress, but feel free to start commenting.

/assign @jsafrane @thockin

### DIFF
--- a/contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
+++ b/contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
@@ -1,0 +1,145 @@
+# CSI Cluster Registry Design Doc
+
+***Status:*** Pending
+
+***Version:*** Alpha
+
+***Author:*** Saad Ali ([@saad-ali](https://github.com/saad-ali), saadali@google.com)
+
+## Terminology
+
+Term | Definition
+---|---
+Container Storage Interface (CSI) | A specification attempting to establish an industry standard interface that Container Orchestration Systems (COs) can use to expose arbitrary storage systems to their containerized workloads.
+CSI Volume Plugin | A new, in-tree volume plugin that acts as an adapter and enables out-of-tree, third-party CSI volume drivers to be used in Kubernetes.
+CSI Volume Driver | An out-of-tree CSI compatible implementation of a volume plugin that can be used in Kubernetes through the Kubernetes CSI Volume Plugin.
+
+# Summary
+
+This document proposes a "CSI cluster registration mechanism".
+
+This mechanism will enable:
+* A CSI volume plugin to register itself with Kubernetes when it is deployed.
+* A CSI volume plugin to customize how Kubernetes interacts with it (e.g. skip attach process because the plugin doesn't support ControllerPublish, etc.).
+* A user or cluster admin to easily discover which CSI volume plugins are deployed on their Kubernetes cluster.
+* Be optional -- a plugin may choose not to use it (and will get the default set of behaviors).
+
+
+## Background & Motivations
+
+Kubernetes supports the Container Storage Interface (CSI) to enable third party storage developers to deploy volume plugins exposing new storage systems in Kubernetes without having to touch the core Kubernetes code.
+Support for CSI was introduced as alpha in Kubernetes v1.9, and moved to beta in v1.10.
+See "CSI Volume Plugins in Kubernetes Design Doc" in the "Links" section below for details on how an arbitrary CSI volume plugin is deployed and interacts with Kubernetes.
+The beta implementation of CSI has a number of limitations, including:
+* CSI drivers must be deployed with the provided CSI external-attacher sidecar container, even if they don’t implement ControllerPublishVolume.
+  * Meaning a CSI volume plugin deployed on kubernetes has to deploy a CSI `external-attacher` container even if the volume plugin doesn't require a "volume attach" operation.
+  * In this case, the CSI `external-attacher` container basically does a no-op in response to Kubernetes `VolumeAttachment` objects to allow Kubernetes to continue with the mounting process.
+* Users and cluster admins have no easy way to discover what CSI volume plugins are deployed on their Kubernetes cluster.
+
+The proposed "CSI cluster registration mechanism" should address these issues.
+
+### Links
+
+* [Container Storage Interface (CSI) Spec](https://github.com/container-storage-interface/spec/blob/master/spec.md)
+* [CSI Volume Plugins in Kubernetes Design Doc](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md)
+
+## Objective
+
+The objective of this document is to document all the requirements for enabling a cluster level registry for CSI compliant volume plugins (a CSI volume driver) in Kubernetes.
+
+## Goals
+
+* Allow a CSI volume plugin (that opts-in to new mechanism) to configure how Kubernetes should interact with it.
+* Improve discoverability by users/cluster admins of a CSI volume plugin (that opts-in to new mechanism) deployed on the Kubernetes cluster.
+
+## Non-Goals
+
+* Define how CSI volume plugin should be deployed on Kubernetes.
+* Require use of new CSI cluster registry mechanism for all CSI volume plugins deployed on Kubernetes (the mechanism will be optional/opt-in).
+
+## Design Overview
+
+A new custom resource will be automatically be installed on Kubernetes clusters.
+
+Upon deployment a plugin must create a new custom resource object.
+
+## Design Details
+
+### `CSIPlugin` Object
+
+#### Proposed API
+
+```go
+// CSIPlugin captures information about a Container Storage Interface (CSI)
+// volume plugin deployed on the cluster.
+//
+// CSIPlugin objects are non-namespaced.
+type CSIPlugin struct {
+    metav1.TypeMeta   `json:",inline"`
+
+	// Standard object metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+    metav1.ObjectMeta `json:"metadata"`
+
+    // Specification describing the CSI volume plugin and any custom
+    // configuration for it.
+	Spec CSIPluginSpec `json:"spec"`
+
+	// Status of the CSI volume plugin.
+	Status CSIPluginStatus `json:"status,omitempty"`
+}
+
+// CSIPluginList is a collection of CSIPlugin objects.
+type CSIPluginList struct {
+    metav1.TypeMeta `json:",inline"`
+    
+	// Standard list metadata
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Items is the list of CSIPlugins
+	Items []CSIPlugin `json:"items"`
+}
+
+// CSIPluginSpec is the specification of a CSIPlugin.
+type CSIPluginSpec struct {
+    // Driver indicates the name of the CSI Plugin that this object refers to.
+    // This MUST be the same name returned by the CSI GetPluginName() call for
+    // that driver.
+    Driver string `json:"driver"`
+    
+    // Indicates this CSI volume plugin does not require an attach operation
+    // (because it does not implement the ControllerPublishVolume() method),
+    // and that Kubernetes should not call or wait for any attach operation and
+    // just skip to mounting.
+	SkipAttach bool `json:"skipAttach"`
+}
+
+// CSIPluginStatus is the status of a CSIPlugin.
+type CSIPluginStatus struct {
+    // Indicates the volume plugin has been successfully deployed on the cluster
+    // and is ready to use.
+	Ready bool `json:"ready"`
+}
+
+```
+
+#### CRD Installation
+
+The `CSIPlugin` object schema will checked in to kubernetes/kubernetes repo under a storage directory.
+The `CSIPlugin` `CustomResourceDefinition` (CRD) will be installed by the exiting Kubernetes attach/detach controller that is part of the kube-controller-manager binary.
+The controller will periodically verify that the CRD is still installed, and recreate it, if it is not.
+
+#### CR Creation
+
+When a CSI volume plugin is deployed on Kubernetes, it may optionally, register it self with Kubernetes by creating a new `CSIPlugin` object.
+The Kubernetes team will modify the CSI [`driver-registrar` container](https://github.com/kubernetes-csi/driver-registrar) to automatically do this.
+The plugin must set the `Driver` field to the same name returned by the CSI `GetPluginName()` call for that driver.
+The plugin may set any optional configuration fields (like `SkipAttach`) as appropriate.
+When the plugin is ready to serve, it must set `Ready` in the status to `true`.
+
+#### Upgrade/Downgrade
+TODO
+
+## Alternatives Considered
+TODO


contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,145 @@
...
+The `CSIPlugin` `CustomResourceDefinition` (CRD) will be installed by the exiting Kubernetes attach/detach controller that is part of the kube-controller-manager binary.
+The controller will periodically verify that the CRD is still installed, and recreate it, if it is not.
+
+#### CR Creation
+
+When a CSI volume plugin is deployed on Kubernetes, it may optionally, register it self with Kubernetes by creating a new `CSIPlugin` object.
+The Kubernetes team will modify the CSI [`driver-registrar` container](https://github.com/kubernetes-csi/driver-registrar) to automatically do this.
+The plugin must set the `Driver` field to the same name returned by the CSI `GetPluginName()` call for that driver.
+The plugin may set any optional configuration fields (like `SkipAttach`) as appropriate.
+When the plugin is ready to serve, it must set `Ready` in the status to `true`.
```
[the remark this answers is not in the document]
sbezverk: @jsafrane If CSIPlugin object had something like a `Registered` field which is set by the kubelet's informer, then driver-registrar could watch for it and then set CSIPlugin as `Ready` once it has been seen. In this case there is no any race.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,145 @@
...
+    // Driver indicates the name of the CSI Plugin that this object refers to.
+    // This MUST be the same name returned by the CSI GetPluginName() call for
+    // that driver.
+    Driver string `json:"driver"`
+    
+    // Indicates this CSI volume plugin does not require an attach operation
+    // (because it does not implement the ControllerPublishVolume() method),
+    // and that Kubernetes should not call or wait for any attach operation and
+    // just skip to mounting.
+	SkipAttach bool `json:"skipAttach"`
```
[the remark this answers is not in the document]
saad-ali: Agree with the renaming.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,145 @@
...
+	// Standard list metadata
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Items is the list of CSIPlugins
+	Items []CSIPlugin `json:"items"`
+}
+
+// CSIPluginSpec is the specification of a CSIPlugin.
+type CSIPluginSpec struct {
```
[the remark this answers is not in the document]
saad-ali: Done.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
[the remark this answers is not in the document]
saad-ali: Good question. It would have to be the kubelet.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,145 @@
...
+    // and is ready to use.
+	Ready bool `json:"ready"`
+}
+
+```
+
+#### CRD Installation
+
+The `CSIPlugin` object schema will checked in to kubernetes/kubernetes repo under a storage directory.
+The `CSIPlugin` `CustomResourceDefinition` (CRD) will be installed by the exiting Kubernetes attach/detach controller that is part of the kube-controller-manager binary.
```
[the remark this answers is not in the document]
saad-ali: I agree. But it would be easier to use an existing controller for the alpha version.

I changed it to say new controller -- if we have enough time, I'll do new controller, otherwise I'll just do existing controller and we can add in a standalone controller later for beta.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,145 @@
...
+    // and that Kubernetes should not call or wait for any attach operation and
+    // just skip to mounting.
+	SkipAttach bool `json:"skipAttach"`
+}
+
+// CSIPluginStatus is the status of a CSIPlugin.
+type CSIPluginStatus struct {
+    // Indicates the volume plugin has been successfully deployed on the cluster
+    // and is ready to use.
+	Ready bool `json:"ready"`
```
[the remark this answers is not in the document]
saad-ali: For the initial version no consumers. In the future we can use this as one of the bits for smarter CSI volume aware scheduling.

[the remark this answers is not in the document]
saad-ali: I'm hesitant on expanding the API more for a hypothetical use case. I'd prefer to leave it simple or drop it altogether until we have a real use case. If I drop it, there would be nothing in `CSIPluginStatus` so I would have to drop that for now too. What do you prefer?

saad-ali:
@jsafrane  Addressed feedback, PTAL

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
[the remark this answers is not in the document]
jsafrane: I am torn on this. Currently I don't see any useful Status to report. But I may come in future. Can we have an object with .Spec and no .Status?

If we use `Ready`, we need clear semantics what it means and most importantly what is it good for. Is it enough to have one node with CSI driver installed to have the global object ready? Is it enough to have one node with CSI driver broken (failed liveness probe, driver not installed there, crash loop...) to have the global object not-ready?

>  In the future we can use this as one of the bits for smarter CSI volume aware scheduling.

For smarter CSI volume aware scheduling I would use decision based on availability / status of drivers on nodes, not a global object. What should scheduler do when the global object is not ready? All it can do is to stop scheduling.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,158 @@
...
+
+```go
+// CSIDriver captures information about a Container Storage Interface (CSI)
+// volume driver deployed on the cluster.
+//
+// CSIDriver objects are non-namespaced.
+type CSIDriver struct {
+    metav1.TypeMeta   `json:",inline"`
+
+	// Standard object metadata.
```
jsafrane: wrong indentation

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,158 @@
...
+* Improve discoverability by users/cluster admins of a CSI volume driver (that opts-in to new mechanism) deployed on the Kubernetes cluster.
+
+## Non-Goals
+
+* Define how CSI volume driver should be deployed on Kubernetes.
+* Require use of new CSI cluster registry mechanism for all CSI volume drivers deployed on Kubernetes (the mechanism will be optional/opt-in).
+
+## Design Overview
+
+A new custom resource will be automatically be installed on Kubernetes clusters.
```
jsafrane: a new custom resource **definition**? CR is optional.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,158 @@
...
+
+Upon deployment a driver must create a new custom resource object.
```
jsafrane: Driver may create a new object? It's optional.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,158 @@
...
+    // will not be passed on mount. 
+    // +optional
+    PodInfoRequiredOnMount *bool `json:"podInfoRequiredOnMount"`
+}
+
+// CSIDriverStatus is the status of a CSIDriver.
+type CSIDriverStatus struct {
+    // Indicates the volume driver has been successfully deployed on the cluster
+    // and is ready to use.
+	Ready bool `json:"ready"`
```
jsafrane: It's still unclear to me what "Ready" means.  "the volume driver has been successfully deployed" - so all containers (DaemonSet with node components and Deployment with controller components) are running? What one of them fails, who sets `Ready=false`? What if I add a new node, shall something set it to `Ready=false`, wait until the DaemonSet runs a new pod with node components there and set it to `Ready=true`?

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,158 @@
...
+The `CSIDriver` object schema will checked in to kubernetes/kubernetes repo under a storage directory.
+The `CSIDriver` `CustomResourceDefinition` (CRD) will be installed by a new Kubernetes controller that is responsible for ensuring required CRDs are installed.
+The controller will periodically verify that the CRD is still installed, and recreate it, if it is not.
+
+#### CR Creation
+
+When a CSI volume driver is deployed on Kubernetes, it may optionally, register it self with Kubernetes by creating a new `CSIDriver` object.
+The Kubernetes team will modify the CSI [`driver-registrar` container](https://github.com/kubernetes-csi/driver-registrar) to automatically do this.
+If kubelet driver registration is enabled, the kubelet will automatically do this as part of plugin registration.
+The driver must set the `Driver` field to the same name returned by the CSI `GetPluginName()` call for that driver.
```
gnufied: So does this mean that a CSI driver has to ship with some mechanism of instantiating CRs when it is deployed? Currently obviously that is not part of CSI spec, so how will that get done?

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,158 @@
...
+// CSIDriver captures information about a Container Storage Interface (CSI)
+// volume driver deployed on the cluster.
+//
+// CSIDriver objects are non-namespaced.
+type CSIDriver struct {
+    metav1.TypeMeta   `json:",inline"`
+
+	// Standard object metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+    metav1.ObjectMeta `json:"metadata"`
```
jsafrane: Shall we require that ObjectMeta.Name is the CSI driver name? How else can we ensure that each CSI driver has just one CSIDriver instance?

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
saad-ali: Fixed

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
saad-ali: Fixed

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
saad-ali: Fixed

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
saad-ali: Will remove

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
saad-ali: > So does this mean that a CSI driver has to ship with some mechanism of instantiating CRs when it is deployed?

CSI driver authors can choose to but don't have to

> Currently obviously that is not part of CSI spec, so how will that get done?

The same way all custom k8s behavior is done, side car containers provided by us (e.g. driver-registrar, external-attacher, etc.).

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
saad-ali: I don't think we should require it. We should have validation webhook to verify that you can't create two objects with the same `CSIDriverSpec.Driver` fields.

saad-ali:
@thockin @kubernetes/api-approvers PTAL
@kubernetes/sig-storage-api-reviews is happy with this.
I'm going to start implementing this to unblock other designs pending on this.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
jsafrane: This validation is IMO racy. There may be several API servers running in parallel.

1. API server A gets POST CSIDriver.name=X for a driver.
2. API server B gets POST CSIDriver.name=Y for the same driver at the same time.
3. Both call validation webhook.
4. The webhook gets two queries, possibly in parallel. Let's say the webhook synchronizes the requests and processes X fist.
5. The webhook can GET CSIDriver to check if it already exists, but in this case it does not return anything useful, X and Y were not created yet.
6. The webhook allows X to be created. It remembers that X exists in local cache.
7. The webhook processes Y, it can check in the API server if X exists. It does not exist yet, either the API server is slow or it was rejected by another webhook / validation.
8. The webhook can check internal cache if it allowed X. It was allowed, so it rejects Y.

And this is wrong, because X might have been rejected by API server (e.g. another webhook failed) and Y should have been allowed.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -152,7 +142,15 @@ The driver may set any optional configuration fields (like `SkipAttach`) as appr
 When the driver is ready to serve, it must set `Ready` in the status to `true`.
 
 #### Upgrade/Downgrade
-TODO
+This change is backwards compatible.
+Existing CSI drivers that are already deployed will not create a `CSIDriver` object.
+And Kubernetes will continue to interact with those drivers as it does today.
+New drivers MAY create a `CSIDriver` object to change how Kubernetes interacts with them.
+
+If a Kubernetes cluster is downgraded to a version that does not support `CSIDriver`, but continues to have a CSI driver deployed on it that creates a `CSIDriver` object and expects non-default behavior, Kubernetes will not be able to interact with it correctly (e.g. it may call attach when driver requests attach not to be called resulting in the volumes not mounting). Therefore, cluster admins must ensure if they downgrade, that they also change the CSI driver, if needed.
```
jsafrane: It's very likely that the driver will be the same in this case. If I have Gluster CSI driver working in 1.11 with external attacher, I can use the same driver in 1.12 with the external attacher or I may create CSIDriver for it and stop the attacher. The driver itself is the same in all three cases, PVs are the same and pod definitions are the same.

Downgrade from a non-attachable driver with a CSIDriver instance to 1.11 is not that easy as you describe, some volumes can be mounted to running pods and we need to make sure they remain mounted after downgrade and unmounted when the pod is deleted. Let's discuss in #2523

kfox1111:
looks good to me.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,156 @@
...
+#### CRD Installation
+
+The `CSIDriver` object schema will checked in to kubernetes/kubernetes repo under a storage directory.
+The `CSIDriver` `CustomResourceDefinition` (CRD) will be installed by a new Kubernetes controller that is responsible for ensuring required CRDs are installed.
+The controller will periodically verify that the CRD is still installed, and recreate it, if it is not.
+
+#### CR Creation
+
+When a CSI volume driver is deployed on Kubernetes, it may optionally, register it self with Kubernetes by creating a new `CSIDriver` object.
+The Kubernetes team will modify the CSI [`driver-registrar` container](https://github.com/kubernetes-csi/driver-registrar) to automatically do this.
```
kfox1111: hmm.... Thinking about this some more, with my operator hat on, I'd rather it not. If the driver-registrar were to do it, then we'd have to rbac it well in order to restrict its access to a safe subset of things. but if you are going through that much effort, you could instead just upload the crd directly in whatever csi packaging is provided, like via k8s yaml / helm chart to the cluster using your admin credentials instead and not worry about it at all. Similar to some of the reasoning Rook took when they tightened security (https://github.com/rook/rook/blob/master/design/security-model.md for details)

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,156 @@
...
+    
+    // Indicates this CSI volume driver requires additional pod information
+    // (like podName, podUID, etc.) during mount operations.
+    // If this is set to true, Kubelet will pass pod information as
+    // VolumeAttributes in the CSI NodePublishVolume() calls.
+    // If value is not specified, default is false -- meaning pod information
+    // will not be passed on mount. 
+    // +optional
+    PodInfoRequiredOnMount *bool `json:"podInfoRequiredOnMount"`
+}
```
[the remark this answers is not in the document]
jsafrane: Should this be limited only to inline volumes? Or handles for PVs should be (optionally) autogenerated too?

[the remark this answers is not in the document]
kfox1111: hmm... pv's too.... say someone used the image driver with a pv (for some reason). It would break horribly if it wasn't given proper handles.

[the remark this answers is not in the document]
vladimirvivien: I think for now the language and implementation should be limited to Inline.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,156 @@
...
+    // If value is not specified, default is true -- meaning attach will be
+    // called.
+    // +optional
+    AttachRequired *bool `json:"attachRequired"`
+    
+    // Indicates this CSI volume driver requires additional pod information
+    // (like podName, podUID, etc.) during mount operations.
+    // If this is set to true, Kubelet will pass pod information as
+    // VolumeAttributes in the CSI NodePublishVolume() calls.
+    // If value is not specified, default is false -- meaning pod information
```
thockin: Why not pass this always?

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,156 @@
...
+    // Driver indicates the name of the CSI driver that this object refers to.
+    // This MUST be the same name returned by the CSI GetPluginName() call for
+    // that driver.
+    Driver string `json:"driver"`
+    
+    // Indicates this CSI volume driver requires an attach operation (because it
+    // implements the CSI ControllerPublishVolume() method), and that Kubernetes
+    // should call attach and wait for any attach operation to complete before
+    // proceeding to mounting.
+    // If value is not specified, default is true -- meaning attach will be
```
thockin: Is this a default as in the normal defaults (will be set in the object) or is it the implied behavior?  To set a default will require admission control...  Just clarify.

contributors/design-proposals/storage/container-storage-interface-cluster-registry.md
```diff
@@ -0,0 +1,156 @@
...
+    // called.
+    // +optional
+    AttachRequired *bool `json:"attachRequired"`
```
jqmichael, line 115: Probably a dump question. 

Isn't this value already exposed from `ControllerServiceCapability` as `PUBLISH_UNPUBLISH_VOLUME`? Can't kubelet make decision based on that value?

kfox1111: The problem is where the call is being made. Kubelet is not doing the attach. A pod floating in the cluster takes attachment requests, converts them into the driver calls for attach/detach.

You could have the attacher sidecar running, probe the driver as you say, and then determine that it doesn't ever have to do anything. That does work today.

But that's a fair amount of extra pluming running in the cluster that isn't needed.

The flag above tells Kubernetes to skip entirely any of the handshaking at the cluster level required to do that attach so the sidecar that does nothing and its driver doesn't even need to exist.

PR #2519, opened 2018-08-11
Add kubernetes-csi/* repos to sigs.yaml

### LINKED ISSUE #2467: kubernetes-csi/* repos need to be added to sigs.yaml and have OWNERS files added
Ownership of the code each of these repos needs to ultimately fall to a single sig.  In this case it seems like it should be sig-storage, which I confirmed with @saad-ali.  Adding them to sigs.yaml will do that.

They also need OWNERS files.  Eventually we'll want to turn on automation to consume them.  For right now they will serve as a reference of who owns the repos.

It also looks like these are all inter-related, so maybe they fall under a single subproject instead of a subproject per-repo. eg: [sig-api-machinery's kubernetes-clients subproject](https://github.com/kubernetes/community/blob/master/sigs.yaml#L95-L108)

/sig storage
@kubernetes/sig-storage-misc 

/assign @saad-ali 

ref: https://github.com/kubernetes/community/issues/1721
ref: https://github.com/kubernetes/community/issues/2464

saad-ali:
Closes #2467

/assign @spiffxp

### DIFF
--- a/sig-storage/README.md
+++ b/sig-storage/README.md
@@ -31,6 +31,17 @@ The Chairs of the SIG run operations and processes governing the SIG.
 ## Subprojects
 
 The following subprojects are owned by sig-storage:
+- **kubernetes-csi**
+  - Owners:
+    - https://raw.githubusercontent.com/kubernetes-csi/csi-test/master/OWNERS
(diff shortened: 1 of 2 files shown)


spiffxp:
/lgtm
/approve
Thanks for moving on this!

PR #2523, opened 2018-08-13
CSI: Skip attach/detach for non-attachable volumes

jsafrane:
In this proposal, we describe a way how to skip creation of `VolumeAttachments` for non-attachable CSI drivers.

Benefits:

* Admins don't need to deploy external attacher for non-attachable CSI drivers. It was very confusing to them, they could not understand why they should run an attacher for non-attachable drivers. Now they won't need to do it.
* Somewhat faster pod startup

This PR depends on ~~`CSIPlugin`~~ `CSIDriver` introduced in https://github.com/kubernetes/community/pull/2514

### DIFF
--- a/contributors/design-proposals/storage/container-storage-interface-skip-attach.md
+++ b/contributors/design-proposals/storage/container-storage-interface-skip-attach.md
@@ -14,24 +14,23 @@ We want to skip creation of `VolumeAttachment` objects in A/D controller for CSI
 In order to skip both A/D controller attaching a volume and kubelet waiting for the attachment, both of them need to know if a particular CSI driver is attachable or not. In this document we expect that proposal #2514 is implemented and both A/D controller and kubelet has informer on `CSIDriver` so they can check if a volume is attachable easily.
 
 ## Design
-### Volume plugins
-
-#### CSI volume plugin
-* Rework [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/43f805b7bdda7a5b491d34611f85c249a63d7f97/pkg/volume/csi/csi_plugin.go#L58) to accept informer that watches CSIDriver. The plugin will store the informer for later.
+### CSI volume plugin
+* Rework [`Init`](https://github.com/kubernetes/kubernetes/blob/43f805b7bdda7a5b491d34611f85c249a63d7f97/pkg/volume/csi/csi_plugin.go#L58) to get or create informer to cache CSIDriver instances.
+  * Depending on where the API for CSIDriver ends up, we may:
+    * Rework VolumeHost to either provide the informer. This leaks CSI implementation details to A/D controller and kubelet
+    * Or the CSI volume plugin can create and run CSIDriver informer by itself. No other component in controller-manager or kubelet needs the informer right now, so a non-shared informer is viable option. Depending on when the API for CSIDriver ends up, `VolumeHost` may need to be extended to provide client interface to the API and kubelet and A/D controller may need to be updated to create the interface (somewhere in `cmd/`, where RESTConfig is still available to create new clients ) and pass it to their `VolumeHost` implementations.
 * Rework `Attach`, `Detach`, `VolumesAreAttached` and `WaitForAttach` to check for `CSIDriver` instance using the informer.
 	* If CSIDriver for the driver exists and it's attachable, perform usual logic.
-	* If CSIDriver for the driver exists and it's not attachable, return success immediately (basically NOOP).
+	* If CSIDriver for the driver exists and it's not attachable, return success immediately (basically NOOP). A/D controller will still mark the volume as attached in `Node.Status.VolumesAttached`.
 	* If CSIDriver for the driver does not exist, perform usual logic (i.e. treat the volume as attachable).
 	  * This keeps the behavior the same as in old Kubernetes version without CSIDriver object.
 	  * This also happens when CSIDriver informer has not been quick enough. It is suggested that CSIDriver instance is created **before** any pod that uses corresponding CSI driver can run.
 	    * In case that CSIDriver informer (or user) is too slow, CSI volume plugin `Attach()` will create `VolumeAttachment` instance and wait for (non-existing) external attacher to fulfill it. The CSI plugin shall recover when `CSIDriver` instance is created and skip attach. Any `VolumeAttachment` instance created here will be deleted on `Detach()`, see the next bullet.
 * In addition to the above, `Detach()` removes `VolumeAttachment` instance even if the volume is not attachable. This deletes `VolumeAttachment` instances created by old A/D controller or before `CSIDriver` instance was created.
 
-### A/D controller
-* A/D controller must pass a (shared) informer to watch CSIDriver and pass it to CSI volume plugin in [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/8db5328c4c1f9467ab0d70ccb991a12d4675b6a7/cmd/kube-controller-manager/app/plugins.go#L82).
 
-### Kubelet / VolumeManager
-* Kubelet must create a (shared) informer to watch CSIDriver and pass it to CSI volume plugin in [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/8db5328c4c1f9467ab0d70ccb991a12d4675b6a7/cmd/kubelet/app/plugins.go#L101).
+### Authorization
+* A/D controller and kubelet must be allowed to list+watch CSIDriver instances. Updating RBAC rules should be enough.
 
 ## API
 No API changes.
@@ -49,7 +48,7 @@ For non-attachable volumes, if the volume was attached by "old" Kubernetes (or "
 ## Downgrade
 This chapter covers:
 * Downgrade from new Kubernetes that has `CSISkipAttach` enabled to old Kubernetes with `CSISkipAttach disabled.
-* Update from Kubernetes that has `CSISkipAttach` enabled to the same Kubernetes with `CSISkipAttach` disabled.
+* Update from Kubernetes that has `CSISkipAttach` feature enabled to the same Kubernetes with `CSISkipAttach` disabled.
 * Deletion of CSIDriver instance with non-attachable CSI driver.
 
 In all cases listed above, a non-attachable CSI driver becomes "attachable" (i.e. requires external attacher).  Downgrade does not affect attachable CSI drivers, both "old" and "new" Kubernetes processes them in the same way.
@@ -75,7 +74,7 @@ Expected timeline:
 
 ## Alternatives considered
 A/D controller and kubelet can be easily extended to check if a given volume is attachable. This would make mounting of non-attachable volumes easier, as kubelet would not need to wait for A/D controller to mark the volume as attached. However, there would be issues when upgrading or downgrading Kubernetes (or marking CSIDriver as attachable or non-attachable, which has basically the same handling).
-* On upgrade (i.e. a previously attachable CSI volume becomes non-attachable), A/D controller could discover that an attached volume is not attachable any longer. A/D controller could clean up `Node.Status.VolumesAttached`, but since A/D controller does not know anything about `VolumeAttachment`, we would either need to introduce a new volume plugin call to clean it up in CSI volume plugin, or something else would need to clean it.
-* On downgrade (i.e. a previously non-attachable CSI volume becomes attachable), kubelet must discover that already mounted volume has changed from non-attachable to attachable and put it into `Node.Status.VolumesInUse`. This would race with A/D controller detaching the volume when a pod was deleted at the same time a CSIDriver instance was made attachable.
+* On upgrade (i.e. a previously attachable CSI volume becomes non-attachable, e.g. when user creates CSIDriver instance while corresponding CSI driver is already running), A/D controller could discover that an attached volume is not attachable any longer. A/D controller could clean up `Node.Status.VolumesAttached`, but since A/D controller does not know anything about `VolumeAttachment`, we would either need to introduce a new volume plugin call to clean it up in CSI volume plugin, or something else would need to clean it.
+* On downgrade (i.e. a previously non-attachable CSI volume becomes attachable, e.g. when user deletes CSIDriver instance or downgrades to old Kubernetes without this feature), kubelet must discover that already mounted volume has changed from non-attachable to attachable and put it into `Node.Status.VolumesInUse`. This would race with A/D controller detaching the volume when a pod was deleted at the same time a CSIDriver instance was made attachable.
 
 Passing all volumes through A/D controller saves us from these difficulties and even races.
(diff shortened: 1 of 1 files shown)


jsafrane:
@saad-ali @vladimirvivien @gnufied @jingxu97 @msau42 @sbezverk, PTAL

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Downgrade
+This chapter covers:
+* Downgrade from new Kubernetes that has `CSISkipAttach` enabled to old Kubernetes with `CSISkipAttach disabled.
+* Update from Kubernetes that has `CSISkipAttach` enabled to the same Kubernetes with `CSISkipAttach` disabled.
+* Deletion of CSIPlugin instance with non-attachable CSI driver.
+
+In all cases listed above, a non-attachable CSI driver becomes "attachable" (i.e. requires external attacher).  Downgrade does not affect attachable CSI drivers, both "old" and "new" Kubernetes processes them in the same way.
+
+For non-attachable volumes, if the volume was mounted by "new" Kubernetes, it has no VolumeAttachment instance. "Old" A/D controller does not know about it. However, it will periodically call plugin's `VolumesAreAttached()` that checks for `VolumeAttachment` presence. Volumes without `VolumeAttachment` will be reported as not attached and A/D controller will call `Attach()` on these.
```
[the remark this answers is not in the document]
jsafrane: If you downgrade Kubernetes (or delete CSIDriver instance or disable the alpha feature), you're required to run an external CSI attacher for non-attachable CSI drivers, just like before this proposal.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+## Performance considerations
+
+* Flow suggested in this proposal adds new `CSIPlugin` informer both to A/D controller and kubelet. We don't expect any high amount of instances of `CSIPlugin` nor any high frequency of updates. `CSIPlugin` should have negligible impact on performance.
+
+* A/D controller will not create `VolumeAttachment` instances for non-attachable volumes. Etcd load will be reduced.
+
+* On the other hand, all CSI volumes still must go though A/D controller. A/D controller **must** process every CSI volume and kubelet **must** wait until A/D controller marks a volume as attached, even if A/D controller basically does nothing. All CSI volumes must be added to `Node.Status.VolumesInUse` and `Node.Status.VolumesAttached`. This does not introduce any new API calls, all this is already implemented, however this proposal won't reduce `Node.Status` update frequency in any way. If *all* volumes will move to CSI eventually, pod startup will be slower than when using in-tree volume plugins that don't go through A/D controller.
```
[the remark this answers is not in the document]
jsafrane: Added a note about it.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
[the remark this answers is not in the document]
jsafrane: Added a note about external attacher.

jsafrane:
Fixed the comments. And renamed `CSIPlugin` to `CSIDriver`, following updates in  #2514

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Upgrade
+This chapter covers:
+* Upgrade from old Kubernetes that has `CSISkipAttach` disabled to new Kubernetes with `CSISkipAttach` enabled.
+* Update from Kubernetes that has `CSISkipAttach` disabled to the same Kubernetes with `CSISkipAttach` enabled.
+* Creation of CSIDriver instance with non-attachable CSI driver.
+
+In all cases listed above, an "attachable" CSI driver becomes non-attachable. Upgrade does not affect attachable CSI drivers, both "old" and "new" Kubernetes processes them in the same way.
+
+For non-attachable volumes, if the volume was attached by "old" Kubernetes (or "new" Kubernetes before CSIDriver instance was created), it has `VolumeAttachment` instance. It will be deleted by `Detach()`, as it deletes `VolumeAttachment` instance also for non-attachable volumes.
```
gnufied: Is there a problem when controller-manager is newer but kubelet is older? The controller-manager will not create `VolumeAttachment` object but kubelet will be stuck waiting for `WaitForAttach` ?

jsafrane: Admin should not enable the alpha feature in this case, I think. Can we solve it by documentation?

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+	    * In case that CSIDriver informer (or user) is too slow, CSI volume plugin `Attach()` will create `VolumeAttachment` instance and wait for (non-existing) external attacher to fulfill it. The CSI plugin shall recover when `CSIDriver` instance is created and skip attach. Any `VolumeAttachment` instance created here will be deleted on `Detach()`, see the next bullet.
+* In addition to the above, `Detach()` removes `VolumeAttachment` instance even if the volume is not attachable. This deletes `VolumeAttachment` instances created by old A/D controller or before `CSIDriver` instance was created.
+
+### A/D controller
+* A/D controller must pass a (shared) informer to watch CSIDriver and pass it to CSI volume plugin in [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/8db5328c4c1f9467ab0d70ccb991a12d4675b6a7/cmd/kube-controller-manager/app/plugins.go#L82).
+
+### Kubelet / VolumeManager
+* Kubelet must create a (shared) informer to watch CSIDriver and pass it to CSI volume plugin in [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/8db5328c4c1f9467ab0d70ccb991a12d4675b6a7/cmd/kubelet/app/plugins.go#L101).
+
```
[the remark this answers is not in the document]
liggitt: and NodeAuthorizer

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
msau42: What about for beta/GA?  The standard upgrade procedure is to update master first

msau42: Or is the requirement to not update your drivers until both master+nodes are beyond some version?

jsafrane: I think that whatever creates CSIDriver instances should create it only when both A/D controller and kubelet understand it. I.e. they're both at the right version and they have necessary alpha features enabled.

On downgrade, A/D controller gets updated first and it starts creating VolumeAttachment instances. It's up to the admin to start external attacher in this case, because CSI required it at that old version (or version with disabled alpha features).

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+## Implementation
+
+Expected timeline:
+* Alpha: 1.12 (behind feature gate `CSISkipAttach`)
+* Beta: 1.13 (enabled by default)
+* GA: 1.14
+
+## Alternatives considered
+A/D controller and kubelet can be easily extended to check if a given volume is attachable. This would make mounting of non-attachable volumes easier, as kubelet would not need to wait for A/D controller to mark the volume as attached. However, there would be issues when upgrading or downgrading Kubernetes (or marking CSIDriver as attachable or non-attachable, which has basically the same handling).
+* On upgrade (i.e. a previously attachable CSI volume becomes non-attachable), A/D controller could discover that an attached volume is not attachable any longer. A/D controller could clean up `Node.Status.VolumesAttached`, but since A/D controller does not know anything about `VolumeAttachment`, we would either need to introduce a new volume plugin call to clean it up in CSI volume plugin, or something else would need to clean it.
```
vladimirvivien: Is this a realistic scenario ? Can a driver that was attachable becomes non-attachable?  That sounds like a misconfiguration.  Again, maybe the feature should be gated during alpha/beta forcing users to buy into the CSIDriver registry thing.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+* On downgrade (i.e. a previously non-attachable CSI volume becomes attachable), kubelet must discover that already mounted volume has changed from non-attachable to attachable and put it into `Node.Status.VolumesInUse`. This would race with A/D controller detaching the volume when a pod was deleted at the same time a CSIDriver instance was made attachable.
```
vladimirvivien: Same question here, is that possible?

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Implementation
+
+Expected timeline:
+* Alpha: 1.12 (behind feature gate `CSISkipAttach`)
+* Beta: 1.13 (enabled by default)
+* GA: 1.14
+
+## Alternatives considered
+A/D controller and kubelet can be easily extended to check if a given volume is attachable. This would make mounting of non-attachable volumes easier, as kubelet would not need to wait for A/D controller to mark the volume as attached. However, there would be issues when upgrading or downgrading Kubernetes (or marking CSIDriver as attachable or non-attachable, which has basically the same handling).
```
vladimirvivien: If this issue is introduced as `alpha` and `feature-gated`, that should help avoid misconfigurations right?  Meaning, this feature will only work when `enabled` along with a properly-configured CSIDriver object (and supporting code) is running.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+	  * This keeps the behavior the same as in old Kubernetes version without CSIDriver object.
+	  * This also happens when CSIDriver informer has not been quick enough. It is suggested that CSIDriver instance is created **before** any pod that uses corresponding CSI driver can run.
+	    * In case that CSIDriver informer (or user) is too slow, CSI volume plugin `Attach()` will create `VolumeAttachment` instance and wait for (non-existing) external attacher to fulfill it. The CSI plugin shall recover when `CSIDriver` instance is created and skip attach. Any `VolumeAttachment` instance created here will be deleted on `Detach()`, see the next bullet.
+* In addition to the above, `Detach()` removes `VolumeAttachment` instance even if the volume is not attachable. This deletes `VolumeAttachment` instances created by old A/D controller or before `CSIDriver` instance was created.
+
+### A/D controller
+* A/D controller must pass a (shared) informer to watch CSIDriver and pass it to CSI volume plugin in [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/8db5328c4c1f9467ab0d70ccb991a12d4675b6a7/cmd/kube-controller-manager/app/plugins.go#L82).
```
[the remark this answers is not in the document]
jsafrane: I'll rephrase it, it does not need to be a/d controller, it can be kube-controller-manager itself. And I tried to add a parameter with the informer to `ProbeVolumePlugins()` and it's too early. kubelet does not have fully established API client at that time. So I'll either create a new non-shared informer in `Init()` of the plugin or add it to `VolumeHost`.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Kubelet / VolumeManager
+* Kubelet must create a (shared) informer to watch CSIDriver and pass it to CSI volume plugin in [`ProbeVolumePlugins`](https://github.com/kubernetes/kubernetes/blob/8db5328c4c1f9467ab0d70ccb991a12d4675b6a7/cmd/kubelet/app/plugins.go#L101).
```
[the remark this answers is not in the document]
jsafrane: Yes, the informer will be in `csiPlugin` struct, so it's available for Mounter and Attacher. I need to figure out who creates the informer, see above.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
```diff
@@ -0,0 +1,81 @@
...
+* Rework `Attach`, `Detach`, `VolumesAreAttached` and `WaitForAttach` to check for `CSIDriver` instance using the informer.
+	* If CSIDriver for the driver exists and it's attachable, perform usual logic.
+	* If CSIDriver for the driver exists and it's not attachable, return success immediately (basically NOOP).
+	* If CSIDriver for the driver does not exist, perform usual logic (i.e. treat the volume as attachable).
+	  * This keeps the behavior the same as in old Kubernetes version without CSIDriver object.
+	  * This also happens when CSIDriver informer has not been quick enough. It is suggested that CSIDriver instance is created **before** any pod that uses corresponding CSI driver can run.
+	    * In case that CSIDriver informer (or user) is too slow, CSI volume plugin `Attach()` will create `VolumeAttachment` instance and wait for (non-existing) external attacher to fulfill it. The CSI plugin shall recover when `CSIDriver` instance is created and skip attach. Any `VolumeAttachment` instance created here will be deleted on `Detach()`, see the next bullet.
```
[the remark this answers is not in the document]
jsafrane: @saad-ali wants CSIDriver instances optional. So if admin decides that they don't want the instance for a driver, Kiubernetes must work. Admin can create the instance at any later time and kubernetes must carry on. Admin can even delete the instance and kubernetes and all running pods must survive.

I'd like to have the instance mandatory and immutable (users can't switch driver from attachable to non-attachable), but that's perhaps unrealistic requirement.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
jsafrane: > Can a driver that was attachable becomes non-attachable? 

Yes, admin creates / modifies CSIDriver for the driver.

Or admin enables alpha feature that gates the CSIDriver logic in Kubernetes. This one is particularly interesting, because there will be some (short?) time when the feature is enabled on some components and disabled on others.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
jsafrane: Yes, admin deletes CSIDriver instance. Or changes it to "attachable". Or disables the alpha feature. Or downgrades Kubernetes to version that does not implement CSIDriver logic. From our point of view (A/D controller + kubelet) , a driver that was not attachable becomes attachable

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
jsafrane: I am sure we need to support downgrade to older Kubernetes that did not have this feature. And IMO we should also support disabling the feature in a running cluster.

contributors/design-proposals/storage/container-storage-interface-skip-attach.md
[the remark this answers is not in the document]
jsafrane: Kubelet needs only read-only access to CSIDrivers, IMO we don't need to update NodeAuthorizer.

jsafrane:
Added some clarification about informer creation and upgrade/downgrade in "Alternatives considered".

jsafrane:
Is there anything missing in this proposal? I'm working on proof-of-concept and it looks good so far. Creating / deleting CSIDriver instances while pod with non-attachable CSI volumes (surprisingly ;-) ) works.

PR #2524, opened 2018-08-13
[KEP] storage api version migration

caesarxuchao:
This KEP proposes a tool that migrates the stored data in etcd to the latest storage version. The KEP is based on the [doc](https://docs.google.com/document/d/16HWEjqnlEc2PJnS4D7TVbHoNyVlyWpjx0zRsFNyxDsI/edit#heading=h.fap52hhsdgdt) which was discussed in the Aug. 1, 2018 sig apimachinery meeting. The KEP added more details on the tool's workflow and development timeline.

/sig api-machinery
/assign @lavalamp @deads2k


keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+      * [Graduation Criteria](#graduation-criteria)
+      * [Open questions](#open-questions)
+      * [Alternatives](#alternatives)
+         * [“Active management”](#active-management)
+         * [update-storage-objects.sh](#update-storage-objectssh)
+
+
+## Summary
+
+We propose to extend and improve the [oc adm migrate stroage][] command line
```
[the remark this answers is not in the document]
caesarxuchao: Done.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+and eventually arrive at a version of Kubernetes that can’t interpret the stored
+resource and crashes. See k8s.io/pr/52185."[1][]. We propose a tool that
+prevents this problem.
+
+[1]:https://docs.google.com/document/d/1eoS1K40HLMl4zUyw5pnC05dEF3mzFLp5TPEEt4PFvsM
+
+### Goals
+
+A successful storage version migration tool must:
+* only cause insignificant load to apiservers. For example, if the master has
```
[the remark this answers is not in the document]
caesarxuchao: Got it. The `oc adm migrate` already has rate limiting based on bandwidth, though I don't know whether bandwidth or qps is a better metric.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+  study how well the default 10 Mbps bandwidth limit in the oc command work).
+* work for big clusters that have ~10^5 instances of some resource types.
+* make progress in flaky environment, e.g., flaky apiservers, or the migration
+  process get preempted.
+* allow system administrators to track the migration progress.
+
+As to the deliverables,
+* in the short term, providing system administrators with a tool to migrate
+  the API objects to the proper storage versions. 
+* in the long term, automating the migration without further burdening
```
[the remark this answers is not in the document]
caesarxuchao: Done. Explicitly say the final goal is automating the migration for built-in APIs, CRD, and aggregated apis. But at alpha, CRD is ignored.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+
+We will create a "kube-migrator" command line tool. 
+
+### Workflow
+After the apiservers are upgraded, the cluster administrator launches
+"kube-migrator", which creates a deployment of "kube-migrator controller" in the
+cluster. "kube-migrator" uses the same kubeconfig file as kubectl does.
+
+The "kube-migrator controller" does the following:
+* creates a `progress` custom resource (CR).
```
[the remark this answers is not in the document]
caesarxuchao: Added `in the migration.k8s.io group`.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+* uses the discovery client to discover the supported APIs.
+* lists all CRDs (custom resource definitions).
+* stores the discovered resource types, excluding resources defined by CRDs, in
+  the "progress" CR.
```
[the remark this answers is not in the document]
caesarxuchao: In the alpha stage, the migration tool will be manually invoked, so I just say `the admin should invoke the tool after all the apiservers are at the desired version, including both kube apiservers and aggregated apiservers`

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+  Completed bool
+}
+
+type GroupVersion struct {
+  Group string
+  Version string
+}
+```
+
+The "migration worker" runs `oc adm migrate storage --include=<resource type>`
```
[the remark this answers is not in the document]
caesarxuchao: Changed to "runs the equivalence of `oc adm migrate storage --include=<resource type>`". I keep the old name to show how the existing oc tool fits in.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Failure recovery
+
+As stated in the goals section, the migration has to make progress even if the
+environment is flaky. This section describe how the migrator recovers from
+failure.
+
+Kubernetes replicaset controller restarts the "migration controller" pod
+if it fails. When the "migration controller" starts,
+* it checks if there exists any `progress` object that matches the current
```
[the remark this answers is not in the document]
caesarxuchao: The migrator for each resource type stores continue tokens. The `progress` object in addition needs to record if all resource types have been migrated.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+  and sets the worker job as the owner. When the "migration controller" deletes
+  the completed job, the configmap is garbage collected.
+
+### Automation
+
+If the Kubernetes deprecation [policy][] is followed, to migrate kubernetes
+native resources, the migrator only needs to run when the apiservers upgrade to
+a different minor version. Thus, the migration can be triggered automatically
+when the version stored in the `progress` object differs from what it returned
+at kube apiserver's `/version` endpoint.
```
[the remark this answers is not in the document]
caesarxuchao: Removed apiserver version from the alpha plan, and discussed in the automation section.

I don't think an incrementing counter is necessary, the command line tool just deletes old deployments before creating a new one. See the updated "workflow" section.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+at kube apiserver's `/version` endpoint. 
+
+[policy]:https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+For HA apiservers, migration should start after all apiservers have upgraded to
+the same version, which is not reflected in today's discovery API. Solving the
+HA apiservers agreement problem is a precondition for the migration automation.
+
+For CRD, because the CRD exposes all versions that are ever persisted in the
+storage since last migration, the migration should migrate any CRD that has
```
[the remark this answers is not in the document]
caesarxuchao: Changed the entire section. Explicitly state that in alpha we don't do automation, and we don't handle CRD.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+more than two persisted storage versions. 
+
+For aggregated apiservers, they can expose their version to trigger the tool
```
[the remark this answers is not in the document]
caesarxuchao: Rewrote the section.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Risks and Mitigations
+
+The migration process does not change the objects, so it will not pollute
+existing data.
+
+If the rate limiting is not tuned well, the migration can overload the
+apiserver. Users can delete the migration controller and the migration
+jobs to mitigate.
+
```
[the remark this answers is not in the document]
caesarxuchao: Added a paragraph suggesting rollback and redo the migration if that happens.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+* beta: integrating the migration into Kubernetes upgrade tests, and it will be a
+  blocking test for 1.13. ETA is 2018 Q4.
+
+* GA: implementing the automation.
+
+We will revisit this section in 2018 Q4. 
+
+## Open questions
+
+* Resources can migrate across API groups. For example, `deployment` was
```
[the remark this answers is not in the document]
caesarxuchao: I'll expand in the next iteration.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Alternatives
+
+### “Active management”
+
+[API Object Storage Version: Active Management][] (for simplicity, we’ll call it
+“active management”) is another propose. "Active management" is different to
+this proposal ("kube-migrator") in that it requires the apiservers to expose the
+storage versions. The advantage is that the "active management" controller only
+has to touch the resource types whose storage versions have changed, thus more
```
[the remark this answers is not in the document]
caesarxuchao: Got it.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,244 @@
...
+  insignificant.
+
+[API Object Storage Version: Active Management]:https://docs.google.com/document/d/1eoS1K40HLMl4zUyw5pnC05dEF3mzFLp5TPEEt4PFvsM
+[policy]:https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+
+### update-storage-objects.sh
+Nevertheless, the Kubernetes repo has a update-storage-objects.sh. It is not
+adopted by the community because it is not production ready: no rate limiting,
+hard-coded resource types, no persisted migration states.
```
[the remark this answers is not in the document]
caesarxuchao: Done.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,291 @@
...
+
+The migration process does not change the objects, so it will not pollute
+existing data.
+
+If the rate limiting is not tuned well, the migration can overload the
+apiserver. Users can delete the migration controller and the migration
+jobs to mitigate.
+
+Before upgrading or downgrading the cluster, the cluster administrator MUST
+check the "progress" object, and make sure every resource type is marked as
```
lavalamp: We can do slightly better than this. It can state separately whether roll-back and roll-forward is safe.

E.g., if you roll-forward, and then roll-back before the tool is finished, that is actually fine. You just can't roll-forward twice in a row without waiting, nor can you roll back twice in a row.

The tool has to track state to understand whether the operations in progress are rolling forward or backwards to support this.

I think it might actually be quite important, especially in large clusters, to be able to do an immediate rollback without waiting for a data migration.

sig-api-machinery/README.md
```diff
@@ -70,6 +70,9 @@ The following subprojects are owned by sig-api-machinery:
     - https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/OWNERS
     - https://raw.githubusercontent.com/kubernetes-sigs/controller-tools/master/OWNERS
     - https://raw.githubusercontent.com/kubernetes-sigs/kubebuilder/master/OWNERS
+- **server-upgrade-downgrade**
```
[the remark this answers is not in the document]
lavalamp: I think this is (going to be) a control plane feature--there's a section for that on line 42.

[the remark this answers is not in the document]
deads2k: > I think this is (going to be) a control plane feature--there's a section for that on line 42.

I must need to re-read.  I remember talking about creating this as a separate tool.  It allowed evolution outside of kube releases and was a clear indication that it was indeed separate from core.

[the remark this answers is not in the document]
deads2k: > I think this is (going to be) a control plane feature--there's a section for that on line 42.
> 
> I must need to re-read. I remember talking about creating this as a separate tool. It allowed evolution outside of kube releases and was a clear indication that it was indeed separate from core.

Or do you just mean, "yes it will be a separate repo, but it falls into an existing bucket and we can link the repo in the existing section"?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,291 @@
...
+* when a job completes, marks completion in the "progress" CR and deletes the
+  job.
+
+The schema of `progress`:
+```golang
+type Progress struct {
+  metav1.TypeMeta
+  metav1.ObjectMeta
+  // A list of all the resources as suggested by the apiservers
+  Resources []Resource
```
deads2k: The multiplicity here seems odd.  I would expect to have one instance per resource with both spec and status.

Also, shall we split the API and design for the controller out of this first pass?  I accept that such a controller can be built, but I don't see a reason to block the creation of the command line tool on the shape of the controller to come.

sig-api-machinery/README.md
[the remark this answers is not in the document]
lavalamp: > it will be a separate repo, but it falls into an existing bucket and we can link the repo in the existing section

Exactly. It will be a control plane feature when it grows up, no matter how it is invoked or where the code lives.

keps/sig-api-machinery/storage-migration.md
liggitt: > can we deprecate --storage-versions? It's defined in pkg/kubeapiserver. Kuberentes is following the 4-step deprecation policy, I don't see in what case the --storage-versions flag is useful

I agree with this. It seems like flexibility we don't need, that actually harms our ability to reason about what is safe to do as we maintain kube-apiserver in future releases.

keps/sig-api-machinery/storage-migration.md
deads2k: I think the API should be drivable by an external actor.  I should be able to indicate, "migrate storage for this resource".  That would bump a generation and then the actual migration would reflect a status generation to indicate that migration was complete.  Given an externally driveable change, I suspect you'd also want to be able indicate why you bumped it.

The granularity of both desired and actual state is at the level of resource.  Combining them makes a singleton resource which is generally odd, makes conflicts more likely, and generally means "bad" updates are more likely.

caesarxuchao:
Moved the repo creation request to https://github.com/kubernetes/community/pull/2568.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: @lavalamp @deads2k what do you think about removing ` --storage-versions`? I opened https://github.com/kubernetes/kubernetes/pull/67678 to remove the flag, but it's more complicated than I expected, can you let me know early if you think it's a bad idea?

lavalamp: removing `--storage-versions` means everyone has to use the default. How would you e.g. do a safe downgrade if someone skipped a step in the version upgrade process?

I think most people should use the default.

We can't remove the flag without a deprecation process unless we exhaustively demonstrate that no one is using it.

keps/sig-api-machinery/storage-migration.md
lavalamp: I agree with @deads2k about no singleton resources. The "drivable by an external actor" part implies further creation of such an actor. It makes each part of the system more focused and hence composable, at the cost of more moving pieces. Overall I think it's probably slightly better to do it that way.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: > How would you e.g. do a safe downgrade if someone skipped a step in the version upgrade process?

Not sure if I follow, could you be more specific?

> We can't remove the flag without a deprecation process unless we exhaustively demonstrate that no one is using it.

How about keeping a noop `--storage-versions` flag for one more release?

side note: if users are using this flag, they need to manually inspect this flag every time they upgrade anyway, because the API version in the flag value can be deprecated.

lavalamp: One more release is not our deprecation policy. Without looking, I think we need at least two, maybe more.

If someone introduces a v2 and also accidentally sets it as the storage version in the same release, it is impossible to downgrade safely without use of the storage version flag.

caesarxuchao: > If someone introduces a v2 and also accidentally sets it as the storage version in the same release, it is impossible to downgrade safely without use of the storage version flag.

We should write tests to prevent this kind of bugs. However, as we don't expose the storage versions, I don't know how to write such tests.

The deprecation [policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli) says 6 months.

caesarxuchao: Just realized that after we remove the `--storage-versions`, a group's preferred version showed in the discovery API equals the storage version, then we don't need to argue if we should expose the storage versions.. Is there any other reason why storage version might diverge from preferred version?

caesarxuchao:
@deads2k @lavalamp PTAL. Thanks.

Mainly changed the following:
* Addressed deads2k's comments on the singleton API. 
* Explicitly separated the alpha workflow (requiring admin intervention) from beta workflow (automated)
* Added the idea of expanding the discovery API with `StorageUID` to identify resources that share storage.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+and eventually arrive at a version of Kubernetes that can’t interpret the stored
+resource and crashes. See k8s.io/pr/52185."[1][]. We propose a tool that
+prevents this problem.
+
+[1]:https://docs.google.com/document/d/1eoS1K40HLMl4zUyw5pnC05dEF3mzFLp5TPEEt4PFvsM
+
+### Goals
+
+A successful storage version migration tool must:
+* work for Kubernetes built-in APIs, custom resources (CR), and aggregated APIs.
```
caesarxuchao: I'm not sure if we should handle aggregated APIs, which can have their own storage conventions. The read-then-write operations might not migrate the stored data. The API can even be non-idempotent.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+In both alternatives, HA apiservers present additional challenges. The migration
+must start after all HA apiservers have agreed on the storage versions or the
+apiserver versions. The agreement protocol is worth its own KEP.
+
+We will revisit this section in 2018 Q4. 
+
+### Risks and Mitigations
+
+The migration process does not change the objects, so it will not pollute
+existing data.
```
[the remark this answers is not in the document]
caesarxuchao: "Before upgrading or downgrading the cluster, the cluster administrator must run
the `kube-migrator progress` to make sure the migration has completed".

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+
+If the rate limiting is not tuned well, the migration can overload the
+apiserver. Users can delete the migration controller and the migration
+jobs to mitigate.
```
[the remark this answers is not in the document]
caesarxuchao: In the workflow added "makes sure there is only one **migration worker** `job` at a time, to avoid overloading the apiserver."

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+
+Before upgrading or downgrading the cluster, the cluster administrator MUST
+list the `migration` objects, and make sure every resource type is marked as
+completed. Otherwise the apiserver can crash. To mitigate, the cluster
```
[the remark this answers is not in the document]
caesarxuchao: Added an explanation, "because it cannot interpret the serialized data in etcd".

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+administrator can rollback the apiserver to the old version, and wait for the
+migration to complete. If the apiserver does not crash, the `migration` will be
+corrupted, administrator needs to re-launch the **kube-migrator** to recover.
```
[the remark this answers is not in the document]
caesarxuchao: We are concerned that there is no one-to-one mapping between apiserver version and storage versions, because user can configure `--storage-versions` flag, though the flag will be removed in 1.13.
Another concern is that the aggregated apiserver doesn't have a server version. If we can agree on either forcing aggregated apiserver follows semver, or that the storage migration doesn't handle aggregated APIs by default, we can implement what you suggested.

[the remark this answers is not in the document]
caesarxuchao: > it is not clear why 'migration' will be corrupted

Expanded. "If the apiserver does not crash after upgrading or downgrading, the `migration` objects lose track of migration progress, because the default storage versions might changed after upgrading".
or downgrading, but no one increments the `migration.spec.generation`.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+
+TODO: it is safe to rollback an apiserver to the previous version without
+waiting for the migration to complete. It is only unsafe to roll-forward or
+rollback twice. We need to design how to record the previous version, especially
+for the aggregated apiservers.
```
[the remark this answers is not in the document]
caesarxuchao: I think the blocker is aggregated apiservers. We need to either expose their versions, or decide not to take care of aggregate apiservers. Or we expose the storage versions of all resources. Basically the two alternatives in section `Beta workflow - Automation`.

Not sure what you mean by "guarded against failures in the migrator tool".

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Graduation Criteria
```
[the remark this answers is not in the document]
caesarxuchao: Yeah, it's a roadmap. Keeping the title as "graduation criteria" because that' required of a KEP.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+of Kubernetes built-in resource types, customer resources, and aggregated
+resources in the same fashion.
+
+Another advantage is that the *active management* controller only
+has to touch the resource types whose storage versions have changed, thus more
+efficient, though the difference is not that significant, because
+* The migration needs to run every time the master is upgraded to a
+  different minor version, which happens once per quarter. 
+* **kube-migrator** ignore resources that only have only one version. If the
+  4-step deprecation [policy][] is followed, v1 and v2 only co-exist for 2
```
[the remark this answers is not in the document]
caesarxuchao: In the "view" mode I can click through the hyperlink. Can you double check?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,360 @@
...
+After all the apiservers are at the desired version, including kube-apiservers
+and aggregated apiservers, the cluster administrator launches **kube-migrator**.
+It
+* deletes any existing `deployment` of **kube-migrator controller** 
+* creates a `deployment` of **kube-migrator controller** 
+* uses the same kubeconfig file as **kubectl** does.
+
+On start-up, the **kube-migrator controller** does the following,
+* discovers all supported APIs.
+* lists all custom resource definitions, excludes them from the API list
```
yliaog: so only CRDs are excluded? what about aggregated apis?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+  // StorageUID=SHA-256(etcd server in the form of scheme://ip:port, etcd path),
+  // obfuscated with a random key. E.g., for pods, it is
+  // SHA-256(scheme://ip:port/registry/pods).
+  // The string is obfuscated to avoid exposing server configuration details.
+  StorageUID string
+}
+```
+
+[APIResource]:https://github.com/kubernetes/kubernetes/blob/d38fb2c6451447ad0ecbdfed971a563419d60883/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L874
+
```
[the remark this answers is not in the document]
yliaog: works now

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+### Failure recovery
+
+As stated in the goals section, the migration has to make progress even if the
+environment is flaky. This section describes how the migrator recovers from
+failure.
+
+Kubernetes **replicaset controller** restarts the **migration controller** `pod`
+if it fails. When the **migration controller** starts, it resumes migration
+by reading the existing `migration` objects.
```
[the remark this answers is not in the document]
yliaog: i'm still confused. Below, it lists all the tasks done at the "kube-migrator controller" start-up time. My question is how does it know it is the first time (started by kube-migrator command) vs a second time (it is killed and restarted by the replicaset controller)

==================================================
On start-up, the kube-migrator controller does the following,

discovers all supported APIs.
lists all custom resource definitions, excludes them from the API list retrieved in the previous step.
for each non-custom resource, creates a migration CR (see the API section for the schema).
for each migration instance, increments the migration.spec.generation, and logs the reason.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+  the completed `job`, the `configmap` is garbage collected.
+
+### Beta workflow - Automation
+
+It is a beta goal to automate the migration workflow. That is, migration does
+not need to be triggered manually by cluster admins, or by custom control loops
+of Kubernetes distributions.
+
+The automated migration should work for Kubernetes built-in resource types,
+custom resources, and aggregated resources.
```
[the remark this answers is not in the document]
yliaog: ok, if custom resources are not that different from native resources, what prevents it from being included in the alpha phase?

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
yliaog: why not having this check automatically done inside kube-migrator? what if the admin forget to run "kube-migrator progress"?

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
yliaog: I can view the link now.

where are the 4-steps? the deprecation policy doc lists a few rules, no 4-steps

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Aggregated apis will be included in the first iteration, see line 90.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
caesarxuchao: Sorry for the confusion.

Updated this section to:
```
Kubernetes **replicaset controller** restarts the **migration controller** `pod`
if it fails. The restarted **migration controller** follows the same bootstrap
workflow as described in the [workflow section][]. Because the **migration
controller** adopts existing `migration` objects, it can resume from where
it left off.
```

and updated the workflow section to
```
On start-up, the **kube-migrator controller** does the following,
...
* for each non-custom resource, creates a `migration` CR (see the [API
  section][] for the schema) if it doesn't exist already.
* for each newly created `migration` instance, increments the
  `migration.spec.generation`, and logs the reason.
...
```

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
caesarxuchao: In the alpha phase, the migrator tool is manually invoked by cluster admins. They don't control the CRD lifecycle, so they don't know when to migrate the stored CRs.

In the beta iteration, where we want to automate the migrator invocation, we will build another component which monitors the CRD spec to figure out when their storage versions change, and then updates `migration.spec.generation`.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
caesarxuchao: Perhaps I missed your point. "upgrading or downgrading the cluster" are done via cloud-specific tools. We cannot force those tools to run `kube-migrator progress`.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
caesarxuchao: Removed this point. v1 and v2 can co-exist 4 quarters. The cost argument isn't critical after all. Safety is.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: The proposal isn't final yet. 

I want to discuss if storage migrator should take care of aggregated apis. If so, let's decide on the two alternatives listed in the `Beta workflow - Automation` section. Then I'll update the proposal.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,360 @@
...
+uses API chunking to retrieve partial lists of a resource type and thus can
+migrate a small chunk at a time. It stores the [continue token] in a `configmap`
+to preserve progress. With the inconsistent continue token introduced in
+[#67284][], the **migrator worker** does not need to worry about expired continue
+token.
+
+[#67284]:https://github.com/kubernetes/kubernetes/pull/67284
+[API section]:#api
+
+The cluster admin can run the `kube-migrator progress` command to check if the
```
yliaog: so kube-migrator has 'install' and 'progress' commands, what about 'uninstall' that can clean up everything related to the migration.

caesarxuchao: Yeah, we can add such a subcommand, would be trivial. I will skip it in the KEP.

caesarxuchao:
I have finished the iterations with @yliaog. 

We think the controversial part of this KEP is how to handle aggregated APIs, which can have their own storage conventions. The read-then-write operations might not migrate the stored aggregated APIs data. The API can even be non-idempotent. So at least we should make aggregated API's participation optional.

To handle aggregated APIs, we need to decide on the two alternatives in the "Beta workflow -  Automation" section.

Handling aggregated API isn't a blocker for the alpha phase. The tool will not distinguish aggregated APIs from built-in APIs in the alpha phase.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,360 @@
...
+every cluster uses the default [storage version configs][], then we can use the
+apiserver version to drive the migration. More specifically,
+* kube-apiserver removes the `--storagve-version` flag, so that every deployment
+  of kube-apiserver uses the default storage versions. [#67678][]
+* aggregated apiservers follow semantic versioning, and expose their version via
+  API.
+* aggregated apiservers follow the same API deprecation [policy][].
+
+Then the migration of built-in resources and aggregated resources can be driven
+by the changes of apiservers versions.
```
lavalamp: I'm pretty against driving storage changes on apiserver binary version changes. Many times when it changes there's no storage change needed.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,360 @@
...
+The automated migration should work for Kubernetes built-in resource types,
+custom resources, and aggregated resources.
+
+Here are two alternatives and their requirements.
+
+#### a. Exposing the configured storage versions via API
+
+This is what the [Active Management][] design proposes. It requires:
+* exposing the configured [storage versions][] of Kubernetes built-in resource
+  types and aggregated resources.
```
lavalamp: I still think this is the way to go. If an aggregated apiserver doesn't want to participate in the storage migrator or doesn't actually store its data or doesn't support ordinary changes or whatever, it can just omit the information in the discovery API.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: > Many times when it changes there's no storage change needed.

Migrating objects whose storage version hasn't changed doesn't impact correctness. The extra workload is also tolerable.

If storage versions can change when apiserver binary versions do *not* change, that will impact correctness. Can you suggest cases where we need to do that?

deads2k: > I'm pretty against driving storage changes on apiserver binary version changes. Many times when it changes there's no storage change needed.

Given the fairly low cost of migration and the fairly low frequency of binary versions, it seems reasonable to kick the storage migration globally when it happens.  As @caesarxuchao noted, it doesn't impact correctness and it can be refined and extended as time goes on.

As @liggitt has noted in the past, the leaking the storage version isn't required by clients and doesn't mean anything to every other client out there.

Choosing to trigger on this now doesn't preclude triggering on other items later on.  For a beta, this seems like a very simple start that doesn't require exposing extraneous information, doesn't limit future choice, and doesn't impact correctness.

lavalamp: I view the kubernetes version and the api machinery version as orthogonal things. The apiserver binary version isn't logically correlated to the api versions. I would like to live in a world where I can keep my kubernetes version constant while picking up new api machinery changes, or vice versa.

Today these things happen to be correlated but that's accidental. It's really bad to make an automatic trigger that adds an unintended semantic to some facet of our api. E.g. is the /version endpoint even a "real" api covered by our deprecation policy? Adding this trigger changes the meaning of a version change from "I'm now running X" to "please touch every object in the system". I don't think that's a good meaning change. You should have to do something more than push a new patch to get every object in your system touched.

Just to be extra clear, my objection has nothing to do with efficiency and everything to do with not adding unintended consequences. 

I don't care about the exact mechanism we use as long as it has no false positives and no false negatives, again not because of efficiency but because that indicates it's being driven from a signal that's not going to change meaning in the future. We don't HAVE to expose the storage version, although that'd be a convenient thing to trigger on--it works great for rolling backwards as well as rolling forwards. Other clients don't need to know the storage version, but I think the storage migration triggerer component does deserve to know.

Are you worried that clients will abuse knowing the storage version? How?

I also really want some way of tracing resources when they move between groups, e.g. deployment. We haven't done too much of that recently, but I bet CRD authors are going to want to do it sooner or later.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,360 @@
...
+  // TBD: maybe resource=hash(etcd instance, etcd path prefix).
+  // Immuntable.
+  Resource string 
+  // GroupVersions are all the groups and versions this resource is served by
+  // the apiserver. The server preferred version should be groupVersions[0].
+  // **Migration workers** GET & PUT via the preferred version by default, to
+  // avoid conversions in the apiserver.
+  GroupVersions []GroupVersion
+  // Generation is the generation of the spec. Incrementing generation triggers
+  // migration. We do not use the objectMeta.generation because its semantic is
```
deads2k: Pretty sure we don't bump it for that.  If we do, we should flag it as not good and avoid doing that.

deads2k:
> I don't care about the exact mechanism we use as long as it has no false positives and no false negatives, again not because of efficiency but because that indicates it's being driven from a signal that's not going to change meaning in the future.

Perhaps this is the conflict.  I don't care about an infrequent false positive.  A false positive just runs a rate limited read write cycle on resources.  

I think we can and should develop an API that can be triggered separately and I think the one described in this KEP is worth starting with.  If we separate the trigger mechanism out of the KEP for now, we can proceed on the guts behind them while we come to an agreement about how to trigger them.  In the simplest case, we simply provide a few simple trigger options and other integrators can share or build their own.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Yeah, for CR only spec change bumps the generation ([code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L118-L126)). I'll update.

The issue I cited was about CRD actually. And FWIW, deployment increments generation for annotations changes, see [code](https://github.com/kubernetes/kubernetes/blob/58d43abc39178dc2efc3b1ed6e4727bb34a4b68a/pkg/registry/apps/deployment/strategy.go#L106).

caesarxuchao: Updated the KEP to use the `metadata.generation`.

caesarxuchao:
Thanks for the input. I removed the contentious discussion on the the migration automation. PTAL.

keps/sig-api-machinery/storage-migration.md
lavalamp: Can this state be kept in a CRD, such that I can use kubectl to read the current progress? It seems contra our general direction to introduce new command line tools.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Graduation Criteria](#graduation-criteria)
+      * [Alternatives](#alternatives)
+         * [“Active management”](#active-management)
+         * [update-storage-objects.sh](#update-storage-objectssh)
+
+
+## Summary
+
+We propose to extend and improve the [oc adm migrate storage][] command line
```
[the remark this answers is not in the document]
lavalamp: missing push?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,346 @@
...
+to preserve progress. With the inconsistent continue token introduced in
+[#67284][], the **migrator worker** does not need to worry about expired continue
+token.
+
+[#67284]:https://github.com/kubernetes/kubernetes/pull/67284
+[API section]:#api
+
+### API
+
+We introduce the `migration` API to record the intention and progress of a
```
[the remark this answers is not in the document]
lavalamp: It'd be OK to have a counter for the number of items migrated so far; that lets you see whether it's stuck or not which is pretty handy.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,297 @@
...
+  Spec StorageVersionMigrationSpec
+  Status StorageVersionMigrationStatus
+}
+
+// Note that the apiserver will increment `metadata.generation` when this spec
+// is updated.
+type StorageVersionMigrationSpec {
+  // Resource is the identifier of the resource that's being migrated.
+  // TBD: maybe resource=hash(etcd instance, etcd path prefix).
+  // Immuntable.
```
lavalamp: Immutable

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,297 @@
...
+  Resource string 
+
+  // GroupVersions are all the groups and versions the apiserver serves this
```
lavalamp: Why is it necessary to pass all? Shouldn't a single one be sufficient? And why do you need to care whether it is preferred or not?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,297 @@
...
+  // resource in. The server preferred version should be groupVersions[0].
+  // **Migration worker** GET & PUT via the preferred version by default, to
+  // avoid conversions in the apiserver.
+  GroupVersions []GroupVersion
+  
+  // LastUpdateTimestamp is the last time this spec is updated. The value is set
```
lavalamp: 1. Why do you need to do this?
2. Please at least use a logical clock (an integer that counts up).

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,297 @@
...
+  // by clients, not managed by the apiserver. The timestamp introduces randomness
+  // to the spec so that every update of the spec changes the contents of the spec,
+  // which leads to the increment of `metadata.generation`.
+  LastUpdateTimestamp metav1.Time
+
+  // Reason of the migration. Examples are "manually triggered", "apiserver
```
lavalamp: I'm somewhat skeptical that this will be useful.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,297 @@
...
+  metav1.TypeMeta
+  metav1.ObjectMeta
+  Spec StorageVersionMigrationSpec
+  Status StorageVersionMigrationStatus
+}
+
+// Note that the apiserver will increment `metadata.generation` when this spec
+// is updated.
+type StorageVersionMigrationSpec {
+  // Resource is the identifier of the resource that's being migrated.
```
lavalamp: I don't understand the relationship between Resource and GroupVersions.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,297 @@
...
+  // LastUpdateTimestamp is the last time this spec is updated. The value is set
+  // by clients, not managed by the apiserver. The timestamp introduces randomness
+  // to the spec so that every update of the spec changes the contents of the spec,
+  // which leads to the increment of `metadata.generation`.
+  LastUpdateTimestamp metav1.Time
+
+  // Reason of the migration. Examples are "manually triggered", "apiserver
+  // updated", etc.
+  Reason string
+}
```
lavalamp: I guess what I expected to see is an object that expresses a migration request for a single resource type (e.g., apps/v1/deployment). With status indicating progress {not started, migrated N objects, completed}, and where the object is removed after completion.

The generation stuff makes me think you want to reuse a single object to make multiple requests, but I don't understand why you would want to do that.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: > Why is it necessary to pass all?

For optimization. The migrator can skip the resource if this resource only has one version.

In the iteration addressing your comments, I updated the KEP to say that at the alpha stage, the tool migrates a hard-coded list of resources, this field became unnecessary.

> And why do you need to care whether it is preferred or not?

I thought if the migrator GET&PUT at the preferred version could save the apiserver a few conversions, but my calculation was wrong.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Removed the generation stuff, so this is removed as well.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Removed.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Removed the command, user can just run `kubectl get migrations -o jsonpath="{.status.completed}"`.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
caesarxuchao: Added `migration.status.continueToken`. User can use the token to check progress.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -81,158 +81,109 @@ As to the deliverables,
...
-We introduce the `migration` API to record the intention and progress of a
+We introduce the `migration` API to record the intention and the progress of a
 migration. The API will be a CRD defined in the `migration.k8s.io` group.
 
 Read the [workflow section][] to understand how the API is used.
 
 ```golang
 type StorageVersionMigration struct {
   metav1.TypeMeta
+  // metadata.name should be "<resource>.<group>" of the resource being
```
lavalamp: Why? I'd probably just say to use generateName to avoid collisions. Requiring this means the old request has to be deleted before I can make a new request.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -81,158 +81,109 @@ As to the deliverables,
...
-  ObeservedGeneration int64
-
-  // CompletedGeneration is the greatest spec generation whose **migration worker**
-  // has completed.
-  CompletedGeneration int64
+  // Completed means the migration of this resource has completed.
+  Completed bool
+
+  // ContinueToken is the token to use in the list options to get the next chunk
+  // of objects to migrate. User can use this token to check the progress of the
```
lavalamp: Can the user use this to tell if the migration has started or not? If yes, how? If no, can we add this to the API?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -81,158 +81,109 @@ As to the deliverables,
...
+  // migration.
+  ContinueToken string
```
lavalamp: Should this be optional?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,249 @@
...
+  the API objects to the proper storage versions.
+* in the long term, automating the migration of Kubernetes built-in APIs, CR,
+  aggregated APIs without further burdening system administrators or Kubernetes
+  distributions.
+
+## Proposal
+
+### Alpha workflow
+
+We will create a **kube-migrator** command line tool. At the alpha stage,
```
lavalamp: I would greatly prefer CRDs or yamls that can be controlled by kubectl. Why would we add a new command line tool for this? The whole point of our control plane model is that it's extensible.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
lavalamp: We probably need the triggering mechanism to cancel an in-progress migration if the trigger changes (and then start a new one). It's e.g. safe for user to go forward and then back without waiting for the migration to complete.

But actually I don't know how to make this safe for a single object write (read on version X, but by the time there's a write it's version Y). You need something like a precondition on the api micro version for that to be 100% safe.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -81,158 +81,109 @@ As to the deliverables,
...
 
-After all the apiservers are at the desired version, including kube-apiservers
-and aggregated apiservers, the cluster administrator runs the `kube-migrator
-install` command. It
+After all the kube-apiservers are at the desired version, the cluster
+administrator runs the `kube-migrator install` command. It
 * deletes any existing `deployment` of **kube-migrator controller** 
 * creates a `deployment` of **kube-migrator controller** 
+* generates a comprehensive list of resource types via the discovery API
+* discovers all custom resources via listing CRDs
```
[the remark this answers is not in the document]
caesarxuchao: We can't tell if it's a CRD or a built-in resource by looking at the apiservices. The apiservices for CRD also have `.spec.service == null`.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -81,158 +81,109 @@ As to the deliverables,
...
+* discovers all aggregated resources via listing all `apiservices` that have
+  `.spec.service != null`
+* removes the custom resources and aggregated resources from the comprehensive
+  resource list. The list now only contains Kubernetes built-in resources.
+* removes resources that share the same storage. At the alpha stage, the
+  information is hard-coded, like in this [list][].
```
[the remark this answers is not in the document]
caesarxuchao: The link works for me. Are you referring to the empty `[]`? That's a neat format of inserting hyperlinks in git markdown. It has a pairing `[list]:https://github.com/openshift/origin/blob/2a8633598ef0dcfa4589d1e9e944447373ac00d7/pkg/oc/cli/admin/migrate/storage/storage.go#L120-L184` at the end of the section. If you `view` the rendered file you'll be able to click through.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: I guess you want to keep the old requests to record history?

I'll have to add a `.status.active bool` to the `migration`, to let clients tell if the `migration` is for history keeping or is active. Sounds good?

keps/sig-api-machinery/storage-migration.md
caesarxuchao: It would be empty if the migration hasn't started. And if I added `.status.active` to `migration` as in https://github.com/kubernetes/community/pull/2524#discussion_r221792276, that would indicate the start as well.

keps/sig-api-machinery/storage-migration.md
lavalamp: No, you already have a .status.completed for that. It's not about keeping history (which I don't want to do) it's about there not being any need to (effectively) have a per-resource lock on creating these things. (which is not even that effective since occasionally we move things between groups...)

Also imagine I have foo in group/v1 and also foo in group/v5, but they don't share key spaces because the one in v5 is a complete rewrite. Now I want to migrate v1 to v2 and also v5 to v6. It's contrived but I don't see any benefit at all from requiring the name match that format.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
lavalamp: ah my bad.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
lavalamp: ...which doesn't matter since you want to exclude both, no?

keps/sig-api-machinery/storage-migration.md
lavalamp: Don't add a bunch of bools if an enum will serve.

Also for this one respond in the comment (responding on github is optional) ;)

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Since kubernetes doesn't support transactions, I think it's easiest to only keep one `migration` for each resource type. We can keep the historic requests in a subfield.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
caesarxuchao: > We probably need the triggering mechanism to cancel an in-progress migration if the trigger changes (and then start a new one). 

I think it's a general requirement to have a single migration job running for each resource type. I don't understand why it's a specific requirement of enabling fast roll back.

I didn't follow the second paragraph.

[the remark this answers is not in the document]
caesarxuchao: Side note: to enable fast rollback, `migration.status` needs to reflect what is the currently stored in etcd, either the stored version, the hashed stored version, or the apiserver version (probably isn't safe in the future). I didn't put it in the proposal because we haven't reached consensus on the triggers.

keps/sig-api-machinery/storage-migration.md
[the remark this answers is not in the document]
caesarxuchao: I want to exclude both the custom resources (CR) and the aggregated resources (AR). I want to only include the built-in resources. From the apiservices, I can tell AR from CR and the built-ins, but I can't tell CR from the built-ins.

keps/sig-api-machinery/storage-migration.md
deads2k: > I would greatly prefer CRDs or yamls that can be controlled by kubectl. Why would we add a new command line tool for this? The whole point of our control plane model is that it's extensible.

This is the basic tool we build up.  This tool would be what is later trigger automatically.  Doing it like this allows us to build up a series of useful pieces.  Nothing precludes adding controller logic around it (this doc describes exactly that), but a standalone CLI tool has value in and of itself.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,249 @@
...
+type StorageVersionMigrationSpec {
+  // Resource is the resource that is being migrated. The migrator sends
+  // requests to the endpoint serving the Resource.
+  // Immutable.
+  Resource GroupVersionResource
+}
+
+type StorageVersionMigrationStatus {
+  // Completed means the migration of this resource has completed.
+  Completed bool
```
deads2k: I expected status to include conditions which are fairly standard in the API and have built in scripting functionality.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,249 @@
...
+  // metadata.name can be hash(etcd instance, etcd path prefix).
+  metav1.ObjectMeta
+  Spec StorageVersionMigrationSpec
+  Status StorageVersionMigrationStatus
+}
+
+type StorageVersionMigrationSpec {
+  // Resource is the resource that is being migrated. The migrator sends
+  // requests to the endpoint serving the Resource.
+  // Immutable.
```
deads2k: if this is immutable, how do I update a field in here to indicate that the generation needs to be bumped?  The reason field made such a thing a possible and allowed me to key a request to an external system with an opaque key.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: If the format of the name is `hash(etcd instance, etcd path prefix)` (as the next line in the comment suggested for the future), there is no issue in the case where we move things between groups, or in the case where group/v1/foo and group/v5/foo don't share key spaces.

It doesn't make sense to run multiple migration for the same resource type at the same time. If we go with your suggestion, for each resource type, we need to mark only one of the `migration` as active. Because Kubernetes doesn't support transactions, I don't see any easy way to do that. So I still favor the current design for simplicity.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: I think Daniel also supports building up the migrator in separable pieces. His concrete [suggestion] (https://github.com/kubernetes/community/pull/2524#discussion_r221402698) was moving the API discovery step to a separate job, that would leave very little logic in the CLI.

I think we should still make the CLI, it will be a thin layer that launches jobs and creates CRs. Users should be able do all the operations via kubectl. The value of the CLI is saving users from editing yamls.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Do you mean making it a enum supporting states like Running, Succeeded, Failed? I agree it should be an enum instead of an boolean.

I'm not sure what you mean by "built in scripting functionality"?

keps/sig-api-machinery/storage-migration.md
caesarxuchao: I'm thinking in the future we will add a mutable field to this spec, like `targetStorageVersion(Hash)`, then the spec will be changing.

My concern with a `Reason string` field is it's too loose. User might edit the `Reason` and inadvertently triggered a new round of migration.

Can the external system to be triggered by the `migration.status.conditions == proposed`?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -81,158 +81,109 @@ As to the deliverables,
...
+* discovers all aggregated resources via listing all `apiservices` that have
+  `.spec.service != null`
+* removes the custom resources and aggregated resources from the comprehensive
+  resource list. The list now only contains Kubernetes built-in resources.
+* removes resources that share the same storage. At the alpha stage, the
+  information is hard-coded, like in this [list][].
+* creates `migration` CRs (see the [API section][] for the schema) for all
+  remaining resources. The `ownerReferences` of the `migration` objects are set
+  to the **kube-migrator controller** `deployment`. Thus, the old `migration`s
+  are deleted with the old `deployment` in the first step.
```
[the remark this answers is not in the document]
caesarxuchao: I don't follow. I'm think that the user needs to recreate the initialization `job`, maybe by `kubectl apply` the `job` with a different name. The `job` will delete the old controller and `migration`s and create new ones.

keps/sig-api-machinery/storage-migration.md
deads2k: > Can the external system to be triggered by the `migration.status.conditions == proposed`?

The external system drives this migration.  It needs the generation of this object to be incremented to do that.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Sure. Added.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: I'll start with adding `.status.phase`.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
 ```golang
 type StorageVersionMigration struct {
   metav1.TypeMeta
-  // metadata.name should be "<resource>.<group>" of the resource being
-  // migrated.
-  // Because resouces can be served at different groups, in the future version,
-  // metadata.name can be hash(etcd instance, etcd path prefix).
+  // At the alpha stage, metadata.name must be "<resource>.<group>" of the
+  // resource being migrated.
+  // We require this name format to avoid duplicate `migration` CRs for the same
```
lavalamp: I still think preventing such a duplicate should be a non-goal; it prevents you from starting a new one if a record of an old one is still around.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
...
+  // resource type. Otherwise we will need to build complicated system to make
+  // sure the latest `migration` CR is honored.
```
lavalamp: You already execute only one migration at a time; why would you need to build anything else?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
...
+  // resource being migrated.
+  // We require this name format to avoid duplicate `migration` CRs for the same
+  // resource type. Otherwise we will need to build complicated system to make
+  // sure the latest `migration` CR is honored. 
+  // Using "<resource>.<group>" isn't entirely safe, for two reasons:
+  // a. resources share key spaces can be served at different groups
+  // b. groupX/v1/foo and groupX/v5/foo may not share key spaces and thus
+  // require two `migration` CRs, though Kubernetes don't have such cases yet. 
+  // To address the two concerns, in the beta version, the name need to be
+  // tied to the key spaces, e.g., a hash of "etcd instance"+"etcd path prefix".
```
lavalamp: The only reason I can think of for layering requirements on the name is for idempotency. Is that why you are doing it?

Why can't you just let the client worry about that? Why do you have to mandate it?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
...
   metav1.ObjectMeta
   Spec StorageVersionMigrationSpec
+  // If status is empty, it means the `migration` is just created. The migrator
+  // controller hasn't launched the **migration worker** `job` for it yet.
```
lavalamp: All of our objects work like this, is it necessary to state?

keps/sig-api-machinery/storage-migration.md
lavalamp: Why does the generation need to change if you are deleting and recreating to request a new migration?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
...
-  // requests to the endpoint serving the Resource.
+  // requests to the endpoint tied to the Resource.
   // Immutable.
   Resource GroupVersionResource
 }
 
+type MigrationPhase string
+
+const (
+  // MigrationPending indicates that the migration is requested, but no
```
lavalamp: This contradicts what you said about "status"?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
...
+const (
+  // MigrationPending indicates that the migration is requested, but no 
+  // migrator job has been created to fulfill the request.
+  MigrationPending MigrationPhase = "Pending"
+  // MigrationRunning indicates that a migrator job is running.
+  MigrationRunning MigrationPhase = "Running"
+  // MigrationSucceed indicates that the migration has completed successfully.
+  MigrationSucceeded MigrationPhase = "Succeeded"
+  // MigrationSucceed indicates that the migration has failed.
+  MigrationRunning MigrationPhase = "Running"
```
lavalamp: Comment, name, and value are copied wrong :)

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
...
+)
+
 type StorageVersionMigrationStatus {
-  // Completed means the migration of this resource has completed.
-  Completed bool
+  // Phase indicates where the migration is in the life-cycle. Do not treat it
+  // as a state machine, because new phases can be added.
```
lavalamp: They can?

I think this is an impossible ask from clients.

There is confusion about this, but I think the latest guidance is still that extensible things like this should be represented as conditions.

But does this need to be extensible?

keps/sig-api-machinery/storage-migration.md
```diff
@@ -141,29 +149,57 @@ Read the [workflow section][] to understand how the API is used.
...
+  Phase MigrationPhase
 
   // ContinueToken is the token to use in the list options to get the next chunk
-  // of objects to migrate. User can use this token to check the progress of the
-  // migration.
+  // of objects to migrate. When the .status.phase is "Running", users can use
+  // this token to check the progress of the migration.
+  // +optional
```
lavalamp: To be extremely technical, continue token isn't status, because it's not reproducible from the environment.

keps/sig-api-machinery/storage-migration.md
lavalamp: Why is that problematic? Do you expect clients to request that? Folks have repeatedly said we don't care about extra migrations.

Also if the client wants to do that, they can always just repeatedly delete and re-add the object.

We don't need to protect them from themselves to that degree.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Ok, I've converted to conditions.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: I think three of us had three ideas on how to drive the migration control loop. I updated the KEP to follow Daniel's idea.

The original three ideas are:

I wanted to maintain one `migration` CR for one resource type, and the control loop would start the migration if `migration.status==nil`. User needs to delete and create `migration` to request for another round of migration.

I think deads2k's model was also one `migration` CR for one resource type, but using the `migration.metadata.generation != observedGeneration` to drive the control loop to do migration.

In lavalamp's model, users *can* create multiple `migration` CRs for a single resource type. Each CR will cause the control loop to do the migration once. This results in wasteful migration, but doesn't affect correctness. This state is recoverable by deleting redundant `migration`s. And we won't reach this state unless users are manually creating duplicate `migration` CRs. I updated the KEP to lavalamp's idea because it's the simplest and low risk.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Right. I forgot to remove it.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Converted to the conditions pattern.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: The migrator can maintain a configmap to record the continue token. That would make this field reproducible, but it doesn't add any value.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Reading the KEP again, I think both the "kube-migrator install" and "kube-migrator progress" commands can be replaced with kubectl one-liners. So I removed the standalone CLI.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Alpha API
+
+We introduce the `migration` API to record the intention and the progress of a
+migration. The API will be a CRD defined in the `migration.k8s.io` group.
+
+Read the [workflow section][] to understand how the API is used.
+
+```golang
+type StorageVersionMigration struct {
```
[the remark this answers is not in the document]
caesarxuchao: Added a line saying that we abbreviated it as `migration` in the text.

I think the API name should be fully qualified, because we might add other kind of migrations in the future.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,277 @@
...
+  metav1.TypeMeta
+  // metadata.generateName will be "<resource>.<group>" of the resource being
```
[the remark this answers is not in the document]
caesarxuchao: It's only intended for the readers of this KEP. I'll remove it in the API PR. Explained in the KEP.

keps/sig-api-machinery/storage-migration.md
caesarxuchao: Done.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,277 @@
...
+As to the deliverables,
+* in the short term, providing system administrators with a tool to migrate
+  the Kubernetes built-in API objects to the proper storage versions.
+* in the long term, automating the migration of Kubernetes built-in APIs, CR,
+  aggregated APIs without further burdening system administrators or Kubernetes
+  distributions.
+
+## Proposal
+
+### Alpha workflow
```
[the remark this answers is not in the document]
caesarxuchao: Added in the workflow section.

keps/sig-api-machinery/storage-migration.md
```diff
@@ -0,0 +1,277 @@
...
+// request another round of migration for the resource, clients need to create
+// another `migration` CR.
+type StorageVersionMigrationSpec {
+  // Resource is the resource that is being migrated. The migrator sends
+  // requests to the endpoint tied to the Resource.
+  // Immutable.
+  Resource GroupVersionResource
+}
+
+type MigrationConditionType string
```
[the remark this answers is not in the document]
caesarxuchao: We might add other conditions like "Pause"? There will be other clients monitoring the `migrations.status`, for example, cloud providers monitor this status to decide if the cluster is safe to upgrade. `conditions` force developers to not think the status as a state machine.

Also the `kubectl wait` function only works with `.status.conditions`. The command is marked as "experimental" but it's quite neat.

lavalamp:
/approve

perhaps you should squash?

@deads2k it's changed a bit since you last saw it, do you want to look again?

keps/sig-api-machinery/storage-migration.md
deads2k: > I think three of us had three ideas on how to drive the migration control loop. I updated the KEP to follow Daniel's idea.
> 
> The original three ideas are:
> 
> I wanted to maintain one `migration` CR for one resource type, and the control loop would start the migration if `migration.status==nil`. User needs to delete and create `migration` to request for another round of migration.
> 
> I think deads2k's model was also one `migration` CR for one resource type, but using the `migration.metadata.generation != observedGeneration` to drive the control loop to do migration.
> 
> In lavalamp's model, users _can_ create multiple `migration` CRs for a single resource type. Each CR will cause the control loop to do the migration once. This results in wasteful migration, but doesn't affect correctness. This state is recoverable by deleting redundant `migration`s. And we won't reach this state unless users are manually creating duplicate `migration` CRs. I updated the KEP to lavalamp's idea because it's the simplest and low risk.

I think that I could use the model multiple migration CRs for a single resource type.  It's a more batch-like way of looking at the problem, but I think it can work.  Proof of the pudding is in the tasting - I'm willing to taste it and see.

lavalamp:
Looks like owners has changed.

/assign @idvoretskyi

PR #2527, opened 2018-08-13
explicit and clear about what majority means

MHBauer:
@carolynvs @duglin @kibbles-n-bytes @pmorie


duglin:
I like line 87 needs to be changed too

MHBauer:
The chairs are enumerated specifically, and lazy consensus is a find a negative response situation. 2/3rds of those lazily responding implies they have to respond when laziness implies the opposite. It just doesn't make sense.

duglin:
fair enough. Personally, I wish we didn't have lazy consensus - to assume a default position of 'yes' is just weird to me and can be easily gamed by wording but that's a different issue I guess.
/approve

PR #2529, opened 2018-08-14
Add Korean l10n repo owners for sig-docs

gochist:
Added a team, Korean l10n repository owners for sig-docs.


zacharysarah:
LGTM from a SIG Docs perspective.

/lgtm

PR #2533, opened 2018-08-14
[WIP] Create cloning.md

erinboyd:
Design proposal for same namespace cloning support in kube

### DIFF
--- a/contributors/design-proposals/storage/cloning.md
+++ b/contributors/design-proposals/storage/cloning.md
@@ -16,14 +16,14 @@ but always provide a full and complete copy.
 * Host-Assisted Clone - using a pod to copy data between two persistent volumes
 
 ## Goals
-* Provide consistent process/annotations to execute cloning for storage technologies both for intree and out of tree plugins
+* Provide consistent process to execute cloning for storage technologies both for intree and out of tree plugins
 
 ## Non Goals
 * The process should not force storage vendors to implement cloning
 * Provide governance for security to clone assets between namespaces
 
 ## Value add to Kubernetes  
- By providing a consistent process/annotation for cloning, users can leverage the native capability of their storage to quickly create a 
+ By providing a consistent method for cloning, users can leverage the native capability of their storage to quickly create a 
  copy of their persistent volumes without having to wait for a snapshot object to be created. This is especially desirable for large 
  data volumes. Cloning also is a common use case in disaster recovery and for pre-seeding storage in a virtualized environment that many 
  users accustomed to operating in those environments might find useful. As plug-ins move out of tree, it will be more difficult to 
--- a/contributors/design-proposals/storage/cloning.md
+++ b/contributors/design-proposals/storage/cloning.md
@@ -83,7 +83,7 @@ spec:
   storageClassName: csi-gce-pd
   dataSource:
     kind: PersistentVolume
-    name: pvc-1
+    name: pv-1
   resources:
     requests:
       storage: 10Gi
--- a/contributors/design-proposals/storage/cloning.md
+++ b/contributors/design-proposals/storage/cloning.md
@@ -20,9 +20,6 @@ but always provide a full and complete copy.
 
 ## Non Goals
 * The process should not force storage vendors to implement cloning
-* An admin wishes to clone a volume to give a copy to another user in a different namespace 
-* An admin creates a ‘golden’ image in a restricted namespace and provides copies of that volumes to other namespaces
-* A user wishes to use a promoted snapshot to create a clone of it in a different namespace
 * Provide governance for security to clone assets between namespaces
 
 ## Value add to Kubernetes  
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/cloning.md
```diff
@@ -44,6 +44,105 @@ Specific use cases around cloning are included in the use cases listed below as
 ## Design Overview
 The proposed design is based on the idea of leveraging well defined concepts for storage in Kubernetes. The cloning will be initiated 
 by creating an annotated PVC to create a new volume using a defined storage class that supports cloning.
-The actual cloning process must be implemented by the provisioning storage and properly understand and react to the annotation to clone.
+The actual cloning process must be implemented by the provisioning storage and properly understand and react to the annotation to clone. In the absence of a storage class that supports cloning a host-assisted clone can also be exected to create the copy. The host-assisted cloning deploys a controller to utlize pod streaming to execute the copy.
```
j-griffith: s/and properly understand and react/, therefore plugins that support cloning need to ensure they're checking ..../  

Also, do we want annotations mentioned here or aren't we using VolumeContentSource from the spec?

minor typo s/exected/executed/

contributors/design-proposals/storage/cloning.md
```diff
@@ -44,6 +44,105 @@ Specific use cases around cloning are included in the use cases listed below as
...
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: clone-pvc
+  Namespace: myns
+spec:
+  accessModes:
+    - ReadWriteOnce
+  storageClassName: csi-gce-pd
+  dataSource:
```
j-griffith: Ahh, yeah that's what I was getting at above; I'd just call out the field name rather than say "annotations" just to be clear.

contributors/design-proposals/storage/cloning.md
```diff
@@ -44,6 +44,105 @@ Specific use cases around cloning are included in the use cases listed below as
...
+```
+
+```
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: csi-gce-pd
+provisioner: kubernetes.io/gce-pd
+parameters:
+  smartclone: "true"
```
j-griffith: Should smartclone just be an inherent/implied *thin* based on the a devices reported capabilities?  In other words if it's reported via capabilities a user doesn't need to specify this; a clone request is made, if capabilities allow great, if not error or fall back to host-assisted if possible.

erinboyd: I think this will change with the introduction of the populator, which I have intentionally left out of this proposal. I can remove this as well so as to reduce confusion

erinboyd:
@j-griffith please re-review to make sure I didn't miss anything from the design meeting. Thanks!

contributors/design-proposals/storage/cloning.md
```diff
@@ -0,0 +1,158 @@
...
+kind: PersistentVolume
+metadata:
+  name: pv-1
+  Namespace: myns
+spec:
+  capacity:
+    storage: 100Gi
+  volumeMode: Filesystem
+accessModes:
+    - ReadOnlyMany
```
humblec: Do we need to think about restricting a clone for a RO source PV?

j-griffith: @humblec I believe that came up in some discussions, but most devices that offer enhanced cloning features also handle things like dealing with consistency internally.  Even for those doing host assisted type cloning they would implicitly deal with this as well if they need to via mount options.  Also, just to be clear this is referencing the PVC and not the PV (just to make sure there's no confusion).

We don't want to restrict the cloning operation to RO I don't think, if we need something to control in-use PVCs we could perhaps use capabilities reporting (CSI Plugin/Spec) and deal with it there.  So something like "can clone in-use volumes" or something like that?

contributors/design-proposals/storage/cloning.md
```diff
@@ -0,0 +1,158 @@
...
+
+### CSI Spec
+This PR does not propose an API change, however the CSI specification will need to be updated as follows to add an additional
+option for the VolumeContentSource:
+
+``` 
+// Specifies what source the volume will be created from. One of the
+// type fields MUST be specified.
+message VolumeContentSource {
+  message SnapshotSource {
```
j-griffith: We've got this one in place so that's good

contributors/design-proposals/storage/cloning.md
```diff
@@ -0,0 +1,158 @@
...
+    // Contains identity information for the existing source snapshot.
+    // This field is REQUIRED. Plugin is REQUIRED to support creating
+    // volume from snapshot if it supports the capability
+    // CREATE_DELETE_SNAPSHOT.
+    string id = 1;
+  }
+  message PersistentVolumeSource {
```
j-griffith: This ones under review, uses a more generic term of "VolumeSource" which I think is preferable.
https://github.com/container-storage-interface/spec/pull/244

PR #2535, opened 2018-08-14
Update reference to SIG Network in SIG Node Readme

jordanbuchman:
/sig node


cblecker:
Thanks for the fix, @jordanbuchman 🚀 🎉

PR #2538, opened 2018-08-15
Refering to OWNERS_ALIASES from the assigning bulletpoint

CsatariGergely:
In the description of how to assign reviewers to a pr it is not
explicitly mentioned, that the aliases used in OWNERS files should
be resolved before assigning.

Signed-off-by: Gergely Csatari <gergely.csatari@nokia.com>

### DIFF
--- a/contributors/guide/owners.md
+++ b/contributors/guide/owners.md
@@ -101,8 +101,8 @@ aspects of this process may be configured on a per-repo basis.
 - Phase 2: Humans approve the PR
   - The PR **author** `/assign`'s all suggested **approvers** to the PR, and optionally notifies
     them (eg: "pinging @foo for approval")
-  - Only people listed in the relevant OWNERS files, either directly or through an alias, can act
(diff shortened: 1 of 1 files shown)


contributors/guide/owners.md
```diff
@@ -102,7 +102,8 @@ aspects of this process may be configured on a per-repo basis.
   - The PR **author** `/assign`'s all suggested **approvers** to the PR, and optionally notifies
     them (eg: "pinging @foo for approval")
   - Only people listed in the relevant OWNERS files, either directly or through an alias, can act
```
tpepper: Isn't this already covered here?  "...either directly or through an alias..."

I can see changing that minimally to:
"...either directly or through an alias as described above..."

tpepper:
With the second commit's text present, is the first commit actually needed?

CsatariGergely:
@tpepper Maybe if I combine the two?

parispittman:
/lgtm
/approve

thanks!

PR #2542, opened 2018-08-15
Add affiliation requirements

castrojo:
First stab at adding some clarity to steering committee diversity requirements.

/sig contributor-experience
cc @jdumars @spiffxp

### DIFF
--- a/events/elections/2018/README.md
+++ b/events/elections/2018/README.md
@@ -37,11 +37,11 @@ The Steering Committee has selected the following people as [election officers](
 
 ### Schedule
 
-August 15 - Announcement of Election
-September 14 - All candidate bios / statement of intents due. Voting exemption forms due.
-September 19 - Election Begins
-October 3 - Election Closes
-October 4 - Announcement of Results
+- August 15 - Announcement of Election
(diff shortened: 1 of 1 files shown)


spiffxp:
One meta comment before I review:

From elections/README.md:

>The voter’s guide is the single source of truth for the election that year! All announcements and notices should link to this document.

If this document is supposed to be the single source of truth, then [kubernetes/steering/elections.md](https://github.com/kubernetes/steering/blob/master/elections.md) should be removed, and the necessary language merged into this doc.  

I'd like to ask @kubernetes/steering-committee about this during our meeting today.

jdumars:
any steering update on this @spiffxp ?

spiffxp:
Yeah I'm going to get started on review/content in a moment.  Steering consensus is we shouldn't be duplicating content.  Policies should be a living document located in a place covered by a steering-committee OWNERS file.

I'm going to update the kubernetes/steering doc, and this doc should become voter-facing.  Aka what are the dates, who's running etc.

spiffxp:
PTAL at https://github.com/kubernetes/steering/pull/68, I'm writing a PR to supersede this by referencing that more explicitly

PR #2545, opened 2018-08-15
Add sig-cli KEP OWNERS file

soltysh:
The contents where copied from https://github.com/kubernetes/community/blob/master/sig-cli/OWNERS


soltysh:
@cblecker ptal

PR #2552, opened 2018-08-16
KEP for TTL-after-finished controller

janetkuo:


janetkuo:
/assign @tnozicka @kow3ns @enisoc 

cc @kubernetes/sig-apps-pr-reviews

keps/sig-apps/0024-ttl-after-finish.md
```diff
@@ -0,0 +1,290 @@
...
+	TTLSecondsAfterFinished *int32
+}
+```
+
+##### Validation
+
+Because Job controller depends on Pods to exist to work correctly. In Job
+validation, `ttlSecondsAfterFinished` of its pod template shouldn't be set, to
+prevent users from breaking their Jobs. Users should set TTL seconds on a Job,
+instead of Pods owned by a Job.
```
liggitt: will job validation prevent this? how does this work with podgc today? does that mess up jobs?

keps/sig-apps/0024-ttl-after-finish.md
```diff
@@ -0,0 +1,290 @@
...
+delete it, even if it hasn't finished or its TTL hasn't expired. 
+
+### Risks and Mitigations
+
+Risks:
+* Time skew may cause TTL controller to clean up resource objects at the wrong
+  time.
+
+Mitigations:
+* In Kubernetes, it's required to run NTP on all nodes ([#6159][]) to avoid time
```
liggitt: does that ensure clocks are correct inside the controller manager container?

janetkuo: Clocks aren't always correct, but the difference should be small. We can also add a validation to disallow non-zero TTL less than 1 minute.

keps/sig-apps/0024-ttl-after-finish.md
```diff
@@ -0,0 +1,290 @@
...
+  skew. We will also document this risk.
+
+[#6159]: https://github.com/kubernetes/kubernetes/issues/6159#issuecomment-93844058
+
+## Graduation Criteria
+
+This can be promoted to beta when it satisfies users' need for cleaning up
+finished resource objects, without regressions.
+
+This will be promoted to GA once it's gone a sufficient amount of time as beta
```
tnozicka: [From the previous dicussions on the google doc](https://docs.google.com/document/d/1U6h1DrRJNuQlL2_FYY_FdkQhgtTRn1kEylEOHRoESTc/edit?disco=AAAACB9yb9k):

*Because metadata is used by all resources, but not all resources are finishable (only Pods/Jobs). We want to implement this for Pods/Jobs first. If users like this feature, and more use cases from custom resources, we can move it to metadata. To avoid the overhead, we can either remove this alpha field in spec and add a field in metadata, or generalize by keeping it in the spec of each finishable resource.*

I am not sure I want us promoting something to beta/GA if we know that we will replace it with a generic mechanism. We should flesh out the path here a bit more in that regard.

tnozicka: > This will only be promoted to beta after this decision is finalized, i.e. whether we decide to replace it with a generic mechanism or not to do it. Adding this feature as alpha first is essential to help us gather feedback before deciding to generalize it or not.

I agree we should try this as alpha to gather feedback. I was just asking to formalize what you just said as beta graduation criteria in the proposal.

keps/sig-apps/0024-ttl-after-finish.md
```diff
@@ -0,0 +1,290 @@
...
+This will be launched as an alpha feature first, with feature gate
+`TTLAfterFinished`.
+
+#### API Object
+
+We will add the following API fields to `JobSpec` (`Job`'s `.spec`).
+
+```go
+type JobSpec struct {
+ 	// ttlSecondsAfterFinished limits the lifetime of a Job that has finished
```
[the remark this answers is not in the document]
liggitt: doc is supposed to reference the serialized name so the generated doc matches what people see in the API

keps/sig-apps/0024-ttl-after-finish.md
liggitt: it's common for higher level resources to call generic pod spec validation, then add their own restrictions on top:

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/validation/validation.go#L121-L127

keps/sig-apps/0024-ttl-after-finish.md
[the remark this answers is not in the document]
tnozicka: Thanks for the clarification, I wasn't aware API is an exception for the Golang rule where every comment should begin with name of the type.

keps/sig-apps/0024-ttl-after-finish.md
```diff
@@ -0,0 +1,290 @@
...
+finished Job/Pod will be cleaned up after the TTL expires.
+
+## Motivation
+
+In Kubernetes, finishable resources, such as Jobs and Pods, are often
+frequently-created and short-lived. If a Job or Pod isn't controlled by a
+higher-level resource (e.g. CronJob for Jobs or Job for Pods), or owned by some
+other resources, it's difficult for the users to clean them up automatically,
+and those Jobs and Pods can accumulate and overload a Kubernetes cluster very
+easily. Even if we can avoid the overload issue by implementing a cluster-wide
```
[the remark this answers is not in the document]
janetkuo: We currently have `ResourceQuota`, which sets quota restrictions per namespace. Is there a per cluster / global quota available today?

keps/sig-apps/0024-ttl-after-finish.md
```diff
@@ -0,0 +1,290 @@
...
+read Jobs and Pods from a local cache.
+
+#### Finished Jobs
+
+When a Job is created or updated:
+
+1. Check its `.status.conditions` to see if it has finished (`Complete` or
+   `Failed`). If it hasn't finished, do nothing. 
+1. Otherwise, if the Job has finished, check if Job's 
+   `.spec.ttlSecondsAfterFinished` field is set. Do nothing if the TTL field is
```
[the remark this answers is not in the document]
janetkuo: Since both steps merely check the fields of the Job, the order is interchangeable. We don't need to do any further computation if either the job hasn't finished or its ttl it's set.

keps/sig-apps/0024-ttl-after-finish.md
janetkuo: Thanks for the advice. Added validation for restart policy.

keps/sig-apps/0024-ttl-after-finish.md
janetkuo: Thanks, updated graduation criteria.

janetkuo:
Thanks for the review. Rebased and addressed comments. PTAL

keps/sig-apps/0024-ttl-after-finish.md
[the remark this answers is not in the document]
tnozicka: yeah, seems like we have only quota per namespace, best you get to protect the etcd/cluster now is that the cluster object count is limited by multiplying the quota per namespace by the number of namespaces.

keps/sig-apps/0024-ttl-after-finish.md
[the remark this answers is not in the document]
tnozicka: I think there is an important distinction here. `.spec.ttlSecondsAfterFinished` is the field that allows this controller to process those jobs. This is what discriminates if a Job is added to a queue or not and we should be very explicit about this. No other action should be taken on a Job if this field is not set. Logically it doesn't make sense to check a status for a Job is you shouldn't be managing it at all. Controllers/code can evolve an we should structure it well from the start even though now the read-only actions don't cause issues.

smarterclayton:
@stevekuznetsov as someone who has actually implemented a variant of this for a CI cluster, can you comment about your usecase here and give some concrete feedback

keps/sig-apps/0024-ttl-after-finish.md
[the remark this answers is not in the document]
janetkuo: To clarify, Jobs are only added to the workqueue (for cleanup) when they (1) finished (2) have a TTL set. No action will be taken if not both conditions are met.

Issue #2464, opened 2018-08-03
Umbrella issue: ensure all repos present in sigs.yaml


spiffxp:
This is a step before https://github.com/kubernetes/community/issues/1721

We have a number of repos which aren't listed in sigs.yaml ([script](https://gist.github.com/spiffxp/c250d281330bbaea6f634ff3dc65efea))

| org/repo | in sigs.yaml? | has owners? | has approve? | in tide? | branch protected? |
| -------- | ------------- | ----------- | ------------ | -------- | ----------------- |

edit 2018-08-06: down from 25 repos to 18
edit 2018-08-10: down from 18 repos to 16
edit 2018-08-13: down from 16 repos to 6
edit 2018-08-16: down from 16 repos to 2
edit 2018-08-21: down from 2 to 0

/area github-management
/committee steering
/sig contributor-experience
/priority important-soon
/assign

spiffxp:
/close
OK, every repo has at least one entry in sigs.yaml.

We're likely missing a number of subprojects that are distinct sets of packages within or across repos, but this is enough to get started.

PR #2553, opened 2018-08-17
Move k/utils to sig-arch, add k/contrib to sig-arch

spiffxp:
Per discussion in https://github.com/kubernetes/utils/issues/44 I'm
suggesting a move of the kubernetes/utils subproject to sig-arch

I also think sig-arch should own deciding where to move code that
currently lives in k/contrib. This is the last repo not yet included in
sigs.yaml (ref: https://github.com/kubernetes/community/issues/2464)

I have added each of these as their own subprojects. Another approach
could be to make a "code-organization" subproject, and place both of
these repos under that

WDYT?

/sig contributor-experience
/sig architecture
/cc @thockin @lavalamp

/hold
for comment


spiffxp:
Related, for whomever owns k/utils, I want to turn on merge automation for it https://github.com/kubernetes/test-infra/pull/9086

PR #2561, opened 2018-08-20
SIG Multicluster charter

pmorie:
Replaces #1989 as a WIP for SIG multicluster charter.

### DIFF
--- a/sig-multicluster/charter.md
+++ b/sig-multicluster/charter.md
@@ -1,0 +1,47 @@
+# SIG Multicluster Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+The scope of SIG Multicluster is limited to the following subprojects:
+
+- The [cluster-registry](https://github.com/kubernetes/cluster-registry)
+- Kubernetes federation:
+  - [Federation v2](https://github.com/kubernetes-sigs/federation-v2)
+  - [Federation v1](https://github.com/kubernetes/federation)
+- [Kubemci](https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress)
+
+### In scope
+
+See [SIG README].
+
+#### Code, Binaries and Services
+
+SIG Multicluster code and binaries are limited to those from one of the SIG subprojects.
+
+#### Cross-cutting and Externally Facing Processes
+
+- Consult with other SIGs and the community on how the in-scope mechanisms
+  should work and integrate with other areas of the wider Kubernetes ecosystem
+
+### Out of scope
+
+- Software that creates or manages the lifecycle of Kubernetes clusters
+
+## Roles and Organization Management
+
+This sig follows adheres to the Roles and Organization Management outlined in [sig-governance]
+and opts-in to updates and modifications to [sig-governance].
+
+### Subproject Creation
+
+SIG Multicluster delegates subproject approval to Technical Leads. See [Subproject creation - Option 1]. 
+
+[sig-governance]: https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md
+[sig-subprojects]: https://github.com/kubernetes/community/blob/master/sig-multicluster/README.md#subprojects
+[sigs.yaml]: https://github.com/kubernetes/community/blob/master/sigs.yaml#L1042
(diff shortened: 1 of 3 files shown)


marun:
lgtm

sig-multicluster/charter.md
```diff
@@ -0,0 +1,45 @@
+# SIG Multicluster Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+The scope of SIG Multicluster includes:
+
+- Solving common challenges related to management of multiple Kubernetes clusters and the applications that exist therein
```
timothysc: This is really broad, could you scope it down?  

e.g. - what aspects?  
- Networking, application mgmt, disaster recovery, etc.

pmorie: I think the most sensible thing to say here is probably to be really specific about the scopes of the existing projects, which are:

- cluster-registry
- federation-v2
- kubemci

I don't think it's especially useful to have a vague prose description of the scope - since this SIG is essentially focused entirely around these projects at this point.

timothysc: > I think the most sensible thing to say here is probably to be really specific about the scopes of the existing projects 

Agreed, we want to make certain the areas of responsibility translate to specific areas of ownership so we start to rationalize the SIGs across the project.

sig-multicluster/charter.md
```diff
@@ -0,0 +1,45 @@
...
+- Designing, discussing, implementing, and maintaining APIs, tools, and documentation related to multi-cluster administration and application management
+
+### In scope
+
+Link to SIG section in [sigs.yaml]
+
+#### Code, Binaries and Services
+
+- Mechanisms in Kubernetes that track information about multiple clusters
```
spiffxp: Needs to be less vague.  spartakus tracks information about multiple clusters, in the sense that they all ping it... but I doubt that's under your purview

sig-multicluster/charter.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Cross-cutting and Externally Facing Processes
+
+- Consult with other SIGs and the community on how the mechanisms in scope for
+  the SIG should work and integrate with other areas of the wider Kubernetes
+  ecosystem
+
+### Out of scope
+
+- Software that creates or manages the lifecycle of Kubernetes clusters
```
spiffxp: Which SIG (if any) would own reconciliation between cluster-api and cluster-registry?

spiffxp:
Ping, any updates on this?

pmorie:
@spiffxp @timothysc changes pushed, tried to be very specific.

timothysc:
@pmorie am I missing something, I don't see an updated PR?

pmorie:
@timothysc it sure helps if you push to the right branch; apologies!

sig-multicluster/charter.md
pmorie: I had hoped to address that by narrowly defining the scope of the SIG. I will add a note that specifically addresses the situation where another part of the community plans an integration between another piece of software and one that is in scope for the SIG.

timothysc:
/hold 

Pending @spiffxp feedback.

I don't really have issues with the current form, but it may be a bit limiting.  I'd leave that to the chairs to decide. 

/approve

spiffxp:
This looks like over-specification to me, especially compared to the description in sigs.yaml.  Things mentioned there I don't see here:
- tools related to multi-cluster administration and application management
- batch-workflow continuous deployment systems like Spinnaker and others
- changes to kubernetes core where appropriate
- anything around common practices or approaches to the use of multiple clusters

/lgtm
I'll LGTM this as is, but I would anticipate some iteration will need to happen here

spiffxp:
Please `/hold cancel` if you're happy with this in its current form @pmorie

pmorie:
@spiffxp @timothysc I agree that it is very specific at this point, I found it hard to determine exactly what the right level of detail was and decided to err on being overly specific rather than overly broad.

I'm fine iterating on this in follow-ups, canceling hold.

Thanks for the reviews!

/hold cancel

spiffxp:
The /hold command wasn't picked up due to an ongoing github incident.  Linking to see if this gets picked back up when github recovers

https://github.com/kubernetes/test-infra/issues/9881

PR #2584, opened 2018-08-24
Change k8s.io/api owners from apimachinery to architecture

nikhita:
Ref: https://github.com/kubernetes/kubernetes/pull/67672#issuecomment-415560829
/cc dims

/assign lavalamp 
/sig api-machinery

/assign bgrant0607 jdumars 
/sig architecture


lavalamp:
LGTM - I'll let someone else from sig arch actually put on the label

PR #2586, opened 2018-08-24
Introduce definition of network programming latency SLI

wojtek-t:
/assign @thockin @dcbw @caseydavenport @bowei 

FYI: @kubernetes/sig-scalability-feature-requests @kubernetes/sig-network-feature-requests 
@countspongebob @shyamjvs


sig-scalability/slos/network_programming_latency.md
```diff
@@ -0,0 +1,89 @@
...
+load-balancing
+
+### Other notes
+- We are consciously focusing on in-cluster load-balancing for the purpose of
+this SLI, as external load-balancing is clearly provider specific (which makes
+it hard to set the SLO for it).
+- However, in the future it should be possible to formulate the SLI for external
+load-balancing in pretty much the same way for consistency.
+- The SLI measuring end-to-end time from pod creation was also considered,
+but rejected due to being application specific, and thusintroducing SLO would
```
shyamjvs: thus introducing

sig-scalability/slos/network_programming_latency.md
```diff
@@ -0,0 +1,89 @@
...
+to service specification (including creation) will be reflected in in-cluster
+load-balancing
+
+### Other notes
+- We are consciously focusing on in-cluster load-balancing for the purpose of
+this SLI, as external load-balancing is clearly provider specific (which makes
+it hard to set the SLO for it).
+- However, in the future it should be possible to formulate the SLI for external
+load-balancing in pretty much the same way for consistency.
+- The SLI measuring end-to-end time from pod creation was also considered,
```
shyamjvs: from -> for

sig-scalability/slos/network_programming_latency.md
wojtek-t: done

sig-scalability/slos/network_programming_latency.md
wojtek-t: That's actually correct: from pod creation to when it's available via in-cluster lb.

shyamjvs: Ok, I misunderstood that.

sig-scalability/slos/network_programming_latency.md
```diff
@@ -0,0 +1,89 @@
...
+interesting for the user). The reason for doing it this is feasibility for
+efficiently computing that:
+  - if we would be doing aggregation at the SLI level (i.e. the SLI would be
+    formulated like "... reflected in in-cluster load-balancing mechanism and
+    visible from 99% of programmers"), computing that SLI would be extremely
+    difficult. It's because in order to decide e.g. whether pod transition to
+    Ready state is reflected, we would have to know when exactly it was reflected
+    in 99% of programmers (e.g. iptables). That requires tracking metrics on
+    per-change base (which we can't do efficiently).
+  - we admit that the SLO is abit weaker in that form (i.e. it doesn't necessary
```
shyamjvs: a bit

shyamjvs:
A couple of comments (based on our offline discussion):

- If we have N pods and M nodes, then we need to get M*N rules programmed. While it is not very clear, one might say that as a user I want 99% of those M*N rules to be programmed within X seconds. Your SLI which offers 99% of nodes to have 99% of their rules programmed within X seconds is a bit weaker (which you've noted already in the doc). But I suggest adding a brief explanation showing how the latter maps to the former (just to make user expectations bit clearer).
- The SLI in it's current form seems to weigh all pods equally, while in practice one might argue that some pods are much more important than others (for e.g the only pod in a 1-pod service compared to a pod in a 3000-pod service). So wdyt about having a service-level SLI too? Or maybe it's not worth it?

wojtek-t:
> If we have N pods and M nodes, then we need to get MN rules programmed. 

With iptables yes. In general - not necessarily.

> While it is not very clear, one might say that as a user I want 99% of those MN rules to be programmed within X seconds. Your SLI which offers 99% of nodes to have 99% of their rules programmed within X seconds is a bit weaker (which you've noted already in the doc). But I suggest adding a brief explanation showing how the latter maps to the former (just to make user expectations bit clearer).

I'm not sure I can add a `brief` explanation of that to avoid introducing more confusion that it's worth it.
Also (given that your suggestion is just some approximation of what users want) I'm not 100% sure it's actually worth making this explanation additionally longer.

> The SLI in it's current form seems to weigh all pods equally, while in practice one might argue that some pods are much more important than others (for e.g the only pod in a 1-pod service compared to a pod in a 3000-pod service). So wdyt about having a service-level SLI too? Or maybe it's not worth it?

As I mentioned offline, I think it may be worth adding. But for me, that's a completely different SLI and I don't want to mix it here. We tried doing too much in one step in the past and that didn't work. well.
So i would treat it more as AI for the future.

shyamjvs:
That sounds reasonable. Though..

> given that your suggestion is just some approximation of what users want

So what is the metric that users want? :)

> So i would treat it more as AI for the future.

Maybe mention that in the TODO at bottom?

sig-scalability/slos/network_programming_latency.md
```diff
@@ -0,0 +1,89 @@
...
+    formulated like "... reflected in in-cluster load-balancing mechanism and
+    visible from 99% of programmers"), computing that SLI would be extremely
+    difficult. It's because in order to decide e.g. whether pod transition to
+    Ready state is reflected, we would have to know when exactly it was reflected
+    in 99% of programmers (e.g. iptables). That requires tracking metrics on
+    per-change base (which we can't do efficiently).
+  - we admit that the SLO is a bit weaker in that form (i.e. it doesn't necessary
+    force that a given change is reflected in 99% of programmers with a given
+		99th percentile latency), but it's close enough approximation.
+
```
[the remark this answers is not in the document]
wojtek-t: So it was initially defined only in the slos.md file (which is linking this file).

The motivation for that was that the slos.md is the "main and goto page" from where you only go to this file (or other files for other SLIs/SLOs if you want more details).
Apparently it's not that obvious so motivated by this comment (and similar comment from @shyamjvs in the past) and I did: https://github.com/kubernetes/community/pull/2591.

So now we still have this main page (that should be goto page), but these per-slo files contain the definition of SLI and SLO too.

sig-scalability/slos/network_programming_latency.md
```diff
@@ -0,0 +1,89 @@
...
+### How to measure the SLI.
+The method of measuring this SLI is not obvious, so for completeness we describe
+it here how it will be implemented with all caveats.
+1. We assume that for the in-cluster load-balancing programmin we are using
```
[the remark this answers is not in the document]
wojtek-t: done

sig-scalability/slos/network_programming_latency.md
```diff
@@ -0,0 +1,89 @@
+## Network programming latency SLIs/SLOs details
+
```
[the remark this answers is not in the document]
wojtek-t: Should be linked from the main slos.md - will open a separate PR for it.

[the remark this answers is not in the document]
wojtek-t: Fixed in https://github.com/kubernetes/community/pull/2591

wojtek-t:
@thockin @dcbw @caseydavenport @kubernetes/sig-network-api-reviews @kubernetes/sig-network-pr-reviews - I'm going to leave it open for comments for next 3 days and with lgtm from @bowei if no other objections will appear I'm going to merge that and treat as approved.

sig-scalability/slos/network_programming_latency.md
```diff
@@ -0,0 +1,95 @@
...
+This may be the case for both Endpoints controller or kube-proxy (or other
+network programmers if used instead). That becomes a problem when a single
+object changed multiple times in the meantime (otherwise informers will
+deliver handlers on relisting). Additionally, this can happen only when
+components are too slow in processing events (that would already be reflected
+in metrics) or (sometimes) after kube-apiserver restart. Given that, we are
+going to neglect this problem to avoid unnecessary complications for little
+or no gain.
+
+**TODO:**
```
caseydavenport: Is there more coming here?

wojtek-t: Not that I'm aware of now - I should probably remove it.

wojtek-t: Just removed that.

wojtek-t:
@bowei - I just removed the "TODO" word from the text (per @caseydavenport comment). If you could re-lgtm it, would be great.

wojtek-t:
As mentioned in https://github.com/kubernetes/community/pull/2586#issuecomment-418087510, since I didn't hear any objections, I'm going to approve that now.

/approve

PR #2588, opened 2018-08-25
Extending resource quota to support node labels

vishh:


vishh:
cc @kubernetes/api-approvers @derekwaynecarr @bsalamat

jbeda:
/approve

Based on form, not content.  Assuming that others from sig-scheduling are on board, feel free to merge early and iterate on the checked in version. I'd add sig-architecture as an involved SIG so that they are consulted wrt API review as this moves along and before it moves to `implementable`.

vishh:
@davidopp

derekwaynecarr:
per our prior discussion on the earlier draft form, i prefer solution A.

/approve
/lgtm

vishh:
@jdumars can you get this KEP queued up for API review?

hex108:
@vishh Thanks for the great feature! Is there any plan to implement it? We need this feature to manage quota for different GPU cards, I could do some work if it's OK.

PR #2595, opened 2018-08-28
Update userns proposal with FsGroup feature support roadmap and more details on default behavior

vikaschoudhary16:
DEMO: https://www.youtube.com/watch?v=2sEmceRNa6Y

/cc @mrunalp @kubernetes/sig-node-api-reviews @kubernetes/sig-node-feature-requests @kubernetes/sig-node-proposals @kubernetes/sig-auth-api-reviews

### DIFF
--- a/contributors/design-proposals/node/node-usernamespace-remapping.md
+++ b/contributors/design-proposals/node/node-usernamespace-remapping.md
@@ -40,6 +40,7 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
 
 ## Non-Goals
 - Non-goal is to support pod/container level user namespace isolation. There can be images using different users but on the node, pods/containers running with these images will share common user namespace remapping configuration. In other words, all containers on a node share a common user-namespace range.
+- To let user facing API request specific mapping between host usernamespace and container usernamespace
 - Remote volumes support eg. NFS
 
 ## User Stories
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -143,39 +175,107 @@ message NamespaceOption {
...
+
+
+## Implementation Roadmap
+### Phase 1: Support in Kubelet, Alpha, [Target: Kubernetes v1.14]
+- Add `hostUserNamespace` in pod API
+- Add feature gate `HostUserNamespace`, disabled by default
 - Add new CRI API, `GetRuntimeConfigInfo()`
 - Add logic in Kubelet to handle pod creation which includes parsing GetRuntimeConfigInfo response and changing file-permissions in /var/lib/kubelet with learned userns mapping.
 - Add changes in dockershim to implement GetRuntimeConfigInfo() for docker runtime
 - Add changes in CRI-O to implement userns support and GetRuntimeConfigInfo() support
```
dchen1107: cc/ @Random-Liu on CRI-Containerd support

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -143,39 +175,107 @@ message NamespaceOption {
...
-- CRI-O: https://github.com/kubernetes-incubator/cri-o/pull/1519
+- CRI-O: CRI-O supports usernamespace remapping. [Here is the PR](https://github.com/kubernetes-sigs/cri-o/pull/1941) for getting supplemental groups(fsgroups) working in namespaced environnment.
 - Containerd: https://github.com/containerd/containerd/blob/129167132c5e0dbd1b031badae201a432d1bd681/container_opts_unix.go#L149
 
-### Implementation Roadmap
-#### Phase 1: Support in Kubelet, Alpha, [Target: Kubernetes v1.11]
-- Add feature gate `NodeUserNamespace`, disabled by default
+## Sandbox type runtimes
+`HostUserNamespace` will work only with runc based runtimes. Techniques such as node selectors will have to be used in pods(that are using `HostuserNamespace`) to correctly assign runtimes/nodes. If such a pod unexpectedly lands on a sandbox type runtime node, pod admit handler will reject it.
```
dchen1107: I think a node has to support both sandboxed pods and non-sandboxed pods in most cases, but I agreed that pod admit handler still can reject a pod if it request both HostUserNamespace & Sandbox=true.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,44 +49,71 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+### Volume Permissions  
+Kubelet will change the file permissions, i.e chown, at `/var/lib/kubelet/pods/<pod-uid>/volumes` after volume plugin is done with provisioning to get updated ownership of the volume directory according to remapped UID and GID.
+This proposal will work only for local volumes and not with remote volumes which have client-server kind of complex architecture where client is not allowed to chown volumes as root such as NFS.
 
-- **Beta:** Explicit API to request host user-namespace in pod spec
-          This is being targeted under Beta graduation plans.
 
 ### CRI API Changes
-Proposed CRI API changes:
+A new CRI API, `GetRuntimeConfigInfo` will be added. At the initialization, Kubelet will use this API to retrieve default {U/G}ID mappings from the container runtime.
```
Random-Liu: Suggestion about the CRI change https://github.com/kubernetes/kubernetes/pull/64005#issuecomment-452429232.

Basically, I don't think we should define CRI based on how user namespace is implemented in Docker today. There is a better way to define this API, and also get the node level remapping support we proposed for alpha.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
-  - https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c#diff-a53fa76e941e0bdaee26dcbc435ad2ffR437 introduced via https://github.com/kubernetes/kubernetes/commit/d0d78f478ce0fb9d5e121db3b7c6993b482af82c.
+* **RunAsUser**: UID to run the entrypoint of the container process. After usernamespace support in kubelet, if Kubelet discovers that usernamespace remapping is enabled at runtime, this will be the UID in the container usernamespace. In such a case, if no mapping(corresponding host UID) is found in the mapping range, pod will be failed admission at Kubelet. If kubelet discovers that usernamespace remapping is not supported at runtime, no validation check will be performed. 
+* **RunAsGroup**: same as **RunAsUser** but for gid mapping ranges.
+* **RunAsNonRoot**: Today, kubelet fails to start the container if image is found to be having UID 0. After usernamespace support in kubelet, kubelet should ignore this check if runtime is found to be running with usernamespace remapping enabled. Kubelet should verify image UID only when it is learned that usernamespace remapping is either not supported or not enabled at runtime.
```
tallclair: > kubelet should ignore this check if runtime is found to be running with usernamespace remapping enabled

I'm not sure I agree with this. There might still be use cases for requiring a non-root users in a user namespace (root users do have elevated privileges within the namespace, e.g. CAP_DAC_OVERRIDE). It also makes auditing difficult, especially in heterogeneous clusters.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
 
-- **Beta:** Explicit API to request host user-namespace in pod spec
-          This is being targeted under Beta graduation plans.
+    If pod has `FsGroup: 2000` and runtime's default gid mappings are `containerid 0,  hostid 100000, size 1000`, runtime should be adding two entries in the `/proc/<container-pid>/gid_map`:
+    - [0 100000 1000]  // From GID 0 to 1000 in the container are mapped with GID range 100k to 100k+1000 on host
+    - [2000 2000 1]    // GID 2000 in container usernamespace is mapped to GID 2000 is host usernamespace
```
tallclair: What if there's overlap? E.g. what if it was `FsGroup: 500` in this example?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
-### CRI API Changes
-Proposed CRI API changes:
+    Once this 1:1 mapping is set for the container process, `runc` already has the logic to call `setgroups()` to set the FsGroupID as the supplemental ID in the container usernamespace.
 
-```golang
-// Runtime service defines the public APIs for remote container runtimes
-service RuntimeService {
+    Setting the permissions on volume is resposibility of volume plugin. Once GID on the volume matches the supplemental GID in the container, access will work.
+
+    NOTE: Pod will be failed admission at kubelet if FsGroup/SupplementalGroups are in the default mapping range and mappings are not 1:1 i.e host id for a container id is different.
```
tallclair: Soft failure (`BLOCKED`) or hard failure? It seems problematic that this failure isn't known until the pod is actually run on a node.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+
+
+### Volume Permissions  
+Kubelet will change the file permissions, i.e chown, at `/var/lib/kubelet/pods/<pod-uid>/volumes` after volume plugin is done with provisioning to get updated ownership of the volume directory according to remapped UID and GID.
```
tallclair: What if the volume is mounted by multiple pods? What if it's mounted by multiple nodes?

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+### Phase 2: Beta Support [Target: Kubernetes v1.15]
 - PSP integration
-- To grow ExperimentalHostUserNamespaceDefaulting from experimental feature gate to a Kubelet flag
-- API changes to allow pod able to request HostUserNamespace in pod spec
-- e2e tests
+- Namespace-Level usernamespace remapping support
 
-### References
+## Rollout Strategy
+Upgrading a cluster to 1.14 will not impact any existing workloads (Assuming that existing cluster nodes were not running runtimes with usernamespace remapping enabled). To be able to use the security benefit of usernamespace remapping on existing node, usernamespace remapping will have to enabled at the runtime. To perform this runtime configuration update, first node will be tainted to make it unschedulable, then it will be drained and runtime will be updated. Next, the taint will removed to make it available to the schedular. Now Kubelet will discover this configuration change and starts creating pods with remapped usernamespace.
```
tallclair: nit: s/schedular/scheduler/

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+As we discussed in above sections, Values of `RunAsUser`, `RunAsGroup`, `SupplementalGroups` and `FsGroup` in the older/existing templates may cause pod admission failure if the values are not consistent with the default id mappings of the runtime as learned by Kubelet.
+
+`RunAsUser` and `RunAsGroup` values are treated as translated IDs in the container usernamespace and if the runtime's default mappings does not have a mapping for these containerIds, pod will fail admission at Kubelet.
+
+`FsGroup` and `SupplementalGroups` values are are not treated as translated IDs. For a FsGroup/SupplementalGroup ID, container usernamespace ID is mapped to the same host usernamespace ID. Therefore for these GIDs in the pod objects in the older templates, either there should be no mapping in runtime's default id mappings or if mapping is there, container id and host id should be same fopr the mapping range i.e 1:1 mapping.
+
+### Recommended default id mappings configuration for the runtime
+- Map `0` {U/G}ID in the container namespace with max {u/g}id i.e `4294967295`
+- Map `1` ID to `4294967294`(max id -1) ID in the container namespace with the exactly same IDs on host user namespace
```
tallclair: sounds good, but then any container using supplemental or fsgroups will fail. will the kubelet detect that no translation is needed, and not cause problems?

contributors/design-proposals/node/node-usernamespace-remapping.md
Random-Liu: Hm, good catch.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+    - Container (U/G)ID: First (U/G)ID of the range in the container namespace and this is mapped to the first (U/G)ID on the host(mentioned in previous point).
+    - Count/Size: Total number of consecutive mapping between host and container user-namespaces, starting from the first one (including) mentioned above.
 
 As an example, `host_id 1000, container_id 0, size 10`
 In this case, 1000 to 1009 on host will be mapped to 0 to 9 inside the container.
 
-User-namespace support should be enabled only when container runtime on the node supports user-namespace remapping and is enabled in its configuration. To enable user-namespaces, feature-gate flag will need to be passed to Kubelet like this `--feature-gates=”NodeUserNamespace=true”`
+2. Enabling usernamespace remapping at runtime will not require any changes in the existing pod yamls. Kubelet will make sure that older pod yamls will continue to work even in usernamespace remapped configuration. There are valid pod configurations which cannot be supported in usernamespace remapped environment, ex:
+  - host namespaces (pid, ipc, net)
+  - non-namespaced capabilities (mknod, sys_time, sys_module)
```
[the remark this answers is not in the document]
vikaschoudhary16: sure, will verify and update if required.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+  - The pod contains a privileged container or using host path volumes.
+To maintain **backward compatibility**, if such a pod configuration is detected by Kubelet, usernamespace remapping will be disabled, implicitly, for each of such pod in isolation by passing case appropriate `NamespaceOption` to the runtime. For more details see the CRI section below. Other pods will be run with remapped usernamespace.
+
+### Pod API Change
+`HostUserNamespace` field will be introduced in the pod spec. Its meaning is similar to other `Host*` fields the pod spec. If set to `true`, pod will share usernamespace from underlying host.
```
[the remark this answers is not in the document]
vikaschoudhary16: It was discussed earlier also and consensus was to granularity of usernamesapce remapping as kubelet implementation detail. Leaving user with the choice of either "hostUsernamespace" or "usernamespace remapping security"..whether its node level, pod level or namespace level, let it be internal to kubelet.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+```
+        // Use host's user namespace for the pod.
+        // This field is alpha-level and can be set to true/false only if HostUserNamespace feature-gate is enabled.
+        // Optional: Defaults to nil which means behavior will be container runtime defined.
+        // +k8s:conversion-gen=false
+        // +optional
+        HostUserNamespace *bool `json:"hostUserNamespace,omitempty" protobuf:"varint,31,opt,name=hostUserNamespace"`
+```
+Today end users are NOT provided any guaranteed behavior by the Kubernetes related to usernamespace remapping support at runtime. If runtime has usernamespace remapping disabled, pods will run otherwise pods will fail. In other words, behavior is container runtime defined. `nil` value of `HostUserNamespace` will show the same behavior i.e runtime defined usernamespace behavior. However, with an added improvement of running pods with remapped usernamespace "when possible". "when possible" means runtime is running with usernamespace remapping enabled and pod does not specify any configuration which is not possible to provide in remapped usernamespace.
```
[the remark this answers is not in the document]
vikaschoudhary16: With existing Kubelet, if userns is enabled at runtime, pod containers will fail to start because of permission errors on volumes.

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: Agree with your point. `RunAsNonRoot`, just like `RunAsUser` and `RunAsGroup`, should be conveying meaning inside container usernamespace and thus existing check should be performed. Will update.

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: Will fail at kubelet admission

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: should be hard failure because pod spec or runtime configuration will not change on retries and does not seem recoverable. WDYT?

contributors/design-proposals/node/node-usernamespace-remapping.md
vikaschoudhary16: yeah, either kubelet can detect and will not set supplemental groups in the pod creation request which is sent to the runtime.
OR
Can runtimes be expected to understand this and behave properly?

I think first is more sane. WDYT?

contributors/design-proposals/node/node-usernamespace-remapping.md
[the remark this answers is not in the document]
tallclair: Ah, makes sense, please update with that language.

contributors/design-proposals/node/node-usernamespace-remapping.md
tallclair: The problem with hard failure is that the controllers will retry and cause churn. This is a general problem though, which we've discussed fixing in the controller layer.

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,248 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
-// Runtime service defines the public APIs for remote container runtimes
-service RuntimeService {
+    Setting the permissions on volume is resposibility of volume plugin. Once GID on the volume matches the supplemental GID in the container, access will work.
+
+    NOTE: Pod will be failed admission at kubelet if FsGroup/SupplementalGroups are in the default mapping range and mappings are not 1:1 i.e host id for a container id is different.
+
+
+### Volume Permissions  
+Kubelet will change the file permissions, i.e chown, at `/var/lib/kubelet/pods/<pod-uid>/volumes` after volume plugin is done with provisioning to get updated ownership of the volume directory according to remapped UID and GID.
+This proposal will work only for local volumes and not with remote volumes which have client-server kind of complex architecture where client is not allowed to chown volumes as root such as NFS.
```
tallclair: Hmm, that seems sort of confusing. So, if I'm using user NS remapping, some volumes will be chown'ed, and others will be left as-is?

vikaschoudhary16: If any of the volumes is found to be NFS volume, pod will be defaulted to use host usernamespace, like in case of host path volumes.

tallclair: I still think that's confusing. The user needs to know whether a given volume type respects user namespace remapping before they use it. That seems a especially problematic when CSI is thrown into the mix.

contributors/design-proposals/node/node-usernamespace-remapping.md
tallclair: Would the chown be recursive, or just apply to the volumes root directory?

saad-ali:
Discussed at SIG Storage Meeting. Notes [here](https://docs.google.com/document/d/1-8KEG8AjAgKznS9NFm3qWqkGyCHmvU6HVl0sk5hwoAE/edit#heading=h.72zss16zktf5).

Summary:
* Counter proposal: Can we just use FSGroups? Since each container maybe running as different UID
  * If no FSGroup specified -- no change of ownership
  * If FSGroup is specified -- adjust it to based on UID mapping
* Perf issues with recursive chown
  * SIG storage looking in to it. https://github.com/kubernetes/enhancements/pull/696

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,286 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+        // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+        Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+        // VolumeSource represents the location and type of the mounted volume.
+        // If not specified, the Volume is implied to be an EmptyDir.
+        // This implied behavior is deprecated and will be removed in a future version.
+        VolumeSource `json:",inline" protobuf:"bytes,2,opt,name=volumeSource"`
+        // Optional: Defaults to value of hostUserNamespace field in the pod spec
+        // +k8s:conversion-gen=false
+        // +optional
+        HostUserNamespace *bool `json:"hostUserNamespace,omitempty" protobuf:"varint,3,opt,
```
nathanjsweet: This troubles me. I think this might be the first feature in K8s where two separate resources will have to coordinate well with one another in order to work. Forcing the user to think about namespaces at this level is a recipe for trouble, IMO. Also, how will K8s handle node failures and premature detaches? If these volumes get chowned, before getting unchowned. Users volumes will be in a broken state.

It would almost be better to fail (i.e. not handle the case of volumes yet).

contributors/design-proposals/node/node-usernamespace-remapping.md
```diff
@@ -48,134 +49,286 @@ Enable user namespace support in a kubernetes cluster so that workloads that wor
...
+    // mapped to uid/gid 200000 on node namespace.
+    POD       = 0;
+    // A CONTAINER namespace is restricted to a single container.
+    // For example, a container with a PID namespace of CONTAINER expects to
+    // view only the processes in that container.
+    CONTAINER = 1; 
+    // A NODE namespace is the namespace of the Kubernetes node.
+    // For example, a container with a PID namespace of NODE expects to view
+    // all of the processes on the host running the kubelet.
+    NODE      = 2;
```
krmayankk: which of these will be default ? What is the current mode when i am not using any of these features ?

PR #2596, opened 2018-08-28
contributors/guide/owners: Document OWNERS' 'filters' key

wking:
Catching up with kubernetes/test-infra#6411.


contributors/guide/owners.md
```diff
@@ -49,6 +51,22 @@ reviewers:
   - sig-foo # this is an alias
 ```
 
+#### Filters
+
+An OWNERS file may also include a `filters` key.
+The `filters` key is a map whose keys are regular expressions and whose values are [simple OWNERS configurations](#owners).
```
wking: It would be nice if we explained the paths that the regular expressions are matched against.  Are they absolute?  Relative to the repo root?  Relative to the `OWNERS` file they came from?  I'm having a hard time figuring this out by reading the source.   @cjwagner, @BenTheElder, do either of you remember off the top of your head?

cjwagner: Ah, I didn't know the regexp filtering was documented anywhere. We had intentionally left it undocumented, because it is not handled properly by the `approve` plugin (and possibly `blunderbuss`).  The `approve` plugin will need a significant rework to work well with regexp filtering.

cjwagner: https://github.com/kubernetes/test-infra/issues/7690

wking: > Ah, I didn't know the regexp filtering was documented anywhere.

It's not.  This PR was my attempt at documenting it, after I ran across `filters` in the wild and wondered what it was ;).

> We had intentionally left it undocumented...

Hmm.  Maybe we should hold on this PR until that gets fixed?  Or land this documenting the limitation?

Do you know the answer to my path question?

cjwagner: > Maybe we should hold on this PR until that gets fixed? Or land this documenting the limitation?

I think the later makes the most sense. For now the filters should only be used with the labels section. 

> Do you know the answer to my path question?

The paths are relative to the OWNERS file they are in.


We should also note that the `filters` section is mutually exclusive with the normal simple OWNERS config at the root of the file.

wking: > We should also note that the filters section is mutually exclusive with the normal simple OWNERS config at the root of the file.

How does that combine with the `approve` limitation?  Does it mean that an `OWNERS` file which sets `filters` should currently only set `labels`, and that you shouldn't do something like:

```yaml
approvers:
- alice
filters:
  "\\.go$":
    labels:
    - re/go
```

wking:
I've pushed 99889e3 -> 8c7e130 documenting the regex dialect, relative path comparison, and the current restriction to `labels`.  Still TODO: documenting the restriction against `filters` an a simple configuration in the same `OWNERS` file (once [I understand what that limitation is][2] ;).

[1]: https://github.com/kubernetes/community/pull/2596#discussion_r213768438
[2]: https://github.com/kubernetes/community/pull/2596#discussion_r213772511

contributors/guide/owners.md
cjwagner: They are unrelated limitations.
 The first is a formatting limitation. You can't specify a simple config if you specify `filters`. You should instead specify an entry in `filters` for the `.*` key. So your example would still be invalid even if it was:
```
labels:
- foo
filters:
  "\\.go$":
    labels:
    - re/go
```
The second is a limitation due to bugs. You should only set the `labels` section if you are using `filters`. So this is invalid:
```
filters:
  "\\.go$":
    approvers:
    - alice
```

wking:
I've pushed 8c7e130 -> 99f251c documenting [the sibling limitation][1].  I think that's everything that's come up in review so far.

[1]: https://github.com/kubernetes/community/pull/2596#discussion_r213788099

PR #2604, opened 2018-08-30
Clarify some things in main slo page

wojtek-t:
/assign @shyamjvs

### DIFF
--- a/sig-scalability/slos/slos.md
+++ b/sig-scalability/slos/slos.md
@@ -11,26 +11,29 @@ in these areas.
 
 ## What do we require from SLIs/SLOs?
 
-We are going to define more SLIs and SLOs based on the most important indicators
-in the system.
+We are in the process of extending the number of SLIs ([Service Level Indicators])
+and SLOs ([Service Level Objectives]) built on top of these SLIs to cover more areas
+of the system and user expectations.
 
-Our SLOs need to have the following properties:
+Our SLIs/SLOs need to have the following properties:
 - <b> They need to be testable </b> <br/>
-  That means that we need to have a benchmark to measure if it's met.
+  Ideally, they (SLIs and SLOs) should be measurable in all running clusters,
+	but if   that isn't possible a benchmark may be enough in some situations.
+  That means that not every SLO may be translatable to SLA ([Service
+  Level Agreement]).
 - <b> They need to be understandable for users </b> <br/>
   In particular, they need to be understandable for people not familiar
   with the system internals, i.e. their formulation can't depend on some
   arcane knowledge.
 
-However, we may introduce some internal (for developers only) SLIs, that
-may be useful for understanding performance characterstic of the system,
-but for which we don't provide any guarantees for users and thus may not
-be fully understandable for users.
+We may also introduce internal(for developers only) SLIs, that may be useful
+for understanding performance characterstic of the system, but for which
+we don't provide any guarantees for users (and thus don't require them to be
+that easily understandable).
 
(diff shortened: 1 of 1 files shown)


sig-scalability/slos/slos.md
```diff
@@ -11,26 +11,29 @@ in these areas.
 
 ## What do we require from SLIs/SLOs?
 
-We are going to define more SLIs and SLOs based on the most important indicators
-in the system.
+We are in the process of extending the number of SLIs ([Service Level Indicators])
+and SLOs ([Service Level Objectives]) built on top of them to cover more areas
```
shyamjvs: What is 'them' in 'on top of them'?

wojtek-t: SLIs - fixed.

sig-scalability/slos/slos.md
```diff
@@ -11,26 +11,29 @@ in these areas.
...
-We are going to define more SLIs and SLOs based on the most important indicators
-in the system.
+We are in the process of extending the number of SLIs ([Service Level Indicators])
+and SLOs ([Service Level Objectives]) built on top of these SLIs to cover more areas
+of the system and user expectations.
 
 Our SLOs need to have the following properties:
 - <b> They need to be testable </b> <br/>
-  That means that we need to have a benchmark to measure if it's met.
+  Ideally, they should be measurable in the running cluster, but if
```
shyamjvs: Could you clarify what is not measurable, SLI or SLO?

wojtek-t: Both :)
But it's mostly about the SLI. SLO is generally defined as some aggregation of SLI <= X, which is generally quite easy to compute. SLI is what might be hard to measure in all running clusters.

wojtek-t: Done.

sig-scalability/slos/slos.md
```diff
@@ -11,26 +11,29 @@ in these areas.
...
+and SLOs ([Service Level Objectives]) built on top of these SLIs to cover more areas
+of the system and user expectations.
 
-Our SLOs need to have the following properties:
+Our SLIs/SLOs need to have the following properties:
 - <b> They need to be testable </b> <br/>
-  That means that we need to have a benchmark to measure if it's met.
+  Ideally, they (SLIs and SLOs) should be measurable in all running clusters,
+	but if   that isn't possible a benchmark may be enough in some situations.
+  In other words, not every SLO may be translantable to SLA ([Service
```
shyamjvs: -> translatable

wojtek-t: done

sig-scalability/slos/slos.md
shyamjvs: > But it's mostly about the SLI. SLO is generally defined as some aggregation of SLI <= X, which is generally quite easy to compute. SLI is what might be hard to measure in all running clusters.

Yes, that's why I was asking :)

Still there's one more thing that I'm not sure about. How do you infer "In other words, not every SLO may be translantable to SLA" from the earlier lines?

wojtek-t: If it's not possible to measure it on every running cluster (i.e. if we only have benchmark), you obviously can't have SLA for it (because you're not able to measure it for user clusters).

shyamjvs: Then I wouldn't say "In other words", but rather "So that means" (as it's not an equivalence, but an implication).

wojtek-t: done.

PR #2192, opened 2018-05-25
Add icons set, docs and slidedeck

etiennecoutaud:
Refers [kubernetes/website #747](https://github.com/kubernetes/website/issues/747)

Icons are from: [https://github.com/octo-technology/kubernetes-icons](https://github.com/octo-technology/kubernetes-icons)

/cc @castrojo

PR #2609, opened 2018-08-30
Add importable diagram libraries for icon sets and logo

scottrigby:
I've been using the community icons in diagrams since #2192 - they're fantastic 🎉 

A lot of the diagrams I'm making now are for [Helm Charts](https://github.com/helm/charts), which I co-maintain.

I'm not sure if these importable libraries are appropriate for the community repo, but I'm opening this PR for discussion about that. Perhaps they are?

This PR currently includes SVG libraries matching the iconset directories in this repo, as well as one for the [SVG Kubernetes logos](https://github.com/kubernetes/kubernetes/tree/master/logo). 

The libraries are currently only for draw.io. I'm not able to do this with Lucidchart because of a [SVG import and Text object](https://lucidchart.zendesk.com/hc/en-us/community/posts/360014618646-SVG-import-and-Text-object) bug. We could export only the unlabeled icons for Lucidchart, or use PNGs for the labeled ones, but I didn't do that yet. Up for discussion.

Is this valuable to others?


thockin:
I don't know what libraries are for?  How will they be maintained over time?

I lean towards less-is-more for this repo (I argued against including slide templates, too).

scottrigby:
Hey @thockin 👋  Ah right, so image libraries are what draw.io calls importable data & metadata for a set of images. People have asked in a few different places how they can get k8s icons into diagram apps - this is just an example to make it easier to use these icons (the import files themselves are just JSON including base64 SVG object data, with some metadata like default height and width etc). Maybe the directory should change (to `icons/tools/draw.io`) so it's less confusing.

I'm adding a quick gif I just made - below - to show how it works:

kubernetes community pull 2609

In general I agree with the less-is-more approach. But this made it much easier for me to whip out k8s diagrams, so maybe this PR will help others, whether or not this seems worth adding to this repo. Maybe others know of more apps out there with similar export/import capabilities, and we can discuss the best place to keep those for reuse? 😄

scottrigby:
>How will they be maintained over time?

Good point, it should be automated. If this seems worthwhile, I could quickly script that (similar to [rasterize.sh](https://github.com/kubernetes/community/blob/master/icons/tools/rasterize.sh)).

scottrigby:
Hey @cblecker 👋 I just added a simple(ish) shell script to create the files, with a note on how to run with docker. It's the same level of automation as the existing icons rasterize script.

It sounds like the goal for this is to make maintenance easy, whenever icons change - right now both the existing `rasterize.sh` script and this new one make that easier, but they still expect users will run them manually to update the SVG derivatives, then add those updates to any PR that changes the base SVG icons.

Ideally we'd fully automate that (whenever icons change in master, generate all derivatives when base SVGs change - both the PNGs and these new draw.io metadata files - then push the derivative updates to master as well). But unless we want to give CI the ability push new commits back to this git repo (security?), what do we want CI to do? Would it fail a PR whenever the base icon SVGs change without also including updates to the derivatives in the same PR? I'm guessing that's outside the scope of this PR though?

cblecker:
@scottrigby The typical pattern for these would be to have CI run a verify script in a temp folder, and then diff to compare and see if anything changed.

If nothing changes, then the script should pass (exit 0). If there is a diff, then the PR author needs to do something (like run the script themselves and commit the changes. In this case, the script should print a message to stderr stating what failed and what the PR author should do, before exiting with a non-zero exit code.

For an example, check `hack/verify-generated-docs.sh`.

scottrigby:
@cblecker OK I added the automated check. Also followed the existing pattern and added a Make target (with and without Docker) 👍 Please let me know if I should adjust in any way.

Note the existing PNGs (also generated by a script from the same base SVG icons) don't have a diff check yet. Would you like me to make a follow-up PR for that (add a similar Make target and CI diff check script)?

cblecker:
/assign
Adding to my review queue. May take a bit as we're in a crunch before the v1.12 release :)

scottrigby:
After a long while, just got back to check on this PR - reopened.

Not sure why the job failed, it was passing until I added the k8s license to the script in 0b30fbe. Any ideas?

cblecker:
Looks like CI is failing based on 3 icons not being up to date.

If you run `make generate-icons-drawio` you can see three files are up to date.

PR #2620, opened 2018-08-31
Initial bug-bounty scope

tallclair:
First attempt at the scope for Kubernetes bug bounty program.

I wasn't sure here to put the file. LMK if you have a better idea.

/assign @cjcullen @liggitt @philips @jessfraz
/cc @destijl @mayakacz

### DIFF
--- a/contributors/guide/bug-bounty.md
+++ b/contributors/guide/bug-bounty.md
@@ -1,0 +1,79 @@
+# Kubernetes Bug Bounty Program
+
+This program is a **work in progress**. This tracks the currently proposed program, but the bug
+bounty is not currently active. If you currently have a bug to submit, follow instructions at https://kubernetes.io/docs/reference/issues-security/security/
+
+## Scope
+
+The scope of the Kubernetes project itself is not well defined, which makes defining a bug bounty
+scope difficult. With that in mind, we've attempted to enumerate the features, configurations, and
+components that clearly fall inside or outside of our scope. However, this leaves a lot of gray area
+in between. For that, we ask you to use your best judgment, and welcome contributions to help
+clarify the scope.
+
+### In Scope
+
+The following items are explicitly in-scope for the bug bounty program:
+
+**Cluster Attacks:**
+
+- Attacks against Beta & GA features, unless explicitly excluded below
+- Privilege escalation due to bugs in RBAC, ABAC, pod security policies
+- Authentication bugs in the in-tree authentication handlers<br>
+  _Including: OIDC, x509 certificates, service accounts, webhook authenticator, bearer token, etc._
+- Privilege escalation through the kubelet APIs
+- Remote code execution in kubelet, api server
+- Unauthorized etcd access via the Kubernetes API
+- Path traversal attacks in API, namespaces, etcd
+- Info leak (e.g. workload names) from publicly accessible unauthenticated endpoints<br>
+  _Excluding intentionally disclosed info, such as Kubernetes version & enabled APIs_
+- Reliable suppression of audit logs for privileged actions
+- Unexpected editing, removal, or permission changes of files on the host filesystems from
+  Kubernetes components (e.g. kubelet)
+- Persistent DoS from within a cluster by an unprivileged container or user.
+
+**Supply Chain:** _(excluding social engineering attacks against maintainers)_
+
+- Unauthorized code commit to any Kubernetes org repository<br>
+  _Including: `github.com/kubernetes{,-client,-csi,-incubator,-retired,-security,-sigs}/*`_
+- Unauthorized access to github.com/kubernetes-security
+- Publishing of unauthorized artifacts
+- Unauthorized modification of github data
+- CI/CD Credential Leaks
+- Execution inside the CI/CD infrastructure
+
+**Components:**
+
+- Attacks against a stable & supported Kubernetes release (most recent 3 releases)
+- Community maintained stable cloud platform plugins<br>
+  _Vulnerabilities in other cloud platform plugins should be reported through the associated provider_
+- In-tree (k8s.io/kubernetes) stable volume plugins
+
+### Out of Scope
+
+The following items are explicitly out-of scope for the bug bounty program. While we still welcome
+vulnerability reports in these areas, they are not (currently) eligible to receive a bounty.
+
+- Alpha features & APIs
+- Kubernetes running on Windows or other non-linux operating systems
+- Non-Kubernetes binaries distributed as cluster addons<br>
+  _Please report vulnerabilities in these components through the appropriate channel for the
+  upstream component_
+- Container escalations and escapes to the host, unless the attack path traverses a Kubernetes
+  process (e.g. kubelet).
+- Linux privilege escalations<br>
+  _Please report these through security@kernel.org_
+- Attacks against containers from the host they are running on
+- Attacks relying on insecure configurations (subject to the [Product Security Team][]'s opinion),
+  such as clusters not utilizing mutual authentication or encryption between Kubernetes components.
+- Attacks relying on or against deprecated components (e.g. gitrepo volumes)
+- Vulnerabilities in etcd<br>
+  _Please report these through [CoreOS's disclosure process][]_
+- Vulnerabilities in CoreDNS<br>
+  _Please report these through [CoreDNS's disclosure process][]_
+- Vulnerabilities specific to a hosted Kubernetes setup<br>
+  _Please report these through the associated provider_
+
+[Product Security Team]: https://github.com/kubernetes/sig-release/blob/master/security-release-process-documentation/security-release-process.md#product-security-team-pst
+[CoreOS's disclosure process]: https://coreos.com/security/disclosure/
+[CoreDNS's disclosure process]: https://github.com/coredns/coredns#security


contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,67 @@
+# Kubernetes Bug Bounty Program
+
+This program is a **work in progress**. This tracks the currently proposed program, but the bug
+bounty is not currently active.
```
mayakacz: Add a note: if you currently have a bug to submit, follow instructions at https://kubernetes.io/docs/reference/issues-security/security/

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Scope
+
+The scope of the Kubernetes project itself is not well defined, which makes defining a bug bounty
+scope difficult. With that in mind, we've attempted to enumerate the featuers, configurations, and
```
mayakacz: typo "features"

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,67 @@
...
+
+The following items are explicitly in-scope for the bug bounty program:
+
+**Cluster Attacks:**
+
+- Privilege escalation due to bugs in RBAC, ABAC, pod security policies
+- Authentication bugs due to OIDC, x509 certificate handling,  
+- Privilege escalation through the kubelet APIs
+- Remote code execution in kubelet, api server
+- Arbitrary key injection to etcd via the Kubernetes API
```
mayakacz: should this be more general?

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,67 @@
...
+- Path traversal attacks in API, namespaces, etcd
+- Info leak from publicly accessible unauthenticated endpoints
```
mayakacz: clarify that you don't just mean discovering the existence of those endpoints

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,67 @@
...
+- Community maintained stable cloud platform plugins<br>
+  _Vulnerabilities in other cloud platform plugins should be reported through the associated provider_
+- In-tree (k8s.io/kubernetes) stable volume plugins
+
+### Out of Scope
+
+The following items are explicitly out-of scope for the bug bounty program. While we still welcome
+vulnerability reports in these areas, they are not (currently) eligible to receive a bounty.
+
+- Alpha features & APIs
```
mayakacz: and Beta? if not, explicitly state above

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,67 @@
...
+- Non-Kubernetes binaries distributed as cluster addons<br>
+  _Please report vulnerabilities in these components through the appropriate channel for the
+  upstream component_
+- Container escalations and escapes to the host, unless the attack path traverses a Kubernetes
+  process (e.g. kubelet).
+- Linux privilege escalations<br>
+  _Please report these through security@kernel.org_
+- Attacks against containers from the host they are running on
+- Attacks relying on insecure configurations, such as clusters not utilizing mutual authentication
+  or encryption between Kubernetes components.
```
mayakacz: this is going to be the hardest to define. Is there a specific hardening or config guide we want to suggest?
else, would make more specific/ remove. This can be too broad

contributors/guide/bug-bounty.md
tallclair: Done.

contributors/guide/bug-bounty.md
tallclair: done

contributors/guide/bug-bounty.md
tallclair: how's this?

> Unauthorized etcd access via the Kubernetes API

contributors/guide/bug-bounty.md
tallclair: how's this?

>Info leak (e.g. workload names) from publicly accessible unauthenticated endpoints
>  _Excluding interntionally disclosed info, such as Kubernetes version & enabled APIs_

contributors/guide/bug-bounty.md
tallclair: done. We decided to include beta features, since they are enabled by default.

contributors/guide/bug-bounty.md
tallclair: Yeah, this ate up a big portion of our meeting, but I don't think we landed on anything.

Anyone have suggestions? Or should we just remove it?

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,69 @@
...
+
+- Attacks against Beta & GA features, unless explicitly excluded below
+- Privilege escalation due to bugs in RBAC, ABAC, pod security policies
+- Authentication bugs due to OIDC, x509 certificate handling,  
+- Privilege escalation through the kubelet APIs
+- Remote code execution in kubelet, api server
+- Unauthorized etcd access via the Kubernetes API
+- Path traversal attacks in API, namespaces, etcd
+- Info leak (e.g. workload names) from publicly accessible unauthenticated endpoints<br>
+  _Excluding interntionally disclosed info, such as Kubernetes version & enabled APIs_
```
liggitt: typo: `interntionally`

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,69 @@
...
+
+### In Scope
+
+The following items are explicitly in-scope for the bug bounty program:
+
+**Cluster Attacks:**
+
+- Attacks against Beta & GA features, unless explicitly excluded below
+- Privilege escalation due to bugs in RBAC, ABAC, pod security policies
+- Authentication bugs due to OIDC, x509 certificate handling,
```
liggitt: more intended to be included in this list? service accounts, webhook auth, and request header auth seem in scope

contributors/guide/bug-bounty.md
liggitt: https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/ hits most of the highlights, but doesn't mention PodSecurityPolicy

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,70 @@
...
+  _Vulnerabilities in other cloud platform plugins should be reported through the associated provider_
+- In-tree (k8s.io/kubernetes) stable volume plugins
+
+### Out of Scope
+
+The following items are explicitly out-of scope for the bug bounty program. While we still welcome
+vulnerability reports in these areas, they are not (currently) eligible to receive a bounty.
+
+- Alpha features & APIs
+- Kubernetes running on Windows or other non-linux operating systems
```
mattfarina: Windows server containers are a beta feature IIRC. How does this relate to the earlier beta comment?

philips: @mattfarina it explicitly makes Windows server containers out of scope despite the beta status.

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,70 @@
...
+scope difficult. With that in mind, we've attempted to enumerate the features, configurations, and
+components that clearly fall inside or outside of our scope. However, this leaves a lot of gray area
+in between. For that, we ask you to use your best judgment, and welcome contributions to help
+clarify the scope.
+
+### In Scope
+
+The following items are explicitly in-scope for the bug bounty program:
+
+**Cluster Attacks:**
```
dims: @tallclair which category would CVE-2017–1002101 and CVE-2017–1002102 fall under in this list?

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,70 @@
...
+- Privilege escalation due to bugs in RBAC, ABAC, pod security policies
+- Authentication bugs in the in-tree authentication handlers<br>
+  _Including: OIDC, x509 certificates, service accounts, webhook authenticator, bearer token, etc._
+- Privilege escalation through the kubelet APIs
+- Remote code execution in kubelet, api server
+- Unauthorized etcd access via the Kubernetes API
+- Path traversal attacks in API, namespaces, etcd
+- Info leak (e.g. workload names) from publicly accessible unauthenticated endpoints<br>
+  _Excluding intentionally disclosed info, such as Kubernetes version & enabled APIs_
+- Reliable suppression of audit logs for privileged actions
```
thockin: DoS attack from within a cluster (e.g. shared-infra overloads like DNS, endpoints)?

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,70 @@
...
+- In-tree (k8s.io/kubernetes) stable volume plugins
+
+### Out of Scope
+
+The following items are explicitly out-of scope for the bug bounty program. While we still welcome
+vulnerability reports in these areas, they are not (currently) eligible to receive a bounty.
+
+- Alpha features & APIs
+- Kubernetes running on Windows or other non-linux operating systems
+- Non-Kubernetes binaries distributed as cluster addons<br>
```
thockin: CoreDNS is now a default but is not a Kubernetes binary...

contributors/guide/bug-bounty.md
raesene: FWIW I think you'd definitely want to exclude any bounty reports that could be mitigated by secure configuration of the cluster (i.e. if it's possible to use existing k8s configuration options to mitigate an issue, it doesn't count).  Otherwise you'll get a load of reports around things like SSL configuration which are easy to test for but can be mitigated with start-up flags. 

If you do want to refer to a guide in terms of "if it's in here as suggest don't report it" you could go for the CIS guide which probably covers *most* of the areas in question (or if it's missing them, they could easily be added to a future version).

philips: @raesene That is a reasonable suggestion and referring to the CIS guide seems like the most expedient option.

It is important that we don't accidentally remove the ability to have people report issues with defaults in RBAC or service accounts, for example, which might be insecure.

liggitt: > referring to the CIS guide seems like the most expedient option

there are recommendations in the CIS guide I do not consider necessary ("disable anonymous auth on the apiserver") and some which make most production clusters non-functional ("disable the ability of kubelets to run privileged containers"). I'm not sure I'd endorse that as a foundational config guide

raesene: If the goal of this is to allow for findings that could relate to default configuration of a cluster (e.g. RBAC, Service accounts), then to be effective, it's got to provide a target.  If that's not the CIS guide, then it needs to be some other concrete entity (e.g. Kubeadm), otherwise you're very likely to get people installing "kubernetes distro x" which has a poor security default and then it'll get reported to this bug bounty as a vuln.  Also if there's non-ideal security settings that are considered necessary for a functional cluster (e.g use of privileged pods instead of specifying required capabilities etc), again that's going to be worth getting into the initial scope definition to avoid lots of "this cluster runs insecure pods because they use the root user and are privileged" style reports.

liggitt: I'm referring to specific infrastructure-level pods like kube-proxy that make use of privileged containers, like 
https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/manifests/kube-proxy.manifest (also set up by kubeadm this way, see 
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/addons/proxy/manifests.go)

raesene: Sure I wasn't disagreeing that all clusters tend to do that, just suggesting that for bug bounty purposes, I'd generally recommend explicitly ruling out things which might seem like a dangerous thing to do (like running privileged containers), to avoid getting a load of bogus reports from security people who have "discovered" this flaw.  

So for example if we use Kubeadm as the target for "Kubernetes Default configuration" in the bug bounty, I'd explicitly exclude the use of privileged containers in infrastructure level pods as a target for bounties.  The clearer it can be the fewer bad reports you're likely to get.

contributors/guide/bug-bounty.md
mayakacz: should we add this explicitly (like etcd)? Their process: https://github.com/coredns/coredns#security

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,70 @@
...
+- Publishing of unauthorized artifacts
+- Unauthorized modification of github data
+- CI/CD Credential Leaks
+- Execution inside the CI/CD infrastructure
+
+**Components:**
+
+- Attacks against a stable & supported Kubernetes release (most recent 3 releases)
+- Community maintained stable cloud platform plugins<br>
+  _Vulnerabilities in other cloud platform plugins should be reported through the associated provider_
```
mayakacz: Vulnerabilities in 'other' cloud platform plugins only, or in all cloud platform providers?

contributors/guide/bug-bounty.md
mayakacz: Given the lack of up-to-date OSS security guides, I think the k8s one Jordan originally suggested works best. This will be the most contentious piece with a researcher, may be worth adding language like, "subject to the Product Security team's opinion" or something.

contributors/guide/bug-bounty.md
```diff
@@ -0,0 +1,70 @@
...
+- Linux privilege escalations<br>
+  _Please report these through security@kernel.org_
+- Attacks against containers from the host they are running on
+- Attacks relying on insecure configurations, such as clusters not utilizing mutual authentication
+  or encryption between Kubernetes components.
+- Attacks relying on or against deprecated components (e.g. gitrepo volumes)
+- Vulnerabilities in etcd<br>
+  _Please report these through [CoreOS's disclosure process](https://coreos.com/security/disclosure/)_
+- Vulnerabilities specific to a hosted Kubernetes setup<br>
+  _Please report these through the associated provider_
```
mayakacz: I'd like to add links to all of these here as well. Here's where to do it for GKE: https://www.google.com/about/appsecurity/reward-program/

contributors/guide/bug-bounty.md
tallclair: I want to be careful with that one, since there are so many known issues in that area. How about:

> Persistent DoS from within a cluster by an unprivileged container or user.

contributors/guide/bug-bounty.md
tallclair: If the a plugin is community maintained, it should be handled by the community. If it is maintained by the cloud provider, it should be reported through them. Does that make sense?

contributors/guide/bug-bounty.md
tallclair: Added below.

contributors/guide/bug-bounty.md
tallclair: That could be a long list, if we avoid being political. We could add an appendix, and just move all the links for related disclosure processes there. WDYT?

contributors/guide/bug-bounty.md
raesene: @tallclair so I agree that the problem is that there's a lot of different options for cluste config.  To me that leaves two options, either 1) avoid any areas that are config dependent (e.g. RBAC) or 2) define a "gold" configuration (e.g. Kubeadm).  anything else will leave the programme open to debatable tickets.  Ofc as @mayakacz says you can add language around "opinion" in there, but then you'll still get the reports and need to triage them. Depending on the resource available to manage reports, that may, or may not, be fine :)

tallclair: Added "subject to the Product Security team's opinion".

> 1) avoid any areas that are config dependent (e.g. RBAC)

The problem is that the security of the entire cluster is dependent on config. Kubernetes is very configurable, and there are very many ways to do it wrong. This is something we should fix, but that's a huge project and I don't think we should give up on the bug bounty until then.

> 2) define a "gold" configuration (e.g. Kubeadm)

I guess the question we ended up with is whether this would be more or less work than triaging the reports. We settled with the idea that we would try to leave the scope more open ended, and trust in the community to use their judgement (and our VRP partner to do a good job). That said, none of us have a lot of experience in this area. If this ends up being a false assumption, we will need to reevaluate this path.

tallclair:
Squashed commits. Can we merge this as-is, and then iterate from there?

PR #2624, opened 2018-09-04
include projects in wording

mikebrow:
To cover projects like cri-o, which arguably isn't so much a tool as it is a runtime project.

Signed-off-by: Mike Brown <brownwm@us.ibm.com>


spiffxp:
/approve
thanks

Issue #2031, opened 2018-04-11
Umbrella issue to sunset SIG Cluster Ops


castrojo:
We actually don't have a process documented for how to turn a SIG into a WG, I suspect it's just a bunch of renaming, but filing this in github so that we can gleam info and write it down. 

Also not sure if Cluster Ops wants to move into the On Prem WG, or roll up under SIG Cluster Lifecycle. 

https://github.com/kubernetes/community/issues/2029

/cc @zehicle 
/sig cluster-ops

spiffxp:
From a github label perspective we can either migrate sig/cluster-ops to sig/cluster-lifecycle or get rid of it. Let me know what you decide.

jdumars:
AFAIK there should be no problem migrating existing labels to cluster-lifecycle. The other question is best answered by @zehicle

cblecker:
Nope, totally possible to migrate. We just need to understand what should migrate where.

spiffxp:
/remove-lifecycle stale
/committee steering
/priority important-soon
I went looking through the meeting notes and I couldn't find a written decision to truly sunset this SIG.  It just sorta tapered off.  We either need a decision on this instance, or a policy on how to treat these sorts of things going forward.

zehicle:
Aaron,

We could not establish a consistent quorum - the group is effectively
sunset.  Let me know what actions I should take to make that official.

Sorry for the slow response.

Rob

<http://t.sidekickopen10.com/e1t/c/5/f18dQhb0S7lC8dDMPbW2n0x6l2B9nMJW7t5XZs3M2bjgW64zKpj3LyR22W1qMCsg56dVpDf9jJGjx02?t=https%3A%2F%2Fwww.rackn.com%2F&si=5660451956457472&pi=b5476891-75f5-465d-8650-7da909995352>

Rob Hirschfeld

Co-Founder and Chief Executive Officer, RackN

(512) 773-7522

<http://t.sidekickopen10.com/e1t/c/5/f18dQhb0S7lC8dDMPbW2n0x6l2B9nMJW7t5XZs3M2bjgW64zKpj3LyR22W1qMCsg56dVpDf9jJGjx02?t=https%3A%2F%2Ftwitter.com%2Fzehicle%3Flang%3Den&si=5660451956457472&pi=b5476891-75f5-465d-8650-7da909995352>
[...]

Issue #2029, opened 2018-04-11
Documentation: Sunset procedures for SIGs/WGs


parispittman:
We should have documentation on the why and how to dissolve a SIG/WG, turn a WG into a SIG (or vice versa), combining efforts, etc. #2025 inspired this as we have a few more areas to fix/remove/modify.

The why should come from steering? (I could have glanced over the why in one of the new governance docs) The how could come from contribex. 

/sig contributor-experience
/steering

jdumars:
https://github.com/kubernetes/community/pull/2629

spiffxp:
I opened an an initial stab #2629 that adds a section, and notes we can retire sigs due to inactivity.  

Based on comments in https://groups.google.com/a/kubernetes.io/d/msg/steering/IRFRNwFjQYA/wYd5WUw8BQAJ and https://github.com/kubernetes/community/pull/2629#discussion_r215471944 it sounds like the steering committee would like to discuss more specifics before opening another PR with more detailed criteria.

parispittman:
@spiffxp in the meantime, me/jorge/both can try to work up a draft of logistical items to add to that after the SC convo happens. I'll add a gdoc link here if we get something together before you do.

parispittman:
me + @castrojo will work on this after election cycle.

dims:
@parispittman #2629 has merged. Do we need to do more here?

parispittman:
@dims - yes, we need to document the logistics of retirement. ie - process to follow to archive any repos, slack channels, mailing lists, etc.

parispittman:
working document until we draft in GH
https://docs.google.com/document/d/1_venhAEU1ndOD57Jcz03E645xEtj4eAkrRIOaWjSsJ4/edit?usp=sharing

castrojo:
I am confused as to why we have two `sig-governance.md` documents? I've linked to the one in the steering directory from the root one but imo it's confusing.

michelleN:
Update: #2832 looks good is waiting on an official lgtm and unhold from @spiffxp before merge. Aaron added a follow up item #2857 as well.

philips:
#2832 is merged.

spiffxp:
/close
I'll close this and take on the followup item separately as it's out of scope for this issue

PR #2629, opened 2018-09-04
Add sig retirement criteria (inactivity only)

spiffxp:
/committee steering
/kind documentation
/priority important-soon

I've heard this referred to as sunsetting elsewhere, but using the term
"retirement" to line up with us retiring projects to the
kubernetes-retired GitHub org.

SIG Cluster Ops is effectively retired due to lack of consistent quorum
(ref: #2031). Here's a swag at describing why, to use as policy to 
retire inactive SIGs in the future.

This is a very tiny start to the "why" of https://github.com/kubernetes/community/issues/2029 not necessarily the "how"

### DIFF
--- a/committee-steering/governance/sig-governance.md
+++ b/committee-steering/governance/sig-governance.md
@@ -28,13 +28,11 @@ Subproject Owner Role.  (this different from a SIG or Organization Member).
   through a [super-majority] vote between Chairs, Tech Leads and Subproject Owners.
 - Membership disagreements may be escalated to the SIG Chairs.  SIG Chair membership
   disagreements may be escalated to the Steering Committee.
-
 - Members *MAY* decide to step down at anytime and propose a replacement.  Use lazy consensus amongst
   other members with fallback on majority vote to accept proposal.  The candidate *SHOULD* be supported by a
   majority of SIG Members or Subproject Contributors (as applicable).
 - Members *MAY* select additional members through a [super-majority] vote amongst members. This
   *SHOULD* be supported by a majority of SIG Members or Subproject Contributors (as applicable).
-
 
 ### Chair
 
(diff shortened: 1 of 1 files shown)


spiffxp:
/hold
for comment

/cc @kubernetes/steering-committee 
For review

/cc @parispittman @castrojo 
Does this capture the scope of the decision you need from steering?

committee-steering/governance/sig-governance.md
```diff
@@ -149,6 +148,13 @@ Issues impacting multiple subprojects in the SIG should be resolved by either:
 - Option 1: SIG Technical Leads
 - Option 2: Federation of Subproject Owners
 
+### SIG Retirement
+
+- In the event that the SIG is unable to regularly establish consistent quorum
+  or otherwise fulfill its Organizational Management responsibilities
+  - after 3 or more months it *SHOULD* be retired
```
timothysc: What does retired mean?  Can they reform?   
Also, how would you pass on ownership, b/c by definition SIGs own code.

spiffxp: What needs to be spelled out here? I'm trying to avoid details because I think the sig contributor experience community management subproject can own the logistics.  I would say retired is: owns no more code, and has been moved to [community/archive](https://github.com/kubernetes/community/tree/master/archive).  Can it be revived? If there are people willing to step up, sounds great.

The specific case that is motivating this has no code, so I'm wary of setting precedent for that without having a concrete example.

timothysc: Is this to rationalize sigs that have not submitted a charter, or just don't meet?  
If so, then let's do that on steering and possibly add some clauses and more details here.

spiffxp: I can't answer yes to an either/or question?  I feel like I was pretty clear what the purpose of this was in the PR description.  I agree there are refinements to be made for more cases, but I don't feel a strong need to do that right now.

timothysc: Could you just open up another issue that cross-references this one spelling out that we need to clarify details on retirement and policy and we can get this one done.

spiffxp: I think #2029 is that issue, so I retitled this, and added a comment there

bgrant0607:
Late, but lgtm

PR #2630, opened 2018-09-05
fix yaml formatting in kep 26 frontmatter

mrbobbytables:
This resolved the issue with the frontmatter for kep-26 failing to render for both github and hugo (contributor site).  This is due to the `@` symbol being a reserved [yaml indicator](http://yaml.org/refcard.html) and must be quoted when it is used as the leading character in a string. 

/cc @castrojo


cblecker:
/lgtm
/approve

✨

PR #2632, opened 2018-09-05
Leadership changes for SIG CLI

seans3:
* Position "Technical Lead" is now a separate role from "Chair" 
* Maciej Szulik and Phillip Wittrock are SIG Technical Leads
* Maciej Szulik and Sean Sullivan are SIG Chairs
* Tony Adohe is added as a SIG Emeritus Lead


pwittrock:
@soltysh This looks good to me.

sigs.yaml
```diff
@@ -512,20 +512,27 @@ sigs:
     charter_link:
     label: cli
     leadership:
-      chairs:
+      technical leads:
```
cblecker: this should be `tech_leads:`

cblecker:
Looks like this was discussed in the sig-cli on May 9, 2018 when discussing the charter.
/approve

sigs.yaml
```diff
@@ -509,23 +509,30 @@ sigs:
       establishment of conventions for writing CLI commands, POSIX compliance,
       and improving the command line tools from a developer and devops user
       experience and usability perspective.
-    charter_link:
+    charter_link: https://github.com/kubernetes/community/blob/master/sig-cli/charter.md
```
cblecker: this can just be `charter_link: charter.md` :)

Also, can you squash into one commit? Thank you!

seans3:
@cblecker Thanks for guidance on these changes

cblecker:
no problem at all! :)

OWNERS_ALIASES
```diff
@@ -29,9 +29,10 @@ aliases:
     - erikerlandson
     - liyinan926
   sig-cli-leads:
+    - soltysh
```
soltysh: I don't need to be called out twice, do I ;)

cblecker: This is because you're listed as both a chair and a tech lead. The aliases are autogenerated (it's just a dumb go text/template, so it doesn't de-dupe).

PR #2636, opened 2018-09-06
Network latency SLI

wojtek-t:
/assign @bowei @thockin @dcbw @caseydavenport 
@kubernetes/sig-network-pr-reviews @kubernetes/sig-scalability-pr-reviews


wojtek-t:
@bowei @thockin @kubernetes/sig-network-api-reviews - friendly ping

sig-scalability/slos/network_latency.md
```diff
@@ -0,0 +1,47 @@
+## In-cluster network latency SLIs/SLOs details
+
+### Definition
+
+| Status | SLI | SLO |
+| --- | --- | --- |
+| __WIP__ | In-cluster network latency from a single node, measured as latency of per second ping to kubernetes.default.svc.cluster.local/, measured as 99th percentile over last 5 minutes. | In default Kubernetes installataion with RTT between nodes <= Y, 99th percentile of (99th percentile over all nodes) per cluster-day <= X |
```
bowei: Should we use the master for this SLI? It is "special" in some sense, as in some implementations, the master lives on a different networking plane.

bowei: Also, ping from where. And you probably mean TCP RTT, not ICMP.

sig-scalability/slos/network_latency.md
```diff
@@ -0,0 +1,47 @@
...
+
+### User stories
+- As a user of vanilla Kubernetes, I want some guarantee how fast my http
```
bowei: We can have this SLI, but it may not be that useful by itself, for example, TCP throughput is a function of loss + RTT.

Also, in your user story, network latency is probably the smallest contribution to the e2e latency (server time/OS time will dominate).

bowei: Now that you remind me of the chaining that we do with IP tables, the SLI makes sense :-)

sig-scalability/slos/network_latency.md
bowei: (Also, we should probably use a non-TLS-based endpoint to eliminate that latency)

bowei: Does it make sense to say "RTT to a null service" where null service is defined to be some hello-world serving container?

sig-scalability/slos/network_latency.md
```diff
@@ -0,0 +1,47 @@
...
+request to some Kubernetes service reaches its endpoint
+- As a user of vanilla Kubernetes, I want some guarantee how fast in-cluster
+DNS resolution happens.
+
+### Other notes
+- We obviously can't give any guarantee in a general case, because cluster
+administrators may configure cluster as the want (e.g. number of DNS replicas).
```
bowei: the => they

sig-scalability/slos/network_latency.md
bowei: Oh -- are you measuring DNS specifically? Shouldn't that be under DNS service SLI, rather than generate networking SLI?

sig-scalability/slos/network_latency.md
wojtek-t: > Also, we should probably use a non-TLS-based endpoint to eliminate that latency

Yes - that absolutely makes sense.

> Does it make sense to say "RTT to a null service" where null service is defined to be some hello-world serving container?

Yes - that would be the ideal goal. Though, what I wanted to avoid is to create a dedicated service for that (or at least, creating a dedicated pods to serve those service), because we would like to measure that in user clusters too, and we don't want to consume additional user resources.

Maybe the workaround for it would be to make this "null service" be exposed from kube-proxy pods and then we would just need to create a service (which is not that big deal).
The drawback of that is that then there is a backend on every single node, so this won't work good in large clusters.
[Ideally, i would say let's expose it from a subset of kube-proxies, but then we would need to add labels on some kube-proxies, which then turns out to be too complicated, i think...]

Maybe you have something better on your mind?

sig-scalability/slos/network_latency.md
wojtek-t: I was already thinking about this and my rationale was:
- users care about end-to-end flow (and in many cases it contains dns resolution)
- dns sli would be useful for debugging (i.e. deciding whether it is dns that is slow or other parts of stack). But that's more for developers, not something that we need to expose to users in a form of SLO guarantee

WDYT?

sig-scalability/slos/network_latency.md
wojtek-t: done

sig-scalability/slos/network_latency.md
wojtek-t: We discussed offline and decided to split those. I left a TODO to consider joining them longer-term.

sig-scalability/slos/network_latency.md
wojtek-t: As mentioned above we don't seem to have anything satisfying our needs now.
Making a service on top of kube-proxies will be problematic in large clusters, and choosing only a subset complicates cluster setup.

So we decided for requiring that null-service to be setup by cluster administrator.

bowei:
cc: @lzang

sig-scalability/slos/network_latency.md
```diff
@@ -0,0 +1,49 @@
...
+  - it is not application-specific
+
+### Caveats
+- The SLI is formulated for a single node, even though users are mostly
+interested in the aggregation across all nodes (that is done only at the SLO
+level). However, that provides very similar guarantees and makes it fairly
+easy to measure.
+- The prober reporting that is fairly trivial and itself needs only negligible
+amount of resources. However, to avoid any visible overhead in the cluster
+(in terms of additionally needed components) **we will make it part of kube-proxy
```
thockin: This is going to test (as you stated in the SLI column) traffic from nodes.  I don't think that is the most representative or useful metric.  It seems more correct to test from a pod, through a service, to a pod, across nodes in the same primary topology.

I don't know how to generically express "same primary topology".  In GCP that is zone, but topology is arbitrary -- maybe that needs to be parameterized?

We might ALSO want to SLI pod-to-pod without a service (or with a headless service) with the same conditions.

@johnbelamaric

wojtek-t: > This is going to test (as you stated in the SLI column) traffic from nodes. I don't think that is the most representative or useful metric. It seems more correct to test from a pod, through a service, to a pod, across nodes in the same primary topology.

Kube-proxy is also a pod. So why this isn't what you want?

Re sam primary topology - given that we currently don't really support that fully, I'm a bit sceptical about doing that. Once we have topology-aware service routing (https://github.com/kubernetes/community/pull/2846) we can update that SLI.

@thockin - WDYT?

wojtek-t: Discussed that offline with @thockin 
The main points were:
- we can't use kube-proxy, because it's in host network - so I switched that to use dedicated prober
- about topologies - making that explicit that we acknowledge very different latencies if nodes are in different topologies - I added explicit point about that.

wojtek-t:
@thockin @bowei - Tim`s comments are now addressed (after discussing that deeper offline). PTAL

lzang:
Are we going to measure this SLI in users' clusters in prod?  From the discussion above, it seems that we would need user to setup this test as it consumes their resource. Do we know if users are willing to do that? Also would the result vary significantly due to user cluster's load level/deployment scenario?  How do we account for that when we get the data? For example, the result could be different if user is running istio side-car or enable network policy.

PR #2637, opened 2018-09-06
Fix bad link

TinySong:


cblecker:
These aren't bad links. They're relative ones in GitHub Markdown.

/close

TinySong:
@cblecker, if this is not bad link, the flexvolume.md should be in ~contributors/design-proposals/storage ~ directory

PR #2638, opened 2018-09-06
Provisional: quotas for ephemeral storage

RobertKrawitz:


RobertKrawitz:
@sjenning

RobertKrawitz:
ping @dchen1107 @derekwaynecarr

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,647 @@
...
+
+### Non-Goals
+
+* Enforcing limits on total pod storage consumption by any means, such
+  that the pod would be hard restricted to the desired storage limit.
+
+## Proposal
+
+This proposal applies project quotas to emptydir volumes on qualifying
+filesystems (ext4fs and xfs with project quotas enabled).  Project
```
dashpole: What do we plan to do for other filesystems?  If we are changing the behavior for exceeding container limits (eviction vs "no space left on device"), ideally we want to keep the behavior the same across filesystems.  If that isn't possible, we should explicitly point it out.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,647 @@
...
+* Determine whether a quota applies to the directory, in a
+  filesystem-dependent manner (see [below](#notes-on-implementation)).
+* If so, determine how much storage or how many inodes are utilized,
+  in a filesystem dependent manner.
+
+If the quota code is unable to retrieve the consumption, it returns an
+error status and it is up to the caller to utilize a fallback
+mechanism (such as the directory walk performed today).
+
+### Operation Flow -- Removing a Quota.
```
dashpole: When do we remove a quota?

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,647 @@
...
+manager, particularly under stress).  It also requires setup in the
+form of a prepared filesystem.  It would be better served by
+appropriate end to end tests.
+
+### Risks and Mitigations
+
+* The SIG raised the possibility of a container being unable to exit
+  should we enforce quotas, and the quota interferes with writing the
+  log.  This can be mitigated by either not applying a quota to the
+  log directory and using the du mechanism, or by applying a separate
```
dashpole: agree that a non-enforcing quota seems like the best solution here.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,647 @@
...
+inherited by any files created under that directory.
+
+If we elect to use the quota as enforcing, we impose a quota
+consistent with the desired limit.  If we elect to use it as
+non-enforcing, we impose a large quota that in practice cannot be
+exceeded (2^63-1 bytes for XFS, 2^58-1 bytes for ext4fs).
+
+### Operation Flow -- Applying a Quota
+
+* Caller (emptydir volume manager or container runtime) creates an
```
dashpole: Is enforcing quotas on container writable layers (or logs) in-scope, or out-of-scope for this proposal?  They are not listed in goals (and empty-dir quota is explicitly called out).  If it is in-scope, you should detail the changes to the CRI that are required for it. If it is not in-scope, it should be added to non-goals.  If it is out-of-scope, I am not sure if the risk about container shutdown still applies?

dashpole:
cc @jingxu97 
who worked on the initial local ephemeral storage proposal and implementation

derekwaynecarr:
@RobertKrawitz - this is a great write-up, and i have a few questions.

i understand that if the kubelet owns the project id mapping for a given directory, the quota code (assumed vendored and managed in kubelet) can handle assignment of a project id to a given directory it manages (i.e. emptyDir).  for container runtime managed layers, its not clear to me if you are saying that a) the container runtime asks the kubelet for a project id, b) the kubelet informs the container runtime of a project id (maybe as part of `LinuxContainerConfig` in StartContainer), c) something else?

can a non-privileged user write a file with an alternate project id and mess up accounting?  basically, what is the security boundary that we can depend upon?

how many project ids would you recommend we have per pod?  is there a project id per container and emptyDir volume?

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: Exceeding container limits will still result in eviction.  And there's always the risk, even today, of running out of storage; we don't (and can't!) promise containers that they will have an unlimited amount of local storage to play with.

I'd be happy to support quotas on as many filesystems as we can, but if the underlying filesystem doesn't support project quotas, there's little we can do.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: When a volume is torn down (including at kubelet restart).

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: I'd like it to be in scope, at least for writable layers, but who should I discuss the CRI changes with?

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,647 @@
...
+  files being held open.
+* Primary: this will not interfere with the more common user and group
+  quotas.
+* Stretch: enforce limits on per-volume storage consumption by using
+  enforced project quotas.  Each volume would be given an enforced
+  quota of the total ephemeral storage limit of the pod.
+
+### Non-Goals
+
+* Enforcing limits on total pod storage consumption by any means, such
```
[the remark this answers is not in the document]
dashpole: "Enforcing Node Allocatable" would mean enforcing a limit over the sum of all pod's disk usage.  For memory, for example, this means the `/kubepods` cgroup has a memory limit of allocatable.  The purpose of it is to protect system daemons even in the case where containers do not set limits.  In this case, it would mean that the abuse vectors you point out cannot cause the node to run out of disk, but at worst cause temporary problems for other pods.  However, currently our image accounting design has problems (images are considered node "overhead", which makes setting kube-reserved rather strange), so we probably don't want to enforce Node Allocatable until that is solved as well.

cc @yujuhong, who drove the initial CRI design and improvements.

RobertKrawitz:
@derekwaynecarr:

1) If we're going to go with non-enforcing quotas, or enforcing quotas with a separate quota on each filesystem (and as I stated, those two alternatives are my preference), I think the container runtime should ask its instance the quota code to allocate it a project ID.  In other words, (c).

2) A non-privileged user cannot create a file with an alternate project ID.  Nor can the user cheat by hard linking a file with a different project ID into the directory tree, for instance.  A privileged user can change the project ID of a file, but a non-privileged user cannot.  See https://www.systutorials.com/docs/linux/man/8-xfs_quota/#lbAK.  I will add a short background and a link to this.

3) One project ID per ephemeral volume.  So that means one per emptyDir volume, one per writable layer (one per container), and one (non-enforcing) per logdir.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
dashpole: > Exceeding container limits will still result in eviction

Is this because we are excluding the enforcement of container limits from the proposal, and leaving the current implementation as is?

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
dashpole: @yujuhong

RobertKrawitz:
@dashpole "IMO, it would make more sense to either tackle just metrics, or to expand the scope to include container limits, and pod limits as well."

Are container and pod limits referring to ephemeral storage or to other resources?  If the former, I agree, but I haven't succeeded thus far in devising a mechanism that would accomplish that that doesn't either add a lot of complexity or break the solution for metrics.  If the other, what other resources have similar limits that could be covered by this mechanism?  Persistent local volumes would be a simple extension (basically, just add the same code as I added to emptydir volumes).

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
dashpole: Is the behavior changing for empty-dirs (eviction-> "no space left on device")?

RobertKrawitz: Hitting the limit is still hitting the limit.  If you hit the limit and quickly got back under you'd avoid eviction, but again, no different from today.  If you hit the limit and stay there, the eviction code will pick up on it.

Consider setting the quota to one filesystem block more than the ephemeral limit (which I've considered doing).  If you write up to the quota limit and stop, you will definitely be above the ephemeral limit, so the eviction manager will toss you when it sees it.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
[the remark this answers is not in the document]
RobertKrawitz: This is strictly referring to local storage capacity isolation aka ephemeral storage, at least at this point.

I will add "enforcing node allocatable" to the non-goals and make clear that the goals apply only to ephemeral storage.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: Also consider what happens if you have two emptydir volumes with a total limit of 1 MiB.  If you write 513 KiB to each you won't hit the quota on either, but you will still be evicted.

This proposal will certainly make "no space left on device" and similar failures more common, but they can still happen today.

dashpole: > If you hit the limit and stay there, the eviction code will pick up on it.

Ah, I see.  I was under the impression that pods that hit the disk quota limit would just stay there.  Might be worth clarifying that this is not replacing evictions for exceeding container limits (or empty-dir, for that matter).

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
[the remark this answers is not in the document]
dashpole: node allocatable exists for ephemeral storage, we just don't want to enforce it anytime soon.

dashpole:
> Are container and pod limits referring to ephemeral storage or to other resources? If the former, I agree, but I haven't succeeded thus far in devising a mechanism that would accomplish that that doesn't either add a lot of complexity or break the solution for metrics. If the other, what other resources have similar limits that could be covered by this mechanism? Persistent local volumes would be a simple extension (basically, just add the same code as I added to emptydir volumes).

I am referring to ephemeral storage.  If we want to scope this to just monitoring + empty-dir volumes, I would be interested in an example of a use-case this solves (other than monitoring), or a goal it accomplishes.  The abuse vector appears to still exist, as I can just write to my container's writable layer instead of the empty-dir.  Is this more about more than preventing abuse (e.g. user error)? Or is there a way for cluster admins to enforce that empty-dir volumes have size limits, and disable container writable layers, and thus prevent this abuse vector?

RobertKrawitz:
@dashpole I agree that in the absence of a solution for writable layers, enabling enforcement by default does not make sense.  A feature gate, off by default, for enabling enforcement may be useful to give administrators and developers experience with the enforcing environment prior to enabling by default, if and when we go that route.  For clusters that choose to go that route, it would provide us with soak.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: Adding an explicit non-goal that eviction will be eliminated.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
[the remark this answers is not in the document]
RobertKrawitz: Adding non-goal.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: This is already listed as a stretch goal.  I am clarifying that it will only be done if we can apply an enforcing quota to the writable layer.

RobertKrawitz:
@derekwaynecarr adding a descriptive section on project quotas.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,731 @@
...
+  gate.
+  
+  A report <https://blog.pythonanywhere.com/110/> notes that an
+  unclean shutdown on Linux kernel versions between 3.11 and 3.17 can
+  result in a prolonged downtime while quota information is restored.
+  Unfortunately, [the link referenced
+  here](http://oss.sgi.com/pipermail/xfs/2015-March/040879.html) is no
+  longer available.
+
+* Bugs in the quota code could result in a variety of regression
```
yujuhong: How stable is the quota feature in general, and how many CVEs are related to it?

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,731 @@
...
+  behavior.  For example, if a quota is incorrectly applied it could
+  result in ability to write no data at all to the volume.  This could
+  be mitigated by use of non-enforcing quotas.  XFS in particular
+  offers the pqnoenforce mount option that makes all quotas
+  non-enforcing.
+
+  We should offer two feature gates, one to enable quotas at all (on
```
yujuhong: A thought: Can we enable quota for monitoring side-by-side with `du`, so that we can compare how stable it is?

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
yujuhong: I think it's worth adding a section to clarify the what number to use when setting the quota, and how this interact with eviction. It's not very clear (at least to me) from reading the proposal.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,731 @@
...
+
+  The only potentially viable combination in my view would be a dense
+  loop filesystem without sync, but that would render any thin
+  provisioning impossible.
+
+## Infrastructure Needed [optional]
+
+* Decision: who is responsible for quota management of all volume
+  types (and especially ephemeral volumes of all types).  At present,
+  emptydir volumes are managed by the kubelet and logdirs and writable
```
yujuhong: We can start a discussion with runtime vendors to set quotas for the writable layers. Prior to that, I think we can explore more on how this is expected to implement in kubelet and what to pass to the runtime. IIUC, the ephemeral storage limit  is shared among writable layers, logs, emptyDirs, etc. If that's the case, what limit are we going to pass to the runtime to set? Passing the project id may not work because the writable layer may be in a separate filesystem (do we still need to include it?). 

Maybe we can have a separate proposal for enforcement, given that this proposal already reaches 700 lines.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: There's no guarantee that the numbers will match up at any given moment in time.  That's partly because of the issue of deleted files, and partly because a quota reflects a snapshot at a given moment in time while du is not atomic.

The caller could use both mechanisms and compare them; the code calling into the quota for this purpose is isolated.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
RobertKrawitz: We don't pass in the project ID; the quota code picks one.  We only pass in the requested quota (or -1 for monitoring only).

I will clarify this.

RobertKrawitz: BTW, I should note that as the writable layer may be in a different filesystem from the emptydir volume(s), it may not be possible to apply an enforcing quota that applies to the entire ephemeral storage.

For example, if there is an ephemeral limit of 10 MB on the pod, and the writable layer is on a different filesystem from the emptydir volume(s), we cannot enforce a quota of 10 MB on the total.  We can enforce a limit of 10 MB on the emptydir volume(s) and a separate limit of 10 MB on the writable layer in this case.  That's inherent in the fact that quotas are always per-filesystem.

RobertKrawitz: There is already a section "Selecting a Project ID" that desribes how project IDs are selected.

RobertKrawitz: The (non)interaction with eviction is listed under non-goals.

keps/sig-node/0028-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,813 @@
...
+
+* `LocalStorageCapacityIsolation` must be enabled for any use of
+  quotas.
+  
+* `FSQuotaForLSCIMonitoring` must be enabled in addition.  If this is
+  enabled, quotas are used for monitoring, but not enforcement.  At
+  present, this defaults to False, but the intention is that this will
+  default to True by initial release.
+  
+* `FSQuotaForLSCIEnforcement` must be enabled, in addition to
```
dashpole: I would prefer if we delayed adding the `FSQuotaForLSCIEnforcement` feature gate until after we have settled on a proposal for how enforcement will work.  Otherwise, what code would we place behind it?

RobertKrawitz: The low level code is very similar; see the PR.  It's not a problem to remove it if it would be problematic, although having it in there makes it possible to experiment with enforcing quotas without additional code changes.

RobertKrawitz: I've removed this from the code and moved it to future in the KEP.

dashpole:
This proposal lgtm wrt monitoring and eviction changes.

keps/sig-node/0030-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,810 @@
+---
+kep-number: 0
+title: My First KEP
```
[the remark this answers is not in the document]
RobertKrawitz: OK

keps/sig-node/0030-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,810 @@
...
+authors:
+  - "@janedoe"
```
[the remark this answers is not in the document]
RobertKrawitz: OK

keps/sig-node/0030-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,810 @@
...
+owning-sig: sig-xxx
```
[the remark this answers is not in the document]
RobertKrawitz: OK

keps/sig-node/0030-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,810 @@
...
+participating-sigs:
+  - sig-aaa
+  - sig-bbb
+reviewers:
+  - TBD
+  - "@alicedoe"
+approvers:
```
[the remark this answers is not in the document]
RobertKrawitz: OK

keps/sig-node/0030-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,810 @@
...
+exceeded (2^63-1 bytes for XFS, 2^58-1 bytes for ext4fs).
+
+### Control over Use of Quotas
+
+At present, two feature gates control operation of quotas:
+
+* `LocalStorageCapacityIsolation` must be enabled for any use of
+  quotas.
+  
+* `FSQuotaForLSCIMonitoring` must be enabled in addition.  If this is
```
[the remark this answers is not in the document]
RobertKrawitz: OK

keps/sig-node/0030-20180906-quotas-for-ephemeral-storage.md
```diff
@@ -0,0 +1,810 @@
...
+  enabled, quotas are used for monitoring, but not enforcement.  At
+  present, this defaults to False, but the intention is that this will
+  default to True by initial release.
+  
+* _`FSQuotaForLSCIEnforcement` must be enabled, in addition to
+  `FSQuotaForLSCIMonitoring`, to use quotas for enforcement._
+
+### Operation Flow -- Applying a Quota
```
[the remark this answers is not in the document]
RobertKrawitz: Adding non-goal for management of anything but emptydir, noting that writable layers will be a future project involving the container runtime.

For monitoring, I don't think any changes to the CRI per se are needed.  Runtimes can use the quota mechanism to implement monitoring; they should follow the same protocol we're using for picking project IDs to use.  For enforcement, changes would be needed at a minimum to support stating the cap.

derekwaynecarr:
Thank you for updates.  I think this is good enough to proceed to work through code.

/lgtm
/approve

PR #2640, opened 2018-09-06
[KEP] Improving Node Health Tracking by tracking health of extensions

vishh:
This KEP is about extending Node Readiness to include health of node level addons and extensions and providing a standard API for cluster level components to discern node readiness.

cc @dchen1107 @derekwaynecarr @kubernetes/sig-node-proposals @kubernetes/sig-scheduling-proposals @thockin @bgrant0607 @mwielgus


keps/sig-node/0028-node-readiness++.md
```diff
@@ -0,0 +1,237 @@
...
+The Node API will include a new field that will track a list of node extensions that are expected on the node.
+This list can be initially populated by the node controller when a node object is created or by the kubelet as it’s extensions register, or by individual extensions themselves (if they have sufficient privileges). 
+
+Here are some sample illustrations:
+
+```yaml
+type Node struct {
+ …
+ Spec:
+  SystemExtensions []SystemExtension
```
johnbelamaric: Is there a real need to add these to the Node spec? Couldn't they be tracked as separate objects each with a reference (i.e., name) to the node?

That would avoid any change to the actual node object here.

keps/sig-node/0028-node-readiness++.md
```diff
@@ -0,0 +1,237 @@
...
+...
+}
+
+type SystemExtension struct {
+   Reference ObjectReference
+   AffectsReadiness bool // Ignore non-critical node extensions from node readiness. Helps identify the overall set of extensions deterministically.
```
johnbelamaric: This only affects "MetaReadiness"?

keps/sig-node/0028-node-readiness++.md
```diff
@@ -0,0 +1,237 @@
...
+}
+```
+
+The following journey attempts to illustrate this solution a bit more in detail:
+
+1. Kubelet starts up on a node and immediately taints the node with a special “kubernetes.io/MetaReadiness” taint with an effect of “NoSchedule” in addition to flipping it’s “Ready” condition to true (when appropriate). 
+2. Kubelet will not re-admit existing pods on the node that do not tolerate the “Readiness” taint. Kubelet will also not evict yet to be admitted pods to avoid causing disruptions. The node is still not fully functional yet.
```
johnbelamaric: s/Readiness/MetaReadiness/

?

keps/sig-node/0028-node-readiness++.md
```diff
@@ -0,0 +1,237 @@
...
+
+```
+
+One of the goals for this proposal is to evaluate the viability and usefullness of combining Status and Health for extensions.
+
+### Approach B - Use Readiness of Node extension pods
+
+This approach assumes that all node level extensions will get deployed as kubernetes pods and that the health of those extension pods can be exposed via existing Readiness probes.
+
+Similar to the previous approach `A`, a special taint will be used to track the overall readiness of the node and the node controller and kubelet will manage the special taint.
```
johnbelamaric: Isn't this approach essentially a special case of Approach A, where the method of reporting readiness is done by looking at the specific pods? I guess in this case it doesn't have to write any per-component state - it is determined by the existing Pod state.

For those cases where pod state is enough, this seems really preferable to avoid additional, frequent writes to the API server. If we need to handle cases where pod state is insufficient, perhaps there is a way to allow the choice to use pod state or stored API objects on a per-extension basis.

keps/sig-node/0028-node-readiness++.md
```diff
@@ -0,0 +1,237 @@
...
+Node selector will take into account `Readiness` all the pods that match the selector(s) specified in the Node Spec `SystemExtensions` field to determine Node Readiness.
+
+The individual system extensions are required to detect kubelet crash (or restart) quickly and reflect that in their Readiness.
+If the Node Controller observes that all the system extension pods are Ready after the kubelet places the special meta readiness taint, it will remove that taint immediately.
+If any of the pods become unready, then the node controller will place the meta readiness taint on the nodes.
+
+This approach has the advantage of not introducing additional API concepts to track state as illustrated in Approach A.
+Side-car containers can be used to expose Readiness more easily than in approach A where it involves updating API objects directly.
+
+Handling of kubelet restarts can be tricky though since there is a chance that the extension pods may not reflect their connections with the kubelet soon after the kubelet restarts.
```
johnbelamaric: Can you explain this more?

keps/sig-node/0028-node-readiness++.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Approach C - Use conditions
```
johnbelamaric: This looks a little more like the [Pod Readiness++ proposal](https://github.com/kubernetes/community/blob/master/keps/sig-network/0007-pod-ready%2B%2B.md). I think we need to reconcile these.

keps/sig-node/0028-node-readiness++.md
```diff
@@ -0,0 +1,237 @@
...
+
+This design continues the existing pattern of expressing health of various node features and components via Conditions.
+For example the Kubelet uses conditions heavily and the Node Problem Detector also uses conditions. Conditions allow for recording heartbeats already.
+
+This design is similar to the approach `A` where instead of recording state to a separate object, kubelet and other system components will record their state via Conditions.
+The node controller will be extended to support a configurable policy on what Conditions influence the meta readiness of a node.
+
+The kubelet and node controller will use a taint similar to the one illustrated in approach `A` to signal node meta readiness.
+
+The scheduler needs to understand that certain special pods (system pods) can be scheduled to nodes even if the node isn’t logically ready.
```
johnbelamaric: This is true for all solutions, isn't it? Essentially, this feature:

* Defines two separate readiness tiers:
    - Ready - existing status indicator, this is more "kubelet ready" or "node core ready" or "ready for control plane workloads" or "ready for system workloads"
    - MetaReadiness - the new state, which represents "ready for user workloads".
 * Provides a mechanism to to stall user workloads (the taint) while system workloads are still coming up. system workloads need to tolerate those taints.

It might be helpful to consider each of these as separate problems that need solving:
1. How do we differentiate "system workload" and "user workload" readiness in a way that works for scheduling? All solutions currently discussed use existing readiness as "system workload readiness" and a taint as "not ready for user workloads", from what I can see.
2. How do we flip each of those readiness flags? This could vary a little based on the solution to (1) - for example the current proposal does not require any changes to handle the "system workload readiness" flag, IIUC. For the "user workload readiness", this can be a stored component-level status + a policy (hardcoded policy in this case using AffectsReadiness) - e.g., "A". Alternatively it could use pod status ("B"), or conditions ("C").

Re-organizing the doc in terms of these may make it easier digest.

keps/sig-node/0028-node-readiness++.md
vishh: > Isn't this approach essentially a special case of Approach A

Yes. Are you proposing we do both `A` and `B`? 
As for handling the additional state, we can consider providing an extensible side car container that deals with health checks for extensions that are not easily modifiable. 
When it comes to tracking internal state of these extensions, a separate object will be necessary at some point though.

keps/sig-node/0028-node-readiness++.md
vishh: Yes. cc @thockin

keps/sig-node/0028-node-readiness++.md
vishh: > This is true for all solutions, isn't it?

All the other approaches uses a taint to express "MetaReadiness" which means no changes are needed to the scheduler. Conditions on the other hand need special logic to be embedded into the scheduler and other controllers like autoscaler.

keps/sig-node/0028-node-readiness++.md
johnbelamaric: Well, I am proposing that we make it easy for extension authors to indicate readiness using the existing control plane features whenever possible. Ideally that means that we don't need to write extension state, although if we do A then you can have a naive implementation that just writes it based on pod status.

justaugustus:

justaugustus:

/close

andrewsykim:
@vishh can I pick up this work from where you left off? There's quite a few issues that could benefit from having node readiness gates and I have cycles to work on it this release. 

ref: https://github.com/kubernetes/kubernetes/issues/57130 https://github.com/kubernetes/kubernetes/pull/33573 https://github.com/kubernetes/kubernetes/pull/76946

chenk008:
> @vishh can I pick up this work from where you left off? There's quite a few issues that could benefit from having node readiness gates and I have cycles to work on it this release.
> 
> ref: [kubernetes/kubernetes#57130](https://github.com/kubernetes/kubernetes/issues/57130) [kubernetes/kubernetes#33573](https://github.com/kubernetes/kubernetes/pull/33573) [kubernetes/kubernetes#76946](https://github.com/kubernetes/kubernetes/pull/76946)

I can work on this too.If I can do anything to help,please ping me.

vishh:
@andrewsykim that's great. I can help shepherd it through. The main changes needed to the KEP as it stands is to switch to using "Approach B" where node extension pods will be identified using label selectors and the node controller is expected to use a configurable taint to ensure the node is ordinarily schedulable only when all the extension pods matched by the label selectors are "Ready". At-least one pod should match a label selector and for simplicity we can start with requiring only one pod to match. 
Does that make sense?

chenk008:
@vishh [kubernetes/kubernetes#76946](https://github.com/kubernetes/kubernetes/pull/76946)
we need support for cloud-controller-manager in the KEP.The node can be scheduable after the cloud provider success to create route

andrewsykim:
@vishh yes that makes sense. I'll update the current KEP to reflect that and open a PR in k/enhancements soon. I'll assign you for review @chenk008

yastij:
@andrewsykim - can you add me as a reviewer on this as kubernetes/kubernetes#75890 needs exactly this.

PR #2648, opened 2018-09-07
Match other staging repos, both OWNERS files listed

spiffxp:
As I was reviewing other staging repos, I noticed two of them
weren't listed here in sigs.yaml. I don't really know whether
listing both the staging OWNERS and published OWNERS is any
better or worse than just listing one, so I leaned in the
direction of consistency


brancz:
cc @DirectXMan12

PR #2656, opened 2018-09-10
update sig-auth mission statement for vernacular readers

deads2k:
sig-auth's existing mission statement indicates how we achieve our objectives and this quote really sums up what our objective is to end users.

@kubernetes/sig-auth-api-reviews

### DIFF
--- a/sig-auth/README.md
+++ b/sig-auth/README.md
@@ -10,6 +10,8 @@ To understand how this file is generated, see https://git.k8s.io/community/gener
 
 Covers improvements to Kubernetes authorization, authentication, and cluster security policy.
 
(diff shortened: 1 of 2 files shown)


liggitt:
Heh, I'd take this. 

/cc @tallclair @mikedanese

enj:
Haha, LGTM!

deads2k:
> Unrelated: @deads2k, was there some other path you modified at some point? I'm curious as to why this is labelled as an api-change

@cblecker I suspect it's because I tagged @kubernetes/sig-auth-api-reviews rather than the leads individually.

cblecker:
Yeah, that'll be it. Thanks for pointing that out :D

mikedanese:
If a tree falls in the woods, does it make a sound? If a feature provides security without usability, are users any more secure?

tallclair:
/lgtm
:clap:

PR #2674, opened 2018-09-14
governance: remove note that SC is still being formed

nikhita:
A minor cleanup.

/committee steering
/kind cleanup

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -31,9 +31,6 @@ The project has 4 main types of groups:
 3. Working Groups, WGs
 4. Committees
 
-Note that the project is also in the process of forming a Steering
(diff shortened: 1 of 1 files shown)


governance.md
```diff
@@ -163,7 +160,7 @@ community meeting.
 
 All repositories under Kubernetes github orgs, such as kubernetes and kubernetes-incubator,
 should follow the procedures outlined in the [incubator document](incubator.md). All code projects
```
nikhita: Also, this line links to https://github.com/kubernetes/community/blob/master/incubator.md but that document is deprecated.

Changing this needs to go through proper channels though and is out of scope for this PR.

PR #2676, opened 2018-09-16
adding more eligible voters

parispittman:
adding eligible voters from for the steering committee election round 1 of the exception form voting by election officials. second round to follow in the next 24 hours.

### DIFF
--- a/events/elections/2018/voters.md
+++ b/events/elections/2018/voters.md
@@ -1,7 +1,7 @@
 ## 2018 Steering Committee Election
 ### Eligible Voters List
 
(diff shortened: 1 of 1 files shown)


events/elections/2018/voters.md
```diff
@@ -538,6 +539,7 @@ The eligible voter process can be found in the steering committee [elections.md]
 1. rmmh
 1. roberthbailey
 1. robinpercy
+1. rodev
```
castrojo: This should be `rdodev`

PR #2679, opened 2018-09-17
Replace kubernetes-users mentions with new location

castrojo:
cc @mrbobbytables @jeefy


jeefy:
Should updates also be in `communication/README.md` and `sig-governance.md` ?

PR #2686, opened 2018-09-18
Clarify rules around defaults across versions

thockin:
Clarify the text a bit and specify that new fields must have equivalent defaults in all API versions.


thockin:
This whole doc needs a revisit

lavalamp:
/lgtm
/approve

Wording can probably be tweaked further but I don't see any reason not to merge now.

PR #2695, opened 2018-09-20
kep: image promoter: process to copy container images

javier-b-perez:
KEP to create a process for publishing container images into the official kubernetes container registry.


dims:
@javier-b-perez wanted to raise this question. As part of the promotion, do we want to sign the images? please see

- https://github.com/kubernetes/kubernetes/issues/63704
- https://docs.docker.com/engine/security/trust/content_trust/

BenTheElder:
Thanks @javier-b-perez! Initial pass LGTM :smile:

javier-b-perez:
> @javier-b-perez wanted to raise this question. As part of the promotion, do we want to sign the images? please see
> 
> * [kubernetes/kubernetes#63704](https://github.com/kubernetes/kubernetes/issues/63704)
> * https://docs.docker.com/engine/security/trust/content_trust/

That is a good point, I will investigate if this is possible and if we can, I think we should sign during the promotion process.

thockin:
Idea LGTM.  Who else do we want to look at this?  @kubernetes/sig-release-members maybe ?

dims:
Needs a small update after looking at signing possibilities. LGTM otherwise.

javier-b-perez:
> Needs a small update after looking at signing possibilities. LGTM otherwise.

Looking at GCR seems it is not supported at the moment, but we can still push to multiple registries, if we want, (like GCR, docker hub) and sign images when possible as part of the promotion process. 
I will update the text with a comment saying that as part of the promotion process, we could sign images.

thockin:
Before approving this, I really want some of the release folks to review.

thockin:
Larger than this proposal?  Inclusive of this or different than this?

dims:
@tpepper what's detailed here in this KEP will be an important part of whatever we come up with for a "larger KEP around an improved release engineering". So i'd like to get this approved.

/cc @AishSundar @calebamiles @kubernetes/k8s-infra-team

keps/sig-release/0028-k8s-image-promoter.md
```diff
@@ -0,0 +1,102 @@
...
+  * [Staging Container Registry](#staging-container-registry)
+  * [Production Container Registry](#production-container-registry)
+  * [Promotion Process](#promotion-process)
+* [Graduation Criteria](#graduation-criteria)
+* [Infrastructure Needed](#infrastructure-needed)
+
+
+## Summary
+
+For security reasons, we cannot allow everyone to publish container images into the official kubernetes container registry. This is why we need a process that allows us to review who built an image and who approve it to be shown in the official channels.
```
AishSundar: nit: s/approve/approved

keps/sig-release/0028-k8s-image-promoter.md
```diff
@@ -0,0 +1,102 @@
...
+### Production Container Registry
+
+A restricted set of members can have push access to override any tool or process if necessary.
+Ideally we only push images that have been approved by the owners of the production container registry, following the promotion process.
+
+### Promotion Process
+
+1. Maintainer create a container image and push it into *staging* GCR.
+1. Maintainer creates a PR in GitHub to add the new image into the *official* container registry.
+1. Once owners of the official container registry approve the change and merge into the master branch, the image should be copied from *staging* into *official* container registry.
```
AishSundar: In this sentence, it not very clear who does the actual push - the person approving the PR or the bot merging the PR or promoter tool. Can you plz clarify

keps/sig-release/0028-k8s-image-promoter.md
```diff
@@ -0,0 +1,102 @@
...
+In the future, we could add more information into the context of the PR like the vulnerability scan and test results of the container image.
+
+## Graduation Criteria
+
+We will know we are done when we have:
+
+* User guide for developers/maintainers: how to build? how to promote?
+* User guide for owners: review and approve PR, how to push images?
+* A repository to host the manifest file.
+* Initial set of repository's owners who can approve changes.
```
spiffxp: Policy: criteria for granting/removing staging access? production access?

thockin:
I don't think we can do permissions less than a whole bucket.  Does it make
sense to consider a federation of buckets that people can use as their
staging and have the promoter be able to promote from any of them?

keps/sig-release/0028-k8s-image-promoter.md
javier-b-perez: done

keps/sig-release/0028-k8s-image-promoter.md
javier-b-perez: done

keps/sig-release/0028-k8s-image-promoter.md
javier-b-perez: done

keps/sig-release/0028-k8s-image-promoter.md
```diff
@@ -0,0 +1,102 @@
...
+
+For security reasons, we cannot allow everyone to publish container images into the official kubernetes container registry. This is why we need a process that allows us to review who built an image and who approve it to be shown in the official channels.
+
+
+## Motivation
+
+There are multiple reasons why we should have a process to publish container images in place:
+
+* We cannot allow all community members to publish images into the official kubernetes container registry.
+* We should restrict who can push images to a small set of members and some systems accounts for automation.
```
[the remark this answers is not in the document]
javier-b-perez: We cannot control individual images or directories in GCR. This is why I say it should be a "small set of members".
We can restrict who can merge or approve PR into the manifest's repository, then leave the rest to the promoter tool). So those who approve PRs don't even need to have write access into the container registry.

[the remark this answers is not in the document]
thockin: @javier-b-perez Would it make sense to have each sub-project get a new staging area of their own, and have the promoter pull from any of those?

I guess we'd need a new GCP project for each bucket (I can't see how to make >1 bucket in a project).  The promoter works on hash, so it should Just Work, right?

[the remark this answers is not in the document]
javier-b-perez: @thockin it is possible to have a per project staging container registry, that means we need to create a GCP project per each GCR we want to handle. 
The promoter tool just need read access to this GCR and it can copy the container image with the same digest from project staging to official container registry.

How do you see the staging domain working? something like:
staging-k8s.gcr.io/project1/imageA -> GCR/project1/imageA
staging-k8s.gcr.io/project2/imageB -> GCR/project2/imageB
(not sure if this is possible, might not)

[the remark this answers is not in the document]
thockin: I'm just thinking out loud.  The options are really 

a) Be careful who can push because anyone who can write can do anything.

b) Give each sub-team their own space.

e.g. gcr.io/k8s-foo-staging for project foo and gcr.io/k8s-bar-staging for project bar, both promote to `gcr.k8s.io/{foo,bar}/{foo,bar}:<tag>`

@dims @mkumatag thoughts?

listx:
I'd like to state for the record (please let it be known!!) that I'm highly interested in creating a demo of a promotion tool to help realize this KEP. It would be responsible for performing various promotions and whatnot.

I should have a proof-of-concept ready before the end of the year.

listx:
@calebamiles What will it take to get this KEP approved?

dims:
@listx i think we need approval from @thockin as mentioned in the proposal itself :)

listx:
@dims Oops, I completely missed the metadata part.

As a reviewer myself, LGTM.

keps/sig-release/0031-k8s-image-promoter.md
```diff
@@ -0,0 +1,103 @@
...
+  - TBD
+reviewers:
+  - TBD
+  - "@listx"
+approvers:
+  - "@thockin"
+editor: TBD
+creation-date: 2018-09-05
+last-updated: 2018-11-05
+status: draft
```
thockin: implementable

PR #2708, opened 2018-09-25
Add sig-cluster-lifecycle/kubespray meeting details

chadswen:
/cc @roberthbailey


roberthbailey:
/lgtm

FYI I've also added an event to the sig cluster lifecycle calendar and it should show up on the universal calendar as well.

chadswen:
@jdumars looks like this one is ready, but let me know if there's anything else we need to do before removing the hold.

chadswen:
@jdumars had to rebase, so this will need /lgtm again once the hold is removed.

chadswen:
Had to rebase again.

PR #2711, opened 2018-09-25
Add max volume attach limit integration with topology

msau42:

### DIFF
--- a/contributors/design-proposals/storage/volume-topology-scheduling.md
+++ b/contributors/design-proposals/storage/volume-topology-scheduling.md
@@ -950,18 +950,36 @@ Currently, there are a few existing scheduler predicates that require the PVC
 to be bound.  The bound assumption needs to be changed in order to work with
 this new workflow.
 
-TODO: how to handle race condition of PVCs becoming bound in the middle of
-running predicates?  One possible way is to mark at the beginning of scheduling
-a Pod if all PVCs were bound.  Then we can check if a second scheduler pass is
-needed.
-
 ##### Max PD Volume Count Predicate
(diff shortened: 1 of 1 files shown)


contributors/design-proposals/storage/volume-topology-scheduling.md
```diff
@@ -950,18 +950,36 @@ Currently, there are a few existing scheduler predicates that require the PVC
...
+CheckVolumeBinding predicate already does:
+* Find an available PV
+* Check if provisioning is possible
+
+If any of these steps are successful, one last check is made to see if the
+current node's volume attach limit plus this unbound PVC (count of 1) is valid.
+The attach limit resource name is determined from the StorageClass's provisioner name,
+which must equal the plugin name. This unbound PVC counting also needs to handle the
+scenario of multiple PVCs in a Pod with the same or different volume plugins.
+
```
ddebroy: Are there any synchronization issues that we need to be aware of in the context of the above for a give node? Is it possible for two threads of the check against the attach limits to be running and both deciding it's okay based on current conditions and leading to one of them failing if there were attach_limit - 1 slots remaining?

msau42: Predicate and assume runs serially per pod, so as long as we can handle assumed pods correctly in the calculation, then this should be safe.

We will have to think carefully about this when the predicate equivalence cache is enabled though (alpha feature).  It caches the result of the predicate, so we need to make sure we invalidate the cached predicate in all appropriate scenarios.

justaugustus:
Could we instead migrate this into a KEP and land it in [k/enhancements](https://github.com/kubernetes/enhancements)?

Issue #2619, opened 2018-08-31
Subproject details in sigs.yaml


timothysc:
Subprojects need to have more details available other then OWNERS link 

e.g. slack and lead details. 

Currently I can't seem to add that to a sub-project.

cblecker:
/sig contributor-experience

Can you mock up what the yaml might look like? Should be easy to implement.

cc: @spiffxp

justaugustus:
@timothysc -- I have a PR up that addresses this: https://github.com/kubernetes/community/pull/2731

spiffxp:
ref: https://github.com/kubernetes/community/issues/1673

nikhita:
This is currently blocked on Steering, since we need docs for what a subproject "lead" means and the processes around it before we can add fields for it in sigs.yaml.

Ref: https://github.com/kubernetes/steering/issues/36#issuecomment-471732144

dims:
proposal
---------
- A sub project can have one or more leads.
- The lead keeps track of things to do
- The lead coordinates work among interested parties
- The lead runs the meetings, ensures there are notes/recordings
- The lead reports back updates to the parent SIG.

Does that help?

nikhita:
@dims :+1: :+1: :+1: Would be cool to see this PR'd in. :)

For reference, we also have docs on what a "Subproject Owner" means today: https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md#subproject-owner

dims:
adding here https://github.com/kubernetes/community/pull/3413

bgrant0607:
Let's please not use the term "lead". I commented on #3413.

nikhita:
@spiffxp @timothysc can this be closed?

spiffxp:
/close
https://github.com/kubernetes/community/pull/3483 addressed this by adding a `Contacts` field to the Subproject struct, which allows for subprojects to have a slack channel, mailing list, and github teams (with examples of all three)

Issue #1913, opened 2018-03-08
Subprojects in sigs.yaml will be instantly out of date


lavalamp:
Anytime anyone moves a directory, the entry in sigs.yaml will suddenly be wrong.

Wouldn't it make more sense for OWNERS files to instead declare the owning sig and subproject, and then we generate sigs.yaml from that?

lavalamp:
Not sure who to tag-- maybe @spiffxp ?

cblecker:
/sig contributor-experience

yeah, there are some updates that are going to have to be made to the generator to accommodate the new structures in https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance-template-short.md

spiffxp:
/area github-management

Yeah I agree until there's a single source of truth this is going to be perpetually stale.  It started in sigs.yaml because it's easier to try and change everything in one place while getting started.

nikhita:
/kind bug
/priority important-soon

xref https://github.com/kubernetes/steering/issues/36

spiffxp:
/assign
ref: https://github.com/kubernetes/community/issues/1808 for owners management

McCoyAle:
/remove-lifecycle stale
/lifecycle frozen

Due to important long term label.

mrbobbytables:
/committee steering
This is similar to https://github.com/kubernetes/community/issues/4125 - should this be rolled into the annual sig/wg checks?

dims:
see new tool for ensuring urls are still valid ( https://github.com/kubernetes/community/issues/4125#issuecomment-1000412060 )

dims:
related https://github.com/kubernetes/community/issues/6307

parispittman:
with the annual report checks and the new tool above that dims linked, @lavalamp do you think we are in a better place now and ok to close this?

Issue #1673, opened 2018-01-24
Formalize the concept of subprojects


spiffxp:
This is based on / inspired by a [SIG Architecture proposal to expand SIG metadata and flesh out OWNERS](https://docs.google.com/document/d/1FHauGII5LNVM-dZcNfzYZ-6WRs9RoPctQ4bw5dczrkk/edit#)

This is driven by a steering committee goal of ensuring that our organizational structure maps to our code structure.

> Every part of the Kubernetes code and documentation should be owned by a Special Interest Group (SIG). However, many SIGs have multiple significant subprojects with distinct (though sometimes overlapping) sets of contributors and owners, who act as subproject’s technical leaders: responsible for vision and direction and overall design, choose/approve change proposal (KEP) approvers, field technical escalations, etc. We should formally recognize that.

> For example, SIG Network potentially could be divided into multiple subprojects: pod networking (CNI, etc.), Service, Ingress, DNS, and Network policy. SIG Apps has the workload APIs, Helm, and Kompose. SIG Cluster Lifecycle has kubeadm, kops, kubespray, minikube, etc. SIG API machinery has apiserver, client libraries, API discovery and OpenAPI, CRD, API aggregation, and admission extension. In addition to Kubelet proper, SIG Node has multiple CRI implementations underway. Maybe some of these subprojects should be grouped, but hopefully you get the idea.

I'd like to start by implementing this concept in [`sigs.yaml`](/sigs.yaml). This ensures the information is machine-readable, generated README's contain human-readable information, and we start with a single source of truth.

Strawman:
- a sig may have N subprojects
- a subproject must have a name
- a subproject must have 1 - N OWNERS files associated with it
- the OWNERS files define the root of the repos/subdirectories owned by the subproject
- a subproject may have a description
- a subproject may have one or more meetings
- meetings can have different uris for video, agenda, recordings, etc.

Looking at SIG Apps as an example:
- A "Workloads API" subproject could potentially span multiple subdirs and repos
- A "Charts" subproject spans the kubernetes/charts repo
- A "Helm" subproject could potentially span kubernetes/helm and kubernetes-helm/* repos

Looking at SIG Contributor Experience as an example:
- A "Mentorship" subproject would own the docs in kubernetes/community that related to mentorship
- A "Community Management" subproject would own the docs in multiple subdirs of kubernetes/community

[...]

PR #2731, opened 2018-09-30
Extend subproject and repo generation for generator app

justaugustus:
Adds a `Repo` struct and additional fields to the `Subproject` and `Group` structs of the generator app.

This adds some granularity to allow you to define:
- Repos / subdirectories not owned by a SIG Subproject
- Subproject ~Chairs~Contacts
- Repo names / descriptions / URLs / OWNERS

...or decide to continue using the OWNERS files for subprojects. :)

I've created a gist that you can pull to test or view an example of the structure:
- sig.yaml: https://gist.github.com/justaugustus/6e862bdd58832315e79c12ef203e6ec8#file-sigs-yaml
- example SIG Release README.md: https://gist.github.com/justaugustus/6e862bdd58832315e79c12ef203e6ec8#file-sig-release-mock-readme-md

It also cleans up some syntax nits.

Fixes: https://github.com/kubernetes/community/issues/2619
Related: https://github.com/kubernetes/community/issues/1913

/kind feature
/sig contributor-experience

Signed-off-by: Stephen Augustus <foo@agst.us>

### DIFF
--- a/generator/app.go
+++ b/generator/app.go
@@ -84,12 +84,23 @@ type GithubTeams struct {
 	Description string
 }
 
+// Repo represents a specific repo or subdirectories owned by the group
+type Repo struct {
+	Name        string
+	Description string
+	URL         string
+	Owners      []string
+}
+
 // Subproject represents a specific subproject owned by the group
 type Subproject struct {
 	Name        string
 	Description string
+	Label       string
+	Repos       []Repo
 	Owners      []string
 	Meetings    []Meeting
+	Contact     *Contact
 }
 
 // LeadershipGroup represents the different groups of leaders within a group
@@ -109,6 +120,7 @@ type Group struct {
 	Leadership       LeadershipGroup `yaml:"leadership"`
 	Meetings         []Meeting
 	Contact          Contact
+	Repos            []Repo
 	Subprojects      []Subproject
 	StakeholderSIGs  []string `yaml:"stakeholder_sigs,omitempty"`
 }
--- a/generator/sig_readme.tmpl
+++ b/generator/sig_readme.tmpl
@@ -1,19 +1,23 @@
 {{- template "header" }}
 # {{.Name}} Special Interest Group
+{{ if .MissionStatement }}
 
 {{ .MissionStatement }}
+{{end}}
 {{- if .CharterLink }}
+
 The [charter]({{.CharterLink}}) defines the scope and governance of the {{.Name}} Special Interest Group.
 {{ end }}
 {{ if .Meetings -}}
 ## Meetings
+
 {{- range .Meetings }}
-* {{.Description}}: [{{.Day}}s at {{.Time}} {{.TZ}}]({{.URL}}) ({{.Frequency}}). [Convert to your timezone](http://www.thetimezoneconverter.com/?t={{.Time}}&tz={{.TZ | tzUrlEncode}}).
+- {{.Description}}: [{{.Day}}s at {{.Time}} {{.TZ}}]({{.URL}}) ({{.Frequency}}) ([Convert to your timezone](http://www.thetimezoneconverter.com/?t={{.Time}}&tz={{.TZ | tzUrlEncode}}))
 {{- if .ArchiveURL }}
-  * [Meeting notes and Agenda]({{.ArchiveURL}}).
+  - [Meeting notes and Agenda]({{.ArchiveURL}})
 {{- end }}
 {{- if .RecordingsURL }}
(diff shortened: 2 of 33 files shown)


cblecker:
/uncc abgworrall andrewsykim
/assign 

Ooo! Will review!

justaugustus:
Did a few more things:
- Updated the open issues / PRs link to search across all kubernetes orgs instead of just k/community
- Added a label field for subprojects to enable search across orgs via an arbitrary label
- Added struct for per subproject contacts

spiffxp:
/hold
What problem is this trying to solve?

https://github.com/kubernetes/community/blob/fccff8fb9bf7ab9495715ba84e033e59ccf62e2c/governance.md#subprojects

Subprojects don't have chairs

generator/app.go
```diff
@@ -84,12 +84,24 @@ type GithubTeams struct {
...
+	Owners      []string
+}
+
 // Subproject represents a specific subproject owned by the group
 type Subproject struct {
 	Name        string
 	Description string
+	Label       string
+	Chairs      []Person
+	Repos       []Repo
```
timothysc: I don't think this is bad, but I can't think of an example of a subproject that has multiple repos.

justaugustus: API Machinery has a bunch: https://github.com/kubernetes/community/tree/master/sig-api-machinery#subprojects

andrewsykim:
👋 just plugging my two cents here. In https://github.com/kubernetes/community/pull/2740 I had proposed to create a WG as part of SIG Cloud provider. I had choosen WG over a subproject (even though subproject makes more sense) only because WGs had a mechanism to specify leads, meeting times, agenda, etc. I'd like to see this PR go in since a SIG such as SIG Cloud Provider has inititiaves across many subprojects (one per cloud provider) that each require their own leads, meeting time, agendas, etc.

justaugustus:
@spiffxp @cblecker -- any holdback on merging this?

spiffxp:
Yeah I'm not ready to merge or review this until we've had a chance to discuss at sig arch.

I agree the OWNERS files are onerous, but how is this solution going to fall any less out of sync than what we already have?  Having an OWNERS file also mandates there must be code or docs associated with the project, there is no such thing as a subproject without code or docs.

spiffxp:
@andrewsykim FWIW subprojects can have their own meetings and agendas today, eg: https://github.com/kubernetes/community/blob/master/sigs.yaml#L1740-L1755

The way it's rendered in SIG README's could definitely be improved https://github.com/kubernetes/community/tree/master/sig-testing

andrewsykim:
So SIG Cloud Provider is maybe an interesting case where we need to support something that resembles "sub SIGs" as we start to fold the current providers into one. The difference being that the "sub SIG" structure let's us define "tech leads" (a.k.a chairs) which are different from OWNERS in that it gives us more flexibility to assign many repos to one set of leads (the previous chairs of provider SIGs) as opposed to the combination of all developers in all OWNER files.

spiffxp:
/committee steering
turns out this isn't a sig arch concern, it's a steering concern, based on me raising this last steering meeting

justaugustus:
@timothysc -- Fixed up.
Setting a lazy consensus timebox for Friday, November 9, 5pm PT.

generator/app.go
```diff
@@ -84,12 +84,23 @@ type GithubTeams struct {
...
+	Owners      []string
+}
+
 // Subproject represents a specific subproject owned by the group
 type Subproject struct {
 	Name        string
 	Description string
+	Label       string
+	Repos       []Repo
 	Owners      []string
```
spiffxp: So here's an example of how you could add the human info, while preventing it from falling out of data with the OWNERS files:
- change this to `OwnersFiles []string`
- add a field `Owners []Person`
- update app/generator to verify that the people listed in `Owners` are actually present in the `OwnersFiles` as `approvers`

You have to do that third step though.  If you don't, then you're left with two sources of truth as to who has the final say, and that's more confusing than the situation today.

nikhita:
For the record: this is currently blocked on Steering, since we need docs for what a subproject "lead" means and the processes around it before we can add fields for it in sigs.yaml.

Ref: https://github.com/kubernetes/steering/issues/36#issuecomment-471732144

spiffxp:
/close
This is in dire need of rebasing.  I've been leaving this open as a reminder that it's out here, but I don't see us getting to this anytime soon.  Please re-open if you disagree

We will come back to this via https://github.com/kubernetes/community/issues/1673#issuecomment-484746226

PR #2739, opened 2018-10-02
KEP: Per QoS-Class CPU Affinity Hints

ipuustin:
This KEP addresses partitioning the CPUs present in the node to help mitigate the side effects caused by SIMD instructions (such as AVX instruction sets). The current version is meant to be a basis for discussion rather than a finalized proposal.

@ConnorDoyle 
@kad 
@klihub


keps/sig-node/0030-qos-class-cpu-affinity.md
```diff
@@ -0,0 +1,152 @@
...
+
+*Figure 1: Example BestEffort cpuset. If AVX instructions are only run as `BE`, then this caps the number of physical cores that can issue AVX instructions to one.*
+
+**Optional extension (out of tree):**
+
+An operator might prefer local users to be explicit about mapping pods that use a particular ISA extension to a target QoS class. A validating admission controller could validate for example that pods labeled "AVX" run in the BestEffort QoS class. A more opinionated admission controller could mutate the pod spec, removing resource requests from "AVX" pods, thereby forcing them into the BestEffort class. This or similar mechanisms could exist outside of core Kubernetes, following recommendations from hardware vendors.
+
+#### Variant 2: Workloads targeting AVX, run as Burstable or Guaranteed
+
+Enable the existing CPU manager static policy. Use existing ability to advertise extended resources to allow a limited number of "AVX cores". Enforce that if a container consumes the "AVX cores" resource:
```
vishh: how will a user choose a `limited` number?

keps/sig-node/0030-qos-class-cpu-affinity.md
```diff
@@ -0,0 +1,152 @@
...
+- AVX Cores = CPU
+- QoS class is Guaranteed
```
vishh: What guarantees that a Gu pods consuming Avx will not disrupt another Avx pod not consuming Avx?

keps/sig-node/0030-qos-class-cpu-affinity.md
ipuustin: This is the mechanism for advertising extended resources for a node: https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/ . The user can check the node CPU capacity, and then based on that number and some policy choose a suitable number of AVX cores. An example policy might say that "half of the cores should be AVX cores" or something similar.

keps/sig-node/0030-qos-class-cpu-affinity.md
ipuustin: They still might, but at least the AVX and non-AVX pods wouldn't run on the same core. The current algorithm for choosing cores (`takeByTopology()` function) tries to first allocate full cores and sockets based on the request size.

ipuustin:
Thanks for the comments @vishh !

Regarding the points you made:

1. This depends a lot how the cluster is used and how much spare nodes it has. We have heard from customers that they do not want to have nodes dedicated only to AVX workloads. Also, splitting the cluster based on the node features or kubelet configuration leads to combinatoric problems: if (let's say) the cluster is split based on whether nodes have a fast or a slow disk, a GPU or no GPU, and AVX or no AVX support, then that's already the cluster split into `2^3` parts. Still, this is definitely a viable alternative and we added it to the Alternatives section.
2. There is a kernel boot parameter which removes AVX flags from `/proc/cpuinfo` so that the applications can choose not to run AVX instructions based on the flag. I think that doesn't disable AVX support from the processor though. Workloads can be of course written or recompiled not use AVX instructions.
3. This is a good idea, but we don't at the moment have a good way to dynamically determine if a process uses AVX instructions or not. The key idea in this proposal is to find a way to indicate that a pod requires special handling, and that is via the QoS classes, so that pod spec can stay untouched. The actual mechanism to separate the containers to different CPUs is another question -- here we use kubelet options to assign QoS classes to CPU sets. However, later I can see a Dynamic CPU manager policy doing a good effort for this. We planned this proposal to leave enough design space for such a policy.
4. Maybe yes -- we already can know if a processor has the AVX pipeline active or not, meaning that there is some process which run AVX instructions there. But it might be difficult to do the heuristics right (for example to prevent the pods from being ping-ponged around), and the complexity would likely be much larger than with this proposal. This would be a very interesting topic for further investigation though.

justaugustus:

justaugustus:

/close

PR #2740, opened 2018-10-02
Create Cloud Provider Extraction Subproject

andrewsykim:
Our work to remove cloud providers out of kubernetes/kubernetes is turning out to be a bigger challenge then we had originally thought and figured it's time to formally create a working group to focus in on this effort. 

cc @d-nishi @cheftako @hogepodge @jagosan


sigs.yaml
```diff
@@ -2073,3 +2073,25 @@ workinggroups:
...
+      - name: Nishi Davidson
+        github: d-nishi
+        company: AWS
+      - name: Walter Fender
+        github: cheftako
+        company: Google
+    meetings:
+    - description: Regular WG Meeting
+      day: Thursday
+      time: "1:30"
```
andrewsykim: Time is open for change

cblecker:
@andrewsykim Was there some sort of community consensus on this?

I wonder if there is some overlap with sig-cloud-provider, or if this might better fit as a subproject.

cc: @kubernetes/steering-committee

andrewsykim:
@cblecker there was consensus for this within SIG Cloud Provider. The only reason for not putting this effort under a subproject is because subprojects (to me at least) are meant for long term projects where as working groups are for shorter term initiatives. The provider extraction we think will be (relatively) short term.

justaugustus:
@cblecker -- I'd say as it's cross-SIG and hopefully ephemeral, it's well-suited as a working group.

andrewsykim:
Also important to note that we are working on having provider subprojects under SIG Cloud Provider going forward. Once the provider code is removed the WG would be deleted and extracted code would fall under the respective provider subprojects.

cblecker:
As working group formation is still an open topic (https://github.com/kubernetes/community/pull/2702), steering should review this.

@andrewsykim Can you send a note to steering@k8s.io and ask them how they would like to proceed?

justaugustus:
Thanks for linking that, Christoph!

spiffxp:
>Once the provider code is removed the WG would be deleted and extracted code would fall under the respective provider subprojects.

Why you couldn't go ahead and do this now?  Make the subprojects under sig-cloud-provider, and have them cover both in-tree and out-of-tree code

sigs.yaml
frapposelli: 👍 from SIG-VMware

andrewsykim:
/hold 

currently trying to make this into a subproject instead, just waiting on https://github.com/kubernetes/community/pull/2731 to go in first

sig-cloud-provider/README.md
```diff
@@ -48,6 +48,10 @@ The following subprojects are owned by sig-cloud-provider:
 - **cloud-provider-vsphere**
   - Owners:
     - https://raw.githubusercontent.com/kubernetes/cloud-provider-vsphere/master/OWNERS
+- **cloud-provider-extraction**
+  - Owners:
```
jagosan: Are @cheftako and @d-nishi proposed as OWNERS? Reasonable to add this now?

spiffxp: Yeah this needs at least one OWNERS file linked

andrewsykim: So this is part of the concern I had raised in https://github.com/kubernetes/community/pull/2731, we don't currently have an OWNERS file because this effort is about migrating/removing code so there is no offical code ownership even though we need to meet weekly.

andrewsykim:
@spiffxp can I move forward with this now without any OWNERS file?

cheftako:
Can we move the meeting from 1 to 1:30?

andrewsykim:
meeting time updated to 1:30

andrewsykim:
@spiffxp added a `sig-cloud-provider/cloud-provider-extraction/OWNERS` file as you suggested, PTAL :)

spiffxp:
/approve
Thanks @andrewsykim that works for me
/lgtm
Based on all of the LGTM's above

andrewsykim:
Thanks @spiffxp!

/hold cancel

PR #2741, opened 2018-10-02
Move kubernetes-incubator/kube-arbitrator to k-sigs/kube-batch

spiffxp:
ref: https://github.com/kubernetes/org/issues/112


spiffxp:
/approve
self-approving

PR #2742, opened 2018-10-03
KEP: proposal for making kube-proxy pluggable

jainsm:

### DIFF
--- a/keps/sig-network/0016-18102018-kubeproxy-plugin.md
+++ b/keps/sig-network/0016-18102018-kubeproxy-plugin.md
@@ -1,0 +1,120 @@
+Enabling Plugins for Kube-Proxy to support different backends
+=============================================================
+
+Table of Contents<br>
+---------------------
+
+-   [Table of Contents](#table-of-contents)
+
+-   [Summary](#summary)
+
+-   [Motivation](#motivation)
+
+    -   [Goals](#goals)
+
+    -   [Non-Goals](#non-goals)
+
+-   [Proposal](#proposal)
+
+    -   [User Stories [optional]](#user-stories-optional)
+
+    -   [Story 1](#risks-and-mitigations)
+
+    -   [Story 2](#story-2)
+
+    -   [Implementation Details/Notes/Constraints
+        [optional]](#implementation-detailsnotesconstraints-optional)
+
+    -   [Risks and Mitigations](#risks-and-mitigations)
+
+-   [Graduation Criteria](#graduation-criteria)
+
+-   [Implementation History](#implementation-history)
+
+-   [Drawbacks [optional]](#drawbacks-optional)
+
+-   [Alternatives [optional]](#alternatives-optional)
+
+Summary
+-------
+
+This proposal aims to make the kube-proxy more pluggable to support different
+proxying mechanisms to be plugged in. This would encourage out of tree
+development and enhancements for the kube-proxy component.
+
+Motivation
+----------
+
+Kube-proxy component is one of the critical component in the data path of the
+service endpoints and abstracts the service endpoint from the consumer. Over a
+period this component has evolved from being a user-space based component to
+iptables and then to LVS based mechanism. This still remains an opinionated
+component and doesn’t provide flexibility for out of tree development of
+kube-proxy.
+
+Since kube-proxy is majorly providing some network functions, the proposal is to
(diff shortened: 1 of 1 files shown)


jainsm:
CLA check done . Still not reflecting

quinton-hoole-2:
The general idea seems very sensible.

thockin:
TL;DR : What you want already exists - it is called "don't run kube-proxy".

Strictly speaking, `Service` is a non-core API in Kubernetes.  Nothing (or almost nothing) in the core depends on Services.  They are a pluggable facet of the kube API (probably the first, in fact!).  It is entirely possible to NOT run kube-proxy and instead run something else (or not!).  Several implementors have done this.  I posit that this is what you're asking for, more or less.

To take a stronger stance - it may have been a mistake to have different proxy modes in kube-proxy -- maybe those would have been better as distinct binaries.  It was almost certainly a mistake to make the Windows modes be the same binary -- they share almost no code with the Linux forms, but the flags and config are clumsily shared.

We already have a plugin API - `Services` and `Endpoints`.

Rather than make kube-proxy even more of a plugin, I'd rather strip it down, move it to a distinct repo or 2 or 5, and EOL the omnibus version.  Maybe make it more configurable statically.  Maybe make core parts of it more composable/reusable as libraries.

Given all that, and the huge backlog of KEPs, I feel like we should close this one -- I love the idea in theory, but I don't think this is the way to achieve the goal.  I would be happy to entertain proposals for how to modularize better.  If you want to mutate this KEP into that, please reopen!

quinton-hoole-2:
@thockin I agree.  In my initial scan-read I misunderstood this to be a proposal to incorporate additional hooks into the rules (ipvs, iptables, whatever) that kubeproxy manages. But it's not.  Hence my comment above does not apply.

jainsm:
I think the idea is little bit misunderstood.  The idea was what @quinton-hoole-2  is referring to. Allow different kind of hooks into the rules . I called it the kube-proxy plugin which can as an example allow people to easily hook an ebpf program which does the filtering of traffic. End goal is to support different implementations not as part of core but as a seperate component. Correct me if such functionality already exists?

jainsm:
@thockin 
The code (https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-proxy/app/server.go) is part of the kube-proxy server implementation. Currently if we see we have all implementations bundled into one.
So any one who intends to change/add to the kube proxy has to make changes to this code.
The proposal was to take the specific functionality (core of how traffic should be forwarded. Like applying iptables or ipvs or userspace) into seperate plugins which can then be invoked by this core kube-proxy component. 
This to my knowledge will allow different and may be more optimal data path technology adaptations.
Please let me know what you think about it ?

quinton-hoole-2:
I wonder whether having standard iptables subchains (or ipvs equivalents)
at the appropriate points in the packet processing pipeline might be
sufficient? That way something morally equivalent to a kubeproxy plugin
could actually be a completely independent process/add-on, that just
managed those subchains (or whatever they're called). Just a thought.

jainsm:
I think thats what I am also referring to. When you say packet processing pipeline, it should allow a pluggable hook (plugin) to be invoked as part of that packet processing. Now this can be an ebpf program someone configures or can use the iptables/ipvs hook.

thockin:
@jainsm 

> The proposal was to take the specific functionality (core of how traffic should be forwarded. Like applying iptables or ipvs or userspace) into seperate plugins which can then be invoked by this core kube-proxy component.

This is very different than:

> Allow different kind of hooks into the rules

I would be OK with the iptables implementation having more defined hook points (call to an empty chain, let users own that chain).  The IPVS implementation does not have the same structure, so the hooks are probably going to be either impossible or so very different as to be not the same idea.  Making this generally standardized across other implementations of Services is not something we can govern (e.g. Cilium does its own thing and is not under our jurisdiction).

If you want to pursue this, I would do so between the 2 main implementations (iptables, ipvs) and see what hooks you think you need (and why), where they need to go, what the semantics would be, etc. and see how consistent you can make them.

To the idea of "take the specific functionality ... into seperate plugins ... invoked by this core kube-proxy" - there's nothing left of kube-proxy once you remove the plugins.  So I am against that idea.

jainsm:
Thanks @thockin . Shall I then redraft the proposal and open the PR again?

PR #2372, opened 2018-07-12
Add wg-networkservicemesh

edwarnicke:
Signed-off-by: Ed Warnicke <hagbard@gmail.com>

PR #2744, opened 2018-10-03
Add SIG Network charter

caseydavenport:

### DIFF
--- a/sig-network/charter.md
+++ b/sig-network/charter.md
@@ -1,0 +1,74 @@
+# SIG Network Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+SIG Network is responsible for the components, interfaces, and APIs which expose networking capabilities to Kubernetes
+users and workloads. SIG Network also provides some reference implementations of these APIs, for
+example kube-proxy as a reference implementation of the Service API.
+
+### In scope
+
+The following topics fall under ownership of this SIG:
+
+- Networking control plane and data paths.
+- Network service abstractions.
+- Service discovery (DNS).
+- Service load balancing (L4, L7).
+- Network security and identity.
+- Cluster connectivity.
+- Cross-cutting concerns such as scalability.
+- Metrics and monitoring associated with networking components.
+
+#### Code, Binaries and Services
+
+- Services
+  - APIs for defining and grouping networking endpoints.
+  - APIs for defining L3/4 loadbalancing.
+  - Reference implementation (kube-proxy).
+- Ingress
+  - APIs for defining L7 loadbalancing.
+- APIs for defining network policy.
+- Cluster DNS.
+- Integration points with networking implementations (e.g. CNI integration).
+- Container runtime interface (CRI) (With [sig-node]).
+- Cloud provider network integrations (With [sig-cloud-provider]).
+
+#### Cross-cutting and Externally Facing Processes
+
+### Out of scope
+
+- The CNI specification itself, which is maintained outside the Kubernetes project
+- Particular implementations of the CNI specification
+- Particular implementations of the NetworkPolicy API
+- Particular implementations of the Ingress API
+
+## Roles and Organization Management
+
+This sig adheres to the Roles and Organization Management outlined in [sig-governance]
+and opts-in to updates and modifications to [sig-governance].
+
+### Additional responsibilities of Chairs
+
+- None
+
+### Additional responsibilities of Tech Leads
+
+- None
+
+### Deviations from [sig-governance]
+
+- None
+
+### Subproject Creation
+
+SIG Technical Leads
+
+[sig-cloud-provider]: https://github.com/kubernetes/community/tree/master/sig-cloud-provider
+[sig-node]: https://github.com/kubernetes/community/tree/master/sig-node
+
+[sig-governance]: https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md
+[sig-subprojects]: https://github.com/kubernetes/community/blob/master/sig-network/README.md#subprojects
+[Kubernetes Charter README]: https://github.com/kubernetes/community/blob/master/committee-steering/governance/README.md


sig-network/charter.md
```diff
@@ -0,0 +1,55 @@
...
+SIG Network is responsible for the components, interfaces, and APIs which provide networking capabilities to Kubernetes
+users and workloads. SIG Network's goal is to provide simple, flexible abstractions on top of raw networking details, and to
+ensure Kubernetes provides the necessary networking features to enable application developers and cluster operators can
+meet their operational requirements.
+
+### In scope
+
+#### Code, Binaries and Services
+
+- Integration points with networking implementations (e.g. CNI integration)
```
dcbw: Might add CRI here too, since we have things to say about the PodSandbox setup stuff that CRI conveys.

dcbw: Also, we do have some overlap with Cloud Providers and their network integration.

sig-network/charter.md
```diff
@@ -0,0 +1,55 @@
...
+- Services / kube-proxy
```
bowei: Maybe it's worth putting a sentence for each one that describes what these terms mean?

Services:
- APIs for defining and grouping networking endpoints
- APIs and implementation for load balancing (L3/4 (kube-proxy) + L7 (Ingress))

sig-network/charter.md
```diff
@@ -0,0 +1,55 @@
+# SIG Network Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
```
danwinship: This section isn't entirely clear about the distinction between what SIG Network itself provides (APIs, interfaces, kube-proxy, ...?), and what it enables other people to provide (CNI plugins, Ingress implementations, ...?)

sig-network/charter.md
```diff
@@ -0,0 +1,55 @@
...
+- Ingress
+- Network policy
+- DNS
+
+#### Cross-cutting and Externally Facing Processes
+
+### Out of scope
+
+- The CNI specification itself, which is maintained outside the Kubernetes project.
+- Particular implementations of the CNI specification.
```
danwinship: - Particular implementations of Ingress

It might also be worth adding "Particular implementations of NetworkPolicy" since it may not be immediately obvious to most readers that that falls under "particular implementations of the CNI spec"

sig-network/charter.md
caseydavenport: I adjusted the wording a bit, but I think the "In scope" and "Out of scope" sections below make this clear.

caseydavenport:
@dcbw @bowei @danwinship thanks for the review, I've pushed a commit with some adjustments.

danwinship:
lgtm

sig-network/charter.md
```diff
@@ -0,0 +1,64 @@
+# SIG Network Charter
+
+This charter adheres to the conventions described in the [Kubernetes Charter README] and uses
+the Roles and Organization Management outlined in [sig-governance].
+
+## Scope
+
+SIG Network is responsible for the components, interfaces, and APIs which expose networking capabilities to Kubernetes
```
bowei: should we give some examples here?

"This includes both the networking control and data paths. Key areas include network service abstractions, service discovery (DNS), service load balancing (L4, L7), network security and identity and cluster connectivity. SIG-NETWORK is also responsible for cross-cutting concerns such as scalability, metrics and monitoring associated with networking components.,"

caseydavenport: I think the more detailed key areas bit is covered by the "In scope" (and to a lesser extent, the "out of scope") subsections below. Probably no need to explicitly call out particulars like DNS / Services, etc both here and below?

Agreed it's worth calling out the broader areas of scalability, metrics, monitoring explicitly in this section.

bowei: I was looking at how the other sigs structured their charters.

IMHO this section gives the broad categories and examples (networking capabilities is quite general).
The in code binaries, services seems like reference to actual artifacts?

caseydavenport: Yep, looking at cluster-lifecycle, they've used this section for a very high level description, but also have a section that we're missing right now which I think covers the examples you're looking for. I'll try modeling it after that and see what we think :)

dims:
@caseydavenport can we trying to get this going soon please? (pinging on behalf of steering)

caseydavenport:
@bowei @dims I believe I've addressed all the latest feedback. 

Please take another look! :)

sig-network/charter.md
```diff
@@ -0,0 +1,74 @@
...
+### Out of scope
+
+- The CNI specification itself, which is maintained outside the Kubernetes project
+- Particular implementations of the CNI specification
+- Particular implementations of the NetworkPolicy API
+- Particular implementations of the Ingress API
+
+## Roles and Organization Management
+
+This sig follows adheres to the Roles and Organization Management outlined in [sig-governance]
```
dims: "follows adheres" ... drop "follows"?

sig-network/charter.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Subproject Creation
+
+SIG Technical Leads
+
+[sig-cloud-provider]: https://github.com/kubernetes/community/tree/master/sig-cloud-provider
+[sig-node]: https://github.com/kubernetes/community/tree/master/sig-node
+
+[sig-governance]: https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md
+[sig-subprojects]: https://github.com/kubernetes/community/blob/master/sig-YOURSIG/README.md#subprojects
```
dims: replace `YOURSIG` with `network` please

sig-network/charter.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Networking control plane and data paths.
+- Network service abstractions.
+- Service discovery (DNS).
+- Service load balancing (L4, L7).
+- Network security and identity.
+- Cluster connectivity.
+- Cross-cutting concerns such as scalability.
+- Metrics and monitoring associated with networking components.
+
```
dims: Do you see the comment from @bgrant0607 in network service mesh discussion? which of the items above would the work there fall under? https://github.com/kubernetes/community/pull/2372#issuecomment-404721484

caseydavenport: I believe it would fall under some combination of "network service abstractions" and "data paths".

As @thockin mentioned in that thread, there are some aspects of that discussion which span multiple SIGs as well, but I think those two cover the parts relevant to SIG network.

dims: Thanks @caseydavenport

caseydavenport:
Who can approve this. @thockin perhaps?

spiffxp:
@philips is listed as the other steering committee reviewer

PR #2752, opened 2018-10-04
Update steering-committee alias per 2018 election results

spiffxp:
ref: https://github.com/kubernetes/steering/pull/79


spiffxp:
/approve
looking for /lgtm

PR #2758, opened 2018-10-05
KEP: NodeLocal DNS Cache

prameshj:
This KEP aims to improve kube-dns performance by running a nodelocal dns cache. 

@thockin @bowei @johnbelamaric @sdodson

### DIFF
--- a/keps/sig-network/0030-nodelocal-dns-cache.md
+++ b/keps/sig-network/0030-nodelocal-dns-cache.md
@@ -1,0 +1,150 @@
+---
+kep-number: 30
+title: NodeLocal DNS Cache
+authors:
+  - "@prameshj"
+owning-sig: sig-network
+participating-sigs:
+  - sig-network
+reviewers:
+  - "@thockin"
+  - "@bowei"
+  - "@johnbelamaric"
+  - "@sdodson"
+approvers:
+  - "@thockin"
+  - "@bowei"
+editor: TBD
+creation-date: 2018-10-05
+last-updated: 2018-10-05
+status: provisional
+---
+
+# NodeLocal DNS Cache
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Rollout Plan](#rollout-plan)
+* [Implementation History](#implementation-history)
+* [Drawbacks [optional]](#drawbacks-optional)
+* [Alternatives [optional]](#alternatives-optional)
+
+[Tools for generating]: https://github.com/ekalinin/github-markdown-toc
+
+## Summary
+
+This proposal aims to improve DNS performance by running a dns caching agent on cluster nodes as a Daemonset. In today's architecture, pods in ClusterFirst DNS mode reach out to a kube-dns serviceIP for DNS queries. This is translated to a kube-dns endpoint via iptables rules added by kube-proxy. With this new architecture, pods will reach out to the dns caching agent running on the same node, thereby avoiding iptables DNAT rules and connection tracking. The local caching agent will query kube-dns for cache misses of cluster hostnames(cluster.local suffix by default).
+
+
+## Motivation
+
+* With the current DNS achitecture, it is possible that pods with the highest DNS QPS have to reach out to a different node, if there is no local kube-dns instance.  
+Having a local cache will help improve the latency in such scenarios. 
+
+* Skipping iptables DNAT and connection tracking will help reduce [conntrack races](https://github.com/kubernetes/kubernetes/issues/56903) and avoid UDP DNS entries filling up conntrack table.
+
+* Connections from local caching agent to kube-dns can be upgraded to TCP. TCP conntrack entries will be removed on connection close in contrast with UDP entries that have to timeout ([default](https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt) `nf_conntrack_udp_timeout` is 30 seconds)
+
+* Upgrading DNS queries from UDP to TCP would reduce tail latency attributed to dropped UDP packets and DNS timeouts usually up to 30s (3 retries + 10s timeout). Since the nodelocal cache listens for UDP DNS queries, applications don't need to be changed.
+
+* Metrics & visibility into dns requests at a node level.
+
+* Neg caching can be re-enabled, thereby reducing number of queries to kube-dns.
+
+* There are several open github issues proposing a local DNS Cache daemonset and scripts to run it:
+	* [https://github.com/kubernetes/kubernetes/issues/7470#issuecomment-248912603](https://github.com/kubernetes/kubernetes/issues/7470#issuecomment-248912603)
+
+	* [https://github.com/kubernetes/kubernetes/issues/32749](https://github.com/kubernetes/kubernetes/issues/32749)
+
+	* [https://github.com/kubernetes/kubernetes/issues/45363](https://github.com/kubernetes/kubernetes/issues/45363)
+
+
+This shows that there is interest in the wider Kubernetes community for a solution similar to the proposal here. 
+
+
+### Goals
+
+Being able to run a dns caching agent as a Daemonset and get pods to use the local instance. Having visibility into cache stats and other metrics.
+
+### Non-Goals
+
+* Providing a replacement for kube-dns/CoreDNS.
+* Changing the underlying protocol for DNS (e.g. to gRPC)
+
+## Proposal
+
+A nodeLocal dns cache runs on all cluster nodes. This is managed as an add-on, runs as a Daemonset. All pods using clusterDNS will now talk to the nodeLocal cache, which will query kube-dns in case of cache misses in cluster's configured DNS suffix and for all reverse lookups(in-addr.arpa and ip6.arpa). User-configured stubDomains will be passed on to this local agent.  
+The node's resolv.conf will be used by this local agent for all other cache misses. One benefit of doing the non-cluster lookups on the nodes from which they are happening, rather than the kube-dns instances, is better use of per-node DNS resources in cloud. For instance, in a 10-node cluster with 3 kube-dns instances, the 3 nodes running kube-dns will end up resolving all external hostnames and can exhaust QPS quota. Spreading the queries across the 10 nodes will help alleviate this.
+
+#### Daemonset, Service, Listen Interface for caching agent
+
+The caching agent daemonset runs in hostNetwork mode in kube-system namespace with a Priority Class of “system-node-critical”. It listens for dns requests on a dummy interface created on the host. A separate ip address is assigned to this dummy interface, so that requests to kube-dns or any other custom service are not incorrectly intercepted by the caching agent. This ip address is obtained by creating a nodelocaldns service, with no endpoints. Kubelet takes the service name as an argument `--localDNS=<namespace>/<svc name>`, looks up the ip address and populates pods' resolv.conf with this value instead of clusterDNS. Each cluster node will have a dummy interface with this service ip assigned to it. This IP will be handled specially because of the NOTRACK rules described in the section below.
+
+#### iptables NOTRACK
+
+NOTRACK rules are added for connections to and from the nodelocal dns service ip. Additional rules in FILTER table to whitelist these connections, since the INPUT and OUTPUT chains have a default DROP policy.
+
+A dnscache nanny, similar to the [dnsmasq nanny](https://github.com/kubernetes/dns/tree/master/pkg/dnsmasq) will create the dummy interface and iptables rules. The nanny gets the nodelocal dns service ip address by querying kube-dns for the service name. The Daemonset runs in privileged securityContext since the nanny container needs to create this dummy interface and add iptables rules.
+ The nanny will also periodically ensure that the iptables rules are present. The resource usage for periodic iptables check needs to be measured. We can reduce the frequency of the checks by having the nanny query the dns cache and checking/adding rules only if the query fails.
+
+
+[Proposal presentation](https://docs.google.com/presentation/d/1c43cZqbVhGAlw3dSNQIOGuvQmDfKaA2yiAPRoYpa6iY), also shared at the sig-networking meeting on 2018-10-04
+
+Slide 5 has a diagram showing how the new dns cache fits in.
+
+#### Choice of caching agent
+
+The current plan is to run [Unbound dns server](https://www.nlnetlabs.nl/projects/unbound/about/) by default, based on these benchmark[ tests](https://github.com/kubernetes/perf-tests/tree/master/dns).
+
+Will rerun benchmark tests with coreDNS after SO_REUSEPORT support and measure QPS.
+
+Tests were run on a 1.9.7 cluster with 2 nodes, using Unbound 1.7.3 and coreDNS 1.2. Unbound QPS was 3600(caching)/3400(no caching), coreDNS 950(caching)/500(no caching)
+
+
+Based on the prototype/test results, these are the recommended defaults: 
+CPU request: 50m
+Memory Limit : 25m  
+Resource usage and QPS to be recomputed after metrics collection has been added.
+
+#### Metrics
+
+Unbound metrics can be extracted from the program unbound-control. This can be run from the nanny container.
+Unbound-control-setup is required to setup keys/certs for SSL.
+
+
+### Risks and Mitigations
+
+* Having the pods query the nodelocal cache introduces a single point of failure.   
+This is mitigated by assigning a PriorityClass of "system-node-critical", so it is always running. The nanny process will periodically check if the dns agent is running.
+Populating both the nodelocal cache ip address and kube-dns ip address in resolv.conf is not a reliable option. Depending on underlying implementation, this can result in kube-dns being queried only if cache ip does not repond, or both queried simultaneously.
+
+
+## Graduation Criteria
+TODO
+
+## Rollout Plan
+This feature will be launched with Alpha support in the first release. Master versions v1.13 and above will deploy the new add-on. Node versions v1.13 and above will have kubelet code to modify pods' resolv.conf. Nodes running older versions will run the nodelocal daemonset, but it will not be used. The user can specify a custom dnsConfig to use this local cache dns server.
+
+## Implementation History
+
+* 2018-10-05 - Creation of the KEP
+
+## Drawbacks [optional]
+
+Additional resource consumption for the Daemonset might not be necessary for clusters with low DNS QPS needs. 
+
+
+## Alternatives [optional]
+
+The listen ip address for the dns cache could be a link-local ip address, if there is a reliable way to reserve a link-local subnet. This will still require iptables rules for connectivity, in order to skip conntrack.
+
+
+Instead of just a dns-cache, a full-fledged kube-dns instance can be run on all nodes. This will consume much more resources since each instance will also watch Services and Endpoints.


keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,141 @@
...
+Unbound-control-setup is required to setup keys/certs for SSL.
+
+
+### Risks and Mitigations
+
+* Having the pods query the nodelocal cache introduces a single point of failure.   
+This is mitigated by assigning a PriorityClass of "system-node-critical", so it is always running. The nanny process will periodically check if the dns agent is running.
+Another option is to populate both the nodelocal cache ip address and kube-dns ip address in resolv.conf. The default behavior will be to try the nameservers in the order they are listed. From the man page of [resolv.conf](https://linux.die.net/man/5/resolv.conf) - "If there are multiple servers, the resolver library queries them in the order listed".
+
+
```
bowei: Can we add a subsection on how this is to be rolled out?

Also, we would need to adjust the kube-dns scaling parameters I would imagine when this is deployed?

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,141 @@
...
+Unbound metrics can be extracted from the program unbound-control. This can be run from the nanny container.
+Unbound-control-setup is required to setup keys/certs for SSL.
+
+
+### Risks and Mitigations
+
+* Having the pods query the nodelocal cache introduces a single point of failure.   
+This is mitigated by assigning a PriorityClass of "system-node-critical", so it is always running. The nanny process will periodically check if the dns agent is running.
+Another option is to populate both the nodelocal cache ip address and kube-dns ip address in resolv.conf. The default behavior will be to try the nameservers in the order they are listed. From the man page of [resolv.conf](https://linux.die.net/man/5/resolv.conf) - "If there are multiple servers, the resolver library queries them in the order listed".
```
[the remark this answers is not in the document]
stp-ip: It should be either the nodelocal cache or the central coredns instance in my view. Mixing might lead to unintended behaviors. One mentioned by John. Others could be harder debugging due to different resolution paths and silent failure of nodelocal cache (fallback to central).

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,141 @@
...
+* Neg caching can be re-enabled, thereby reducing number of queries to kube-dns.
+
+* There are several open github issues proposing a local DNS Cache daemonset and scripts to run it:
+	* [https://github.com/kubernetes/kubernetes/issues/7470#issuecomment-248912603](https://github.com/kubernetes/kubernetes/issues/7470#issuecomment-248912603)
+
+	* [https://github.com/kubernetes/kubernetes/issues/32749](https://github.com/kubernetes/kubernetes/issues/32749)
+
+	* [https://github.com/kubernetes/kubernetes/issues/45363](https://github.com/kubernetes/kubernetes/issues/45363)
+
+
```
stp-ip: * Controlling the central kube-dns service and the node local cache enables iterating the protocol used. Assuming both kube-dns and the cache are running CoreDNS a move to GRPC and watch based DNS querying might be possible.

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,141 @@
...
+## Motivation
+
+* With the current DNS achitecture, it is possible that pods with the highest DNS QPS have to reach out to a different node, if there is no local kube-dns instance.  
+Having a local cache will help improve the latency in such scenarios. 
+
+* Skipping iptables DNAT and connection tracking will help reduce [conntrack races](https://github.com/kubernetes/kubernetes/issues/56903) and avoid UDP DNS entries filling up conntrack table.
+
+* Connections from local caching agent to kube-dns can be upgraded to TCP. TCP conntrack entries will be removed on connection close in contrast with UDP entries that have to timeout ([default](https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt) `nf_conntrack_udp_timeout` is 30 seconds)
+
```
stp-ip: * Upgrading all non node local DNS queries from UDP to TCP would reduce tail latency attributed to dropped UDP packets and DNS timeouts usually up to 30s (3 retries + 10s timeout). Due to node local support for UDP applications don't need to be changed.

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,141 @@
...
+Being able to run a dns caching agent as a Daemonset and get pods to use the local instance. Having visibility into cache stats and other metrics.
+
+### Non-Goals
+
+Providing a replacement for kube-dns.
+
+## Proposal
+
+A nodeLocal dns cache runs on all cluster nodes. This is managed as an add-on, runs as a Daemonset. All pods using clusterDNS will now talk to the nodeLocal cache, which will query kube-dns in case of cache misses in cluster.local domain. The node's resolv.conf will be used by this local agent for all other cache misses.
```
stp-ip: Any specific reason to use the node's resolv.conf for all misses instead of using kube-dns for all query misses?

keps/sig-network/0029-nodelocal-dns-cache.md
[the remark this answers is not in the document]
bowei: Yes this is a non-started given the inconsistent behavior of *libc

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,141 @@
...
+
+[Proposal presentation](https://docs.google.com/presentation/d/1c43cZqbVhGAlw3dSNQIOGuvQmDfKaA2yiAPRoYpa6iY), also shared at the sig-networking meeting on 2018-10-04
+
+Slide 5 has a diagram showing how the new dns cache fits in.
+
+#### Choice of caching agent
+
+The current plan is to run [Unbound dns server](https://www.nlnetlabs.nl/projects/unbound/about/) by default, based on these benchmark[ tests](https://github.com/kubernetes/perf-tests/tree/master/dns).
+
+Will rerun benchmark tests with coreDNS after SO_REUSEPORT support and measure QPS.
```
stp-ip: If CoreDNS benchmark results are comparable to unbound after this change, I think using CoreDNS gives us some added features such as GRPC and watch based queries and also doesn't introduce another DNS component into the stack as CoreDNS would most likely already be running within kube-dns.

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+Based on the prototype/test results, these are the recommended defaults: 
+CPU request: 50m
+Memory Limit : 25m  
+Resource usage and QPS to be recomputed after metrics collection has been added.
+
+#### Metrics
+
+Unbound metrics can be extracted from the program unbound-control. This can be run from the nanny container.
```
stp-ip: Using CoreDNS would make exposing prometheus metrics possible.

keps/sig-network/0029-nodelocal-dns-cache.md
bowei: Wouldn't that cause problems with this?

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-config

keps/sig-network/0029-nodelocal-dns-cache.md
bowei: unbound is currently ~4x more efficient in CPU usage, which looks to the constrained resource.

keps/sig-network/0029-nodelocal-dns-cache.md
johnbelamaric: Yes, there still needs to be some examination of the memory impact with unbound. Since it needs a separate nanny process it may end up needing more memory. That's being evaluated, along with the re-running the QPS tests for CoreDNS. Obviously I would also prefer to see this be CoreDNS!

@prameshj, please be sure to use a variety of queries when doing the caching tests, as using the same query over and over for CoreDNS won't really get the fully benefit of the cache.

johnbelamaric:
@fturib @chrisohaver fyi

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Added.

Yes, we would need to adjust autoscaling parameters and also maybe resources to kube-dns and dnsmasq. I was hoping to get some more data once the feature is in alpha and then determine changes to the parameters.

keps/sig-network/0029-nodelocal-dns-cache.md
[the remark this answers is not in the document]
prameshj: Thanks for pointing it out. I hadn't realised musl would query all. I agree, mixing the 2 will be confusing and hard to debug.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: The reason to use node's resolv.conf is to make use of any additional caching that the node's upstream servers might have. Since kubeDNS pods use node's resolv.conf for looking up hostnames outside cluster.local, querying those servers directly reduces queries to kube-dns, so fewer DNAT. Unless nodes running kube dns have different /etc/resolv.conf contents than other nodes.

This shouldn't affect the custom dnsConfig since only pods in clusterFirst dnsPolicy mode will get nodelocal cache ip as their dns server. custom dnsConfig pods have a dnsPolicy of "None"

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Will do, @johnbelamaric

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+
+Being able to run a dns caching agent as a Daemonset and get pods to use the local instance. Having visibility into cache stats and other metrics.
+
+### Non-Goals
+
+Providing a replacement for kube-dns.
+
+## Proposal
+
+A nodeLocal dns cache runs on all cluster nodes. This is managed as an add-on, runs as a Daemonset. All pods using clusterDNS will now talk to the nodeLocal cache, which will query kube-dns in case of cache misses in cluster.local domain. The node's resolv.conf will be used by this local agent for all other cache misses.
```
sdodson: Don't we also need to account for reverse lookups?

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+* [Rollout Plan](#rollout-plan)
+* [Implementation History](#implementation-history)
+* [Drawbacks [optional]](#drawbacks-optional)
+* [Alternatives [optional]](#alternatives-optional)
+
+[Tools for generating]: https://github.com/ekalinin/github-markdown-toc
+
+## Summary
+
+This proposal aims to improve DNS performance by running a dns caching agent on cluster nodes as a Daemonset. In today's architecture, pods in ClusterFirst DNS mode reach out to a kube-dns serviceIP for DNS queries. This is translated to a kube-dns endpoint via iptables rules added by kube-proxy. With this new architecture, pods will reach out to the dns caching agent running on the same node, thereby avoiding iptables DNAT rules and connection tracking. The local caching agent will query kube-dns for cache misses of cluster.local hostnames.
```
thockin: Let's be really clear:  `cluster.local` is the default suffix, but it is actually configurable.

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: Precisely.  It may be possible to reuse TCP connections, reducing latency and conntrack footprint even further.

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+Having a local cache will help improve the latency in such scenarios. 
+
+* Skipping iptables DNAT and connection tracking will help reduce [conntrack races](https://github.com/kubernetes/kubernetes/issues/56903) and avoid UDP DNS entries filling up conntrack table.
+
+* Connections from local caching agent to kube-dns can be upgraded to TCP. TCP conntrack entries will be removed on connection close in contrast with UDP entries that have to timeout ([default](https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt) `nf_conntrack_udp_timeout` is 30 seconds)
+
+* Metrics & visibility into dns requests at a node level.
+
+* Neg caching can be re-enabled, thereby reducing number of queries to kube-dns.
+
```
thockin: Another motivation is to do the non-cluster lookups on the nodes from which they are happening, rather than the kube-dns instances.  We have evidence of this exhausting per-node DNS resources in clouds.  (e.g. 10 nodes all end up resolving all external names via 3 kube-dns instances).

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: I wanted to avoid this topic here - it is certainly possible, but the benefits are unknown, and it's distracting.  Let's fix the worst the problems we KNOW we have first. :)

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+This shows that there is interest in the wider Kubernetes community for a solution similar to the proposal here. 
+
+
+### Goals
+
+Being able to run a dns caching agent as a Daemonset and get pods to use the local instance. Having visibility into cache stats and other metrics.
+
+### Non-Goals
+
+Providing a replacement for kube-dns.
```
thockin: I would say kube-dns or CoreDNS.

I would also add a non-goal of changing the app protocol (e.g. to gRPC)

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: Change "cluster.local" to "the cluster's configured DNS suffix" or something like that.

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: Does this mean we need to push all the stub-domain configs down to the per-node caches?

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: Can you explain the question a bit more?

keps/sig-network/0029-nodelocal-dns-cache.md
stp-ip: Protocol changes are a non-goal in this KEP, but should be taken into consideration as a pro for using CoreDNS over Unbound here at least imo.

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+
+#### Daemonset, Service, Listen Interface for caching agent
+
+The caching agent daemonset runs in hostNetwork mode in kube-system namespace with a Priority Class of “system-node-critical”. It listens for dns requests on a dummy interface created on the host. A separate ip address is assigned to this dummy interface, so that requests to kube-dns or any other custom service are not incorrectly intercepted by the caching agent. This ip address is obtained by creating a nodelocaldns service, with no endpoints. Kubelet takes the service name as an argument `--localDNS=<namespace>/<svc name>`, looks up the ip address and populates pods' resolv.conf with this value instead of clusterDNS. Each cluster node will have a dummy interface with this service ip assigned to it. The Daemonset runs in privileged securityContext since it needs to create this dummy interface and add iptables rules.
```
thockin: As a transition, can you add a one-sentence paragraph to the end of this section, to the effect of: "This IP will be handled specially because of the NOTRACK rules described below"

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+
+#### iptables NOTRACK
```
thockin: we need to think about how this would or would-not work with ipvs kube-proxy, and document that VERY CLEARLY.  If this only works with iptables mode, we probably want to think harder about it.

We should maybe get some guidance from non-core implementations of services - could they work in a mode like this?

I don't think we have to make this a universal solution - if it only works with the upstream kube-proxy, that is OK.  Probably.  But we need to be confident how widely this technique applies vs just the idea being translatable.

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+
+NOTRACK rules are added for connections to and from the nodelocal dns service ip. Additional rules in FILTER table to whitelist these connections, since the INPUT and OUTPUT chains have a default DROP policy.
+
+A dnscache nanny, similar to the [dnsmasq nanny](https://github.com/kubernetes/dns/tree/master/pkg/dnsmasq) will create the dummy interface and iptables rules. The nanny gets the nodelocal dns service ip address by querying kube-dns for the service name. The nanny will also periodically ensure that the iptables rules are present. The resource usage for periodic iptables check needs to be measured. We can reduce the frequency of the checks by having the nanny query the dns cache and checking/adding rules only if the query fails.
```
thockin: As discussed, the resource footprint for iptables calls can be significant.  I call this out as a major risk to being able to deploy this idea.

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: memory footprint also matters.  Presumably the cache will be O(MB) (like 2-4 MB) so we need to look at TCO for CoreDNS vs Unbound+Nanny

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+
+The current plan is to run [Unbound dns server](https://www.nlnetlabs.nl/projects/unbound/about/) by default, based on these benchmark[ tests](https://github.com/kubernetes/perf-tests/tree/master/dns).
+
+Will rerun benchmark tests with coreDNS after SO_REUSEPORT support and measure QPS.
+
+Tests were run on a 1.9.7 cluster with 2 nodes, using Unbound 1.7.3 and coreDNS 1.2. Unbound QPS was 3600(caching)/3400(no caching), coreDNS 950(caching)/500(no caching)
+
+
+Based on the prototype/test results, these are the recommended defaults: 
+CPU request: 50m
```
thockin: that's a lot more than I expected.  We'll need to see if we can rely on burstable CPU for this and once we have proper autoscaling, use it.  

@justinsb has been looking at cluster-proportional scaling again, but this would almost certainly need to be scaled based on the size of the node it lands on, which is more complicated.

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+#### Metrics
+
+Unbound metrics can be extracted from the program unbound-control. This can be run from the nanny container.
+Unbound-control-setup is required to setup keys/certs for SSL.
+
+
+### Risks and Mitigations
+
+* Having the pods query the nodelocal cache introduces a single point of failure.   
+This is mitigated by assigning a PriorityClass of "system-node-critical", so it is always running. The nanny process will periodically check if the dns agent is running.
```
thockin: We should talk about how upgrades of the cache will work and what the implications are for outages.  E.g. during a rolling update of the cache queries can be lost and since they are UDP, will be subject to long timeouts.

How will we mitigate this?  

* Consider changing the default timeout paramt from 5s to 0.5s ? Or less.
* Can we do some other tricks to do hitless upgrades?
* Do we strongly recommend draining a node before upgrading this daemonset?

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+Populating both the nodelocal cache ip address and kube-dns ip address in resolv.conf is not a reliable option. Depending on underlying implementation, this can result in kube-dns being queried only if cache ip does not repond, or both queried simultaneously.
+
+
+## Graduation Criteria
+TODO
+
+## Rollout Plan
+This feature will be launched with Alpha support in the first release. Master versions v1.13 and above will deploy the new add-on. Node versions v1.13 and above will have kubelet code to modify pods' resolv.conf. Nodes running older versions will run the nodelocal daemonset, but it will not be used. The user can specify a custom dnsConfig to use this local cache dns server.
```
thockin: This seems to be more a deployment-tool choice, doesn't it?  E.g. are we saying that Kubeadm will support it?

keps/sig-network/0029-nodelocal-dns-cache.md
sdodson: I thought there were tests that verified that service ips and cluster ips had PTR entries. If that's the case doesn't the local caching agent need to forward in-addr.arpa to kube-dns at least for those zones? Maybe this is an OpenShift specific thing?

```
# host 172.30.0.1
1.0.30.172.in-addr.arpa domain name pointer kubernetes.default.svc.cluster.local.
```

johnbelamaric: Yes, that would be needed. So:

cluster.local
in-addr.arpa
ip6.arpa

Plus the stub-domains need to be on each local cache.

johnbelamaric: (and as tim says below, cluster.local is configurable as well)

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Added this to the list of benefits, thanks for pointing it out.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Yes, stubDomains and custom upstream servers will need to be passed down to the per-node caches.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Yes, I am working on measuring memory footprint and qps, with nanny doing periodic iptables check as well.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Thanks for pointing it out, we do need the reverse lookups.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Done

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Done

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Done

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: I haven't tried this with ipvs mode, but since the exposed service has no endpoints, there shouldn't be a difference. I will try it out.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: I agree. The tables we need to check are Raw(which has no rules added by kube-proxy) and Filter(much smaller than NAT table, kube proxy does not add rules for each service/endpoint)

keps/sig-network/0029-nodelocal-dns-cache.md
```diff
@@ -0,0 +1,145 @@
...
+* 2018-10-05 - Creation of the KEP
+
+## Drawbacks [optional]
+
+Additional resource consumption for the Daemonset might not be necessary for clusters with low DNS QPS needs. 
+
+
+## Alternatives [optional]
+
+The listen ip address for the dns cache could be a link-local ip address, if there is a reliable way to reserve a link-local subnet. This will still require iptables rules for connectivity, in order to skip conntrack.
```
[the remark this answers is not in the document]
prameshj: We would also need to whitelist those connections in filter table.  Not sure we want to add a generic rule to allow all connections to link local addresses to be ACCEPTed. If we did, would kube-proxy init be the place to add it?

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Likely not in the first release, but can be added later. I see Kubeadm supports kube-dns, coreDNS add-ons.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: That's a good point. 

I am worried setting the timeout too low might cause actual dns queries that are slow to never succeed, if there is a custom upstream server configured.
Can the recommendation be to use custom dns config and point at cluster dns ip for all clusterFIrst mode pods, before upgrading this daemonset?

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: That what does it mean to have a rollout plan?  If kubeadm doesn't support it, it's just a specification of how a user could set up their DNS, right?

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: That's a pretty heavy-handed approach - first mutate all of your daemonsets, then update DNS, then mutate them all back.  It would be simpler just to evict.

We could configure clients to have 2 nameservers, and fall back on the cluster-scope server, but clients are unpredictable and some are dumb.  Still might be OK - worth trying, perhaps?  Or understanding more.

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: @m1093782566

m1093782566: @prameshj You can ping me if you have any trouble in trying IPVS mode.

keps/sig-network/0029-nodelocal-dns-cache.md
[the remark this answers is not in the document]
thockin: hmm, you're right.  That's pretty ugly.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: In the first version of the KEP, I added the option of specifying 2 upstream servers in pods resolv.conf, first one the nodelocal ip followed by cluster dns ip. However, as other reviewers pointed out, only glibc implementation queries the next server if the first one does not respond. musl implementation queries them all in parallel.

One idea is to create the nodelocaldns service with the same selectors as kube-dns/coredns service. That way, it will have the same endpoints. Reaching out to nodelocal ip will hit the kube-dns/coredns pods, as long as the nanny container does not NOTRACK rules. Once nanny container is up and adds the NOTRACK rules and filter table rules, traffic will reach the local instance. If nanny process exits, dns will continue to work. Does this sound like a reasonable approach?

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Not sure I understand. Kubeadm is one of the ways to bring up a cluster, right? You mean for users of kubeadm, this feature will just be a spec without actual support to run it? I can add the manifests for kubeadm to run this addon, it just might not happen by 1.13 code freeze.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Thanks, I will reach out.

keps/sig-network/0029-nodelocal-dns-cache.md
prameshj: Need to verify if this will work in ipvs mode as well.

johnbelamaric: @thockin putting both in the resolv.conf will not work well for alpine images, it will end up hitting the cluster DNS and defeat the purpose. @prameshj that sounds like a viable option, good idea. It may be a little confusing to see the endpoints in the API server but have the traffic stolen by the new iptables rules, but it does solve the HA problem pretty cleanly.

keps/sig-network/0029-nodelocal-dns-cache.md
johnbelamaric: Tim's point is that deployment of each addon is specific to whatever is being used to create/manage the cluster. We have the same problem in making CoreDNS the "default". It requires adoption and updates to each of the cluster provisioning tools, not really to core Kubernetes itself. So, kubeadm, kops, kube-spray, etc. For CoreDNS, @rajansandeep updated several of these, I believe kube-up, kubeadm, and kops. The community followed with the rest.

quinton-hoole-2:
The general idea here seems sensible and useful.  Can we declare this to only be in the purview of sig-network and sig-node, or does it have significant impact beyond those sigs?

prameshj:
Thanks @quinton-hoole-2 , sig-network and sig-node are the sufficient, imo. I have modified the labels in this PR.

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: It doesn't solve HA entirely.  It solves it in the case of a clean upgrade, but not in the case of a crash.  I am not sure there's a great answer for that, really.  We could make the nanny add/remove NOTRACK when it loses connection to the DNS container, which might cover some of the issues.

Does openshift handle this case?

keps/sig-network/0029-nodelocal-dns-cache.md
thockin: Exactly.  It's worthwhile to think through how someone would do this rollout, but all we can produce is a doc.

thockin:
I propose we merge this KEP soon and get last details as followups.

prameshj:
Sounds good, I will follow up with changes once this is merged. Thanks for the feedback, everyone.

keps/sig-network/0029-nodelocal-dns-cache.md
krmayankk: Is there an update planned to handle HA and upgrade issue in this doc ?

PR #2772, opened 2018-10-09
Introduce definition of "DNS programming latency" SLI

wojtek-t:
/assing @bowei @caseydavenport @thockin @dcbw @countspongebob @shyamjvs


bowei:
cc: @lzang

thockin:
I think this is OK.  I guess these docs are not supposed to have details about HOW to get that info?

/lgtm

wojtek-t:
> I guess these docs are not supposed to have details about HOW to get that info?

You mean how to compute it? There is a section "how to measure the SLI" which is referencing to the network programming latency SLI, which is a sibling one. But yeah - the purpose was to briefly describe it, not really provide the details.

If you're asking about how to gather the metrics and present them to user/cluster admin/.,,, then yes - it's not in a scope of that doc.

@thockin - which one you meant?

PR #2774, opened 2018-10-09
[WIP] KEP: Vendor Cinder Volume Provider

hogepodge:


hogepodge:
/sig cloud-provider
/sig storage

cc: @dims @andrewsykim @jagosan @cheftako 

Related PRs
https://github.com/kubernetes/cloud-provider-openstack/pull/317
https://github.com/kubernetes/kubernetes/pull/69529

WIP, please comment

/hold

hogepodge:
I have this a bit backwards and need to do some updating.

hogepodge:
cc @cheftako

keps/sig-cloud-provider/0029-vendor-cinder-driver.md
```diff
@@ -0,0 +1,136 @@
...
+  - "@hogepodge"
+  - "@dims"
+owning-sig: sig-cloud-provider
+participating-sigs:
+  - sig-openstack
+  - sig-storage
+reviewers:
+  - "@hogepodge"
+  - "@dims"
+  - "@cheftaco"
```
dims: cheftako :)

keps/sig-cloud-provider/0029-vendor-cinder-driver.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Non-Goals
+
+* Removal of in-tree Cinder code is out of scope for this KEP.
+* Use of the CSI interface is out of scope for thie KEP.
+
+## Proposal
+
+An in-tree interface will be created for the Kubernetes Cinder code. This
+interface will call out to the external Cinder provider, allowing for the
```
dims: @hogepodge it's not the `external` provider, we are just moving the code to another repo and just vendoring that repo into main k/k.

the issue here is there is a recursion k/k repends on k/c-p-o and k/c-p-o depends on k/k.... this situation will get better over time as @andrewsykim and others are extracting the code into staging/ repos (into separate repos). eventually this recursion will be no more. but until then we have to be careful when changes are being made in both repos

cheftako:
@dims @hogepodge a few thoughts. 

While I seem some advantages for this approach in the short term, I have a question about the longer term approach. This has you making a library for your CP in its own repo and having K8s/K8s depend on your library/repo. Longer term I think you end up needing your own repo for building CCM with your cloud-provider code and then packaging it with the output of K8s/K8s. For that to work you will need to depend on K8s/K8s. That suggests either you are going to need to 2 repos (at least) or do some sort of instant switch over (which seems impractical). I think it would make more sense to have cloud-provider-openstack be the longer term repo out of which your distro builds come.

Also interested in the benefit of breaking this all the way out of staging to its own repo right now. Either way it will need to be vendored in to K8s/K8s in the short term. The biggest problem with that in mind is depending on/vendoring core K8s/K8s code. Moving it to its own repo will not break that dependency/vendoring. Either way will end up with something like https://github.com/kubernetes/kubernetes/blob/04b7baf43ad8ce68210e3a69bd395f467aee7275/vendor/k8s.io/cloud-provider-openstack/pkg/volume/cinder/BUILD#L16 where we have introduced vendor code depending on K8s/K8s. I believe release is in process of eradicating such dependencies.

justaugustus:

PR #2778, opened 2018-10-09
Update SIG Release leadership

justaugustus:
Chairs:
- Move Jaice Singer DuMars to Emeritus status
- Add Stephen Augustus
- Add Tim Pepper

/assign @calebamiles @jdumars @tpepper 
/sig release

Signed-off-by: Stephen Augustus <foo@agst.us>


calebamiles:
@justaugustus,

IMO this charter should live with the rest of the (non-generated) SIG Release information in the `github.com/kubernetes/sig-release` namespace like our other documentation. Generally, the charter needs to square the content that already exists in the [README][]. We're also proposing a change of leadership without documenting how these changes should be proposed and are approved

[README]: https://github.com/kubernetes/sig-release/blob/master/README.md

tpepper:
Elaborating a bit on Caleb's comment, this week's SIG Release meeting (see [Oct. 9, 2018 recording](https://www.youtube.com/watch?v=MIoXky4Mz3A)) started with discussion of this draft.

Regarding location of charter files:  other SIGs have theirs in k/community repo, with subdir and file of sig-${NAME}/charter.md.  It seems confusing for SIG Release to then put ours in k/sig-release.  Are there other sigs which place their charter in a non-k/community repo location?  My suggestion would be to reduce the content in k/sig-release/README.md and have it mostly link to the charter in k/community/sig-release/charter.md.

Leadership changes:  my reading of the current contents of the directory at https://github.com/kubernetes/community/tree/master/committee-steering/governance including the README.md, sig-governance.md, sig-governance-requirements.md, and the new shorter template (which Stephen has copied as is the guidance) is that the intent is for individual SIG charters to mostly say "we follow the normal documented path, plus the following deviations".  In particular https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance.md covers a basic path for continuity across members of defined roles in SIGs.  If nothing else is proposed for this SIG, the charter could explicitly state that.

That said, this PR introduces changes to SIG chair membership.  That should be a separate PR.

I'll note a few other things in the charter directly.

justaugustus:
(following the thread, btw. I'm just away at a company event at the moment, so I won't be able to address feedback until later.)

jdumars:
I am sad to be a deletion in the file, but also glad to see the next wave of project leadership emerge. 

/lgtm
/meow

justaugustus:
Announced on `k-release`, `k-release-leads`, and `steering`: https://groups.google.com/d/topic/kubernetes-sig-release/CIe4ekG7GkE/discussion
Maintaining lazy consensus hold until Friday, November 2nd, EOD PT. 
/hold

timothysc:
LGTM - But what does this mean you two plan to stay in this roll for some period of time across (N) releases, b/c historically release managers rotate.

sigs.yaml
```diff
@@ -1386,16 +1386,23 @@ sigs:
       - https://raw.githubusercontent.com/kubernetes/features/master/OWNERS
   - name: Release
     dir: sig-release
-    charter_link:
+    charter_link: https://git.k8s.io/sig-release/charter.md
```
cblecker: /cc @timothysc @spiffxp
I don't know if there are specific guidelines for charter location, but CC'ing steering to ensure this is cool with them.

timothysc: The charter typically lives in the community repo next to all the other sig details, not in an external repo.

jberkus:
@timothysc SIG-Release Lead != Version Release Lead.

That is, the leads for the SIG are different from the leads for each of the releases.  The terminology makes it confusing.  But the SIG leads are supposed to focus on long-term release needs, instead of the current release.

sigs.yaml
calebamiles: all of the important documentation for SIG release lives at `github.com/kubernetes/sig-release` we are unique in that regard but I would strongly like to continue in that direction. I would say that we shouldn't use a vanity URL in the link though.

justaugustus: Hmm... we provide the field to specify `charter_link`. I feel that implies that charters don't necessarily need to live in k/community as `charter.md`. If that's not the case, we should remove that field from the generator.

As Caleb mentioned, `k/sig-release` is the canonical source for all SIG Release information, so I think having the SIG's defining document adjacent to other SIG Release info is important. 
While I can get the need of not wanting break current convention, SIG Release is one of the few (maybe only) SIGs that has their own repo, and it's well-organized. Let me know what you think!

justaugustus:
Yep. What Josh said.
The official terminology would be `SIG Release Chair` (SIG wide) and `Kubernetes x.y Release Team Lead` (Release Team subproject-scoped).

sigs.yaml
justaugustus: Also, any reason to not use the vanity URL, @calebamiles?
It's the suggested method (for markdown, at least): https://github.com/kubernetes/community/blob/master/sig-contributor-experience/markdown-link-style-guide.md

timothysc: Charters should really be homed in a single location from a discoverability and policy enforcement perspective.  If there are bulk edits to how we do charters, the person responsible would have to track across (N).  I don't think sig-release is unique from a charter perspective.

justaugustus: I do think SIG Release is unique insofar as where the SIGs documentation is constantly put. That said, I don't have a strong opinion on where we land our governing document, outside of saying I hope it would stay close to the day-to-day docs we craft.

If Steering would prefer the charter remain in the community repo, as opposed to in k/sig-release, I'd want that to be explicitly said, as well as having `charter_link` be removed from the generator template. If we're saying that charters will always be within the SIG dir, named as `charter.md`, than the additional field has nonuse outside of allowing confusion.

cblecker: @justaugustus The field allowed for transition when some charter docs (before there was formal structure) were in docs and the like. I don't think we need to force a technical limitation on the generator here, when a policy one will do just fine. Setting that value to `charter.md` like most other sigs have works just fine.

justaugustus:
Removed the charter link. Needs a re-`/lgtm`.

cblecker:
/lgtm
/approve
/hold

Feel free to remove hold once lazy consensus period expires.

justaugustus:
Thanks @timothysc!

justaugustus:
Removing lazy consensus hold.
/hold cancel

justaugustus:
Thanks everyone who reviewed this!

